PR #744, opened 2016-05-01
blucat: set max to Yosemite

ilovezfs:


ilovezfs:
Author has already fixed it for us and tagged a new version. That has to be some kind of record.

UniqMartin:
> Author has already fixed it for us and tagged a new version. That has to be some kind of record.

Not only that, but the interaction in the related issue is just awesome. There are definitely great upstreams out there. :tada: Sorry for the spam …

PR #746, opened 2016-05-01
blucat 0.91

ilovezfs:


Formula/blucat.rb
```diff
@@ -17,6 +17,13 @@ def install
   end
 
   test do
-    system "#{bin}/blucat", "doctor"
+    begin
+      io = IO.popen("#{bin}/blucat scan 0")
```
tdsmith: we can pass an Array into `popen` to avoid invoking a shell right?

Formula/blucat.rb
```diff
@@ -1,12 +1,12 @@
...
-  url "http://blucat.sourceforge.net/blucat/wp-content/uploads/blucat-aa3e02.zip"
-  version "0.9"
-  sha256 "6dcd6bf538a06c2f29d21a9e94d859d91667a7014244462bffca9767bba5307d"
+  url "https://github.com/ieee8023/blucat/archive/v0.91.tar.gz"
+  sha256 "3e006d25b7e82689826c89ffbbfa818f8b78cced47e6d0647e901578d330a2f4"
+  head "https://github.com/ieee8023/blucat.git"
 
   depends_on "ant" => :build
-  depends_on :java => "1.6"
+  depends_on :java
```
tdsmith: I forget how this works; does `:java => "1.x"` set a minimum acceptable Java version? What does dropping the version argument do?

UniqMartin: IIRC, `:java => "1.x"` makes the formula depend on the specific major release, `:java => "1.x+"` on the specified major release and newer. Not sure what happens when the version is completely omitted …

Formula/blucat.rb
```diff
@@ -17,6 +17,13 @@ def install
   end
 
   test do
-    system "#{bin}/blucat", "doctor"
+    begin
+      io = IO.popen(bin/"blucat", "scan", "0")
```
tdsmith: I expected this to look like `[bin/"blucat", "scan", "0"]` but maybe i'm confused

Formula/blucat.rb
ilovezfs: It forces a specific version.

UniqMartin: … it uses the newest available Java installed on the system. The version requirement is transparently forwarded to `/usr/libexec/java_home --version` and `java_home` implements the selection logic. If `--version` is omitted, there's neither a lower nor upper limit. (My gut feeling is that we should always specify a minimum as documented by upstream.)

Formula/blucat.rb
ilovezfs: Maybe you're thinking of Open3.popen3?

ilovezfs: Probably thinking of this one?
```
    Utils.popen(["ronn", format_flag] + shared_args, "rb+") do |ronn|
      ronn.write markup
      ronn.close_write
      target.atomic_write ronn.read
    end

```

UniqMartin: @ilovezfs `Utils.popen` is rarely used directly. Most often we use `Utils.popen_read` and that one accepts a variable number of arguments. `IO.popen` just like `Utils.popen` expects the command and all its arguments to be either passed as a string (executed in a shell) or as an array of strings (executed without a shell).

Formula/blucat.rb
ilovezfs: iMac-TMP:~ joe$ for i in 1.4 1.5 1.6 1.7 1.8 1.9 ; do printf "$i: "; /usr/libexec/java_home --version $i &>/dev/null ; echo $? ; done
1.4: 2
1.5: 2
1.6: 0
1.7: 2
1.8: 0
1.9: 2
iMac-TMP:~ joe$

ilovezfs: Josephs-MacBook-Pro:Formula joe$ ag ':java'|grep -v '1\.'|wc -l
      64
Josephs-MacBook-Pro:Formula joe$ ag ':java'|grep '1\.'|wc -l
      92

I wonder to what extent these assignments or lack of assignments are random :)

ilovezfs: I guess I'll make it 1.6+

ilovezfs:
@tdsmith @UniqMartin Unfortunately the array version of IO.popen is not available with ruby 1.8.7, which is the cause of the test failure.

UniqMartin:
> Unfortunately the array version of IO.popen is not available with ruby 1.8.7, which is the cause of the test failure.

That's indeed unfortunate. I think this limitation of `IO.popen` could very well be the reason why `Utils.popen` and its sister functions exist.

ilovezfs:
@UniqMartin The main point of changing the test is that the old test fails on the test bot since Bluetooth isn't available, which may or may not be true in any given test environment. So the expectation here is that it will print `#Scanning RFCOMM Channels 1-30` and then proceed to fail. Should I just make it a non-test test that does `assert_match /blucat - by Joseph Paul Cohen 2012 - josephpcohen.com/, shell_output("#{bin}/blucat")` instead of the added complexity introduced by the test in the PR? Or is it worth it to at least have it attempt to do something functional other than printing a usage message?

UniqMartin:
@ilovezfs Let's use the test you just built! We can always easily go back to a simpler test once this one starts to act up, but currently it's in good shape and works both on CI and locally on my system. In my opinion, go ahead and merge this PR in its current form!

Also, whether or not we're passing the arguments directly or via some intermediate shell doesn't really matter here as the arguments are fixed strings, so there's not much that could go wrong.

tdsmith:
Oops; I was looking at the 1.9 API docs. Thanks for the education, all!

PR #752, opened 2016-05-01
no-more-secrets 0.1.0 (new formula)

apjanke:
### Description

Brian Barto's [no-more-secrets](https://github.com/bartobri/no-more-secrets) TTY toy. Recreates the SETEC ASTRONOMY effect from the movie "Sneakers".

687474703a2f2f692e696d6775722e636f6d2f657a4633786b4e2e676966

(This is basically a `curses` demo. Does not actually support universal decryption of all secret data, at least not in the 0.x release series.)

Hat tip to [this week's Changelog Weekly](http://email.changelog.com/t/t-F9D87CABB20E607A) mailing list for pointing me at this project.

### DIFF
--- a/Formula/no-more-secrets.rb
+++ b/Formula/no-more-secrets.rb
@@ -1,0 +1,16 @@
+class NoMoreSecrets < Formula
+  desc 'Recreates the SETEC ASTRONOMY effect from "Sneakers"'
+  homepage "https://github.com/bartobri/no-more-secrets"
+  url "https://github.com/bartobri/no-more-secrets/archive/v0.1.0.tar.gz"
+  sha256 "9e42f359bda578716d176245d416b5924ed0b321390a28221467301bc10e537b"
+
+  def install
+    system "make", "all"
+    bin.install "bin/nms", "bin/sneakers"
+  end
+
+  test do
+    # nms is an interactive-only program, so we can't do a real test
+    assert (bin/"nms").executable?
+  end
+end


apjanke:
Build failure is just because the repository is young. Which it is. But it's already got almost 1,000 stars. Any objections to pulling this in now?

```
failed: brew audit no-more-secrets --strict --online

Stacktrace
        Error: 1 problem in 1 formula
no-more-secrets:
```

Formula/no-more-secrets.rb
```diff
@@ -0,0 +1,16 @@
+class NoMoreSecrets < Formula
+  desc 'Recreates the SETEC ASTRONOMY effect from "Sneakers"'
```
UniqMartin: Not too happy with the single quotes we have banned almost everywhere else, but oh well …

DomT4: Should probably be reversed, so the single quotes are inside the `desc` element rather than wrapping it.

DomT4: Or just escape it and use double quotes in both places, but escaping is a bit messy in this situation.

UniqMartin: In my opinion, the inner quotes aren't even needed. After all it's *Sneakers* and not *sneakers*, so the quotes don't convey any additional information.

apjanke: But it is rendered in plain text in a terminal, so I can't italicize in this case. (Well, I *could* in some terminals, but we don't have markup for that, and I'm loath to put raw ANSI terminal-control codes in a description.) I think *some* typography is called for to indicate that it's a title, and not just a proper noun like a restaurant name. I chose double-quotes as a second-best substitute for italics; this seems to be a common convention when more advanced typesetting isn't available, at least in American English. And I've usually seen it in double quotes, not single quotes.

Happy to switch to escaped double quotes inside double quotes — it started out like this but I flipped it to single quotes on the outside just because it looked ugly and I thought folks would actually prefer the single-quote/double-quote inversion to backslashes. In the Ruby books I've read, that use is touted as one of the motivations for the interchangeability of double and single quotes in Ruby.

One more item for the Prose Style Guide, I guess. ;)

UniqMartin: I wasn't suggesting to use italics (and definitely am for sticking with plain text in the description), merely trying to say that the capitalized title sufficiently sticks out from the sentence case that is used in the rest of the description. Of course that's just my subjective feeling and as a non-native speaker it's difficult to judge how this is perceived by others.

DomT4: I think we should rewrite all formulae in Markdown to resolve this issue 😜. Would prefer the single quotes to be inside double quotes, but I'm not overly passionate about it.

apjanke: I switched it to single-quotes-in-double-quotes in 60eb33f7456dd7c605a8ca56929c98ddd05d8545, since that seems to be preferred and it still reads fine as both code and `brew info no-more-secrets` output.

PR #768, opened 2016-05-02
dwarf 0.3.1

ilovezfs:
- version bump
- new test: compile hello world and inspect Macho main header
- fix clang return-type errors
- fix parallelized build failure with missing target y.tab.h
- work around header issue with unknown types intmax_t and uintmax_t

### DIFF
--- a/Formula/dwarf.rb
+++ b/Formula/dwarf.rb
@@ -1,29 +1,51 @@
 class Dwarf < Formula
   desc "Object file manipulation tool"
   homepage "https://code.google.com/p/dwarf-ng/"
-  url "https://dwarf-ng.googlecode.com/files/dwarf-0.3.0.tar.gz"
-  sha256 "85062d0d3e8aa31374dd085cb79ce02c2b8737e9b143f640a262556233715763"
+  url "https://github.com/elboza/dwarf-ng/archive/dwarf-0.3.1.tar.gz"
+  sha256 "921667018e0edb057d695cdb6b7ed3bd8922a4050506252c21fffe4f7e77be2e"
 
+  depends_on "automake" => :build
+  depends_on "autoconf" => :build
   depends_on "readline"
 
-  # Imports inttypes.h in a generated lex file instead of stdint.h
-  # Reported upstream: https://code.autistici.org/trac/dwarf/ticket/8
-  # Also has missing & unwanted return values in some functions
-  # If the above are fixed, the newer 0.3.1 fails with a missing make target
-  fails_with :clang do
-    cause "error: unknown type name 'intmax_t'"
+  # Build fails with clang due to -Wreturn-type errors
+  # Reported 2nd May 2016: https://github.com/elboza/dwarf-ng/issues/2
+  # Opened PR with a fix 2nd May 2016: https://github.com/elboza/dwarf-ng/pull/4
+  patch do
+    url "https://github.com/elboza/dwarf-ng/pull/4.patch"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 you were the original reporter of these issues, so would be great if you could look this over when you have a chance. Thanks :)

DomT4:
Would be nice to get the patch submitted to and used from the [patches](https://github.com/Homebrew/patches) repo but otherwise LGTM.

I'd forgotten I'd reported problems upstream here, to be honest. It's not exactly filling me with hope those never got a response, but this formula remains useful so I guess let's give upstream some more time.

ilovezfs:
@DomT4 Yes, using the patches repo instead is a good idea. I was looking into opening a PR with upstream to apply the specific patch I have here in the __DATA__ section right now, but one of the files has changed since the tag. So it might be possible if I split it into two separate patches, one just for the file upstream changed, and the other for the other three that upstream did not change. The latter could be turned into a PR. Unfortunately, upstream appears to be somewhat of a :ghost: town, so I'm not sure it would even be noticed, but it wouldn't hurt to try.

ilovezfs:
@DomT4 OK, I opened a PR with upstream: https://github.com/elboza/dwarf-ng/pull/4

It's applied directly as a patch here, https://github.com/elboza/dwarf-ng/pull/4.patch. Would an additional instance of it in the patches repo be desirable and then use that, or is it preferable to leave it as it is right now?

DomT4:
Thanks, nice work here!

It's fine as-is, IMO. The goal of `patches` is twofold in 1) Having the patch somewhere that won't suddenly vanish one day, 2) Avoiding inline patches which are fond of breaking in odd ways. The second is fixed here and the first isn't particularly likely to happen in this case.

ilovezfs:
Thanks :)

Yup, that makes sense.

DomT4:
:shipit: when desired.

ilovezfs:
@DomT4 made one small change:
```
%w[src/cmdline src/libdwarf].each { |d| system "make", "-C", d, "y.tab.c" }
```
instead of the same thing twice :)

DomT4:
👍

Formula/dwarf.rb
```diff
@@ -1,29 +1,56 @@
...
   homepage "https://code.google.com/p/dwarf-ng/"
-  url "https://dwarf-ng.googlecode.com/files/dwarf-0.3.0.tar.gz"
-  sha256 "85062d0d3e8aa31374dd085cb79ce02c2b8737e9b143f640a262556233715763"
+  url "https://github.com/elboza/dwarf-ng/archive/dwarf-0.3.1.tar.gz"
+  sha256 "921667018e0edb057d695cdb6b7ed3bd8922a4050506252c21fffe4f7e77be2e"
 
+  bottle do
+    cellar :any
+    sha256 "cdecfa5a24f724649ec6880df2381eba77cd74d57329f6c637aadfda3318b13e" => :el_capitan
+  end
```
UniqMartin: I'm a bit confused where this `bottle do` block is coming from.

ilovezfs: I ran test-bot, and didn't notice that had snuck in. Thanks for catching it.

ilovezfs:
@UniqMartin fixed. Any tweaks you'd like to see?

UniqMartin:
> Any tweaks you'd like to see?

Nope, I'm fine with the formula as-is. But I better not look again at the upstream source. That's not what I would hope most C-based projects look like. (Or maybe my expectations are just off.) :wink:

ilovezfs:
@UniqMartin LOL! No comment.

PR #769, opened 2016-05-02
pelikan 0.1.1 (new formula)

thinkingfish:
### Description

Pelikan is a cache framework written in C. Currently it provides three
binaries that run as servers, including:
- pelikan_twemcache: a memcached/twemcache compatible server
- pelikan_slimcache: a memcached server using cuckoo hashing
- pelikan_pingserver: a ping server

### DIFF
--- a/Formula/pelikan.rb
+++ b/Formula/pelikan.rb
@@ -1,0 +1,21 @@
+class Pelikan < Formula
+  desc "production-ready cache services"
+  homepage "http://pelikan.io"
+  url "https://github.com/twitter/pelikan/archive/0.1.1.tar.gz"
+  sha256 "24a84bd9be7bb25da507a355b7ed3a35f7cb2a8c9447092f5acc91ff32f98775"
+  head "https://github.com/twitter/pelikan.git"
+
+  depends_on "cmake" => :build
+
+  def install
+    mkdir "_build" do
+      system "cmake", "..", *std_cmake_args
+      system "make"
+      system "make", "install"
+    end
+  end
+
+  test do
+    system "#{bin}/pelikan_twemcache", "-c"
+  end
+end


Formula/Pelikan.rb
```diff
@@ -0,0 +1,23 @@
...
+  url "https://github.com/twitter/pelikan/archive/0.1.1.tar.gz"
+  sha256 "98ab698b80e6e7634245e82798a9672b4a0b9f7f603879b398a66a4d97550e7f"
+  head "https://github.com/twitter/pelikan.git"
+
+  depends_on "cmake" => :build
+  depends_on "check" => :build
+
+  def install
+    mkdir "_build"
+    cd "_build" do
```
bfontaine: Use `mkdir "_build" do` instead of `mkdir` + `cd`.

Formula/Pelikan.rb
```diff
@@ -0,0 +1,23 @@
...
+      system "cmake", "..", *std_cmake_args # TODO: adjust args
```
bfontaine: `TODO`? 😉

Formula/Pelikan.rb
```diff
@@ -0,0 +1,23 @@
...
+      system "make", "-j"
```
bfontaine: Is there any reason to pass it `-j`?

thinkingfish: @bfontaine if compiling from source `make -j` is almost always faster than `make`, because the former uses multiple threads. I do not know if this is kosher with homebrew, but that's what I run when build locally.

Would follow whatever is recommended for homebrew.

Formula/Pelikan.rb
```diff
@@ -0,0 +1,22 @@
...
+  sha256 "98ab698b80e6e7634245e82798a9672b4a0b9f7f603879b398a66a4d97550e7f"
+  head "https://github.com/twitter/pelikan.git"
+
+  depends_on "cmake" => :build
+  depends_on "check" => :build
+
+  def install
+    mkdir "_build" do
+      system "cmake", "..", *std_cmake_args
+      system "make", "-j"
```
UniqMartin: Please remove the `-j` argument. Homebrew automatically passes it via `MAKEFLAGS`.

Formula/Pelikan.rb
```diff
@@ -0,0 +1,22 @@
+class Pelikan < Formula
+  desc "Pelikan: production-ready cache services"
```
UniqMartin: Please remove the `Pelikan: ` from the description. When appearing in a context where multiple formulae are listed, Homebrew will automatically prefix the formula name. In other contexts (like `brew info`) this prefix would look weird.

Formula/Pelikan.rb
```diff
@@ -0,0 +1,22 @@
...
+  homepage "http://pelikan.io"
+  url "https://github.com/twitter/pelikan/archive/0.1.1.tar.gz"
+  sha256 "98ab698b80e6e7634245e82798a9672b4a0b9f7f603879b398a66a4d97550e7f"
+  head "https://github.com/twitter/pelikan.git"
+
+  depends_on "cmake" => :build
+  depends_on "check" => :build
```
UniqMartin: Is this dependency really needed? The upstream documentation lists it as optional (only required for unit testing), but I don't see us running the test suite anywhere inside this formula.

thinkingfish:
@UniqMartin Thanks for the feedback. I have adopted all of your recommended changes.

mikemcquaid:
Looks like a `check` dependency may be missing: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/1281/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_Pelikan/

UniqMartin:
@thinkingfish Please have a look at <https://github.com/twitter/pelikan#build>. For the build to succeed without `check` you need to pass additional arguments to CMake.

@mikemcquaid No, it's not. See the above lines in this comment and now hidden code comments.

thinkingfish:
@UniqMartin I wrote the CMake file and the README.md. And the current error is actually a bug of the CMakelist.txt – pelikan should build w/o check, too, the correct behavior is skipping the test. I will correct this in the original repo(s) to avoid pulling in unnecessary dependencies.

UniqMartin:
@thinkingfish Sorry, I wasn't aware of that. We can also move forward with the `check` dependency added back. It's not exactly heavyweight and it will only affect people who build from source. But of course it's always better if these kinds of things can be resolved upstream. I suspect you'll be contributing formula updates in the future, so we can defer eliminating the `check` dependency till then.

thinkingfish:
@UniqMartin I've already fixed the bug upstream :smiley: 

I'll update the formula shortly to pull the next release, which will be void of this issue.

thinkingfish:
OK, upstream fixed, new tarball being pulled. I tested it locally and let's see if the CI god is happy this time.

UniqMartin:
> I tested it locally and let's see if the CI god is happy this time.

Thanks, that worked. Unfortunately you managed to brake things in a more subtle way because you capitalized the file name. Can you please squash your commits down to a single commit and while doing that, also rename the formula file from `Pelikan.rb` to `pelikan.rb`? OS X itself doesn't care about this, but Git and some of our infrastructure does.

thinkingfish:
@UniqMartin Just to be sure – are you asking me to basically overwrite branch history so all changes up till now (and the rename) is in one commit? Or do you prefer I start a new PR and leave this one alone (and potentially close it).

UniqMartin:
> Just to be sure – are you asking me to basically overwrite branch history so all changes up till now (and the rename) is in one commit?

Yes, please. Standard procedure for us is to squash all changes to a single commit per formula, unless the changes are orthogonal and worth preserving for the sake of history. Which cannot be the case here, as the change is basically “introduce new formula”.

thinkingfish:
@UniqMartin Interesting, thanks for the explanation. My previous experiences all had commit history be squashed when merging, so this is quite the opposite :)

UniqMartin:
The reason I'm explicitly asking for the squash here is because I envision some trouble if we try to handle a set of commits where the last one is only about changing the case of a file name. (As all our CI infrastructure runs OS X with the usual case-insensitive file system.)

> My previous experiences all had commit history be squashed when merging, […]

If contributors don't do it (not all are as confident using Git), we usually do it for them when pulling their changes. But we actually ask our contributors to squash in the “Following up” section of our [How To Open a Homebrew Pull Request (and get it merged)](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-%28and-get-it-merged%29.md#following-up) document:

> Squash new commits into one commit per formula with `git rebase --interactive origin/master`.

… but few people actually read that document or remember every detail they may have read a while ago.

thinkingfish:
@UniqMartin I'll make sure you don't have to paste it again for any of _my_ future PRs.

TBH I actually read that doc quickly before I started, but that line didn't stick. Maybe consider making it _**bold**_ (bold and italic at the same time)? :P

UniqMartin:
Everything worked out finally. Thanks for working with us. 😸

Merged in 518b0166be3067507f6e263f143b93d0522f9028. Thank you for this contribution to Homebrew, @thinkingfish! :tada:

> Maybe consider making it bold (bold and italic at the same time)? :P

No amount of formatting will ever be able to solve that. 😜

PR #783, opened 2016-05-02
schismtracker: url updates

vszakats:

### DIFF
--- a/Formula/schismtracker.rb
+++ b/Formula/schismtracker.rb
@@ -13,6 +13,7 @@ class Schismtracker < Formula
   end
 
   depends_on "autoconf" => :build
+  depends_on "automake" => :build
   depends_on "sdl"
 
   # CC BY-NC-ND licensed set of five mods by Keith Baylis/Vim! for testing purposes
--- a/Formula/schismtracker.rb
+++ b/Formula/schismtracker.rb
@@ -1,8 +1,9 @@
 class Schismtracker < Formula
   desc "Portable reimplementation of Impulse Tracker"
-  homepage "http://schismtracker.org"
-  url "http://schismtracker.org/dl/schismtracker-20120105.tar.bz2"
-  sha256 "31cef23e6806027618aef03675a5e4681043a99afa6e9e92d82ad4cec5e6c7db"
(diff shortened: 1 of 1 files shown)


Formula/schismtracker.rb
```diff
@@ -11,16 +12,11 @@ class Schismtracker < Formula
     sha256 "f892d60e3e74f352d7d9efbbf84b6470e4e751ffafe55c9b6ab9913bd611dba2" => :mavericks
   end
 
-  head do
-    url "http://schismtracker.org/hg/", :using => :hg
-
-    depends_on "autoconf" => :build
-  end
-
+  depends_on "autoconf" => :build
```
UniqMartin: Like I mentioned in the other PR, `autoconf` is actually only used by the HEAD build, so it should stay in the `head do` block, just like the HEAD `url` should.

vszakats: Isn't the new source archive created from a source tag?

I'll revert this, but the build failed last time with the new source archive.

vszakats:
Reverted. Let's see the CI results.

UniqMartin:
I was hoping you could try and build that locally to see what changes are needed in the `install` method to accommodate the changes caused by the relocated archive.

It seems my code comment didn't make sense and you actually instinctively/accidentally did the right thing. Sorry for making you revert that! There's no longer a `configure` in the archive, so we will have to run `autoreconf` unconditionally. I think replaceing (in `def install`):

```ruby
system "autoreconf", "-ivf" if build.head?
```

with

```ruby
system "autoreconf", "-ivf"
```

could do the trick.

Formula/schismtracker.rb
UniqMartin: Yes, you're right. (See my other non-code comment below.)

vszakats:
Oops, nice find, I missed that. Let's see how it works after that.

UniqMartin:
Build failed, looks like adding `depends_on "automake" => :build` should fix that.

UniqMartin:
Merged in 470e7aabade01e56cc0ae4a36e19e258c4b035ca. Thank you for yet another contribution to Homebrew, @vszakats! :tada:

PR #790, opened 2016-05-03
akka 2.4.4

yous:
### Description

Bump [akka](http://akka.io/) to 2.4.4.

### DIFF
--- a/Formula/akka.rb
+++ b/Formula/akka.rb
@@ -1,10 +1,12 @@
 class Akka < Formula
   desc "Toolkit for building concurrent, distributed, and fault tolerant apps"
   homepage "http://akka.io/"
-  url "https://downloads.typesafe.com/akka/akka_2.11-2.4.0.zip"
-  sha256 "22155144e5828a1dc40fa0a74031d5bf10292e8fb574d1c7fc5fc0ddebd03667"
+  url "https://downloads.typesafe.com/akka/akka_2.11-2.4.4.zip"
+  sha256 "f449ac4a9f401d2c25618c7d49a2e63f9c1c0e608f44854c06a604eec5f72d66"
 
   bottle :unneeded
+
+  depends_on :java
 
   def install
     # Remove Windows files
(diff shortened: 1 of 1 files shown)


UniqMartin:
The “builds” succeeds, so that's good. 😸

The formula unfortunately doesn't have a test yet, which makes it harder for our build infrastructure to check whether it's broken despite building fine and makes it more likely to break in the future. Would you be willing to try and add a test to the formula? [This section of the Formula Cookbook](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula) should get you started, and I'm happy to answer any related questions you might have. It's okay if you rather not, but it would be much appreciated if you could give this a try.

yous:
Thank you for pointing that out. I found a nice HelloWorld example for Akka, ["Akka Main in Java"](http://www.lightbend.com/activator/template/akka-sample-main-java). Although it can't be ran without Java, akka formula actually doesn't have `depends_on :java` line. The example run with Lightbend Activator, but we can run it Maven as the post describes at the last. I can confirm it runs with Maven, but we have another dependency now. This is the sample test script, but depends on Java and Maven. Do you have any idea?

``` diff
diff --git a/Formula/akka.rb b/Formula/akka.rb
index 90f7323..d730c9b 100644
--- a/Formula/akka.rb
+++ b/Formula/akka.rb
@@ -24,4 +24,98 @@ class Akka < Formula
     bin.install_symlink libexec/"bin/akka"
     bin.install_symlink libexec/"bin/akka-cluster"
   end
+
+  test do
+    (testpath/"pom.xml").write <<-EOS.undent
+      <?xml version="1.0" encoding="UTF-8"?>
+      <project xmlns="http://maven.apache.org/POM/4.0.0"
+               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+               xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+        <modelVersion>4.0.0</modelVersion>
+        <artifactId>akka-sample-main-java</artifactId>
+        <groupId>com.typesafe.akka.samples</groupId>
+        <name>Akka Main in Java</name>
+        <version>1.0</version>
+
+        <properties>
+          <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        </properties>
+
+        <dependencies>
+          <dependency>
+            <groupId>com.typesafe.akka</groupId>
+            <artifactId>akka-actor_2.11</artifactId>
+            <version>2.4.4</version>
+          </dependency>
+        </dependencies>
+
+      </project>
+    EOS
+    (testpath/"src/main/java/sample/hello/HelloWorld.java").write <<-EOS.undent
+      package sample.hello;
+
+      import akka.actor.Props;
+      import akka.actor.UntypedActor;
+      import akka.actor.ActorRef;
+
+      public class HelloWorld extends UntypedActor {
+
+        @Override
+        public void preStart() {
+          // create the greeter actor
+          final ActorRef greeter = getContext().actorOf(Props.create(Greeter.class), "greeter");
+          // tell it to perform the greeting
+          greeter.tell(Greeter.Msg.GREET, getSelf());
+        }
+
+        @Override
+        public void onReceive(Object msg) {
+          if (msg == Greeter.Msg.DONE) {
+            // when the greeter is done, stop this actor and with it the application
+            getContext().stop(getSelf());
+          } else
+            unhandled(msg);
+        }
+      }
+    EOS
+    (testpath/"src/main/java/sample/hello/Greeter.java").write <<-EOS.undent
+      package sample.hello;
+
+      import akka.actor.UntypedActor;
+
+      public class Greeter extends UntypedActor {
+
+        public static enum Msg {
+          GREET, DONE;
+        }
+
+        @Override
+        public void onReceive(Object msg) {
+          if (msg == Msg.GREET) {
+            System.out.println("Hello World!");
+            getSender().tell(Msg.DONE, getSelf());
+          } else
+            unhandled(msg);
+        }
+
+      }
+    EOS
+    (testpath/"src/main/java/sample/hello/Main.java").write <<-EOS.undent
+      package sample.hello;
+
+      public class Main {
+
+        public static void main(String[] args) {
+          akka.Main.main(new String[] { HelloWorld.class.getName() });
+        }
+      }
+    EOS
+    (testpath/"src/main/resources/application.conf").write <<-EOS.undent
+      akka {
+        loglevel = INFO
+      }
+    EOS
+    system "mvn compile exec:java -Dexec.mainClass='akka.Main' -Dexec.args='sample.hello.HelloWorld'"
+  end
 end
```

apjanke:
I like the test overall.

If `akka` can't run without Java, then it should have a `depends_on :java` in the formula. Let's add it. This was probably an oversight in the original formula, and the sort of thing adding tests is designed to catch. :)

And the Maven use here is trivial: it's functioning mainly to download Akka, which isn't needed in this case because we already know where it's downloaded and located. The Maven config file and `mvn` call could be replaced with a simple `javac` call (maybe that sets `-classpath` appropriately) and then a `java` call to actually run it.

yous:
@apjanke Thanks for explanation! I think I got this. The test works with my local machine, OS X 10.11. I introduced `depends_on :java` line as Akka 2.4.4 seems [work with Scala 2.11 and Java 8+](http://akka.io/downloads/). If it's wrong, please let me know.

apjanke:
LGTM at first glance. Let's see how BrewTestBot likes it...

apjanke:
TestBot likes it!

UniqMartin:
@yous Thanks for taking the time to contribute a good test for this formula! You rock!

PR #793, opened 2016-05-03
mplayer: add (back) freetype OSD support

jakeonfire:
### Description

The OSD is no longer enabled by default without including `freetype` and building with the `--enable-freetype` option (at least on OS X 10.11.4). Seems essential to me.

### DIFF
--- a/Formula/mplayer.rb
+++ b/Formula/mplayer.rb
@@ -20,6 +20,7 @@ class Mplayer < Formula
   end
 
   depends_on "yasm" => :build
+  depends_on "freetype" => :build
   depends_on "libcaca" => :optional
 
   unless MacOS.prefer_64_bit?
@@ -46,6 +47,7 @@ class Mplayer < Formula
       --disable-cdparanoia
       --prefix=#{prefix}
       --disable-x11
+      --enable-freetype
     ]
 
     args << "--enable-caca" if build.with? "libcaca"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Did you read the original reason this was removed? It requires X11 on OS X. Is this no longer the case? Thanks!

jakeonfire:
I have it working on my machine without X12, just `freetype`. I'm unaware of the history of the X12 requirement. was it perhaps for the `--enable-menu` option and not the `--enable-freetype` option??

Either way, I can't get the jenkins tests to pass. It installs fine and passes all tests locally...

Formula/mplayer.rb
```diff
@@ -20,6 +20,8 @@ class Mplayer < Formula
   end
 
   depends_on "yasm" => :build
+  depends_on "freetype" => :build
+  depends_on "libpng" => :build # required by freetype
```
DomT4: The CI is failing because `libpng` isn't a pure build-time dependency. I'd be surprised if `freetype` was as well, to be honest, but that's mildly more plausible.

DomT4:
Ping?

josheghani-dev:
what should i do with installation mongodb on my mac v10.12 ????? :((

DomT4:
@josheghani-dev Not sure this was the most relevant place for this comment, but only `brew install go --devel` works on macOS Sierra. If you have an existing `go` installation you should `brew rm go` & then reinstall it using `--devel` as directed. After that you'll be able to install `mongodb`.

MikeMcQuaid:
Closing due to red build and no response.

PR #796, opened 2016-05-03
BIND 9.10.4

aabdnn:
This pull request updates BIND to version 9.10.4

### DIFF
--- a/Formula/bind.rb
+++ b/Formula/bind.rb
@@ -1,10 +1,9 @@
 class Bind < Formula
   desc "Implementation of the DNS protocols"
   homepage "https://www.isc.org/downloads/bind/"
-  url "https://ftp.isc.org/isc/bind9/9.10.3-P4/bind-9.10.3-P4.tar.gz"
-  mirror "https://fossies.org/linux/misc/dns/bind9/9.10.3-P4/bind-9.10.3-P4.tar.gz"
-  version "9.10.3-P4"
-  sha256 "2ac044b5fbdf45fb45107af0df961b3b7cb5262a3bf1948ed3fe7a170dd13e3e"
+  url "https://ftp.isc.org/isc/bind9/9.10.4/bind-9.10.4.tar.gz"
+  mirror "http://fossies.org/linux/misc/dns/bind9/9.10.4/bind-9.10.4.tar.gz"
+  sha256 "f8d412b38d5ac390275b943bde69f4608f67862a45487ec854b30e4448fcb056"
   head "https://source.isc.org/git/bind9.git"
 
   bottle do
--- a/Formula/bind.rb
+++ b/Formula/bind.rb
@@ -2,7 +2,7 @@ class Bind < Formula
   desc "Implementation of the DNS protocols"
   homepage "https://www.isc.org/downloads/bind/"
   url "https://ftp.isc.org/isc/bind9/9.10.4/bind-9.10.4.tar.gz"
-  mirror "http://fossies.org/linux/misc/dns/bind9/9.10.4/bind-9.10.4.tar.gz"
+  mirror "https://fossies.org/linux/misc/dns/bind9/9.10.4/bind-9.10.4.tar.gz"
   sha256 "f8d412b38d5ac390275b943bde69f4608f67862a45487ec854b30e4448fcb056"
   head "https://source.isc.org/git/bind9.git"
 


apjanke:
Audit failure. Just needs a URL change.

```
Error Message

failed: brew audit bind

Stacktrace

        Error: 1 problem in 1 formula
bind:
  * Stable: Please use https:// for http://fossies.org/linux/misc/dns/bind9/9.10.4/bind-9.10.4.tar.gz

      
```

aabdnn:
I've updated the URL to use HTTPS

apjanke:
It's good now.

Habbie:
Please note that ISC has retracted this release. I suggest reverting this update.

apjanke:
Okay. I see the deprecation note at http://www.isc.org/downloads/. Sounds like we should roll back to 9.10.3-P4

screen shot 2016-05-12 at 9 21 38 am

Do you have a link to any announcement or discussion of the deprecation? I'm curious why it happened, and I couldn't find anything by Googling.

Habbie:
https://lists.isc.org/pipermail/bind-users/2016-May/096851.html

Habbie:
After discussion on IRC: rollback might be suboptimal as that won't cause 'brew upgrade' to 'go back'. I've asked ISC what their plan number-wise is so we can maybe fake an in-between version.

apjanke:
Oh gosh. That's right. You know, the lack of ability to "roll back" should probably be considered a bug in Homebrew, because of cases like this.

Per that announcement, this BIND bug is just a rare crash, and not a security vulnerability, so it's probably not critical to get all users to roll back automatically. At least not at the expense of adding a bogus version in the Homebrew history that could cause other problems later.

Best approach might be to roll back the Homebrew formula, and for Homebrew users who have picked up the 9.10.4 upgrade, rely on them noticing the crash and researching it or reporting it as a bug to Homebrew or ISC, in which case we or they can instruct them to just reinstall 9.10.3-P4.

Habbie:
@apjanke that also seems sensible

I wrote something here too: https://lists.isc.org/pipermail/bind-users/2016-May/096852.html

apjanke:
Okay. I'm going to roll the Homebrew formula back to 9.10.3-P4. That will prevent any new installs or upgrades of 10.9.4, and it won't rule out other future actions, since we have no mechanism for them in the first place.

Habbie:
I was just about to make a PR, but yes, please do.

apjanke:
This is better done without a PR: that will trigger a bottle build and attempted republish, which will bork because bottles for that exact version are already published. I'll do a direct commit to do the rollback without rebuilding, and so it retains the reference to the prior bottles.

ilovezfs:
Perhaps the bottles on bintray should be wished out to the corn fields as well.

apjanke:
Yeah, probably. That could catch some additional bogus installs. I'll do so.

DomT4:
> Oh gosh. That's right. You know, the lack of ability to "roll back" should probably be considered a bug in Homebrew, because of cases like this.

It already is really; ties into the discussions we had before about "borrowing" Debian's epoch system or something similar.

> I'll do a direct commit to do the rollback without rebuilding, and so it retains the reference to the prior bottles.

We tend to revert each necessary commit in turn for the sake of history (i.e. revert the bottle changes, then the Bind update), but the way you did it is reasonable enough.

apjanke:
Rolled back to 9.10.3-P4 in 33cc32379aab9d6f91b197da90585be4e2d81121.

Deleted the 9.10.4 bottle files from Bintray.

Bintray showed 399 downloads by the time I deleted them.

Let's send an announcement out to Homebrew-Discuss. (Or do we have a separate Homebrew-Announce mailing list yet?) A rollback to a significant piece of software that can't be automated probably justifies an announcement mailing. And maybe a separate bug posting so that people watching the Issues activity see it.

apjanke:
Posted an issue to serve as an announcement: #1099

Thoughts on sending out an email? Does this justify an announcement? I'm thinking yes.

ilovezfs:
I think a one-off version class like https://github.com/Homebrew/homebrew-x11/blob/c3c636cd1550c0b7c9c0cd8fbf6c6025051e7e82/scantailor.rb#L1-L19 might be able to force a downgrade until there's a DSL for it.

DomT4:
Well, _theoretically_ you can force a downgrade fairly easily by inserting a fake version greater than the previous version people had installed, but if the next release isn't greater than our fake version you have to retain & bump the fake version until it is, and at that point things get a bit dicey.

mikemcquaid:
> Deleted the 9.10.4 bottle files from Bintray.

Too late now but FWIW I think this is a bad idea; it's just going to mean people on the outdated version end up having to build from source.

ilovezfs:
If the next version is 9.10.4-P1 or 9.10.5, either of those will be greater that 9.10.4-P0.1

apjanke:
> Too late now but FWIW I think this is a bad idea...

Will keep that in mind for next time something like this happens. (Which is hopefully never. :wink:)

ilovezfs:
The solution to that is to have upstream kill the download link.

PR #799, opened 2016-05-03
ammonite-repl-scala-2.10 0.5.7 (new formula)

Atry:

---

This formula is like https://github.com/Homebrew/homebrew-core/blob/master/Formula/ammonite-repl.rb , except this formula is for Scala 2.10 instead of Scala 2.11

### DIFF
--- a/Formula/ammonite-repl-scala-2.10.rb
+++ b/Formula/ammonite-repl-scala-2.10.rb
@@ -1,0 +1,18 @@
+class AmmoniteReplScala210 < Formula
+  desc "cleanroom re-implementation of the Scala 2.10 REPL"
+  homepage "http://www.lihaoyi.com/Ammonite/#Ammonite-REPL"
+  url "https://github.com/lihaoyi/Ammonite/releases/download/0.5.7/ammonite-repl-0.5.7-2.10.5", :using => :nounzip
+  version "0.5.7"
+  sha256 "9f8e67466f58f6490c2d4b4d0f4ff9ad555ce29909843ba63f5c7533b993642b"
+
+  depends_on :java => "1.7+"
+
+  def install
+    bin.install "ammonite-repl-0.5.7-2.10.5" => "amm210"
+  end
+
+  test do
+    ENV.java_cache
+    assert_equal "hello world!", shell_output("#{bin}/amm210 -c 'print(\"hello world!\")'")
+  end
+end


apjanke:
The `scala` brew formula is already at Scala 2.11. Maybe we should just bump the version in the existing `ammonite-repl` instead? (And maybe make the command name simply `ammonite-repl`? We normally don't accept multiple versions of formulae in core Homebrew.

And should `ammonite-repl` depend on `scala`? It's a "cleanroom reimplementation of the REPL", but it looks like the implementation is at least partly in the Scala language, even if it's not using the Scala REPL.

Atry:
The `ammonite-repl` script embedded Scala compiler and standard library in itself, hence it does not depend on `scala` formula.

0.5.7 is the version of `ammonite-repl`.
2.11 or 2.10 is the embedded Scala version.

Atry:
I think we did accept multiple versions of formulae in core Homebrew. For example, [Python 2](https://github.com/Homebrew/homebrew-core/blob/master/Formula/python.rb) vs [Python 3](https://github.com/Homebrew/homebrew-core/blob/master/Formula/python3.rb).

apjanke:
> I think we did accept multiple versions of formulae in core Homebrew. For example, Python 2 vs Python 3.

Only for a few exceptions where there are multiple major versions in simultaneous use by various programs. E.g. Python 2 and 3, GTK+ 2 and 3, and that's about it. The general rule is "only one version of a given package in core".

Let's just bump the existing version of `ammonite-repl` in this case. If there is really a need for multiple versions of Scala and `ammonite` to coexist, that would probably be a case for an external Tap.

Atry:
The point is backward compatibility. Scala 2.10 and 2.11 are different major versions, as the version does not follow [semver](http://semver.org)



The upstream releases different binaries of the same Ammonite version for different Scala versions: https://github.com/lihaoyi/Ammonite/releases

apjanke:
Thoughts from other maintainers on this? Will we accept multiple major versions of Scala and Scala-related tools in core Homebrew?

I lean toward "no", because making those exceptions is a pretty big deal, and I see this ballooning out in to a lot of formula variations because Scala is currently young and fast-moving. Might be a case for a Scala-specific tap run by people who actually know Scala. I don't think we have any Scala hackers among the Homebrew maintainers.

Added "maintainer feedback" label to solicit input from other folks.

Atry:
I found that Debian also treat backward incompatible stuffs as different packages, for example:


- https://packages.debian.org/search?keywords=libboost1.60-dev
- https://packages.debian.org/search?keywords=libboost1.58-dev
- https://packages.debian.org/search?keywords=libboost1.55-dev
- https://packages.debian.org/search?keywords=libboost1.54-dev

This approach allows other packages of Debian depends on different versions of boost in one Debian release.

apjanke:
True, and we recognize the advantages of that approach. But compared to Debian, Homebrew is faster-moving, a smaller organization, and doesn't have the version-specific dependency support that Debian does. So we generally stick with one major version in core Homebrew, and push support for multiple major versions (like that Boost 1.54/1.55/1.58/1.60) out to independently-maintained taps.

We've attempted support for multiple versions with the [homebrew-versions tap](https://github.com/homebrew/homebrew-versions), and the result has not been great; we don't have the resources or tech to support it adequately.

I think this still points to just bumping the version of `ammonite-repl` in core Homebrew, and leaving multiple versions to an outside Tap.

Atry:
You are right, I found Scala 2.10 there: https://github.com/Homebrew/homebrew-versions/blob/master/scala210.rb

apjanke:
Yeah. If you'd like to have both of these available, submit this on homebrew/versions instead of here on core; we'll merge it there. If there's a Scala 2.10 version there, it's legit to have a corresponding ammonite-repl.

mikemcquaid:
Agreed on homebrew-versions.

PR #806, opened 2016-05-03
henplus: use :java requirement

ilovezfs:


DomT4:
This won't test on CI because one of the dependencies wants Java 1.6, somehow.

Personally given the need for an ancient Java I'd just been tempted to kill this and `libreadline-java` but if you want to keep it I don't mind.

ilovezfs:
@DomT4 I've fixed the libreadline-java issue and reported it upstream: https://github.com/Homebrew/homebrew-core/pull/837

In addition, henplus upstream has agreed to create a new tag: https://github.com/neurolabs/henplus/issues/41

So I don't think either need be removed.

DomT4:
Alright, as long as you're working on it. Anything using Java 1.6 these days frankly seems unsupported by upstream.

ilovezfs:
@DomT4 :ship:?

DomT4:
Seems reasonable 👍

Issue #681, opened 2016-04-28
ccze crashes a lot


onlynone:
I'm trying to add `hollywood` which uses `ccze`. But `ccze` crashes a lot which takes down all of `hollywood`. I get an OS X crash reporter window when `ccze` crashes. There seem to be two different cases that cause a crash. The first is when `ccze` is passed a color name with a space in it, like `ccze -c bin="bold green"`. I get `Segmentation fault: 11` printed to the terminal, and the following error report from OS X:

```
Process:               ccze [96719]
Path:                  /usr/local/Cellar/ccze/0.2.1/bin/ccze
Identifier:            ccze
Version:               0
Code Type:             X86-64 (Native)
Parent Process:        bash [22708]
Responsible:           Terminal [72798]
User ID:               501

Date/Time:             2016-04-28 10:44:16.185 -0400
OS Version:            Mac OS X 10.10.5 (14F1605)
Report Version:        11
Anonymous UUID:        C3D36F96-11A2-5880-7B10-5EABA98AEF85

Sleep/Wake UUID:       6A5FF312-5247-44BC-9B97-E2797C02B293

Time Awake Since Boot: 1200000 seconds
Time Since Wake:       4700 seconds

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000

VM Regions Near 0:
--> 
    __TEXT                 000000010105e000-0000000101069000 [   44K] r-x/rwx SM=COW  /usr/local/Cellar/ccze/0.2.1/bin/ccze

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_c.dylib             	0x00007fff98b96f92 strlen + 18
1   libsystem_c.dylib             	0x00007fff98bf0dd1 strdup + 18
2   ccze                          	0x00000001010601e4 parse_opt + 241
3   ccze                          	0x000000010106269c argp_parse + 554
4   ccze                          	0x000000010105fbb4 main + 100
5   libdyld.dylib                 	0x00007fff940805c9 start + 1

Thread 0 crashed with X86 Thread State (64-bit):
  rax: 0x00000000ffffffff  rbx: 0x0000000000000000  rcx: 0x0000000000000000  rdx: 0x0000000000000000
[...]

PR #810, opened 2016-05-03
ccze: Fix C pre-processor directive

onlynone:
See issue #681

### DIFF
--- a/Formula/ccze.rb
+++ b/Formula/ccze.rb
@@ -43,3 +43,16 @@ index c6f9892..9b93b65 100644
 		-Wmissing-noreturn -Wmissing-declarations @WFLAGS_3X@
  WFLAGS_ICC	= -Wall -wd193,279,810,869,1418,1419
  WFLAGS_3X	= -Wsequence-point -Wdiv-by-zero -W -Wunused \
+diff --git a/src/ccze-compat.c b/src/ccze-compat.c
+index 0a3c335..5afdc20 100644
+--- a/src/ccze-compat.c
++++ b/src/ccze-compat.c
+@@ -275,7 +275,7 @@ ccze_getsubopt (char **optionp, char *const *tokens,
+ 		char **valuep)
+ {
+   int i = getsubopt (optionp, tokens, valuep);
+-#if HAVE_SUBOPTARg
++#if HAVE_SUBOPTARG
+   if (!*valuep && suboptarg)
+     *valuep = strdup (suboptarg);
+ #endif


DomT4:
Where was this reported upstream? We probably want to bump the `revision` as well if we're fixing breakage.

onlynone:
It hadn't been previously (until I just did today).
On May 3, 2016 13:21, "Dominyk Tiller" <notifications@github.com> wrote:

Where was this reported upstream? We probably want to bump the revision as
well if we're fixing breakage.

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/Homebrew/homebrew-core/pull/810#issuecomment-216602128>

UniqMartin:
> It hadn't been previously (until I just did today).

I think we have a misunderstanding here. By “reported to upstream” we mean reported to the developers of `ccze`. That's not us (Homebrew), as we're just packaging other people's software, but not maintaining it ourselves. For Homebrew to accept a patch to a formula, it needs to be sent to the upstream developers and accepted by them. (If you did, please provide a URL to the mailing list message, issue tracker entry, or whatever channel of communication you used to submit your patch.)

onlynone:
I know what reported to upstream means.

1. I hadn't reported this issue anywhere except for here at first.
2. Then I found the issue with `HAVE_SUBOPTARg`.
3. Then I created the PR here.
4. Then I [created this bug report at Debian](http://bugs.debian.org/823334) after you said "Alright, so this seems to be an upstream issue. Thanks for reporting it to them!"

Now, a true upstream doesn't really seem to exist. The package seems to be abandoned and only maintained at various distros like debian. Since this formula points to debian for the source, I figured that's the closest to "upstream" this package has.

However, upon further investigation I've found that this bug was independently fixed in freebsd 13 years ago:

http://lists.freebsd.org/pipermail/freebsd-ports-bugs/2003-September/013364.html

And in gentoo 6 years ago:

https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/app-admin/ccze/files/ccze-fbsd.patch?hideattic=0&revision=1.1&view=markup

I'm thinking that it might make more sense to base this formula on either the freebsd port (since OS X has a bsd userland) or gentoo since they seem to have even more patches to ccze that address building and segfaulting:

https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/app-admin/ccze/files/

UniqMartin:
> I know what reported to upstream means.

Sorry if I implied otherwise and thanks for clarifying things! That's just what I gathered from the little information that I had available. Now I have a much clearer picture.

> Now, a true upstream doesn't really seem to exist.

That's somewhat unfortunate and doesn't really fit Homebrew's expectations for fomulae. Our [usual acceptance criteria](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Acceptable-Formulae.md) require upstream to exist and be active. Of course since the formula is already in Homebrew, we might judge differently than with new formulae being submitted.

As the formula basically never changed since it was added three years ago and seems to be both low maintenance and somewhat popular, we might as well accept this patch. I'm not sure about the other patches (the ones from Gentoo) and whether they apply here. Thanks for the through investigation!

Can you please extend the comment above the `patch :DATA` line with a brief summary of this and links to the respective FreeBSD/Gentoo resources, so future maintainers/contributors can easier understand the situation? Thanks!

@DomT4 Any major objections to pulling this as-is? An upstream is unlikely to appear out of nowhere, but I think it's worth applying the patch instead of boneyarding this formula.

DomT4:
Given it's only a single line and we may well be carrying this indefinitely we should be using an `inreplace` rather than a patch, i.e.:

```ruby
def install
  # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=823334
  inreplace "src/ccze-compat.c", "#if HAVE_SUBOPTARg", "#if HAVE_SUBOPTARG"
```

And we probably want to bump the revision since it fixes current breakage.

Otherwise, 👍.

UniqMartin:
> Given it's only a single line and we may well be carrying this indefinitely we should be using an `inreplace` rather than a patch, i.e.:

What's the advantage of using `inreplace` over extending the already existing patch? I can only see disadvantages: no context and a greater likelihood of matching stuff it's not supposed to match …

> And we probably want to bump the revision since it fixes current breakage.

Indeed, good point.

DomT4:
> What's the advantage of using inreplace over extending the already existing patch?

It's Homebrew "policy" that if we have to carry something indefinitely with potentially no chance of it ever being accepted upstream we heavily prefer `inreplace`s, especially for single-line changes.

I'm way too tired to get particularly passionate about if it you are, but let's at least ship the patch to the patches repo given the awful dance that is inline patching.

UniqMartin:
@onlynone Please follow @DomT4's advice on using `inreplace` instead of extending the inline patch and please bump the revision, i.e. add a `revision 1` line below the topmost `sha256` line in the formula, so existing installations will be updated and users get the fixed version. Once this is done we're good to go.

DomT4:
Applied additional changes in a follow-up commit.

Merged in https://github.com/Homebrew/homebrew-core/commit/0a02e10fefcfd1e79839fa9c0892ee5d27bfc0b1. Thank you for your contribution to Homebrew @onlynone; we appreciate it! :smiley_cat:

PR #813, opened 2016-05-03
opentsdb 2.2.0 (new formula)

CamJN:

### DIFF
--- a/Formula/opentsdb.rb
+++ b/Formula/opentsdb.rb
@@ -76,8 +76,8 @@ class Opentsdb < Formula
 
   def post_install
     (var/"cache/opentsdb").mkpath
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
     begin
-      system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
       sleep 2
       system "#{pkgshare}/tools/create_table_with_env.sh"
     ensure
@@ -124,28 +124,30 @@ class Opentsdb < Formula
     ENV["HBASE_LOG_DIR"]  = testpath/"logs"
     ENV["HBASE_CONF_DIR"] = testpath/"conf"
     ENV["HBASE_PID_DIR"]  = testpath/"pid"
+
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
     begin
-      system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
       sleep 2
 
       system "#{pkgshare}/tools/create_table_with_env.sh"
 
       tsdb_err = "#{testpath}/tsdb.err"
       tsdb_out = "#{testpath}/tsdb.out"
-      pid = fork do
+      tsdb_daemon_pid = fork do
         $stderr.reopen(tsdb_err, "w")
         $stdout.reopen(tsdb_out, "w")
         exec("#{bin}/start-tsdb.sh")
       end
-
       sleep 15
 
-      pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0\n")
+      begin
+        pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0\n")
 
-      system "#{bin}/tsdb", "query", "1356998000", "1356999000", "sum", "homebrew.install.test", "host=webserver01", "cpu=0"
+        system "#{bin}/tsdb", "query", "1356998000", "1356999000", "sum", "homebrew.install.test", "host=webserver01", "cpu=0"
+      ensure
+        Process.kill(9, tsdb_daemon_pid)
+      end
     ensure
-      Process.kill(9, pid) if defined? pid && pid > 1
-
       system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
     end
   end
--- a/Formula/opentsdb.rb
+++ b/Formula/opentsdb.rb
@@ -144,7 +144,7 @@ class Opentsdb < Formula
 
       system "#{bin}/tsdb", "query", "1356998000", "1356999000", "sum", "homebrew.install.test", "host=webserver01", "cpu=0"
     ensure
-      Process.kill(9, pid) if defined? pid and pid > 1
+      Process.kill(9, pid) if defined? pid && pid > 1
 
       system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
     end
--- a/Formula/opentsdb.rb
+++ b/Formula/opentsdb.rb
@@ -1,0 +1,149 @@
+class HbaseLZORequirement < Requirement
+  fatal true
+
+  satisfy(:build_env => false) { Tab.for_name("hbase").with?("lzo") }
+
+  def message; <<-EOS.undent
+    hbase must not have disabled lzo compression to use it in opentsdb:
+      brew install hbase
+      not
+      brew install hbase --without-lzo
+    EOS
+  end
+end
+
+class Opentsdb < Formula
+  desc "Scalable, distributed Time Series Database."
+  homepage "http://opentsdb.net/"
+  url "https://github.com/OpenTSDB/opentsdb/releases/download/v2.2.0/opentsdb-2.2.0.tar.gz"
+  sha256 "5689d4d83ee21f1ce5892d064d6738bfa9fdef99f106f45d5c38eefb9476dfb5"
+
+  depends_on "hbase"
+  depends_on "lzo" => :recommended
+  depends_on HbaseLZORequirement if build.with?("lzo")
+  depends_on :java => "1.6+"
+  depends_on "gnuplot" => :optional
+
+  def install
+    # submitted to upstream: https://github.com/OpenTSDB/opentsdb/pull/711
+    # pulled to next branch: https://github.com/OpenTSDB/opentsdb/commit/5d0cfa9b4b6d8da86735efeea4856632581a7adb.patch
+    # doesn't apply cleanly on this release though
+    # mkdir_p is called from in a subdir of build so needs an extra ../ and there is no rule to create $(classes) and
+    # everything builds without specifying them as dependencies of the jar.
+    inreplace "Makefile.in" do |s|
+      s.sub!(/(\$\(jar\): manifest \.javac-stamp) \$\(classes\)/, '\1')
+      s.sub!(/(echo " \$\(mkdir_p\) '\$\$dstdir'"; )/, '\1../')
+    end
+
+    mkdir "build" do
+      system "../configure",
+             "--disable-silent-rules",
+             "--prefix=#{prefix}",
+             "--mandir=#{man}",
+             "--sysconfdir=#{etc}",
+             "--localstatedir=#{var}/opentsdb"
+      system "make"
+      bin.mkpath
+      system "make", "install-exec-am", "install-data-am"
+    end
+
+    env = {
+      :HBASE_HOME => Formula["hbase"].opt_libexec,
+      :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE"),
+    }
+    env = Language::Java.java_home_env.merge(env)
+    create_table = pkgshare/"tools/create_table_with_env.sh"
+    create_table.write_env_script pkgshare/"tools/create_table.sh", env
+    create_table.chmod 0755
+
+    inreplace pkgshare/"etc/opentsdb/opentsdb.conf", "/usr/share", "#{HOMEBREW_PREFIX}/share"
+    etc.install Dir["#{pkgshare}/etc/opentsdb"]
+    (pkgshare/"plugins/.keep").write ""
+
+    (bin/"start-tsdb.sh").write <<-EOS.undent
+      #!/bin/sh
+      exec "#{opt_bin}/tsdb" tsd \\
+        --config="#{etc}/opentsdb/opentsdb.conf" \\
+        --staticroot="#{opt_pkgshare}/static/" \\
+        --cachedir="#{var}/cache/opentsdb" \\
+        --port=4242 \\
+        --zkquorum=localhost:2181 \\
+        --zkbasedir=/hbase \\
+        --auto-metric \\
+        "$@"
+    EOS
+  end
+
+  def post_install
+    (var/"cache/opentsdb").mkpath
+
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
+    sleep 2
+    system "#{pkgshare}/tools/create_table_with_env.sh"
+    system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
+  end
+
+  plist_options :manual => "#{HOMEBREW_PREFIX}/opt/opentsdb/bin/start-tsdb.sh"
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+    <dict>
+      <key>KeepAlive</key>
+      <dict>
+        <key>OtherJobEnabled</key>
+        <string>#{Formula["hbase"].plist_name}</string>
+      </dict>
+      <key>Label</key>
+      <string>#{plist_name}</string>
+      <key>ProgramArguments</key>
+      <array>
+        <string>#{opt_bin}/start-tsdb.sh</string>
+      </array>
+      <key>WorkingDirectory</key>
+      <string>#{HOMEBREW_PREFIX}</string>
+      <key>StandardOutPath</key>
+      <string>#{var}/opentsdb/opentsdb.log</string>
+      <key>StandardErrorPath</key>
+      <string>#{var}/opentsdb/opentsdb.err</string>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    cp_r (Formula["hbase"].opt_libexec/"conf"), testpath
+    inreplace (testpath/"conf/hbase-site.xml") do |s|
+      s.gsub! /(hbase.rootdir.*)\n.*/, "\\1\n<value>file://#{testpath}/hbase</value>"
+      s.gsub! /(hbase.zookeeper.property.dataDir.*)\n.*/, "\\1\n<value>#{testpath}/zookeeper</value>"
+    end
+
+    ENV["HBASE_LOG_DIR"]  = testpath/"logs"
+    ENV["HBASE_CONF_DIR"] = testpath/"conf"
+    ENV["HBASE_PID_DIR"]  = testpath/"pid"
+
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
+    sleep 2
+
+    system "#{pkgshare}/tools/create_table_with_env.sh"
+
+    tsdb_err = "#{testpath}/tsdb.err"
+    tsdb_out = "#{testpath}/tsdb.out"
+    pid = fork do
+      $stderr.reopen(tsdb_err, "w")
+      $stdout.reopen(tsdb_out, "w")
+      exec("#{bin}/start-tsdb.sh")
+    end
+
+    sleep 15
+
+    pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0\n")
+
+    system "#{bin}/tsdb", "query", "1356998000", "1356999000", "sum", "homebrew.install.test", "host=webserver01", "cpu=0"
+
+    Process.kill(9, pid)
+
+    system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
+  end
+end
--- a/Formula/opentsdb.rb
+++ b/Formula/opentsdb.rb
@@ -57,7 +57,7 @@ class Opentsdb < Formula
     create_table.chmod 0755
 
     inreplace pkgshare/"etc/opentsdb/opentsdb.conf", "/usr/share", "#{HOMEBREW_PREFIX}/share"
-    etc.install Dir["#{pkgshare}/etc/opentsdb"]
+    etc.install pkgshare/"etc/opentsdb"
     (pkgshare/"plugins/.keep").write ""
 
     (bin/"start-tsdb.sh").write <<-EOS.undent
@@ -76,11 +76,13 @@ class Opentsdb < Formula
 
   def post_install
     (var/"cache/opentsdb").mkpath
-
-    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
-    sleep 2
-    system "#{pkgshare}/tools/create_table_with_env.sh"
-    system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
+    begin
+      system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
+      sleep 2
+      system "#{pkgshare}/tools/create_table_with_env.sh"
+    ensure
+      system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
+    end
   end
 
   plist_options :manual => "#{HOMEBREW_PREFIX}/opt/opentsdb/bin/start-tsdb.sh"
@@ -122,28 +124,29 @@ class Opentsdb < Formula
     ENV["HBASE_LOG_DIR"]  = testpath/"logs"
     ENV["HBASE_CONF_DIR"] = testpath/"conf"
     ENV["HBASE_PID_DIR"]  = testpath/"pid"
+    begin
+      system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
+      sleep 2
 
-    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
-    sleep 2
+      system "#{pkgshare}/tools/create_table_with_env.sh"
 
-    system "#{pkgshare}/tools/create_table_with_env.sh"
+      tsdb_err = "#{testpath}/tsdb.err"
+      tsdb_out = "#{testpath}/tsdb.out"
+      pid = fork do
+        $stderr.reopen(tsdb_err, "w")
+        $stdout.reopen(tsdb_out, "w")
+        exec("#{bin}/start-tsdb.sh")
+      end
 
-    tsdb_err = "#{testpath}/tsdb.err"
-    tsdb_out = "#{testpath}/tsdb.out"
-    pid = fork do
-      $stderr.reopen(tsdb_err, "w")
-      $stdout.reopen(tsdb_out, "w")
-      exec("#{bin}/start-tsdb.sh")
+      sleep 15
+
+      pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0\n")
+
+      system "#{bin}/tsdb", "query", "1356998000", "1356999000", "sum", "homebrew.install.test", "host=webserver01", "cpu=0"
+    ensure
+      Process.kill(9, pid) if defined? pid and pid > 1
+
+      system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
     end
-
-    sleep 15
-
-    pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0\n")
-
-    system "#{bin}/tsdb", "query", "1356998000", "1356999000", "sum", "homebrew.install.test", "host=webserver01", "cpu=0"
-
-    Process.kill(9, pid)
-
-    system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
   end
 end


Formula/opentsdb.rb
```diff
@@ -0,0 +1,121 @@
...
+  homepage "http://opentsdb.net/"
+  url "https://github.com/OpenTSDB/opentsdb/releases/download/v2.2.0/opentsdb-2.2.0.tar.gz"
+  sha256 "5689d4d83ee21f1ce5892d064d6738bfa9fdef99f106f45d5c38eefb9476dfb5"
+
+  option "without-lzo", "Don't use LZO Compression"
+
+  if build.without? "lzo"
+    depends_on "hbase"
+  else
+    depends_on "hbase" => "with-lzo"
```
mistydemeo: This is the default; hbase defines a `--without-lzo` (which is probably what you want above).

Formula/opentsdb.rb
```diff
@@ -0,0 +1,121 @@
...
+    mkdir "build" do
+      system "../configure",
+             "--disable-silent-rules",
+             "--prefix=#{prefix}",
+             "--mandir=#{man}",
+             "--sysconfdir=#{etc}",
+             "--localstatedir=#{var}/opentsdb"
+      system "make"
+      system "make", "install-exec-am"
+      system "make", "install-data-am"
```
bfontaine: Can’t you do those two with one command?
```rb
       system "make"
       system "make", "install-exec-am", "install-data-am"
```

Formula/opentsdb.rb
CamJN: I just wanted to verify the correct way to specify this. If you installed hbase with lzo but (for some reason) don't want lzo for opentsdb you shouldn't be forced to reinstall hbase without lzo, that's silly. But using lzo with opentsdb does require that you installed hbase with lzo.

My reasoning is a bit convoluted, sorry.

DomT4: We don't actually have a simple way to mandate you must have built with a default option at this point. You can _kind of_ force it with a Requirement [like this](https://github.com/Homebrew/homebrew-core/blob/ed890fc46b99f54e387caae92d7e5dd01bfdbbac/Formula/coq.rb#L1-L11) but it's a dance.

CamJN:
I'm getting a `java.lang.Error: Properties init: Could not determine current working directory.` in the CI.

Formula/opentsdb.rb
mikemcquaid: @CamJN Just a note on this.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,139 @@
...
+    end
+
+    env = Language::Java.java_home_env.merge(:HBASE_HOME => Formula["hbase"].opt_libexec, :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE"))
+    (bin/"create_table.sh").write_env_script opt_pkgshare/"tools/create_table.sh", env
+  end
+
+  def post_install
+    (var/"cache/opentsdb").mkpath
+    inreplace opt_pkgshare/"etc/opentsdb/opentsdb.conf", "usr/share", "usr/local/share"
+    etc.install Dir["#{opt_pkgshare}/etc/opentsdb"]
```
mikemcquaid: Can this line and the `inreplace` go into `install`?

Formula/opentsdb.rb
```diff
@@ -0,0 +1,139 @@
...
+      system "make", "install-data-am"
+    end
+
+    env = Language::Java.java_home_env.merge(:HBASE_HOME => Formula["hbase"].opt_libexec, :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE"))
+    (bin/"create_table.sh").write_env_script opt_pkgshare/"tools/create_table.sh", env
+  end
+
+  def post_install
+    (var/"cache/opentsdb").mkpath
+    inreplace opt_pkgshare/"etc/opentsdb/opentsdb.conf", "usr/share", "usr/local/share"
```
UniqMartin: What is this replacing? If that's replacing `"/usr/share"` with `"/usr/local/share"`, then the latter should probably be `"#{HOMEBREW_PREFIX}/share"` so that things continue to work for Homebrew users who install outside of `/usr/local`.

Formula/opentsdb.rb
CamJN: I don't think so, I had it in `install` to begin with but it didn't do anything. I have no clue why it didn't work, and would welcome suggestions. But I tried several different variations before giving up and putting it here.

Formula/opentsdb.rb
CamJN: Nope:
```
Last 15 lines from /Users/camdennarzt/Library/Logs/Homebrew/opentsdb/03.make:
mkdir: /usr/local/Cellar/opentsdb/2.2.0/share: File exists
mkdir: /usr/local/Cellar/opentsdb/2.2.0/share/opentsdb: File exists
mkdir: /usr/local/Cellar/opentsdb/2.2.0/share: File exists
mkdir: /usr/local/Cellar/opentsdb/2.2.0/share/opentsdb: File exists
 /usr/bin/install -c ../src/mygnuplot.sh ../src/mygnuplot.bat /usr/local/Cellar/opentsdb/2.2.0/bin/tsdb /usr/local/Cellar/opentsdb/2.2.0/share/opentsdb/bin
 /usr/bin/install -c ../tools/check_tsd ../tools/clean_cache.sh ../tools/opentsdb_restart.py ../tools/tsddrain.py ../src/create_table.sh ../src/upgrade_1to2.sh /usr/local/Cellar/opentsdb/2.2.0/share/opentsdb/tools
 ../build-aux/install-sh -c -d /usr/local/Cellar/opentsdb/2.2.0/share/opentsdb/etc/init.d
install: /usr/local/Cellar/opentsdb/2.2.0/bin/tsdb: No such file or directory
 /usr/bin/install -c tsdb '/usr/local/Cellar/opentsdb/2.2.0/bin'
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,140 @@
...
+    env = Language::Java.java_home_env.merge(:HBASE_HOME => Formula["hbase"].opt_libexec, :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE"))
+    (bin/"create_table.sh").write_env_script opt_pkgshare/"tools/create_table.sh", env
+
+    inreplace pkgshare/"etc/opentsdb/opentsdb.conf", "/usr/share", "#{HOMEBREW_PREFIX}/share"
+    etc.install Dir["#{pkgshare}/etc/opentsdb"]
+  end
+
+  def post_install
+    (var/"cache/opentsdb").mkpath
+    (pkgshare/"plugins").mkpath
```
UniqMartin: Please move this line into `install`, too. A formula is not supposed to make any changes in its `prefix` (of which `pkgshare` is a subdirectory) after `install` finishes.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,140 @@
...
+
+    system "#{Formula["hbase"].opt_libexec}/bin/start-hbase.sh"
+
+    system "#{bin}/create_table.sh"
```
UniqMartin: Purely stylistic nit, but I think you can safely remove the empty lines surrounding this one without negatively affecting readability.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+
+    ENV["HBASE_LOG_DIR"]  = (testpath/"logs")
+    ENV["HBASE_CONF_DIR"] = (testpath/"conf")
+    ENV["HBASE_PID_DIR"]  = (testpath/"pid")
+
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
+    sleep 2
+
+    pid = fork { exec "#{bin}/tsdb tsd --config=#{HOMEBREW_PREFIX}/etc/opentsdb/opentsdb.conf --staticroot=#{HOMEBREW_PREFIX}/opt/opentsdb/share/opentsdb/static/ --cachedir=#{HOMEBREW_PREFIX}/var/cache/opentsdb --port=4242 --zkquorum=localhost:2181 --zkbasedir=/hbase --auto-metric" }
```
UniqMartin: Try to shorten this by wrapping the line and passing the arguments individually to `exec`. Something like this should work (and avoids an intermediate shell for argument parsing/splitting):

```ruby
pid = fork do
  exec "#{bin}/tsdb", "tsd", "--config=#{etc}/opentsdb/opentsdb.conf",
                             "--staticroot=#{opt_pkgshare}/static/",
                             "--cachedir=#{var}/cache/opentsdb",
                             "--port=4242",
                             "--zkquorum=localhost:2181",
                             "--zkbasedir=/hbase",
                             "--auto-metric"
end
```

(Note that I'm using some convenience methods, like you already used in your `.plist` file, to avoid using `HOMEBREW_PREFIX`.)

Formula/opentsdb.rb
CamJN: Just tried, didn't work. What's with that?

Formula/opentsdb.rb
CamJN: @UniqMartin done.

Formula/opentsdb.rb
CamJN: @UniqMartin done.

Formula/opentsdb.rb
CamJN: @UniqMartin specifically:
```
diff --git a/Formula/opentsdb.rb b/Formula/opentsdb.rb
index fde4031..89dbfef 100644
--- a/Formula/opentsdb.rb
+++ b/Formula/opentsdb.rb
@@ -58,11 +58,11 @@ class Opentsdb < Formula
 
     inreplace pkgshare/"etc/opentsdb/opentsdb.conf", "/usr/share", "#{HOMEBREW_PREFIX}/share"
     etc.install Dir["#{pkgshare}/etc/opentsdb"]
+    (pkgshare/"plugins").mkpath
   end
 
   def post_install
     (var/"cache/opentsdb").mkpath
-    (pkgshare/"plugins").mkpath
 
     system "#{Formula["hbase"].opt_libexec}/bin/start-hbase.sh"
     system "#{bin}/create_table.sh"
```
results in no plugins directory in `/usr/local/share/opentsdb/`

Formula/opentsdb.rb
mikemcquaid: Looks like you may just need a `bin.mkpath` first?

Formula/opentsdb.rb
UniqMartin: Sorry, you're right about that. I forgot that empty directories are automatically removed in a cleanup step after `install` completes. This raises a question though: Why do you need an empty `plugins` directory in that location? It's not going to be of any use as you're not allowed to place anything in it once `install` has finished. So what's the original motivation for doing this?

CamJN: I need a real directory to point to for plugins to reside or opentsdb won't start. It's dumb, but that's the way it is. And I figured I should create one because I don't know what would happen if there were non-plugins in whatever dir I pointed at.

UniqMartin: I see. You will need to create a file to prevent the directory from being removed in this case:

```ruby
(pkgshare/"plugins/.keep").write ""
```

(This is the equivalent of `touch` and will automatically create the containing directory, too.)

CamJN:
@mikemcquaid @mistydemeo @bfontaine @DomT4 @UniqMartin 
All feedback addressed, Yosemite and El Capitan are fine in the CI, but mavericks hangs during the test. Is there something weird about the mavericks worker? Or some way to diagnose this? I don't have a mavericks box myself.

UniqMartin:
> All feedback addressed, Yosemite and El Capitan are fine in the CI, […]

:+1:

> […] but mavericks hangs during the test. Is there something weird about the mavericks worker? Or some way to diagnose this? I don't have a mavericks box myself.

I suspect that's not an issue related to the OS X version, but to the fact that we're using Ruby 1.8.7 for testing on 10.9. Thus you could try to reproduce the issue locally by doing approximately the following:

```sh
brew install homebrew/versions/ruby187
export HOMEBREW_DEVELOPER=1 # HOMEBREW_RUBY_PATH will be ignored otherwise.
export HOMEBREW_RUBY_PATH="$(brew --prefix)/opt/ruby187/bin/ruby"
brew config # Check the “System Ruby” line.
brew test --verbose opentsdb
```

CamJN:
I just ran the test with ruby 1.8.7-p375 and it ran through at normal speed and succeeded.

CamJN:
@mikemcquaid @mistydemeo @bfontaine @DomT4 @UniqMartin 
Since ruby 1.8.7 isn't the problem (at least on my 10.11 box) what else could this be?

CamJN:
bump

UniqMartin:
@BrewTestBot test this please

UniqMartin:
@CamJN Sorry for the delay! I just re-queued the job in the hopes that a recent Jenkins update has changed something. If this is not going to resolve the issue on OS X 10.9, I'll try to figure out what's going on there and how we can modify the test to make it pass.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,151 @@
...
+    depends_on "lzo"
+    depends_on HbaseLZORequirement
+  end
+  depends_on :java => "1.6+"
+  depends_on "gnuplot" => :optional
+
+  def install
+    # submitted to upstream https://github.com/OpenTSDB/opentsdb/pull/711
+    # mkdir_p is called from in a subdir of build so needs an extra ../ and there is no rule to create $(classes) and
+    # everything builds without specifying them as dependencies of the jar.
```
mikemcquaid: Looks like they merged it; is there an upstream commit we can use in a `patch do`?

CamJN: It's not on the master branch, does that matter?

patch url: https://github.com/OpenTSDB/opentsdb/commit/5d0cfa9b4b6d8da86735efeea4856632581a7adb.patch

UniqMartin: No, it doesn't. I suspect the `next` branch is going to be merged into `master` at some point in time? Let's use the patch URL in a `patch do` block (instead of the `inreplace`) if it applies cleanly.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  satisfy(:build_env => false) { !Tab.for_name("hbase").without?("lzo") }
+
+  def message; <<-EOS.undent
+  hbase must not have disabled lzo compression to use it in opentsdb:
+    brew install hbase
+    not
+    brew install hbase --without-lzo
+  EOS
+  end
```
UniqMartin: The indentation is slightly off here. We prefer it to be as follows (see also most `caveats` methods of other formulae):

```ruby
def message; <<-EOS.undent
  hbase must not have disabled lzo compression to use it in opentsdb:
    brew install hbase
    not
    brew install hbase --without-lzo
  EOS
end
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,151 @@
+class HbaseLZORequirement < Requirement
+  fatal true
+
+  satisfy(:build_env => false) { !Tab.for_name("hbase").without?("lzo") }
```
UniqMartin: You can use `Tab.for_name("hbase").with?("lzo")` instead of `!Tab.for_name("hbase").without?("lzo")` to avoid the double negation.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  option "without-lzo", "Don't use LZO Compression"
+
+  if build.without? "lzo"
+    depends_on "hbase"
+  else
+    depends_on "hbase"
+    depends_on "lzo"
+    depends_on HbaseLZORequirement
+  end
```
UniqMartin: Please replace this `if`-`else` block with the following code snippet. The benefit is that there's less duplication and the dependencies can be more easily discovered by our dependency resolution code (`depends_on "lzo" => :recommended` is always preferable over `depends_on "lzo" if build.with?("lzo")`).

```ruby
depends_on "hbase"
depends_on "lzo" => :recommended
depends_on HbaseLZORequirement if build.with?("lzo")
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,151 @@
...
+             "--prefix=#{prefix}",
+             "--mandir=#{man}",
+             "--sysconfdir=#{etc}",
+             "--localstatedir=#{var}/opentsdb"
+      system "make"
+      bin.mkpath
+      system "make", "install-exec-am", "install-data-am"
+    end
+
+    env = Language::Java.java_home_env.merge(:HBASE_HOME => Formula["hbase"].opt_libexec, :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE"))
```
UniqMartin: Let's break this into multiple lines to avoid this overly long line and make the code more readable. Here's a suggestion:

```ruby
env = {
  :HBASE_HOME => Formula["hbase"].opt_libexec,
  :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE")),
}
env = Language::Java.java_home_env.merge(env)
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,151 @@
...
+    (bin/"create_table.sh").write_env_script opt_pkgshare/"tools/create_table.sh", env
+
+    inreplace pkgshare/"etc/opentsdb/opentsdb.conf", "/usr/share", "#{HOMEBREW_PREFIX}/share"
+    etc.install Dir["#{pkgshare}/etc/opentsdb"]
+    (pkgshare/"plugins").mkpath
```
UniqMartin: This line can be dropped. The following `(pkgshare/"plugins/.keep").write ""` takes care of creating all directories in that path that don't exist yet.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,151 @@
...
+
+    system "#{Formula["hbase"].opt_libexec}/bin/start-hbase.sh"
+    system "#{bin}/create_table.sh"
+    system "#{Formula["hbase"].opt_libexec}/bin/stop-hbase.sh"
+  end
+
+  plist_options :manual => "tsdb tsd --config=#{HOMEBREW_PREFIX}/etc/opentsdb/opentsdb.conf "\
+                           "--staticroot=#{HOMEBREW_PREFIX}/opt/opentsdb/share/opentsdb/static/ "\
+                           "--cachedir=#{HOMEBREW_PREFIX}/var/cache/opentsdb --port=4242 "\
+                           "--zkquorum=localhost:2181 --zkbasedir=/hbase --auto-metric"
```
UniqMartin: Should this `tsdb` be `#{opt_bin}/tsdb` instead? Given the long list of arguments here I wonder if we can refactor this to avoid duplicating all those arguments here, in the `plist` method, and in the `test do` block. I'm thinking something along the lines of:

```ruby
def tsdb_args
  %W[
    #{opt_bin}/tsdb
    tsd
    --config=#{HOMEBREW_PREFIX}/etc/opentsdb/opentsdb.conf
    --staticroot=#{HOMEBREW_PREFIX}/opt/opentsdb/share/opentsdb/static/
    --cachedir=#{HOMEBREW_PREFIX}/var/cache/opentsdb
    --port=4242
    --zkquorum=localhost:2181
    --zkbasedir=/hbase
    --auto-metric
  ]
end
```

It could be used here as follows:

```ruby
plist_options :manual => tsdb_args.join(" ")
```

And in the `plist` method:

```ruby
def plist
  program_args = tsdb_args.map { |s| "    <string>#{s}</string>" }.join("\n")
  <<-EOS.undent
    … snip …
      <key>ProgramArguments</key>
      <array>
    #{program_args}
      </array>
    … snip …
  EOS
end
```

UniqMartin:
> I just re-queued the job in the hopes that a recent Jenkins update has changed something. If this is not going to resolve the issue on OS X 10.9, I'll try to figure out what's going on there and how we can modify the test to make it pass.

Unfortunately, this didn't work. I'm now trying to reproduce the test bot failure locally. I'll report back as soon as there's something to report on.

UniqMartin:
@CamJN Can you address the code comments Mike and I left in the mean time? Not strictly necessary, but it would also help if you squashed down your branch to a single commit titled `opentsdb 2.2.0 (new formula)` and rebased it on an up-to-date master before your next push. Thanks!

CamJN:
Changes applied, but there's an issue with the manual plist bit

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+    ENV["HBASE_CONF_DIR"] = (testpath/"conf")
+    ENV["HBASE_PID_DIR"]  = (testpath/"pid")
+
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
+    sleep 2
+
+    system "#{opt_bin}/create_table.sh"
+
+    pid = fork do
+      exec tsdb_args.join(" ")
```
UniqMartin: I'm pretty sure this should be `exec(*tsdb_args)` instead, as we want to pass the arguments directly to the executable (the one in the first array element) without involving an intermediate shell.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    cp_r (Formula["hbase"].libexec/"conf"), testpath
+    inreplace (testpath/"conf/hbase-site.xml") do |s|
+      s.gsub! /(hbase.rootdir.*)\n.*/, "\\1\n<value>file://#{testpath}/hbase</value>"
+      s.gsub! /(hbase.zookeeper.property.dataDir.*)\n.*/, "\\1\n<value>#{testpath}/zookeeper</value>"
+    end
+
+    ENV["HBASE_LOG_DIR"]  = (testpath/"logs")
+    ENV["HBASE_CONF_DIR"] = (testpath/"conf")
+    ENV["HBASE_PID_DIR"]  = (testpath/"pid")
```
UniqMartin: You can omit the parentheses that surround the right-hand sides here.

UniqMartin:
> Changes applied, but there's an issue with the manual plist bit

Can you elaborate? I'm not sure what the issue is you're referring to in this statement.

CamJN:
```
Error: undefined local variable or method `tsdb_args' for #<Class:0x007fb7c333c778>
Please report this bug:
/Users/camdennarzt/Developer/Ruby/homebrew-core/formula/opentsdb.rb:82:in `<class:Opentsdb>'
/Users/camdennarzt/Developer/Ruby/homebrew-core/formula/opentsdb.rb:15:in `load_formula'
/usr/local/Library/Homebrew/formulary.rb:21:in `module_eval'
/usr/local/Library/Homebrew/formulary.rb:21:in `load_formula'
/usr/local/Library/Homebrew/formulary.rb:38:in `load_formula_from_path'
/usr/local/Library/Homebrew/formulary.rb:87:in `load_file'
/usr/local/Library/Homebrew/formulary.rb:78:in `klass'
/usr/local/Library/Homebrew/formulary.rb:74:in `get_formula'
/usr/local/Library/Homebrew/formulary.rb:211:in `factory'
/usr/local/Library/Homebrew/extend/ARGV.rb:29:in `block in resolved_formulae'
/usr/local/Library/Homebrew/extend/ARGV.rb:27:in `map'
/usr/local/Library/Homebrew/extend/ARGV.rb:27:in `resolved_formulae'
/usr/local/Library/Homebrew/cmd/audit.rb:57:in `audit'
/usr/local/Library/brew.rb:87:in `<main>'
```

mikemcquaid:
@CamJN Please inline `tsdb_args` in each of the uses instead, thanks.

CamJN:
@mikemcquaid @UniqMartin can you two hash that out? I was just told to do the opposite.

UniqMartin:
> Please inline `tsdb_args` in each of the uses instead, thanks.

[This was my suggestion](https://github.com/Homebrew/homebrew-core/pull/813#discussion_r64073637) because the exact same sequence of arguments (a non-trivial list) was duplicated in three places. I wanted to avoid this kind of duplication as this increases the likelihood of them drifting apart in the future and centralizing this makes the formula more concise. (I didn't realize this creates issues when used with `plist_options`.) Is there a better solution than just inlining this?

Formula/opentsdb.rb
mikemcquaid: @UniqMartin Duplication is not a problem in formulae as they should be human readable DSL rather than DRY software.

mikemcquaid:
@UniqMartin In this case it doesn't work in one of the cases and makes the other two harder to read. We could consider writing these into a dedicated script.

UniqMartin:
> can you two hash that out? I was just told to do the opposite.

@CamJN [Mike has a good argument](https://github.com/Homebrew/homebrew-core/pull/813#discussion_r64084064) and his opinion carries more weight, thus it's back to inlining. Sorry for suggesting a problematic solution and for this back and forth.

> We could consider writing these into a dedicated script.

@CamJN If you think it's a good idea to hard-code all (or most) of the arguments in a dedicated script, I think this would create a happy middle ground. The script could be created in the `install` method and look something like (just a rough and untested sketch):

```ruby
(bin/"script_name").write <<-EOS.undent
  #!/bin/sh
  exec "#{opt_bin}/tsdb" tsd \
    --config="#{HOMEBREW_PREFIX}/etc/opentsdb/opentsdb.conf" \
    --staticroot="#{HOMEBREW_PREFIX}/opt/opentsdb/share/opentsdb/static/" \
    --cachedir="#{HOMEBREW_PREFIX}/var/cache/opentsdb" \
    --port=4242 \
    --zkquorum=localhost:2181 \
    --zkbasedir=/hbase \
    --auto-metric \
    "$@"
EOS
```

CamJN:
Ok, script is made, i had to revert the `patch do` bit b/c it didn't apply for some reason. Fixed the latest style things.

Mavericks still FUBAR.

UniqMartin:
> Mavericks still FUBAR.

Yes, I'm looking into this right now. I'll report back as soon as possible.

> Ok, script is made, i had to revert the `patch do` bit b/c it didn't apply for some reason. Fixed the latest style things.

Thanks! There are still some minor issues that will need to be addressed, but those can wait until the Mavericks issue has been sorted out. (At the moment, pushing new commits will just trigger new builds and those will waste CI time without achieving much.)

UniqMartin:
When testing things locally, I wasn't able to reproduce the problem exactly, but I had some random failures and fixing them seems to have also fixed the test on CI (I experimented a bit in PR #1321, see also the corresponding CI builds). The problem is twofold:

- The main issue seems to be that Jenkins is somewhat overwhelmed by the amount of output generated by the TSDB daemon and possibly by the fact that both the daemon and the Ruby process running the test concurrently write to standard output/error. Redirecting the daemon output to log files helped. (See UniqMartin/homebrew-core@da63e30fe695cecac04d7658ba5966151d308b6f.)

- Sometimes connecting to the TSDB daemon only 2 seconds after it had started wouldn't be sufficient to get the daemon into a state where it would accept connections. Bumping this to 15 seconds makes this a lot more robust, though ideally there was a cleaner way to detec when the daemon is actually ready. (See UniqMartin/homebrew-core@fadec6d2696989caaf4a56e30a3568e0874c3aa9.)

Furthermore there are some pending fixes to the remaining formula code:

- Some indentation is off, backslashes need to be escaped, some paths can be simplified, and the `.plist` file is broken. (See UniqMartin/homebrew-core@dd858ad0084f9d62852a2859395e47fa01ea1dce.)

And finally there are some new/open questions:

- Does the wrapper script `create_table.sh` actually need to live in `bin`? It seems to be only needed during setup (in `postinstall`) and I have the feeling it's undesirable to have that permanently in PATH (and mixed with all the other binaries from all installed formulae). The name is also quite generic. My suggestion would be to place it somewhere else, e.g. in `libexec`.

- Similarly, the newly created wrapper script `start-tsdb.sh` probably doesn't need to live in `bin` and can be relocated to `libexec`, though its name is more specific.

(Feel free to cherry pick the listed commits into your branch and then squash them down again.)

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    pid = fork do
+      exec("#{opt_bin}/start-tsdb.sh")
+    end
+
+    sleep 2
+
+    socket = TCPSocket.new "localhost", 4242
+    socket.puts "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0"
+    socket.close
```
UniqMartin: Can we replace this block and the `require "socket"` (further up) with a call to `nc` like we already do [in `hbase.rb`](https://github.com/Homebrew/homebrew-core/blob/f5016a81d421eda8960974819e742d47dab24fc5/Formula/hbase.rb#L163)? We could simply discard the output of `pipe_output`, but if we can check the server response (if any) for something, that would be even better …

CamJN:
A couple things:

- The `start-tsdb.sh` script should stay in bin IMO, just like `mysql.server`, `hbase-start.sh`, `pgctl`, `startx` etc.
- And there's no output from the `nc`, so nothing to check.

CamJN:
Moving the `create_table.sh` script outside of bin broke the formula.

UniqMartin:
> Moving the `create_table.sh` script outside of bin broke the formula.

I don't think that's the issue here. The problem seems to be a confusion about when to use the `opt_<path>` methods and when to use the `<path>` methods. The former reference the version-independent path accessible via the `opt` symlink, e.g. `opt/opentsdb/<path>`, and the latter point directly into the formula prefix, e.g. `Cellar/opentsdb/2.2.0/<path>`.

When accessing stuff that is related to the formula currently being built, the non-`opt` version should be preferred (unless this is a path that is going to end up in some configuration file like the `.plist` file). The `opt/opentsdb` symlink is only created once `install` completes.

When accessing dependencies (like `hbase` in this case), we usually want to reference the version that is currently installed and linked by the user, thus the `opt_*` variants should be used in all instances.

(If you give me a few minutes, I'll review the formula in regards to this and point out all the places that need to be changed. I have to admit that I haven't paid too much attention to this in previous reviews.)

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+      bin.mkpath
+      system "make", "install-exec-am", "install-data-am"
+    end
+
+    env = {
+      :HBASE_HOME => Formula["hbase"].opt_libexec,
+      :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE"),
+    }
+    env = Language::Java.java_home_env.merge(env)
+    (bin/"create_table.sh").write_env_script opt_pkgshare/"tools/create_table.sh", env
```
UniqMartin:
```diff
-    (bin/"create_table.sh").write_env_script opt_pkgshare/"tools/create_table.sh", env
+    create_table = pkgshare/"tools/create_table_with_env.sh"
+    create_table.write_env_script pkgshare/"tools/create_table.sh", env
+    create_table.chmod 0755
```

The problem here was that `write_env_script` doesn't bother to make the file executable (likely a bug), but instead relies on a post-install cleanup step to make files executable that are in locations like `bin`. Relocating it to `#{pkgshare}/tools/` caused it to end up in a location where that magic wouldn't kick in (and hence break the build).

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+        --zkbasedir=/hbase \\
+        --auto-metric \\
+        "$@"
+    EOS
+  end
+
+  def post_install
+    (var/"cache/opentsdb").mkpath
+
+    system "#{Formula["hbase"].opt_libexec}/bin/start-hbase.sh"
```
UniqMartin:
```diff
-    system "#{Formula["hbase"].opt_libexec}/bin/start-hbase.sh"
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+    sleep 2
+    system "#{opt_bin}/create_table.sh"
+    system "#{Formula["hbase"].opt_libexec}/bin/stop-hbase.sh"
```
UniqMartin:
```diff
-    system "#{opt_bin}/create_table.sh"
-    system "#{Formula["hbase"].opt_libexec}/bin/stop-hbase.sh"
+    system "#{pkgshare}/tools/create_table_with_env.sh"
+    system "#{Formula["hbase"].opt_bin}/stop-hbase.sh"
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+      <string>#{var}/opentsdb/opentsdb.log</string>
+      <key>StandardErrorPath</key>
+      <string>#{var}/opentsdb/opentsdb.err</string>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    cp_r (Formula["hbase"].libexec/"conf"), testpath
```
UniqMartin:
```diff
-    cp_r (Formula["hbase"].libexec/"conf"), testpath
+    cp_r (Formula["hbase"].opt_libexec/"conf"), testpath
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+    end
+
+    ENV["HBASE_LOG_DIR"]  = testpath/"logs"
+    ENV["HBASE_CONF_DIR"] = testpath/"conf"
+    ENV["HBASE_PID_DIR"]  = testpath/"pid"
+
+    system "#{Formula["hbase"].opt_bin}/start-hbase.sh"
+    sleep 2
+
+    system "#{opt_bin}/create_table.sh"
```
UniqMartin:
```diff
-    system "#{opt_bin}/create_table.sh"
+    system "#{pkgshare}/tools/create_table_with_env.sh"
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    tsdb_err = "#{testpath}/tsdb.err"
+    tsdb_out = "#{testpath}/tsdb.out"
+    pid = fork do
+      $stderr.reopen(tsdb_err, "w")
+      $stdout.reopen(tsdb_out, "w")
+      exec("#{opt_bin}/start-tsdb.sh")
```
UniqMartin:
```diff
-      exec("#{opt_bin}/start-tsdb.sh")
+      exec("#{bin}/start-tsdb.sh")
```

Formula/opentsdb.rb
```diff
@@ -0,0 +1,147 @@
...
+    end
+
+    sleep 15
+
+    pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0")
```
UniqMartin:
```diff
-    pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0")
+    pipe_output("nc localhost 4242 2>&1", "put homebrew.install.test 1356998400 42.5 host=webserver01 cpu=0\n")
```

The newline at the end is really important here! Otherwise the request won't be properly terminated and the following `tsdb query` will fail as can be seen in the latest failed test bot build.

UniqMartin:
> (If you give me a few minutes, I'll review the formula in regards to this and point out all the places that need to be changed. I have to admit that I haven't paid too much attention to this in previous reviews.)

Done (even if it took a few more minutes), including two additional fixes that caused the build to break at one point or another. Have tested this locally and works for me, so hopefully CI will be happy, too.

CamJN:
mavericks CI did something weird:
```
+ brew update
Formula/opentsdb.rb: needs merge
Formula/opentsdb.rb: needs merge
Formula/opentsdb.rb: unmerged (a55b82f4564fa88afc0e175b4f25de98e403b93b)
Formula/opentsdb.rb: unmerged (f9c2cfeef802f14a71a70f91ad4783fcedd6ba83)
fatal: git-write-tree: error building trees
Cannot save the current index state
To restore the stashed changes to /usr/local/Library/Taps/homebrew/homebrew-core run:
  'cd /usr/local/Library/Taps/homebrew/homebrew-core && git stash pop'
Already up-to-date.
+ brew test-bot --ci-pr
HEAD is now at 0123e04 mirror: add command to mirror to Bintray. (#263)
Already on 'master'
```

UniqMartin:
> mavericks CI did something weird:

I agree that this looks weird, but it is harmless. However, there was a different issue with that particular machine that caused the formula install and test runs to fail. I've been able to fix that and now the formula passes with ✅. Unless one of the other participants here objects, I'll be merging it later today.

Formula/opentsdb.rb
```diff
@@ -0,0 +1,149 @@
...
+      :HBASE_HOME => Formula["hbase"].opt_libexec,
+      :COMPRESSION => (build.with?("lzo") ? "LZO" : "NONE"),
+    }
+    env = Language::Java.java_home_env.merge(env)
+    create_table = pkgshare/"tools/create_table_with_env.sh"
+    create_table.write_env_script pkgshare/"tools/create_table.sh", env
+    create_table.chmod 0755
+
+    inreplace pkgshare/"etc/opentsdb/opentsdb.conf", "/usr/share", "#{HOMEBREW_PREFIX}/share"
+    etc.install Dir["#{pkgshare}/etc/opentsdb"]
```
DomT4: Small nit here, should either be:

```ruby
etc.install Dir["#{pkgshare}/etc/opentsdb/*"]
```
Or
```ruby
etc.install pkgshare/"etc/opentsdb"
```

DomT4:
It feels like `start-hbase.sh` and `stop-hbase.sh` belong inside `begin` and `ensure` blocks respectively, but apologies if I missed a discussion on that in one of the now-hidden comment threads here. Otherwise beside the nit, seems 👍.

UniqMartin:
> It feels like `start-hbase.sh` and `stop-hbase.sh` belong inside begin and ensure blocks respectively, but apologies if I missed a discussion on that in one of the now-hidden comment threads here.

Thanks @DomT4 for spotting this and @CamJN for promptly addressing it! I think the general structure for these `begin`-`ensure` blocks should be as follows:

```ruby
# Start service A.

begin
  # Use service A.
ensure
  # Stop service A.
end
```

Or to be more specific, for the case of the `test do` block of this formula:

```ruby
# Start HBase daemon (including guarding `sleep`).

begin
  # Create TSDB tables.

  # Start TSDB daemon (including guarding `sleep`).

  begin
    # Actual test code.
  ensure
    # Stop (kill) TSDB daemon.
  end
ensure
  # Stop HBase daemon.
end
```

This will also get rid of the somewhat ugly `if defined? pid && pid > 1` therein.

----

Thanks for working with us, @CamJN! I think the formula is now very close to being merged.

CamJN:
Now I admit to being more familiar with other languages' ways of doing this, but the docs sound like the same trap exists in ruby.

from the Kernel.exec docs:

> If the command fails to execute (typically Errno::ENOENT when it was not found) a SystemCallError exception is raised.

The docs don't specify the return value but if the return value is just passed from C's exec (-1 for an error) or is unset/nil then the next bit becomes a problem.

from the Process.kill docs:

> Sends the given signal to the specified process id(s), or to the current process if pid is zero. signal may be an integer signal number or a POSIX signal name (either with or without a SIG prefix). If signal is negative (or starts with a minus sign), kills process groups instead of processes.

so even if you use an extra begin/ensure block you still have to check if the pid is safe to kill, because it might have been set on a failure to exec which would give it a -1 or nil/0 value (and killing -1 or 0 is very bad).

UniqMartin:
I don't think that's an issue. `Kernel.exec` doesn't have a return value because it either fails, in which case an exception is thrown, or it succeeds, in which case the Ruby process will be replaced by the `exec`ed process. Either case will result in a valid value for `pid` as all of this is already happening in the `fork`ed Ruby process. `Kernel.fork` itself may fail (this is fairly unlikely), but this will raise an exception, thus in all circumstances `pid` is guaranteed to be a positive integer.

The only thing that could happen is that the process referenced by `pid` exits before we get to the point where `Process.kill` is called. But `kill`ing a process that no longer exists is fine, too, and the likelihood of killing an unrelated process that was started in the mean time and got the same PID is very unlikely.

Please use this (very likely final) adjustment to the formula code as an opportunity to rename `pid` to something like `tsdb_daemon_pid`. Thanks!

CamJN:
Sorry one last thing, why start the service before the begin/ensure block? Seems like that should be in the begin section?

UniqMartin:
> Sorry one last thing, why start the service before the begin/ensure block? Seems like that should be in the begin section?

No worries! The idea is that when starting the service fails, the service isn't running and it doesn't make sense to invoke the script (or any equivalent code) responsible for stopping the service. If we move the service start into the `begin` part, `ensure` will try to stop something that couldn't be started successfully.

CamJN:
Ok, I was hung up on the following pattern:
```
begin
  start db
rescue
  #don't care
else
  use db
ensure
  stop db
end
```
where you only want to try and use the service if it didn't freak out being started, and then make sure you shut it down after.

UniqMartin:
Took a while to get everything sorted out, but now I'm happy and the CI is, too. 😺 Thanks for working with us! Merged in 6d8379c0f434f4d26ea375bb1e89aded8772cb8c. Thank you for this contribution to Homebrew, @CamJN! :tada:

PR #820, opened 2016-05-03
spidermonkey 38.2.1

dholm:


DomT4:
Looks like `couchdb` needs a revision bump.

Formula/spidermonkey.rb
```diff
@@ -1,19 +1,13 @@
 class Spidermonkey < Formula
   desc "JavaScript-C Engine"
   homepage "https://developer.mozilla.org/en/SpiderMonkey"
-  url "https://archive.mozilla.org/pub/mozilla.org/js/js185-1.0.0.tar.gz"
-  version "1.8.5"
-  sha256 "5d12f7e1f5b4a99436685d97b9b7b75f094d33580227aa998c406bbae6f2a687"
-  revision 1
+  url "https://people.mozilla.org/~sstangl/mozjs-38.2.1.rc0.tar.bz2"
+  version "38.2.1"
```
DomT4: Apologies if you commented on this in an earlier thread and I missed it, but why the suddenly huge jump in version here?

Formula/spidermonkey.rb
```diff
@@ -1,19 +1,13 @@
 class Spidermonkey < Formula
   desc "JavaScript-C Engine"
   homepage "https://developer.mozilla.org/en/SpiderMonkey"
-  url "https://archive.mozilla.org/pub/mozilla.org/js/js185-1.0.0.tar.gz"
-  version "1.8.5"
-  sha256 "5d12f7e1f5b4a99436685d97b9b7b75f094d33580227aa998c406bbae6f2a687"
-  revision 1
+  url "https://people.mozilla.org/~sstangl/mozjs-38.2.1.rc0.tar.bz2"
```
DomT4: Is this a stable release?

Formula/spidermonkey.rb
```diff
@@ -22,28 +16,24 @@ class Spidermonkey < Formula
...
-      # Remove the broken *(for anyone but FF) install_name
-      inreplace "config/rules.mk",
-        "-install_name @executable_path/$(SHARED_LIBRARY) ",
-        "-install_name #{lib}/$(SHARED_LIBRARY) "
-    end
-
     mkdir "brew-build" do
+      # Configure needs python to figure out which version of libmozjs it is
+      # building.
+      ENV["PYTHON"] = which "python"
```
DomT4: FWIW this will always be the Python in `/usr/bin` on OS X.

Formula/spidermonkey.rb
dholm: Yes, they refer to it as "38.2".

Formula/spidermonkey.rb
dholm: I've submitted an update to this formula several times for (I think) a couple of years.  Usually it takes a very long time before someone reviews it and in the past it coincided with me being very busy at work and by the time I got around to replying to the comments someone had already closed the pull request due to inactivity so I had to resubmit and the cycle restarted.

Formula/spidermonkey.rb
dholm: I will change it if you think it's better?

Formula/spidermonkey.rb
DomT4: 👍. The `rc0` is a bit quirky but it's what the upstream website references, so happy enough.

Formula/spidermonkey.rb
DomT4: Two PRs in the space of a year, the first being April 2015 apparently. Was reviewed within two days initially by Mike. Second was reviewed after a day by Alex. We try and wrap things quickly but sometimes it does take a while with back and forth comments if it's a confusing or problematic change.

If you leave a comment along the lines of "I'm busy with work, I'll pick this back up in a couple of weeks" it likely won't get closed, but otherwise we've no indication that you're still interested. Sadly, most people who don't respond for more than 10 days don't at all, over any length of time 😕.

Formula/spidermonkey.rb
DomT4: I think this is fine as-is, unless @tdsmith has a stronger opinion than me.

Formula/spidermonkey.rb
dholm: I thought that I had a pull request that was even older (as in 2 years or so), which was the one I was referring to.  But maybe my memory is failing me in which case the blame is all on me!

DomT4: Couldn't [find](https://github.com/Homebrew/legacy-homebrew/pulls?utf8=%E2%9C%93&q=is%3Apr+spidermonkey+author%3Adholm+is%3Aclosed+) anything older than April 2015 with a quick check.

DomT4:
This seems fine, but `couchdb` will need the `revision` bump and will _hopefully_ build against this okay. If it does we should be able to ship.

apjanke:
I tried doing a `couchdb` revision bump in #861, and it can't find the headers.

```
checking js/jsapi.h usability... no
checking js/jsapi.h presence... no
checking for js/jsapi.h... no
configure: error: Could not find the jsapi header.

Are the Mozilla SpiderMonkey headers installed?
```

Looks like CouchDB itself might need an update to catch up with the current SpiderMonkey.

dholm:
Looks like the spidermonkey headers don't get installed properly.  Will investigate..

Issue #775, opened 2016-05-02
LBDB does not install the abook module (m_abook)


djwf:
When installing lbdb using `brew install lbdb` the file `/usr/local/Cellar/lbdb/0.41/libexec/m_abook` is not installed.  To check to see if it's a brew problem or an original source problem I downloaded the source tarball for lbdb 0.41, tried `./configure --prefix=/Users/dave/lbdb && make install`, and the abook module does install.

My solution (so far) is to change the formula to allow `configure` to use the `--with-abook` option, thus ensuring the module is installed.

However, before putting up a pull request I figured I'd see if anyone else is having the problem.  I've tried searching on a number of terms, but haven't found anyone else reporting this problem.

ilovezfs:
If you think it should be in the bottles,
```
depends_on "abook" => :recommended
```
otherwise
```
depends_on "abook" => :optional
```

No need to manually change the configure options.

djwf:
Sorry: I should have been more specific.  My current solution is to apply the attached patch.

[lbdb.txt](https://github.com/Homebrew/homebrew-core/files/245625/lbdb.txt)

The weirdness is that the current tarball does provide this module when I configure/make *without* requiring the `--with-abook` option, but something in the homebrew configure/make process is preventing the `m_abook` module from being made/installed.

ilovezfs:
@djwf no need to be any more specific than you were the first time. Before responding above, I both reproduced the behavior you described, and verified that the `depends_on` statements alone were sufficient to resolve it without further intervention in configure. All behavior described is normal, though it may be somewhat surprising if you haven't dealt with Homebrew's `superenv` before.

The reason you do *not* need to intervene with the configure is that `superenv` changes the environment to expose the dependency if and only if you declare it. When it's undeclared, even if the dependency is installed, it will be hidden from view during formula installation, and configure won't see it there.

mikemcquaid:
@djwf Your patch looks good. Could you try and open a pull request? [This document](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged) should help and I'm happy to walk you through anything else.

Thanks!

djwf:
@ilovezfs thanks for the clarification

@mikemcquaid I'll put that in tonight

ilovezfs:
Closing in favor of #826

PR #826, opened 2016-05-04
lbdb: add abook module to bottle

djwf:
See issue #775.

This makes abook a recommended dependency (thus causes it to be installed by default, along with the m_abook module, as it is when using the lbdb tarball).


ilovezfs:
Thank you for your contribution to Homebrew @djwf :tada:

PR #835, opened 2016-05-04
pastebinit: improve test

apjanke:

---

Checks the output of `pastebinit` to see if it's actually a URL, indicating the paste succeeded. Unfortunately, we can't check the results on pastebin because it blocks our `curl` fetches, probably based on user agent, and also detects these simple pastes as spammy, so you need to pass a captcha to see them.

### DIFF
--- a/Formula/pastebinit.rb
+++ b/Formula/pastebinit.rb
@@ -25,6 +25,9 @@ class Pastebinit < Formula
   end
 
   test do
-    system "date | pastebinit"
+    url = pipe_output("#{bin}/pastebinit", "Hello, world!").chomp
+    assert_match %r{^http://pastebin\.com/}, url
+    # We can't actually fetch the URL to check the paste's success because pastebin
+    # blocks our fetches with curl, probably based on the user agent
   end
 end


Formula/pastebinit.rb
```diff
@@ -25,6 +25,9 @@ def install
   end
 
   test do
-    system "date | pastebinit"
+    url = shell_output("echo \"Hello world at $(date)\" | pastebinit").chomp
+    assert_match %r{^http://pastebin\.com/}, url
+    # We can't actually fetch the URL to check the paste's success because pastebin
+    # blocks our fetches with curl, probably based on the user agent
```
ilovezfs: `curl` can set its own user agent with `-A/--user-agent` option, so that can probably be defeated without too much trouble.

Formula/pastebinit.rb
```diff
@@ -25,6 +25,9 @@ def install
   end
 
   test do
-    system "date | pastebinit"
+    url = shell_output("echo \"Hello world at $(date)\" | pastebinit").chomp
```
ilovezfs: Probably should use `#{bin}/patebinit`, no?

ilovezfs: Should this use `pipe_output` with the optional `input` argument instead of an inline `|`?

Formula/pastebinit.rb
apjanke: Yeah, but we don't want to defeat it. Our policy is to report homebrew/curl as the user agent, and respect the server side's handling of it.

Formula/pastebinit.rb
apjanke: Yes on both counts. Amended to do so. Thanks.

PR #837, opened 2016-05-04
libreadline-java: disable javadoc doclint checks

ilovezfs:
The javadocs in the current code aren't compliant with JDK 8 javadoc's
doclint, so disable that feature when it's detected.


apjanke:
:+1:

PR #839, opened 2016-05-04
devil: GCC 5 fix from NetBSD

ilovezfs:
Also, re-enable bottles.

### DIFF
--- a/Formula/devil.rb
+++ b/Formula/devil.rb
@@ -1,7 +1,7 @@
 class Devil < Formula
   desc "Cross-platform image library"
   homepage "https://sourceforge.net/projects/openil/"
-  revision 1
+  revision 2
 
   stable do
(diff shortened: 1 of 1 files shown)


Formula/devil.rb
```diff
@@ -47,26 +47,35 @@ class Devil < Formula
...
     args = %W[
       --disable-debug
       --disable-dependency-tracking
       --prefix=#{prefix}
       --enable-ILU
     ]
     args << "--enable-ILUT" if build.stable? || build.with?("ilut")
+    args << "--disable-tiff" if build.without? "libtiff"
+
+    # Fixes the test for HEAD builds without the ilut option
```
dunn: Has this also been reported?

ilovezfs:
@dunn now it has :smiley:

dunn:
🙆

apjanke:
:+1:

PR #841, opened 2016-05-04
gawk: Fix arithmetic by building against MPFR. (And readline)

kmarekspartz:

---

Before:

```
    gawk -M 'BEGIN { printf "%d\n", 9007199254740991+9007199254740992 }'
    gawk: warning: -M ignored: MPFR/GMP support not compiled in
    18014398509481984
```

After:

```
    gawk -M 'BEGIN { printf "%d\n", 9007199254740991+9007199254740992 }'
    18014398509481983
```

### DIFF
--- a/Formula/gawk.rb
+++ b/Formula/gawk.rb
@@ -4,6 +4,7 @@ class Gawk < Formula
   url "http://ftpmirror.gnu.org/gawk/gawk-4.1.3.tar.xz"
   mirror "https://ftp.gnu.org/gnu/gawk/gawk-4.1.3.tar.xz"
   sha256 "e3cf55e91e31ea2845f8338bedd91e40671fc30e4d82ea147d220e687abda625"
+  revision 1
 
   bottle do
     sha256 "ed2c5f0b20e4b4af151177a750b1287435aad66bb0c8c4b60daf753f3955332c" => :el_capitan
@@ -17,12 +18,13 @@ class Gawk < Formula
     cause "Undefined symbols when linking"
   end
 
+  depends_on "mpfr"
+  depends_on "readline"
+
   def install
     system "./configure", "--disable-debug",
                           "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
-                          "--without-readline",
-                          "--without-mpfr",
                           "--without-libsigsegv-prefix"
     system "make"
     system "make", "check"


dunn:
Should these be hard dependencies or just `:recommended`?

ilovezfs:
If the only way to get the correct arithmetic is with them, then probably hard dependencies, unless in the typical case it's not worth it since the problem is probably pretty rare?

apjanke:
`mpfr` and `readline` are both cheap dependencies, so if it's a correctness issue and not optional added functionality, they should be hard dependencies and not `:recommended`.

```
$ du -sh mpfr readline
3.6M	mpfr
2.1M	readline
```

Either way, yeah, please add a `revision` bump, since this changes the contents of the default installation. (The `revision 1` line should go after the `sha256` in the main formula, not inside the `bottle` block.)

zeckalpha:
Added revision bump and squashed.

apjanke:
LGTM

DomT4:
👍

chdiza:
`mpfr` may be cheap wrt bottle file size.  It takes a long time to build, though, doesn't it?

Also, according to `brew deps` it depends on gmp, which definitely takes a long time to build.

This seems like some heavy deps for what should be an optional feature.

DomT4:
> It takes a long time to build, though, doesn't it?

A little over a minute on a recent rMBP.

I'd be fine with these being `:recommended` I guess.

chdiza:
But one would have to build (if one is building) both `gmp` and `mpfr`, which I timed at a total of six minutes and change on my MBP.

ilovezfs:
That's not actually true unless you actually want to build everything from source, as opposed to just gawk. If you pass --build-bottle instead of --build-from-source, only gawk will be built from source. Alternatively, what is nearly equivalent, brew install --only-dependencies gawk && brew install --build-from-source gawk, will only install gawk and xz from source, though usually people already have xz installed anyway.

DomT4:
Both those bottles are `cellar :any`, so the only way anyone is going to be building them from source is if you're running 10.8 or less or deliberately invoking from-source, FWIW.

ilovezfs:
@DomT4 Is there any option other than the (undocumented) `--build-bottle` that lets you only install the named formulae from source (but everything else from bottles to the extent possible) in a single command?

mikemcquaid:
@ilovezfs There is not. There's an outstanding TODO to make `--build-from-source` behave like that (and differently to `HOMEBREW_BUILD_FROM_SOURCE`) as it's almost certainly what you actually want (i.e. just build the named stuff from source rather than all the dependencies as well).

ilovezfs:
@mikemcquaid OK. Glad I wasn't missing something blatantly obvious :) I wonder if as a stop-gap `--build-bottle` should just be documented as "almost that."

DomT4:
You can do something like:

```bash
brew install wget --only-dependencies && brew install wget -s
```
It's not "one command" though.

mikemcquaid:
@ilovezfs Unfortunately it also stops `post_install` from running.

apjanke:
I didn't know this `-s` behavior wasn't by design! I'm frequently running in to inconveniences with this behavior. Here's a PR to change it to apply only to the explicitly requested formulae, and use bottles for dependencies by default. https://github.com/Homebrew/brew/pull/205

apjanke:
Merged Homebrew/brew#205. If you do a `brew update`, `brew install --build-from-source` will now only build the specifically-requested formulae from source, and will install any dependencies from binary bottles instead of building them.

Does that change your view on the required vs `:recommended` issue?

geoff-codes:
👍
👎👎  `=> :recommended`

zeckalpha:
I have no opinion either way on this, so I'll wait for the dust to settle on what I need to do for this PR before updating.

apjanke:
I'm think the dust has settled, and we're largely agreed on merging as is, with us mostly agreed on merging as-is with it being a required dependency, since installing from source is a less common use case. So you don't need to do anything.

Unless someone hollers in the next few hours, I'm going to pull this as is. And if @chdiza or someone else runs in to a use case where they're actually installing it from source and the build time is an inconvenience, they can put in a separate PR to change the deps to `:recommended` and I'll pull that too.

mikemcquaid:
Yep, I'm good with pulling this 👍

dunn:
@BrewTestBot test this please

dunn:
Thanks, merged as 0404aa8068ebf546fb8c49463c4982646c56beb7.

Issue #801, opened 2016-05-03
libgphoto2 fails to compile


ruychi:
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

### Bug reports:

brew upgrade to libgphoto2 2.5.10 fails to compile, preventing upgrade to wine 1.8.2.

https://gist.github.com/4f62cf0f275d63919f39ebbb9cb662c4

mikemcquaid:
```
Warning: Unbrewed dylibs were found in /usr/local/lib.
If you didn't put them there on purpose they could cause problems when
building Homebrew formulae, and may need to be deleted.
```

This is likely your issue. Please remove those libraries.

ruychi:
Hi mikemcquaid,
Thanks for your note.
Although I was not suspicious of those unbrewed dylibs, I removed them properly and then updated homebrew.
However, the problem still persists and libgphoto2 2.5.10 still fails to compile, when installing/upgrading wine.
The new gist-log for libgphoto2 2.5.10:
https://gist.github.com/53b5c451a3bce40f70cef91d9c62de50

Apparently the problem comes from wine install, as libgphoto2 is a dependency. I experimented to uninstall wine and libgphoto2, then installed libgphoto2 alone. All went well.
But as soon as tried to install wine, the formula required libgphoto2 to be re-installed, and it is there when the errors occur on compiling libgphoto2.

The issued commands and the results:

**$ brew uninstall libgphoto2**
_Uninstalling /usr/local/Cellar/libgphoto2/2.5.10... (168 files, 4.3M)_
**$ brew install libgphoto2**
_==> Downloading https://homebrew.bintray.com/bottles/libgphoto2-2.5.10.el_capitan.bottle.tar.gz
Already downloaded: /Library/Caches/Homebrew/libgphoto2-2.5.10.el_capitan.bottle.tar.gz
==> Pouring libgphoto2-2.5.10.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/libgphoto2/2.5.10: 168 files, 4.3M_
**$ brew install wine**
_==> Installing dependencies for wine: libgphoto2
==> Installing wine dependency: libgphoto2
==> Downloading https://downloads.sourceforge.net/project/gphoto/libgphoto/2.5.10/libgphoto2-2.5.10.tar.bz2
[...]

PR #843, opened 2016-05-04
wine: make libgphoto2, sane-backends optional

ghost:
- [ ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)? -- wine has no tests defined

---

libgphoto2 is, according to the [Wine Source Tree Structure](https://wiki.winehq.org/Source_Tree_Structure), a driver to use TWAIN.
TWAIN is used for scanners and digital cameras.

Since I don't know what is the typical use case for people who use Wine in their Macs (I for one use games and some simple Windows apps), I've edited the formula to make libgphoto2 optional. 

Not sure if it should be the other way around (optional without), so some time should be given for people to comment.

The formula builds OK and I was able to run Wine.

EDIT: does it make sense to deploy sane-backends without libgphoto2? Added new commit to remove it.

### DIFF
--- a/Formula/wine.rb
+++ b/Formula/wine.rb
@@ -41,7 +41,7 @@ class Wine < Formula
   depends_on "little-cms2"
   depends_on "libicns"
   depends_on "libtiff"
-  depends_on "sane-backends"
+  depends_on "sane-backends" => :optional
   depends_on "gnutls"
   depends_on "libgsm" => :optional
   depends_on "samba" => :optional
@@ -122,6 +122,8 @@ class Wine < Formula
     args << "--without-x" if build.without? "x11"
     
     args << "--without-gphoto" if build.without? "libgphoto2"
+    
+    args << "--without-sane" if build.without? "sane-backends"
 
     system "./configure", *args
 
--- a/Formula/wine.rb
+++ b/Formula/wine.rb
@@ -37,7 +37,7 @@ class Wine < Formula
   depends_on "pkg-config" => :build
   depends_on "freetype"
   depends_on "jpeg"
-  depends_on "libgphoto2"
+  depends_on "libgphoto2" => :optional
   depends_on "little-cms2"
   depends_on "libicns"
   depends_on "libtiff"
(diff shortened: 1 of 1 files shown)


Formula/wine.rb
```diff
@@ -37,11 +37,11 @@ def require_universal_deps?
   depends_on "pkg-config" => :build
   depends_on "freetype"
   depends_on "jpeg"
-  depends_on "libgphoto2"
+  depends_on "libgphoto2" => :optional
```
DomT4: I'd probably make these `:recommended` for now. We try not to suddenly downgrade dependencies by default between versions in case someone's relying on the status quo.

Formula/wine.rb
```diff
@@ -120,6 +120,10 @@ def install
     args << "--without-mpg123" if Hardware.is_64_bit?
 
     args << "--without-x" if build.without? "x11"
+
```
DomT4: Looks like there's some whitespace on 123 & 125.

ilovezfs:
This looks like a solution in search of a problem to me.

apjanke:
These are lightweight dependencies – a few MB each, and they have no conflicts – and Homebrew prioritizes keeping formulae simple and minimizing options over conserving hard drive space. I think we would rather keep these as unconditional dependencies. Optional dependencies are usually for things which are very large, have dependencies which are hard to satisfy, or have conflicts with other formulae.

Marking this for rejection, but we appreciate your submission regardless.

ilovezfs:
@apjanke Hmm ... based on https://github.com/Homebrew/homebrew-core/issues/801 perhaps libgphoto2's bottles ought to be `--universal` to avoid rebuilding `libgphoto2` during every (even bottled) `wine` install.

ilovezfs:
But I guess there are actually several wine dependencies in the same camp, so probably just need to figure out the specific problem in that issue unless all of wine's dependencies' bottles are going to go universal by default.

neuralshock:
Thanks all for the constructive feedback! My suggestion was going the opposite way to the project's priorities: less dependencies, but more choices to do when deploying.
I'll take these into account the next time I send a pull request :)

dunn:
Closing this, but thank you very much for the submission! ❤️

PR #846, opened 2016-05-04
hqx: revision for devil

ilovezfs:
Also, add a test.


ilovezfs:
Now that #839 is merged, this was able to generate bottles, as expected.

However, given the status of upstream – it seems to be cryogenically frozen on the defunct Google Code platform – I'm wondering if this should be boneyarded.

apjanke:
I think we're fine to keep this until it actually breaks.

:+1:

ilovezfs:
@apjanke great

PR #865, opened 2016-05-05
juise 0.7.2

ilovezfs:


mikemcquaid:
👍

Formula/juise.rb
```diff
@@ -24,12 +24,22 @@ class Juise < Formula
   depends_on "sqlite"
 
   def install
-    system "sh ./bin/setup.sh" if build.head?
+    system "sh", "./bin/setup.sh" if build.head?
+
+    # Prevent sandbox violation
+    inreplace "configure",
+      "SLAX_EXTDIR=\"`$SLAX_CONFIG --extdir | head -1`\"",
```
mikemcquaid: Might be worth reporting upstream? What's the actual violation?

ilovezfs: I think they consider it a feature. That's a search path for extensions that `libslax` sets up for this purpose.

ilovezfs: The actual violation is
```
/Library/Developer/CommandLineTools/usr/bin/make  install-exec-hook
Install link libjuise.0.dylib -> jcs.prefix:http%3A%2F%2Fxml.juniper.net%2Fjunos%2Fcommit-scripts%2F1.0.ext ...
rm: jcs.prefix: Operation not permitted
rm: http%3A%2F%2Fxml.juniper.net%2Fjunos%2Fcommit-scripts%2F1.0.ext: Operation not permitted
make[3]: *** [install-exec-hook] Error 1
make[2]: *** [install-exec-am] Error 2
make[1]: *** [install-am] Error 2

```
which is this
```
UGLY_NAME = jcs.prefix:http%3A%2F%2Fxml.juniper.net%2Fjunos%2Fcommit-scripts%2F1.0.ext

install-exec-hook:
        @DLNAME=`sh -c '. ./libjuise.la ; echo $$dlname'`; \
                if [ x"$$DLNAME" = x ]; \
                    then DLNAME=${LIBNAME}.${SLAX_LIBEXT}; fi ; \
                echo Install link $$DLNAME "->" ${UGLY_NAME} "..." ; \
                mkdir -p ${DESTDIR}${SLAX_EXTDIR} ; \
                cd ${DESTDIR}${SLAX_EXTDIR} \
                && chmod +w . \
                && prefix=`echo ${UGLY_NAME} | awk -F: '{ print $$1 }'` \
                && url=`echo ${UGLY_NAME} | awk -F: '{ print $$2 }'` \
                && rm -f $$prefix $$url \
                && ${LN_S} ${JUISE_LIBDIR}/$$DLNAME $$url \
                && ${LN_S} $$url $$prefix

```

mikemcquaid: Ok, I see. I think it might still be worth asking upstream if we can configure that `extdir` or the call somehow?

mikemcquaid: 👍 to the PR otherwise

ilovezfs: Yeah, having upstream `juise` give us a configure option is probably the best compromise.

`libslax` itself allows the extension directory to be set during configure, but when I tried to set it to `"#{HOMEBREW_PREFIX}/lib/slax-0.20.1/extensions`, that causes `libslax` installation to trigger sandbox violations as well because it installs its own extensions there, too. So the `libslax` build would need to be changed to use that dir but only actually install its own extensions in `post_install`.

This is `libslax`'s guidance: https://github.com/Juniper/libslax/blob/4f35fba631166ade66e97c8c1880375572674d5a/doc/slax.txt#L3771-L3780
```

Extension libraries should be placed in the directory named that is
either:
- the /usr/local/lib/slax/extensions directory
- the "lib/slax/extensions" directory under the "configure --prefix"
  option given at build time
- the directory specified with the "configure --with-extension-dir"
  option given at build time
- one of the directories listed in the environment variable
  SLAX_EXTDIR (colon separated)
- one of the directories provided via the --lib/-L argument to "slaxproc"
```
So it may be somewhat inconvenient for `juise` users to have to set `SLAX_EXTDIR` or pass `--lib/-L`, but that seems preferable to altering the `libslax` build system.

ilovezfs: I guess another option would be to vendor `libslax` into `juise`'s `libexec`, which would make the problem moot. The `juise` configure already has
```
  --with-libslax-prefix=PFX           Specify location of libslax config
  --with-libslax-include-prefix=PFX   Specify location of libslax headers
  --with-libslax-libs-prefix=PFX      Specify location of libslax libs
  --with-libslax-src=DIR              For libslax thats not installed yet (sets all three above)
```
So a vendored libslax should be trivial. But I guess that's overkill?

mikemcquaid: Yeh, feels overkill. Thanks for the exploration here 👍

ilovezfs: OK, I'll ping `juise` for a configure option.

ilovezfs:
@mikemcquaid OK, new configure option requested upstream: https://github.com/Juniper/juise/issues/34

ilovezfs:
Upstream has agreed to add such an option: https://github.com/Juniper/juise/issues/34#issuecomment-217344228

mikemcquaid:
Nice 👏

PR #870, opened 2016-05-05
kjell: fix sandbox violation

ilovezfs:


apjanke:
Smells bad to have something called `CFG_DIR` pointed at `pkgshare`, but the resulting file layout looks sensible, with a `#{pkgshare}/extensions` directory. At least initially: if this is used as a config directory, then additional files are going to be dropped right in `#{pkgshare}`.

I'm not sure it's working, though. When I install and run `kjell` from this PR, it goes ahead and creates a `~/.kjell` directory where the . Looks like its code may not be properly configured to use that `CFG_DIR` throughout. Which I think makes sense: I only see `CFG_DIR` used in the `install-extensions` target, not `install` or `compile`

```
[~]
$ ls -d .k*
.kjell
[~]
$ find .kjell
.kjell
.kjell/default.profile
.kjell/extensions
.kjell/kjell.config
```

I think this means that this will fix the sandbox violation, but those extensions installed at build time will not be available when `kjell` is run.

Maybe `kjell` could be taught to have a separate extensions path? Or auto-install those extensions each time a user's config directory is created? Doing per-user operations with `make install-extensions` at build time is not very Unixy. Or to use `$HOME` instead of `~${USER}`?

DomT4:
I'd be tempted to do something like this:

```diff
diff --git a/Formula/kjell.rb b/Formula/kjell.rb
index 56afd96..f905f69 100644
--- a/Formula/kjell.rb
+++ b/Formula/kjell.rb
@@ -17,13 +17,19 @@ class Kjell < Formula
   depends_on "erlang"
 
   def install
+    inreplace "Makefile", "~${USER}/.kjell", buildpath/".kjell"
+
     system "make"
     system "make", "configure", "PREFIX=#{prefix}"
     system "make", "install", "NO_SYMLINK=1"
     system "make", "install-extensions"
+    pkgshare.install Dir[buildpath/".kjell/*"]
   end
 
   def caveats; <<-EOS.undent
+    Before initial run you may wish to do:
+      cp -r #{opt_pkgshare}/ ~/.kjell
+
     Extension 'kjell-prompt' requires a powerline patched font.
     See https://github.com/Lokaltog/powerline-fonts
     EOS
```

apjanke:
Upstream issue: https://github.com/karlll/kjell/issues/33

apjanke:
Hmm. That sounds promising. Maybe `#{pkgshare}/skel` instead of plain `pkgshare`, though.

(This could be an `#{etc}/skel` for other programs, so you could do site configuration of the startup files, but that sounds wrong here due to the presence of extensions, and the fact that kjell doesn't support this natively, so it belongs better under the per-version Cellar than a persistent `etc` area.)

DomT4:
`pkgshare` is already nested, so it isn't dropped directly into `/usr/local/share` for example, if that's your concern?

```
/usr/local/Cellar/kjell/0.2.6/bin/kjell
/usr/local/Cellar/kjell/0.2.6/CONTRIBUTORS
/usr/local/Cellar/kjell/0.2.6/ebin/ (11 files)
/usr/local/Cellar/kjell/0.2.6/EPLICENCE
/usr/local/Cellar/kjell/0.2.6/ext/extensions/ (3 files)
/usr/local/Cellar/kjell/0.2.6/ext/profiles/ (2 files)
/usr/local/Cellar/kjell/0.2.6/ext/ (6 files)
/usr/local/Cellar/kjell/0.2.6/include/kjell.hrl
/usr/local/Cellar/kjell/0.2.6/Makefile
/usr/local/Cellar/kjell/0.2.6/priv/prefix
/usr/local/Cellar/kjell/0.2.6/rebar
/usr/local/Cellar/kjell/0.2.6/rebar.config
/usr/local/Cellar/kjell/0.2.6/share/kjell/ (4 files)
/usr/local/Cellar/kjell/0.2.6/src/ (11 files)
/usr/local/Cellar/kjell/0.2.6/test/kjell_extension_SUITE_data/ (5 files)
/usr/local/Cellar/kjell/0.2.6/test/kjell_profile_SUITE_data/ (2 files)
/usr/local/Cellar/kjell/0.2.6/test/ (2 files)
```

apjanke:
No, my concern is it colliding with other stuff that `kjell` might add to its own pkgshare in the future. Even if there's no name collision, a user wouldn't want to copy those additional files in to their config directory. Since this would be a nonstandard part of the `kjell` installation, seems cleaner to wall it off in its own directory. And `skel` is a conventional name for this kind of "skeleton" config file, so it may make the intent clearer.

DomT4:
Ah, okay, my apologies for misunderstanding. Using `(pkgshare/"blah")` instead of `pgkshare` directly seems fine to me then.

apjanke:
No worries. Though I suppose it's just as unlikely that upstream would pick `.kjell` for a new directory as it would pick `skel`, so it probably doesn't matter. But it might also be nice to avoid using a hidden file in the system installation – could just call it `skel` instead of `.kjell`.

ilovezfs:
I think the issue is that the string needs to be replaced here too:
https://github.com/karlll/kjell/blob/0848ad2d2ddefc74774f0d793f4aebd260efb052/src/start_kjell.erl#L15-L20

Then
```
inreplace "Makefile", "CFG_DIR=~${USER}/.kjell", "CFG_DIR=#{pkgshare}"
```
should work as intended.

apjanke:
I d t think that'll work either, because then the user-specific config files which do make sense to go n `$HOME/.kjell` would get written to `pkgshare` at run time, which isn't what we want either. This looks like it needs a deeper upstream change.

The thing to do in Homebrew might be to just drop the `install-extensions` step.  We can't make it quite work right. And because the older bottles wouldn't have picked up the `.kjell` directory from pre-sandbox builds, we wouldn't be getting a regression for most users.

DomT4:
It's perhaps worth noting that the only thing `install-extensions` does is try to create that directory and then move the three extension files there. We could drop that `make` step, stick the files in `pkgshare/xyz` ourselves and tell people if they want them to move them there.

ilovezfs:
@apjanke if you look at the source, extensions can be separate from the others and can be an absolute path.

mikemcquaid:
👍 to merging either way.

Formula/kjell.rb
```diff
@@ -17,6 +17,11 @@ class Kjell < Formula
   depends_on "erlang"
 
   def install
+    # mkdir: $HOME/.kjell: Operation not permitted
+    # Prevent sandbox violation
+    # Reported 5th May 2016: https://github.com/karlll/kjell/issues/33
+    inreplace "Makefile", "CFG_DIR=~${USER}/.kjell", "CFG_DIR=#{pkgshare}"
```
MikeMcQuaid: Could we make this just `${HOME}/.kjell`? That'll work in the sandbox and after installation by using different directories.

ilovezfs:
I'll leave this one to someone else's judgment.

PR #871, opened 2016-05-05
libcanberra: depend on libtool

ilovezfs:
libcanberra: depend on libtool
- Fix "configure: error: Unable to find libltdl"
- Remove patch for the "--as-needed" and "--gc-sections" LDFLAGS bug and
  use `inreplace` on "config.status" instead, which also works for HEAD
- Add Autotools deps for HEAD builds
- Require `gtk-doc` for HEAD builds to prevent failure during configure
- Add a test, adapted from http://www.gtkforums.com/viewtopic.php?t=6753

### DIFF
--- a/Formula/libcanberra.rb
+++ b/Formula/libcanberra.rb
@@ -1,27 +1,53 @@
 class Libcanberra < Formula
   desc "Implementation of XDG Sound Theme and Name Specifications"
   homepage "http://0pointer.de/lennart/projects/libcanberra/"
-  url "http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz"
-  sha256 "c2b671e67e0c288a69fc33dc1b6f1b534d07882c2aceed37004bf48c601afa72"
 
-  head "git://git.0pointer.de/libcanberra"
+  stable do
+    url "http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz"
+    sha256 "c2b671e67e0c288a69fc33dc1b6f1b534d07882c2aceed37004bf48c601afa72"
+
+    depends_on "gtk-doc" => :optional
+  end
+
+  head do
+    url "git://git.0pointer.de/libcanberra"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "gtk-doc"
+  end
 
   depends_on "pkg-config" => :build
+  depends_on "libtool" => :run
   depends_on "libvorbis"
   depends_on "pulseaudio" => :optional
   depends_on "gstreamer" => :optional
   depends_on "gtk+" => :optional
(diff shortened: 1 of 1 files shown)


DomT4:
👍. Would be nice to add a test here, if you can think of one at all.

mikemcquaid:
👍 and 👍 on a basic test, if possible.

ilovezfs:
@mikemcquaid @DomT4 I wrote a test if you'd like to look it over.

Formula/libcanberra.rb
```diff
@@ -24,4 +25,30 @@ def install
     system "./configure", "--prefix=#{prefix}"
     system "make", "install"
   end
+
+  test do
+    (testpath/"test.c").write <<-'EOS'.undent
```
DomT4: Did this not work with your standard `<<-EOS.undent`?

DomT4:
It looks like your test may well have done its job and revealed the `libtool` dependency isn't a pure build-time one:
```
==> /usr/bin/clang test.c -I/usr/local/Cellar/libcanberra/0.30/include -L/usr/local/Cellar/libcanberra/0.30/lib -lcanberra -o test
==> ./test
dyld: Library not loaded: /usr/local/opt/libtool/lib/libltdl.7.dylib
  Referenced from: /usr/local/opt/libcanberra/lib/libcanberra.0.dylib
  Reason: image not found
Error: libcanberra: failed
<0> expected but was
<nil>.
```

ilovezfs:
nothing is interpolated so I was avoiding "\\n" but final version ended up with none anyway, so makes no difference

ilovezfs:
@DomT4 fixed

Formula/libcanberra.rb
```diff
@@ -7,6 +7,7 @@ class Libcanberra < Formula
   head "git://git.0pointer.de/libcanberra"
 
   depends_on "pkg-config" => :build
+  depends_on "libtool"
```
DomT4: This will raise an audit failure unless you tag it `=> :run`. Should have pointed that out in my initial comment, apologies.

ilovezfs:
@DomT4 @mikemcquaid I refreshed the PR with the following changes:
- added `:run` to fix the libtool audit failure
- simplified the test
- reported the preexisting `LDFLAGS` bug upstream
- removed the old patch for the `--as-needed` and `--gc-sections` `LDFLAGS` bug, and substituted a simpler `inreplace` on "config.status," which has the benefit of working with `HEAD`, too, rather than requiring an equivalent, but separate, patch
- fixed the `HEAD` build, which had undeclared dependencies on Autotools and `gtk-doc`

ilovezfs:
@DomT4 @mikemcquaid :ship: ?

mikemcquaid:
👍

DomT4:
:shipit:

PR #875, opened 2016-05-05
libmarisa: new upstream repo and another clang fix

ilovezfs:
Using clang leads to this failure during `make check`:
  TestKey(): 115: Assertion `r_key.ptr() == NULL' failed.

### DIFF
--- a/Formula/libmarisa.rb
+++ b/Formula/libmarisa.rb
@@ -1,8 +1,24 @@
 class Libmarisa < Formula
   desc "Static and space-efficient trie data structure"
-  homepage "https://code.google.com/p/marisa-trie/"
-  url "https://marisa-trie.googlecode.com/files/marisa-0.2.4.tar.gz"
-  sha256 "67a7a4f70d3cc7b0a85eb08f10bc3eaf6763419f0c031f278c1f919121729fb3"
+  homepage "https://github.com/s-yata/marisa-trie"
+  revision 1
+
+  stable do
+    url "https://marisa-trie.googlecode.com/files/marisa-0.2.4.tar.gz"
+    sha256 "67a7a4f70d3cc7b0a85eb08f10bc3eaf6763419f0c031f278c1f919121729fb3"
+
+    # trie-test.cc:71: TestKey(): 115: Assertion `r_key.ptr() == NULL' failed.
+    # Both upstream patches are needed since one unwinds the other.
+    patch do
+      url "https://github.com/s-yata/marisa-trie/commit/80f812304bcf6d2ca2f7d614cbb7b5fb07ac44f5.patch"
+      sha256 "e7882c93b470c1a079ee22805108f976b2a460759ea7656cf5264a793427cc8c"
+    end
+
+    patch do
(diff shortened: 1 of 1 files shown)


ilovezfs:
Upstream appears dead: https://code.google.com/p/marisa-trie

There doesn't seem to be anyone to report this to.

DomT4:
Possibly a good 💀 candidate. No apparent upstream so likely to never be fixed, not flagged by `brew uses` at all and less than 100 downloads in the last year.

ilovezfs:
To be fair, there's no El Capitan bottle, so the download stat may not mean too much unless extrapolated based on the period between May and October 2015.

Nonetheless, grandfathering formulae that would not pass review if resubmitted today doesn't make much sense to me even when gcc isn't needed.

mikemcquaid:
Agreed to boneyard if it doesn't build with Clang 👍

ilovezfs:
@mikemcquaid @DomT4 OK, it's back to working with clang. I'll see if I can find some way to report the fix upstream.

ilovezfs:
Ah ha! Found upstream, and the new site for this software ... and this bug already reported and already fixed: https://github.com/s-yata/marisa-trie/pull/7

DomT4:
Same upstream maintainer or?

ilovezfs:
Yep, same guy.

ilovezfs:
@DomT4 so it turned out there way yet-another-bug of the same kind, which I patched using his method and opened a PR https://github.com/s-yata/marisa-trie/pull/9

The patch from that PR is used in the formula.

Formula/libmarisa.rb
```diff
@@ -1,8 +1,24 @@
 class Libmarisa < Formula
   desc "Static and space-efficient trie data structure"
-  homepage "https://code.google.com/p/marisa-trie/"
-  url "https://marisa-trie.googlecode.com/files/marisa-0.2.4.tar.gz"
-  sha256 "67a7a4f70d3cc7b0a85eb08f10bc3eaf6763419f0c031f278c1f919121729fb3"
+  homepage "https://github.com/s-yata/marisa-trie"
+  revision 1
+
+  stable do
+    url "https://marisa-trie.googlecode.com/files/marisa-0.2.4.tar.gz"
```
DomT4: I wonder if it's worthwhile asking upstream for a new tag now we've discovered they're still alive?

ilovezfs: Yes, definitely, but I wanted to hold off on that request until after s-yata/marisa-trie#9 is merged.

ilovezfs: Also, it might be a good idea to (re-)define 0.2.4 now in terms of the equivalent GitHub repo commit to get rid of the Google Code tarball altogether immediately, but I figured we might as well wait for s-yata/marisa-trie#9 to be merged, request a new tag, and then use that.

ilovezfs:
@DomT4 :ship:?

mikemcquaid:
👍

DomT4:
LGTM. I'd have just boneyarded it, I admire your persistence and patience in getting this workable again! 😃.

ilovezfs:
@DomT4 Thanks :)

PR #883, opened 2016-05-05
ee: Add test.

woodruffw:
~~**Important**: Please don't merge this yet!~~

~~I'm going to be adding a few tests to formulae that I noticed were missing them in this branch.
I'm opening this PR to get feedback about the quality/rigour of the tests (if individual PRs are preferred for each test, I'll split them up for review).~~

This PR is just for `ee` now.

cc @UniqMartin

### DIFF
--- a/Formula/ee.rb
+++ b/Formula/ee.rb
@@ -19,4 +19,10 @@ class Ee < Formula
     bin.install "ee"
     man1.install "ee.1"
   end
+
+  test do
+    ENV["TERM"] = "xterm"
+    # escape + a + b is the exit sequence for `ee`
+    pipe_output("#{bin}/ee", "\\033[ab", 0)
+  end
 end


UniqMartin:
> I'm opening this PR to get feedback about the quality/rigour of the tests

This highly depends on the formula and what can be tested reasonably. Specifically for `ee` I think that's about as good as a test for this formula can be. :+1:

> if individual PRs are preferred for each test, I'll split them up for review

Yes, please one PR per formula, as it's much easier to keep track of that. Unless your batching together a bunch of really trivial tests, but I hope there won't be too many of these. In any case it's easier to keep track of things if they are in individual PRs and they can also be pulled independently, even if some debate arises around a particular formula.

As for the commit message format: You're close, but we prefer them to be even more terse. That means (a) omit the `.rb` extension, i.e. `<formula>` instead of `<formula>.rb` and (b) the “missing” in “Add missing test.” is somewhat redundant. This second advice is rather nitpicky and you'd also be fine if you left this part as is. To summarize: Commit subject here should preferably be “ee: add test” (my personal favorite) or “ee: Add test.” (seems to be your favorite and other maintainers use this style too).

woodruffw:
> Yes, please one PR per formula, as it's much easier to keep track of that. Unless your batching together a bunch of really trivial tests, but I hope there won't be too many of these. In any case it's easier to keep track of things if they are in individual PRs and they can also be pulled independently, even if some debate arises around a particular formula.

Sure thing. This PR will just be for `ee`, then.

I'll amend the commit message to your specifications and open new PRs for other tests :smile:

UniqMartin:
Merged in 3b44daadb4c3bef8d67497d8109f4aa8f7ef2d1a. Thank you for this contribution to Homebrew, @woodruffw! :tada:

> I'll amend the commit message to your specifications and open new PRs for other tests :smile:

I re-amended the commit message to strip the `.rb`, too. :wink: Thanks; that's a good start! :+1:

woodruffw:
Oops, missed that :disappointed_relieved:

UniqMartin:
> Oops, missed that :disappointed_relieved:

Don't worry! It's not super important (the actual test is more important), it wouldn't have made the commit harder to find, and I noticed and adjusted, so all is good. 😺

mikemcquaid:
> This highly depends on the formula and what can be tested reasonably. Specifically for ee I think that's about as good as a test for this formula can be.

Personally I'd favour adding `--help` or `--version` tests to 10 formulae rather than a great test to a single one. Long term it'd be nice to have great tests on them all but I'd love to see some test on every formula first 👍

UniqMartin:
> Personally I'd favour adding `--help` or `--version` tests to 10 formulae rather than a great test to a single one. Long term it'd be nice to have great tests on them all but I'd love to see some test on every formula first 👍

My thinking is that once there's already a test in place (no matter how trivial it is), there's less incentive to improve it. It will no longer be flagged as missing and a human needs to check whether it's a minimal test or a good (or even great) test.

Additionally, most `--help` or `--version` tests only check whether the program starts at all, something that in the past helped to detect broken linkage for compiled programs, but little beyond that. Since Andrew added `brew linkage --test` to our `test-bot` we have a more comprehensive check for linkage issues. (Of course it doesn't cover formulae built on top of interpreted languages, Java, etc.)

woodruffw:
This is just an idea, but maybe an optional parameter like `test :trivial do` for tests that only check `--help`/`--version`? That way we could test just about every (executable-installing formula) trivially and still have an easy-to-find tag for more in-depth testing.

UniqMartin:
@woodruffw I think that's a good idea. And it would be just a matter of adding an optional argument to the [`Formula.test` method](https://github.com/Homebrew/brew/blob/d887dd39ec60c70e76a696e5fc9309172962d3c8/Library/Homebrew/formula.rb#L2048-L2050) that is responsible for storing the code between `do` and `end` for later use. The simplest possible implementation could ignore this argument, but those trivial tests would still be easily greppable. Or we could expose this as something that can be checked programmatically, e.g. via `Formula#test_tags`, `Formula#test_trivial?`, or something similar.

What do you think about transforming your idea into code and making that a [Homebrew/brew](https://github.com/Homebrew/brew) PR? The easiest possible implementation should be sufficient to get the discussion started and to gauge how other maintainers feel about this idea.

woodruffw:
> What do you think about transforming your idea into code and making that a Homebrew/brew PR? The easiest possible implementation should be sufficient to get the discussion started and to gauge how other maintainers feel about this idea.

Sounds good to me! I'll have someone up by tonight  :+1:

PR #885, opened 2016-05-06
bsdiff: Add test.

woodruffw:
Test specifics: generates a patchfile from two (tiny) binary files.

This should be sufficient to ensure `bsdiff` actually executes, but it may be desirable to add `bspatch` reconstitution and a quick file comparison. Since `bspatch` isn't installed by the formula (it's distributed with OS X in `/usr/bin`), I skipped doing that unless someone wants me to.

cc @UniqMartin


mikemcquaid:
Thanks @woodruffw!

PR #887, opened 2016-05-06
duff: Add test.

woodruffw:
Test specifics: compares two equal files and asserts expected output.

cc @UniqMartin

### DIFF
--- a/Formula/duff.rb
+++ b/Formula/duff.rb
@@ -17,4 +17,17 @@ class Duff < Formula
                           "--mandir=#{man}"
     system "make", "install"
   end
+
+  test do
+    expected = <<-EOS.undent
+      2 files in cluster 1 (6 bytes, digest 8843d7f92416211de9ebb963ff4ce28125932878)
+      cmp1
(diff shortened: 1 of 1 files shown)


Formula/duff.rb
```diff
@@ -17,4 +17,17 @@ def install
...
+    expected = <<-EOS.undent
+      2 files in cluster 1 (6 bytes, digest 8843d7f92416211de9ebb963ff4ce28125932878)
+      cmp1
+      cmp2
+    EOS
+
+    (testpath/"cmp1").write "foobar"
+    (testpath/"cmp2").write "foobar"
+
+    assert_equal expected, `"#{bin}/duff" cmp1 cmp2`
```
DomT4: We shouldn't be using backtics in tests directly. You want to use `shell_output` or `pipe_output` depending on circumstance.

woodruffw: Fixed it, thanks for the tip!

mikemcquaid:
Thanks @woodruffw!

PR #896, opened 2016-05-06
qfs 1.1.4 (new formula)

fsareshwala:

---

### DIFF
--- a/Formula/qfs.rb
+++ b/Formula/qfs.rb
@@ -1,0 +1,38 @@
+class Qfs < Formula
+  desc "High-performance, fault-tolerant, distributed file system and its tools"
+  homepage "https://github.com/quantcast/qfs"
+  url "https://github.com/quantcast/qfs/archive/c7f6950df21780bb02222c311fc797fa37a3b66d.tar.gz"
+  version "1.1.4-1"
+  sha256 "fbced6b93ef4d74bc1deb66e89f67b3c2f8f6de902276ad1d443881343d9ac20"
+
+  depends_on "boost"
+  depends_on "cmake" => :build
+  depends_on "maven" => :build
+  depends_on "openssl" => :build
+  depends_on "wget"
+
+  def install
+    ENV.java_cache
+
+    system "make", "build", "hadoop-jars", "BUILD_TYPE=release"
+    rm_rf "build/release/bin/examples"
+    rm_rf "build/release/bin/tests"
+
+    bin.install Dir["build/release/bin/tools/*"]
+    bin.install Dir["build/release/bin/emulator/*"]
+    bin.install Dir["build/release/bin/devtools/*"]
+    bin.install Dir["build/release/bin/*"]
+
+    libexec.install Dir["build/java/hadoop-qfs/*.jar"]
+    libexec.install "build/java/qfs-access/qfs-access-master.jar"
+
+    lib.install Dir["build/release/lib/*.dylib"]
+    lib.install Dir["build/release/lib/static/*.a"]
+
+    include.install Dir["build/release/include/kfs/*"]
+  end
+
+  test do
+    system "#{bin}/qfs", "-ls", "/"
+  end
+end


fsareshwala:
It looks like the build failed due to the following error:

```
+ exec mvn -Dqfs.release.version=master -Dqfs.source.revision=753a76706afa3ab22fd38cc26a7f5fe898a2f8eb -Dtest.build.data=/tmp --projects qfs-access package
[ERROR] Could not create local repository at /Users/brew/.m2/repository -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/LocalRepositoryNotAccessibleException
make: *** [java] Error 1
```

Does using `mvn` in homebrew require me to do something special?

mikemcquaid:
> Does using mvn in homebrew require me to do something special?

You may need to add `ENV.java_cache` to the top of the `install` method.

fsareshwala:
@mikemcquaid Thank you! That seems to have done the trick. I've updated the patch and it should hopefully be good now to merge.

Formula/qfs.rb
```diff
@@ -0,0 +1,101 @@
...
+  depends_on "boost"
+  depends_on "cmake" => :build
+  depends_on "maven" => :build
+  depends_on "openssl" => :build
+  depends_on "wget"
+
+  def install
+    ENV.java_cache
+
+    opoo "Building qfs. This may take some time..."
```
mikemcquaid: Please remove this line, thanks.

Formula/qfs.rb
```diff
@@ -0,0 +1,101 @@
...
+    system "make", "build", "hadoop-jars", "BUILD_TYPE=release"
+
+    bin.install "build/release/bin/chunkscrubber"
```
mikemcquaid: Is there no `make install` we can use? If not, can you go `prefix.install "build/release/bin"`?

Formula/qfs.rb
fsareshwala: Removed.

Formula/qfs.rb
fsareshwala: The `make install` in qfs unfortunately installs to the `build` directory and not the system directories. That's a decision made upstream unfortunately.

Wouldn't `prefix.install "build/release/bin"` install to the `/usr/local/Cellar/qfs/1.1.4-1` directory instead of the `/usr/local/Cellar/qfs/1.1.4-1/bin` like we want? I just want to make sure I understand what you're asking. Would `bin.install Dir["build/release/bin/*"]` be a better solution?

mikemcquaid: It'll install the `bin` directory into the prefix so it should be in the right place. I think the glob would be fine too; you could maybe just delete the stuff you don't want before installation.

I think this is worth filing an issue upstream about, though, this is pretty weird behaviour for a `make` based buildsystem.

fsareshwala: Got it. I've updated the code to do that now. Let me know what you think.

I agree on the `make install` point. I work at Quantcast and have had this discussion internally and I'm not sure if we are going to change it right now. We might in the future.

Formula/qfs.rb
```diff
@@ -0,0 +1,37 @@
+class Qfs < Formula
+  desc "High-performance, fault-tolerant, distributed file system and its tools"
+  homepage "https://github.com/quantcast/qfs"
+  url "https://github.com/quantcast/qfs.git"
+  version "1.1.4-1"
+  head "https://github.com/quantcast/qfs.git"
+
+  bottle :unneeded
```
mikemcquaid: The bottle is needed unfortunately as this uses `make`.

Formula/qfs.rb
mikemcquaid: > I work at Quantcast and have had this discussion internally and I'm not sure if we are going to change it right now. We might in the future.

It'd be great if there was at least an upstream tracking issue for this. A `make install` is pretty standard practice to make life easier for packagers.

Formula/qfs.rb
fsareshwala: Ah, okay. I've removed it. Thanks!

Formula/qfs.rb
fsareshwala: Agreed. I've created https://quantcast.atlassian.net/browse/QFS-313 for that.

Formula/qfs.rb
```diff
@@ -0,0 +1,37 @@
...
+  depends_on "cmake" => :build
+  depends_on "maven" => :build
+  depends_on "openssl" => :build
+  depends_on "wget"
+
+  def install
+    ENV.java_cache
+
+    system "make", "build", "hadoop-jars", "BUILD_TYPE=release"
+    system "rm", "-rf", "build/release/bin/examples", "build/release/bin/tests"
```
mikemcquaid: Can just use the `rm_rf` method instead of `system` here.

Formula/qfs.rb
```diff
@@ -0,0 +1,37 @@
+class Qfs < Formula
+  desc "High-performance, fault-tolerant, distributed file system and its tools"
+  homepage "https://github.com/quantcast/qfs"
+  url "https://github.com/quantcast/qfs.git"
+  version "1.1.4-1"
```
mikemcquaid: This needs to be a tarball rather than a Git checkout e.g. https://github.com/quantcast/qfs/archive/1.1.4.tar.gz If we're waiting for another release we can either apply patches that have been applied upstream or just wait for a new release.

Formula/qfs.rb
fsareshwala: Done.

Formula/qfs.rb
fsareshwala: Done. We can wait for the new release, but I've used the hash for the latest master which should be a good stopgap for now until the next version is released.

fsareshwala:
I've updated the code to reflect all comments and made sure that the audit works. Let me know if you guys want me to change anything else. Thanks!

MikeMcQuaid:
@fsareshwala We'll wait on a new release but other than that 👍 thanks!

fsareshwala:
@MikeMcQuaid just to make sure, what are we waiting for for the new release? Is there anything specific that we need?

MikeMcQuaid:
A version number, basically. We can't say `master` is a version that upstream doesn't, unfortunately.

fsareshwala:
@MikeMcQuaid Ah got it. I'm not sure when the next version of qfs will be released so we might be waiting for some time.

MikeMcQuaid:
@fsareshwala That's fine, thanks

MikeMcQuaid:
Closing but will consider including it when a new version is released.

fsareshwala:
@MikeMcQuaid QFS 1.2.1 is now available and it looks like some other repositories are looking to bind against it through homebrew. Can we reopen this pull request or should I make another one with the change?

MikeMcQuaid:
@fsareshwala I can't reopen this but yes: please create a new PR. Thanks!

PR #907, opened 2016-05-06
czmqpp 1.2.0 (new formula)

alexreg:


mikemcquaid:
Various audit failures (although the notability one I think we can ignore): http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/1439/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_czmqpp___strict___online/

alexreg:
@mikemcquaid Oops, sorry about that. I managed to get rid of the 2nd and 3rd audit failures. Adding a test-case for the library is non-trivial unfortunately, as I can't find a pre-existing one, and zeromq is by its nature tricky to test insolation (certainly I wouldn't know how to do this – I just use this library as a dependency of a dependency).

Good to know we can ignore the one about notability, thanks.

mikemcquaid:
> Adding a test-case for the library is non-trivial unfortunately, as I can't find a pre-existing one, and zeromq is by its nature tricky to test insolation (certainly I wouldn't know how to do this – I just use this library as a dependency of a dependency).

We need some sort of test case unfortunately; even if it's just verifiying it can link against the library and import a header.

UniqMartin:
> Adding a test-case for the library is non-trivial unfortunately, as I can't find a pre-existing one, and zeromq is by its nature tricky to test insolation (certainly I wouldn't know how to do this – I just use this library as a dependency of a dependency).

It's actually not that difficult to test in isolation if you use an in-process channel. I just added a test to `czmq` (see #993). Maybe you can make a C++ version of that test built on top of `czmqpp` instead?

alexreg:
@UniqMartin Thanks. I'm having trouble porting it, due to the lack of documentation in czmqpp (it's a dependency of a dependency) and my poor C++, but at least it's a good place to start.

alexreg:
Added test case, which succeeds on my machine. Hope all is in order now.

Formula/czmqpp.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  head "https://github.com/zeromq/czmqpp.git"
+
+  option :universal
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "czmq"
```
UniqMartin: The test nicely shows that C++11 is a requirement for this library. Please add `needs :cxx11` after this line, separated from this `depends_on` by an empty line.

DomT4: You may also want `ENV.cxx11` in `def install`, IIRC.

Formula/czmqpp.rb
```diff
@@ -0,0 +1,81 @@
...
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <iostream>
+      #include <string>
+      #include <vector>
```
UniqMartin: I think you can drop this include. `std::vector` isn't directly used in your test code and that it's used by `czmqpp` is an implementation detail.

Formula/czmqpp.rb
```diff
@@ -0,0 +1,81 @@
...
+        czmqpp::socket push_sock(context, ZMQ_PUSH);
+        push_sock.connect(addr);
+
+        {
+          czmqpp::message send_msg;
+          const czmqpp::data_chunk msg_data(msg.begin(), msg.end());
+          send_msg.append(msg_data);
+          if (!send_msg.send(push_sock))
+            return 1;
+        }
```
UniqMartin: If I'm interpreting things correctly, you added the braces to limit the scope of the variables in this block. That's a noble goal, but probably overkill in a simple test like this one. I'd suggest to remove them here and in the following block to make the code a bit shorter. To avoid having two `msg_data` you could rename them to `send_data` and `recv_data`.

Formula/czmqpp.rb
```diff
@@ -0,0 +1,81 @@
...
+        return 0;
+      }
+    EOS
+
+    flags = ENV.cxxflags.to_s.split + %W[
+      -std=c++11
+      -I#{include}
+      -L#{lib}
+      -lczmq++
+      -lczmq
```
UniqMartin: Please add `-L#{Formula["czmq"].opt_lib}` above this line so that `libczmq.dylib` is also found on systems where Homebrew is not installed in `/usr/local` (`/usr/local/lib` is a default linker search path, but e.g. `~/.homebrew/lib` isn't).

Formula/czmqpp.rb
UniqMartin: It's not strictly necessary to add `ENV.cxx11` because the `configure` script of this library takes care of this and figures out the correct compiler flags, but it doesn't seem to hurt adding it.

UniqMartin:
> Added test case, which succeeds on my machine. Hope all is in order now.

I added a few more comments I'd like to see addressed, but overall the test looks fantastic and is pretty much what I was hoping for when I linked you to the `czmq` test. Good work! :+1:

Formula/czmqpp.rb
alexreg: czmqpp::data_chunk is just a typedef for `std::vector<...>` (which I initially wrote instead of the typedef), so I thought it wouldn't hurt... but I'll remove it if you still prefer.

Formula/czmqpp.rb
alexreg: I actually tried using `ENV.cxx11` before before, but it seems to fail on this. I'm not sure it's even valid to use in this context.

```
==> Using the sandbox
==> ["CXX"] -o test test.cpp -Os -w -pipe -march=native -mmacosx-version-min=10.11 -std=c++11 -I/usr
Error: czmqpp: failed
no implicit conversion of Array into String
```

alexreg:
@UniqMartin Thanks for the feedback. I've incorporated all of your changes but one minor one (about including `vector` – see comment).

Incidentally, just to confirm: if `system ENV.cxx ...` fails, the test will early-exit, right? So it won't try to run the test subsequently?

Formula/czmqpp.rb
UniqMartin: It should work just fine in `def install`. In `test do` the failure is caused by `ENV.cxx` no longer being an just an executable, but an executable with additional arguments. Thus it needs a treatment similar to `ENV.cxxflags`. Something along the lines of:

```ruby
args = ENV.cxx.split + ENV.cxxflags.to_s.split + %W[
  -o test test.cpp
  -I#{include} -L#{lib} -lczmq++
  -L#{Formula["czmq"].opt_lib} -lczmq
]
system(*args)
```

That's not exactly beautiful, but I don't think we have a more elegant solution.

Formula/czmqpp.rb
UniqMartin: Yes, please remove it. Its presence doesn't break anything, but it's also not needed. Since `czmqpp::data_chunk` is defined in one of the `czmqpp` headers, that's the one that uses `std::vector` and should be including `vector` (as it does).

Formula/czmqpp.rb
alexreg: You mean I should add something to the `install` block too – how would that work?

UniqMartin:
> Incidentally, just to confirm: if `system ENV.cxx ...` fails, the test will early-exit, right? So it won't try to run the test subsequently?

Correct. The assumption is that running subsequent statements after a failed one probably doesn't make sense (as is the case in this particular example, where running `./test` will be impossible if the compile fails). Thus the test aborts on first failure. – But see my code comment for why it failed and how to fix it.

Formula/czmqpp.rb
alexreg: Also, I don't see `ENV.cxx11` in the above code.

alexreg: Okay, I'm just confused now... all sorts of weird things happening.

UniqMartin: > You mean I should add something to the `install` block too – how would that work?

It could start off something like this (but as I explained, it is not actually necessary because the `configure` script will do the right thing without it):

```ruby
def install
  ENV.cxx11
  ENV.universal_binary if build.universal?

  # Rest of the code …
```

> Also, I don't see `ENV.cxx11` in the above code.

Sorry, my bad. I just assumed it would be clear that it's sufficient to stick this line somewhere in `test do` *before* invoking the compiler, e.g., like this (that's incomplete test code):

```ruby
ENV.cxx11
args = ENV.cxx.split + ENV.cxxflags.to_s.split + %W[
  -o test test.cpp
  -I#{include} -L#{lib} -lczmq++
  -L#{Formula["czmq"].opt_lib} -lczmq
]
system(*args)
```

UniqMartin: > Okay, I'm just confused now... all sorts of weird things happening.

Sorry for confusing you! Where did I lose you?

alexreg: > Sorry for confusing you! Where did I lose you?

Oh, well I was getting all sorts of weird output trying your first example (but I hadn't inserted the `ENV.cxx11` in the appropriate places). Your latest post clarified that, so all is good now, thanks.

alexreg:
Okay, all your suggestions are integrated. Read to merge, I hope!

UniqMartin:
> Okay, all your suggestions are integrated. Read to merge, I hope!

Sorry, only almost, but that's completely my fault. I wrote the following before, but apparently forgot to hit the “Comment” button. After this has been addressed we're ready to go.

Please replace

```ruby
url "https://github.com/zeromq/czmqpp.git", :tag => "v1.2.0", :revision => "a3c38eb6e3018887a8ecabc77f3029f7edb7b2d0"
```

with

```ruby
url "https://github.com/zeromq/czmqpp/archive/v1.2.0.tar.gz"
sha256 "4ed983c3cfa7c5b0f035c2868357887f5663a7fce75c55da4b0dc47f37d83e2a"
```

The latter is more friendly towards users that want (or have to) build from source and it also allows us to verify the integrity of the source tarball.

alexreg:
Okay, that's done too now. (I assume @mikemcquaid prefers the tarball over Git with tag and revision too?)

mikemcquaid:
> (I assume @mikemcquaid prefers the tarball over Git with tag and revision too?)

Yep, thanks.

mikemcquaid:
Thanks @alexreg!

alexreg:
No problem. This was a good learning experience.

PR #912, opened 2016-05-06
libao: Add HEAD and tests

rdj:

---

Adds head URL for libao, which is of interest recently because they've finally moved off of a deprecated osx API (Carbon Component Manager). Also adds a test which builds a simple test.cpp that initializes the ao library, based on the tinyxml2 test that is recommended in the formula contribution doc.

### DIFF
--- a/Formula/libao.rb
+++ b/Formula/libao.rb
@@ -12,12 +12,35 @@ class Libao < Formula
     sha256 "21aa15e92c5577a4a610de8fbb3f5a72638a0c37a40c4ebebc14826359932efa" => :mountain_lion
   end
 
+  head do
+    url "https://git.xiph.org/libao.git"
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+    depends_on "libtool" => :build
+  end
+
   depends_on "pkg-config" => :build
 
   def install
+    if build.head?
+      ENV["AUTOMAKE_FLAGS"] = "--include-deps"
+      system "./autogen.sh"
+    end
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
                           "--enable-static"
     system "make", "install"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <ao/ao.h>
+      int main() {
+        ao_initialize();
+        return 0;
+      }
+    EOS
+    system ENV.cc, "test.cpp", "-L#{lib}", "-lao", "-o", "test"
+    system "./test"
   end
 end


rdj:
The test failure definitely doesn't look related to my change: 

```
Error Message

failed: brew linkage --test vorbis-tools

Stacktrace

        Missing libraries:
  /usr/lib/libnetwork.dylib
```

Formula/libao.rb
```diff
@@ -13,11 +14,30 @@ class Libao < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "automake" => :build
```
DomT4: Presuming these are only required for the `head` you need to put them in a block so not everyone has to install them to build the formula:

```ruby
head do
  url "https://git.xiph.org/libao.git"
end
```

rdj:
Hey, thanks, I was trying to figure out how to do that, but I really had very little luck finding documentation for the Formula DSL and couldn't find an example of head-specific dependencies.

rdj:
Rebased on top of d09b135b941578df836e5a3697230edd7f3dff17 but still getting that libvorbix linkage failure.

DomT4:
Seems fine. @apjanke If you get a chance if you've any thoughts on why `vorbis-tools` declares missing linkage to a system library, even post-revision bump, let me know. Curious one that.

Merged in https://github.com/Homebrew/homebrew-core/commit/a4a6707dbb1e57fb379862537a17863abcbf2fe0. Thank you for your contribution to Homebrew @rdj; we appreciate it! :smiley_cat:

apjanke:
Will look at it. Dropping a link to the failed build here as a reference: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/1514/

apjanke:
I don't know.

I can reproduce this linkage error when I `brew install --build-from-source vorbis-tools` on my Yosemite VM, so it's not test-bot specific.

`/usr/lib/libnetwork.dylib` does not exist on any of my OS X machines that I checked, including my 10.10 and 10.9 machines. They have CLT installed (I checked by doing `sudo xcode-select --install` again), so it's not a matter of it being provided by the CLT installation.

Might be nice to have a `-v` or `--show-linkers` option on `brew linkage` so we could easily see exactly which binaries are linking to a given lib.

When I googled for `/usr/lib/libnetwork.dylib` and the like, I mostly got results about iOS Simulator stuff, and a few bug reports for other software that had that library in stack traces but didn't have anything directly relevant to it.

It's `bin/ogg123` in `vorbis-tools` that's linking to it.

```
$ for i in *; do echo $i; otool -L $i | grep libnetwork; done
ogg123
	/usr/lib/libnetwork.dylib (compatibility version 1.0.0, current version 583.20.10)
oggdec
oggenc
ogginfo
vcut
vorbiscomment
[✘ /usr/local/Cellar/vorbis-tools/1.4.0_1/bin on ⇄ master]
```

The `CHANGES` file in the `vorbis-tools` distribution has this item in the latest release.
```
vorbis-tools 1.4.0 -- 2010-03-25
[...snip...]
 * Add -lnetwork check for socket in configure
```

In the source tree:
```
[~/tmp/vorbis-tools/vorbis-tools-1.4.0]
$ grep -r lnetwork *
CHANGES: * Add -lnetwork check for socket in configure
configure:{ echo "$as_me:$LINENO: checking for socket in -lnetwork" >&5
configure:echo $ECHO_N "checking for socket in -lnetwork... $ECHO_C" >&6; }
configure:LIBS="-lnetwork  $LIBS"
configure:  SOCKET_LIBS="-lnetwork"
configure.ac:AC_CHECK_LIB(network, socket, SOCKET_LIBS="-lnetwork")
```

My Yosemite VM has `usr/lib/libnetwork.dylib` present in the iOS frameworks in Xcode 7.1 and 7.2.

```
$ find / -name libnetwork.dylib 2>/dev/null
/Applications/Xcode-7.1.1.app/Contents/Developer/Platforms/AppleTVSimulator.platform/Developer/SDKs/AppleTVSimulator.sdk/usr/lib/libnetwork.dylib
/Applications/Xcode-7.1.1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/lib/libnetwork.dylib
/Applications/Xcode-7.1.1.app/Contents/Developer/Platforms/WatchSimulator.platform/Developer/SDKs/WatchSimulator.sdk/usr/lib/libnetwork.dylib
/Applications/Xcode-7.2.1.app/Contents/Developer/Platforms/AppleTVSimulator.platform/Developer/SDKs/AppleTVSimulator.sdk/usr/lib/libnetwork.dylib
/Applications/Xcode-7.2.1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/lib/libnetwork.dylib
/Applications/Xcode-7.2.1.app/Contents/Developer/Platforms/WatchSimulator.platform/Developer/SDKs/WatchSimulator.sdk/usr/lib/libnetwork.dylib
[✘ /usr/local/Cellar/vorbis-tools/1.4.0_1/bin on ⇄ master]
```

Sooooo, it looks like... on Yosemite/Xcode 7.2.x, `vorbis-tools` is for some reason linking against an iOS SDK instead of an OS X SDK, maybe due to some `autoconf` issue?

UniqMartin:
> Might be nice to have a `-v` or `--show-linkers` option on `brew linkage` so we could easily see exactly which binaries are linking to a given lib.

For the time being you might want to check out my [`brew macho-dump $(brew --prefix vorbis-tools)`](https://github.com/UniqMartin/homebrew-tools/blob/master/cmd/brew-macho-dump.rb). See `brew macho-dump --help` for more details (`--wide` makes the output easier to grep).

> `/usr/lib/libnetwork.dylib` does not exist on any of my OS X machines that I checked, including my 10.10 and 10.9 machines.

It does exist on my OS X 10.11.4 machine and seems to be a core part of the OS according to `pkgutil`:

```
$ pkgutil --file-info /usr/lib/libnetwork.dylib
volume: /
path: /usr/lib/libnetwork.dylib

pkgid: com.apple.pkg.Essentials
pkg-version: 10.11.0.1.1.1442471843
install-time: 1443677583
uid: 0
gid: 0
mode: 755

pkgid: com.apple.pkg.update.os.10.11.1Auto.15B42
pkg-version: 1.0.0.0.1.1445123301
install-time: 1445499275
uid: 0
gid: 0
mode: 755

pkgid: com.apple.pkg.update.os.10.11.2.15C50
pkg-version: 1.0.0.0.1.1449123250
install-time: 1449626885
uid: 0
gid: 0
mode: 755

pkgid: com.apple.pkg.update.os.10.11.3.15D21
pkg-version: 1.0.0.0.1.1452734829
install-time: 1453292868
uid: 0
gid: 0
mode: 755

pkgid: com.apple.pkg.update.os.10.11.4.15E65
pkg-version: 1.0.0.0.1.1457768702
install-time: 1458615124
uid: 0
gid: 0
mode: 755
```

> Sooooo, it looks like... on Yosemite/Xcode 7.2.x, `vorbis-tools` is for some reason linking against an iOS SDK instead of (or in addition to) an OS X SDK for some symbols, maybe due to some `autoconf` issue?

It builds against the OS X 10.11 SDK that is the one provided by Xcode 7.2.x (even on Yosemite) and apparently falsely assumes that a library present in the SDK will also be available on the 10.10 host (specified via `MACOS_DEPLOYMENT_TARGET` by `superenv`). Not sure who's to blame here, but that's about what I think is going on and causing this issue.

apjanke:
Ah! Okay, I also see `/usr/lib/libnetwork.dylib` on my El Cap 10.11.4 machine. And I had only been looking for `*.dylib` files in Xcode.app, not `*.tbd` files, which are [apparently a new kind of stub for dylib files](http://stackoverflow.com/questions/31450690/why-xcode-7-shows-tbd-instead-of-dylib). Using that file name, I see it in the OS X SDK now on Xcode 7.2.1 running on OS X 10.10.

```
[/Applications/Xcode-7.2.1.app/Contents/Developer/Platforms]
$ find MacOSX.platform -name 'libnetwork.dylib'
[/Applications/Xcode-7.2.1.app/Contents/Developer/Platforms]
$ find MacOSX.platform -name 'libnetwork.tbd'
MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/lib/libnetwork.tbd
```

Sounds like the iOS thing is a red herring and it's a deployment target issue like you suggest. Thanks for clearing that up.

I suspect it is maybe a bug in `vorbis-tools`'s use of autoconf's `AC_CHECK_LIB`. They don't account for the case of `socket()` being available by default. Here's part of their `configure.ac`, starting at line 173.

```
SOCKET_LIBS=
AC_CHECK_LIB(socket, socket, SOCKET_LIBS="-lsocket")
AC_CHECK_LIB(network, socket, SOCKET_LIBS="-lnetwork")
AC_CHECK_LIB(nsl, gethostbyname, SOCKET_LIBS="-lnsl $SOCKET_LIBS")
```

Based on reading the autoconf 2.69 source code and [`AC_CHECK_LIB`'s doco](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Libraries.html), I think the `AC_CHECK_LIB` macro works by attempting to compile a program which uses the named function, in this case `socket`, with the given library linked. But on OS X, based on my testing, `socket()` is available in one of the libraries that's linked by default.

```
[~/tmp/socket]
$ cat test_socket.c
#include <sys/socket.h>
#include <stdio.h>

int main() {
    socket(0, 0, 0);
    printf("Hello, world!\n");
}
[~/tmp/socket]
$ cc test_socket.c
[~/tmp/socket]
$ ./a.out
Hello, world!
[~/tmp/socket]
$
```

So the `AC_CHECK_LIB` test for it will always succeed if `-lnetwork` can be done at all, which means `SOCKET_LIBS` ends up being the value of the last `AC_CHECK_LIB` test for `socket`, which in this case is the spurious `-lnetwork`.

If this is what's going on, I think you could consider this to partially be a bug in autoconf's `AC_CHECK_LIB`, since it just checks whether something compiles with a given lib, and not whether it differentially compiles with a given lib _and doesn't_ compile without that lib.

But why does compiling with `-lnetwork` succeed in the first place, in either the autoconf or actual build contexts? Maybe all the `.tbd` files are unconditionally available in the Xcode build environment, and `MACOS_DEPLOYMENT_TARGET` doesn't affect the `-l` success, just the stuff that's defined in the headers? And this just hasn't been a problem because usually OS X doesn't add new libraries that happen to have the same names as libraries on Linux or other systems and that thus get probed for by tools like autoconf? This I really don't know.

I think we could clear this bad linkage up just by patching `configure.ac` to comment out the `socket`-related `AC_CHECK_LIB` tests. But this is an interesting underlying problem. Should be reported upstream, probably to both vorbis-tools and autoconf (and maybe even to Apple?). I'll do some more research and give this a try if I have spare time this evening; anyone feel free to get ahead of me on this, though.

apjanke:
(What a fun puzzle! Stuff like this is why I like programming.)

UniqMartin:
> And I had only been looking for `*.dylib` files in Xcode.app, not `*.tbd` files, which are [apparently a new kind of stub for dylib files](http://stackoverflow.com/questions/31450690/why-xcode-7-shows-tbd-instead-of-dylib).

Those were primarily introduced to save space, as far as I know. They initially caused some grief for hand-crafted build systems that manually looked for `lib<something>.dylib` instead of relying on the magic behind `-l<something>`, which was extended to work with `.tbd` files transparently.

If you care about some Mach-O trivia: Previously those files in the SDKs had a `.dylib` extension, but had a file type of `MH_DYLIB_STUB` instead of `MH_DYLIB` (to be found in the `filetype` member of the `mach_header` struct). Just compare the files from an Xcode 6.x with those available in Xcode 7.x.

> But why does compiling with `-lnetwork` succeed in the first place, in either the autoconf or actual build contexts? *[…snip…]* Maybe all the `.tbd` files are unconditionally available in the Xcode build environment, and `MACOS_DEPLOYMENT_TARGET` doesn't affect the `-l` success, just the stuff that's defined in the headers?

That's because I didn't mention `SDKROOT` which wasn't entirely relevant to my previous comment. But if you set it up properly, or just use `xcrun --sdk macosx10.11 cc` instead of `cc`, the link should succeed.

> And this just hasn't been a problem because usually OS X doesn't add new libraries that happen to have the same names as libraries on Linux or other systems and that thus get probed for by tools like autoconf? This I really don't know.

That's hard to tell. Surprisingly, `/usr/lib/libnetwork.dylib` is actually a real independent library, unlike many others like `libc.dylib`, `libdl.dylib`, `libgcc_s.1.dylib`, `libm.dylib`, `libpthread.dylib` and a few others in the same directory, that are obviously only there for cross-platform compatibility (and possibly POSIX compliance), but all are just symlinks to `libSystem.dylib`. Some build systems get quite confused that so many things are centralized in a single library on OS X.

apjanke:
> If you care about some Mach-O trivia: Previously those files in the SDKs had a .dylib extension, but had a file type of MH_DYLIB_STUB instead of MH_DYLIB (to be found in the filetype member of the mach_header struct). Just compare the files from an Xcode 6.x with those available in Xcode 7.x.

My head just popped off.

UniqMartin:
> My head just popped off.

I hope in a good way. 😀

Both the new `.tbd` files and the old `.dylib` files that were actually dylib stub files are the magic behind why you can link to stuff deeply hidden inside `/Applications/Xcode.app/…` or wherever your Xcode happens to be installed, but end up with binaries that link to stuff in `/usr/lib`, `/System/Frameworks`, etc.

PR #918, opened 2016-05-06
cairomm: use secure homepage

vszakats:

---
- add canonical source url

### DIFF
--- a/Formula/cairomm.rb
+++ b/Formula/cairomm.rb
@@ -1,11 +1,8 @@
 class Cairomm < Formula
   desc "Vector graphics library with cross-device output support"
   homepage "https://cairographics.org/cairomm/"
-  # canonical source (same content):
-  # url "https://cairographics.org/releases/cairomm-1.12.0.tar.gz"
-  # sha256 "50435aec6fdd976934b791e808993160113ad19ca53a5634a9b64ccbe55874cc"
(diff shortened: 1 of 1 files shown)


vszakats:
Should the source package changed to the canonical location? (it's a `.gz` vs. current `.xz`, internal content being the same)

apjanke:
I don't know. Anyone know why it's downloading from this alternate location in the first place?

vszakats:
It was changed in 045dd7ccdb29b6110d93bb1cb9b6b058fcbb3aa9, possibly because the new release was made available on the other URL first, or else.

I'm updating to use the canonical one again.

apjanke:
Sounds reasonable. I've also asked on that commit to see if they can confirm.

Can you verify that the SHA sums for the individual files inside the two archive files are the same? If not, we should probably bump the revision, strictly speaking.

vszakats:
Yes, they do have the exact same set files with the exact same content, verified by `diff` and `shasum -a 256`. Only difference is in directory timestamps - but not content.

apjanke:
Cool. No revision bump necessary then; will merge when green.

vszakats:
Thank you!

DomT4:
Bit of a shame to lose the `.xz` compression but otherwise seems okay.

apjanke:

PR #939, opened 2016-05-07
icon-naming-utils: Get tarball from Debian.

hfiguiere:

---

Fixes Homebrew/brew#209

### DIFF
--- a/Formula/icon-naming-utils.rb
+++ b/Formula/icon-naming-utils.rb
@@ -1,7 +1,7 @@
 class IconNamingUtils < Formula
   desc "Script to handle icon names in desktop icon themes"
   homepage "https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html"
-  url "http://tango.freedesktop.org/releases/icon-naming-utils-0.8.90.tar.gz"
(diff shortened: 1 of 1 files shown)


hfiguiere:
I'll be honest I'm not sure if it is a good idea to pull from Debian and use their resources.

DomT4:
An imaginary $1 says if you run `brew doctor` it nudges you to `brew install curl` 😉.

DomT4:
Oh, that's actually not the problem for once. I can reproduce.

The system `curl` on 10.8 is heavily useless when it comes to a lot of secure connections these days, so that's become something of my go-to response when reports come from 10.8 about secure connections failing. My apologies for jumping to that conclusion here!

DomT4:
Fixed via https://github.com/Homebrew/homebrew-core/commit/2c96f339133529eb82baba6be0d0340715a16907. Apologies once again for the presumption 🙈, and thank you for reporting it to us.

hfiguiere:
I forgot to mention I did test in the browser :-D, but then the file is 404.

DomT4:
Yeah. Upstream have been working towards getting SSL/TLS across that whole domain for a little while now, which is great, but they seem to have turned it on by default (by forcing the redirect to `https`) without perhaps noticing the certificate is for `www.freedesktop.com` rather than a wildcard or specific subdomain, which has consequently broken things like `tango.freedesktop.com`.

PR #240, opened 2016-04-12
tmux 2.2

seryl:
- [√] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [√] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [√] Does your submission pass `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [√] Have you built your formula locally prior to submission with `brew install <formula>`?

PR #941, opened 2016-05-07
tmux 2.2

pandada8:


DomT4:
There was a big discussion on an OS X issue in https://github.com/Homebrew/homebrew-core/pull/240 that may be relevant here.

pandada8:
i see……

PR #950, opened 2016-05-07
unittest-cpp 1.6.1

pjohnmeyer:

---

New versions have a pre-built configure script so dependencies
on autotools were removed. A test block was also added to satisfy
brew audit, although it doesn't do much.

### DIFF
--- a/Formula/unittest-cpp.rb
+++ b/Formula/unittest-cpp.rb
@@ -1,8 +1,8 @@
 class UnittestCpp < Formula
   desc "Unit testing framework for C++"
   homepage "https://github.com/unittest-cpp/unittest-cpp"
-  url "https://github.com/unittest-cpp/unittest-cpp/archive/v1.5.1.tar.gz"
-  sha256 "7021008424054fe17dbd10307b72367f172846173d20848e555f64b8a6b4a8a3"
+  url "https://github.com/unittest-cpp/unittest-cpp/releases/download/v1.6.1/unittest-cpp-1.6.1.tar.gz"
+  sha256 "5b00a217f90fe262d91d09405ef5a8b5d28063da2f701cc58af48d560c4792af"
   head "https://github.com/unittest-cpp/unittest-cpp.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Formula/unittest-cpp.rb
```diff
@@ -12,13 +12,12 @@ class UnittestCpp < Formula
     sha256 "67f3d01e1d608e9f27d230e8e2bb35851248a69d526f120e78b13d9fdc2874b4" => :mavericks
   end
 
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
```
ilovezfs: These are actually for the head build.

ilovezfs:
It looks like upstream is using `cmake`. Is that not possible on OS X?

DomT4:
For some reason the CMake build scripts aren't present in the release tarball, only the autotool-based build system.

Formula/unittest-cpp.rb
DomT4: Initially they weren't, but yeah, still needed for head.

pjohnmeyer: Ah, sorry, I did not understand that. I can fix sometime soon.

Formula/unittest-cpp.rb
```diff
@@ -12,13 +12,12 @@ class UnittestCpp < Formula
...
-  depends_on "libtool" => :build
-
   def install
-    system "autoreconf", "-fvi"
     system "./configure", "--prefix=#{prefix}", "--disable-silent-rules"
     system "make", "install"
   end
+
+  test do
+    assert_match version.to_s, shell_output("cat /usr/local/lib/pkgconfig/UnitTest++.pc | grep Version: | cut -d' ' -f2")
```
DomT4: You don't need to use shell utilities directly here, Ruby can manage this itself:

```ruby
assert_match version.to_s, File.read(lib/"pkgconfig/UnitTest++.pc")
```

It'd be nice to write a more comprehensive test though, if you can think of one.

Formula/unittest-cpp.rb
DomT4: 👍 Just a case of doing:

```ruby
head do
  url "https://github.com/unittest-cpp/unittest-cpp.git"
end
```

pjohnmeyer:
The CMake scripts are in git and it GitHub's automated tag tarballs, they are not in the autotools-generated tarball that I uploaded as a release artifact.

I stuck with autotools on this PR because that's what the existing formula used, but both are viable options depending on which tarball is downloaded.

DomT4:
Seems fine using the autotools-based one given it has a pre-generated `configure` script; means people don't need to install any dependencies to install the stable formula, which is always nice.

Formula/unittest-cpp.rb
pjohnmeyer: The package has unit tests, but they aren't currently in the stable tarballs. That would be the best test -- is that an appropriate thing to do in this step for future updates?

mikemcquaid: This isn't for unit tests but packaging tests. You can use `assert_match version.to_s, IO.read("#{lib}/pkgconfig/UnitTest++.pc"}`

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/90b3594db8154d8ef1e4e21d2827064c5c22b63b. Thank you for your contribution to Homebrew @pjohnmeyer; we appreciate it! :smiley_cat:

PR #971, opened 2016-05-08
openssl@1.1 1.1.0 (new formula)

DomT4:
  _(Apart from one issue, which I've deliberately ignored for consistency)._

---

It's going to take us a long time to transition every formula across to the new OpenSSL release _(when it's stable)_ given the not insignificant API changes, as well as upstream deprecation of zlib compression and SSLv3.

Consequently, we're going to be running two OpenSSL releases for the foreseeable. The OpenSSL formula will remain for now the 1.0.2 branch which is supported until 2019, this formula will track 1.1.0 which is supported upstream till ~2018 under current plans.

Some things in this formula are on by default, such as the overdue death of SSLv3 and zlib compression, but since those are fairly big departures from OpenSSL releases to date it's explicitly stated in-formula so when people report _"This is broken using this"_ some of the very obvious breakage points are highlighted at cursory glance.

This OpenSSL release also introduces a fairly high Perl requirement. I'd like to note at this point this isn't our choice and is something we're trying to minimise, but to be somewhat blunt if you're running 10.8 or less you're going to have to accept the system utilities on those OS X versions are increasingly problematic today.

This formula is a fairly large departure from the way people have come to expect Homebrew to behave in terms of always running the latest. OpenSSL is a particularly punchy example of this, but going forwards incompatible releases are going to be given a little more leeway to remain around until a transition can happen in a mildly smoother manner than we've perhaps managed previously at times.

### DIFF
--- a/Formula/openssl@1.1.rb
+++ b/Formula/openssl@1.1.rb
@@ -1,0 +1,171 @@
+class OpensslAT11 < Formula
+  desc "Cryptography and SSL/TLS Toolkit"
+  homepage "https://openssl.org/"
+  url "https://www.openssl.org/source/openssl-1.1.0.tar.gz"
+  mirror "https://www.mirrorservice.org/sites/ftp.openssl.org/source/openssl-1.1.0.tar.gz"
+  sha256 "f5c69ff9ac1472c80b868efc1c1c0d8dcfc746d29ebe563de2365dd56dbd8c82"
+
+  keg_only :provided_by_osx,
+    "Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries"
+
+  option :universal
+  option "without-test", "Skip build-time tests (not recommended)"
+
+  # Only needs 5.10 to run, but needs >5.13.4 to run the testsuite.
+  # https://github.com/openssl/openssl/blob/4b16fa791d3ad8/README.PERL
+  # The MacOS ML tag is same hack as the way we handle most :python deps.
+  if build.with? "test"
+    depends_on :perl => "5.14" if MacOS.version <= :mountain_lion
+  else
+    depends_on :perl => "5.10"
+  end
+
+  def arch_args
+    {
+      :x86_64 => %w[darwin64-x86_64-cc enable-ec_nistp_64_gcc_128],
+      :i386   => %w[darwin-i386-cc],
+    }
+  end
+
+  # SSLv2 died with 1.1.0, so no-ssl2 no longer required.
+  # SSLv3 & zlib are off by default with 1.1.0 but this may not
+  # be obvious to everyone, so explicitly state it for now to
+  # help debug inevitable breakage.
+  def configure_args; %W[
+    --prefix=#{prefix}
+    --openssldir=#{openssldir}
+    no-ssl3
+    no-ssl3-method
+    no-zlib
+    enable-cms
+  ]
+  end
+
+  def install
+    # This could interfere with how we expect OpenSSL to build.
+    ENV.delete("OPENSSL_LOCAL_CONFIG_DIR")
+
+    # This ensures where Homebrew's Perl is needed the Cellar path isn't
+    # hardcoded into OpenSSL's scripts, causing them to break every Perl update.
+    # Whilst our env points to opt_bin, by default OpenSSL resolves the symlink.
+    if which("perl") == Formula["perl"].opt_bin/"perl"
+      ENV["PERL"] = Formula["perl"].opt_bin/"perl"
+    end
+
+    if build.universal?
+      ENV.permit_arch_flags
+      archs = Hardware::CPU.universal_archs
+    elsif MacOS.prefer_64_bit?
+      archs = [Hardware::CPU.arch_64_bit]
+    else
+      archs = [Hardware::CPU.arch_32_bit]
+    end
+
+    dirs = []
+
+    archs.each do |arch|
+      if build.universal?
+        dir = "build-#{arch}"
+        dirs << dir
+        mkdir dir
+        mkdir "#{dir}/engines"
+      end
+
+      ENV.deparallelize
+      system "perl", "./Configure", *(configure_args + arch_args[arch])
+      system "make", "clean" if build.universal?
+      system "make"
+      system "make", "test" if build.with?("test")
+
+      next unless build.universal?
+      cp "include/openssl/opensslconf.h", dir
+      cp Dir["*.?.?.dylib", "*.a", "apps/openssl"], dir
+      cp Dir["engines/**/*.dylib"], "#{dir}/engines"
+    end
+
+    system "make", "install", "MANDIR=#{man}", "MANSUFFIX=ssl"
+
+    if build.universal?
+      %w[libcrypto libssl].each do |libname|
+        system "lipo", "-create", "#{dirs.first}/#{libname}.1.1.dylib",
+                                  "#{dirs.last}/#{libname}.1.1.dylib",
+                       "-output", "#{lib}/#{libname}.1.1.dylib"
+        system "lipo", "-create", "#{dirs.first}/#{libname}.a",
+                                  "#{dirs.last}/#{libname}.a",
+                       "-output", "#{lib}/#{libname}.a"
+      end
+
+      Dir.glob("#{dirs.first}/engines/*.dylib") do |engine|
+        libname = File.basename(engine)
+        system "lipo", "-create", "#{dirs.first}/engines/#{libname}",
+                                  "#{dirs.last}/engines/#{libname}",
+                       "-output", "#{lib}/engines-1.1/#{libname}"
+      end
+
+      system "lipo", "-create", "#{dirs.first}/openssl",
+                                "#{dirs.last}/openssl",
+                     "-output", "#{bin}/openssl"
+
+      confs = archs.map do |arch|
+        <<-EOS.undent
+          #ifdef __#{arch}__
+          #{(buildpath/"build-#{arch}/opensslconf.h").read}
+          #endif
+        EOS
+      end
+      (include/"openssl/opensslconf.h").atomic_write confs.join("\n")
+    end
+  end
+
+  def openssldir
+    etc/"openssl@1.1"
+  end
+
+  def post_install
+    keychains = %w[
+      /System/Library/Keychains/SystemRootCertificates.keychain
+    ]
+
+    certs_list = `security find-certificate -a -p #{keychains.join(" ")}`
+    certs = certs_list.scan(
+      /-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----/m
+    )
+
+    valid_certs = certs.select do |cert|
+      IO.popen("#{bin}/openssl x509 -inform pem -checkend 0 -noout", "w") do |openssl_io|
+        openssl_io.write(cert)
+        openssl_io.close_write
+      end
+
+      $?.success?
+    end
+
+    openssldir.mkpath
+    (openssldir/"cert.pem").atomic_write(valid_certs.join("\n"))
+  end
+
+  def caveats; <<-EOS.undent
+    A CA file has been bootstrapped using certificates from the system
+    keychain. To add additional certificates, place .pem files in
+      #{openssldir}/certs
+
+    and run
+      #{opt_bin}/c_rehash
+    EOS
+  end
+
+  test do
+    # Make sure the necessary .cnf file exists, otherwise OpenSSL gets moody.
+    assert (HOMEBREW_PREFIX/"etc/openssl@1.1/openssl.cnf").exist?,
+            "OpenSSL requires the .cnf file for some functionality"
+
+    # Check OpenSSL itself functions as expected.
+    (testpath/"testfile.txt").write("This is a test file")
+    expected_checksum = "e2d0fe1585a63ec6009c8016ff8dda8b17719a637405a4e23c0ff81339148249"
+    system bin/"openssl", "dgst", "-sha256", "-out", "checksum.txt", "testfile.txt"
+    open("checksum.txt") do |f|
+      checksum = f.read(100).split("=").last.strip
+      assert_equal checksum, expected_checksum
+    end
+  end
+end


DomT4:
This is still very much a WIP. I need to work out the system Perl bug and whether that's an upstream problem (and then go report it if it is), but here's floating a discussion 🎈.

UniqMartin:
Given what I read on the OpenSSL homepage about their release policy and compatibility guarantees, I wonder whether this formula should be named `openssl11` instead. (A release of, e.g., OpenSSL 1.1.1 should in theory not break any formulae compiled against 1.1.0 unless we happen to disable features in a future release that are currently enabled.)

Formula/openssl110.rb
```diff
@@ -0,0 +1,184 @@
...
+      system "lipo", "-create", "#{dirs.first}/openssl",
+                                "#{dirs.last}/openssl",
+                     "-output", "#{bin}/openssl"
+
+      confs = archs.map do |arch|
+        <<-EOS.undent
+          #ifdef __#{arch}__
+          #{(buildpath/"build-#{arch}/opensslconf.h").read}
+          #endif
+          EOS
```
UniqMartin: This `EOS` seems to be misaligned.

DomT4: Yeah, it is 👍. I'm not sure who put it there originally, but I missed that alignment issue when I re-used it here. Will fix.

geoff-codes:
Yeah. This sucks. I gotta ask, what is the impetus for 1.1.x at all? Why are they doing this? What's being gained? Sure, zlib and SSLv3 should have probably been disabled a while ago actually, upstream, and/or in the homebrew formula, but that's all water under the bridge at this point. But I definitely think this should be openssl110 (and be relegated to `-versions` to boot).

I wonder how much stuff would break if the openssl deps were all changed to libressl.
That's really "the future" of this library, IMHO.

Formula/openssl110.rb
```diff
@@ -0,0 +1,184 @@
...
+      #{openssldir}/misc/CA.pl
+      #{openssldir}/misc/tsget
+      #{bin}/c_rehash
+    ]
+    perl_scripts.each do |f|
+      next unless File.exist?(f)
+      if File.read(f).include?(Formula["perl"].bin)
+        inreplace f, Formula["perl"].bin, Formula["perl"].opt_bin
+      end
+    end
```
xu-cheng: IMO, this part should be in `install` instead of `post_install`.

DomT4: Yeah, that shouldn't be a problem. I dumped it here for easier testing, and then didn't move it, but `install` makes sense to me.

Formula/openssl110.rb
```diff
@@ -0,0 +1,184 @@
...
+  option :universal
+  option "without-test", "Skip build-time tests (not recommended)"
+
+  # Only needs 5.10 to run, but needs >5.13.4 to run the testsuite.
+  # https://github.com/openssl/openssl/blob/4b16fa791d3ad8/README.PERL
+  if build.with? "test"
+    depends_on :perl => "5.14"
+  else
+    depends_on :perl => "5.10"
+  end
```
xu-cheng: This will cause similar problem to formulae which declare `depends_on :python`. It will introduce homebrew perl as hard dependency for our bottles.

DomT4: Indeed. I'm not quite sure what to do about that, truth be told, unless like `:python` we hack our inherently problematic system of default dependencies by doing something like:

```ruby
depends_on :perl => "5.14" if MacOS.version <= :mountain_lion
```

DomT4: Realistically, long-term, `default_formula` is an idea that needs significant work. We're running into multiple areas now where it mandates dancing around with less than elegant solutions.

xu-cheng: I think that would be one of options.

The another solution is that you could try `depends_on :perl => ["5.14", :run]`. However, in this case, you need to make sure the bottle can work on all of four situations:
* It's built by Homebrew perl and run with system perl.
* It's built by system perl and run with system perl.
* It's built by Homebrew perl and run with Homebrew perl.
* It's built by system perl and run with Homebrew perl.

mikemcquaid: > Realistically, long-term, default_formula is an idea that needs significant work. We're running into multiple areas now where it mandates dancing around with less than elegant solutions.

I've said it before and I'll say it again: we should switch the default on it so the "use `default_formula` for all bottles" behaviour is no longer in effect or make it something that's decided per-requirement (defaulting to not the current behaviour). The current behaviour is stupid and now causes more problems than it solves (and I implemented it originally which hopefully gives my opinion _some_ weight).

DomT4:
> I gotta ask, what is the impetus for 1.1.x at all? Why are they doing this?

1.1.x is a _(relative to OpenSSL)_ cleaner & more consistently-styled codebase, which long-term should make life a bit easier for the upstream devs and people who have to package it downstream, and less "cruft" brings with it the chances of less vulnerabilities from areas of the codebase not touched for years. They're also introducing new ciphers that other crypto libraries have had for a while.

The 1.1.x release branch is, IMHO, a step forwards for OpenSSL and will eventually make life easier for a whole bunch of other tools and libraries that require it. But yeah, short-term, this is going to be hell for most package managers whilst we wait for the OpenSSL ecosystem to catch up on 1.1.x and patch their projects for it.

> But I definitely think this should be openssl110 (and be relegated to -versions to boot).

Versions, as a tap, is more or less a dying pony at this point. There's already a fairly major cleanup going on behind the scenes there but long-term the core problem is that almost nobody in terms of either contributors or maintainers either wants or has the time to promise to look after versioned formulae adequately.

This sort of thing is going to become more commonplace, where upstream release major breaking changes the old version sticks around in the core for a while beside the new version. It'll likely happen with the next Boost release, for example.

> I wonder how much stuff would break if the openssl deps were all changed to libressl.

Lots. _Lots_. The LibreSSL team have made their fair share of breaking changes themselves, and for now is a less stable library than OpenSSL, as fond as I am of LibreSSL's work.

Formula/openssl110.rb
xu-cheng: > I've said it before and I'll say it again: we should switch the default on it so the "use default_formula for all bottles" behaviour is no longer in effect or make it something that's decided per-requirement (defaulting to not the current behaviour). The current behaviour is stupid and now causes more problems than it solves (and I implemented it originally which hopefully gives my opinion some weight).

I don't agree. I think the default behavior of `default_formula` has to be what it is now. i.e. if the formula's bottle is built by `default_formula`. The `default_formula` has to be installed unless explicitly told not to (as current `:run` tag). Because this is only the safe assumption, as the bottle may hardcode link to the `default_formula`(one of example is this openssl formula itself).

What could be improved is to let bottle default to build against to system instead of homebrew tools when both of them are  satisfied. Right now, it just chooses whatever come up with first.

geoff-codes:
> 1.1.x is a (relative to OpenSSL) cleaner & more consistently-styled codebase, which long-term should make life a bit easier for the upstream devs and people who have to package it downstream, and less "cruft" brings with it the chances of less vulnerabilities from areas of the codebase not touched for years. They're also introducing new ciphers that other crypto libraries have had for a while.

I see. Well, that's good news!

> Versions, as a tap, is more or less a dying pony at this point.

That makes sense. My, a lot has changed while I was gone for a bit.

> This sort of thing is going to become more commonplace, where upstream release major breaking changes the old version sticks around in the core for a while beside the new version.

I think that's really quite prudent, honestly. Personally, I think that bleeding-edge culture homebrew had going for a while may have been useful to a point, but its probably time to settle down a little bit. Perfectly fine with me.

> Lots. Lots. The LibreSSL team have made their fair share of breaking changes themselves, and for now is a less stable library than OpenSSL, as fond as I am of LibreSSL's work.

Yeah, I suspected as much. Bummer.

Formula/openssl110.rb
```diff
@@ -0,0 +1,184 @@
...
+
+    openssldir.mkpath
+    (openssldir/"cert.pem").atomic_write(valid_certs.join("\n"))
+  end
+
+  def caveats; <<-EOS.undent
+    A CA file has been bootstrapped using certificates from the system
+    keychain. To add additional certificates, place .pem files in
+      #{openssldir}/certs
+
```
geoff-codes: While we're kinda on the topic: are you guys aware that this doesn't actually work?

DomT4: Can you elaborate? It seems to work for me locally.

Formula/openssl110.rb
```diff
@@ -0,0 +1,184 @@
...
+    ]
+    perl_scripts.each do |f|
+      next unless File.exist?(f)
+      if File.read(f).include?(Formula["perl"].bin)
+        inreplace f, Formula["perl"].bin, Formula["perl"].opt_bin
+      end
+    end
+
+    keychains = %w[
+      /Library/Keychains/System.keychain
```
geoff-codes: And I've been sitting on this one for _a while actually_ because, honestly ,I didn't know what would be an appropriately-private forum. So this is kinda **burn-after-reading**. If there's a better place to talk about this, let me know.

But essentially, this line is not such a good idea.
As in:

Say I have a self-signed root CA and a leaf SSL cert in the System keychain (not uncommon). In Apple-land, these get big ❌s, warnings, etc. But the `.pem` file we're dumping is a just _blind whitelist_.

Example:
1. go to https://geoff.codes
You should get a big invalid certificate modal dialog.

2. `wget https://geoff.codes`
You should also get stopped here because of certificate problems.

3. `cat [1.pem.txt](https://github.com/Homebrew/homebrew-core/files/254015/1.pem.txt) [2.pem.txt](https://github.com/Homebrew/homebrew-core/files/254016/2.pem.txt)  >> "$(brew --prefix)/etc/openssl/cert.pem"

4. `wget https://geoff.codes`
😨

(Github made me add .txt to the pem files to upload, but if you rename them back to .pem, Spotlight will quickly show you that the emperor has no clothes.)

Formula/openssl110.rb
geoff-codes: If you put `.pem`s in `#{openssldir}/certs`, they do nothing.

Formula/openssl110.rb
UniqMartin: If anyone has missed the discussion and the potential problems of changing the current behavior for bottles: Homebrew/legacy-homebrew#47890

Formula/openssl110.rb
DomT4: I wouldn't consider this a common security issue so I'm fine leaving it here. Give me some time to look into this.

Formula/openssl110.rb
mikemcquaid: > Because this is the only safe assumption, as the bottle may hardcode link to the default_formula(one of examples is this openssl formula itself).

We'll catch that with `brew linkage` now, though.

Formula/openssl110.rb
geoff-codes: @DomT4 Fair enough — but know this: I'm going to be offline for a bit.
Because, (I shit you not), someone already used this vector to plant some very gnarly malware 😞 on my machine. I literally just had this demo up for 8 hours; there was no server up before that.

----
From a tor exit node:
```
default 162.247.72.27 - - [08/May/2016:14:19:29 -0700] "GET / HTTP/1.1" 302 265 "http://burger-imperia.com/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.120 Safari/537.36"
...
http://apple.com-systemnotice.com/eryp/index.html ... (includes my IP, geoip stuff, etc)... =BASE64dmlkLi4wMDAwMDAwNC0xYzIzLTQ5NDUtODAwMC0wMDAwMDAwMDAwMDBfX3ZwaWQuLjM1ODNmMDAwLTE1NjgtMTFlNi04NDdiLTc1MzY4ZDdmYTRmM19fY2FpZC4uMTQ5ZDNhYjMtNDg3Ny00N2M2LWEzYWEtYjJjYzE0YmFiYzA1X19ydC4uRF9fbGlkLi43NjUzYmZhNS0yNjA1LTQ1ZmMtYmQ0Yy1iNmJlMzA1MjgzZDZfX29pZDEuLmI4OGM4ZDlhLWNjMzktNDg2NC05ZDI0LWI0ZWQ5NWNmOThmN19fdmFyMS4udGFuZ28tcHllLVZmVThHdXlvX192YXIyLi50d2VlZGVoYW5kc19fdmFyNC4uTk9OLUFEVUxUX192YXI1Li5ET01BSU5fX3JkLi5fX2FpZC4uX19hYi4uX19zaWQuLg
...
vid..00000004-1c23-4945-8000-000000000000__vpid..3583f000-1568-11e6-847b-75368d7fa4f3__caid..149d3ab3-4877-47c6-a3aa-b2cc14babc05__rt..D__lid..7653bfa5-2605-45fc-bd4c-b6be305283d6__oid1..b88c8d9a-cc39-4864-9d24-b4ed95cf98f7__var1..tango-pye-VfU8Guyo__var2..tweedehands__var4..NON-ADULT__var5..DOMAIN__rd..__aid..__ab..__sid.
```

..and I'm still grepping through a billion logs for what happend after that, but somehow, I'm already on three blacklists already.

I had thought this would be a relatively benign way to demo this issue; but I guess not.
But please understand I intentionally omitted mentioning the real attack vectors in my commend above.

So, while you're investigating, you might wan't to use a VM.

I, (stupidly), just pointed my A record to my laptop.

So I'm gonna be doing some serious digging and reformatting; I may need sterilize the whole machine, depending on how deep this is. (I do have current backups, thankfully.)

So, a parting thought. Perhaps you may want to think of it this way: what relevant certificates exist in `/Library/Keychains/System.keychain` (any and all of which will be treated as root CAs) do we _need_ to have in `cert.pem`? What purpose do they serve? There is a reason the root CA list on operating systems are usually **locked down** (including OS X now, with SIU — but that doesn't cover `/Library/...`).

Anyway, if you need any pointers replicating/investigating, I'm happy to help, either here in this thread, and I can jump on IRC too if you give me some notice.

🍻

DomT4: I've pruned and saved the comments here elsewhere, for now. Never trust an exposed connection facing the internet, sadly 😕.

geoff-codes: Good. 

> Never trust an exposed connection facing the internet, sadly.
I feel you, but note that (in theory) off-the-shelf Server.app is supposed to be secure.

I was in the middle of editing my "demo comment" when you cleaned up, so you can take it or leave it as you like, if you think its helpful.

This line is not such a good idea.

As in:

Example:
1. Create a self-signed CA, and then create and sign a SSL cert with that CA. Launch a https server that uses these (Server.app is a quick way to do this, especially as we're using SecureTransport to contrast with OpenSSL). Once configure correctly, when you browse to your server, you should get a big invalid certificate modal dialog.

2. User the system `curl` to hit it over HTTPS.
You should also get stopped here because of certificate problems.

3. Reinstall openssl.rb, so the keychains get dumped again.  Then, using Homebrew curl (or wget, etc., anything linked to brew's OpenSSL), provided everything is where it should be, you should now no longer get an error. 
Not a good thing. 😨

Formula/openssl110.rb
xu-cheng: > We'll catch that with brew linkage now, though

`brew linkage` only catch linkage paths for dylib files. It cannot catch, for example, hard code path in the scripts.

DomT4:
Updated to reflect Xu's comments on `post_install` and Martin's comments on the naming and EOS misplacement. Still working on the Perl issues and the other raised points.

vszakats:
How is migration planned to be done for `openssl` dependent formulae?

Ideally, the switch to `openssl11` from `openssl` should be made once support is available for a given formula. But, this may still break dependent projects which depend on that given formula _and_ some other ones as well, that _don't yet have_ 1.1.x support. What would be the solution to handle this scenario?

Mixing old and new `openssl`s in those projects don't seem like a very safe idea. Holding off OpenSSL upgrades until every dependent formulae and all their related subcomponents support it, may mean a very slow upgrade process. Using alternate names for `openssl11` dependent formulae would look quite messy (f.e. `libssh2-openssl11`), plus it'd need heavy patching to avoid colliding names.

[ Apologize if this is an already resolved problem. ]

UniqMartin:
@vszakats This hasn't been completely formalized and we don't have a coherent document yet we can point people at. This is partially a result of the discussion in https://github.com/Homebrew/brew/issues/60. A summary is in the top post and it's probably worth reading the few comments from https://github.com/Homebrew/brew/issues/60#issuecomment-216687487 to the end. Though if you're curious, I'd say reading the whole thread for more context is worthwhile.

vszakats:
Thanks for the pointers @UniqMartin!

DomT4:
Although for full disclosure, even I haven't read that thread yet, knowing full well once I do I'll end up writing a minor dissertation 😅.

> Mixing old and new openssls in those projects don't seem like a very safe idea.

Agreed. They shouldn't end up in the same recursive dependency tree in an ideal world.

> Holding off OpenSSL upgrades until every dependent formulae and all their related subcomponents support it, may mean a very slow upgrade process. 

I'm not even going to set a timetable for this, to be honest. 1.0.2 is supported until 2019 and it wouldn't surprise me if it takes _that long_ for our entire package ecosystem that needs OpenSSL to be shunted across to what will presumably by then by 1.1.1 or something.

If we're lucky Debian will take the jump sooner rather than later, significant chunks of the rest of Linux will follow and the ecosystem will go "Linux has moved over, we need to get on with fixing any incompatibility sharpish". We saw something like this happen when Debian killed off SSLv3, although on a smaller scale than would be required here.

The primary goal of this package isn't necessarily to move everything right out of the gate; it's to provide an easy way to test what works and what doesn't locally & also fulfil the loose promise that somewhere, somehow Homebrew supplies the latest and greatest.

> Using alternate names for openssl11 dependent formulae would look quite messy (f.e. libssh2-openssl11), plus it'd need heavy patching to avoid colliding names.

Yeah, have no desire to go down this path whatsoever. Even if Homebrew had 50 maintainers and a large CI farm I'd have my concerns about this, let alone as we are, heh.

vszakats:
Thank you @DomT4, I've also read into the other thread and now have a picture of how this is planned.

Here's hoping the 1.1.x upgrade can happen relatively fast and smooth. It's an important upgrade, with clear benefits. I could drop some ugly old cruft from existing code while updating, and most attention needed to go into keeping compatibility with previous/current OpenSSL versions.

geoff-codes:
Hey - I was wondering if anyone knew where the original discussion that led from from changing from using the Curl CA bundle to using the certs generated by the `security.` lives these days? It's probably several years old by now. I know that @mistydemeo's tigerbrew ~~still uses this~~ still uses this _in some cases_. I'm not necessarily advocating that approach exactly per se, but https://www.apple.com/certificateauthority (with documents like https://www.apple.com/certificateauthority/ca_program.html) are pretty interesting.

DomT4:
@geoff-codes Your comments were investigated further & resulted in changes:

* https://github.com/Homebrew/homebrew-core/commit/d5633692693accf05f0c644674407648bf645f4f
* https://github.com/Homebrew/homebrew-core/commit/092e7608b286d124d689cac2bd59cd217a2e195f
* https://github.com/Homebrew/homebrew-core/commit/30d3766453347f6e22b3ed6c74bb926d6def2eb5
* https://github.com/Homebrew/homebrew-versions/commit/28dc0e4932689086544e5c3d78a4383572e618b9

Thanks for highlighting the potential issue there.

geoff-codes:
@DomT4 No problem!
Thanks for taking it seriously (I appreciate the rev. bump), and taking care of it so thoroughly. 

I'm considering opening an issue in [brew](https://github.com/Homebrew/brew) for discussion of possible robust, long-term approaches to this sort of thing. While I wouldn't consider it _at all_ critical, its still not ideal how we're reliant on the Apple CA list _at the time of install_; when Apple updates the list with a software update, brew's list doesn't get updated without reinstalling. Which is to say nothing of older versions of OS X, which don't get updated at all. And it would be nice to have some type of mechanism/game-plan in place for if and when something [like this](https://blog.mozilla.org/security/2011/09/02/diginotar-removal-follow-up) happens again.

@owners, would you guys be open to some type of discussion on this?

Formula/openssl110.rb
DomT4: Since there's a fair bit of disagreement on how we get past this recurring issue for now I've adopted the `:python` hack:

```ruby
  if build.with? "test"
    depends_on :perl => "5.14" if MacOS.version <= :mountain_lion
  else
    depends_on :perl => "5.10"
  end
```

Still doesn't build properly against the system Perl though, working on it.

DomT4:
> I'm considering opening an issue in brew for discussion of possible robust, long-term approaches to this sort of thing. 

Don't worry about the Issue. I've thought about this before and did so again after [re-reading the original discussion a few days back](https://github.com/Homebrew/legacy-homebrew/pull/21065). I have 90% of a PR ready to go already. Give me today to file that, PRs generate more discussion than Issues by their nature, especially once I threaten to start merging mine as-is 😜.

geoff-codes:
@DomT4 Fantastic. 😉

DomT4:
^ https://github.com/Homebrew/brew/pull/241

geoff-codes:
You read my mind. 💯

DomT4:
Pushed a bunch of changes here, contacted OpenSSL about the Perl hashbang oddities, etc. PR may be slow, but it's not dead yet 😉.

CamJN:
Just FYI https://twitter.com/richsalz/status/759418686608474112

DomT4:
Yeah, I'm aware, I'm on the mailing list, heh. Thanks for the unintentional nudge here though!

Keep meaning to bump this to the current beta; need to check the status around the Perl hashbang, which by reporting upstream I managed to spark a wider debate which I quietly backed away from & left them to decide the best path forwards 😄.

DomT4:
This be stable now. PR updated.

Formula/openssl11.rb
```diff
@@ -0,0 +1,176 @@
...
+  # The MacOS ML tag is same hack as the way we handle most :python deps.
+  if build.with? "test"
+    depends_on :perl => "5.14" if MacOS.version <= :mountain_lion
+  else
+    depends_on :perl => "5.10"
+  end
+
+  # This isn't technically needed where Clang exists, but is a small
+  # dep and is hard to reliably express non-Clang-only in dependency format.
+  depends_on "makedepend" => :build
```
chdiza: Doesn't clang automatically exist if either the CLT or Xcode is installed?  If so, can we just drop this?

OTOH something rings a bell that `makedepend` entered the picture for reasons totally unconnected to compilers.

DomT4: If you changed the default algorithms you needed to run `make depend` to rejiggle _(technical term that)_ things to build correctly before. I _think_ it would still be necessary if you've got a dirty source. Documentation upstream is kind of "eh" on the `makedepend` situation. There's a commit [here](https://github.com/openssl/openssl/commit/d20bb611d90df0dc3561a6972ba6bf0e3c65b3e5) that adds some detail, but not much.

I'll try poking it locally. If it breaks I guess we could always add the dep in later.

chdiza: Yeah, I *think* makedepend was brought in because we activated the `ec-nistp` stuff.  I *think* that without the `make depend` step we *silently* weren't getting the `ec-nistp`.

But then how could it be true that `makedepend` is unnecessary when clang is present?

DomT4: https://github.com/openssl/openssl/commit/30752dd7df7c699b380043cb73e384dd84cab539
https://github.com/openssl/openssl/commit/09aa263a143b006650a0f879dc2348609587b806

Apparently Clang (since always) and some GCC versions support dependency generation natively without having to lean on `makedepend` to do that magic.

mithrandi:
> If we're lucky Debian will take the jump sooner rather than later, significant chunks of the rest of Linux will follow and the ecosystem will go "Linux has moved over, we need to get on with fixing any incompatibility sharpish". We saw something like this happen when Debian killed off SSLv3, although on a smaller scale than would be required here.

Debian has had OpenSSL 1.1.0 prereleases in experimental for a while now, and there is a [transition tracking bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=827061) open (the lengthy list of blocking bugs there are packages that failed to rebuild against 1.1.0). It'll probably take a while to sort out the broken packages, but I expect the next stable release of Debian to have 1.1.0. So at worst, you can probably borrow Debian's patches in many cases, but hopefully most of these problems will be fixed upstream, rather than requiring distribution-specific patches.

DomT4:
> and there is a transition tracking bug open

Yeah, I've been keeping an eye on some of them. It's a few months old now so less useful but https://breakpoint.cc/openssl-1.1-rebuild-2016-05-29/ is interesting & an easier format to read through.

> So at worst, you can probably borrow Debian's patches in many cases, but hopefully most of these problems will be fixed upstream, rather than requiring distribution-specific patches.

Yeah, I don't have a lot of interest in carrying around what is essentially a feature patch (support for 1.1.0) given OpenSSL are supporting 1.0.2 as an LTS branch. Not planning to leap over upstream releases unless absolutely necessary for whatever reason.

Transition is also complicated by potentially overlapping linkage, so for example, if you `brew install curl --with-openssl` against OpenSSL 1.1.0 but you add `--with-libssh2` as well which remains linked against OpenSSL 1.0.2, we're not really sure how predictably that'll behave. Ideally, the plan is:

* Anything standalone (i.e. nothing in `brew uses`) that supports OpenSSL 1.1.0 can use this new formula as the dependency.
* Anything else needs to be upgraded by tree, so to update `libqalculate` you'd want `wget` to be using OpenSSL 1.1.0 first, etc.

We may well change that plan as time passes; If I could I'd have the entire ecosystem using OpenSSL 1.1.0 tomorrow but it really depends on how quickly we can get this done without breaking significant things for users.

DomT4:
This is blocked on https://github.com/Homebrew/brew/pull/812 temporarily.

DomT4:
Now blocked on https://github.com/Homebrew/brew/pull/832. Making progress!

MikeMcQuaid:
@DomT4 :shipit:!

DomT4:
Can't quite yet. There's a new blocker around `pull`, which might actually be a deeper issue.

```
No available formula with the name "homebrew/core/openssl@1.1"
```

I poked at it a bit in `irb` but haven't had the chance to see whether it's a purely `pull` issue or pointing at a deeper problem yet:
```
irb(main):058:0> orig_revision = "05256bca5b18a84f98258e3da847c2bb8c4f53ec"
=> "05256bca5b18a84f98258e3da847c2bb8c4f53ec"
irb(main):059:0> changed_formulae_names = []
=> []
irb(main):060:0>  Utils.popen_read("git", "diff-tree", "-r", "--name-only","--diff-filter=AM", orig_revision, "HEAD", "--", tap.formula_dir.to_s).each_line do |line|
irb(main):061:1* next unless line.end_with? ".rb\n"
irb(main):062:1> name = "#{tap.name}/#{File.basename(line.chomp, ".rb")}"
irb(main):063:1> changed_formulae_names << name
irb(main):064:1> end
=> "Formula/openssl@1.1.rb\n"
irb(main):065:0> changed_formulae_names
=> ["homebrew/core/openssl@1.1"]
irb(main):072:0> changed_formulae_names.each do |name|
irb(main):073:1* begin
irb(main):074:2* f = Formula[name]
irb(main):075:2> end
irb(main):076:1> end
FormulaUnavailableError: No available formula with the name "homebrew/core/openssl@1.1"
	from /usr/local/Library/Homebrew/formulary.rb:188:in `get_formula'
	from /usr/local/Library/Homebrew/formulary.rb:216:in `factory'
	from /usr/local/Library/Homebrew/formula.rb:1279:in `[]'
irb(main):089:0> changed_formulae_names.to_s.gsub!(/@/, "AT") { |f| Formula[f].name }
=> "[\"homebrew/core/opensslAT1.1\"]"
```

MikeMcQuaid:
@DomT4 Weirdly worked fine for me the second time I did a `brew pull` after a `git reset --hard`. Can you confirm if that works for you? If not, can I see the `brew pull --debug` output?

DomT4:
It `brew pull`s okay initially, but you can't get a bottle from it, even though there's one sat on Bintray.

```
~> brewpb https://github.com/Homebrew/homebrew-core/pull/971
==> Fetching patch
Patch: https://github.com/Homebrew/homebrew-core/pull/971.patch
==> Applying patch
Applying: openssl@1.1 1.1.0 (new formula)
==> Patch closes issue #971
Error: No available formula with the name "homebrew/core/openssl@1.1"

~> gloga
* 3313499 (HEAD -> master) openssl@1.1 1.1.0 (new formula)
```

```
~> brewpb https://github.com/Homebrew/homebrew-core/pull/971 --debug
==> Fetching patch
Patch: https://github.com/Homebrew/homebrew-core/pull/971.patch
==> Applying patch
Applying: openssl@1.1 1.1.0 (new formula)
==> Patch closes issue #971
Error: No available formula with the name "homebrew/core/openssl@1.1"
/usr/local/Library/Homebrew/formulary.rb:188:in `get_formula'
/usr/local/Library/Homebrew/formulary.rb:216:in `factory'
/usr/local/Library/Homebrew/formula.rb:1279:in `[]'
/usr/local/Library/Homebrew/cmd/pull.rb:171:in `block in pull'
/usr/local/Library/Homebrew/cmd/pull.rb:49:in `each'
/usr/local/Library/Homebrew/cmd/pull.rb:49:in `pull'
/usr/local/Library/Homebrew/brew.rb:91:in `<main>'
```

DomT4:
Because someone will notice & ask, the `postinstall` step is hilariously verbose now. That's an upstream change, not something on our end. I'll look into trying to do that step a bit more quietly, but this PR has sat for a long, long time now so let's get it merged & let people play with it and the new core-versioned formulae system.

Merged in https://github.com/Homebrew/homebrew-core/commit/663e4792e1f355ef2571a06a9101606027cfedf9.

MikeMcQuaid:
@DomT4 Hooray! 👏

DomT4:
@MikeMcQuaid Thanks for your help finishing up the `@` support! Couldn't have got it merged without the final piece in the puzzle https://github.com/Homebrew/brew/commit/0c44ce2a38d42876d02275b4db719930860beb03.

I'll try and spin up a VM soon locally to rebuild as much as possible in the core that currently uses `openssl` & we can track what's horribly broken & what works already.

PR #980, opened 2016-05-08
file-roller: use upstream patch for schema bug

ilovezfs:
- Add undeclared dependencies: make the undeclared dependencies reported by `brew linkage` explicit: atk, cairo, gdk-pixbuf, gettext, glib, pango
- Use assert_match in the version test: Verify the version is correct instead of just running the command; redirect stderr to stdout so that it will only show the following harmless, but distracting, message when the test is run verbosely:

```
    file-roller[80748:8815155] *** WARNING: Method userSpaceScaleFactor
    in class NSView is deprecated on 10.7 and later. It should not be
    used in new applications. Use convertRectToBacking: instead. 
```
- Use upstream patch for schema bug: This patch forces autoconf, automake, libtool, and yelp-tools to be
  build-time dependencies temporarily. Both the patch and these temporary dependencies can be removed when >13.16.4 is released.

### DIFF
--- a/Formula/file-roller.rb
+++ b/Formula/file-roller.rb
@@ -30,18 +30,32 @@ class FileRoller < Formula
   depends_on "glib"
   depends_on "pango"
 
+  # These can be removed when the next version (>3.16.4) is released, since they
+  # are only required to support the patch below, which will no longer be needed
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "yelp-tools" => :build
+
+  # Remove when >3.16.4 is released
+  # Build fails during schema validation when using glib >= 1.45.1
+  # Upstream patch for https://bugzilla.gnome.org/show_bug.cgi?id=756607
+  patch do
+    url "https://git.gnome.org/browse/file-roller/patch/?id=37df6d640817127dbad0f9ba94141cfb1e6a8360"
+    sha256 "5dee926832a607fc703518cd35b8fe0846c6bf760faf19982a89fe5f5c21fe75"
+  end
+
   def install
+    ENV.append "CFLAGS", "-I#{Formula["libmagic"].opt_include}"
+    ENV.append "LIBS", "-L#{Formula["libmagic"].opt_lib}"
+
+    # Remove when >3.16.4 is released
+    system "autoreconf", "-fiv"
+
     # forces use of gtk3-update-icon-cache instead of gtk-update-icon-cache. No bugreport should
     # be filed for this since it only occurs because Homebrew renames gtk+3's gtk-update-icon-cache
     # to gtk3-update-icon-cache in order to avoid a collision between gtk+ and gtk+3.
     inreplace "data/Makefile.in", "gtk-update-icon-cache", "gtk3-update-icon-cache"
-    ENV.append "CFLAGS", "-I#{Formula["libmagic"].opt_include}"
-    ENV.append "LIBS", "-L#{Formula["libmagic"].opt_lib}"
-
-    # Upstream bug: https://bugzilla.gnome.org/show_bug.cgi?id=756607
-    # A more elaborate, "correct" fix would be similar to this:
-    # https://github.com/mate-desktop/mate-utils/commit/c4df12f12d21ea7d4bc0d656bd5f93539c078d93
-    inreplace "configure", "$(GLIB_COMPILE_SCHEMAS) --strict", "$(GLIB_COMPILE_SCHEMAS)"
 
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
--- a/Formula/file-roller.rb
+++ b/Formula/file-roller.rb
@@ -22,6 +22,14 @@ class FileRoller < Formula
   depends_on "hicolor-icon-theme"
   depends_on "gnome-icon-theme"
 
+  # Add linked-library dependencies
+  depends_on "atk"
+  depends_on "cairo"
+  depends_on "gdk-pixbuf"
+  depends_on "gettext"
+  depends_on "glib"
+  depends_on "pango"
+
   def install
     # forces use of gtk3-update-icon-cache instead of gtk-update-icon-cache. No bugreport should
     # be filed for this since it only occurs because Homebrew renames gtk+3's gtk-update-icon-cache
--- a/Formula/file-roller.rb
+++ b/Formula/file-roller.rb
@@ -59,6 +59,6 @@ class FileRoller < Formula
   end
 
   test do
-    system "#{bin}/file-roller", "--version"
+    assert_match version.to_s, shell_output("#{bin}/file-roller --version 2>&1")
   end
 end


Formula/file-roller.rb
```diff
@@ -11,30 +11,48 @@ class FileRoller < Formula
...
+  depends_on "gnome-icon-theme"
   depends_on "gtk+3"
+  depends_on "hicolor-icon-theme"
   depends_on "json-glib"
-  depends_on "libmagic"
   depends_on "libarchive"
-  depends_on "hicolor-icon-theme"
-  depends_on "gnome-icon-theme"
+  depends_on "libmagic"
+  depends_on "pango"
```
UniqMartin: Apart from sorting things, there are a lot of new dependencies here and in the build dependencies above. How come? This change is neither mentioned in the PR body nor in the commit message.

ilovezfs: I added the "possibly undeclared" dependencies reported by brew linkage.

DomT4: It looks like this is the full recursive deps tree, rather than the immediate level of dependencies.

Formula/file-roller.rb
```diff
@@ -51,6 +69,6 @@ def post_install
   end
 
   test do
-    system "#{bin}/file-roller", "--version"
+    assert_match version.to_s, shell_output("#{bin}/file-roller --version 2>/dev/null")
```
DomT4: Why the need for `2>/dev/null`?

Formula/file-roller.rb
```diff
@@ -11,30 +11,48 @@ class FileRoller < Formula
     sha256 "05ca1ccdd5771e574bf4b1bad58fa0ddc6d2db65d2fda7999912ced5e0cb704e" => :mavericks
   end
 
-  depends_on "pkg-config" => :build
+  depends_on "autoconf" => :build
```
DomT4: Probably worth noting the autotool deps are only required for the patch, but can be removed once the patch is in stable?

Formula/file-roller.rb
ilovezfs: The mandatory new ones are the Autotools and yelp-tools. The other new ones are things it's actually linked against.

ilovezfs: depends_on "autoconf" => :build
  depends_on "yelp-tools" => :build

  depends_on "atk"
  depends_on "cairo"
  depends_on "gdk-pixbuf"
  depends_on "gettext"
  depends_on "glib"
  depends_on "pango"

ilovezfs:
```
iMac-TMP:Formula joe$ HOMEBREW_DEVELOPER=1 brew linkage file-roller
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/atk/lib/libatk-1.0.0.dylib (atk)
  /usr/local/opt/cairo/lib/libcairo-gobject.2.dylib (cairo)
  /usr/local/opt/cairo/lib/libcairo.2.dylib (cairo)
  /usr/local/opt/gdk-pixbuf/lib/libgdk_pixbuf-2.0.0.dylib (gdk-pixbuf)
  /usr/local/opt/glib/lib/libgthread-2.0.0.dylib (glib)
  /usr/local/opt/gtk+3/lib/libgdk-3.0.dylib (gtk+3)
  /usr/local/opt/gtk+3/lib/libgtk-3.0.dylib (gtk+3)
  /usr/local/opt/json-glib/lib/libjson-glib-1.0.0.dylib (json-glib)
  /usr/local/opt/libarchive/lib/libarchive.13.dylib (libarchive)
  /usr/local/opt/libmagic/lib/libmagic.1.dylib (libmagic)
  /usr/local/opt/pango/lib/libpango-1.0.0.dylib (pango)
  /usr/local/opt/pango/lib/libpangocairo-1.0.0.dylib (pango)
  atk
  cairo
  gdk-pixbuf
  gettext
  glib
  pango
```

Formula/file-roller.rb
ilovezfs: I thought that was self evident since there's no head build but happy to add a comment. ;)

DomT4: Most people don't bother even checking dependencies unless there's something notable about them, alas 😅.

Formula/file-roller.rb
ilovezfs: To shut it up.

```
iMac-TMP:Formula joe$ brew test -v file-roller
Testing file-roller
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160508-80740-1a2nm7r.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/file-roller.rb -v
==> /usr/local/Cellar/file-roller/3.16.4/bin/file-roller --version
2016-05-08 06:57:05.066 file-roller[80748:8815155] *** WARNING: Method userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be used in new applications. Use convertRectToBacking: instead. 
iMac-TMP:Formula joe$
```

DomT4: 😆 You'll be hard pressed to find GTK-based applications that don't throw errors similar to this. I think it's okay without personally.

ilovezfs:
@UniqMartin Based on
```
iMac-TMP:Formula joe$ HOMEBREW_DEVELOPER=1 brew linkage file-roller
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/atk/lib/libatk-1.0.0.dylib (atk)
  /usr/local/opt/cairo/lib/libcairo-gobject.2.dylib (cairo)
  /usr/local/opt/cairo/lib/libcairo.2.dylib (cairo)
  /usr/local/opt/gdk-pixbuf/lib/libgdk_pixbuf-2.0.0.dylib (gdk-pixbuf)
  /usr/local/opt/glib/lib/libgthread-2.0.0.dylib (glib)
  /usr/local/opt/gtk+3/lib/libgdk-3.0.dylib (gtk+3)
  /usr/local/opt/gtk+3/lib/libgtk-3.0.dylib (gtk+3)
  /usr/local/opt/json-glib/lib/libjson-glib-1.0.0.dylib (json-glib)
  /usr/local/opt/libarchive/lib/libarchive.13.dylib (libarchive)
  /usr/local/opt/libmagic/lib/libmagic.1.dylib (libmagic)
  /usr/local/opt/pango/lib/libpango-1.0.0.dylib (pango)
  /usr/local/opt/pango/lib/libpangocairo-1.0.0.dylib (pango)
  atk
  cairo
  gdk-pixbuf
  gettext
  glib
  pango
iMac-TMP:Formula joe$  brew macho-dump $(brew --cellar file-roller)
/usr/local/Cellar/file-roller/3.16.4/bin/file-roller:
    kind=executable
    arch=[x86_64]
    dylib_load=/usr/lib/libSystem.B.dylib
    dylib_load=/usr/local/opt/atk/lib/libatk-1.0.0.dylib
    dylib_load=/usr/local/opt/cairo/lib/libcairo-gobject.2.dylib
    dylib_load=/usr/local/opt/cairo/lib/libcairo.2.dylib
    dylib_load=/usr/local/opt/gdk-pixbuf/lib/libgdk_pixbuf-2.0.0.dylib
    dylib_load=/usr/local/opt/gettext/lib/libintl.8.dylib
    dylib_load=/usr/local/opt/glib/lib/libgio-2.0.0.dylib
    dylib_load=/usr/local/opt/glib/lib/libglib-2.0.0.dylib
    dylib_load=/usr/local/opt/glib/lib/libgobject-2.0.0.dylib
    dylib_load=/usr/local/opt/glib/lib/libgthread-2.0.0.dylib
    dylib_load=/usr/local/opt/gtk+3/lib/libgdk-3.0.dylib
    dylib_load=/usr/local/opt/gtk+3/lib/libgtk-3.0.dylib
    dylib_load=/usr/local/opt/json-glib/lib/libjson-glib-1.0.0.dylib
    dylib_load=/usr/local/opt/libarchive/lib/libarchive.13.dylib
    dylib_load=/usr/local/opt/libmagic/lib/libmagic.1.dylib
    dylib_load=/usr/local/opt/pango/lib/libpango-1.0.0.dylib
    dylib_load=/usr/local/opt/pango/lib/libpangocairo-1.0.0.dylib
/usr/local/Cellar/file-roller/3.16.4/libexec/file-roller/rpm2cpio:
    kind=executable
    arch=[x86_64]
    dylib_load=/usr/lib/libSystem.B.dylib
    dylib_load=/usr/local/opt/atk/lib/libatk-1.0.0.dylib
    dylib_load=/usr/local/opt/cairo/lib/libcairo-gobject.2.dylib
    dylib_load=/usr/local/opt/cairo/lib/libcairo.2.dylib
    dylib_load=/usr/local/opt/gdk-pixbuf/lib/libgdk_pixbuf-2.0.0.dylib
    dylib_load=/usr/local/opt/gettext/lib/libintl.8.dylib
    dylib_load=/usr/local/opt/glib/lib/libgio-2.0.0.dylib
    dylib_load=/usr/local/opt/glib/lib/libglib-2.0.0.dylib
    dylib_load=/usr/local/opt/glib/lib/libgobject-2.0.0.dylib
    dylib_load=/usr/local/opt/glib/lib/libgthread-2.0.0.dylib
    dylib_load=/usr/local/opt/gtk+3/lib/libgdk-3.0.dylib
    dylib_load=/usr/local/opt/gtk+3/lib/libgtk-3.0.dylib
    dylib_load=/usr/local/opt/pango/lib/libpango-1.0.0.dylib
    dylib_load=/usr/local/opt/pango/lib/libpangocairo-1.0.0.dylib
```

do you want 
```
  atk
  cairo
  gdk-pixbuf
  gettext
  glib
  pango
```
removed or not? I'm happy either way.

Formula/file-roller.rb
UniqMartin: I'm fine with adding these in this case, but I'd like to see that reflected in the PR body and commit message. And that should probably be a separate commit as it's not immediately related to the patch stuff. Currently that's a change nobody could understand from reading the commit history.

Formula/file-roller.rb
ilovezfs: Since the version test passes with or without it, I figured I'd save people time and brain cycles re-reading that same gibberish every time they look at the CI, but there's no harm in removing the `2>/dev/null` if you'd like.

UniqMartin:
> @UniqMartin Based on `<snip>` do you want `<snip>` removed or not? I'm happy either way.

No need to invoke me by name. :wink: I'm reading and responding (I did a few minutes ago) to code comments, just not as quickly as, e.g., Dom usually does.

Formula/file-roller.rb
DomT4: On balance I think it's probably sensible to not shove it to `/dev/null` in case it spits out a message one day that's actually useful and not just _"Hey, these upstream devs are using laughably outdated methods"_.

ilovezfs: The thought there was that if anything of interest actually occurred, the exit code would be non zero, and then the investigation could ensue, but I'll take it out :)

ilovezfs:
@UniqMartin OK. /* irony */

Formula/file-roller.rb
apjanke: That sounds optimistic about how careful people are with exit statuses. Poor hygiene on this could also mean poor hygiene on return value checking etc. Let's leave it visible. I'd only use the `>/dev/null` squashing if it's to silence output intentionally generated by the tested program itself, and even then, using a more specific `grep` filter might be better.

Formula/file-roller.rb
mikemcquaid: Comment would be great 👍

Formula/file-roller.rb
mikemcquaid: Maybe just redirect to `stdout` to shut it up but spit it out on `--verbose`.

ilovezfs:
PR refreshed with the requested changes. Additional tweak-requests are welcome.

Formula/file-roller.rb
```diff
@@ -22,18 +22,40 @@ class FileRoller < Formula
   depends_on "hicolor-icon-theme"
   depends_on "gnome-icon-theme"
 
+  # Add undeclared dependencies reported by `brew linkage`
```
mikemcquaid: Could maybe just say `add linked library dependencies`

Formula/file-roller.rb
```diff
@@ -22,18 +22,40 @@ class FileRoller < Formula
...
+  depends_on "gettext"
+  depends_on "glib"
+  depends_on "pango"
+
+  # These can be removed when the next version (>3.16.4) is released, since they
+  # are only required to support the patch below, which will no longer be needed
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "yelp-tools" => :build
```
mikemcquaid: 👍

Formula/file-roller.rb
```diff
@@ -22,18 +22,40 @@ class FileRoller < Formula
...
+
+  # Remove when >3.16.4 is released
```
mikemcquaid: As an aside, this made me wonder about a DSL or `audit` or similar that would handle this. Quite often we forget to remove bits like this. Not blocking this PR but just general musings.

ilovezfs: I wonder if wrapping these in `if version == "13.16.4"` and then having audit fail if such a conditional is in the file but version != the RHS literal would work.

mikemcquaid: Yeh; maybe we have some sort of helper method that throws an exception if the version doesn't match the one passed?

ilovezfs: Maybe something like this?

```
diff --git a/Formula/file-roller.rb b/Formula/file-roller.rb
index 9ebe38b..36c41a5 100644
--- a/Formula/file-roller.rb
+++ b/Formula/file-roller.rb
@@ -37,20 +37,19 @@ class FileRoller < Formula
   depends_on "yelp-tools" => :build
 
-  # Remove when >3.16.4 is released
   # Build fails during schema validation when using glib >= 1.45.1
   # Upstream patch for https://bugzilla.gnome.org/show_bug.cgi?id=756607
   patch do
     url "https://git.gnome.org/browse/file-roller/patch/?id=37df6d640817127dbad0f9ba94141cfb1e6a8360"
     sha256 "5dee926832a607fc703518cd35b8fe0846c6bf760faf19982a89fe5f5c21fe75"
+    assert_version "3.16.4"
   end
 
   def install
     ENV.append "CFLAGS", "-I#{Formula["libmagic"].opt_include}"
     ENV.append "LIBS", "-L#{Formula["libmagic"].opt_lib}"
 
-    # Remove when >3.16.4 is released
-    system "autoreconf", "-fiv"
+    assert_version("3.16.4") { system "autoreconf", "-fiv" }
 
     # forces use of gtk3-update-icon-cache instead of gtk-update-icon-cache. No bugreport should
     # be filed for this since it only occurs because Homebrew renames gtk+3's gtk-update-icon-cache

```

```
diff --git a/Library/Homebrew/formula.rb b/Library/Homebrew/formula.rb
index 3e6006b..15e1e9e 100644
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -217,6 +217,19 @@ class Formula
     active_spec == stable
   end
 
+  def assert_version(expected, strict=true, &block)
+    matches = version == Version.new(expected)
+    unless matches
+      if strict
+        raise "verify change(s) still needed for #{version}"
+      else
+        opoo "verify change(s) still needed for #{version}"
+      end
+    end
+    yield if block_given?
+    matches
+  end
+
   # Is the currently active {SoftwareSpec} a {#devel} build?
   # @private
   def devel?
@@ -918,6 +931,11 @@ class Formula
   def patch
     unless patchlist.empty?
       ohai "Patching"
+      patchlist.each do |p|
+        if p.respond_to?(:assert_version) && !p.assert_version.nil? && Version.new(p.assert_version) != version
+          raise "verify change(s) still needed for #{version}"
+        end
+      end
       patchlist.each(&:apply)
     end
   end
diff --git a/Library/Homebrew/resource.rb b/Library/Homebrew/resource.rb
index 3fcdf46..613c952 100644
--- a/Library/Homebrew/resource.rb
+++ b/Library/Homebrew/resource.rb
@@ -185,6 +185,10 @@ class Resource
       @patch_files.concat(paths)
       @patch_files.uniq!
     end
+
+    def assert_version(v=nil)
+      v
+    end
   end
 end
 
```

mikemcquaid: Certainly worth opening to prompt some discussion 👍

UniqMartin: I agree. An interesting idea I'd be interested to see fleshed out further.

UniqMartin:
:+1:

Formula/file-roller.rb
xu-cheng: I can see this is useful not only for patch of certain install method, but for every possible DSL in the formulae(like option, dependency, resources, caveats, test etc). However, it would be hard to declare similar DSL for all of them. So may I suggest a different approach? 

Instead of leveraging ruby DSL, why not use Ruby comment. For example:
```ruby
# VERSION 1.0.0 ONLY 
patch do
  ...
end

# VERSION 1.0.0 ONLY 
depends_on "foo"
```

Then we can use simply grep in `brew audit` to check this. The benefit is we don't need to introduce new DSL. And Ruby comment can pretty much be applied to every possible fields.

ilovezfs: It would be nice not to modify the DSL, but how would you define the scope of where the relevant part ends? Or just continue with educated guessing?

Also how would you "disable" all version-bound sections en masse to see if it builds without them? That's not strictly necessary but more efficient than manually editing them out one by one just to test. That said, often removing them requires additional changes, so probably less useful than it might seem at first blush.

I'm also not thrilled with copying the literal version string all over the place anyway. Maybe a convention that does `# stable restrict 1.2.3` and `# devel restrict 4.5.6` at the top and then at each affected specific location `# stable restrict` or `# devel restrict`.

mikemcquaid: Let's pull this discussion into a PR or another issue.

ilovezfs:
Closing this in favor of the new version.

PR #999, opened 2016-05-09
tinyscheme: migrate to boneyard

woodruffw:

---
- [ ] Merge `mit-scheme` into core first (#1030) so that `conflicts_with` in this formula can be updated.

Tests the output of the `-?` flag.


mikemcquaid:
👍

ilovezfs:
Current version looks like it's actually tinyscheme-1.41 if you'd like to do a version bump here or a separate PR.

woodruffw:
Updated. Do you prefer two commits or one?

woodruffw:
Looks like the makefile needs to be re-patched for 1.41.

ilovezfs:
I just realized that this is the Herman Cain PR.

Yes, I can check it on 10.9

ilovezfs:
It looks like this formula also needs a `conflicts_with` for `mit_scheme` and vice versa, unless its binary is renamed:
```
==> make INITDEST=/usr/local/Cellar/tinyscheme/1.41/share
The formula built, but is not symlinked into /usr/local
Could not symlink bin/scheme
Target /usr/local/bin/scheme
is a symlink belonging to mit-scheme. You can unlink it:
  brew unlink mit-scheme
etc.
```

ilovezfs:
I think the problem is with the build server not your test. I cannot reproduce the issue and the error makes no sense.

woodruffw:
> It looks like this formula also needs a conflicts_with for mit_scheme and vice versa, unless its binary is renamed

Got it, I'll add the appropriate lines in `mit_scheme` and `tinyscheme`.

> I think the problem is with the build server not your test. I cannot reproduce the issue and the error makes no sense.

That's what I was afraid of. Is there a standard procedure from here?

ilovezfs:
This one seems to work
```
assert_match "Usage: tinyscheme", shell_output("#{bin}/scheme -?", 1)
```
so I'd go with that for now.

woodruffw:
Changed it to `-?`, hopefully that does it.

woodruffw:
The `conflicts_with` line for `mit-scheme` causes the test to fair, presumably because the `x11` tap isn't tapped.
Should I remove it? The `mit-scheme` formula already has a `conflicts_with` for `tinyscheme`.

Formula/tinyscheme.rb
```diff
@@ -20,35 +22,38 @@ def install
     share.install("init.scm")
     bin.install("scheme")
   end
+
+  test do
+    assert_match "Usage: tinyscheme", shell_output("#{bin}/scheme -?", 1)
```
UniqMartin: I would expect the `?` needs to be escaped since `shell_output` uses an intermediate shell.

UniqMartin:
> The `conflicts_with` line in `tinyscheme` for `mit-scheme` causes the test to fail, presumably because the `x11` tap isn't tapped. Should I remove it?

Yes. None of our core formulae have a `conflicts_with` line that references a formula from a different tap.

woodruffw:
Got it :+1:

Formula/tinyscheme.rb
UniqMartin: Also, any reason not to use a small Hello World example here like in the other PRs?

woodruffw: The CI seemed to have problems with the `hello.scm` I had originally (non repro on Mavericks), so @ilovezfs advised changing it.

UniqMartin: Sorry, I overlooked that information. It's a bit concerning that something as simple would fail on 10.9 and makes me think the resulting binary could be broken, but I'm not sure how feasible it is to debug this failure. (Though my take is that if a perfectly reasonable test fails, then the required action is to fix<sup>#</sup> the root cause revealed by the test instead of adapting the test to the failure.)

<sup>#</sup>) If possible. Or boneyard the formula or do something else, but not weaken the test.

woodruffw: We're not certain if it's a broken binary, since @ilovezfs couldn't reproduce it locally. The failure message also didn't make any sense (even for a broken binary), so there's at least a good chance that the fault lies somewhere in CI rather than `tinyscheme`. Until resolved either way, it's probably a good idea to hold off on merging this one.

ilovezfs:
I had tested with the supposedly broken bottle and separately from that on 10.8 and 10.9 and 10.11. All were fine. I think it's having problem on Jenkins opening files for some reason, as the output generated was "" there and was correct elsewhere. The history of https://github.com/Homebrew/homebrew-core/pull/1016 on the build server shows the results of various tests.

ilovezfs:
If the conflicts with is being removed, it would be good if this were installed with a name other than scheme to avoid the conflict, since scheme users may have both, and the only reason mit-scheme is in the other tap is no one has bothered to make the x11 dependency optional yet.

woodruffw:
I think it might be better to work on trimming `mit-scheme`'s X11 dependency and getting it into core, as people probably already expect a `scheme` binary from `tinyscheme` and the `conflicts_with` can be used. Plus, renaming it raises the question of which formula provides the canonical `scheme` in Homebrew, something that users should probably be deciding instead.

I'll take a look at `mit-scheme` in a second.

ilovezfs:
@woodruffw yes, trimming the x11 dependency would nice, because mit-scheme really deserves to have a home in core. Also, just a heads up that mit-scheme takes a while to build, so you should cancel a build of it on Jenkins if you force push another commit before a prior one runs to avoid monopolizing the build servers.

woodruffw:
> Also, just a heads up that mit-scheme takes a while to build, so you should cancel a build of it on Jenkins if you force push another commit before a prior one runs to avoid monopolizing the build servers.

Thanks for the tip!

woodruffw:
Has anybody successfully separated `mit-scheme` from X11 before?

Going through the source, I see no normal autoconf flags for disabling it. No luck with `--without-x`, `--disable-xlib`, or `--disable-x11`.

Formula/tinyscheme.rb
```diff
@@ -20,35 +20,38 @@ def install
...
+  test do
+    assert_match "Usage: tinyscheme", shell_output("#{bin}/scheme -\?", 1)
+  end
 end
 
 __END__
---- a/makefile  2011-01-16 20:51:17.000000000 +1300
-+++ b/makefile  2012-04-08 22:38:11.000000000 +1200
+--- a/makefile
++++ b/makefile
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

woodruffw: I sent an email to the upstream maintainers.

Someone else wrote an explanation of the earlier patch on line 14. The changes I've introduced are just due to offset, so do you want me to write a new explanation?

mikemcquaid: Thanks. I'm just generally concerned when we have patches that don't seem to be getting merged upstream.

woodruffw: No problem! I've forwarded the email I sent them to you. It looks like the last activity on `tinyscheme` was about 3 years ago, so the project itself might be dormant.

DomT4: It might be nice to push this into `homebrew/patches` and use the `patch do` blob instead. It's fairly hefty to carry inline, especially indefinitely.

ilovezfs: I'm wondering if this might be preferable to using a patch:
```
    mac = <<-EOS.undent
      SOsuf=dylib
      CC = gcc -fpic -pedantic
      LD = gcc
      LDFLAGS = -dynamiclib
      DEBUG=-g -Wno-char-subscripts -O
      SYS_LIBS= -ldl -lm
      PLATFORM_FEATURES= -DUSE_STRLWR=1 -D__APPLE__=1 -DOSX=1
    EOS
    inreplace "makefile",
     "\nFEATURES = $(PLATFORM_FEATURES)",
     "#{mac}\nFEATURES = $(PLATFORM_FEATURES)"
```

woodruffw: @DomT4 
Moving this to `homebrew/patches` sounds good to me. 

Is the standard operating procedure just to use the `raw.githubusercontent.com` URL?

@ilovezfs I'm hesitant about that approach, mainly because it'll most likely execute successfully but produce the wrong build specifications if the `tinyscheme` maintainers decide to tweak the makefile again. With the diff, the patch will fail early and we won't be stuck debugging compilation/behaviour errors.

ilovezfs: I think the same could be said about just about any use of `inreplace`.

My concern is that this patch is basically just un-commenting the little OS X section but was understandably characterized by @DomT4 as "fairly hefty to carry inline". It's also difficult to see what parts of the patch are the uncommenting of what's already there and what's parts are actually a substantive change and why.

Based on https://gist.github.com/anonymous/5520f144682f37711d95bf3a7cb271d3 I'm tempted to suggest we not even use the makefile and run the commands inline so it's transparent what this actually does, which in its entirety is

```
gcc -fpic -pedantic -I. -c -g -Wno-char-subscripts -O  -DUSE_DL=1 -DUSE_MATH=1 -DUSE_ASCII_NAMES=0 -DOSX -DInitFile="\"/usr/local/Cellar/tinyscheme/1.41/share/init.scm"\"  scheme.c
gcc -fpic -pedantic -I. -c -g -Wno-char-subscripts -O  -DUSE_DL=1 -DUSE_MATH=1 -DUSE_ASCII_NAMES=0 -DOSX -DInitFile="\"/usr/local/Cellar/tinyscheme/1.41/share/init.scm"\"  dynload.c
gcc -shared -o libtinyscheme.dylib scheme.o dynload.o -ldl -lm
ar crs libtinyscheme.a scheme.o dynload.o
gcc -fpic -pedantic -o scheme -g -Wno-char-subscripts -O scheme.o dynload.o -ldl -lm
```

ilovezfs: @woodruffw in any case, go ahead and proceed with opening the pull request in https://github.com/Homebrew/patches/pulls

Here's an example:
https://github.com/Homebrew/patches/pull/11
https://github.com/Homebrew/legacy-homebrew/pull/49827

woodruffw: > I'm tempted to suggest we not even use the makefile and run the commands inline so it's transparent what this actually does

This sounds good to me. You're definitely right in that the patch is fairly opaque, and the compilation steps are so simple in this case that they're hardly worth the added complexity of patching-then-running make.

@DomT4 thoughts?

ilovezfs: @woodruffw what really would be nice if you documented why each of those variables is set as it is and which ones are set as the makefile says they should be for OS X and which ones you needed to change, regardless of whether the final outcome is a patch, heredoc, directly running the commands, whatever.

The one gotcha is that you will need to add  ENV.refurbish_args to the formula if you run the commands inline. When you call "make", superenv does that automatically, but that is not true for inline commands unless you add  ENV.refurbish_args.

DomT4: I don't really want to inline the actual `make` steps. It'll just confuse future contributors and is only an option we ever really deliberately exercise if upstream doesn't provide a make system at all.

I don't have a massively strong preference on whether we use a patch or an `inreplace` but it should be one of those options.

ilovezfs: I'm happy with the patch as long as it's clear what came from the OS X snippet in the existing makefile and what we changed and why in the comments above the `patch do`.

woodruffw: Opened as Homebrew/patches#18.

DomT4: Merged it. Please use the commit reference rather a direct link to `master`; doing so is a small safety step in case we accidentally remove a patch prematurely or accidentally 😄.

woodruffw: > Please use the commit reference rather a direct link to master; doing so is a small safety step in case we accidentally remove a patch prematurely or accidentally :smile:.

Got it :+1:

woodruffw: Ugh, I left git-style a/b information in the patch. I created a PR at Homebrew/patches/pull/19 to fix it.

DomT4: Oh, that's not usually problematic. Have merged your revised PR.

woodruffw: Yeah, I don't know why it didn't crop up during my testing. 
Oh well, it should be working now.

ilovezfs:
@woodruffw The patch is so much clearer now! Nice job. Should that be `--shared` or just `-dynamiclib` instead of their bogus `--dynamiclib`?

woodruffw:
@ilovezfs Thank you!

I think it should technically be `-dynamic`, but the linker doesn't seem to care. Do you think this is worth a change?

ilovezfs:
http://lists.apple.com/archives/xcode-users/2003/Dec/msg00395.html
https://github.com/Homebrew/brew/blob/master/Library/ENV/4.3/cc#L151-L202

So `-dylib` gets special treatment.

After looking at the superenv output, it looks like in this case the ultimate consequence is only an effect on the order of the arguments to ld when you use gcc5, so I'd leave it.

woodruffw:
> So -dylib gets special treatment.

Interesting, thanks for the links.

> After looking at the superenv output, it looks like in this case the ultimate consequence is only an effect on the order of the arguments to ld when you use gcc5, so I'd leave it.

Got it :+1:

woodruffw:
Any reason why `brew readall --aliases --syntax` would fail in the CI?

DomT4:
Yeah, there's a single piece of syntax it doesn't like, the floating `retry_count` on 555 in this commit: https://github.com/Homebrew/brew/commit/b5a44161a4b2af3bf6b08970d3f8fc151ee1796a. I'm just ignoring it for now.

I'm not quite sure how to handle `possibly useless use of a variable in void context` warnings and the internet didn't provide a clear answer, so I've been trying to give Andrew time to look it over.

DomT4:
I suppose could possibly do something like `retry_count = retry_count` but that feels a bit silly.

woodruffw:
> I suppose could possibly do something like retry_count = retry_count but that feels a bit silly.

Yeah, I think that would look a bit strange.

What's the point of the floating `retry_count`?

DomT4:
Looking at the code and the similar usage further up I'm not actually sure. It looks like `retry_count = 0` could be within the `begin` block if it has to be there in some form rather than above, since you can't get to the `rescue` before the `begin` so it should always be defined in that situation. 

I haven't really tracked all the changes to the `pull` command lately as I've not had as much time to be around Homebrew, so I'm starting from a point of understanding further back than I probably need to be to be honest.

DomT4:
Went for the questionable syntax hack: https://github.com/Homebrew/brew/commit/745a1312dce936ddd519c2e151e889459243c543.

I think `retry_count = 0` _inside_ `begin` is reasonable but I try not to tweak other maintainers' changes more than cosmetically as a point of principle, which is why I've let this fail every CI job for half a day before getting fed up with it 😅.

ilovezfs:
The original tests that fail on the Mavericks build server still fail, but if I add `fails_with :clang` they pass:
https://github.com/Homebrew/homebrew-core/pull/1016

So even though the supposedly broken bottles seem to work fine when I test them, I think we may want to add `fails_with :clang` until the segfault is fully understood.

mikemcquaid:
Given the patches that aren't going upstream and tag failures I'd be tempted to boneyard it.

woodruffw:
> I'd be tempted to boneyard it.

Works for me; I haven't gotten a response from the maintainers.

apjanke:
FWIW: that `retry_count` was in fact a bug, and I suspect any similar `possibly useless use of a variable in void context` would be as well. I think it's fine to consider those to be errors, the way we are now, and require them to be fixed.

> What's the point of the floating retry_count?

Silly copy-and-paste error. It did nothing, and shouldn't have been there.

Thanks for going with the minimal workaround, DomT4. And it's fixed cleanly now.

mikemcquaid:
> I did some quick research on TinyScheme: Apparently it's very popular as an embedded language, used by GIMP and a few other large projects for scripting. That may explain why it's basically inactive except for major (infrequent) releases. That alone might not be enough to justify not sending it to the boneyard, but it's probably worth considering.

@woodruffw If we don't have anything that depends on it: boneyard feels best.

woodruffw:
> If we don't have anything that depends on it: boneyard feels best.

Got it, I'll move it tonight :+1:

woodruffw:
This PR now removes `tinyscheme` from `core`.

ilovezfs:
@woodruffw You'll also want to update `tap_migrations.json` in this PR.

ilovezfs:
@woodruffw Also, you can remove
```
 conflicts_with "tinyscheme", :because => "both install a `scheme` binary"
```
from `mit-scheme` in this PR or in a separate one, now that mit-scheme has found its way back to core after a long eX11e.

woodruffw:
@ilovezfs That should do it!

ilovezfs:
@woodruffw 🎉 the *real* scheme is back! Awesome.

PR #1001, opened 2016-05-09
scheme48: Add test.

woodruffw:

---

Tests the output of a trivial Scheme program.

### DIFF
--- a/Formula/scheme48.rb
+++ b/Formula/scheme48.rb
@@ -21,4 +21,19 @@ class Scheme48 < Formula
     system "make"
     system "make", "install"
   end
+
+  test do
+    (testpath/"hello.scm").write <<-EOS.undent
+      (display "Hello, World!") (newline)
+    EOS
+
+    expected = <<-EOS.undent
+      Hello, World!\#{Unspecific}
+
+      \#{Unspecific}
+
+    EOS
+
+    assert_equal expected, shell_output("#{bin}/scheme48 -a batch < hello.scm")
+  end
 end


Formula/scheme48.rb
```diff
@@ -21,4 +21,12 @@ def install
     system "make"
     system "make", "install"
   end
+
+  test do
+    (testpath/"hello.scm").write <<-EOS.undent
+      (display "Hello, World!") (newline)
+    EOS
+
+    assert_equal "Hello, World!\#{Unspecific}\n\n\#{Unspecific}\n\n", shell_output("#{bin}/scheme48 -a batch < hello.scm")
```
mikemcquaid: Would probably use `EOS.undent` rather than embedding newlines here.

Formula/scheme48.rb
```diff
@@ -21,4 +21,19 @@ def install
...
+
+  test do
+    (testpath/"hello.scm").write <<-EOS.undent
+      (display "Hello, World!") (newline)
+    EOS
+
+    expected = <<-EOS.undent
+      Hello, World!\#{Unspecific}
+
+      \#{Unspecific}
```
UniqMartin: What's the meaning behind this and the above `\#{Unspecific}` output? And what's going on with the many newlines?

woodruffw: This particular interpreter doesn't have a file input argument, so it has to be fed lines of Scheme one-by-one in batch mode. AFAICT, the superfluous newlines and `#{Unspecifc}`s are the product of that one-by-one interpretation.

The other option is to generate an image file and load the interpreter with that, but that's a 3-4 step process compares to the two steps here (write file, execute).

UniqMartin: Makes sense. Thank you for the explanation!

UniqMartin:
Merged. Thank you for yet another contribution to Homebrew, @woodruffw! :tada:

Formula/scheme48.rb
woodruffw: No problem!

PR #1002, opened 2016-05-09
zzuf 0.15

ilovezfs:
- bump the version
- remove the patches since the underlying issues are now fixed
- no longer conflicts with libzzip: "zzcat" was renamed "zzat"
- fix the test:
  (1) "hd" doesn't exist on OS X
  (2) zzuf doesn't work on OS X when invoking other commands directly
      even if "hd" were to exist (e.g., substituting "hexdump")


mikemcquaid:
👍

PR #1004, opened 2016-05-09
dssim 1.3.2 (new formula)

swernerx:

---

Added Kornels `dssim` tool to compare RGBA structural similarity of different images / compressions.

### DIFF
--- a/Formula/dssim.rb
+++ b/Formula/dssim.rb
@@ -13,6 +13,6 @@ class Dssim < Formula
   end
 
   test do
-    system "false"
+    system "#{bin}/dssim", test_fixtures("test.png"), test_fixtures("test.png")
   end
 end
--- a/Formula/dssim.rb
+++ b/Formula/dssim.rb
@@ -1,8 +1,8 @@
 class Dssim < Formula
   desc "RGBA Structural Similarity C implementation (with a Rust API)"
   homepage "https://github.com/pornel/dssim"
-  url "https://github.com/pornel/dssim/archive/1.1.tar.gz"
-  sha256 "bad5bde3846e499be16e551248168b98195ca1e5893f948739ecc0cf2e8d52f6"
+  url "https://github.com/pornel/dssim/archive/1.3.2.zip"
+  sha256 "393a5431f08c747ac19b00176f2b710010d87215aa6f5105e65e9fa7afed63be"
 
   depends_on "pkg-config" => :build
   depends_on "libpng"
--- a/Formula/dssim.rb
+++ b/Formula/dssim.rb
@@ -1,0 +1,18 @@
+class Dssim < Formula
+  desc "RGBA Structural Similarity C implementation (with a Rust API)"
+  homepage "https://github.com/pornel/dssim"
+  url "https://github.com/pornel/dssim/archive/1.1.tar.gz"
+  sha256 "bad5bde3846e499be16e551248168b98195ca1e5893f948739ecc0cf2e8d52f6"
+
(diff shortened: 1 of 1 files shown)


Formula/dssim.rb
```diff
@@ -0,0 +1,18 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "libpng"
+
+  def install
+    system "make"
+    bin.install Dir["bin/*"]
+  end
+
+  test do
+    system "false"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Formula/dssim.rb
```diff
@@ -0,0 +1,18 @@
...
+  homepage "https://github.com/pornel/dssim"
+  url "https://github.com/pornel/dssim/archive/1.1.tar.gz"
+  sha256 "bad5bde3846e499be16e551248168b98195ca1e5893f948739ecc0cf2e8d52f6"
+
+  depends_on "pkg-config" => :build
+  depends_on "libpng"
+
+  def install
+    system "make"
+    bin.install Dir["bin/*"]
```
mikemcquaid: Is there no `make install` we can use? If not, could you file an issue asking upstream for one?

swernerx: There is no "install" target in the makefile of that tool. What do you mean by asking "upstream"? It's really just a single binary.

Formula/dssim.rb
swernerx: I don't think there is a lot we can do here. For check whether it's working we would need two images to compare as far as I understand. I am not the author of the tool - I just want to use it :)

Formula/dssim.rb
mikemcquaid: I mean file an issue on `https://github.com/pornel/dssim` asking requesting (or, ideally, adding) a `make install` target so if more files should be installed in future we don't miss them. Also, if it's currently a single binary then `bin.install "bin/binary"` is preferable to globbing.

Formula/dssim.rb
mikemcquaid: We can check two images. Check out https://github.com/Homebrew/homebrew-core/blob/4c37b28a39c9cd066edb7024e9e75ab8d22ab7d4/Formula/imagemagick.rb#L128. A test is a requirement to merge this formula, unfortunately.

ilovezfs:
https://github.com/pornel/dssim/blob/master/README.md

> On OS X make USE_COCOA=1 will compile without libpng.

You should probably make that the default, and then offer libpng as an optional dependency:
```
depends_on "libpng" => :optional
```

And then skip the `USE_COCOA=1` when it's `build.with? "libpng"`

swernerx:
I tried first using without libpng without success. Therefor I used it as a normal dependency.

UniqMartin:
Mostly looking good, but please address the following two audit issues:

```
dssim:
  * Stable: Use GitHub tarballs rather than zipballs (url is https://github.com/pornel/dssim/archive/1.3.2.zip).
  * Dir["bin/dssim"] is unnecessary; just use "bin/dssim"
```

(And as mentioned in the checklist in the first post, don't forget that you can and should check locally whether you have resolved those issues by running `brew audit --strict --online dssim`.)

swernerx:
Fine. I changed the part and also checked using the audit tool.

mikemcquaid:

PR #1008, opened 2016-05-09
offlineimap: fix audit errors

hvoigt:

---

This is a preparation for a small fix/extension I will send in a seperate pull request on top of this.

Not really sure if startup => true is the correct choice since the documentation here: http://www.rubydoc.info/github/Homebrew/brew/Formula.plist_options is not that extensive.

### DIFF
--- a/Formula/offlineimap.rb
+++ b/Formula/offlineimap.rb
@@ -24,6 +24,8 @@ class Offlineimap < Formula
     EOS
   end
 
+  plist_options :startup => true
+
   def plist; <<-EOS.undent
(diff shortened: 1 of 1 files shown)


UniqMartin:
This PR seems to be redundant as its changes are also contained in #1009. I suppose that's not how you planned things, but that's how things turned out. Thus closing in favor of #1009.

hvoigt:
I wanted to keep those two seperate so this could be merged independently since this PR only contains fixes and the other one has a feature extension on top. So I thought this would be reviewed/merged first and then the other (which would then fast forward since they are the same commits). But whatever works better for you is fine with me.

PR #1009, opened 2016-05-09
offlineimap: fix audit errors and plist issues

hvoigt:

---

This is based on a seperate pull request fixing the existing audit errors: https://github.com/Homebrew/homebrew-core/pull/1008 which needs to be reviewed and merged first.

### DIFF
--- a/Formula/offlineimap.rb
+++ b/Formula/offlineimap.rb
@@ -31,6 +31,11 @@ class Offlineimap < Formula
     <!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
     <plist version="1.0">
       <dict>
+        <key>EnvironmentVariables</key>
+        <dict>
+          <key>PATH</key>
+          <string>/usr/bin:/bin:/usr/sbin:/sbin:#{HOMEBREW_PREFIX}/bin</string>
+        </dict>
         <key>KeepAlive</key>
         <false/>
         <key>Label</key>
--- a/Formula/offlineimap.rb
+++ b/Formula/offlineimap.rb
@@ -24,6 +24,8 @@ class Offlineimap < Formula
     EOS
   end
 
+  plist_options :manual => "offlineimap"
+
   def plist; <<-EOS.undent
     <?xml version="1.0" encoding="UTF-8"?>
     <!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
--- a/Formula/offlineimap.rb
+++ b/Formula/offlineimap.rb
@@ -24,10 +24,6 @@ class Offlineimap < Formula
     EOS
   end
 
-  test do
-    system bin/"offlineimap", "--version"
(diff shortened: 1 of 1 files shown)


Formula/offlineimap.rb
```diff
@@ -24,15 +24,18 @@ def caveats; <<-EOS.undent
...
 
   def plist; <<-EOS.undent
     <?xml version="1.0" encoding="UTF-8"?>
     <!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
     <plist version="1.0">
       <dict>
+        <key>EnvironmentVariables</key>
+        <dict>
+          <key>PATH</key>
+          <string>/usr/bin:/usr/local/bin</string>
```
mikemcquaid: Are these not there by default?

Formula/offlineimap.rb
```diff
@@ -24,15 +24,18 @@ def caveats; <<-EOS.undent
     EOS
   end
 
-  test do
-    system bin/"offlineimap", "--version"
-  end
+  plist_options :startup => true
```
mikemcquaid: Needs a `:manual` bit and I'm not sure this needs to be `:startup`?

Formula/offlineimap.rb
hvoigt: No, /usr/local/bin is not there by default.

Formula/offlineimap.rb
hvoigt: What does manual actually mean? I figured startup means that the service should be started upon login if configured by the user and manual describes some manual method the user can use to manually start the service. What command should go into the manual part just the same as the one in the plist without the full path?

Formula/offlineimap.rb
mikemcquaid: Ok. You probably want `#{HOMEBREW_PREFIX}/bin` or `#{HOMEBREW_PREFIX}/opt/offlineimap/bin`. Was it broken before?

hvoigt: It was not broken, but as explained in the commit message it was inconvenient that you could not call homebrew tools from offlineimap without full path. E.g. notmuch which indexes all messages for quicker searching.

Formula/offlineimap.rb
hvoigt: The only documentation I found for plist_options is [this](http://www.rubydoc.info/github/Homebrew/brew/Formula.plist_options). But it does not really explain this option for me.

Formula/offlineimap.rb
mikemcquaid: Ok, that's cool, thanks.

Formula/offlineimap.rb
mikemcquaid: Yeh, that's the best docs we have for now (but we should improve them). In this case it doesn't need `:startup` and should specify a `:manual` on how to run this outside `launchd`

hvoigt:
Ok pushed another version with the requested changes incorporated. I additionally checked what the default path set by launchd is (/usr/bin:/bin:/usr/sbin:/sbin) and added those as well so we do not break any existing use cases of others.

mikemcquaid:

PR #1013, opened 2016-05-09
zpaq 7.13 (devel)

ilovezfs:
- devel spec for the latest version 
- zpaq >=7.06 does not yet have an OS X compatible Makefile 
- use upstream's test case

### DIFF
--- a/Formula/zpaq.rb
+++ b/Formula/zpaq.rb
@@ -1,11 +1,20 @@
 class Zpaq < Formula
   desc "Incremental, journaling command-line archiver"
   homepage "http://mattmahoney.net/dc/zpaq.html"
-  url "http://mattmahoney.net/dc/zpaq705.zip"
-  sha256 "d8abe3e3620d4c6f3ddc1da149acffa4c24296fd9c74c9d7b62319e308b63334"
-  version "7.05"
-
+  revision 1
   head "https://github.com/zpaq/zpaq.git"
+
+  stable do
+    url "http://mattmahoney.net/dc/zpaq705.zip"
+    sha256 "d8abe3e3620d4c6f3ddc1da149acffa4c24296fd9c74c9d7b62319e308b63334"
+    version "7.05"
+
+    # Should be removed once >7.05 ships as stable
+    resource "backport_makefile" do
+      url "http://mattmahoney.net/dc/zpaq713.zip"
+      sha256 "9120cf4fb1afdecea3ac4f690d7b0577f7cb004ca6b152856edd8ac444f0d919"
+    end
+  end
 
   bottle do
     cellar :any_skip_relocation
@@ -15,20 +24,53 @@ class Zpaq < Formula
     sha256 "a2d2215b51cc18370f5ebe700160d1d7143ce5d680b2365a5adb4bb1622d06fc" => :mountain_lion
   end
 
-  def install
-    system "make"
-    include.install "libzpaq.h"
-    bin.install "zpaq"
+  devel do
+    url "http://mattmahoney.net/dc/zpaq713.zip"
+    sha256 "9120cf4fb1afdecea3ac4f690d7b0577f7cb004ca6b152856edd8ac444f0d919"
+    version "7.13"
+  end
 
-    system "pod2man", "zpaq.pod", "zpaq.1"
-    man1.install "zpaq.1"
+  resource "test" do
+    url "http://mattmahoney.net/dc/calgarytest2.zpaq"
+    sha256 "b110688939477bbe62263faff1ce488872c68c0352aa8e55779346f1bd1ed07e"
+  end
+
+  def install
+    # Should be removed once >7.05 ships as stable
+    if build.stable?
+      resource("backport_makefile").stage do
+        buildpath.install "Makefile"
+      end
+    end
+
+    # Makefile introduced for >7.05 has not yet been adapted for OS X
+    # Reported 12th May 2016 to mattmahoneyfl@gmail.com
+    inreplace "Makefile" do |s|
+      # Use OS X style dylib names
+      s.gsub! "libzpaq.so.0.1", "libzpaq.0.1.dylib"
+      s.gsub! "libzpaq.so", "libzpaq.dylib"
+
+      # ld: unknown option: -soname
+      # ld: file not found: libzpaq.0.1.dylib
+      # use the correct arguments to compile a dylib
+      s.gsub! "-Wl,-soname,$(SONAME) -o $@ $<",
+              "-Wl -dynamiclib libzpaq.o zpaq.o -o $(SONAME)"
+
+      # clang: error: no such file or directory: 'zpaq.o'
+      # libzpaq.0.1.dylib needs both libzpaq.o and zpaq.o
+      s.gsub! "$(SONAME): libzpaq.o", "$(SONAME): libzpaq.o zpaq.o"
+
+      # OS X `install` command doesn't have `-t`
+      s.gsub! /(install -m.* )-t (.*) (.*)(\r)/, "\\1 \\3 \\2\\4"
+    end
+
+    system "make"
+    system "make", "check"
+    system "make", "install", "PREFIX=#{prefix}"
   end
 
   test do
-    archive = testpath/"test.zpaq"
-    zpaq = bin/"zpaq"
-    system zpaq, "a", archive, "#{include}/libzpaq.h"
-    system zpaq, "l", archive
-    assert_equal "7kSt", archive.read(4)
+    resource("test").stage testpath
+    assert_match /all OK/, shell_output("#{bin}/zpaq x calgarytest2.zpaq 2>&1")
   end
 end


Formula/zpaq.rb
```diff
@@ -1,27 +1,49 @@
...
   head "https://github.com/zpaq/zpaq.git"
 
   bottle do
-    cellar :any_skip_relocation
-    sha256 "f0998826effe4dae407d041c4143a39b3645dc46a82456b4e86ac88f106babd3" => :el_capitan
-    sha256 "bdd4716595b8eed1c52d2a2ef43abb095a996d05a0489f81f0b22ae43a26d51d" => :yosemite
-    sha256 "a71ad4808da8c69844f72322ff7633b895d034028129b30c3d25409d91bd4213" => :mavericks
-    sha256 "a2d2215b51cc18370f5ebe700160d1d7143ce5d680b2365a5adb4bb1622d06fc" => :mountain_lion
+    cellar :any
+    sha256 "c75fda6498ab5c8535654bd1f8057b6fec6717bf62daab743e5fb0142bead204" => :el_capitan
```
mikemcquaid: What's with the `bottle` changes?

Formula/zpaq.rb
```diff
@@ -1,27 +1,49 @@
...
   end
 
   def install
-    system "make"
-    include.install "libzpaq.h"
-    bin.install "zpaq"
+    # Makefile introduced for zpaq >= 7.06 was never adapted for OS X
+    # Reported 9th May 2016: https://github.com/zpaq/zpaq/pull/13
```
mikemcquaid: Is it possible to just apply this as a `.patch`?

Formula/zpaq.rb
ilovezfs: PEBKAC. test-bot bit me again.

mikemcquaid: No worries.

Formula/zpaq.rb
ilovezfs: It is. It's just adding the same file generated after these replacements. Including the two "unnecessary" ones that are commented out. But it would need to go in patches repo since apparently the `head` url in this formula only points to a mirror, and they already closed the PR: https://github.com/zpaq/zpaq/pull/13#issuecomment-217909974

I'm wondering if the head url should be removed if it's just a mirror, but upstream does link to it on his site http://mattmahoney.net/dc/zpaq.html where it's described, as follows: "Github Repository maintained by Evan Nemerson and Yonggang Luo. Includes history and convert script."

I suppose I'll need to email upstream directly.

mikemcquaid: Gotcha 👍

Formula/zpaq.rb
```diff
@@ -1,27 +1,49 @@
...
+      s.gsub! "$(SONAME): libzpaq.o", "$(SONAME): libzpaq.o zpaq.o"
 
-    system "pod2man", "zpaq.pod", "zpaq.1"
-    man1.install "zpaq.1"
+      # OS X `install` command doesn't have `-t`
+      s.gsub! /(install -m.* )-t (.*) (.*)\r/, "\\1 \\3 \\2"
+
+      # Repair the corruption the prior replace just caused to the line endings
+      # Not necessary but may be useful
+      # s.gsub! /(.*)\r/, "\\1"
```
ilovezfs: I think this is working around either a bug in `s.gsub! /(install -m.* )-t (.*) (.*)\r/, "\\1 \\3 \\2"` or a bug in `inreplace ... do` or both. Not sure what gives, but without this, the file has `^M` all over (except on the lines with the replacement, which are only OK because I through in the `\r`).

UniqMartin: Not a bug in `inreplace`, but a mistake in the preceding regular expression. The file being edited seems to have CRLF line endings, thus using `\r` as an end of line anchor in the preceding regular expression works. But because the `\r` is not included in the replacement pattern, lines affected by the substitution now only end in LF.

Most editors (and maybe `make`, too) are just fine with either all lines ending in CRLF or ending in LF, but you're creating a mix of CRLF and LF and so the interpretation ends up being that the file has LF line endings, causing the CR characters that are still present in some (most) lines to become visible.

Replacing

```ruby
s.gsub! /(install -m.* )-t (.*) (.*)\r/, "\\1 \\3 \\2"
```

with (or something similar)

```ruby
s.gsub! /(install -m.* )-t (.*) (.*)(\r)/, "\\1 \\3 \\2\\4"
```

should do the trick. (I'm not sure though, why the last two install arguments are reversed.) There's definitely some room for improvement left here …

ilovezfs: Thanks for explaining why that's happening.

They're reversed because of the difference between OS X's install command and GNU's. The simplest fix is to depend on coreutils and prepend the gnubin directory to the PATH (or just add a wrapper in the PATH for the ginstall command only – makes no difference). The point of the replacement is to avoid the additional dependency and to avoid manual DSL foo.install methods in lieu of the Makefile.

ilovezfs:
Changes since last review:
- adopted Martin's regex
- modified the test to use the upstream test case
- switched 7.13 to be a devel spec because the homepage still refers to 7.05 as the latest stable version
- reported the Makefile patch to the upstream developer himself (instead of the PR I'd opened on GitHub mirror)

In addition, the following diff would allow the new Makefile to be used with 7.05, if that's preferred to the old "manual" installation method that the formula has been using for 7.05.
```
diff --git a/Formula/zpaq.rb b/Formula/zpaq.rb
index 8727b98..6c603ec 100644
--- a/Formula/zpaq.rb
+++ b/Formula/zpaq.rb
@@ -1,12 +1,17 @@
 class Zpaq < Formula
   desc "Incremental, journaling command-line archiver"
   homepage "http://mattmahoney.net/dc/zpaq.html"
+  revision 1
   head "https://github.com/zpaq/zpaq.git"
 
   stable do
     url "http://mattmahoney.net/dc/zpaq705.zip"
     sha256 "d8abe3e3620d4c6f3ddc1da149acffa4c24296fd9c74c9d7b62319e308b63334"
     version "7.05"
+    resource("backport_makefile") do
+      url "http://mattmahoney.net/dc/zpaq713.zip"
+      sha256 "9120cf4fb1afdecea3ac4f690d7b0577f7cb004ca6b152856edd8ac444f0d919"
+    end
   end
 
   bottle do
@@ -30,35 +35,35 @@ class Zpaq < Formula
 
   def install
     if build.stable?
-      system "make"
-      include.install "libzpaq.h"
-      bin.install "zpaq"
-    else
-      # Makefile introduced for zpaq >= 7.06 has not yet been adapted for OS X
-      # Reported 12th May 2016 to mattmahoneyfl@gmail.com
-      inreplace "Makefile" do |s|
-        # Use OS X style dylib names
-        s.gsub! "libzpaq.so.0.1", "libzpaq.0.1.dylib"
-        s.gsub! "libzpaq.so", "libzpaq.dylib"
+      resource("backport_makefile").stage do
+        buildpath.install "Makefile"
+      end
+    end
 
-        # ld: unknown option: -soname
-        # ld: file not found: libzpaq.0.1.dylib
-        # use the correct arguments to compile a dylib
-        s.gsub! "-Wl,-soname,$(SONAME) -o $@ $<",
-                "-Wl -dynamiclib libzpaq.o zpaq.o -o $(SONAME)"
+    # Makefile introduced for zpaq >= 7.06 has not yet been adapted for OS X
+    # Reported 12th May 2016 to mattmahoneyfl@gmail.com
+    inreplace "Makefile" do |s|
+      # Use OS X style dylib names
+      s.gsub! "libzpaq.so.0.1", "libzpaq.0.1.dylib"
+      s.gsub! "libzpaq.so", "libzpaq.dylib"
 
-        # clang: error: no such file or directory: 'zpaq.o'
-        # libzpaq.0.1.dylib needs both libzpaq.o and zpaq.o
-        s.gsub! "$(SONAME): libzpaq.o", "$(SONAME): libzpaq.o zpaq.o"
+      # ld: unknown option: -soname
+      # ld: file not found: libzpaq.0.1.dylib
+      # use the correct arguments to compile a dylib
+      s.gsub! "-Wl,-soname,$(SONAME) -o $@ $<",
+              "-Wl -dynamiclib libzpaq.o zpaq.o -o $(SONAME)"
 
-        # OS X `install` command doesn't have `-t`
-        s.gsub! /(install -m.* )-t (.*) (.*)(\r)/, "\\1 \\3 \\2\\4"
-      end
+      # clang: error: no such file or directory: 'zpaq.o'
+      # libzpaq.0.1.dylib needs both libzpaq.o and zpaq.o
+      s.gsub! "$(SONAME): libzpaq.o", "$(SONAME): libzpaq.o zpaq.o"
 
-      system "make"
-      system "make", "check"
-      system "make", "install", "PREFIX=#{prefix}"
+      # OS X `install` command doesn't have `-t`
+      s.gsub! /(install -m.* )-t (.*) (.*)(\r)/, "\\1 \\3 \\2\\4"
     end
+
+    system "make"
+    system "make", "check"
   end
 
   test do

```

mikemcquaid:
Yeh, may as well fix the old version too 👍

Formula/zpaq.rb
```diff
@@ -15,20 +17,52 @@ class Zpaq < Formula
     sha256 "a2d2215b51cc18370f5ebe700160d1d7143ce5d680b2365a5adb4bb1622d06fc" => :mountain_lion
   end
 
+  devel do
+    url "http://mattmahoney.net/dc/zpaq713.zip"
+    sha256 "9120cf4fb1afdecea3ac4f690d7b0577f7cb004ca6b152856edd8ac444f0d919"
+    version "7.13"
+  end
+
+  resource("test") do
```
DomT4: Style nit but we don't normally paren-wrap resources at this point. I'd prefer to keep that consistent unless you feel like changing them all :trollface:.

ilovezfs: A couple of other miscreants:
```
mr.rb:13:  resource("test-repo") do
svg2pdf.rb:18:  resource("svg.svg") do
```

DomT4: Bleh 🙈. I'll be looking for those the next time they come through a PR 😉.

ilovezfs: You'll be waiting a while on svg2pdf I think:
```
svg2pdf-0.1.3.tar.gz	2005-08-04 18:27 	77K	
```
LOL

DomT4: I'll, er, live in hope, heh.

PR #1030, opened 2016-05-10
mit-scheme: migrate from x11

woodruffw:

---

Moves `mit-scheme` from the x11 tap to core with modifications for disabling the X11 dependency.

Corresponding PR in homebrew-x11: Homebrew/homebrew-x11#208

cc @UniqMartin @ilovezfs

### DIFF
--- a/Formula/mit-scheme.rb
+++ b/Formula/mit-scheme.rb
@@ -1,0 +1,116 @@
+class MitScheme < Formula
+  desc "MIT/GNU Scheme development tools and runtime library"
+  homepage "https://www.gnu.org/software/mit-scheme/"
+  url "http://ftpmirror.gnu.org/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
+  mirror "https://ftp.gnu.org/gnu/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
+  sha256 "4f6a16f9c7d4b4b7bb3aa53ef523cad39b54ae1eaa3ab3205930b6a87759b170"
+  revision 1
+
+  bottle do
+    revision 1
+    sha256 "1a449d9c1f1043c1a557962332c7ee18664f5f15ea9eca8e754c0c3cb3c4aef4" => :el_capitan
+    sha256 "7966d841f92346a68ee088b0e25394e541b690aafbb306e27b352ee35a1e24c7" => :yosemite
+    sha256 "a782017f01f5c385e343c802a33cd634b1cfd89ec7eed94a43d753034eaa15ab" => :mavericks
+  end
+
+  conflicts_with "tinyscheme", :because => "both install a `scheme` binary"
+
+  # Has a hardcoded compile check for /Applications/Xcode.app
+  # Dies on "configure: error: SIZEOF_CHAR is not 1" without Xcode.
+  # https://github.com/Homebrew/homebrew-x11/issues/103#issuecomment-125014423
+  depends_on :xcode => :build
+  depends_on "openssl"
+  depends_on :x11 => :optional
+
+  def install
+    # Setting -march=native, which is what --build-from-source does, can fail
+    # with the error "the object ..., passed as the second argument to apply, is
+    # not the correct type." Only Haswell and above appear to be impacted.
+    # Reported 23rd Apr 2016: https://savannah.gnu.org/bugs/index.php?47767
+    # Note that `unless build.bottle?` avoids overriding --bottle-arch=[...].
+    ENV["HOMEBREW_OPTFLAGS"] = "-march=#{Hardware.oldest_cpu}" unless build.bottle?
+
+    # The build breaks __HORRIBLY__ with parallel make -- one target will
+    # erase something before another target gets it, so it's easier to change
+    # the environment than to change_make_var, because there are Makefiles
+    # littered everywhere
+    ENV.deparallelize
+
+    # Liarc builds must launch within the src dir, not using the top-level
+    # Makefile
+    cd "src"
+
+    # Take care of some hard-coded paths
+    %w[
+      6001/edextra.scm
+      6001/floppy.scm
+      compiler/etc/disload.scm
+      edwin/techinfo.scm
+      edwin/unix.scm
+      swat/c/tk3.2-custom/Makefile
+      swat/c/tk3.2-custom/tcl/Makefile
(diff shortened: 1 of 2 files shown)


ilovezfs:
Yup this worked for me. Without this change it'll build with or without X11, and link against X11 when you have it, and not link against it when you don't have it. With this change, even if you have X11, it will not link against X11 unless you ask it to. Nice job.

If you want to keep going with improving the `mit-scheme` formula, the next thing to do is to find a way to eliminate the Xcode requirement and get it building with just the CLT.

woodruffw:
> If you want to keep going with improving the mit-scheme formula, the next thing to do is to find a way to eliminate the Xcode requirement and get it building with just the CLT.

Definitely in the future; for now I just want to get it into core so that I can square away `tinyscheme`.

ilovezfs:
Sounds good to me.

woodruffw:
The CI is going to continue to fail the audit until Homebrew/homebrew-x11#208 is closed, so that half of the process should probably be performed first.

ilovezfs:
Yeah, we just need to see if anyone is going to howl that the bottled version won't have X support. I suspect not, but you never know.

woodruffw:
Sounds good to me :+1:

Formula/mit-scheme.rb
```diff
@@ -0,0 +1,119 @@
...
+      # Fixes "configure: error: No MacOSX SDK for version: 10.10"
+      # Reported 23rd Apr 2016: https://savannah.gnu.org/bugs/index.php?47769
+      s.gsub! /SDK=MacOSX\${MACOSX}$/, "SDK=MacOSX#{MacOS.sdk.version}"
+    end
+
+    if build.without? "x11"
+      inreplace "etc/make-liarc.sh" do |s|
+        s.gsub! "run_configure", "run_configure --without-x"
+      end
+    end
```
mikemcquaid: Can you please submit these patches to the upstream developers of this project and add a link to the upstream patch submission in a comment in the formula file. Thanks!

Formula/mit-scheme.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    # The configure script will add "-isysroot" to CPPFLAGS, so it didn't
+    # check .h here by default even Homebrew is installed in /usr/local. This
+    # breaks things when gdbm or other optional dependencies was installed
+    # using Homebrew
+    ENV.prepend "CPPFLAGS", "-I#{HOMEBREW_PREFIX}/include"
```
mikemcquaid: This should instead reference the include directories of individual dependencies that are needed.

woodruffw: I'm having some trouble figuring out what those individual dependencies are. The comment above mentions `gdbm`, but it's not specified in the formula itself. The only dependencies actually mentioned in the formula don't have `include`s in `/usr/local`, so I'm not sure this line is actually performing anything.

mikemcquaid: Ok, probably worth just removing it then and seeing what breaks

woodruffw: Yep, doing that right now :+1:

Formula/mit-scheme.rb
woodruffw: Sure :+1:

Formula/mit-scheme.rb
woodruffw: I built `mit-scheme` successfully locally without this line.

Looks like it was introduced in Homebrew/homebrew-x11@9506f453 as part of Homebrew/homebrew-x11#84. I'm going to ping the committer below.

woodruffw:
@L2G Do you remember the circumstances that required this addition in `mit-scheme`?

```ruby
    # The configure script will add "-isysroot" to CPPFLAGS, so it didn't
    # check .h here by default even Homebrew is installed in /usr/local. This
    # breaks things when gdbm or other optional dependencies was installed
    # using Homebrew
    ENV.prepend "CPPFLAGS", "-I#{HOMEBREW_PREFIX}/include"
```

I've been testing a few builds locally, and haven't been able to replicate it. Thanks!

ilovezfs:
@woodruffw he's actually not the original author of that addition, which was there in the original:
https://github.com/Homebrew/legacy-homebrew/commit/3048e2f249f7768f706f722379303d1d73ee2e3a

The bug that it was fixing seems to be a section that has changed significantly between 9.0.1 and 9.2:
```
     if test "${MACOSX}" = 10.4; then
       SDK=MacOSX${MACOSX}u
     else
       SDK=MacOSX${MACOSX}
     fi
-    MACOSX_SYSROOT=/Developer/SDKs/${SDK}.sdk
-    if test ! -d "${MACOSX_SYSROOT}"; then
-      { { echo "$as_me:$LINENO: error: No MacOSX SDK for version: ${MACOSX}" >&5
-echo "$as_me: error: No MacOSX SDK for version: ${MACOSX}" >&2;}
-   { (exit 1); exit 1; }; }
+    MACOSX_SYSROOT=
+    for dir in /Developer/SDKs \
+      /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs
+    do
+      if test -d "${dir}"; then
+          MACOSX_SYSROOT=${dir}
+          break
+      fi
+    done
+    if test -z "${MACOSX_SYSROOT}"; then
+      { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: Can't find SDKs directory, not checking for SDK" >&5
+$as_echo "$as_me: WARNING: Can't find SDKs directory, not checking for SDK" >&2;}
+    else
+      MACOSX_SYSROOT=${MACOSX_SYSROOT}/${SDK}.sdk
+      if test ! -d "${MACOSX_SYSROOT}"; then
+          as_fn_error $? "No MacOSX SDK for version: ${MACOSX}" "$LINENO" 5
+      fi
     fi
     MACOSX_CFLAGS="${MACOSX_CFLAGS} -isysroot ${MACOSX_SYSROOT}"
```
So if removing it doesn't seem to be breaking anything, I wouldn't worry about it.

woodruffw:
> So if removing it doesn't seem to be breaking anything, I wouldn't worry about it.

Sounds good to me :tada:

dunn:
Merged as 0293c55e272acfff9b24f93f6a2a9827774479d1, thanks so much!

ilovezfs:
Nice work @woodruffw!

woodruffw:
Same to you, thanks for the help!

PR #1032, opened 2016-05-10
bookloupe 2.0 (new formula)

tangledhelix:


Formula/bookloupe.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/bookloupe", "--help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

tangledhelix: @mikemcquaid The package turns out to have a number of tests bundled, I have added them to the formula. They are passing (for me).

Formula/bookloupe.rb
```diff
@@ -0,0 +1,62 @@
...
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    ln_s("#{bin}/bookloupe", "./bookloupe")
```
UniqMartin: Why is this `ln_s` needed? We also tend to prefer a style that omits the (optional) parentheses, i.e.:

```ruby
ln_s "#{bin}/bookloupe", "./bookloupe"
```

Formula/bookloupe.rb
```diff
@@ -0,0 +1,62 @@
...
+      "scannos",
+      "short-line",
+      "single-quotes",
+      "spaced-doublequote",
+      "spaced-punctuation",
+      "unmarked-paragraph",
+      "unusual-characters",
+      "user-defined-typo",
+      "windows-1252",
+    ]
```
UniqMartin: You can make this array a bit more concise by using Ruby's `%w[` array syntax e.g.:

```ruby
test_list = %w[
  abbreviation
  before-comma
  before-period
  brackets
  control-characters
  dashes
  digits
  double-punctuation
  ellipsis
]
```

However, I wonder if this is really needed. Can't we just iterate over all `*.tst` files in `pkgshare` and use them? E.g., do something like this:

```ruby
Dir["#{pkgshare}/*.tst"].each do |test_file|
  system "#{bin}/loupe-test", test_file
end
```

Formula/bookloupe.rb
```diff
@@ -0,0 +1,62 @@
+class Bookloupe < Formula
+  desc "Lists common formatting errors in a Project Gutenberg candidate file"
```
UniqMartin: Tiny nitpick, but I think it reads better not to use the 3<sup>rd</sup> person singular here, i.e. use instead:

```ruby
desc "List common formatting errors in a Project Gutenberg candidate file"
```

Formula/bookloupe.rb
tangledhelix: The tests failed if `bookloupe` was not present in the cwd. After reviewing the source code, I found this can be done with an environment variable, I will squash and re-push with that instead.

Formula/bookloupe.rb
tangledhelix: No, you can't unfortunately. One test fails on Mac (passes on Linux). I omitted that test here since homebrew only is used on Macs, and opened a bug with the author about it. I'll switch to `%w[]`.

Formula/bookloupe.rb
tangledhelix: I updated that phrasing.

Formula/bookloupe.rb
UniqMartin: > One test fails on Mac (passes on Linux). I omitted that test here since homebrew only is used on Macs, and opened a bug with the author about it.

Thanks for the explanation! That's good to know. We should include that information and I think my suggestion is still valid with a minor modification:

```ruby
Dir["#{pkgshare}/*.tst"].each do |test_file|
  # Skip test that fails on OS X
  # <url-to-upstream-bug-report>
  next if test_file.end_with?("/name-of-failing-test.tst")

  system "#{bin}/loupe-test", test_file
end
```

Still a lot more concise than hard-coding an almost complete directory listing. What do you think?

tangledhelix: I agree; squashed and re-pushed.

UniqMartin:
Merged in cf5d1f55b6b4f094dc30891fbff7bf011867b606. Thank you for this contribution to Homebrew, @tangledhelix! :tada:

PR #1039, opened 2016-05-10
sile 0.9.3

ilovezfs:
add a test
disable bottling due to LuaRocks requirements

### DIFF
--- a/Formula/sile.rb
+++ b/Formula/sile.rb
@@ -1,8 +1,8 @@
 class Sile < Formula
   desc "Modern typesetting system inspired by TeX"
   homepage "http://www.sile-typesetter.org/"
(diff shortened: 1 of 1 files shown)


Formula/sile.rb
```diff
@@ -11,6 +11,8 @@ class Sile < Formula
     depends_on "libtool" => :build
   end
 
+  bottle :disable, "LuaRocks requirements preclude bottling"
```
mikemcquaid: 😭 Should see if we can vendor these at some point.

mikemcquaid:
👍

Formula/sile.rb
ilovezfs: Yeah: https://github.com/Homebrew/legacy-homebrew/issues/42273

PR #1040, opened 2016-05-10
dmd: fix VERSION file

John-Colvin:

---

I would have done this with a patch, but it seemed easier this way., especially as it will probably happen again and so we want to stay up to date regardless.

### DIFF
--- a/Formula/dmd.rb
+++ b/Formula/dmd.rb
@@ -1,6 +1,7 @@
 class Dmd < Formula
   desc "D programming language compiler for OS X"
   homepage "https://dlang.org/"
+  revision 1
 
   stable do
     url "https://github.com/D-Programming-Language/dmd/archive/v2.071.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/dmd.rb
```diff
@@ -47,6 +48,11 @@ class Dmd < Formula
   def install
     make_args = ["INSTALL_DIR=#{prefix}", "MODEL=#{Hardware::CPU.bits}", "-f", "posix.mak"]
 
+    # VERSION file is wrong upstream, has happened before, so we just overwrite it here.
+    version_file = (buildpath/"VERSION")
+    rm version_file
+    version_file.write version
```
mikemcquaid: Does `(buildpath/"VERSION").write version` work?

mikemcquaid:
Could you complain to upstream about this happening in an issue? Probably up for merging this anyway, though.

Formula/dmd.rb
John-Colvin: Have to remove first because otherwise it complains about overwriting an
existing file.

John-Colvin:
It's recognised as a problem upstream, but keeps happening nonetheless...

mikemcquaid:
Thanks again @John-Colvin!

For future reference the preferred commit message format is `dmd: fix VERSION file.`. Please check the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md). Thanks!

PR #1041, opened 2016-05-10
idnits 2.14.01

larseggert:

---

### DIFF
--- a/Formula/idnits.rb
+++ b/Formula/idnits.rb
@@ -1,14 +1,19 @@
 class Idnits < Formula
   desc "Looks for problems in internet draft formatting"
   homepage "https://tools.ietf.org/tools/idnits/"
-  url "https://tools.ietf.org/tools/idnits/idnits-2.13.02.tgz"
-  sha256 "6e42b044c79dc4e616d10ee9e283c20acc741601811a6acfc0c0d310afdf0823"
+  url "https://tools.ietf.org/tools/idnits/idnits-2.14.01.tgz"
+  sha256 "5b49749b0e4dc610fae9c12780c7a4d02723a5f4b701271bdd374d909c186654"
 
   bottle :unneeded
 
   depends_on "aspell"
+  depends_on "languagetool"
 
   def install
     bin.install "idnits"
+  end
+
+  test do
+    system "idnits --version"
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/idnits.rb
```diff
@@ -1,14 +1,19 @@
...
   homepage "https://tools.ietf.org/tools/idnits/"
-  url "https://tools.ietf.org/tools/idnits/idnits-2.13.02.tgz"
-  sha256 "6e42b044c79dc4e616d10ee9e283c20acc741601811a6acfc0c0d310afdf0823"
+  url "https://tools.ietf.org/tools/idnits/idnits-2.14.01.tgz"
+  sha256 "5b49749b0e4dc610fae9c12780c7a4d02723a5f4b701271bdd374d909c186654"
 
   bottle :unneeded
 
   depends_on "aspell"
+  depends_on "languagetool"
```
mikemcquaid: Can you elaborate a bit on why this is needed now? Thanks!

Formula/idnits.rb
```diff
@@ -1,14 +1,19 @@
...
 
   def install
     bin.install "idnits"
   end
+
+  test do
+    system "idnits --version"
```
mikemcquaid: `system "#{bin}/idnits", "--version"`. Ideally though this would run on an input file?

Formula/idnits.rb
larseggert: To enable the --grammarcheck option

Formula/idnits.rb
larseggert: Yes, but no valid input is included with the distribution. Is it OK to wget one during the test?

Formula/idnits.rb
mikemcquaid: Cool. Maybe worth making `:recommended` so users can opt-out?

Formula/idnits.rb
mikemcquaid: Ideally write one like https://github.com/Homebrew/homebrew-core/blob/master/Formula/cmake.rb#L65. Use `EOS.undent` to write a multiple line file.

larseggert: A minimal valid I-D is several pages of text, however

mikemcquaid: In that case making one a `resource` and downloading it with `.stage` maybe be better 👍

UniqMartin:
Merged. Thank you for this contribution to Homebrew, @larseggert! :tada:

PR #1044, opened 2016-05-10
soccerwindow2: fix the build

ilovezfs:
Fixes "ld: framework not found Security-lz"

### DIFF
--- a/Formula/soccerwindow2.rb
+++ b/Formula/soccerwindow2.rb
@@ -16,12 +16,22 @@ class Soccerwindow2 < Formula
   depends_on "librcsc"
 
   def install
+    # ld: framework not found Security-lz; also affects the rcssserver formula
+    # Reported 10 May 2016:
+    #   https://osdn.jp/ticket/browse.php?group_id=213&tid=36316
+    #   https://sourceforge.net/p/sserver/discussion/76440/thread/e61e21e8/
+    # Previously reported 14 Jan 2016:
+    #   https://sourceforge.net/p/sserver/mailman/message/34765272/
+    inreplace "configure",
+      "$QT4_REQUIRED_MODULES)$($PKG_CONFIG",
+      "$QT4_REQUIRED_MODULES) $($PKG_CONFIG"
+
     system "./configure", "--disable-debug",
                           "--prefix=#{prefix}"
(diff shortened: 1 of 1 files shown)


ilovezfs:
The current version of this software is here: https://sourceforge.net/projects/sserver/ and seems to have changed substantially including being broken up into multiple repositories. I'm not sure if anyone has an interest in trying to get this current, as development has actually continued into 2016, but I have no reason to believe this bug still exists.

Formula/soccerwindow2.rb
```diff
@@ -16,12 +16,17 @@ class Soccerwindow2 < Formula
   depends_on "librcsc"
 
   def install
+    # ld: framework not found Security-lz
+    inreplace "configure",
```
DomT4: Heh, this is identical to the problem is another of the upstream packages that [I patched](https://github.com/Homebrew/homebrew-core/blob/master/Formula/rcssserver.rb#L70-L77) a while back now.

ilovezfs: I like that our string replacements started and stopped at the same place.

ilovezfs: Wow this bug does still exist! It's so persistent because the problem is in the m4 file.
https://sourceforge.net/p/sserver/code/HEAD/tree/rcss/trunk/rcssmonitor_qt4/m4/qt.m4

DomT4: Don't you just love it when projects share buggy code and it ends up everywhere.

ilovezfs: I wonder where that m4 file actually originates.

mikemcquaid:
👍

Formula/soccerwindow2.rb
mikemcquaid: Usual comment that it'd be good to submit upstream if it's not already.

ilovezfs: https://sourceforge.net/p/sserver/discussion/76440/thread/e61e21e8/

ilovezfs: https://osdn.jp/ticket/browse.php?group_id=213&tid=36316

PR #1046, opened 2016-05-10
h2o 1.7.2

DomT4:

---

Note that the security fix for LibreSSL upstream didn't impact Homebrew because we don't use the bundled LibreSSL library.

### DIFF
--- a/Formula/h2o.rb
+++ b/Formula/h2o.rb
@@ -1,8 +1,8 @@
 class H2o < Formula
   desc "HTTP server with support for HTTP/1.x and HTTP/2"
   homepage "https://github.com/h2o/h2o/"
-  url "https://github.com/h2o/h2o/archive/v1.7.1.tar.gz"
-  sha256 "a9488667f0b73a66d5ef593660f6c0f66311d6eb4bf9378c6b7e74ab7ec9eea2"
+  url "https://github.com/h2o/h2o/archive/v1.7.2.tar.gz"
+  sha256 "fd75eab905dff9bcfd6c9c45d83a4cd0c133cc82a7d3dd48537244b12c2327a8"
   head "https://github.com/h2o/h2o.git"
 
   bottle do
@@ -22,6 +22,10 @@ class H2o < Formula
   depends_on "wslay" => :optional
 
   def install
+    # https://github.com/Homebrew/homebrew-core/pull/1046
+    # https://github.com/Homebrew/brew/pull/251
+    ENV.delete("SDKROOT")
+
     args = std_cmake_args
     args << "-DWITH_BUNDLED_SSL=OFF"
     args << "-DWITH_MRUBY=OFF" if build.without? "mruby"


mikemcquaid:
Yosemite hates you: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/1655/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_h2o/

DomT4:
Huh, Thanks Yosemite:
```
rake aborted!
The compiler failed to generate an executable file.
You have to install development tools first.
rakefile:26:in `load'
make[2]: *** [CMakeFiles/mruby] Error 1
make[1]: *** [CMakeFiles/mruby.dir/all] Error 2
```
Wonder if that's a race or something 💭.

DomT4:
The Yosemite error reproduced. Terrific. Guess I'm spinning up a VM later tonight then.

DomT4:
Can't reproduce on vanilla Yosemite with CLT-only. Will install Xcode and see if that's the issue, again, here.

DomT4:
Seems to be the increasingly well-known 10.10 SDK missing Vs protective `superenv` situation. In `env=std` and outside of Homebrew it realises its missing and falls back pretty gracefully:
```
==> cmake -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/h2o/1.7.2 -DCMAKE_BUILD_TYPE=Release -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_VERBOSE_MAKEFILE=ON -Wno-dev -DWITH_BUNDLED_SSL=OFF
CMake Warning at /usr/local/Cellar/cmake/3.5.2/share/cmake/Modules/Platform/Darwin-Initialize.cmake:97 (message):
  CMAKE_OSX_DEPLOYMENT_TARGET is '10.10' but the matching SDK does not exist
  at:

   "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk"

  Instead using SDK:

   "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk".
  /usr/local/Cellar/cmake/3.5.2/share/cmake/Modules/CMakeSystemSpecificInitialize.cmake:18 (include)
  CMakeLists.txt:24 (PROJECT)


-- The C compiler identification is AppleClang 7.0.2.7000181
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang
```

In `superenv` we remove the correct SDK and don't put anything back in its place:
```
clang called with: -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk -mmacosx-version-min=10.10 -o CMakeFiles/cmTC_f1b2f.dir/testCCompiler.c.o -c /tmp/h2o-20160515-2052-10pymmd/h2o-1.7.2/CMakeFiles/CMakeTmp/testCCompiler.c
superenv removed:  -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk
superenv added:    -pipe -w -Os -march=native -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -I/usr/local/opt/openssl/include
superenv executed: clang -pipe -w -Os -march=native -mmacosx-version-min=10.10 -o CMakeFiles/cmTC_f1b2f.dir/testCCompiler.c.o -c /tmp/h2o-20160515-2052-10pymmd/h2o-1.7.2/CMakeFiles/CMakeTmp/testCCompiler.c -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -I/usr/local/opt/openssl/include
```

mikemcquaid:
@DomT4 Correct me if I'm wrong @ilovezfs but it seems we should either always add the latest SDK or at least never remove it?

ilovezfs:
It is using the latest SDK. This is just a bug with finding that particular file.

ilovezfs:
Sorry wrong PR! I thought this was the vim one.

Yes, basically there is a whole set of bugs caused by developers thinking the 10.10 SDK still exists (or simply not having updated their build system yet), which only does if you're using Xcode 6*, no longer the latest version of Xcode for Yosemite. For Xcode 7 and above, there is no such thing as a separate 10.10 SDK, and any build system that attempts to use and find it is written with the expectation that it's dealing with Xcode 6.

ilovezfs:
```
superenv removed:  -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk
```
That is weird.

ilovezfs:
https://github.com/Homebrew/brew/blob/master/Library/ENV/4.3/cc#L182-L185
Removes it there.

It seems that it's not put back in here:
https://github.com/Homebrew/brew/blob/master/Library/ENV/4.3/cc#L91-L97

which I believe implies HOMEBREW_SDKROOT is not set. And that seems likely:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/ENV/super.rb#L110-L112

That looks like a bug to me, since when both Xcode and CLT are there, we go with the 10.11 SDK but then apparently don't set isysroot. If that's what's going on I think the bug won't occur if only Xcode is installed, so I'll check that to confirm.

mikemcquaid:
> That looks like a bug to me, since when both Xcode and CLT are there, we go with the 10.11 SDK but then apparently don't set isysroot.

This also sounds like a bug to me 👍

ilovezfs:
OK, this one can be fixed by adding
```
ENV["SDKROOT"] = "" 
```
at the top of the formula.

ilovezfs:
Annoyingly, that is another way to fix the vim bug too.

mikemcquaid:
Feels like a 🐛 worth 🔨

ilovezfs:
I'm not sure there's any obvious way to do so. Some stuff wants it set, other stuff doesn't.

mikemcquaid:
What is/was the value? Does it work if set to 10.11's SDK on 10.10?

ilovezfs:
It's getting confused because because on Yosemite `/usr/bin/ruby` rbconfig tells it that ruby is `/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby`, and to look for the folder `/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/universal-darwin14/ruby`, which contains `config.h`. However, we set `SDKROOT` to the 10.11 SDK folder `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk`, which contains its own `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/Ruby.framework`, but that one only has `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/universal-darwin15/ruby`, and doesn't have a `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/universal-darwin14/ruby`, so it fails to find config.h.

A hack solution is to add a symlink in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0` named `universal-darwin14` that points at `universal-darwin15`, which is not something I'm suggesting we do, but demonstrates the problem.

The cases where setting SDKROOT explicitly helps are when something is getting confused because it sees Xcode is there but can't find a MacOSX10.10.sdk folder.

mikemcquaid:
I see. It seems in this case that setting SDKROOT to the 10.11 SDK is explicitly the opposite of what we want. Given this is something provided by the system and not the CLT even perhaps we shouldn't be setting it globally at all?

ilovezfs:
The "somthing" in this case is Ruby, because `/usr/bin/ruby` is `/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby` and there's no `bin` folder at all in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr`.

It's a choice between two different sets of bugs. Setting it creates some bugs and prevents others. Not setting it avoids creating some bugs but leads to other bugs. On balance, not setting it is probably the "right" thing to do, and reporting all instances of "looking for non-existent 10.10 sdk with Xcode 7 folder" to upstreams as outright bugs, which they are.

Practically, though, it's a question of which type of bug is more plentiful. So far this Ruby config.h kind seems more rare than the other kind because it's pretty Ruby specific with this hard coded darwin version problem compounded with Apple's failure to provide a universal-darwin14 symlink folder in the unitary SDK.

mikemcquaid:
I'd err towards whichever creates less work for us for now which sounds like unsetting `SDKROOT` in the relevant cases. Out of interest, could superenv having a `ruby` shim help here?

ilovezfs:
Probably not. Adding a shim is what I did to discover it was using rbconfig. But there's no alternate ruby for us to point it at so it would still end up in the same place in /System unless/until we vendor our own Ruby, in which case presumably the vendored `ruby` would be earlier in the PATH and providing non-problematic answers, so the shim wouldn't be adding value in that case either.

ilovezfs:
BTW another answer is to cap all the Yosemite CI at Xcode 6.4, and then deal with formulae that both have a hard >=Xcode 7 dependency and are Yosemite compatible on a case-by-case basis, or relegate them to a separate YosemiteXcode7 CI. This problem of course will continue when Xcode 8 comes out and only has a 10.12 SDK and yet is El Capitan compatible :)

mikemcquaid:
Yeh, would rather just leave it at the same version as people are going to autoupdate to that anyway.

ilovezfs:
Yup, agreed. It would basically mean the end of supporting source builds except on the latest OS.

DomT4:
I left this alone last night to work on it tomorrow (today) and came back to discover you two had mostly worked it out. I should let things sit for a while more often 😉.

Simply unsetting `SDKROOT` on Yosemite where Xcode 7 is installed makes sense to me. IIRC we don't set `SDKROOT` where _only_ the CLT is installed and on a CLT-only system this (along with other formulae I believe we've had issue with) works without any issue.

ilovezfs:
@DomT4 It is always *set* if Xcode is installed and Xcode is >= 7, regardless of the presence or absence of the CLT, specifically for the sake of Yosemite. On an as needed basis we can turn that behavior off if forced to, which is what this formula, but not the vast majority, needs.

ilovezfs:
@DomT4 BTW I think it's just this mruby thing where the problem occurs. Do you happen to know why that's a `--without` option?

Also, your idea about *only* unsetting it on Yosemite is interesting. The only downside to that I can see is it will mean that the exception case will have to be updated for this formula at each Xcode major version bump.

DomT4:
> It is always set if Xcode is installed and Xcode is >= 7, regardless of the presence or absence of the CLT, specifically for the sake of Yosemite.

Yeah, that's what I was saying. Where _only_ the CLT is installed we don't set that variable and there don't seem to be so many build problems (subjectively speaking).

> On an as needed basis we can turn that behavior off if forced to, which is what this formula, but not the vast majority, needs.

Can you point to other formulae which break if `SDKROOT` isn't set where Xcode is installed? I'm sure there are a bunch, not questioning that, just wondering if we can nail down differences in the build process and work out if there are things we could fix on our side.

> I guess ENV.delete("SDKROOT") should work, too.

Yeah, that's the preferred syntax 👍.

> BTW I think it's just this mruby thing where the problem occurs. Do you happen to know why that's a --without option?

It's on by default upstream, and we try to respect that when it doesn't require anything not provided by the system, which in this case it hasn't so far _(Needs `bison`, `ruby`, and `mkmf` which has been provided since at least the Ruby 1.8 series)_.

You're correct that if you turn that off on Yosemite the build process is content.

> Also, your idea about only unsetting it on Yosemite is interesting. The only downside to that I can see is it will mean that the exception case will have to be updated for this formula at each Xcode major version bump.

Would this be a significant problem these days given Yosemite has apparently stopped receiving Xcode updates?

ilovezfs:
>Can you point to other formulae which break

https://github.com/Homebrew/legacy-homebrew/pull/50137
https://github.com/Homebrew/legacy-homebrew/pull/50229

>Would this be a significant problem these days given Yosemite has apparently stopped receiving Xcode updates?

Yes, because it's not a Yosemite issue per se. El Capitan will be "the new Yosemite" when Xcode 8 ships. If you want to restrict it in some way, which probably isn't necessary, it would be better to do `ENV.delete("SDKROOT") if MacOS::Xcode.installed? && MacOS::Xcode.version.to_i >= 7`.

DomT4:
@ilovezfs If you feel like it IMO it'd be a good idea for us to document this in the `brew` repo, which kinds of errors frequently mean the `ENV.delete("SDKROOT")` should be used, etc, especially if we're going to carry this system for El Cap in future as well.

mikemcquaid:
Maybe I'm overly optimistic but I wonder if there's something we can't do to fix this, given `stdenv` gets it right. Could we avoid filtering the preceding/following SDK with `superenv` or something?

ilovezfs:
```
    if MacOS::Xcode.without_clt? || (MacOS::Xcode.installed? && MacOS::Xcode.version.to_i >= 7)
      self["MACOSX_DEPLOYMENT_TARGET"] = MacOS.version.to_s
      self["SDKROOT"] = MacOS.sdk_path
    end
```
I'm tempted to change this to

```
    if MacOS::Xcode.without_clt? || (MacOS::Xcode.installed? && MacOS::Xcode.version.to_i >= 7)
      self["MACOSX_DEPLOYMENT_TARGET"] = MacOS.version.to_s
    end
```
and wait for the other class of bugs to show themselves, which can then be fixed with
```
ENV["SDKROOT"] = MacOS.sdk_path
```
at the top of `def install`, and make it opt-in as opposed to opt-out.

mikemcquaid:
Remind me: why are we only setting this for Xcode 7?

> In superenv we remove the correct SDK and don't put anything back in its place:

This feels like it may be a mistake?

ilovezfs:
Because only Xcode 7 has the unitary SDK. The old practice was to set SDKROOT and MACOSX_DEPLOYMENT_TARGET only if did not have the CLT. As of Xcode 7, it became necessary to set MACOSX_DEPLOYMENT_TARGET if Xcode was installed, regardless of the presence or absence of the CLT. SDKROOT came along for the ride.

I think this same class of bugs we're currently seeing existed under the old rules, but only presented on systems that had Xcode and lacked the CLT, whereas as now they present on any system with Xcode.

>In superenv we remove the correct SDK and don't put anything back in its place:
>
This feels like it may be a mistake?

The isysroot thing may very well be a bug in its own right but changing it made no difference here, and ended up being a red herring.

mikemcquaid:
> The isysroot thing may very well be a bug in its own right but changing it made no difference here, and ended up being a red herring.

Ok, thanks for humouring me here. I'm easy with either opt-in or opt-out but I'd probably favour whichever is going to result in the least issues for us. Might be interesting seeing how MacPorts is handling it.

ilovezfs:
It feels "safest" from a superenv perspective to always set the SDKROOT to the SDK folder within Xcode whenever it's available, and make accommodations for build systems that, for whatever valid or invalid reasons cannot handle that, but I only feel strongly about MACOSX_DEPLOYMENT_TARGET, which is a separable topic.

mikemcquaid:
> but I only feel strongly about MACOSX_DEPLOYMENT_TARGET, which is a separable topic.

Separate but related. What are your thoughts on that?

ilovezfs:
>Separate but related. What are your thoughts on that?

For Xcode 7, you have one SDK only. Therefore, always setting MACOSX_DEPLOYMENT_TARGET is necessary whenever that SDK is present (i.e., whenever >=Xcode 7 is installed).

But always setting MACOSX_DEPLOYMENT_TARGET neither requires, nor prevents, us from also always setting SDKROOT. If always setting SDKROOT is causing more problems than it's preventing, we can simply stop doing that, and only always set MACOSX_DEPLOYMENT_TARGET.

That said, since setting SDKROOT to the SDK folder "should" always be OK, that's probably "safest." However, it does introduce an asymmetry between what we do when Xcode is installed (set it) and what we do when Xcode is not installed (not set it, since there's nothing to set it to), which is not necessarily desirable.

mikemcquaid:
> If always setting SDKROOT is causing more problems than it's preventing, we can simply stop doing that, and only always set MACOSX_DEPLOYMENT_TARGET.
> 
> That said, since setting SDKROOT to the SDK folder "should" always be OK, that's probably "safest." However, it does introduce an asymmetry between what we do when Xcode is installed (set it) and what we do when Xcode is not installed (not set it, since there's nothing to set it to), which is not necessarily desirable.

Seems like a good call to just remove it, in that case.

ilovezfs:
>Seems like a good call to just remove it, in that case.

Yes, I think it's worth trying. Things that break can be worked around by adding `ENV["SDKROOT"] = MacOS.sdk_path` at the beginning of `def install`. Off the top of my head, I believe `portmidi` will need that, but we'll see what else comes up.

ilovezfs:
@mikemcquaid https://github.com/Homebrew/brew/pull/251

PR #1056, opened 2016-05-10
elm 0.17

jtwebman:

---

### DIFF
--- a/Formula/elm.rb
+++ b/Formula/elm.rb
@@ -69,7 +69,7 @@ class Elm < Formula
       main = text "Hello, world!"
     EOS
 
-    system bin/"elm", "package", "install", "evancz/elm-html", "--yes"
+    system bin/"elm", "package", "install", "elm-lang/html", "--yes"
 
     out_path = testpath/"index.html"
     system bin/"elm", "make", src_path, "--output=#{out_path}"
--- a/Formula/elm.rb
+++ b/Formula/elm.rb
@@ -7,27 +7,27 @@ class Elm < Formula
   homepage "http://elm-lang.org"
 
   stable do
-    url "https://github.com/elm-lang/elm-compiler/archive/0.16.tar.gz"
-    sha256 "ea4ff37ec6a1bfb8876e7a9b2aa0755df9ac92f5e5c8bfcc611b1886fb06bb13"
+    url "https://github.com/elm-lang/elm-compiler/archive/0.17.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
Do you know if all the gymnastics in `def install` are actually still required for 0.17?

jtwebman:
@ilovezfs what do you mean by gymnastics?

ilovezfs:
@jtwebman Compare, for example, git-annex which only needs to do
```
install_cabal_package :using => ["alex", "happy", "c2hs"], :flags => ["webapp"] 
```
and agda which only needs to do
```
install_cabal_package :using => ["alex", "happy", "cpphs"]
```
Currently elm is not making any use of that newer method and I'm wondering if that still has to remain true.

jtwebman:
@ilovezfs I am not sure, I was just updating it to the latest packages as it was released a few days ago now. The main people on Elm don't see brew as one of the official packages.

jtwebman:
@ilovezfs Looks like this is still needed from the build from source hs file. 

https://github.com/elm-lang/elm-platform/blob/master/installers/BuildFromSource.hs

ilovezfs:
@jtwebman 

>The main people on Elm don't see brew as one of the official packages.

Do you have citation(s) for this? Are there particular concerns they've expressed?

ilovezfs:
@jtwebman Thank you for that link it. That helps explain why this install method is the way it is. We should probably have a comment with that link in the formula.

Reading through that, it does make me wonder if we could simplify this formula by relying more heavily on the upstream build script which a) uses a sandbox just as we do b) handles the vendoring of each of the resources the formula is doing manually c) performs the directories' gymnastics the formula is doing manually.

jtwebman:
@ilovezfs I have no concern with it I was just updating this because our team uses brew more then the npm way to install Elm. I am sure changes could be made but this works now so I hope we can get this pull request merged and then we can work on a better way to build it.

ilovezfs:
Thank you for your contribution to Homebrew @jtwebman :tada:

jtwebman:
@ilovezfs No merge, just closed?

ilovezfs:
@jtwebman Yup it gets rebased on origin/master so that there aren't any merge commits. Also, in this case I squashed it down to a single commit, and set the standard bump subject.

DomT4:
> I guess you guys just move the commit.

We cherry-pick commits that need bottle updates, in particular. Commit credit is still fully assigned to you.

jtwebman:
@ilovezfs @DomT4 Awesome, I actually kind of like it this way!

PR #1069, opened 2016-05-11
gst-plugins-bad: fix applemedia

### LINKED ISSUE #1006: Installing gst-plugins-bad does not find corevideomemory.h
When installing [gst-plugins-bad](../tree/master/Formula/gst-plugins-bad.rb) with the `with-applemedia` flag, building fails (but installation completes) with the following error:
```
plugin.m:25:10: fatal error: 'corevideomemory.h' file not found
```

This seems to be caused by [gst-plugins-bad-1.8.1.tar.xz](https://gstreamer.freedesktop.org/src/gst-plugins-bad/gst-plugins-bad-1.8.1.tar.xz) not containing said `corevideomemory.h` file. The file is however present in [the gst-plugins-bad git repo](https://github.com/GStreamer/gst-plugins-bad/blob/1.8.1/sys/applemedia/corevideomemory.h).

I know this is an issue with the package, not with Homebrew, but having this issue should at least create some visibility for people encountering the same error.

Reported in Gnome's Bugzilla as [Bug 766163](https://bugzilla.gnome.org/show_bug.cgi?id=766163)

ilovezfs:
- add resources for the two missing headers to fix applemedia option
- make applemedia a "without"/recommended option

Since a fix has already been merged upstream, the two new resource
blocks should be removed when >1.8.1 is released.

https://bugzilla.gnome.org/show_bug.cgi?id=766163
https://github.com/GStreamer/gst-plugins-bad/commit/43487482e5c5ec71867acb887d50b8c3f813cd63

Closes #1006.

### DIFF
--- a/Formula/gst-plugins-bad.rb
+++ b/Formula/gst-plugins-bad.rb
@@ -1,8 +1,27 @@
 class GstPluginsBad < Formula
-  desc "GStreamer plugins (less supported, missing docs, not fully tested)"
+  desc "GStreamer plugins less supported, not fully tested"
   homepage "https://gstreamer.freedesktop.org/"
-  url "https://gstreamer.freedesktop.org/src/gst-plugins-bad/gst-plugins-bad-1.8.1.tar.xz"
-  sha256 "0bbd58f363734fc0c4a620b2d6fb01d427fdafdbda7b90b4e15d03b751ca40f5"
+
+  stable do
+    url "https://gstreamer.freedesktop.org/src/gst-plugins-bad/gst-plugins-bad-1.8.1.tar.xz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@mikemcquaid is it reasonable to turn this into always-on, even though the feature must be disabled at compile time for 10.9, or would it better to put in a hard `depends_on :macos => :yosemite`, or better to leave it as-is?

Seems odd not to compile in the apple-media option whenever possible, since it doesn't add additional dependencies if the only reason is that it introduces a feature discrepancy between the 10.9 bottles and the 10.10 and 10.11 bottles.

Original discussion was here: https://github.com/Homebrew/legacy-homebrew/pull/35284

mikemcquaid:
I think as it's only affecting one version I'm 👍 on enabling it where possible. 🚢 it.

DomT4:
👍 as-is.

PR #1072, opened 2016-05-11
openssl: PEM generation tweaks

DomT4:

---

Due to the more or less blind trust external tools place in the PEM generated and stored in `/usr/local/etc/openssl` (Presuming your Homebrew is in the normal place) it's possible for self-signed or explicitly not trusted certificates to be treated as perfectly valid and safe.

This results in some potentially unexpected differences between system-provided tools and Homebrew-provided tools linked against different crypto libraries.

By default both the system curl (Since Mountain Lion) and Homebrew's curl (For ML and above) are linked to Apple's Secure Transport, which handles self-signed certificates in the Keychain as invalid unless they are explicitly trusted and throws the following error on connection: `* SSL certificate problem: Invalid certificate chain`.

If you mark the certificate in the Keychain as explicitly trusted both the system and default Homebrew curl respect that and will permit the connection to be established:

```
Connected to abc.def (xxx.xxx.xxx.xx) port 443 (#0)
TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
Server certificate: abc.def
...
Connection #0 to host abc.def left intact
```

This is also the situation for at least Safari and Google Chrome on OS X, where Keychain trust status guides whether or not the self-signed certificate should be trusted.

Where things are potentially more dicey is where you have built Homebrew's curl against OpenSSL or LibreSSL, which is a fairly common configuration, and you have a self-signed certificate in the Keychain when OpenSSL/LibreSSL is installed/upgraded/etc. That self-signed cert is folded into the PEM and treated as perfectly valid, even if it was marked explicitly not trusted in the Keychain at time of PEM creation.

This enables self-signed cert connections to go ahead without notification of the potential issue. It's mildly easier to spot on verbose curl connections, but no obvious warning is included that you're using a self-signed certificate for the connection. The example below was explicitly generated as a test cert and tagged as such:

```
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=GB; O=TEST.ORG; OU=TESTING; CN=abc.def; emailAddress=test@abc.def
*  start date: May  9 20:28:52 2016 GMT
*  expire date: June  9 20:28:52 2016 GMT
*  common name: abc.def (matched)
*  issuer: C=GB; O=TEST.ORG; OU=TESTING; CN=abc.def; emailAddress=test@abc.def
*  SSL certificate verify ok.
```

Although curl is used for testing here, wget and other utilities that trust OpenSSL's PEM file are impacted as well.

This change results in the PEM file only containing system root certificates, which is a more secure default position to take.

If you relied on the previous behaviour you can re-enable it by copying PEM files into `/usr/local/etc/openssl/certs` & running `$(brew --prefix openssl)/bin/c_rehash`. You may need to pass the directory to specific tools or by setting the SSL_CERT_DIR variable.

Credit to Geoff Nixon for the original report. Extra testing done by me against an Ubuntu server running the latest nginx release.

### DIFF
--- a/Formula/openssl.rb
+++ b/Formula/openssl.rb
@@ -5,6 +5,7 @@ class Openssl < Formula
   mirror "https://dl.bintray.com/homebrew/mirror/openssl-1.0.2h.tar.gz"
   mirror "https://www.mirrorservice.org/sites/ftp.openssl.org/source/openssl-1.0.2h.tar.gz"
   sha256 "1d4007e53aad94a5b2002fe045ee7bb0b3d98f1a47f8b2bc851dcd1c74332919"
+  revision 1
 
   bottle do
     sha256 "28320f60a208f1aba5420957148458d59b00bcac62fcb550ba88e737c07c27b4" => :el_capitan
@@ -121,7 +122,6 @@ class Openssl < Formula
 
   def post_install
     keychains = %w[
-      /Library/Keychains/System.keychain
       /System/Library/Keychains/SystemRootCertificates.keychain
     ]
 


DomT4:
For some reason the `openssl` bottles aren't generating correctly anymore, hmm. There should be an `etc` directory in `.bottle` and neither exist. This isn't a problem on the from source build, so it's an exclusively bottling issue apparently.

DomT4:
Apparently the problem was doing OpenSSL and LibreSSL in the same PR, for some reason.

derektamsen:
This doesn't actually solve the issue. You could also have untrusted certs in the System Roots keychain. Instead their should be a filter on the export to only export certs that are trusted from they system and system roots keychain.

DomT4:
> This doesn't actually solve the issue.

Find me anything in security that completely resolves an issue without wrecking usability and I'll find you a gold pot at the end of any given rainbow.

> You could also have untrusted certs in the System Roots keychain. 

Indeed, but at least for El Cap you're barred from playing with that keychain so anything there has been put there by Apple. The chances of ending up with a rogue cert in the System Root keychain is significantly lower _(and frankly if something does you've got bigger problems)_.

> Instead their should be a filter on the export to only export certs that are trusted from the system and system roots keychain.

I agree. Someone should ask Apple to write this into the `security` tool. It doesn't seem to exist today and is the first thing we checked for here.

mikemcquaid:
@derektamsen We'll accept PRs to improve this but comments that don't suggest an alternative implementation aren't really helpful at this point.

derektamsen:
Sorry about that. The apple security tool is definitely crappy. I'll see if their is something I can tack on to have a better solution.

mikemcquaid:
@derektamsen No worries, thanks for the apology ❤️

md5:
Can someone clarify for me where this leaves users with explicitly trusted corporate CA certificates in `System.keychain`? Are we expected to add them to `/usr/local/etc/certs` in addition to having them in the System keychain?

md5:
I suppose it could be expected that our corporate IT could image our machines in such a way that the internal root CA ends up under `/System`, but that's not how they're currently imaged and doesn't help all the developers who have these certificates trusted in `System.keychain` currently.

DomT4:
If you wanted something completely programmatic that didn't require any extra steps:
```bash
echo -n "\r$(security find-certificate -a -c "MYCERT" -p /Library/Keychains/System.keychain)" >> $(brew --prefix)/etc/openssl/cert.pem
```
Would do the trick.

Ultimately, what's desired here is for Apple to add a way to _not_ export certs that aren't trusted rather than exporting absolutely everything regardless of if that cert is expired, untrusted, etc. Sadly does not seem forthcoming.

md5:
@DomT4 Thanks. It seems like `verify-cert` could be useful.

PR #716, opened 2016-04-30
file-roller: fix for gschema

ilovezfs:
Closes #652.

PR #267, opened 2016-04-13
Gnome 3.20.0

tschoonj:
This is the continuation of PR Homebrew/legacy-homebrew#50257

Not to be merged yet as gtk+3.0 (and a couple of others too) is still quite buggy.

PR #1094, opened 2016-05-12
file-roller 3.20.2

ilovezfs:
- version bump
- schema validation bug is fixed (#716), so yank the `inreplace`
- add linked-library dependencies
- use assert_match in the version test


mikemcquaid:
👍

DomT4:
I'm not sure if this bump needs to go in with the rest of the new GNOME [stuff](https://github.com/Homebrew/homebrew-core/pull/267) but otherwise 👍.

PR #1106, opened 2016-05-12
pysvn 1.8.0 (new formula)

IlyaSemenov:
`pysvn` is not available via `pip`, so I made a formula for it.

I had to apply two small patches to the original source:
1. Fix `configure` to respect `CPPFLAGS`, `CXXFLAGS` and `LDDFLAGS` (Gentoo does the similar)
2. Prevent explicit linking to Python framework

### DIFF
--- a/Formula/pysvn.rb
+++ b/Formula/pysvn.rb
@@ -1,0 +1,111 @@
+class Pysvn < Formula
+  desc "Python interface to Subversion"
+  homepage "http://pysvn.tigris.org/"
+  url "http://pysvn.barrys-emacs.org/source_kits/pysvn-1.8.0.tar.gz"
+  sha256 "39596f4884ed689cdb5a4e210e421724302a566c7ba756cc4d46bbfeb0c8326b"
+
+  option "without-python", "Build without python2 support"
+
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
+  depends_on :python3 => :recommended
+  depends_on "openssl"
+  depends_on "subversion"
+
+  # Fix configure to respect CPPFLAGS, CXXFLAGS and LDFLAGS
+  # http://pysvn.tigris.org/issues/show_bug.cgi?id=206
+  patch :DATA
+
+  def install
+    if build.without?("python") && build.without?("python3")
+      odie "pysvn: at least one of --with-python or --with-python3 must be specified"
+    end
+
+    cd "Source" do
+      # Replace explicit linking against Python framework with dynamic lookup
+      # http://pysvn.tigris.org/issues/show_bug.cgi?id=207
+      inreplace "setup_configure.py", "%(PYTHON_FRAMEWORK)s", "-undefined dynamic_lookup"
+
+      i = 0
+      Language::Python.each_python(build) do |python, version|
+        if i > 0
+          # Remove trails of the previous compile
+          system "make", "clean"
+        end
+        system python, "setup.py", "configure"
+        system "make"
+        (lib/"python#{version}/site-packages/pysvn").install "pysvn/__init__.py", Dir["pysvn/_pysvn*.so"]
+        i += 1
+      end
+    end
+  end
+
+  test do
+    system "svnadmin", "create", "test"
+    Language::Python.each_python(build) do |python, _version|
+      system python, "-c", "import os, pysvn; pysvn.Client().info2('file://' + os.path.join(os.getcwd(), 'test'))"
+    end
+  end
+end
+__END__
+diff --git a/Source/setup_configure.py b/Source/setup_configure.py
+index afeee2d..1680de1 100644
+--- a/Source/setup_configure.py
++++ b/Source/setup_configure.py
+@@ -849,8 +849,8 @@ class CompilerGCC(Compiler):
+     def __init__( self, setup ):
+         Compiler.__init__( self, setup )
+ 
+-        self._addVar( 'CCC',            'g++' )
+-        self._addVar( 'CC',             'gcc' )
++        self._addVar( 'CCC',            'g++ $(CPPFLAGS) $(CXXFLAGS)' )
++        self._addVar( 'CC',             'gcc $(CPPFLAGS)' )
+ 
+     def getPythonExtensionFileExt( self ):
+         return '.so'
+@@ -976,8 +976,8 @@ class MacOsxCompilerGCC(CompilerGCC):
+         else:
+             arch_options = ''
+ 
+-        self._addVar( 'CCC',            'g++ %s' % (arch_options,) )
+-        self._addVar( 'CC',             'gcc %s' % (arch_options,) )
++        self._addVar( 'CCC',            'g++ $(CPPFLAGS) $(CXXFLAGS) %s' % (arch_options,) )
++        self._addVar( 'CC',             'gcc $(CPPFLAGS) %s' % (arch_options,) )
+ 
+         self._find_paths_pycxx_dir = [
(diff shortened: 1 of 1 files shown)


Formula/pysvn.rb
```diff
@@ -0,0 +1,70 @@
+class Pysvn < Formula
+  desc "Python interface to Subversion"
+  homepage "http://pysvn.tigris.org/"
+  url "http://pysvn.barrys-emacs.org/source_kits/pysvn-1.8.0.tar.gz"
+  sha256 "39596f4884ed689cdb5a4e210e421724302a566c7ba756cc4d46bbfeb0c8326b"
+
+  depends_on "openssl"
+  depends_on "subversion"
+  depends_on :python => :recommended
```
DomT4: This needs tweaking a little bit so we don't mandate anyone pouring the bottle installs Homebrew's Python to do so. You want something like:

```ruby
  option "without-python", "Build without python2 support"

  depends_on :python => :recommended if MacOS.version <= :snow_leopard
  depends_on :python3 => :optional
  depends_on "openssl"
  depends_on "subversion"
```

Formula/pysvn.rb
```diff
@@ -0,0 +1,70 @@
...
+  depends_on :python3 => :recommended
+
+  # Fix configure to respect CPPFLAGS, CXXFLAGS and LDFLAGS
```
DomT4: Is upstream aware of this?

Formula/pysvn.rb
```diff
@@ -0,0 +1,70 @@
...
+  patch :DATA
+
+  def install
+    if build.without?("python") && build.without?("python3")
+      odie "pysvn: at least one of --with-python or --with-python3 must be specified"
+    end
+    cd "Source" do
+      inreplace "setup_configure.py", "%(PYTHON_FRAMEWORK)s", "-undefined dynamic_lookup"
```
DomT4: Please report this upstream as well. We consider it an issue that upstreams can & should be aware of: http://blog.tim-smith.us/2015/09/python-extension-modules-os-x/

Formula/pysvn.rb
```diff
@@ -0,0 +1,70 @@
...
+      Language::Python.each_python(build) do |python, version|
+        system python, "setup.py", "configure"
+        system "make"
+        (lib/"python#{version}/site-packages/pysvn").install "pysvn/__init__.py", Dir["pysvn/_pysvn*.so"]
```
DomT4: Why can't this be done with `make install`? Does one not exist?

Formula/pysvn.rb
```diff
@@ -0,0 +1,70 @@
...
+        system "make", "clean"
```
DomT4: Why `make clean`?

Formula/pysvn.rb
IlyaSemenov: This is how upstream wants us to do: http://pysvn.stage.tigris.org/source/browse/*checkout*/pysvn/trunk/pysvn/Extension/INSTALL.html?revision=1297
Yes, there is no Makefile until `configure` has completed.

Formula/pysvn.rb
IlyaSemenov: Because we need to start from scratch (reconfigure & recompile) for every new Python in cycle (perhaps `make clean` can be moved before running `make` and not after install).

Formula/pysvn.rb
```diff
@@ -0,0 +1,70 @@
...
+  url "http://pysvn.barrys-emacs.org/source_kits/pysvn-1.8.0.tar.gz"
+  sha256 "39596f4884ed689cdb5a4e210e421724302a566c7ba756cc4d46bbfeb0c8326b"
+
+  depends_on "openssl"
+  depends_on "subversion"
+  depends_on :python => :recommended
+  depends_on :python3 => :recommended
+
+  # Fix configure to respect CPPFLAGS, CXXFLAGS and LDFLAGS
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/pysvn.rb
mikemcquaid: 👍

Formula/pysvn.rb
mikemcquaid: Could check `Language::Python.each_python(build).length > 1`

Formula/pysvn.rb
IlyaSemenov: Fixed (not sure how to test it without bottled Pythons though).

Formula/pysvn.rb
IlyaSemenov: Done.

Formula/pysvn.rb
IlyaSemenov: It's not like that. `Language::Python.each_python` does not return a collection or an Enumerator *(it better had do!)*, it's a simple function that accepts a block and returns nil. I couldn't invent anything better than plain old counter loop.

Formula/pysvn.rb
IlyaSemenov: Done.

Formula/pysvn.rb
IlyaSemenov: I reported the issue and updated the formula.

Formula/pysvn.rb
IlyaSemenov: I think I misread your question yesterday. Yes, there is no `make install` in the `Makefile`:

```
~/tmp/pysvn-1.8.0/Source $ make install
make: *** No rule to make target `install'.  Stop.
```

Upstream specifically tells to copy 2 files to `site-packages` (see the link above).

Formula/pysvn.rb
```diff
@@ -0,0 +1,111 @@
...
+  option "without-python", "Build without python2 support"
+
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
+  depends_on :python3 => :recommended
+  depends_on "openssl"
+  depends_on "subversion"
+
+  # Fix configure to respect CPPFLAGS, CXXFLAGS and LDFLAGS
+  # http://pysvn.tigris.org/issues/show_bug.cgi?id=206
+  patch :DATA
```
mikemcquaid: I'd like to see upstream merge this before we include it given it's non-trivial and a good indication of how maintained this project is. Sorry for the delay.

Formula/pysvn.rb
xu-cheng: Noted, this is a python library, instead of python CLI. So `depends_on :python` may be required. And this formula probably be more suitable for `homebrew/python`.

Formula/pysvn.rb
mikemcquaid: 😭

Formula/pysvn.rb
mikemcquaid: A good point @xu-cheng; I just noticed the `test` so seems this should be in homebrew/python, yeh.

IlyaSemenov:
Created a PR at https://github.com/Homebrew/homebrew-python/pull/299 instead

pbauer:
@IlyaSemenov pysvn has not survived the consolidation of homebrew/python into homebrew (https://github.com/Homebrew/homebrew-python/issues/388). Is there some way to install it with homebrew? What other way would you recommend?

IlyaSemenov:
I believe you may simply drop the raw formula file into /usr/local/... along with other formulas. I didn’t follow the latest Homebrew development so not completely sure the formula will work straight out of the box but it’s quite probable.

PR #1110, opened 2016-05-13
certbot 0.6.0

DomT4:

---

### DIFF
--- a/Formula/certbot.rb
+++ b/Formula/certbot.rb
@@ -1,8 +1,8 @@
 class Certbot < Formula
   desc "Tool to obtain certs from Let's Encrypt and autoenable HTTPS"
-  homepage "https://letsencrypt.org/"
-  url "https://github.com/certbot/certbot/archive/v0.5.0.tar.gz"
-  sha256 "143a7eb5bc6b1336ead2add37425170f73da471c18fb4d36c7e983e8db5251df"
+  homepage "https://certbot.eff.org/"
+  url "https://github.com/certbot/certbot/archive/v0.6.0.tar.gz"
+  sha256 "58eaa6be4ae90af07d682296e42dbac924b5602c51002445059596313e182a09"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


jmagnusson:
This version seems to want to write outside of the `brew --prefix` location.

```
$ certbot
An unexpected error occurred:
OSError: [Errno 13] Permission denied: '/var/log/letsencrypt'
Please see the logfile 'certbot.log' for more details.
```

When I run `sudo certbot` it shows this dialog with the text `2 Nginx Web Server - currently doesn't work (nginx) [Misconfigure`. If I select `2 Nginx Web Server` and go to `<More Info>` I get the following info:
```
 ┌──────────────────────────────────────────────────────────────────────┐
 │ Reported Error: Error while running nginx -c /etc/nginx/nginx.conf   │
 │ -t.                                                                  │
 │                                                                      │
 │ nginx: [emerg] open() "/etc/nginx/nginx.conf" failed (2: No such     │
 │ file or directory)                                                   │
 │ nginx: configuration file /etc/nginx/nginx.conf test failed          │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 │                                                                      │
 ├──────────────────────────────────────────────────────────────────────┤
 │                               <  OK  >                               │
 └──────────────────────────────────────────────────────────────────────┘
```

It doesn't seem to respect installed brew software at all. If I look at the formula it specifies all of the Python dependencies manually. Why not just depend on python and `pip install certbot`?

DomT4:
Have you used `letsencrypt` (now `certbot`) before? That behaviour is normal.

It's fairly well documented upstream `certbot` _expects_ to be treated like an operating system component and consequently requires `sudo`. Not only for the folders it prefers to write to but because for some operations it binds to port 443. This is unusual for `brew`-installable software but not unheard of.

> If I select 2 Nginx Web Server and go to <More Info> I get the following info:

I'll look into the `nginx` situation but that plugin (and indeed the entire `certbot` software) is still deemed "experimental" and isn't expected to behave entirely coherently for every use case. Last time I checked you could feed it an alternative path to the `nginx` configuration directory.

> If I look at the formula it specifies all of the Python dependencies manually. Why not just depend on python and pip install certbot?

This isn't the way Homebrew handles Python formulae. There's documentation on this [here](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md), but essentially the Python resources are sandboxed into the prefix so it's both a sanitised environment we know works & it doesn't interfere with what a user has `pip install`'ed previously.

jmagnusson:
@DomT4 thanks for the info and sorry for taking your time. I'll try to read up more on Homebrew's documentation until next time 😅

PR #1117, opened 2016-05-13
prometheus 0.18.0

ilovezfs:


ilovezfs:
Seems fine building from the tarball instead of the tag.

mikemcquaid:
👍

DomT4:
`ENV["GO15VENDOREXPERIMENT"] = "1"` can be removed now, but feel free to do so in post. 👍

PR #1141, opened 2016-05-13
ospac 1.0 (new formula)

sritterbusch:
Ospac will take a multi-channel recording of an audio podcast conversation
and master this to a high-quality mix-down with support for intro and outro.

### DIFF
--- a/Formula/ospac-gui.rb
+++ b/Formula/ospac-gui.rb
@@ -1,0 +1,20 @@
+class OspacGui < Formula
+  desc "Open Source Podcast Audio Chain - GUI"
+  homepage "http://ospac.net/"
+  url "https://github.com/sritterbusch/ospac/archive/v1.0.tar.gz"
+  sha256 "793f0454cebe48dfa8364f26fa48469717954c31cb632a9c0b9dd24920d83f4c"
+
+  depends_on "libsndfile"
+  depends_on "fltk"
+  depends_on "ospac"
+
+  def install
+    system "make", "ospac-gui"
+    mkdir bin.to_s
+    cp "GUI-Release/ospac-gui", bin.to_s
+  end
+
+  test do
+    system "ospac", "--help"
+  end
+end
--- a/Formula/ospac-gui.rb
+++ b/Formula/ospac-gui.rb
@@ -9,9 +9,12 @@ class OspacGui < Formula
   depends_on "ospac"
 
   def install
-    system "make", "ospac-gui"
+    inreplace "makefile.targets" do |s|
+      s.gsub! "/usr/local", prefix.to_s
+      s.gsub! "cp ../ospac.1", "#cp ../ospac.1"
+    end
     mkdir bin.to_s
-    cp "GUI-Release/ospac-gui", bin.to_s
+    system "make", "ospac-gui", "install-gui"
   end
 
   test do
--- a/Formula/ospac.rb
+++ b/Formula/ospac.rb
@@ -7,13 +7,12 @@ class Ospac < Formula
   depends_on "libsndfile"
 
   def install
-    system "make", "ospac"
+    inreplace "makefile.targets", "/usr/local", prefix.to_s
     mkdir bin.to_s
     mkdir share.to_s
     mkdir man.to_s
     mkdir man1.to_s
-    cp "Release/ospac", bin.to_s
-    cp "ospac.1", man1.to_s
+    system "make", "ospac", "install"
   end
 
   test do
--- a/Formula/ospac.rb
+++ b/Formula/ospac.rb
@@ -1,0 +1,22 @@
+class Ospac < Formula
+  desc "Open Source Podcast Audio Chain"
+  homepage "http://ospac.net/"
+  url "https://github.com/sritterbusch/ospac/archive/v1.0.tar.gz"
+  sha256 "793f0454cebe48dfa8364f26fa48469717954c31cb632a9c0b9dd24920d83f4c"
+
+  depends_on "libsndfile"
+
+  def install
+    system "make", "ospac"
+    mkdir bin.to_s
+    mkdir share.to_s
+    mkdir man.to_s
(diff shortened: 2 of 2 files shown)


Formula/ospac-gui.rb
```diff
@@ -0,0 +1,20 @@
...
+  sha256 "793f0454cebe48dfa8364f26fa48469717954c31cb632a9c0b9dd24920d83f4c"
+
+  depends_on "libsndfile"
+  depends_on "fltk"
+  depends_on "ospac"
+
+  def install
+    system "make", "ospac-gui"
+    mkdir bin.to_s
+    cp "GUI-Release/ospac-gui", bin.to_s
```
ilovezfs: this should use `bin.install`

Formula/ospac-gui.rb
```diff
@@ -0,0 +1,20 @@
...
+  homepage "http://ospac.net/"
+  url "https://github.com/sritterbusch/ospac/archive/v1.0.tar.gz"
+  sha256 "793f0454cebe48dfa8364f26fa48469717954c31cb632a9c0b9dd24920d83f4c"
+
+  depends_on "libsndfile"
+  depends_on "fltk"
+  depends_on "ospac"
+
+  def install
+    system "make", "ospac-gui"
```
ilovezfs: Is there no `make install`?

Formula/ospac.rb
```diff
@@ -0,0 +1,22 @@
...
+  sha256 "793f0454cebe48dfa8364f26fa48469717954c31cb632a9c0b9dd24920d83f4c"
+
+  depends_on "libsndfile"
+
+  def install
+    system "make", "ospac"
+    mkdir bin.to_s
+    mkdir share.to_s
+    mkdir man.to_s
+    mkdir man1.to_s
```
ilovezfs: none of these mkdirs should be needed.

Formula/ospac.rb
```diff
@@ -0,0 +1,22 @@
...
+    cp "Release/ospac", bin.to_s
+    cp "ospac.1", man1.to_s
```
ilovezfs: bin.install and man1.install

Formula/ospac.rb
sritterbusch: When following https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md none of the target directories where existent, therefore the mkdir were indeed needed.

Formula/ospac.rb
sritterbusch: Following https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md the audit recommended using "cp".

Formula/ospac.rb
ilovezfs: This is why you use, for example, `bin.install "..."`

https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/pathname.rb#L97

It calls `mkpath` before calling `FileUtils.mv`.

Formula/ospac-gui.rb
sritterbusch: Of course there is "make install" to /usr/local/bin. Since the homebrew build process does not seem to be in a jail, this would not result in a target where home-brew expects it to be. Therefore I added the two install commands directly into the script taking care of the demanded destinations.

ilovezfs: It would be better to adapt the Makefile using `inreplace "Makefile", /somepattern/, "blah blah #{prefix}"`

Formula/ospac-gui.rb
sritterbusch: If so, please adapt https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md . The command bin.install is indeed mentioned there, but to date only as a remark to explain this "stuff".

ilovezfs: You're welcome to open a pull request.

sritterbusch: With all your respect, I am trying to follow the given guidelines- not to change the rules. And thanks btw for your fast response. =)

ilovezfs: You're welcome, and it is in there:

https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#file-level-operations

https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#just-moving-some-files

Formula/ospac-gui.rb
sritterbusch: For the next release, I will provide some kind of prefix-variable to the Makefile. Therefore, hacking the Makefile seems to be more out-of-place than just installing the two files manually.

But if this kind of string-replacement hacking is desired for homebrew, I will try to follow your recommendations.

Formula/ospac.rb
ilovezfs: That was probably rubocop talking.

Formula/ospac-gui.rb
ilovezfs: Yes, when it's possible to use `inreplace` or a `patch do`, and then call `make install`, that is preferable to manual `blah.install`s.

ilovezfs: most likely you can just do
```
inreplace "Makefile", "/usr/local", prefix
```

sritterbusch: Thanks. I followed this advice, but now I had to leave the `mkdir`-commands in: The `make install` assumes existing directories.

DomT4:
Sorry, this remains quite a bit below our eligibility guidelines at the moment as detailed by `brew audit --strict --online`. You may wish to consider hosting it in your [own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) until it meets those requirements.

Thank for for the submission to Homebrew nonetheless ❤️

PR #1153, opened 2016-05-14
JSON for Modern C++ 1.1.0 (new formula)

nlohmann:
JSON for Modern C++ is a single-header library for C++11. Its goal is to introduce JSON as first-class data type. JSON is implemented as C++ container and plays well with other STL containers, algorithms, or other aspects of C++.

### DIFF
--- a/Formula/nlohmann_json.rb
+++ b/Formula/nlohmann_json.rb
@@ -1,0 +1,65 @@
+class NlohmannJson < Formula
+  desc "JSON for Modern C++"
+  homepage "https://github.com/nlohmann/json"
+  url "https://github.com/nlohmann/json/archive/v1.1.0.tar.gz"
+  sha256 "ee3825841e6d6915428caf2cea53927c3e4d56315a23ee7d1a64bfe1c19a656f"
+  head "https://github.com/nlohmann/json.git"
+
+  def install
+    include.install "src/json.hpp"
+    ohai "to use the library, please set your include path accordingly:"
+    ohai "CPPFLAGS: -I#{include}"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <json.hpp>
+
+      using nlohmann::json;
+
+      int main() {
+        // create an empty structure (null)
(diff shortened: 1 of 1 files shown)


Formula/nlohmann_json.rb
```diff
@@ -0,0 +1,65 @@
+class NlohmannJson < Formula
+  desc "JSON for Modern C++"
+  homepage "https://github.com/nlohmann/json"
+  url "https://github.com/nlohmann/json/archive/v1.1.0.tar.gz"
+  sha256 "ee3825841e6d6915428caf2cea53927c3e4d56315a23ee7d1a64bfe1c19a656f"
+  head "https://github.com/nlohmann/json.git"
+
+  def install
+    include.install "src/json.hpp"
+    ohai "to use the library, please set your include path accordingly:"
```
DomT4: Please don't `ohai` here.

Formula/nlohmann_json.rb
```diff
@@ -0,0 +1,65 @@
+class NlohmannJson < Formula
+  desc "JSON for Modern C++"
+  homepage "https://github.com/nlohmann/json"
+  url "https://github.com/nlohmann/json/archive/v1.1.0.tar.gz"
+  sha256 "ee3825841e6d6915428caf2cea53927c3e4d56315a23ee7d1a64bfe1c19a656f"
+  head "https://github.com/nlohmann/json.git"
+
+  def install
+    include.install "src/json.hpp"
```
DomT4: Various maintainers have feelings on single-header formulae, so I'll open this one up for comments.

mikemcquaid: Yeh, personally 👎 on these formulae; feels like a better fit for projects to vendor them themselves.

Formula/nlohmann_json.rb
nlohmann: Should the message be removed altogether our just be displayed with a command different than `ohai`?

dunn:
Passing on this, but thank you for the submission! ❤️

PR #1158, opened 2016-05-14
compcert: fix ocaml 4.03.0 induced build failure

ilovezfs:
ocaml's old default "warn-error" list was empty, so all warnings were
non-fatal, but as of ocaml 4.03.0, the default setting is now
"-warn-error -a+31" (all warnings are non-fatal except 31). This causes
the vendored Coq 8.4 build to fail, so restore the old default.

Also, restore parallelization for the `make world` part of the Coq 8.4
build, as only `make install` seems to require deparallelization.


DomT4:
LGTM 👍.

Issue #342, opened 2016-04-15
Formulae fail to build


xu-cheng:
I have finished the massive bottle build. Below is the list of formulae fail to build.

 - [x] blucat
 - [x] cvs2svn
 - [x] dwarf
 - [x] file-roller
 - [x] fsw (test timeout)
 - [x] geocouch — #475
 - [x] gexiv2
 - [x] henplus
 - [x] hqx
 - [x] juise
 - [x] kjell — #870
 - [x] libcanberra
 - [x] libmarisa
 - [x] libreadline-java
 - [x] mapserver
 - [x] mercury
 - [x] mikmod
 - [x] mmix
 - [x] ninja-ide
 - [x] ocamlsdl  (build timeout like 12hrs)
 - [x] pg_top
 - [x] pillar
 - [x] pylucene
 - [x] recode
 - [x] sam2p
 - [x] sdf
 - [x] sile
 - [x] skytools
 - [x] slony
 - [x] snow
 - [x] soccerwindow2
 - [x] sproxy
 - [x] ssreflect
 - [x] suricata
 - [x] swatchbooker
 - [x] valkyrie
 - [x] vcodex
 - [x] vmalloc
 - [x] wartremover
 - [x] wsmancli — fails because its dependency `openwsman` is broken: Homebrew/legacy-homebrew#44047
 - [x] zint

Without an El Capitan bottle, but not necessarily broken:

- [x] apt-cacher-ng — `depends_on :osxfuse`
- [x] class-dump — `MaximumMacOSRequirement => :mavericks`
- [x] ctunnel — `depends_on :tuntap`
- [x] fastd — `depends_on :tuntap`
- [x] horndis — `depends_on UnsignedKextRequirement`
- [x] iodine — `depends_on :tuntap`
- [x] ld64 — `MaximumMacOSRequirement => :snow_leopard`
- [x] osxfuse — `depends_on UnsignedKextRequirement`
- [x] peervpn — `depends_on :tuntap`
- [x] shrewsoft-vpn-client — `depends_on :tuntap`
- [x] tuntap — `depends_on UnsignedKextRequirement`
- [x] vpnc — `depends_on :tuntap`

mikemcquaid:
Failed to build on what platform/platforms? Do they have sandbox violations?

DomT4:
I presume `apt-cacher-ng` was a 404, was being worked on [here](https://github.com/Homebrew/legacy-homebrew/pull/49987) but hit other issues.

[...]

PR #1166, opened 2016-05-14
ssreflect: vendor Coq 8.4

ilovezfs:
Fixes

```
  File "src/ssrmatching.mli", line 76, characters 14-34:
  Error: Unbound type constructor glob_constr_and_expr
  make[1]: *** [src/ssrmatching.cmi] Error 2
  make: *** [all] Error 2
```

Also, add a test.

Upgrading the formula to mathcomp/ssreflect 1.6 will allow Coq 8.5 to
displace the vendored Coq 8.4.

### DIFF
--- a/Formula/ssreflect.rb
+++ b/Formula/ssreflect.rb
@@ -1,20 +1,58 @@
+class Camlp5TransitionalModeRequirement < Requirement
+  fatal true
+
+  satisfy(:build_env => false) { !Tab.for_name("camlp5").with?("strict") }
+
+  def message; <<-EOS.undent
+    camlp5 must be compiled in transitional mode (instead of --strict mode):
+      brew install camlp5
+    EOS
+  end
+end
+
 class Ssreflect < Formula
   desc "Virtual package provided by libssreflect-coq"
   homepage "https://www.msr-inria.fr/projects/mathematical-components-2/"
   url "http://ssr.msr-inria.inria.fr/FTP/ssreflect-1.5.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Seems a bit heavy but seems reasonable if it's just a short-term fix per https://github.com/Homebrew/homebrew-core/issues/342#issuecomment-219235922.

Formula/ssreflect.rb
```diff
@@ -36,7 +76,27 @@ def install
...
+        done.
+      Qed.
+
+      Check helloworld.
+    EOS
+    (testpath/"expected").write <<-EOS.undent
+      helloworld
+           : P -> P
+    EOS
+    assert_equal File.read(testpath/"expected"), pipe_output("ssrcoq -compile helloworld")
```
DomT4: If `ssrcoq` is part of this package this should be `("#{bin}/ssrcoq -compile helloworld")` though 😉.

ilovezfs:
Yeah, just as a matter of principle, it feels best to leave it in a working state before moving to the new version which is pretty different. I'm not sure if it should be a rename or a new formula altogether:

https://github.com/math-comp/math-comp/tree/master/mathcomp

Note that ssreflect is just one component of the thing that is called mathcomp 1.6

Formula/ssreflect.rb
ilovezfs: nice catch

DomT4:
Cool, no worries. :shipit: as desired. Can always fix the bin call in `test do` in post if you want to skip another CI run.

PR #1182, opened 2016-05-15
kore: El Capitan fixes

DomT4:

---

Currently the test (including outside Homebrew) fails because it can't find our OpenSSL outside of the build due to not "remembering" the path, and also doesn't respect user CFLAGS, so this is currently broken pretty nicely for El Cap users. This should fix it.

Also fixes for Homebrew users who haven't placed their Homebrew in the default prefix.

### DIFF
--- a/Formula/kore.rb
+++ b/Formula/kore.rb
@@ -1,10 +1,24 @@
 class Kore < Formula
   desc "Web application framework for writing web APIs in C"
   homepage "https://kore.io/"
-  url "https://kore.io/release/kore-1.2.3-release.tgz"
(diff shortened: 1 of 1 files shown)


DomT4:
I need to check the `CFLAGS` and `LDFLAGS` situation outside `/usr/local` because IIRC `superenv` should be doing that automagically, so don't 🚢 just yet.

mikemcquaid:
👍

DomT4:
Pulled as-is for now, but have made a note to go back and check whether the OpenSSL pointers are required when Homebrew is outside `/usr/local`.

PR #1193, opened 2016-05-16
tomcat 8.0.35 (stable), tomcat 8.5.2 (beta)

makigumo:

---

I included the newly introduced 8.5.x beta series instead of the 9.x alpha series. As per announcement of 8.5.x

> Apache Tomcat 8.5.0 is intended to replace 8.0.x and includes new features pulled forward from Tomcat 9.0.x.

### DIFF
--- a/Formula/tomcat.rb
+++ b/Formula/tomcat.rb
@@ -3,31 +3,30 @@ class Tomcat < Formula
   homepage "https://tomcat.apache.org/"
 
   stable do
-    url "https://www.apache.org/dyn/closer.cgi?path=tomcat/tomcat-8/v8.0.33/bin/apache-tomcat-8.0.33.tar.gz"
-    mirror "https://archive.apache.org/dist/tomcat/tomcat-8/v8.0.33/bin/apache-tomcat-8.0.33.tar.gz"
-    sha256 "c77873c1861ed81617abb8bedc392fb0ff5ebf871de33cd1fcd49d4c072e38b7"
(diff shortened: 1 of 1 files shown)


DomT4:
> I included the newly introduced 8.5.x beta series instead of the 9.x alpha series. 

The only problem with this is that nobody currently on the `devel` series will be upgraded until the version number is greater than it is currently (`9.0.0.M4`).

makigumo:
Yes, this is unfortunate but I don't see a way around it. The 8.5.x series was only introduced after the 9.x development began, so this situation couldn't have been foreseen then. For what is's worth it is expected for 8.5.x to replace 8.x in the near future, which then probably will reintroduce 9.x as the devel series. Any advice on how to proceed?

dunn:
> it is expected for 8.5.x to replace 8.x in the near future

If this is going to happen in, say, a week, let's hold off until then and 8.5 can be stable and `devel` can remain at 9.x.  Otherwise I think your current solution is the best we can manage.

makigumo:
AFAIK the tomcat team is aiming for monthly releases. So depending on their voting for stable it might well take at least another month or two.

MikeMcQuaid:
@DomT4 I think with issues like this it's probably worth just setting the version number to whatever we want it to be and adding comments to the "fix this versioning edge case" issue.

DomT4:
I merged https://github.com/Homebrew/homebrew-core/pull/1358 so people weren't waiting for us to make a decision on this.

@makigumo If you wish to refile the PR to downgrade devel back to beta rather than the alpha branch we can discuss a bit more whether playing with the version on our side is the best path forwards here.

PR #1211, opened 2016-05-16
the_silver_searcher: build without xz by disabling LZMA support

chriso:

---

Make `xz` a recommended dependency of `the_silver_searcher`. Configure `the_silver_searcher` with `--disable-lzma` if installing `--without-xz`.

### DIFF
--- a/Formula/the_silver_searcher.rb
+++ b/Formula/the_silver_searcher.rb
@@ -17,7 +17,7 @@ class TheSilverSearcher < Formula
 
   depends_on "pkg-config" => :build
   depends_on "pcre"
-  depends_on "xz"
+  depends_on "xz" => :recommended
 
   def install
     # Stable tarball does not include pre-generated configure script
@@ -26,8 +26,14 @@ class TheSilverSearcher < Formula
     system "autoheader"
     system "automake", "--add-missing"
 
-    system "./configure", "--disable-dependency-tracking",
-                          "--prefix=#{prefix}"
+    args = %W[
+      --disable-dependency-tracking
+      --prefix=#{prefix}
+    ]
+
+    args << "--disable-lzma" if build.without?("xz")
+
+    system "./configure", *args
     system "make"
     system "make", "install"
 


ilovezfs:
Since pretty much everyone ends up xz installed, I'm not sure why this is useful.

chriso:
I don't use `xz`, and none of the other packages I have installed depend on `xz`, hence the PR :wink:

chriso:
I'm just following the conventions from other formulas. I can install python, curl, whatever without the things I don't need/use, so why not this too?

chriso:
FYI:

```sh
$ for pkg in $(brew uses xz); do echo $pkg && brew info $pkg | grep -- --without-xz; done
apt-cacher-ng
binwalk
dpkg
folly
gimp
gst-libav
imagemagick
innoextract
io
libarchive
--without-xz
pixz
pxz
python3
--without-xz
rpm
rpm2cpio
squashfs
termrec
the_silver_searcher
xar-mackyle
zbackup
zbar
homebrew/versions/cmake31
```

ilovezfs:
```
gettext.rb:40:                          "--without-xz"
libgetdata.rb:36:    args << "--without-liblzma" if build.without? "xz"
tarsnap.rb:33:    args << "--without-lzma" << "--without-lzmadec" if build.without? "xz"
```
are currently the only formulae with an opt out for xz. I'd be tempted to remove those options.

The 10.11 bottle for `xz` is 450K, it takes about 17s to build from source, and there are 256 formulae in core with an xz dependency.

Is xz causing problems of some kind for you?

ilovezfs:
>FYI

I'd be tempted to remove all of those.

chriso:
Why have opt-outs at all then? `xz` isn't causing problems, nor is `<some_other_package_I_don't_use>`. Should I install that too? In general, installing things that you don't need/use increases your attack surface. That's a problem.

I count 32 packages that depend on `xz`, and two of them are opt-in (including your tarsnap example above):

```sh
$ ag -l 'depends_on "xz"' | wc -l
      32
$ ag -l 'depends_on "xz" => :optional' | wc -l
       2
```

Your call, of course, but this is a pretty trivial patch that follows convention and retains the default behavior.

ilovezfs:
Anything that has an foo.tar.xz url also has an implicit xz dependency.

chriso:
Ok, well that'd be a build dependency only.

ilovezfs:
Yes, and build-time dependencies are not automatically uninstalled at the end of the build, which is why I'm saying almost everyone ends up with xz.

ilovezfs:
So I'd be curious of
```
apt-cacher-ng
binwalk
dpkg
folly
gimp
gst-libav
imagemagick
innoextract
io
pixz
pstoedit
pxz
rpm
rpm2cpio
squashfs
termrec
the_silver_searcher
xar-mackyle
zbackup
zbar
```

I wonder how many could be made to have xz as a recommended instead of required option.

chriso:
I can uninstall a build dependency though, so there are really only 30 packages that depend on `xz` being installed at all times, and 29 if this PR is merged.

ilovezfs:
You can uninstall it, but most people will not do so, so as I said, most people will have xz.

chriso:
Citation needed of course, but "pretty much" everyone ends up with packages like `readline`, `pcre` and `openssl` installed. You still provide opt-outs/opt-ins for these?

ilovezfs:
All three of those are also provided by OS X.

chriso:
So opt-in/opt-out are only provided when a package clashes with a library provided by the OS? A quick search through various packages (e.g. `python`, `curl`) shows that this isn't the case, and again there are already packages in the tree that are opt-in and opt-out for `xz`.

chriso:
So even though I've:

- shown that there aren't that many packages that need `xz` installed at all times
- given you a reason why one might not want to have the package installed at all times
- given you examples of other packages in homebrew that are both opt-in (`--with-xz`) and opt-out (`--without-xz`) of the same package
- provided a 1-1 mapping to an existing `./configure` option (`--disable-lzma`)
- stuck with homebrew conventions, and used a homebrew feature (opt-out) as intended
- kept the default `brew install the_silver_searcher` behavior

Your only argument against is that: *pretty much everyone has xz installed* ¯\_(ツ)_/¯

ilovezfs:
It has a bottle, the bottle is less than 1M, building from source instead of using the bottle takes under 30s, most people already have the dependency installed, and we try to avoid adding superfluous options. That's my "argument."

That said, since this seems to be important to you, and the dependency on xz is neither a security nor correctness requirement, I'm happy to accept this PR.

mikemcquaid:
I think if someone wants to disable building with LZMA and it's a simple dependency option I'm 👍 on including this.

ilovezfs:
Yup, agreed.

mikemcquaid:

chriso:
Awesome, thanks.

PR #1218, opened 2016-05-17
translate-shell: backport Google API fix

dunn:

---

### DIFF
--- a/Formula/translate-shell.rb
+++ b/Formula/translate-shell.rb
@@ -1,9 +1,18 @@
 class TranslateShell < Formula
   desc "Command-line translator using Google Translate and more"
   homepage "https://www.soimort.org/translate-shell"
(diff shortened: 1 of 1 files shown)


zmwangx:
No revision?

zmwangx:
And could you please remind me why setting `TERM` is necessary? Even when I unset `TERM`, I only get a warning `tput: No value for $TERM and no -T specified` on stderr, which doesn't affect the exit code.

dunn:
> No revision?

Woops, good catch.

> And could you please remind me why setting TERM is necessary?

It's just to silence that warning.

zmwangx:
> It's just to silence that warning.

It might get outdated though, if upstream ever decides to silence the warning. Then it's just one stray line that was unnecessary to begin with.

dunn:
OK. I don't have strong feelings either way so I'll remove it.

PR #1242, opened 2016-05-17
add x86 support for proxychains

firefart:
This PR adds x86 support for proxychains. See here for the error for an older version:
https://gist.github.com/y3dips/0972e13256f947b0004a

Tried with `proxychains-ng-4.11.tar.bz2`

### DIFF
--- a/Formula/proxychains-ng.rb
+++ b/Formula/proxychains-ng.rb
@@ -15,7 +15,7 @@ class ProxychainsNg < Formula
   option :universal
 
   def install
-    args = ["--prefix=#{prefix}", "--sysconfdir=#{prefix}/etc"]
+    args = ["--prefix=#{prefix}", "--sysconfdir=#{prefix}/etc", "-arch i386"]
     if build.universal?
       ENV.universal_binary
       args << "--fat-binary"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Does this build universally or always build in x86 mode? Regardless it should be guarded behind either a universal option or `MacOS.prefer_64_bit?`

FireFart:
@mikemcquaid thanks I moved the argument to the universal section

mikemcquaid:
@FireFart does this build just 32-bit or both 32 and 64-bit? Are you sure this is needed (can you let me know how you've verified it)? Thanks!

FireFart:
@mikemcquaid 
```
lipo -info /usr/local/Cellar/proxychains-ng/4.11/lib/libproxychains4.dylib
Architectures in the fat file: /usr/local/Cellar/proxychains-ng/4.11/lib/libproxychains4.dylib are: x86_64 i386
```
This builds for both x86 and x64. You can test this out by installing nmap und running through proxychains like `proxychains nmap`. The error is the same as in the linked gist. After building it again nmap runs correctly.

mikemcquaid:
This doesn't seem to be necessary. Without this PR if I run `brew install proxychains-ng --universal` I get:

```
$ lipo -info /usr/local/Cellar/proxychains-ng/4.11/lib/libproxychains4.dylib
Architectures in the fat file: /usr/local/Cellar/proxychains-ng/4.11/lib/libproxychains4.dylib are: x86_64 i386 
```

PR #1246, opened 2016-05-17
riemann 0.2.11

mnuessler:

---

### DIFF
--- a/Formula/riemann.rb
+++ b/Formula/riemann.rb
@@ -1,8 +1,8 @@
 class Riemann < Formula
   desc "Event stream processor"
   homepage "http://riemann.io"
-  url "https://aphyr.com/riemann/riemann-0.2.10.tar.bz2"
-  sha256 "ec697519e80781dc478983de04204760f2790da1715acfd233b35da071be5455"
+  url "https://aphyr.com/riemann/riemann-0.2.11.tar.bz2"
+  sha256 "5e7982002685f6fa7716782c57e227e4b4eeaa4c14a77ab586aac5ecf114e933"
 
   bottle :unneeded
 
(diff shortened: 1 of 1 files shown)


mnuessler:
I think the build fails only because it tries to download riemann several times in a row and there is some kind of download limit for a time period.

mikemcquaid:
@mnuessler A bunch of Homebrew people are going to end up doing that so we'll need to work out some solution to this.

mikemcquaid:
Perhaps finding a `mirror` we can add?

mnuessler:
Unfortunately I am not able to find a mirror. The download limit is not a problem for homebrew users, just for the build because it does several downloads within a very short time.

mikemcquaid:
There's only one download per (virtual) machine so it'll be the fact they share the same external IP address. This will be a problem for Homebrew users who share an external IP address.

mikemcquaid:
We can always consider mirroring ourselves if that's a better option.

apjanke:
A version of the releases are also hosted on their [GitHub releases](https://github.com/riemann/riemann/releases). E.g. https://github.com/riemann/riemann/archive/0.2.11.tar.gz.

apjanke:
I [asked upstream](https://github.com/riemann/riemann/issues/716) if they would include the distribution tarballs with the GitHub releases, too. That seems like an ideal mirroring situation.

apjanke:
There's [already a mirror out there](https://github.com/riemann/riemann/issues/716#issuecomment-241855800):  https://wintermute.skunkwerks.at/riemann/releases/

Added it in #4207.

MikeMcQuaid:
Replaced with https://github.com/Homebrew/homebrew-core/pull/4207 (@apjanke, probably worth using "Closes" in future subsuming PRs to autoclone).

apjanke:
Thanks.

I had the "Closes" in the PR, but since the PR gets closed by `brew pull` instead of a normal merge, it didn't trigger. Forgot to follow it manually. And I've avoided putting "Closes" in the commit comments for a PR, because IIRC, those get spuriously triggered when the test-bot runs them. Is there a good way to get auto-closing behavior to work with `brew pull`?

PR #1259, opened 2016-05-18
libirecovery 0.2.0 (new formula)

ebekebe:
- [Y ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [Y ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [Y ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [Y ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

### DIFF
--- a/Formula/libirecovery.rb
+++ b/Formula/libirecovery.rb
@@ -1,0 +1,24 @@
+class Libirecovery < Formula
+  desc "library for communication to iBoot/iBSS on iOS devices via USB."
+  homepage "https://github.com/libimobiledevice/libirecovery"
+  url "https://github.com/libimobiledevice/libirecovery/archive/master.tar.gz"
+  version "0.2.0"
+  sha256 "05c6fec3a1aff14111c4ae8252c802e2d583114474a9ae1b30e89fdb0f4edee4"
+
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "autoconf" => :build
+
+  def install
+    system "./autogen.sh", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system "ls", "#{lib}/libirecovery.dylib"
+  end
+end


Formula/libirecovery.rb
```diff
@@ -0,0 +1,24 @@
...
+    system "./autogen.sh", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system "ls", "#{lib}/libirecovery.dylib"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Formula/libirecovery.rb
```diff
@@ -0,0 +1,24 @@
+class Libirecovery < Formula
+  desc "library for communication to iBoot/iBSS on iOS devices via USB."
+  homepage "https://github.com/libimobiledevice/libirecovery"
+  url "https://github.com/libimobiledevice/libirecovery/archive/master.tar.gz"
```
mikemcquaid: This needs to be a tagged, versioned release.

Formula/libirecovery.rb
ebekebe: I am note sure how such a test should look like, since this is just a library. It should be enough if it is just there or can you point me at some similar library with "substantial" tests?

mikemcquaid: > tinyxml2.rb for an example of a library formula with a good test

😉

DomT4:
Will happily review a fresh PR when the issues here are resolved but this one seems to have gone dead, particularly around getting upstream to cut a new tag which you requested back in May, so closing 😢. Thank you nonetheless for the submission to Homebrew ❤️

PR #1281, opened 2016-05-19
logstash 5.0.0-alpha2

Jerzerak:

---
- Add logstash devel version 5.0.0-alpha2

### DIFF
--- a/Formula/logstash.rb
+++ b/Formula/logstash.rb
@@ -1,8 +1,19 @@
 class Logstash < Formula
   desc "Tool for managing events and logs"
   homepage "https://www.elastic.co/products/logstash"
-  url "https://download.elastic.co/logstash/logstash/logstash-2.3.2.tar.gz"
-  sha256 "b3c9d943fa273c8087386736ef6809df9c5959bab870a6ab4723f58d48dd38c1"
+
+  stable do
+    url "https://download.elastic.co/logstash/logstash/logstash-2.3.2.tar.gz"
+    sha256 "b3c9d943fa273c8087386736ef6809df9c5959bab870a6ab4723f58d48dd38c1"
+    depends_on :java => "1.7+"
+  end
+
+  devel do
+    url "https://download.elastic.co/logstash/logstash/logstash-5.0.0-alpha2.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/logstash.rb
```diff
@@ -9,6 +9,13 @@ class Logstash < Formula
     depends_on :java => "1.8"
   end
 
+  devel do
+    url "https://download.elastic.co/logstash/logstash/logstash-5.0.0-alpha2.tar.gz"
+    sha256 "575615d7d0e6b5dbf8805dc0a7420753e093ddd6f14cbe3229f73fb5eae14d80"
+    version "5.0.0-alpha2"
+    depends_on :java => "1.8"
+  end
```
UniqMartin: This block should be placed before `head do`. Use `brew audit --strict logstash` next time. :wink:

Formula/logstash.rb
```diff
@@ -9,6 +9,13 @@ class Logstash < Formula
     depends_on :java => "1.8"
   end
 
+  devel do
+    url "https://download.elastic.co/logstash/logstash/logstash-5.0.0-alpha2.tar.gz"
+    sha256 "575615d7d0e6b5dbf8805dc0a7420753e093ddd6f14cbe3229f73fb5eae14d80"
+    version "5.0.0-alpha2"
+    depends_on :java => "1.8"
```
UniqMartin: This was already a problem before with the HEAD version, but I just noticed it now. This line effectively means that we have two Java requirements that need to be satisfied: This one and the one from further below (`depends_on :java => "1.7+"`). It works, but isn't ideal. This can be fixed by adding a `stable do` block that looks something like this:

```ruby
stable do
  url "https://download.elastic.co/logstash/logstash/logstash-2.3.2.tar.gz"
  sha256 "b3c9d943fa273c8087386736ef6809df9c5959bab870a6ab4723f58d48dd38c1"
  depends_on :java => "1.7+"
end
```

(It replaces the `url` and `sha256` at the very top of the formula and the `depends_on` below.)

Jerzerak:
@UniqMartin done and done, let me know if there's anything else that needs doing!

UniqMartin:
Looking good! Merged. Thank you for one more contribution to Homebrew, @Jerzerak! :tada:

PR #1302, opened 2016-05-19
cairo: Added HEAD

davenquinn:

---

Added HEAD option to cairo formula

### DIFF
--- a/Formula/cairo.rb
+++ b/Formula/cairo.rb
@@ -18,7 +18,7 @@ class Cairo < Formula
   end
 
   head do
-    url "http://anongit.freedesktop.org/git/cairo", :using => :git
+    url "https://anongit.freedesktop.org/git/cairo", :using => :git
     depends_on "automake" => :build
     depends_on "autoconf" => :build
     depends_on "libtool" => :build
--- a/Formula/cairo.rb
+++ b/Formula/cairo.rb
@@ -54,10 +54,7 @@ class Cairo < Formula
       args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
     end
 
-    if build.head?
-      system "./autogen.sh", *args
-    end
-
+    system "./autogen.sh", *args if build.head?
     system "./configure", *args
     system "make", "install"
   end
--- a/Formula/cairo.rb
+++ b/Formula/cairo.rb
@@ -54,8 +54,11 @@ class Cairo < Formula
       args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
     end
 
-    system "./autogen.sh", *args if build.head?
-    system "./configure", *args
+    if build.head?
+      system "./autogen.sh", *args
+    else
+      system "./configure", *args
+    end
     system "make", "install"
   end
 
--- a/Formula/cairo.rb
+++ b/Formula/cairo.rb
@@ -17,6 +17,13 @@ class Cairo < Formula
     sha256 "268cc265a7f807403582f440643064bf52896556766890c8df7bad02d230f6c9"
   end
 
+  head do
+    url "http://anongit.freedesktop.org/git/cairo", :using => :git
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+    depends_on "libtool" => :build
+  end
+
   keg_only :provided_pre_mountain_lion
 
   option :universal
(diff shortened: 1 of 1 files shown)


Formula/cairo.rb
```diff
@@ -47,6 +54,10 @@ def install
       args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
     end
 
+    if build.head?
+      system "./autogen.sh", *args
+    end
+
```
UniqMartin: Let's make this `system "./autogen.sh", *args if build.head?` and put this directly in front of the other `system` calls for brevity.

UniqMartin: Also, the arguments that are being passed to `./configure` (the ones in `args`) are probably not the ones you will want to pass to `./autogen.sh`?!

DomT4: They might be, some `autogen` scripts immediately throw themselves into `configure`. If that's the case you can usually opt-out via a variable or you'll need to `if/else` the `autogen` and `configure` steps.

davenquinn:
I reconfigured the line as suggested. The `autogen.sh` script is supposed to take the same args as `configure` and the build is successful.

Formula/cairo.rb
```diff
@@ -47,6 +54,7 @@ def install
       args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
     end
 
+    system "./autogen.sh", *args if build.head?
```
DomT4: If this also expects the arguments to be passed directly to it it seems unlikely you want to repeat the process again below...

Formula/cairo.rb
```diff
@@ -17,6 +17,13 @@ class Cairo < Formula
     sha256 "268cc265a7f807403582f440643064bf52896556766890c8df7bad02d230f6c9"
   end
 
+  head do
+    url "http://anongit.freedesktop.org/git/cairo", :using => :git
```
DomT4: `url "https://anongit.freedesktop.org/git/cairo.git"` works fine. No need for the extra `:using => :gut` bit here.

DomT4: Er, `:using => :git` even. We don't offer support for fetching URLs with one's gut 😅.

davenquinn:
I just checked this; the repository cannot be found at
`https://anongit.freedesktop.org/git/cairo.git`; I think they are using a weird URL signature.

I did add HTTPS though.

DomT4:
Welp, you're right. Most of the `freedesktop` repos follow a convention like this:
```
https://anongit.freedesktop.org/git/pulseaudio/pulseaudio.git
```

Or like:
```
https://anongit.freedesktop.org/git/pixman.git/
```

Apparently Cairo decided to shun both those schemes.

Formula/cairo.rb
UniqMartin: Indeed, after inspecting `autogen.sh`, I'm pretty sure this should be something like:

```ruby
if build.head?
  system "./autogen.sh", *args
else
  system "./configure", *args
end
```

Otherwise we're effectively running `./configure` twice for HEAD builds.

UniqMartin:
There's also another issue (I have `transfer.fsckObjects = true` in my Git configuration). Not sure how much this should bother us, as this configuration value is not the default yet.

```
$ git clone https://anongit.freedesktop.org/git/cairo
Cloning into 'cairo'...
remote: Counting objects: 69540, done.
remote: Compressing objects: 100% (24960/24960), done.
error: object e5b831b107ad18e78ec9ea5e6da68eb30bf71126: missingSpaceBeforeDate: invalid author/committer line - missing space before date
fatal: Error in object
fatal: index-pack failed
```

DomT4:
Yeah, I run into the same thing because I use the same setting. It's pretty common to set glancing through a quick GitHub search but whether it's typical in our user base is hard to know for sure.

davenquinn:
Anything that can be done about this? I'm not familiar with that config setting

dunn:
> Anything that can be done about this?

Not really. It's probably innocuous, but users with strict integrity checking won't be able to clone the repository, and there's not much to do about it because Cairo would have to rewrite their git history to remove the invalid object.

davenquinn:
OK: fixed double configure issue and tested. There's still a problem with strict integrity checking on checkout, but that's upstream in Cairo's git repository. I think this is as good as we're going to get.

MikeMcQuaid:

PR #1304, opened 2016-05-19
Instructions for installing older formulae/packages

fsanggang:
This adds instructions for installing older versions of packages. From this [SO answer](http://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula/17757092#17757092).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,13 @@ Core formulae for the Homebrew package manager.
 ## How do I install these formulae?
 Just `brew install <formula>`. This is the default tap for Homebrew and is installed by default.
 
+### Installing older formulae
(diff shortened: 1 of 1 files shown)


DomT4:
I'd rather we don't wedge this right in the README here. It feels quite out-of-place, especially considering if you follow those steps and then come to us for support in fixing an issue caused by that you'll pretty much get told we don't officially support that mechanism.

It's worthwhile noting Homebrew did have official versioned support [similar to this](https://github.com/Homebrew/legacy-homebrew/pull/5027) for a fairly long time, but almost nobody wanted to help support it and consequently it was eventually removed. It was discussed previously [here](https://github.com/Homebrew/legacy-homebrew/issues/42036) and [here](https://github.com/Homebrew/legacy-homebrew/issues/38414) amongst other places.

mikemcquaid:
Agreed, we specifically don't want to support this but we're looking at better ways of doing this in future.

PR #1307, opened 2016-05-19
git-annex 6.20160511

### LINKED ISSUE #1268: Latest git-annex Formula Loses s3 remote
### Bug reports:

`brew install git-annex` installs 6.2.20160418.  Attempt to create an [S3 remote](http://git-annex.branchable.com/special_remotes/S3/) and you'll get the error `git-annex: Unknown remote type S3`.  

It appears the latest bottle has been compiled without S3 remote type support?

I reverted to 6.20160126, the version I was using before my latest `brew update`, and S3 remote types work again:
```
brew uninstall git-annex
brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/714f0346476022d2b874fbeb3784bc5db4e7cd98/Formula/git-annex.rb
```

ilovezfs:
git-annex 6.20160511
- version bump
- restore s3 support

The "strong" (i.e., explicit) flag is currently needed to stop Cabal's solver from dropping S3 support. There's no need to remove the explicit flag in the future unless its name changes.

Reported 20 May 2016 to git-annex upstream:
http://git-annex.branchable.com/bugs/default_cabal_install_on_OSX_lacks_S3/

aws upstream has already fixed the underlying cause:
https://github.com/aristidb/aws/commit/402bfe5aa9ef4bec84186880faafcbfdae1ad91d

New aws version (including that commit) requested 20 May 2016:
https://github.com/aristidb/aws/issues/202

Issue https://github.com/Homebrew/legacy-homebrew/issues/47737.
Closes #1268.

### DIFF
--- a/Formula/git-annex.rb
+++ b/Formula/git-annex.rb
@@ -5,8 +5,8 @@ class GitAnnex < Formula
 
   desc "Manage files with git without checking in file contents"
   homepage "https://git-annex.branchable.com/"
-  url "https://hackage.haskell.org/package/git-annex-6.20160418/git-annex-6.20160418.tar.gz"
-  sha256 "9d13586cc38d78bcd94c1f3a245d5283e67f43b0ea88033c40d54e78f6544fa2"
+  url "https://hackage.haskell.org/package/git-annex-6.20160511/git-annex-6.20160511.tar.gz"
+  sha256 "85fc8853166fe57d91dc2776d5df4acb5911a91815f8aa12881928a1afe8ba01"
 
   head "git://git-annex.branchable.com/"
 
@@ -28,7 +28,7 @@ class GitAnnex < Formula
   depends_on "quvi"
 
   def install
-    install_cabal_package :using => ["alex", "happy", "c2hs"], :flags => ["webapp"] do
+    install_cabal_package :using => ["alex", "happy", "c2hs"], :flags => ["s3", "webapp"] do
       # this can be made the default behavior again once git-union-merge builds properly when bottling
       if build.with? "git-union-merge"
         system "make", "git-union-merge", "PREFIX=#{prefix}"


ilovezfs:
@zmwangx As usual, it also works to put in a cabal.config, but I'd like your opinion on the solution in this PR. Also, strangely, adding "s3" as a flag didn't seem to help at all.

Formula/git-annex.rb
```diff
@@ -28,15 +28,29 @@ class GitAnnex < Formula
...
-        bin.install "git-union-merge"
-        system "make", "git-union-merge.1", "PREFIX=#{prefix}"
+    # Outermost `cabal_sandbox do` can be removed when >aws-0.13.0 is released
+    cabal_sandbox do
+      # Remove when >aws-0.13.0 is released
+      # Gets S3 support working again; aws upstream has already patched this:
+      # https://github.com/aristidb/aws/commit/402bfe5aa9ef4bec84186880faafcbfdae1ad91d
+      # https://github.com/Homebrew/homebrew-core/issues/1268
+      # https://github.com/Homebrew/legacy-homebrew/issues/47737
+      system Formula["cabal-install"].opt_bin/"cabal", "get", "aws"
```
ilovezfs: @mikemcquaid I'm thinking we may want `cabal_get` in `language/haskell.rb` to avoid this "manual" invocation.

ilovezfs:
Least fancy solution is
```
(buildpath/"cabal.config").write("constraints: data-default ==0.5.3\n")
```
in case anyone objects to the fix in the PR (which allows it to use `data-default 0.6`). I'll be happy either way.

zmwangx:
Wow there's no end. I'm actually in favor of the least fancy solution here. Doesn't mean I object to this PR.

ilovezfs:
@zmwangx Well we could just use lts-5.5, but where's the fun in that?

Formula/git-annex.rb
mikemcquaid: 👍 if it gets used in a couple of formulae. FWIW I think you can just use `system "cabal"` as superenv will handle the PATH correctly.

ilovezfs: Since cabal already fetches its own packages, it's the simplest way to grab something that needs patching.

ilovezfs:
Dropped the absolute cabal path.

Also, these are the upstream reports I just created:

http://source.git-annex.branchable.com/?p=source.git;a=commitdiff;h=bc87b2d9f9df8d5db5a6858a711fb6aaeefc8033

and

https://github.com/aristidb/aws/issues/202

ilovezfs:
Ah, looks like the issue here is really that `language/haskell.rb` has a bug in the flags code. Changing it to pass `system …, "-f", "webapp", "-f", "s3"` instead of `system …, "--flags='webapp s3'"` gets it to work. I'm not exactly sure why it's not happy with the latter, though, but probably something to do with the way the quoting is handled.

ilovezfs:
This fixes the overquoting: https://github.com/Homebrew/brew/pull/267

ilovezfs:
Now that Homebrew/brew#267 is merged, I've simplified the PR to pass the "s3" flag as originally proposed by @csstaub here: https://github.com/Homebrew/legacy-homebrew/issues/47737#issuecomment-213145186

Also, @DomT4 (RE: https://github.com/Homebrew/legacy-homebrew/issues/47737#issuecomment-213145801)

DomT4:
@ilovezfs Nice work here 👍. Fixing the flags is certainly a lot cleaner in-formula than having to dance. Surprised we'd not noticed that bug anywhere before now.

ilovezfs:
@DomT4 Thanks. Not too surprising actually, since this is the only formula where they've been used so far, though we should probably use them more liberally elsewhere as well to guard against feature regressions.

PR #1317, opened 2016-05-20
dpkg: fix pkg-config file installed to wrong place

ashb:
libdpkg includes a pkg-config file -- it works if I manually specified
PKG_CONFIG_PATH=$(brew --prefix dpkg)/libexec/lib/pkgconfig so this
fixes it to install the .pc file into the right default search path.

### DIFF
--- a/Formula/dpkg.rb
+++ b/Formula/dpkg.rb
@@ -40,6 +40,7 @@ class Dpkg < Formula
 
     bin.install Dir["#{libexec}/bin/*"]
     man.install Dir["#{libexec}/share/man/*"]
+    (lib/"pkgconfig").install Dir["#{libexec}/lib/pkgconfig/*.pc"]
     bin.env_script_all_files(libexec+"bin", :PERL5LIB => ENV["PERL5LIB"])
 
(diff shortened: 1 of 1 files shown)


Formula/dpkg.rb
```diff
@@ -40,6 +40,7 @@ def install
 
     bin.install Dir["#{libexec}/bin/*"]
     man.install Dir["#{libexec}/share/man/*"]
+    (lib/"pkgconfig").install Dir["#{libexec}/lib/pkgconfig/*.pc"]
```
DomT4: This should probably be a symlink to avoid potentially breaking current workflows.

ashb: Oh yes, good call. Will fix and rebase

ashb: @DomT4 added as d234b2d

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/3d0c554723a69f0278ad43d3b2d8f4df14f85f51. Thank you for your contribution to Homebrew @ashb; we appreciate it! :smiley_cat:

PR #1322, opened 2016-05-21
Delete plustache.rb

mrtazz:
I don't really have a lot of time to maintain the software. So releases have been and will be infrequent. It's also somewhat of a niche thing. So I moved the formula to my tap (https://github.com/mrtazz/homebrew-oss) and will maintain updates there. My main intention here is to take some maintenance work off your shoulders that I can somewhat easily do myself. If you think there is any reason this would be better suited to stay in homebrew-core, I'm happy to have it stay here.

Otherwise thank you for maintaining this for the last couple of years! And thank you for maintaining a package manager that has made my life tremendously easier!


dunn:
One benefit of having it in our core tap is the precompiled bottles; I don't know how long it takes to build from source.

If you do want to migrate it to your tap, though, it'd be great to add a line to `tap_migrations.json` so users can see where it's gone: https://github.com/Homebrew/homebrew-core/blob/master/tap_migrations.json

MikeMcQuaid:
Yeh, for the binary packages and discoverability I'd probably rather this stuck here.

mrtazz:
It should be somewhat quick to build, but I see the point. I'm happy to have it stay in homebrew-core in this case. Gonna send a pull request to update the version then. Thanks!

PR #1326, opened 2016-05-21
apache-brooklyn-cli 0.9.0 (new formula)

johnmccabe:

---

Adds formula for the [Apache Brooklyn `br` client](https://github.com/apache/brooklyn-client).

### DIFF
--- a/Formula/apache-brooklyn-cli.rb
+++ b/Formula/apache-brooklyn-cli.rb
@@ -3,9 +3,8 @@ require "language/go"
 class ApacheBrooklynCli < Formula
   desc "Apache Brooklyn command-line interface"
   homepage "https://brooklyn.apache.org"
-  url "https://github.com/apache/brooklyn-client.git",
-      :tag => "rel/apache-brooklyn-0.9.0",
-      :revision => "bc8593a933fcb76327ae4a511643e39d25a87ba2"
+  url "https://github.com/apache/brooklyn-client/archive/rel/apache-brooklyn-0.9.0.tar.gz"
+  sha256 "873804a145aed33de86e3928df05bad31a921f73984fed06ecdeb11e799d9c01"
 
   depends_on "go" => :build
 
@@ -43,9 +42,9 @@ class ApacheBrooklynCli < Formula
         socket = server.accept
         response = '{"version":"1.2.3","buildSha1":"dummysha","buildBranch":"1.2.3"}'
         socket.print "HTTP/1.1 200 OK\r\n" \
-                      "Content-Type: application/json\r\n" \
-                      "Content-Length: #{response.bytesize}\r\n" \
-                      "Connection: close\r\n"
+                     "Content-Type: application/json\r\n" \
+                     "Content-Length: #{response.bytesize}\r\n" \
+                     "Connection: close\r\n"
         socket.print "\r\n"
         socket.print response
         socket.close
--- a/Formula/apache-brooklyn-cli.rb
+++ b/Formula/apache-brooklyn-cli.rb
@@ -36,28 +36,27 @@ class ApacheBrooklynCli < Formula
 
   test do
     require "socket"
-    require "timeout"
 
     server = TCPServer.new("localhost", 0)
     pid_mock_brooklyn = fork do
-      Timeout.timeout(10) do
-        loop do
-          socket = server.accept
-          response = '{"version":"1.2.3","buildSha1":"dummysha","buildBranch":"1.2.3"}'
-          socket.print "HTTP/1.1 200 OK\r\n" \
+      loop do
+        socket = server.accept
+        response = '{"version":"1.2.3","buildSha1":"dummysha","buildBranch":"1.2.3"}'
+        socket.print "HTTP/1.1 200 OK\r\n" \
                       "Content-Type: application/json\r\n" \
                       "Content-Length: #{response.bytesize}\r\n" \
                       "Connection: close\r\n"
-          socket.print "\r\n"
-          socket.print response
-          socket.close
-        end
+        socket.print "\r\n"
+        socket.print response
+        socket.close
       end
     end
 
-    mock_brooklyn_url = "http://localhost:#{server.addr[1]}"
-    assert_equal "Connected to Brooklyn version 1.2.3 at #{mock_brooklyn_url}", shell_output("br login #{mock_brooklyn_url}").strip
-
-    Process.kill("KILL", pid_mock_brooklyn)
+    begin
+      mock_brooklyn_url = "http://localhost:#{server.addr[1]}"
+      assert_equal "Connected to Brooklyn version 1.2.3 at #{mock_brooklyn_url}\n", shell_output("#{bin}/br login #{mock_brooklyn_url}")
+    ensure
+      Process.kill("KILL", pid_mock_brooklyn)
+    end
   end
 end
--- a/Formula/apache-brooklyn-cli.rb
+++ b/Formula/apache-brooklyn-cli.rb
@@ -1,6 +1,4 @@
 require "language/go"
-require "socket"
-require "timeout"
 
 class ApacheBrooklynCli < Formula
   desc "Apache Brooklyn command-line interface"
@@ -37,6 +35,9 @@ class ApacheBrooklynCli < Formula
   end
 
   test do
+    require "socket"
+    require "timeout"
+
     server = TCPServer.new("localhost", 0)
     pid_mock_brooklyn = fork do
       Timeout.timeout(10) do
--- a/Formula/apache-brooklyn-cli.rb
+++ b/Formula/apache-brooklyn-cli.rb
@@ -1,0 +1,62 @@
+require "language/go"
+require "socket"
+require "timeout"
+
+class ApacheBrooklynCli < Formula
+  desc "Apache Brooklyn command-line interface"
+  homepage "https://brooklyn.apache.org"
+  url "https://github.com/apache/brooklyn-client.git",
+      :tag => "rel/apache-brooklyn-0.9.0",
+      :revision => "bc8593a933fcb76327ae4a511643e39d25a87ba2"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/codegangsta/cli" do
+    url "https://github.com/codegangsta/cli.git",
+      :revision => "5db74198dee1cfe60cf06a611d03a420361baad6"
+  end
+
+  go_resource "golang.org/x/crypto" do
+    url "https://github.com/golang/crypto.git",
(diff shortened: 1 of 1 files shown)


Aliases/br
```diff
@@ -0,0 +1 @@
+../Formula/brooklyn-client.rb
```
MikeMcQuaid: I think this is probably a little too generic and risks conflicting with another tool.

Formula/brooklyn-client.rb
```diff
@@ -0,0 +1,40 @@
...
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd "src/github.com/apache/brooklyn-client/br" do
+      system "go", "build"
+      bin.install brooklyn_client_path/"br/br"
+    end
+  end
+
+  test do
+    system "#{bin}/br", "--version"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Aliases/br
johnmccabe: Switchings not ideal, we'd explicitly named to avoid a clash with any commonly used tools and to stay consistent with the likes of `cf`, `aws` etc. Is sticking with the same name as the binary out of the question?

Formula/brooklyn-client.rb
johnmccabe: @MikeMcQuaid was in the process of updating as you commented :)

Aliases/br
MikeMcQuaid: I don't think this is as widely used as they are. `aws` is pretty unambiguous at this point.

Formula/brooklyn-client.rb
MikeMcQuaid: Great.

Formula/brooklyn-client.rb
```diff
@@ -0,0 +1,66 @@
...
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd "src/github.com/apache/brooklyn-client/br" do
+      system "go", "build"
+      bin.install brooklyn_client_path/"br/br"
+    end
+  end
+
+  test do
+    server = TCPServer.new('localhost', 0)
```
MikeMcQuaid: Use double-quoted strings here: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/2148/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_brooklyn_client___strict___online/

Formula/brooklyn-client.rb
```diff
@@ -0,0 +1,66 @@
...
+          socket.print "\r\n"
+          socket.print response
+          socket.close
+        end
+      end
+    end
+
+    mock_brooklyn_url = "http://localhost:#{server.addr[1]}"
+    output = `br login #{mock_brooklyn_url}`
+    exitcode = $?.exitstatus
```
MikeMcQuaid: Use the `shell_output` function here.

Aliases/br
johnmccabe: At least for now.. 😄 

Is there an accepted convention for naming when the command name itself isn't available, I'll need to run it past the dev list to get consensus on any alternative.

MikeMcQuaid: The current naming (`brooklyn-client`) is fine for now.

johnmccabe: @MikeMcQuaid going to go with `apache-brooklyn-cli`, just holding off on the commit for a bit to allow additional feedback on the dev list.

Do you want me to squash the commits when I'm done?

MikeMcQuaid: That's even longer. We don't need (or usually add) an alias for a new formula so let's just skip one for this PR, thanks.

MikeMcQuaid: > Do you want me to squash the commits when I'm done?

Yes, please.

johnmccabe: > That's even longer. We don't need (or usually add) an alias for a new formula so let's just skip one for this PR, thanks.

Ah sorry should have been clearer, I'm nuking the alias and renaming the class/file

MikeMcQuaid: Ah, I see. Yeh, that sounds good then 👍

johnmccabe:
Just noticed Description length is too long following name change, fixing now.

johnmccabe:
That should be ready to go now @MikeMcQuaid, thanks for the feedback.

andreaturli:
+1

Formula/apache-brooklyn-cli.rb
```diff
@@ -0,0 +1,62 @@
+require "language/go"
+require "socket"
+require "timeout"
```
UniqMartin: `socket` and `timeout` are only needed in the test, thus let's move these `require`s to the `test do` block. We want to avoid `require`ing them every time the formula is loaded (e.g. during `brew info`).

johnmccabe: Good point, have relocated them @UniqMartin

Formula/apache-brooklyn-cli.rb
```diff
@@ -0,0 +1,63 @@
...
+                      "Connection: close\r\n"
+          socket.print "\r\n"
+          socket.print response
+          socket.close
+        end
+      end
+    end
+
+    mock_brooklyn_url = "http://localhost:#{server.addr[1]}"
+    assert_equal "Connected to Brooklyn version 1.2.3 at #{mock_brooklyn_url}", shell_output("br login #{mock_brooklyn_url}").strip
```
UniqMartin: - This still needs to reference the `br` binary as `#{bin}/br`.
- Any reason you can't simply include the `\n` in the string and avoid the `.strip`?

Formula/apache-brooklyn-cli.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    Process.kill("KILL", pid_mock_brooklyn)
```
UniqMartin: It would be preferable to wrap this in a `begin`-`ensure` block so that the mock server gets killed even if the assert above fails. The general structure for these `begin`-`ensure` blocks should be as follows:

```ruby
# Start service.

begin
  # Use service.
ensure
  # Stop service.
end
```

(I believe this will also allow you to drop `Timeout.timeout` from your mock server code.)

Formula/apache-brooklyn-cli.rb
```diff
@@ -0,0 +1,63 @@
...
+    server = TCPServer.new("localhost", 0)
+    pid_mock_brooklyn = fork do
+      Timeout.timeout(10) do
+        loop do
+          socket = server.accept
+          response = '{"version":"1.2.3","buildSha1":"dummysha","buildBranch":"1.2.3"}'
+          socket.print "HTTP/1.1 200 OK\r\n" \
+                      "Content-Type: application/json\r\n" \
+                      "Content-Length: #{response.bytesize}\r\n" \
+                      "Connection: close\r\n"
```
UniqMartin: The indentation is slightly off here. (I assume you intended to align this with `"HTTP/1.1 200 OK\r\n"`?)

Formula/apache-brooklyn-cli.rb
johnmccabe: - Thats weird... brew test apache-brooklyn-cli appeared to pass for me as-is, possibly via the local $PATH, but the jenkins build job passed as well..
- Re: `\n` no specific reason, agree it makes more sense to test the output explicitly so will add it back

Formula/apache-brooklyn-cli.rb
johnmccabe: Looks good, dropped the Timeout as well, also just to be paranoid, I deleted the `/usr/local/bin/br` that was on the path and ran the test again (passed).

Formula/apache-brooklyn-cli.rb
johnmccabe: Yes, updated.
Cheers for the detailed review @UniqMartin, much appreciated !!

Formula/apache-brooklyn-cli.rb
```diff
@@ -0,0 +1,62 @@
+require "language/go"
+
+class ApacheBrooklynCli < Formula
+  desc "Apache Brooklyn command-line interface"
+  homepage "https://brooklyn.apache.org"
+  url "https://github.com/apache/brooklyn-client.git",
+      :tag => "rel/apache-brooklyn-0.9.0",
+      :revision => "bc8593a933fcb76327ae4a511643e39d25a87ba2"
```
UniqMartin: Can we use [the release tarball](https://github.com/apache/brooklyn-client/archive/rel/apache-brooklyn-0.9.0.tar.gz) instead of directly cloning from the repository? A single-file download is more resource-friendly towards both GitHub and users that build from source. It also allows us to check the integrity of the download (you'll need to add a `sha256` statement).

johnmccabe: Can you point me to an example I can crib from

johnmccabe: nvm got one

Formula/apache-brooklyn-cli.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    server = TCPServer.new("localhost", 0)
+    pid_mock_brooklyn = fork do
+      loop do
+        socket = server.accept
+        response = '{"version":"1.2.3","buildSha1":"dummysha","buildBranch":"1.2.3"}'
+        socket.print "HTTP/1.1 200 OK\r\n" \
+                      "Content-Type: application/json\r\n" \
+                      "Content-Length: #{response.bytesize}\r\n" \
+                      "Connection: close\r\n"
```
UniqMartin: It's off by one again. :wink: I guess your editor insists on using an indentation that is a multiple of two. (Don't change this if there's nothing else to be changed. I can also fix this when pulling your PR.)

johnmccabe: I mustn't have staged the change before committing, that *should* be it now

johnmccabe: Turns that its Visual Studio Code doing it, if you use tab to shift the indent level (I'd done so after removing the Timeout block) then it offsets the indent as you saw.

Sublime Text behaves itself tho.

UniqMartin:
Thanks for addressing all our feedback so swiftly!

Merged in ee0bfdbc6b811d9aed381baf9d6795e820a96644. Thank you for this contribution to Homebrew, @johnmccabe! :tada:

johnmccabe:
Thanks @UniqMartin, @MikeMcQuaid , been very impressed with the whole process.

PR #1330, opened 2016-05-21
fop: depend on Java (6 or higher)

bw-yuhogg:

---

`fop` depends on Java 1.6 or higher (source: https://xmlgraphics.apache.org/fop/2.1/compiling.html).

Homebrew doesn't compile it from source, so the JAR files _might_ work with Java 5, but I think it's reasonable to depend on 6+.

### DIFF
--- a/Formula/fop.rb
+++ b/Formula/fop.rb
@@ -11,6 +11,8 @@ class Fop < Formula
     sha256 "38602cef629a33f05149c3411ea6b82451deec872aa6cbe1fa8203ad2ee875fb" => :mavericks
   end
 
+  depends_on :java => "1.6+"
+
   resource "hyph" do
     url "https://downloads.sourceforge.net/project/offo/offo-hyphenation-utf8/0.1/offo-hyphenation-fop-stable-utf8.zip"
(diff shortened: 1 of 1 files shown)


Formula/fop.rb
```diff
@@ -3,6 +3,7 @@ class Fop < Formula
   homepage "https://xmlgraphics.apache.org/fop/index.html"
   url "https://www.apache.org/dyn/closer.cgi?path=/xmlgraphics/fop/binaries/fop-2.1-bin.tar.gz"
   sha256 "a93b59aa4d0b6d573c9090d8f21dee6c7d0c449a4bd2d48a1723e233dfb423ea"
+  revision 1
```
DomT4: We're probably fine without the `revision` bump here.

UniqMartin: Definitely fine without a `revision` bump!

UniqMartin:
Merged. Thank you for another contribution to Homebrew, @rwhogg! :tada:

> […], but I think it's reasonable to depend on 6+.

Very reasonable indeed. Even Java 7 is no longer supported by Oracle except for enterprise customers.

PR #1338, opened 2016-05-22
googler 2.4.1

zmwangx:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

https://github.com/jarun/googler/releases/tag/v2.4
- Python 2 support deprecated and no longer works without manually editing shebang; switch to Python 3.
- Use the new `--noprompt` option for noninteractive testing, <s>and unset env var `PYTHONIOENCODING` which was no longer necessary since v2.3</s>.

PR #1342, opened 2016-05-22
googler 2.4

herrbischoff:
Current formula installation does not work because of Python 2. v2.4 requires Python 3 and works.


DomT4:
There's already an open PR for this in https://github.com/Homebrew/homebrew-core/pull/1338, but thank you for the submission to Homebrew.

herrbischoff:
Ahh, sorry. I probably forgot to look thoroughly enough.

PR #1346, opened 2016-05-22
luajit: apply upstream patch to devel

### LINKED ISSUE #1274: luajit --devel compiled with options that break correctness
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

The issue is tracked and explained in depth on this issue that I filed against luaJIT.
https://github.com/LuaJIT/LuaJIT/issues/180

Some aspect of the formula's build process causes the compiled artifact to use AVX instead of integer operations which clobber a register than LuaJIT expects to be available. This problem cannot be reproduced without changing the `CCOPT_x64` make flag to `-march=native` when building from source.

This leads to wildly incorrect behavior on trivial examples.

DomT4:

---

Also:
- Switch to using a versioned release for devel rather than one which moves around constantly. We're trying to move in this direction because the end goal is to bottle devel as well, which we can't do if those can move around.
- Fix the dylib vanilla symlinking to work for stable/devel.
- Minor code compacting.

Closes #1274.

### DIFF
--- a/Formula/luajit.rb
+++ b/Formula/luajit.rb
@@ -14,8 +14,14 @@ class Luajit < Formula
   end
 
   devel do
-    url "http://luajit.org/git/luajit-2.0.git", :branch => "v2.1"
-    version "2.1"
+    url "http://luajit.org/download/LuaJIT-2.1.0-beta2.tar.gz"
+    sha256 "713924ca034b9d99c84a0b7b701794c359ffb54f8e3aa2b84fad52d98384cf47"
+
+    # https://github.com/LuaJIT/LuaJIT/issues/180
+    patch do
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍 :shipit:?

DomT4:
Hoping someone who reported the original issue can help test the fix since I can't reproduce the original error locally even before the change.

MikeMcQuaid:
@DomT4 I'd maybe just 🚢 it and hope.

DomT4:
If I don't hear anything back over the next 24 hours or so I'll push it and see what happens in terms of response.

PR #1361, opened 2016-05-23
keybase: add gnupg dependency

kevit:
Gnupg was not in dependencies, but must be presented in the system for keybase

### DIFF
--- a/Formula/keybase.rb
+++ b/Formula/keybase.rb
@@ -14,6 +14,7 @@ class Keybase < Formula
   end
 
   depends_on "go" => :build
+  depends_on "gnupg"
 
   def install
     ENV["GOPATH"] = buildpath


DomT4:
@BrewTestBot test this please

DomT4:
This didn't use to be the case, and I believe that hasn't changed. What's happening for you locally?

screen shot 2016-05-23 at 11 42 01

kevit:
for me _keybase login_ not working without gpg

MikeMcQuaid:
CC @patrickxb

patrickxb:
There is one path in the device provisioning process that will require gpg for some users.  It is only required for the first device they provision, but if they don't have gpg installed on the device, we tell them to install it.

So for the vast majority of cases, gpg is not required, and once a device is provisioned, gpg is not required to use keybase.

UniqMartin:
Thanks for the PR, @kevit, but passing on it per @patrickxb's comment.

PR #1314, opened 2016-05-20
dspdfviewer 1.15

wwwdata:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

Updated the version to 1.15 and now also compiles with qt5

PR #1365, opened 2016-05-23
poppler 0.44.0

edporras:

---


UniqMartin:
Merged in 575786b98191576a55512bf9762e82d91db30e64. Thank you for this contribution to Homebrew, @edporras! :tada:

----

**Note to other maintainers:** Pulled this despite the build failure as that one is either a bug in our `test-bot` code or in our dependency resolution logic. Steps to reproduce locally:

- Install `poppler` without any options.
- Install `dspdfviewer`. This will cause a from-source build of `poppler` with `--with-qt5`.
- Install `frescobaldi` (the failing reverse dependency). This will cause a from-source build of `poppler` with `--with-qt`, but because `poppler` is already installed with different options, the list of options gets merged and causes the equivalent of `brew install poppler --with-qt5 --with-qt`.
  - `poppler` rightfully complains: `--with-qt and --with-qt5 cannot be used at the same time`.

This wasn't an issue in the past because the `dspdfviewer` formula only very recently (#1314) switched from the Qt 4 to the Qt 5 backend of `poppler`.

PR #1366, opened 2016-05-23
genders 1.22

zioproto:
compile with --with-java-extensions=no because the build is broken.
note that current version 1.20 lead to segmentation fault when
using psdh with genders

### DIFF
--- a/Formula/genders.rb
+++ b/Formula/genders.rb
@@ -1,8 +1,8 @@
 class Genders < Formula
   desc "Static cluster configuration database for cluster management"
   homepage "https://computing.llnl.gov/linux/genders.html"
-  url "https://downloads.sourceforge.net/project/genders/genders/1.20-1/genders-1.20.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/genders.rb
```diff
@@ -16,6 +16,8 @@ class Genders < Formula
   def install
     args = ["--prefix=#{prefix}"]
 
+    args << "--with-java-extensions=no"
```
UniqMartin: Since this isn't a conditional `configure` argument (unlike the one just below), you can merge it with the initialization of the `args` array from above like this:

```ruby
args = %W[
  --prefix=#{prefix}
  --with-java-extensions=no
]
```

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/0870f9905bfe9997411ba46e0f0c6ee3fae9d531. Thank you for your contribution to Homebrew @zioproto; we appreciate it! :smiley_cat:

Formula/genders.rb
DomT4: Changed this in post.

PR #1367, opened 2016-05-23
lua: move sigaction patch to patches repository

maxim-belkin:

---

Currently, there is something wrong with [lua-users](http://lua-users.org/files) website that stores the patch, so the url for the patch specified in the formula is no longer valid. 
I found the patch at http://lua-users.org/files/wiki_insecure/jZgoWXtW/5.2/lua-5.2.3-sig_catch.patch and copied it to gist.github.com. That's it.

### DIFF
--- a/Formula/lua.rb
+++ b/Formula/lua.rb
@@ -45,7 +45,8 @@ class Lua < Formula
   # sigaction provided by posix signalling power patch
   if build.with? "sigaction"
     patch do
-      url "http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch"
+		 # url "http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch"
+			url "https://gist.githubusercontent.com/anonymous/99614c27bcac2aa86181fabe78c85915/raw/7abdf0234ffd38fc9094078d8128938f5e2accd4/lua-5.2.3-sig_catch.patch"
       sha256 "f2e77f73791c08169573658caa3c97ba8b574c870a0a165972ddfbddb948c164"
     end
   end
--- a/Formula/lua.rb
+++ b/Formula/lua.rb
@@ -45,8 +45,8 @@ class Lua < Formula
   # sigaction provided by posix signalling power patch
   if build.with? "sigaction"
     patch do
-		 # url "http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch"
-			url "https://gist.githubusercontent.com/anonymous/99614c27bcac2aa86181fabe78c85915/raw/7abdf0234ffd38fc9094078d8128938f5e2accd4/lua-5.2.3-sig_catch.patch"
+      # url "http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch"
+      url "https://gist.githubusercontent.com/anonymous/99614c27bcac2aa86181fabe78c85915/raw/7abdf0234ffd38fc9094078d8128938f5e2accd4/lua-5.2.3-sig_catch.patch"
       sha256 "f2e77f73791c08169573658caa3c97ba8b574c870a0a165972ddfbddb948c164"
     end
   end
--- a/Formula/lua.rb
+++ b/Formula/lua.rb
@@ -45,8 +45,8 @@ class Lua < Formula
   # sigaction provided by posix signalling power patch
   if build.with? "sigaction"
     patch do
-      # url "http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch"
-      url "https://gist.githubusercontent.com/anonymous/99614c27bcac2aa86181fabe78c85915/raw/7abdf0234ffd38fc9094078d8128938f5e2accd4/lua-5.2.3-sig_catch.patch"
+      # original patch file (not available): http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch
+      url "https://raw.githubusercontent.com/Homebrew/patches/d674e02d1097b21032198854080204680b616b61/lua/lua-5.2.3-sig_catch.patch"
       sha256 "f2e77f73791c08169573658caa3c97ba8b574c870a0a165972ddfbddb948c164"
     end
   end


Formula/lua.rb
```diff
@@ -45,7 +45,8 @@ class Lua < Formula
   # sigaction provided by posix signalling power patch
   if build.with? "sigaction"
     patch do
-      url "http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch"
+      # url "http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch"
+      url "https://gist.githubusercontent.com/anonymous/99614c27bcac2aa86181fabe78c85915/raw/7abdf0234ffd38fc9094078d8128938f5e2accd4/lua-5.2.3-sig_catch.patch"
```
DomT4: Mind submitting this to https://github.com/Homebrew/patches instead?

maxim-belkin:
```
$ brew tap -v homebrew/patches
==> Tapping homebrew/patches
git clone https://github.com/Homebrew/homebrew-patches [linuxbrew]/Library/Taps/homebrew/homebrew-patches --depth=1 --config core.autocrlf=false
Cloning into '[linuxbrew]/Library/Taps/homebrew/homebrew-patches'...
Username for 'https://github.com': maxim-belkin
Password for 'https://maxim-belkin@github.com': 
remote: Repository not found.
fatal: repository 'https://github.com/Homebrew/homebrew-patches/' not found
Error: Failure while executing: git clone https://github.com/Homebrew/homebrew-patches [linuxbrew]/Library/Taps/homebrew/homebrew-patches --depth=1 --config core.autocrlf=false
```

maxim-belkin: I guess I'll just fork it and commit a PR...

DomT4: Yeah, it's not really intended to be tapped to be honest, although as an idea that's an interesting one. It's more intended as a standard fork/commit deal.

MikeMcQuaid: @DomT4 Maybe worth renaming to `patches-homebrew`?

DomT4: @MikeMcQuaid Yeah, that's a good idea. I presume the redirect will follow okay for existing patches?

MikeMcQuaid: @DomT4 Aye.

UniqMartin: @MikeMcQuaid @DomT4 What's your thinking here? Isn't one “Homebrew” (the one from the organization part) in the repository URL sufficient? I don't really understand what the longer repository name buys us …

DomT4: From my POV it breaks up the format we do use for tappable public taps.

Pretty much everything intended to be tapped is `homebrew/homebrew-xyz` or more commonly written in terminals as `homebrew/x11` for example. I think that usage can give the impression everything that you can `brew tap homebrew/xyz` is supposed to be tapped.

I agree it's not ideal, but I think it's slightly more obvious that things named `homebrew/brew-evolution` aren't supposed to be tappable.

maxim-belkin: Make homebrew/patches tappable => problem solved.
Add a note to README.md => problem solved.
Rename the repo => more things to know/remember

UniqMartin: Not that I need to be convinced, but this didn't really convince me. I'm aware that we have a certain ambiguity thanks to our tap-to-repository mapping, but I don't feel this rename improves on this. Mostly, I'm a bit meh on renaming repositories on a whim, as I think they should be pretty stable identifiers and the rename has the potential to cause its own share of confusion. (Thankfully redirects work very well across GitHub.)

DomT4: @maxim-belkin Adding a note to the README is easy enough to do, but there's no user upside to putting the work into making it tappable to users since we're not going to put the work into rewriting formulae to fetch offline patches rather than ones from GH.

UniqMartin:
Merged. Thank you for this contribution to Homebrew, @maxim-belkin! :tada:

Formula/lua.rb
MikeMcQuaid: @UniqMartin It's not "on a whim", it's because multiple users have been confused by this. As you've said: GitHub provides redirects.

@DomT4 `formulae-patches` is perhaps my preference given these are never patches for core itself.

DomT4: The alternative to playing with renames is introducing tap blacklisting, or a system where if you tap something under the `Homebrew` user isn't included in the `OFFICIAL_TAPS` output it raises with a message, but I don't know how quickly I could find the time to get that done & properly tested.

UniqMartin: @MikeMcQuaid It's been proposed and implemented in less than three hours, based on an exchange between two maintainers that is buried in the code comment of a formula PR. That's why I chose “on a whim”. Maybe I'm asking for too much bureaucracy/communication here.

I'm not objecting a rename at all costs, but as I stated the benefit wasn't clear to me and the initial new name `patches-homebrew` was more confusing. For what it's worth, I did like the briefly used `brew-patches`, though that might carry the wrong message that these are patches for the `brew` repository (they are not). Of all the names suggested so far `formula-patches` is probably my favorite (no plural “e” for readability and I believe also grammar).

I'm still not entirely convinced that this is such a pressing issue, as few contributors have to work with the patches repository and (I believe) few are confused by it. A rename will probably just shift confusion from one group of people to another, at the expense of those who have already familiarized themselves with it and will be surprised to no longer find it in our repository list.

DomT4: To be fair, all commit discussions are pinged onto the team chat, so it's not like it was intentionally buried, and the change is so incredibly minor in terms of consequences; URLs still work, existing clones still work, and you can rename it to something else or rename it back to what it was pretty much endlessly without issue.

Short of deleting the entire repo, or force-pushing over it with emptiness, it's really pretty hard to break patches for the current use case.

MikeMcQuaid: > Of all the names suggested so far formula-patches is probably my favorite (no plural “e” for readability and I believe also grammar).

`formula-patches` it is. OK with you @DomT4?

> A rename will probably just shift confusion from one group of people to another, at the expense of those who have already familiarized themselves with it and will be surprised to no longer find it in our repository list.

Those who search for it will see it with the a similar name and all previous links redirect.

DomT4: > formula-patches it is. OK with you @DomT4?

No objection here.

UniqMartin: I believe I had sufficient room to raise awareness for some thoughts I had regarding this rename and don't intend to discuss this for discussion's sake. I'm fine with `formula-patches`.

UniqMartin: I apologize if my comments sounded dismissive. They were not intended to, so I'm blaming this on the medium (text) and my inability to express my thoughts adequately. My initial comment was sparked by genuine curiosity as to how the rename would help and the conviction that an already well established repository shouldn't be renamed lightly, even if the repository in question is probably among the least critical to Homebrew and mostly invisible to its users and contributors.

(Just wanted to hopefully clarify a few things. I'm not expecting to see a response and not at all intending to re-spark the discussion.)

PR #676, opened 2016-04-28
gcc 6.1.0

zbeekman:
Update gcc formula for 6.1 release. Passes tests locally, and brew audit --strict etc.

AFAICT ISL, cloog, etc. are all fine as is.

PR #1368, opened 2016-05-23
Boneyard GASNet and sam2p

zbeekman:
See PR #676 for some additional context.

This PR goes hand in hand with https://github.com/Homebrew/homebrew-boneyard/pull/120

GASNet is pretty broken and since it is blocking GCC 6.1 I think it should either be boneyard-ed or completely dropped until I can determine a sane way to patch/install it.


UniqMartin:
Merged in 05ce2548bad01807c6be2aece21ab70d221755e6. Thank you for another contribution to Homebrew, @zbeekman! :tada:

PR #1339, opened 2016-05-22
ghc 8.0.1

ilovezfs:
Also,
- run the upstream tests if building a bottle
- add a workaround for a segfault that occurs on at least Skylake
- simplify the workaround for the nm POSIX compliance bug

PR #1374, opened 2016-05-24
idris: trigger CI

ilovezfs:
**Added by UniqMartin:** See https://github.com/Homebrew/homebrew-core/pull/1339#issuecomment-221021992 for details.


ilovezfs:
0.11.2 seems to be OK: https://github.com/Homebrew/homebrew-core/pull/1382

PR #1256, opened 2016-05-18
breakpad 1.0.0 (new formula)

flier:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

PR #1377, opened 2016-05-24
butteraugli 0.1 (new formula)

nosyjoe:

---

Butteraugli is a tool for measuring differences between images developed by
Google.

### DIFF
--- a/Formula/butteraugli.rb
+++ b/Formula/butteraugli.rb
@@ -1,0 +1,20 @@
+class Butteraugli < Formula
+  desc "Measuring differences between images."
+  homepage "https://github.com/google/butteraugli"
+  url "https://github.com/google/butteraugli/archive/2f8d6a0b1466b003d610271ef1f0ae3dd0d28998.tar.gz"
+  version "0.1"
+  sha256 "79ce7ead4ccb159b84bbab05e8a3e569c3b98b032ee58bdfd13df25acc85d5a2"
+
+  depends_on "libpng"
+
+  def install
+    cd "src"
+    system "make"
+    bin.install "compare_pngs"
+  end
+
+  test do
+    # unfortunatley, butteraugli fails when applied to the test fixtures
+    # system "#{bin}/compare_pngs", test_fixtures("test.png"), test_fixtures("test.png")
+  end
+end


Formula/butteraugli.rb
```diff
@@ -0,0 +1,20 @@
+class Butteraugli < Formula
+  desc "Measuring differences between images."
+  homepage "https://github.com/google/butteraugli"
+  url "https://github.com/google/butteraugli/archive/2f8d6a0b1466b003d610271ef1f0ae3dd0d28998.tar.gz"
+  version "0.1"
```
UniqMartin: This looks somewhat arbitrary, but I guess that's because upstream doesn't provide any tagged releases? I think we'll need to find something better to put here. (If interested, check out the discussion in #1256, particularly the now hidden code comments, for some context.)

Formula/butteraugli.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def install
+    cd "src"
+    system "make"
+    bin.install "compare_pngs"
+  end
+
+  test do
+    # unfortunatley, butteraugli fails when applied to the test fixtures
+    # system "#{bin}/compare_pngs", test_fixtures("test.png"), test_fixtures("test.png")
```
UniqMartin: If that fails, that's a strong indication that something is broken with the build (though there might be other reasons). Can you provide more context on how exactly this test is failing for you? Ultimately, we'll want to have a test that is both non-trivial and passes successfully.

Formula/butteraugli.rb
nosyjoe: it's arbitrary, you're right. they don't have any tagged releases yet. as far as I could follow, in #1256 they have not decided on a solution on how to handle software that has no release versions yet. is it possible to use the commit id instead of a numeric version number?

Formula/butteraugli.rb
nosyjoe: the test png from the fixtures is 8x8 pixels, and butteraugli works only on pngs >= 32x32. from `butteraugli.cc`:

```
if (xsize < 32 || ysize < 32) {
  return false;  // Butteraugli is undefined for small images.
}
```

Formula/butteraugli.rb
UniqMartin: Yes, it's true that we haven't settled on a solution yet. It's a bit of a delicate matter because whether a formula needs to be upgraded is immediately tied to its version number. That's also the reason why we can't just use the commit ID, as we need something that we can compare numerically and put in an order that makes sense (this probably applies to all package managers).

We normally have a fairly strict requirement that upstream tag releases and that we only use tagged releases (though exception to this rule may have been made, even more so in the past). But we're obviously unlikely to be able to convince Google, if that's a policy across all their projects, and the project is sufficiently popular that we probably want to have it in Homebrew proper. So that calls for some pragmatic solution, that we haven't found yet.

Formula/butteraugli.rb
UniqMartin: I see. I think the best solution would be to switch the test to a short C++ program that exercises the library instead of this simple command line wrapper. It shouldn't be too difficult to generate some appropriately sized dummy images in memory. (As I'm understanding, the software is more about the library and the CLI is mostly of a demo use of that library.)

An alternative would be to generate a PNG of sufficient size and encode it directly in the `test do` block, write it to files in the test directory, and then use that. (A PNG stripped of all optional metadata and that compresses well should be doable in a few lines of Base64 encoding.)

Formula/butteraugli.rb
nosyjoe: ok, I asked the author for a release, maybe we're lucky and he does not know about the policy ;)

Formula/butteraugli.rb
nosyjoe: I'll look into that!

DomT4:
Will happily review a fresh PR if we ever have any luck getting upstream to cut a release but this one seems to have gone dead, so closing. Thank you nonetheless for the submission to Homebrew ❤️ 

You may wish to host this in your [own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) until that release becomes available, where you aren't bound by our submission guidelines.

PR #1378, opened 2016-05-24
dub 0.9.25

cmeury:

---
- upgrade to dub release 0.9.25
- remove devel target for the moment
- fake git tag to compile in version

### DIFF
--- a/Formula/dub.rb
+++ b/Formula/dub.rb
@@ -1,9 +1,9 @@
 class Dub < Formula
   desc "Build tool for D projects"
   homepage "https://code.dlang.org/about"
-  url "https://github.com/D-Programming-Language/dub/archive/v0.9.24.tar.gz"
-  sha256 "88fe9ff507d47cb74af685ad234158426219b7fdd7609de016fc6f5199def866"
-  head "https://github.com/D-Programming-Language/dub.git", :shallow => false
+  url "https://github.com/dlang/dub/archive/v0.9.25.tar.gz"
+  sha256 "e0c759d4b170cb0a57c2a4b35b700ef9e2eb4714aa27968be9403fae7bb14c86"
+  head "https://github.com/dlang/dub.git", :shallow => false
 
   bottle do
     sha256 "bf14b900869d28bc8140731ee81d04d9ee5b456603dea51353863bd76358f49d" => :el_capitan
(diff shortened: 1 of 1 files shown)


Formula/dub.rb
```diff
@@ -1,26 +1,34 @@
...
-    sha256 "577e127e5e361d1ed7fe6ef40100e91e361dd7063f304afeab7ae5a6274005de"
-    version "0.9.25-rc.1"
-  end
+  # No devel tag currently available in the repository.
+  #
+  # devel do
+  #   url "https://github.com/rejectedsoftware/dub/archive/v0.9.25-rc.1.tar.gz"
+  #   sha256 "577e127e5e361d1ed7fe6ef40100e91e361dd7063f304afeab7ae5a6274005de"
+  #   version "0.9.25-rc.1"
+  # end
```
UniqMartin: You can remove this comment. It's easy to look up this block in Git history. If there's currently no devel version and the stable version is newer anyway, just remove the `devel do` block.

Formula/dub.rb
```diff
@@ -1,26 +1,34 @@
...
   depends_on "pkg-config" => :build
   depends_on "dmd" => :build
 
   def install
+    # Create a dummy commit and tag it because build.sh uses "git describe" to compile
+    # in the application version
+    system "git", "init"
+    system "git", "add", "dub.json"
+    system "git", "commit", "-a", "-m", "'dummy commit'"
+    system "git", "tag", "-a", "-m", "'dummy message'", "v#{version}"
```
UniqMartin: This doesn't look like a solution we want to carry around. Is upstream aware of this issue (the release tarballs apparently being unusable out of the box)? Or is there a possibility this already works, but requires some argument be passed to `./build.sh`?

Formula/dub.rb
cmeury: I'll remove it.

Formula/dub.rb
cmeury: I wholeheartedly agree. Created a pull request upstream (https://github.com/dlang/dub/pull/850) and added a comment here to fix it once a new version of dub is released.

MikeMcQuaid: In the short-term we can just use a Git checkout instead of the tarball.

cmeury: Done.

UniqMartin: @cmeury Awesome, thanks! :heart:

An alternative short-term fix would have been to stick with the tarball and just `inreplace` the `GITVER` variable (e.g. like in the snippet below). But now that you've already switched this to a Git checkout, let's keep it like that to avoid going back and forth too much. (Release tarballs, IMO, have the advantage of being more resource friendly and allow us to independently validate the checksum, though GitHub is a reasonably safe source.)

```ruby
inreplace "build.sh", /^GITVER=.*$/, "GITVER=#{version}"
```

cmeury: Oh, that's another good approach. Still learning, new to brew :) Thanks for taking the time to elaborate and guide me.

UniqMartin:
Merged in 0f3b5201f5955640a2f29f15e7d72ead9cf751b3. Thank you for this contribution to Homebrew, @cmeury! :tada:

PR #1380, opened 2016-05-24
pandoc-crossref 0.2.1.3

ilovezfs:
Also,
- make `pandoc` a run-time dependency since cabal builds its own
- add `require "language/haskell"` instead of relying on the
  `depends_on "pandoc"` to do that for us as a side-effect
- use absolute paths in the test
- verify the output is as expected in the test, but use `cat -et` so
  that we don't need non-breaking spaces in the formula


MikeMcQuaid:
👍

PR #1386, opened 2016-05-24
wireshark: reference homebrew-cask wireshark-chmodbpf

jasmas:
Update wireshark formula to reference new homebrew-cask wireshark-chmodbpf which installs only the chmodbpf package from the most recent Wireshark binary distribution. This should be better than referencing the static fix attached to a bug.

### DIFF
--- a/Formula/wireshark.rb
+++ b/Formula/wireshark.rb
@@ -125,16 +125,13 @@ class Wireshark < Formula
 
   def caveats; <<-EOS.undent
     If your list of available capture interfaces is empty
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:

jasmas:
Thanks! This is easily one of my favorite projects. I'm saving time, even with updating formula and submitting pull requests over the time I'd put in without homebrew.

MikeMcQuaid:
@jasmas Thanks for the kind words ❤️

PR #1396, opened 2016-05-25
swi-prolog 7.3.21 (devel)

honnix:
- [ x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

Upgrade to latest development version

### DIFF
--- a/Formula/swi-prolog.rb
+++ b/Formula/swi-prolog.rb
@@ -11,8 +11,8 @@ class SwiProlog < Formula
   end
 
   devel do
-    url "http://www.swi-prolog.org/download/devel/src/swipl-7.3.19.tar.gz"
-    sha256 "693cfeac8f31d050ddcec27e550586359e98a64dc820398997d0ce6b8f564354"
+    url "http://www.swi-prolog.org/download/devel/src/swipl-7.3.21.tar.gz"
+    sha256 "a439fada1f5f2291ec51d500abedd10f0e0112e295fea2cffd6dd8f2ae274859"
   end
 
   head do


Formula/swi-prolog.rb
```diff
@@ -69,7 +69,7 @@ def install
     # './prepare' prompts the user to build documentation
     # (which requires other modules). '3' is the option
     # to ignore documentation.
-    system "echo '3' | ./prepare" if build.head?
+    system "echo", "'3' | ./prepare" if build.head?
```
MikeMcQuaid: Almost 100% sure this will no longer work the way you expect/it did before.

DomT4: Yeah, it won't. I presume `audit` objected to the current layout?

honnix: Yes it complained. So I should change this line back then?

UniqMartin: Yes, please. Our audit check is unfortunately unable to fully understand this line, so it gives bad advice. Would be nice if we had an equivalent to `pipe_output` (often used in `test do`) that we could use in the `install` method for cases like these.

DomT4: It looks like system calls with `|` in them may supposed to be exempt [already](https://github.com/Homebrew/brew/blob/a0d008af77785071ffba27dd995e5c794042f649/Library/Homebrew/cmd/audit.rb#L908).

DomT4: (Although agree about `pipe_output`, it would be cleaner to glance over)

honnix: OK. I changed it back. Thank you guys for reviewing.

MikeMcQuaid:

Formula/swi-prolog.rb
UniqMartin: > It looks like system calls with `|` in them may supposed to be exempt [already](https://github.com/Homebrew/brew/blob/a0d008af77785071ffba27dd995e5c794042f649/Library/Homebrew/cmd/audit.rb#L908).

@DomT4 That check trips over the quotes in quotes issue, because it doesn't use a back-reference to the opening quote character to find the end and thus stops capturing before it hits the pipe (`|`):

```
swi-prolog:
  * Use `system "echo", "'` instead of `system "echo '`
```

Of course the inner single quotes are completely pointless in this case and can be omitted, which will also silence the false warning without altering behavior. (Addressed in 7039eedc2fd645f2dfa90925f745b83fb29f2199.)

DomT4: Nice, Thank you for fixing that ❤️. I'd put it on my list to look at this afternoon and it's already 5pm and I hadn't yet done so, sigh.

PR #1401, opened 2016-05-25
python3: skip the xcode-stubs patch if using --HEAD

myyc:
I was playing around with python 3.6 and it doesn't install by default (`--HEAD`) because of the `xcode-stubs` patch. Ignoring it works fine. 

Not sure whether this deserves a version bump. It's a very quick fix anyway. This is my first pull request in this project so don't be too harsh :)

### DIFF
--- a/Formula/python3.rb
+++ b/Formula/python3.rb
@@ -60,9 +60,11 @@ class Python3 < Formula
   # Fix extension module builds against Xcode 7 SDKs
   # https://github.com/Homebrew/homebrew/issues/41085
   # https://bugs.python.org/issue25136
-  patch do
-    url "https://bugs.python.org/file40478/xcode-stubs.diff"
-    sha256 "029cc0dc72b1bcf4ddc5f913cc4a3fd970378073c6355921891f041aca2f8b12"
+  stable do
+    patch do
+      url "https://bugs.python.org/file40478/xcode-stubs.diff"
+      sha256 "029cc0dc72b1bcf4ddc5f913cc4a3fd970378073c6355921891f041aca2f8b12"
+    end
   end
 
   def lib_cellar


rakkesh:
This change seems fine but `brew audit --strict --online python3` has some errors,

```
python3:
  * `stable block` (line 63) should be put before `bottle block` (line 7)
  * `url` should be put inside `stable block`
  * `checksum` should be put inside `stable block`
  * Use #{pkgshare} instead of #{share}/python3
```

For errors 1, 2 and 3, changes would be like the following,

```py
class Python3 < Formula
  desc "Interpreted, interactive, object-oriented programming language"
  homepage "https://www.python.org/"
  
  stable do
    url "https://www.python.org/ftp/python/3.5.1/Python-3.5.1.tar.xz"
    sha256 "c6d57c0c366d9060ab6c0cdf889ebf3d92711d466cc0119c441dbf2746f725c9"

    # Fix extension module builds against Xcode 7 SDKs
    # https://github.com/Homebrew/homebrew/issues/41085
    # https://bugs.python.org/issue25136
    patch do
      url "https://bugs.python.org/file40478/xcode-stubs.diff"
      sha256 "029cc0dc72b1bcf4ddc5f913cc4a3fd970378073c6355921891f041aca2f8b12"
    end
  end

  bottle do
  ...
```

For the final error, need to change this [line](https://github.com/myyc/homebrew-core/blob/9fc943a42a14920781f5964ed2bbb294a17928fc/Formula/python3.rb#L180) to 
`system "make", "frameworkinstallextras", "PYTHONAPPSDIR=#{pkgshare}"`

rodrigc:
I submitted this PR which addresses the audit issues:
https://github.com/Homebrew/homebrew-core/pull/1500

DomT4:
Merged https://github.com/Homebrew/homebrew-core/pull/1500.

PR #1407, opened 2016-05-25
git-annex: GHC 8 compatibility

ilovezfs:
All of the changes needed have been accepted upstream.

For additional details, see the upstream report here:
http://git-annex.branchable.com/bugs/ghc_8.0.1_build_fixes

Fix "Couldn't match expected type 't' with actual type ...":
https://github.com/joeyh/git-annex/commit/fe944a96d3e2b8c755970bd28641925617f19613

Also, upstream added a cabal.config to "allow-newer" for the packages "base,"
"time," and "transformers," primarily to work around a strict depedency issues
with the aws package, which just needs to bump the versions:
https://github.com/joeyh/git-annex/commit/7b61c7f5d0fe76e9d2db9a61848611eeaa66c3d4

This is applied by installing the new cabal.config file as a resource rather
than using a patch, since the patch does not apply cleanly due to prior
unrelated changes in git-annex.cabal.

Bump base from < 4.9 to < 5.0:
https://github.com/joeyh/git-annex/commit/f8e71e1a52a42260568796487fa8cd768478538a

We don't need to do this explicitly in the formula because the cabal.config
mentioned above already has this effect.

### DIFF
--- a/Formula/git-annex.rb
+++ b/Formula/git-annex.rb
@@ -5,10 +5,33 @@ class GitAnnex < Formula
 
   desc "Manage files with git without checking in file contents"
   homepage "https://git-annex.branchable.com/"
-  url "https://hackage.haskell.org/package/git-annex-6.20160511/git-annex-6.20160511.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
👍

PR #1408, opened 2016-05-25
brightness: renamed from "screenbrightness"

chdiza:

---

This also finally fixes, or at least makes no longer relevant, Homebrew/legacy-homebrew#42533

### DIFF
--- a/Aliases/brightness
+++ b/Aliases/brightness
@@ -1,1 +1,0 @@
-../Formula/screenbrightness.rb
--- a/Formula/brightness.rb
+++ b/Formula/brightness.rb
@@ -1,4 +1,4 @@
-class Screenbrightness < Formula
+class Brightness < Formula
   desc "Change OS X display brightness from the command-line"
   homepage "https://github.com/nriley/brightness"
   url "https://github.com/nriley/brightness/archive/1.2.tar.gz"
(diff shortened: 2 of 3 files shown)


UniqMartin:
The build currently fails because you forgot to remove the `brightness` alias in `Aliases/`.

> This also finally fixes, or at least makes no longer relevant, Homebrew/legacy-homebrew#42533

Could the issue there have been that you added a `brightness` formula but then never removed the `brightness` alias in `Aliases/`? I suspect that's the main reason why you have been seeing errors, since when you type `brew uninstall brightness` the alias takes precedence and continues to resolve to `screenbrightness`.

Where did you add your custom `brightness` formula? If it's been in a custom tap (I sure hope so), you should still have been able to remove it with `brew remove chdiza/<tap-name>/brightness`.

chdiza:
> The build currently fails because you forgot to remove the brightness alias in Aliases/.

Whoops, I completely forgot that existed.  I'll fix.

You may or may not be right about the old issue.  I haven't seen it in ages, because I've been using a custom, local (i.e., not a git repo) tap.  At the time of that error report, IIRC, there were no such things as custom local repos.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/7fbe9ff90a95b83b0b3a80167c1b92ce4c108788 & https://github.com/Homebrew/homebrew-core/commit/27c1e87086727039e00397db04aef696bd48f455. Thanks again @chdiza! 😃

PR #1417, opened 2016-05-25
n98-magerun 1.97.22 (new formula)

mgeoffray:
Add formula for Magento tool n98-magerun

---

### DIFF
--- a/Formula/n98-magerun.rb
+++ b/Formula/n98-magerun.rb
@@ -1,0 +1,15 @@
+class N98Magerun < Formula
+  desc "The swiss army knife for Magento developers, sysadmins and devops."
+  homepage "http://magerun.net/"
+  url "https://files.magerun.net/n98-magerun.phar"
+  version "1.0.0"
+  sha256 "199dd0d416260c68e691580ffd3f7a448d60f616529aadadcfee584992d6253e"
+
+  def install
+    bin.install "n98-magerun.phar" => "n98-magerun"
+  end
+
+  test do
+    system "#{bin}/n98-magerun", "--version"
+  end
+end
--- a/Formula/n98-magerun.rb
+++ b/Formula/n98-magerun.rb
@@ -1,5 +1,5 @@
 class N98Magerun < Formula
-  desc "The swiss army knife for Magento developers, sysadmins and devops."
+  desc "Swiss army knife for Magento developers, sysadmins and devops."
   homepage "http://magerun.net/"
   url "https://files.magerun.net/n98-magerun.phar"
   version "1.97.22"
--- a/Formula/n98-magerun.rb
+++ b/Formula/n98-magerun.rb
@@ -1,7 +1,7 @@
 class N98Magerun < Formula
   desc "Swiss army knife for Magento developers, sysadmins and devops."
   homepage "http://magerun.net/"
-  url "https://files.magerun.net/n98-magerun.phar"
+  url "https://files.magerun.net/n98-magerun-1.97.22.phar"
   version "1.97.22"
   sha256 "199dd0d416260c68e691580ffd3f7a448d60f616529aadadcfee584992d6253e"
 
--- a/Formula/n98-magerun.rb
+++ b/Formula/n98-magerun.rb
@@ -7,7 +7,7 @@ class N98Magerun < Formula
   bottle :unneeded
 
   def install
-    bin.install "n98-magerun.phar" => "n98-magerun"
+    bin.install "n98-magerun-#{version}.phar" => "n98-magerun"
   end
 
   test do
--- a/Formula/n98-magerun.rb
+++ b/Formula/n98-magerun.rb
@@ -5,11 +5,13 @@ class N98Magerun < Formula
   version "1.97.22"
   sha256 "199dd0d416260c68e691580ffd3f7a448d60f616529aadadcfee584992d6253e"
 
+  bottle :unneeded
+
   def install
     bin.install "n98-magerun.phar" => "n98-magerun"
   end
 
   test do
-    system "#{bin}/n98-magerun", "--version"
+    system "#{bin}/n98-magerun", "list"
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/n98-magerun.rb
```diff
@@ -0,0 +1,15 @@
+class N98Magerun < Formula
+  desc "The swiss army knife for Magento developers, sysadmins and devops."
+  homepage "http://magerun.net/"
+  url "https://files.magerun.net/n98-magerun.phar"
+  version "1.0.0"
```
UniqMartin: Can you comment on how you came up with this URL and version? I can see a lot of releases https://files.magerun.net/ and it doesn't look like 1.0.0 is the latest version. Actually, it seems the latest version is 1.97.22 and we'll need to reference that file explicitly. Otherwise installation will break if the file referenced by the current URL will be updated and the checksum no longer matches.

Formula/n98-magerun.rb
```diff
@@ -0,0 +1,15 @@
+class N98Magerun < Formula
+  desc "The swiss army knife for Magento developers, sysadmins and devops."
```
UniqMartin: You can shorten this a bit to `Swiss army knife for Magento developers, sysadmins and devops`.

Formula/n98-magerun.rb
```diff
@@ -0,0 +1,15 @@
...
+  homepage "http://magerun.net/"
+  url "https://files.magerun.net/n98-magerun.phar"
+  version "1.0.0"
+  sha256 "199dd0d416260c68e691580ffd3f7a448d60f616529aadadcfee584992d6253e"
+
+  def install
+    bin.install "n98-magerun.phar" => "n98-magerun"
```
UniqMartin: Does this run out of the box on OS X, with no dependencies needed? (The CI test run indicates it does.) If so, we don't need to create a bottle (binary package of a formula) here and you can add a `bottle :unneeded` before this method (with a blank line in between).

Formula/n98-magerun.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  test do
+    system "#{bin}/n98-magerun", "--version"
```
UniqMartin: Can this test be modified to do something more substantial than just `--version`? [This section of the Formula Cookbook](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula) should get you started, and I'm happy to answer any related questions you might have.

DomT4:
I'm not completely sure this isn't a better fit for https://github.com/Homebrew/homebrew-php than here, but I'll let others chip in on that.

Formula/n98-magerun.rb
mgeoffray: OK actually I wrote `1.0.0` because I did not really understand what the formula version number reference for. I will write `1.97.22`

Formula/n98-magerun.rb
mgeoffray: Yes it is a standalone file/script, the only need is PHP. Do I have to specify this or the `bottle :unneeded` is fine ?

Formula/n98-magerun.rb
mgeoffray: Actually all commands need a Magento installation to be ran. The only one that have no dependencies on Magento are `--version`, `--help` and `list` which is displaying all the available commands. I will change for `list` then.

Formula/n98-magerun.rb
```diff
@@ -0,0 +1,17 @@
+class N98Magerun < Formula
+  desc "Swiss army knife for Magento developers, sysadmins and devops."
+  homepage "http://magerun.net/"
+  url "https://files.magerun.net/n98-magerun.phar"
```
DomT4: Just to check, is this a versioned URL or does it always point to the latest release?

mgeoffray: No there are URLs for specifics versions like `https://files.magerun.net/n98-magerun-1.93.0.phar`    
The URL `https://files.magerun.net/n98-magerun.phar` is like `https://files.magerun.net/n98-magerun-latest.phar` it is the latest release.    
Do I have to remove version line and specify `https://files.magerun.net/n98-magerun-latest.phar` to make it more simple ?

MikeMcQuaid: This needs to be a URL for a specific version e.g. https://files.magerun.net/n98-magerun-1.93.0.phar

mgeoffray: OK thanks for the precisions I will make the changes during the day.   
Thank you.

mgeoffray:
I fix the version and the URL

Formula/n98-magerun.rb
```diff
@@ -0,0 +1,17 @@
+class N98Magerun < Formula
+  desc "Swiss army knife for Magento developers, sysadmins and devops."
+  homepage "http://magerun.net/"
+  url "https://files.magerun.net/n98-magerun-1.97.22.phar"
+  version "1.97.22"
```
UniqMartin: You can remove this line as the version will be auto-detected from the URL. You can check for issues like these locally by running `brew audit --strict --online n98-magerun`.

Formula/n98-magerun.rb
```diff
@@ -0,0 +1,17 @@
...
+  sha256 "199dd0d416260c68e691580ffd3f7a448d60f616529aadadcfee584992d6253e"
+
+  bottle :unneeded
+
+  def install
+    bin.install "n98-magerun.phar" => "n98-magerun"
```
UniqMartin: You'll need to adjust this line, so that it takes the versioned file name into account. (The installation currently fails.) You can use the following, so this line won't need to be changed with every update:

```ruby
bin.install "n98-magerun-#{version}.phar" => "n98-magerun"
```

mgeoffray:
Version issue fixed

UniqMartin:
Merged in 7f9de48549939e534ab3207ba81a585b45abf878. Thank you for this contribution to Homebrew, @mgeoffray! :tada:

PR #1428, opened 2016-05-26
agda: revision for GHC 8

ilovezfs:


MikeMcQuaid:
@ilovezfs I'm confused, why does this need a `revision`? Is it broken without one?

ilovezfs:
Discussion is here: https://github.com/Homebrew/homebrew-core/pull/1440#issuecomment-221805442

PR #1434, opened 2016-05-26
mighttpd2: revision for GHC 8

ilovezfs:

~~Note: there is now a 3.3.1 which I have not checked yet~~ https://github.com/Homebrew/homebrew-core/pull/1471


ilovezfs:

PR #1435, opened 2016-05-26
purescript: revision for GHC 8

ilovezfs:

~~Note: there is a pre-release https://github.com/purescript/purescript/releases/tag/v0.9.0 for which we might want to add a `:devel` spec~~ https://github.com/Homebrew/homebrew-core/pull/1474


ilovezfs:

PR #1440, opened 2016-05-26
cless: GHC 8 compatibility

ilovezfs:
We've been using a patch to raise the base requirement from < 4.8 to <5
for the sake of GHC 7.10, but without also raising the cap for
transformers, I get the following build failure under GHC 8:

```
trying: cless-0.3.0.0 (user goal)
trying: transformers-0.5.2.0/installed-0.5... (dependency of
optparse-applicative-0.12.1.0)
trying: nats-1.1 (dependency of wl-pprint-extras-3.5.0.5)
next goal: wl-pprint-terminfo (dependency of cless-0.3.0.0)
rejecting: wl-pprint-terminfo-3.7.1.3 (conflict:
transformers==0.5.2.0/installed-0.5..., wl-pprint-terminfo =>
transformers>=0.2 && <0.5)
etc. etc.
```

For simplicity I'm creating a cabal.config rather than using a patch to
set "allow-newer" for transformers. Adding base to that "allow-newer"
list makes it so that the earlier patch for just base can be removed
since it would be redundant.

We can remove base from the cabal.config if tanakh/cless#2 is merged and
we can stop creating the cabal.config entirely as soon as the
dependencies setting unnecessary restrictions on transformers catch up,
or as soon as cless itself has a built-in workaround.

### DIFF
--- a/Formula/cless.rb
+++ b/Formula/cless.rb
@@ -9,10 +9,6 @@ class Cless < Formula
   sha256 "382ad9b2ce6bf216bf2da1b9cadd9a7561526bfbab418c933b646d03e56833b2"
   revision 1
 
-  # fix compilation with GHC 7.10
-  # to be removed once https://github.com/tanakh/cless/pull/2 is merged
-  patch :DATA
-
   bottle do
     sha256 "69b6e6441633e58e2c48483b2bf6122daed6d1dfe3d7ce31525024dc0ce2d4d6" => :el_capitan
     sha256 "49b15946ec65f85e5b94333485ba8a8eee1b7ec6d2f53c4619d894c9aaf3e6a8" => :yosemite
@@ -24,6 +20,12 @@ class Cless < Formula
   depends_on "cabal-install" => :build
 
   def install
+    # GHC 8 compat
+    # Reported 25 May 2016: https://github.com/tanakh/cless/issues/3
+    # Also see "fix compilation with GHC 7.10", which has the base bump but not
+    # the transformers bump: https://github.com/tanakh/cless/pull/2
+    (buildpath/"cabal.config").write("allow-newer: base,transformers\n")
+
     install_cabal_package
   end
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'm confused: is this formula broken now https://github.com/Homebrew/homebrew-core/issues/1339 is merged?

ilovezfs:
Yes, anything that says "compatibility" is broken.

Anything that says "revision for" is validating that I'm correct that it builds without changes, and making sure that people have packages built with the latest ghc. All of them except `agda` (which has a non-build-time dep) can be closed if desired when they pass, though I figure it's best for people to be upgraded to packages built with the latest GHC, all things being equal.

MikeMcQuaid:
> Yes, anything that says "compatibility" is broken.

Ok. We should have merged these in the same PR as the GHC 8 stuff so we're not breaking things.

> Anything that says "revision for" is validating that I'm correct that it builds without changes, and making sure that people have packages built with the latest ghc. All of them except agda (which has a non-build-time dep) can be closed if desired when they pass, though I figure it's best for people to be upgraded to packages built with the latest GHC, all things being equal.

I don't think these are necessary, FWIW if the binary packages still work. I'd probably say it's worth just testing these locally or with the "Testing" CI job (but I don't think either is actually necessary). The `agda` should also be part of the GHC 8 PR.

ilovezfs:
>We should have merged these in the same PR as the GHC 8

Given that GHC 8 alone took three hours to build and there's significant variation in the extent of patching needed for each formula, that seemed like a bad idea.

MikeMcQuaid:
@ilovezfs As mentioned, those formulae don't all need to be `revision`d and the ones that do we can bump the timeout for. Given the current situation until we roll out the new versioning process: we need to ensure we don't merge PRs that knowingly break other formulae in the core/same tap.

ilovezfs:
As I mentioned in the GHC 8 PR, all necessary patches to get things happy again are here: https://github.com/ilovezfs/homebrew-ghc/commits/master

I'm opening each PR as quickly as I can ;)

MikeMcQuaid:
Sorry, I should have been more explicit in https://github.com/Homebrew/homebrew-core/pull/1339#issuecomment-221625951 that we always want to avoid merging PRs that knowingly break _anything_; particularly formulae in the same tap. I appreciate that can't be changed now but it's just so we all understand the process for next time.

ilovezfs:
> just so we all understand the process for next time.

OK.

> want to avoid merging PRs that knowingly break anything

In this case, the only thing that is broken assuming the patches are accepted is `idris`: https://github.com/idris-lang/Idris-dev/pull/3128#issuecomment-221280724

Is that one popular or boneyardy?

MikeMcQuaid:
> In this case, the only thing that is broken assuming the patches are accepted is idris: idris-lang/Idris-dev#3128 (comment)

If upstream are reasonably responsive we can try to just wait and fix it up.

ilovezfs:
Yeah, upstream has engaged in discussion in the PR I mentioned above, so that sounds fine.

ilovezfs:
:ship:?

MikeMcQuaid:
👍

PR #1445, opened 2016-05-26
hledger 0.27.1

ilovezfs:
- version bump
- builds with GHC 8 without patching

### DIFF
--- a/Formula/hledger.rb
+++ b/Formula/hledger.rb
@@ -5,8 +5,8 @@ class Hledger < Formula
 
   desc "Command-line accounting tool"
   homepage "http://hledger.org"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍

ilovezfs:
Not sure what's up with the test, but I opened an issue for that: https://github.com/simonmichael/hledger/issues/352

ilovezfs:
Looks like upstream knows what the problem is and is going to tag a new release for us! https://github.com/simonmichael/hledger/issues/351#issuecomment-221908883

MikeMcQuaid:
@ilovezfs 🎉 Happy to apply upstream commit as a patch if it makes it faster.

ilovezfs:
Woo hoo :green_apple:

simonmichael:
Thanks!

PR #1446, opened 2016-05-26
cryptol: GHC 8 compatibility

ilovezfs:
Uses an upstream commit already in Cryptol's master plus a PR opened by
Tom Kerckhove to patch the closely coupled dependency SBV:

https://github.com/GaloisInc/cryptol/pull/317
https://github.com/LeventErkok/sbv/pull/219

### DIFF
--- a/Formula/cryptol.rb
+++ b/Formula/cryptol.rb
@@ -5,10 +5,21 @@ class Cryptol < Formula
 
   desc "Domain-specific language for specifying cryptographic algorithms"
   homepage "http://www.cryptol.net/"
-  url "https://github.com/GaloisInc/cryptol.git",
-      :tag => "2.3.0",
-      :revision => "eb51fab238797dfc10274fd60c68acd4bdf53820"
+  revision 1
   head "https://github.com/GaloisInc/cryptol.git"
+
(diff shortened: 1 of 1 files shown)


ilovezfs:
:8ball: :fire: 

```
 cd("sbv-5.11") { system "patch", "-p1", "-i", buildpath/"219.diff" }
```

In the absence of the ability to add `patch do` blocks to `resource do` blocks, does anyone have a better idea? I'm perfectly happy with it as-is, but don't want to be :gun:

DomT4:
Do either of those need applying to `HEAD` in the formula?

The mechanism is kind of grim but I think it's the best we can hope for until we get resource patching sorted, so 👍.

ilovezfs:
Nice catch. The first one is from HEAD, so should be in a stable block. The one for the dependency is needed for both stable and HEAD.

MikeMcQuaid:
👍 

> In the absence of the ability to add patch do blocks to resource do blocks, does anyone have a better idea? I'm perfectly happy with it as-is, but don't want to be 🔫

That's the best thing for now but 👍 to making `resource do` support a `patch do`

PR #1449, opened 2016-05-26
appscale-tools: bump dependencies

nijikon:

---

### DIFF
--- a/Formula/appscale-tools.rb
+++ b/Formula/appscale-tools.rb
@@ -47,19 +47,19 @@ class AppscaleTools < Formula
   end
 
   resource "argparse" do
-    url "https://pypi.python.org/packages/source/a/argparse/argparse-1.3.0.tar.gz"
-    sha256 "b3a79a23d37b5a02faa550b92cbbbebeb4aa1d77e649c3eb39c19abf5262da04"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍

DomT4:
@nijikon Make sure you pull the bottles on resource updates so the bottled and from-source installations produce the same result. Thanks 😃

nijikon:
@DomT4 Will it work properly without bumping the revision?

DomT4:
Which `revision`? 😄. Bottle revisions get bumped automatically when you pull new ones.

If you mean formula revisions, unless the dependencies are vital for functionality doesn't need a `revision` bump but future installations should produce the same outcome whether you install from source or pour the bottle. Otherwise we can see issues reported where different installations are using different dependencies and it's hard to debug.

nijikon:
I meant formula revision. So basically for this pulling it with bottle would be enough. It would bump the bottle revision. Correct?

DomT4:
Yup. Just pulling the bottle in these situations is enough, unless you're fixing existing breakage, in which case bump the formula revision.

nijikon:
Very clear now, thanks!

PR #1453, opened 2016-05-26
pandoc-crossref: GHC 8 compatibility

ilovezfs:
Fixes to accommodate API changes made to template haskell in GHC 8.

In addition, I needed to lift constraints on the base and time packages
with a cabal.config, and lift the template-haskell constraint set by the
data-accessor-template dependency, which itself needed a similar patch
for the same API changes.

See the following references:
https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#template-haskell-2.11.0.0
http://git.haskell.org/ghc.git/commitdiff/575abf42e218925e456bf765abb14f069ac048a0


MikeMcQuaid:
👍 but let's make sure these changes also get submitted upstream.

PR #1456, opened 2016-05-26
pandoc: GHC 8 compatibility

ilovezfs:
Without adding "allow-newer" for the packages base and time, I get the
build failure

```
Resolving dependencies...
trying: pandoc-1.17.0.3 (user goal)
trying: base-4.9.0.0/installed-4.9... (dependency of pandoc-1.17.0.3)
next goal: haddock-library (dependency of pandoc-1.17.0.3)
rejecting: haddock-library-1.4.1 (conflict: pandoc => haddock-library>=1.1 &&
<1.3)
etc. etc.
```

Can be removed as soon as upstream and its dependencies lift the
unnecessary restrictions or pandoc itself has a builtin workaround.

### DIFF
--- a/Formula/pandoc.rb
+++ b/Formula/pandoc.rb
@@ -21,6 +21,11 @@ class Pandoc < Formula
   depends_on "gmp"
 
   def install
+    # GHC 8 compat
+    # Fixes "cabal: Could not resolve dependencies"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍

PR #1457, opened 2016-05-26
postgrest: GHC 8 compatibility

ilovezfs:
- Compiler option needed to get two of postgrest's dependencies to build
  due to increased strictness under GHC 8.
- Lift caps on base and transformers for postgrest itself and on base
  for another of its dependencies.


ilovezfs:
@MikeMcQuaid this is the last of the PRs.

MikeMcQuaid:
👍

PR #1459, opened 2016-05-26
Vdirsyncer 0.11.0

benjaminfrank:

---

Includes upstream launchd .plist

### DIFF
--- a/Formula/vdirsyncer.rb
+++ b/Formula/vdirsyncer.rb
@@ -1,8 +1,9 @@
+# coding: utf-8
 class Vdirsyncer < Formula
   desc "Synchronize calendars and contacts"
   homepage "https://github.com/pimutils/vdirsyncer"
-  url "https://pypi.python.org/packages/0b/fb/c42223e1e9169e4770194e62143d431755724b080d8cb77f14705b634815/vdirsyncer-0.10.0.tar.gz"
-  sha256 "e8b894022beab6f98bde80c919e5fa99cb72698e2327e8d5271c70d39636c8bd"
+  url "https://pypi.python.org/packages/39/e5/1e7097b5f0cd6de79ec9014f162a6000b77ca2a369ea8a1588a2eebff570/vdirsyncer-0.11.0.tar.gz"
+  sha256 "dbe1d139abf576ccf0b8220a30c80803c7ecaa52c7088241565c379cc4dcc0ad"
   head "https://github.com/pimutils/vdirsyncer"
 
   bottle do
@@ -67,11 +68,21 @@ class Vdirsyncer < Formula
 
     bin.install Dir["#{libexec}/bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+
+    prefix.install "contrib/vdirsyncer.plist"
+    inreplace prefix/"vdirsyncer.plist" do |s|
+      s.gsub! "@@WORKINGDIRECTORY@@", bin
+      s.gsub! "@@VDIRSYNCER@@", bin/name
+      s.gsub! "@@SYNCINTERVALL@@", "60"
+    end
+  end
+
+  def post_install
+    inreplace prefix/"vdirsyncer.plist", "@@LOCALE@@", ENV["LC_ALL"] || ENV["LANG"] || "en_US.UTF-8"
   end
 
   test do
     ENV["LC_ALL"] = "en_US.UTF-8"
-    ENV["LANG"] = "en_US.UTF-8"
     (testpath/".config/vdirsyncer/config").write <<-EOS.undent
       [general]
       status_path = #{testpath}/.vdirsyncer/status/
@@ -92,13 +103,14 @@ class Vdirsyncer < Formula
       BEGIN:VCARD
       VERSION:3.0
       EMAIL;TYPE=work:username@example.org
-      FN:User Name
+      FN:User Name Ö φ 風 ض
       UID:092a1e3b55
       N:Name;User
       END:VCARD
     EOS
     (testpath/".contacts/b/foo/").mkpath
+    system "#{bin}/vdirsyncer", "discover"
     system "#{bin}/vdirsyncer", "sync"
-    assert_match /BEGIN:VCARD/, (testpath/".contacts/b/foo/092a1e3b55.vcf").read
+    assert_match /Ö φ 風 ض/, (testpath/".contacts/b/foo/092a1e3b55.vcf").read
   end
 end


Formula/vdirsyncer.rb
```diff
@@ -67,6 +67,27 @@ def install
 
     bin.install Dir["#{libexec}/bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+
+    prefix.install "contrib"
```
benjaminfrank: This is my kind of workaround so I can read the upstream .plist later in ```def plist```. I tried using instance variables but that did not work. Class variables produced an empty .plist file.
Any suggestions on how this part can be improved?

DomT4:
@BrewTestBot test this please

Formula/vdirsyncer.rb
```diff
@@ -67,6 +67,27 @@ def install
...
+  def caveats; <<-EOS.undent
+      The sync intervall in the launchd .plist file is set to #{SYNC_INTERVAL}s
+      and should be adjusted to your when enabled.
+    EOS
+  end
+
+  plist_options :manual => "vdirsyncer"
+
+  def plist
+    p = File.read(prefix/"contrib/vdirsyncer.plist")
```
MikeMcQuaid: Instead use `prefix.install "contrib/vdirsyncer.plist"` above and then use `inreplace` on that file.

Formula/vdirsyncer.rb
```diff
@@ -67,6 +67,27 @@ def install
...
+    prefix.install "contrib"
+  end
+
+  SYNC_INTERVAL = 60
+
+  def caveats; <<-EOS.undent
+      The sync intervall in the launchd .plist file is set to #{SYNC_INTERVAL}s
+      and should be adjusted to your when enabled.
+    EOS
+  end
```
MikeMcQuaid: Don't need these caveats.

Formula/vdirsyncer.rb
```diff
@@ -67,6 +67,27 @@ def install
 
     bin.install Dir["#{libexec}/bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+
+    prefix.install "contrib"
+  end
+
+  SYNC_INTERVAL = 60
```
MikeMcQuaid: Just embed this in the `inreplace`.

Formula/vdirsyncer.rb
MikeMcQuaid: The `inreplace` for `LOCALE` should be in `post_install`.

MikeMcQuaid:
Installation is failing: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/2382/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_vdirsyncer/

benjaminfrank:
Not sure why the installation fails, maybe related to the plist file?

benjaminfrank:
```
Warning: The post-install step did not complete successfully
You can try again using `brew postinstall vdirsyncer`
```
Any additional info I could get on what the issue is? The only thing that comes to my mind is that `ENV['LANG']` is not defined somehow

DomT4:
I believe `LANG` is unset by default on OS X?

benjaminfrank:
Hm, any hint how to get the locale then? Otherwise I could set the locale via env params in the .plist itself and then tell vdirsyncer about it...but that sounds somewhat strange. I'd rather get whatever locale is default on osx.

UniqMartin:
What is the purpose of setting the locale in the `.plist` file? It feels a bit arbitrary to hard-code this value to whatever the user happened to have configured during installation. I would tend to use a fixed, but safe default instead (like `en_US.UTF-8`).

On a side note, the `.plist` file sets both `LANG` and `LC_ALL` and this is pretty pointless as they are both essentially the same setting, just with different priorities. `LANG` is used for all `LC_*` values that haven't been explicitly set and `LC_ALL` overrides all `LC_*` values, even if `LANG` or individual `LC_*` variables have been set to something else. (At least that's my understanding of the locale system. You can try setting and/or unsetting various combinations of these and then run `locale` to see the effect.)

benjaminfrank:
Vdirsyncer requires a proper locale set, that is locale != C. Hence I need to set the locale to whatever is the default of the system. it seems getting `ENV["LANG"]` is not how to do that. I'll try to find how to do this. Any hint is very welcome :)

about LC_ALL / LANG, you'r right, it should suffice to set one of them.

benjaminfrank:
Sadly, all of the thing I found to get the  locale return empty when brew install runs:
- osascript -e user locale of (get system info)
- defaults read .GlobalPreferences AppleLocale
However, I get the feeling that osx simply supports many locales anyway, so using something like `en_US.UTF-8`should just work. not sure what this would mean for foreign characters in the application files though

UniqMartin:
I'm not sure if I can provide a good suggestion, but both the OS X Terminal and the somewhat popular iTerm2 have an option named “Set locale environment variables on startup”. With this unchecked, none of the `LANG` and `LC_*` are set and a plain `locale` is reporting the C locale for all categories. With it checked, both set `LANG` to whatever best matches the user's configured locale on OS X, which happens to be `en_US.UTF-8` on my system. I think your best bet is to look at the output of `locale` (for the category that best matches the use case) and then take that or some fallback if this happens to be the C locale. E.g. here is my output of `locale`:

```
$ locale
LANG="en_US.UTF-8"
LC_COLLATE="C"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL=
```

Which of those lines would you tend to use? (Notice that I don't use `en_US.UTF-8` for everything.) The question you'd need know how to answer is what this locale information is used for. (I'm not familiar with it, so I cannot answer this myself.)

MikeMcQuaid:
I think relying on the system variable or falling back to `en_US.UTF-8` is probably a reasonable default 👍

UniqMartin:
I agree. That's probably the best solution without over-complicating things. In Ruby code this would be something like `ENV["LC_ALL"] || ENV["LANG"] || "en_US.UTF-8"` if you wanted to still respect that `LC_ALL` takes precedence over `LANG`.

benjaminfrank:
Ok, I incorporated your suggestions, thanks for the help! I also added some non-western chars and test for them, just to make sure the locale is a sensible choice. locally it works as it should now: pick the locale the user has set and fall back to en_US.UTF-8 (which still supports chinese/arabic/etc chars as far as I understand)

UniqMartin:
> (which still supports chinese/arabic/etc chars as far as I understand)

Yes, every `.UTF-8` locale supports this (and there are hardly any non-UTF-8 locales in use nowadays), as the displayable characters really only depend on the character encoding used, not on the selected language. The language itself then affects sorting order, date/time formatting, and a few other things.

MikeMcQuaid:
And if you haven't set a locale I think it's totally reasonable for us to default to English (like most other software).

MikeMcQuaid:
Thanks again @benjaminfrank for this PR, your previous contributions and your patience and hard work getting this merged 👏

Formula/vdirsyncer.rb
```diff
@@ -1,8 +1,9 @@
+# coding: utf-8
```
DomT4: Minor belated notation, this shouldn't be necessary any more. We've logic rammed into the core to treat formulae files as UTF-8.

benjaminfrank: That is most likely a leftover from me simply copying some other Formula as basis for this one. I'll try to remember to remove it in the next version bump

DomT4: I'll squash it out of existence in a minute. Was just an FYI really 😺

DomT4: https://github.com/Homebrew/homebrew-core/commit/e63d284c793debd8b5011973caf7060a635c0c5b

PR #1466, opened 2016-05-27
noti 2.2.0

variadico:

---

This updates `noti` to version 2.2.0.

### DIFF
--- a/Formula/noti.rb
+++ b/Formula/noti.rb
@@ -1,11 +1,10 @@
 require "language/go"
 
 class Noti < Formula
-  desc "Displays a notification after a terminal process finishes."
+  desc "Trigger notifications when a process completes"
   homepage "https://github.com/variadico/noti"
-  url "https://github.com/variadico/noti/archive/v2.1.1.tar.gz"
-  sha256 "c31c763eb92b4600371b2a62f6dc1dbb3390d89c09e0baab7af0dc427b55a8c5"
-  head "https://github.com/variadico/noti.git", :branch => "dev"
+  url "https://github.com/variadico/noti/archive/v2.2.0.tar.gz"
+  sha256 "3acb1cb0c352e6387b172867e5187f9241b66f9104d95c93ad8dc9a908937626"
 
   bottle do
     cellar :any_skip_relocation
@@ -17,15 +16,14 @@ class Noti < Formula
   depends_on "go" => :build
 
   def install
-    mkdir_p "#{buildpath}/src/github.com/variadico"
-    ln_s buildpath, "#{buildpath}/src/github.com/variadico/noti"
     ENV["GOPATH"] = buildpath
-    cd "#{buildpath}/src/github.com/variadico/noti" do
-      if build.stable?
-        system "go", "build", "-o", "#{bin}/noti"
-      else
-        system "go", "build", "-o", "#{bin}/noti", "cmd/noti/main.go", "cmd/noti/osx.go"
-      end
+
+    notipath = buildpath/"src/github.com/variadico/noti"
+    notipath.install Dir["*"]
+
+    cd "src/github.com/variadico/noti/cmd/noti" do
+      system "go", "build"
+      bin.install "noti"
     end
   end
 


Formula/noti.rb
```diff
@@ -17,19 +16,20 @@ class Noti < Formula
   depends_on "go" => :build
 
   def install
-    mkdir_p "#{buildpath}/src/github.com/variadico"
-    ln_s buildpath, "#{buildpath}/src/github.com/variadico/noti"
```
MikeMcQuaid: Does the symlink approach no longer work?

Formula/noti.rb
```diff
@@ -17,19 +16,20 @@ class Noti < Formula
...
+
+    cd "src/github.com/variadico/noti/cmd/noti" do
+      system "go", "build"
+      bin.install "noti"
     end
   end
 
   test do
-    system "#{bin}/noti", "-t", "Noti", "-m", "'Noti recipe installation test has finished.'"
+    output = shell_output(bin/"noti -version")
```
MikeMcQuaid: Please use the existing, better test instead.

Formula/noti.rb
variadico: It probably still did. This just seemed more readable to me.

Formula/noti.rb
variadico: Ah, okay. I'm not familiar with how you guys do automated testing. The original test actually creates an OS X notification. Does Homebrew verify that notification during your tests? Or does a human run each test to verify it works?

Formula/noti.rb
```diff
@@ -1,11 +1,10 @@
 require "language/go"
 
 class Noti < Formula
-  desc "Displays a notification after a terminal process finishes."
+  desc "Trigger notifications when a process completes"
   homepage "https://github.com/variadico/noti"
-  url "https://github.com/variadico/noti/archive/v2.1.1.tar.gz"
-  sha256 "c31c763eb92b4600371b2a62f6dc1dbb3390d89c09e0baab7af0dc427b55a8c5"
-  head "https://github.com/variadico/noti.git", :branch => "dev"
```
DomT4: Is removing `head` an intentional decision here?

Formula/noti.rb
DomT4: We don't verify the notification itself. Does the tool throw a bad exit code if that command fails?

variadico: `noti` is a proxy for another command. So, `noti` always returns the exit code of the program it ran. Regardless of whether `noti` itself had any problems. If the proxied command returns `1`, then `noti` will return `1`.

`noti -t Noti -m 'Noti recipe installation test has finished.'` will always succeed because it's not running any other program. If the notification fails to trigger, an error will be written to stderr, but the exit status will be `0`.

Formula/noti.rb
variadico: It is. I keep the latest release on `master` and make sure that's stable, but I don't make any guarantees for `dev`. This could be pulling bad code.

Formula/noti.rb
```diff
@@ -17,19 +16,20 @@ class Noti < Formula
...
-    ln_s buildpath, "#{buildpath}/src/github.com/variadico/noti"
     ENV["GOPATH"] = buildpath
-    cd "#{buildpath}/src/github.com/variadico/noti" do
-      if build.stable?
-        system "go", "build", "-o", "#{bin}/noti"
-      else
-        system "go", "build", "-o", "#{bin}/noti", "cmd/noti/main.go", "cmd/noti/osx.go"
-      end
+
+    # Move files to right location in GOPATH.
```
variadico: @DomT4 Do you mean this comment? I added this because I don't touch Ruby/Homebrew very often so I had to look up what `buildpath` and `install Dir` do. To me, it's not very obvious what those lines do.

Formula/noti.rb
DomT4: Okay. If you want to switch it over to use master that's fine, but if you don't want to support Homebrew uses building `head` I've no issues with that either.

Formula/noti.rb
DomT4: Yeah, I did. We use it in quite a lot of places and I'd rather not encourage people to start documenting it in every single one. That's a problem we should really resolve with core documentation rather than in-formula notations.

variadico: KK. You got it.

Formula/noti.rb
variadico: Yup, I think it'd be easier on me to not allow users to build from `head` on any branch.

Formula/noti.rb
DomT4: Ah, I see. That's interesting knowledge, thanks. Does the version output fail hard if there's an issue loading that information?

Formula/noti.rb
DomT4: 👍

Formula/noti.rb
```diff
@@ -17,19 +16,19 @@ class Noti < Formula
...
     ENV["GOPATH"] = buildpath
-    cd "#{buildpath}/src/github.com/variadico/noti" do
-      if build.stable?
-        system "go", "build", "-o", "#{bin}/noti"
-      else
-        system "go", "build", "-o", "#{bin}/noti", "cmd/noti/main.go", "cmd/noti/osx.go"
-      end
+
+    notipath = buildpath/"src/github.com/variadico/noti"
+    notipath.install Dir["{*}"]
```
DomT4: You should be able to get away with a more vanilla `notipath.install Dir["*"]` here I'd imagine.

variadico: Sweet. Nice catch!

Formula/noti.rb
variadico: Na, all it does is print a string to stdout. But, I was thinking the assert match would detect if `-version` failed to write to stdout.

MikeMcQuaid: We'd still prefer the previous test as if `noti` itself fails to run that existing test would have failed.

variadico: Okay. Switched back to original test.

MikeMcQuaid:

variadico:
Sweet! Thank you!

PR #1467, opened 2016-05-27
update tmux to 2.2 with unicode patch

joshuarubin:

---

As requested in #240

### DIFF
--- a/Formula/tmux.rb
+++ b/Formula/tmux.rb
@@ -7,7 +7,7 @@ class Tmux < Formula
     sha256 "bc28541b64f99929fe8e3ae7a02291263f3c97730781201824c0f05d7c8e19e4"
 
     patch do
-      # required for the following patch
+      # required for the following unicode patch
       url "https://github.com/tmux/tmux/commit/d303e5.patch"
       sha256 "a3ae96b209254de9dc1f10207cc0da250f7d5ec771f2b5f5593c687e21028f67"
     end
@@ -18,6 +18,14 @@ class Tmux < Formula
       url "https://github.com/tmux/tmux/commit/23fdbc.patch"
       sha256 "7ec4e7f325f836de5948c3f3b03bec6031d60a17927a5f50fdb2e13842e90c3e"
     end
+  end
+
+  bottle do
+    cellar :any
+    revision 2
+    sha256 "815920cd38a8102360f7d667271d9c724f41087dd79be433db29259390ef8011" => :el_capitan
+    sha256 "93e2156c3c7e1c9f3f4b86dd84617e7519e9bee630f1e8769e00a91aa341d274" => :yosemite
+    sha256 "03c4ca001f72a1623393c0ec9406dfd82b7e449d745762a6e761da6a95d0fbd9" => :mavericks
   end
 
   head do
@@ -47,7 +55,6 @@ class Tmux < Formula
     system "make", "install"
 
     pkgshare.install "example_tmux.conf"
-
     bash_completion.install resource("completion")
   end
 
--- a/Formula/tmux.rb
+++ b/Formula/tmux.rb
@@ -3,30 +3,21 @@ class Tmux < Formula
   homepage "https://tmux.github.io/"
 
   stable do
-    url "https://github.com/tmux/tmux/releases/download/2.1/tmux-2.1.tar.gz"
-    sha256 "31564e7bf4bcef2defb3cb34b9e596bd43a3937cad9e5438701a81a5a9af6176"
+    url "https://github.com/tmux/tmux/releases/download/2.2/tmux-2.2.tar.gz"
+    sha256 "bc28541b64f99929fe8e3ae7a02291263f3c97730781201824c0f05d7c8e19e4"
 
     patch do
-      # This fixes the Tmux 2.1 update that broke the ability to use select-pane [-LDUR]
-      # to switch panes when in a maximized pane https://github.com/tmux/tmux/issues/150#issuecomment-149466158
-      url "https://github.com/tmux/tmux/commit/a05c27a7e1c4d43709817d6746a510f16c960b4b.diff"
-      sha256 "2a60a63f0477f2e3056d9f76207d4ed905de8a9ce0645de6c29cf3f445bace12"
+      # required for the following patch
+      url "https://github.com/tmux/tmux/commit/d303e5.patch"
+      sha256 "a3ae96b209254de9dc1f10207cc0da250f7d5ec771f2b5f5593c687e21028f67"
     end
 
     patch do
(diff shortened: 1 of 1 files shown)


Formula/tmux.rb
```diff
@@ -3,32 +3,23 @@ class Tmux < Formula
...
     end
   end
 
-  bottle do
-    cellar :any
-    revision 2
-    sha256 "815920cd38a8102360f7d667271d9c724f41087dd79be433db29259390ef8011" => :el_capitan
-    sha256 "93e2156c3c7e1c9f3f4b86dd84617e7519e9bee630f1e8769e00a91aa341d274" => :yosemite
-    sha256 "03c4ca001f72a1623393c0ec9406dfd82b7e449d745762a6e761da6a95d0fbd9" => :mavericks
-  end
```
MikeMcQuaid: Please leave the bottle block as-is; we'll update it automatically when we pull.

Formula/tmux.rb
```diff
@@ -3,32 +3,23 @@ class Tmux < Formula
...
-    sha256 "31564e7bf4bcef2defb3cb34b9e596bd43a3937cad9e5438701a81a5a9af6176"
+    url "https://github.com/tmux/tmux/releases/download/2.2/tmux-2.2.tar.gz"
+    sha256 "bc28541b64f99929fe8e3ae7a02291263f3c97730781201824c0f05d7c8e19e4"
 
     patch do
-      # This fixes the Tmux 2.1 update that broke the ability to use select-pane [-LDUR]
-      # to switch panes when in a maximized pane https://github.com/tmux/tmux/issues/150#issuecomment-149466158
-      url "https://github.com/tmux/tmux/commit/a05c27a7e1c4d43709817d6746a510f16c960b4b.diff"
-      sha256 "2a60a63f0477f2e3056d9f76207d4ed905de8a9ce0645de6c29cf3f445bace12"
+      # required for the following patch
```
MikeMcQuaid: `required for the following unicode patch`

Formula/tmux.rb
```diff
@@ -50,12 +46,9 @@ def install
...
     system "make", "install"
 
-    if build.head?
-      pkgshare.install "example_tmux.conf"
-    else
-      bash_completion.install "examples/bash_completion_tmux.sh" => "tmux"
-      pkgshare.install "examples"
-    end
+    pkgshare.install "example_tmux.conf"
+
```
MikeMcQuaid: Could remove this newline.

joshuarubin:
updated with suggested changes

Formula/tmux.rb
```diff
@@ -40,6 +39,11 @@ class Tmux < Formula
   depends_on "pkg-config" => :build
   depends_on "libevent"
 
+  resource "completion" do
+    url "https://raw.githubusercontent.com/przepompownia/tmux-bash-completion/v0.0.1/completions/tmux"
```
DomT4: Where did this come from? Is it official at all?

Formula/tmux.rb
```diff
@@ -3,21 +3,20 @@ class Tmux < Formula
...
     end
 
     patch do
-      # This fixes the Tmux 2.1 update that breaks "tmux killw\; detach"
-      # https://github.com/tmux/tmux/issues/153#issuecomment-150184957
-      url "https://github.com/tmux/tmux/commit/3ebcf25149d75977ea97e9d4f786e0508d1a0d5e.diff"
-      sha256 "65a8bc0b2f6a8b41ad27605fd99419fff36314499969adc9d17dd3940a173508"
+      # workaround for bug in system unicode library reporting negative width
+      # for some valid characters
+      url "https://github.com/tmux/tmux/commit/23fdbc.patch"
```
DomT4: Can you use slightly longer commit references here? Thanks.

MikeMcQuaid: @DomT4 This is fine; if this wasn't unique GitHub would refuse to serve them so there's no attack vector given we have a checksum and you cannot "delete" commits like this (even with a force-push).

Formula/tmux.rb
MikeMcQuaid: Let's remove this from this PR, thanks.

Formula/tmux.rb
DomT4: I wasn't worried about things on the GitHub side; was more worried that if you build up enough stuff in `HOMEBREW_CACHE` _eventually_ you could run into some other patch that also is `23fdbc.patch`.

MikeMcQuaid: @DomT4 it's also named based on the formula, I believe?

DomT4: Ha, forgot that. I stylistically prefer slightly longer commit hashes, but on the technical side I'll have to shut up I guess 😄.

UniqMartin: More so, the cached file is named `<formula>--patch-<sha256><suffix>`, so the original URL doesn't even matter except for the `.patch` suffix. :wink:

Formula/tmux.rb
joshuarubin: It is officially recommended by tmux: https://github.com/tmux/tmux/blob/master/README

Formula/tmux.rb
xu-cheng: I think the full commit is still preferable. Otherwise, `brew fetch` will fail if a new commit with same prefix is added by upstream. Also this is more consistent with style in rest of our code.

MikeMcQuaid: > I think the full commit is still preferable. Otherwise, brew fetch will fail if a new commit with same prefix is added by upstream. Also this is more consistent with style in rest of our code.

@xu-cheng `brew fetch` will never fail in that situation; GitHub will fail to deliver the patch. Six characters is more than enough to be unique in all but the largest Git histories. Eight characters is definitely sufficient. I don't think this is more consistent; I've always used shorter patch URLs when possible.

xu-cheng: Sorry, but if Github fails to deliver the patch, won't that mean the `brew fetch` will fail.

MikeMcQuaid: Yes. That's not necessary here as it is unique.

Formula/tmux.rb
MikeMcQuaid: 🆒 thanks.

Formula/tmux.rb
DomT4: Will take this as-is but would prefer we at least stick to the 8 characters in future as a comfort margin.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/8d6b93af7a1b2125e15814d59242983e561bba44. Thank you for your contribution to Homebrew @joshuarubin; we appreciate it! :smiley_cat:

PR #1470, opened 2016-05-27
nvc 1.0.0 (new formula)

robertgzr:
I have have a little problem with script at line 23:

``` ruby
    system "./tools/fetch-ieee.sh"
```

It asks to respond with yes/no and this interrupts the install process.

Is there a way to go around this without asking the author for a change?

### DIFF
--- a/Formula/nvc.rb
+++ b/Formula/nvc.rb
@@ -1,0 +1,50 @@
+class Nvc < Formula
+  desc "VHDL compiler and simulator"
+  homepage "https://github.com/nickg/nvc"
+
+  stable do
+    url "https://github.com/nickg/nvc/releases/download/r1.0.0/nvc-1.0.0.tar.gz"
+    sha256 "a60478636268a7d1cad2d1c90e775cfd938199dba6b9c407badbef32753dc30f"
+
+    # Fixes nickg/nvc#265
+    # nvc needs external vhdl libraries at build
+    # PR adds https and checksums to the download script - fetch-ieee.sh
+    patch do
+      url "https://github.com/nickg/nvc/pull/296.patch"
+      sha256 "68cc9d23c286945a63464f9aa71d25d1e1b140e4652845c837cae3babbc45ab6"
+    end
+  end
+
+  head do
+    url "https://github.com/nickg/nvc.git"
+
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+  end
+
+  depends_on "llvm" => :build
+  depends_on "check" => :build
+
+  resource "vim-hdl-examples" do
+    url "https://github.com/suoto/vim-hdl-examples.git",
+    :revision => "232d28a9279bc80d2797afef212749434a78be6a"
+  end
+
+  def install
+    args = %W[
+      --with-llvm=#{Formula["llvm"].opt_bin}/llvm-config
+      --prefix=#{prefix}
+    ]
+
+    system "./autogen.sh" unless build.stable?
+    system "./tools/fetch-ieee.sh"
+    system "./configure", *args
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    resource("vim-hdl-examples").stage testpath
+    system "#{bin}/nvc", "-a", "#{testpath}/basic_library/very_common_pkg.vhd"
+  end
+end


MikeMcQuaid:
> Is there a way to go around this without asking the author for a change?

Yes, we need to duplicate the behaviour in https://github.com/nickg/nvc/blob/master/tools/fetch-ieee.sh with using `resource`s to download, checksum and cache these files ourselves rather than through this script (which, worryingly, does not checksum these files or use HTTPS).

robertgzr:
Oh alright, can you point me somewhere I can see how I would go about this.

MikeMcQuaid:
https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#specifying-gems-python-modules-go-projects-etc-as-dependencies and `grep`ping for `resource` is probably your best bet.

robertgzr:
I tried my best. What do you think?

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+  end
+
+  depends_on "llvm" => :build
+  depends_on "check" => :build
+
+
+  ieee_url="http://standards.ieee.org/downloads/1076/1076.2-1996"
```
MikeMcQuaid: Is there an HTTPS URL we can use?

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+  vital_url="http://svn.gna.org/svn/ghdl/trunk/libraries/vital2000"
```
MikeMcQuaid: This will need to be a non-`trunk` download. Also, it may be worth just using an `svn` resource and cloning this directory (depending on how much is in there).

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+  end
+  
+  resource "memory_b" do
+    url "#{vital_url}/memory_b.vhdl", :using => :curl
+    sha256 "03878edc834050f67ec5d3e0b49bef883065dc429700a174e93850274c63e458"
+  end
+  
+  def install
+    args = [
+      "--with-llvm=/usr/local/opt/llvm/bin/llvm-config",
```
MikeMcQuaid: Are you sure this is needed? It isn't autodetected by the buildsystem? Need to use `Formula["llvm"].bin` instead of hard-coding `/usr/local` if it's needed.

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+      "--prefix=#{prefix}",
+    ]
+
+    resources.each do |r|
+      r.stage buildpath/"lib/ieee"
+      
+      if r.name == "std_logic_1164"
+        opoo "Uncommenting XNOR operator in #{r.name}.vhdl"
```
MikeMcQuaid: Can remove these comments.

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+        system "mv #{buildpath}/lib/ieee/#{r.name}.vhdl #{buildpath}/lib/ieee/#{r.name}.vhdl.bak"
+        system "sed -e '119,120 s/^--/  /' -e '89 s/^--/  /' #{buildpath}/lib/ieee/#{r.name}.vhdl.bak > #{buildpath}/lib/ieee/#{r.name}.vhdl"
```
MikeMcQuaid: Can use `inreplace` for this and don't need the `.bak`. Might also be worth doing these after the `.stage`

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+        system "mv #{buildpath}/lib/ieee/#{r.name}.vhdl #{buildpath}/lib/ieee/#{r.name}.vhdl.bak"
+        system "sed -e '367,382 s/^--/  /' -e '384,399 s/^--/  /' -e '165,168 s/^--/  /' #{buildpath}/lib/ieee/#{r.name}.vhdl.bak > #{buildpath}/lib/ieee/#{r.name}.vhdl"
+      end
+    end
+
+    system "rm #{buildpath}/lib/ieee/*.vhdl.bak"
+
+    if build.head?
+      system "./autogen.sh"
+      mkdir "build"
```
MikeMcQuaid: `mkdir "build" do` will create it and `cd` to it.

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+      Dir.chdir "build"
+      system "../configure", *args
+    end
+
+    system "./configure", *args unless build.head?
```
MikeMcQuaid: Put this in an `else` instead.

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+    system "make"
+    system "make", "install"
+  end
+
+  test do
```
MikeMcQuaid: Need a test for this 😀

MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+    url "https://github.com/nickg/nvc.git"
+
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+  end
+
+  depends_on "llvm" => :build
+  depends_on "check" => :build
+
+
```
DomT4: No need for this empty line.

Formula/nvc.rb
DomT4: Style nit but would prefer it if you wrote these with spacing:
```ruby
ieee_url = "http://standards.ieee.org/downloads/1076/1076.2-1996"
```
It's easier to read at a glance.

Formula/nvc.rb
```diff
@@ -0,0 +1,140 @@
...
+    sha256 "d06dd5f5d5c4f8b19770db8e06b6302f87cb8e30e027feef201cc780152e3b8b"
+  end
+  
+  resource "memory_b" do
+    url "#{vital_url}/memory_b.vhdl", :using => :curl
+    sha256 "03878edc834050f67ec5d3e0b49bef883065dc429700a174e93850274c63e458"
+  end
+  
+  def install
+    args = [
```
DomT4: Using an `%W[]` style array is a bit neater since it allows you to cut the `""` and `,` out.

Formula/nvc.rb
DomT4: `Formula["llvm"].opt_bin`?

MikeMcQuaid: yeh 😀

Formula/nvc.rb
robertgzr: The makefile includes a `make check`

Would it be possible to use this?

MikeMcQuaid: We can use that as well but not instead. This is a runtime/packaging check.

Formula/nvc.rb
robertgzr: @MikeMcQuaid @DomT4 
I don't do a `system "sed ..."` right?

Is there a Ruby function something to edit specific lines?

MikeMcQuaid: There's not for specific lines but you could use a multline regex perhaps if that's what's needed.

robertgzr: I basically need to get rid of the `--` commenting on a few ranges of lines...
Do I need to regex the specific content on these lines to do that?

robertgzr:
What do you think about the test? I don't think it is a good idea to rely on a random repo for the test but I wanted to see if that's what you're after

Formula/nvc.rb
```diff
@@ -0,0 +1,120 @@
...
+    sha256 "10f391f90869db4a05c2087d434d270b24fa8427218b64b68a827e270c47acc1"
+  end
+
+  resource "std_logic_1164" do
+    url "#{ieee_url}/std_logic_1164.vhdl"
+    sha256 "c6359df78524b69052c8a6f8c485effbedb152b2bb5eecd0d38043b8013658c3"
+  end
+
+  resource "vital" do
+    url "http://svn.gna.org/svn/ghdl/trunk/libraries/vital2000"
```
DomT4: This will need a matching checksum, or people will see integrity warnings.

Formula/nvc.rb
```diff
@@ -0,0 +1,120 @@
...
+    else
+      system "./configure", *args
+    end
+
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system "#{Formula["wget"].bin}/wget", "https://raw.githubusercontent.com/suoto/vim-hdl-examples/master/basic_library/very_common_pkg.vhd"
```
DomT4: You won't be able to call `wget` here, it won't exist on CI for testing. Fetching in a `resource` with a commit reference and staging it in the `testpath` is probably reasonable.

Formula/nvc.rb
```diff
@@ -0,0 +1,120 @@
...
+    url "https://github.com/nickg/nvc.git"
+
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+  end
+
+  depends_on "llvm" => :build
+  depends_on "check" => :build
+
+  ieee_url="https://standards.ieee.org/downloads/1076/1076.2-1996"
```
DomT4: Same comment as before on spacing around `=` 😄

Formula/nvc.rb
robertgzr: This pulls down a directory of files... How do I calculate the checksum, every file individually?

MikeMcQuaid: @DomT4 This is an SVN checkout.

DomT4: Yeah, noticed that, just hadn't been around much today to reply to this mini-discussion yet. It should be fine then.

DomT4: Although would be nice to peg to a specific `:revision` for stability, perhaps?

MikeMcQuaid: Agreed. @robertgzr can you specify a `:revision` argument here? Thanks!

robertgzr:
I don't know what's wrong with the test block... 

It says it can't find the `test-vhdl_examples` resource
Yesterday the same code worked fine for me

Formula/nvc.rb
```diff
@@ -0,0 +1,125 @@
...
+    sha256 "c6359df78524b69052c8a6f8c485effbedb152b2bb5eecd0d38043b8013658c3"
+  end
+
+  resource "vital2000" do
+    url "http://svn.gna.org/svn/ghdl/trunk/libraries/vital2000", :revision => "151"
+  end
+
+  resource "test-vdhl_examples" do
+    url "https://github.com/suoto/vim-hdl-examples.git"
+    sha256 "bc0df6d6b694f3de4261e956042a902cf4d7f9c2f28d990c3d99f9595a48892c"
```
DomT4: This won't work as-is. You can't checksum entire git repos. Will need to do something like:

```ruby
resource "test-vdhl_examples" do
  url "https://github.com/suoto/vim-hdl-examples.git",
  :revision => "232d28a9279bc80d2797afef212749434a78be6a"
end
```

MikeMcQuaid:
@BrewTestBot test this please

Formula/nvc.rb
```diff
@@ -0,0 +1,125 @@
...
+  resource "std_logic_1164" do
+    url "#{ieee_url}/std_logic_1164.vhdl"
+    sha256 "c6359df78524b69052c8a6f8c485effbedb152b2bb5eecd0d38043b8013658c3"
+  end
+
+  resource "vital2000" do
+    url "http://svn.gna.org/svn/ghdl/trunk/libraries/vital2000", :revision => "151"
+  end
+
+  resource "test-vdhl_examples" do
```
MikeMcQuaid: Try `vim-hdl-examples` as the name instead.

robertgzr:
I assume you want all the commits to be squashed once everything is sorted out... 

Since this has been open for quite a while this branch is behind the master by some 6000 commits. 
Are you going to resolve that or should I rebase?

MikeMcQuaid:
@robertgzr Squashing and rebasing would be great, thanks.

dunn:
Could you rebase, please? Thanks!

ilovezfs:
@BrewTestBot test this please

Formula/nvc.rb
```diff
@@ -0,0 +1,125 @@
...
+    args = %W[
+      --with-llvm=#{Formula["llvm"].opt_bin}/llvm-config
+      --prefix=#{prefix}
+    ]
+
+    resources.each do |r|
+      r.stage buildpath/"lib/ieee"
+    end
+
+    inreplace "#{buildpath}/lib/ieee/std_logic_1164.vhdl" do |s|
```
dunn: Apologies if this was answered earlier in the conversation, but why are we performing these replacements and why can't they be done by upstream or by the build scripts?

robertgzr: I am replicating what the [fetch-ieee.sh](https://github.com/nickg/nvc/blob/master/tools/fetch-ieee.sh) script is doing... Basically grabbing some official libraries where the XOR operator functions are commented out and removing the comments.

If there was a way to run that script this could be removed. But it needs user interaction to run.

Another option would be to comment out the prompt there but then we lose checksumming on the external files.

MikeMcQuaid: I've filed an issue about that script: https://github.com/nickg/nvc/issues/295

I think we need a better solution to be handled upstream before this is merged; this approach with hard-coded line numbers is going to be very brittle.

MikeMcQuaid (changes requested):

Formula/nvc.rb
MikeMcQuaid: Looks like it was merged upstream. We could apply that PR as a patch and skip the `resource`s?

MikeMcQuaid: Add a `patch do` block with https://github.com/nickg/nvc/pull/296.patch as the `url`.

Formula/nvc.rb
```diff
@@ -0,0 +1,57 @@
+class Nvc < Formula
+  desc "VHDL compiler and simulator"
+  homepage "https://github.com/nickg/nvc"
+
+  stable do
+    url "https://github.com/nickg/nvc/releases/download/r1.0.0/nvc-1.0.0.tar.gz"
+    sha256 "a60478636268a7d1cad2d1c90e775cfd938199dba6b9c407badbef32753dc30f"
+
+    # fixes nickg/nvc#265
```
MikeMcQuaid: Can you please add an explanation of why the patch is needed in a bit more depth. Thanks!

Formula/nvc.rb
MikeMcQuaid: Thanks!

Formula/nvc.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def install
+    args = %W[
+      --with-llvm=#{Formula["llvm"].opt_bin}/llvm-config
+      --prefix=#{prefix}
+    ]
+
+    if build.head?
+      system "./autogen.sh"
+      mkdir "build" do
```
MikeMcQuaid: Any idea why `head` needs run in a subdirectory?

MikeMcQuaid (changes requested):

MikeMcQuaid (approved):

MikeMcQuaid:

PR #1473, opened 2016-05-27
texmath 0.8.6.3

ilovezfs:
Also,
- use absolute path to `texmath` in the test
- use the new `:flags` DSL instead of `-f executable`
- build and run the upstream test suite (only takes a couple minutes)


MikeMcQuaid:
👍

PR #1480, opened 2016-05-27
haskell-stack: bootstrap stack with stack

ilovezfs:
The idea for this commit is from Emanuel Borsboom, who has suggested that bootstrapping should help prevent Homebrew-specific regressions caused by Cabal's Solver choosing different versions of stack's
dependencies as compared to the versions used when building the upstream binaries.

Note that the stage 1 stack gathers its own resources including even the compiler. Regarding the security implications of this approach, see
https://www.fpcomplete.com/blog/2015/07/package-security-in-stack
https://www.fpcomplete.com/blog/2016/05/stack-security-gnupg-keys

A non-default option to skip bootstrapping and just install the stage 1 build is included as well.

### DIFF
--- a/Formula/haskell-stack.rb
+++ b/Formula/haskell-stack.rb
@@ -5,57 +5,11 @@ class HaskellStack < Formula
 
   desc "The Haskell Tool Stack"
   homepage "http://haskellstack.org"
-  revision 2
+  url "https://github.com/commercialhaskell/stack/releases/download/v1.1.2/stack-1.1.2-sdist-2.tar.gz"
+  version "1.1.2"
+  sha256 "8197e055451437218e964ff4a53936a497a2c1ed4818c17cf290c9a59fff9424"
+  revision 3
   head "https://github.com/commercialhaskell/stack.git"
-
-  stable do
-    url "https://github.com/commercialhaskell/stack/archive/v1.1.2.tar.gz"
-    sha256 "8f43d69a00a8861b156705a634e55179524cefbd98e6c29182e7bdcb57d8b3be"
-
-    # Fix build with newer cabal
-    # Due to per-component cabal_macros.h, the executable code doesn't have
-    # access to VERSION_hpack
-    patch do
-      url "https://github.com/commercialhaskell/stack/commit/93cedb7ce20d1fcc4f46d4b0b38b7842d44bad20.patch"
-      sha256 "378a30878cc25e3b1580e24f832ed171c4fb67786dd46a5d2653863c1b01ea3f"
-    end
-
-    # stack.cabal: adjust dependency bounds
-    # Allow generic-deriving < 1.11 and monad-unlift < 0.3
-    # https://github.com/fpco/stackage/issues/1467
-    patch do
-      url "https://github.com/commercialhaskell/stack/commit/3a7122e61b0263f44b35761ff093d134f9166899.patch"
-      sha256 "e188b457883da73888119772f308828f8f34090f294afabf9d49f7072c365ab3"
-    end
-
-    # Have `stack ghci` skip build if there are no targets
-    # https://github.com/commercialhaskell/stack/issues/2176
-    patch do
-      url "https://github.com/commercialhaskell/stack/commit/0b58282cd97be66b22019832eb5906c29b4c8971.patch"
-      sha256 "6a2f52e6b389ad3e17892ea584bef46e07ad9a0c7598cd857832f650eb6464ac"
-    end
-
-    # Remove debug info from default output
-    # https://github.com/commercialhaskell/stack/issues/2175
-    patch do
-      url "https://github.com/commercialhaskell/stack/commit/0060443d53673ee035e1a755b7082b892a47cb56.patch"
-      sha256 "6e91794e3cdebb3cafa74f58a4338c867cb9cd981eb500bdf8e545ba23725d3e"
-    end
-
-    # Bump GHC-included package upper bounds
-    # https://github.com/fpco/stackage/issues/1476
-    patch do
-      url "https://github.com/commercialhaskell/stack/commit/f848ceb16b147636e112068d346267b1de661c1b.patch"
-      sha256 "d7229380ec9c4a6de8abaeaf58afab71dfb434e5c65ed20cde1fb35c2dd4f306"
-    end
-
-    # Restrict persistent-sqlite < 2.5.0.1
-    # https://github.com/commercialhaskell/stack/issues/2182
-    patch do
-      url "https://github.com/commercialhaskell/stack/commit/b388d11e2013c3e755573c44e40788461a26116f.patch"
-      sha256 "dfcf888ff715a4843fbd4730912b640445e28a4440445d6198112acdb89f56ed"
-    end
-  end
 
   bottle do
     sha256 "4ee81138d47a78fb6cd416491d1c6f1f9c610473870dcba8617c6625d82b7d20" => :el_capitan
@@ -63,14 +17,29 @@ class HaskellStack < Formula
     sha256 "0cb19aded681b0b4a002236f0e0ed992e229c1a5598adf2748a5272328dd5b41" => :mavericks
   end
 
+  option "without-bootstrap", "Don't bootstrap a stage 2 stack"
+
   depends_on "ghc" => :build
   depends_on "cabal-install" => :build
 
   def install
-    install_cabal_package
+    if build.with? "bootstrap"
+      cabal_sandbox do
+        cabal_install
+        # Let `stack` handle its own parallelization
+        # Prevents "install: mkdir ... ghc-7.10.3/lib: File exists"
+        jobs = ENV.make_jobs
+        ENV.deparallelize do
+          system "stack", "-j#{jobs}", "setup"
+          system "stack", "-j#{jobs}", "--local-bin-path=#{bin}", "install"
+        end
+      end
+    else
+      install_cabal_package
+    end
   end
 
   test do
-    system "#{bin}/stack", "new", "test"
+    system bin/"stack", "new", "test"
   end
 end


ilovezfs:
Also, see https://github.com/commercialhaskell/stack/issues/2182 and https://github.com/commercialhaskell/stack/issues/2192 for the prior conversations.

ilovezfs:
CC @zmwangx

zmwangx:
system "stack", "--local-bin-path=#{bin}", "install"

works only because `stack` installs a single fully-linked executable and no more. I wonder if this is future-proof.

I'm 👍  on this If it provably solves a problem though.

borsboom:
We (the Stack release team) very much intend to keep `stack` as a single executable as ease-of-distribution is a major goal.  Our new version release process also includes testing and sending a PR to homebrew-core, so we'll definitely notice if it stops working.

I'm building this PR now, so I'll let you know whether it solves the original problem.

ilovezfs:
@borsboom If you don't mind, please also check the current bottle that's distributed in master as well since it differs from the one the user reporting the original issue was using, so we don't actually know at this point whether the current bottle is or isn't broken in the relevant sense.

borsboom:
Ok, I just confirmed the existing bottle against the new build, and I can confirm that the existing bottle fails and the new build succeeds:

```
$ ~/fpco/tmp/stack-1.1.2-homebrew-old --docker exec echo hello
SQLite3 returned ErrorCan'tOpen while attempting to perform open "/Users/manny/.stack/docker.db".

$ ~/fpco/tmp/stack-1.1.2-homebrew-new --docker exec echo hello
hello
```

So this solves the original problem and I expect it will prevent other problems in the future.  Incidentally, other Haskell-built programs may also benefit from being built by Stack rather than cabal-install if their source code includes a `stack.yaml`, for the same reason: stack is all about reproducible builds.

ilovezfs:
@borsboom That's with https://github.com/Homebrew/homebrew-core/blob/master/Formula/haskell-stack.rb#L15 ?

borsboom:
Yes, my local Formula/haskell-stack.rb has the same SHA256 for the bottle and I just installed that bottle before running my test.

ilovezfs:
@borsboom https://github.com/Homebrew/legacy-homebrew/pull/49158 for the prior conversation on whether using stack more generally might be a viable option for Homebrew.

ilovezfs:
@borsboom In a local test, not the CI for the PR which as you can see passed, I just got this:
```
install: mkdir /private/tmp/haskell-stack-20160527-86482-aj8hzg/stack-1.1.2/stack_root/programs/x86_64-osx/ghc-7.10.3/lib: File exists
make[1]: *** [install_libffi_headers] Error 71
make: *** [install] Error 2

/usr/local/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: stack setup
```
Do I need to deparallelize stack setup in some way? Is that a known bug?

borsboom:
`stack setup` doesn't currently expect to be run in parallel in the same STACK_ROOT.  So either de-parallelizing, or making sure concurrent runs go in different STACK_ROOTs should fix that.

ilovezfs:
@borsboom does stack install have the same issue?

borsboom:
It potentially can if parallel installs of the same package go into the same shared sandbox (in the STACK_ROOT).

ilovezfs:
@borsboom PR refreshed. I think that will fix it.

ilovezfs:
@DomT4 I wonder if it would be wise to specify a folder within HOMEBREW_CACHE for the stack downloads to accumulate, on analogy to the java cache thingy.

Formula/haskell-stack.rb
```diff
@@ -26,7 +27,16 @@ def install
     # Reported 25 May 2016: https://github.com/commercialhaskell/stack/issues/2192
     (buildpath/"cabal.config").write("allow-newer: base,transformers\n")
 
-    install_cabal_package
+    cabal_sandbox do
+      cabal_install
+
+      # Let `stack` handle its own parallelization
```
MikeMcQuaid: What happens without this? Seems weird.

ilovezfs: https://github.com/Homebrew/homebrew-core/pull/1480#issuecomment-222192300

MikeMcQuaid: Does it work to just have it for `stack setup` and leave `stack install` as normal? No worries if not, just seems odd.

ilovezfs: It's possible that the only problem was that I was explicitly setting STACK_ROOT. I should try with just passing `-j#{jobs}` but not wrapped in the ENV.deparallelize.

ilovezfs: Yeah, we may not need to pass `-j#{jobs}` at all either, but it seems that's probably the "right" thing to do, as it's the same thing we do with cabal: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/language/haskell.rb#L58

ilovezfs: Note that the breakage occurred *without* setting `-j` at all, but *with* setting `STACK_ROOT`.

ilovezfs:
Without the wrapping in ENV.deparallelize, the error showed up again during stack setup, this time on CI.
```
install: mkdir /private/tmp/haskell-stack-20160527-31238-lzl7mu/stack-1.1.2/.stack/programs/x86_64-osx/ghc-7.10.3/lib: File exists
make[1]: *** [install_includes] Error 71
make: *** [install] Error 2
```
So I'll put that back in.

ilovezfs:
On the one hand, I'm pleased that this bootstrapping technique works without much fuss. On the other hand, I kind of wish we knew what was causing `SQLite3 returned ErrorCan'tOpen while attempting to perform open "/Users/manny/.stack/docker.db".` and could address that directly.

DomT4:
It's mentioned in https://github.com/commercialhaskell/stack/blob/master/doc/docker_integration.md#configuration, if helpful at all.

ilovezfs:
I'm also thinking this formula should be renamed `stack` because I cannot seem to stop myself from typing `brew install stack`, which seems to be impervious to `:ilovezfs-self-fix`.

borsboom:
> On the one hand, I'm pleased that this bootstrapping technique works without much fuss. On the other hand, I kind of wish we knew what was causing SQLite3 returned ErrorCan'tOpen while attempting to perform open "/Users/manny/.stack/docker.db". and could address that directly.

I actually think of this bootstrapping technique fixing a general problem (that there's no guarantee that the Homebrew version of `stack` has been built the same way as other `stack` binaries), and it's just a side-effect that it fixes the SQLite error.  As for the original cause of the SQLite error: IMO that's for the Stack team to determine, and has nothing to do with Homebrew.  Bugs get introduced _all the time_ by changing Haskell dependency versions and this is just the latest example, which is why we're kind of anal about wanting consistent, reproducible builds.

ilovezfs:
Sure, but this bug precedes the monkeying with `(buildpath/"cabal.config").write("allow-newer: base,transformers\n")`, so it implies there's a bug with https://github.com/commercialhaskell/stack/blob/master/stack.cabal which must be permitting solver to find an invalid solution.

borsboom:
Yup, that's why I say it's a problem for the Stack team to solve, not for Homebrew.  If Homebrew builds Stack consistently to the official builds, it should never have a bug that the official builds don't have.

borsboom:
Incidentally, to help narrow this down: is it possible to get access to the logs from the Homebrew builds to see which version sqlite-persistent it used?

ilovezfs:
It's not executing cabal install with verbosity, so that info appears not to make it to any logs, but doing a local build, it appears to be persistent-sqlite-2.5.0.1

ilovezfs:
@borsboom Here's the full log for that: https://gist.github.com/ilovezfs/9411d2976e2617ef2b9f89a98edbbc6f

borsboom:
FYI I've narrowed it down to being introduced with persistent-sqlite-2.5.0.1.  See https://github.com/yesodweb/persistent/pull/561#issuecomment-222329087, so we could fix this particular error by adding a `persistent-sqlite < 2.5.0.1` bound (and I will be adding one to stack.cabal).  That said, I still think this bootstrapping is the right approach to building Stack on Homebrew to prevent such issues in the future.

ilovezfs:
@borsboom Thank you for tracking that down. It helps to make this change not feel like a shotgun wedding hehe.

I wonder if bootstrapping should be a (possibly default) option?

ilovezfs:
https://github.com/Homebrew/homebrew-core/pull/1525 brings in the upstream fix for the sqlite problem and replaces our GHC 8 compatibility hack with the corresponding upstream fixes. This should allow us to consider the merits of bootstrapping on its own terms without having it tied to either of those specific issues.

@zmwangx Therefore, this commit no longer "provably solves a problem."

zmwangx:
I would be hesitant because some other packages could be installed through stack too, and it just goes back to our original discussion about stack and Stackage on the legacy-homebrew tracker (don't bother to pull up the issue/PR number now). The conclusion there was wait and see.

The situation has hardly changed since then (right?) — packages break now and then, all quite fixable 😉 

I'm not an expert. 😉

ilovezfs:
@zmwangx yup just wondering if you were still :+1: on this idea or not, given that "the facts on the ground" have now changed.

>The situation has hardly changed since then (right?) — packages break now and then, all quite fixable

Yup, agreed.

> would be hesitant because some other packages could be installed through stack too

Homebrew's sandbox (i.e., OS X sandboxing) will prevent that in the case of bottling/CI, but the `HOMEBREW_SANDBOX=1` / `--sandbox` option is not on by default when users do local `--build-from-source` installs (although we intend to turn that on at some point in the future), so that is definitely a real concern.

I'm intrigued by the possibility of having Homebrew use stack more generally not just the curated cabal files where needed, but I believe your main concern was that there isn't any clear mapping between Homebrew's Cellar target installation prefixes for a given piece of Haskell software and any notion of "installation" Stack could perform. (I'm also really curious whether we'll be able to make use of http://blog.ezyang.com/2016/05/announcing-cabal-new-build-nix-style-local-builds/)

In this particular case, since it's only one file, there's a clear mapping between "local-bin-path" and Homebrew's #{prefix}/bin, but I definitely now see what you mean about specifically STACK_ROOT not being the same thing as $(brew --prefix) or a Formula["foo"].prefix.

@borsboom when software includes a stack yaml file, and it needs to install include header files, dylibs, multiple binaries, man pages, etc., is there a `stack` command line option that corresponds to e.g., `./configure --prefix=#{prefix} ; make ; make install`.

zmwangx:
> yup just wondering if you were still 👍 on this bootstrapping idea or not

I've always been 👍 on the idea. Not only bootstrapping `haskell-stack`, but also use it for more packages. (Was it me or you that brought it up the stack/Stackage discussion in the first place? Probably you but not too sure. Anyway, IIRC I was definitely one of the first to jump on board.) I just don't feel too strongly about it.

> when software includes a stack yaml file, and it needs to install include header files, dylibs, multiple binaries, man pages, etc., is there a stack command line option that corresponds to e.g., ./configure --prefix=#{prefix} ; make ; make install.

I thought the answer is no and that was an explicit wontfix?

ilovezfs:
@zmwangx I'd originally brought up the idea of using the stack tool itself in some way, and you'd jumped in with
>Talking about stack, I once thought about integrating stack into language/haskell.rb, but stack install lacks a --prefix option which makes it pretty much useless (see enhancement commercialhaskell/stack#848). @ilovezfs Do you have another idea to make it work?

(https://github.com/Homebrew/legacy-homebrew/pull/49158#issuecomment-184932302)

which I still think is exactly on point in terms of how much use we can productively make of the tool itself (as opposed to just the curated cabal files).

zmwangx:
Yeah, and I think lack of a `--prefix` option is still the problem.

ilovezfs:
Well I'm glad we have @borsboom in the conversation too now since https://github.com/commercialhaskell/stack/issues/848 was opened by him!

@zmwangx Have you looked at http://blog.ezyang.com/2016/05/announcing-cabal-new-build-nix-style-local-builds/ yet? Looks exciting if it actually works …

zmwangx:
> ...cabal-new-build...

Looks interesting. Too tired today, bookmarked for tomorrow 😃

ilovezfs:
>Too tired today, bookmarked for tomorrow 

Nighty night!

zmwangx:
By the way, I was aware of https://github.com/commercialhaskell/stack/issues/848 and had it subscribed. Not sure how my memory got screwed up and thought it was a wontfix.

zmwangx:
And I was even responsible for the single upvote on that issue!

ilovezfs:
@zmwangx See, this is totally bogus:
>I'm not an expert. 😉 

:)

ilovezfs:
I have refreshed the PR rebased on https://github.com/Homebrew/homebrew-core/commit/55ac991a66bed86410e2311ab701add9ed10dc9c and made not bootstrapping available as a non-default option.

ilovezfs:
@borsboom The new bottles pushed out this morning should no longer have the docker bug, if you'd like to check.

borsboom:
Hmm, I tried reinstalling the bottle but still getting the original error.  This is in my /usr/local/Library/Taps/homebrew/homebrew-core/Formula/haskell-stack.rb:

```
  bottle do
    sha256 "891fa9dc2ac357980b6140d5771d60b5056d7845ad03b580b1206a5234cb8e76" => :el_capitan
    sha256 "e1c3e21c829930cad1870bada9a90969bf32ee866bb6379c1bc8004759cb0b54" => :yosemite
    sha256 "a3ab294bf25d6424ecdc70b7eee2f2928674e51098a73c4eef6ae746fb6f5eb4" => :mavericks
  end
```

I do still have misgivings about recommending a package built using not only different dependency versions but even a different GHC version (which means we can't fix dependency versions with a stack.cabal), so I'd love to see this get merged.  Until then, I'll retain a warning recommending against using Homebrew to install Stack on our download/install instructions.

ilovezfs:
I have to say that I'm inclined to boneyard this formula if upstream is going to be warning about using our version.

borsboom:
If that happens, we could bring it back once official Stack builds move to GHC 8.0.1 and we can use a cabal.config to ensure that Homebrew uses the right dependency versions.  That'll probably happen once the next Stackage LTS major version is released in a few months.

ilovezfs:
I think stack may be a better fit for Homebrew Cask, if upstream needs that level of control over its binaries.

ilovezfs:
This is weird
```
iMac-TMP:Formula joe$ ls -l /usr/local/Cellar/haskell-stack/1.1.2_1/bin/stack 
-r-xr-xr-x  1 joe  admin  62315676 May 20 15:10 /usr/local/Cellar/haskell-stack/1.1.2_1/bin/stack
```

borsboom:
Exactly the same `ls -l` output for me.

ilovezfs:
I'm going to rebuild a new revision just to be sure this isn't some silly CI thing.

ilovezfs:
@borsboom So it looks like the files get dated based on the upstream tarball release date. So that's probably a red herring. Nonetheless, I have bumped the revision, rebuilt the bottles, and pushed: https://github.com/Homebrew/homebrew-core/commit/026dd68589efc3fd452961ad0833a7739cd12c49

Please let me know if 1.1.2_2 still exhibits the problematic behavior. If so, then I believe the problem is not persistent sqlite 2.5.0.1.

borsboom:
Seems to be fixed.

ilovezfs:
@borsboom How peculiar … but glad to hear it.

ilovezfs:
@borsboom If that's the case, I'd appreciate the removal of the warnings regarding the Homebrew version.

borsboom:
@ilovezfs: Warning removed.

ilovezfs:
@borsboom Thank you. Also, I made some more changes to this PR. Is that the proper use of sdist in this context?

borsboom:
Not sure if I'm reading it correctly, but it looks like you're only using the sdist for the `stack install` part.  The sdist is actually more important for the `cabal install` part because it has full pvp-bounds in `stack.cabal` (whereas the plain source code from Github does not), but there's no harm using it for the `stack install` part as well.

ilovezfs:
@boorsboom OK, that makes sense. I did that because the patches here (and in the future) are much more likely to apply cleanly to the GitHub tag. Would it be correct to simply replace the cabal.config with the one from sdist?

ilovezfs:
I suppose just using sdist's for the stage 1 build will be problematic, though, because it will be missing https://github.com/commercialhaskell/stack/commit/f848ceb16b147636e112068d346267b1de661c1b.patch

ilovezfs:
@borsboom OK, refreshed the PR using sdist's stack.cabal. How's that looking?

borsboom:
I guess you're just copying the `stack.cabal`?  I'd probably recommend just using the whole sdist verbatim for the sake of simplicity, but either way works so I don't feel really strongly about it.

ilovezfs:
@borsboom That's probably a good idea. My concern is just that if an important GitHub commit "foo.patch" is needed in the future between releases that someone might get confused when the commits refuse to apply cleanly, but I suppose we can cross that bridge if/when we get there.

ilovezfs:
@borsboom OK, updated to use sdist only. Note that the version needs to be specified manually since brew's version detector thingy is confused by the sdist url.

borsboom:
LGTM!

ilovezfs:
@borsboom Cool. BTW, I suspect this will also work out-of-the-box on Linuxbrew.

ilovezfs:
@borsboom Is there an environment variable for specifying a cache location for the downloads? I think we might be able to accomplish something similar to the `ENV.java_cache` feature we have.

borsboom:
Those are all kept in the STACK_ROOT.

DomT4:
@ilovezfs Is `haskell-stack` used for other Homebrew formulae? If it isn't we don't "care" about how it behaves outside Homebrew's usage.

ilovezfs:
@DomT4 The current version here is downloading a ton of external resources every run... including the compiler, and then throwing it all out.

ilovezfs:
@DomT4 Hence https://github.com/Homebrew/homebrew-core/pull/1480#issuecomment-222224562

DomT4:
@ilovezfs I see. Sorry, didn't see the original comment, this thread got kind of busy 😄. If usage is limited to this formula I don't have an objection to an `ENV` variable limited to this formula specifically.

ilovezfs:
@DomT4 Longer-term I think we're going to want to switch from the current cabal sandbox approach to some combination of haskell stack (the current blocker to that is https://github.com/commercialhaskell/stack/issues/848) and cabal new-build, but at the moment the concern is only this formula and users with slow Internet connections and/or bandwidth caps, so I'm not imaging anything for superenv at this point, just a one-off analogue for this formula alone.

However, I'm not sure if there's any way to "reset" a STACK_ROOT but preserve the downloads to avoid "contamination" between builds.

borsboom:
You could probably delete everything except `$STACK_ROOT/programs` if you just want to keep GHC.  Then you lose the package caching, though.

DomT4:
@ilovezfs Cool. You're more than welcome to introduce a tweak for the single formula here as long as we avoid introducing DSL where the application is narrow, is basically what I was getting at vaguely. I'll trust you on longer-term Cabal changes, I've done my utmost to avoid getting involved in Haskell stuff heh.

ilovezfs:
> I've done my utmost to avoid getting involved in Haskell stuff heh.

:laughing: :laughing: 

>introducing DSL where the application is narrow

Wouldn't dream of it :)

MikeMcQuaid:
PR seems good to me 👍

ilovezfs:
Setting `ENV["STACK_ROOT"] = "#{HOMEBREW_CACHE}/stack_root"` at the top, once it's installed and the dust has settled:
```
iMac-TMP:~ joe$ du -sh ~/Library/Caches/Homebrew/stack_root/
1.8G	/Users/joe/Library/Caches/Homebrew/stack_root/
```
Worth exploring further perhaps, but I don't think that needs to block this PR.

ilovezfs:
@borsboom It has :ship:'ed if you'd like to verify it's :100: .

PR #1498, opened 2016-05-28
n98-magerun2 1.1.12 (new formula)

mgeoffray:
Add formula for Magento 2 tool n98-magerun2


UniqMartin:
Check out [the build error](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/2465/version=mavericks/testReport/junit/brew-test-bot/mavericks/test_n98_magerun2/). It seems that the PHP that ships with Mavericks is too old to be compatible with `n98-magerun2`. (We could make the formula require Yosemite or above, but that would be unfortunate for people that have installed an up-to-date PHP from a third party or the [`homebrew/php` tap](https://github.com/Homebrew/homebrew-php).)

UniqMartin:
On second thought (and after asking someone from the `homebrew/php` tap) this is definitely something that should live in the `homebrew/php` tap instead of here. Thanks for your submission, but please open a new pull request with this formula at https://github.com/Homebrew/homebrew-php.

**Maintainers:** I was probably too quick to accept the related `n98-magerun` (in PR #1417). Sorry about that! I'll follow up with a migration of `n98-magerun` to the `homebrew/php` tap once this formula has been submitted and accepted there. (Please speak up if this is not desirable.)

mgeoffray:
OK not problem at all I will reopen these pull requests on PHP tap ! Thank you for you replies

PR #1507, opened 2016-05-28
moreutils 0.59

nijikon:

---

### DIFF
--- a/Formula/moreutils.rb
+++ b/Formula/moreutils.rb
@@ -1,9 +1,9 @@
 class Moreutils < Formula
   desc "Collection of tools that nobody wrote when UNIX was young"
   homepage "https://joeyh.name/code/moreutils/"
-  url "https://mirrors.ocf.berkeley.edu/debian/pool/main/m/moreutils/moreutils_0.58.orig.tar.gz"
-  mirror "https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/m/moreutils/moreutils_0.58.orig.tar.gz"
-  sha256 "9c4bba745518f1cc4d5594b1276793bd4f737fedb2e5b14beeab4b73f106c00a"
+  url "https://mirrors.ocf.berkeley.edu/debian/pool/main/m/moreutils/moreutils_0.59.orig.tar.gz"
+  mirror "https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/m/moreutils/moreutils_0.59.orig.tar.gz"
+  sha256 "a48e11c3197bf79a7bfaa121423e64626e3381d9fedc91d606e9724ae498d1b4"
 
   head "git://git.kitenet.net/moreutils"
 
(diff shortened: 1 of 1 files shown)


Formula/moreutils.rb
```diff
@@ -22,15 +22,13 @@ class Moreutils < Formula
   conflicts_with "task-spooler", :because => "Both install a 'ts' executable."
 
   resource "Time::Duration" do
-    url "https://cpan.metacpan.org/authors/id/A/AV/AVIF/Time-Duration-1.1.tar.gz"
```
DomT4: Please don't remove the `https://` URLs here, they still work with the update:

```
https://cpan.metacpan.org/authors/id/N/NE/NEILB/Time-Duration-1.20.tar.gz
```

Formula/moreutils.rb
```diff
@@ -22,15 +22,13 @@ class Moreutils < Formula
...
-    mirror "http://search.cpan.org/CPAN/authors/id/A/AV/AVIF/Time-Duration-1.1.tar.gz"
-    sha256 "a69c419c4892f21eba10002e2ab8c55b657b6691cf6873544ef99ef5fd188f4e"
+    url "http://search.cpan.org/CPAN/authors/id/N/NE/NEILB/Time-Duration-1.20.tar.gz"
+    sha256 "458205b528818e741757b2854afac5f9af257f983000aae0c0b1d04b5a9cbbb8"
   end
 
   resource "IPC::Run" do
-    url "https://cpan.metacpan.org/authors/id/T/TO/TODDR/IPC-Run-0.92.tar.gz"
```
DomT4: Same notation as above.

nijikon:
@DomT4 sure, sorry for that. I've tweaked audit to handle this better.

DomT4:
What was the audit saying/doing prior to the tweak? Seems okay here now, although it's nice to avoid `ftp` where possible because firewalls can be prone to closing that port.

nijikon:
Nothing. That's why I tweaked it. Maybe we could allow http in mirrors and add an audit that we prefer http over ftp in strict mode?

DomT4:
Sorry, you've lost me a bit here, which albeit is quite easy at 5am 😄. If `https://` is available it should be used. Ideally mirrors should be using a different website because if one protocol is down it's not regularly an unsafe bet others will be dead as well.

Was something amiss with the usual `http://search.cpan.org/CPAN/authors` mirrors we use for Perl stuff?

nijikon:
No worries 😄  I was just wondering if we could tweak audit a bit for this. So for ```url``` we would allow https or http only no ftp. For this particular case we can make it ```http://search.cpan.org/CPAN/authors``` but than I would have to change this https://github.com/Homebrew/brew/commit/7bbcf9cacc18aabf78a21ec5cc5892d440631e08 so it would only work on url not on mirror. What do you think?

DomT4:
https://github.com/Homebrew/brew/pull/302

DomT4:
👍 :shipit:.

PR #1518, opened 2016-05-28
racer 1.1.0 (new formula)

CamJN:

---

I know I bend some of the unwritten rules with this formula but come on how awesome would this setup be? You can just bump the revision and add a line to the map each time there's a new rust.

And I feel I should point out that the rust folks have [explicitly stated](https://github.com/rust-lang/cargo/issues/2082) that `cargo install` is not meant to be a package manager and therefore they won't be supporting a bunch of necessary functionality that one can get from homebrew, which is why I am submitting this formula here.

### DIFF
--- a/Formula/racer.rb
+++ b/Formula/racer.rb
@@ -9,8 +9,8 @@ class Racer < Formula
   depends_on "rust" => :build
 
   resource "rust_source" do
-    url Formula['rust'].stable.url
-    sha256 Formula['rust'].stable.checksum.hexdigest
+    url Formula["rust"].stable.url
+    sha256 Formula["rust"].stable.checksum.hexdigest
   end
 
   def install
@@ -19,12 +19,12 @@ class Racer < Formula
     (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
     resource("rust_source").stage do
       rm_rf ["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"]
-      (share/"rust_src/#{Formula['rust'].stable.version}").install Dir["src/*"]
+      (share/"rust_src/#{Formula["rust"].stable.version}").install Dir["src/*"]
     end
   end
 
   def post_install
-    ln_sf (share/"rust_src/#{Formula['rust'].stable.version}"), (share/"rust_src/current")
+    ln_sf (share/"rust_src/#{Formula["rust"].stable.version}"), (share/"rust_src/current")
   end
 
   test do
--- a/Formula/racer.rb
+++ b/Formula/racer.rb
@@ -6,7 +6,7 @@ class Racer < Formula
   sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
   head "https://github.com/phildawes/racer.git"
 
-  depends_on :rust => ["1.9.0", :build]
+  depends_on :rust => :build # rust MUST be up to date or else racer will produce invalid completions
 
   resource "rust_source" do
     url "https://static.rust-lang.org/dist/rustc-1.9.0-src.tar.gz"
@@ -24,11 +24,11 @@ class Racer < Formula
   end
 
   def post_install
-    ln_s (share/"rust_src/1.9.0"), (share/"rust_src/current"), :force => true
+    ln_sf (share/"rust_src/1.9.0"), (share/"rust_src/current")
   end
 
   test do
-    ENV["RUST_SRC_PATH"] = (share/"rust_src/src")
+    ENV["RUST_SRC_PATH"] = share/"rust_src/src"
     system "#{bin}/racer", "complete", "std::io::B"
   end
 end
--- a/Formula/racer.rb
+++ b/Formula/racer.rb
@@ -1,0 +1,34 @@
+class Racer < Formula
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on :rust => ["1.9.0", :build]
+
+  resource "rust_source" do
+    url "https://static.rust-lang.org/dist/rustc-1.9.0-src.tar.gz"
+    sha256 "b19b21193d7d36039debeaaa1f61cbf98787e0ce94bd85c5cbe2a59462d7cfcd"
+  end
+
+  def install
+    system "cargo", "build", "--release"
+    (libexec/"bin").install "./target/release/racer"
+    bin.write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
+    resource("rust_source").stage do
+      rm_rf(["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"])
+      (share/"rust_src/1.9.0").install Dir["./src/*"]
+    end
+  end
+
+  def post_install
+    ln_s (share/"rust_src/1.9.0"), (share/"rust_src/current"), :force => true
+  end
+
+  test do
+    ENV["RUST_SRC_PATH"] = (share/"rust_src/src")
+    system "#{bin}/racer", "complete", "std::io::B"
+  end
+end
--- a/Formula/racer.rb
+++ b/Formula/racer.rb
@@ -9,8 +9,8 @@ class Racer < Formula
   depends_on "rust" => :build
 
   resource "rust_source" do
-    url "https://static.rust-lang.org/dist/rustc-1.9.0-src.tar.gz"
-    sha256 "b19b21193d7d36039debeaaa1f61cbf98787e0ce94bd85c5cbe2a59462d7cfcd"
+    url Formula['rust'].stable.url
+    sha256 Formula['rust'].stable.checksum.hexdigest
   end
 
   def install
@@ -19,12 +19,12 @@ class Racer < Formula
     (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
     resource("rust_source").stage do
       rm_rf ["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"]
-      (share/"rust_src/1.9.0").install Dir["src/*"]
+      (share/"rust_src/#{Formula['rust'].stable.version}").install Dir["src/*"]
     end
   end
 
   def post_install
-    ln_sf (share/"rust_src/1.9.0"), (share/"rust_src/current")
+    ln_sf (share/"rust_src/#{Formula['rust'].stable.version}"), (share/"rust_src/current")
   end
 
   test do
--- a/Formula/racer.rb
+++ b/Formula/racer.rb
@@ -6,7 +6,7 @@ class Racer < Formula
   sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
   head "https://github.com/phildawes/racer.git"
 
-  depends_on :rust => :build # rust MUST be up to date or else racer will produce invalid completions
+  depends_on "rust" => :build # rust MUST be up to date or else racer will produce invalid completions
 
   resource "rust_source" do
     url "https://static.rust-lang.org/dist/rustc-1.9.0-src.tar.gz"
--- a/Formula/racer.rb
+++ b/Formula/racer.rb
@@ -1,7 +1,6 @@
 class Racer < Formula
   desc "Rust Code Completion utility"
   homepage "https://github.com/phildawes/racer"
-
   url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
   sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
   head "https://github.com/phildawes/racer.git"
@@ -15,19 +14,20 @@ class Racer < Formula
 
   def install
     system "cargo", "build", "--release"
-    (libexec/"bin").install "./target/release/racer"
-    (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{share}/rust_src/current")
+    (libexec/"bin").install "target/release/racer"
+    (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src/current")
+
     resource("rust_source").stage do
-      rm_rf ["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"]
-      (share/"rust_src/#{Formula["rust"].stable.version}").install Dir["src/*"]
+      rm_rf Dir["src/{llvm,test,librustdoc,etc/snapshot.pyc}"]
+      (pkgshare/"rust_src/#{Formula["rust"].stable.version}").install Dir["src/*"]
     end
   end
 
   def post_install
-    (share/"rust_src").install_symlink Formula["rust"].stable.version.to_s => "current"
+    (pkgshare/"rust_src").install_symlink Formula["rust"].stable.version.to_s => "current"
   end
 
   test do
-    system "#{bin}/racer", "complete", "std::io::B"
+    system bin/"racer", "complete", "std::io::B"
   end
 end
--- a/Formula/racer.rb
+++ b/Formula/racer.rb
@@ -16,7 +16,7 @@ class Racer < Formula
   def install
     system "cargo", "build", "--release"
     (libexec/"bin").install "./target/release/racer"
-    (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
+    (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{share}/rust_src/current")
     resource("rust_source").stage do
       rm_rf ["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"]
       (share/"rust_src/#{Formula["rust"].stable.version}").install Dir["src/*"]
@@ -24,7 +24,7 @@ class Racer < Formula
   end
 
   def post_install
-    ln_sf (share/"rust_src/#{Formula["rust"].stable.version}"), (share/"rust_src/current")
+    (share/"rust_src").install_symlink Formula["rust"].stable.version.to_s => "current"
   end
 
   test do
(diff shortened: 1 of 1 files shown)


Formula/racer.rb
```diff
@@ -0,0 +1,48 @@
...
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on "rust" => :build
+
+  resource "rust_source" do
+    url "https://static.rust-lang.org/dist/rustc-#{Formula["rust"].installed_version.to_str}-src.tar.gz"
```
MikeMcQuaid: This only needs to support the newest version of `rust`

Formula/racer.rb
```diff
@@ -0,0 +1,48 @@
...
+      (share/"rust_src/#{Formula["rust"].installed_version.to_str}").install Dir["./src/*"]
+    end
+  end
+
+  def post_install
+    ln_s (share/"rust_src/#{Formula["rust"].installed_version.to_str}"), (share/"rust_src/current"), :force => true
+  end
+
+  def caveats
+    "You must set RUST_SRC_PATH=#{HOMEBREW_PREFIX}/share/rust_src/current in your ENV/editor."
```
MikeMcQuaid: Instead it'd be nicer to write a wrapper script to always set this correctly.

frewsxcv:
What's wrong with doing `cargo install racer`?

CamJN:
@frewsxcv there's no way to update it, or be notified of updates, and you have to add yet another dir to your path.

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+    (libexec/"bin").install "./target/release/racer"
+    bin.write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
+    resource("rust_source").stage do
+      rm_rf(["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"])
+      (share/"rust_src/1.9.0").install Dir["./src/*"]
+    end
+  end
+
+  def post_install
+    ln_s (share/"rust_src/1.9.0"), (share/"rust_src/current"), :force => true
```
ilovezfs: probably want
```
ln_sf share/"rust_src/1.9.0", share/"rust_src/current"
```

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  test do
+    ENV["RUST_SRC_PATH"] = (share/"rust_src/src")
```
ilovezfs: don't need the ()

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  resource "rust_source" do
+    url "https://static.rust-lang.org/dist/rustc-1.9.0-src.tar.gz"
+    sha256 "b19b21193d7d36039debeaaa1f61cbf98787e0ce94bd85c5cbe2a59462d7cfcd"
+  end
+
+  def install
+    system "cargo", "build", "--release"
+    (libexec/"bin").install "./target/release/racer"
+    bin.write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
```
ilovezfs: Upstream's readme seems to suggest it can just be installed in bin. This seems awfully complex.

MikeMcQuaid: @ilovezfs does `RUST_SRC_PATH` not need to be set manually in that case?

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on :rust => ["1.9.0", :build]
+
+  resource "rust_source" do
```
MikeMcQuaid: Should this maybe be installed as part of the Rust formula?

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
+class Racer < Formula
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on :rust => ["1.9.0", :build]
```
MikeMcQuaid: Just do `depends_on "rust"`

Formula/racer.rb
CamJN: RUST_SRC_PATH does need to be set if you do that, per the docs. https://github.com/phildawes/racer#configuration

Formula/racer.rb
CamJN: There's been suggestions both ways on this point, I was told though that keeping the source around for a package wasn't allowed. So I did it here. Keeping it in rust is IMO a cleaner way to do this if it's allowed.

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
+class Racer < Formula
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on :rust => :build # rust MUST be up to date or else racer will produce invalid completions
```
DomT4: `:rust` isn't a valid special dependency.

CamJN:
Hmm the link to the last build is broken, but I recall not understanding what went wrong. Could someone take a look? And if the build is gone would you prefer I push an empty commit or someone just ask @BrewTestBot to help?

DomT4:
@BrewTestBot test this please

CamJN:
Oops, I think I got it figured out.

CamJN:
Yup, I know what I did, and it's fixed now. Works locally hopefully on CI as well.

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on "rust" => :build # rust MUST be up to date or else racer will produce invalid completions
+
+  resource "rust_source" do
+    url "https://static.rust-lang.org/dist/rustc-1.9.0-src.tar.gz"
```
MikeMcQuaid: Does this need to be updated every time we update Rust? Need to figure out how best to handle that.

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
+class Racer < Formula
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on "rust" => :build # rust MUST be up to date or else racer will produce invalid completions
```
MikeMcQuaid: Don't need this comment, it will always be up to date.

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+    url "https://static.rust-lang.org/dist/rustc-1.9.0-src.tar.gz"
+    sha256 "b19b21193d7d36039debeaaa1f61cbf98787e0ce94bd85c5cbe2a59462d7cfcd"
+  end
+
+  def install
+    system "cargo", "build", "--release"
+    (libexec/"bin").install "./target/release/racer"
+    (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
+    resource("rust_source").stage do
+      rm_rf(["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"])
```
MikeMcQuaid: Don't need the `()` or perhaps even `[]`.

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+      (share/"rust_src/1.9.0").install Dir["./src/*"]
```
MikeMcQuaid: Think you can just do `src/*`

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+  end
+
+  def post_install
+    ln_sf (share/"rust_src/1.9.0"), (share/"rust_src/current")
```
MikeMcQuaid: Can this not be done in `install`?

Formula/racer.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  test do
+    ENV["RUST_SRC_PATH"] = share/"rust_src/src"
```
MikeMcQuaid: Can we write a wrapper script so users don't need to set this?

Formula/racer.rb
CamJN: Yes it does, and it would work WAAAAY better if it was done on the rust formula, but I haven't gotten a go ahead for doing it that way.

Formula/racer.rb
CamJN: Not really, can't trust users to stay up to date, but I could move it to a caveats section.

Formula/racer.rb
CamJN: Oops, already did, forgot to remove redundant step in test.

Formula/racer.rb
MikeMcQuaid: We can because depending on something means it's always upgraded before installation.

Formula/racer.rb
MikeMcQuaid: How big is the extracted source?

CamJN: 55 MB, and it could totally be an optional thing for the rust formula, it's just that by pulling it down there it's automatically the correct version for the rust that's installed. Whether the user has rust up to date or not.

MikeMcQuaid: I think we could probably do something hacky where we make use of the existing Rust formula's tarball instead until we've got at least 2-3 formulae that need the source.

CamJN: I don't necc. follow, what if they installed from a bottle? Then the tarball wouldn't contain the sources. I tried an approach a while ago that forced a fetch of the rust source tarball ala `brew fetch rust --force --build-from-source` but wound up switching to the resource version because it seemed cleaner.

Formula/racer.rb
CamJN: Didn't know that, cool.

Formula/racer.rb
CamJN: `[]` wound up being necc. but () weren't.

Formula/racer.rb
MikeMcQuaid: It's fine if it downloads it for some people but that would save having to remember to keep this up-to-date on the `rust` formula changes.

CamJN: and if we are doing something hacky why not just do this?
```
resource "rust-src" do
    url Formula['rust'].stable.url
    sha256 Formula['rust'].stable.checksum.hexdigest
end
```

Formula/racer.rb
CamJN: It results in a weird broken symlink if you do it in install.

The `/usr/local/share/rust_src/current -> /usr/local/share/rust_src/1.9.0` symlink is there and works, but for some reason there's also a `/usr/local/share/rust_src/1.9.0/1.9.0 -> /usr/local/share/rust_src/1.9.0/1.9.0` symlink too.

Formula/racer.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  resource "rust_source" do
+    url Formula["rust"].stable.url
+    sha256 Formula["rust"].stable.checksum.hexdigest
+  end
+
+  def install
+    system "cargo", "build", "--release"
+    (libexec/"bin").install "./target/release/racer"
+    (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{HOMEBREW_PREFIX}/share/rust_src/current")
```
DomT4: Should this not just be `#{share}/rust_src/current` rather than `HOMEBREW_PREFIX`? I tend to feel like `HOMEBREW_PREFIX` usage makes it sound the directory will exist across updates, which it won't here.

Apologies if I missed a discussion here. GH squashes "old" diffs which makes it hard to glance through everything.

Formula/racer.rb
```diff
@@ -0,0 +1,33 @@
...
+    resource("rust_source").stage do
+      rm_rf ["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"]
+      (share/"rust_src/#{Formula["rust"].stable.version}").install Dir["src/*"]
+    end
+  end
+
+  def post_install
+    ln_sf (share/"rust_src/#{Formula["rust"].stable.version}"), (share/"rust_src/current")
```
DomT4: Please use:

```ruby
(share/"rust_src/current").install_symlink share/"rust_src/#{Formula["rust"].stable.version}"
```

Formula/racer.rb
DomT4: I'm not sure all of the `share` usage here shouldn't actually be `pkgshare`, to be honest.

Formula/racer.rb
CamJN: This doesn't work. it makes a dir named current containing a symlink to 1.9.0 not a symlink named current pointing to 1.9.0.


```
ls -l /usr/local/Cellar/racer/1.1.0/share/rust_src/current
total 8
lrwxr-xr-x  1 camdennarzt  admin     8B 18 Jun 15:09 1.9.0 -> ../1.9.0
```

UniqMartin: Indeed. He probably meant to suggest:

```ruby
(share/"rust_src").install_symlink Formula["rust"].stable.version.to_s => "current"
```

(The relative path on the left-hand side of `=>` is resolved relative to `share/"rust_src"`.)

DomT4: Yeah, I did, sorry. I should stop reviewing PRs with less than an optimum amount of caffeine in my system.

MikeMcQuaid:
Seems good to me now. @ilovezfs, @DomT4, thoughts?

Formula/racer.rb
```diff
@@ -0,0 +1,33 @@
...
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.1.0.tar.gz"
+  sha256 "f969e66d5119f544347e9f9424e83d739eef0c75811fa1a5c77e58df621e066d"
+  head "https://github.com/phildawes/racer.git"
+
+  depends_on "rust" => :build
+
+  resource "rust_source" do
+    url Formula["rust"].stable.url
```
ilovezfs: I'm not thrilled with making this dynamic since it means a given (formula version, formula revision, bottle revision) tuple can actually mean more than one thing, and the additional information about which specific version of the sources was used wouldn't make it into the bottle or receipt.

Formula/racer.rb
```diff
@@ -0,0 +1,33 @@
...
+    sha256 Formula["rust"].stable.checksum.hexdigest
+  end
+
+  def install
+    system "cargo", "build", "--release"
+    (libexec/"bin").install "./target/release/racer"
```
ilovezfs: This can be `"target/release/racer"` instead of `"./target/release/racer"`

Formula/racer.rb
```diff
@@ -0,0 +1,33 @@
...
+    (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => "#{share}/rust_src/current")
```
ilovezfs: To be consistent it would be better as `share/"rust_src/current"`. Also I'm wondering if this would be better off in `pkgshare` than `share`?

Formula/racer.rb
MikeMcQuaid: Yeh. Another option would doing this first inside `install`:

```ruby
rust_src = Formula["rust"].fetch
Formula["rust"].verify_download_integrity(rust_src)
...
rust_src.stage do
  ...
end
```

ilovezfs: Maybe we should just have the `rust` formula keep its sources around and expose them in the rust pkgshare directory if the rust sources are generally supposed to be available to client software like this?

Formula/racer.rb
```diff
@@ -0,0 +1,33 @@
...
+    resource("rust_source").stage do
+      rm_rf ["src/llvm", "src/test", "src/librustdoc", "src/etc/snapshot.pyc"]
```
ilovezfs: `rm_rf Dir["src/{llvm,test,etc/snapshot.pyc}"]`

Formula/racer.rb
MikeMcQuaid: I think if we had more than one formula using it I'd agree.

Formula/racer.rb
```diff
@@ -0,0 +1,33 @@
...
+      (share/"rust_src/#{Formula["rust"].stable.version}").install Dir["src/*"]
+    end
+  end
+
+  def post_install
+    (share/"rust_src").install_symlink Formula["rust"].stable.version.to_s => "current"
+  end
+
+  test do
+    system "#{bin}/racer", "complete", "std::io::B"
```
ilovezfs: `bin/"racer"` to be consistent with the rest of the formula.

Formula/racer.rb
ilovezfs: In that case, the `.fetch` approach looks more "honest".

However, I'm still be concerned about corner cases here though, because the resource is living with the "wrong" formula. So for example, if I install a `rust` `--HEAD` or `--devel` version, it would be best if racer used the sources for the version I actually have installed, not some random unrelated stable spec.

MikeMcQuaid: Yeh, good point.

CamJN: @ilovezfs The thing is that currently AFAIK only racer uses the rust sources and there's no general consensus that they should be available. BUT that said, it is WAAAAY better IMO to do things the way you suggested and keep the source of the rust you installed around instead of downloading a good guess but not knowing for sure that the versions match. It could even be behind a flag or something, (though rust takes ages to compile). I mean a lot of rust programmers use racer, because it's the only completion tool for the language (aside from paying intellij for their intellij-rust IDE/plugin/experiment), so it might be worth supporting by default for an extra ~52-55 MB, I mean mysql keeps almost half a gig of garbage around that I doubt very many people use.

MikeMcQuaid:
The best option for this I think may be adding a `with-racer` option and installing this as part of the Rust formula which also knows to install sources. That let's people install them if they want them and we can evaluate if this option is used often so it can be the default.

MikeMcQuaid:
> The best option for this I think may be adding a with-racer option and installing this as part of the Rust formula which also knows to install sources. That let's people install them if they want them and we can evaluate if this option is used often so it can be the default.

@CamJN Thoughts?

CamJN:
I'm totally fine with that approach.

PR #866, opened 2016-05-05
sourcekitten: remove :build tag from Xcode dep

jpsim:
as recommended in https://github.com/Homebrew/legacy-homebrew/pull/50211#issuecomment-217037901 /cc @apjanke

PR #1520, opened 2016-05-28
sourcekitten 0.13.0

jpsim:
Thanks!

### DIFF
--- a/Formula/sourcekitten.rb
+++ b/Formula/sourcekitten.rb
@@ -2,8 +2,8 @@ class Sourcekitten < Formula
   desc "Framework and command-line tool for interacting with SourceKit"
   homepage "https://github.com/jpsim/SourceKitten"
   url "https://github.com/jpsim/SourceKitten.git",
-      :tag => "0.12.2",
-      :revision => "cb35dd8a3850b69bf612ef6fd7d5078123662bd6"
+      :tag => "0.13.0",
+      :revision => "3a70e6b718908d20553df2351f209d966df0ead7"
   head "https://github.com/jpsim/SourceKitten.git"
 
   bottle do


jpsim:
Failed due to the change in #866.

DomT4:
Isn't actually related to that change. https://github.com/Homebrew/homebrew-core/pull/933 also failed at the time, without that change included, and IIRC both Andrew & I managed to reproduce the failure locally, although I'm still not sure _why_ it happens:

```
xcodebuild -workspace 'SourceKitten.xcworkspace' -scheme 'sourcekitten' DSTROOT=/private/tmp/sourcekitten-20160529-4828-1dg1c1t/SourceKitten.dst -configuration Release clean
Build settings from command line:
    DSTROOT = /private/tmp/sourcekitten-20160529-4828-1dg1c1t/SourceKitten.dst

fatal: /usr/local/Cellar/git/2.8.3/libexec/git-core/git-submodule cannot be used without a working tree.
=== CLEAN TARGET Yaml OSX OF PROJECT Yaml WITH CONFIGURATION Release ===

Check dependencies
Stopping at 'Carthage/Checkouts/Nimble'; script returned non-zero status.

Stopping at 'Carthage/Checkouts/Commandant'; script returned non-zero status.
make: *** [bootstrap] Error 1
make: *** Waiting for unfinished jobs....
```

UniqMartin:
> […], and IIRC both Andrew & I managed to reproduce the failure locally, although I'm still not sure *why* it happens:

I can tell you why, though I'm not sure whether `git` or [our `GitDownloadStrategy#stage`](https://github.com/Homebrew/brew/blob/01b6787aa8969eeae26be401371e890d5924b992/Library/Homebrew/download_strategy.rb#L574-L577) is to blame. If one was okay with hacking the formula, here's the somewhat self-documenting workaround:

```diff
diff --git i/Formula/sourcekitten.rb w/Formula/sourcekitten.rb
index 01d4ab5b..108a28f4 100644
--- i/Formula/sourcekitten.rb
+++ w/Formula/sourcekitten.rb
@@ -14,6 +14,17 @@ class Sourcekitten < Formula
   depends_on :xcode => "7.3"

   def install
+    # Fix absolute `gitdir:` entries in `.git` files of submodules that (after
+    # staging) still point into the Homebrew cache and break the gitdir/worktree
+    # relationship (`git rev-parse --is-inside-work-tree` returns false there).
+    Dir["#{buildpath}/**/.git"].each do |f|
+      next if File.directory?(f) # Skip if `.git` is directory.
+      next if IO.read(f) !~ %r{^gitdir: /} # Skip if path is relative.
+      ohai "Fixing: #{f.strip_prefix("#{buildpath}/")}"
+      inreplace f, %r{^gitdir: /.+?/\.git/modules/},
+                   "gitdir: #{buildpath}/.git/modules/"
+    end
+
     system "make", "prefix_install", "PREFIX=#{prefix}", "TEMPORARY_FOLDER=#{buildpath}/SourceKitten.dst"
   end

```

This problem seems only to affect submodules of submodules, but I don't know why. For me, on OS X 10.11.5 with Xcode 7.3.1 and Git 2.7.4 (Apple Git-66) and after wiping my cache, this outputs:

```
$ brew install sourcekitten
[…]
==> Fixing: Carthage/Checkouts/Commandant/Carthage/Checkouts/Nimble/.git
==> Fixing: Carthage/Checkouts/Commandant/Carthage/Checkouts/Quick/.git
==> Fixing: Carthage/Checkouts/Commandant/Carthage/Checkouts/Quick/Externals/Nimble/.git
==> Fixing: Carthage/Checkouts/Commandant/Carthage/Checkouts/Result/.git
==> Fixing: Carthage/Checkouts/Commandant/Carthage/Checkouts/xcconfigs/.git
[…]
```

MikeMcQuaid:
@UniqMartin Seems that could perhaps be a more generic patch to our download strategy code?

UniqMartin:
> Seems that could perhaps be a more generic patch to our download strategy code?

Yes, I agree. This seems to be an issue with a broader range of Git versions and we don't do anything out of the ordinary when checking out a Git repository with submodules. While it might be desirable to get this fixed in Git proper (if it isn't already, I haven't investigated), we'll probably have to deal with this behavior for quite a while, so adding this workaround to our `GitDownloadStrategy#stage` makes sense.

I'll try to follow up with a PR shortly, but today has been quite busy. The above is really just a brain dump and proof of concept that I wanted to get out before going to bed.

jpsim:
Many thanks for digging into this @UniqMartin! This release isn't pressing, so I'll wait until you get the chance to make a patch to Homebrew with this change and then try the release again.

UniqMartin:
Homebrew/brew#303 addresses the Git problem and contains some more background information.

UniqMartin:
Fix applied in Homebrew/brew, build rescheduled and it now finishes fine on all machines. :tada:

UniqMartin:
Merged in 707143d6d9ad23c8e3a9ff184100b1939203dec0. Thank you for another contribution to Homebrew, @jpsim! :tada:

jpsim:
Thank you for digging into this and making a fix so quickly! 👏

PR #1533, opened 2016-05-29
haskell-stack: bump revision

ilovezfs:
Something appears to be wrong with the revision 1 bottles, so
regenerating just to be safe.


ilovezfs:
This seems to have resolved whatever was going on: https://github.com/Homebrew/homebrew-core/pull/1480#issuecomment-222384273

Issue #889, opened 2016-05-06
Formulae should also handle exporting env vars like ${FORMULA}_HOME


alb-i986:
Consider gradle. Part of its setup is to export the environment variable GRADLE_HOME.
If I use brew to install gradle, I'll need to add to my .bash_profile `export GRADLE_HOME=/usr/local/Cellar/gradle/2.13/libexec`. Then, when `brew update && brew upgrade` installs a new version of gradle, I'll need to (remember to!) manually update the export line in my `.bash_profile`.

As far as I understand, the best it is currently done is to add a caveat message saying something like "Please add #{lib exec} to your .bash_profile" (see e.g. https://github.com/Homebrew/homebrew-core/blob/master/Formula/wildfly-as.rb#L19)

How can we automate this?

One file, managed by brew, with all of the exports required by formulae.
With something like this in place, the final user would then only need to source this one file:

```
# .bash_profile

. /usr/local/Cellar/exports.sh
```

`exports.sh` may be actually sourcing other exports.sh files: each formula may have its own `exports.sh `, e.g. `/usr/local/Cellar/gradle/exports.sh `, which stores the actual export statements:

```
# /usr/local/Cellar/gradle/export.sh
export GRADLE_HOME=/usr/local/Cellar/gradle/2.13/libexec
```

Upon upgrade, a formula may completely overwrite the associated exports.sh file (as it may be easier than doing search&replace).

PS: Disclaimer: I'm just a brew user, so please bear with me, you may have better ideas for a concrete implementation

PPS: Sorry, maybe this should better go into the brew repo.

ilovezfs:
Not automating this is by design since users can have different shells and different desires for their environment. In really critical cases, a wrapper script is installed.

DomT4:
> One file, managed by brew, with all of the exports required by formulae.
With something like this in place, the final user would then only need to source this one file:

I don't want to go down this rabbit hole, personally.

We tend to handle this on a per-formula basis with env scripting. It's not uncommon, but doesn't work in 100% of cases. I don't think `GRADLE_HOME` has been mentioned to us, at least as so far as I remember, so that's possibly why discussion hasn't come up on this particular formula yet.

DomT4:
> Then, when brew update && brew upgrade installs a new version of gradle, I'll need to (remember to!) manually update the export line in my .bash_profile.

FYI, if you use for example `/usr/local/opt/gradle/libexec` you'll only need to add it the once and won't need to update it across every version.

[...]

PR #1543, opened 2016-05-29
gradle: set GRADLE_HOME in wrapper script.

MikeMcQuaid:

---

Also, cleanup the test a little.

As mentioned in #889 CC @alb-i986.

### DIFF
--- a/Formula/gradle.rb
+++ b/Formula/gradle.rb
@@ -12,14 +12,16 @@ class Gradle < Formula
 
   bottle :unneeded
 
+  depends_on :java => "1.6+"
+
   def install
     libexec.install %w[bin lib]
-    bin.install_symlink libexec+"bin/gradle"
+    (bin/"gradle").write_env_script libexec/"bin/gradle",
+      Language::Java.java_home_env.merge(:GRADLE_HOME => libexec)
   end
 
   test do
     ENV.java_cache
-    output = shell_output("#{bin}/gradle --version")
-    assert_match /Gradle #{version}/, output
+    assert_match(/Gradle #{version}/, shell_output("#{bin}/gradle --version"))
   end
 end


DomT4:
>  Does your submission pass brew audit --strict --online <formula> (after doing brew install <formula>)?

```
failed: brew audit gradle
Stacktrace

        Error: 1 problem in 1 formula
gradle:
  * Use `depends_on :java` to set JAVA_HOME
```

:trollface:

Formula/gradle.rb
```diff
@@ -12,14 +12,16 @@ class Gradle < Formula
 
   bottle :unneeded
 
+  depends_on :java => "1.6+"
+
   def install
     libexec.install %w[bin lib]
-    bin.install_symlink libexec+"bin/gradle"
+    (bin/"gradle").write_env_script libexec/"bin/gradle",
+      Language::Java.java_home_env.merge(:GRADLE_HOME => libexec)
```
jasontedor: This change breaks homebrew-installed `gradle`. At the least, it causes `gradle` to no longer respect my `JAVA_HOME`. More specifically, if I have JDK8 and JDK9-ea builds installed and point my `JAVA_HOME` to the JDK8 build for everyday development, but `gradle` will now pick up the JDK9-ea install.

```
21:30:42 ⌁49% [jason:~] $ echo $JAVA_HOME
/Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home
21:30:45 ⌁49% [jason:~] $ gradle --version

------------------------------------------------------------
Gradle 2.13
------------------------------------------------------------

Build time:   2016-04-25 04:10:10 UTC
Build number: none
Revision:     3b427b1481e46232107303c90be7b05079b05b1c

Groovy:       2.4.4
Ant:          Apache Ant(TM) version 1.9.6 compiled on June 29 2015
JVM:          9-ea (Oracle Corporation 9-ea+120)
OS:           Mac OS X 10.11.5 x86_64

21:30:48 ⌁49% [jason:~] $ 
```

Of course, this is not what I want, and `gradle` is broken on JDK9-ea builds right now.

Moreover, this change is unnecessary. The [built-in gradle script will already try to detect what `JAVA_HOME` should be *if* it is not set, but it will respect what you have set if you do have it set](https://github.com/gradle/gradle/blob/65b309141421f4aaeae35d7eab830bee9ea4b790/subprojects/plugins/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt#L70-L89). This changes completely breaks that.

Finally, looking at #889, it appears this was done for `GRADLE_HOME`. But having `GRADLE_HOME` set is completely unnecessary, and [does not do anything useful](https://github.com/gradle/gradle/search?q=GRADLE_HOME&type=Code&utf8=✓). Since `/usr/local/bin` is on the path, there's no need to set `GRADLE_HOME` and add `$GRADLE_HOME/bin` to the path.

At the least, setting `JAVA_HOME` and `GRADLE_HOME` should be wrapped in conditionals. But I think that this change should just be reverted. I opened #1650.

PR #1570, opened 2016-05-30
aircrack-ng 1.2-rc4

jellyfishcoder:
- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open Pull Requests for the same formula update/change?
- [x] Have you built your formula locally prior to submission with brew install <formula> (where -<formula> is the name of the formula you're submitting)?
- [x] Does your submission pass brew audit --strict --online <formula> (after doing brew install <formula>)?

Previously, the aircrack-ng formula was held back and considered for deletion because the newest version needed linux-only libraries, but it seems this can be fixed by compiling it from the source with certain options included. This pull request updates aircrack-ng to the latest version and adds the appropriate compiling method and dependancies. Also, this adds a test do to the formula.

### DIFF
--- a/Formula/aircrack-ng.rb
+++ b/Formula/aircrack-ng.rb
@@ -2,11 +2,10 @@ class AircrackNg < Formula
   desc "Next-generation aircrack with lots of new features"
   homepage "https://aircrack-ng.org/"
 
-  # We can't update this due to linux-only dependencies in >1.1.
-  # See https://github.com/Homebrew/homebrew/issues/29450
-  url "https://download.aircrack-ng.org/aircrack-ng-1.1.tar.gz"
-  sha256 "b136b549b7d2a2751c21793100075ea43b28de9af4c1969508bb95bcc92224ad"
-  revision 2
+  # If you compile aircrack-ng from source using Gmake, it is compatible with OS X
+  # Issue fixed by compiling aircrack-ng from source
+  url "http://download.aircrack-ng.org/aircrack-ng-1.2-rc4.tar.gz"
+  sha256 "d93ac16aade5b4d37ab8cdf6ce4b855835096ccf83deb65ffdeff6d666eaff36"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Formula/aircrack-ng.rb
```diff
@@ -2,11 +2,10 @@ class AircrackNg < Formula
...
   homepage "https://aircrack-ng.org/"
 
-  # We can't update this due to linux-only dependencies in >1.1.
-  # See https://github.com/Homebrew/homebrew/issues/29450
-  url "https://download.aircrack-ng.org/aircrack-ng-1.1.tar.gz"
-  sha256 "b136b549b7d2a2751c21793100075ea43b28de9af4c1969508bb95bcc92224ad"
-  revision 2
+  # If you compile aircrack-ng from source using Gmake, it is compatible with OS X
+  # Issue fixed by compiling aircrack-ng from source
+  url "http://download.aircrack-ng.org/aircrack-ng-1.2-rc4.tar.gz"
```
MikeMcQuaid: Need a stable version here rather than a RC.

Formula/aircrack-ng.rb
```diff
@@ -18,25 +17,26 @@ class AircrackNg < Formula
   depends_on "pkg-config" => :build
   depends_on "sqlite"
   depends_on "openssl"
+  depends_on "make"
```
MikeMcQuaid: `make` is provided by OS X.

Formula/aircrack-ng.rb
```diff
@@ -18,25 +17,26 @@ class AircrackNg < Formula
...
+  depends_on "pcre"
 
   # Remove root requirement from OUI update script. See:
   # https://github.com/Homebrew/homebrew/pull/12755
   patch :DATA
```
MikeMcQuaid: Would be good to submit this to the upstream developers.

DomT4:
Last attempt at this, for the sake of history: https://github.com/Homebrew/legacy-homebrew/pull/47728

Formula/aircrack-ng.rb
jellyfishcoder: Sorry, this is my first pull here, what do you mean by that?

Formula/aircrack-ng.rb
jellyfishcoder: This is the only version for download on the aircrack-ng official website under sources. @MikeMcQuaid

MikeMcQuaid: It's maybe worth asking them why this isn't considered to be stable?

Formula/aircrack-ng.rb
MikeMcQuaid: There's a patch here that is needed for aircrack-ng on OS X and it would be good if it was sent to its developers.

jellyfishcoder: The patch has to do with aircrack-ng usually being root only. I think its just a home-brew thing and probably is on purpose that it is root only otherwise.

MikeMcQuaid: We need patches like this to be submitted upstream regardless.

Formula/aircrack-ng.rb
jellyfishcoder: I did, they said:

Hi,

it's rc because it's not final release quality in my opinion. It still contain some bugs I want to fix. Nothing too serious though.

The reason we're using GNU is so that we don't have to write 2 different make files.

Best regards,

Thomas

MikeMcQuaid: If it's not final release quality I think we need to wait for a release that is.

jellyfishcoder: Ok, I understand.

MikeMcQuaid:
Passing on this due to mentioned reasons. We should probably consider boneyarding this formula, too.

PR #1369, opened 2016-05-23
JSON-Fortran and FORD new versions

zbeekman:
JSON-Fortran 5.0.0 released. Includes work around for regression in GFortran 6.1 (See PR #676 for more info/context)
FORD 4.6.0 released

PR #1578, opened 2016-05-30
ford 4.6.2

zbeekman:
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change? This is part I of II to supersede PR #1369

---

### DIFF
--- a/Formula/ford.rb
+++ b/Formula/ford.rb
@@ -1,8 +1,8 @@
 class Ford < Formula
   desc "Automatic documentation generator for modern Fortran programs"
   homepage "https://github.com/cmacmackin/ford/"
-  url "https://pypi.python.org/packages/source/F/FORD/FORD-4.5.4.tar.gz"
-  sha256 "fc32eb17c2aa6bfd89e4168f237b7f66a9892c00dcc94a7ce6af08d05cdfdfc1"
+  url "https://pypi.python.org/packages/85/11/7a70d720fcceddad67001dfa36bbc202d6172ba0b17e7d955db88516a33a/FORD-4.6.2.tar.gz"
+  sha256 "48062d658c712e22e459d9a2460f7fd584013289a24d13e0fc011bdbe05c99b4"
 
   head "https://github.com/cmacmackin/ford.git"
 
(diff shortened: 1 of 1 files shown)


DomT4:
Cross-posting my question from https://github.com/Homebrew/homebrew-core/pull/1580#issuecomment-222572174 as either me or someone else will forget to check in the morning I'm sure 😄.

zbeekman:
Let's hold off on merging this until after #1580 is merged and I can do some more testing with JSON-Fortran 5.0.0 to figure out why the new version of FORD causes it to fail.

zbeekman:
I just confirmed FORD 4.6.0 also causes the older 4.x.x JSON-Fortran to fail as well...

zbeekman:
Oddly, this works if I pass `--env=std` to build JSON-Fortran, but then the JSON-Fortran tests fail... (my test case to see whether FORD is working or not...)

Do you guys have any thoughts on why this is the case?

I have filed an issue upstream, and it seems like perhaps this is a regression that was re-introduced: https://github.com/cmacmackin/ford/issues/139#issuecomment-223836258

In my mind there are three options for how we can proceed:

1. wait until this issue has been resolved upstream, and new release including this fix is available
2. proceed without abandon and merge this formula upgrade... cross our fingers and hope that no one else triggers this bug, and that everyone is installing JSON-Fortran from a bottle...
3. merge this, and create a revision bump in JSON-Fortran that disables building and installing documentation by default, including a caveat about recent releases of FORD....

Thoughts?

zbeekman:
ping @DomT4 @UniqMartin

zbeekman:
Hi @DomT4 and @UniqMartin,

I have isolate the problem: Some changes were introduced to FORD to use `cpp -E -traditional... ` etc. to do any source pre-processing. In the Homebrew environment cpp resolves to clang, rather than gcc provided cpp... Additionally there was [another issue](https://github.com/cmacmackin/ford/issues/139#issuecomment-225375316) which artificially disabled `__GFORTRAN__` macro from being defined (because `cpp` does not define it)

So in summary, a bug (or at least unexpected behavior) in FORD has been fixed with the 4.6.2 release, BUT building JSON-Fortran from source will still fail if we update FORD, due to the fact that `cpp` resolves to clang in the Homebrew build environment. As such, we need to patch the JSON-Fortran build system to generate documentation using FORD more intelligently, based on which compiler is being used to build JSON-Fortran...

TL;DR: This PR is ready to be merged, IMO, however building JSON-Fortran from source will be broken if it is merged until we can put out a patch release (of JSON-Fortran, to work around the cpp ==> clang issue) which should happen pretty soon. Also, the JSON-Fortran patch is not only due to Homebrew, we want to use the compiler's companion Fortran preprocessor to be used in generating the documentation too...

UniqMartin:
Thanks for the update and sorry for not following up sooner! I was meaning to get to this earlier, but got carried away by other issues. Glad you were able to identify the cause of the issue in the mean time! :tada:

To be honest, I'm not entirely sure what your conclusion is. Is this a problem that can and should be fixed upstream by either FORD or JSON-Fortran? To me this sounds like in our build environment we shouldn't blindly forward `cpp` invocations to Apple's `cpp`, but instead to `cpp-N` if `gcc-N` is the currently selected compiler. (Not currently the case, but would make sense.) Unfortunately, this will not help with Fortran code, as the selection of the Fortran compiler is independent from what is used for C/C++ code.

zbeekman:
OK, https://github.com/jacobwilliams/json-fortran/pull/201 looks like it should be good to go, and as soon as @jacobwilliams authors a 5.0.1 release I can create a companion PR for JSON-Fortran that will prevent the build failures...

@UniqMartin:

> To be honest, I'm not entirely sure what your conclusion is. Is this a problem that can and should be fixed upstream by either FORD or JSON-Fortran?

There was a bug in FORD that needed to be fixed, and there is a new release 4.6.2 that fixed it. In addition, JSON-Fortran wants to add functionality to perform the preprocessing more intelligently. A consequence of this improvement to JSON-Fortran is that it fixes the `cpp` related issue with the Homebrew superenv.

> To me this sounds like in our build environment we shouldn't blindly forward cpp invocations to Apple's cpp, but instead to cpp-N if gcc-N is the currently selected compiler. (Not currently the case, but would make sense.) Unfortunately, this will not help with Fortran code, as the selection of the Fortran compiler is independent from what is used for C/C++ code.

I didn't look into how cpp is forwarded, but yes, I would strongly recommend matching it to the C/C++ compiler at least... As you mentioned, it won't do much good for Fortran...

With the latest JSON-Fortran improvements, if someone sets `FC=ifort` before installing from source with homebrew things should work, and the documentation will be built with Intel's `fpp` (https://github.com/jacobwilliams/json-fortran/pull/201)

Formula/ford.rb
```diff
@@ -13,52 +13,63 @@ class Ford < Formula
...
+
   depends_on "graphviz"
   depends_on :python if MacOS.version <= :snow_leopard
 
+  if build.with? "lxml"
+    resource "lxml" do
+      url "https://pypi.python.org/packages/11/1b/fe6904151b37a0d6da6e60c13583945f8ce3eae8ebd0ec763ce546358947/lxml-3.6.0.tar.gz"
+      sha256 "9c74ca28a7f0c30dca8872281b3c47705e21217c8bc63912d95c9e2a7cac6bdf"
+    end
+  end
```
UniqMartin: Please remove the outer `if`. Resources should always be listed unconditionally, though they might not always be needed depending on the build flags (and how the `install` method handles them).

zbeekman: Derps, sorry about that. Will attend to it shortly.

UniqMartin:
> There was a bug in FORD that needed to be fixed, and there is a new release 4.6.2 that fixed it. In addition, JSON-Fortran wants to add functionality to perform the preprocessing more intelligently. A consequence of this improvement to JSON-Fortran is that it fixes the `cpp` related issue with the Homebrew superenv.

Thanks! It's clear to me now. Let's wait until your JSON-Fortan fix has been merged upstream (which I understand should happen very soon). Once this has happened, feel free to cherry pick this fix into a revision-bumped `json-fortran` formula, in case a release that includes this fix takes a bit longer to be tagged (but I trust your judgment here how quickly this should happen).

> I didn't look into how cpp is forwarded, but yes, I would strongly recommend matching it to the C/C++ compiler at least... As you mentioned, it won't do much good for Fortran…

I'll see if I can address this quickly. Should be a pretty straightforward and small fix, though it might meet some opposition as this is (to my knowledge) not a bug that has been reported to affect users.

zbeekman:
> feel free to cherry pick this fix into a revision-bumped json-fortran formula

Not sure I know how to do this... I suspect there will be a quick release, but if you have the time to explain how to accomplish this I would love to learn.

Formula/ford.rb
UniqMartin: No worries! That's what review is good for. 😸

UniqMartin:
> Not sure I know how to do this... I suspect there will be a quick release, but if you have the time to explain how to accomplish this I would love to learn.

The short summary is: If your change is just a single commit and applies cleanly to the most recently released stable version, you can just take the commit URL from GitHub, append `.patch` to it, and stick it into the `url` line of a `patch do` block. If the situation is a bit more involved, creating this patch might become a bit more complicated (but hopefully not by a lot). To make sure the fix makes it to all users, you'd add a `revision 1` line to the formula or increment an already existing formula revision. Finally, depending on the build system being used and the nature of the changes you might need to regenerate the `configure` script and related files (not necessary for `json-fortran` as that uses CMake).

Feel free to ping me once the time has come and these instructions turn out to be a tad too brief.

Formula/ford.rb
UniqMartin: Ping. 😸 Per https://github.com/Homebrew/homebrew-core/pull/1914#issuecomment-225459562 I believe this is the only thing that stops us from merging this …

zbeekman: Double derps! I committed but never pushed last night...

ilovezfs:
@zbeekman #1914 is now merged. Please let us know when you've done the additional testing you mentioned above.

zbeekman:
@ilovezfs Yup already tested locally... :ship: it!

UniqMartin:
Merged in 16ef52b742b3c6ec2db1b903c6ce5663b47898e9. Thank you for another contribution to Homebrew, @zbeekman! :tada:

PR #1579, opened 2016-05-30
crystal-icr 0.2.8 (new formula)

genki:
Interactive console for Crystal Programming Language.

### DIFF
--- a/Formula/crystal-icr.rb
+++ b/Formula/crystal-icr.rb
@@ -18,7 +18,7 @@ class CrystalIcr < Formula
   end
 
   test do
-    assert_equal "icr version #{version}",
+    assert_equal "icr version #{version}\n",
       shell_output("#{bin/"icr"} -v").lines.first
   end
 end
--- a/Formula/crystal-icr.rb
+++ b/Formula/crystal-icr.rb
@@ -7,13 +7,18 @@ class CrystalIcr < Formula
   depends_on "crystal-lang"
   depends_on "readline"
 
+  patch do
+    url "https://github.com/greyblake/crystal-icr/commit/25a348bb.patch"
+    sha256 "1ed507cdda3635c998e7613e0bb4d4172af73bd7b10431d341c3559e6c01ceb0"
+  end
+
   def install
     system "make"
-    bin.install "bin/icr"
+    system "make", "install", "PREFIX=#{prefix}"
   end
 
   test do
-    assert_equal "icr version 0.2.8",
-      shell_output("#{bin/"icr"} -v").split("\n").first
+    assert_equal "icr version #{version}",
+      shell_output("#{bin/"icr"} -v").lines.first
   end
 end
--- a/Formula/crystal-icr.rb
+++ b/Formula/crystal-icr.rb
@@ -1,0 +1,19 @@
+class CrystalIcr < Formula
+  desc "Interactive console for Crystal programming language"
+  homepage "https://github.com/greyblake/crystal-icr"
+  url "https://github.com/greyblake/crystal-icr/archive/v0.2.8.tar.gz"
+  sha256 "5a235ee2f3065c6cf9479a7c32a11ba3d11f5e185d6c1458a97bd60eb11f88a3"
+
+  depends_on "crystal-lang"
+  depends_on "readline"
+
+  def install
+    system "make"
+    bin.install "bin/icr"
+  end
+
+  test do
+    assert_equal "icr version 0.2.8",
+      shell_output("#{bin/"icr"} -v").split("\n").first
+  end
+end


Formula/crystal-icr.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage "https://github.com/greyblake/crystal-icr"
+  url "https://github.com/greyblake/crystal-icr/archive/v0.2.8.tar.gz"
+  sha256 "5a235ee2f3065c6cf9479a7c32a11ba3d11f5e185d6c1458a97bd60eb11f88a3"
+
+  depends_on "crystal-lang"
+  depends_on "readline"
+
+  def install
+    system "make"
+    bin.install "bin/icr"
```
MikeMcQuaid: Is there no `make install` we can use? Would be good to submit a request for one upstream, if not.

genki: @MikeMcQuaid There isn't. Currently it makes only a single binary file, so `make install` is overkill.
If in the future, it provides other files such as man files, there should be.

MikeMcQuaid: @genki It'd be good if possible to submit a PR or issue there to request it so we don't accidentally miss added files in future.

genki: Yes, I will do so for the future versions that have such files.

MikeMcQuaid: @genki If you're not the one to update this package: how will we know if new files that should be installed have been added?

genki: This formula is only for the version 0.2.8 that has no such files (for now, and/or in the future).
If the update would have come, we can see the Makefile and check if there are such files.

MikeMcQuaid: As a user of this software I'd appreciate if you can make this issue (or ideally pull request) on behalf of Homebrew to make our future packaging easier, even if you don't agree. If you'd rather not, I'll add this to my TODO list and will do it before merging this PR.

genki: I think that making communities provided `make install` is not a business of package manager.
If there is it, we shall use it. But there isn't, `bin.install` is for that purpose, isn't it?

MikeMcQuaid: We're not "making" people do things but we do ask for things that make lives easier for all package managers. With respect, I've been helping run a package manager for 7 years now so I have a good idea on what our business should and shouldn't be 😉

genki: I respect and appreciate the great effort.
But I wonder why there is `bin.install` because it is not necessary if all packages are respecting your manner.
Is it deprecated or obsoleted?

MikeMcQuaid: Added in https://github.com/greyblake/crystal-icr/pull/20. We'll see what they say and, if accepted, we can apply it as a patch.

MikeMcQuaid: @genki And to answer your question: it's not deprecated but we do always prefer `make install` when possible but some buildsystems do not provide a native installation method like `Makefile`s do.

genki: I understood. Thank you. I will make PR again if newer version would come.

Formula/crystal-icr.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  test do
+    assert_equal "icr version 0.2.8",
+      shell_output("#{bin/"icr"} -v").split("\n").first
```
MikeMcQuaid: Instead of `.split("\n")` here you can do `lines.first` or just use `assert_match`.

MikeMcQuaid: (on the full string with lines and all)

Formula/crystal-icr.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "crystal-lang"
+  depends_on "readline"
+
+  def install
+    system "make"
+    bin.install "bin/icr"
+  end
+
+  test do
+    assert_equal "icr version 0.2.8",
```
MikeMcQuaid: `assert_equal "icr version #{version}",`

Formula/crystal-icr.rb
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Perhaps you could use `pipe_output` or similar to write in some Crystal? Thanks!

Formula/crystal-icr.rb
MikeMcQuaid: I'm fine with 🚢ing this as-is before the PR. Could you add a comment saying `PR to add 'make install': https://github.com/greyblake/crystal-icr/pull/20` above `bin.install`? thanks!

MikeMcQuaid: It was merged so instead you can add:
```ruby
# Add `make install` target.
patch do
  url "https://github.com/greyblake/crystal-icr/commit/25a348bb.patch"
  sha256 "..."
end
```
and use `make install PREFIX=#{prefix}`

Formula/crystal-icr.rb
genki: I've tested but it causes error like this:

```
make install PREFIX=/usr/local/Cellar/crystal-icr/0.2.8
mkdir -p /usr/local/Cellar/crystal-icr/0.2.8/bin
cp ./bin/icr /usr/local/Cellar/crystal-icr/0.2.8/bin
cp: ./bin/icr: No such file or directory
make: *** [install] Error 1
```

Investigating...

MikeMcQuaid: Did you run `make` first? I'll test again.

MikeMcQuaid: Opened another PR: https://github.com/greyblake/crystal-icr/pull/21

genki:
Thank you for reviewing. I've updated my PR.

Formula/crystal-icr.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def install
+    system "make"
+    system "make", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    assert_equal "icr version #{version}",
+      shell_output("#{bin/"icr"} -v").lines.first
```
MikeMcQuaid: Needs a `.chomp` on the end: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/2697/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_crystal_icr/

Also, as mentioned before: can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. It'd be good to try and run some Crystal code. Thanks!

MikeMcQuaid: `"#{bin/"icr"} -v"` should be `"#{bin}/icr -v"`

genki: Hmm, It's kinda 隔靴掻痒.
It might be difficult for me to know clearly what you expect...
I wish you would fix it directly instead of me.

MikeMcQuaid: Sorry, I don't know what 隔靴掻痒 means? CC @bfontaine who normally has better ideas on how to write better tests for such things.

Formula/crystal-icr.rb
```diff
@@ -0,0 +1,24 @@
...
+  depends_on "readline"
+
+  patch do
+    url "https://github.com/greyblake/crystal-icr/commit/25a348bb.patch"
+    sha256 "1ed507cdda3635c998e7613e0bb4d4172af73bd7b10431d341c3559e6c01ceb0"
+  end
+
+  def install
+    system "make"
+    system "make", "install", "PREFIX=#{prefix}"
```
MikeMcQuaid: `make install` patch was merged too so you can run it without `make` before: https://github.com/greyblake/crystal-icr/pull/21

MikeMcQuaid: https://github.com/greyblake/crystal-icr/commit/b18e391c.patch

DomT4: Ping on Mike's comment here. Looks like the last thing that needs fixing before this can 🚢. Would be nice to add a short comment above the patches explaining what they do as well.

Sija:
👍

MikeMcQuaid:
@genki Ping?

DomT4:
Will happily review a fresh PR but this one seems to have gone dead, so closing. Thank you nonetheless for the submission to Homebrew ❤️

PR #1580, opened 2016-05-31
json-fortran 5.0.0

zbeekman:

This is part II of II replacing PR #1369 I can't reproduce the error (locally) caused by upgrading FORD. Let's see what happens if the CI tests get fired off...

### DIFF
--- a/Formula/json-fortran.rb
+++ b/Formula/json-fortran.rb
@@ -1,8 +1,8 @@
 class JsonFortran < Formula
   desc "Fortran 2008 JSON API"
   homepage "https://github.com/jacobwilliams/json-fortran"
-  url "https://github.com/jacobwilliams/json-fortran/archive/4.3.0.tar.gz"
-  sha256 "c97f8de53e2ca9ee91fb3148bfa971b00eaea6e757b5e3d67cc4b4ff197e392e"
+  url "https://github.com/jacobwilliams/json-fortran/archive/5.0.0.tar.gz"
+  sha256 "0a1bf8788bdf8bdc6af009ae078ed374b2c95e0cfc3f30354b5dc5d3d35e5e66"
 
   head "https://github.com/jacobwilliams/json-fortran.git"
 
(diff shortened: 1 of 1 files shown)


DomT4:
Do we need to hold off on this till `FORD` is working as-expected or can this be 🚢ed?

zbeekman:
@DomT4 Unclear... I'm having trouble reproducing the issue outside of the homebrew build environment, and because of this it is difficult to determine where the problem lies... in the build environment or in FORD. I tried to patch JSON-Fortran's CMakeLists.txt to pass the `--debug` flag to FORD for a better trace, but can't seem to get the patch to apply in the homebrew formula. (It applies cleanly outside of the homebrew build...)

I have two different ideas for moving forwards:

1. merge JSON-Fortran, but hold off on FORD, so that the GCC formula upgrade can progress, and then I can continue to try to run down the problem.
2. merge FORD first, and then re-run these CI tests after FORD has been upgraded.

DomT4:
> I tried to patch JSON-Fortran's CMakeLists.txt to pass the --debug flag to FORD for a better trace, but can't seem to get the patch to apply in the homebrew formula

Inside Homebrew if you're using `std_cmake_args` you'll need to do something like:

```ruby
args = std_cmake_args
args -= %w[-DCMAKE_BUILD_TYPE=Release]
args << "-DCMAKE_BUILD_TYPE=Debug"
```
To override our release defaults.

> I have two different ideas for moving forwards

Are there downsides to either system? If we merge the updated `ford` for example could the current version of `json-fortran` outside this PR still build okay, etc?

zbeekman:
> Inside Homebrew if you're using std_cmake_args you'll need to do something like:
> 
> args = std_cmake_args
> args -= %w[-DCMAKE_BUILD_TYPE=Release]
> args << "-DCMAKE_BUILD_TYPE=Debug"
> To override our release defaults.

Thanks for the tip. It's not so much that I need a CMake Debug release type, it's that I want to modify the top level CMakeLists.txt file because FORD craps during configure time and I want to pass a special flag to FORD to get a backtrace/debug output to see  why it craps out. (The CMake build system runs FORD at configure time to determine the outputs so that proper dependency tracking etc. can happen via Makefiles... It is a lengthy process, so it also saves the documentation output for reuse.


> Are there downsides to either system? If we merge the updated ford for example could the current version of json-fortran outside this PR still build okay, etc?

Right now, upgrading FORD is preventing homebrew from building JSON-Fortran, even the older version, in the Homebrew environment. Also, JSON-Fortran is blocking #676, so my preference is merge JSON-Fortran 5.0.0 and hold off while I investigate the source of the problem with FORD.

It is possible that somehow the problem lies within JSON-Fortran's build system, and was introduced after the major refactor and improvements, but I think this is somewhat unlikely, and would prefer to un-block GCC 6.1.0.

UniqMartin:
> Also, JSON-Fortran is blocking #676, so my preference is merge JSON-Fortran 5.0.0 and hold off while I investigate the source of the problem with FORD.

Sounds like a good plan. :+1:

UniqMartin:
Sorry for the double post, but does that mean that you consider this PR as-is to be ready?

zbeekman:
@UniqMartin yes, I'm ready to :shipit: so long as you guys are...

UniqMartin:
Great! Merged in 439c0a989223b782aabe8fc5cfa025edfefd070e. Thank you for another contribution to Homebrew, @zbeekman! :tada:

PR #109, opened 2016-04-06
aws-sdk-cpp 0.10.6 (new formula)

zwass:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Does your submission pass `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Have you built your formula locally prior to submission with `brew install <formula>`?

PR #1582, opened 2016-05-31
aws-sdk-cpp add logging-only and minimize-size build options

theopolis:

---

These 2 options are highly desired for the osquery package/build. Using the AWS SDKs has become mandatory for the osquery build and the embedded logging-related SDKs are the only ones we link. Including a homebrew option to only build/install these SDKs will be constantly tested and used by osquery. This "greatly" enhances our CI by reducing the time to build by about 30 minutes and reduces the size of build host disks by about 300MB.

### DIFF
--- a/Formula/aws-sdk-cpp.rb
+++ b/Formula/aws-sdk-cpp.rb
@@ -14,6 +14,8 @@ class AwsSdkCpp < Formula
 
   option "with-static", "Build with static linking"
   option "without-http-client", "Don't include the libcurl HTTP client"
+  option "with-logging-only", "Only build logging-related SDKs"
+  option "with-minimize-size", "Request size optimization"
 
   depends_on "cmake" => :build
 
@@ -21,6 +23,8 @@ class AwsSdkCpp < Formula
     args = std_cmake_args
     args << "-DSTATIC_LINKING=1" if build.with? "static"
     args << "-DNO_HTTP_CLIENT=1" if build.without? "http-client"
+    args << "-DBUILD_ONLY=firehose;kinesis" if build.with? "logging-only"
+    args << "-DMINIMIZE_SIZE=ON" if build.with? "minimize-size"
 
     mkdir "build" do
       system "cmake", "..", *args


DomT4:
Don't have an objection to either of these, but do have a question: Are there downsides to `minimize-size`?

theopolis:
Good question!

For their CMake configuration it doesn't have any side effects on OS X. There are several for Android builds. But, when compiling for size you typically miss out on performance optimizations. The normal `-Os` versus `-O{2,3}`. Setting `MINIMIZE_SIZE` switches off these optimizations.

Formula/aws-sdk-cpp.rb
```diff
@@ -14,13 +14,17 @@ class AwsSdkCpp < Formula
 
   option "with-static", "Build with static linking"
   option "without-http-client", "Don't include the libcurl HTTP client"
+  option "with-logging-only", "Only build logging-related SDKs"
```
MikeMcQuaid: Do `without-http-client` and `with-logging-only` interact as expectect?

MikeMcQuaid:
> But, when compiling for size you typically miss out on performance optimizations. The normal -Os versus -O{2,3}. Setting MINIMIZE_SIZE switches off these optimizations.

So is it just doing `-Os`? If so, we do that already: https://github.com/Homebrew/brew/blob/f12932cf3176ffea5f4004984a390e8dc929ebbf/Library/Homebrew/extend/ENV/super.rb#L52

> Including a homebrew option to only build/install these SDKs will be constantly tested and used by osquery.

What are the options `osquery` uses by default? FWIW for C/C++ projects I've normally found when you hit these cases it makes sense to vendor dependencies in your build tree so you can have more control over how they are installed.

theopolis:
Yes, the options matrix has no hidden intra-dependence.

No, it does more than just setting -Os. And a test build shows about 8/9 size minification.

osquery uses all of the options when building statically, and only without-http-client otherwise, osquery CI would like to use the two additional.

Can you point to an example? I'm very happy with using brew to setup build environments. If an option here doesn't work then it'll be a significant effort to replicate the setup outside of brew, or outside of homebrew managed formulas-- seems like a bad time all around.

MikeMcQuaid:
> Yes, the options matrix has no hidden intra-dependence.

👍 

> No, it does more than just setting -Os. And a test build shows about 8/9 size minification.

👍 

> osquery uses all of the options when building statically, and only without-http-client otherwise, osquery CI would like to use the two additional.

👍 

> Can you point to an example? I'm very happy with using brew to setup build environments. If an option here doesn't work then it'll be a significant effort to replicate the setup outside of brew, or outside of homebrew managed formulas-- seems like a bad time all around.

An option in this case would be [creating your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) with a version of this formula that you want the configurations for so you can e.g. create a bottle with a static build and host it yourselves somewhere.

I'm sure you can appreciate that it's a tougher sell for us to maintain a new option because a single organisation wants it to be there, even if we personally like the organisation and project its for. That's not to say it won't happen but just so you're aware that if it does it's a favour rather than usual practice.

As an unrelated nudge (sadly we don't take bribes 😆) it'd be great if Facebook could consider a recurring donation to Homebrew if it finds it useful: https://github.com/homebrew/brew#donations

Thanks again @theopolis!

theopolis:
> An option in this case would be creating your own tap with a version of this formula that you want the configurations for so you can e.g. create a bottle with a static build and host it yourselves somewhere.
> 

Right!

For the above example request, I'm really interested in a real-world situation where either (1) Homebrew pushes back on a developer or maintainer, or (2) the dev/maintainer chooses to support a tap for their packages/dependency configuration or because their build/distribution situation does not fit the Homebrew model.

> I'm sure you can appreciate that it's a tougher sell for us to maintain a new option because a single organisation wants it to be there, even if we personally like the organisation and project its for. That's not to say it won't happen but just so you're aware that if it does it's a favour rather than usual practice.
> 

What is the governing manifesto for what constitutes the "tougher sell"? I look at projects I'm familiar with, Debian/FreeBSD ports/U-Boot/Linux and they all would rather accommodate within-project support rather than encourage fragmentation. I consider maintaining an out-of-band Formula fragmentation. That is IMO is the tougher sell.

Personally, I absolutely love when I can make someone's life easier with a single option. ❤️ 

A single option to support a REAL use case that a forward dependency has is slightly odd, but when the build/packaging system enables this with 1 or 2 lines of code change, that's absolutely brilliant. I must HT to you folks for making that possible! To me, it's exciting to support tight coupling of packages and their build options.

> As an unrelated nudge (sadly we don't take bribes 😆) it'd be great if Facebook could consider a recurring donation to Homebrew if it finds it useful: https://github.com/homebrew/brew#donations

99% of the work I do in Homebrew is my own time. IIRC Facebook has had some trouble with Homebrew and a general lack of support for fleet management (odd `sudo` dependence, installing to `/usr/local`, and conflicting with Chef and the machine owner's needs). Since it's not part of Facebook's infrastructure and they don't contribute/develop for Homebrew it's difficult to ask them to donate at this time.

MikeMcQuaid:
> For the above example request, I'm really interested in a real-world situation where either (1) Homebrew pushes back on a developer or maintainer, or (2) the dev/maintainer chooses to support a tap for their packages/dependency configuration or because their build/distribution situation does not fit the Homebrew model.

This is an example of us pushing back. It seems the other options we've added for `osquery` were both not sufficient and are only used together (presumably by `osquery`) in ~14% of installs. It's looking like our formula only needs one option: `--with-osquery-ci` which I hope helps convey the reason I'm not really happy with adding it.

In terms of real world situations, large projects like Qt and Chrome regularly vendor dependencies in their build tree for this exact use case: they want to build with a specific set of options.

> What is the governing recommendation for what constitutes the "tougher sell"? I look at projects I'm familiar with, Debian/FreeBSD ports/U-Boot/Linux and they all would rather accommodate within-project support rather than encourage fragmentation. I consider maintaining an out-of-band Formula fragmentation. That is IMO is the tougher sell.
> 
> Personally, I absolutely love when I can make someone's life easier with a single option. ❤️
> 
> A single option to support a REAL use case that a forward dependency has is slightly odd, but when the build/packaging system enables this with 1 or 2 lines of code change, that's absolutely brilliant. I must HT to you folks for making that possible! To me, it's exciting to support tight coupling of packages and their build options.

It's not a single option, though, it's two more options just for `osquery` on top of the previous two options just for `osquery`. Each additional option we add may just be a single line of code but doubles the number of possible build configurations we need to support for the software.

> 99% of the work I do in Homebrew is my own time. IIRC Facebook has had some trouble with Homebrew and a general lack of support for fleet management (odd sudo dependence, installing to /usr/local, and conflicting with Chef and the machine owner's needs). Since it's not part of Facebook's infrastructure and they don't contribute/develop for Homebrew it's difficult to ask them to donate at this time.

It's not so much the work _in_ Homebrew but the work _using_ Homebrew that I'd like to see companies donate to Homebrew. We're completely volunteer run and have to waste a lot of our time due to having no recurring income or corporate sponsorship (unlike `osquery` which is a Facebook sponsored or at least affiliated project).

On those specific points though: Homebrew never requires you to use `sudo` but recommends it on first install into the default location (which is optional) and it's possible to happily run Homebrew inside Chef or Puppet (see Boxen for an example of the latter).

theopolis:
@MikeMcQuaid you've lost me, I'm having way too much trouble following your arguments/rational.

I don't understand `--with-osquery-ci` suggestion. That doesn't make sense and would not work. The only option that is added to help osquery is `--with-logging-only` and it would help CI immensely as well as everyone who tries to build osquery themselves.

If you don't want to add the option I will be very sad. But let's figure this out quickly so we can stop wasting each other's time.

MikeMcQuaid:
> I don't understand --with-osquery-ci suggestion. That doesn't make sense and would not work. The only option that is added to help osquery is --with-logging-only and it would help CI immensely as well as everyone who tries to build osquery themselves.

`s/--with-osquery-ci/--with-options-only-osquery-uses` if that's clearer.

The context is https://github.com/Homebrew/legacy-homebrew/pull/50360#discussion_r57219502 and https://github.com/Homebrew/homebrew-core/pull/109#discussion_r58828512 in this discussion here. The only two options in this formula before this PR were added because `osquery` needed them. You're suggesting we add another two options because `osquery` needs them. This introduces 2^4 option combinations we need to support for everyone purely because `osquery` needs them which adds to our support burden. As mentioned before: two approaches we suggest instead are to create your own tap or vendor this library in your project. If you do decide to create a tap: please let us know so we can monitor the usage of those options and see if they are still used outside of `osquery`'s build.

> If you don't want to add the option I will be very sad. But let's figure this out quickly so we can stop wasting each other's time.

I'm not really sure what to make of this but I would say that given this is your PR you could close it at any time when you feel it's a waste of your time with no hard feelings.

theopolis:
> The context is Homebrew/legacy-homebrew#50360 (comment) and #109 (comment) in this discussion here. The only two options in this formula before this PR were added because osquery needed them. You're suggesting we add another two options because osquery needs them. This introduces 2^4 option combinations we need to support for everyone purely because osquery needs them which adds to our support burden. As mentioned before: two approaches we suggest instead are to create your own tap or vendor this library in your project. If you do decide to create a tap: please let us know so we can monitor the usage of those options and see if they are still used outside of osquery's build.

Really, two options adds to your support burden?

If merging two options to enable this much awesome helpfulness for a project in brew does not give you excitement-- I'm at a loss for why you're volunteering in the first place.

If we start supporting a tap then the support burden for me increases significantly as well as the test-ability for the osquery bottle brew distributes goes to 0. If we have to support a tap I will need to remove the osquery Formula (I don't want to do this). As well as support burden for running a tap.

theopolis:
> The context is Homebrew/legacy-homebrew#50360 (comment) and #109 (comment) in this discussion here. The only two options in this formula before this PR were added because osquery needed them. You're suggesting we add another two options because osquery needs them.

These options are sound:
```
option "with-static", "Build with static linking"
```
Is in-step with tons of other Formulas, occasionally folks want to build a static library locally. Your customers extend beyond forward-dependent packages.

```
option "without-http-client", "Don't include the libcurl HTTP client"
```
This prevents several additional dependencies for the SDK. It is used by osquery, but added to the AWS SDK because often folks want to install and build the SDK with their own HTTP implementation libraries.

The ones we want to add:
```
option "with-logging-only", "Only build logging-related SDKs"
```
Yes, this is for the good of the realm.

```
option "with-minimize-size", "Request size optimization"
```
Yes, also a nice to have, abstracted by the SDK for a reason. If building for constrained environments like Android, this is very helpful.

theopolis:
I'm still confused about the moral dilemma here. @MikeMcQuaid your asking a package to fragment and use/manage a tap rather than merge two options? Is the 2x4 options the heart of the problem-- is that something you CI test, or config all with?

Do you not want to cater to a package? Who does the Homebrew project optimize for then?

theopolis:
I thought this PR was ready to go, until @MikeMcQuaid made a friendly note:

> I'm sure you can appreciate that it's a tougher sell for us to maintain a new option because a single organisation wants it to be there, even if we personally like the organisation and project its for.

Then made a reference to taps, mentioned a donation, and I asked for clarification.How did we arrive at him closing the PR?

MikeMcQuaid:
> Really, two options adds to your support burden?

It increases our support surface which increases our support burden, yes.

> I'm still confused about the moral dilemma here. @MikeMcQuaid your asking a package to fragment and use/manage a tap rather than merge two options? Is the 2x4 options the heart of the problem-- is that something you CI test, or config all with?

We do not do CI tests on options which is a big part of the problem. If we have 2^4 options (not 2*4) then the overall build time to tests every option combination is 2^4 * 31m (the build time for this package)  -/+ whatever the differences in build times are for options. Untestable code is bad so we want to avoid it. I'm afraid it's insufficient for us to have your guarantees that `osquery` will uses these options as we want more than a single consumer for them.

> If we start supporting a tap then the support burden for me increases significantly as well as the test-ability for the osquery bottle brew distributes goes to 0. If we have to support a tap I will need to remove the osquery Formula (I don't want to do this). As well as support burden for running a tap.

We test this bottle fine already. Whether we have a `osquery` formula or not is up to the Homebrew maintainers and community and not you solely.

> Then made a reference to taps, mentioned a donation, and I asked for clarification.How did we arrive at him closing the PR?

We regularly make references to taps as alternatives. I said the donation was unrelated. You can see prior discussion of options for this specific formula in the references I linked you to previously.

I'll now reply to your email. Please hold off with more replies until you've read my replies to both and the previously referenced PRs.

theopolis:
Sounds good, let's forget these options for now. I'll try to come up with an alternative that works for both projects. 👍

PR #1586, opened 2016-05-31
osh 4.2.1

nijikon:

---

### DIFF
--- a/Formula/osh.rb
+++ b/Formula/osh.rb
@@ -1,8 +1,10 @@
 class Osh < Formula
   desc "Two ports of /bin/sh from V6 UNIX (circa 1975)"
   homepage "http://v6shell.org"
-  url "http://v6shell.org/src/osh-20160108.tar.gz"
-  sha256 "51980473429cd1f8dad255c30749b8588278caf42b28185e22b9455e8f381bf7"
+  url "http://v6shell.org/src/osh-4.2.1.tar.gz"
+  # TODO: fix this when epochs exist
+  version "20160515"
+  sha256 "2e2855c58b88d96146accbdc60f39a5745dea571b620b5f38ebf3e43d9b0ca74"
   head "https://github.com/JNeitzel/v6shell.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Formula/osh.rb
```diff
@@ -1,8 +1,8 @@
 class Osh < Formula
   desc "Two ports of /bin/sh from V6 UNIX (circa 1975)"
   homepage "http://v6shell.org"
-  url "http://v6shell.org/src/osh-20160108.tar.gz"
-  sha256 "51980473429cd1f8dad255c30749b8588278caf42b28185e22b9455e8f381bf7"
+  url "http://v6shell.org/src/osh-4.2.1.tar.gz"
```
MikeMcQuaid: What was/is the version here? Unfortunately this is another case where people may not get upgraded. CC @DomT4 on epoch thoughts (but, then again, I wonder if just assuming `20160108` is always older than `1.0.0` will fix the majority of cases in which the epoch is needed).

nijikon: Actually I was about to ask about this. I was not able to upgrade locally, I had to remove and install new version. Can we somehow handle a case like this?

MikeMcQuaid: We should but currently don't, unfortunately.

DomT4: Yeah, `20160108` beats `4.2.1` in a fight of "Which number is the bigger version". This would be an epoch situation, when implemented. More immediately you could just fake the version by switching it to the release date again, with a notation to fix that when possible.

> (but, then again, I wonder if just assuming 20160108 is always older than 1.0.0 will fix the majority of cases in which the epoch is needed)

Possibly, but that mandates almost as much extensive testing as much as an epoch system given it'll impact how we read/handle every single existing date-style version in Homebrew.

UniqMartin: I wonder if we can devise an `audit` check that at least complains that a PR downgrades a formula, so this won't be flagged as ✅ and won't be merged without scrutiny. (Could probably be part of the existing `audit` check that walks Git history and makes sure that formula `revision`s are properly incremented and reset on version bump.)

MikeMcQuaid: @UniqMartin Probably can't be in `audit` as it'll need to know before/after but could be in `brew pull` or `brew test-bot`.

DomT4: I think the revision check is part of `audit` but doesn't attempt it unless the git history is available to hunt through or something? Haven't checked for a while, but that's my vague recollection.

DomT4: @nijikon Recommend sticking a fake date `version` in for now, along with a `# TODO: fix this when epochs exist` notation for me to find later.

PR #1587, opened 2016-05-31
postgrest: GHC 8 compat with only upstream patches

ilovezfs:
Working with postgrest upstream on more surgical patching:
https://github.com/begriffs/postgrest/pull/619
- remove the -XNoImpredicativeTypes patch for bytestring-tree-builder
  and postgresql-binary, and the build-depends patch for
  hasql-transaction
- instead, use postgrest upstream patch bumping constraints on
  bytestring-tree-builder, postgresql-binary, and hasql-transaction
- remove the blanket cabal.config allow-newer for base and transformers
- instead patch wai-cors with an upstream commit that removes the
  parsers dependency and patch jwt with an upstream commit that bumps
  the base constraint to < 5

This commit allows for GHC 8 compatibility using only upstream patches,
which can be removed as soon as postgrest, wai-cors, and jwt each
release their next versions, or as soon as postgrest has a built-in fix
that removes the need to patch any of its dependencies.

### DIFF
--- a/Formula/postgrest.rb
+++ b/Formula/postgrest.rb
@@ -6,9 +6,18 @@ class Postgrest < Formula
 
   desc "Serves a fully RESTful API from any existing PostgreSQL database"
   homepage "https://github.com/begriffs/postgrest"
-  url "https://github.com/begriffs/postgrest/archive/v0.3.1.1.tar.gz"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍 when green.

PR #1589, opened 2016-05-31
orientdb 2.2.0

arxpoetica:
Update orientdb version. Added test.

### DIFF
--- a/Formula/orientdb.rb
+++ b/Formula/orientdb.rb
@@ -24,14 +24,14 @@ class Orientdb < Formula
     inreplace %W[bin/orientdb.sh bin/console.sh bin/gremlin.sh],
       '"YOUR_ORIENTDB_INSTALLATION_PATH"', libexec
 
-    inreplace "#{libexec}/config/orientdb-server-log.properties", "../log", "#{var}/log"
-
     chmod 0755, Dir["bin/*"]
     libexec.install Dir["*"]
 
-    mkpath "#{var}/log"
-    touch "#{var}/log/orientdb.err"
-    touch "#{var}/log/orientdb.log"
+    mkpath "#{var}/log/orientdb"
+    touch "#{var}/log/orientdb/orientdb.err"
+    touch "#{var}/log/orientdb/orientdb.log"
+    inreplace "#{libexec}/config/orientdb-server-log.properties", "../log", "#{var}/log/orientdb"
+    inreplace "#{libexec}/bin/orientdb.sh", "../log", "#{var}/log/orientdb"
 
     bin.install_symlink "#{libexec}/bin/orientdb.sh" => "orientdb"
     bin.install_symlink "#{libexec}/bin/console.sh" => "orientdb-console"
--- a/Formula/orientdb.rb
+++ b/Formula/orientdb.rb
@@ -1,9 +1,9 @@
 class Orientdb < Formula
   desc "Graph database"
   homepage "https://orientdb.com"
-  url "https://orientdb.com/download.php?email=unknown@unknown.com&file=orientdb-community-2.1.16.tar.gz&os=mac"
-  version "2.1.16"
-  sha256 "41ad0db53c418459d0efbf6a7f7e2b39f48467f1ec582efa925ceb38de3f3cc6"
+  url "https://orientdb.com/download.php?email=unknown@unknown.com&file=orientdb-community-2.2.0.tar.gz&os=mac"
+  version "2.2.0"
+  sha256 "32db9f634433563b15654a3f726750f35b53af77f0b9bd4fb15ca57434c252e4"
 
   bottle do
     cellar :any_skip_relocation
@@ -38,5 +38,11 @@ class Orientdb < Formula
 
   def caveats
     "Use `orientdb <start | stop | status>`, `orientdb-console` and `orientdb-gremlin`."
+  end
+
+  test do
+    # quick check if file is there
+    result = `if [ -f #{libexec}/bin/orientdb.sh ]; then echo 'true'; else echo 'false'; fi`.strip
+    result == "true" ? true : false
   end
 end
--- a/Formula/orientdb.rb
+++ b/Formula/orientdb.rb
@@ -43,6 +43,7 @@ class Orientdb < Formula
   test do
     begin
       system "#{bin}/orientdb", "start"
+      sleep 2
       assert_match "OrientDB Server v.2.2.0", shell_output("curl -I localhost:2480")
     ensure
       system "#{bin}/orientdb", "stop"
--- a/Formula/orientdb.rb
+++ b/Formula/orientdb.rb
@@ -41,8 +41,11 @@ class Orientdb < Formula
   end
 
   test do
-    # quick check if file is there
-    result = `if [ -f #{libexec}/bin/orientdb.sh ]; then echo 'true'; else echo 'false'; fi`.strip
-    result == "true" ? true : false
+    begin
+      system "#{bin}/orientdb", "start"
+      assert_match "OrientDB Server v.2.2.0", shell_output("curl -I localhost:2480")
+    ensure
+      system "#{bin}/orientdb", "stop"
+    end
   end
 end
--- a/Formula/orientdb.rb
+++ b/Formula/orientdb.rb
@@ -24,12 +24,14 @@ class Orientdb < Formula
     inreplace %W[bin/orientdb.sh bin/console.sh bin/gremlin.sh],
       '"YOUR_ORIENTDB_INSTALLATION_PATH"', libexec
 
+    inreplace "#{libexec}/config/orientdb-server-log.properties", "../log", "#{var}/log"
+
     chmod 0755, Dir["bin/*"]
     libexec.install Dir["*"]
 
-    mkpath "#{libexec}/log"
-    touch "#{libexec}/log/orientdb.err"
-    touch "#{libexec}/log/orientdb.log"
+    mkpath "#{var}/log"
+    touch "#{var}/log/orientdb.err"
+    touch "#{var}/log/orientdb.log"
 
     bin.install_symlink "#{libexec}/bin/orientdb.sh" => "orientdb"
     bin.install_symlink "#{libexec}/bin/console.sh" => "orientdb-console"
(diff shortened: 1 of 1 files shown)


Formula/orientdb.rb
```diff
@@ -39,4 +39,10 @@ def install
   def caveats
     "Use `orientdb <start | stop | status>`, `orientdb-console` and `orientdb-gremlin`."
   end
+
+  test do
+    # quick check if file is there
```
MikeMcQuaid: Can do this with `assert File.exist?` instead. It'dbe preferable to do something with e.g. `orientdb-console` instead?

arxpoetica: Preferable in what way, not sure I follow. Do something with the console and check the output? I'm not sure how that could be done.

MikeMcQuaid: In an ideal case, start it, test it, stop it. If not, checking any message from e.g. `orientdb-console` would be good (redirecting stderr with `2>&1 if necessary)

arxpoetica: Sorry, I'll be clearer. 😺 I'm not sure how to start/stop a database from Ruby (using Ruby system commands?). I work with Ruby, but I'm no expert. Do you have an example test that does such a thing. I looked at redis.rb, mongodb.rb, mysql.db, and I didn't see any such thing, but if you could point me to an example, I'd be happy to give it a go.

DomT4: It won't be an exact fit I suspect but the mechanism itself would likely be similar to [mpd](https://github.com/Homebrew/homebrew-core/blob/55183eb837f5535ff1b57051dbd0f2bd282ea3e0/Formula/mpd.rb#L127-L139)'s test.

arxpoetica: Great! I'll look at it.

MikeMcQuaid: `system "#{bin}/orientdb", "start"` and in an `ensure` `system "#{bin}/orientdb", "stop"`

arxpoetica: @MikeMcQuaid made some changes to the test.

MikeMcQuaid:
@arxpoetica Some test errors: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/2590/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_orientdb/

arxpoetica:
Darn, I was having trouble with the same fail last night, but thought I solved it. I'm a little at a loss for why `shell_output("curl -I localhost:2480")` isn't working. It's responding with `curl: (7) Failed to connect to localhost port 2480: Connection refused`. I tried searching the Interwebs for an answer; one jo thought it might be permissions, but I have nothing to substantiate that.

Any ideas why curl would fail? I'm noticing it's a common call in lots of the formula tests. Why would it be failing for this? (`localhost:2480` is the interface for the db, btw.)

arxpoetica:
Btw, technically, I don't need to `curl`—it seems to be starting and stopping just fine, but I thought that would be a good way to measure that the db is actually running, since it runs the interface also on port 2480.

MikeMcQuaid:
@arxpoetica Try adding a `sleep 2` in there.

arxpoetica:
I had sleep 0.5. I'll up it.

arxpoetica:
Done. https://github.com/Homebrew/homebrew-core/pull/1589/files#diff-8382d0e5c0b6b9bfa00336219cc630a5R46

arxpoetica:
I see it failed; same result.

MikeMcQuaid:
@arxpoetica Run `brew test orientdb` locally and you should see the same issue: `/usr/local/Cellar/orientdb/2.2.0/bin/orientdb: line 28: ../log/orientdb.log: Operation not permitted`; it's trying to write to a log file directory it shouldn't be writing to during the `test` (it should only write to `testpath`).

arxpoetica:
Ah. It's a logging issue. Somehow, it's not erroring locally, so chalk it up to environmental discrepancies. I think I `chowned` some folders I probably shouldn't have sometime. I'll keep tinkering, see if I can get around that problem.

arxpoetica:
Just for the record:

```brew test orientdb
Testing orientdb
==> Using the sandbox
==> /usr/local/Cellar/orientdb/2.2.0/bin/orientdb start
==> curl -I localhost:2480
  0   120    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
==> /usr/local/Cellar/orientdb/2.2.0/bin/orientdb stop```

Just so you know I'm not crazy. ;)

**Comment by UniqMartin** [comment]
@arxpoetica Try `brew test --sandbox orientdb` to get closer to what is done on our CI.

This test has nicely revealed that the formula is doing stuff it shouldn't be doing. In its default installation it is apparently logging to `/usr/local/Cellar/orientdb/2.2.0/libexec/log` but this directory should be somewhere in `var`, e.g. `/usr/local/var/log/orientdb`. The reason for that is that, strictly speaking, the formula prefix (`/usr/local/Cellar/orientdb/2.2.0`) is off-limits for writing operations once an installation finishes. Ideally, nothing should be modifying anything in there (except for Homebrew itself). Can you see whether the installation of `orientdb` can be fixed in this regard (which should fix the test, too).

File: Formula/orientdb.rb [hunk 2]
```diff
@@ -39,4 +39,14 @@ def install
   def caveats
     "Use `orientdb <start | stop | status>`, `orientdb-console` and `orientdb-gremlin`."
   end
+
+  test do
+    begin
+      system "#{bin}/orientdb", "start"
```
UniqMartin: This line should be outside of this block, immediately preceding the `begin`. The reason for that is that if the start fails already, there's no point in running the `ensure` and attempting to stop the server that has never been started. (Another subtlety revealed by the currently failing test.)

arxpoetica: I'm exposing all my inexperience with Ruby 😳 😁

arxpoetica: I just figured out that there's a `server.sh` file that I can run instead of the daemon, and the logging errors are now getting output correctly to the console on test. `/usr/local/Cellar/orientdb/2.2.0/libexec/bin/server.sh: line 95: /usr/local/Cellar/orientdb/2.2.0/libexec/bin/orient.pid: Operation not permitted
Can't load log handler "java.util.logging.FileHandler"...` Hopefully solution to follow shortly.

arxpoetica: It looks like there's a way to turn off the system file logging using something like `java -Dlog.file.level=OFF` http://orientdb.com/docs/2.2/Logging.html#using-the-jvm

Any tips on how I can run this command from Ruby would be appreciated.

MikeMcQuaid: Seems like updating the `orientdb-server-log.properties` file to use `#{var}/log` would be a good idea and may fix this test also.

arxpoetica:
@UniqMartin @MikeMcQuaid Would appreciate feedback on this latest commit: https://github.com/Homebrew/homebrew-core/pull/1589/commits/e2b5985aefc1f883325467ec6c68b5776fcf2028

I'm still having trouble replicating the testing environement (even after using `--sandbox`). The test just doesn't fail locally for me, so, until I can replicate the right environment, it's going to be trial and error. :(

MikeMcQuaid:
@arxpoetica Can I see your output of `env | grep -i log` and `env | grep -i orient`? Thanks!

Formula/orientdb.rb
```diff
@@ -24,12 +24,14 @@ def install
     inreplace %W[bin/orientdb.sh bin/console.sh bin/gremlin.sh],
       '"YOUR_ORIENTDB_INSTALLATION_PATH"', libexec
 
+    inreplace "#{libexec}/config/orientdb-server-log.properties", "../log", "#{var}/log"
```
MikeMcQuaid: This file doesn't seem to exist yet: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/2734/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_orientdb/

arxpoetica:
@MikeMcQuaid:

```
env | grep -i log
LOGNAME=americanyak
```
both `env | grep -i orient` and `env | grep -i orientdb` return blank, zip, zilch. (I assumed you meant `orientdb`, so I did both, in case.)

Formula/orientdb.rb
arxpoetica: Noted, and fixed. I'll push up when I get the rest working.

arxpoetica:
@MikeMcQuaid somehow in uninstalling and reinstalling the environment started failing (not good) appropriately again (good, so I can fix it). Also, I noticed that `brew test --sandbox --verbose orientdb` gives me more detail on the fail, so I should be able to work on this locally now until I get it fixed. Thanks for your help and patience.

arxpoetica:
Okay. So. It looks like I solved the log file problem by getting things in the `var` folder. But there appears to be a bigger problem now. OrientDB tries to install a pid and a couple db files into `libexec`:

```
Jun  4 20:31:56 sandboxd[132]: sh(89602) deny file-write-create /usr/local/Cellar/orientdb/2.2.0/libexec/bin/orient.pid
Jun  4 20:31:57 sandboxd[132]: java(89602) deny file-write-create /usr/local/Cellar/orientdb/2.2.0/libexec/databases/OSystem
Jun  4 20:31:57 sandboxd[132]: java(89602) deny file-write-create /usr/local/Cellar/orientdb/2.2.0/libexec/databases/OSystem
```

This is getting a little hairy. I'm worried about patching such things that are so basic to OrientDB. I might be getting in over my head now...

MikeMcQuaid:
@arxpoetica Great work! I think it may still be worth pushing through; currently these files are all being installed into locations that, it looks like, means all data will be destroyed on upgrade. Happy to help however I can.

arxpoetica:
Ha ha, spoken like a true evangelist! I haven't given up, but I'm definitely worried about doing this the wrong way. I'll keep tinkering and let you know what I come up with...

austinsmorris:
@arxpoetica, any update on this?  I'm excited to get v2.2 into homebrew.  What are the remaining issues?  I'm happy to help if I can, but I don't want to duplicate work, and I don't ruby very well....

arxpoetica:
I haven't made a lot of progress, unfortunately. I'm happy to hand it off to whomever.

austinsmorris:
I think I have at least replicated @arxpoetica's progress, but for the latest version (2.2.2) in #2004.  Therefore, this PR can be closed.

UniqMartin:
Closing in favor of #2004 as @arxpoetica seems to be happy to hand this over to @austinsmorris.

PR #1592, opened 2016-05-31
filebeat 1.2.3 (new formula)

chadrien:

---

Hi,

I know that `filebeat --version` for the test block isn't really great, but I tried to do better (like creating a temp config file, writing into a temp log file and output the result to another file) but couldn't get it working. If anyone have an idea how to test filebeat in a better way, I'd be open to it.

So apart from the test it works, the installation goes through and the service is working (tested with `brew services`).

Cheers 🍻

### DIFF
--- a/Formula/filebeat.rb
+++ b/Formula/filebeat.rb
@@ -47,6 +47,31 @@ class Filebeat < Formula
   end
 
   test do
-    system bin/"filebeat", "--version"
+    logFile = testpath/"log"
+    logFile.write ""
+
+    (testpath/"filebeat.yml").write <<-EOS.undent
+      filebeat:
+        prospectors:
+          -
+            paths:
+              - #{logFile}
+            scan_frequency: 0s
+      output:
+        file:
+          path: #{testpath}
+        console:
+          pretty: true
+    EOS
+
+    begin
+      fork { system bin/"filebeat", "-c", testpath/"filebeat.yml" }
+      sleep 5
+      system "echo 'foo bar baz' > #{logFile}"
+      sleep 10
+      assert File.exists? testpath/"filebeat"
+    ensure
+      system "pkill", "filebeat"
+    end
   end
 end
--- a/Formula/filebeat.rb
+++ b/Formula/filebeat.rb
@@ -15,7 +15,6 @@ class Filebeat < Formula
     ENV["GOPATH"] = gopath
 
     cd gopath/"src/github.com/elastic/beats/filebeat" do
-      system "go", "get"
       system "make"
       libexec.install "filebeat"
       etc.install "etc/filebeat.yml"
--- a/Formula/filebeat.rb
+++ b/Formula/filebeat.rb
@@ -1,18 +1,16 @@
 class Filebeat < Formula
   desc "File harvester, used to fetch logs files and feed them into logstash"
   homepage "https://www.elastic.co/products/beats/filebeat"
-  url "https://github.com/elastic/beats.git",
-    :tag => "v1.2.3",
-    :revision => "e849e45f713a6112145ac8309fc23bd69bc41ef8"
+  url "https://github.com/elastic/beats/archive/v1.2.3.tar.gz"
+  sha256 "8eea85de415898c362144ba533062651d8891241c738799e54cc9b17040c1fc9"
 
   head "https://github.com/elastic/beats.git"
 
   depends_on "go" => :build
 
   def install
-    contents = Dir["{*,.git,.gitignore}"]
     gopath = buildpath/"gopath"
-    (gopath/"src/github.com/elastic/beats").install contents
+    (gopath/"src/github.com/elastic/beats").install Dir["{*,.git,.gitignore}"]
 
     ENV["GOPATH"] = gopath
 
@@ -23,8 +21,10 @@ class Filebeat < Formula
       etc.install "etc/filebeat.yml"
     end
 
-    sh = bin/"filebeat"
-    sh.write "#!/usr/bin/env bash\n\n#{libexec}/filebeat -c #{etc}/filebeat.yml \"$@\""
+    (bin/"filebeat").write <<-EOS.undent
+      #!/bin/bash
+      #{libexec}/filebeat -c #{etc}/filebeat.yml "$@"
+    EOS
   end
 
   plist_options :manual => "filebeat"
@@ -47,15 +47,15 @@ class Filebeat < Formula
   end
 
   test do
-    logFile = testpath/"log"
-    logFile.write ""
+    log_file = testpath/"log"
+    touch log_file
 
     (testpath/"filebeat.yml").write <<-EOS.undent
       filebeat:
         prospectors:
           -
             paths:
-              - #{logFile}
+              - #{log_file}
             scan_frequency: 0s
       output:
         file:
@@ -67,9 +67,9 @@ class Filebeat < Formula
     begin
       fork { system bin/"filebeat", "-c", testpath/"filebeat.yml" }
       sleep 5
-      system "echo 'foo bar baz' > #{logFile}"
+      log_file.append_lines "foo bar baz"
       sleep 10
-      assert File.exists? testpath/"filebeat"
+      assert File.exist? testpath/"filebeat"
     ensure
       system "pkill", "filebeat"
     end
--- a/Formula/filebeat.rb
+++ b/Formula/filebeat.rb
@@ -1,0 +1,52 @@
+class Filebeat < Formula
+  desc "File harvester, used to fetch logs files and feed them into logstash"
+  homepage "https://www.elastic.co/products/beats/filebeat"
+  url "https://github.com/elastic/beats.git",
+    :tag => "v1.2.3",
+    :revision => "e849e45f713a6112145ac8309fc23bd69bc41ef8"
+
+  head "https://github.com/elastic/beats.git"
+
+  depends_on "go" => :build
+
+  def install
+    contents = Dir["{*,.git,.gitignore}"]
+    gopath = buildpath/"gopath"
+    (gopath/"src/github.com/elastic/beats").install contents
+
+    ENV["GOPATH"] = gopath
+
+    cd gopath/"src/github.com/elastic/beats/filebeat" do
+      system "go", "get"
+      system "make"
+      libexec.install "filebeat"
+      etc.install "etc/filebeat.yml"
+    end
+
+    sh = bin/"filebeat"
+    sh.write "#!/usr/bin/env bash\n\n#{libexec}/filebeat -c #{etc}/filebeat.yml \"$@\""
+  end
+
+  plist_options :manual => "filebeat"
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN"
+    "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>Program</key>
+        <string>#{opt_bin}/filebeat</string>
+        <key>RunAtLoad</key>
+        <true/>
+      </dict>
+    </plist>
+  EOS
+  end
+
+  test do
+    system bin/"filebeat", "--version"
+  end
+end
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> I know that filebeat --version for the test block isn't really great, but I tried to do better (like creating a temp config file, writing into a temp log file and output the result to another file) but couldn't get it working. If anyone have an idea how to test filebeat in a better way, I'd be open to it.

Could you change it to the version you couldn't get working and we'll help you get there? Thanks!

chadrien:
Sure, will do

chadrien:
OK I actually even got the tests running. Only small drawback is that they're a bit "long" (in total there is `sleep 15`) but that's because it's a mandatory time for filebeat to notice the file change and actually do its job.

Formula/filebeat.rb
```diff
@@ -0,0 +1,77 @@
+class Filebeat < Formula
+  desc "File harvester, used to fetch logs files and feed them into logstash"
+  homepage "https://www.elastic.co/products/beats/filebeat"
+  url "https://github.com/elastic/beats.git",
+    :tag => "v1.2.3",
+    :revision => "e849e45f713a6112145ac8309fc23bd69bc41ef8"
```
MikeMcQuaid: Does this use submodules? If not, we prefer tarballs.

Formula/filebeat.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  head "https://github.com/elastic/beats.git"
+
+  depends_on "go" => :build
+
+  def install
+    contents = Dir["{*,.git,.gitignore}"]
```
MikeMcQuaid: Inline this variable as it's only used once.

Formula/filebeat.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    cd gopath/"src/github.com/elastic/beats/filebeat" do
+      system "go", "get"
+      system "make"
+      libexec.install "filebeat"
+      etc.install "etc/filebeat.yml"
+    end
+
+    sh = bin/"filebeat"
+    sh.write "#!/usr/bin/env bash\n\n#{libexec}/filebeat -c #{etc}/filebeat.yml \"$@\""
```
MikeMcQuaid: `(bin/"filebeat").write`. Also, what if someone wants to specify another configuration file? Also, `EOS.undent` and `/bin/bash` are preferable.

Formula/filebeat.rb
```diff
@@ -0,0 +1,77 @@
...
+        <string>#{opt_bin}/filebeat</string>
+        <key>RunAtLoad</key>
+        <true/>
+      </dict>
+    </plist>
+  EOS
+  end
+
+  test do
+    logFile = testpath/"log"
```
MikeMcQuaid: `log_file`

Formula/filebeat.rb
```diff
@@ -0,0 +1,77 @@
...
+    logFile.write ""
```
MikeMcQuaid: `touch log_file`

Formula/filebeat.rb
```diff
@@ -0,0 +1,77 @@
...
+        file:
+          path: #{testpath}
+        console:
+          pretty: true
+    EOS
+
+    begin
+      fork { system bin/"filebeat", "-c", testpath/"filebeat.yml" }
+      sleep 5
+      system "echo 'foo bar baz' > #{logFile}"
```
MikeMcQuaid: Could use a Ruby write in append mode here.

MikeMcQuaid:
Great test, thanks!

Formula/filebeat.rb
DomT4: Our `.append_lines` method would work as well, if the file already exists.

Formula/filebeat.rb
chadrien: Will do the changes for `EOS.undent`. 

Regarding the user specifying the configuration file it would work as the `-c` option can be passed multiple times and the last one is going to be the one actually used (as you can see in the test). But you may have a point. I did that because, when compiling, the generated binary expects the default configuration to be in `$HOME/filebeat.yml`. I though it would be better to have it in `$HOMEBREW/etc/filebeat.yml` (because it's more the "unix way"). So what's your opinion on this? Let the default behaviour alone and remove this wrapper? Or maybe try to see if I can change the default path during compilation to avoid using this wrapper? Or simply keep it this way?

MikeMcQuaid: If it can be set multiple times 👍 to this way, thnaks.

chadrien:
I applied the changes from your code review ✔︎💯

Formula/filebeat.rb
```diff
@@ -0,0 +1,79 @@
...
+      system "go", "get"
+      system "make"
+      libexec.install "filebeat"
+      etc.install "etc/filebeat.yml"
+    end
+
+    sh = bin/"filebeat"
+    sh.write <<-EOS.undent
+      #!/bin/bash
+
```
MikeMcQuaid: Tiny nit: you can kill this empty line.

Formula/filebeat.rb
```diff
@@ -0,0 +1,79 @@
...
+    cd gopath/"src/github.com/elastic/beats/filebeat" do
+      system "go", "get"
+      system "make"
+      libexec.install "filebeat"
+      etc.install "etc/filebeat.yml"
+    end
+
+    sh = bin/"filebeat"
+    sh.write <<-EOS.undent
```
MikeMcQuaid: `(bin/"filebeat").write`

Formula/filebeat.rb
```diff
@@ -0,0 +1,79 @@
...
+  depends_on "go" => :build
+
+  def install
+    gopath = buildpath/"gopath"
+    (gopath/"src/github.com/elastic/beats").install Dir["{*,.git,.gitignore}"]
+
+    ENV["GOPATH"] = gopath
+
+    cd gopath/"src/github.com/elastic/beats/filebeat" do
+      system "go", "get"
```
DomT4: Please don't `go get` directly in a formula. If external resources are needed and aren't vendored should be using `go_resource`s.

chadrien: After testing it the `go get` wasn't actually required

Formula/filebeat.rb
```diff
@@ -0,0 +1,76 @@
...
+    EOS
+
+    begin
+      fork { system bin/"filebeat", "-c", testpath/"filebeat.yml" }
+      sleep 5
+      log_file.append_lines "foo bar baz"
+      sleep 10
+      assert File.exist? testpath/"filebeat"
+    ensure
+      system "pkill", "filebeat"
```
UniqMartin: Could you modify this a little bit such that `filebeat` is started as follows (this can also be moved outside of the `begin`-`ensure` block):

```ruby
filebeat_pid = fork { exec bin/"filebeat", "-c", testpath/"filebeat.yml" }
```

This would allow to use the following line instead of the one here to avoid killing other unrelated `filebeat` processes (e.g. a user has it running locally and then decides to run the test locally):

```ruby
Process.kill("TERM", filebeat_pid)
```

chadrien: I tried it already and this is not working. A process will still be alive after the test is ran.

```bash
$ ps aux | grep filebeat | grep -v grep # nothing
$ brew test filebeat
Testing filebeat
==> Using the sandbox
$ ps aux | grep filebeat | grep -v grep
chadrien        45466 103.4  0.1 573413280  11320 s002  R     5:45PM   0:17.06 /opt/homebrew/Cellar/filebeat/1.2.3/libexec/filebeat -c /opt/homebrew/etc/filebeat.yml -c /tmp/filebeat-test-20160531-45457-6i51j8/filebeat.yml
```

I'm not sure why it's not working. I've got a hunch that's it's caused by the usage of a wrapper though.

chadrien: Maybe I could just use the "real" binary (from `libexec`) instead of the wrapper in the test context?

Formula/filebeat.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    cd gopath/"src/github.com/elastic/beats/filebeat" do
+      system "make"
+      libexec.install "filebeat"
+      etc.install "etc/filebeat.yml"
+    end
+
+    (bin/"filebeat").write <<-EOS.undent
+      #!/bin/bash
+      #{libexec}/filebeat -c #{etc}/filebeat.yml "$@"
```
UniqMartin: Let's modify the wrapper to look as follows instead (the `exec` should fix the test problem):

```sh
#!/bin/sh
exec "#{libexec}/filebeat" -c "#{etc}/filebeat.yml" "$@"
```

The `exec` will make sure to replace the wrapper shell process with the `filebeat` process (there's no need to run it as a child process of the wrapper), quoting file names makes sure to avoid issues with the file names (though Homebrew is unlikely to be installed in a location with problematic file paths, e.g. including whitespace), and `/bin/sh` is sufficient (and more portable) as this isn't using any Bash-specific features.

Formula/filebeat.rb
UniqMartin: I suggested some improvements to the wrapper script that should fix this issue (and benefit everyone, not just this test). Sorry for missing this in the first place!

UniqMartin: Also I just realized that both the wrapper and this test code pass a configuration file via the `-c` argument? How does `filebeat` deal with that? Does it read both configuration files in order, does the later one override the preceding one, or does it do something else?

Formula/filebeat.rb
chadrien: I didn't know that trick. You learn everyday 🙂

UniqMartin:
@chadrien Can you please comment on https://github.com/Homebrew/homebrew-core/pull/1592#discussion_r65212469 relating to the fact that two configuration files are being passed to the program in the `test do`? Depending on your answers we might want to make some adjustments to the wrapper script and/or `.plist` file. Thanks!

Formula/filebeat.rb
chadrien: @UniqMartin sorry didn't saw this comment. The later override the preceding one. I discussed this in https://github.com/Homebrew/homebrew-core/pull/1592#discussion_r65188046

> Regarding the user specifying the configuration file it would work as the -c option can be passed multiple times and the last one is going to be the one actually used (as you can see in the test). But you may have a point. I did that because, when compiling, the generated binary expects the default configuration to be in $HOME/filebeat.yml. I though it would be better to have it in $HOMEBREW/etc/filebeat.yml (because it's more the "unix way"). So what's your opinion on this? Let the default behaviour alone and remove this wrapper? Or maybe try to see if I can change the default path during compilation to avoid using this wrapper? Or simply keep it this way?

@MikeMcQuaid was OK to keep things as they are (https://github.com/Homebrew/homebrew-core/pull/1592#discussion_r65188305).

UniqMartin: Thanks! And sorry I missed this discussion! In this case everything is fine and we can keep things as they are now. 😄

UniqMartin:
Merged in 729f5f37284b76e217c68ef2c34d78794968bd72. Thank you for this contribution to Homebrew, @chadrien! :tada:

I allowed myself to make some minor adjustment and moved the log write back into the `begin`-`ensure` block, just in case the write to the log file fails (though that's unlikely). I hope that's fine with you.

Formula/filebeat.rb
```diff
@@ -0,0 +1,75 @@
...
+    log_file = testpath/"log"
+    touch log_file
+
+    (testpath/"filebeat.yml").write <<-EOS.undent
+      filebeat:
+        prospectors:
+          -
+            paths:
+              - #{log_file}
+            scan_frequency: 0s
```
ruflin: I strongly recommend NOT to use scan_frequency: 0s even for tests. Please use something like 50ms or similar.

ruflin:
@chadrien Thanks a lot for brining filebeat to brew.

Formula/filebeat.rb
```diff
@@ -0,0 +1,75 @@
...
+  def install
+    gopath = buildpath/"gopath"
+    (gopath/"src/github.com/elastic/beats").install Dir["{*,.git,.gitignore}"]
+
+    ENV["GOPATH"] = gopath
+
+    cd gopath/"src/github.com/elastic/beats/filebeat" do
+      system "make"
+      libexec.install "filebeat"
+      etc.install "etc/filebeat.yml"
```
ruflin: Be aware that this file is not in `etc/filebeat.yml` anymore in master, but directly in the top directory.

DomT4:
@ruflin You may wish to submit a PR with any changes you wish to make. Will be the quickest way to get 'em reviewed.

PR #1597, opened 2016-05-31
git: set keychain credential helper by default.

MikeMcQuaid:

---

Apple's CLT's git also does this. Excuse the embedded tab; that's the format used by `.gitconfig`.

Also, remove the caveats that are redundant at this stage.

CC @DomT4 for thoughts.

### DIFF
--- a/Formula/git.rb
+++ b/Formula/git.rb
@@ -146,15 +146,14 @@ class Git < Formula
     # To avoid this feature hooking into the system OpenSSL, remove it.
     # If you need it, install git --with-brewed-openssl.
     rm "#{libexec}/git-core/git-imap-send" if build.without? "brewed-openssl"
-  end
 
-  def caveats; <<-EOS.undent
-    The OS X keychain credential helper has been installed to:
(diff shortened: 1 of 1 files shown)


DomT4:
Not sure why I only just got the email telling me I'd been pinged, apologies for the half hour response time, slacking for me 😉. Seems reasonable enough.

Is this potentially problematic at all in terms of someone having set up another credential helper in their global config and that suddenly being changed by this tweak? I think git obeys system before global, from memory, so just curious on that.

MikeMcQuaid:
> Is this potentially problematic at all in terms of someone having set up another credential helper in their global config and that suddenly being changed by this tweak? I think git obeys system before global, from memory, so just curious on that.

@DomT4 Nope; specified file > local > global > system so this is the lowest priority there is.

DomT4:
👍 Just edited my original comment after actually checking the documentation:

> Edit - I went and checked the documentation. Apparently local => global => system, so IMO this is more than safe enough.

It's good with me.

PR #1600, opened 2016-05-31
fltk: fix build with 10.11 SDK

ilovezfs:
Use upstream patch for 10.11 SDK, fixing the build on Yosemite when
building with SDKROOT set to Xcode 7.x's unitary 10.11 SDK.

Also, remove the <= Yosemite restriction, since this incidentally
happens to let it build on El Capitan as well. Note that there are still
El Capitan integration issues with the stable spec, which are already
fixed upstream and available in the devel spec.


MikeMcQuaid:
👍 👏

DomT4:
Really nice work here 👍. Backporting other El Cap specific fixes available is probably a good idea, but to have stable available again is great.

ilovezfs:
Thanks :)

I wonder why upstream is dragging its feet with a new release.

PR #1603, opened 2016-05-31
ejabberd 16.04

marktran:

---

### DIFF
--- a/Formula/ejabberd.rb
+++ b/Formula/ejabberd.rb
@@ -62,8 +62,8 @@ class Ejabberd < Formula
 
     resource "lager" do
       url "https://github.com/basho/lager.git",
-        :tag => "3.0.3",
-        :revision => "c7737da3a1892dbee0eb7554e236e1b041809269"
+        :tag => "3.0.2",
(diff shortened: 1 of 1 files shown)


marktran:
I removed all of the resources because the build script will actually download any needed dependencies so it isn't necessary to maintain it in the formula.

DomT4:
> I removed all of the resources since the build script will actually download any needed dependencies.

We'd prefer the current system.

We don't knowingly permit build scripts to fetch dependencies on the fly because it breaks caching for users, and often upstreams perform a lesser level of verification in terms of tags matching expected commits, etc than we do.

marktran:
Ok, I'll update this shortly with the latest dependencies since they've changed a bit from the previous version.

marktran:
Added the resources back with the latest dependencies.

nijikon:
@marktran Can you please squash this to one commit?

MikeMcQuaid:
@nijikon Probably worth just doing that locally when we pull.

MikeMcQuaid:

PR #1607, opened 2016-06-01
emojify 1.0.0 (new formula)

johnhampton:

---

### DIFF
--- a/Formula/emojify.rb
+++ b/Formula/emojify.rb
@@ -1,0 +1,15 @@
+class Emojify < Formula
+  desc "Emoji on the command-line :scream:"
+  homepage "https://github.com/mrowa44/emojify"
+  url "https://github.com/mrowa44/emojify/archive/v1.0.0.tar.gz"
+  sha256 "fabefc4767428a2634a77e7845e315725b75b50f282d0943c5b65789650c25d1"
+  head "https://github.com/mrowa44/emojify.git"
+
+  def install
+    bin.install "emojify"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


Formula/emojify.rb
```diff
@@ -0,0 +1,15 @@
+class Emojify < Formula
+  desc "Emoji on the command-line :scream:"
+  homepage "https://github.com/mrowa44/emojify"
+  url "https://github.com/mrowa44/emojify/archive/v1.0.0.tar.gz"
+  sha256 "fabefc4767428a2634a77e7845e315725b75b50f282d0943c5b65789650c25d1"
+  head "https://github.com/mrowa44/emojify.git"
+
+  def install
```
UniqMartin: This seems to be a simple shell script that doesn't need to be built. In this case we don't need to create bottles (binary packages). To denote this, please add `bottle :unneeded` between the `head …` and this line (leave an empty line before and after).

Formula/emojify.rb
```diff
@@ -0,0 +1,15 @@
...
+    bin.install "emojify"
+  end
+
+  test do
+    assert_equal("Hey, I just 🙋 you, and this is 😱 , but here's my 📲 , so 📞 me, maybe?", shell_output("#{bin}/emojify \"Hey, I just :raising_hand: you, and this is :scream: , but here's my :calling: , so :telephone_receiver: me, maybe?\"").strip)
```
UniqMartin: Please put the input and output into separate variables to make this line a bit shorter and more readable. This could look something like this:

```ruby
input = "Hey, I just :raising_hand: you, and this is :scream: , but here's my :calling: , so :telephone_receiver: me, maybe?"
expect = "Hey, I just 🙋 you, and this is 😱 , but here's my 📲 , so 📞 me, maybe?\n"
assert_equal expect, shell_output("#{bin}/emojify \"#{input}\"")
```

Also, are the somewhat odd looking spaces before the commas a limitation of the script?

johnhampton:
Yes, the spaces before the commas are needed for the script.

UniqMartin:
Merged. Thank you for this contribution to Homebrew, @johnhampton! :tada:

PR #1610, opened 2016-06-01
moco 0.11.0

dreamhead:
Upgrade moco to 0.11.0

### DIFF
--- a/Formula/moco.rb
+++ b/Formula/moco.rb
@@ -1,15 +1,15 @@
 class Moco < Formula
   desc "Stub server with Maven, Gradle, Scala, and shell integration"
   homepage "https://github.com/dreamhead/moco"
-  url "https://search.maven.org/remotecontent?filepath=com/github/dreamhead/moco-runner/0.10.2/moco-runner-0.10.2-standalone.jar"
-  version "0.10.2"
-  sha256 "ef946d090d3108843708c194809c57f192e3623dbbbcf86bdee54ce93c299a41"
+  url "https://repo1.maven.org/maven2/com/github/dreamhead/moco-runner/0.11.0/moco-runner-0.11.0-standalone.jar"
+  version "0.11.0"
+  sha256 "d3c772333f6a35fe4bc168d0541e97a0d36071afe343ced840f7afd1c037b661"
 
   bottle :unneeded
 
   def install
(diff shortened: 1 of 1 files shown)


Formula/moco.rb
```diff
@@ -1,15 +1,15 @@
 class Moco < Formula
   desc "Stub server with Maven, Gradle, Scala, and shell integration"
   homepage "https://github.com/dreamhead/moco"
-  url "https://search.maven.org/remotecontent?filepath=com/github/dreamhead/moco-runner/0.10.2/moco-runner-0.10.2-standalone.jar"
-  version "0.10.2"
-  sha256 "ef946d090d3108843708c194809c57f192e3623dbbbcf86bdee54ce93c299a41"
+  url "https://repo1.maven.org/maven2/com/github/dreamhead/moco-runner/0.11.0/moco-runner-0.11.0-standalone.jar"
```
UniqMartin: Can you please retain the original URL and just update the version components in the URL? I believe we want to continue using `https://search.maven.org/` instead of hard-coding a specific server.

dreamhead: I'm not quite sure **https://search.maven.org/** is a good idea? Could you please give some background?

If you don't want to use specific server, I would like to use the central maven.

http://central.maven.org/maven2/com/github/dreamhead/moco-runner/0.11.0/moco-runner-0.11.0-standalone.jar

UniqMartin: A few very simple reasons:

- It was used before in this formula.
- We use it for the majority of our formulae that we fetch from maven.org.
- It will redirect to an appropriate download server and thus probably better distribute the load. (And it is documented in [Maven Central's API](https://search.maven.org/#api).)
- central.maven.org has the disadvantage that it's not accessible via HTTPS.

That said, I have no definite resource that I can quote. What are your reasons to move away from the previously used URL in this formula? (If you have good reasons, I'd like to hear them.)

dreamhead: Great. I'd like to follow the convention.

I'm the Moco creator. I use that URL in Moco document. But I will use **search** here.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/8cd92ea9077b6f69fa4d53e28f6b9fe0730b7490. Thank you for your contribution to Homebrew @dreamhead; we appreciate it! :smiley_cat:

PR #1622, opened 2016-06-01
gd: remove gnu-sed dependency via upstream patch

ilovezfs:
Upstream patch applies cleanly if the source archive tarball is used

### DIFF
--- a/Formula/gd.rb
+++ b/Formula/gd.rb
@@ -1,6 +1,7 @@
 class Gd < Formula
   desc "Graphics library to dynamically manipulate images"
   homepage "https://libgd.github.io/"
+  revision 1
 
   stable do
     url "https://github.com/libgd/libgd/archive/gd-2.2.1.tar.gz"
@@ -52,7 +53,7 @@ class Gd < Formula
   depends_on "jpeg" => :recommended
   depends_on "libpng" => :recommended
   depends_on "libtiff" => :recommended
-  depends_on "webp" => :optional
+  depends_on "webp" => :recommended
 
   fails_with :llvm do
     build 2326
--- a/Formula/gd.rb
+++ b/Formula/gd.rb
@@ -3,8 +3,17 @@ class Gd < Formula
   homepage "https://libgd.github.io/"
 
   stable do
-    url "https://github.com/libgd/libgd/releases/download/gd-2.2.1/libgd-2.2.1.tar.xz"
-    sha256 "708762ae483e5fe46b58659f622c3e8f820c7ce0b3ae4e10ad0fbf17d4c4b976"
+    url "https://github.com/libgd/libgd/archive/gd-2.2.1.tar.gz"
+    sha256 "06b0f2ef45fbdde7b05816a8a988868b11ac348f77ffa15a958128a8106b1e08"
+
+    # Prevents need for a dependency on gnu-sed; fixes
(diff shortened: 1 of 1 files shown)


DomT4:
This was actually a deliberate choice by me, FWIW.

It's often the case that when we switch from release tarballs to archive tarballs "temporarily" we never, ever unswitch and then someone follows the notations to remove the deps, wonders why the formula doesn't build anymore and I end up letting out a long sigh 😉.

Given the bottle is `cellar :any` it's very unlikely more than a tiny amount of people are getting whacked with `gnu-sed` installations, as well. I'm fine with you changing it if you really want, but the choice on this mechanism wasn't a random one 😄.

MikeMcQuaid:
👍

ilovezfs:
My general theory has always been that most @DomT4 actions are intentional.

Since the other patch already created an Autotools dependency, I don't see the harm. In this case anyone who had libvpx or who wants webp is getting whacked with the gnu-sed dependency I believe.

We sure do need a more general solution to the "this formula has changes that are only needed due to the included patches, which when removed will eliminate the need for those changes, so please don't forget to unwind those changes as soon as possible if you are the next maintainer to update this formula due to a release bump" problem.

ilovezfs:
@DomT4  Question: why is `depends_on "webp"` `:optional` not `:recommended`?

MikeMcQuaid:
@ilovezfs Fewer dependencies = betterer.

ilovezfs:
Hehe … it's not like webp is obscure …

ilovezfs:
Looking at
```
  depends_on "fontconfig" => :recommended
  depends_on "freetype" => :recommended
  depends_on "jpeg" => :recommended
  depends_on "webp" => :optional
```
personally I'd be annoyed if the only reason I had to do a source build was that I wanted the brand new webp option, given all that other cruft that's already being "recommended"

MikeMcQuaid:
Is WebP a widely used format at this point? We have analytics now to help decide on these types of options, at least.

DomT4:
> In this case anyone who had libvpx or who wants webp is getting whacked with the gnu-sed dependency I believe.

Shouldn't be the case about `libvpx`? Dead options are just discarded now, I believe. Mike or Xu or someone fixed that a while back.

> We sure do need a more general solution to the

Yeah, we do 👍. A lot of the time I try to get to the notation-heavy formula before others do to be honest because inevitably a fair number of people either don't read the notes and just update the version, which I can't really blame them for but does make life more complicated for us in the long run.

> is WebP a widely used format at this point?

Almost nobody apparently (8k hits total over last month) executes `brew install webp`, but it is dragged in heavily from other formulae.

Because the option is new and there's no proven demand for it yet I left it as `:optional` but my feelings on that aren't particularly strong.

ilovezfs:
You're probably right about it not triggering it automatically. However webp is upstream's replacement for libvpx, so that probably should have been treated as a deprecated option pointing at webp rather than just removed, since now all those people have to notice and re-opt-in, especially since it was made optional.

ilovezfs:
https://github.com/libgd/libgd/issues/129

DomT4:
Yeah, I'm aware of the history, I spend too much of my life reading Changelogs.

Frankly I wouldn't have made all of the options `:recommended` if creating this formula from scratch, but since the damage has already been done on this dance I don't object if you want to go ahead and make it `:recommended`.

ilovezfs:
@DomT4 Done.

PR #1628, opened 2016-06-01
judy et al.: boneyard

### LINKED ISSUE #1562: Judy: undefined behavior in source code, not yet triggered by clang
Judy is a C library where upstream is dead (no activity since 2009). Homebrew packages the latest upstream stable release.

The latest release contains code that triggers an undefined behavior. On GCC 4.9+, it generates wrong code. You can see here a test that checks whether wrong code is generated (bottom of the script):
https://github.com/traildb/traildb/blob/master/wscript

The fix for this bug was written by the original author of the library, but a new release is unlikely to happen:
Ticket for upstream: https://sourceforge.net/p/judy/patches/5/
Patch in readable format: https://github.com/rasky/judy/commit/2a75dfd07094a8e33eb83b7550906c8950760200

This patch has been applied by Debian/Ubuntu at least in their repositories, to fix the wrong-code issue.

Currently, clang (as of 10.11.5) doesn't trigger the wrong code generation. I wanted to know from Homebrew maintainers how to handle this. 

Should we pre-emptively add the patch (and the test) in case a newer clang breaks it as GCC does?
Should we just add the test to the formula, so that if it breaks at some point, the test will tell us (and somebody will patch it at that point)?
Should we do nothing and cross fingers?

DomT4:

---

Closes #1562


DomT4:
I'm crushingly disappointed to discover you can't use emojis as PR titles on GitHub 😄.

PR #1629, opened 2016-06-01
liquidprompt HEAD, devel

augmentedfourth:
Moving HEAD into its own block & adding a separate devel block

### DIFF
--- a/Formula/liquidprompt.rb
+++ b/Formula/liquidprompt.rb
@@ -3,9 +3,18 @@ class Liquidprompt < Formula
   homepage "https://github.com/nojhan/liquidprompt"
   url "https://github.com/nojhan/liquidprompt/archive/v_1.10.tar.gz"
   sha256 "5a1e01b86e81002ac20c61cbd96f5fd276db7f3fa09ef220f3db0fb7aa614e24"
-  head "https://github.com/nojhan/liquidprompt.git", :branch => "develop"
 
   bottle :unneeded
+
+  head do
+    url "https://github.com/nojhan/liquidprompt.git", :branch => "master"
+    version "head"
+  end
+
+  devel do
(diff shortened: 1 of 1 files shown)


augmentedfourth:
The project has changed its structure & now maintains head/master as separate from devel/develop.

Formula/liquidprompt.rb
```diff
@@ -3,10 +3,19 @@ class Liquidprompt < Formula
   homepage "https://github.com/nojhan/liquidprompt"
   url "https://github.com/nojhan/liquidprompt/archive/v_1.10.tar.gz"
   sha256 "5a1e01b86e81002ac20c61cbd96f5fd276db7f3fa09ef220f3db0fb7aa614e24"
-  head "https://github.com/nojhan/liquidprompt.git", :branch => "develop"
 
   bottle :unneeded
 
+  head do
```
DomT4: This doesn't need moving, or the explicit `version`. You just need to remove the `:branch => "develop"` from the original line here.

Formula/liquidprompt.rb
```diff
@@ -3,10 +3,19 @@ class Liquidprompt < Formula
...
+    url "https://github.com/nojhan/liquidprompt.git", :branch => "master"
+    version "head"
+  end
+
+  devel do
```
DomT4: We don't accept new `devel` blocks that aren't pinned to a tarball or specific git tag/revision. We plan to eventually bottle `devel` stuff and we can't do that if they can move around freely.

augmentedfourth:
fixed according to suggestions from @DomT4.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/b89da345b00c9ee69c62e5307371185603153b2d. Thank you for your contribution to Homebrew @augmentedfourth; we appreciate it! :smiley_cat:

Issue #1486, opened 2016-05-27
idris does not build under GHC 8


ilovezfs:
I've reported the problem upstream, but putting this here so it's not forgotten: https://github.com/idris-lang/Idris-dev/issues/3193

MikeMcQuaid:
We should boneyard this if it's known to fail.

ilovezfs:
Yeah, I think that's where this probably belongs at least for the short term. I poked my HEAD into #idris on IRC freenode yesterday. No one seems to know what's wrong yet. I was directed to follow a particular guide to do a from scratch GHC 8 HEAD build to rule out any anomalies with the Homebrew formula, which I did, and ended up hitting the same error.

ilovezfs:
Since upstream acknowledges this is a serious bug and approves of the workaround we're using for the time being, I'm closing this for now in favor of the confirmed upstream report. With any luck, the next major release will have GHC 8 support.

PR #1630, opened 2016-06-01
idris: install using haskell-stack

ilovezfs:
- use haskell-stack to build idris to work around GHC 8 incompatibility
- minor style changes to the test-do block

Issue #1486.

### DIFF
--- a/Formula/idris.rb
+++ b/Formula/idris.rb
@@ -7,6 +7,7 @@ class Idris < Formula
   homepage "http://www.idris-lang.org"
   url "https://github.com/idris-lang/Idris-dev/archive/v0.11.2.tar.gz"
   sha256 "eddc8d8d6401d5c5743df43003a11d031be42eb6b09800dc1111606c39a3c8bc"
+  revision 1
   head "https://github.com/idris-lang/Idris-dev.git"
 
   bottle do
@@ -15,18 +16,31 @@ class Idris < Formula
     sha256 "652da19c6f08b65483df86768813f6ce7eca2d8c31cf0a0f4f624a22344cdbef" => :mavericks
   end
 
-  depends_on "ghc" => :build
-  depends_on "cabal-install" => :build
+  depends_on "haskell-stack" => :build
+  depends_on "pkg-config" => :build
   depends_on "gmp"
-
-  depends_on "libffi" => :recommended
-  depends_on "pkg-config" => :build if build.with? "libffi"
+  depends_on "libffi"
 
   def install
-    args = []
-    args << "-f FFI" if build.with? "libffi"
-    args << "-f release" if build.stable?
-    install_cabal_package *args
+    ENV["STACK_ROOT"] = "#{HOMEBREW_CACHE}/stack_root"
+    (libexec/"stack").install Dir["#{buildpath}/*"]
+
+    cd libexec/"stack" do
+      system "stack", "setup"
+      system "stack", "build"
+
+      bin.mkpath
+      system "stack", "install", "--local-bin-path=#{bin}"
+
+      install_dir = Utils.popen_read("stack", "path", "--local-install-root").chomp
+
+      share.mkpath
+      mv "#{install_dir}/share/man", share
+      mv "#{install_dir}/doc", share
+
+      libexec.install_symlink "#{install_dir}/lib"
+      libexec.install_symlink "#{install_dir}/share"
+    end
   end
 
   test do
@@ -44,14 +58,10 @@ class Idris < Formula
       main : IO ()
       main = puts "Hello, interpreter!"
     EOS
-    shell_output "#{bin}/idris #{testpath}/hello.idr -o #{testpath}/hello"
-    result = shell_output "#{testpath}/hello"
-    assert_match /Hello, Homebrew!/, result
+    system bin/"idris", testpath/"hello.idr", "-o", testpath/"hello"
+    assert_match /Hello, Homebrew!/, shell_output(testpath/"hello")
 
-    if build.with? "libffi"
-      shell_output "#{bin}/idris #{testpath}/ffi.idr -o #{testpath}/ffi"
-      result = shell_output "#{testpath}/ffi"
-      assert_match /Hello, interpreter!/, result
-    end
+    system bin/"idris", testpath/"ffi.idr", "-o", testpath/"ffi"
+    assert_match /Hello, interpreter!/, shell_output(testpath/"ffi")
   end
 end


MikeMcQuaid:
👍

ilovezfs:
@MikeMcQuaid That would be lovely, but I expect it to fail. I want to see it fail on CI not just locally, to rule out any local issues.

However, I'm experimenting locally with a stack-based workaround:
```
  def install
    ENV["STACK_ROOT"] = "#{HOMEBREW_CACHE}/stack_root"
    libexec.install Dir["#{buildpath}/*"]
    cd libexec
    system "stack", "setup"
    system "stack", "build", "--work-dir=stack_work_dir"
    system "stack", "install", "--work-dir=stack_work_dir", "--local-bin-path=#{bin}"
    doc.install_symlink Dir["#{libexec}/stack_work_dir/install/x86_64-osx/lts-5.11/7.10.3/doc/*"]
    share.install_symlink Dir["#{libexec}/stack_work_dir/install/x86_64-osx/lts-5.11/7.10.3/share/*"]
  end
```

ilovezfs:
Yup, same failure as here:
```
Building libraries...
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C prelude build
../../dist/dist-sandbox-a2ddac0/build/idris/idris --build prelude.ipkg
./Builtins.idr:6:1: error: expected: declaration,
    end of input
||| The canonical single-element type, also known as the trivially 
^                                                                  
Type checking ./Builtins.idr
make[1]: *** [build] Error 1
make: *** [build] Error 2
cabal: Leaving directory '.'
cabal: Error: some packages failed to install:
idris-0.11.2 failed during the building phase. The exception was:
ExitFailure 2
```

ilovezfs:
Stack-based workaround installs and passes the test:
```
==> Summary
🍺  /usr/local/Cellar/idris/0.11.2_2: 2,324 files, 936.0M, built in 8 minutes 39 seconds
iMac-TMP:Idris-dev joe$ brew test -v idris
Testing idris
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160601-56408-1fj3ukz.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/idris.rb -v
==> /usr/local/Cellar/idris/0.11.2_2/bin/idris /tmp/idris-test-20160601-56409-1w546mh/hello.idr -o /tmp/idris-test-20160601-56409-1w546mh/hello
==> /tmp/idris-test-20160601-56409-1w546mh/hello
iMac-TMP:Idris-dev joe$
```

ilovezfs:
@MikeMcQuaid I've updated the PR with the `haskell-stack` based workaround.

Also, CC @zmwangx and @borsboom (RE: https://github.com/commercialhaskell/stack/issues/848)

Formula/idris.rb
```diff
@@ -15,18 +16,31 @@ class Idris < Formula
...
+    (libexec/"stack").install Dir["#{buildpath}/*"]
+
+    cd libexec/"stack" do
+      system "stack", "setup"
+      system "stack", "build"
+
+      mkdir_p bin
+      system "stack", "install", "--local-bin-path=#{bin}"
+
+      install_dir = libexec/"stack/.stack-work/install/x86_64-osx/lts-5.11/7.10.3"
```
zmwangx: This is why we need `stack install --prefix`...

ilovezfs: indeed.

borsboom:
You might want to use the output of `stack path --local-install-root` instead of hard-coding `.stack-work/install/x86_64-osx/lts-5.11/7.10.3` which could change in the future.

It looks like you're copying out of the `.stack-work` directory as a workaround for the lack of a `stack --prefix` option.  I don't know if this is a problem with Idris in particular, but I'll just warn you that some Haskell packages are not relocatable because they end up with compiled-in hard-coded paths pointing to the original build directory (finding an elegant solution to this is the main holdup of supporting a `stack --prefix option`, but we recently got some code contributed that attempts to make these relocatable so hopefully we'll have a solution soon).

ilovezfs:
>You might want to use the output of stack path --local-install-root instead of hard-coding .stack-work/install/x86_64-osx/lts-5.11/7.10.3 which could change in the future.

Good idea.

> for the lack of a stack --prefix option

Pretty please, with sugar on top.

> I don't know if this is a problem with Idris in particular

It is, which is why I had to leave the `.stack-work` intact for everything except the docs and executables.

> we recently got some code contributed that attempts to make these relocatable.

Nice! Do you happen to have a link?

ilovezfs:
Woo hoo :green_apple:

Formula/idris.rb
```diff
@@ -15,18 +16,31 @@ class Idris < Formula
...
-    args << "-f release" if build.stable?
-    install_cabal_package *args
+    ENV["STACK_ROOT"] = "#{HOMEBREW_CACHE}/stack_root"
+    (libexec/"stack").install Dir["#{buildpath}/*"]
+
+    cd libexec/"stack" do
+      system "stack", "setup"
+      system "stack", "build"
+
+      mkdir_p bin
```
MikeMcQuaid: Tiniest nit that normally we go for `bin.mkpath`

Formula/idris.rb
```diff
@@ -15,18 +16,31 @@ class Idris < Formula
...
+      system "stack", "install", "--local-bin-path=#{bin}"
+
+      install_dir = libexec/"stack/.stack-work/install/x86_64-osx/lts-5.11/7.10.3"
+
+      mkdir_p share
+      mv install_dir/"share/man", share
```
MikeMcQuaid: `share.install` here and below.

MikeMcQuaid:
Tiny nits otherwise 👍

ilovezfs:
@MikeMcQuaid Sweet.

Formula/idris.rb
ilovezfs: @MikeMcQuaid 
```
install_dir = IO.popen("stack path --local-install-root") { |io| io.gets.chomp }
```
or
```
install_dir = `stack path --local-install-root`.chomp
```
?

ilovezfs:
@MikeMcQuaid Also, we can optionally make the `.stack-work` visible using `--work-dir=whatever_subfolder_name_you_want`. Any preference between the default `.stack-work` or something else?

Formula/idris.rb
MikeMcQuaid: `install_dir = Utils.popen("stack path --local-install-root").chomp`

ilovezfs: I guess that should be 
```
Utils.popen_read("stack", "path", "--local-install-root").chomp
```
or
```
Utils.popen("stack path --local-install-root", "r").chomp
```
preference?

ilovezfs:
@DomT4 Any changes desired?

ilovezfs:
@MikeMcQuaid any thoughts on https://github.com/Homebrew/homebrew-core/pull/1630#issuecomment-223112449? I don't know if there's any precedent for a significant invisible dot directory that's actually installed within the Cellar, but I don't mind the `.stack-work` default if you don't.

MikeMcQuaid:
@ilovezfs I think I'd rather it's not in the Cellar if possible? I'm not sure I understand the contents or purpose, though, so I'm lacking context currently.

ilovezfs:
It's required that it be installed since it contains the installation directory ("work-dir" is somewhat of a misnomer, or at least an incomplete description). The core issue is that some paths can get hard coded (which they indeed do in the case of Idris), and the stack work-dir must be a subfolder of the directory in which you're building. It can have whatever name you'd like, but the default is `libexec/"stack/.stack-work"` since we're building in `libexec/"stack"`. Here's what's ultimately installed: https://gist.github.com/ilovezfs/0dd1839e08440f29471ff6aa40554762

MikeMcQuaid:
I see, yuck. Normally I'm pretty 👎 on installation happening in the prefix but if it's unavoidable to fix up this package that's otherwise broken then a begruding 👍. I think in this case it's better to be a hidden/dotted work directory then, yeh.

ilovezfs:
>I see, yuck. Normally I'm pretty :-1: on installation happening in the prefix 

@MikeMcQuaid hehe. Yes indeed. That's why I flagged what's happening here with a neon sign for you!

ilovezfs:
On the bright side, this is a proof-of-concept for using stack with brew, even with its current limitations, and there's reason to be hopeful that the situation will improve in the future as there are upstream `stack` plans for the ability to locate the `work-dir` outside of the build path and, even more significantly, to add a `--prefix` option for installation.

ilovezfs:
:ship: 'd

PR #1658, opened 2016-06-02
libchewing 0.5.1 (new formula)

Chocobo1:

---

### DIFF
--- a/Formula/libchewing.rb
+++ b/Formula/libchewing.rb
@@ -1,0 +1,49 @@
+class Libchewing < Formula
+  desc "Intelligent phonetic input method library"
+  homepage "http://chewing.im/"
+  url "https://github.com/chewing/libchewing/releases/download/v0.5.1/libchewing-0.5.1.tar.bz2"
+  sha256 "9708c63415fa6034435c0f38100e7d30d0e1bac927f67bec6dfeb3fef016172b"
+
+  head do
+    url "https://github.com/chewing/libchewing.git"
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+  end
+
+  depends_on "texinfo" => :build
+
+  def install
+    if build.head?
+      system "./autogen.sh"
+    end
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+    #include <stdlib.h>
+    #include <chewing/chewing.h>
+    int main()
+    {
+        ChewingContext *ctx = chewing_new();
+        chewing_handle_Default(ctx, 'x');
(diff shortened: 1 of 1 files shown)


Formula/libchewing.rb
```diff
@@ -0,0 +1,48 @@
...
+  homepage "http://chewing.im/"
+  url "https://github.com/chewing/libchewing/releases/download/v0.5.1/libchewing-0.5.1.tar.bz2"
+  sha256 "9708c63415fa6034435c0f38100e7d30d0e1bac927f67bec6dfeb3fef016172b"
+  head "https://github.com/chewing/libchewing.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "texinfo" => :build
+  depends_on "sqlite" => :build
```
MikeMcQuaid: This isn't linked against? Can the system version be used instead?

Chocobo1: >Can the system version be used instead?

You are right, PR updated.

MikeMcQuaid: @Chocobo1 To use the system one you remove `depends_on "sqlite" altogether. Does that work locally?

Formula/libchewing.rb
```diff
@@ -0,0 +1,47 @@
...
+  head "https://github.com/chewing/libchewing.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "texinfo" => :build
+  depends_on "sqlite"
+
+  def install
+    if build.head?
```
DomT4: You've wrapped this in a `build.head?` but you've left all the autotool deps for both stable & head. Intentional or?

Chocobo1: Thanks for pointing out, will fix it.

Formula/libchewing.rb
Chocobo1: >@Chocobo1 To use the system one you remove `depends_on "sqlite" altogether. Does that work locally?

You mean the one installed in system but not from `brew` right?
But what if some user didn't install `sqlite` and just `brew install libchewing`, guess that way the dependency is broken, what is right way to do for this situation?

ps. pardon me, just my first time toying with osx.

DomT4: `sqlite` comes shipped with OS X as standard, yeah:

```
~> /usr/bin/sqlite3 --version
3.8.10.2 2015-05-20 18:17:19 2ef4f3a5b1d1d0c4338f8243d40a2452cc1f7fe4
```

Formula/libchewing.rb
DomT4: 👍

Formula/libchewing.rb
Chocobo1: lol, then it's safe to remove it.

Chocobo1:
Updated.

MikeMcQuaid:

PR #1660, opened 2016-06-02
cmake: linux fix libcurl tls

heinemml:

---

Fixes the following problem:
On Linux cmake is build with --no-system-curl instead of
--system-curl due to an unclear if statement
this breaks https support


DomT4:
Since this is Linux-specific it'll need to live in https://github.com/Linuxbrew/homebrew-core. Please re-file this PR there, Thanks!

heinemml:
Ah, sorry. I wasn't aware there has been a split of OS versions. Will check there. Thx.

DomT4:
No worries at all 😃. For core formulae that split has always existed. `homebrew/dupes`, `homebrew/science` and to a much lesser extent `homebrew/versions` will accept some `OS` abstraction but we don't in core formulae. 

Eventually `homebrew/brew`, which is the codebase everything `brew` runs off of will eventually be fully cross-platform but due to the amount of different handling required across Linux and OS X it's likely the core formulae will always be separate entities.

PR #1671, opened 2016-06-02
sipp 3.5.1 - 3.4.1 is 2 years old and missing options, namely -screen_file

phatjmo:
The existing formula for sipp is version 3.4.1 which is 2 years old and is missing options that I need for my test scripts. I'm trying to setup a simple test suite for our QA department and rather than have each compile the latest version, I prefer them using homebrew. I've run the entire gamut of conditions under https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md and even added a "test do" that wasn't there before. It's checking version output, but that's really the only test I can come up with for this utility since a proper test requires an active SIP endpoint.

### DIFF
--- a/Formula/sipp.rb
+++ b/Formula/sipp.rb
@@ -23,9 +23,6 @@ class Sipp < Formula
   end
 
   test do
-    # sipp_output = `#{bin}/sipp -v`.split(/$/)
-    # sipp_output.each { |o| print o }
-    # assert_match "SIPp v#{version}", sipp_output[0]
-    assert_match "SIPp v#{version}", shell_output("#{bin}/sipp -v 2>&1 || true")
+    assert_match "SIPp v#{version}", shell_output("#{bin}/sipp -v", 99)
   end
 end
--- a/Formula/sipp.rb
+++ b/Formula/sipp.rb
@@ -23,9 +23,9 @@ class Sipp < Formula
   end
 
   test do
-    sipp_output = `#{bin}/sipp -v`.split(/$/)
-    sipp_output.each { |o| print o }
-    assert_match "SIPp v#{version}", sipp_output[0]
-    # assert_match "SIPp", shell_output("#{bin}/sipp -v 2>&1")
+    # sipp_output = `#{bin}/sipp -v`.split(/$/)
+    # sipp_output.each { |o| print o }
+    # assert_match "SIPp v#{version}", sipp_output[0]
+    assert_match "SIPp v#{version}", shell_output("#{bin}/sipp -v 2>&1 || true")
   end
 end
--- a/Formula/sipp.rb
+++ b/Formula/sipp.rb
@@ -25,6 +25,7 @@ class Sipp < Formula
   test do
     sipp_output = `#{bin}/sipp -v`.split(/$/)
     sipp_output.each { |o| print o }
-    assert_match "SIPp v3.5.1", sipp_output[0]
+    assert_match "SIPp v#{version}", sipp_output[0]
+    # assert_match "SIPp", shell_output("#{bin}/sipp -v 2>&1")
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/sipp.rb
```diff
@@ -22,4 +21,8 @@ def install
     system "make", "DESTDIR=#{prefix}"
     bin.install "sipp"
   end
+
+  test do
+    assert_equal "SIPp", shell_output("#{bin}/sipp -v | grep SIPp | awk '{ print $1 }'").strip
```
MikeMcQuaid: Instead just do `assert_match "SIPp", shell_output("#{bin}/sipp -v")`

phatjmo: Unfortunately, the test fails with that. The "sipp -v" output is a multi-line block of text with the first line being blank and shell_output doesn't return it properly. 

So I had to improvise with:

  test do
    sipp_output = `#{bin}/sipp -v`.split(/$/)
    sipp_output.each { |o| print o }
    assert_match "SIPp v3.5.1", sipp_output[0]
  end

I've commited these changes for review.

DomT4: It's possible Mike's suggestion with a `2>&1` tacked onto the end after `-v` will do the trick, and is quite a bit cleaner.

MikeMcQuaid: Can you show the failure? `assert_match "SIPp v#{version}", shell_output("#{bin}/sipp -v")` should definitely match the same as your version above.

phatjmo: @MikeMcQuaid using your original recommendation:
```
Error: sipp: failed
<0> expected but was
<99>.
```

@DomT4 after appending `2>&1`:
```
Error: sipp: failed
<0> expected but was
<99>.
```

For reference, this is what `sipp -v` spits out:
```

 SIPp v3.5.1-TLS-PCAP-RTPSTREAM built Jun  2 2016, 12:31:50.

 This program is free software; you can redistribute it and/or
 modify it under the terms of the GNU General Public License as
 published by the Free Software Foundation; either version 2 of
 the License, or (at your option) any later version.

 This program is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.

 You should have received a copy of the GNU General Public
 License along with this program; if not, write to the
 Free Software Foundation, Inc.,
 59 Temple Place, Suite 330, Boston, MA  02111-1307 USA

 Author: see source files.

```
I was not successful in parsing out the lines of the version output using shell_output so the only code that I could make work without failing was the following (using @MikeMcQuaid's version variable recommendation):
```
    sipp_output = `#{bin}/sipp -v`.split(/$/)
    sipp_output.each { |o| print o }
    assert_match "SIPp v#{version}", sipp_output[0]
```
This prints out the `sipp -v` output and passes the test.

phatjmo: Ahh, it's because sipp -v exits with a return code 99 instead of 0. Wasn't paying attention to the detail of the output.

phatjmo: Confirmed: `assert_match "SIPp v#{version}", shell_output("#{bin}/sipp -v 2>&1 || true")` passes the test.

MikeMcQuaid: Pass `99` as the second argument to `shell_output` instead.

phatjmo: That works much better. Test passed, audit passed, commited and pushed.

MikeMcQuaid:

For future reference the preferred commit message format for simple version updates is `sipp 3.5.1`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md). Thanks!

PR #1680, opened 2016-06-03
traildb 0.5 (new formula)

rasky:

---

### DIFF
--- a/Formula/traildb.rb
+++ b/Formula/traildb.rb
@@ -30,6 +30,24 @@ class Traildb < Formula
   end
 
   test do
+    # Check that the library has been installed correctly
+    (testpath/"test.c").write <<-EOS.undent
+      #include <traildb.h>
+      #include <assert.h>
+      int main() {
+        const char *path = "test.tdb";
+        const char *fields[] = {};
+        tdb_cons* c1 = tdb_cons_init();
+        assert(tdb_cons_open(c1, path, fields, 0) == 0);
+        assert(tdb_cons_finalize(c1) == 0);
+        tdb* t1 = tdb_init();
+        assert(tdb_open(t1, path) == 0);
+      }
+    EOS
+    system ENV.cc, "test.c", "-L#{lib}", "-ltraildb", "-o", "test"
+    system "./test"
+
+    # Check that the provided tdb binary works correctly
     (testpath/"in.csv").write("1234 1234\n")
     system "#{bin}/tdb", "make", "-c", "-i", "in.csv", "--tdb-format", "pkg"
   end
--- a/Formula/traildb.rb
+++ b/Formula/traildb.rb
@@ -20,6 +20,7 @@ class Traildb < Formula
     resource("judy").stage do
       system "./configure", "--disable-debug", "--disable-dependency-tracking",
           "--disable-shared", "--prefix=#{judyprefix}"
+      # Build with -j1 because parallel builds are broken
       system "make", "-j1", "install"
     end
 
--- a/Formula/traildb.rb
+++ b/Formula/traildb.rb
@@ -7,32 +7,25 @@ class Traildb < Formula
   depends_on "libarchive"
   depends_on "pkg-config" => :build
 
-  # only required because we run autoreconf for judy
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
-  depends_on "libtool" => :build
-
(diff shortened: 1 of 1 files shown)


Formula/traildb.rb
```diff
@@ -0,0 +1,21 @@
+
```
MikeMcQuaid: Can remove this empty line.

Formula/traildb.rb
```diff
@@ -0,0 +1,21 @@
...
+class Traildb < Formula
+  desc "Blazingly-fast database for log-structured data"
+  homepage "http://traildb.io"
+  url "https://github.com/traildb/traildb/archive/0.5.tar.gz"
+  sha256 "4d1b61cc7068ec3313fe6322fc366a996c9d357dd3edf667dd33f0ab2c103271"
+
+  depends_on "libarchive"
+  depends_on "homebrew/boneyard/judy"
```
MikeMcQuaid: Can't depend on things in the boneyard; they are there because they are broken in some form.

rasky: This was just moved to boneyard, see https://github.com/Homebrew/homebrew-core/issues/1562. The reason for the move is that it's unmaintained. But it's a stable smallish C library that implements a single data structure. It's unmaintained because it's "done". What should we do then? Move it back from boneyard?

MikeMcQuaid: If it's got known bugs but isn't getting a new release: it should live in the boneyard and we cannot package this software if it depends on it.

rasky: None of the known bugs trigger on OS X though. They might never trigger. Can we at least agree on a solution until those bugs trigger, if ever? 

For instance: what if I publish a judy fork (possibly patched for the Linux-only-for-now bug) in the traildb organization? Can I pull it and install it as a dependency of Traildb from within the Traildb formula? That's closer to having a vendored copy.

MikeMcQuaid: You could install it as a `resource` in this formula, that's probably the best fit I think. Will let other @Homebrew/maintainers express views here too.

UniqMartin: If `judy` is essentially done and dead by now, then vendoring it into this formula sounds like the most practical solution to me. (Of course having a maintained `judy` project would be even nicer. While the data structure and algorithms may be done, software never is as it always has to adapt to changes in compilers and environments, however small those changes might be if something is implemented in a safe subset of a pretty stable and standardized language.)

DomT4: 👍 on resource. If upstream want to fork `judy` and become the official maintainers of the new package & it gains some usage traction I'd be happy for us to switch to that eventually, but I don't want to reintroduce an unsupported formula only for it to become unsupported again if a new issue crops up.

rasky:
Thanks for the comments, I've pushed an update to the formula that install judy as a `resource`. I also compile it statically, so that there is no dylib files that gets installed into the user prefix.

rasky:
Sorry, hold on a second on this, I'm double-checking whether building judy statically is a good idea wrt traildb language bindings and build systems...

rasky:
Ok, false warning, everything looks fine. Let me know if there are other issues with the PR.

Formula/traildb.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def install
+    # We build judy as static library, so we don't need to install it
+    # into the real prefix
+    judyprefix = "#{buildpath}/resources/judy"
+
+    resource("judy").stage do
+      system "./configure", "--disable-debug", "--disable-dependency-tracking",
+          "--disable-shared", "--prefix=#{judyprefix}"
+      system "make", "-j1", "install"
```
DomT4: Why the hardcoded `-j1` here?

rasky: Inherited from the Judy formula in boneyard. It looks like the build system fails with parallel builds; might be a missing dependency in the upstream makefile.

MikeMcQuaid: Could you make an inline comment about that. That'd be a good candidate for the type of things we want fixed if you decided to maintain an upstream fork 👍

rasky: Done.

MikeMcQuaid:
Thanks for your contribution to Homebrew and patience here! Without people like you submitting PRs we couldn't run this project. You rock!

PR #1693, opened 2016-06-03
elasticsearch: add devel and fix head

jasontedor:
This commit adds a devel install for Elasticsearch, and fixes the head
install for Elasticsearch to reflect changes that have occurred in
master (e.g., changing the name of the plugin script to
"elasticsearch-plugin").

### DIFF
--- a/Formula/elasticsearch.rb
+++ b/Formula/elasticsearch.rb
@@ -4,6 +4,12 @@ class Elasticsearch < Formula
   url "https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.3.3/elasticsearch-2.3.3.tar.gz"
   sha256 "5fe0a6887432bb8a8d3de2e79c9b81c83cfa241e6440f0f0379a686657789165"
 
+  devel do
+    url "https://download.elastic.co/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/5.0.0-alpha3/elasticsearch-5.0.0-alpha3.tar.gz"
+    sha256 "8fc606d56bdec83dc5219b5ad2346f2fe7c95feec9ad1d92c2bc1268323885e5"
+    version "5.0.0-alpha3"
+  end
+
   head do
     url "https://github.com/elasticsearch/elasticsearch.git"
     depends_on :java => "1.8"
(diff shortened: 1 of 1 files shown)


Formula/elasticsearch.rb
```diff
@@ -21,10 +27,11 @@ def cluster_name
   def install
     if build.head?
       # Build the package from source
-      system "gradle", "clean", "assemble"
-      # Extract the package to the current directory
-      targz = Dir["distribution/tar/build/distributions/elasticsearch-*.tar.gz"].first
-      system "tar", "--strip-components=1", "-xf", targz
+      system "gradle", "clean", ":distribution:tar:assemble"
+      # Extract the package to the tar directory
+      mkdir "tar"
```
DomT4: Could use this style of syntax:
```ruby
if build.head?
  system "gradle", "clean", ":distribution:tar:assemble"
  mkdir "tar" do
    system "tar", "--strip-components=1", "-xf", Dir["../distribution/tar/build/distributions/elasticsearch-*.tar.gz"].first
  end
end
```

Formula/elasticsearch.rb
```diff
@@ -113,11 +128,18 @@ def plist; <<-EOS.undent
   end
 
   test do
-    system "#{libexec}/bin/plugin", "list"
+    if FileTest.exist?("#{libexec}/bin/plugin")
```
DomT4: You can use `devel` and `head` as the modifier down here as well if you want to, they just become `if head?` and `if devel?` in `test do` rather than the `if build.head?` style equivalents.

Formula/elasticsearch.rb
jasontedor: Maybe I'm misunderstanding, but I'm not sure about this? We need to remain in the `tar` subdirectory for the later statements that install from `bin`, `config`, `lib`, and `modules`.

Formula/elasticsearch.rb
jasontedor: Thanks @DomT4, that's great. I pushed 94b0efeb066694172b1054dd592e8e520b325a37.

MikeMcQuaid:
👍  once @DomT4's happy.

Formula/elasticsearch.rb
DomT4: Apologies, the compressed diff made that difficult to spot. It's alright as-is then.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/5f2ed9214d2e09dffcb74c1e29b3bab7ffc1afbe. Thanks again @jasontedor!

jasontedor:
@DomT4 Thanks again! :smile:

Formula/elasticsearch.rb
jasontedor: No worries, I've been tricked by diffs before too. :smile:

Bobcui001:
How can i solve this problem?
When i execute ./bin/elasticsearch-plugin install elasticsearch-head
It wil show this
A tool for managing installed elasticsearch plugins

Commands
--------
list - Lists installed elasticsearch plugins
install - Install a plugin
remove - Removes a plugin from elasticsearch

Non-option arguments:
command

Option         Description
------         -----------
-h, --help     show help
-s, --silent   show minimal output
-v, --verbose  show verbose output
ERROR: Unknown plugin elasticsearch-head
Why ? 
I install ES(5.0.2) by the tool mac brew

jasontedor:
This pull request has nothing to do with elasticsearch-head. If you require assistance using Elasticsearch and installing plugins, I recommend checking out the [Elastic forum](https://discuss.elastic.co).

PR #1703, opened 2016-06-04
vasm 1.7e (new formula)

qu1j0t3:
- [Y] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

### DIFF
--- a/Formula/vasm.rb
+++ b/Formula/vasm.rb
@@ -36,7 +36,7 @@ class Vasm < Formula
     cpu_options.each do |cpu|
       syntax_options.each do |syntax|
         prog = "vasm#{cpu}_#{syntax}"
-        system "make", prog, "CPU="+cpu, "SYNTAX="+syntax
+        system "make", prog, "CPU=#{cpu}", "SYNTAX=#{syntax}"
         bin.install prog
       end
     end
--- a/Formula/vasm.rb
+++ b/Formula/vasm.rb
@@ -10,7 +10,7 @@ class Vasm < Formula
   option "with-arm", "Enable ARM CPU target (binary vasmarm_SYNTAX)"
   option "with-c16x", "Enable c16x/st10 microcontroller target (binary vasmc16x_SYNTAX)"
   option "with-jagrisc", "Enable Atari Jaguar GPU/DSP RISC target (binary vasmjagrisc_SYNTAX)"
-  option "with-m68k", "Enable Motorola 68K CPU target (binary vasmm68k_SYNTAX)"
+  option "with-m68k", "Enable Motorola 68K CPU target (binary vasmm68k_SYNTAX) (default)"
   option "with-ppc", "Enable PowerPC CPU target (binary vasmppc_SYNTAX)"
   option "with-tr3200", "Enable Trillek TR3200 CPU target (binary vasmtr3200_SYNTAX)"
   option "with-vidcore", "Enable VideoCore IV CPU target (binary vasmvidcore_SYNTAX)"
@@ -23,7 +23,8 @@ class Vasm < Formula
   option "with-oldstyle", "Enable oldstyle (8-bit) syntax (binary vasmCPU_oldstyle)"
 
   def cpu_options
-    %w[6502 6800 arm c16x jagrisc m68k ppc tr3200 vidcore x86 z80].select { |c| build.with? c }
+    opts = %w[6502 6800 arm c16x jagrisc m68k ppc tr3200 vidcore x86 z80].select { |c| build.with? c }
+    opts.empty? ? ["m68k"] : opts
   end
 
   def syntax_options
@@ -32,20 +33,21 @@ class Vasm < Formula
   end
 
   def install
-    configs = 0
     cpu_options.each do |cpu|
       syntax_options.each do |syntax|
         prog = "vasm#{cpu}_#{syntax}"
         system "make", prog, "CPU="+cpu, "SYNTAX="+syntax
         bin.install prog
-        configs += 1
       end
     end
 
-    odie "Please specify at least one cpu with --with-<cpu>" unless configs > 0
-
     system "make", "vobjdump"
     bin.install "vobjdump"
+  end
+
+  def caveats
+    opts = %w[6502 6800 arm c16x jagrisc m68k ppc tr3200 vidcore x86 z80].select { |c| build.with? c }
+    "No target CPU was specified with --with-<cpu>, so #{cpu_options} has been chosen by default." if opts.empty?
   end
 
   test do
--- a/Formula/vasm.rb
+++ b/Formula/vasm.rb
@@ -19,15 +19,24 @@ class Vasm < Formula
   option "with-z80", "Enable Zilog Z80 CPU target (binary vasmz80_SYNTAX)"
 
   # syntax options
-  option "with-std", "Enable standard syntax (binary vasmCPU_std)"
+  option "with-std", "Enable standard syntax (binary vasmCPU_std) (default)"
   option "with-madmac", "Enable MadMac (Atari) syntax (binary vasmCPU_madmac)"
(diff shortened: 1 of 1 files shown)


Formula/vasm.rb
```diff
@@ -0,0 +1,60 @@
+class Vasm < Formula
+  desc "Portable and retargetable assembler"
+  homepage "http://sun.hasenbraten.de/vasm/"
+  url "http://server.owl.de/~frank/tags/vasm1_7e.tar.gz"
+  version "1.7e"
+  sha256 "2878c9c62bd7b33379111a66649f6de7f9267568946c097ffb7c08f0acd0df92"
+
+  # target options
+  option "with-6502", "Enable 6502 CPU target (binary vasm6502_SYNTAX)"
```
DomT4: We're not going to want to carry around all of these options with no proven demand. Options = significantly higher maintenance burden, sadly. Can we narrow this list down a bit?

Formula/vasm.rb
```diff
@@ -0,0 +1,60 @@
...
+    %w[6502 6800 arm c16x jagrisc m68k ppc tr3200 vidcore x86 z80].select { |c| build.with? c }.each do |cpu|
+      %w[std madmac mot oldstyle].select { |s| build.with? s }.each do |syntax|
+        prog = "vasm#{cpu}_#{syntax}"
+        system "make", prog, "CPU="+cpu, "SYNTAX="+syntax
+        bin.install prog
+        configs += 1
+      end
+    end
+
+    odie "Please specify at least one cpu with --with-<cpu> and at least one syntax with --with-<syntax>" unless configs > 0
```
DomT4: Typically in these situations we pick some on the behalf of users that are "sane defaults".

qu1j0t3: Because of the nature of the project (a cross-assembler), it is not really possible to pick defaults. Each choice of CPU is a distinct assembler.

An alternative to options would be to produce N x M formulae, which doesn't seem an improvement.

MikeMcQuaid: We need to pick a default for the CPU or this it not a useful formula.

Formula/vasm.rb
```diff
@@ -0,0 +1,60 @@
+class Vasm < Formula
+  desc "Portable and retargetable assembler"
+  homepage "http://sun.hasenbraten.de/vasm/"
+  url "http://server.owl.de/~frank/tags/vasm1_7e.tar.gz"
+  version "1.7e"
+  sha256 "2878c9c62bd7b33379111a66649f6de7f9267568946c097ffb7c08f0acd0df92"
+
+  # target options
```
MikeMcQuaid: Don't need this comment.

Formula/vasm.rb
```diff
@@ -0,0 +1,60 @@
...
+        system "make", prog, "CPU="+cpu, "SYNTAX="+syntax
+        bin.install prog
+        configs += 1
+      end
+    end
+
+    odie "Please specify at least one cpu with --with-<cpu> and at least one syntax with --with-<syntax>" unless configs > 0
+
+    system "make", "vobjdump"
+    bin.install "vobjdump"
```
MikeMcQuaid: No `make install`?

toby5box: Indeed, there is no `make install`.

qu1j0t3: Indeed, there is no `make install`.

qu1j0t3:
Comments above have been addressed.

UniqMartin:
I feel like offering all those options and forcing users that need something that deviates from the default to build from source is not the best thing we can do. Instead, we could drop all options and simply build all combinations of CPUs and syntaxes. Granted, this takes a bit longer (100 instead of 5 seconds) and requires a bit more disk space (7 instead of 0.5 megabytes), but these numbers are still very low and in my opinion justify enabling all combinations by default. Objections? Opinions?

Formula/vasm.rb
```diff
@@ -0,0 +1,69 @@
...
+  def syntax_options
+    opts = %w[std madmac mot oldstyle].select { |s| build.with? s }
+    opts.empty? ? ["std"] : opts
+  end
+
+  def install
+    cpu_options.each do |cpu|
+      syntax_options.each do |syntax|
+        prog = "vasm#{cpu}_#{syntax}"
+        system "make", prog, "CPU="+cpu, "SYNTAX="+syntax
```
UniqMartin: Please use string interpolation here, i.e. `system "make", prog, "CPU=#{cpu}", "SYNTAX=#{syntax}"`

toby5box: OK

qu1j0t3: The reason I don't think we should build all combinations is that nearly all users will only want one CPU/syntax combination. I don't think it's an undue burden to ask them to pick it with options; they would have to do the same thing if building from source (it's mandatory to pick a combination).

I have however introduced defaults as suggested above.

qu1j0t3:
@UniqMartin String interpolation used. Also rebased.

Formula/vasm.rb
UniqMartin: *It's not helpful that you commented here and this is now hidden by default because you addressed the unrelated string interpolation issue. Can you please re-post this as a regular PR comment?*

toby5box:
(reposting in main thread, in reply to @UniqMartin) The reason I don't think we should build all combinations is that nearly all users will only want one CPU/syntax combination. I don't think it's an undue burden to ask them to pick it with options; they would have to do the same thing if building from source (it's mandatory to pick a combination). Or, in other words, building with all combinations doesn't really make the package more useful. Compare to say a C compiler; you normally only want one target (e.g. the developer's machine), not all the targets of which the compiler is capable. An assembler user is generally targeting one piece of physical hardware.

I have however introduced defaults as suggested above.

qu1j0t3:
(reposting in main thread, in reply to @UniqMartin) The reason I don't think we should build all combinations is that nearly all users will only want one CPU/syntax combination. I don't think it's an undue burden to ask them to pick it with options; they would have to do the same thing if building from source (it's mandatory to pick a combination). Or, in other words, building with all combinations doesn't really make the package more useful. Compare to say a C compiler; you normally only want one target (e.g. the developer's machine), not all the targets of which the compiler is capable. An assembler user is generally targeting one piece of physical hardware.

I have however introduced defaults as suggested above.

UniqMartin:
> The reason I don't think we should build all combinations is that nearly all users will only want one CPU/syntax combination. I don't think it's an undue burden to ask them to pick it with options; […]

It's not an undue burden, but I think it will make the formula immediately useful for more people without forcing them to make a choice and forcing them to do a build from source. I believe this convenience trumps the slight overhead caused by offering all CPU/syntax combinations.

A fully-featured C/C++ compiler is probably a bit more involved, making compile time significantly longer if all supported targets/backends were to be compiled. I don't think this is a good comparison. And just think about the toolchain provided by Apple. Even though most (Homebrew) users only care about x86_64, they get compilers capable of producing i386 as well as various flavors of ARM binaries.

But let's wait for the other maintainers that have already commented here to chime in and see how they feel about this. (I'm fine with accepting the single-architecture default if they disagree with me.)

DomT4:
Agree with Martin that I'd at least like _some_ of these options to be in by default. In by default also means they'll be CI tested, less likely to break for users and less likely consequently to cause us trouble. I don't think we need to turn them all on by default, but I would really like to slim down that list of `option`s.

qu1j0t3:
I don't think that makes sense for the type of tool this is. Closing PR.

DomT4:
Alright. You're more than welcome to support it in your [own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) as desired. Where you assume the burden of maintaining the formula you can make any decision about options you wish to.

qu1j0t3:
I'm going to talk to upstream and I'll reopen if that conversation produces a compromise solution.

PR #1712, opened 2016-06-04
s6 2.3.0.0 (new formula)

remko:

---


Formula/execline.rb
```diff
@@ -17,7 +17,6 @@ class Execline < Formula
   def install
     system "./configure",
       "--prefix=#{prefix}",
-      "--disable-static",
```
MikeMcQuaid: Can you explain this change a bit more?

Formula/s6.rb
```diff
@@ -0,0 +1,39 @@
...
+      "--prefix=#{prefix}",
+      "--bindir=#{libexec}/bin",
+      "--sbindir=#{libexec}/sbin",
+      "--with-sysdeps=#{Formula["skalibs"].opt_prefix}/lib/skalibs/sysdeps",
+      "--with-lib=#{Formula["skalibs"].opt_prefix}/lib",
+      "--with-lib=#{Formula["execline"].opt_prefix}/lib/execline",
+      "--disable-shared"
+    system "make", "install"
+
+    # Some S6 tools expect execline binaries to be on the path
```
MikeMcQuaid: Are they not already in `bin`? If not: why not?

Formula/s6.rb
```diff
@@ -0,0 +1,39 @@
...
+      (libexec/bindir).entries.select(&:file?).each do |entry|
+        (bin/entry.basename).write_env_script libexec/bindir/entry.basename, :PATH => path
+      end
+    end
+  end
+
+  test do
+    mkdir(testpath/"log")
+    IO.popen(["#{bin}/s6-log", testpath/"log"], "r+") do |pipe|
+      pipe.puts("This is a test")
```
MikeMcQuaid: Can use `pipe_output` for this.

remko: Thanks, I'll change this.

Formula/s6.rb
remko: Unlike other shells, execline doesn't have any real builtins; instead, it comes with external commands. 

- Many of these are original (`multidefine`, `forbacktickx`), no problem there
- Others (such as `cd`, `exec`, `umask`) overlap with shell builtins, but have different semantics  than what you would expect (since execline relies on Bernstein chaining); this is actually also fine, since they are caught by shell builtins, so won't be picked from the path.
- One command is problematic: `import`; this one conflicts with ImageMagick.

So I had the choice of putting everything except `import` in PATH, or only `execline`, and make `execline` set the path before executing its script. Since these commands are probably mostly only used in the context of `execline` scripts, I took the second option. However, `s6` is an exception, as it also uses some execline binaries here and there, so I had to also set the path there.

I'm open to taking the other path, but it would mean `import` is either not available to execline scripts (which is a bit annoying).

Formula/execline.rb
remko: The execline build also generates a static library; `--disable-static` disables building this library.
I don't know why I explicitly asked it to not build when submitting the formula, but it's a prerequisite for `s6`.

Formula/s6.rb
MikeMcQuaid: Ok, that makes sense. `bin.env_script_all_files` is probably your best bet here; see `ansible.rb` as an example for usage.

Formula/execline.rb
MikeMcQuaid: @remko It sounds like this should be told to build shared libraries somehow given that you need a runtime dependency on it anyway.

remko: It does. However, the upstream of both `execline` and `skalibs` don't support shared libraries on OS X (and explicitly don't want to, cfr their README and questions on the mailing list). That said, the static libs are built so they don't incur much waste of space, so the net result shouldn't be that different in size aspect. 

It could probably be patched to build and use shared libs on OS X, but it would never be integrated upstream, so it would need to stay maintained by homebrew, which isn't worth the effort in this case i think.

Formula/s6.rb
```diff
@@ -0,0 +1,30 @@
...
+  sha256 "49c9a0476d93c6279c776ca5ae9c49b3fd39c356eda674828aebf78f8488e8e6"
+  head "git://git.skarnet.org/s6"
+
+  depends_on "skalibs"
+  depends_on "execline"
+
+  def install
+    system "./configure",
+      "--prefix=#{prefix}",
+      "--with-sysdeps=#{Formula["skalibs"].opt_prefix}/lib/skalibs/sysdeps",
```
DomT4: `#{Formula["skalibs"].opt_lib}/skalibs/sysdeps`

Formula/s6.rb
```diff
@@ -0,0 +1,30 @@
...
+      "--with-lib=#{Formula["skalibs"].opt_prefix}/lib",
```
DomT4: `#{Formula["skalibs"].opt_lib}`

Formula/s6.rb
```diff
@@ -0,0 +1,30 @@
...
+      "--with-lib=#{Formula["execline"].opt_prefix}/lib/execline",
```
DomT4: `#{Formula["execline"].opt_lib}/execline`

Formula/s6.rb
DomT4: I appreciate you likely just mimicked the lines from `execline` but they're wrong there too, sadly 😄.

Formula/s6.rb
```diff
@@ -0,0 +1,30 @@
...
+      "--disable-shared"
+    system "make", "install"
+
+    # Some S6 tools expect execline binaries to be on the path
+    bin.env_script_all_files(libexec/"bin", :PATH => "#{Formula["execline"].libexec}:$PATH")
```
DomT4: `"#{Formula["execline"].opt_libexec}:$PATH"`

Formula/s6.rb
```diff
@@ -0,0 +1,30 @@
...
+    sbin.env_script_all_files(libexec/"sbin", :PATH => "#{Formula["execline"].libexec}:$PATH")
```
DomT4: Same notation as above.

Formula/s6.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  test do
```
DomT4:
```ruby
test do
  (testpath/"log").mkpath
  shell_output("#{bin}/s6-log #{testpath}/log")
  assert File.exist?(testpath/"log/current")
end
```
Should work, I'd imagine?

remko: Thanks, the return status check was indeed redundant; it has to be `pipe_output`, though, the command takes input until the pipe closes.

Formula/execline.rb
MikeMcQuaid: Well, the big problem is that this means that to use the latest `skalibs` the `execline` formula needs to be rebuilt every time. I think what would make more sense in this case is we remove these libraries from Homebrew and use `resource`s on the software that needs them (which seems to be just `s6` at this point, possibly ever).

MikeMcQuaid: So: you can delete the existing formulae and install the non-static parts (e.g. binaries) into `libexec` of the `s6` cellar.

remko: Ok for `skalibs`, but the thing is that `execline` is useful on its own. What would this mean: i remove the `skalibs` formula, but keep the static linking of s6 against `execline`?

remko: Btw, there's always the alternative for patching the build to use dynamic libraries, with the mentioned caveat that the patch won't be accepted upstream.

MikeMcQuaid: If upstream won't accept it it's a non-starter for us, unfortunately. I think we can leave `execline` as is but I'd rather we leave static linking disabled. Although it may be useful on its own I wonder if it's actually going to see widespread enough usage to leave it around.

remko: Well, leaving the static linking disabled for `execline` is not really possible, as `s6` depends on it :(

Since `s6` depends on `skalib`+`execline` lib+`execline` binaries, and `execline` in turn depends on `skalib`, the only options I still see at the moment are:

- Leave it as it is: static `skalibs` library, static `execline` library, `s6` binary statically linked against both. You can't upgrade an upstream package without upgrading the downstream packages and reinstalling.
- Package `skalibs` with `execline`, and make `s6` depend and (statically) link against libraries from the `execline` package. Same problem as above, but one package less in the game; however, a weird split.
- Build everything in one package, and don't install any libraries. Not sure how you build different tarballs in one package in Homebrew.

MikeMcQuaid: > Build everything in one package, and don't install any libraries. Not sure how you build different tarballs in one package in Homebrew.

I'd go for this. You can do this with `resource`s (`grep` so see example usage)

Formula/s6.rb
```diff
@@ -0,0 +1,86 @@
...
+      url "git://git.skarnet.org/skalibs"
+    end
+
+    resource "execline" do
+      url "git://git.skarnet.org/execline"
+    end
+  end
+
+  def install
+    resources.each { |r| r.stage("#{buildpath}/#{r.name}") }
```
DomT4: Tiny tiny style nit, but this can probably be:

```ruby
resources.each { |r| r.stage(buildpath/r.name) }
```
?

Formula/s6.rb
```diff
@@ -0,0 +1,86 @@
...
+    build_dir = buildpath/"build"
+
+    chdir "skalibs" do
```
DomT4: Please use `cd` rather than `chdir` throughout, Thanks!

Formula/s6.rb
```diff
@@ -0,0 +1,86 @@
...
+      system "./configure",
+        "--prefix=#{build_dir}",
+        "--bindir=#{libexec}/execline",
+        "--with-include=#{build_dir}/include",
+        "--with-lib=#{build_dir}/lib",
+        "--with-sysdeps=#{build_dir}/lib/skalibs/sysdeps",
+        "--disable-shared"
+      system "make", "install"
+      (bin/"execlineb").write_env_script libexec/"execline/execlineb", :PATH => "#{libexec}/execline:$PATH"
+      (bin/"execlineb").chmod 0555
```
DomT4: This shouldn't be necessary?

Formula/s6.rb
```diff
@@ -0,0 +1,86 @@
...
+    test_script = testpath/"test.eb"
+    test_script.write <<-EOS.undent
+     import PATH
+     if { [ ! -z ${PATH} ] }
+       true
+    EOS
+    system "#{bin}/execlineb", test_script
+
+    # Test s6
+    mkdir(testpath/"log")
```
DomT4: Another tiny nit, but we mildly prefer `(testpath/"log").mkpath`

DomT4:
Few nits but we're nearly there at this point, IMO.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/d11bdff056b38f5e95b8ea9846e1698ccb296d4b. Thank you for your contribution to Homebrew @remko; we appreciate it! :smiley_cat:

PR #1716, opened 2016-06-04
abcde 2.7.2

ilovezfs:
Also,
- remove patch since change was merged upstream
- match against the version in the test
- etcdir is now sysconfdir in the Makefile


DomT4:
👍

PR #1717, opened 2016-06-05
pandoc 1.17.1

dunn:

---

cc @ilovezfs

### DIFF
--- a/Formula/pandoc.rb
+++ b/Formula/pandoc.rb
@@ -19,7 +19,6 @@ class Pandoc < Formula
 
   depends_on "ghc" => :build
   depends_on "cabal-install" => :build
-  depends_on "gmp"
 
   def install
     args = []
--- a/Formula/pandoc.rb
+++ b/Formula/pandoc.rb
@@ -5,8 +5,8 @@ class Pandoc < Formula
 
   desc "Swiss-army knife of markup format conversion"
   homepage "http://pandoc.org"
-  url "https://hackage.haskell.org/package/pandoc-1.17.0.3/pandoc-1.17.0.3.tar.gz"
-  sha256 "7b14e1bcb78a7e2ad1e585f127be7efd20225c9f9b5131d507b376b62cd77e32"
+  url "https://hackage.haskell.org/package/pandoc-1.17.1/pandoc-1.17.1.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
:+1:

Formula/pandoc.rb
```diff
@@ -22,11 +22,6 @@ class Pandoc < Formula
   depends_on "gmp"
```
rakkesh: Is this `dependency` really required here? Pandoc builds and works fine without `depends_on "gmp"`.

dunn: Looks like it's a runtime dep, according to https://github.com/jgm/pandoc/issues/2943#issuecomment-221423238

dunn: Also https://github.com/jgm/pandoc/issues/11#issuecomment-632751

rakkesh: Thanks for providing the reference. I haven't tried it on GNU/Linux but on Mac OS X I don't notice such an issue without installing `gmp`. 

```
$ pandoc -o test.html test.md      # works on all my machines
```

Also, `pandoc` doesn't link with `gmp` even with `depends_on "gmp"` in the formula.

```
$ otool -L `which pandoc`
/usr/local/bin/pandoc:
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

I have just now built the current version `1.17.1` without `depends_on "gmp"`. The generated binary works fine.

Here are the logs, https://gist.github.com/d0235638893c6074fa80e906589a0232

dunn: Alright, let's give it a shot.  If anything breaks we can revert.  Thanks for looking closer!

ilovezfs: I think this is a consequence of https://github.com/Homebrew/legacy-homebrew/pull/39134. I'm not sure why pandoc wasn't included in https://github.com/Homebrew/legacy-homebrew/pull/39709.

ilovezfs: @geoff-codes Perhaps you recall?

PR #1721, opened 2016-06-05
cpputest 3.8

ilovezfs:
- bump version
- add a test
- use CMake instead of Autotools for head

### DIFF
--- a/Formula/cpputest.rb
+++ b/Formula/cpputest.rb
@@ -1,8 +1,9 @@
 class Cpputest < Formula
   desc "C /C++ based unit xUnit test framework"
   homepage "http://www.cpputest.org/"
-  url "https://github.com/cpputest/cpputest/releases/download/3.7.2/cpputest-3.7.2.tar.gz"
-  sha256 "8c5d00be3a08ea580e51e5cfe26f05d05c6bf546206ff67dbb3757d48c109653"
+  url "https://github.com/cpputest/cpputest/releases/download/v3.8/cpputest-3.8.tar.gz"
+  sha256 "c81dccc5a1bfc7fc6511590c0a61def5f78e3fb19cb8e1f889d8d3395a476456"
+  head "https://github.com/cpputest/cpputest.git"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍 . Presumably CMake doesn't work for non-HEAD?

ilovezfs:
It probably does, actually, but since the release tarball comes with a configure script, I was thinking that would be an unnecessary extra dependency.

MikeMcQuaid:
@ilovezfs For formulae simplification let's just use CMake consistently 👍

MikeMcQuaid:
(for `cellar :any` stuff we don't care about build time dependencies, particularly stuff like CMake that's really widely used)

ilovezfs:
@MikeMcQuaid PR refreshed.

MikeMcQuaid:
👍

ilovezfs:
The build failure is mysterious to me. Not seeing it locally.

MikeMcQuaid:
@ilovezfs Looks sandbox related.

ilovezfs:
locally: https://gist.github.com/ilovezfs/7445fb292fbc378f553f9f0c2f8152c9

MikeMcQuaid:
Trying a rebuild just in case.

ilovezfs:
I think it may be caused by a case insensitivity bug since my file system on that machine is case sensitive. I'll dig deeper.

ilovezfs:
Yup, reproduced the bug with a case insensitive volume. lol

ilovezfs:
Well that "fixed" it.

Formula/cpputest.rb
```diff
@@ -12,16 +13,24 @@ class Cpputest < Formula
...
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
-  end
+  depends_on "cmake" => :build
 
   def install
-    system "./autogen.sh" if build.head?
-    system "./configure", "--prefix=#{prefix}"
-    system "make", "install"
+    cd "build" do
```
DomT4: `build` already exists?

ilovezfs: Yes, which is probably the root of the reason it otherwise fails on a case-insensitive volume, though I've not proven that yet.

DomT4: Strange. Oh well, if it works, feel free to ship it I guess 🙈.

PR #1722, opened 2016-06-05
aide 0.16b1 (devel)

ilovezfs:
Also,
- devel and head now depend on pcre
- add linked library dependency libgpg-error


DomT4:
Seems okay 👍.

PR #1724, opened 2016-06-05
prettyping: fix ipv6 support

daenney:

---

The BSD/OS X ping6 utility produces slightly different output than the
GNU utility does. This patch has been pending upstream for a while but
the maintainer has been very quiet in the past half year.

Since this results in a broken experience on OS X including the patch
here until it gets merged seems like a good way forward.

### DIFF
--- a/Formula/prettyping.rb
+++ b/Formula/prettyping.rb
@@ -6,6 +6,11 @@ class Prettyping < Formula
 
   bottle :unneeded
 
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/848211f/prettyping/ipv6.patch"
+    sha256 "263113acb0c2c99d0fefe979ac450cf00361cb0283df21564958cc4c38e98aad"
+  end
+
   def install
     bin.install "prettyping"
   end


Formula/prettyping.rb
```diff
@@ -6,6 +6,9 @@ class Prettyping < Formula
 
   bottle :unneeded
 
+  # Fix IPv6 handling on BSD/OSX: https://github.com/denilsonsa/prettyping/pull/11
+  patch :DATA
```
DomT4: Any chance the upstream patch applies cleanly?

```ruby
patch do
  url "https://github.com/denilsonsa/prettyping/pull/11.patch"
  sha256 ""
end
```
etc.

daenney: Afraid not, I tried that. There's been a few changes to master since the 1.0.0 release so the hunks all fail if you try that.

DomT4: I see, that's a shame. Would you mind submitting the patch here to https://github.com/Homebrew/formula-patches then and reworking it like [this](https://github.com/Homebrew/homebrew-core/blob/7f5799fd2178001b715827e3a84e021923d357c2/Formula/cdparanoia.rb#L27-L30)? We try to avoid inline patches because people find novel ways to break them.

daenney: Oh yes, that does seem much better. I had no idea that was a thing!

daenney:
@DomT4 Done. I generated the shasum with `-a 256 -p`, I believe that ought to do it.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/d13efa8bd5ec2a060f201bc084a570a34c3b74ca. Thank you for your contribution to Homebrew @daenney; we appreciate it! :smiley_cat:

daenney:
Thanks for helping getting the patch in!

PR #1726, opened 2016-06-05
aescrypt: Implement an effective test

crd:

---

This new test ensures that a message can be encrypted and decrypted
successfully.

### DIFF
--- a/Formula/aescrypt.rb
+++ b/Formula/aescrypt.rb
@@ -20,8 +20,8 @@ class Aescrypt < Formula
   test do
     (testpath/"key").write "kk=12345678901234567890123456789abc0"
     original_text = "hello"
-    cipher_text = pipe_output("echo #{originalText} | #{bin}/aescrypt -k #{testpath}/key -s 128").chomp
-    deciphered_text = pipe_output("echo #{originalText} | #{bin}/aescrypt -k #{testpath}/key -s 128 | #{bin}/aesget -k #{testpath}/key -s 128").chomp
+    cipher_text = pipe_output("echo #{original_text} | #{bin}/aescrypt -k #{testpath}/key -s 128").chomp
+    deciphered_text = pipe_output("echo #{original_text} | #{bin}/aescrypt -k #{testpath}/key -s 128 | #{bin}/aesget -k #{testpath}/key -s 128").chomp
     assert_not_equal original_text, cipher_text
     assert_equal original_text, deciphered_text
   end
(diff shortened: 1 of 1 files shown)


Formula/aescrypt.rb
```diff
@@ -19,13 +19,10 @@ def install
...
-
-    require "open3"
-    Open3.popen3("#{bin}/aescrypt", "-k", testpath/"key") do |stdin, stdout, _|
-      stdin.write("hello")
-      stdin.close
-      # we can't predict the output
-      stdout.read.length > 0
-    end
+    original_text = "hello"
+    cipher_text = pipe_output("echo #{original_text} | #{bin}/aescrypt -k #{testpath}/key -s 128").chomp
```
UniqMartin: `pipe_output` is here to avoid using `echo` and having to deal with possible problems due to quoting etc. For that you have to use it as `pipe_output(<command>, <input>)`, or to be more specific:

```ruby
cipher_text = pipe_output("#{bin}/aescrypt -k #{testpath}/key -s 128", original_text)
```

Is the `.chomp` at the end really needed? Note that this will produce a slightly different result as the string `hello` is passed without a newline that is automatically appended by `echo`.

Formula/aescrypt.rb
```diff
@@ -19,13 +19,10 @@ def install
...
+    deciphered_text = pipe_output("echo #{original_text} | #{bin}/aescrypt -k #{testpath}/key -s 128 | #{bin}/aesget -k #{testpath}/key -s 128").chomp
```
UniqMartin: Similar story here and I think it would be better to directly use the output of the previous line instead of running the unencrypted text again through `aescrypt`:

```ruby
deciphered_text = pipe_output("#{bin}/aesget -k #{testpath}/key -s 128", cipher_text)
```

Formula/aescrypt.rb
UniqMartin: Nope, just checked. The `.chomp` is really not needed. Neither here nor in the next line.

crd:
Thanks @UniqMartin -- I didn't know that pipe_output could take that second optional parameter. Should have RTFM. For my own future reference: [pipe_output-instance_method](http://www.rubydoc.info/github/Homebrew/brew/Homebrew/Assertions#pipe_output-instance_method)

UniqMartin:
Merged. Thank you for this contribution to Homebrew, @crd! :tada:

PR #1731, opened 2016-06-05
nutcracker 0.4.1

ilovezfs:
Also,
- use pkgshare
- match version in the test

### DIFF
--- a/Formula/nutcracker.rb
+++ b/Formula/nutcracker.rb
@@ -1,8 +1,8 @@
 class Nutcracker < Formula
   desc "Proxy for memcached and redis"
   homepage "https://github.com/twitter/twemproxy"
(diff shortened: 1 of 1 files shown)


DomT4:
Identical comment to https://github.com/Homebrew/homebrew-core/pull/1732#issuecomment-223862006 😄

ilovezfs:
I've removed the symlink because I think that's a bad idea.

ilovezfs:
@DomT4 :ship:?

DomT4:
👍. Nice spot that some audit checks weren't looking at the `+`-style syntax as well, by the way! Always the little things that creep through testing, however extensive.

ilovezfs:
Thanks. I was pretty surprised, actually.

PR #1732, opened 2016-06-05
sslh 1.18

ilovezfs:
Also,
- remove patch since issue was fixed upstream
- match version in the test

### DIFF
--- a/Formula/sslh.rb
+++ b/Formula/sslh.rb
@@ -1,17 +1,10 @@
 class Sslh < Formula
   desc "Forward connections based on first data packet sent by client"
   homepage "http://www.rutschle.net/tech/sslh.shtml"
+  url "http://www.rutschle.net/tech/sslh/sslh-v1.18.tar.gz"
+  sha256 "1601a5b377dcafc6b47d2fbb8d4d25cceb83053a4adcc5874d501a2d5a7745ad"
+  head "https://github.com/yrutschle/sslh.git"
 
-  stable do
-    url "http://www.rutschle.net/tech/sslh-v1.17.tar.gz"
-    sha256 "4f3589ed36d8a21581268d53055240eee5e5adf02894a2ca7a6c9022f24b582a"
-
-    # fixes `make install`, fixed in HEAD
-    patch do
-      url "https://github.com/yrutschle/sslh/commit/7c35ef8528d47b97894a6495275b57dc1ae3f8c7.diff"
-      sha256 "a6f8e1c3f9776d7344ea839de1d8a40e9925101528bd3beee50a0c60c62872cf"
-    end
-  end
   bottle do
     cellar :any
     sha256 "57377bc2f5df6479428b757741a61ab2d3aa1fc899772f732f602bd9d4be9dd8" => :el_capitan
@@ -20,16 +13,14 @@ class Sslh < Formula
     sha256 "2a712be56b116244717fb4e414846b6b9373bf2b0482b764b2142096cffbac18" => :mountain_lion
   end
 
-  head "https://github.com/yrutschle/sslh.git"
-
   depends_on "libconfig"
 
   def install
-    ENV.j1
+    ENV.deparallelize
     system "make", "install", "PREFIX=#{prefix}"
   end
 
   test do
-    system sbin/"sslh", "-V"
+    assert_match version.to_s, shell_output("#{sbin}/sslh -V")
   end
 end


DomT4:
> set BINDIR=#{bin} to avoid installing in sbin

I'm not massively in love with this until we actually decide some sort of policy on `sbin`, unless other package managers are doing the same thing here.

At the very least, if you're going to start attempting to transition things, to avoid breaking scripts ideally we should be leaving symlinks present in `sbin` for a while.

ilovezfs:
Symlink sounds fine to soften the blow as long as the real one is in `bin` :smirk:

ilovezfs:
However, it's obnoxious that then still triggers the warning.

ilovezfs:
I think I prefer it without the symlink unless we're going to shut the warning up when the target of the symlink is indeed in the PATH.

MikeMcQuaid:
> I'm not massively in love with this until we actually decide some sort of policy on sbin, unless other package managers are doing the same thing here.

I think it would be good to have a policy here but I don't think it should be dependent on other package managers almost all of which requiring running as root/support setuid binaries so `superuser binaries` makes sense for them.

ilovezfs:
@MikeMcQuaid @DomT4 This shuts it up: https://github.com/Homebrew/brew/pull/323

DomT4:
> but I don't think it should be dependent on other package managers almost all of which requiring running as root/support setuid binaries so superuser binaries makes sense for them.

Sorry, I evidently did a bad job explaining my initial comment. I'll expand it a bit to clarify.

> I'm not massively in love with this until we actually decide some sort of policy on sbin, unless other package managers are doing the same thing here.

I'm not massively in love with this until we actually decide some sort of policy on `sbin`, but I'll accept jumping that shark for this particular formula if it has become the norm in other package managers to install _this_ into `bin` rather than `sbin` despite the upstream "default".

ilovezfs:
This doesn't need to be so complicated. /usr/local/bin is a standard path that is in everyone's PATH by default. /usr/local/sbin is not. So unless there's a really good reason specific to a given formula that has been articulated by upstream or technically justifiable with evidence, we should use the standard path.

ilovezfs:
I've removed the symlink because I think that's a bad idea.

DomT4:
> we should use the standard path.

Except as far as upstreams are concerned, they expect their executable to live in `sbin` or they wouldn't be using it. Maybe you'd like to volunteer to "educate" upstream developers, but I think it's a bad idea to arbitrarily deviate from upstream documentation highlighting they expect their executable to live in `sbin`.

And I note, Homebrew already has form on being inconsistent with this particular formula. Adam V moved it from `sbin` to `bin` in 2013, and then Alex moved it back to `sbin` in 2015. That's the kind of thing that hacks off people and shows why we should adopt policy rather than piecemeal.

Formula/sslh.rb
```diff
@@ -20,16 +13,16 @@ class Sslh < Formula
...
-
   depends_on "libconfig"
 
   def install
-    ENV.j1
-    system "make", "install", "PREFIX=#{prefix}"
+    ENV.deparallelize
+    system "make", "install", "PREFIX=#{prefix}", "BINDIR=#{bin}"
+
+    prefix.install "README.MacOSX"
```
DomT4: FWIW, this file literally only tells people to install `sslh` with MacPorts and offers up an example `launchctl` plist. We can write the latter into the formula ourselves.

ilovezfs:
I consider this an integration issue for OS X when the chosen prefix is /usr/local. If we were on Linux and the prefix was / or /usr, I'd have no problem /sbin or /usr/sbin. We're not on Linux. The prefix is not /. The prefix is not /usr. Nor can it be now that SIP is in place. Therefore, unless Apple adds /usr/local/sbin to the default set of paths, I consider it a serious packaging bug to be putting stuff in non-standard locations, especially when the sole justification is an attribution of intentionality to upstream defaults, even when upstream deliberately makes them configurable.

Formula/sslh.rb
ilovezfs: Yes, I read it.

DomT4:
> I consider this an integration issue for OS X when the chosen prefix is /usr/local

So you'll be adding logic to handle the situation for all our users _not using_ `/usr/local` then, I presume? 😉. Bearing in mind Homebrew may very well not live in `/usr/local` forever it's going to look pretty silly if we move & a year down the line our justification suddenly dies out.

I don't think it makes sense to implement a system on a _per-formula basis_ where you can expect these executables to live in `sbin` on Linux, MacPorts, probably Fink, FreeBSD, and then you start using Homebrew and suddenly that executable lives in a different subdirectory.

ilovezfs:
No special logic is required for non-standard prefixes. They've already added foo_prefix/"bin" to their PATH. If they've also added foo_prefix/"sbin", then it's a matter of indifference for them.

If we move prefixes elsewhere, then it's a question of whether we're adding two paths or one. Not as interesting of an issue in that case, but one is still simpler than two.

As for what happens on Linux, FreeBSD, or MacPorts I think is irrelevant, and I doubt anyone would even notice the difference, but if they did, it be because Linux, et. al were using /sbin or /usr/sbin, not /usr/local/sbin, so it differs regardless.

MikeMcQuaid:
If we're to have any `bin` and `sbin` distinction the one that makes sense in Homebrew's context is having `sbin` those that always require root to run them e.g. `mtr`.

ilovezfs:
Let's just leave it the way it is, since it means less work for us, other than supporting users who get confused when something isn't in their default PATH.

MikeMcQuaid:
@ilovezfs We'll leave it as-is for this PR and I'll make a proposal at some point to detail the version in my comment if you're 👍 on it

DomT4:
> If we're to have any bin and sbin distinction the one that makes sense in Homebrew's context is having sbin those that always require root to run them e.g. mtr.

This would be fine with me, personally.

ilovezfs:
I think we should just leave it as upstream defaults unless there's some specific reason for a particular formula. There are plenty of system binaries that do not require root on OS X. (e.g., diskutil)

ilovezfs:
```
bash-3.2$ which diskutil
/usr/sbin/diskutil
bash-3.2$
```

ilovezfs:
As for things that always require root, yes upstreams will usually be targeting sbin for those if they're not already, and we think they should, we can always ask them in an individual case why they're not.

Of course, if I were they, and someone showed up questioning why I wasn't using sbin for this or that given that it always requires root, I would refer her to
```
$ which sudo
/usr/bin/sudo
```

MikeMcQuaid:
> /sbin is a standard subdirectory of the root directory in Linux and other Unix-like operating systems that contains executable (i.e., ready to run) programs. They are mostly administrative tools, that should be made available only to the root (i.e., administrative) user.

http://www.linfo.org/sbin.html

> `/sbin` ... for scripts with superuser (root) privileges required.

http://askubuntu.com/questions/308045/differences-between-bin-sbin-usr-bin-usr-sbin-usr-local-bin-usr-local/308048#308048

This seems to be most consistent with the system above.

MikeMcQuaid:
Also worth nothing `/etc/paths`:
```
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin
```

ilovezfs:
It's probably worth filing an upstream bug report about if we feel strongly that a particular utility should be targeting sbin not bin. I can't imagine there are many where that actually applies that won't already have an sbin default.

xu-cheng:
It worth noting that:
> /usr/sbin Non-essential system binaries, e.g., daemons for various network-services.
> https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

Also, I share the same opinion among other maintainers that this is a thing should be decided by upstream rather than us. Why to complicate formula, when previous a `./configure --prefix=<prefix>; make install` work just as upstream intended.

MikeMcQuaid:
@xu-cheng Do you think upstreams are considering the default PATH for the default Homebrew installation location on OS X? I don't think they are nor should they.

ilovezfs:
On OS X there is no consistency about what "system" utilities require root. Generally there's a lot more stuff that requires root in /sbin than in /usr/sbin, but there are plenty of utilities that sometimes require root in /bin and /usr/bin, though almost nothing that always requires root.

@xu-cheng only reason would be convenience due to the /etc/paths issue, which is definitely not on upstreams' minds.

If we ever move prefixes, the relevance of the default contents of /etc/paths goes out the window, but we could always add a file to /etc/paths.d/foo if we wanted our new default prefix/"bin" and prefix/"sbin" to be included. That is of course another option here. We could improve UX by having the brew installer do

```
mkdir -p /etc/paths.d
echo /usr/local/sbin > /etc/paths.d/homebrew
```

xu-cheng:
> Do you think upstreams are considering the default PATH for the default Homebrew installation location on OS X? I don't think they are nor should they.

In comparison, on a fresh Linux machine, you still need set this PATH.

```
[root@bot ~]# echo $PATH
/usr/local/bin:/usr/bin
```

Even more, since lots of binaries on sbin are for daemons. They are  never mean to be called by users directly but by `init`.

ilovezfs:
@xu-cheng That's 100% distro dependent:
```
if [ "$EUID" = "0" ] || [ "$USER" = "root" ] ; then
        PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:${ROOTPATH}"
else
        PATH="/usr/local/bin:/usr/bin:/bin:${PATH}"
fi
```

MikeMcQuaid:
@xu-cheng Generally on Linux distros you need to set it as a non-root user but not as a root user.

ilovezfs:
Note that in the example above it's EUID not UID:
```
openzfs@openzfs ~ $ bash -c 'echo $PATH'
/usr/local/bin:/usr/bin:/bin:/opt/bin:/usr/x86_64-pc-linux-gnu/gcc-bin/4.9.3:/usr/games/bin
openzfs@openzfs ~ $ sudo bash -c 'echo $PATH'
/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/opt/bin:/opt/Citrix/ICAClient:/opt/freerapid:/opt/blackdown-jdk-1.4.2.03/bin:/opt/blackdown-jdk-1.4.2.03/jre/bin:/opt/dmd/bin:/usr/lib/plan9/bin:/opt/nvidia-cg-toolkit/bin:/opt/tivoli/tsm/client/ba/bin:/opt/eagle/bin:/opt/vmware/bin:/opt/aws-as-tools/bin:/opt/aws-cw-tools/bin:/opt/aws-elb-tools/bin:/opt/aws-iam-tools/bin:/opt/aws-rds-tools/bin:/usr/bin/cdsclient:/opt/cuda/bin:/opt/ekopath/bin:/usr/ti-linux-gnu/tigcc-bin/4.1.2:/usr/ti-linux-gnu/bin
```

ilovezfs:
```
openzfs@openzfs ~ $ which zfs
which: no zfs in (/usr/local/bin:/usr/bin:/bin:/opt/bin:/usr/x86_64-pc-linux-gnu/gcc-bin/4.9.3:/usr/games/bin)
openzfs@openzfs ~ $ sudo which zfs
/sbin/zfs
openzfs@openzfs ~ $
```

ilovezfs:
@DomT4 :shipit: ?

DomT4:
Aye 👍. We've got at least a few `sbin` conversations ongoing, it'd be nice to centralise them somewhere but a formula PR isn't the ideal place.

ilovezfs:
@DomT4 I think we should just treat the sbin cases where we "disagree" with upstream's choice for whatever reason (a game in sbin, a system utility with no non-root use cases in bin, etc.) as upstream bug report territory. If we change it prospectively, comment with issue link. Or wait for them to do it. Just like any other bug. :)

PR #1735, opened 2016-06-06
grass: 6.4.5; remove gcc dependency; support compilation on 10.11

dakcarto:

---

Fixups so that #676 can continue, also as patch until `homebrew-geospatial` is migrated from `osgeo4mac`.

### DIFF
--- a/Formula/grass.rb
+++ b/Formula/grass.rb
@@ -1,13 +1,13 @@
 class Grass < Formula
   desc "Geographic Resources Analysis Support System"
   homepage "https://grass.osgeo.org/"
-  revision 1
 
   stable do
-    url "https://grass.osgeo.org/grass64/source/grass-6.4.4.tar.gz"
-    sha256 "5ddba27b4e5495f602ee5249a07e287f342dd8e1422ea5d490c04311c731d274"
+    url "https://grass.osgeo.org/grass64/source/grass-6.4.5.tar.gz"
+    sha256 "f501da62807eb08efcb85820859fe5ade9bc392e20641b606273c956bb678f3e"
 
-    # Patches that files are not installed outside of the prefix.
+    # Patches to keep files from being installed outside of the prefix.
+    # Also, quick patch for compiling with clang (as yet, unreported issue)
     patch :DATA
   end
 
@@ -29,7 +29,6 @@ class Grass < Formula
   option "without-gui", "Build without WxPython interface. Command line tools still available."
 
   depends_on :macos => :lion
-  depends_on "gcc" if MacOS.version >= :mountain_lion
   depends_on "pkg-config" => :build
   depends_on "gettext"
   depends_on "readline"
@@ -38,20 +37,14 @@ class Grass < Formula
   depends_on "unixodbc"
   depends_on "fftw"
   depends_on "cairo"
-  depends_on "freetype"
   depends_on :x11 # needs to find at least X11/include/GL/gl.h
-  depends_on "wxpython" => :recommended
+  depends_on "wxpython" if build.with? "gui"
   depends_on :postgresql => :optional
   depends_on :mysql => :optional
 
-  fails_with :clang do
-    cause "Multiple build failures while compiling GRASS tools."
-  end
-
   def headless?
-    # The GRASS GUI is based on WxPython. Unfortunately, Lion does not include
-    # this module so we have to drop it.
-    build.without?("gui") || MacOS.version == :lion
+    # The GRASS GUI is based on WxPython.
+    build.without? "gui"
   end
 
   def install
@@ -70,6 +63,7 @@ class Grass < Formula
       "--with-sqlite",
       "--with-odbc",
       "--with-geos=#{Formula["geos"].opt_bin}/geos-config",
+      "--with-proj-share=#{Formula["proj"].opt_share}/proj",
       "--with-png",
       "--with-readline-includes=#{readline}/include",
       "--with-readline-libs=#{readline}/lib",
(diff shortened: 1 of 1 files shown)


Formula/grass.rb
```diff
@@ -1,13 +1,13 @@
 class Grass < Formula
   desc "Geographic Resources Analysis Support System"
   homepage "https://grass.osgeo.org/"
-  revision 1
```
dunn: This should either be left as-is, or changed to `revision 2` if everyone should receive the updated version.

dakcarto: Why? The base version has changed.

Formula/grass.rb
```diff
@@ -1,13 +1,13 @@
...
   stable do
-    url "https://grass.osgeo.org/grass64/source/grass-6.4.4.tar.gz"
-    sha256 "5ddba27b4e5495f602ee5249a07e287f342dd8e1422ea5d490c04311c731d274"
+    url "https://grass.osgeo.org/grass64/source/grass-6.4.5.tar.gz"
+    sha256 "f501da62807eb08efcb85820859fe5ade9bc392e20641b606273c956bb678f3e"
 
-    # Patches that files are not installed outside of the prefix.
+    # Patches to keep files from being installed outside of the prefix.
+    # Also, quick patch for compiling with clang (as yet, unreported issue)
     patch :DATA
```
MikeMcQuaid: Can you please submit all these patches to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/grass.rb
```diff
@@ -117,6 +109,12 @@ def install
       args << "--with-mysql"
     end
 
+    if MacOS.version >= :el_capitan
+      # handle stripping of DYLD_* env vars by SIP when passed to utilities;
+      # HOME env var is .brew_home during build, so it is still checked for lib
+      ln_sf "#{buildpath}/dist.x86_64-apple-darwin#{`uname -r`.strip}/lib", ".brew_home/lib"
```
MikeMcQuaid: Can you elaborate a bit more on why this is needed?

Formula/grass.rb
```diff
@@ -127,35 +125,19 @@ def install
   def caveats
     if headless?
       <<-EOS.undent
-        This build of GRASS has been compiled without the WxPython GUI. This is
-        done by default on Lion because there is no stable build of WxPython
-        available to compile against.
+        This build of GRASS has been compiled without the WxPython GUI.
```
MikeMcQuaid: Can remove this caveat.

Formula/grass.rb
```diff
@@ -70,6 +63,7 @@ def install
       "--with-sqlite",
       "--with-odbc",
       "--with-geos=#{Formula["geos"].opt_bin}/geos-config",
+      "--with-proj-share=#{Formula["proj"].opt_share}/proj",
```
UniqMartin: This should be `#{Formula["proj"].opt_pkgshare}` instead. Also, am I overlooking something because I don't see `proj` among the dependencies of this formula?

Formula/grass.rb
```diff
@@ -38,20 +37,14 @@ class Grass < Formula
...
-  fails_with :clang do
-    cause "Multiple build failures while compiling GRASS tools."
-  end
-
   def headless?
-    # The GRASS GUI is based on WxPython. Unfortunately, Lion does not include
-    # this module so we have to drop it.
-    build.without?("gui") || MacOS.version == :lion
+    # The GRASS GUI is based on WxPython.
+    build.without? "gui"
```
UniqMartin: The check is now sufficiently simple and no longer warrants a helper method. Please inline this code.

Formula/grass.rb
```diff
@@ -78,7 +72,8 @@ def install
       "--with-nls-libs=#{gettext}/lib",
       "--with-nls",
       "--with-freetype",
-      "--without-tcltk" # Disabled due to compatibility issues with OS X Tcl/Tk
+      "--without-tcltk", # Disabled due to compatibility issues with OS X Tcl/Tk
+      "--with-includes=#{gettext}/include"
```
UniqMartin: Please add a comma after the last element of an array.

Formula/grass.rb
dunn: Oh, I didn't see the version change down in the stable block, my apologies!

Formula/grass.rb
UniqMartin: This has been fixed in upstream's `trunk` in http://trac.osgeo.org/grass/changeset/58434 (patch does not apply at all thanks to new files and all `*.cc` files having been renamed to `*.cpp`). A backport of this to the 6.4 branch has been requested in http://trac.osgeo.org/grass/ticket/2170 (already two years ago; the target milestone recently changed to 6.4.6, but the change hasn't made it into the upstream `releasebranch_6_4` branch yet).

MikeMcQuaid: 👍 to making all this an inline comment.

DomT4:
Will need to rebase this now given it was moved to the `boneyard` so the `gcc` PR could go ahead.

Formula/grass.rb
DomT4: Should probably also be using `"#{ENV["HOME"]}/lib"` rather than `.brew_home` directly. IMO we make promises on `ENV["HOME"]` being stable but no promises on not changing that directory name at some future point if necessary.

dakcarto:
If it is moved to `boneyard` then there is no reason for this PR. The main `grass` formulae from osgeo4mac take precedence (at least relative to where I'd prefer to spend my time working 😏 ). Feel free to close if that makes sense.

Formula/grass.rb
dakcarto: @DomT4 Thanks for the info on that.

DomT4:
I think we'd accept it again if you can get it working, but if you'd prefer to hold off until you're ready to migrate everything to the new tap I personally don't have an objection to that.

DomT4:
Closing here. Will look forwards to this being revived in the geospatial tap.

PR #1752, opened 2016-06-06
libconfig 1.6

ilovezfs:
- bump version
- force regeneration of two outdated BUILT_SOURCES files, which upstream
  has under revision control


DomT4:
Seems reasonable 👍.

PR #1753, opened 2016-06-06
mupdf-tools 1.9a (and add 'mudraw' symlink)

### LINKED ISSUE #1747: mupdf-tools: Create symlink from mudraw to mutool
I installed mupdf-tools looking for the mudraw command. `mutool` is installed, and you can access mudraw via `mutool draw`, but this is problematic for third party tools that expect `mudraw`.

This can be solved by creating a symlink from `mutool` to `mudraw`, which works great.

Can that be made a part of the official package?

UniqMartin:

---

Other distributions (e.g. Debian) provide a `mudraw` binary by default. Appease users and scripts accustomed to this by providing a matching symbolic link that behaves identically to `mutool draw`.

Also address a Ruby syntax warning by replacing an unnecessary regular expression with a plain string.

Closes #1747.

### DIFF
--- a/Formula/mupdf-tools.rb
+++ b/Formula/mupdf-tools.rb
@@ -23,6 +23,10 @@ class MupdfTools < Formula
            "HAVE_X11=no",
            "CC=#{ENV.cc}",
            "prefix=#{prefix}"
+
+    # Symlink `mutool` as `mudraw` (a popular shortcut for `mutool draw`).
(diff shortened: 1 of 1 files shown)


UniqMartin:
**Additionally:** Drop `openssl` dependency as it is no longer expected on Darwin and not used, even if forced with `HAVE_OPENSSL=yes` as a `Makefile` argument.

PR #1758, opened 2016-06-06
homeshick 1.0.0 (new formula)

benjaminfrank:

---

### DIFF
--- a/Formula/homeshick.rb
+++ b/Formula/homeshick.rb
@@ -1,0 +1,60 @@
+class Homeshick < Formula
+  desc "Git dotfiles synchronizer written in bash"
+  homepage "https://github.com/andsens/homeshick"
+  url "https://github.com/andsens/homeshick/archive/1.0.0.tar.gz"
+  sha256 "8bd3c46f1cfd68d82d97fa72a68a07c966092c77f276d1335cb390b2ec6062bf"
+  head "https://github.com/andsens/homeshick.git"
+
+  option "with-fish", "Build fish bindings"
+  option "with-csh", "Build csh bindings"
+
+  def install
+    inreplace "bin/homeshick", /^homeshick=.*/, "homeshick=#{opt_prefix}"
+
+    prefix.install "lib", "homeshick.sh"
+    prefix.install "homeshick.fish" if build.with? "fish"
+    bin.install "bin/homeshick"
+    bin.install "bin/homeshick.csh" if build.with? "csh"
+    zsh_completion.install "completions/_homeshick"
+    bash_completion.install "completions/homeshick-completion.bash"
+    if build.head? && build.with?("fish")
+      fish_completion.install "completions/homeshick.fish"
+    end
+  end
+
+  def caveats
+    s = <<-EOS.undent
+      To enable the `homeshick cd <CASTLE>` command, you need to
+      `export HOMESHICK_DIR=#{opt_prefix}`
+      and
+      `source "#{opt_prefix}/homeshick.sh"`
+      in your $HOME/.bash_rc
+    EOS
+    if build.with? "fish"
+      s += <<-EOS.undent
+        and
+        `#{opt_prefix}.fish`
+        in your $HOME/.config/fish/config.fish
+      EOS
+    end
+    if build.with? "csh"
+      s += <<-EOS.undent
+        and
+        `alias homeshick source "#{opt_bin}/homeshick.csh"`
+        in your $HOME/.cshrc
+      EOS
+    end
+    s
+  end
+
+  test do
+    (testpath/"test.sh").write <<-EOS.undent
+      #!/bin/sh
+      export HOMESHICK_DIR="#{opt_prefix}"
+      source "#{opt_prefix}/homeshick.sh"
+      homeshick generate test
+      homeshick list
+    EOS
+    assert_match "test", shell_output("bash #{testpath}/test.sh")
+  end
+end


benjaminfrank:
The patch is needed because due to the fact that bin/homeshick is not simply a binary you can execute. The installation guide mentions:
```
Note: The reason we want to define a function/use source instead of aliasing
bin/homeshick to homeshick is because of the cd command. homeshick 
cannot change the working directory of your current shell if it is invoked 
as a subprocess.
```

Formula/homeshick.rb
```diff
@@ -0,0 +1,122 @@
...
+  homepage "https://github.com/andsens/homeshick"
+  url "https://github.com/andsens/homeshick/archive/1.0.0.tar.gz"
+  sha256 "8bd3c46f1cfd68d82d97fa72a68a07c966092c77f276d1335cb390b2ec6062bf"
+  head "https://github.com/andsens/homeshick.git"
+
+  option "with-fish", "Build fish bindings"
+  option "with-csh", "Build csh bindings"
+  option "without-bash", "Build without bash bindings"
+
+  if (build.without? "bash") && (build.without? "fish") && (build.without? "csh")
```
DomT4: This should be inside `def install`, but either way I'd probably rather we just build at least one of these options. It seems reasonable to assume people would want `bash` support?

Formula/homeshick.rb
```diff
@@ -0,0 +1,122 @@
...
+    odie <<-EOS.undent
+      You need to enable at least one of `--with-csh` or `--with-fish` or
+      omit `--without-bash` for a functional homeshick installation.
+    EOS
+  end
+
+  def install
+    prefix.install "lib"
```
DomT4: An array might work here:

```ruby
installs = %w[ lib bin/homeshick]
installs << "homeshick.sh" if build.with? "bash"
installs << "homeshick.fish" if build.with? "fish"
installs << "bin/homeshick.csh" if build.with? "csh"
prefix.install installs
```

DomT4:
> The patch is needed because due to the fact that bin/homeshick is not simply a binary you can execute. However, homebrew makes files in bin/ available in PATH leading to strange results before the required export and sourcing happens. 

This doesn't sound particularly easy to package for Homebrew given the need for the not small patch that won't ever be accepted upstream in the current format. It seems like a good middle ground would be to work with upstream for better configurability.

Formula/homeshick.rb
```diff
@@ -0,0 +1,122 @@
...
+    if build.head? && (build.with? "fish")
+      fish_completion.install "completions/homeshick.fish"
+    end
+  end
+
+  patch :DATA
+
+  def caveats
+    s = <<-EOS.undent
+      `export HOMESHICK_DIR=/usr/local/opt/homeshick`
```
DomT4: Homebrew won't always be in `/usr/local`. Should be using the `opt_prefix` style helpers.

benjaminfrank:
I will raise the issue with homeshick dev @andsens and see what could be done here

Formula/homeshick.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  def caveats
+    s = <<-EOS.undent
+      `export HOMESHICK_DIR=#{opt_prefix}`
+      is required to enable homeshick.
+    EOS
+    if build.with? "bash"
+      s += <<-EOS.undent
+
+        Eable homeshick in bash by appending
```
andsens: Enable

benjaminfrank:
Regarding the patch, would a one-line patch be acceptable? This way we could keep everything as is. For details see https://github.com/andsens/homeshick/issues/127#issuecomment-224410797

MikeMcQuaid:
@benjaminfrank We could use `inreplace` to set the value of a single variable but anything more than that should be merged upstream.

benjaminfrank:
Ok, sounds reasonable. This would allow homeshick to work. The only catch would be that the wrapper scripts behave a bit different then upstream in the case someone sets HOMESHICK_DIR to something else than the homebrew location. But then, why would someone install it via homebrew and then change the location to something else.

Formula/homeshick.rb
```diff
@@ -0,0 +1,68 @@
+class Homeshick < Formula
+  desc "Git dotfiles synchronizer written in bash"
+  homepage "https://github.com/andsens/homeshick"
+  url "https://github.com/andsens/homeshick/archive/1.0.0.tar.gz"
+  sha256 "8bd3c46f1cfd68d82d97fa72a68a07c966092c77f276d1335cb390b2ec6062bf"
+  head "https://github.com/andsens/homeshick.git"
+
+  option "with-fish", "Build fish bindings"
+  option "with-csh", "Build csh bindings"
+  option "without-bash", "Build without bash bindings"
```
DomT4: This was made a default, so this option doesn't do anything now.

Formula/homeshick.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def install
+    inreplace "bin/homeshick", /^homeshick=.*/, "homeshick=#{opt_prefix}"
+
+    prefix.install "lib"
```
DomT4: You'd make me a mildly happier human if you replaced the two unconditional `prefix.install`s with an array 😉

```ruby
prefix.install %w[ lib homeshick.sh ]
```
Or even just:
```ruby
prefix.install "lib", "homeshick.sh"
```

Formula/homeshick.rb
```diff
@@ -0,0 +1,68 @@
...
+    zsh_completion.install "completions/_homeshick"
+    bash_completion.install "completions/homeshick-completion.bash"
+    if build.head? && (build.with? "fish")
+      fish_completion.install "completions/homeshick.fish"
+    end
+  end
+
+  def caveats
+    s = <<-EOS.undent
+      `export HOMESHICK_DIR=#{opt_prefix}`
```
DomT4: Is this still the case, given the `inreplace`?

Formula/homeshick.rb
```diff
@@ -0,0 +1,68 @@
...
+    EOS
+    if build.with? "bash"
```
DomT4: Will need to tweak this a bit if `bash` is unconditional now.

Formula/homeshick.rb
```diff
@@ -0,0 +1,68 @@
...
+    s += <<-EOS.undent
+      is required to enable 'homeshick cd <CASTLE>'.
+    EOS
+    s
+  end
+
+  test do
+    (testpath/"test.sh").write <<-EOS.undent
+      #!/bin/sh
+      export HOMESHICK_DIR="/usr/local/opt/homeshick/"
```
DomT4: Don't assume `/usr/local` here.

Formula/homeshick.rb
benjaminfrank: To enable the subcommand `homeshick cd <CASTLE>` yes as it required the shell wrappers which in turn use $HOMESHICK_DIR to execute the binary homeshick. Changeing the shell wrapper would take a couple more `inreplace` commands.
The only other option I'd see would be to tell users in the caveat the content of those wrappers (which is a small shell function). @andsens @DomT4 @MikeMcQuaid what would be your take on something like this:

instead of those shell wrappers (export & sourceing) we could print something like:
```
To enable `homeshick dir <CASTLE>` command append

function homeshick() {
	if [ "$1" = "cd" ] && [ -n "$2" ]; then
		cd "$HOME/.homesick/repos/$2"
	else
		"#{opt_prefix}/bin/homeshick" "$@"
	fi
}

to your $HOME/.bashrc
```

Formula/homeshick.rb
```diff
@@ -0,0 +1,60 @@
...
+  def install
+    inreplace "bin/homeshick", /^homeshick=.*/, "homeshick=#{opt_prefix}"
+
+    prefix.install "lib", "homeshick.sh"
+    prefix.install "homeshick.fish" if build.with? "fish"
+    bin.install "bin/homeshick"
+    bin.install "bin/homeshick.csh" if build.with? "csh"
+    zsh_completion.install "completions/_homeshick"
+    bash_completion.install "completions/homeshick-completion.bash"
+    if build.head? && (build.with? "fish")
```
MikeMcQuaid: `build.with?("fish")`

MikeMcQuaid:
LGTM. @DomT4?

andsens:
Yup, looks good. This is the smallest changeset that gets bin/homeshick working, while letting the users through a single `export` make the shell function "modification" themselves.

Formula/homeshick.rb
```diff
@@ -0,0 +1,60 @@
...
+      s += <<-EOS.undent
+        and
+        `#{opt_prefix}.fish`
+        in your $HOME/.config/fish/config.fish
+      EOS
+    end
+    if build.with? "csh"
+      s += <<-EOS.undent
+        and
+        `alias homeshick source "#{opt_prefix}/bin/homeshick.csh"`
```
DomT4: `opt_bin` rather than `opt_prefix/bin`

Formula/homeshick.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  test do
+    (testpath/"test.sh").write <<-EOS.undent
+      #!/bin/sh
+      export HOMESHICK_DIR="#{opt_prefix}"
+      source "#{opt_prefix}/homeshick.sh"
+      homeshick generate test
+      homeshick list
+    EOS
+    assert_match /test/, shell_output("bash #{testpath}/test.sh")
```
DomT4: Unless you're actually regex matching here just use:
```ruby
assert_match "test", shell_output("bash #{testpath}/test.sh")
```

DomT4:
Few style nits but essentially 👍.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/8622237b2a499762cb3e1607dddfccc4854b9e10. Thank you for your contribution to Homebrew @benjaminfrank; we appreciate it! :smiley_cat:

PR #1764, opened 2016-06-06
hugo 0.16

ilovezfs:


DomT4:
They'll be doing us a big favour if/when they eventually move to Go's vendor system 😅. LGTM.

ilovezfs:
@DomT4 A question about gdm. When there's a discrepancy between what the Godeps file checked in to git says and what gdm says, I've been assuming we just go with what gdm pukes out assuming it builds for both the given stable tag and head. Is that the "correct" procedure, or should I manually be intervening in the gdm output to make the revisions match what's in the Godeps file even if it doesn't seem to matter one or the other?

DomT4:
Hmm. Interesting question, possibly one for @sparrc given `gdm` is their creation. I'd lean towards "obeying" `godeps` files personally.

ilovezfs:
Example:
https://github.com/monochromegane/the_platinum_searcher/blob/master/Godeps/Godeps.json
https://github.com/Homebrew/homebrew-core/pull/1728/files

Yeah, I felt funny about overriding those explicit choices.

ilovezfs:
@DomT4 I wonder if we can somehow just use the Godeps.json directly rather than go_resource blocks. One problem is that the Godeps file can be a moving target so unless we're directly using the upstream file in def install, I don't see how having head specs makes any sense. If we're just going to use hard coded go_resource blocks, I'd rather have no head spec than one that can be blatantly incorrect.

DomT4:
Yeah, where `godep` is supported upstream we should ideally be using it. In most cases it's as simple as adding the `godep` dependency and then plonking `godep` in front of the `go build` command, but sometimes it is a bit more involved than that, and sometimes it doesn't work at all for completely mysterious reasons.

It's not necessarily a great system and I'm quite glad the vendor system is gaining steam instead, to be honest.

ilovezfs:
Yeah invoking godep directly seems like a much better approach if the vendor system isn't available.

ilovezfs:
@DomT4 So looks like if we wanted to continue to use the resource blocks to cache the resources, the revisions in Godeps.json can be restored by doing this:
```
  def install
    ENV["GOPATH"] = buildpath
    dir = buildpath/"src/github.com/monochromegane/the_platinum_searcher"
    dir.install Dir.glob("*", File::FNM_DOTMATCH) - %w[. .. src]
    Language::Go.stage_deps resources, buildpath/"src"
    ln_s buildpath/"src", dir
    cd dir do
      resources.each { |r| cd("src/#{r.name}") { puts "#{r.name}: #{`git rev-parse HEAD`}" } }
      system "godep", "restore"
      resources.each { |r| cd("src/#{r.name}") { puts "#{r.name}: #{`git rev-parse HEAD`}" } }
      system "go", "build", "-o", bin/"pt", "cmd/pt/main.go"
    end
  end
```
Those `resources.each ...` just demonstrate the revisions have been restored. The dir.install thing prevents the godeps from throwing the annoying (but I suppose harmless) warning
```
godep: [WARNING]: godep should only be used inside a valid go package directory and
godep: [WARNING]: may not function correctly. You are probably outside of your $GOPATH.
```

In order for the revision restoring to work, the following change is required to resource.rb:
```
iMac-TMP:local joe$ git diff
diff --git a/Library/Homebrew/resource.rb b/Library/Homebrew/resource.rb
index 3fcdf46..188afdd 100644
--- a/Library/Homebrew/resource.rb
+++ b/Library/Homebrew/resource.rb
@@ -98,7 +98,7 @@ class Resource
         yield ResourceStageContext.new(self, staging)
       elsif target
         target = Pathname.new(target) unless target.is_a? Pathname
-        target.install Dir["*"]
+        target.install Dir.glob("*", File::FNM_DOTMATCH) - %w[. ..]
       end
     end
   end
```
which I think actually makes it more correct. I've always thought we should install the entire resource contents and not just randomly skip the invisible files and directories. If you don't do this, then the .git directory isn't installed, which means that `godep restore` has no way of checking out the correct revisions.

Basically this demonstrates that the `:revision =>` part of the go resource blocks is unnecessary if you have a `Godeps.json` file.

Of course, `system "godep", "restore"` on its own works without the resource blocks, but then nothing is cached, which isn't necessarily a big deal, but I was curious if it could be done using the existing tools, which clearly it can.

Perhaps we can find another way of caching that doesn't involve the cumbersome resource blocks, though.

ilovezfs:
@DomT4 Regarding `hugo` vendoring specifically: https://github.com/spf13/hugo/issues/2119

Do we have any preferences to express over there in terms of which of the solutions they're considering is best for our purposes?

DomT4:
> Of course, system "godep", "restore" on its own works without the resource blocks, but then nothing is cached, which isn't necessarily a big deal, but I was curious if it could be done using the existing tools, which clearly it can.

Aye. It can be cached, but I'm not massively fussed about doing so to be honest, perhaps because I naively hope Go's push towards the vendor system proves to be a significant success for still-developed projects.

> Perhaps we can find another way of caching that doesn't involve the cumbersome resource blocks, though.

The `ENV.java_cache` situation works surprisingly well for Java, I'm pretty fond of that. Not sure if Go has anything similar, though.

> Do we have any preferences to express over there in terms of which of the solutions they're considering is best for our purposes?

It's their decision pretty much, but I'm glad to see they're discussing using a full vendor system. Go doesn't seem fussy about how that `/vendor` directory is created, and that's more or less the important thing from a packaging point of view.

We could probably add a formula for `govendor` given it has tagged releases in the hopes of making that more accessible.

ilovezfs:
> We could probably add a formula for govendor given it has tagged releases in the hopes of making that more accessible.

I'm surprised no one has shown up with a PR for that already.

PR #1768, opened 2016-06-07
the_platinum_searcher: use godep

ilovezfs:
Also, bumping revision since this now uses the git commits specified by
upstream, instead of whatever gdm happened to output.

### DIFF
--- a/Formula/the_platinum_searcher.rb
+++ b/Formula/the_platinum_searcher.rb
@@ -5,6 +5,7 @@ class ThePlatinumSearcher < Formula
   homepage "https://github.com/monochromegane/the_platinum_searcher"
   url "https://github.com/monochromegane/the_platinum_searcher/archive/v2.1.2.tar.gz"
   sha256 "db8cbe30381e7f0a6cdb4b8ab306194483df16197b268790b6d8cee3f8d32d05"
+  revision 1
   head "https://github.com/monochromegane/the_platinum_searcher.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Formula/the_platinum_searcher.rb
```diff
@@ -15,59 +16,15 @@ class ThePlatinumSearcher < Formula
...
-    mkdir_p buildpath/"src/github.com/monochromegane"
-    ln_s buildpath, buildpath/"src/github.com/monochromegane/the_platinum_searcher"
-
     ENV["GOPATH"] = buildpath
-    Language::Go.stage_deps resources, buildpath/"src"
-
+    dir = buildpath/"src/github.com/monochromegane/the_platinum_searcher"
+    dir.install Dir["*"]
+    ln_s buildpath/"src", dir
+    cd dir
```
DomT4: Would be nice to make this `cd dir do` to match most usage elsewhere. We try to avoid vanilla `cd`s unless using them takes up a ridiculous amount of space or is problematic in some other way.

DomT4:
Beside the style nit, if it works and tests okay, 👍.

Formula/the_platinum_searcher.rb
ilovezfs: That's how I had it to begin with, but figured I'd see if I could get away with one fewer line of code or if you'd notice :smirk_cat:

DomT4: When it comes to formulae, I notice everything 😈 😄.

ilovezfs:
The interesting thing to note here is why the dir.install is necessary to shut up the warning, which is that if you `cd` into the ordinary symlinked directory as other formulae usually have it, Dir.pwd does not change and it's as if you did nothing. The reason is that it's actually just shell magic that makes it look like that ever works. So in order to actually change Dir.pwd in terms of how `godeps` sees it, the real dir has to be the one with the "src/github.com/monochromegane/the_platinum_searcher" in the path.

PR #1770, opened 2016-06-07
memo: boneyard

ilovezfs:


ilovezfs:
The reason for boneyarding is that the repository is gone and no where to be found elsewhere on the web as far as I can tell.

DomT4:
Seems fine, both the homepage and GitHub URL are now 404s. Only downloaded ~100 times from bottles over the last year, and most of that was clustered back in July 2015 for some reason.

Issue #1743, opened 2016-06-06
gitup refers to out of date GitPython


dotwaffle:
The "gitup" formula refers to a python module called "GitPython": https://github.com/Homebrew/homebrew-core/blob/fb81645562f7e495a49561065912788af2d1f14f/Formula/gitup.rb#L33-36

Unfortuantely, this particular version referred to is quite old, and has a number of bugs -- including not closing file handles that makes gitup occasionally crash with larger numbers of repositories.

https://github.com/earwig/git-repo-updater/issues/9 says that using a later version should fix the issue.

Altering the formula to depend on a later version of GitPython should fix the issue. There has not been a new version of gitup or GitPython released, supposedly just using the HEAD of GitPython works... I am unsure as to the way to proceed as the current version refers to a pre-built package -- happy to help if/where I can.

ilovezfs:
@dotwaffle This should fix it: https://github.com/Homebrew/homebrew-core/pull/1773

`brew pull 1773` if you'd like to test that it does.

ilovezfs:
@dotwaffle the fix is now merged to master, so if you'd like to test it, you just need to brew update  and brew upgrade.

dotwaffle:
Seems to work great, thank you!

ilovezfs:
@dotwaffle Excellent news. Thanks for reporting this and the testing follow-up!

PR #1773, opened 2016-06-07
gitup: refresh Python resources

ilovezfs:
Also, use absolute paths in the test.

Closes https://github.com/Homebrew/homebrew-core/issues/1743.


DomT4:
Maybe worth bumping the revision if it resolves breakage? Otherwise 👍.

ilovezfs:
Good point.

PR #1782, opened 2016-06-07
doctl 1.2.0

ilovezfs:
Also,
- fix audit error about redundant to_s inside interpolation
- fix warning about trying to stage empty resources array

### DIFF
--- a/Formula/doctl.rb
+++ b/Formula/doctl.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Doctl < Formula
   desc "Command-line tool for DigitalOcean"
   homepage "https://github.com/digitalocean/doctl"
-  url "https://github.com/digitalocean/doctl/archive/v1.1.0.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Someone filed a PR earlier FWIW: https://github.com/Homebrew/homebrew-core/pull/1778 😜

ilovezfs:
@DomT4 ah ok. Then I guess we can fix the audit error and warning separately.

DomT4:
You could always `brew pull` the older PR into your branch, add a separate commit on top of it with your audit changes, push here & then merge 'em together. I do that sometimes if prior PRs are "incomplete".

ilovezfs:
Since the two changes here don't actually affect the bottle, I think we can pull his, and then pull this without updating the bottle?

ilovezfs:
ah now I understand

UniqMartin:
@ilovezfs #1778 has been merged in case you want to follow up with your tweaks.

ilovezfs:
@UniqMartin Thanks for the heads up. Voilà: https://github.com/Homebrew/homebrew-core/pull/1789

PR #1789, opened 2016-06-07
doctl: audit fix and empty resources array fix

ilovezfs:
- fix audit error about redundant to_s inside interpolation
- fix warning about trying to stage empty resources array

I would also like to note that 1789 is the year the American Constitution went into force, so this is a very special pull request :)

### DIFF
--- a/Formula/doctl.rb
+++ b/Formula/doctl.rb
@@ -1,5 +1,3 @@
-require "language/go"
-
 class Doctl < Formula
   desc "Command-line tool for DigitalOcean"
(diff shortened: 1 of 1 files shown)


UniqMartin:
:+1:

ilovezfs:
@UniqMartin hmm I suppose I should probably remove `require "language/go"` too

UniqMartin:
@ilovezfs Indeed, you should. Sorry for missing this and thanks for paying attention!

PR #1793, opened 2016-06-07
Conan 0.9.3

lasote:

---

### DIFF
--- a/Formula/conan.rb
+++ b/Formula/conan.rb
@@ -1,8 +1,8 @@
 class Conan < Formula
   desc "Distributed, open source, package manager for C/C++"
   homepage "https://github.com/conan-io/conan"
-  url "https://pypi.python.org/packages/0b/34/5b03ad307911ef4221dd965a8867502dc96107cf58ed83f3274421e6808d/conan-0.9.2.tar.gz"
-  sha256 "750cef39e504d6d1e267a4fd6fc42ff4d9f4a3a25bbe3dfdbdff573025932db0"
+  url "https://pypi.python.org/packages/9d/e3/1e8082abbe379e430c6e70fd03ac08b443a065b8f33ff171efa2ff629069/conan-0.9.3.tar.gz"
+  sha256 "f8eb096507ac47f1c83636d4ac0cdba4b291b0293794bc74eb6edba593b14f0a"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


Formula/conan.rb
```diff
@@ -11,7 +11,11 @@ class Conan < Formula
     sha256 "aae8d317e2e9add74d893d0183c3f93539f98b81f59f67bed1d1f2a9466198bd" => :mavericks
   end
 
-  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on :python
+  resource "backport_ipaddress" do
```
UniqMartin: Please leave an empty line between the `depends_on` and the `resource` line. Also, why did you drop the `if MacOS.version <= :snow_leopard` on the `depends_on :python` line?

Formula/conan.rb
```diff
@@ -33,16 +42,41 @@ class Conan < Formula
     sha256 "e043c8d32527607223652021ff648fbb394d5e19cba9f1a698670b338c9d782b"
   end
 
+  resource "cryptography" do
+    url "https://pypi.python.org/packages/a9/5b/a383b3a778609fe8177bd51307b5ebeee369b353550675353f46cb99c6f0/cryptography-1.4.tar.gz"
+    sha256 "bb149540ed90c4b2171bf694fe6991d6331bc149ae623c8ff419324f4222d128"
+  end
```
UniqMartin: Does this also mean the formula now depends on `openssl` and thus needs a `depends_on "openssl"` line? At least that's my interpretation of the build failure where it fails to find an OpenSSL header …

Formula/conan.rb
```diff
@@ -75,7 +129,7 @@ class Conan < Formula
 
   def install
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
-    %w[boto bottle cfgparse colorama fasteners monotonic passlib patch PyJWT PyYAML requests six].each do |r|
+    %w[backport_ipaddress boto bottle cffi cfgparse colorama cryptography enum34 fasteners idna ipaddress monotonic ndg-httpsclient passlib patch pyasn pyasn1 pycparser PyJWT pyOpenSSL PyYAML requests six].each do |r|
```
UniqMartin: If the order of these doesn't matter, let's simplify this by doing something like:

```ruby
resources.each do |r|
  r.stage do
  end
end
```

If the order is important, let's at least make this a bit more readable by formatting the `%w[` style array like this:

```ruby
%w[
  backport_ipaddress
  boto
  bottle
  cffi
  (remaining resources)
].each do |r|
  # …
```

lasote:
Hi, thanks for the comments and your time. We realize that 0.9.3 is a bad version, so I'm closing the PR.

PR #1798, opened 2016-06-07
qt5: add -separate-debug-info configure option

jakepetroules:

---

This generates .dSYM bundles for each Qt module and thus allows Qt
applications to be properly debugged with full stack traces and source
info.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -66,6 +66,8 @@ class Qt5 < Formula
       -verbose
       -prefix #{prefix}
       -release
+      -separate-debug-info
+      -force-debug-info
       -opensource -confirm-license
       -system-zlib
       -qt-libpng


UniqMartin:
I aborted the build for now since it won't finish with the normal CI timeout anyway. Can you provide a bit more information about this change and why you think it's worth enabling for *all users*? Namely:

- What is the build time impact, i.e. how long does the build take without and with this option added?
- What is the impact on disk space, i.e. what is the size of the formula prefix before and after?
- Is it actually that useful given that Qt is built in release mode and thus (hopefully) heavily optimized by the compiler, thus stack traces and source locations might be less useful?

Thanks!

jakepetroules:
- The build time impact should be near zero.
- The impact on disk space may be a bit larger.
- Yes, it's quite useful to have this additional information when debugging or reporting crashes or bugs in applications relying on Qt. We plan to do this for official packages of Qt beginning with our next release.

UniqMartin:
> - The build time impact should be near zero.

Can you be more specific, i.e. provide concrete numbers for the machine you tested this with? My experience is that linking binaries in a projects of that size is a non-negligible factor. Merging non-trivial amounts of debug information will very likely take some time, not to mention the slowdown because the linker has to read and process a lot more information.

> - The impact on disk space may be a bit larger.

For full disclosure: This is the factor that concerns me most, as it affects all Homebrew users. Like above, I'd be interested to see the exact numbers. `qt5` is already a pretty big formula and if this change means that this doubles the size of the bottle download and the size of the formula on disk, then that's something to ponder a bit longer. (Don't get me wrong. I have a lot of sympathy for this PR, but that's a rather developer-focused view and the cost/benefit ratio for the average user isn't clear to me yet.)

> - Yes, it's quite useful to have this additional information when debugging or reporting crashes or bugs in applications relying on Qt. We plan to do this for official packages of Qt beginning with our next release.

That's interesting. Thanks for this insight!

UniqMartin:
@BrewTestBot test this please

UniqMartin:
From the log of the failed CI build:

```
==> brew install --build-bottle --verbose qt5
[…snip…]
==> ./configure -verbose -prefix /usr/local/Cellar/qt5/5.6.0 -release -separate-debug-info -opensource -confirm-license -system-zlib -qt-libpng -qt-libjpeg -qt-freetype -qt-pcre -nomake tests -no-rpath -nomake examples -no-dbus
[…snip…]
ERROR: -separate-debug-info needs -debug, -debug-and-release, or -force-debug-info
```

I have to admit that I have trouble understanding this given this checked item from the top comment:

> - [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?

jakepetroules:
Missed the -release in there.

@BrewTestBot test this please

UniqMartin:
> `@BrewTestBot` test this please

No need for this. Force pushing to the PR branch will automatically trigger a new CI build. Additionally, the bot only listens to Homebrew maintainers when addressed like that.

jakepetroules:
Why aborted?

UniqMartin:
> Why aborted?

Sorry if this sounds harsh, but to save our CI from wasting time on something that's broken and untested.

----

I'm very tempted to flat out close this PR because it shows total disrespect for our work at Homebrew, but for transparency's sake I'll “spend” a bit more time explaining why I'm feeling that way:

- The “Have you built your formula locally prior to submission […]” box has been ticked from the start, though the original build already aborted at `./configure` time. This makes it evident that it hasn't been tested at all. (But no big deal, many people tick those boxes without even reading the text.)

- A hastily added `-force-debug-info` option rectified this, but didn't change the fact that this PR has been filed against Qt 5.6.0 (I have since bumped it to 5.6.1 in #1840), a Qt release that doesn't even properly support this option on Apple platforms before 5.6.1, if [QTBUG-37952](https://bugreports.qt.io/browse/QTBUG-37952) is accurate. This is even more surprising since you are the one who authored the fix for QTBUG-37952.

- My questions about build time and disk space impact have been answered rather vaguely and the (extensively motivated) follow-up questions for clarification and specific numbers have been ignored.

- Being a bit curious and maybe too lenient, I simply applied the proposed change to Qt 5.6.1 and tried it myself. Turns out: the `-separate-debug-info` option is still broken even though the `-g` flag is added to the build flags and the debug information is moved to `.dSYM` bundles for frameworks, dylibs, and app bundles. The problem is that `make install` fails to copy **all* the generated `.dSYM` bundles to the installation prefix, effectively causing only the static libraries (`.a` files) to have debug information.

  Now this could be related to how Homebrew builds Qt, but having looked at the build logs I highly doubt this. (I can provide the logs if that's of any help with debugging. My `qmake` skills are insufficient to do that myself and my personal interest in this is fading.)

All of this could have been avoided (or handled better), if the PR was actually built and tested locally. It could have been really simple to check whether the `.dSYM` bundles are actually there or one could have done more extensive testing by crashing a Qt app and making sure the debug information is actually usable and (automatically) found by the debugger.

----

Since I went to the trouble of experimenting with the proposed change to check if this is something we might want to fold into the 5.6.1 version bump, here are answers to my very own questions from above.

*Disclaimer: Since my resources are limited, I excluded the `qtwebengine` Qt module from my test builds. Thus my numbers might not perfectly extrapolate to a full build.*

**Build time:** About 15-20% longer build times (82 instead of 70 minutes). Not exactly what I would call “near zero”, but not a deal-breaker either. If that extrapolates, 210 versus 180 minutes for a full build.

**Disk space:** An increase of about 200-300%. “The impact on disk space may be a bit larger.” is quite an understatement. If this feature wasn't broken, I'm afraid this would disqualify it from being default-enabled in Homebrew. If that extrapolates, this is something like 900-1200 MB just for the `qt5` formula instead of the current ~300 MB. The debug information compresses fairly well, so the impact on bottles would be lower (but still around 100% increase in size).

Below are some numbers from a minimal non-GUI build of `qtbase` without any other Qt modules.

```
$ du -shc *.framework
296K	QtConcurrent.framework
 11M	QtCore.framework
2.3M	QtNetwork.framework
528K	QtSql.framework
888K	QtTest.framework
520K	QtXml.framework
 16M	total

$ du -shc *.framework.dSYM
560K	QtConcurrent.framework.dSYM
 28M	QtCore.framework.dSYM
 11M	QtNetwork.framework.dSYM
2.3M	QtSql.framework.dSYM
1.9M	QtTest.framework.dSYM
2.5M	QtXml.framework.dSYM
 46M	total
```

----

Maybe I'm just overreacting. Feel free to correct me if anything I've written above is wrong or misleading.

DomT4:
Agree with Martin's assessment here, unless we're missing something. I'm a bit bemused how the original box was ticked when it pretty much instantly failed CI, to be honest.

Qt5 is a chunk to build, taking about a touch north of ~2hrs to build per VM, almost 7 hours of build time in total. For a project such as our's which has fairly slim resources, and the number of PRs we see a day, effectively wasted big builds are a significant burden on CI.

We added those checkboxes as a way to counter that unnecessary build churn, so it is appreciated if care is taken to lessen the stress on our resources.

jakepetroules:
On hold. [QTBUG-54035](https://bugreports.qt.io/browse/QTBUG-54035), [QTBUG-54036](https://bugreports.qt.io/browse/QTBUG-54036)

MikeMcQuaid:
Let's put this on hold until at least the upstream Qt packages use it.

PR #1800, opened 2016-06-07
webp: avoid build issue with freeglut installed

### LINKED ISSUE #1736: webp fails to build on OS X 10.11.5
Looks like it's failing with a linking error:

```
  "_glutLeaveMainLoop", referenced from:
      _HandleKey in vwebp-vwebp.o
  "_glutSetOption", referenced from:
      _main in vwebp-vwebp.o

```
Gist: https://gist.github.com/0566a0aa5a18ff15bc6a0b07378f31b4

UniqMartin:

---

This disables checking for OpenGL/GLUT and effectively results in the `vwebp` binary (a simple WebP viewer) not being built.

The problem being worked around by this is: If `freeglut` (from the `homebrew/x11` tap) is installed and was built without the `--universal` option but then `webp` is built with the `--universal` option (either directly or as a dependency of another formula), then it will find the GLUT headers and libraries provided by `freeglut` (instead of the `GLUT` framework in `/System/Library`) which appears to be fine during the
`./configure` run, but then causes failure during the build due to lacking architecture support.

Closes #1736.

cc @mistydemeo

### DIFF
--- a/Formula/webp.rb
+++ b/Formula/webp.rb
@@ -30,6 +30,7 @@ class Webp < Formula
 
     ENV.universal_binary if build.universal?
     system "./configure", "--disable-dependency-tracking",
+                          "--disable-gl",
                           "--enable-libwebpmux",
                           "--enable-libwebpdemux",
                           "--enable-libwebpdecoder",


ilovezfs:
Don't we usually allow inheritance to handle this as wine and all its deps do? Or is that not possible with these two formulae?

Also, since this results in a component not being built, shouldn't that at least be made optional if there's no way to handle this via the regular inheritance mechanism?

DomT4:
Could perhaps do:

```ruby
args << "--disable-gl" if build.universal?
```
But no objections to the current pure disable here. I can't imagine people banging down our doors to demand its return, so.

UniqMartin:
> Don't we usually allow inheritance to handle this as wine and all its deps do? Or is that not possible with these two formulae?

This only works when there's a dependency relationship between the formulae in question. As it stands, `freeglut` files are picked up unintentionally by `webp` and `webp` neither can nor should depend on `freeglut`.

> Could perhaps do: […]

This quickly crossed my mind, but I consciously decided not to do this, as this will give users a different set of binaries depending on whether they have a universal build or not, and that is likely to cause more irritation than taking away this binary from everyone.

> Also, since this results in a component not being built, shouldn't that at least be made optional if there's no way to handle this via the regular inheritance mechanism?

My thinking here is that the binary is obscure enough to not be missed by 99.9% of our user base. We can still reconsider and try to find a more elegant solution once users come banging at our door.

PR #1801, opened 2016-06-08
telegraf 1.0.0-beta1

gunnaraasen:

---

This is the beta release in preparation for hitting 1.0. It is more akin to a full minor release than true beta, hence making this the default version instead of a `--devel` release.

### DIFF
--- a/Formula/telegraf.rb
+++ b/Formula/telegraf.rb
@@ -6,6 +6,8 @@ class Telegraf < Formula
   url "https://github.com/influxdata/telegraf/archive/0.13.1.tar.gz"
   sha256 "96d67c203eb3820d3cde5bd89c87521ad7a404a495db9f250084f767c26d8990"
 
+  head "https://github.com/influxdata/telegraf.git"
+
   bottle do
     cellar :any_skip_relocation
     sha256 "ff7f3f1f8f4cbdf708eada5ab3ef1e0eb0d594a22690777cb224fe71d6c608d3" => :el_capitan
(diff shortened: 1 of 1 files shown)


ilovezfs:
:-1: See comments here: https://github.com/Homebrew/homebrew-core/pull/1802#issuecomment-224524007

DomT4:
Would prefer it as a `devel` block as per my comment in another thread Joe commented in 😸.

ilovezfs:
Yeah, I'm just mainly concerned that these three formulae go together including a dependency relationship, but there's no way to force `depends_on "foo" => [:devel, "with-blah"] `at this point.

DomT4:
Aye, the inability to do that is something of a pain. @gunnaraasen Do things break if one builds say ` kapacitor 1.0.0-beta1` against an earlier `influxdb` or is there complete backwards compatibility?

gunnaraasen:
I've updated this formula to use `--devel`.

Formula/telegraf.rb
```diff
@@ -6,23 +6,48 @@ class Telegraf < Formula
...
-    url "https://github.com/influxdata/telegraf.git"
+  devel do
+    url "https://github.com/influxdata/telegraf/archive/1.0.0-beta1.tar.gz"
+    sha256 "8b8d9fdcd49bddd9d0a5c62fdaad15c40effa25bd0de031a6004859af3542e4b"
+  end
 
+  depends_on "go" => :build
+
+  if build.stable?
+    go_resource "github.com/gobwas/glob" do
```
MikeMcQuaid: Put these in a `stable do` block.

Formula/telegraf.rb
```diff
@@ -6,23 +6,48 @@ class Telegraf < Formula
...
+      url "https://github.com/gobwas/glob.git",
+      :revision => "d877f6352135181470c40c73ebb81aefa22115fa"
+    end
+
+    go_resource "github.com/shirou/gopsutil" do
+      url "https://github.com/shirou/gopsutil.git",
+      :revision => "83c6e72cbdef6e8ada934549abf700ff0ba96776"
+    end
+  else
     go_resource "github.com/hashicorp/consul" do
```
MikeMcQuaid: Duplicate these in a `devel do` and `head do` block.

Formula/telegraf.rb
```diff
@@ -318,9 +333,19 @@ def install
...
-      system "go", "build", "-o", "telegraf",
-             "-ldflags", "-X main.Version=#{version}",
-             "cmd/telegraf/telegraf.go"
+      if build.head?
+        system "go", "build", "-o", "telegraf",
+               "-ldflags", "-X main.version=1.0.0-HEAD -X main.commit=#{`git rev-parse HEAD`.strip}",
+               "cmd/telegraf/telegraf.go"
+      elsif build.devel?
+        system "go", "build", "-o", "telegraf",
+               "-ldflags", "-X main.version=1.0.0-beta1 -X main.commit=ad88a9421a16456c1e3029392a29c5bb7ad907e7",
```
MikeMcQuaid: Similarly: can we avoid hardcoding the commit if we but from a Git repo?

gunnaraasen:
Replaced by https://github.com/Homebrew/homebrew-core/pull/2258

Formula/telegraf.rb
gunnaraasen: Fixed in https://github.com/Homebrew/homebrew-core/pull/2258

Formula/telegraf.rb
gunnaraasen: Fixed in https://github.com/Homebrew/homebrew-core/pull/2258

PR #1802, opened 2016-06-08
kapacitor 1.0.0-beta1

gunnaraasen:

---

This is the beta release in preparation for hitting 1.0. It is more akin to a full minor release than true beta, hence making this the default version instead of a `--devel` release.

### DIFF
--- a/Formula/kapacitor.rb
+++ b/Formula/kapacitor.rb
@@ -4,8 +4,8 @@ class Kapacitor < Formula
   desc "Open source time series data processor"
   homepage "https://github.com/influxdata/kapacitor"
   url "https://github.com/influxdata/kapacitor.git",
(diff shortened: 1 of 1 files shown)


ilovezfs:
I think this may be problematic because it depends on influxdb, which also presumably should be using the corresponding 1.0.0-beta1 development version, and we have no way of enforcing that. So I think I'd prefer we not have the development versions for this set of formulae unless we're also adding a custom requirement block to make sure the versions are consistent.

ilovezfs:
Also, even if we were going to accept this 1.0.0-beta1 would need to be in a devel-do block not substituted for the stable spec, which I just noticed is what you have here.

gunnaraasen:
Replaced by https://github.com/Homebrew/homebrew-core/pull/2255

PR #1806, opened 2016-06-08
erlang: don't build jinterface by default

enilsen16:

---

### DIFF
--- a/Formula/erlang.rb
+++ b/Formula/erlang.rb
@@ -23,6 +23,7 @@ class Erlang < Formula
   option "without-hipe", "Disable building hipe; fails on various OS X systems"
   option "with-native-libs", "Enable native library compilation"
   option "with-dirty-schedulers", "Enable experimental dirty schedulers"
+  option "with-java", "Build jinterface application"
   option "without-docs", "Do not install documentation"
 
   deprecated_option "disable-hipe" => "without-hipe"
@@ -33,6 +34,7 @@ class Erlang < Formula
   depends_on "libtool" => :build
   depends_on "openssl"
   depends_on "fop" => :optional # enables building PDF docs
+  depends_on :java => :optional
   depends_on "wxmac" => :recommended # for GUI apps like observer
 
   fails_with :llvm
@@ -86,6 +88,12 @@ class Erlang < Formula
       args << "--disable-hipe"
     else
       args << "--enable-hipe"
+    end
+
+    if build.with? "java"
+      args << "--with-javac"
+    else
+      args << "--without-javac"
     end
 
     system "./configure", *args


ilovezfs:
I wonder if we should be doing
```
option "with-java", "Build jinterface application"
depends_on :java => :optional
```

DomT4:
They don't need to be clustered quite that closely, but 👍 on ILZ's suggestion. What's the justification for turning this off by default here by the way?

ilovezfs:
Hehe, yeah I didn't mean to imply the two lines of code should be adjacent.

@DomT4 It does seem the formula currently has a bug insofar as it doesn't declare a `:java` requirement, but I guess it's harmless if it builds it even if you don't actually have Java, in which case it does seem the option would be superfluous.

DomT4:
I don't mind turning it off by default if it's currently problematic in some way, but that's what we need to know here. Agree if there's a Java element in place there probably should be a `:java` dependency in some shape or form.

ilovezfs:
If building the jinterface application doesn't itself require java, then I'm not sure there's any harm in having it sit there inert, but yeah let's see what @enilsen16's reason is.

enilsen16:
When you build the jinterface application without java, OSX prompts the user to go and install java. Which I found fairly confusing, I had to spend time to try and figure out why java would be a dependency of erlang...

Formula/erlang.rb
```diff
@@ -32,6 +33,7 @@ class Erlang < Formula
   depends_on "libtool" => :build
   depends_on "openssl"
   depends_on "fop" => :optional # enables building PDF docs
+  depends_on :java => :optional
```
DomT4: There's a mismatch here, you've called the option `with-javac` but the dependency is `java`. One won't trigger the other.

enilsen16: Isn't javac part of the JDK? I named it `with-javac` to just reflect the build flag that erlang provides....

DomT4: Yes, but `brew install erlang --with-javac` won't trigger the `java` dependency, because the latter doesn't grasp you meant to call it. It's an explicit match rather than word association, essentially. Probably easiest to just rename the option `with-java`, etc.

enilsen16:
@DomT4 I went ahead and renamed it to `with-java`. 😄

Formula/erlang.rb
```diff
@@ -87,6 +89,12 @@ def install
       args << "--enable-hipe"
     end
 
+    if build.with? "javac"
```
DomT4: Need to fix this as well, to match the option rename 🙈.

DomT4:
Last thing, but then we can finally ship this!

enilsen16:
Awesome! Thanks for the help @DomT4

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/4b4a1f5e54a832c0eb91e149e73680bc347cdbfb. Thank you for your contribution to Homebrew @enilsen16, and thank you for your patience on the back-and-forth; we appreciate it! :smiley_cat:

PR #1811, opened 2016-06-08
loudmouth 1.5.3

ilovezfs:
Also,
- add linked library dependency `gettext`
- run `make check`
- install the examples
- add a test

### DIFF
--- a/Formula/loudmouth.rb
+++ b/Formula/loudmouth.rb
@@ -1,9 +1,8 @@
 class Loudmouth < Formula
   desc "Lightweight C library for the Jabber protocol"
   homepage "https://mcabber.com"
-  url "https://mcabber.com/files/loudmouth/loudmouth-1.5.1.tar.bz2"
-  sha256 "ffb493b085c1d40176ecbe1c478f05932f265e0e5ba93444b87d3cd076267939"
-  revision 1
+  url "https://mcabber.com/files/loudmouth/loudmouth-1.5.3.tar.bz2"
+  sha256 "54329415cb1bacb783c20f5f1f975de4fc460165d0d8a1e3b789367b5f69d32c"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Formula/loudmouth.rb
```diff
@@ -24,11 +23,25 @@ class Loudmouth < Formula
...
     system "make", "install"
+    (pkgshare/"examples").install Dir["examples/*.c"]
+  end
+
+  test do
+    cp pkgshare/"examples/lm-send-async.c", testpath
+    system ENV.cc, "lm-send-async.c", "-o", "test",
+      "-lloudmouth-1", "-lglib-2.0",
+      "-I#{include}/loudmouth-1.0",
+      "-I#{Formula["glib"].include}/glib-2.0",
```
DomT4: Should possibly be `.opt_include` so we're consistent with a lot of other usage.

Formula/loudmouth.rb
```diff
@@ -24,11 +23,25 @@ class Loudmouth < Formula
...
+      "-I#{Formula["glib"].lib}/glib-2.0/include"
```
DomT4: (and `.opt_lib` here, as well, if the above is changed)

DomT4:
Couple of nits but LGTM otherwise.

ilovezfs:
What's the logic for the `opt_*` in `test do`? I'm interested if it breaks against the current version not the randomly installed possibly outdated version. During CI it seems they'd be identical. On a local machine, non-opt would avoid false negatives/positives caused by testing against a non-current version. Also, since it's a direct dependency in this case, it seems the current version would have to be the same as the opt version, but maybe I'm missing something that's good about opt here.

DomT4:
In testing it's not actually massively important, it's really more about consistency there than much else. `brew test xyz` is incredibly rarely ran (GA backs this up), and anything submitted as a PR will be tested by CI without that user configuration issue, so it's not something we need to worry about particularly.

PR #1812, opened 2016-06-08
stlink 1.2.0

ilovezfs:
Also,
- switch to CMake build system
- add option for gtk+3

### DIFF
--- a/Formula/stlink.rb
+++ b/Formula/stlink.rb
@@ -1,8 +1,8 @@
 class Stlink < Formula
   desc "stm32 discovery line Linux programmer"
   homepage "https://github.com/texane/stlink"
-  url "https://github.com/texane/stlink/archive/1.1.0.tar.gz"
-  sha256 "3ac4dfcf1da0da40a1b71a8789ff0f1e7d978ea0222158bebd2de916c550682c"
+  url "https://github.com/texane/stlink/archive/1.2.0.tar.gz"
+  sha256 "acfdd52e350a61c14910f3c14b9ed232a79febcf35b38479b011d5cd2d4af688"
 
   head "https://github.com/texane/stlink.git"
 
@@ -14,16 +14,13 @@ class Stlink < Formula
     sha256 "36ad00988ff74f08da8be3653ed7070b206ac53c9720e5c6e2a5476b2d1fa7cf" => :mavericks
   end
 
-  depends_on "libusb"
-
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
+  depends_on "cmake" => :build
   depends_on "pkg-config" => :build
+  depends_on "libusb"
+  depends_on "gtk+3" => :optional
 
   def install
-    system "./autogen.sh"
-    system "./configure", "--prefix=#{prefix}"
-    system "make"
+    system "cmake", ".", *std_cmake_args
     system "make", "install"
   end
 


Formula/stlink.rb
```diff
@@ -15,15 +15,13 @@ class Stlink < Formula
   end
 
   depends_on "libusb"
+  depends_on "gtk+3" => :optional
 
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
+  depends_on "cmake" => :build
```
DomT4: I get far more bothered by `:build` dependencies being below everything else than I should do 😆

DomT4:
Does the GTK+3 option need to be told to be off by default? CMake is kind of prone to just scooping things up from the environment, even when the option isn't explicitly invoked.

ilovezfs:
It seems to get this one right.
```
iMac-TMP:~ joe$ brew install -s stlink
==> Downloading https://github.com/texane/stlink/archive/1.2.0.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/stlink-1.2.0.tar.gz
==> cmake . -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFI
==> make install
🍺  /usr/local/Cellar/stlink/1.2.0: 13 files, 334.6K, built in 3 seconds
iMac-TMP:~ joe$ brew linkage stlink
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/libusb/lib/libusb-1.0.0.dylib (libusb)
```
vs.
```
iMac-TMP:~ joe$ brew install stlink --with-gtk+3
==> Downloading https://github.com/texane/stlink/archive/1.2.0.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/stlink-1.2.0.tar.gz
==> cmake . -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFI
==> make install
🍺  /usr/local/Cellar/stlink/1.2.0: 15 files, 459.7K, built in 4 seconds
iMac-TMP:~ joe$ brew linkage stlink
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/atk/lib/libatk-1.0.0.dylib (atk)
  /usr/local/opt/cairo/lib/libcairo-gobject.2.dylib (cairo)
  /usr/local/opt/cairo/lib/libcairo.2.dylib (cairo)
  /usr/local/opt/gdk-pixbuf/lib/libgdk_pixbuf-2.0.0.dylib (gdk-pixbuf)
  /usr/local/opt/gtk+3/lib/libgdk-3.0.dylib (gtk+3)
  /usr/local/opt/gtk+3/lib/libgtk-3.0.dylib (gtk+3)
  /usr/local/opt/libusb/lib/libusb-1.0.0.dylib (libusb)
  /usr/local/opt/pango/lib/libpango-1.0.0.dylib (pango)
  /usr/local/opt/pango/lib/libpangocairo-1.0.0.dylib (pango)
  atk
  cairo
  gdk-pixbuf
  gettext
  glib
  pango
```

DomT4:
Crumbs, miracles do happen. 👍 then.

Formula/stlink.rb
ilovezfs: I agree. But the real question is whether recommended must always go above optional and whether unconditional must always go above recommended :)

UniqMartin: There's always room for improvement in `brew audit`. :joy:

DomT4: Don't even tempt me to make this a mandatory audit check 😅 😈.

ilovezfs: I think it's an excellent idea :)

Also, while I'm thinking about pet peeves, what in the world is the rule concerning 

```
desc "blah"
etc. etc.

head "https://foourl"
```
OR
```
desc "blah"
etc. etc.
head "https://foourl"
```

and whether there should be a blank line above `head ...` or not?

There seems to be a silent war going on between the blank line faction and the no blank line faction.

DomT4: > Also, while I'm thinking about pet peeves, what in the world is the rule concerning

Is it worrying I instantly already have the answer to this in my mind? 😆.

The rule is generally if there's no formula `revision` there shouldn't be an empty line, if there is a formula `revision` there should be an empty line before the `head` to make it a little more obvious `revision`s don't apply to `head`.

ilovezfs: > Is it worrying I instantly already have the answer to this in my mind? :laughing:.

Only slightly.

>The rule is generally if there's no formula revision there shouldn't be an empty line, if there is a formula revision there should be an empty line before the head to make it a little more obvious revisions don't apply to head.

No wonder there's no consistency. LOL

UniqMartin: It's a fine rule, but it's not easily discoverable and I guess it's easy to forget to remove the empty line before `head` again when the `revision` line gets removed due to a version bump.

ilovezfs: I think it shows yet again how awkward it is having the revision line be something that appears and disappears.

DomT4: There's no way in hell we're adding `revision 0` to every formula in Homebrew 😜. If people want to harmonise the style on single `head` lines even further I'd settle on there always being an empty line between the initial block and `head`.

ilovezfs: >There's no way in hell we're adding revision 0 to every formula in Homebrew

Nor was that what I was implying would be the solution.

DomT4: Was only gently poking you there 😺

PR #1817, opened 2016-06-08
Chromedriver 2.22

rowanbeentje:

---

http://chromedriver.storage.googleapis.com/2.22/notes.txt

`brew audit` came up with an error so added what I hope the fix is...

### DIFF
--- a/Formula/chromedriver.rb
+++ b/Formula/chromedriver.rb
@@ -1,9 +1,9 @@
 class Chromedriver < Formula
   desc "Tool for automated testing of webapps across many browsers"
   homepage "https://sites.google.com/a/chromium.org/chromedriver/"
-  url "https://chromedriver.storage.googleapis.com/2.21/chromedriver_mac32.zip"
-  version "2.21"
-  sha256 "41f21005b498adc5896e4c3e264d718e4ade09609a7544ea8753d4e8ef1e787f"
+  url "https://chromedriver.storage.googleapis.com/2.22/chromedriver_mac32.zip"
+  version "2.22"
+  sha256 "086943786ce736c72f790d85bc810cd541d97abdf81d51e30da62a6d3e2152e3"
 
   bottle :unneeded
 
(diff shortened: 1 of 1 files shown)


rowanbeentje:
Looking more closely at the plist and the key settings therein, have opted for both `manual` and `startup` plist_options settings - but open to any guidance there!

DomT4:
Will trust you on the `plist` decision, if anyone thinks differently I'm sure they'll let us know 😄.

Merged in https://github.com/Homebrew/homebrew-core/commit/14c7f7415bd2feaf49def7ad53fe5f6eb32f9265. Thank you for your contribution to Homebrew @rowanbeentje; we appreciate it! :smiley_cat:

Formula/chromedriver.rb
```diff
@@ -1,16 +1,18 @@
...
+  version "2.22"
+  sha256 "086943786ce736c72f790d85bc810cd541d97abdf81d51e30da62a6d3e2152e3"
 
   bottle :unneeded
 
   def install
     bin.install "chromedriver"
   end
 
+  plist_options :startup => "true", :manual => "chromedriver"
```
MikeMcQuaid: @DomT4 For future reference: `:startup` implies it'll run as root on system boot (rather than login) which is probably not what we want here.

DomT4: Aware of the difference in variable, just not aware of whether or not it was applicable to this particular piece of software.

rowanbeentje: @MikeMcQuaid My fault!  I started off with just `:manual` in 7dd393f0b6d969ca9881e6a777bf7f0b76311479, but after looking at the pre-existing `plist` def and its `RunAtLoad` `<true/>`, I added the `:startup` to try and match.  Is the launchagent per-user but the plist_options a boot setting then?

http://www.rubydoc.info/github/Homebrew/brew/Formula.plist_options suggested that having both gave the user a choice, so I didn't think it was harmful... (but I am aware that's a third-party site - I struggled to find what the audit failure meant when I was just trying to update the version :( )

MikeMcQuaid: No worries, it's not a big deal, just wanted to reply to @DomT4 from earlier. We almost never want `:startup` unless it's some sort of system daemon that definitely needs to be running at all times (e.g. SSH protection)

PR #1824, opened 2016-06-08
bup 0.28

ilovezfs:
- version bump
- always run the tests if building a bottle
- add `without-web` option to enable `bup web` by default, which fixes
  `make test`; upstream is still working on fixing the test when tornado
  is not available
- The patch works around another `make test` failure caused by bup's not
  handling unicode normalization; unfortunately, the patch is not
  suitable for release, so we'll need to wait on a proper upstream fix.
- Switch to building from the tag to prevent the following:
  Makefile:6: config/config.vars: No such file or directory
  Makefile:43: **\* Version configuration failed.  Stop.

The problem building from a source archive tarball is known to upstream,
and has been addressed already in https://github.com/bup/bup/tree/tmp/fix-release-archives-v2

### DIFF
--- a/Formula/bup.rb
+++ b/Formula/bup.rb
@@ -1,8 +1,9 @@
 class Bup < Formula
   desc "Backup tool"
   homepage "https://github.com/bup/bup"
-  url "https://github.com/bup/bup/archive/0.27.tar.gz"
-  sha256 "ae8a744a3415ce4766d1d896f3d48b9b2ae6167c7bc65d8d1a112f37b42720fb"
+  url "https://github.com/bup/bup.git",
+    :tag => "0.28",
+    :revision => "017af24d11d45541dd20d48568588650cc5ed598"
 
   head "https://github.com/bup/bup.git"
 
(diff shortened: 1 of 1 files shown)


DomT4:
Do the tests take long? Would be nice to enable those by default, or at least for bottle generation.

ilovezfs:
@DomT4 I'm seeing the same test failure here. Odd that it succeeded on Mavericks.

ilovezfs:
Interesting. It also passes here on Mountain Lion. So this test failure seems to be a >= 10.10 regression.

ilovezfs:
@DomT4 Test failure turns out to be a regression caused by https://github.com/bup/bup/commit/506eaeb0d32a240c1056c2168b9b6f335d034951 but it's not clear why it only affects >= 10.10

ilovezfs:
@DomT4 I've reported the issue upstream: https://groups.google.com/forum/#!topic/bup-list/caoQLpqrVPk

ilovezfs:
PR refreshed. The patch :DATA is not suitable for release, so this will need to wait on 0.28.1, or we can remove the `bup web` test (namely, `test-web.sh`) from `make check` as well as remove the web option and the vendored tornado in the meantime.

ilovezfs:
:green_apple: now, but since the patch is not suitable to ship, closing.

PR #1826, opened 2016-06-08
docx2txt 1.4 (new formula)

proinsias:
  - <s>I'm getting an error about an "Empty installation"</s> (resolved)

<s>I've also added a test, but when I run it I get:</s>

<s>`Error: Testing requires the latest version of docx2txt` </s>
<s>`Check if </tmp/docx2txt-test-20160617-27321-1dgayqb/demo.docx> exists, is readable and has non-zero size!`</s>

<s>It seems as if the `demo.docx` is not copied to the `testpath`.

Any guidance on these issues would be appreciated.</s> (resolved)

### DIFF
--- a/Formula/docx2txt.rb
+++ b/Formula/docx2txt.rb
@@ -1,0 +1,20 @@
+class Docx2txt < Formula
+  desc "Converts Microsoft Office docx documents to equivalent text documents"
+  homepage "http://docx2txt.sourceforge.net/"
+  url "https://downloads.sourceforge.net/docx2txt/docx2txt-1.4.tgz"
+  sha256 "b297752910a404c1435e703d5aedb4571222bd759fa316c86ad8c8bbe58c6d1b"
+
+  resource "sample_doc" do
+    url "http://calibre-ebook.com/downloads/demos/demo.docx", :using => :nounzip
+    sha256 "269329fc7ae54b3f289b3ac52efde387edc2e566ef9a48d637e841022c7e0eab"
+  end
+
+  def install
+    system "make", "install", "CONFIGDIR=#{etc}", "BINDIR=#{bin}"
+  end
+
+  test do
+    testpath.install resource("sample_doc")
+    system "#{bin}/docx2txt.sh", "#{testpath}/demo.docx"
+  end
+end


Formula/docx2txt.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "perl"
+  depends_on "unzip"
+  depends_on "bash" => :optional
+
+  def install
+    system "make", "install", "CONFIGDIR=/usr/local/etc"
+  end
+
+  test do
+    system "wget", "http://calibre-ebook.com/downloads/demos/demo.docx"
```
dunn: This should be a vendored `resource` to enable checksumming; see [`aften.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/aften.rb) for an example.

dunn: This should be a vendored `resource` to enable checksumming; see [`aften.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/aften.rb) for an example.

Formula/docx2txt.rb
```diff
@@ -0,0 +1,19 @@
...
+    system "docx2txt.sh", "demo.docx"
```
dunn: Use `bin/"docx2txt.sh"` to enure it's using the one we just installed.

dunn:
> Error: Testing requires the latest version of docx2txt

Usually that means it's not installed locally.  Did you uninstall it before trying to run the test?

Formula/docx2txt.rb
```diff
@@ -0,0 +1,19 @@
+class Docx2txt < Formula
+  desc "Converts Microsoft Office docx documents to equivalent text documents"
+  homepage "http://docx2txt.sourceforge.net/"
+  url "http://downloads.sourceforge.net/docx2txt/docx2txt-1.4.tgz"
+  sha256 "b297752910a404c1435e703d5aedb4571222bd759fa316c86ad8c8bbe58c6d1b"
+
+  depends_on "perl"
```
DomT4: Please don't hardcode our `perl`. If it needs a version of Perl a release of OS X doesn't ship use:

```ruby
depends_on :perl => "5.14" if MacOS.version <= :mavericks
```
For example, bending that style of syntax to your needs. If it's fine with OS X's shipped `perl` on OS X as far as you know, you can just remove the dependency completely.

Formula/docx2txt.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "unzip"
+  depends_on "bash" => :optional
```
DomT4: What does this change if it is built with our `bash` rather than the system's?

Formula/docx2txt.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def install
+    system "make", "install", "CONFIGDIR=/usr/local/etc"
```
DomT4: Don't presume `/usr/local`. Should be `CONFIGDIR=#{etc}`

proinsias:
> > Error: Testing requires the latest version of docx2txt

> Usually that means it's not installed locally. Did you uninstall it before trying to run the test?

No, though when I try to uninstall, I get `Error: No such keg: /usr/local/Cellar/docx2txt`.

Formula/docx2txt.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  depends_on "unzip"
+
+  resource "sample_doc" do
+    url "http://calibre-ebook.com/downloads/demos/demo.docx"
+    sha256 "269329fc7ae54b3f289b3ac52efde387edc2e566ef9a48d637e841022c7e0eab"
+  end
+
+  def install
+    system "make", "install", "CONFIGDIR=#{etc}"
```
DomT4: I suspect your errors in running `brew test` here stem from the fact you've not set a prefix here, so it'll be installing into the default prefix rather than the Cellar as intended.

dunn: That would also explain the `Error: No such keg: /usr/local/Cellar/docx2txt` error.

proinsias:
Thanks for the tips – this is my first homebrew submission!

Now I'm just left with a testing error. When I run the test I get:

    Check if </tmp/docx2txt-test-20160617-27321-1dgayqb/demo.docx> exists, is readable and has non-zero size!

It seems as if the `demo.docx` is not copied to the `testpath`. I followed the pattern from the `aften.rb` formula. Any ideas?

Formula/docx2txt.rb
```diff
@@ -0,0 +1,22 @@
...
+    url "http://calibre-ebook.com/downloads/demos/demo.docx"
+    sha256 "269329fc7ae54b3f289b3ac52efde387edc2e566ef9a48d637e841022c7e0eab"
+  end
+
+  def install
+    system "make", "install", "CONFIGDIR=#{etc}", "BINDIR=#{bin}"
+  end
+
+  test do
+    resource("sample_doc").stage testpath
```
dunn: This should be `testpath.install resource("sample_doc")`

proinsias:
I've changed the method of installing the test resource to `testpath.install resource("sample_doc")`, but still no luck. I've also added a `assert File.exist? "#{testpath}/demo.docx"` statement to verify the file is there. This is now where the test fails – so I'm still not copying the test resource properly.

dunn:
Oh, I see the problem.  Homebrew is trying to unzip the `docx`.  You can prevent that though:
```diff
diff --git a/Formula/docx2txt.rb b/Formula/docx2txt.rb
index 8c2d495..9886c33 100644
--- a/Formula/docx2txt.rb
+++ b/Formula/docx2txt.rb
@@ -7,7 +7,7 @@ class Docx2txt < Formula
   depends_on "unzip"

   resource "sample_doc" do
-    url "http://calibre-ebook.com/downloads/demos/demo.docx"
+    url "http://calibre-ebook.com/downloads/demos/demo.docx", :using => :nounzip
     sha256 "269329fc7ae54b3f289b3ac52efde387edc2e566ef9a48d637e841022c7e0eab"
   end
```
Then you can remove the `File.exist?` statements.

proinsias:
I've updated. Thanks for the help!

dunn:
The test failure now is because (and I should have realized this) `unzip` is not a core formula: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/3420/version=el_capitan/console

It's from the homebrew-dupes tap, and we don't allow cross-tap dependencies in core.  Why can't we use the system `unzip`?

proinsias:
Removed dependency on `unzip`.

dunn:
Thanks! We're so close:
```
        Error: 1 problem in 1 formula
docx2txt:
  * Stable: Please use https:// for http://downloads.sourceforge.net/docx2txt/docx2txt-1.4.tgz

proinsias:
Fixed.

dunn:
Thanks! Merged as 449314903243247eab5630d0c3918d3a5a914768.

PR #1836, opened 2016-06-09
watchman 4.6.0-rc2 (devel)

ilovezfs:
- add devel spec
- make test work for both stable and devel
- set statedir to a directory outside of the prefix

### DIFF
--- a/Formula/watchman.rb
+++ b/Formula/watchman.rb
@@ -11,6 +11,12 @@ class Watchman < Formula
     sha256 "74c19390f79a0ca19faa5290f1e494cda14390606bb891a43cf8957d0098a3dd" => :mavericks
   end
 
+  devel do
+    url "https://github.com/facebook/watchman/archive/v4.6.0-rc2.tar.gz"
+    version "4.6.0-rc2"
+    sha256 "2b8cae97615cfa83178fdf0cc1e320a6b4c9277b07217f2c642ae240cc39be19"
+  end
+
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "autoconf" => :build
   depends_on "automake" => :build
@@ -23,7 +29,8 @@ class Watchman < Formula
                           "--with-pcre",
                           # we'll do the homebrew specific python
                           # installation below
-                          "--without-python"
+                          "--without-python",
+                          "--enable-statedir=#{var}/run/watchman"
     system "make"
     system "make", "install"
 
@@ -36,10 +43,12 @@ class Watchman < Formula
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
   end
 
+  def post_install
+    (var/"run/watchman").mkpath
+    chmod 042777, var/"run/watchman"
+  end
+
   test do
-    list = shell_output("#{bin}/watchman -v")
-    if list.index(version).nil?
-      raise "expected to see #{version} in the version output"
-    end
+    assert_equal /(\d+\.\d+\.\d+)/.match(version)[0], shell_output("#{bin}/watchman -v").chomp
   end
 end


Formula/watchman.rb
```diff
@@ -23,7 +29,8 @@ def install
                           "--with-pcre",
                           # we'll do the homebrew specific python
                           # installation below
-                          "--without-python"
+                          "--without-python",
+                          "--enable-statedir=#{var}/run/watchman"
```
DomT4: We'll want to make sure this is created in `post_install`.

DomT4:
Single nit but 👍 otherwise.

Formula/watchman.rb
ilovezfs: Seems to create itself on demand as needed ...

ilovezfs: Actually, it's created by the formula at build time. So not sure what to do about that. If `/usr/local/var` exists at the beginning and it creates `/usr/local/var/run` and `/usr/local/var/run/watchman` during the build, do we then need to `rm -rf /usr/local/var/run` at the end of the build in `install`, and then `mkdir_p /usr/local/var/run/watchman` during `post_install`? But if at the beginning `/usr/local/var/run` already existed, and during `install` the build only creates `/usr/local/var/run/watchman`, then do we need to `rm -rf  /usr/local/var/run/watchman` only and then during post_install `mkdir /usr/local/var/run/watchman`?

DomT4: It it's created during build that's pretty much fine. The problems tend to start when it isn't and then the executable is called with root, either via `sudo` or `launchctl` and that directory becomes unwritable for Homebrew.

ilovezfs: This is what you get by the end of `install` if neither `/usr/local/var/run` nor `/usr/local/var/run/watchman` exist at the beginning:
```
$ ls -ld /usr/local/var/run
drwxr-xr-x  3 joe  admin  102 Jun  9 06:56 /usr/local/var/run
$ ls -ld /usr/local/var/run/watchman
drwxrwsrwx  2 joe  admin  102 Jun  9 06:56 /usr/local/var/run/watchman
```

DomT4: Seems alright as-is then 👍.

ilovezfs: I'll check to make sure it works via bottle install.

ilovezfs: Interestingly the bottle screws up the permissions:
```
$ ls -ld /usr/local/var/run
drwxr-xr-x  3 joe  admin  102 Jun  9 07:06 /usr/local/var/run
$ ls -ld /usr/local/var/run/watchman
drwxr-xr-x  2 joe  admin  68 Jun  9 07:06 /usr/local/var/run/watchman
```

DomT4: I suspect it's attempting to "be helpful".

ilovezfs: I guess we want
```
  def post_install
    (var/"run/watchman").mkpath
    chmod "uo=rwx,g=rwxs", var/"run/watchman"
  end
```

ilovezfs:
```
iMac-TMP:local joe$ brew install ./watchman-4.6.0-rc1.el_capitan.bottle.tar.gz 
==> Pouring watchman-4.6.0-rc1.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/watchman/4.6.0-rc1: 21 files, 387.5K
iMac-TMP:local joe$ ls -ld /usr/local/var/run
drwxr-xr-x  3 joe  admin  102 Jun  9 07:23 /usr/local/var/run
iMac-TMP:local joe$ ls -ld /usr/local/var/run/watchman
drwxrwsrwx  2 joe  admin  68 Jun  9 07:23 /usr/local/var/run/watchman
```

ilovezfs: OK, added a `post_install` method.

ilovezfs: which of course causes CI failure.

ilovezfs: OK, using 042777 instead of "uo=rwx,g=rwxs" should make it happy on Mavericks CI, too.

DomT4: Smells like it's worth hunting why we change permissions on bottled `var` and whether doing so has any downsides.

ilovezfs: I agree. Bottles should preserve the fidelity of the permissions of what's installed (of course, including whatever permissions "fixes" brew may have done during install).

PR #1837, opened 2016-06-09
iso-codes 3.68

ilovezfs:
Also,
- fix audit error about the description length
- update the head url
- add :python3 requirement since python3 is now invoked by make
- run make check
- add run-time dependency on pkg-config
- add a test

### DIFF
--- a/Formula/iso-codes.rb
+++ b/Formula/iso-codes.rb
@@ -1,10 +1,10 @@
 class IsoCodes < Formula
-  desc "ISO language, territory, currency, script codes, and their translations"
+  desc "Provides lists of various ISO standards"
   homepage "https://pkg-isocodes.alioth.debian.org/"
-  url "https://pkg-isocodes.alioth.debian.org/downloads/iso-codes-3.64.tar.xz"
-  sha256 "5ef061381e37e9576760df1ad504d4bbc84c270da30512b2891baed9add70729"
-
-  head "git://git.debian.org/git/iso-codes/iso-codes.git", :shallow => false
+  url "https://pkg-isocodes.alioth.debian.org/downloads/iso-codes-3.68.tar.xz"
+  mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/i/iso-codes/iso-codes_3.68.orig.tar.xz"
+  sha256 "5881cf7caa5adfffb14ade99138949324c28a277babe8d07dafbff521acef9d1"
+  head "https://anonscm.debian.org/git/pkg-isocodes/iso-codes.git"
 
   bottle do
     cellar :any_skip_relocation
@@ -14,9 +14,19 @@ class IsoCodes < Formula
   end
 
   depends_on "gettext" => :build
+  depends_on :python3
+  depends_on "pkg-config" => :run
 
   def install
     system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "check"
     system "make", "install"
+  end
+
+  test do
+    pkg_config = Formula["pkg-config"].opt_bin/"pkg-config"
+    output = shell_output("#{pkg_config} --variable=domains iso-codes")
+    assert_match "iso_639-2 iso_639-3 iso_639-5 iso_3166-1", output
   end
 end


Formula/iso-codes.rb
```diff
@@ -1,10 +1,11 @@
...
+  desc "Provides lists of various ISO standards"
   homepage "https://pkg-isocodes.alioth.debian.org/"
-  url "https://pkg-isocodes.alioth.debian.org/downloads/iso-codes-3.64.tar.xz"
-  sha256 "5ef061381e37e9576760df1ad504d4bbc84c270da30512b2891baed9add70729"
+  url "https://pkg-isocodes.alioth.debian.org/downloads/iso-codes-3.68.tar.xz"
+  mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/i/iso-codes/iso-codes_3.68.orig.tar.xz"
+  sha256 "5881cf7caa5adfffb14ade99138949324c28a277babe8d07dafbff521acef9d1"
 
-  head "git://git.debian.org/git/iso-codes/iso-codes.git", :shallow => false
+  head "git://anonscm.debian.org/pkg-isocodes/iso-codes.git", :shallow => false
```
DomT4: Switching to `https://` might work here?

```
https://anonscm.debian.org/git/pkg-isocodes/iso-codes.git
```

Formula/iso-codes.rb
```diff
@@ -14,9 +15,17 @@ class IsoCodes < Formula
...
 
   def install
     system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "check"
     system "make", "install"
   end
+
+  test do
+    assert_match /iso_639-2 iso_639-3 iso_639-5 iso_3166-1/, shell_output("#{Formula["pkg-config"].opt_bin}/pkg-config --variable=domains iso-codes")
```
DomT4: Mind wrapping this a bit closer to the 80 characters?

```ruby
output = shell_output("#{Formula["pkg-config"].opt_bin}/pkg-config --variable=domains iso-codes")
assert_match "iso_639-2 iso_639-3 iso_639-5 iso_3166-1", output
```
etc.

Formula/iso-codes.rb
ilovezfs: Indeed.

DomT4:
Couple of nits but 👍 otherwise.

Formula/iso-codes.rb
ilovezfs: not sure I follow

DomT4: The current test stretches the formula out quite a bit. Just asking that we split it onto the two lines like that to make sure it isn't stretched beyond 83 characters in the test rather than, er, 145. Helps with readability.

ilovezfs: But I'm morally opposed to useless temporary variables :)

DomT4: I'm morally opposed to our formulae being ridiculously difficult to read at a glance 😉

DomT4: Plus at 140+ examples of that usage in the core the ship may have sailed on objection here 😄.

ilovezfs: Can we do this please?
```
    pkg_config = "pkg-config"
    pkg_config_formula = Formula[pkg_config]
    pkg_config_opt_bin = pkg_config_formula.opt_bin
    pkg_config_prog = pkg_config_opt_bin/pkg_config
    args = %w[--variable=domains iso-codes]
    separator = " "
    command = "#{pkg_config_prog} #{args.join(separator)}"
    output = shell_output(command)
    match = /iso_639-2 iso_639-3 iso_639-5 iso_3166-1/
    assert_match match, output
```

DomT4: Yes, that's clearly proportional to my suggestion that we add a single extra line used in 140+ other places.

ilovezfs: I'm glad you agree!

DomT4: One day you'll succeed in giving me a brain aneurysm.

ilovezfs: Hmm... I'm obviously not trying hard enough.

DomT4: You're plenty trying, fret not 😜.

Formula/iso-codes.rb
```diff
@@ -14,9 +14,18 @@ class IsoCodes < Formula
...
 
   def install
     system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "check"
     system "make", "install"
   end
+
+  test do
+    output = shell_output("#{Formula["pkg-config"].opt_bin}/pkg-config --variable=domains iso-codes")
```
ilovezfs: Hmm this still wraps onto the next line. Surely we can do better.

DomT4: 🙈 It's fine.

ilovezfs: I'm quite disappointed in your lack of standards today.

DomT4: What can I say, sometimes I try to be nice to you 😉.

ilovezfs: :blush:

Formula/iso-codes.rb
```diff
@@ -14,9 +14,18 @@ class IsoCodes < Formula
...
+    assert_match /iso_639-2 iso_639-3 iso_639-5 iso_3166-1/, output
```
UniqMartin: Please use a string (as already suggested by @DomT4) instead of a regular expression when a simple substring match is sufficient.

Formula/iso-codes.rb
UniqMartin: Surely we can introduce `pkg_config` to fix this:

```ruby
pkg_config = Formula["pkg-config"].opt_bin/"pkg-config"
```

Formula/iso-codes.rb
UniqMartin: Very entertaining! 🎭 On a more serious note: You realize that you can wrap function arguments across multiple lines when temporary variables are undesired?

ilovezfs: Of course, but is that permissible style?

Formula/iso-codes.rb
ilovezfs: I like it.

Formula/iso-codes.rb
UniqMartin: Don't know if it is permissible in formulae, but we do it all the time in other Ruby code (and in Homebrew test code), thus I assume it's at least reasonable.

PR #1842, opened 2016-06-09
git-stree: added deprecation warning and fixed line order

drothmaler:
According to [it's homepage](https://github.com/deliciousinsights/git-stree#warning-this-project-is-deprecated) `git-stree` is no longer maintained and has been deprecated in favor of `git-subrepo`. 
So I added a warning to the caveats of this formula, to inform (new) users about it.

### DIFF
--- a/Formula/git-stree.rb
+++ b/Formula/git-stree.rb
@@ -1,9 +1,9 @@
 class GitStree < Formula
   desc "Git subtree helper command"
   homepage "https://github.com/tdd/git-stree"
-  head "https://github.com/tdd/git-stree.git"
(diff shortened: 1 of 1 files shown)


DomT4:
We'd probably rather just boneyard the formula, to be honest.

drothmaler:
That's fine to me... But I'm not the original maintainer. I just thought, this would be a nice way to provide a migration path to the users. But **I** would not care about it's removal.

MikeMcQuaid:
Thanks for the PR but passing on this due to reasons above.

PR #1858, opened 2016-06-09
goaccess 1.0.2

ilovezfs:
- stable spec also needs Autotools now
- remove glib and pkg-config dependencies
- enable utf8 support
- update the test

### DIFF
--- a/Formula/goaccess.rb
+++ b/Formula/goaccess.rb
@@ -1,8 +1,9 @@
 class Goaccess < Formula
   desc "Log analyzer and interactive viewer for the Apache Webserver"
   homepage "https://goaccess.io/"
-  url "http://tar.goaccess.io/goaccess-0.9.8.tar.gz"
-  sha256 "45bea0a2d87130f23869a26b103eb1b752a2b4a5b0f6f210cb6ffe8d7b2a18f1"
+  url "http://tar.goaccess.io/goaccess-1.0.2.tar.gz"
+  sha256 "58a83e201f29f0127b89032fbd40677558a643f6141b8c8413afd1e182f104f1"
+  head "https://github.com/allinurl/goaccess.git"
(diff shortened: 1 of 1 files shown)


ilovezfs:
So just how much do we all hate this approach?

UniqMartin:
It's not technically patching, but the `open_memstream` workaround feels pretty invasive and I wonder whether there's a better and cleaner solution for that …

ilovezfs:
Upstream is aware of the issue per the issue report link above, and is deciding how to proceed, but in the meantime this seemed like the most straightforward workaround to me. Also, this worked equally well: http://piumarta.com/software/memstream

ilovezfs:
For what it's worth, the breaking commit https://github.com/allinurl/goaccess/commit/487ef906f8a7c7a7bff807c252c60c4fe2f537ca did not `git revert` cleanly. But maybe you have another idea?

UniqMartin:
If the currently available version doesn't have any issues, how about just waiting a little bit longer until upstream has had a fair chance to ponder this a bit more and maybe even come up with an official fix?

ilovezfs:
@UniqMartin we could do that, but upstream seemed to be happy with the PR in the interim ...

ilovezfs:
Time wasted, I guess. Moving on.

UniqMartin:
:+1: Thanks for your patience and letting upstream fix this and create a new release.

ilovezfs:
@UniqMartin https://github.com/allinurl/goaccess/issues/422#issuecomment-225333338, and following :)

It's nice to have upstreams that care about their software.

UniqMartin:
:tada:

> It's nice to have upstreams that care about their software.

Indeed. That reads a lot like a constructive and respectful conversation. :)

PR #1861, opened 2016-06-09
geckodriver 0.8.0, rename from wires

leonyu:
- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?

---

`wires` has been adopted by Mozilla and renamed `geckodriver`

### DIFF
--- a/Formula/geckodriver.rb
+++ b/Formula/geckodriver.rb
@@ -1,4 +1,4 @@
-class Wires < Formula
+class Geckodriver < Formula
   desc "WebDriver <-> Marionette proxy"
   homepage "https://github.com/mozilla/geckodriver"
   url "https://github.com/mozilla/geckodriver/archive/v0.8.0.tar.gz"
--- a/Formula/wires.rb
+++ b/Formula/wires.rb
@@ -1,8 +1,8 @@
 class Wires < Formula
   desc "WebDriver <-> Marionette proxy"
-  homepage "https://github.com/jgraham/wires"
-  url "https://github.com/jgraham/wires/archive/v0.6.2.tar.gz"
-  sha256 "bbc73c1014dd218b424a82da6e656f638aa672fd0990229eff8c1a005166db07"
+  homepage "https://github.com/mozilla/geckodriver"
+  url "https://github.com/mozilla/geckodriver/archive/v0.8.0.tar.gz"
+  sha256 "a3b2cc5fd48622cdc811f952565ab235e5b64b5776e1f03c889a5c5c02367e1a"
 
   bottle do
     cellar :any_skip_relocation
@@ -15,10 +15,11 @@ class Wires < Formula
 
   def install
     system "cargo", "build"
-    bin.install "target/debug/wires"
+    bin.install "target/debug/geckodriver"
+    bin.install_symlink bin/"geckodriver" => "wires"
   end
 
   test do
-    system "#{bin}/wires", "--help"
+    system "#{bin}/geckodriver", "--help"
   end
 end
--- a/formula_renames.json
+++ b/formula_renames.json
@@ -22,5 +22,6 @@
   "offline-imap": "offlineimap",
   "plt-racket": "racket",
   "polarssl": "mbedtls",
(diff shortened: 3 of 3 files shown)


leonyu:
`brew audit --strict --online geckodriver` is complaining that `geckodriver` "is not notable enough", even though earlier version of it as `wires` was a formula and that it is now a Mozilla project.

BrianHawley:
The geckodriver project has precompiled binaries as well. Can there be a bottle install too, without rust?

DomT4:
@leonyu We can ignore the `audit` failure here since technically speaking it's a formula that already exists. Would you mind splitting your commits up a little bit though, to help distinguish the rename from the version bump in the git history?

leonyu:
Would you prefer me do the rename first and then up the version, or the reverse? 

I think it's more logical to me to update the version and then rename.

DomT4:
No strong preference. The important thing is making the rename more obvious in the history, whether that's before or after the version change happy to leave to you.

Aliases/wires
```diff
@@ -0,0 +1 @@
+../Formula/geckodriver.rb
```
UniqMartin: You don't need to add an alias. The information is extracted from `formula_renames.json` and the formula will still be accessible via its old name without the alias.

Formula/geckodriver.rb
```diff
@@ -15,10 +15,12 @@ class Wires < Formula
...
   def install
     system "cargo", "build"
-    bin.install "target/debug/wires"
+    bin.install "target/debug/geckodriver"
+    bin.install_symlink bin/"geckodriver" => "wires"
   end
 
   test do
     system "#{bin}/wires", "--help"
+    system "#{bin}/geckodriver", "--help"
```
UniqMartin: These are effectively the same file (you're symlinking one to the other in the `install` method). Thus it seems unnecessary to test both. I'd suggest dropping the old test in favor of the new one.

leonyu:
i have updated the PR to only include version update and use the compiled binary. I will have a follow up PR to rename it.

DomT4:
Please don't switch over to the upstream binary. We don't support those unless it's something like a `.jar`. We'd prefer to keep building from source.

leonyu:
Ah, OK, it was suggested by the other reviewer which I thought would be preferable. I will switch it back tomorrow.

leonyu:
Updated. I added a symlink to the older name because according to documentation: 

> Even though the project has been renamed to GeckoDriver, many of the selenium clients look for the old name. You need to rename the binary file to 'wires' (the old name) and ensure it is executable.

Formula/wires.rb
```diff
@@ -1,8 +1,9 @@
 class Wires < Formula
   desc "WebDriver <-> Marionette proxy"
-  homepage "https://github.com/jgraham/wires"
-  url "https://github.com/jgraham/wires/archive/v0.6.2.tar.gz"
-  sha256 "bbc73c1014dd218b424a82da6e656f638aa672fd0990229eff8c1a005166db07"
+  homepage "https://github.com/mozilla/geckodriver"
+  url "https://github.com/mozilla/geckodriver/archive/v0.8.0.tar.gz"
+  version "0.8.0"
```
UniqMartin: You can drop this line. The version is automatically inferred from the `url` field.

UniqMartin:
This is no longer renaming the formula, just updating it and adopting the new binary name. Any reason why you decided to do this, @leonyu? I believe there was no objection to the rename and we just asked you to do the rename in one commit and the update to the latest version in a separate commit.

leonyu:
oh, i though i was going need to make a separate PR. I misread.

leonyu:
pushed

UniqMartin:
Merged in 2b07bd4af2bbf8548ac5c3328520c995ccda65dd and 6145914a98d108b7b83e10549f07cff2d4545434. Thank you for this contribution to Homebrew, @leonyu! :tada:

PR #1862, opened 2016-06-09
govendor 1.0.3 (new formula)

DomT4:

---

New formula for https://github.com/kardianos/govendor. Adding in the naive hope developers discover this formula & consider migrating their go projects over to the vendor system ❤️.

`gvt` by Filippo is a pretty solid alternative too, but doesn't do tagged releases.

### DIFF
--- a/Formula/govendor.rb
+++ b/Formula/govendor.rb
@@ -1,0 +1,35 @@
+class Govendor < Formula
+  desc "Go vendor tool that works with the standard vendor file."
+  homepage "https://github.com/kardianos/govendor"
+  url "https://github.com/kardianos/govendor/archive/v1.0.3.tar.gz"
+  sha256 "7d7a032355ee4117822150db72a9db745d9a147a2e32972e58a65fe7545444f2"
+  head "https://github.com/kardianos/govendor.git"
+
+  depends_on "go"
+
+  def install
+    ENV["GOPATH"] = buildpath
+    ENV["GOOS"] = "darwin"
+    ENV["GOARCH"] = MacOS.prefer_64_bit? ? "amd64" : "386"
+
+    (buildpath/"src/github.com/kardianos/").mkpath
+    ln_sf buildpath, buildpath/"src/github.com/kardianos/govendor"
+    system "go", "build", "-o", bin/"govendor"
+  end
+
+  test do
+    # Default HOMEBREW_TEMP is /tmp, which is actually a symlink to /private/tmp.
+    # `govendor` bails without `.realpath` as it expects $GOPATH to be "real" path.
+    ENV["GOPATH"] = testpath.realpath
+    commit = "89d9e62992539701a49a19c52ebb33e84cbbe80f"
+    (testpath/"src/github.com/project/testing").mkpath
+
+    cd "src/github.com/project/testing" do
+      system bin/"govendor", "init"
+      assert File.exist?("vendor"), "Failed to init!"
+      system bin/"govendor", "fetch", "-tree", "golang.org/x/crypto@#{commit}"
+      assert_match commit, File.read("vendor/vendor.json")
+      assert_match "golang.org/x/crypto/blowfish", shell_output("#{bin}/govendor list")
+    end
+  end
+end


Formula/govendor.rb
```diff
@@ -0,0 +1,39 @@
...
+  url "https://github.com/kardianos/govendor/archive/v1.0.3.tar.gz"
+  sha256 "7d7a032355ee4117822150db72a9db745d9a147a2e32972e58a65fe7545444f2"
+  head "https://github.com/kardianos/govendor.git"
+
+  depends_on "go"
+
+  def install
+    ENV["GOPATH"] = buildpath
+    ENV["GOOS"] = "darwin"
+    ENV["GOARCH"] = MacOS.prefer_64_bit? ? "amd64" : "386"
```
ilovezfs: Is there a universal option in their build system?

DomT4: I'm not sure universal binaries are a concept `go` even grasps. Don't think I've ever seen a universal go-generated executable.

Formula/govendor.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    (buildpath/"src/github.com/kardianos/").mkpath
+    ln_sf buildpath, buildpath/"src/github.com/kardianos/govendor"
+    system "go", "build", "-o", bin/"govendor"
+  end
+
+  test do
+    # Can't run directly in testpath because `govendor` expects $GOPATH to be
+    # a "real" path & panics that /tmp is actually symlink to /private/tmp.
```
ilovezfs: It's not clear to me how the below changed that at all, so I'm not sure that's what actually going on. In any case, the directory handling looks much more complicated than it needs to be.

DomT4: `testpath` resolves to `/tmp/blah` whereas `pwd` resolves to `/private/tmp/blah`, which is the "true" location and consequently makes `govendor` happy.

You're dealing with `go`. If you wanted simple, it's the wrong language 😄.

ilovezfs: I don't think ruby `cd` understands that distinction. You probably want to use realpath here and avoid unneeded directory changes that don't do anything.

DomT4: I actually used `realpath` initially, but I wasn't sure how robust that would be across every possible `HOMEBREW_TEMP`, whereas this shouldn't break either way. I'm happy to go back to `.realpath` but if it fails I'll be looking at you to cover the cost of the mug of whiskey my patience will need to reassemble this 😉.

ilovezfs: I'd be less vague if I weren't AFK right now, but yeah realpath is your friend.

DomT4: I'll trust you. If you're right, we get to save a few lines. If you're wrong I get to gloat, it's a win-win.

Formula/govendor.rb
```diff
@@ -0,0 +1,36 @@
...
+    (buildpath/"src/github.com/kardianos/").mkpath
+    ln_sf buildpath, buildpath/"src/github.com/kardianos/govendor"
+    system "go", "build", "-o", bin/"govendor"
+  end
+
+  test do
+    # Default HOMEBREW_TEMP is /tmp, which is actually a symlink to /private/tmp.
+    # `govendor` bails without `.realpath` as it expects $GOPATH to be "real" path.
+    ENV["GOPATH"] = testpath.realpath
+    (testpath/"src/github.com/project/testing").mkpath
```
ilovezfs: You can now probably use the ln_sf idiom instead of mkpath+cd

DomT4: I personally find our `ln_sf` style as clear as mud in terms of what is getting executed where & when, at a glance.

ilovezfs: We can certainly consider changing all of them to something else, but I know I don't need to preach the merits of consistency in the meantime to you.

ilovezfs: Yeah, not sure the links can work in this case. How do you feel about this:
```
  test do
    ENV["GOPATH"] = testpath.realpath
    repo = "golang.org/x/crypto"
    commit = "89d9e62992539701a49a19c52ebb33e84cbbe80f"
    (project = testpath/"src/github.com/testing/project").mkpath
    cd project do
      system bin/"govendor", "init"
      system bin/"govendor", "fetch", "-tree", "#{repo}@#{commit}"
      assert_match commit, File.read("vendor/vendor.json")
      assert_match "#{repo}/blowfish", shell_output("#{bin}/govendor list")
    end
  end
```

ilovezfs: Or with one more variable,
```
test do
  ENV["GOPATH"] = testpath.realpath
  gv = bin/"govendor"
  repo = "golang.org/x/crypto"
  commit = "89d9e62992539701a49a19c52ebb33e84cbbe80f"
  (testing = testpath/"src/github.com/project/testing").mkpath
  cd testing do
    system gv, "init"
    system gv, "fetch", "-tree", "#{repo}@#{commit}"
    assert_match commit, File.read("vendor/vendor.json")
    assert_match "#{repo}/blowfish", shell_output("#{gv} list")
  end
end
```

ilovezfs: I wonder if we should extend the `Pathname.mkpath` method to accept a block. That would certainly make this even nicer:
```
test do
  ENV["GOPATH"] = testpath.realpath
  gv = bin/"govendor"
  repo = "golang.org/x/crypto"
  commit = "89d9e62992539701a49a19c52ebb33e84cbbe80f"
  (testpath/"src/github.com/project/testing").mkpath do
    system gv, "init"
    system gv, "fetch", "-tree", "#{repo}@#{commit}"
    assert_match commit, File.read("vendor/vendor.json")
    assert_match "#{repo}/blowfish", shell_output("#{gv} list")
  end
end
```

ilovezfs: Or even just allow mkdir_p to take a block,
```
mkdir_p testpath/"src/github.com/project/testing" do
  etc.
end
```

ilovezfs: Or if you'd like to keep it mostly as is, which I think is fine, I'd probably at least factor out `commit`:
```
test do
  ENV["GOPATH"] = testpath.realpath
  commit = "89d9e62992539701a49a19c52ebb33e84cbbe80f"
  (testpath/"src/github.com/project/testing").mkpath
  cd testpath/"src/github.com/project/testing" do
    system bin/"govendor", "init"
    system bin/"govendor", "fetch", "-tree", "golang.org/x/crypto@#{commit}"
    assert_match commit, File.read("vendor/vendor.json")
    assert_match "golang.org/x/crypto/blowfish", shell_output("#{bin}/govendor list")
  end
end
```

ilovezfs:
:tada: Looks good! Thanks for making the changes.

DomT4:
No worries. Agree `mkpath` being able to take a block in the way `mkdir` can would be useful.

ilovezfs:
@UniqMartin Any thoughts on `(mypathname/"foo/bar").mkpath do` vs. `mkdir_p mypathname/"foo/bar" do` vs. both vs. neither?

As a hack I was also thinking about `cd (mypathname/"foo/bar").mkpath do` which somewhat surprisingly also doesn't work out of the box since `(mypathname/"foo/bar").mkpath` returns nil, whereas I'd have "expected" it to return a Pathname instance corresponding to the "bar" directory.

UniqMartin:
I agree that `Pathname#mkpath` accepting a block would be a neat thing to have. If you really want to have an ugly (but nonetheless working) hack, here's one:

```ruby
(mypathname/"foo/bar").tap(&:mkpath).cd { puts Dir.pwd }
```

ilovezfs:
Clever!

PR #739, opened 2016-05-01
SDL_image: exactly link with libpng and other recommended libraries

s172262:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

I'm sorry that this is re-open #281.

PR #1864, opened 2016-06-10
sdl_image: fix graphical glitching

tomyun:
- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?

---

The current version of `sdl_image` has graphical problems on El Capitan and affects many applications including Homebrew/homebrew-python#281.

It was fixed in `sdl2_image`, but never back-ported. Here we borrow a patch from [MacPorts](https://trac.macports.org/ticket/37453).
#739 may fix the same issue by disabling ImageIO, but that was a solution for another similar issue on 10.5 according to [the discussion](https://trac.macports.org/ticket/38012). As I don't have an access to 10.5 for testing, I'll just leave it for now.

### DIFF
--- a/Formula/sdl_image.rb
+++ b/Formula/sdl_image.rb
@@ -3,7 +3,7 @@ class SdlImage < Formula
   homepage "https://www.libsdl.org/projects/SDL_image"
   url "https://www.libsdl.org/projects/SDL_image/release/SDL_image-1.2.12.tar.gz"
   sha256 "0b90722984561004de84847744d566809dbb9daf732a9e503b91a1b5a84e5699"
-  revision 2
+  revision 3
 
   bottle do
     cellar :any
@@ -12,6 +12,8 @@ class SdlImage < Formula
     sha256 "84272135304e7c475e8000a90af113127858273a4449ce6566805f5b4c2cb476" => :mavericks
   end
 
+  option :universal
+
   depends_on "pkg-config" => :build
   depends_on "sdl"
   depends_on "jpeg" => :recommended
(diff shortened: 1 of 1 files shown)


Formula/sdl_image.rb
```diff
@@ -12,14 +12,22 @@ class SdlImage < Formula
...
   depends_on "libpng" => :recommended
   depends_on "libtiff" => :recommended
   depends_on "webp" => :recommended
 
-  option :universal
+  # Fix graphical glitching
+  # https://github.com/Homebrew/homebrew-python/issues/281
+  # https://trac.macports.org/ticket/37453
+  patch :p0 do
+    url "https://trac.macports.org/export/148198/trunk/dports/devel/libsdl_image/files/IMG_ImageIO.m.patch"
```
DomT4: Mind submitting this to `https://github.com/Homebrew/formula-patches` instead? Feel free to credit MacPorts still. 

We've just had past situations where Homebrew users haven't been able to install formula for 3+ days because MacPorts were having server problems, and it's hard to not feel like MacPort's servers shouldn't be a Homebrew dependency in any situation.

tomyun: Submitted to Homebrew/formula-patches#30.

DomT4: Merged there, Thanks for that.

tomyun: Formula updated to use the patch from `formula-patches`.

DomT4:
Tiny note for future: we prefer the patches linked to the commit reference rather than `master` so if the patch is updated or accidentally removed in future things already using it don't suddenly break. Fixed that in post here.

Merged in https://github.com/Homebrew/homebrew-core/commit/1ed223c6ad01ca76cca97b94ecce91bceb8e0459. Thank you for your contribution to Homebrew @tomyun; we appreciate it! :smiley_cat:

tomyun:
That makes more sense. Thanks for the note!

PR #1869, opened 2016-06-10
ooniprobe 1.5.1

hellais:

---
- Also update the sub-dependencies of it

### DIFF
--- a/Formula/ooniprobe.rb
+++ b/Formula/ooniprobe.rb
@@ -167,6 +167,11 @@ class Ooniprobe < Formula
     (var/"lib/ooni").mkpath
   end
 
+  def post_install
+   system bin/"ooniresources"
+   system bin/"oonideckgen", "-o", "#{HOMEBREW_PREFIX}/share/ooni/decks/"
+  end
+
   def caveats; <<-EOS.undent
     Decks are installed to #{HOMEBREW_PREFIX}/share/ooni.
     EOS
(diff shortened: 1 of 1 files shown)


Formula/ooniprobe.rb
```diff
@@ -169,10 +167,6 @@ def install
     (var/"lib/ooni").mkpath
   end
 
-  def post_install
-    system bin/"ooniresources", "--update-inputs", "--update-geoip"
```
DomT4: No longer needed?

hellais: Yes, it's now run as part of `setup.py` (https://github.com/TheTorProject/ooni-probe/blob/master/setup.py#L166)

DomT4: I'm not sure in our case, given we bottle things, that doesn't mean we don't want to retain this. Most people will pour the bottles, and for example the bottles are generated today so someone installing in a couple of months would still get the `inputs` and `geoip` from today as-is.

hellais: Oh I see. I didn't think about the bottles.

Will add it back then and take note of the fact that in the future ooniprobe should support disabling the fetching of resources via setup.py for use-cases like this.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/bf71e3cb77535d08f7596865449339b758b8ce4b. Thank you for your contribution to Homebrew @hellais; we appreciate it! :smiley_cat:

PR #1875, opened 2016-06-10
dvm 0.6.2 (new formula)

carolynvs:

---

FYI, dvm has an in-place upgrade command which I have disabled for homebrew users so that they don't bork their installation. 😄

### DIFF
--- a/Formula/dvm.rb
+++ b/Formula/dvm.rb
@@ -1,0 +1,34 @@
+class Dvm < Formula
+  desc "Docker Version Manager"
+  homepage "https://github.com/getcarina/dvm"
+  url "https://github.com/getcarina/dvm/archive/0.6.2.tar.gz"
+  sha256 "57a14fc85c37d30c3c2265eb0f06d2fdc18273306052a5ccc748b54a5c38660c"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    dvmpath = buildpath/"src/github.com/getcarina/dvm"
+    dvmpath.install Dir["{*,.git}"]
+
+    cd dvmpath do
+      system "make", "VERSION=#{version}", "COMMIT=6467884b2e25d080c25aa647bbcbbf8c12dc45bd", "UPGRADE_DISABLED=true"
+
+      prefix.install "dvm.sh"
+      (prefix/"dvm-helper").install "dvm-helper/dvm-helper"
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    dvm is a shell function, and must be sourced before it can be used.
+    Add the following command to your bash profile:
+      . "$(brew --prefix dvm)/dvm.sh"
+    EOS
+  end
+
+  test do
+    output = shell_output("bash -c 'source #{prefix}/dvm.sh && dvm --version'")
+    assert_match "Docker Version Manager version #{version}", output
+  end
+end


ruurd:
\o\\o//o/

Formula/dvm.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    dvmpath = buildpath/"src/github.com/getcarina/dvm"
+    dvmpath.install Dir["{*,.git}"]
+
+    cd dvmpath do
+      system "make", "VERSION=#{version}", "UPGRADE_DISABLED=true"
+
+      prefix.install "dvm.sh"
+      mkdir "#{prefix}/dvm-helper"
+      cp "dvm-helper/dvm-helper", "#{prefix}/dvm-helper/"
```
DomT4: You can squash these two lines into:

```ruby
(prefix/"dvm-helper").install "dvm-helper/dvm-helper"
```
If I'm reading that correctly.

DomT4:
> FYI, dvm has an in-place upgrade command which I have disabled for homebrew users so that they don't bork their installation. 

Thanks for doing this! Self-updating formulae are a pretty big no-no on our side as well because they cause pain for everyone involved 😄.

carolynvs:
@DomT4  Thank you for the tip on how to install a file into a subdirectory! I figured my hack wasn't "the homebrew way" but wasn't having much luck figuring out on my own how that `install` method works. 😊 

I've updated the PR with your suggestion.

rgbkrk:
:smile:

Formula/dvm.rb
```diff
@@ -0,0 +1,35 @@
+class Dvm < Formula
+  desc "Docker Version Manager"
+  homepage "https://github.com/getcarina/dvm"
+  url "https://github.com/getcarina/dvm.git",
```
DomT4: Does this need to be a git checkout or can it use the automatically generated tarball? https://github.com/getcarina/dvm/archive/0.6.2.tar.gz

Formula/dvm.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def caveats; <<-EOS.undent
+    dvm is a shell function, and must be sourced before it can be used.
+    Add the following command to your bash profile:
+      . "$(brew --prefix dvm)/dvm.sh"
+    EOS
+  end
+
+  test do
+    system "bash -c 'source $(brew --prefix dvm)/dvm.sh && dvm --version'"
```
DomT4: Might be nice to check the version output is correct, just as a teeny bit more rigorous a test:
```ruby
test do
  output = shell_output("bash -c 'source $(brew --prefix dvm)/dvm.sh && dvm --version'")
  assert_match version.to_s, output
end
```

(I didn't test this, but it should pass)

DomT4:
@carolynvs No worries at all, thanks for the quick response. Couple more nits but we're nearly good here 🙈.

carolynvs:
I have updated the formula to download an archive instead of using git, specifying the commit sha and release version explicitly.

Also the test now verifies the output of `dvm --version` and ensures that it matches the expected version.

carolynvs:
These errors popped up when I updated the formula to not use git as requested. The Makefile previously was able to figure out these values itself, but now that I am trying to set them manually in the formula, I've run afoul brew's auditing. I'd appreciate some feedback on how to fix the formula to pass the audit! 😄 

`Stable: version 0.6.2 is redundant with version scanned from URL`
That I'm not allowed to use a variable to hold the version, and have to put it in the url is disappointing. It's easier visually to have a set of variables (version, commit and sha256) to update with each new release, but oh well...

`C: 5: col 3: Replace class var @@commit with a class instance var.`
I am confused about why this is an error and how to address it? If I change `@@commit` to a class instance variable, how do I initialize it so that I can use it in multiple places in the formula? I tried just using `@commit = "abc123"` but the value isn't persisted, breaking the formula.

Formula/dvm.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    dvmpath = buildpath/"src/github.com/getcarina/dvm"
+    dvmpath.install Dir["{*,.git}"]
+
+    cd dvmpath do
+      system "make", "VERSION=#{version}", "COMMIT=#{@@commit}", "UPGRADE_DISABLED=true"
```
UniqMartin: Please replace `#{@@commit}` directly with the string. Ideal abstraction and avoiding all repetition is actually not the most important goal in formulae. They are intended to be easy to read by a human, even one not very familiar with Ruby.

Formula/dvm.rb
```diff
@@ -0,0 +1,36 @@
...
+  def caveats; <<-EOS.undent
+    dvm is a shell function, and must be sourced before it can be used.
+    Add the following command to your bash profile:
+      . "$(brew --prefix dvm)/dvm.sh"
+    EOS
+  end
+
+  test do
+  output = shell_output("bash -c 'source $(brew --prefix dvm)/dvm.sh && dvm --version'")
+  assert_match "Docker Version Manager version #{version} (#{@@commit})", output
```
UniqMartin: Simply drop `#{@@commit}` here. A partial match that just checks the version is sufficient.

Formula/dvm.rb
```diff
@@ -0,0 +1,36 @@
+class Dvm < Formula
+  desc "Docker Version Manager"
+  homepage "https://github.com/getcarina/dvm"
+  version "0.6.2"
+  @@commit = "6467884b2e25d080c25aa647bbcbbf8c12dc45bd"
```
UniqMartin: Please remove this line.

Formula/dvm.rb
```diff
@@ -0,0 +1,36 @@
...
+    dvm is a shell function, and must be sourced before it can be used.
+    Add the following command to your bash profile:
+      . "$(brew --prefix dvm)/dvm.sh"
+    EOS
+  end
+
+  test do
+  output = shell_output("bash -c 'source $(brew --prefix dvm)/dvm.sh && dvm --version'")
+  assert_match "Docker Version Manager version #{version} (#{@@commit})", output
+end
```
UniqMartin: The indentation of the `test do` block up to this `end` is off.

Formula/dvm.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def caveats; <<-EOS.undent
+    dvm is a shell function, and must be sourced before it can be used.
+    Add the following command to your bash profile:
+      . "$(brew --prefix dvm)/dvm.sh"
+    EOS
+  end
+
+  test do
+  output = shell_output("bash -c 'source $(brew --prefix dvm)/dvm.sh && dvm --version'")
```
UniqMartin: Please use the path directly instead of making the call invoke `brew --prefix`. You can replace `$(brew --prefix dvm)` with `#{prefix}`.

Formula/dvm.rb
```diff
@@ -0,0 +1,36 @@
+class Dvm < Formula
+  desc "Docker Version Manager"
+  homepage "https://github.com/getcarina/dvm"
+  version "0.6.2"
```
UniqMartin: Drop this line and instead inline the version into the `url` line. You will still be able to call `version` to query the version (that is auto-detected from the URL) in, e.g., `test do` or the `install` method.

UniqMartin:
> That I'm not allowed to use a variable to hold the version, and have to put it in the url is disappointing.

We prefer it to be part of the URL, as it's often more convenient to replace the entire URL when a new release arrives. And for consistency, we're not going to change that now.

carolynvs:
I have updated the formula with your feedback. Thanks!

UniqMartin:
Merged in 904dbbc0a973e67280198afd93cb3916af1b4d62. Thank you for this contribution to Homebrew, @carolynvs! :tada:

ruurd:
\o\\o//o/ Congrats Carolyn!

PR #1882, opened 2016-06-11
libtiff: add LZMA support

qu1j0t3:
- [Y] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

### DIFF
--- a/Formula/libtiff.rb
+++ b/Formula/libtiff.rb
@@ -15,10 +15,10 @@ class Libtiff < Formula
 
   option :universal
   option :cxx11
-  option "with-lzma", "Include support for LZMA compression"
+  option "with-xz", "Include support for LZMA compression"
 
   depends_on "jpeg"
-  depends_on "xz" if build.with? "lzma"
+  depends_on "xz" => :optional
 
   # Backports of various security/potential security fixes from Debian.
   # Already applied upstream in CVS but no new release yet.
@@ -35,15 +35,20 @@ class Libtiff < Formula
     ENV.universal_binary if build.universal?
     ENV.cxx11 if build.cxx11?
 
-    jpeg = Formula["jpeg"].opt_prefix
-    lzma = Formula["xz"].opt_prefix
-    system "./configure", "--disable-dependency-tracking",
-                          "--prefix=#{prefix}",
-                          "--without-x",
-                          "--with-jpeg-include-dir=#{jpeg}/include",
-                          "--with-jpeg-lib-dir=#{jpeg}/lib",
-                          "--with-lzma-include-dir=#{lzma}/include",
-                          "--with-lzma-lib-dir=#{lzma}/lib"
+    args = %W[
+      --disable-dependency-tracking
+      --prefix=#{prefix}
+      --without-x
+      --with-jpeg-include-dir=#{Formula["jpeg"].opt_include}
+      --with-jpeg-lib-dir=#{Formula["jpeg"].opt_lib}
+    ]
+    if build.with? "xz"
+      args << "--with-lzma-include-dir=#{Formula["xz"].opt_include}"
+      args << "--with-lzma-lib-dir=#{Formula["xz"].opt_lib}"
+    else
+      args << "--disable-lzma"
+    end
+    system "./configure", *args
     system "make", "install"
   end
 
--- a/Formula/libtiff.rb
+++ b/Formula/libtiff.rb
@@ -15,8 +15,10 @@ class Libtiff < Formula
 
   option :universal
   option :cxx11
+  option "with-lzma", "Include support for LZMA compression"
 
   depends_on "jpeg"
+  depends_on "xz" if build.with? "lzma"
 
   # Backports of various security/potential security fixes from Debian.
   # Already applied upstream in CVS but no new release yet.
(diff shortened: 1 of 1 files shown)


Formula/libtiff.rb
```diff
@@ -15,8 +15,10 @@ class Libtiff < Formula
 
   option :universal
   option :cxx11
+  option "with-lzma", "Include support for LZMA compression"
```
DomT4: Please just make this `with-xz` and then `depends_on "xz" => :optional`.

Formula/libtiff.rb
```diff
@@ -34,12 +36,14 @@ def install
...
     jpeg = Formula["jpeg"].opt_prefix
+    lzma = Formula["xz"].opt_prefix
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
                           "--without-x",
-                          "--disable-lzma",
                           "--with-jpeg-include-dir=#{jpeg}/include",
-                          "--with-jpeg-lib-dir=#{jpeg}/lib"
+                          "--with-jpeg-lib-dir=#{jpeg}/lib",
+                          "--with-lzma-include-dir=#{lzma}/include",
```
DomT4: Given you've made this optional, you shouldn't automatically throw it into the configure arguments as it'll break.

DomT4: You'll want to switch this over to using an `args = %W[]` style array.

UniqMartin: To be more specific, this part should probably look something like (untested):

```ruby
jpeg = Formula["jpeg"].opt_prefix
args = %W[
  --disable-dependency-tracking
  --prefix=#{prefix}
  --without-x
  --with-jpeg-include-dir=#{jpeg}/include
  --with-jpeg-lib-dir=#{jpeg}/lib
]
if build.with?("xz")
  args << "--with-lzma-include-dir=#{Formula["xz"].opt_include}"
  args << "--with-lzma-lib-dir=#{Formula["xz"].opt_lib}"
else
  args << "--disable-lzma"
end
system "./configure", *args
```

Formula/libtiff.rb
qu1j0t3: I considered that, but the libtiff option is actually `lzma`, which is what I'd be looking for if I were installing lzma. The fact this is packaged as `xv` in Homebrew seems arbitrary. Are you sure?

Formula/libtiff.rb
qu1j0t3: Good pickup, thanks.

Formula/libtiff.rb
DomT4: It's consistent with every other usage of `xz` for LZMA in Homebrew's core, where there's an explicitly defined `option`.

UniqMartin: I believe upstream changed the name from `lzma` to `xz` at some point, so our naming isn't that arbitrary. We could say that `libtiff` hasn't entirely caught up, but that's beside the point. The problem with `--with-lzma` is that this will make it harder for our dependency handling to handle the `xz` dependency and why we strongly prefer `depends_on "xz" => :optional` over `depends_on "xz" if <condition>`. `brew options libtiff` will help with finding the correct option name.

That said, is there a strong argument to be made against enabling LZMA support by default in `libtiff`? If not, I'd prefer we do just that and avoid all complications the option and dependency handling creates. (And provide a more feature-complete `libtiff` by default.)

qu1j0t3: I don't think very many users actually need it, and it adds a dep. So I'd argue against making it default.

LZMA doesn't seem to be supported by many viewers, either. (e.g. OS X Preview doesn't support it, and nor does Adobe Photoshop). 

However I happened to need it for some testing, and it's pretty easy to have it there just in case.

qu1j0t3: Re naming: Happy to go with `xv`, but I would mention that `xv` and `lzma` really refer to two different things (a tool and a compression method). It is unlikely libtiff would "catch up" as its options are compression methods.

DomT4: `xz` to be clear. We're aware there's different schools of thought on the whole naming ordeal, but on OS X it has been the case for a fair while, whether using MacPorts, Homebrew or Fink that `<package manager> install xz` gets you `xz` and `lzma`.

There's only a minimal distinction on Linux as well. `apt-get install xz-utils` for example, installs both the `xz-utils` and `lzma`. Both are technically separate packages, but if you install `xz` you get the latter, and both use the same source tarball.

UniqMartin: You're right, LZMA is a compression algorithm and there are multiple projects implementing it, though most distributions and package managers nowadays tend to default to the implementation provided by the `xz` project. I was a bit confused here. Thanks for the clarification!

qu1j0t3: @DomT4  Sure, but I think "what package the library comes from" is not relevant to enabling the LZMA compression method. Brew metadata abstracts that as intended.

But if the consensus is `xz` then that's what I'll use.

UniqMartin: @qu1j0t3 Yes, please use `with-xz`. You can still retain the description that makes it clear that this is actually referring to LZMA compression. Maybe one day we'll be able to handle this better in our dependency system, i.e. be able to better separate option names from formula names where it makes sense without all the downsides `depends_on <formula> if <condition>` currently has.

qu1j0t3: No worries.

UniqMartin: Yes, making it optional as you originally proposed is fine with me.

qu1j0t3:
Changes made as suggested.

UniqMartin:
Merged in 7dbd201adeaf96435d28fce8057c6204e0812504. Thank you for this contribution to Homebrew, @qu1j0t3! :tada:

PR #1899, opened 2016-06-11
consul-template 0.15.0

ilovezfs:
Thanks to whoever upstream vendored everything.

### DIFF
--- a/Formula/consul-template.rb
+++ b/Formula/consul-template.rb
@@ -1,12 +1,9 @@
-require "language/go"
-
 class ConsulTemplate < Formula
   desc "Generic template rendering and notifications with Consul"
   homepage "https://github.com/hashicorp/consul-template"
   url "https://github.com/hashicorp/consul-template.git",
-      :tag => "v0.14.0",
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 ` +14 −119` LOL.

DomT4:
They've moved to vendoring? Nice. I'll trust you this all builds correctly, I've had enough `go` for the week 😄.

ilovezfs:
Yeah, it works beautifully, modulo the fact that it still needs requires `make bootstrap`, which sneaks in that old gox thing through the backdoor.

DomT4:
We could always keep just the `gox` resource. I vaguely recall that's the situation in at least a couple other formulae as well.

ilovezfs:
Oh, no it's fine as is. The vendoring of gox is the same as the rest of it. I'm just still disappointed it needs that tool at all.

DomT4:
Thanks for the clarification. Initially thought you meant that bit wasn't vendored for some reason and online fetching was still happening. Nice work here.

PR #1924, opened 2016-06-13
bup 0.28.1

ilovezfs:
Also,
- always run `make test` when bottling
- enable a default `bup web` option by vendoring Tornado
- use upstream fix for `make test` getting stuck without Tornado

### DIFF
--- a/Formula/bup.rb
+++ b/Formula/bup.rb
@@ -1,9 +1,8 @@
 class Bup < Formula
   desc "Backup tool"
   homepage "https://github.com/bup/bup"
-  url "https://github.com/bup/bup/archive/0.27.tar.gz"
-  sha256 "ae8a744a3415ce4766d1d896f3d48b9b2ae6167c7bc65d8d1a112f37b42720fb"
-
+  url "https://github.com/bup/bup/archive/0.28.1.tar.gz"
+  sha256 "fd962dbdade1b8ea257ac0e95d771ba11e6da4ef6f8ca6bee498a5b1bce8c817"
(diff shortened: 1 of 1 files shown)


Formula/bup.rb
```diff
@@ -14,20 +13,81 @@ class Bup < Formula
...
-    system "make", "test" if build.with? "test"
+    system "make", "test" if build.bottle? || build.with?("test")
     system "make", "install", "DESTDIR=#{prefix}", "PREFIX="
   end
 
+  def caveats
+    if build.with? "web"
+      <<-EOS.undent
+        In order to use `bup web` you must set your PYTHONPATH in your ~/.bashrc:
+        export PYTHONPATH="#{opt_libexec}/vendor/lib/python2.7/site-packages:$PYTHONPATH"
```
MikeMcQuaid: Seems like something we should set in a wrapper script.

MikeMcQuaid:
One comment otherwise 👍

ilovezfs:
@MikeMcQuaid As in, like this?
```
    if build.with? "web"
      mv bin/"bup", libexec/"bup.py"
      (bin/"bup").write_env_script libexec/"bup.py", :PYTHONPATH => ENV["PYTHONPATH"]
    end
```

MikeMcQuaid:
aye 👍

MikeMcQuaid:
LGTM 👍 death to caveats!

ilovezfs:
@MikeMcQuaid Yes, indeed: they create too much controversy concerning the letter after "r" but before "t"

PR #1939, opened 2016-06-13
perl-build 1.13

ilovezfs:
Also, add a test.

### DIFF
--- a/Formula/perl-build.rb
+++ b/Formula/perl-build.rb
@@ -1,13 +1,18 @@
 class PerlBuild < Formula
   desc "Perl builder"
   homepage "https://github.com/tokuhirom/Perl-Build"
-  url "https://github.com/tokuhirom/Perl-Build/archive/1.12.tar.gz"
-  sha256 "2a3c33b7ea63c511db20778ebcedc35a0cc53fb780fc61b23d48ef3cc216bfc8"
+  url "https://github.com/tokuhirom/Perl-Build/archive/1.13.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Did upstream forget to update the version line with the new release?

ilovezfs:
@DomT4 clearly, nor is it updated in HEAD. I checked and it was the scenario last time. i.e., current version report 11 not 12.

DomT4:
Sigh. It's probably worth giving them a nudge about that to try and stop us inreplacing it every time, but 👍.

ilovezfs:
Yeah, I couldn't decide whether to do that. It's so ridiculously wrong and for so long that I assume they know and I'd just be annoying. But I suppose that's their problem not mine :)

DomT4:
If we're having to effectively patch it you've every right to be annoying 😉.

ilovezfs:
I wonder if the "real" issue here is that we should be building it, not installing the prebuilt thingy

DomT4:
If it has the correct version after manual building, possibly so.

ilovezfs:
Not sure what this output means yet, but it does suggest it's cognizant of the correct version
```
iMac-TMP:Perl-Build joe$  perl Build.PL --install-base=`pwd`/IB
Created MYMETA.yml and MYMETA.json
Creating new 'Build' script for 'Perl-Build' version '1.13'
cp META.json MYMETA.json
```

ilovezfs:
Yeah, this is just confusing, since the 1.13 tag is from April 1 but the pre-built thing was last updated January 8. So I'm closing this PR for now.

PR #1941, opened 2016-06-13
xorriso 1.4.2

ilovezfs:


Formula/xorriso.rb
```diff
@@ -1,9 +1,9 @@
 class Xorriso < Formula
   desc "ISO9660+RR manipulation tool"
   homepage "https://www.gnu.org/software/xorriso/"
-  url "http://ftpmirror.gnu.org/xorriso/xorriso-1.4.0.tar.gz"
-  mirror "https://ftp.gnu.org/gnu/xorriso/xorriso-1.4.0.tar.gz"
-  sha256 "0bd1e085015b28c24f57697d6def2fe84517967dc417554c0c3ccf1685ed0e56"
+  url "https://www.gnu.org/software/xorriso/xorriso-1.4.2.tar.gz"
```
UniqMartin: I wonder whether `ftpmirror` was intentionally favored over the primary site. @DomT4 probably knows …

Formula/xorriso.rb
```diff
@@ -13,13 +13,21 @@ class Xorriso < Formula
...
+    url "http://bazaar.launchpad.net/~libburnia-team/libisofs/scdbackup/diff/1319"
+    sha256 "c548c59b143efb87ccb3dc08f80f527ae93918fab56894c9f8876f31dccce142"
+  end
+
   def install
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make", "install"
+    system "make"
+    ENV.deparallelize { system "make", "install" }
```
UniqMartin: A regression from the previous version where it still worked? That's sad and maybe worth an upstream bug report, if that's not a known issue.

ilovezfs: More likely it's the machine/configuration as opposed to a regression.

UniqMartin: How so?

ilovezfs: I just mean I suspect it will die with the prior version too. I've not checked yet.

UniqMartin: I see. You mean it might have been just luck that it went through last time and the `install` target wasn't safe before? That would make sense …

ilovezfs: Yes, exactly.

Formula/xorriso.rb
DomT4: Hmm. Kind of. GNU have a standing request that where the software is served by them to use mirrors, officially because it's closer & quicker for users but also partially I suspect to stop their servers getting utterly rammed:

> GNU software is available on ftp.gnu.org under the directory /gnu. Please try to use one of the many mirrors of our site listed below: the mirrors will give you faster response. Some mirrors also provide a copy of the rest of ftp.gnu.org; try the parent directory if you're interested. You can use the generic url http://ftpmirror.gnu.org to automatically choose a nearby and up-to-date mirror.

However, in this case, the homepage doesn't make any mention of the FTP mirrors (most GNU "software" pages do) and directs people to the `software` site directly. Part of me thinks this is consequently safe to change, part of me disagrees.

Formula/xorriso.rb
DomT4: It's probably worth reporting it, just don't necessarily say it only impacts xyz version I guess.

ilovezfs: Bizarre. It appears to be a regression and a problem with the Makefile.in. If I regenerate with autoreconf -fiv, the problem does not occur. Lovely.

ilovezfs: OK, the problem is with the included `install-sh` file that included in the tarball.
```
cp Formula["automake"].share/"automake-1.15/install-sh", buildpath
```
fixes it. Ahem.

ilovezfs: Looks like we can do this as a workaround:
```
system "make", "MKDIR_P=mkdir -p", "install"
```

UniqMartin: Bizarre and lovely, indeed. I'm amazed at your willingness (and ability) to dissect this. I think it's best to stick with the `ENV.deparallelize` workaround and to forward what you've learned to upstream (if you haven't already), so that they can hopefully fix that for good in future releases.

Gotta love Autotools build systems …

ilovezfs: Yeah, I agree that deparallelizing is the right path for now. Only other thing I'd actually consider is re-running autoconf, but that ends up wasting more time than it saves.

I put in a request to join the mailing list, but may end up just emailing scdbackup at gmx.net if Mailman doesn't decide to wake up.

UniqMartin: Thanks! Some part of me wants to just `autoreconf` every formula that has both a `configure` and `configure.ac` file, but this is probably not a sane course of action. And it would add quite a few dependencies to quite a few formulae …

ilovezfs: That would certainly simplify all the goofy stable vs. head logic to prevent – heaven forbid – having to run autoreconf -fi when *unnecessary*, and it sure is annoying suffering under the yoke of some old version of install-sh when we have a perfectly good one in hand. LOL

ilovezfs: Woo hoo Mailman woke up.

ilovezfs:
:ship:?

Formula/xorriso.rb
```diff
@@ -13,13 +13,26 @@ class Xorriso < Formula
...
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make", "install"
+    system "make"
+
+    # `make install` has to be deparallelized due to the following error:
+    #   mkdir: /usr/local/Cellar/xorriso/1.4.2/bin: File exists
+    #   make[1]: *** [install-binPROGRAMS] Error 1
+    # Reported 14 Jun 2016: http://lists.gnu.org/archive/html/bug-xorriso/2016-06/msg00003.html
+    ENV.deparallelize { system "make", "install" }
```
ilovezfs: @DomT4 another one liner!

UniqMartin:
> 🚢?

:shipit:

DomT4:
👍

PR #1964, opened 2016-06-14
taglib 1.11

ilovezfs:
Also, add a test.

### DIFF
--- a/Formula/taglib.rb
+++ b/Formula/taglib.rb
@@ -1,9 +1,8 @@
 class Taglib < Formula
   desc "Audio metadata library"
   homepage "https://taglib.github.io/"
-  url "https://taglib.github.io/releases/taglib-1.10.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/taglib.rb
```diff
@@ -1,9 +1,8 @@
 class Taglib < Formula
   desc "Audio metadata library"
   homepage "https://taglib.github.io/"
-  url "https://taglib.github.io/releases/taglib-1.10.tar.gz"
-  sha256 "24c32d50042cb0ddf162eb263f8ac75c5a158e12bf32ed534c1d5c71ee369baa"
-
+  url "http://taglib.github.io/releases/taglib-1.11.tar.gz"
```
DomT4: Is this not available over TLS? It should be.

ilovezfs: That's weird. Wonder how that happened

DomT4:
👍

PR #1965, opened 2016-06-14
arangodb 2.8.9 and fixed --HEAD install

fceller:

### DIFF
--- a/Formula/arangodb.rb
+++ b/Formula/arangodb.rb
@@ -1,15 +1,18 @@
 class Arangodb < Formula
   desc "Universal open-source database with a flexible data model"
   homepage "https://www.arangodb.com/"
-  url "https://www.arangodb.com/repositories/Source/ArangoDB-2.8.7.tar.gz"
-  sha256 "cda284af7e1d45c2250dcb2acd253e805e9aaad3504f1d09cfcb8afa5ad87020"
-
-  head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
+  url "https://www.arangodb.com/repositories/Source/ArangoDB-2.8.9.tar.gz"
+  sha256 "cff21ca654056bed08781c5e462966f5f15acec7b6522191d286dee3339e327e"
 
   bottle do
     sha256 "b048abb7d6c7ed967885375341dc141512cd498ec004c823f21979a9b9e96315" => :el_capitan
     sha256 "e55fc36e8e167732f2160b1366edcee901fe380ad9deef92d963c2eb745f69c7" => :yosemite
     sha256 "b3b0b09665fef2cc8dabfacd15110e8f1d72680aeeffa6e1dc72ab3cc2fa5376" => :mavericks
+  end
+
+  head do
+    url "https://github.com/arangodb/arangodb.git", :branch => "unstable"
+    depends_on "cmake" => :build
   end
 
   depends_on "go" => :build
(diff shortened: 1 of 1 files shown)


Formula/arangodb.rb
```diff
@@ -23,41 +26,91 @@ class Arangodb < Formula
...
-      --datadir=#{share}
-      --localstatedir=#{var}
-    ]
+    if build.head?
+      args = %W[
+        -DVERBOSE=On
+        -DHOMEBREW=On
+        -DUSE_OPTIMIZE_FOR_ARCHITECTURE=Off
+        -DASM_OPTIMIZATIONS=Off
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo
```
DomT4: Why `RelWithDebInfo` specifically?

Formula/arangodb.rb
```diff
@@ -23,41 +26,91 @@ class Arangodb < Formula
...
 
-    args = %W[
-      --disable-dependency-tracking
-      --prefix=#{prefix}
-      --disable-relative
-      --datadir=#{share}
-      --localstatedir=#{var}
-    ]
+    if build.head?
+      args = %W[
```
DomT4: I'm not _enormously_ in love with a significant list of custom `cmake` args, to be honest. We try to nudge people towards `std_cmake_args` because it provides us with consistency.

Formula/arangodb.rb
```diff
@@ -23,41 +26,91 @@ class Arangodb < Formula
...
+        --disable-dependency-tracking
+        --prefix=#{prefix}
+        --disable-relative
+        --datadir=#{share}
+        --localstatedir=#{var}
+      ]
 
-    system "./configure", *args
-    system "make", "install"
+      args << "--program-suffix=-unstable" if build.head?
```
DomT4: Is this still supposed to be here? You've tucked the `if build.head?` conditional inside what is effectively a `if build.stable?` block.

Formula/arangodb.rb
```diff
@@ -23,41 +26,91 @@ class Arangodb < Formula
...
+
+      system "./configure", *args
+      system "make", "install"
+    end
   end
 
-  def post_install
-    (var/"arangodb").mkpath
-    (var/"log/arangodb").mkpath
+  if build.head?
```
DomT4: `build.head`, etc is nil outside of install, so I'm a bit surprised this actually gets triggered correctly? Duplicating the entire block is a bit grim, as well. It probably makes sense to combine this a chunk using `if head?` and `else` blocks inside a singular `def post_install`.

Formula/arangodb.rb
```diff
@@ -23,41 +26,91 @@ class Arangodb < Formula
...
+    def post_install
+      (var/"arangodb").mkpath
+      (var/"log/arangodb").mkpath
 
-    system "#{sbin}/arangod" + (build.head? ? "-unstable" : ""), "--upgrade", "--log.file", "-"
+      pw = (0...16).map { (65 + rand(26)).chr }.join
+
+      begin
+        system "#{sbin}/arango-init-database", "--database.password", pw
```
DomT4: We typically create databases without passwords, sadly, because I can promise 99% of users won't pay a bit of notice to the `post_install` output and we'll get Issues like _"I can't access my `arangodb` database, what did you do"_ 😕.

Formula/arangodb.rb
```diff
@@ -23,41 +26,91 @@ class Arangodb < Formula
...
+        system "echo", "default password is #{password}"
```
DomT4: `ohai` rather than `echo`. Calling the system to echo isn't necessary in Ruby, lots of alternatives, including our internal ones.

Formula/arangodb.rb
```diff
@@ -23,41 +26,91 @@ class Arangodb < Formula
...
+      system "#{sbin}/arangod", "--upgrade", "--log.file", "-"
+    end
   end
 
-  def caveats; <<-EOS.undent
-    Please note that clang and/or its standard library 7.0.0 has a severe
-    performance issue. Please consider using '--cc=gcc-5' when installing
-    if you are running on such a system.
-    EOS
+  if build.head?
```
DomT4: Again, please avoid duplicating entire blocks rather than using conditionals inside of them. [This sort of syntax](https://github.com/Homebrew/homebrew-core/blob/c11358f26c963a2c24563935db855a4deb285562/Formula/node.rb#L128-L150) will serve you well.

Formula/arangodb.rb
fceller: sound good, will add std_cmake_args. However, USE_OPTIMIZE_FOR_ARCHITECTURE, ASM_OPTIMIZATIONS, ETCDIR, VARDIR are specific to ArangoDB.

Formula/arangodb.rb
fceller: Missed that.

Formula/arangodb.rb
fceller: To be honest, I did not know that "head?" could be used inside the post_install (in %post in spec file that is sometimes a problem). Will change

Formula/arangodb.rb
fceller: @DomT4 is there a way (like debian packages) to ask users for a password?

Formula/arangodb.rb
fceller: will change

Formula/arangodb.rb
fceller: will change

Formula/arangodb.rb
fceller: Normally that allows people to debug crashes without compromising performance. But with std_cmake_args I assume it should not specified at all.

Formula/arangodb.rb
fceller: @DomT4 I tried the following:

Formula/arangodb.rb
MikeMcQuaid: Nope, let's just create without, thanks.

fceller: Ok, will do

fceller:
@DomT4 Does this fix all the issues?

MikeMcQuaid:
Thanks again @fceller!

DomT4:
@MikeMcQuaid Heads up, we needed a new bottle here. The stable version changed, and bottle is now 404.

MikeMcQuaid:
@DomT4 Yeh, not sure what I was doing here 😞. Pulling one now.

DomT4:
https://github.com/Homebrew/homebrew-core/commit/8f78859975b3e1bbb736ab1fa0fb48f838889811

MikeMcQuaid:
Cheers @DomT4!

PR #1968, opened 2016-06-14
Conan 0.9.4

lasote:

---

The python dep is required always, if don't this error is raised:

```
/usr/local/Library/Formula$ conan --version
  File "/usr/local/Cellar/conan/0.0.15/libexec/bin/conan", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 2797, in <module>
    parse_requirements(__requires__), Environment()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 576, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: setuptools>=11.3
```

OpenSSL is now needed too.

### DIFF
--- a/Formula/conan.rb
+++ b/Formula/conan.rb
@@ -1,8 +1,8 @@
 class Conan < Formula
   desc "Distributed, open source, package manager for C/C++"
   homepage "https://github.com/conan-io/conan"
-  url "https://pypi.python.org/packages/0b/34/5b03ad307911ef4221dd965a8867502dc96107cf58ed83f3274421e6808d/conan-0.9.2.tar.gz"
-  sha256 "750cef39e504d6d1e267a4fd6fc42ff4d9f4a3a25bbe3dfdbdff573025932db0"
+  url "https://pypi.python.org/packages/27/47/9098ac3ad4416c5a2ecb6bfbd45dd65e59c01a1f1281493c737a95245ce2/conan-0.9.4.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
We don't need to enforce a hard `:python` dependency to fix that; can add a `setuptools` resource like [this](https://github.com/Homebrew/homebrew-core/blob/237c5eb054962af82a6cfb03e6a39fd9407ec689/Formula/protobuf.rb#L53-L56).

lasote:
Hi, I've tried to add setuptools and remove python dependency. Now the setuptools error is not raised but it fails when we try to download conan packages from TLS servers (I've changed the test to download OpenSSL sources and build the conan package from the official server that requires a modern security protocol). It happens when python is using an old version of OpenSSL with old protocols. What do you think?

```
OpenSSL: EOF occurred in violation of protocol (_ssl.c:590)
```

The curious thing is if I execute `brew test conan --devel --keep-tmp` all is working fine in the sandbox...

UniqMartin:
I have honestly no idea why this is failing. Maybe @tdsmith can help …

lasote:
Could we keep the python dependency?

lasote:
Well, I'm closing this PR because it's stopped with no exit. I'll submit another one with the python dependency, will see if CI is ok and if you have any concern about it.

PR #1971, opened 2016-06-14
gauge 0.5.0

ilovezfs:
Also,
- use `godep restore` to replace the existing go_resource blocks
- add a new go_resource block for gauge_screenshot to avoid a `go get`
- match the version in the test


ilovezfs:
@DomT4 More Go! Your favorite!

DomT4:
_runs around screaming_. Seems reasonable enough 👍.

PR #1977, opened 2016-06-14
pick 1.4.0 (new formula)

fornwall:

### DIFF
--- a/Formula/pick.rb
+++ b/Formula/pick.rb
@@ -1,0 +1,18 @@
+class Pick < Formula
+  desc "Utility to choose one option from a set of choices"
+  homepage "https://github.com/thoughtbot/pick"
+  url "https://github.com/thoughtbot/pick/releases/download/v1.4.0/pick-1.4.0.tar.gz"
+  sha256 "46f46b0df54cf27e8dd19ae291d5534cb55ef37d9cdb3cc774cd88c809f718fd"
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "check"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/pick", "-v"
+  end
+end


DomT4:
@calleerlandsson @gfontenot Would you be happy with us supporting this or does upstream want to remain the go-to place for accessing this software via `brew`?

fornwall:
Sorry - I just switched to homebrew from macports and missed the more distributed philosophy of homebrew to keep packages in upstream-maintained taps, compared to the central package repository of macports.

DomT4:
We're perfectly happy to carry this in the core, we just try to avoid treading on upstreams if they've chosen not to submit to Homebrew core for a specific reason. If upstream respond to my ping and say _"We'd like to be part of Homebrew"_ more than happy to accept this PR.

calleerlandsson:
Making core the go-to place for getting pick sounds good to me.

How can we help people who're currently installing pick from the https://github.com/thoughtbot/homebrew-formulae tap to switch to the one in core?

fornwall:
Ok, re-opening this after the above feedback.

DomT4:
@calleerlandsson Effectively if you delete your `pick` formula once this has been merged, this `pick` formula will fall into its place after users `brew update` and all the `brew install/upgrade/info` style commands will execute on this `pick` formula.

Formula/pick.rb
```diff
@@ -0,0 +1,20 @@
+class Pick < Formula
+  desc "Utility to choose one option from a set of choices"
+  homepage "https://github.com/thoughtbot/pick"
+  url "https://github.com/thoughtbot/pick/releases/download/v1.4.0/pick-1.4.0.tar.gz"
+  sha256 "46f46b0df54cf27e8dd19ae291d5534cb55ef37d9cdb3cc774cd88c809f718fd"
+
+  depends_on "autoconf" => :build
```
DomT4: I think since you're using the release tarball here neither of these `:build` deps should be needed?

calleerlandsson: I think so too. For reference here's the formula we're currently using in the thoughtbot tap: https://github.com/thoughtbot/homebrew-formulae/blob/3e52b878635553c9a493808782151e41a3fdb99b/Formula/pick.rb

calleerlandsson:
Sounds good @DomT4!

Formula/pick.rb
fornwall: Removed the `:build` deps and squashed the commits.

DomT4:
@calleerlandsson If you could leave a comment here when you're online I'll go ahead and merge this then so the add here/remove on your end is fairly simultaneous.

calleerlandsson:
@DomT4, I'm online now and ready to remove the formula on our end.

calleerlandsson:
As @mptre pointed out in thoughtbot/pick#134, the `test do` block should now be able to call `system "make", "check"` to run the test suite.

fornwall:
Grepping for `"make", "check"` in Formula/ shows that this tends to be run during installation, not by the `test do` block.

Is that the right way @DomT4, or should `make check` only be run by `test do` (if so: how do one access the build directory from that block)?

DomT4:
@fornwall is correct that `make check` in `test do` wouldn't work. The `buildpath` and `testpath` are unique directories & scrubbed at the end of `brew install` and `brew test`'s execution. We could run `make check` during the install process though.

calleerlandsson:
I see. I'm not sure running `make check` on installation is helpful to people installing pick?

DomT4:
It's fairly standard for `make check` to be run before `make install` in most situations. Is `pick` not designed that way?

calleerlandsson:
Pick's `make check` will build and run a test runner that when executed will run a bunch of self-tests on the `pick` binary. It's not something we tell users to do when installing from source but is rather something we do during development to make sure things work before shipping new code.

The reason I'm not sure if it is helpful to users installing Pick is that the user can't really do much in case `make check` fails other than reporting the issue.

If you think it would be a good addition, I'm happy to try it though!

DomT4:
It's useful to us as packagers to know `make check` works; it helps us know with a little more confidence we're not accidentally passing bugs onto users. If the `make check` stage is particularly hefty we could scope it so it only applies to our CI rather than end users, but if it's quick we don't tend to bother.

calleerlandsson:
That makes sense. In Pick's case `make check` runs fairly quick.

fornwall:
Added `make check` before `make install` and squashed the commits, so this can hopefully be merged now!

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/36b5954d79515ccbb542923f510fdfe42621484b. Thank you for your contribution to Homebrew @fornwall; we appreciate it! :smiley_cat:. Thank you also @calleerlandsson for the input throughout.

calleerlandsson:
I've removed the formula from the thoughtbot tap, updated the installation instructions in the Pick README, and tried installing it from core and it worked great. Thanks both of you for working on this!

DomT4:
Thanks @calleerlandsson! 🎉

PR #1981, opened 2016-06-14
gimp, horndis, osxfuse, tuntap migrated

AnastasiaSulyagina:
_In progress_

Formulae removed due to [deduplication](https://github.com/caskroom/homebrew-cask/issues/15603).


Formula/gimp.rb
```diff
@@ -1,87 +0,0 @@
-class Gimp < Formula
```
xu-cheng: Why do we want remove this formula? As far as I can tell, some users prefer to install software from homebrew even it's availabile from cask. The reasons can be preferring building from source, preferring homebrew's sandbox, preferring DIY build options, etc.

MikeMcQuaid: @xu-cheng See the linked issue: https://github.com/caskroom/homebrew-cask/issues/15603

xu-cheng: I expressed my thought on that issue.

AnastasiaSulyagina: Thanks! 
As for this PR, it is not to be merged, it is just to see why do I have issues with debugging. So hopefully in the related issue we'll decide on what is finally to be deduplicated and I will do necessary migrations.

PR #1982, opened 2016-06-14
gox 0.3.0 (new formula)

ilovezfs:
consul-template, nomad, otto, packer, pup, terraform, and vault each
compile gox from scratch at build time, so I thought it might be simpler
to have a formula for gox that they could depend on. In more than one
case, this is the only non-vendored resource.

### DIFF
--- a/Formula/gox.rb
+++ b/Formula/gox.rb
@@ -1,0 +1,38 @@
+require "language/go"
+
+class Gox < Formula
+  desc "Go cross compile tool"
+  homepage "https://github.com/mitchellh/gox"
+  head "https://github.com/mitchellh/gox.git"
+
+  stable do
+    url "https://github.com/mitchellh/gox/archive/v0.3.0.tar.gz"
+    sha256 "29dc6b689f670a5444cc54cd9111549ccb01501901bc9197d0e1325a35157802"
+
+    # Equivalent to upstream commit "Allow go build tags to be passed as well":
+    # https://github.com/mitchellh/gox/commit/e557cfcb6e3c7f63c3abdae5dd354931814b0bdb
+    patch do
+      url "https://raw.githubusercontent.com/ilovezfs/formula-patches/bf09354fc67b4ec6ed2744613f9e2897e4764b60/gox/go-build-tags.diff"
+      sha256 "673adfacf3fb4e0a224ee735392f4681715d0530905bfb88cd126574ee50930a"
+    end
+  end
+
+  depends_on "go" => :build
+
+  go_resource "github.com/mitchellh/iochan" do
+    url "https://github.com/mitchellh/iochan.git",
+    :revision => "87b45ffd0e9581375c491fef3d32130bb15c5bd7"
+  end
+
+  def install
+    ENV["GOPATH"] = buildpath
+    (buildpath/"src/github.com/mitchellh/gox").install buildpath.children
+    Language::Go.stage_deps resources, buildpath/"src"
+    cd("src/github.com/mitchellh/gox") { system "go", "install" }
+    bin.install "bin/gox"
+  end
+
+  test do
+    assert_match "Usage: gox", shell_output("#{bin}/gox -h 2>&1", 2)
+  end
+end


ilovezfs:
I've requested a newer tag https://github.com/mitchellh/gox/issues/66 since the most recent one is significantly behind HEAD, which is what the formulae that depend on `gox` tend to go with.

ilovezfs:
This is an example of where `depends_on "gox" => :build` could be used: https://github.com/Homebrew/homebrew-core/pull/1983

ilovezfs:
@DomT4 By the way your initial hunch was correct about consul-template. Contrary to what I thought, `make bootstrap` runs `go get` https://github.com/hashicorp/consul-template/blob/master/Makefile#L52 so that should probably either be made like #1983 or it can use this PR.

DomT4:
Feared it would, I think it's baked into the make step where `gox` isn't available in the `$PATH`. I think this formula in theory is plenty reasonable.

ilovezfs:
Only issue I've seen so far is that the old 0.3.0 tag doesn't support the `-tags=` parameter which is used here: https://github.com/hashicorp/vault/blob/b45aa72f71e5ebae4231dfc23966bbfd79085a70/scripts/build.sh#L52

But just removing that line made it work, and the problem goes away altogether if upstream gox agrees to make a new tag.

DomT4:
Yes, the existing tag is kind of ancient at this point, would certainly be good to get a newer one in place. Thanks for asking there.

ilovezfs:
hehe `pup.rb` is still hard coding the old tag :)

DomT4:
Given that `pup` has 4 bottles it might have not been updated since gox `0.3.0` 😆

ilovezfs:
Yeah, that's always a dead giveaway, isn't it?

ilovezfs:
@DomT4 OK applying this in a patch do solves the `-tags` issue: https://gist.githubusercontent.com/ilovezfs/761d7294acf2fc435d5f2acaecde528c/raw/86993f2ccf0b0d875a247cf79cf5f0b093897c6e/gistfile1.txt

That's the same as https://github.com/mitchellh/gox/commit/e557cfcb6e3c7f63c3abdae5dd354931814b0bdb, which doesn't apply cleanly for unrelated reasons.

So I think it makes sense to put that in formula patches until gox makes a new tag.

ilovezfs:
Here's the formula-patches PR: https://github.com/Homebrew/formula-patches/pull/31

ilovezfs:
Refreshed this PR to apply the patch.

ilovezfs:
Net impact not including the gox formula itself:
```
iMac-TMP:Formula joe$ git diff -- nomad.rb otto.rb packer.rb terraform.rb vault.rb consul-template.rb|grep -E '^\+'
+++ b/Formula/consul-template.rb
+  depends_on "gox" => :build
+    ENV.prepend_create_path "PATH", buildpath/"bin"
+++ b/Formula/nomad.rb
+  depends_on "gox" => :build
+++ b/Formula/otto.rb
+  depends_on "gox" => :build
+++ b/Formula/packer.rb
+  depends_on "gox" => :build
+++ b/Formula/terraform.rb
+  depends_on "gox" => :build
+++ b/Formula/vault.rb
+  depends_on "gox" => :build
iMac-TMP:Formula joe$ git diff -- nomad.rb otto.rb packer.rb terraform.rb vault.rb consul-template.rb|grep -E '^\-'
--- a/Formula/consul-template.rb
-    ENV.prepend_create_path "PATH", buildpath/"bin"
-      system "make", "bootstrap"
--- a/Formula/nomad.rb
-  go_resource "github.com/mitchellh/gox" do
-    url "https://github.com/mitchellh/gox.git",
-        :revision => "39862d88e853ecc97f45e91c1cdcb1b312c51ea"
-  end
-
-  go_resource "github.com/mitchellh/iochan" do
-    url "https://github.com/mitchellh/iochan.git",
-        :revision => "87b45ffd0e9581375c491fef3d32130bb15c5bd7"
-  end
-
-    cd gopath/"src/github.com/mitchellh/gox" do
-      system "go", "install"
-    end
-
--- a/Formula/otto.rb
-    github.com/mitchellh/iochan 87b45ffd0e9581375c491fef3d32130bb15c5bd7
-    github.com/mitchellh/gox 39862d88e853ecc97f45e91c1cdcb1b312c51eaa
-    cd gopath/"src/github.com/mitchellh/gox" do
-      system "go", "install"
-    end
-
--- a/Formula/packer.rb
-
-  go_resource "github.com/mitchellh/gox" do
-    url "https://github.com/mitchellh/gox.git",
-      :revision => "6e9ee79eab7bb1b84155379b3f94ff9a87b344e4"
-  end
-  go_resource "github.com/mitchellh/iochan" do
-    url "https://github.com/mitchellh/iochan.git",
-      :revision => "87b45ffd0e9581375c491fef3d32130bb15c5bd7"
-  end
-
-    cd "src/github.com/mitchellh/gox" do
-      system "go", "build"
-      buildpath.install "gox"
-    end
-
--- a/Formula/terraform.rb
-
-  terraform_deps = %w[
-    github.com/mitchellh/gox 770c39f64e66797aa46b70ea953ff57d41658e40
-    github.com/mitchellh/iochan 87b45ffd0e9581375c491fef3d32130bb15c5bd7
-  ]
-
-  terraform_deps.each_slice(2) do |x, y|
-    go_resource x do
-      url "https://#{x}.git", :revision => y
-    end
-  end
-    cd "src/github.com/mitchellh/gox" do
-      system "go", "build"
-      buildpath.install "gox"
-    end
-
--- a/Formula/vault.rb
-  go_resource "github.com/mitchellh/gox" do
-    url "https://github.com/mitchellh/gox.git",
-        :revision => "39862d88e853ecc97f45e91c1cdcb1b312c51eaa"
-  end
-
-  # gox dependency
-  go_resource "github.com/mitchellh/iochan" do
-    url "https://github.com/mitchellh/iochan.git",
-        :revision => "87b45ffd0e9581375c491fef3d32130bb15c5bd7"
-  end
-
-    cd gopath/"src/github.com/mitchellh/gox" do
-      system "go", "install"
-    end
-
```
So possibly not worth it.

ilovezfs:
If upstream creates a new tag, I'll reconsider this option.

DomT4:
👍 If we can get a new tag I think this is worthwhile doing, personally.

PR #1983, opened 2016-06-14
vault 0.6.0

ilovezfs:
Also,
- add an option to run `make test`, and always do so when bottling
- all resources except `gox` have been vendored upstream

### DIFF
--- a/Formula/vault.rb
+++ b/Formula/vault.rb
@@ -7,9 +7,8 @@ class Vault < Formula
   desc "Secures, stores, and tightly controls access to secrets"
   homepage "https://vaultproject.io/"
   url "https://github.com/hashicorp/vault.git",
-      :tag => "v0.5.2",
-      :revision => "77f2b8a2fa408e0fc77ed7402d51cf0cfa0335d7"
-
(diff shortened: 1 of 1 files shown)


sl4m:
It looks like there's a newer tag / revision that fixes the version reported via `vault version`. The tag / revision you're using now reports the version as `Vault v0.6.0-rc1`.

New tag: `v0.6.0-rebuild`
New revision: `d2d0aa07d079836c9a5bd258479b019ca00eb6b8`

ilovezfs:
@sl4m last I checked, there was no difference between the tags with those names, but maybe that has since changed.

sl4m:
I have tried both versions and the newer tag outputs the correct version.

ilovezfs:
@sl4m yes as you can see I've already updated the PR ;)

sl4m:
Sorry, was replying directly via email and haven't looked. :+1:

ilovezfs:
@DomT4 :shipit:?

DomT4:
Aye, seems reasonable to me.

PR #1665, opened 2016-06-02
godep 74

Jerzerak:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---
- Bump stable version from 73 to 74

PR #1985, opened 2016-06-14
godep 74

ilovezfs:
Also,
- due to upstream changes, `go build` needs to be executed from a path
  containing "src/github.com/tools/godep" or it will fail
- change the test so that it doesn't cause warnings

Closes #1665.

### DIFF
--- a/Formula/godep.rb
+++ b/Formula/godep.rb
@@ -1,8 +1,8 @@
 class Godep < Formula
   desc "Dependency tool for go"
   homepage "https://godoc.org/github.com/tools/godep"
-  url "https://github.com/tools/godep/archive/v73.tar.gz"
-  sha256 "c2a2b37c20620d19cf341c479442067434049171e9cc3e226cdddde34aa6b3d9"
+  url "https://github.com/tools/godep/archive/v74.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
I know it's a chunk more lengthy but I do prefer our `cd do` blocks, I think people have a more instinctive understanding of that than brace relationships. 👍 otherwise, though.

ilovezfs:
It is a block ... hehe

DomT4:
Okay, technically 😉, but I fear that's less obvious to new users... and that someone's going to use this style of syntax on a less simple situation & we end up with hilariously stretched formulae 🙈.

ilovezfs:
I'd recommend the rule be that if it fits on a single line, use {}, else use do ... end. I'd definitely not recommend a more general use of {}.

DomT4:
I'd be fine with that, if it wraps to ~80 characters. My concern primarily is that contributors tend to follow the same style regardless of whether it's appropriate or not.

ilovezfs:
Here's another instance of the same: https://github.com/Homebrew/homebrew-core/pull/1983/files#diff-d7846406d4f4249bf301abe5d0185e56R57

ilovezfs:
I'll leave it up to you, but I suspect the one-liner rule shouldn't be too hard to police or communicate.

DomT4:
If you can keep it below the 80 character mark, 👍 generally.

PR #1990, opened 2016-06-14
juju 2.0-beta8 (devel)

tomck:

---

Note: the bash_completion.d filename changed. I also tried to make this work with --HEAD but that does not use the github.com/juju/juju/ prefix and I couldn't determine a good non-hacky way to do that without adding even more conditionals.

A ridiculous amount of testing went into this! Made sure it not only built with --devel but with -s to ensure the conditional was working.


DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/b810cfd2fc1c9409e1b91186aa5cbce5485755f9. Thank you for your contribution to Homebrew @tomck; we appreciate it! :smiley_cat:

PR #1993, opened 2016-06-14
vim, macvim: add analyics on Python/Ruby/Perl.

MikeMcQuaid:

---

We temporarily want to measure these to decide whether or not a bottle would make sense for either of these formulae.

CC @xu-cheng and @DomT4 who talked about this.

### DIFF
--- a/Formula/macvim.rb
+++ b/Formula/macvim.rb
@@ -25,6 +25,12 @@ class Macvim < Formula
   env :std if MacOS.version <= :snow_leopard
 
   def install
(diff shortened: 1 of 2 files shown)


DomT4:
Since I completely misread this at first, to clarify, this doesn't provide any information to Homebrew beyond `python: false, ruby: false perl: true` for example where you have a custom Python & Ruby installed but only the system Perl installed. There's no PII still.

I make a sour face at myself whilst saying this, but I don't have an objection here given the limited scope & the problem it helps us solve.

PR #2003, opened 2016-06-15
awscli 1.10.38

ilovezfs:
Also,
- minor style updates taken from `poet -f` output.
- have test match against the version

### DIFF
--- a/Formula/awscli.rb
+++ b/Formula/awscli.rb
@@ -1,8 +1,8 @@
 class Awscli < Formula
   desc "Official Amazon AWS command-line interface"
   homepage "https://aws.amazon.com/cli/"
-  url "https://pypi.python.org/packages/0e/27/1189968c900f9ad8f62c1483479bde13088d15172f68a81bfd5cfc3014dd/awscli-1.10.37.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Seems fine 👍. Maybe @tdsmith knows more about the change from `https://pypi.python.org` to `https://files.pythonhosted.org`... Is that part of the previously discussed URL changes or are we hitting a mirroring system or such?

tdsmith:
@dstufft says these are good and stable URLs so I'm happy; thanks for checking

DomT4:
Thanks Tim! 🙇

nijikon:
Always update botocore as well.

nijikon:
#2034

ilovezfs:
@nijikon As you can see I *did* update botocore (note that the URL changed). The change you just made is for a release of botocore that occurred after I opened this PR.

nijikon:
Sure, I just wanted to make sure that all are aware that the best approach is to update both awscli and botocore as they are dependent on each other. No worries :)

DomT4:
@nijikon Maybe this is worth noting in the formula if it needs an update _every_ time.

nijikon:
@DomT4 "Make sure that you also bump botocore"?

PR #2006, opened 2016-06-15
commandbox 3.1.1

jclausen:


DomT4:
Since what was the `devel` release has become the stable release you can go ahead and cull the `devel do` block entirely.

dunn:
:bell

bdw429s:
@jclausen Do we know what needs to be done here?  Are they waiting for us to remove something from the brew file?

jclausen:
@DomT4 :  Sorry, I must have missed the email notifications on this PR.  I'll check my settings.  I've bumped develop to the latest snapshot.  Going forward, when dev is the same as master, I'll omit the `devel do`  block.

cc/ @dunn @bdw429s

bdw429s:
Thanks @jclausen .

jclausen:
Update:  I just removed the `devel do` block altogether, as  your audit checks were failing because of the timestamp of the 3.2 snapshot was older that the 3.1.1 master.  We'll just maintain the `devel` blocks in our tap, from now on.

UniqMartin:
Merged in 7e3757930d07120c428d5b57cecc756669cb7cb4. Thank you for this contribution to Homebrew, @jclausen! :tada:

DomT4:
@jclausen FWIW, the problem with `devel` wasn't timestamps. Our versioning logic somewhat struggles with certain types of versions used in certain ways. In this case if you'd added a manual:
```ruby
version "3.2.0-snapshot"
```
to the `devel do` block CI would have been 🍏.

PR #2010, opened 2016-06-15
qt5: Add PostgreSQL plugin support

uranusjr:

---

The audit process gives an error about the MySQL plugin being linked against system OpenSSL, but that is unrelated to this patch, and should probably be fixed with another pull request.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -57,6 +57,7 @@ class Qt5 < Formula
 
   depends_on "dbus" => :optional
   depends_on :mysql => :optional
+  depends_on :postgresql => :optional
   depends_on :xcode => :build
 
   depends_on OracleHomeVarRequirement if build.with? "oci"
@@ -79,6 +80,7 @@ class Qt5 < Formula
     args << "-nomake" << "examples" if build.without? "examples"
 
     args << "-plugin-sql-mysql" if build.with? "mysql"
+    args << "-plugin-sql-psql" if build.with? "postgresql"
 
     if build.with? "dbus"
       dbus_opt = Formula["dbus"].opt_prefix


UniqMartin:
Merged in 4145ba04be7c47913166f789de2c0fb238d1ae8e. Thank you for this contribution to Homebrew, @uranusjr! :tada:

> The audit process gives an error about the MySQL plugin being linked against system OpenSSL, but that is unrelated to this patch, and should probably be fixed with another pull request.

Can you provide a bit more information here? The output of `brew audit` would be useful and the logs for this build of would be helpful, too. Maybe you can even investigate this problem a bit more, find out where it comes from, and how it can be fixed? Thanks!

uranusjr:
This is the output I got running `brew audit qt5 --strict --online`:

~~~
qt5:
    These object files were linked against the deprecated system OpenSSL.
      /usr/local/Cellar/qt5/5.6.1/plugins/sqldrivers/libqsqlmysql.dylib
Error: 1 problem in 1 formula
~~~

Since Qt 5.7 has been released, I’ll build it this weekend with the OpenSSL dependency to see if it fixes the error.

UniqMartin:
Thanks, @uranusjr! I'll try to check this as soon as time permits. (Qt is huge and a bit annoying to rebuild constantly to debug these kinds of problems.) Qt 5.7.0 is in preparation in #2087, but is currently blocked by build issues on Mavericks. If you happen to have any ideas why that could be, I'd appreciate your help.

uranusjr:
Update: I tried to rebuild with `depends_on "openssl" if builds.with? "mysql"`, but that didn’t work—the error still came up with the new build. I inspected the configure output, and it seems that OpenSSL might not have been picked up:

~~~
  Networking: 
    CoreWlan ............. yes
    getaddrinfo .......... yes
    getifaddrs ........... yes
    IPv6 ifname .......... yes
    libproxy.............. no
    OpenSSL .............. no
    SecureTransport ...... yes
~~~

Maybe I’ll find some time to try fixing this (and take a look at #2087). It really is not easy to work on this monster.

UniqMartin:
I think for OpenSSL to be used by Qt 5 you have to both enable OpenSSL, but also explicitly disable SecureTransport via corresponding `configure` flags. I think you might even have to pass the paths to the include and library directory [like is done in the Qt 4 formula](https://github.com/Homebrew/homebrew-core/blob/abe47453ba50810da185f2212629e1a012ce3bfb/Formula/qt.rb#L70-L72). But I don't think that's something we want to do, as we prefer for Qt to use SecureTransport on Apple platforms (and Qt upstream does, too).

Maybe the MySQL plugin is picking up the system OpenSSL because of the `mysql_config` that is provided by the installed MySQL and that is used by Qt to configure the MySQL plugin? Can you check what `mysql_config --include`, `mysql_config --libs`, and `mysql_config --libs_r` returns?

uranusjr:
~~~
$ mysql_config --include
-I/usr/local/Cellar/mysql/5.7.13/include/mysql
$ mysql_config --libs
-L/usr/local/Cellar/mysql/5.7.13/lib -lmysqlclient -lssl -lcrypto
$ mysql_config --libs_r
-L/usr/local/Cellar/mysql/5.7.13/lib -lmysqlclient -lssl -lcrypto
~~~

UniqMartin:
So this explains things. That's where Qt is picking up the libraries, but because the `mysql_config` output doesn't contain the paths that point to the Homebrew OpenSSL, it ends up picking up the (old) system OpenSSL. Not entirely sure how to best fix this.

My gut feeling is that this needs to be fixed in the `mysql` formula, i.e. by either removing the `-lssl -lcrypto` part (this is not needed for dynamically linking to `libmysqlclient`) or by adding the path to the Homebrew OpenSSL (`-L#{Formula["openssl"].opt_lib}` or something like that) to `mysql_config`.

PR #2012, opened 2016-06-15
digdag 0.8.2 (new formula)

darklore:

---

Adds new formula for digdag (https://github.com/treasure-data/digdag)

### DIFF
--- a/Formula/digdag.rb
+++ b/Formula/digdag.rb
@@ -11,10 +11,11 @@ class Digdag < Formula
   def install
     libexec.install "digdag-#{version}" => "digdag"
     chmod 0755, (libexec/"digdag")
-    bin.install_symlink (libexec/"digdag") => "digdag"
+    bin.install_symlink "#{libexec}/digdag" => "digdag"
   end
 
   test do
-    assert_match(/#{version}/, shell_output("#{bin}/digdag --version"))
+    ENV.java_cache
+    system "#{bin}/digdag", "--version"
   end
 end
--- a/Formula/digdag.rb
+++ b/Formula/digdag.rb
@@ -1,8 +1,8 @@
 class Digdag < Formula
   desc "Workload Automation System"
   homepage "https://github.com/treasure-data/digdag"
-  url "https://dl.digdag.io/digdag-0.8.1.jar"
-  sha256 "92e79f5eb26b8992c9d510206c71f64fe4765c0c51a4e4cb62279aa1a4231ec7"
+  url "https://dl.digdag.io/digdag-0.8.2.jar"
+  sha256 "9e56fe2009a8f3ef62d2e5d3d1b786d0a2a38686c846402acfdcfbb06eae2843"
 
   bottle :unneeded
 
--- a/Formula/digdag.rb
+++ b/Formula/digdag.rb
@@ -19,6 +19,6 @@ class Digdag < Formula
 
   test do
     ENV.java_cache
-    system "#{bin}/digdag", "--version"
+    assert_match version.to_s, shell_output("#{bin}/digdag --version")
   end
 end
--- a/Formula/digdag.rb
+++ b/Formula/digdag.rb
@@ -1,7 +1,7 @@
 class Digdag < Formula
   desc "Workload Automation System"
   homepage "https://github.com/treasure-data/digdag"
-  url "https://dl.digdag.io/digdag-0.8.1"
+  url "https://dl.digdag.io/digdag-0.8.1.jar"
   sha256 "92e79f5eb26b8992c9d510206c71f64fe4765c0c51a4e4cb62279aa1a4231ec7"
 
   bottle :unneeded
@@ -9,9 +9,9 @@ class Digdag < Formula
   depends_on :java => "1.8+"
 
   def install
-    libexec.install "digdag-#{version}" => "digdag"
-    chmod 0755, (libexec/"digdag")
-    bin.install_symlink "#{libexec}/digdag" => "digdag"
+    libexec.install "digdag-#{version}.jar" => "digdag.jar"
+    chmod 0755, libexec/"digdag.jar"
+    bin.install_symlink "#{libexec}/digdag.jar" => "digdag"
   end
 
   test do
--- a/Formula/digdag.rb
+++ b/Formula/digdag.rb
@@ -9,12 +9,13 @@ class Digdag < Formula
   depends_on :java => "1.8+"
 
   def install
-    # Add prefix a shebang before install for OS X 10.9.
-    # See https://github.com/Homebrew/homebrew-core/pull/2012#issuecomment-226407948 for the detail.
-    inreplace "digdag-#{version}.jar", /\A/, "#!/bin/sh\n"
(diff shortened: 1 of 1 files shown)


Formula/digdag.rb
```diff
@@ -0,0 +1,21 @@
+class Digdag < Formula
+  desc "Workload Automation System"
+  homepage "https://github.com/treasure-data/digdag"
+  url "https://dl.digdag.io/digdag-0.8.1"
```
DomT4: Mind using `https://dl.digdag.io/digdag-0.8.1.jar` instead? That seems to be the full path, and helps make it a bit more obvious what is being downloaded.

darklore: Thank you. I fixed it.

darklore:
I have no idea about the reason of mavericks test failure...
Does anyone know it?

UniqMartin:
> I have no idea about the reason of mavericks test failure...
> Does anyone know it?

This download is a really weird hybrid of a Windows batch file and a shell script with an appended JAR file. Unfortunately, the shell script part doesn't have a shebang line. OS X 10.9 is not as permissive as other systems in this regard and doesn't recognize that as a valid executable. Prefixing a shebang line should fix the problem. Add this line to the very beginning of the `install` method:

```ruby
inreplace "digdag-#{version}.jar", /\A/, "#!/bin/sh\n"
```

darklore:
Thank you!
I added the replacement.

Formula/digdag.rb
```diff
@@ -0,0 +1,24 @@
...
+    # See https://github.com/Homebrew/homebrew-core/pull/2012#issuecomment-226407948 for the detail.
+    inreplace "digdag-#{version}.jar", /\A/, "#!/bin/sh\n"
+    libexec.install "digdag-#{version}.jar" => "digdag.jar"
+    chmod 0755, libexec/"digdag.jar"
+    bin.install_symlink "#{libexec}/digdag.jar" => "digdag"
+  end
+
+  test do
+    ENV.java_cache
+    system "#{bin}/digdag", "--version"
```
UniqMartin: Now that things are working again on all systems, can we improve the test just a tiny bit?

```ruby
assert_match version.to_s, shell_output("#{bin}/digdag --version")
```

darklore: OK. I fixed it.

Formula/digdag.rb
```diff
@@ -0,0 +1,24 @@
...
+  sha256 "92e79f5eb26b8992c9d510206c71f64fe4765c0c51a4e4cb62279aa1a4231ec7"
+
+  bottle :unneeded
+
+  depends_on :java => "1.8+"
+
+  def install
+    # Add prefix a shebang before install for OS X 10.9.
+    # See https://github.com/Homebrew/homebrew-core/pull/2012#issuecomment-226407948 for the detail.
+    inreplace "digdag-#{version}.jar", /\A/, "#!/bin/sh\n"
```
MikeMcQuaid: Can you report this to the upstream repository as an issue? It's needed if they want to support OS X.

darklore: OK. I'll do it.

UniqMartin: I'm not sure they'll be willing to change this. They hacked together a nice header that works on both Windows and UNIX-like systems by exploiting compatibility behavior of most shells (if the kernel cannot execute it because it's not a binary and also has no shebang, try `/bin/sh` as an interpreter). If they added a shebang line this would more reliably work on UNIX-like systems, but it would no longer work on Windows and they had to distribute two files instead of one “universal” file.

I'd say the sane thing to do would be to distribute the plain `.jar` with two wrapper scripts for those two major families of OSes, but obviously that's just my opinion …

MikeMcQuaid: If we need to patch it: it doesn't support OS X, no?

UniqMartin: It does work when started from a shell. It doesn't work when started directly from Ruby via `system "#{bin}/digdag"`. This possibly only affects old Rubies as it was fine on 10.10/10.11 and just failed on our 10.9 test bot. Still, having a shebang line is desirable in my opinion. (I'm just not sure we can sell adding one to upstream for the reasons explained above.)

MikeMcQuaid: Ok, I see. @darklore ignore me for now!
@UniqMartin thoughts on using a wrapper script instead?

UniqMartin: Now that I had more time thinking about this, I believe the best thing we can do is treat the `.jar` just like any other `.jar` file in our Java-using formulae and generate our own wrapper script. I cannot think of a formula where we do that off the top of my head and a quick search didn't reveal anything, thus not sure if this would be appropriate. Most Java-using software is distributed with a somewhat sane wrapper script where we tend to only modify the `JAVA_HOME` variable.

darklore: I suggested to add a shebang but it is rejected.
(I'm not sure about hybrid script mechanism...)
https://github.com/treasure-data/digdag/pull/125

My suggestion is to separate test by MacOS.version like below.

```ruby
  test do
    ENV.java_cache
    if MacOS.version > :mavericks
      assert_match version.to_s, shell_output("#{bin}/digdag --version")
    else
      assert (bin/"digdag").executable?
    end
  end
```

UniqMartin: @darklore Thanks! I'd personally prefer not to include such test code. I'd commented on that PR you created. Let's see what the answer will be. I think adding the shebang line ourselves (as you did following my suggestion) is fine when done in a Homebrew formula as a program installed via Homebrew doesn't need to be executable on a non-OS-X system.

darklore: Thank you!

> a program installed via Homebrew doesn't need to be executable on a non-OS-X system.

That's true.
OK. I wait for the answer and leave this formula PR as it is.

MikeMcQuaid: Yeh, just remove the `inreplace` and use `bin.write_jar_script` instead of `.install_symlink` and the `chmod`.

Formula/digdag.rb
```diff
@@ -0,0 +1,25 @@
...
+  depends_on :java => "1.8+"
+
+  def install
+    libexec.install "digdag-#{version}.jar" => "digdag.jar"
+    # Create a wrapper script to support OS X 10.9.
+    # See https://github.com/Homebrew/homebrew-core/pull/2012#issuecomment-226407948 for the detail.
+    (bin/"digdag").write <<-EOS.undent
+      #!/bin/bash
+      exec /bin/bash "#{libexec}/digdag.jar" "$@"
+    EOS
```
darklore: I modified according to frsyuki's suggestion ( https://github.com/treasure-data/digdag/pull/125#issuecomment-227354246 )
If it suitable to use `write_jar_script`as discussion at the embulk formula https://github.com/Homebrew/homebrew-core/pull/2261#discussion_r67982956 , I'll fix it.

UniqMartin:
Merged in e372ca5b6dc58332ab377bfb3b9472f15b94a409. Thank you for this contribution to Homebrew, @darklore! :tada:

Thanks for your patience while we worked out how to best handle this! :bow:

PR #2029, opened 2016-06-16
ghc: fix build on 10.12

mistydemeo:

---

Fixes an issue with building GHC on 10.12 because of the `clock_gettime` function. See #1957.

### DIFF
--- a/Formula/ghc.rb
+++ b/Formula/ghc.rb
@@ -48,6 +48,13 @@ class Ghc < Formula
     sha256 "bc57163656ece462ef61072559d491b72c5cdd694f3c39b80ac0f6b9a3dc8151"
   end
 
+  # fix clock_gettime support on 10.12
+  # https://ghc.haskell.org/trac/ghc/ticket/12195
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/9eef46168ca5daa1629082af5532fc4b521d1a8d/ghc/clock_gettime.patch"
+    sha256 "42444b381840d9d90202ff619f0070c28a753243af5d7e6b60efd227c6cf308c"
+  end
+
   def install
     # Setting -march=native, which is what --build-from-source does, fails
     # on Skylake (and possibly other architectures as well) with the error


ilovezfs:
Timed out. This one took about 3 hours last time, so probably needs about a 4 hour timeout from now on.

MikeMcQuaid:
@mistydemeo Given how much Apple changes between the first developer preview and final release I think it may be worth waiting at least until the public beta before we start applying patches. Thoughts?

ilovezfs:
@MikeMcQuaid maybe just apply the patch with `system "/usr/bin/patch", …` if the detected OS is 10.12?

MikeMcQuaid:
We can do conditional patches without needing that. My concern is that we apply a patch that turns out to be unnecessary in future and we never realise. Maybe that's paranoid, I don't know, I don't feel strongly either way (except that it's a good idea to only patch this on 10.12).

ilovezfs:
How can a patch-do be limited to a specific OS?

MikeMcQuaid:
With an `if` around it; doesn't end up having the same problems as `option`s do.

ilovezfs:
Ah OK. @mistydemeo filed an upstream issue about it, so I think it's a question of what upstream decides to do, but I'd not be paranoid about this from the Apple side. They've decided to implement clock_gettime for the first time, and there's definitely no reason to imagine they'd later want to go back to having no implementation for it. Several of the programs on opensource.apple.com patch around the fact that macOS has not previously had clock_gettime, so they're probably already undoing their various patches for that.

Inevitably, the patch in this PR *will* become unnecessary as soon as upstream stops hard coding the notion that macOS won't ever have the function, which clearly turned out to be false :)

mistydemeo:
@MikeMcQuaid Regardless of whether Apple changes this or not, this is definitely still an upstream bug; the `#ifdef`s are buggy, and there's no harm from patching it like this. Of the changes that Apple makes, I also doubt that they'll roll back adding a new POSIX function. :)

MikeMcQuaid:
@mistydemeo 👍 then 😀

DomT4:
My position loosely on 10.12 patches is that we'll take them if submitted as PRs & shown to be working, but I'm not all that prepared myself to go chasing issues around until things are a _little_ more stable, FWIW. This one makes sense to me, and much ❤️ to Misty for the willingness to chase it.

mistydemeo:
@BrewTestBot test this please

MikeMcQuaid:
@mistydemeo I think the patch should just be scoped to 10.12 if it's only needed there and if you've tested it locally it can go in without CI (as the jobs aren't testing 10.12 anyway).

mistydemeo:
@MikeMcQuaid Since it's likely to be taken upstream, I'll apply the patch on all platforms. I'll pull without bottles though.

MikeMcQuaid:
@mistydemeo I'd rather we avoid applying patches on platforms they aren't needed 😢

mistydemeo:
@MikeMcQuaid OK, scoped it to only 10.12. My preference though, for things that will be applied upstream, is to patch on all platforms even if they only affect certain platforms.

MikeMcQuaid:
@mistydemeo 🙇

PR #2032, opened 2016-06-16
consul-template: avoid go get for gox

ilovezfs:
Use go_resource blocks for gox and iochan to avoid an insecure go get.


DomT4:
👍

PR #2033, opened 2016-06-16
libgetdata 0.9.2.1

ilovezfs:
Also, fix sandbox violation when building with the perl option.

### DIFF
--- a/Formula/libgetdata.rb
+++ b/Formula/libgetdata.rb
@@ -1,8 +1,8 @@
 class Libgetdata < Formula
   desc "Reference implementation of the Dirfile Standards"
   homepage "http://getdata.sourceforge.net/"
-  url "https://downloads.sourceforge.net/project/getdata/getdata/0.9.0/getdata-0.9.0.tar.xz"
-  sha256 "b38de059ff21df873e95978867eb82f716b89fc7e8e503e2cc7cef93d22685a2"
+  url "https://downloads.sourceforge.net/project/getdata/getdata/0.9.2.1/getdata-0.9.2.1.tar.xz"
+  sha256 "e41d82abdb86577ce41600f187a4271e17e544b67dc0e0518d42f285cd809499"
 
   bottle do
     cellar :any
@@ -20,6 +20,7 @@ class Libgetdata < Formula
   deprecated_option "zzip" => "with-libzzip"
 
   depends_on :fortran => :optional
+  depends_on :perl => ["5.3", :optional]
   depends_on "xz" => :optional
   depends_on "libzzip" => :optional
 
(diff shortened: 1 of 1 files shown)


Formula/libgetdata.rb
```diff
@@ -12,14 +12,14 @@ class Libgetdata < Formula
   end
 
   option "with-fortran", "Build Fortran 77 bindings"
-  option "with-perl", "Build Perl binding"
```
UniqMartin: Maybe keep the option nonetheless so that it is clearer what type of Perl support this formula has to offer?

ilovezfs: Doesn't seem necessary. What other kind of perl support is there besides building a pm for random program foo?

UniqMartin: Something that uses Perl as opposed to providing an interface to itself from Perl …

ilovezfs: If it were something that uses perl, would perl be optional or required?

ilovezfs: @UniqMartin I sympathize with what you're saying, but it seems like the meaning of a perl *option* is obvious, so if anything your concern would suggest the default wording of :perl => :optional/:recommended (or, for that matter, :python => :optional/:recommended) should be changed if it's not sufficiently clear what building with perl support or building with python support means without further customized option descriptions.

UniqMartin: I guess that depends. Anyway, just wanted to check. I'm not feeling overly passionate about this option, thus feel free to keep the PR as-is.

ilovezfs: I'm actually happy to change it back if you'd like. Just thought I'd give @DomT4's cool new :perl thingy and spin, and see if we could "get away with" the default option wording it provides.

UniqMartin:
:+1:

ilovezfs:
@DomT4 any thoughts on this use of `:perl => optional` vs. providing a customized option description? It would be good, I think, to have a "policy" concerning perl/python "bindings" and whether the default descriptions suffice, or if they should be customized, since optional bindings are a pretty common thing.

Formula/libgetdata.rb
DomT4: IMO it's worth introducing the `:perl` dependency _where_ it's optional anyway. It makes sense to build bindings against the version of Perl actually first in the user's `$PATH`, providing that's a suitable version of Perl to meet the requirement.

The only area we need to be particularly wary is where it isn't `:optional`, given that that drags in the need to install our `perl` (as is the situation with `python` as well).

DomT4:
Left a nested comment. Generally we do permit `option` usage on stuff like `:python` and `:perl` to highlight building with those is useful for bindings, when that is the case. This hasn't been applied consistently, though.

Formula/libgetdata.rb
ilovezfs: OK, I'll put ""Build Perl binding" back in then

DomT4:
👍

PR #2039, opened 2016-06-16
ghc: fix installation on Lion

mvkorpel:

---

The error with the previous version of this Formula was:
checking version of ghc... 7.6.3
configure: error: GHC version 7.8 or later is required to compile GHC.

### DIFF
--- a/Formula/ghc.rb
+++ b/Formula/ghc.rb
@@ -35,6 +35,8 @@ class Ghc < Formula
     end
   end
 
+  # https://www.haskell.org/ghc/download_ghc_8_0_1#macosx_x86_64
+  # "This is a distribution for Mac OS X, 10.7 or later."
   resource "binary" do
     url "https://downloads.haskell.org/~ghc/8.0.1/ghc-8.0.1-x86_64-apple-darwin.tar.xz"
     sha256 "06ec33056b927da5e68055147f165f873088f6812fe0c642c4c78c9a449fbc42"
--- a/Formula/ghc.rb
+++ b/Formula/ghc.rb
@@ -34,13 +34,8 @@ class Ghc < Formula
   end
 
   resource "binary" do
-    if MacOS.version <= :lion
-      url "https://downloads.haskell.org/~ghc/7.6.3/ghc-7.6.3-x86_64-apple-darwin.tar.bz2"
-      sha256 "f7a35bea69b6cae798c5f603471a53b43c4cc5feeeeb71733815db6e0a280945"
-    else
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
This now needs a `depends_on :macos => :lion` but otherwise 👍

ilovezfs:
Do we know if 8.0.1 actually works on Lion? Did upstream actually fix the issues that resulted in the hard coded 7.6.3?

mvkorpel:
I'm on Lion. Using the downloads.haskell.org 8.0.1 binary works for `brew install ghc`. After that, `brew install pandoc` and `brew test pandoc` succeed. Please pardon me for not redoing `brew install ghc` after commit `b061f38` to my branch, as the compilation takes some 2.5 hours on my aging MacBook.

I only know that – for whatever reason – the haskell.org binaries for ghc 7.8.1 and 7.8.2 required 10.9 Mavericks minimum, but all following binaries up to and including 8.0.1 claim to support Lion and later.

ilovezfs:
Interesting. It would sure be nice to know exactly what was fixed, but that doesn't have to block this PR.

It would also be good for someone to verify this change actually fixes the formula and leads to a successful build on 10.7, but I suppose we'll "hear about it" if it doesn't, and the PR is definitely not making anything *worse* than the status quo.

mvkorpel:
Well, I successfully built the Formula as in the first commit `dccece3` of this pull request. On Lion. The following changes, i.e. the "depends_on" and the Sierra patch, should have no effect. I would really prefer not compiling ghc again.

ilovezfs:
>Well, I successfully built the Formula as in the first commit dccece3 of this pull request.

Ah, OK. As long you tested the actual build not just the upstream binaries, that's great :)

Formula/ghc.rb
```diff
@@ -15,6 +15,10 @@ class Ghc < Formula
   deprecated_option "tests" => "with-test"
   deprecated_option "with-tests" => "with-test"
 
+  # https://www.haskell.org/ghc/download_ghc_8_0_1#macosx_x86_64
+  # "This is a distribution for Mac OS X, 10.7 or later."
```
ilovezfs: I'm not sure we need the comment and link. Are we going to update the link every release since that refers to 8_0_1 specific release notes?

MikeMcQuaid:
> Do we know if 8.0.1 actually works on Lion?

They documented it on their download page, I also saw this.

Formula/ghc.rb
MikeMcQuaid: This comment should live above the `resource` I think as that's the thing that we need to check the supported versions on.

mvkorpel: I moved the comment above the `resource`.

MikeMcQuaid:

PR #2044, opened 2016-06-16
ipfs 0.4.2

ilovezfs:
- remove language/go and Go.stage_deps, which weren't being used
- shorten description to please audit
- fix the build, which no longer works with the symlink trick


DomT4:
Seems alright 👍.

PR #2061, opened 2016-06-17
python 2.7.12-rc1 (devel)

ilovezfs:
Also,
- update python resources
- fix pkgshare audit issue

Note that the xcode-stubs-2.7.patch was merged by upstream already, so
isn't needed for 2.7.12-rc1.


ilovezfs:
CC @tdsmith

tdsmith:
awesome thanks!

PR #2074, opened 2016-06-17
bam 0.5.0

ilovezfs:
Also,
- use full path in the test
- minor style changes in the test
- patch a bug with bam's new CC, CXX, etc. detection

### DIFF
--- a/Formula/bam.rb
+++ b/Formula/bam.rb
@@ -1,9 +1,8 @@
 class Bam < Formula
   desc "Build system that uses Lua to describe the build process"
   homepage "https://matricks.github.io/bam/"
-  url "https://github.com/downloads/matricks/bam/bam-0.4.0.tar.gz"
-  sha256 "5e4e4920b4d265da582f66774e9b1ec8ddfbe75ddc028fba86c12f686ea18db3"
-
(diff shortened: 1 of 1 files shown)


ilovezfs:
An alternative to the patch is to work around it in the test with
```
    ENV["CC"] = ENV.cc.split("/")[-1]
    ENV["CXX"] = ENV.cxx.split("/")[-1]
```

ilovezfs:
@UniqMartin Any opinion on the patch vs. the workaround?

UniqMartin:
I'm definitely leaning towards the patch, though ideally we'd hear back from upstream how they feel about this and whether they wish to fix the issue as suggested. (I'd be surprised if they object, but who knows.)

ilovezfs:
PR refreshed with a patch in formula-patches. Upstream has yet to comment on the issue.

ilovezfs:
upstream fixed it! :fireworks: https://github.com/matricks/bam/commit/27b28f0956a78844e73aa94fee7d0187c9b68bbd

ilovezfs:
PR refreshed using upstream's patch instead.

PR #2075, opened 2016-06-17
tuntap deleted

AnastasiaSulyagina:


ilovezfs:
This would be more palatable if the Casks supported source builds where possible as well. It seems like going back to the Dark Ages to take "formula knowledge" of how to build things and just toss it into the boneyard because some pkg/dmg is "preferred."

MikeMcQuaid:
@ilovezfs Given we're relying on an upstream signed kernel extension in this case: it doesn't really matter if we have a formula as it can't be used anyway. Regardless, I agree if upstream do not provide binaries and disagree if they do.

ilovezfs:
That's like saying because a stable tag exists, a head spec should not be provided. I don't see how one follows from the other.

MikeMcQuaid:
@ilovezfs Well, in this case you cannot use the formula on 10.10 and above and we've killed many others for far less. As to the other point: Homebrew was never intended for GUI applications and we handle them far worse than Cask does.

ilovezfs:
I guess since I have kext signing certificates my perspective is completely warped beyond repair at this point. Oh well.

DomT4:
Cross-posting myself since the comment applies here too: https://github.com/Homebrew/homebrew-core/pull/2076#issuecomment-226948639

MikeMcQuaid:
@Homebrew/maintainers Any final (very strong) objections before I merge this?

DomT4:
We need to fix the Requirements first. That's all from me.

MikeMcQuaid:
Cool, thanks @DomT4. @AnastasiaSulyagina we'll talk more about that later.

MikeMcQuaid:
Requirements fix in https://github.com/Homebrew/brew/pull/487.

tap_migrations.json
```diff
@@ -279,5 +279,6 @@
   "xspringies": "homebrew/x11",
   "xulrunner": "homebrew/boneyard",
   "yarp": "homebrew/x11",
-  "ydict": "homebrew/boneyard"
+  "ydict": "homebrew/boneyard",
+  "tuntap": "caskroom/cask"
```
MikeMcQuaid: Can you make sure this is in alphabetical order? Thanks!

MikeMcQuaid:
@AnastasiaSulyagina 😻

PR #2076, opened 2016-06-17
osxfuse deleted

AnastasiaSulyagina:


ilovezfs:
I'm starting to think there should be a kext tap. This PR makes me want to :cry:.

MikeMcQuaid:
@ilovezfs You could create one but not inside the Homebrew org as they aren't supported on 10.10 and above.

DomT4:
We're going to lose the ability to CI test them at all in a few months, and generally, nudging towards signed kernel extensions feels like a good thing to encourage people to do. Agree `tuntap`, `osxfuse`, there's a couple of others I think, need to die off in favour of the Casks.

We'll need to rework our Requirements involved here before these can actually 💀, but I presume Anastasia is on top of that.

ilovezfs:
The fact that you can rebuild your kernel cache with an usigned kext *and keep SIP enabled* seems to be missed here.

DomT4:
That's not something we support, though. Frankly, the previous default where kernel extensions didn't have to be signed by upstream was a risky default & I'm glad Apple changed that. "Don't blindly trust kernel extensions" seems a pretty reasonable policy to me.

ilovezfs:
It strikes me as quiet short-sighted to cut Homebrew off from the entire domain of xnu kernel extension development as if it doesn't exist.

DomT4:
Is already the case for ~89% of Homebrew's userbase, which zero complaints about that to date. We've not exactly had people banging down the door in the last two years saying _"Please change this!!!"_.

If people want to support it in external taps, that's fine, we don't police those, but enabling the use of unsigned kernel extensions where a solid majority of users won't know or consider the risk factor was never a great policy to begin with.

MikeMcQuaid:
> If people want to support it in external taps, that's fine, we don't police those, but enabling the use of unsigned kernel extensions where a solid majority of users won't know or consider the risk factor was never a great policy to begin with.

Completely agree. It'd be good if we could reserve "why is Homebrew stopping doing X" for stuff we're making it impossible for third parties (or maintainers on taps they run outside the Homebrew org) to do rather than the stuff we're not maintaining any more.

MikeMcQuaid:
@Homebrew/maintainers Any final (very strong) objections before I merge this?

DomT4:
We need to fix the Requirements first. That's all from me.

MikeMcQuaid:
Cool, thanks @DomT4. @AnastasiaSulyagina we'll talk more about that later.

MikeMcQuaid:
Requirements fix in https://github.com/Homebrew/brew/pull/487.

tap_migrations.json
```diff
@@ -279,5 +279,6 @@
   "xspringies": "homebrew/x11",
   "xulrunner": "homebrew/boneyard",
   "yarp": "homebrew/x11",
-  "ydict": "homebrew/boneyard"
+  "ydict": "homebrew/boneyard",
+  "osxfuse": "caskroom/cask"
```
MikeMcQuaid: Can you make sure this is in alphabetical order? Thanks!

MikeMcQuaid:
This PR needs rebased on `master` (I suspect there's been a formula change).

MikeMcQuaid:
@AnastasiaSulyagina 😍

PR #2078, opened 2016-06-17
yara: update URLs and remove PCRE dependencies

plusvic:
YARA's official site is now https://github.com/VirusTotal/yara and it doesn't depend on PCRE since version 2.x


apjanke:
Merged. Thanks for your contribution to Homebrew!

Issue #1950, opened 2016-06-13
Recommend cask installs for GUI counterparts of CLI-only formulas?


vitorgalvao:
Opening this issue in the name of consistency, and to see if this is somehow desirable.

[`fontforge`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/fontforge.rb), [`platypus`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/platypus.rb), and [`transmission`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/transmission.rb) are the formulas I found that made me think of this.

Both `platypus` and `transmission` have `caveats` that tell the user only the CLI was installed, and that to install the GUI they should **visit the website**.

`fontforge`, on the other hand, has a comment (not a `caveat`) telling that to get the GUI, they should **use homebrew-cask**.

`depends_on :java`, tells the user to **use homebrew-cask** or **visit the website**.

This is inconsistent, and there may be other formulas like this, but right now, these are what seem like the right changes<sup id="a1">[1](#f1)</sup>:

- [x] Add to `platypus` a note about installing the GUI via homebrew-cask (https://github.com/Homebrew/homebrew-core/pull/2080)
- [x] Add to `transmission` a note about installing the GUI via homebrew-cask (https://github.com/Homebrew/homebrew-core/pull/2081)
- [x] Add a `caveat` to `fontforge` (https://github.com/Homebrew/homebrew-core/pull/2083)
  - [x] Mention getting GUI via website
  - [x] Mention getting GUI via homebrew-cask

---

<sup id="f1">[1](#a1)</sup> I’m not necessarily pushing for homebrew-cask. Either way, consistency could be improved on these.

MikeMcQuaid:
Having caveats for all these sounds good.

DomT4:
Agreed.

DomT4:
All three now have the discussed caveats, per @vitorgalvao's PRs. Closing as fixed.

PR #2083, opened 2016-06-17
fontforge: add caveat about installing GUI

vitorgalvao:

---

Refs #1950.

### DIFF
--- a/Formula/fontforge.rb
+++ b/Formula/fontforge.rb
@@ -89,8 +89,8 @@ class Fontforge < Formula
     system "make"
     system "make", "install"
 
-    # The app here is not functional. You should install Fontforge
-    # via the Cask if you want GUI/App support.
+    # The app here is not functional.
+    # If you want GUI/App support, check the caveats to see how to get it.
     (pkgshare/"osx/FontForge.app").rmtree
 
     if build.with? "extra-tools"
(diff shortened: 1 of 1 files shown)


Formula/fontforge.rb
```diff
@@ -101,6 +101,18 @@ def install
     end
   end
 
+  def caveats
+    <<-EOS.undent
```
DomT4: I wildly appreciate how anal I am, but heavily prefer `def caveats; <<-EOS.undent` with the caveats starting underneath that.

It's the most consistent style throughout formula and I'd generally like to start nudging non-conforming formulae towards it rather than allowing others to drift away. It also saves us from a pointless double indent where people indent two spaces for `<<-EOS.undent` and then indent a further two spaces to actually start the caveats on the line under `<<-EOS.undent`.

vitorgalvao: Sure thing. Done. Altering the other PRs now.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/255396e14730f7d953f9b1f0246af204bf8b3b6c. Thanks @vitorgalvao!

Formula/fontforge.rb
MikeMcQuaid: @vitorgalvao Don't worry about changing but FYI @DomT4 I preferred the way before.

vitorgalvao: I also preferred the way before, though as long as it’s consistent it doesn’t bother me at all.

My reasoning: though the previous way needs an extra line and extra indentation (both undesirable things), this way use a `;` (absolutely a matter of personal preference, but they’re a bit uglier to me in ruby) and the terminating `EOS` is aligned with the text (the alternative would be to align it with the `end`, but that also seems wrong).

But again, consistency above preference.

DomT4: @MikeMcQuaid Essentially, if you want to change a majority of caveats to match your preferred style, that's fine, I don't have an objection. My objection is a slow drift where nobody really knows what to use and we get an inconsistent splattering of style.

PR #2087, opened 2016-06-17
qt5 5.7.0

LRFLEW:

---

I know that major version changes to qt5 have been more involved in the past, such as Homebrew/legacy-homebrew#50234. However, I saw that there was no open PR for this, so I went ahead and made one.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -14,11 +14,11 @@ end
 class Qt5 < Formula
   desc "Version 5 of the Qt framework"
   homepage "https://www.qt.io/"
-  url "https://download.qt.io/official_releases/qt/5.6/5.6.1-1/single/qt-everywhere-opensource-src-5.6.1-1.tar.xz"
-  mirror "https://www.mirrorservice.org/sites/download.qt-project.org/official_releases/qt/5.6/5.6.1-1/single/qt-everywhere-opensource-src-5.6.1-1.tar.xz"
-  sha256 "ce08a7eb54661705f55fb283d895a089b267c688fabe017062bd71b9231736db"
+  url "https://download.qt.io/official_releases/qt/5.7/5.7.0/single/qt-everywhere-opensource-src-5.7.0.tar.xz"
+  mirror "https://www.mirrorservice.org/sites/download.qt-project.org/official_releases/qt/5.7/5.7.0/single/qt-everywhere-opensource-src-5.7.0.tar.xz"
+  sha256 "a6a2632de7e44bbb790bc3b563f143702c610464a7f537d02036749041fd1800"
 
-  head "https://code.qt.io/qt/qt5.git", :branch => "5.6", :shallow => false
+  head "https://code.qt.io/qt/qt5.git", :branch => "5.7", :shallow => false
 
   bottle do
     sha256 "2aaa410f2ab2fbbddbc8c3438e43bc9f4271774c794bcae8f935fb6b1b5a82ed" => :el_capitan
@@ -48,6 +48,20 @@ class Qt5 < Formula
     sha256 "2cf77b820f46f0c404284882b4a4a97bf005b680062842cdc53e107a821deeda"
   end
 
+  # Fix failure to build qtwebengine on Xcode versions 6.1 and 6.2,
+  # which are the latest versions that can run on OS X 10.9.
+  # See <https://codereview.qt-project.org/#/c/164100/>,
+  # <https://codereview.qt-project.org/#/c/164101/>,
+  # and <https://codereview.qt-project.org/#/c/163076>.
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/a49592c440962558737b6881cd291576b960028c/qt5/QTBUG-54486.patch"
+    sha256 "493574df1f69b7a23a04b5d96b11326e4867f237b2495e765d457644020aec8e"
+  end
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/4be35441d1fbc8d86f2f7397aae267afaf951db0/qt5/QTBUG-54190.patch"
+    sha256 "a2f6cc311bfc7a1f5939bf78e8da7464d7d41d3cfdb4ef564cf11c682a3444a2"
+  end
+
   keg_only "Qt 5 conflicts Qt 4 (which is currently much more widely used)."
 
   option "with-docs", "Build documentation"
@@ -59,10 +73,9 @@ class Qt5 < Formula
   deprecated_option "qtdbus" => "with-dbus"
   deprecated_option "with-d-bus" => "with-dbus"
 
-  # OS X 10.7 Lion is still supported in Qt 5.5, but is no longer a reference
-  # configuration and thus untested in practice. Builds on OS X 10.7 have been
-  # reported to fail: <https://github.com/Homebrew/homebrew/issues/45284>.
-  depends_on :macos => :mountain_lion
+  # While Qt 5.7 supports OS X 10.8 Mountain Lion, Building QtWebEngine
+  # requires Xcode 6.1 or later, which needs OS X 10.9 to run
+  depends_on :macos => :mavericks
 
   depends_on "dbus" => :optional
   depends_on :mysql => :optional
--- a/Formula/pyqt5.rb
+++ b/Formula/pyqt5.rb
@@ -3,6 +3,7 @@ class Pyqt5 < Formula
   homepage "https://www.riverbankcomputing.com/software/pyqt/download5"
   url "https://downloads.sourceforge.net/project/pyqt/PyQt5/PyQt-5.6/PyQt5_gpl-5.6.tar.gz"
   sha256 "2e481a6c4c41b96ed3b33449e5f9599987c63a5c8db93313bd57a6acbf20f0e1"
+  revision 1
 
   bottle do
     sha256 "4815f60529eadc829c704adbfb6a01ea2daec7acae4c4f7871579d3b01bdbc63" => :el_capitan


UniqMartin:
Thanks! You beat me to it this time! :wink: I was still in the process of building things locally and performing some tests. Anyway, the build is re-queued with sufficient buffer before it times out. Let's what our build bot has to say about this new major release … I guess we'll know more in about three hours. ⌛

LRFLEW:
One of the builds (mavericks) has already failed. It failed on the `brew linkage --test pyqt5` call. Along with complaining about not finding QtWebEngineCore and QtWebEngineWidgets (which I can see on my local machine with the install I did of this before making the PR), it's complaining about not finding Enginio.framework, which was officially removed from the final release packages with 5.7 ([source](http://wiki.qt.io/New_Features_in_Qt_5.7)). This means that the `pyqt5` package probably needs to be edited to no longer rely on Engine by default.

UniqMartin:
First, let's wait until the other two builds finish before taking any action or jumping to any conclusions.

> One of the builds (mavericks) has already failed.

Unfortunately, this is not looking good. Our Mavericks box is frequently the fastest, but this was way too fast for a Qt build and given the other errors, it makes me think that—for reasons yet unknown to me—the whole QtWebEngine module wasn't built. This would be consistent with the observed build time.

> This means that the `pyqt5` package probably needs to be edited to no longer rely on Engine by default.

Since you already have a working Qt 5.7.0 locally, can you try building `pyqt5` from source? It normally automatically adapts to the frameworks available in a given Qt installation, thus if there is no other breakage, it should be sufficient to bump the formula revision of `pyqt5` (add a `revision 1` line after the first `sha256` and put this in a separate commit titled `pyqt5: revision for qt5` in this PR's branch).

UniqMartin:
When you make changes to the branch, please also remove [these lines (currently 53–55)](https://github.com/Homebrew/homebrew-core/blob/eb78afb0804a66082a887c8d0db66384789926f4/Formula/qt5.rb#L53-L55) in `qt5.rb`. OS X 10.8 is now officially the oldest supported OS X release, so this comment no longer makes sense.

LRFLEW:
`pyqt5` appears to be working after compiling from source on my local machine. I'll go ahead and make the changes now, and let the test run to see if it fails again.

DomT4:
I'm always hopeful Qt has made significant strides in build time reduction, but yeah, from over 2 hours to 50 minutes with those failures seems deeply suspicious.

The generated Mavericks bottle is also only 46.78MB compared to 91.31MB for El Cap and 91.08MB for Yosemite. 91MB is about consistent with the bottles for recent Qt5 releases.

LRFLEW:
Yeah, there's something going wrong with the web engine build on Mavericks. I checked the bottles myself, and the only missing files are related to QtWebEngine, QtWebEngineCore, QtWebEngineWidgets, and QtWebView. I'd like to see the output of the `./configuration` call in the build on Mavericks. I'll see if I know anybody online still on Mavericks to see if I can figure something out.

LRFLEW:
Hmm. While looking up to see if the Web Engine had any additional requirements on top of Qt itself, I found this: http://doc.qt.io/qt-5/qtwebengine-platform-notes.html#os-x

It says that 10.9 (Mavericks) should be able to build the Web Engine, though the page does list some caveats that I'm not sure the build server meets. It says that it needs Xcode 5.1 or later, which seems like it might be a newer version than the one that came with Mavericks. Earlier on the page, it also mentions requiring "Bison, Flex" and "GPerf". From what I can tell, Mavericks (with the command-line tools installed) should have all of those, but I wonder if there was a version difference for one of those libraries between 10.9 and 10.10 that's causing the issue.

UniqMartin:
Yes, there's something weird going on with the QtWebEngine build on Mavericks. This much is clear.

> I'd like to see the output of the ./configuration call in the build on Mavericks.

I re-scheduled the build on our machines and collected the configure logs. Have replaced the somewhat random temporary directory with `@TMP@`, but otherwise these are the full logs. Maybe that helps finding the cause: https://gist.github.com/UniqMartin/2885ddf3bed0e98715d376689380f0bf

> Hmm. While looking up to see if the Web Engine had any additional requirements on top of Qt itself, I found this: http://doc.qt.io/qt-5/qtwebengine-platform-notes.html#os-x

None of the stuff listed there sounds like it should create a problem on Mavericks.

> It says that 10.9 (Mavericks) should be able to build the Web Engine, though the page does list some caveats that I'm not sure the build server meets.

Shouldn't be the issue. All our machines are updated to the latest Xcode still supported on that OS X release, i.e. for 10.9 this is Xcode 6.2 and the matching Command Line Tools (see `00.config-*` in Gist).

LRFLEW:
I've scanned over the logs you provided and didn't find anything incredibly useful unfortunately. I was hoping that the configuration output specified what modules were being compiled, but it only names a few in that output. I'm worried the answer I'm looking for is from the output of the make command, but I don't personally want to scan through an hour of logs. Maybe doing something like "grep error" would reveal something.

The weren't many differences in the configurations, but one difference did stand out. 10.10 and 10.11 differed from 10.9 with the support for c++1z and AVX512, but neither of those should be an issue. The other difference is that the 10.9 config had the "Using gold linker" option turned on. This had me really confused, as my research on the "gold linker" says that it only supports ELF formats. I don't know if this is a *cause* for the issues, but it's the only thing I see so far.

UniqMartin:
The C++1z and AVX stuff is to be expected, as that's related to what the compiler supports and of course the slightly older Clang distributed with Xcode 6.2 doesn't support everything the one from Xcode 7.3.1 supports.

The full build log might help, but it is a bit tricky to capture on our test bot because the test bot makes sure to automatically clean up after itself. Maybe that's something we need to improve so that the logs of a build can be more easily retrieved in case there is no obvious error or not in the last few lines that are printed on a failed build. (But here it doesn't fail, it just skips building one part of the library.)

I agree that the “Using gold linker” stuff is really suspicious, as it doesn't make sense on OS X. To me this sounds like a bug that was introduced in the `configure` script or the related configure test. Not sure if this is the cause, but if it is, we could just pass `-no-use-gold-linker` to `configure` to disable the automatic detection. Would you mind trying that?

LRFLEW:
Well, adding the `-no-use-gold-linker` didn't fix it. I'm kind of at a loss here. The only thing I can think to do right now is to check the Makefile to see if it's a problem during configuration or during the build itself.

UniqMartin:
Yes, sadly no luck there. Thanks for trying and keep the ideas coming if they pop into your head. I'll try to diagnose this further in the coming days, but don't have the time to do that right now.

Ashish-Bansal:
Hi, I would like to inform that Qt 5.6.1 link is dead[0] at[1] as it has been replaced by new revision[2]. I know you are trying to fix problems with Qt 5.7 and will update formula for that but I guess for the time being, master should be fixed for not having that problem :)

[0] https://download.qt.io/official_releases/qt/5.6/5.6.1/single/qt-everywhere-opensource-src-5.6.1.tar.xz
[1] https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt5.rb#L17
[2] https://download.qt.io/official_releases/qt/5.6/5.6.1-1/single/

PS: I have just shifted to Mac from my linux machine and Thanks to homebrew for rescuing me!

UniqMartin:
@Ashish-Bansal Thank you for letting us know! I created #2350 to address this, but it will still take a few hours until this has passed CI, bottles have been built, and the PR can be merged. It's rather annoying when upstream decides to completely remove a previously published release from their download servers.

Ashish-Bansal:
@UniqMartin Thanks! And yes, I completely agree with that, they shouldn't really delete the previous revision from their servers.

The-Compiler:
I tried to take a look at this, but I'm a bit clueless as well. I tried asking in the `#qtwebengine` IRC channel on Freenode, I'll let you know if something comes up.

I also diffed the configure output between 10.9 and 10.11 - other than some failing checks, here's the difference:

```diff
--- 01.configure-10.11	2016-06-30 09:29:48.596819069 +0200
+++ 01.configure-10.9	2016-06-30 09:29:48.610152403 +0200
@@ -1,17 +1,17 @@
 Build options:
-  Configuration .......... accessibility audio-backend avx avx2 avx512cd avx512er avx512f c++11 c++14 c++1z compile_examples concurrent corewlan cups doubleconversion freetype full-config getaddrinfo getifaddrs harfbuzz iconv ipv6ifname large-config largefile medium-config minimal-config nis no-pkg-config opengl pcre png poll_poll precompile_header qpa qpa qt_framework reduce_exports release securetransport shared small-config sse2 sse3 sse4_1 sse4_2 ssl ssse3 system-zlib
+  Configuration .......... accessibility audio-backend avx avx2 c++11 c++14 compile_examples concurrent corewlan cups doubleconversion freetype full-config getaddrinfo getifaddrs harfbuzz iconv ipv6ifname large-config largefile medium-config minimal-config nis no-pkg-config opengl pcre png poll_poll precompile_header qpa qpa qt_framework reduce_exports release securetransport shared small-config sse2 sse3 sse4_1 sse4_2 ssl ssse3 system-zlib use_gold_linker
   Build parts ............ libs tools
   Mode ................... release
   Using sanitizer(s)...... none
-  Using C++ standard ..... c++1z
-  Using gold linker....... no
+  Using C++ standard ..... c++14
+  Using gold linker....... yes
   Using new DTAGS ........ no
   Using PCH .............. yes
   Using LTCG ............. no
   Target compiler supports:
     SSE .................. SSE2 SSE3 SSSE3 SSE4.1 SSE4.2
     AVX .................. AVX AVX2
-    AVX512 ............... AVX512F AVX512ER AVX512CD
+    AVX512 ............... <none>
 
 Qt modules and options:
   Qt D-Bus ............... no
```

So nothing new to what you already mentioned above...

The-Compiler:
FWIW, [this change](https://codereview.qt-project.org/#/c/151109/4//ALL,unified) mentions it needs SDK 10.10*.3* or later - I only can see it's using 10.10 from the build log, maybe with an older patch version?

The-Compiler:
Seems like that's the culprit indeed - IRC logs:

- \<alcroito\> The-Compiler: Afaik webengine 5.7 is not going to be officially supported  for 10.9 anymore
- \<alcroito\> The-Compiler: Also without the build log, it's kind of a guessing game what went wrong
- \<alcroito\> The-Compiler: My initial guess would be that if the build is done with the 10.10 sdk shipped with xcode 6.2, then it will not be enough, because it requires a newer 10.10 sdk
- \<The-Compiler\> alcroito: yeah, I just mentioned that guess as well - seems like 10.10.3 is required?
- \<alcroito\> The-Compiler: Yes, which ships minimum with xcode 6.3, which is not supported on 10.9, which is why we will probably not support it either
- \<The-Compiler\> alcroito: FWIW http://doc.qt.io/qt-5/qtwebengine-platform-notes.html#os-x says "On OS X, Xcode version 5.1 or later on OS X 10.9 or later is required." for 5.7 which implies that'd work
- \<alcroito\> The-Compiler: That is outdated, and we will have to update it. We discovered that 10.9 won't work fairly recently
- \<alcroito\> The initial assumption was that 10.10.3 sdk would be shipped with xcode 6.2, but that's not the case.
- \<The-Compiler\> So homebrew would have to drop QtWebEngine for 10.9 with Qt 5.7?
- \<alcroito\> You can of course take the 10.10.3 from a newer xcode, and build with the older one, but that might not work forever

LRFLEW:
I believe the formula scripts can be fairly complicated. It sounds to me like we may need to have a rule, such that building qt5 without a "no webengine" argument would be unsupported on 10.9. We might be able to get past that by manually replacing the bottle for 10.9 with the 10.10 version, but I don't know how easy it would be to do that...

The other thing we could do is manually update the SDK version. I remember reading something about setting a custom SDK version in Xcode before. It may not be that easy with a build script (instead of an .xcodeproj file), but it might work (and save us the headache later).

UniqMartin:
> FWIW, [this change](https://codereview.qt-project.org/#/c/151109/4//ALL,unified) mentions it needs SDK 10.10*.3* or later - I only can see it's using 10.10 from the build log, maybe with an older patch version?

@The-Compiler Thanks for finding this and for investigating this further! This commit pretty much explains it all. So the bottom line is that the QtWebEngine module cannot be built with an SDK prior to 10.10.3, meaning this automatically excludes OS X 10.9 as the latest officially supported Xcode is 6.2.

The saddening part is that this nontrivial change to the build requirements isn't documented anywhere (as far as I can see), neither in [the documentation you pointed out](http://doc.qt.io/qt-5/qtwebengine-platform-notes.html#os-x) nor the [Qt 5.7.0 Change Files](https://wiki.qt.io/Qt_5.7.0_Change_Files).

Building on a newer machine (but with a deployment target of 10.9) and then using this is theoretically possible, but doesn't work with our CI setup. Modifying the Xcode that we use on our 10.9 CI is not an option (and not practical either), as that could create more trouble for other formulae and will only help with the bottles, but 10.9 users building from source (with Xcode 6.2) would still be left out.

I'm currently seeing these options:

1. Delay shipping Qt 5.7.0 until macOS 10.12 Sierra has been released. At this point we no longer care about OS X 10.9 and stop providing bottles for it anyway. (This is at least practical in the sense that Qt 5.6 is a long-term release and will receive future updates.)

2. Ship Qt 5.7.0 now, but live with the problem that the 10.9 bottle will provide an inferior experience, namely the missing QtWebEngine module and related modules. (This sounds problematic as it is likely to create confusion for users and thus additional support burden for us. And it impacts all other formulae that depend on Qt 5 and might be depending on QtWebEngine.)

3. Ship Qt 5.7.0 now, but raise the minimum required OS X version to 10.10. (This provides a more consistent user experience, but will be annoying to those who want other bits of Qt and don't necessarily require the QtWebEngine module. We'd also have to figure out how to deal with this situation in all formulae that depend on Qt 5.)

If I had to order these options by preference, that would be: 1, 3, 2. @DomT4 What are your thoughts on the situation and the options we have here? Any other options that I'm overlooking?

The-Compiler:
1. sounds quite reasonable indeed if nothing is blocked by (i.e. waiting for) Qt 5.7.0. Seems like it should be released in 3 months (with some optimism)?

I opened [QTBUG-54486](https://bugreports.qt.io/browse/QTBUG-54486) now as I agree this should be documented properly.

LRFLEW:
> Building on a newer machine (but with a deployment target of 10.9) and then using this is theoretically possible, but doesn't work with our CI setup.

Actually, the building is not a problem. The Qt build script sets the minimum OS version to 10.8 for Qt 5.7.0. Look at the binary files in the frameworks in the bottles (or just on your local machine). Using `otool -l`, you get
```
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.8
      sdk 10.11
```
Meaning that the binary will work on any OS X down to (and including) 10.8. As long as you can manually change the bottle, it will work just fine. If we do that, we should be able to make the formula bottle-only on 10.9, right? I don't want to have to wait ~3 months to get the additional features, so 1 just doesn't sound like the best idea to me. Maybe option 3 if we must, but I still think there's a solution to this.

UniqMartin:
> I opened [QTBUG-54486](https://bugreports.qt.io/browse/QTBUG-54486) now as I agree this should be documented properly.

@The-Compiler Thanks!

> Actually, the building is not a problem. The Qt build script sets the minimum OS version to 10.8 for Qt 5.7.0. Look at the binary files in the frameworks in the bottles (or just on your local machine). Using `otool -l`, you get […] Meaning that the binary will work on any OS X down to (and including) 10.8. As long as you can manually change the bottle, it will work just fine.

@LRFLEW It is a problem for us and how our build system is set up. Bottles for 10.9 are built on a 10.9 VM and this means the bottles for 10.9 end up losing the functionality unsupported by upstream on that build system. And we're not going to overthrow our setup just because Qt is acting up.

Personally, I find this more than surprising, as according to the [Qt 5.7 Tools and Versions page](https://wiki.qt.io/Qt_5.7_Tools_and_Versions), their CI and/or build machines are OS X 10.8-10.10, which would mean they don't even test on the most recent OS X release and two out of three configurations are completely useless when it comes to QtWebEngine.

> If we do that, we should be able to make the formula bottle-only on 10.9, right?

Again, as I already tried to make this clear in my previous comment, this would only solve the bottling problem, but create new logistic problems for us. It also doesn't help our 10.9 users that would get Qt if they installed from a bottle, but will be unable to build it themselves. This would create an unacceptable imbalance between bottle and from-source installs. (Some people have to build from source because they use Homebrew in a nonstandard location and the Qt bottles are not relocatable, just to name one valid use case for from-source builds.)

> I don't want to have to wait ~3 months to get the additional features, so 1 just doesn't sound like the best idea to me. Maybe option 3 if we must, but I still think there's a solution to this.

I feel your frustration and I'm not happy with the situation myself. I don't think the wait is going to create problems for projects depending on Qt, as most projects are fairly conservative about picking up newer Qt releases. And there's always the option of using the binaries distributed by upstream.

One other option is to stick with 5.6 in core, but to add a `qt57` formula to the `homebrew/versions` tap (and make that formula require OS X 10.10 or newer). This would allow you (and other users) immediate access to the new features for personal development without making 10.9 users sad.

The-Compiler:
> One other option is to stick with 5.6 in core, but to add a qt57 formula to the homebrew/versions tap (and make that formula require OS X 10.10 or newer). This would allow you (and other users) immediate access to the new features for personal development without making 10.9 users sad.

That sounds like a nice compromise, IMHO.

(not that I'm going to use it - I currently disabled OS X tests for my software until I can get rid of QtWebKit, and maintaning my own formula/bottle just turned out to be too much of a hassle - but that's another story)

LRFLEW:
> This would create an unacceptable imbalance between bottle and from-source installs. (Some people have to build from source because they use Homebrew in a nonstandard location and the Qt bottles are not relocatable, just to name one valid use case for from-source builds.)

Oh, I never really realized that. I guess I was under the false impression that "bottle-only" was a thing. 

That does put us in a tight spot. Unless there was some interest in dropping total support for 10.9 ahead of a 10.12 release (and I understand why you wouldn't want to), that really does make option 1 the only real choice we have.

I think the idea of creating a `homebrew/versions` tap should happen if we don't merge this into the main tap right now. It would be frustrating at the least (I've always liked brew for keeping the taps up-to-date), but it may just be the compromise we need.

DomT4:
This is an irritating change on Qt's part, particularly with minimal announcement. Not the first time they've silently broken something, to say the least.

I've lost count of the amount of hours Martin, me & other contributors have plowed into chasing Qt5 breakage over the past year alone, including where feedback from the community is utterly disregarded such as the removal of `pkg-config` files.

In conversations with @MikeMcQuaid on prior potentially-halting issues with Qt5 the answer has roundly ended up as _"Ship the latest at any cost, it's what people want"_, and that's something we've worked to stick to even when it has been painful. I don't really see an easy way out of this, @UniqMartin accurately summarised the options available to us earlier.

10.9.5 (CI) is our second heaviest Qt5 installer in terms of Homebrew's "known" users, so I think we're likely to step on toes if we kill it completely. On the flipside, 10.9.5 (Non-CI) barely registers, so I think we're on safer ground in terms of making modifications than perhaps the initial stats highlight.

> Ship Qt 5.7.0 now, but live with the problem that the 10.9 bottle will provide an inferior experience, namely the missing QtWebEngine module and related modules. (This sounds problematic as it is likely to create confusion for users and thus additional support burden for us. And it impacts all other formulae that depend on Qt 5 and might be depending on QtWebEngine.)

We need to explore how many `qt5`-using formulae require `QtWebEngine`, but I'm leaning towards this personally. Sierra is only three months away, so unless Qt would consider reversing this upstream I think we're better making a decisive decision here rather than punting.

DomT4:
screen shot 2016-06-30 at 23 02 36

LRFLEW:
@DomT4 I'm a little confused by terminology here. Does "10.9.5 (CI)" mean installed using the 10.9 bottle on OS X 10.9.5? If not, what *does* it mean?

bfontaine:
@LRFLEW “CI” means on Jenkins.

DomT4:
Yeah, the ones not tagged CI are more likely than not humans. Those are the numbers that matter more.

LRFLEW:
Oh, so "CI" means downloaded through a CI like Jenkins or Travis... actually, that would explain a lot.

[Travis uses 10.9.5 by default](https://docs.travis-ci.com/user/osx-ci-environment#OS-X-Version), which is something that can cause a lot of issues for users that don't understand how to specify a version for the Travis builds.

UniqMartin:
Not being a 10.9 user, I don't mind leaving this OS X release behind sooner than later. The problem with a bold move like making Qt require 10.10+ is that we have to figure out what to do with all the formulae that depend on Qt and to be honest I loathe the work this might require. (Making those with a hard dependency also require 10.10+. And the ones with an optional dependency: I don't even know how to deal with them, as I don't think we have a concept of optional dependencies limited to certain OS X releases, or do we?)

> We need to explore how many `qt5`-using formulae require `QtWebEngine`, but I'm leaning towards this personally. Sierra is only three months away, so unless Qt would consider reversing this upstream I think we're better making a decisive decision here rather than punting.

The only one I'm aware of is `pyqt5` as it basically tries to make all of Qt available to Python users. But it just needs to be revision bumped (already done in this PR) and the test modified not to require the Python counterpart of the QtWebEngine module (that's why the build was failing and we actually noticed that something's wrong, aside from the small build time and the much smaller bottle; thanks to Qt for not even printing something useful about why a certain module was auto-disabled). Still, I'd be not too happy shipping an incomplete Qt on 10.9 instead of not shipping it at all.

> [Travis uses 10.9.5 by default](https://docs.travis-ci.com/user/osx-ci-environment#OS-X-Version), which is something that can cause a lot of issues for users that don't understand how to specify a version for the Travis builds.

I never thought about this and assumed everyone would be selecting the latest and greatest OS X image (confusingly named after Xcode releases), but this comment makes a lot of sense and is a really good explanation for the CI numbers we're seeing.

MikeMcQuaid:
I also think we should ship the latest version unless it breaks stuff within Homebrew. When upstream projects don't support OS X versions: that's their problem more than it is ours. Similarly, if people can't rely on the current Homebrew/Travis combination to keep working: they will figure out an alternative. We have a rolling release cycle with no decent multiversion support (for now, at least) so I think our priority should just be ensuring we don't break our own ecosystem.

Placinta:
Hi. I have a patch that will allow compiling QWebEngine 5.7 on OSX 10.9, with the usual requirements of an XCode that has a 10.10 sdk (not 10.10.3), which is 6.1-6.2 for Mavericks. It should be in for the Qt 5.7.1 release. So if you can wait until 5.7.1 is released, I guess you will have less problems with the CI and homebrew stuff.

The-Compiler:
@Placinta nice! Out of curiousity, do you have a link to Qt's gerrit or something?

Placinta:
@The-Compiler 
https://codereview.qt-project.org/#/c/164101/
https://codereview.qt-project.org/#/c/164100/

UniqMartin:
@Placinta This looks great! Thanks for working on that and for letting us know. One change has already been merged and I think as soon as the second one has been merged into the 5.7 branch, too, we can cherry pick those commits and ship a 5.7.0, then remove the patch as soon as 5.7.1 comes around (and includes those changes). Does that sound reasonable?

(To me it seems like the pending change is likely to be merged soon and we can wait until this happens.)

ilovezfs:
Do we know when 5.7.1 is due out?

LRFLEW:
@ilovezfs Not that I know of. Qt is never really regular with the minor updates. Just look at when 5.6.1 came out. Homebrew constantly patches qt5 (there are two patches in the formula right now), so adding another shouldn't be an issue.

ilovezfs:
@LRFLEW have you tested the patch actually builds on 10.9 yet?

LRFLEW:
@ilovezfs No, I don't have a machine with 10.9. I'm trying to get the patch into the PR so that the build bot can give it a try.

UniqMartin:
Patch looks good to me, actually applies against the 5.7.0 sources, and looks reasonable after manual inspection of the two Qt Gerrit changes. I agree we should just let our build bot check whether things work out as intended …

Formula/qt5.rb
```diff
@@ -59,9 +68,6 @@ class Qt5 < Formula
   deprecated_option "qtdbus" => "with-dbus"
   deprecated_option "with-d-bus" => "with-dbus"
 
-  # OS X 10.7 Lion is still supported in Qt 5.5, but is no longer a reference
-  # configuration and thus untested in practice. Builds on OS X 10.7 have been
-  # reported to fail: <https://github.com/Homebrew/homebrew/issues/45284>.
   depends_on :macos => :mountain_lion
```
UniqMartin: Let's bump this to `:mavericks` given what we've learned from the QtWebEngine issue.

UniqMartin:
@BrewTestBot test this please

UniqMartin:
@ilovezfs Why did you abort the [most recent build](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4085/)?

ilovezfs:
I will trigger it as soon as https://github.com/Homebrew/homebrew-core/pull/2555 starts.

ilovezfs:
@UniqMartin Build is now off to the races.

LRFLEW:
*fingers crossed it doesn't time out*

ilovezfs:
@LRFLEW Martin set the time out at four hours, so I sure hope not!

DomT4:
Argh. There are further 10.9 issues with `qtwebengine` apparently:

```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk -O2 -gdwarf-2 -fvisibility=hidden -mmacosx-version-min=10.7 -arch x86_64 -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wno-selector-type-mismatch -Wheader-hygiene -Wno-char-subscripts -Wno-unneeded-internal-declaration -Wno-covered-switch-default -Wstring-conversion -Wno-c++11-narrowing -Wno-deprecated-register -Wno-inconsistent-missing-override -Wno-tautological-compare -Wno-unknown-warning-option -std=c++11 -stdlib=libc++ -fno-rtti -fno-exceptions -fvisibility-inlines-hidden -fno-threadsafe-statics -fcolor-diagnostics -fno-strict-aliasing -Wobjc-missing-property-synthesis -fobjc-call-cxx-cdtors  -c /private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/device/bluetooth/bluetooth_socket_mac.mm -o obj/src/3rdparty/chromium/device/bluetooth/device_bluetooth.bluetooth_socket_mac.o
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/device/bluetooth/bluetooth_socket_mac.mm:603:14: error: use of overloaded operator '=' is ambiguous (with operand types 'scoped_ptr<device::BluetoothChannelMac>' and 'scoped_ptr<device::BluetoothRfcommChannelMac>')
    channel_ = BluetoothRfcommChannelMac::OpenAsync(
    ~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/base/memory/scoped_ptr.h:241:3: note: candidate function has been explicitly deleted
  DISALLOW_COPY_AND_ASSIGN_WITH_MOVE_FOR_BIND(scoped_ptr)
  ^
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/base/move.h:38:8: note: expanded from macro 'DISALLOW_COPY_AND_ASSIGN_WITH_MOVE_FOR_BIND'
  void operator=(const type&) = delete;                         \
       ^
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/base/memory/scoped_ptr.h:312:15: note: candidate function
  scoped_ptr& operator=(scoped_ptr&& rhs) {
              ^
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/device/bluetooth/bluetooth_socket_mac.mm:607:14: error: use of overloaded operator '=' is ambiguous (with operand types 'scoped_ptr<device::BluetoothChannelMac>' and 'scoped_ptr<device::BluetoothL2capChannelMac>')
    channel_ =
    ~~~~~~~~ ^
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/base/memory/scoped_ptr.h:241:3: note: candidate function has been explicitly deleted
  DISALLOW_COPY_AND_ASSIGN_WITH_MOVE_FOR_BIND(scoped_ptr)
  ^
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/base/move.h:38:8: note: expanded from macro 'DISALLOW_COPY_AND_ASSIGN_WITH_MOVE_FOR_BIND'
  void operator=(const type&) = delete;                         \
       ^
/private/tmp/qt5-20160701-75582-10cq4n8/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/base/memory/scoped_ptr.h:312:15: note: candidate function
  scoped_ptr& operator=(scoped_ptr&& rhs) {
              ^
2 errors generated.
make[4]: *** [invoke_ninja] Error 1
make[3]: *** [sub-gyp_run-pro-make_first] Error 2
make[2]: *** [sub-core-make_first] Error 2
make[1]: *** [sub-src-make_first] Error 2
make: *** [module-qtwebengine-make_first] Error 2
```

LRFLEW:
I noticed that. I can't do it right now, but I plan on trying to build the web engine locally manually disabling the force touch code to see if the patch is the problem.

DomT4:
Thanks. I've killed the build for now since we know this builds on 10.10 & up, 10.9 was the question mark. Let us know how the local testing goes, and thank you for your work here to date.

UniqMartin:
Yeah, noticed those failures too, but figured there was no harm in letting the build continue nonetheless. (But made sense to kill it; thanks Dom!) Qt isn't making things easy for us with this release. 😭

@LRFLEW ~~Additionally, you could also try adding back the `-no-use-gold-linker` flag to `configure`. It is probably not related to the failures we're seeing now, but the linker misdetection is still bothering me a bit and maybe it turns out it has some impact.~~ Actually, I guess the `gold` linker misdetection has been here for quite a while and I don't think this particular issue is Qt's fault. Turns our that Clang prior to some version (haven't researched further) silently ignores the `-fuse-ld=<whatever>` flag while newer versions respect it and return with a nonzero exit code if the requested linker is unavailable.

DomT4:
> Yeah, noticed those failures too, but figured there was no harm in letting the build continue nonetheless. 

I was going to let it continue, but then I realised due to the `pyqt5` revision bump it'd attempt to build `qt5` from source twice in the case where the first `qt5` build failed, and I couldn't convince myself on the merits of another pointless 70 minutes for the job to fail 🙈.

LRFLEW:
In my attempt to avoid compiling the entire Qt library, I accidentally set make to run in a single thread, so this is going to take a while. In the mean time, I was looking at the error, and found it peculiar. It's seems to be suggesting that there's an ambiguity of which `operator=` to use: the one that's defined and the one explicitly deleted. It really doesn't make sense to me, so I'm going to keep looking into it.

UniqMartin:
Could it be that the build is exposing lacking C++11/C++14 support in the Clang version shipped with Xcode 6.2? The error arises in the WebEngine code so I'm not at all positive that this is the cause, but then again the major difference between Qt 5.6 and Qt 5.7 is that the latter fully embraces C++11 and makes it a requirement. Maybe using `-c++std c++11` as a configure option instead of the auto-detected `c++14` to limit the C++ version it tries to use during the build makes a difference?

LRFLEW:
I think it's worth trying. I'm not 100% sure how to do that (I have some, but limited, ruby experience). Would the appropriate line be `args << "-c++std c++11" if :macos == :mavericks`?

UniqMartin:
> Would the appropriate line be `args << "-c++std c++11" if :macos == :mavericks`?

Almost, though fortunately this should cause the build to fail early in the `configure` phase. You would need to pass this as two separate flags and the check for the OS X release looks a bit different:

```ruby
args << "-c++std" << "c++11" if MacOS.version <= :mavericks
```

LRFLEW:
Ok, I've pushed the change as a new commit. If the build succeeds, I'll rebase it to remove the commit. If not, I'll remove the commit. Let's just see what happens.

LRFLEW:
Ok, I have no clue what's going on at this point. I'm doubting that adding that flag actually did anything, given that DomT4's comment showed that it was already passing `-std=c++11` to the compiler. I think we're also getting into race conditions for what error is shown, given that we've had 3 builds with the patch, and they each reported a different error.

UniqMartin:
> Ok, I have no clue what's going on at this point. I'm doubting that adding that flag actually did anything, given that DomT4's comment showed that it was already passing `-std=c++11` to the compiler.

Right, I didn't notice this. The idea with C++11/C++14 was really speculative and apparently not true.

Though there's still some hope. Apparently we're hitting a known problem and it even has a solution:

- <https://bugreports.qt.io/browse/QTBUG-54190>
- <https://codereview.qt-project.org/#/c/163076>

The comments in the bug indicate otherwise, but the errors we're seeing are exactly in the spots where the patch introduces changes. I think we're only seeing somewhat different errors in every run because the build is parallelized and thus the order of compilation is different on every run.

Placinta:
The ambiguity operator= was a bug in the clang compiler.
Like @UniqMartin mentioned, it was fixed in 5.7 git HEAD, with the https://codereview.qt-project.org/#/c/163076 patch.

LRFLEW:
> Build timed out (after 60 minutes). Marking the build as aborted.

I thought this was set to 4 hours? Can someone restart this with a longer timeout?

DomT4:
I reset it after I cancelled last night's build. Sadly Jenkins isn't flexible enough to assign a timeout _per_ PR, you either bump them all or bump none. I'll raise it again & requeue now.

LRFLEW:
It was SO CLOSE!

```
In file included from /private/tmp/qt5-20160703-84038-tmj2ne/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/content/common/gpu/ca_layer_tree_mac.mm:5:
/private/tmp/qt5-20160703-84038-tmj2ne/qt-everywhere-opensource-src-5.7.0/qtwebengine/src/3rdparty/chromium/content/common/gpu/ca_layer_tree_mac.h:172:46: error: cannot initialize a member subobject of type 'CAEdgeAntialiasingMask' with an rvalue of type 'int'
    CAEdgeAntialiasingMask ca_edge_aa_mask = 0;
                                             ^
1 error generated.
```

I will try to see if patch has been made for this yet, but this seems to be getting ridiculous to me.

MikeMcQuaid:
Let's just ship this without Mavericks support. It doesn't make sense for us to be upstream's guinea pig for every release when it's this broken.

LRFLEW:
With the CI statistics, I'm hesitant to say that we should completely drop support, but I will bring up my suggestion from before. Have separate "bottle" and "build" version requirements could be the way to go. Brew already has a system in place for having build-only dependencies. It would take a change in the code for brew, but a way of being able to set a distinct max-build OS version from a max-use OS version could be useful. If we did that, we could also see about creating a single bottle version that works for everybody (compiled on the latest SDK, but with the `-mmacosx-min-version` attribute set to the minimum supported runtime). Doing that would help me personally for projects where I want to distribute builds of libraries with a project.

MikeMcQuaid:
> If we did that, we could also see about creating a single bottle version that works for everybody (compiled on the latest SDK, but with the -mmacosx-min-version attribute set to the minimum supported runtime). Doing that would help me personally for projects where I want to distribute builds of libraries with a project.

We're not going to do that, sorry.

We don't want to try and attempt to support something as vast as Qt when upstream do not support it. It'll just mean we have to effectively do their 10.9 QA for them and fix whatever issues crop up on a given release. We're also not going to catch any runtime issues (e.g. rendering bugs that may be specific to <=10.9) this way.

UniqMartin:
> Let's just ship this without Mavericks support. It doesn't make sense for us to be upstream's guinea pig for every release when it's this broken.

Sad, but true. Given the current state of this, I pretty much agree. I'm still not thrilled with the implications this has for other formulae and the resulting work for us. If I'm not mistaken, this decision means:

- The `depends_on :macos` line in the `qt5` formula needs to be bumped to `:yosemite`.
- All formulae with a mandatory dependency on `qt5` need a `depends_on :macos => :yosemite` line.
- All formulae with an optional/recommended dependency on `qt5` need their `depends_on "qt5"` line extended with a trailing `if MacOS.version >= :yosemite`.

I this accurate or am I overlooking something?

MikeMcQuaid:
@UniqMartin Sounds right 😭 👍

ilovezfs:
Perhaps that could be handled with an audit rule on a "just in time" basis in case this actually get fixed.

DomT4:
Killing it completely on Mavericks somewhat pains me, especially since we're planning to ditch `qt` dependencies as much as possible in favour of `qt5` dependencies.

Either we're effectively bumping vast majority of Qt usage in Homebrew formulae to Yosemite & above, or we're going to end up introducing a system where x formula uses `qt` on 10.9 & below, and `qt5` above there. Neither option smells great.

Disabling `QtWebEngine` seems a reasonable compromise, but does require extra work, and as Mike points out we can't catch runtime problems that may slip through if Qt aren't bothered about 10.9 now 😢.

Placinta:
I'd want to re-iterate that QWebEngine 5.7 should still compile with 10.9, I have tried this personally with the latest 5.7.x branch. This should be confirmed in the coming days when the Qt 10.9 CI box does a build of WebEngine. Note that I mention 5.7.x HEAD (which will eventually be 5.7.1), and not 5.7.0. There are probably some code difference between 5.7.0 and HEAD, which led to the additional compilation issues you've encountered.

ilovezfs:
Not sure if I've missed any, but these appear to be the 99 directly or indirectly impacted formulae:
https://gist.github.com/ilovezfs/aba90a055a181ec6a15481ec12445eb3

MikeMcQuaid:
> I'd want to re-iterate that QWebEngine 5.7 should still compile with 10.9, I have tried this personally with the latest 5.7.x branch. This should be confirmed in the coming days when the Qt 10.9 CI box does a build of WebEngine.

Why does this CI box not get used before releases get made?

Placinta:
The box was used, but it didn't build WebEngine because of the check to have the required SDK 10.10.3, which the box didn't have (just like yours didn't).

UniqMartin:
I guess the question then is: Why are there no checks in place that make sure that all the expected frameworks (Qt modules) have been built that are supposed to be available on a given platform (at least on those CI boxes where the environment is clear)? Maybe I'm being naive here, but not having these in place and not noticing that QtWebEngine is skipped on those CI boxes feels like a major omission.

If you have any insight in the CI process, do you know why [according to this wiki page](https://wiki.qt.io/Qt_5.7_Tools_and_Versions) there's still a box (or test configuration) for the legacy 10.8 release, but none for the 10.11 release (despite it being pretty seasoned now and macOS 10.12 already on the horizon)?

In any case: Thanks for your work on actually resolving the issues we've come across here and for commenting on this formula PR! That's much appreciated!

Placinta:
The 10.11 CI box is (to my knowledge) in the review process. Qt uses a custom still-developed CI solution, so it takes longer to deploy a new box compared to other CI solutions. And because Qt has a big codebase, and a change has to be successfully built on all CI boxes, introducing new platforms to the CI takes longer (making sure it builds everywhere).

QWebEngine was successfully built on the CI platforms we currently have, except for 10.9 where as I have mentioned, the official SDK was not enough. The fact that the documentation was not updated, was due to my belief that the CI box just had an older Xcode version which needed to be updated. Unfortunately I missed that the latest supported Xcode still didn't have the required SDK. And as that happened too late, the 5.7.0 release didn't get documentation / configuration changes to warn about the situation.

Which is why I hope to make 5.7.1 build properly on 10.9.

jakepetroules:
> We're not going to do that, sorry.
> We don't want to try and attempt to support something as vast as Qt when upstream do not support it.

What are you talking about? This is *exactly* what we support and in fact it's basically the only use case we support. We are also moving towards making the latest Xcode SDKs a hard requirement for building Qt, so you guys are at some point going to have to solve the problem of not being able to have separate build and runtime requirements. It's how you're supposed to build macOS software in general.

ilovezfs:
`depends_on :macos => :sierra` should solve that for us.

DomT4:
> We are also moving towards making the latest Xcode SDKs a hard requirement for building Qt, so you guys are at some point going to have to solve the problem of not being able to have separate build and runtime requirements.

With respect, unless you're willing to help implement it, we can live without advice on what we must or must not do from Qt.

MikeMcQuaid:
> What are you talking about? This is exactly what we support

OS X 10.9? Then why were there build failures?

> We are also moving towards making the latest Xcode SDKs a hard requirement for building Qt, so you guys are at some point going to have to solve the problem of not being able to have separate build and runtime requirements. 

We already have these.

> It's how you're supposed to build macOS software in general.

Thanks but I/we have been building OS X software for quite some time.

MikeMcQuaid:
>> We are also moving towards making the latest Xcode SDKs a hard requirement for building Qt, so you guys are at some point going to have to solve the problem of not being able to have separate build and runtime requirements.
> We already have these.

I assume you're meaning "you're going to need to start support building software on 10.11 and distributing on 10.9". No, we don't have to solve this problem. If Qt starts to require the latest Xcode then the entire Qt ecosystem can now only can have their dependency tree built from source on the versions of OS X that can run that Xcode which are normally the current and maybe previous version. We will just cap our support to the versions you claim you support as we are primarily a source-based package manager that provides packages to users using our default prefix and relatively new OSs but still an acceptable fallback for those that do not use one or both of these.

jakepetroules:
> OS X 10.9? Then why were there build failures?

I was referring to building with the latest SDK and setting -mmacosx-version-min to the latest version we support at runtime; that's the intended use case for Qt (and any binary distribution(s) of macOS software).

Build failures are because QtWebEngine supports macOS 10.9 but requires building with at least the 10.10.3 SDK.

> We already have these. I assume you're meaning "you're going to need to start support building software on 10.11 and distributing on 10.9". No, we don't have to solve this problem. If Qt starts to require the latest Xcode then the entire Qt ecosystem can now only can have their dependency tree built from source on the versions of OS X that can run that Xcode which are normally the current and maybe previous version. We will just cap our support to the versions you claim you support as we are primarily a source-based package manager that provides packages to users using our default prefix and relatively new OSs but still an acceptable fallback for those that do not use one or both of these.

Are you saying you have this feature but you're unwilling to restrict building to the latest OS version while still providing bottles for older OS versions? Why is that not possible?

jakepetroules:
Also note that building on older OSes significantly increases our testing burden -- 4 OSes times 4 SDKs is 16 possible configurations. We cannot possibly test all of those. 4 OSes times 1 SDK (the latest) is 4 configurations, which is much more manageable.

We don't want people reporting bugs because they built with the 10.9 SDK and ran on 10.11 and encountered some behavioral difference as a result (and 99% of the time no one will have thought to provide this information).

LRFLEW:
> 4 OSes times 4 SDKs is 16 possible configurations. We cannot possibly test all of those. 4 OSes times 1 SDK (the latest) is 4 configurations, which is much more manageable.

Um, not really. All that really matters with SDKs is what version comes with what Xcode version. Each version of OS X has a max-supported version of Xcode, and that's all you need to test against with each OS. So it's 4 OSes, 4 SDKs, but still just 4 configurations.

jakepetroules:
I'm talking about runtime requirements in addition to build requirements. You still need to test every one of those builds you create, against every OS version it supports (which is all of them). Perhaps for Homebrew this is less of a concern because you only ever use the build you create on the OS version you're running on, but for bottles, upstream, and for distributing Qt to users, it's important.

ilovezfs:
If you're running on an OS that can only run the software, but cannot build it, then that means Homebrew source builds (--build-from-source) won't work for you, and so it also means you won't be able to build with any non-default options (--with-foo --without-bar). So in those cases, brew would be solely a fixed binary distribution tool.

In the brew ecosystem, fixed binary distributions are primarily handled by Homebrew Cask, which might be a more natural fit for Qt going forward.

DomT4:
> Why is that not possible?

For a start, unlike Qt, Homebrew doesn't offer a commercial product & consequently like the vast majority of FOSS projects runs off extremely limited funds & resources.

Essentially, what you're asking Homebrew to do is invest time rewriting a system that pretty much works for the vast majority of other formulae primarily for the benefit of keeping the `qt5` formula _(which happens to rank 70th in our analytics for direct installations, `qt` is 76th)_ supported across older OS X versions.

I'm not arguing that Homebrew's SDK logic is perfect, but I take some slight umbrage at Qt complaints about limited resources whilst demanding Homebrew redesign existing mechanisms to suit y'all.

I note that more than once Homebrew has had to dance around Qt decisions to ensure other packages can continue to build against Qt/Qt5, including most prominently recently Qt5's decision to [throw pkg-config files in the trash](https://codereview.qt-project.org/#/c/140954/). That broke at least Wireshark, Poppler & mkvtoolnix being able to build their Qt5 options, and indeed one of the Wireshark developers alerted Qt upstream this was going to be a problematic decision.

I note also that Homebrew was _already_ working around the Frameworks issue in two different ways:
```ruby
    # Some config scripts will only find Qt in a "Frameworks" folder
    frameworks.install_symlink Dir["#{lib}/*.framework"]

    # The pkg-config files installed suggest that headers can be found in the
    # `include` directory. Make this so by creating symlinks from `include` to
    # the Frameworks' Headers folders.
    Pathname.glob("#{lib}/*.framework/Headers") do |path|
      include.install_symlink path => path.parent.basename(".framework")
    end
```

I highlight those two issues because there's an implication to your comment that Homebrew's current build system will result in bugs being filed upstream to you, and that _may_ be fair, but Homebrew has received & handled it's fair share of dealing with Qt5-caused breakage with no significant support from Qt in doing so.

We are _happy_ to try and support Qt/Qt5 to the best of our ability but a little recognition that Qt makes some decisions that makes life harder for us without our concerns being swatted away wouldn't go amiss Jake.

MikeMcQuaid:
> I was referring to building with the latest SDK and setting -mmacosx-version-min to the latest version we support at runtime; that's the intended use case for Qt (and any binary distribution(s) of macOS software).

As mentioned: we are not a binary package manager but a from-source one that provides binary packages for some situations.

> Build failures are because QtWebEngine supports macOS 10.9 but requires building with at least the 10.10.3 SDK.

Ok, you support 10.9 but not the tools required to build on 10.9. In Homebrew-land that's not supporting 10.9.

> Are you saying you have this feature but you're unwilling to restrict building to the latest OS version while still providing bottles for older OS versions? Why is that not possible?

Our binary package system (which I built) builds each binary package on the OS on which it'll be used. This may produce redundant packages on occasion and may produce more packages than is needed but, given we want to do build CI on each platform it only really increases our binary package storage space while optimising for safety. Your way would be possible if you were designing another package manager but that's not the way things were built in Homebrew and it doesn't seem to be worth the effort to make the change (and, as was pointed out above: certainly isn't worth the effort for Qt alone).

> Also note that building on older OSes significantly increases our testing burden -- 4 OSes times 4 SDKs is 16 possible configurations. We cannot possibly test all of those. 4 OSes times 1 SDK (the latest) is 4 configurations, which is much more manageable.
> 
> We don't want people reporting bugs because they built with the 10.9 SDK and ran on 10.11 and encountered some behavioral difference as a result (and 99% of the time no one will have thought to provide this information).

> I'm talking about runtime requirements in addition to build requirements. You still need to test every one of those builds you create, against every OS version it supports (which is all of them). Perhaps for Homebrew this is less of a concern because you only ever use the build you create on the OS version you're running on, but for bottles, upstream, and for distributing Qt to users, it's important.

That's up to you. We don't do that and don't require that. On the flip side, as @DomT4 mentioned, we're all volunteers working in our free time and all our CI hardware was funded by a Kickstarter and setup in my house before I physically took them to a donated data centre. Between us we've submitted a fair number of issues (usually ignored) and patches for build issues on your Qt stable and beta releases that don't build on OS X (sometimes even the latest version you claim to support). You should probably decide at some point to either cap your `configure` to not build on OS X versions you support, clearly document what OS X versions you support (apologies if you've done this already, I haven't looked recently) and/or get CI hardware to handle these configurations rather than leaving it to downstream projects to fix these problems.

UniqMartin:
> You should probably decide at some point to either cap your `configure` to not build on OS X versions you support, clearly document what OS X versions you support (apologies if you've done this already, I haven't looked recently) and/or get CI hardware to handle these configurations rather than leaving it to downstream projects to fix these problems.

This. Often enough we might overlook information that is available, but sadly often enough those changes/requirements are neither mentioned in the documentation nor pointed out by `configure` nor mentioned in the release notes or change logs. Frankly, if 10.9 is not supported as a build configuration (as opposed to run time), `configure` should be really clear about this and produce a hard failure, but it currently doesn't even print a warning. Silently skipping QtWebEngine during the build because some additional requirement isn't met is more than just surprising.

haraldF:
ok, loads of information :) Can we get a quick tl;dr?

From what I can see, Qt 5.7 is not in homebrew as QtWebEngine doesn't compile on older OS X versions.

However, there's also a comment indicating that this is fixed in the 5.7 - branch, so it's a matter of backporting a couple of fixes? Or is is generally not supported by the Qt folks any more? (In which case I might be able to help by driving over to the Qt Adlershof office with a crate of beer ;))

LRFLEW:
> Can we get a quick tl;dr?

@haraldF The "quick tl;dr" is that an ideological war is starting on how these sorts of projects should be distributed. 

As to the status of the PR, we're already tried 2 patches from the 5.7 branch, but are still experiencing compile errors. There has been some discussion about just dropping support for the versions of OS X that are failing, but that would involve editing a large number of formulas dependent on it. If you know of what other patches are needed to get this working, link them here, and I'd be willing to update the PR with them and let the bot take another attempt at it (if the project members are willing to let it run for long enough).

UniqMartin:
Yes, @LRFLEW summarized it nicely. I think the simple/desirable solution is one of:

- Identify the missing changes between 5.7.0 and the 5.7 branch that hopefully restore compatibility with 10.9 as a build configuration, for reasons explained above.
- Wait for 5.7.1 that should include all those fixes.

All other solutions require more work (and will probably result in some grief for the 10.9 folks) and also mean someone would need to step up and do this work.

The-Compiler:
FYI: Looks like that requirement [gets bumped](https://codereview.qt-project.org/#/c/166828/) to 10.10 again with Qt 5.8 as Chromium doesn't build anymore otherwise.

LRFLEW:
Hmm, if that's true, then we may want to reconsider what option we take. @UniqMartin What's your thoughts?

The-Compiler:
Though Alexandru Croitor (== @Placinta?) apparently did another patch which is [in review](https://codereview.qt-project.org/#/c/166541/) to get it to build again, so there's hope... What a roller coaster :wink:

UniqMartin:
Since 10.9 is no longer a supported build environment for the QtWebEngine module (inherited from Chromium that it is built on top of) but we want to either ship a complete Qt will all of its modules and we want/have to continue to support from-source builds, I'm increasingly of the opinion that we should just bite the bullet and raise the minimum macOS requirement to 10.10.

Though Alexandru Croitor has so far done a wonderful job of restoring compatibility with older build environments on macOS, this will always be a game of catching up and may very soon turn out to be impractical. And I fear it's always going to be an afterthought in the overall Qt release process, so future breakage seems likely.

If @DomT4 and @MikeMcQuaid don't have strong feelings that we should pursue a different path, let's just raise the minimum macOS requirement and kill 10.9 support in `qt5` and all formulae that have a hard dependency on `qt5`. 10.9 support will soon be phased out anyway as 10.12 is already on the horizon and our <= 10.9 demographic is fairly small at this point.

Placinta:
To be fair, if I do end up fixing 10.9 compilation on Qt 5.8, and it is fixed for 5.7.1, the only catching up will have to be for patch releases of 5.8. And given that the Chromium version will not be updated anymore for the 5.8 patch releases of Qt, there is little chance something will break after the initial hurdle. And in Qt 5.9, 10.9 support will be officially dropped for all Qt modules.

MikeMcQuaid:
> Since 10.9 is no longer a supported build environment for the QtWebEngine module (inherited from Chromium that it is built on top of) but we want to either ship a complete Qt will all of its modules and we want/have to continue to support from-source builds, I'm increasingly of the opinion that we should just bite the bullet and raise the minimum macOS requirement to 10.10.

> And I fear it's always going to be an afterthought in the overall Qt release process, so future breakage seems likely.

I agree 100%. If/when our users complain we can point them to Qt to voice their complaints.

DomT4:
Aye. It might be good to give Travis a heads-up given a decent chunk of our Qt5 usage is CI based and Travis' default CI is 10.9, but in principle I agree we're fighting a losing battle here.

MikeMcQuaid:
@DomT4 We probably owe Travis an email anyway to try and convince them to update their images to the split brew/core repos.

DomT4:
Yeah. It'd be helpful to everyone if they upgrade some of their preinstalled formulae as well; at the moment a wad of people using Travis have had to add `upgrade` checks to `libtool` & `pkg-config` after the changes we made internally.

A general update & raising the default OS X image to at least 10.10 would be ideal on all sides, IMO. We can at least reassure them that once Homebrew is 1.0 the breaking changes to core code are going to be less frequent.

MikeMcQuaid:
> A general update & raising the default OS X image to at least 10.10 would be ideal on all sides, IMO. We can at least reassure them that once Homebrew is 1.0 the breaking changes to core code are going to be less frequent.

Alternatively: we wait until 1.0 and then guarantee it.

MikeMcQuaid:
@LRFLEW Any news on this? Thanks!

MikeMcQuaid:
Will wait for Qt 5.7.1.

schoeps:
... or ship a snapshot?

schoeps:
qt5-qtconnectivity seems to be fixed https://trac.macports.org/ticket/52203

MikeMcQuaid:
If we can get something that compiles on our now three supported platforms (10.10 - 10.12) I'm happy to include it.

PR #2095, opened 2016-06-18
stormpath-cli 0.1.1

ilovezfs:


Formula/stormpath-cli.rb
```diff
@@ -14,65 +14,87 @@ class StormpathCli < Formula
...
+  resource "termcolor" do
+    url "https://files.pythonhosted.org/packages/8a/48/a76be51647d0eb9f10e2a4511bf3ffb8cc1e6b14e9e4fab46173aa79f981/termcolor-1.1.0.tar.gz"
+    sha256 "1d6d69ce66211143803fbc56652b41d73b4a400a2891d7bf7a1cdf4c02de613b"
   end
 
   def install
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
-    resources.each do |r|
-      r.stage { system "python", *Language::Python.setup_install_args(libexec/"vendor") }
+    %w[cssselect docopt isodate lxml oauthlib PyDispatcher PyJWT pyquery python-dateutil requests six stormpath termcolor].each do |r|
```
DomT4: Unless one of the resources needs special attention we do actually prefer the:
```ruby
    resources.each do |r|
      r.stage do
      end
    end
```
style syntax.

ilovezfs: Yeah, this is taken directly from `poet -f` output, so I was curious how it would go over with you. @tdsmith perhaps the `poet -f` default output should be as @DomT4 has it above not the explicit list?

tdsmith: I prefer the explicit form but I don't feel strongly; if others don't like it maybe that's a good decision.

DomT4: Oh, I just use `poet` to grab the resource blocks and do the rest myself. Explains why I hadn't noticed this before.

I don't have a strong objection to being explicit but I've noticed generally with explicit blocks when people add/remove resources they initially forget to update the array. The line should be wrapped as close to ~80 characters as possible, as well, to avoid unnecessarily stretchy formulae.

ilovezfs:
@DomT4 :ship: ?

DomT4:
👍

PR #2098, opened 2016-06-18
planck 2.0-alpha1-devel

ilovezfs:
Also, fix a case sensitivitity issue.

### DIFF
--- a/Formula/planck.rb
+++ b/Formula/planck.rb
@@ -12,16 +12,26 @@ class Planck < Formula
     sha256 "7e2b80e0c2ce8269c6614209fe796b2286c217fc31a560b3491680ba655a490b" => :mavericks
   end
 
+  devel do
+    url "https://github.com/mfikes/planck/archive/2.0-alpha1.tar.gz"
+    version "2.0-alpha1"
+    sha256 "3a1438bbefa5bb6c48b8c5470a53b200198c21cb8186e42f89f93a71e9f6a072"
+  end
+
   depends_on "leiningen" => :build
   depends_on :xcode => :build
   depends_on :macos => :mavericks
 
   def install
+    # Fixes "No such file or directory - build/Release/planck"
+    # Reported 17 Jun 2016: https://github.com/mfikes/planck/issues/322
+    inreplace "script/build", "$(PWD)", "$PWD"
+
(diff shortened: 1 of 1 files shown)


mfikes:
I am the maintainer of Planck. I'd prefer that this alpha not be released this way. I can't guarantee that the tags will remain stable, and additionally I have a PR open for Planck 1.15. Can this PR be declined?

ilovezfs:
@mfikes certainly, but you're aware this can only be installed if the user deliberately passes `--devel` right?

mfikes:
@ilovezfs No. Thanks. I wasn't aware of the `--devel` flag. But, my preference would be to not have to maintain the alpha tags (I plan on deleting them), nor the consequences to `planck.rb`, nor dealing with Planck being released this way. (Of course, I don't control Hombebrew—just expressing my preference.)

ilovezfs:
@mfikes I'm definitely closing this, then :) I just wanted to make sure you were aware the `:stable` and `:devel` specs are orthogonal. If you'd like to do a `:devel` release of the `2.*` series at some point before blessing it `:stable`, let us know!

mfikes:
@ilovezfs Thanks! Perhaps I'll consider this when Planck 2.0 reaches beta quality. (Have been using `--HEAD` in general to install master without having to deal with tags.)

ilovezfs:
>Have been using --HEAD in general to install master

Nice! Always cool to hear when upstreams find ways to use brew productively themselves, not solely for distribution.

mfikes:
@ilovezfs Cool. In hindsight, FWIW, I don't even think this formula was building the 2.0 alpha. I haven't looked at it carefully, but it appears to be building the 1.x (Objective-C) release, not the new 2.0 C port.

ilovezfs:
Ah so they coexist in the same source tree but `$ ./script/build-sandbox` triggers the 1.x build?

mfikes:
Right. The 1.x train is still being maintained for a while. Ultimately we will sort out how to build 2.x via Homebrew (which currently uses make and clang, which may be fine), sort out what packages/libs it depends on (if any, or none if statically linked, etc).

ilovezfs:
Yup, simple make is fine, as is CMake, Autools, etc.

PR #2102, opened 2016-06-18
makepkg 5.0.1

ilovezfs:
Also,
- fix regression caused by calling `fstatat`, which is only available for >= Yosemite by limiting formula to only >= Yosemite
- fix the test by adding pkgver and pkgrel, which are now required
- match against output in the test

### DIFF
--- a/Formula/makepkg.rb
+++ b/Formula/makepkg.rb
@@ -2,10 +2,8 @@ class Makepkg < Formula
   desc "Compile and build packages suitable for installation with pacman"
   homepage "https://wiki.archlinux.org/index.php/makepkg"
   url "https://projects.archlinux.org/git/pacman.git",
-      :tag => "v4.2.1",
-      :revision => "068f8cec42057751f528b19cece37db13ae92541"
-  revision 1
-
+      :tag => "v5.0.1",
+      :revision => "f38de43eb68f1d9c577b4378310640c1eaa93338"
   head "https://projects.archlinux.org/git/pacman.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
A couple of alternatives to restricting to Yosemite:
1. gnulib offers an fstatat
2. https://github.com/openzfsonosx/zfs/blob/master/lib/libspl/getmntany.c#L137-L152
3. "beg" upstream to code around this

DomT4:
Steal [my PRs](https://github.com/Homebrew/homebrew-core/pull/1345) why don't you 😉. I was trying to think of a better alternative to just killing off <Yosemite users, but I haven't yet found the time to seriously explore that.

ilovezfs:
In this case, they only call the non-portable function one time (literally) so we might be able to talk them out of it.

DomT4:
Seems reasonable to ask at least.

ilovezfs:
@DomT4 Requested an upstream dispensation: https://bugs.archlinux.org/task/49771

DomT4:
Cheers ILZ. `gnulib` in the `buildpath` is a decent alternative otherwise.

ilovezfs:
Yeah, but ew

ilovezfs:
@DomT4 Upstream pointed out that fstatat is used in libalpm, which makepkg itself doesn't actually use. So I added a patch that comments out the section in libalpm that uses it, since it's irrelevant.

ilovezfs:
bump.

DomT4:
This smells a bit like something they won't or don't need to fix as far as they're concerned, and we'll keep this patch forever, which is a bit worrying.

ilovezfs:
@DomT4 Yeah, I'm honestly wondering if they're just enjoying being difficult.

ilovezfs:
@DomT4 Remember that the patch is only necessary for <= Mavericks. Mind checking the analytics on that?

MikeMcQuaid:
If it's a patch upstream won't take then let's just force this to require Mavericks.

DomT4:
Turns out almost nobody has been downloading this formula since we started tracking stats. 17 unique installations, all using 10.11.5. I'd go ahead and chop it on Mavericks & below, aye.

ilovezfs:
@MikeMcQuaid Yup, that makes sense.
@DomT4 Thank you for checking!

ilovezfs:
:ship:?

MikeMcQuaid:
:shipit:!

DomT4:
@ilovezfs Apologies, I'd had enough of my emails last night so closed them down & refused to reopen them till later 😄. This is fine, thanks for your efforts to resolve it with upstream here.

ilovezfs:
>Apologies, I'd had enough of my emails last night so closed them down & refused to reopen them till later :smile:

Wait, you know that's not allowed!

>This is fine, thanks for your efforts to resolve it with upstream here.

No problem.

PR #2109, opened 2016-06-18
serialosc 1.4

ilovezfs:
- needs to be built from the tag due to submodules
- add a test
- fix audit issue about avoiding spaces in `system`
- update homepage URL


DomT4:
Seems fine 👍.

PR #2114, opened 2016-06-18
gnu-apl 1.5

ilovezfs:
- upstream patch for clang build failures
- add a head spec

### DIFF
--- a/Formula/gnu-apl.rb
+++ b/Formula/gnu-apl.rb
@@ -1,9 +1,19 @@
 class GnuApl < Formula
   desc "GNU implementation of the programming language APL"
   homepage "https://www.gnu.org/software/apl/"
-  url "http://ftpmirror.gnu.org/apl/apl-1.4.tar.gz"
-  mirror "https://ftp.gnu.org/gnu/apl/apl-1.4.tar.gz"
-  sha256 "69da31db180e6ae17116122758c266e19f62e256f04e3c4959f6f2b224a1893a"
+
+  stable do
+    url "http://ftpmirror.gnu.org/apl/apl-1.5.tar.gz"
+    mirror "https://ftp.gnu.org/gnu/apl/apl-1.5.tar.gz"
+    sha256 "013addd15cab829d3212fb787023a723e01f2e58e4836acbedbc2dc7078e8d40"
+
+    # Upstream fixes for clang build failures, the first of which is
+    # "non-ASCII characters are not allowed outside of literals and identifiers"
+    patch do
+      url "https://raw.githubusercontent.com/Homebrew/formula-patches/69c8268d/gnu-apl/clang-fixes.patch"
+      sha256 "1d2c3e6d97f26ea25bc1e0126a4968537e6d54044cf0a991e0477d462c0158bf"
+    end
+  end
 
   bottle do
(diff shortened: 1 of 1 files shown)


DomT4:
The build is fairly impressively broken 😄

ilovezfs:
The pseudo good news is that it's fine with gcc.

ilovezfs:
OK, found upstream link about it: https://lists.gnu.org/archive/html/bug-apl/2015-08/msg00111.html and sure enough it's fixed in HEAD.
```
```

DomT4:
Nice. Can we grab that as a patch, if it applies cleanly?

ilovezfs:
Yeah, this fixes it:
```
    patch do
      url "https://github.com/ilovezfs/gnu-apl/commit/d3b5ce9f4a079af8f03c664d6f455cb8eb7f722f.patch"
    end

    patch do
      url "https://github.com/ilovezfs/gnu-apl/commit/23dd9084c6227a2f4c01857d5c03e5f6988b8645.patch"
    end

    patch do
      url "https://github.com/ilovezfs/gnu-apl/commit/d69d4ba3774ccc7e9f351a6cc5dd82762e5b9bef.patch"
    end

    patch do
      url "https://github.com/ilovezfs/gnu-apl/commit/7fafa990cf5bed607d3b0a2c820a21a02d964339.patch"
    end

    patch do
      url "https://github.com/ilovezfs/gnu-apl/commit/e82978d154f22ebe782bc034109d93c0df58cc55.patch"
    end

    patch do
      url "https://github.com/ilovezfs/gnu-apl/commit/ae9f34a216a046b29c4b69a87c315509688b6366.patch"
    end

    patch do
      url "https://github.com/ilovezfs/gnu-apl/commit/408da4d7ef6f066d10d9eea81cf0cda745f29d90.patch"
    end
```

ilovezfs:
They're all directly sequential, so could be put into a single patch using the revision range, or separately as above. Not sure which you prefer.

ilovezfs:
Another option (unorthodox) would be to fake it and just define 1.5.0 as the last of these since these commits all directly followed the tag.

ilovezfs:
@DomT4 In other words, I think we should consider doing this:
```
  url "http://svn.savannah.gnu.org/svn/apl/trunk", :revision => 586
  version "1.5"
```

ilovezfs:
PR refreshed with a combo patch in formula-patches.

DomT4:
Apologies, I wasn't around much last night & completely missed the initial emails here. This LGTM. Feel free to :shipit:.

ilovezfs:
Woo hoo 12,258 line patch for the win

DomT4:
😆 Sometimes I wish you wouldn't tell me these things. Should be alright given people can't screw up the whitespacing with external patches, unless we want to wait for the next stable release instead, don't mind either way.

ilovezfs:
>Sometimes I wish you wouldn't tell me these things.

Well that is indeed the point. I don't ever want to hear, "Well if you'd only told me *that*, then ..."

I actually think it's fine in this case. For instance, 8,607 of it is just aclocal.m4 being added.

DomT4:
👍

PR #2119, opened 2016-06-18
macvim: permit non-system ruby/perl

DomT4:

---

CC @MikeMcQuaid - This should ensure your changes the other day here are representative to what most users can actually obtain (i.e. prior to this PR there's no automatic way for any user running >10.6 to use a custom Ruby/Perl here).

### DIFF
--- a/Formula/macvim.rb
+++ b/Formula/macvim.rb
@@ -25,7 +25,7 @@ class Macvim < Formula
   # Help us! We'd like to use superenv in these environments, too
   env :std if MacOS.version <= :snow_leopard
 
-  def language_type language
+  def language_type(language)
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> (i.e. prior to this PR there's no automatic way for any user running >10.6 to use a custom Ruby/Perl here).

Actually, in that case I think we're fine anyway because I'm just wanting to query what we should do for bottles and, with these missing, we can ignore Ruby and Python for bottling anyway.

DomT4:
Alright. Just didn't know if you wanted to be sure the analytics were actually capturing custom/Homebrew Perl/Ruby.

MikeMcQuaid:
@DomT4 Thanks anyway pal 🙇

PR #2126, opened 2016-06-18
mpv: remove --enable-gpl3

ArPrt93:
The option, enabled by default, was removed in git master and it causes the HEAD build to fail.


kokowang:
How long does this usually take to merge? I kinda need MPV...

kokowang:
Umm. Build still fails after commit. Same error.

PR #2132, opened 2016-06-19
git: install diff-highlight

robsonpeixoto:


DomT4:
I'm not sure this merits the `revision` bump unless it's something the vast majority of users are going to want.

dunn:
Agreed; also unless we're open to installing more or less everything from `contrib` into `bin`, we already install the entire folder into `share`, which may be good enough: https://github.com/Homebrew/homebrew-core/pull/2132/files#diff-3e84bae646d908b93e043833873d316dL135

xu-cheng:
I'm not sure we should do this. There are 73 binaries under contrib. It's not consistency to install only some of them to `bin`.
```
$ gfind /usr/local/share/git-core/contrib -type f -executable |wc -l
      73
``` 

Also this creates a breakage for users who already use `/usr/local/share/git-core/contrib/diff-highlight/diff-highlight` to locate diff-highlight.

DomT4:
Heh, this also introduces a conflict with `diff-so-fancy`, as it turns out. Perhaps it's better remaining where it is, otherwise it seems plausible we end up linking them all into `bin` for consistency and creating potential conflicts/breaking existing workflows.

dunn:
Closing per the above; but thank you for the PR! ❤️

PR #2135, opened 2016-06-19
ibex 2.3.0

ilovezfs:
Needed to apply upstream fix for ibex-team/ibex-lib#204.

### DIFF
--- a/Formula/ibex.rb
+++ b/Formula/ibex.rb
@@ -1,8 +1,8 @@
 class Ibex < Formula
   desc "C++ library for constraint processing over real numbers."
   homepage "http://www.ibex-lib.org/"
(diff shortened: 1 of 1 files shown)


DomT4:
One of the tests is 💀.

ilovezfs:
Yeah, it's a regression. Not sure what the exact cause is yet, but seems to be some crazy exception that can be ignored but isn't.

ilovezfs:
Reported upstream: https://github.com/ibex-team/ibex-lib/issues/204

ilovezfs:
@DomT4 or would it be preferable to apply the patch itself? Either way is fine with me.

DomT4:
The `inreplace` is minor enough that I think that's fine, personally. Feel free to patch if you're happier doing so though.

PR #2144, opened 2016-06-19
headphones 0.5.16

ilovezfs:
Also, add a test.

### DIFF
--- a/Formula/headphones.rb
+++ b/Formula/headphones.rb
@@ -1,9 +1,8 @@
 class Headphones < Formula
   desc "Automatic music downloader for SABnzbd"
   homepage "https://github.com/rembo10/headphones"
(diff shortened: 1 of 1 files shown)


Formula/headphones.rb
```diff
@@ -72,4 +71,8 @@ def plist; <<-EOS.undent
     </plist>
     EOS
   end
+
+  test do
+    assert_match "Music add-on", shell_output("headphones -h")
```
DomT4: _cough_ Please use the full path to executable in testing _cough_ 😉

ilovezfs: Just seeing if you're paying attention.

DomT4: It's Sunday, I'd forgive you if you didn't keep me on my toes 😄

PR #2146, opened 2016-06-19
jsdoc3 3.4.0

ilovezfs:
- use language/node
- update description
- switch to registry.npmjs.org URL
- remove devel since it's older than stable
- remove bottle :unneeded since it has node deps
- add "node" dependency


DomT4:
Seems alright 👍

PR #2150, opened 2016-06-19
cheat 2.1.25

ilovezfs:
And style tweaks from poet


DomT4:
Ideally the `+` would be changed to `/` here as that's the style we tend to prefer these days, but not your fault that exists. Once you're happy with resources, 👍.

ilovezfs:
@DomT4 Refreshed with style tweaks from poet

DomT4:
LGTM, Thanks!

PR #2153, opened 2016-06-19
gupnp 0.20.17

ilovezfs:
~~remove the patch~~ patch still needed. Responded to upstream inquiry about that here: https://github.com/GNOME/gupnp/pull/1#issuecomment-183446124


ilovezfs:
Build failure is fixed with https://github.com/Homebrew/homebrew-core/pull/2185

DomT4:
👍

ilovezfs:
@nijikon @DomT4 actually looks like the patch (https://github.com/GNOME/gupnp/pull/1) isn't needed anymore.

ilovezfs:
Hmm looks like something was wrong with my test. Indeed still needed.

PR #2157, opened 2016-06-19
libproxy 0.4.13

ilovezfs:
- clang issues were fixed upstream, so remove the patch
- verify output is as expected in the test


DomT4:
👍

PR #2137, opened 2016-06-19
gnu-complexity 1.10

ilovezfs:
Created with `brew bump-formula-pr`.

PR #2160, opened 2016-06-19
mtools 4.0.18

ilovezfs:
punt with the UNUSED macro if ENV.cc is clang

### DIFF
--- a/Formula/mtools.rb
+++ b/Formula/mtools.rb
@@ -1,9 +1,9 @@
 class Mtools < Formula
   desc "Tools for manipulating MSDOS files"
   homepage "https://www.gnu.org/software/mtools/"
-  url "https://ftpmirror.gnu.org/mtools/mtools-4.0.17.tar.gz"
-  mirror "https://ftp.gnu.org/gnu/mtools/mtools-4.0.17.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Can we report that as being problematic for Clang upstream?

ilovezfs:
@DomT4 perhaps a better idea would be to do a check in superenv for whether the URL matches "gnu" and then slipstream in `fails_with :clang` if it does. :innocent:

DomT4:
😆 However [distasteful Stallman](https://gcc.gnu.org/ml/gcc/2014-01/msg00247.html) personally finds Clang, most of the GNU folk have been happy enough to resolve Clang issues before thankfully.

ilovezfs:
@DomT4 Yeah I just have [this dung pile](https://github.com/Homebrew/homebrew-core/pull/2137#issuecomment-227069300) in mind  and oh yeah [this one](https://github.com/Homebrew/formula-patches/pull/33), too.

DomT4:
Heh. It is slightly annoying that so many of these are creeping in recently.

ilovezfs:
It was reported: https://lists.gnu.org/archive/html/info-mtools/2014-02/msg00000.html

And ignored.

DomT4:
Hilarious. I think GNU have a more formal bug reporting system beside mailing lists, maybe worth filing there as well 😕

MikeMcQuaid:
:shipit:

PR #2167, opened 2016-06-19
watch 3.3.11

ilovezfs:
also needed to use upstream fix for the build failure
  "error: conflicting types for 'user_from_uid'"

### DIFF
--- a/Formula/watch.rb
+++ b/Formula/watch.rb
@@ -1,8 +1,22 @@
 class Watch < Formula
   desc "Executes a program periodically, showing output fullscreen"
   homepage "https://gitlab.com/procps-ng/procps"
-  url "https://download.sourceforge.net/project/procps-ng/Production/procps-ng-3.3.10.tar.xz"
-  sha256 "a02e6f98974dfceab79884df902ca3df30b0e9bad6d76aee0fb5dce17f267f04"
+  head "https://gitlab.com/procps-ng/procps.git"
+
+  stable do
+    url "https://gitlab.com/procps-ng/procps.git",
+      :tag => "v3.3.11",
+      :revision => "de985eced583f18df273146b110491b0f7404aab"
+
+    # Upstream commit, which (probably inadvertently) fixes the error
+    # "conflicting types for 'user_from_uid"
+    # Commit subject is "watch: Correctly process [m Remove lib dependency"
+    # Preexisting upstream report: https://gitlab.com/procps-ng/procps/issues/34
+    patch do
+      url "https://gitlab.com/procps-ng/procps/commit/99fa7f9f.diff"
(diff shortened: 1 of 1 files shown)


ilovezfs:
This is looking boneyard worthy in terms of how many porting issues there are. It doesn't appear upstream is paying any attention to macOS.

DomT4:
It's relatively well used. 3.5k+ downloads in the last 40 days or so. Unless there's new issues I'd probably :shipit: for now.

ilovezfs:
There's nothing to ship since it no longer builds without significant patching for macOS. :disappointed:

DomT4:
Oh, I thought the CI build was green for some reason, apologies, I'm obviously seeing things again. It's probably worth reporting this upstream, staying on the current version temporarily & if we don't get a _"We'll try to fix this"_ from upstream soon kill it then.

ilovezfs:
@DomT4 I determined it was fixed in HEAD, and bisected it to discover that the fix occurred in https://gitlab.com/procps-ng/procps/commit/99fa7f9f57f52afdf648584879f37980731215d5

Since the part of that commit that resolves the build failure was in Makefile.am, I generated the good and bad Makefile.in files, and have applied the difference between them using `inreplace`. Alternatively, we could apply the commit itself as a patch but that would require adding dependencies on Autotools. Which approach do you prefer?

ilovezfs:
Note that one benefit of applying the patch itself is that it would fix the unrelated bug also fixed in that commit. Of course, that could also be done in a patch :DATA or formula patch if desirable.
```
diff --git a/watch.c b/watch.c
index de46730..58b4dbc 100644
--- a/watch.c
+++ b/watch.c
@@ -228,10 +228,12 @@ static void process_ansi(FILE * fp)
 	 * attributes to apply, but typically there are between 1 and 3.
 	 */
 
-	if (*endptr == '\0') set_ansi_attribute(0); /* [m treated as [0m */
 
-	for (endptr = numstart = buf; *endptr != '\0'; numstart = endptr + 1)
+	for (endptr = numstart = buf; *endptr != '\0'; numstart = endptr + 1) {
 		set_ansi_attribute(strtol(numstart, &endptr, 10));
+        if (numstart == endptr)
+            set_ansi_attribute(0); /* [m treated as [0m */
+    }
 }
 
 static void __attribute__ ((__noreturn__)) do_exit(int status)
```

DomT4:
Using a formula patch if possible seems ideal, if you're happy with that.

ilovezfs:
@DomT4 yeah, that's fine with me. Also, it means we build from the gitlab tag, hence no SourceForge. I'd be tempted to leave the Autotools deps in place even after the patch is no longer needed.

ilovezfs:
@DomT4 :ship:?

DomT4:
Seems reasonable enough.

PR #2171, opened 2016-06-19
aptly 0.9.7

ilovezfs:
- let gom handle the vendoring (it uses revisions from the Gomfile)
- hg dep no longer needed
- use full path in the test


DomT4:
The tests need to be changed to use a full path to the executable, but otherwise 👍 when you're happy.

PR #2172, opened 2016-06-19
gollum 0.4.3

ilovezfs:
use upstream vendor support

### DIFF
--- a/Formula/gollum.rb
+++ b/Formula/gollum.rb
@@ -1,12 +1,8 @@
-require "language/go"
-
 class Gollum < Formula
   desc "n:m message multiplexer written in Go"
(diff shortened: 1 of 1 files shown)


DomT4:
👍 whenever you're happy with the resource situation.

DomT4:
It looks like upstream introduced vendor support with this release, so you might actually be able to kill most/all of the resources.

ilovezfs:
@DomT4 Refreshed PR to use upstream's vendor support.

DomT4:
Looks so much nicer this way ❤️. LGTM 👍.

ilovezfs:
@DomT4 I wish they would all do this!

DomT4:
In time, hopefully.

PR #2181, opened 2016-06-19
qt5:  Add missing install targets to Mac QtBase widget examples.  

### LINKED ISSUE #1948: Qt 5.6.1 fails to build with '--with-examples'

- [ ] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?  (N/A)  

My logs are [here](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/tree/master/Qt/First%20Attempt) and [here](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/tree/master/Qt/Second%20Attempt).  My [first attempt](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/tree/master/Qt/First%20Attempt)'s logs are just those created by Homebrew itself upon failure, but my [second attempt](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/tree/master/Qt/Second%20Attempt)'s logs also include the following files:  

* [`reinstall.log`](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/blob/master/Qt/Second%20Attempt/reinstall.log)
* [`doctor.log`](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/blob/master/Qt/Second%20Attempt/doctor.log)
* [`config.log`](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/blob/master/Qt/Second%20Attempt/config.log)

These files contain the results of redirecting standard output and standard error to them from the commands…:  

* `HOMEBREW_MAKE_JOBS=1 brew reinstall -vd --build-from-source qt5 --with-dbus --with-docs --with-examples --with-mysql`
* `brew doctor -vd`
* `brew config`

…, respectively.

RandomDSdevel:

---

Otherwise, building from source with the `--with-examples` option set fails.  

Closes #1948.  

<hr />

NOTE:  In regard to whether or not I've attempted to build Qt v5.x after making my changes to its formula, I actually _have_ done that.  I did _not_, however, check the relevant checkbox because the build in question _did not succeed_ (it failed while attempting to apply the patch that this pull request introduces.)  I did not collect any build logs the first time around, but I can produce some with another installation attempt if requested.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -39,6 +39,18 @@ class Qt5 < Formula
     sha256 "48ff18be2f4050de7288bddbae7f47e949512ac4bcd126c2f504be2ac701158b"
   end
 
+  # Fix for a build error involving missing Mac QtBase widget example targets
+  # detected by the logic introduced in <https://codereview.qt-project.org/#/c/156610/1>,
+  # possibly affecting Poppler, et. al..
+  #
+  # https://codereview.qt-project.org/#/c/161001/
+  #
+  # Should land in either v5.6.2 and/or v5.7.1.
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/6ffd0e250d374193613a51beda8830dda9b67e56/qt5/QTBUG-54110.patch"
+    sha256 "2cf77b820f46f0c404284882b4a4a97bf005b680062842cdc53e107a821deeda"
+  end
+
   keg_only "Qt 5 conflicts Qt 4 (which is currently much more widely used)."
 
   option "with-docs", "Build documentation"


UniqMartin:
Thank you for preparing this PR! Patching Qt is slightly more complicated than for most other formulae for the simple reason that Qt development is distributed across several Git repositories (one repository per module), but the source tarball then combines all of them. This means there's an additional level in the directory hierarchy that is not represented in the patch file. To make this work, you'd need to:

1. Download the patch to some local path on your machine. Use this path in the `patch do` block to test locally that the patch applies, i.e. put `file://<absolute-path-to-patch>` in the `url` field.
2. Modify the patch in an editor and replace all instances of `a/examples/` and `b/examples/`  where files are referenced with either `a/qtbase/examples/` or `b/qtbase/examples/`.
3. Try if the patch applies cleanly by running `brew unpack --patch qt5`. (This will unpack the source tarball in the current working directory and apply the patches.)
4. Submit this patch as a PR to the [formula-patches](https://github.com/Homebrew/formula-patches) repository. (Place the patch in the `qt5/` directory and preferably name it after the corresponding QTBUG.)
5. Once the patch PR has been merged, use the URL to that patch in the `patch do` block in this PR.

I hope all of this made some sense. If not, feel free to ask for clarification. Maybe some of the stuff is also pretty obvious, but I decided to be as explicit as possible just in case.

Formula/qt5.rb
```diff
@@ -39,6 +39,19 @@ class Qt5 < Formula
...
   end
 
+  # Add missing install targets to Mac QtBase widget examples.  This fixes an
+  # issue with building from source with the `--with-examples` option set which
+  # was introduced between the releases of Qt versions 5.6.0 and 5.6.1 by
+  # <http://code.qt.io/cgit/qt/qtbase.git/commit/?id=5b5b7f97b833e9f91c41285718dc7b566cc37247>.
+  # (Remove this as soon as it is feasible and possible to do so!)
+  #
+  # Core formulae known to fail without this patch (as of 2016-06-19):
+  #   (None known.)
```
UniqMartin: You can simplify this comment a bit. The other currently used patch is a bit of a special case and thus not the best template. Try to model your patch comment after [this one](https://github.com/Homebrew/homebrew-core/blob/25154bf004f0e2dbc4b9d38dd594605398eacb78/Formula/qt5.rb#L32-L34) that was used in an older version of this formula. (As for when this patch can be expected to be removed, both 5.6.2 and 5.7.1 will most likely contain it.)

RandomDSdevel:
@UniqMartin on your earlier reply (not the diff comment:)  
> Thank you for preparing this PR!  …

Again, you're very welcome!  

> …Patching Qt is slightly more complicated than for most other formulae for the simple reason that Qt development is distributed across several Git repositories (one repository per module), but the source tarball then combines all of them.  This means there's an additional level in the directory hierarchy that is not represented in the patch file.  

Ah, of course!  That _would_ cause a bit of a problem, wouldn't it?  

> To make this work, you'd need to:  
> 
> 1. Download the patch to some local path on your machine.  Use this path in the patch do block to test locally that the patch applies, i.e. put file://<absolute-path-to-patch> in the url field.  
> 2. Modify the patch in an editor and replace all instances of a/examples/ and b/examples/ where files are referenced with either a/qtbase/examples/ or b/qtbase/examples/.  
> 3. Try if the patch applies cleanly by running brew unpack --patch qt5.  (This will unpack the source tarball in the current working directory and apply the patches.)  
> 4. Submit this patch as a PR to the formula-patches repository.  (Place the patch in the qt5/ directory and preferably name it after the corresponding QTBUG.)  
> 5. Once the patch PR has been merged, use the URL to that patch in the patch do block in this PR.  
> 
> I hope all of this made some sense.  If not, feel free to ask for clarification.  Maybe some of the stuff is also pretty obvious, but I decided to be as explicit as possible just in case.  

Thank you for these very helpful instructions, @UniqMartin!  I'll follow them as soon as I can, which _should_ be when I next get on this device&thinsp;—&thinsp;I'd do it now, but I have to get off for the evening so that other people can use this computer since my family shares it and I only boot to an external drive; I apologize both retroactively for any long delays in response time that this may have caused in the past and in advance for any similar lag that may occur because of both these circumstances and the fact that Qt takes _a really long time_ to build on my machine (I'm on a 24" mid-2007 iMac, and it _definitely_ doesn't help me any that my external drive has to stay connected via USB 2.1 since, for _some_ reason, it won't boot said iMac via FireWire 800.)

Formula/qt5.rb
RandomDSdevel: OK, I'll do that, then!  Also, your assumption on when we'll find ourselves able to remove this patch from the Qt formula sounds correct given what little I know about the software's development process.

UniqMartin:
Don't worry about past and future response times! My ability to respond quickly also fluctuates due to external factors and other issues/PRs requiring my attention. We all have our responsibilities and constraints that we operate under. 😸

As for the build times: If you haven't discovered `--without-webengine` yet, I can highly recommend using this option for from-source builds. The patch only affects the `qtbase` module, thus skipping `qtwebengine` shouldn't make any difference in testing the fix, but it will reduce your build time massively.

Formula/qt5.rb
```diff
@@ -39,6 +39,19 @@ class Qt5 < Formula
...
+  patch do
+    url "http://code.qt.io/cgit/qt/qtbase.git/patch/?id=58408ffa1b9c0b42a1719d3c8a4d4c62dec4fce6"
```
DomT4: Should probably be `https://code.qt.io`...

MikeMcQuaid:
A suggestion either for this or a future PR: let's just enable `"with-docs"` and `"with-examples"` by default as they don't add extra dependencies or much build time and mean people can get these in our binary packages.

UniqMartin:
> A suggestion either for this or a future PR: let's just enable `"with-docs"` and `"with-examples"` by default as they don't add extra dependencies or much build time […]

I don't remember what the build time impact was when I last tried this a few releases ago, but `with-docs` has a significant impact on disk usage (and thus also bottle size). I believe the installation was at least twice as big as without the documentation. All of it is also accessible via http://doc.qt.io/ and given that most of our users that install `qt5` are probably not developers who want to develop with Qt 5, this sounds like a lot of wasted traffic and disk space.

I'd have to check the impact of `with-examples`, but this sounds more like something we might want to enable by default (though it again is only interesting to Qt 5 developers).

Formula/qt5.rb
UniqMartin: Agreed, if only we could use those patches directly. Sadly, we cannot.

RandomDSdevel:
@UniqMartin:  
> Don't worry about past and future response times! My ability to respond quickly also fluctuates due to external factors and other issues/PRs requiring my attention. We all have our responsibilities and constraints that we operate under. 😸
> 
> ⋮

Thanks for being so understanding!  

> As for the build times: If you haven't discovered --without-webengine yet, I can highly recommend using this option for from-source builds. The patch only affects the qtbase module, thus skipping qtwebengine shouldn't make any difference in testing the fix, but it will reduce your build time massively.

That's a neat tip, but I've also noticed that Homebrew seems to try to detect that some formula's associated package has already been updated by an in-flight pull request from a contributor's end and skips installing said package if it has, so I'd rather just use the same options as the ones I'd set if it weren't me putting this contribution together now to save me that time later instead.

Formula/qt5.rb
RandomDSdevel: My apologies:  I didn't know Qt's code repositories were available to browse via HTTPS.  I'll go ahead in fix that in my next commit, then.

RandomDSdevel:
@UniqMartin:  I'm having quite a bit of trouble running `brew unpack --patch qt5`.  It seems that Homebrew caches the patches it loads in for application somehow, so the changes I made to my patch incrementally (I kept accidentally overlooking some file path descriptors involved in patch application) fail to apply.  How can I clear this patch cache?

UniqMartin:
You can remove the offending files in the directory printed by `brew --cache`. The file name should start with `qt5--patch-`. Though that shouldn't be happening, because every change to the file also means that its SHA-256 checksum will change (but maybe you forgot to update in the `patch do` block), meaning it will be “downloaded” anew, because the file name in the cache includes the checksum.

RandomDSdevel:
@UniqMartin:  Ah, doh!  You're right:  I forgot to update the checksum!

RandomDSdevel:
@UniqMartin:  [Pull request](https://github.com/Homebrew/formula-patches/pull/35) submitted!

UniqMartin:
Thanks, merged! You can now reference it in the formula as:

```ruby
url "https://raw.githubusercontent.com/Homebrew/formula-patches/6ffd0e250d374193613a51beda8830dda9b67e56/qt5/QTBUG-54110.patch"
```

RandomDSdevel:
@UniqMartin:  Thanks!  I'll make that change right now.

RandomDSdevel:
Alright, I should be good to try a build now.  Wish me luck!

RandomDSdevel:
@UniqMartin:  I apologize for the sudden appearance of that giant glob of commits that I didn't make; my local copy of the Homebrew Core got behind this one here on GitHub, so I thought I'd rebase mine, and I had to re-synchronize my local and GitHub-based pull request branches as well.  I hope we can clean that up…

UniqMartin:
> P. S.: Looks like we'll have to since this PR's CI build failed to brew-test-bot.*.pull --clean … it. How would I go about doing that?

Yes, please always rebase your branch on top of master if you need to synchronize your changed with what has happened in Homebrew instead of merging stuff. (But it wouldn't have been necessary here as there were no conflicting changes, in fact no upstream changes to the `qt5` formula at all.) Here's a short recipe for the future (not a solution for the current “mess”):

```sh
git checkout YOUR_BRANCH
# Do some work.
git checkout master
brew update # To update the master branch and benefit from the reporting of `brew update`.
git checkout YOUR_BRANCH
git rebase master # To rebase your branch on top of master.
# Make adjustments, add commits, test changes.
git push --force-with-lease # To update an already published branch and the associated PR.
```

To fix the current branch (assuming your local branch is called `Qt__v5_6_0-brew_formula_revision_1`):

```sh
git checkout Qt__v5_6_0-brew_formula_revision_1
git rebase -s recursive -X theirs master # Options help auto-resolve some conflicts.
# Remove the `revision 1` line. It's not needed for this change.
git add -p
git commit
git rebase -i master # To squash all commits to a single commit (assuming you are familiar with interactive rebase).
git push --force-with-lease # To update an already published branch and the associated PR.
```

I hope this helps.

RandomDSdevel:
@UniqMartin:  Thanks for that.  I'll have to wait to follow your advice, though, since I'm already doing a test build of Qt down here on my end and don't want to change things up and out from under Homebrew and create yet _another_ mess.

RandomDSdevel:
@UniqMartin:  YES!!!!  My build succeeded here on my end, taking 6 hours, 18 minutes, and 46 seconds with the options I mentioned above enabled.  It probably won't take _nearly_ as long for the continuous integration servers, though, since they're newer machines and won't be building bottles with all of the options I had enabled on.  Now I just have to properly rebase my pull request and we'll be all set!

RandomDSdevel:
@UniqMartin:  OK, I've followed your instructions and you should be good to merge my changes in once they pass continuous integration.

UniqMartin:
Merged in 5df5ef7ba9fcf7f6b2cf2824a7533b26c6093689. Thank you for this contribution to Homebrew, @RandomDSdevel! :tada:

I allowed myself to trim both your in-code comment and your commit message quite a bit. I hope that's fine with you. (Also feel free not to mention me in future commit messages. This only means I get even more notification mails than I already do. 😉)

RandomDSdevel:
@UniqMartin:  Yeah, that's perfectly fine by me.  Also, sorry for not realizing what the consequences of my mentioning you in my commit messages!  I'll try not to do that again, thereby leaving you in relative peace.

RandomDSdevel:
@UniqMartin:  Hey, remember how I said that I usually don't have to reinstall formulas' related software when I contribute the updates?  Well, it turns out that this only works when revision numbers match up, and I did my local build before I undid the revision bump that I originally intended to commit as part of my contribution because I thought it might be necessary but which you assured me was not.  In any case, Homebrew wanted to reinstall Qt 5 when I went to upgrade Poppler, and I definitely do _not_ want to have to go through _that_ insanely long build again!  Will Homebrew recognize Qt 5 as having been installed if I just rename `/usr/local/Cellar/qt5/5.6.1_1` back to `/usr/local/Cellar/qt5/5.6.1`, or would fixing this be more involved than that?

UniqMartin:
Homebrew will recognize that, if you rename the directory, but unfortunately Qt will have problems because the full path (including the revision number) is baked into most of the Qt binaries and frameworks. I'm sorry about that.

The reason for omitting the revision bump is because the fix only affects people building a non-standard configuration of the `qt5` formula and that were unable to do so prior to the fix, so in theory couldn't already have Qt 5.6.1 installed. If we had bumped the revision, *everyone* (even people not using the `--with-examples` option) would have had to either download a new bottle or rebuild all of `qt5`.

RandomDSdevel:
OK, that makes sense.  I guess I'll just have to rebuild, then (sigh…)

PR #2182, opened 2016-06-20
urweb 20160621 (migrate from boneyard)

eranmeir:
urweb.rb was moved to the boneyard because it used old binary-only mlton
formula (see Homebrew/legacy-homebrew#21780).

A new mlton formula was already merged (Homebrew/legacy-homebrew#48694)
and the urweb formula is now updated as well.

### DIFF
--- a/Formula/urweb.rb
+++ b/Formula/urweb.rb
@@ -1,0 +1,28 @@
+class Urweb < Formula
+  desc "Ur/Web programming language"
+  homepage "http://www.impredicative.com/ur/"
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Looks good. Could you make a PR to the boneyard removing the formula there too?

UniqMartin:
Merged in c1f3fe480a5327e2bb795484203c6eae3893d9bf. Thank you for this contribution to Homebrew, @eranmeir! :tada:

PR #2189, opened 2016-06-20
Build bottles for vim/macvim.

MikeMcQuaid:

---

Based on the analytics results it looks like the most popular option is to just use a Homebrew python for these bottles so: let's do it.

### DIFF
--- a/Formula/vim.rb
+++ b/Formula/vim.rb
@@ -7,8 +7,6 @@ class Vim < Formula
 
   head "https://github.com/vim/vim.git"
 
-  bottle :disable, "To use the user's Python."
-
   deprecated_option "disable-nls" => "without-nls"
   deprecated_option "override-system-vi" => "with-override-system-vi"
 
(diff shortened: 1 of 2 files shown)


Formula/vim.rb
```diff
@@ -19,6 +17,12 @@ class Vim < Formula
   LANGUAGES_OPTIONAL = %w[lua mzscheme python3 tcl]
   LANGUAGES_DEFAULT  = %w[perl python ruby]
 
+  if MacOS.version >= :mavericks
```
DomT4: The `ruby` and `perl` ones won't actually use custom versions unless we add the `:perl` and `:ruby` dependencies. Superenv prevents finding them, even from source.

MikeMcQuaid: This is `vim` rather than `macvim` so it does have these already.

DomT4: Heh, right you are. I may just give up reviewing PRs, it's not going well lately 🙈.

PR #2191, opened 2016-06-20
mimetic 0.9.8 (new formula)

croessner:

---

New formula mimetic.rb

### DIFF
--- a/Formula/mimetic.rb
+++ b/Formula/mimetic.rb
@@ -1,0 +1,41 @@
+class Mimetic < Formula
+  desc "C++ MIME library"
+  homepage "http://www.codesink.org/mimetic_mime_library.html"
+  url "http://www.codesink.org/download/mimetic-0.9.8.tar.gz"
+  sha256 "3a07d68d125f5e132949b078c7275d5eb0078dd649079bd510dd12b969096700"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:

Formula/mimetic.rb
```diff
@@ -0,0 +1,41 @@
...
+            MimeEntity me;
+            me.header().from("me <me@domain.com>");
+            me.header().to("you <you@domain.com>");
+            me.header().subject("my first mimetic msg");
+            me.body().assign("hello there!");
+            cout << me << endl;
+            return 0;
+      }
+    EOS
+    system ENV.cxx, "test.cpp", "-L/usr/local/lib", "-lmimetic", "-o", "test"
```
DomT4: Why was this hardcoded as `/usr/local/lib`?

croessner: On my Mac, I had to add it, or the linker would tell me about not finding all symbols. Maybe something is wrong with my environment, because other libs in /usr/local/{include,lib} also need -I and -L. If you can check the test on your system and it works without the hardcoded part, feel free to remove it. Maybe you can drop me a hint, what could be wrong on my side. Thanks

UniqMartin: Homebrew can be installed in locations other than `/usr/local`, thus using `"-L#{lib}"` (that's referencing this formula's `lib` directory) would be preferable over hard-coding `/usr/local/lib`.

PR #1883, opened 2016-06-11
Flow v0.27.0

jeffmo:
- [ ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [ ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

PR #2196, opened 2016-06-20
flow 0.27.0

turadg:
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change? (yes, #1883 which doesn't work)

---
#1883 downloading the zip doesn't build (`make` hangs). As @ilovezfs points out the git checkout works.

### DIFF
--- a/Formula/flow.rb
+++ b/Formula/flow.rb
@@ -1,8 +1,10 @@
 class Flow < Formula
   desc "Static type checker for JavaScript"
   homepage "https://flowtype.org/"
-  url "https://github.com/facebook/flow/archive/v0.26.0.tar.gz"
-  sha256 "7813c98f7509d89e5a187df4252dbf8e6c429b1d711e10156eccb1ac793fb571"
+  url "https://github.com/facebook/flow.git",
+    :tag => "v0.27.0",
+    :revision => "ceea4c3616824b9ad7cd7028dece2d55a7f3b539"
+  sha256 "c2e2b9d398815825a7bf06c019652a3e3daa8d859e939dbf1d8ed6cb03635462"
   head "https://github.com/facebook/flow.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@turadg could I convince you to report that upstream? :angel:

turadg:
Convinced, https://github.com/facebook/flow/issues/1981

Of course a fix will require another release. Can we merge this for 0.27.0?

Formula/flow.rb
```diff
@@ -1,8 +1,10 @@
 class Flow < Formula
   desc "Static type checker for JavaScript"
   homepage "https://flowtype.org/"
-  url "https://github.com/facebook/flow/archive/v0.26.0.tar.gz"
-  sha256 "7813c98f7509d89e5a187df4252dbf8e6c429b1d711e10156eccb1ac793fb571"
+  url "https://github.com/facebook/flow.git",
+    :tag => "v0.27.0",
+    :revision => "ceea4c3616824b9ad7cd7028dece2d55a7f3b539"
+  sha256 "c2e2b9d398815825a7bf06c019652a3e3daa8d859e939dbf1d8ed6cb03635462"
```
DomT4: You can scrap the `sha256` if we're using a git checkout. The `:revision` serves as the checksum in this case. Possibly worth adding a short note like:

```ruby
# Switch back to archive tarball when possible.
# link/to/upstream/bug/report
```

mroch:
fix is coming upstream. the commit should apply to 0.27, if you'd rather use `patch` to backport it.

ilovezfs:
No need to back port it since the preference for tarball and sha256 over tag and revision is not very great. Only reason would be if upstream would like us to participate in testing/verifying the fix.

MikeMcQuaid:
Thanks again @turadg (or should I say: Robert Downey Junior [I'm sure you've never heard that before 😉])!

turadg:
😆  @MikeMcQuaid , yes but not on Github!

PR #2203, opened 2016-06-20
signing-party: boneyard

DomT4:

---

I'm only intending for this to be temporary, but we're stuck between a rock and a hard place on this formula & it's holding up work that's really more significant. Debian purged the old version from their mirrors, so we can't simply bump the revision & ignore the new release for a time.

Essentially, for the new release Debian moved to requiring `libmd`, which is a cryptographic library that is portable on BSD and Linux, and was included in OS X until 10.8 when it was removed.

Various implementations are available to us:
- We could use the "original" `libmd` that today needs heavy patching to build on OS X.
- We could try to revive the old Apple `libmd` implementation, which doesn't build as-is.
- We could attempt to use the one Debian uses, which although despite being slightly modernised by Debian still doesn't build on OS X.
- We could also ask Debian to go back to using a completely portable mechanism and drop the `libmd` dependency.

Given none of those options are simple, and I've already invested a fair bit of time trying to make them play nicely in sane ways, let's remove this, hopefully temporarily.


UniqMartin:
:+1:

MikeMcQuaid:
👍

DomT4:
Thanks both. https://github.com/Homebrew/homebrew-boneyard/commit/dec7cf46e5a6ed1c3a20922adf82f3e78ef9f246 is the boneyard commit.

PR #2208, opened 2016-06-20
b2sum 20160619 (new formula)

grempe:

---

Installs the reference implementation of the b2sum command
line tool from the official source repository.

b2sum is a command line tool to hash files using the
BLAKE2 hash function.

https://github.com/BLAKE2/BLAKE2

### DIFF
--- a/Formula/b2sum.rb
+++ b/Formula/b2sum.rb
@@ -1,0 +1,18 @@
+class B2sum < Formula
+  desc "BLAKE2 b2sum reference binary"
+  homepage "https://github.com/BLAKE2/BLAKE2"
+  url "https://github.com/BLAKE2/BLAKE2/archive/20160619.tar.gz"
+  sha256 "cbac833ccae56b5c6173dbeaf871aa99b32745cf7a994c7451d4533ecda55633"
+
+  def install
+    cd "b2sum" do
+      system "make", "NO_OPENMP=1"
+      system "make", "install", "PREFIX=#{prefix}", "MANDIR=#{man}"
+    end
+  end
+
+  test do
+    assert_equal "ba80a53f981c4d0d6a2797b69f12f6e94c212f14685ac4b74b12bb6fdbffa2d17d87c5392aab792dc252d5de4533cc9518d38aa8dbf1925ab92386edd4009923  -",
+             pipe_output("#{bin}/b2sum -", "abc").chomp
+  end
+end


Formula/b2sum.rb
```diff
@@ -0,0 +1,20 @@
...
+  homepage "https://github.com/BLAKE2/BLAKE2"
+  url "https://github.com/BLAKE2/BLAKE2/archive/20160619.tar.gz"
+  sha256 "cbac833ccae56b5c6173dbeaf871aa99b32745cf7a994c7451d4533ecda55633"
+
+  def install
+    cd "b2sum" do
+      # https://github.com/BLAKE2/BLAKE2/issues/25
+      system "make", "NO_OPENMP=1"
+      bin.install "b2sum"
+      man1.install "b2sum.1"
```
UniqMartin: Is there no `install` target in the `Makefile`? If not, it might be worthwhile to ask upstream for one or even submit a pull request to add it.

Formula/b2sum.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+  end
+
+  # brew test b2sum
```
UniqMartin: You can drop this comment.

Formula/b2sum.rb
```diff
@@ -0,0 +1,20 @@
...
+  test do
+    assert_equal(shell_output("echo 'abc' | b2sum -").strip, "426526ad3aa99ac6bbb98555c6c30b5a55189d0d7537d9214053b886f365a84bd7e68c91b544a8f485fa842c0ea9fbc4fb92277290264e685eed9e67b7621fc9  -")
```
UniqMartin: The expected value should be the first argument to `assert_equal`, you should use `pipe_output` instead of `shell_output`, and you need to reference the `b2sum` binary with its full path. In summary, something like this will probably work:

```ruby
assert_equal "426526ad3aa99ac6bbb98555c6c30b5a55189d0d7537d9214053b886f365a84bd7e68c91b544a8f485fa842c0ea9fbc4fb92277290264e685eed9e67b7621fc9  -",
             pipe_output("#{bin}/b2sum -", "abc").chomp
```

Formula/b2sum.rb
grempe: There is a makefile install target, but it installs to /usr/local/bin.

Not sure why using the makefile install target is important since there is only a single compilation output (b2sum) and a single pre-generated man.1 file.

See : https://github.com/BLAKE2/BLAKE2/blob/master/b2sum/makefile

Do you have any example of a formula making use of a similar makefile install target?

Formula/b2sum.rb
grempe: Done. I had it there because that is the only place it is documented that you need to pass `"NO_OPENMP=1"` to compile on OS X.

grempe:
I've made the changes requested by @UniqMartin, with the exception of using the makefile install target.

I have force pushed the amended commit onto my b2sum branch for this pull request.

Formula/b2sum.rb
grempe: Changed.

Formula/b2sum.rb
UniqMartin: I only explicitly referenced the `# brew test b2sum` comment. That's why I wrote “You can drop **this** comment.” The other comment referencing the `NO_OPENMP=1` issue was completely fine (feel free to add it back).

Formula/b2sum.rb
UniqMartin: Using the `install` target helps us stay in sync with upstream (e.g. if a new file is added), though I admit that this is unlikely to happen in this case.

But no need to further discuss this, as there is a somewhat usable `install` target. Sadly, the dependencies between targets are not properly declared, so we can't just run `make install` to get both compilation and installation in one step. Here's what should work (not tested myself):

```ruby
system "make", "NO_OPENMP=1"
system "make", "install", "PREFIX=#{prefix}", "MANDIR=#{man}"
```

UniqMartin:
Added a few follow-up comments in the now already by default hidden code comments, but other than that we're almost good to go. Thank you!

grempe:
Force pushed again to include changes to make install setup as requested.

UniqMartin:
Merged in e016cda47dfa298c6628de3e9d0fd976eecd91be. Thank you for this contribution to Homebrew, @grempe! :tada:

grempe:
Awesome.  Thanks for the help. Always impressed with the attention to detail and responsiveness of homebrew maintainers.  Its always a pain, which means you are doing a good job.  ;-)

UniqMartin:
Thank you for the kind words! It's great to hear such feedback every once in a while.

PR #2219, opened 2016-06-21
ruby 2.4.0-preview1 (devel)

ilovezfs:


ilovezfs:
@DomT4 :ship:?

DomT4:
Seems reasonable.

UniqMartin:
The only potential future problem with this I can think of:

```
$ brew irb
irb(main):001:0> Version.new("2.4.0-preview1") > Version.new("2.4.0")
=> true
```

ilovezfs:
Sounds like `-preview*` needs to be added to the list of magic words.

UniqMartin:
> Sounds like `-preview*` needs to be added to the list of magic words.

So where exactly does it belong in the spectrum of `alpha`, `beta`, `rc`, and final releases?

DomT4:
Sigh, of course. We could hack it with `2.4.0-a_prelease1` 😄.

ilovezfs:
@DomT4 Forget that.

DomT4:
More realistically though, `preview` is probably a realistic term to expect/handle. I'd say usually `preview` pre-dates `alpha` but I guess it depends on how upstreams are using each term.

DomT4:
@ilovezfs It was tongue-in-cheek 😉.

ilovezfs:
@DomT4 phew. I was wondering if you were feeling OK.

DomT4:
😄 Was definitely intended to make you laugh more than curl up in a corner.

If we follow Gem's `Version` conventions `preview` would be regarded as "newer" than `alpha` or `beta`. I'm not sure that's an entirely sane route either, so we probably do want to sculpt out something more Homebrew-specific.

ilovezfs:
That sounds reasonable. I think when "preview" is used in a less than beta context, it's usually "technology preview" or "tech preview" as opposed to just an unqualified "preview."

UniqMartin:
Long term we might want to be a bit more strict about what we consider a valid version number, and force everything that doesn't comply into a schema we can work with. Otherwise we'll run into these issues every once in a while and nobody (not even maintainers) will have a good understanding of how versions compare to each other. Here's another potentially surprising comparison:

```
irb(main):002:0> Version.new("1.0.2a") < Version.new("1.0.2")
=> true
```

MikeMcQuaid:
> Long term we might want to be a bit more strict about what we consider a valid version number, and force everything that doesn't comply into a schema we can work with.

I agree; I'd say the semver perhaps plus alpha/beta/rc is sufficient.

ilovezfs:
I think I would treat preview as a synonym for rc.

ilovezfs:
Shall I change the version on this to
```
version "2.4.0-rc1"
```
to make this moot for the time being?

MikeMcQuaid:
@ilovezfs Aye.

PR #2222, opened 2016-06-21
audit fixes for formulae with available upgrades

ilovezfs:
Note that the libmicrohttpd commit was moved to https://github.com/Homebrew/homebrew-core/pull/2230.

### DIFF
--- a/Formula/lcdproc.rb
+++ b/Formula/lcdproc.rb
@@ -24,4 +24,8 @@ class Lcdproc < Formula
     system "make"
     system "make", "install"
   end
+
+  test do
+    assert_match version.to_s, shell_output("#{bin}/lcdproc -v 2>&1")
+  end
 end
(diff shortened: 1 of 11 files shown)


ilovezfs:
libmodbus download is failing, yet I have no problems downloading it locally.

ilovezfs:
Also, not sure why this happened, but we're distributing 3.1.1 as current stable, yet upstream designates 3.0.6 as current stable.

ilovezfs:
Moved libmodbus to a separate PR: https://github.com/Homebrew/homebrew-core/pull/2226

MikeMcQuaid:
👍

ilovezfs:
@MikeMcQuaid would you pull this with or without the bottles?

MikeMcQuaid:
Without as it doesn't seem like it'll affect any bottle contents.

ilovezfs:
@MikeMcQuaid it will for libmicrohttpd

MikeMcQuaid:
@ilovezfs I see. Probably need to split that into a separate PR, then.

ilovezfs:
OK

PR #2225, opened 2016-06-21
ccextractor 0.81

ilovezfs:
- use upstream mac build script
- remove libpng dependency because it's vendored by upstream

### DIFF
--- a/Formula/ccextractor.rb
+++ b/Formula/ccextractor.rb
@@ -1,8 +1,8 @@
 class Ccextractor < Formula
   desc "Free, GPL licensed closed caption tool"
   homepage "http://ccextractor.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


DomT4:
I'm not completely sure `share` shouldn't be `pkgshare` here, but 👍 if you're happy.

ilovezfs:
@DomT4 Oh I'm not happy at all, actually, because upstream should make this work with CMake.

DomT4:
Yeah, that build script is a bit of a hack to say the least. From the README this is barely supported upstream, so I guess that it works at all is something.

ilovezfs:
@DomT4 The CMake build fails because the sha2 lib isn't built. Compile that and voilà it works. They're not even trying :)

PR #2226, opened 2016-06-21
libmodbus: add test

ilovezfs:
also update URL

### DIFF
--- a/Formula/libmodbus.rb
+++ b/Formula/libmodbus.rb
@@ -1,7 +1,8 @@
 class Libmodbus < Formula
   desc "Portable modbus library"
   homepage "http://libmodbus.org"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@ilovezfs I think you know this already but: feel free to self-merge without needing a 👍 any PRs that change only the URL/mirror/checksum/revision and are green.

ilovezfs:
@MikeMcQuaid Hmm... so I split libmodbus out from the others because the download was failing. The URL tweak seems not to have helped, and it's exhibiting the same problem again. Any idea what's going on here? I have no trouble downloading the URL here locally including via `brew install`.

ilovezfs:
Seems server can't access that website for some reason so I added a mirror.

PR #2232, opened 2016-06-21
lcdproc 0.5.7

ilovezfs:
also had to map uint -> u_int to fix the build

### DIFF
--- a/Formula/lcdproc.rb
+++ b/Formula/lcdproc.rb
@@ -1,8 +1,8 @@
 class Lcdproc < Formula
   desc "Display real-time system information on a LCD"
   homepage "http://www.lcdproc.org/"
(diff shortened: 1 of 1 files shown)


ilovezfs:
```
diff --git a/server/drivers/rawserial.c b/server/drivers/rawserial.c
index 433273a..3fcb5aa 100644
--- a/server/drivers/rawserial.c
+++ b/server/drivers/rawserial.c
@@ -45,6 +45,11 @@
 # include <sys/time.h>
 #endif
 
+#ifdef DARWIN
+#include <sys/types.h>
+#define uint u_int
+#endif
+
 #include "lcd.h"
 #include "rawserial.h"
 #include "report.h"

```

DomT4:
You submitted it upstream so 👍.

PR #2243, opened 2016-06-21
diff-so-fancy 0.10.0

stevemao:


DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/2b0b8d3fe9d01aeedad31d658880311e84495aa0. Thanks again @stevemao! 😺

PR #2250, opened 2016-06-21
various: transition formulae over to `gpg2` default.

DomT4:

---

Formula sibling of https://github.com/Homebrew/brew/pull/386. This will follow the other in ASAP post-merge.

Eventually the batch generation of a `gpg` key for proper functionality testing will be split into a `Utils` method to call, rather than having to write the whole blob into every formula each time, so that is a temporary measure rather than a permanent thing.

Where the `gpg` requirement is only `:run`, which is the majority of cases, the `revision` bump is designed to get people moved across harmoniously rather than in dribs & drabs

Beside needing to install `gnupg2` instead of `gnupg` _(which will happen automatically for from-bottle installations)_ there should be minimal change noticed by end users given that `gpg` 1.x & `gpg` 2.0 _(which the new requirement is bound to)_ share a mutual keychain and can be swapped in & out without issue.

This will wildly fail until the `brew` PR is merged, for obvious reasons.

### DIFF
--- a/Formula/blackbox.rb
+++ b/Formula/blackbox.rb
@@ -3,18 +3,22 @@ class Blackbox < Formula
   homepage "https://github.com/StackExchange/blackbox"
   url "https://github.com/StackExchange/blackbox/archive/v1.20160122.tar.gz"
   sha256 "ac5de1d74fdbe88604b34949f3949e53cb72e55e148e46b8c2be98806c888a10"
+  revision 1
 
   bottle :unneeded
 
-  depends_on :gpg
+  depends_on :gpg => :run
 
   def install
+    # Prefer GPG2 by default.
+    inreplace "bin/_blackbox_common.sh", ":=gpg", ":=gpg2"
+
     libexec.install Dir["bin/*"]
     bin.write_exec_script Dir[libexec/"*"].select { |f| File.executable? f }
   end
 
   test do
     system "git", "init"
-    system "#{bin}/blackbox_initialize", "yes"
+    system bin/"blackbox_initialize", "yes"
   end
 end
--- a/Formula/blackbox.rb
+++ b/Formula/blackbox.rb
@@ -18,7 +18,24 @@ class Blackbox < Formula
   end
 
   test do
+    # Note that test do is always in temporary path, with HOME captured,
+    # so this doesn't interfere with a user's pre-existing keychain.
+    (testpath/"batchgpg").write <<-EOS.undent
+    Key-Type: RSA
+    Key-Length: 2048
+    Subkey-Type: RSA
+    Subkey-Length: 2048
+    Name-Real: Testing
+    Name-Email: testing@foo.bar
+    Expire-Date: 1d
+    Passphrase: brew
+    %commit
+    EOS
+    system "gpg2", "--batch", "--gen-key", "batchgpg"
+
     system "git", "init"
     system bin/"blackbox_initialize", "yes"
+    add_created_key = shell_output("#{bin}/blackbox_addadmin Testing 2>&1")
+    assert_match "<testing@foo.bar>", add_created_key
   end
 end
--- a/Formula/duplicity.rb
+++ b/Formula/duplicity.rb
@@ -334,6 +334,37 @@ class Duplicity < Formula
   end
 
   test do
-    system bin/"duplicity", "--dry-run", "--no-encryption", testpath, "file:///#{testpath}/test"
+    # Note that test do is always in temporary path, with HOME captured,
+    # so this doesn't interfere with a user's pre-existing keychain.
+    (testpath/"batchgpg").write <<-EOS.undent
+    Key-Type: RSA
+    Key-Length: 2048
+    Subkey-Type: RSA
+    Subkey-Length: 2048
+    Name-Real: Testing
+    Name-Email: testing@foo.bar
+    Expire-Date: 1d
+    Passphrase: brew
+    %commit
+    EOS
+    system "gpg2", "--batch", "--gen-key", "batchgpg"
+
+    (testpath/"test/hello.txt").write "Hello!"
+    (testpath/"command.sh").write <<-EOS.undent
+      #!/usr/bin/expect -f
+      set timeout -1
+      spawn #{bin}/duplicity #{testpath} "file://test"
+      expect -exact "Local and Remote metadata are synchronized, no sync needed."
+      expect -exact "Last full backup date: none"
+      expect -exact "GnuPG passphrase:"
+      send -- "brew\n"
+      expect -exact "Retype passphrase to confirm:"
+      send -- "brew\n"
+      expect -exact "No signatures found, switching to full backup."
+      expect eof
+    EOS
+    chmod 0755, testpath/"command.sh"
+    system "./command.sh"
+    assert_match "duplicity-full-signatures", Dir["test/*"].to_s
   end
 end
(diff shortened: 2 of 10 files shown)


Formula/blackbox.rb
```diff
@@ -3,18 +3,39 @@ class Blackbox < Formula
...
+  revision 1
 
   bottle :unneeded
 
-  depends_on :gpg
+  depends_on :gpg => :run
 
   def install
+    # Prefer GPG2 by default.
+    inreplace "bin/_blackbox_common.sh", ":=gpg", ":=gpg2"
```
MikeMcQuaid: Think we should submit this upstream.

Formula/duplicity.rb
```diff
@@ -324,10 +325,46 @@ def install
     # OSX doesn't provide a /usr/bin/python2. Upstream has been notified but
     # cannot fix the issue. See:
     #   https://github.com/Homebrew/homebrew/pull/34165#discussion_r22342214
-    inreplace "#{libexec}/bin/duplicity", "python2", "python"
+    inreplace libexec/"bin/duplicity", "python2", "python"
+
+    # We prefer gpg2, so set a default. This can still be overriden by
+    # passing the `--gpg-binary=` command line flag.
+    inreplace libexec/"lib/python2.7/site-packages/duplicity/globals.py",
+              "gpg_binary = None", "gpg_binary = 'gpg2'"
```
MikeMcQuaid: Think we should submit this upstream (or just install our `gpg2` as `gpg`)

Formula/blackbox.rb
```diff
@@ -3,18 +3,39 @@ class Blackbox < Formula
...
+
     libexec.install Dir["bin/*"]
     bin.write_exec_script Dir[libexec/"*"].select { |f| File.executable? f }
   end
 
   test do
+    # Note that test do is always in temporary path, with HOME captured,
+    # so this doesn't interfere with a user's pre-existing keychain.
+    (testpath/"batchgpg").write <<-EOS.undent
+    Key-Type: RSA
```
MikeMcQuaid: Indent this by two chars. Also, pull this into a test utility function if it's shared elsewhere?

Formula/blackbox.rb
```diff
@@ -3,18 +3,39 @@ class Blackbox < Formula
...
+    Key-Length: 2048
+    Subkey-Type: RSA
+    Subkey-Length: 2048
+    Name-Real: Testing
+    Name-Email: testing@foo.bar
+    Expire-Date: 1d
+    Passphrase: brew
+    %commit
+    EOS
+    system "gpg2", "--batch", "--gen-key", "batchgpg"
```
MikeMcQuaid: `batch.gpg`?

Formula/git-secret.rb
```diff
@@ -15,12 +17,44 @@ class GitSecret < Formula
   depends_on :gpg => :recommended
 
   def install
+    # Prefer GPG2 by default.
+    inreplace "src/_utils/_git_secret_tools.sh", ':="gpg"', ':="gpg2"'
```
MikeMcQuaid: Again, let's upstream or change our binary name.

Formula/blackbox.rb
DomT4: It's planned:

> Eventually the batch generation of a gpg key for proper functionality testing will be split into a Utils method to call, rather than having to write the whole blob into every formula each time, so that is a temporary measure rather than a permanent thing.

Formula/git-secret.rb
```diff
@@ -15,12 +17,44 @@ class GitSecret < Formula
...
     system "bash", "utils/install.sh", prefix
   end
 
+  def caveats; <<-EOS.undent
+    Homebrew defaults `git-secret` to using `gpg2`. If you
+    wish to use alternative GPG executables you should add:
+      export SECRETS_GPG_COMMAND="gpg"
+    to your shell profile.
+  EOS
+  end
```
MikeMcQuaid: Another argument I think for changing our binary name.

Formula/zero-install.rb
```diff
@@ -95,6 +102,9 @@ def install
     archives = opamroot/"repo/default/archives"
     modules = []
     resources.each do |r|
+      if build.without? "gtk+"
```
MikeMcQuaid: `next if build.without?("gtk+") && r.name == "lablgtk" `

Formula/duplicity.rb
DomT4: I tend to lean towards thinking how we name `gpg`s is a packager-specific quirk rather than an upstream problem, in the same way we've had to `inreplace` a few `python2` hashbangs out for OS X.

Sadly this is only going to get a little more tricky, Debian are moving towards `gpg1` for `gnupg` which I think we'll want to mimic eventually, and installing `gnupg2` as `gpg2` with a symlink to `gpg`. I think that's a sane system but I'm pretty reluctant to hurdle Debian.

MikeMcQuaid:
> Eventually the batch generation of a gpg key for proper functionality testing will be split into a Utils method to call, rather than having to write the whole blob into every formula each time, so that is a temporary measure rather than a permanent thing.

Think there's enough usage that it'd be good to do this now.

Would be good to elaborate on where this fits into the larger project.

As an aside I think between both this and https://github.com/Homebrew/brew/pull/386 there's a pretty strong argument that we should just symlink our `gpg2` as `gpg` rather than patching and needing caveats.

Formula/duplicity.rb
MikeMcQuaid: I think we definitely should hurdle Debian rather than patching. I don't think they need to lead the way on this.

Formula/blackbox.rb
DomT4: `batchgpg` is the input script name, rather than anything outputted.

Formula/duplicity.rb
DomT4: I'm mostly appreciative that when Debian leads the way on things that carries more weight, and upstream developers take notice. We've had less "natural" (i.e. unprompted) success using Homebrew's muscle in the same way. Currently `gnupg` conflicts with MacGPG2 as well, because of the `gpg` symlink situation, whereas `gnupg2` doesn't. That's something that needs to be considered, given we're introducing more formal support for MacGPG2 here.

DomT4: I agree it'd make things considerably more simple though in terms of inreplacements.

Formula/zero-install.rb
DomT4: I tried something _like this_ & Rubocop yelled at me. Perhaps I wrote it slightly wrong though, will try this way.

Formula/git-secret.rb
DomT4: Perhaps worth noting if we rename `gpg` to `gpg1` and let `gpg2` symlink itself onto `gpg` we're going to end up with `inreplace`s anyway unless we drop support for `gpg` 1.x in the requirement completely. No upstream script/usage/etc is looking for a `gpg1` binary, as far as I saw during the rewrites and testing here.

The initial round of feedback makes me suspect simply saying "We won't support GPG 1.x any more" is perhaps a rational solution.

Formula/blackbox.rb
MikeMcQuaid: @DomT4 Yeh, I think it would be better to have in this PR if at all possible given the number of repeats.

Formula/blackbox.rb
MikeMcQuaid: Yeh, I just mean I think the script could have a better extension

Formula/duplicity.rb
MikeMcQuaid: If Debian are moving towards it that carries enough weight. The irony is that actually we'd need to have more upstream contact now because our policies suggest these `inreplace`s should all be submitted upstream.

Formula/git-secret.rb
MikeMcQuaid: I think it'd be fine to just make them conflict so you can't have them installed side-by-side.

Formula/zero-install.rb
MikeMcQuaid: If it does again: paste the warning as I'm curious.

PR #2253, opened 2016-06-21
idris: GHC 8 compatibility

ilovezfs:
- restore the Cabal-based build
- use upstream PR for GHC 8 compatibility

### DIFF
--- a/Formula/idris.rb
+++ b/Formula/idris.rb
@@ -7,6 +7,8 @@ class Idris < Formula
   homepage "http://www.idris-lang.org"
   url "https://github.com/idris-lang/Idris-dev/archive/v0.12.tar.gz"
   sha256 "dfc31dffd1bafd996a951cbcc551a69337f16a3fa5c2974ec872d62a38bd7c75"
+  revision 1
+
   head "https://github.com/idris-lang/Idris-dev.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍

ilovezfs:
@DomT4 um this isn't good... did you know that our dependencies are order dependent such that
```
  depends_on "pkg-config" => :build if build.with? "libffi"
  depends_on "libffi" => :recommended
```
does  not work, but
```
  depends_on "libffi" => :recommended
  depends_on "pkg-config" => :build if build.with? "libffi"
```
does work?

So much for all build deps go first unless that's fixed!

DomT4:
Yeah, that's a known issue, and yeah, it's annoying. It works if you write an explicit `option` IIRC but otherwise dependency order does matter.

ilovezfs:
That's quite a spectacular bug.

DomT4:
The problem predates me, it's been around for a while. It tends to get tied up in the _"How do we rewrite our dependency system"_ discussion that happens every few months.

ilovezfs:
That's unfortunate, since in this case it's probably going to be a question of how the formula file is parsed as opposed to anything to do with the mechanics of the dependency system itself.

DomT4:
It gets dragged in because fundamentally `if build.with?` style dependencies are broken. They work, of course, but only if you use them in a certain way, certain order, they're hidden from users in vanilla `brew info` calls, etc.

ilovezfs:
I'm still just "in shock" that something as simple as "list all `=> :build` dependencies first" isn't actually possible in the current system.

UniqMartin:
> did you know that our dependencies are order dependent such that

This is almost completely unrelated to our dependency resolution system. All dependencies hidden behind a trailing `if`/`unless` are problematic, particularly if the condition references other dependencies, they need to be known beforehand. There's no other way given how formula code is evaluated by Ruby.

One way we could make that work, but this taps into the “rewrite our dependency system” discussion, is if we completely banned this type of dependency and delayed the evaluation of these conditions until all dependencies of the formula are known. DSLs of varying flexibility have been proposed for that:

```ruby
# Variant 1.
option "with-libffi" do
end
depends_on "libffi" => :recommended

# Variant 2.
depends_on "pkg-config" => :build, :if => proc { build.with? "libffi" }
depends_on "libffi" => :recommended
```

ilovezfs:
@UniqMartin what about scanning the file to collect the options set before actually evaluating the file?

UniqMartin:
@ilovezfs This really depends on what your understanding of “scanning” is, but I can't think of a flexible way of doing that without actually evaluating the file. (How would scanning account for implicitly defined options and options introduced through helper modules or custom base classes?)

The only feasible approach I can think of that doesn't completely overthrow the concept would be to move to a more declarative DSL as demonstrated in the examples in [my previous comment](https://github.com/Homebrew/homebrew-core/pull/2253#issuecomment-227631708).

ilovezfs:
Your previous comment makes sense to me except that it's not clear to me whether or not it undermines grouping all of the option descriptions together neatly.

PR #2254, opened 2016-06-21
ruby: use "beta" instead of "preview"

ilovezfs:
2.4.0-preview1 > 2.4.0 > 2.4.0-beta1, so fix it by using "beta"

### DIFF
--- a/Formula/ruby.rb
+++ b/Formula/ruby.rb
@@ -25,7 +25,7 @@ class Ruby < Formula
 
   devel do
     url "https://cache.ruby-lang.org/pub/ruby/2.4/ruby-2.4.0-preview1.tar.xz"
-    version "2.4.0-preview1"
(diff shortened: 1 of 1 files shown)


UniqMartin:
I hate to be this picky, but up until the 2.2 series, there have been `-preview*` releases that were followed by `-rc*` releases, so they are not exactly equivalent. There was no `-rc*` release in the 2.3 series, but I'm not sure if this is just an outlier or a general trend that will continue in future releases.

https://bugs.ruby-lang.org/projects/ruby/wiki/ReleaseEngineering suggests that using `alpha` or `beta` to represent `preview` would be more appropriate.

ilovezfs:
@UniqMartin :ship:?

UniqMartin:
:shipit:

PR #2261, opened 2016-06-21
embulk: support Mac OS 10.9 using the same exec script with digdag

frsyuki:

---

This PR improves exec script of embulk using the same strategy discussed at #2012 and treasure-data/digdag#125.
Embulk and Digdag use the same hack to make it executable (because creator is same).

### DIFF
--- a/Formula/embulk.rb
+++ b/Formula/embulk.rb
@@ -8,12 +8,10 @@ class Embulk < Formula
 
   depends_on :java
 
-  skip_clean "libexec"
-
   def install
-    (libexec/"bin").install "embulk-#{version}.jar" => "embulk"
-    chmod 0755, libexec/"bin/embulk"
-    bin.write_exec_script libexec/"bin/embulk"
+    # Execute through /bin/bash to be compatible with OS X 10.9.
+    libexec.install "embulk-#{version}.jar" => "embulk.jar"
+    (bin/"embulk").write "#!/bin/bash\nexec /bin/bash \"#{libexec/"embulk.jar"}\" \"$@\"\n"
   end
 
   test do
--- a/Formula/embulk.rb
+++ b/Formula/embulk.rb
@@ -6,12 +6,15 @@ class Embulk < Formula
 
   bottle :unneeded
 
-  depends_on :java
+  depends_on :java => "1.7+"
 
   def install
     # Execute through /bin/bash to be compatible with OS X 10.9.
     libexec.install "embulk-#{version}.jar" => "embulk.jar"
-    (bin/"embulk").write "#!/bin/bash\nexec /bin/bash \"#{libexec/"embulk.jar"}\" \"$@\"\n"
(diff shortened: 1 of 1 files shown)


Formula/embulk.rb
```diff
@@ -8,12 +8,10 @@ class Embulk < Formula
...
 
-  skip_clean "libexec"
-
   def install
-    (libexec/"bin").install "embulk-#{version}.jar" => "embulk"
-    chmod 0755, libexec/"bin/embulk"
-    bin.write_exec_script libexec/"bin/embulk"
+    # Execute through /bin/bash to be compatible with OS X 10.9.
+    libexec.install "embulk-#{version}.jar" => "embulk.jar"
+    (bin/"embulk").write "#!/bin/bash\nexec /bin/bash \"#{libexec/"embulk.jar"}\" \"$@\"\n"
```
UniqMartin: Please replace this line with the following to make the code easier to read:

```ruby
(bin/"embulk").write <<-EOS.undent
  #!/bin/bash
  exec /bin/bash "#{libexec}/embulk.jar" "$@"
EOS
```

Also, would it make sense to use [`Pathname#write_jar_script`](https://github.com/Homebrew/brew/blob/6dc72f2679208f34d9d09f547dda5a2ba0714a9f/Library/Homebrew/extend/pathname.rb#L425-L432) instead? This would of course skip the Java-specific setup in your binary and leave this to us, but maybe that's fine from your perspective. Here's how that would look like:

```ruby
bin.write_jar_script libexec/"embulk.jar", "embulk"
```

Formula/embulk.rb
```diff
@@ -8,12 +8,10 @@ class Embulk < Formula
 
   depends_on :java
```
UniqMartin: Should we maybe be a bit more specific about the supported Java version, e.g. `depends_on :java => "1.7+"`? (Don't know what the minimum requirement is.) On OS X, an unspecified version essentially means 1.6 or newer.

Formula/embulk.rb
frsyuki: Updated to use EOS :-)
`Pathname#write_jar_script` doesn't work well for embulk because the init script of embulk sets (some parameters)[https://github.com/embulk/embulk/blob/master/embulk-cli/src/main/sh/selfrun.sh#L150-L154] to java. Those parameters are important especially for shorter startup time.

Formula/embulk.rb
frsyuki: It's a good point. I updated to use 1.7+.

Formula/embulk.rb
UniqMartin: Thanks for having a look! Let's stick with the shell wrapper then.

UniqMartin:
Merged in ea98100b3568ea15afe9c9fd64db2b813c48237b. Thank you for this contribution to Homebrew, @frsyuki! :tada:

PR #2263, opened 2016-06-22
emacs: Add option to enable dynamic modules.

atheriel:

---

As of version 25.0.95 Emacs supports dynamic (native code) modules, but they are disabled by default. The `--with-modules` option must be passed to the `./configure` script to enable them. This PR adds
this option to the formula, and warns the user if they try to use it for pre-25.0.95 builds.

Note that when Emacs 25 is released, this option will be available to all builds, and the formula will need to be updated.

### DIFF
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -28,6 +28,7 @@ class Emacs < Formula
   option "with-cocoa", "Build a Cocoa version of emacs"
   option "with-ctags", "Don't remove the ctags executable that emacs provides"
   option "without-libxml2", "Don't build with libxml2 support"
+  option "with-modules", "Build with dynamic module support (devel/HEAD only)"
 
   deprecated_option "cocoa" => "with-cocoa"
   deprecated_option "keep-ctags" => "with-ctags"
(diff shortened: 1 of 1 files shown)


DomT4:
I'm reasonably reluctant to carrying around head/devel options only, truth be told. Feelings on this @dunn?

Formula/emacs.rb
```diff
@@ -87,6 +88,14 @@ def install
     args << "--with-imagemagick" if build.with? "imagemagick"
     args << "--without-popmail" if build.with? "mailutils"
 
+    if build.with? "modules"
```
DomT4: You can probably just condense this into:
```ruby
args << "--with-modules" if !build.stable && build.with?("modules")
```
If we decide we're happy with this.

dunn:
> I'm reasonably reluctant to carrying around head/devel options only, truth be told. Feelings on this @dunn?

Yeah, we should probably wait until 25 is stable.  That should be within a month or two anyway.

atheriel:
I think that's a reasonable response. The one comment that I would make is that the community is now starting to try and build dynamic modules in preparation for 25's release, and this patch would make that much, much easier on Mac (where I believe most installs are through Homebrew).

MikeMcQuaid:
Will reconsider this when 25 is stable, thanks.

PR #2264, opened 2016-06-22
elasticsearch: fix plugin script path in caveats

kawamoto:
ex.
before: `plugin script: /usr/local/Cellar/elasticsearch/2.3.3/libexec/plugin`
after: `plugin script: /usr/local/Cellar/elasticsearch/2.3.3/libexec/bin/plugin`

### DIFF
--- a/Formula/elasticsearch.rb
+++ b/Formula/elasticsearch.rb
@@ -88,14 +88,20 @@ class Elasticsearch < Formula
   end
 
   def caveats
-    plugin=(devel? || head?) ? "#{libexec}/bin/elasticsearch-plugin" : "#{libexec}/bin/plugin"
-    <<-EOS.undent
-    Data:    #{var}/elasticsearch/#{cluster_name}/
-    Logs:    #{var}/log/elasticsearch/#{cluster_name}.log
-    Plugins: #{libexec}/plugins/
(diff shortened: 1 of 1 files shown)


DomT4:
@jasontedor Does this look okay to you?

jasontedor:
@DomT4 Yes, this is a mistake. The reason that the caveat is there is because in the 2.x series the name of the plugin script is just `plugin`. We felt that was too generic to leave on the path so it's not symlinked from `/usr/local/bin`. Thus, the caveat is there to point people to the correct place. This work was done in Homebrew/hombrew#47941. For the 5.x series (i.e., the current `devel`), the script has been renamed to `elasticsearch-plugin` in elastic/elasticsearch#16454. Thus, we now are okay with it being on the path and so symlinked from `/usr/local/bin`. Thus, the caveat is not needed anymore, and instead I think should just be completely removed for `devel`.

kawamoto:
@jasontedor sure, it makes sense. how about now?

jasontedor:
> sure, it makes sense. how about now?

@kawamoto +1, thanks!

UniqMartin:
Merged in 24032a642f1979e402adaac86bda10a1874c4dcb. Thank you for this contribution to Homebrew, @kawamoto! :tada:

PR #2265, opened 2016-06-22
expat 2.2.0

ilovezfs:
Fix for http://seclists.org/oss-sec/2016/q2/360 already in the release


DomT4:
Hurrah for releases that allow us to stop backporting security fixes. :shipit:.

PR #2268, opened 2016-06-22
ats2-postiats: build with clang

ilovezfs:
- build no longer fails with clang
- githwxi/ATS-Postiats/#76 is fixed so need to disable GC support

### DIFF
--- a/Formula/ats2-postiats.rb
+++ b/Formula/ats2-postiats.rb
@@ -13,10 +13,6 @@ class Ats2Postiats < Formula
 
   depends_on "gmp"
 
-  fails_with :clang do
-    cause "Trying to compile this with Clang is failure-galore."
-  end
-
   # error: expected declaration specifiers or '...' before '__builtin_object_size'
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 :ship:?

DomT4:
Nice! If it builds with Clang now that's certainly an improvement 👍.

ilovezfs:
It only came to my attention because the current bottles are broken since it has an undeclared dependency on gcc and thus missed the revision bump.

DomT4:
Yeah, that happens a fair bit on `fails_with :clang` stuff. Feel free to 🚢 when happy.

ilovezfs:
Interesting. So it sounds like it might be worth running `brew linkage` on everything with a `fails_with` to search for other extant dead men walking.

DomT4:
Probably. I've found & fixed a couple before now. I presume people don't really notice it because `gcc` is a massively popular formula.

PR #2273, opened 2016-06-22
passenger 5.0.29

FooBarWidget:


ilovezfs:
@FooBarWidget Shouldn't the fix for https://github.com/phusion/passenger/issues/1288 be removed from the formula?

FooBarWidget:
@ilovezfs I don't know? Why should it be removed from the formula? Did something change recently?

ilovezfs:
@FooBarWidget The upstream issue https://github.com/phusion/passenger/issues/1288 is closed and yet we're still patching it, so I'm confused.

FooBarWidget:
The upstream issue is for the homebrew formula itself. The fix referred to
by the upstream issue *is* the very fix inside the formula.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/97f648f2171a457a6ba4aee28bb7173884d1d1ff. Thank you for your contribution to Homebrew @FooBarWidget; we appreciate it! :smiley_cat:

PR #2286, opened 2016-06-22
gammaray 2.4.1

ilovezfs:
Also,
- always use Qt 5 (and remove the Qt 4 option).
- add option to run `make test` and always run it during CI
- fix case sensitivity issue with the old test

### DIFF
--- a/Formula/gammaray.rb
+++ b/Formula/gammaray.rb
@@ -1,9 +1,8 @@
 class Gammaray < Formula
   desc "Examine and manipulate Qt application internals at runtime"
   homepage "https://www.kdab.com/kdab-products/gammaray/"
-  url "https://github.com/KDAB/GammaRay/releases/download/v2.3.0/gammaray-2.3.0.tar.gz"
-  mirror "https://mirrors.kernel.org/debian/pool/main/g/gammaray/gammaray_2.3.0.orig.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
This might be a good time to kill off that `qt4` option here, but if you want to do that elsewhere, 👍.

ilovezfs:
@DomT4 Qt 4 option dead as a doornail now.

DomT4:
I'm sure that'll immensely annoy some people, but upstream have been building against Qt5 by default for a while now & `qt4` is completely unsupported by Qt themselves, so. 👍 when 🍏.

ilovezfs:
@DomT4 also I just modified the test because it was borked on case-sensitive file system (and because it was a silly test to begin with).

DomT4:
Oh, it only generates an `.app` bundle? Heh. May well end up being Casked at some point soon anyway then, but 👍 on the changes and new test.

MikeMcQuaid:
Upstream are my former coworkers and definitely won't mind the Qt4 option being killed. It doesn't seem they provide binaries otherwise I agree it'd be a good Cask candidate. @frankosterfeld are there any KDAB-provided OS X binaries for this?

UniqMartin:
https://github.com/KDAB/homebrew-tap may be relevant and worth inspecting.

ilovezfs:
been there, done that

ilovezfs:
@DomT4 seems new test caused CI to hang. Let's see if `make test` does any better.

PR #2290, opened 2016-06-22
ironcli 0.1.2

ilovezfs:
- adopt upstream's use of glide for vendoring
- use upstream's description
- upstream patch that corrects the version
- minor tweaks to the install method
- match the version in the test

### DIFF
--- a/Formula/ironcli.rb
+++ b/Formula/ironcli.rb
@@ -1,12 +1,18 @@
-require "language/go"
-
 class Ironcli < Formula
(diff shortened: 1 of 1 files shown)


DomT4:
Looks like upstream expect `glide` to manage the dependencies for them.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Seems reasonable 👍.

PR #2291, opened 2016-06-22
knot 2.2.1

ilovezfs:
- enable dnstap support
- install bash completions
- run `make check`
- install single-html documentation
- fix the test

### DIFF
--- a/Formula/knot.rb
+++ b/Formula/knot.rb
@@ -1,10 +1,23 @@
 class Knot < Formula
   desc "High-performance authoritative-only DNS server"
   homepage "https://www.knot-dns.cz/"
-  url "https://secure.nic.cz/files/knot-dns/knot-2.1.1.tar.xz"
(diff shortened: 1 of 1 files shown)


Formula/knot.rb
```diff
@@ -13,31 +26,63 @@ class Knot < Formula
...
-    system "autoreconf", "-i", "-f" if build.head?
-    system "./configure", "--disable-debug",
-                          "--disable-dependency-tracking",
+    resource("fstrm").stage do
+      if build.stable?
+        Pathname.pwd.install resource("fstrm_clock_patch")
+        system "/usr/bin/patch", "-p1", "-i", "c5f09123.patch"
+      end
+
+      system "autoreconf", "-fvi"
```
dunn: This is required for stable builds too now/

ilovezfs: Correct. https://github.com/farsightsec/fstrm/commit/c5f09123.patch touches configure.ac.
Note https://github.com/farsightsec/fstrm/issues/23#issuecomment-228505430.

PR #2297, opened 2016-06-22
rabbitmq-c 0.8.0

ilovezfs:
Also, use full path in the test.

### DIFF
--- a/Formula/rabbitmq-c.rb
+++ b/Formula/rabbitmq-c.rb
@@ -1,9 +1,8 @@
 class RabbitmqC < Formula
   desc "RabbitMQ C client"
   homepage "https://github.com/alanxz/rabbitmq-c"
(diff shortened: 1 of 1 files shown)


DomT4:
Test needs changing so it uses a full path, but feel free to do in post. 👍.

ilovezfs:
@DomT4 Why haven't we added that to audit?

DomT4:
We discussed it a fair bit, but nobody proposed a sane way to handle every possible thing a formula does. Some formulae install only `.app`s, some formulae don't install executables, some executables don't need to be called with a full path _(If you're calling a system tool, for example)_.

I guess you could scan the keg & do a if x name called in `test do` matches x name found in `bin/sbin` & the test doesn't include the words `bin` or `sbin` throw a tantrum, but I'm sure there's edge cases even there.

PR #2306, opened 2016-06-22
mypy 0.4.2

ilovezfs:
use tag since it has submodules now


ilovezfs:
CC @tdsmith

MikeMcQuaid:
LGTM 👍

tdsmith:
lgtm

PR #2311, opened 2016-06-22
mvtools 14

ilovezfs:
needed upstream commit "Depan: fix compilation with clang/libc++"

### DIFF
--- a/Formula/mvtools.rb
+++ b/Formula/mvtools.rb
@@ -1,9 +1,19 @@
 class Mvtools < Formula
   desc "Filters for motion estimation and compensation"
   homepage "https://github.com/dubhater/vapoursynth-mvtools"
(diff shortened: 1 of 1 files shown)


ilovezfs:
This compiles with gcc but not clang. The regression occurred in https://github.com/dubhater/vapoursynth-mvtools/commit/10c370e19968fef2235e17eb9bad6b3431d46697, and when this commit is reverted, it compiles with clang again.

ilovezfs:
Reported https://github.com/dubhater/vapoursynth-mvtools/issues/27

ilovezfs:
OK, upstream fixed it for us.

DomT4:
👍. Another helpful upstream, we're on a roll this week.

PR #2314, opened 2016-06-22
burp 2.0.44 (devel)

ilovezfs:
- add devel spec and use same steps for head spec

### DIFF
--- a/Formula/burp.rb
+++ b/Formula/burp.rb
@@ -1,10 +1,13 @@
 class Burp < Formula
   desc "Network backup and restore"
   homepage "http://burp.grke.org/"
-  url "https://github.com/grke/burp/archive/1.4.40.tar.gz"
-  sha256 "2e6a9a28453a11f3e36d0beefa185e72e7781a8718b55d3101144c9900752d6f"
   revision 1
-  head "https://github.com/grke/burp.git"
+
+  stable do
+    url "https://github.com/grke/burp/archive/1.4.40.tar.gz"
+    sha256 "2e6a9a28453a11f3e36d0beefa185e72e7781a8718b55d3101144c9900752d6f"
+    patch :DATA
+  end
 
   bottle do
     cellar :any
@@ -13,19 +16,64 @@ class Burp < Formula
     sha256 "d3b8cc95839835dd9ba9574444d5cf7394d398376fe2a06fcd3ba9a00b395476" => :mavericks
   end
 
+  devel do
+    url "https://github.com/grke/burp/archive/2.0.44.tar.gz"
+    sha256 "6c296d8b667c4c5a4569cd1f2d22ce3a4fffdf5e04d014ed0b8d9ef7c2edbff2"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+
+    resource "uthash" do
+      url "https://github.com/troydhanson/uthash.git",
+          :tag => "v2.0.1",
+          :revision => "539b4504b052cfca54ed66b82ca99e3aed403d46"
+    end
+  end
+
+  head do
+    url "https://github.com/grke/burp.git"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+
+    resource "uthash" do
+      url "https://github.com/troydhanson/uthash.git"
+    end
+  end
+
   depends_on "librsync"
   depends_on "openssl"
 
-  # patches to change directories to brew conventions in Makefile and config
-  # files
-  patch :DATA
-
   def install
+    unless build.stable?
+      resource("uthash").stage do
+        system "make", "-C", "libut"
+        (buildpath/"uthash/lib").install "libut/libut.a"
+        (buildpath/"uthash/include").install Dir["src/*"]
+      end
+
+      ENV.prepend "CPPFLAGS", "-I#{buildpath}/uthash/include"
+      ENV.prepend "LDFLAGS", "-L#{buildpath}/uthash/lib"
+
+      system "autoreconf", "-fiv"
+    end
+
     system "./configure", "--prefix=#{prefix}",
                           "--sysconfdir=#{etc}/burp",
                           "--sbindir=#{bin}",
-                          "--localstatedir=#{var}/burp"
-    system "make", "install"
+                          "--localstatedir=#{var}"
+    if build.stable?
+      system "make", "install"
+    else
+      system "make", "install-all"
+    end
+  end
+
+  def post_install
+    (var/"run").mkpath
+    (var/"spool/burp").mkpath
   end
 
   def caveats; <<-EOS.undent
@@ -63,7 +111,7 @@ class Burp < Formula
   end
 
   test do
-    system "#{bin}/burp", "-v"
+    system bin/"burp", "-v"
   end
 end
 


ilovezfs:
CC @bfontaine RE: https://github.com/Homebrew/legacy-homebrew/pull/40393

Formula/burp.rb
```diff
@@ -13,14 +17,51 @@ class Burp < Formula
...
-  # files
-  patch :DATA
-
   def install
+    if build.devel? || build.head?
+      resource("uthash").stage do
+        cd "libut" do
+          system "make"
+          (buildpath/"uthash/lib").install "libut.a"
+        end
```
bfontaine: That’s probably nitpicking but you can do this without `cd`:

```ruby
        system "make", "-C", "libut"
        (buildpath/"uthash/lib").install "libut/libut.a"
```

ilovezfs: Yeah it's deliberate, since I'm not a big fan of "-C" for git or make :)

Formula/burp.rb
```diff
@@ -13,14 +17,51 @@ class Burp < Formula
...
+
   depends_on "librsync"
   depends_on "openssl"
 
-  # patches to change directories to brew conventions in Makefile and config
-  # files
-  patch :DATA
-
   def install
+    if build.devel? || build.head?
```
DomT4: I'd just feel free to use `unless build.stable?` here, if you wanted to. Where it applies to both devel & head `unless build.stable?` is a valid syntax choice.

ilovezfs:
@DomT4 @bfontaine PR refreshed.

DomT4:
Test is failing. Related to the moved executable apparently.

ilovezfs:
LOL yeah, stable decided that it should put them in `/usr/local/burp/1.4.40_1/usr/sbin/`.

DomT4:
Wow 😓.

ilovezfs:
I guess I'll do this
```
diff --git a/Formula/burp.rb b/Formula/burp.rb
index b8a594c..08c8409 100644
--- a/Formula/burp.rb
+++ b/Formula/burp.rb
@@ -61,9 +61,14 @@ class Burp < Formula
       system "autoreconf", "-fiv"
     end
 
-    system "./configure", "--prefix=#{prefix}",
-                          "--sysconfdir=#{etc}/burp",
-                          "--localstatedir=#{var}/burp"
+    args = %W[
+      --prefix=#{prefix}
+      --sysconfdir=#{etc}/burp
+      --localstatedir=#{var}/burp
+    ]
+    args << "--sbindir=#{sbin}" if build.stable?
+
     system "make", "install"
   end
```

Or could just do `--sbindir=#{sbin}` for both stable and devel, even though devel gets it right out of the box.

DomT4:
Proposed diff seems alright 👍.

ilovezfs:
:green_apple:

DomT4:
Is all the stuff we've had to patch fixes in the `devel`, etc releases?

ilovezfs:
The patch applies cleanly because I think all it's doing is switching up installation locations as opposed to actually fixing bugs.

ilovezfs:
Ah nevermind, I see I moved :DATA into stable... I'll investigate further.

ilovezfs:
devel has
```
        $(AM_V_at)$(MKDIR_P) $(DESTDIR)$(localstatedir)/spool/burp
        $(AM_V_at)$(MKDIR_P) $(DESTDIR)$(runstatedir)
./configs/client/burp.conf.in:pidfile = @runstatedir@/burp.client.pid
./configs/server/burp.conf.in:pidfile = @runstatedir@/burp.server.pid
clientconfdir
clientconfdir = $(sysconfdir)/clientconfdir
./configs/client/burp.conf.in:ca_burp_ca = @sbindir@/burp_ca
./configs/server/burp.conf.in:ca_burp_ca = @sbindir@/burp_ca
./configs/client/burp.conf.in:ca_csr_dir = @sysconfdir@/CA-client
./configs/client/burp.conf.in:ssl_cert_ca = @sysconfdir@/ssl_cert_ca.pem
./configs/server/burp.conf.in:ssl_cert_ca = @sysconfdir@/ssl_cert_ca.pem
./configs/client/burp.conf.in:ssl_key = @sysconfdir@/ssl_cert-client.key
```
So all is "properly" prefixed now instead of hard coded.

ilovezfs:
@DomT4 :ship:?

DomT4:
You may possibly want to create `#{var}/burp` in `postinstall`, but if not the case 👍.

ilovezfs:
@DomT4 PR refreshed. Found a couple of bugs. Upstream changed `make install` to not include `install-configs`. So you can either do `make install ; make install-configs` or `make install-all` which does both. I opted for the latter in case they move anything else into that umbrella. The other bug is that in stable, even though the configure sets var to var/burp, if you actually look at what the patch does, it ends up using simply var not var/burp. So stable has been doing this:
```
/usr/local/var
/usr/local/var/run
/usr/local/var/spool
/usr/local/var/spool/burp
```
Do we prefer that or what stable was supposedly doing according to configure (but wasn't) and what devel was indeed doing:
```
/usr/local/var
/usr/local/var/burp
/usr/local/var/burp/run
/usr/local/var/burp/spool
/usr/local/burp/spool/burp
```
If the latter is preferred, then the patch for stable can be fixed accordingly.

Either way, it doesn't need to be part of post install because the bottles include them in the `.bottle` directory, which then gets installed into the prefix. So they get installed by the bottle because they got created during the build.

Formula/burp.rb
```diff
@@ -13,19 +17,60 @@ class Burp < Formula
...
+      system "make", "install"
+    else
+      resource("uthash").stage do
+        system "make", "-C", "libut"
+        (buildpath/"uthash/lib").install "libut/libut.a"
+        (buildpath/"uthash/include").install Dir["src/*"]
+      end
+
+      ENV.prepend "CFLAGS", "-I#{buildpath}/uthash/include"
+      ENV.prepend "CXXFLAGS", ENV.cflags
```
ilovezfs: Hmm I wonder if this should be just CPPFLAGS instead of CFLAGS and CXXFLAGS separately since it's setting an include.

DomT4:
> If the latter is preferred, then the patch for stable can be fixed accordingly.

Either are fine really, but if you want to enforce consistency between stable/devel here I've no objections. All the `var` paths should be created in `post_install` as well, ideally.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Not a massive fan of the lack of spacing between `sha256` & the next line in `stable do`, whether that's a patch or deps, but I seem to be losing that argument in Homebrew so 👍.

ilovezfs:
>Not a massive fan of the lack of spacing between sha256 & the next line in stable do, whether that's a patch or deps, but I seem to be losing that argument in Homebrew so :+1:

I agree with you, but I thought you'd already said that was a lost cause so this was the "correct" way now hehe.

DomT4:
Well, between us two, I'm pretty sure we have more commits to `core` this year than everyone else combined.. so if we wanted to change it I suspect we could 😆.

ilovezfs:
>... so if we wanted to ...

Shhh someone might here our plotting.

ilovezfs:
@DomT4 fixed the grotesque lack of spacing in post: https://github.com/Homebrew/homebrew-core/commit/6ee197091fa7f09232c95556e319679b3b99e6c8#diff-4fb9179650da103a71e3042ee5f3221fR9

DomT4:
> Shhh someone might hear our plotting.

I think at this point the coup of `homebrew/core` is pretty complete 😄.

> fixed the grotesque lack of spacing in post

👍 🎉

ilovezfs:
>I think at this point the coup of homebrew/core is pretty complete :smile:.

:v:

PR #2317, opened 2016-06-22
lmdb 0.9.18

ilovezfs:
- use GitHub https mirror for head
- remove man page location inreplace (default is now already "correct")
- remove explicit calls to `mkpath` since `make install` now works
- run `make test`
- match version in the test

### DIFF
--- a/Formula/lmdb.rb
+++ b/Formula/lmdb.rb
@@ -1,10 +1,9 @@
 class Lmdb < Formula
   desc "Lightning memory-mapped database: key-value data store"
   homepage "https://symas.com/mdb-and-sqlite/"
-  url "https://github.com/LMDB/lmdb/archive/LMDB_0.9.14.tar.gz"
-  sha256 "6447d7677a991e922e3e811141869421a2b67952586aa68a26e018ea8ee3989c"
-
-  head "git://git.openldap.org/openldap.git", :branch => "mdb.master"
+  url "https://github.com/LMDB/lmdb/archive/LMDB_0.9.18.tar.gz"
+  sha256 "dd35b471d6eea84f48f2feece13d121abf59ef255308b8624a36223ffbdf9989"
+  head "https://github.com/LMDB/lmdb.git", :branch => "mdb.master"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Formula/lmdb.rb
```diff
@@ -16,20 +15,15 @@ class Lmdb < Formula
   end
 
   def install
-    inreplace "libraries/liblmdb/Makefile" do |s|
-      s.gsub! ".so", ".dylib"
-      s.gsub! "$(DESTDIR)$(prefix)/man/man1", "$(DESTDIR)$(prefix)/share/man/man1"
+    cd "libraries/liblmdb" do
+      inreplace "Makefile", ".so", ".dylib"
```
DomT4: Would be good to get this reported upstream. Switching extensions around based on platforms is a pretty easy thing to support.

MikeMcQuaid: Agreed 👍

kssreeram:
Hello brewers. Any idea when this will be merged? Thanks.

ilovezfs:
@kssreeram I'll do it next just for you.

Formula/lmdb.rb
ilovezfs: This is upstream's fix: https://github.com/LMDB/lmdb/commit/c6510147cee88d182af431c14a61fad6fae18be3

I'm not sure that's what you two had in mind.

ilovezfs:
@kssreeram Shipped! Now you just need to brew update and upgrade :)

kssreeram:
Thank you!

ilovezfs:
You're welcome :)

Formula/lmdb.rb
DomT4: Not entirely, but eh, passing `"SOEXT=.dylib"` to `make` in future isn't the worst thing in the world.

ilovezfs: > isn't the worst thing in the world

Yeah, it'll work. I won't look a gift horse in the mouth if you're satisifed.

yiakwy:
I just don't understand why did you use "cd" ? I think the old codes work already.

PR #2320, opened 2016-06-22
dbus 1.11.2 (devel)

ilovezfs:
also restore the 1.10.8 bottles


DomT4:
👍 Try to get this in ASAP, because obviously anyone who `brew update`s between your prior `d-bus` merge and this one will be stuck on that `d-bus` version until `1.12.x` is stable. Generally why I prefer reverts, but each to their own.

ilovezfs:
:green_apple:

PR #2321, opened 2016-06-22
protozero 1.3.0 (new formula)

smellman:

---

### DIFF
--- a/Formula/protozero.rb
+++ b/Formula/protozero.rb
@@ -15,5 +15,16 @@ class Protozero < Formula
     files.each do |f|
       assert_equal(File.exist?("#{include}/protozero/#{f}"), true)
     end
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <protozero/version.hpp>
+      int main() {
+        if (PROTOZERO_VERSION_STRING == "#{version}") {
+          return 0;
+        }
+        return 1;
+      }
+    EOS
+    system "c++", testpath/"test.cpp", "-o", "test"
+    system testpath/"test"
   end
 end
--- a/Formula/protozero.rb
+++ b/Formula/protozero.rb
@@ -9,22 +9,32 @@ class Protozero < Formula
   end
 
   test do
-    files = ["byteswap.hpp", "config.hpp", "exception.hpp",
-             "pbf_builder.hpp", "pbf_message.hpp", "pbf_reader.hpp",
-             "pbf_writer.hpp", "types.hpp", "varint.hpp", "version.hpp"]
-    files.each do |f|
-      assert_equal(File.exist?("#{include}/protozero/#{f}"), true)
-    end
     (testpath/"test.cpp").write <<-EOS.undent
-      #include <protozero/version.hpp>
+      #include <protozero/pbf_reader.hpp>
       int main() {
-        if (PROTOZERO_VERSION_STRING == "#{version}") {
-          return 0;
+        const char raw[] = { 0x12, 0x06 };
+        std::string check = "foobar";
+        std::string input(raw, 2);
+        input += check;
+        protozero::pbf_reader message(input);
+        while (message.next()) {
+          switch(message.tag()) {
+            case 2:
+            {
+              std::string s = message.get_string();
+              if (s == check) {
+                return 0;
+              }
+              break;
+            }
+            default:
+              message.skip();
+          }
         }
         return 1;
       }
     EOS
-    system "c++", testpath/"test.cpp", "-o", "test"
-    system testpath/"test"
+    system ENV.cxx, "-I#{include}", "-std=c++11", "test.cpp", "-o", "test"
+    system "./test"
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/protozero.rb
```diff
@@ -0,0 +1,19 @@
...
+    (include/"protozero").install Dir["include/protozero/*"]
+  end
+
+  test do
+    files = ["byteswap.hpp", "config.hpp", "exception.hpp",
+             "pbf_builder.hpp", "pbf_message.hpp", "pbf_reader.hpp",
+             "pbf_writer.hpp", "types.hpp", "varint.hpp", "version.hpp"]
+    files.each do |f|
+      assert_equal(File.exist?("#{include}/protozero/#{f}"), true)
+    end
```
UniqMartin: I'm sorry, but just checking whether the files are there isn't a very useful test. Can you instead create a simple C++ program that uses some functionality of this library, gets compiled, and then executed to verify that it is working as expected?

Formula/protozero.rb
```diff
@@ -0,0 +1,19 @@
+class Protozero < Formula
+  desc "Minimalistic protocol buffer decoder and encoder in C++"
+  homepage "https://github.com/mapbox/protozero"
+  url "https://github.com/mapbox/protozero/archive/v1.3.0.tar.gz"
+  sha256 "85f9238fa662ff06a1e364f1461846a9d377846274e7f98407307e31086cab2b"
+
+  def install
+    (include/"protozero").install Dir["include/protozero/*"]
```
UniqMartin: The upstream `Makefile` doesn't currently have an `install` target, but it would be useful for us and probably also for other package managers if they decide to distribute this library. Would you mind asking upstream for one or (even better) contribute one so that we can make use of it here?

Formula/protozero.rb
smellman: OK, I will try to write test program.

smellman: I write it. https://github.com/Homebrew/homebrew-core/pull/2321/commits/0c0a5d7dc7a59bf0cdbf970fac9fe897d930e676

Formula/protozero.rb
```diff
@@ -0,0 +1,30 @@
...
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <protozero/version.hpp>
+      int main() {
+        if (PROTOZERO_VERSION_STRING == "#{version}") {
+          return 0;
+        }
+        return 1;
+      }
+    EOS
+    system "c++", testpath/"test.cpp", "-o", "test"
```
UniqMartin: Please use `ENV.cxx` instead of `"c++"` and please add another argument `"-I#{include}"` to the invocation, so that the test still works even if Homebrew is not installed to `/usr/local`. You can also use `"test.cpp"` instead of `testpath/"test.cpp"` as `testpath` is the working directory when the test executes. To summarize:

```ruby
system ENV.cxx, "-I#{include}", "test.cpp", "-o", "test"
```

Formula/protozero.rb
```diff
@@ -0,0 +1,30 @@
...
+    system testpath/"test"
```
UniqMartin: You can make that:

```ruby
system "./test"
```

Formula/protozero.rb
UniqMartin: Thanks! The added test is better, though ideally it would do more than just check the version number. Can you maybe make it decode a small hand-crafted message? In any case, you can now remove the check for the existence of all those files (i.e. the old part of the test).

smellman: I write new test program. https://github.com/Homebrew/homebrew-core/pull/2321/commits/c715808bc9dd7edac0b0b7c5381d114eb125ffcc

Formula/protozero.rb
DomT4: We've made something of a habit out of rejecting header-only formulae arguing that sort of situation often means people should be embedding in their projects rather than distributing globally. Not sure how other maintainers feel about this particular case.

UniqMartin: Thanks for your input @DomT4! Now that you mention it, it makes a lot of sense and I should have thought about this before making @smellman do all this work. I think I'm with you that this isn't necessarily a useful addition to Homebrew. Sorry for not realizing this earlier!

MikeMcQuaid:
Passing on this header-only software, sorry. Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

smellman:
I want to make other package requires protozero.
In this case, I need to make all packages require protozero in my tap?

UniqMartin:
@smellman Sorry again for making you do the work without realizing that we probably don't want to have this as a formula in Homebrew.

> I want to make other package requires protozero.
> In this case, I need to make all packages require protozero in my tap?

If you intend to contribute those packages to this repository, then you cannot make them depend on a formula in any other tap (including your tap, if you set one up). You will need to either include `protozero` as a resource in the formula that uses it or if that makes sense, upstream could vendor a copy of `protozero` into their source tree themselves, so it's already a part of the source code of the software that depends on `protozero`. I hope this helps; feel free to ask for clarification, if you still have questions.

PR #2323, opened 2016-06-22
pike: revision to remove gdbm linkage

ilovezfs:
The current lot of pike bottles are linked against libgdbm.4.dylib, but
gdbm is only an optional dependencies and, at least currently, shouldn't
be present in the CI environment while building.

This should fix the `brew linkage pike` failure when gdbm doesn't happen
to be installed.

### DIFF
--- a/Formula/pike.rb
+++ b/Formula/pike.rb
@@ -3,7 +3,7 @@ class Pike < Formula
   homepage "https://pike.lysator.liu.se"
   url "https://pike.lysator.liu.se/pub/pike/all/7.8.866/Pike-v7.8.866.tar.gz"
   sha256 "0b12e1a99bd8bdd9c8a2daa46e623ac718bc0737290236a0c8474091359b594e"
-  revision 2
+  revision 3
 
   bottle do
     sha256 "df8e5edbe83a457ee1c204379a5baa08ec93d4e9e3e8d73450e25e2300d2ddc8" => :el_capitan


ilovezfs:
We should also probably explicitly exclude gdbm with ` --without-gdbm` when `build.without? "gdbm"`, but I first want to verify the revision bump alone is sufficient to fix this for the bottles.

ilovezfs:
Yup, now only jpeg mentioned as an undeclared dependency and no sign of gdbm:
```
iMac-TMP:~ joe$ brew linkage pike
System libraries:
  /System/Library/Frameworks/GLUT.framework/Versions/A/GLUT
  /System/Library/Frameworks/Kerberos.framework/Versions/A/Kerberos
  /System/Library/Frameworks/OpenGL.framework/Versions/A/Libraries/libGL.dylib
  /System/Library/Frameworks/OpenGL.framework/Versions/A/Libraries/libGLU.dylib
  /usr/lib/libSystem.B.dylib
  /usr/lib/libbz2.1.0.dylib
  /usr/lib/libgcc_s.1.dylib
  /usr/lib/libiconv.2.dylib
  /usr/lib/libsqlite3.dylib
  /usr/lib/libz.1.dylib
Homebrew libraries:
  /usr/local/opt/gmp/lib/libgmp.10.dylib (gmp)
  /usr/local/opt/jpeg/lib/libjpeg.8.dylib (jpeg)
  /usr/local/opt/libtiff/lib/libtiff.5.dylib (libtiff)
  /usr/local/opt/nettle/lib/libnettle.6.dylib (nettle)
  /usr/local/opt/pcre/lib/libpcre.1.dylib (pcre)
  jpeg

```

Issue #2274, opened 2016-06-22
gd fails to build on 10.11.5


copmill:
gd fails to build with the following error:

gdxpm.c:29:10: fatal error: 'X11/xpm.h' file not found

`brew gist-logs gd` output:

https://gist.github.com/6452f68ba9833f3ce8b1261f716486b0

tdsmith:
I can't reproduce; configure is finding lXpm on your system and isn't on mine. What's the output of `mdfind -name libXpm` for you?

copmill:
mdfind -name libXpm output:

/opt/X11/lib/libXpm.4.dylib

tdsmith:
Aha, this should help: can you add this to the top of the `install` method, try to reinstall gd, and post the gist-logs again?

`ENV.append_to_cflags "-Wl,-t"`

copmill:
New brew gist-logs gd output:

https://gist.github.com/158faa6355d1b5bc6eae722bacd1d35e

tdsmith:
You have a /usr/local/lib/libXpm.dylib that configure is finding, and once it finds that it's assuming that you need the X11 headers to build against it. https://gist.github.com/anonymous/158faa6355d1b5bc6eae722bacd1d35e#file-config-log-L2398

We should probably just pass `--without-xpm` to gd's `configure` but (re)moving that file might save you from future grief.

copmill:
It's a symlink to /opt/X11/lib/libXpm.4.dylib. I don't know where it came from but I just removed it. gd builds fine now.

tdsmith:
:+1: Great! I'll close this in favor of the PR. Let us know if we can help with anything else.

PR #2327, opened 2016-06-23
gd: disable XPM support

tdsmith:
Prevents build failures like #2274.


DomT4:
Crumbs, I thought when I added `--without-x11` that would imply `--also-without-other-x11-bits` but apparently not. 👍 when you're happy.

PR #2345, opened 2016-06-23
ncmpcpp 0.7.4

smancill:

---

I have removed the required dependency on ICU, which should be optional. There was a [bug in the configure script for 0.7.x](https://github.com/arybczak/ncmpcpp/issues/127) that forced to use boost with ICU support, but it is fixed on this release.

This ICU dependency was forcing a full compilation of boost just to install ncmpcpp, which it is a pain.

### DIFF
--- a/Formula/ncmpcpp.rb
+++ b/Formula/ncmpcpp.rb
@@ -1,9 +1,8 @@
 class Ncmpcpp < Formula
   desc "Ncurses-based client for the Music Player Daemon"
   homepage "http://rybczak.net/ncmpcpp/"
-  url "http://rybczak.net/ncmpcpp/stable/ncmpcpp-0.7.3.tar.bz2"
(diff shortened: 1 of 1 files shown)


Formula/ncmpcpp.rb
```diff
@@ -34,10 +33,10 @@ class Ncmpcpp < Formula
   depends_on "fftw" if build.with? "visualizer"
 
   if MacOS.version < :mavericks
-    depends_on "boost" => ["with-icu4c", "c++11"]
+    depends_on "boost" => ["c++11"]
```
ilovezfs: You should also drop the brackets since it not longer needs to be an array.

smancill:
Updated the commit to remove the brackets.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/2d39ac15fefa99f510d582dd48360565834f8453. Thank you for your contribution to Homebrew @smancill; we appreciate it! :smiley_cat:

PR #2365, opened 2016-06-24
percol 0.2.1 (new formula)

zmwangx:

---

A word on the test block: I can't do much better than a version test due to the interactive nature of `percol`.

### DIFF
--- a/Formula/percol.rb
+++ b/Formula/percol.rb
@@ -1,0 +1,42 @@
+class Percol < Formula
+  desc "Interactive grep tool"
+  homepage "https://github.com/mooz/percol"
+  url "https://github.com/mooz/percol/archive/v0.2.1.tar.gz"
+  sha256 "75056ba1fe190ae4c728e68df963c0e7d19bfe5a85649e51ae4193d4011042f9"
+  head "https://github.com/mooz/percol.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  resource "six" do
+    url "https://files.pythonhosted.org/packages/b3/b2/238e2590826bfdd113244a40d9d3eb26918bd798fc187e2360a8367068db/six-1.10.0.tar.gz"
+    sha256 "105f8d68616f8248e24bf0e9372ef04d3cc10104f1980f54d57b2ce73a5ad56a"
+  end
+
+  resource "cmigemo" do
+    url "https://files.pythonhosted.org/packages/2f/e4/374df50b655e36139334046f898469bf5e2d7600e1e638f29baf05b14b72/cmigemo-0.1.6.tar.gz"
+    sha256 "7313aa3007f67600b066e04a4805e444563d151341deb330135b4dcdf6444626"
+  end
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+    resources.each do |r|
+      r.stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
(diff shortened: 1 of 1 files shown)


Formula/percol.rb
```diff
@@ -0,0 +1,37 @@
+class Percol < Formula
+  desc "Interactive grep tool"
+  homepage "https://github.com/mooz/percol"
+  url "https://github.com/mooz/percol/archive/v0.2.1.tar.gz"
+  sha256 "75056ba1fe190ae4c728e68df963c0e7d19bfe5a85649e51ae4193d4011042f9"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  resource "six" do
+    url "https://pypi.python.org/packages/source/s/six/six-1.10.0.tar.gz"
```
ilovezfs: current preferred url is https://files.pythonhosted.org/packages/b3/b2/238e2590826bfdd113244a40d9d3eb26918bd798fc187e2360a8367068db/six-1.10.0.tar.gz

Formula/percol.rb
```diff
@@ -0,0 +1,37 @@
...
+    sha256 "105f8d68616f8248e24bf0e9372ef04d3cc10104f1980f54d57b2ce73a5ad56a"
+  end
+
+  resource "cmigemo" do
+    url "https://pypi.python.org/packages/source/c/cmigemo/cmigemo-0.1.6.tar.gz"
```
ilovezfs: https://files.pythonhosted.org/packages/2f/e4/374df50b655e36139334046f898469bf5e2d7600e1e638f29baf05b14b72/cmigemo-0.1.6.tar.gz

ilovezfs:
@zmwangx You can do something like this:
```
  test do
    (testpath/"expect_script").write <<-EOS.undent
      spawn #{bin}/percol --auto-match --query=_windowserver psaux.txt
      expect "QUERY> _windowserver"
      send "\r\003"
    EOS
    shell_output("/bin/ps aux >psaux.txt")
    output = shell_output("/usr/bin/expect -f expect_script")
    assert_match "CoreGraphics.framework", output
  end
```

Formula/percol.rb
zmwangx: @ilovezfs Why not https://pypi.python.org/packages/b3/b2/238e2590826bfdd113244a40d9d3eb26918bd798fc187e2360a8367068db/six-1.10.0.tar.gz?

ilovezfs: @zmwangx https://github.com/tdsmith/homebrew-pypi-poet is generally the way of generating these blocks and @tdsmith has recently switched over the URLs to the pythonhosted.org ones.

zmwangx:
Good call on using expect. I updated the test to grep Homebrew's README, which is more cross platform.

ilovezfs:
Looks good. Did you want to add `head` too?

zmwangx:
Well, I'm generally hesitant to add head, because it's really a horrible experience when it comes to updates. But it doesn't hurt to include it in the formula, so I'll add it if you say so.

ilovezfs:
Makes no difference to me. Leave out it if you'd like.

zmwangx:
Added anyway.

MikeMcQuaid:
Thanks again @zmwangx!

Formula/percol.rb
tdsmith: FWIW I do not prefer one URL over the other; homebrew-pypi-poet switched to the new Warehouse PyPI API endpoint in order to get hashes over the wire and those happen to be the URLs Warehouse is returning

zmwangx: Fair enough. (So Warehouse's API is actually usable at this point?)

PR #2369, opened 2016-06-24
juju 2.0-beta10 (devel)

tomck:

---

[Changelog](https://launchpad.net/juju-core/trunk/2.0-beta10#milestone-specs-bugs)

p.s. would be good to add beta9 and such to -versions, because they break things every release, a cloud controller build on beta9 doesn't work with beta10.

### DIFF
--- a/Formula/juju.rb
+++ b/Formula/juju.rb
@@ -13,7 +13,7 @@ class Juju < Formula
 
   devel do
     version "2.0-beta10"
-    url "https://launchpad.net/juju-core/trunk/#{version}/+download/juju-core_#{version}.tar.gz"
+    url "https://launchpad.net/juju-core/trunk/2.0-beta10/+download/juju-core_2.0-beta10.tar.gz"
     sha256 "ca1ed827c27b345884aff7e316d54b4929a7e38d6442ffd72d05dd797e8870c8"
   end
 
--- a/Formula/juju.rb
+++ b/Formula/juju.rb
@@ -12,9 +12,9 @@ class Juju < Formula
   end
 
   devel do
-    url "https://launchpad.net/juju-core/trunk/2.0-beta9/+download/juju-core_2.0-beta9.tar.gz"
-    sha256 "0f201909de0c77be21097f7749a32c131606e86a4b5940484d2fe668c108c22b"
-    version "2.0-beta9"
+    version "2.0-beta10"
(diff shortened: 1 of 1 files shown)


sinzui:
I think we need a new formula for juju 2.0 because it is not and will not be compatible with juju 1.x.

When Juju 2.0 because the sable release in a few weeks, we don't want to update this formula to use it. If we didi, users who get updates will discover that cannot manage their cloud deployments. This would be a disaster for any organisation running production applications.

I suggest we create a new formula (juju2.rb) that users can choose to install. We may need to make the two formulas co-installable so that users can migrate deployments from Juju1 to juju2. We can expect juju 1.25.x to be supported for about 2 more years.

tomck:
I agree with @sinzui, but was skeptical the separate item would be accepted. My reasoning for using juju2 is the support for Azure is pretty broken in juju1.

sinzui:
Yeah, juju2 + azure ARM is much more reliable. I think beta10 as a --devel in the formula is fine...I had just pushed up the same change as you @tomck, then I saw your PR. Users opt in when they specifiy --devel. In the next 2 weeks though, I think we need another formula to solve the compatibility issue. I already have juju 1.25.5 and juju 2.0-beta10 on my localhost so that I can manage new and old deployments,

tomck:
Is a beta only version acceptable, though? Or would it need to rest in homebrew/devel-only?

MikeMcQuaid:
> Is a beta only version acceptable, though? Or would it need to rest in homebrew/devel-only?

It would, yes.

Formula/juju.rb
```diff
@@ -12,9 +12,9 @@ class Juju < Formula
   end
 
   devel do
-    url "https://launchpad.net/juju-core/trunk/2.0-beta9/+download/juju-core_2.0-beta9.tar.gz"
-    sha256 "0f201909de0c77be21097f7749a32c131606e86a4b5940484d2fe668c108c22b"
-    version "2.0-beta9"
+    version "2.0-beta10"
```
MikeMcQuaid: Leave `version` after the `url`, thanks.

tomck:
I would like to close this request per @sinzui, and possibly remove the --devel option for juju due to the imminent release (fingers crossed) of juju2, and request homebrew/homebrew-devel-only#47 be approved.

PR #2377, opened 2016-06-24
sane-backends 1.0.25

yurikoles:
Resubmission of previous old patch https://github.com/Homebrew/legacy-homebrew/pull/44655

### DIFF
--- a/Formula/sane-backends.rb
+++ b/Formula/sane-backends.rb
@@ -1,18 +1,22 @@
 class SaneBackends < Formula
   desc "Backends for scanner access"
   homepage "http://www.sane-project.org/"
-  url "https://fossies.org/linux/misc/sane-backends-1.0.24.tar.gz"
-  mirror "https://mirrors.kernel.org/debian/pool/main/s/sane-backends/sane-backends_1.0.24.orig.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Looks like we may need to work on the patch more here, given upstream have [rejected](https://lists.alioth.debian.org/pipermail/sane-devel/2015-October/034011.html) the existing one:

>  However, I don't think that your fix is the right way to go about fixing this. It is really something that configure should check for.

> Feel free to keep using this patch yourself but I will not commit it to Alioth's master.

I can't speak to how fair or unfair the judgement of your patch is by upstream, but since we try very hard not to carry patches forever we would prefer a solution that upstream hasn't shot down already.

yurikoles:
@DomT4 in fact those issues was addressed in upstream's master, but they are introduced new one. So I have sent another patch, that is already merged. So we can now build master (HEAD) without patches.

yurikoles:
I don't need it now, but I have question: could one formula contain different sets of patches for stable/devel/head?

MikeMcQuaid:
> I don't need it now, but I have question: could one formula contain different sets of patches for stable/devel/head?

Yes, just put the `patch do` inside e.g. `stable do`/`devel do`.

PR #2382, opened 2016-06-25
mpv 0.18.0

ilovezfs:
- always build the shared library
- move "applications" and "icons" to `pkgshare`
- set confdir to `"#{etc}/mpv"`
- gpl3 option no longer exists
- libarchive support

### DIFF
--- a/Formula/mpv.rb
+++ b/Formula/mpv.rb
@@ -1,8 +1,8 @@
 class Mpv < Formula
   desc "Free, open source, and cross-platform media player"
   homepage "https://mpv.io"
-  url "https://github.com/mpv-player/mpv/archive/v0.17.0.tar.gz"
-  sha256 "602cd2b0f5fc7e43473234fbb96e3f7bbb6418f15eb8fa720d9433cce31eba6e"
+  url "https://github.com/mpv-player/mpv/archive/v0.18.0.tar.gz"
+  sha256 "b656638d4f6bce2621baaacb60d8be384aa492fcd86dfd43996aaa2c16fee02b"
   head "https://github.com/mpv-player/mpv.git"
 
   bottle do
@@ -11,7 +11,6 @@ class Mpv < Formula
     sha256 "6df154963d1757f6abb2b5255049fd3e91c7585fa72bd8b02834e3448e4117fe" => :mavericks
   end
 
-  option "with-shared", "Build libmpv shared library."
   option "with-bundle", "Enable compilation of the .app bundle."
 
   depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


gtm87x:
> always build the shared library

The shared library is only useful to devs or programs that make use of it (that are pretty much absent on Os X as of now, AFAIK), it's useless for the vast majority of "plain" mpv's users, why build it by default?

> libarchive support

Libarchive is disabled by default on upstream due to uncertain reliability. Maybe it's better to add it as an option instead of enable it by default?

ilovezfs:
I don't see any harm in building the library by default. It's tiny and it's one less option to deal with. As for libarchive being "unstable," do you have a reference?

gtm87x:
From the commit in the mpv repo: https://github.com/mpv-player/mpv/commit/2b280f4522a288429253b396b778d60ed018312c 

> This is disabled by default for now. One reason is that we try
> libarchive on every file we open, before trying libavformat, and I'm not
> sure if I trust libarchive that much yet. Another reason is that this
> breaks multivolume rar support. While libarchive supports seeking in
> rar, and (probably) supports multivolume archive, our support of
> libarchive (probably) does not.

Now it does support multivolume archives, but it's still disabled by default in the build script, so I think the main concern still stands.

ilovezfs:
thanks, I'll make it an option.

ilovezfs:
@gtm87x ok libarchive is now just a (non-default) option.

Formula/mpv.rb
```diff
@@ -11,20 +11,21 @@ class Mpv < Formula
...
 
-  option "with-shared", "Build libmpv shared library."
   option "with-bundle", "Enable compilation of the .app bundle."
 
   depends_on "pkg-config" => :build
   depends_on :python3
 
   depends_on "libass"
   depends_on "ffmpeg"
+  depends_on "cmocka"
```
MikeMcQuaid: Is this just for mocking and only needed at build time?

MikeMcQuaid:
Otherwise 👍

Formula/mpv.rb
ilovezfs: yes only for the test suite, but it ends up linked.

MikeMcQuaid: Think it's worth avoiding the test suite, then.

ilovezfs: okie dokie

jnozsc:
@ilovezfs is there any reason to close this pr instead of make it ready to merge?

ilovezfs:
@jnozsc it's merged ... the messages are misleading since the PR is rebased before merge, so the exact same commit from the PR is not the actual commit that gets merged into master:
https://github.com/Homebrew/homebrew-core/commit/6285427d26e8de2b950ead20d2d6e82fa61d5f5e
https://github.com/Homebrew/homebrew-core/commit/dbf910e2ff5f0910842383302add8054137d19ae

jnozsc:
gotta thanks

PR #2386, opened 2016-06-25
liquidprompt 1.11

ilovezfs:
- add a test

### DIFF
--- a/Formula/liquidprompt.rb
+++ b/Formula/liquidprompt.rb
@@ -1,8 +1,8 @@
 class Liquidprompt < Formula
   desc "Adaptive prompt for bash and zsh shells"
   homepage "https://github.com/nojhan/liquidprompt"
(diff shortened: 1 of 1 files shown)


Formula/liquidprompt.rb
```diff
@@ -26,4 +26,10 @@ def caveats; <<-EOS.undent
     that will break things.
     EOS
   end
+
+  test do
+    liquidpromp = "#{HOMEBREW_PREFIX}/share/liquidprompt"
```
MikeMcQuaid: `liquidpromp`, no 🍵 for you?

MikeMcQuaid:
typo but otherwise 👍

Formula/liquidprompt.rb
ilovezfs: Wow. :see_no_evil:.

PR #2388, opened 2016-06-25
glade 3.20

neffs:

---

This changes depencencies from gtk+ to gtk+3.
Currently needs a patched Makefile (using inreplace). This is already reported upstream: https://bugzilla.gnome.org/show_bug.cgi?id=768032

### DIFF
--- a/Formula/glade.rb
+++ b/Formula/glade.rb
@@ -1,27 +1,19 @@
 class Glade < Formula
   desc "RAD tool for the GTK+ and GNOME environment"
   homepage "https://glade.gnome.org/"
-  url "https://download.gnome.org/sources/glade3/3.8/glade3-3.8.5.tar.xz"
-  sha256 "58a5f6e4df4028230ddecc74c564808b7ec4471b1925058e29304f778b6b2735"
-  revision 1
-
-  bottle do
-    sha256 "5c910f1666f6e261369aa166bea2e48e5b03856928900448c2c577ec4f04eca2" => :el_capitan
-    sha256 "8f0c1d019b71e76381c0f4437395f1e1eb3da24e8ef1b16051535209aeeaf5d7" => :yosemite
-    sha256 "f4eae837311b7aadfff8a27720ce2bbc71871e9b095d31e90d4b833ec725205f" => :mavericks
-  end
+  url "https://download.gnome.org/sources/glade/3.20/glade-3.20.0.tar.xz"
+  sha256 "82d96dca5dec40ee34e2f41d49c13b4ea50da8f32a3a49ca2da802ff14dc18fe"
 
   depends_on "pkg-config" => :build
   depends_on "intltool" => :build
+  depends_on "itstool" => :build
+  depends_on "docbook-xsl" => :build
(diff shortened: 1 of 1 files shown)


DomT4:
We'll want to add this as a separate `glade-3` formula and keep the old one as `glade` if it requires using `gtk+3`. We maintain a lot of side-by-side formulae for the `gtk+`/`gtk+3` split.

My confusion around this, particularly since the current formula installs a `glade-3` executable & this one installs a `glade` executable _(when theoretically it should be the other way around?)_ is why I left this bump out of my GNOME PR recently.

neffs:
Do you really need the old one if GTK3 libs are available?

Maybe we can rename the old one to glade3.8 and create the new version as glade3?

DomT4:
I'll defer to @tschoonj who initially submitted this update as a new formula, but generally we try not to yank the same formula across different `gtk` lines as people may be relying on `glade` being tied into the `gtk+` dependency tree. I'm not sure why we'd need to consider a rename here.

tschoonj:
Not sure what to do here: 3.8.5 is the last version that can be used to generate gtk+2 user interfaces so it may make sense to keep it around a separate formula. On the other hand no one should be writing gtk+2 user interfaces anymore nowadays...

Formula/glade.rb
```diff
@@ -1,42 +1,38 @@
...
     # Find our docbook catalog
     ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
 
     system "./configure", "--disable-debug",
                           "--disable-dependency-tracking",
-                          "--prefix=#{prefix}"
+                          "--prefix=#{prefix}",
+                          "--enable-gladeui"
+    # objective-c is needed for glade-registration.c. unfortunately build fails if -x objective-c is added to global CFLAGS.
+    inreplace "src/Makefile", "-c -o glade-glade-registration.o", "-x objective-c -c -o glade-glade-registration.o"
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

DomT4:
@tschoonj Can you clear up a bit why upstream named the version that links against `gtk+` `glade-3` & the version that links against `gtk+3` a vanilla `glade`, if you know at all? I'm a bit confused by what that means.

Formula/glade.rb
neffs: Could you be more specific? 

Bugzilla link is in the commit message and the PR. Explanation is in the formula.

MikeMcQuaid: If this patch was submitted upstream to their bugzilla could you make that an inline comment pointing to it in the Ruby file too? Thanks.

MikeMcQuaid:

For future reference the preferred commit message format for simple version updates is `glade 3.20`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md). Thanks!

PR #2409, opened 2016-06-27
opencsg 1.4.0 (new formula)

kintel:

---

New formula, OpenCSG. This is needed by OpenSCAD (which is not yet in Homebrew).

### DIFF
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -15,9 +15,9 @@ class Opencsg < Formula
     system "make", "install"
   end
 
-  def patches
-    # Patch to make the library install correctly
-    DATA
+  patch do
+    url "https://raw.githubusercontent.com/openscad/homebrew-tap/master/patches/opencsg.diff"
+    sha256 "9d710cf6c2d5495ca5ba51c0319785cefc21477c85fa3aacb9ccd3473fee54f3"
   end
 
   test do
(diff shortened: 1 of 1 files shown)


Formula/opencsg.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def install
+    system "qmake", "-r", "INSTALLDIR=#{prefix}",
+      "INCLUDEPATH+=#{HOMEBREW_PREFIX}/include",
+      "LIBS+=-L#{HOMEBREW_PREFIX}/lib -lGLEW"
+
+    system "make", "install"
+  end
+
+  def patches
```
DomT4: We haven't used `def patches` syntax for a _long_ time, heh. You'll want to submit the patch to https://github.com/Homebrew/formula-patches & then use a `patch do` blob. Has this patch been submitted upstream though?

kintel: I'll update the patch. In terms of patch location, it's a bit of a chicken & egg problem - will formula-patches accept the patch prior to the formula, or do I just pull it from a private location in the interim?

I'm submitting the patch upstream once I'm confident that it will work with Homebrew, as the upstream author tends to take time to review such patches.

kintel:
Updated to ```patch do```. Could someone sanity-check this before I submit the patch to _formula-patches_ ?

DomT4:
The blob should be above `def install`, and when accepted to `formula-patches` we'd want it pegged to the commit instead of the master branch, in case it's ever accidentally deleted, etc.

Has this patch been submitted anywhere upstream though? That's the big factor for us.

kintel:
@DomT4 I'll submit the patch upstream. I'm hoping someone from the Homebrew team could take a glance at it for compliance first, as I'm uncertain if I've added all that's needed.

kintel:
@DomT4 Fixed patch location. Pull patch from _formula_patches_. Submitted the patch upstream.

Formula/opencsg.rb
```diff
@@ -0,0 +1,38 @@
+class Opencsg < Formula
+  desc "The CSG rendering library"
+  homepage "http://www.opencsg.org"
+  url "http://www.opencsg.org/OpenCSG-1.4.0.tar.gz"
+  sha256 "ecb46be54cfb8a338d2a9b62dec90ec8da6c769078c076f58147d4a6ba1c878d"
+
+  depends_on "qt5" => :build
+  depends_on "glew"
+
+  patch do
```
MikeMcQuaid: Can you please add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/opencsg.rb
```diff
@@ -0,0 +1,38 @@
...
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/814a29d3ce4c6e7e919cd2fcd64bf45d421e821b/opencsg/patch-build.diff"
+    sha256 "9d710cf6c2d5495ca5ba51c0319785cefc21477c85fa3aacb9ccd3473fee54f3"
+  end
+
+  def install
+    system "qmake", "-r", "INSTALLDIR=#{prefix}",
+      "INCLUDEPATH+=#{HOMEBREW_PREFIX}/include",
```
MikeMcQuaid: You don't want to include all of `#{HOMEBREW_PREFIX}/include` but instead the specific directories for specific dependencies.

Formula/opencsg.rb
```diff
@@ -0,0 +1,38 @@
+class Opencsg < Formula
+  desc "The CSG rendering library"
+  homepage "http://www.opencsg.org"
+  url "http://www.opencsg.org/OpenCSG-1.4.0.tar.gz"
+  sha256 "ecb46be54cfb8a338d2a9b62dec90ec8da6c769078c076f58147d4a6ba1c878d"
+
+  depends_on "qt5" => :build
```
MikeMcQuaid: It doesn't link to this at runtime?

kintel: No, it only needs qmake from qt5

Formula/opencsg.rb
kintel: Do you have an example of formulas that do what you suggest? I've looked around and all formulas I can find either use ```#{HOMEBREW_PREFIX}``` directly, or ```#{HOMEBREW_PREFIX}/include```.
The directory in question that we want to use is (on my system) ```/usr/local/include```, not any deeper directory.

Formula/opencsg.rb
kintel: done

Formula/opencsg.rb
dunn: What is the dependency it's searching for?

kintel: glew

dunn: Then you should be able to use `#{Formula["glew"].opt_include}` instead.

kintel: Thx, I've updated it. Btw., I couldn't find that documented anywhere. Perhaps a good thing to add to the formula cookbook?

MikeMcQuaid:

PR #2298, opened 2016-06-22
rrdtool 1.6.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #2416, opened 2016-06-27
ffmpeg 3.1

zmwangx:

---

https://github.com/FFmpeg/FFmpeg/releases/tag/n3.1.

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -1,9 +1,19 @@
 class Ffmpeg < Formula
   desc "Play, record, convert, and stream audio and video"
   homepage "https://ffmpeg.org/"
-  url "https://ffmpeg.org/releases/ffmpeg-3.0.2.tar.bz2"
-  sha256 "30e3c77c2f4c358ed087869455a7496cbd7753a5e1b98d20ba49c1004009fd36"
   head "https://github.com/FFmpeg/FFmpeg.git"
+
+  stable do
+    url "https://ffmpeg.org/releases/ffmpeg-3.1.tar.bz2"
+    sha256 "2100fca81627e6cbe937fd6a071ae89277c02350538944b2b0c3c2cc71d9402a"
+
+    # See https://github.com/Homebrew/homebrew-core/pull/2416 and
+    # https://trac.ffmpeg.org/ticket/5670
+    patch do
+      url "https://git.videolan.org/?p=ffmpeg.git;a=patch;h=c5566f0a944e376b39c8f994659060ca036c441d"
+      sha256 "408d11d8ae496ed4f83e36e1da0db4b373e8c016cb3723e029ba2fd25b1e7611"
+    end
+  end
 
   bottle do
     sha256 "427e874885b03997de009cec923ed9872a498ff431bc24d007c191658e4633b3" => :el_capitan
@@ -82,17 +92,18 @@ class Ffmpeg < Formula
   depends_on "xz" => :optional
 
   def install
-    args = ["--prefix=#{prefix}",
-            "--enable-shared",
-            "--enable-pthreads",
-            "--enable-gpl",
-            "--enable-version3",
-            "--enable-hardcoded-tables",
-            "--enable-avresample",
-            "--cc=#{ENV.cc}",
-            "--host-cflags=#{ENV.cflags}",
-            "--host-ldflags=#{ENV.ldflags}",
-           ]
+    args = %W[
+      --prefix=#{prefix}
+      --enable-shared
+      --enable-pthreads
+      --enable-gpl
+      --enable-version3
+      --enable-hardcoded-tables
+      --enable-avresample
+      --cc=#{ENV.cc}
+      --host-cflags=#{ENV.cflags}
+      --host-ldflags=#{ENV.ldflags}
+    ]
 
     args << "--enable-opencl" if MacOS.version > :lion
 
--- a/Formula/mpv.rb
+++ b/Formula/mpv.rb
@@ -3,6 +3,7 @@ class Mpv < Formula
   homepage "https://mpv.io"
   url "https://github.com/mpv-player/mpv/archive/v0.18.0.tar.gz"
   sha256 "b656638d4f6bce2621baaacb60d8be384aa492fcd86dfd43996aaa2c16fee02b"
+  revision 1
   head "https://github.com/mpv-player/mpv.git"
 
   bottle do


zmwangx:
Looks like `unpaper` and `mpv` need a bump. (I don't know about `unpaper`, but I can verify locally that without recompiling `mpv` against FFmpeg 3.1, `mpv` won't play audio.)

zmwangx:
Turns out `mpv` needs a bump, but the `unpaper` situation is more complicated. There might be a regression in FFmpeg 3.1 that prevents `test.pbm` from being decoded (it decodes fine under FFmpeg 3.0.2). Try `ffmpeg -i test.pbm out.png`.

ilovezfs:
@zmwangx This will address the problem: https://github.com/Homebrew/homebrew-core/pull/2425

zmwangx:
@ilovezfs Actually, I wouldn't call this an incompatibility between FFmpeg and unpaper, because this is a pbm decoding issue intrinsic to FFmpeg, that is,

```bash
$ cat >test.pbm <<EOF
P1
6 10
0 0 0 0 1 0
0 0 0 0 1 0
0 0 0 0 1 0
0 0 0 0 1 0
0 0 0 0 1 0
0 0 0 0 1 0
1 0 0 0 1 0
0 1 1 1 0 0
0 0 0 0 0 0
0 0 0 0 0 0
EOF
$ brew switch ffmpeg 3.0.2
...
$ ffmpeg -y -i test.pbm out.png # works
...
$ brew switch ffmpeg 3.1
...
$ ffmpeg -y -i test.pbm out.png # fails
...
Error while decoding stream #0:0: Invalid data found when processing input
...
```

The conclusion is not definitive, because I haven't got around to actually investigate the issue (haven't even tested FFmpeg's HEAD).

But of course, if upstream develops against Libav, linking against Libav should be better.

ilovezfs:
@zmwangx Yeah, I'm not sure if it's an "incompatibility" or not. What I had in mind specificially is this: https://github.com/Flameeyes/unpaper/pull/42

I tried that patch, and it didn't fix the problem you identified, but I wouldn't be surprised at all if they're related issues.

zmwangx:
I'd say they're not related, because the simple test I gave above doesn't involve unpaper at all.

zmwangx:
I would have investigated the issue, but I'm on some shitty Internet connection right now and can't even clone https://github.com/FFmpeg/FFmpeg.git... I'll see if there's a patch on master or report it when I can.

retokromer:
FYI: with your example `ffmpeg -y -i test.pbm out.png` does **not** work on the current HEAD.

zmwangx:
Yeah, I just tested on HEAD too, same thing. Assuming the code is in `libavcodec/iff.c`, there have been too many changes since 3.0.2.

zmwangx:
Ticket: https://trac.ffmpeg.org/ticket/5670.

ilovezfs:
thanks for reporting it upstream

zmwangx:
For now we can probably switch to a more common format in `unpaper`'s test to bypass this regression, e.g.,

```ruby
  test do
    system bin/"unpaper", test_fixtures("test.png"), "out.png"
    File.exist? testpath/"out.png"
  end
```

How prevalent is the PBM format? I'm not familiar with it.

ilovezfs:
@zmwangx regression was caused by https://github.com/FFmpeg/FFmpeg/commit/48ac4532d437790e56b8ed5d0d467dc88685f035

Here's a patch reverting it: https://gist.githubusercontent.com/ilovezfs/54b9d646671378424ad03869116abeab/raw/40043222489df998c60ffa374637b58f8dc2d22e/gistfile1.txt

zmwangx:
Official patch is in:

https://trac.ffmpeg.org/ticket/5670#comment:1
https://git.videolan.org/?p=ffmpeg.git;a=commitdiff;h=c5566f0a944e376b39c8f994659060ca036c441d

ilovezfs:
Excellent.

zmwangx:
I verified that the patch does work, and have patched the ffmpeg formula. Although I'm not sure if we should inline or vendor the patch instead.

ilovezfs:
you can probably use inreplace

zmwangx:
I inlined the patch. I don't think inreplace helps here, because we need to insert two lines.

ilovezfs:
With inreplace, you can insert multiple lines by using `\n` as necessary. Otherwise, this belongs in formula-patches, unfortunately. See discussion in #2298.

zmwangx:
I mean, using inreplace here would be rather ugly... I would rather use a patch.

> Otherwise, this belongs in formula-patches, unfortunately.

Got the `:DATA` point. But is it true that all patches have to be vendored now? I don't see it in the discussion (or missed it).

ilovezfs:
@zmwangx not, they can be links to patch files elsewhere, as long the urls are expected to be reasonably stable.

ilovezfs:
In this case you should use https://github.com/FFmpeg/FFmpeg/commit/c5566f0a944e376b39c8f994659060ca036c441d.patch

zmwangx:
I just reverted to the git.videolan.org patch. Should be equally good.

ilovezfs:
Yes that's fine.

Formula/ffmpeg.rb
```diff
@@ -1,10 +1,19 @@
...
   homepage "https://ffmpeg.org/"
-  url "https://ffmpeg.org/releases/ffmpeg-3.0.2.tar.bz2"
-  sha256 "30e3c77c2f4c358ed087869455a7496cbd7753a5e1b98d20ba49c1004009fd36"
   head "https://github.com/FFmpeg/FFmpeg.git"
 
+  stable do
+    url "https://ffmpeg.org/releases/ffmpeg-3.1.tar.bz2"
+    sha256 "2100fca81627e6cbe937fd6a071ae89277c02350538944b2b0c3c2cc71d9402a"
+
+    # See https://github.com/Homebrew/homebrew-core/pull/2416 and https://trac.ffmpeg.org/ticket/5670
```
ilovezfs: This goes passed the eightieth column so let's split it into two lines.

Formula/ffmpeg.rb
```diff
@@ -82,17 +91,18 @@ class Ffmpeg < Formula
...
-            "--enable-pthreads",
-            "--enable-gpl",
-            "--enable-version3",
-            "--enable-hardcoded-tables",
-            "--enable-avresample",
-            "--cc=#{ENV.cc}",
-            "--host-cflags=#{ENV.cflags}",
-            "--host-ldflags=#{ENV.ldflags}",
-           ]
+    args = [
```
ilovezfs: Let's use `%W[` instead

zmwangx:
@ilovezfs Done

ilovezfs:
@zmwangx Excellent, that'll queue up some bottles, thanks.

zmwangx:
Oops, accidentally extended the mpv commit... Repushed.

ilovezfs:
No problem. I killed the first build.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/8f9d220ad355a0774bca879ac01f950b81f6e9cb & https://github.com/Homebrew/homebrew-core/commit/ea704b262561d7ff0e9ac1c1370f72f52f0f9f15. Thank you once again @zmwangx 😺.

ened:
* ffmpeg 3.1.1 released
* currently fails to install due to removed `configure` option `--enable-dcadec`:
```
$ brew upgrade ffmpeg
ffmpeg 3.1
==> Upgrading ffmpeg
==> Downloading https://ffmpeg.org/releases/ffmpeg-3.1.tar.bz2
Already downloaded: /Users/ened/Library/Caches/Homebrew/ffmpeg-3.1.tar.bz2
==> Downloading https://git.videolan.org/?p=ffmpeg.git;a=patch;h=c5566f0a944e376b39c8f994659060ca036c441d
Already downloaded: /Users/ened/Library/Caches/Homebrew/ffmpeg--patch-408d11d8ae496ed4f83e36e1da0db4b373e8c016cb3723e029ba2fd25b1e7611.git;a=patch;h=c5566f0a944e376b39c8f994659060ca036c441d
==> Patching
==> Applying p=ffmpeg.git;a=patch;h=c5566f0a944e376b39c8f994659060ca036c441d
patching file libavcodec/pnm_parser.c
==> ./configure --prefix=/usr/local/Cellar/ffmpeg/3.1 --enable-shared --enable-pthreads --enable-gpl --enable-version3 --enable-hardcoded-tables --enable-avresample --cc=clang --host-cflags= --host-ldflags= --enable-opencl --enable-libx264
Last 15 lines from /Users/ened/Library/Logs/Homebrew/ffmpeg/01.configure:
--enable-libzmq
--enable-libbs2b
--enable-libdcadec
--enable-librubberband
--enable-libzimg
--enable-libopenh264
--disable-lzma
--enable-libopenjpeg
--disable-decoder=jpeg2000
--extra-cflags=-I/usr/local/Cellar/openjpeg/1.5.2_1/include/openjpeg-1.5
--enable-nonfree
--enable-vda
```

zmwangx:
@ened Please, open a new PR. Don't comment on a merged PR, especially when the comment is not really relevant (as in this PR is for FFmpeg 3.1 only, it doesn't cover 3.1.x).

ened:
The second point is relevant to this PR, the first one is not. 
Thank you for the bump & will do it later.

zmwangx:
Ah sorry, I thought the second point was for 3.1.1...

ilovezfs:
I guess we'll just need to deprecate that option. It might be worth checking if there have been any other changes in the option space.

@ened the fix, if you're just looking to get things working again locally, is to brew uninstall + brew install instead of brew upgrade, which attempts to preserve options, like brew reinstall.

ened:
@ilovezfs then I'd lose the previous compile options. `brew info` thankfully shows them, so I can remove the single option (`--with-dcadec`) and pass them all to `brew install`. :-) Thank you all.

zmwangx:
@ilovezfs 

> I guess we'll just need to deprecate that option.

I think `deprecated_option` only works for renames? That one has been removed in 3.1, see https://github.com/FFmpeg/FFmpeg/blob/n3.1/Changelog. Unfortunately I didn't notice.

There's also "libutvideo wrapper removed" but I don't think we have it to begin with.

ilovezfs:
Yeah, if it wasn't replaced with anything else, then it just needs to go away, which is itself a bit of an iffy practice if we ever started validating options, but I digress …

zmwangx:
Okay, I submitted #2553. @ened Thanks for the update.

PR #2417, opened 2016-06-27
minikube 0.3.0 (new formula)

ltiao:

---

Minikube is a CLI tool that provisions and manages single-node Kubernetes clusters optimized for development workflows.
This Homebrew Formula is equivalent to the official Google release installation instructions https://github.com/kubernetes/minikube/releases

### DIFF
--- a/Formula/minikube.rb
+++ b/Formula/minikube.rb
@@ -1,0 +1,17 @@
+class Minikube < Formula
+  desc "Tool that provisions and manages local single-node Kubernetes clusters"
+  homepage "https://github.com/kubernetes/minikube"
+  url "https://storage.googleapis.com/minikube/releases/v0.3.0/minikube-darwin-amd64"
+  version "0.3.0"
+  sha256 "cada05315d6643a3d8813b75c36b703773df739f32a576b25b8572a51ca90ee6"
+  head "https://github.com/kubernetes/minikube.git"
+
+  def install
+    mv "minikube-darwin-amd64", "minikube"
+    bin.install "minikube"
+  end
+
+  test do
+    assert_match /^minikube version: v0.3.0/, shell_output("#{bin}/minikube version 2>&1", 0)
+  end
+end


DomT4:
We'll want to build this from source. Homebrew (the formula side) doesn't accept upstream binaries unless they're something like a `.jar`.

Formula/minikube.rb
```diff
@@ -0,0 +1,17 @@
...
+  sha256 "cada05315d6643a3d8813b75c36b703773df739f32a576b25b8572a51ca90ee6"
+  head "https://github.com/kubernetes/minikube.git"
+
+  def install
+    mv "minikube-darwin-amd64", "minikube"
+    bin.install "minikube"
+  end
+
+  test do
+    assert_match /^minikube version: v0.3.0/, shell_output("#{bin}/minikube version 2>&1", 0)
```
DomT4:
```ruby
assert_match version.to_s, shell_output("#{bin}/minikube version 2>&1")
```

evilmarty:
FYI latest version is now 0.5.0

trinitronx:
@DomT4: As it currently appears, the build process for `minikube` has a build dependency of docker, probably provided by `docker-machine` + `virtualbox` on Mac.  That's one route, and provides a simplified bootstrap build environment with those 2 dependencies.  However, it's quite heavyweight as it requires a VM to start up with ~4GiB of RAM.  That process uses cross-compilation for `Mach-O x64`, however it's most likely possible to trim the build dependencies down to a version of golang.  Not sure what the standard for building `go` binaries in Homebrew is... but I figured I'd surface the issue for discussion.

DomT4:
If this is going to be a pain to build from source, particularly mandating things like Virtualbox, it's perhaps a better fit for https://github.com/Homebrew/homebrew-binary, where we will take a binary.

We'd prefer to build from source, but from source software has to be testable by our CI & mandating a Virtualbox installation to build isn't something we can cater for on CI at this point in time.

> Not sure what the standard for building go binaries in Homebrew is... but I figured I'd surface the issue for discussion.

In terms of the vendor system, which upstream seems to have adopted, [this](https://github.com/Homebrew/homebrew-core/blob/ceafc348d5fa4d7c776a223949d4dc7a6717688b/Formula/rancher-compose.rb) is fairly standard. For most `go` stuff we don't _tend_ to use the makefiles/build scripts/etc themselves because they tend to make software more difficult to package than it needs to be _(From our point of view, obviously, rather than upstream's who have different, entirely valid priorities)_.

Formula/minikube.rb
```diff
@@ -0,0 +1,17 @@
...
+  desc "Tool that provisions and manages local single-node Kubernetes clusters"
+  homepage "https://github.com/kubernetes/minikube"
+  url "https://storage.googleapis.com/minikube/releases/v0.3.0/minikube-darwin-amd64"
+  version "0.3.0"
+  sha256 "cada05315d6643a3d8813b75c36b703773df739f32a576b25b8572a51ca90ee6"
+  head "https://github.com/kubernetes/minikube.git"
+
+  def install
+    mv "minikube-darwin-amd64", "minikube"
+    bin.install "minikube"
```
MikeMcQuaid: Instead: `bin.install "minikube-darwin-amd64" => "minikube"`

MikeMcQuaid:
> If this is going to be a pain to build from source, particularly mandating things like Virtualbox, it's perhaps a better fit for https://github.com/Homebrew/homebrew-binary, where we will take a binary.

Agreed 👍

ltiao:
Since I can't even execute the build process manually to build from source (run out of memory), I'm gonna go ahead and open a PR at https://github.com/Homebrew/homebrew-binary as suggested. Also thanks to all for suggested changes.

tback:
Do you mind to look at https://github.com/Homebrew/homebrew-binary/pull/343?

PR #2425, opened 2016-06-27
unpaper: vendor libav instead of using ffmpeg

ilovezfs:
Upstream actually develops against libav not ffmpeg. Since enabling
shared libraries on libav would result in a conflict between ffmpeg
and libav, we can either vendor libav with unpaper or make libav
`keg_only`. This commit pursues the vendoring option but I certainly
wouldn't object to making libav `keg_only` instead if that's acceptable
to others.

The impetus for this change is an unaddressed (by either upstream)
incompatibility between ffmpeg 3.1 and unpaper 6.1. See https://github.com/Homebrew/homebrew-core/pull/2416.

### DIFF
--- a/Formula/unpaper.rb
+++ b/Formula/unpaper.rb
@@ -1,9 +1,17 @@
 class Unpaper < Formula
   desc "Post-processing for scanned/photocopied books"
   homepage "https://www.flameeyes.eu/projects/unpaper"
-  url "https://www.flameeyes.eu/files/unpaper-6.1.tar.xz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @zmwangx

ilovezfs:
And here's the alternative making libav `keg_only`: https://github.com/Homebrew/homebrew-core/pull/2429

zmwangx:
See https://github.com/Homebrew/homebrew-core/pull/2416#issuecomment-228731087.

PR #2434, opened 2016-06-27
ntopng 2.4

ilovezfs:
- use GitHub tarball for ntopng
- use GitHub 1.8-stable branch (four commits beyond the 1.8 tag) of the
  nDPI resource

### DIFF
--- a/Formula/ntopng.rb
+++ b/Formula/ntopng.rb
@@ -1,9 +1,17 @@
 class Ntopng < Formula
   desc "Next generation version of the original ntop"
   homepage "http://www.ntop.org/products/ntop/"
-  url "https://downloads.sourceforge.net/project/ntop/ntopng/ntopng-2.2.tar.gz"
-  sha256 "4fccfc9e9f333addcd3c957b4520c471117bc2df5655d6eabf328c7385fb255e"
-  revision 1
+
+  stable do
+    url "https://github.com/ntop/ntopng/archive/2.4.tar.gz"
+    sha256 "86f8ed46983f46bcd931304d3d992fc1af572b11e461ab9fb4f0f472429bd5dd"
+
+    resource "nDPI" do
+      # tip of 1.8-stable branch; four commits beyond the 1.8 tag
+      url "https://github.com/ntop/nDPI.git",
+        :revision => "6fb81f146e2542cfbf7fab7d53678339c7747b35"
+    end
+  end
 
   bottle do
     sha256 "21c5efd12b1c2871d076bed943c23a337fa626503d793c570894da8c9a85a464" => :el_capitan
@@ -39,12 +47,10 @@ class Ntopng < Formula
   depends_on "mariadb" => :optional
 
   def install
-    if build.head?
-      resource("nDPI").stage do
-        system "./autogen.sh"
-        system "make"
-        (buildpath/"nDPI").install Dir["*"]
-      end
+    resource("nDPI").stage do
+      system "./autogen.sh"
+      system "make"
+      (buildpath/"nDPI").install Dir["*"]
     end
     system "./autogen.sh"
     system "./configure", "--prefix=#{prefix}"


Formula/ntopng.rb
```diff
@@ -1,9 +1,18 @@
...
-  revision 1
+
+  stable do
+    url "https://github.com/ntop/ntopng/archive/2.4.tar.gz"
+    sha256 "86f8ed46983f46bcd931304d3d992fc1af572b11e461ab9fb4f0f472429bd5dd"
+
+    resource "nDPI" do
+      # tip of 1.8-stable branch; four commits beyond the 1.8 tag
+      url "https://github.com/ntop/nDPI.git",
+        :tag => "1.8-stable",
```
ilovezfs: Interestingly, even though this is actually the name of a branch not of a tag, brew will silently fail to verify the `:revision` if this is specified "accurately" with `:branch =>` instead of inaccurately with `:tag =>`.

ilovezfs: @UniqMartin I wonder if this should be considered a bug or not.

UniqMartin: Being able to specify a branch with `:tag =>` sounds like a bug and should probably be fixed [in `GitDownloadStrategy#checkout`](https://github.com/Homebrew/brew/blob/3c5f007bfa1a469ef3e1a048ee163f3eaa73cb71/Library/Homebrew/download_strategy.rb#L679-L682) by using a fully qualified reference like `"refs/heads/#{@ref}"` or `"refs/tags/#{@ref}"` instead of just `@ref`.

Using `:branch` is probably designed to actually track a moving target, thus checking the `:revision` in this case doesn't make much sense ([and thus isn't done](https://github.com/Homebrew/brew/blob/3c5f007bfa1a469ef3e1a048ee163f3eaa73cb71/Library/Homebrew/download_strategy.rb#L138-L145)). Printing a warning or even generating an error if `:revision` is specified nonetheless or simply always checking the revision if it is specified sound like possible solutions to me.

What do you think?

ilovezfs: All of that makes sense except that it seems to foreclose the possibility of using a branch plus revision for a stable or devel spec, which isn't unreasonable when upstreams have release branches to which they backport fixes, but don't necessarily create separate new tags. Of course, the equivalent can be mirrored on our end by using a sequences of patches, but I'm not sure there's a good philosophical justification for not supporting using a "release branch" for one of the specs as long as the revision is hard coded and upstream has designated it as the release foo branch corresponding to release tag foo.

UniqMartin: If one wants a fixed revision (at which point it no longer matters if it comes from a certain branch or not), one would just use `:revision =>` without any extra `:tag` or `:branch` specifiers.

(This is really based on the quite reasonable assumption that tags are expected to be fixed, thus are not expected to be moved to a different revision and it should be noticed if they do, but branches are a moving target, thus fixing them to a single revision hardly makes sense.)

ilovezfs: Do we have any cases where a stable or devel spec is specified as solely a revision? Does that still work even if the repository's default branch is not master?

UniqMartin: `colortail` and `x264` apparently, but I only checked the stable spec and only the primary resource, ignoring all resources specified via `resource do`:

```
brew for-each-formula 'puts "#{f.full_name}: #{f.stable.specs.inspect}" if f.stable && f.stable.specs.key?(:revision) && !f.stable.specs.key?(:tag)'
```

I can't think of a reason why this wouldn't work if the default branch isn't `master`, but I haven't actually investigated that. What problem do you foresee?

ilovezfs: Yep, seems to be no issue there.

So I think in this case, the resource should be
```
resource "nDPI" do
  # 1.8-stable branch
  url "https://github.com/ntop/nDPI.git",
  :revision => "6fb81f146e2542cfbf7fab7d53678339c7747b35"
end 
```

So I think
a) we should always confirm a tag is a tag
b) we should always confirm a branch is a branch
c) we should always rev-parse matches :revision
d) we should always raise an error when (a branch is specified) && (a revision is specified || a tag is specified). Or perhaps (equivalently?) raise an error when a branch is used outside of a head spec.

So, for example, this appears to be bogus: https://github.com/Homebrew/homebrew-games/blob/master/ufoai.rb#L4

ilovezfs:
PR refreshed.

Formula/ntopng.rb
```diff
@@ -1,9 +1,17 @@
...
-  revision 1
+
+  stable do
+    url "https://github.com/ntop/ntopng/archive/2.4.tar.gz"
+    sha256 "86f8ed46983f46bcd931304d3d992fc1af572b11e461ab9fb4f0f472429bd5dd"
+
+    resource "nDPI" do
+      # tip of 1.8-stable branch; four commits beyond the 1.8 tag
+      url "https://github.com/ntop/nDPI.git",
+      :revision => "6fb81f146e2542cfbf7fab7d53678339c7747b35"
```
UniqMartin: I think this should be either indented by two spaces or aligned with `"https://…`. But definitely not at the same level like `url` of which this is an argument. Sorry for the pedantry! 🙈

ilovezfs: Hmm... this indentation is the same as the way the go_resource blocks are currently all indented no?

Formula/ntopng.rb
UniqMartin: > So I think
> a) we should always confirm a tag is a tag
> b) we should always confirm a branch is a branch
> c) we should always confirm rev-parse matches :revision
> d) we should always raise an error when (a branch is specified) && (a revision is specified || a tag is specified). Or perhaps (equivalently?) raise an error when a branch is used outside of a head spec.

I agree with all of these.

> So, for example, this appears to be bogus: https://github.com/Homebrew/homebrew-games/blob/master/ufoai.rb#L4

Indeed, that's definitely a bogus case.

Formula/ntopng.rb
ilovezfs: Of course it's possible they're all wrong, and "the older style" ones not like that are the correct ones :)

UniqMartin: A quick search revealed at least three different styles for `go_resource`:

```ruby
url "URL", :revision => "REVISION"

url "URL",
  :revision => "REVISION"

url "URL",
:revision => "REVISION"
```

Haven't checked which one is the most popular, but I find the third to be the most aesthetically disturbing. (But since I have mostly shied away from dealing with Go-using formulae, I shall not be the authority for the preferred style.)

ilovezfs: The inconsistency with those has been driving me crazy. I think the third one may have become "popular" because gdm screws them up, and then that became the supposed standard, when in fact the second one is correct since it doesn't pretend the comma isn't there, and is also superior to the first because it makes a best effort to avoid going over the 80 character column limit, whereas the first doesn't unless the URL happens to be particularly short, in which case deciding between the first and second might require a neutral third party arbitrator.

UniqMartin: So the short-term fix would be:

1. Make sure `gdm` no longer screws them up.

And the long-term fix could be:

1. Agree on the preferred style.
2. Re-implement `brew audit --strict` as an extension for RuboCop and enable it by default.
  1. Implement a cop for alignment of the `url` method arguments.
  2. Or just re-enable the existing default cop for method argument alignment.
3. Mass-change all formulae by using RuboCop's auto-fix feature.

ilovezfs: Yes, that sounds right, except I think the mass change could be implemented immediately by replacing all `'^    :revision'` with `'^      :revision' :)

ilovezfs: Also humorous
```
Josephs-MacBook-Pro:homebrew joe$ grep -r -E '^    :tag' .|wc -l
      47
Josephs-MacBook-Pro:homebrew joe$ grep -r -E '^      :tag' .|wc -l
      44

```

UniqMartin: The problem both with the statistics and the mass replacement is that it doesn't distinguish between properly aligned `url` arguments at the formula top-level and misaligned arguments to an `url` method inside a `(stable|devel|head|resource) do` block …

ilovezfs: Correct, which is exactly why I mentioned the tag indentation problem. But it looks like the set of go violators is small enough to handle manually.

PR #2446, opened 2016-06-28
progress 0.13

ilovezfs:
/dev/null can no longer be used for the test's output file due to
upstream commit Xfennec/progress@a6f4358 so instead use a real file.

### DIFF
--- a/Formula/progress.rb
+++ b/Formula/progress.rb
@@ -1,8 +1,8 @@
 class Progress < Formula
   desc "Progress: Coreutils Progress Viewer"
   homepage "https://github.com/Xfennec/progress"
-  url "https://github.com/Xfennec/progress/archive/v0.9.tar.gz"
-  sha256 "63e1834ec114ccc1de3d11722131b5975e475bfd72711d457e21ddd7fd16b6bd"
+  url "https://github.com/Xfennec/progress/archive/v0.13.tar.gz"
+  sha256 "160cb6156a0b8df32a3944f3dcecba956ae3e5579e91d53c9d7417bc4956718c"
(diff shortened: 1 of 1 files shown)


Formula/progress.rb
```diff
@@ -18,14 +18,15 @@ def install
...
+      system "/bin/dd", "if=/dev/urandom", "of=foo", "bs=512", "count=1048576"
     end
     sleep 1
     begin
-      assert_match(/dd/, shell_output("#{bin}/progress"))
+      assert_match "dd", shell_output("#{bin}/progress")
     ensure
       Process.kill 9, pid
       Process.wait pid
+      rm "foo"
```
DomT4: Where's this coming from?

ilovezfs: It comes from
```
system "/bin/dd", "if=/dev/urandom", "of=foo", "bs=512", "count=1048576"
```
`of` means "output file." In this case, `foo` is a 512M file, or at least however much of 512M can be written during the 1 second before the process is killed in `ensure`.

DomT4: Oh, of course. Who do you think wrote [this blob](https://github.com/Homebrew/homebrew-core/blob/edf98fecc606cb4cc360ac4dad8beb76f7115e9d/Formula/stunnel.rb#L33-L37) 😉. Is what I get for reviewing PRs that late at night. Seems fine.

ilovezfs: Frankly it's somewhat disturbing that upstream managed to rewrite that code such that `/dev/null` isn't sufficiently "real," but I digress …

PR #2450, opened 2016-06-28
thefuck 3.10

ilovezfs:
- fix the test as output to non-tty is now unadorned text
- style tweaks in the test

### DIFF
--- a/Formula/thefuck.rb
+++ b/Formula/thefuck.rb
@@ -1,10 +1,18 @@
 class Thefuck < Formula
   desc "Programatically correct mistyped console commands"
   homepage "https://github.com/nvbn/thefuck"
-  url "https://pypi.python.org/packages/e3/59/eba3a2b16a9a12869edb7858dbe5e84f3734b8fb58b9c6921f9720c01c8b/thefuck-3.9.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
👍 when 🍏

ilovezfs:
@DomT4 it will be soon thanks to @tdsmith https://github.com/nvbn/thefuck/pull/525

DomT4:
Looks like there's an additional issue 😢
```
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/thefuck/3.10/libexec --single-version-externally-managed --record=installed.txt
error in thefuck setup command: Invalid environment marker: python_version<"3.4"
```

ilovezfs:
Probably the `':python_version<"3.4"'` should be `":python_version<'3.4'"`

ilovezfs:
Nope, it's something else.

ilovezfs:
@DomT4 PR refreshed. :green_apple: 

It requires a newish version of setuptools. Same issue as https://github.com/docker/docker-py/issues/1019.

DomT4:
🎉 Thanks ILZ!

ilovezfs:
*ilovezfs wags his tail*

PR #2471, opened 2016-06-28
cloudbreak-shell 1.4.0-rc.15 (new formula)

seanorama:

---

### DIFF
--- a/Formula/cloudbreak-shell.rb
+++ b/Formula/cloudbreak-shell.rb
@@ -1,0 +1,33 @@
+class CloudbreakShell < Formula
+  desc "CLI shell for the Cloudbreak project"
+  homepage "https://github.com/sequenceiq/cloudbreak/tree/master/shell"
+  url "https://s3-eu-west-1.amazonaws.com/maven.sequenceiq.com/releases/com/sequenceiq/cloudbreak-shell/1.4.0-rc.15/cloudbreak-shell-1.4.0-rc.15.jar"
+  version "1.4.0-rc.15"
+  sha256 "79b00c1744f26346f5a01417e984ce87582b64e741377589d4cf2f66e5f3d548"
+
+  bottle :unneeded
+
+  depends_on :java => "1.8+"
+
+  def bin_wrapper; <<-EOS.undent
+    #!/bin/sh
+    exec java -jar #{libexec}/cloudbreak-shell.jar "$@"
+    EOS
+  end
+
+  def install
+    libexec.install "cloudbreak-shell-#{version}.jar" => "cloudbreak-shell.jar"
+    (bin/"cloudbreak-shell").write(bin_wrapper)
+  end
+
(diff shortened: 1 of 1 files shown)


Formula/cloudbreak-shell.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def bin_wrapper; <<-EOS.undent
+    #!/bin/sh
+    exec java -jar #{libexec}/cloudbreak-shell.jar "$@"
+    EOS
+  end
+
+  def install
+    libexec.install "cloudbreak-shell-#{version}.jar" => "cloudbreak-shell.jar"
+    (bin/"cloudbreak-shell").write(bin_wrapper)
```
DomT4: You can just use:
```ruby
bin.write_jar_script libexec/"cloudbreak-shell.jar", "cloudbreak-shell"
```
here.

Formula/cloudbreak-shell.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def caveats; <<-EOS.undent
```
DomT4: This will be enforced by the `:java` dependency. Can remove these caveats.

Formula/cloudbreak-shell.rb
```diff
@@ -0,0 +1,33 @@
...
+    CloudBreak Shell requires the Java Runtime Engine (JRE) version 8.x or
+    newer; it will not run on older JRE versions.
+    EOS
+  end
+
+  test do
+    File.exist? "#{bin}/cloudbreak-shell"
```
DomT4: Can this test be modified to do something more substantial? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

DomT4:
We'll need a stable release here before we can consider this, FYI.

Formula/cloudbreak-shell.rb
seanorama: Fixed. Thanks.

Formula/cloudbreak-shell.rb
seanorama: Fixed. Thanks.

Formula/cloudbreak-shell.rb
seanorama: Changed to check that application usage is printed. Can't test any more than that without a server & credentials for the client to connect to.

seanorama:
@DomT4 Are you referring to the build name having "rc.15"?

This project seems to only ever be in rc. But it's what they use for stable environments.

DomT4:
Yeah, presuming they've stuck to the standard definition of `rc` in this context.

Have they placed an explanation this is a stable release somewhere handy? It seems like there's a new release every day, which is, er, rapid for stable.

MikeMcQuaid:
Passing on this due to reasons above.

PR #2477, opened 2016-06-29
Updated ffmpeg.rb with new url and sha256 for FFmpeg 3.1 stable

bbgdzxng1:


ilovezfs:
Thanks for taking the time to put together this PR. This is actually already a work in progress here: https://github.com/Homebrew/homebrew-core/pull/2416

PR #2498, opened 2016-06-29
coq: fix emacs caveat

hrldcpr:

---

The emacs major mode bundled with Coq is called `gallina.el`, not `coq.el` file, so the current caveat doesn't work.

From the [Coq Reference Manual](https://coq.inria.fr/refman/Reference-Manual017.html) section 15.6.1:

> Coq comes with a Major mode for GNU Emacs, `gallina.el`.
> Add the following lines to your .emacs file:
> 
> ```
> (setq auto-mode-alist (cons '("\\.v$" . coq-mode) auto-mode-alist))
> (autoload 'coq-mode "gallina" "Major mode for editing Coq vernacular." t)
> ```

### DIFF
--- a/Formula/coq.rb
+++ b/Formula/coq.rb
@@ -40,7 +40,7 @@ class Coq < Formula
                           "-coqide", "no",
                           "-with-doc", "no"
     system "make", "world"
-    system "make", "install"
+    ENV.deparallelize { system "make", "install" }
   end
 
   test do
--- a/Formula/coq.rb
+++ b/Formula/coq.rb
@@ -43,13 +43,6 @@ class Coq < Formula
     system "make", "install"
   end
 
-  def caveats; <<-EOS.undent
-    To use the Coq Emacs mode, add the following to your init file:
-      (setq auto-mode-alist (cons '("\\\\.v$" . coq-mode) auto-mode-alist))
-      (autoload 'coq-mode "coq" "Major mode for editing Coq vernacular." t)
-    EOS
-  end
-
   test do
     (testpath/"testing.v").write <<-EOS.undent
       Inductive nat : Set :=


dunn:
I wonder if the entire caveats section should be removed; it's not Homebrew-specific information, so there's no point replicating upstream docs (and getting them wrong).

hrldcpr:
Good question. I personally found it useful, since the files only end up in that location because of the specific `--emacslib` option passed to the build by this formula.

dunn:
The caveats message doesn't tell you where it's been installed, does it?

hrldcpr:
Yeah you're right, that message (which for the record is `Emacs Lisp files have been installed to: /usr/local/share/emacs/site-lisp/coq`) must come from homebrew.

I don't have much an opinion either way—no caveats, or corrected caveats, both are probably better than the current incorrect situation.

dunn:
Ah, yeah; we print that message when elisp is installed in `share/formula_name`.

Let's go ahead and remove the caveats, thanks!

hrldcpr:
Sounds good, removed.

ilovezfs:
That build failure smells like a parallelization problem.

hrldcpr:
Yeah weird… I've just pushed again in hopes that will trigger the CI again, we shall see.

hrldcpr:
Weird, the CI failed again… It definitely works on my machine, plus it's hard to see how removing the caveats could cause a build failure. I'll leave it to you experts.

ilovezfs:
Well, it was 10.11 not 10.10 so that pretty much confirms it's a parallelization issue :)

Please add a commit making it
```
ENV.deparallelize { system "make", "install" }
```

hrldcpr:
Thanks for looking into it, committed.

dunn:
Thanks! Merged as 13c57c3651489007e9e72cd7b0f0c156d0e0d605 and a12ab38e246ef5c6c26d0c8198d9349ad08656f0.

PR #2524, opened 2016-06-30
jdep 1.3 (new formula)

pdex:

---

Jdep is a tool for analyzing Java .class file dependencies, so that the
peculiar compilation behavior of many Java compilers can be tamed to be
compatible with conventional Unix style Makefiles.

### DIFF
--- a/Formula/jdep.rb
+++ b/Formula/jdep.rb
@@ -32,4 +32,41 @@ index b415cb5..15573d6 100644
  #include <string.h>
  #include <strings.h>
  #include <dirent.h>
+diff --git a/jdep.c b/jdep.c
+index 15573d6..8555936 100644
+--- a/jdep.c
++++ b/jdep.c
+@@ -67,8 +67,11 @@ PackageInfo *IncludedPackages = NULL;
+ #define CONSTANT_Float                   4
+ #define CONSTANT_Integer                 3
+ #define CONSTANT_InterfaceMethodref     11
++#define CONSTANT_InvokeDynamic          18
+ #define CONSTANT_Long                    5
+ #define CONSTANT_Methodref              10
++#define CONSTANT_MethodHandle           15
++#define CONSTANT_MethodType             16
+ #define CONSTANT_NameAndType            12
+ #define CONSTANT_String                  8
+ #define CONSTANT_Utf8                    1
+@@ -670,6 +673,20 @@ readConstantPoolInfo(FILE *fyle, char *filename)
+             char *str = (char *) readByteArray(fyle, length);
+             return (cp_info *) build_constant_utf8_info(str);
+         }
++        case CONSTANT_MethodHandle:{
++            readByte(fyle); /* reference_kind */
++            readWord(fyle); /* reference_index */
++            return NULL;
++        }
++        case CONSTANT_MethodType:{
++            readWord(fyle); /* descriptor_index */
++            return NULL;
++        }
++        case CONSTANT_InvokeDynamic:{
++            readWord(fyle); /* bootstrap_method_attr_index */
++            readWord(fyle); /* name_and_type_index */
++            return NULL;
++        }
+         default:
+             fprintf(stderr, "invalid constant pool tag %d in %s\n", tag,
+                     filename);
 
--- a/Formula/jdep.rb
+++ b/Formula/jdep.rb
@@ -69,4 +69,36 @@ index 15573d6..8555936 100644
          default:
              fprintf(stderr, "invalid constant pool tag %d in %s\n", tag,
                      filename);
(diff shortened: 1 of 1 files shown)


Formula/jdep.rb
```diff
@@ -0,0 +1,35 @@
...
+  url "http://www.fudco.com/software/jdep-1.3.tar.gz"
+  sha256 "3dd925ae2523cca59e4aaa7cbb8daa28927b40520f0f8dccead91d9edea1dc65"
+
+  patch :DATA
+
+  def install
+    system "make"
+    bin.install "bin/jdep"
+    bin.install "bin/touchp"
+    prefix.install "example"
```
tdsmith: It would make sense to put this in `pkgshare` instead of `prefix`.

tdsmith:
Thanks for the PR! This looks like a useful tool and the formula looks complete and correct. We'd prefer not to accept new software that requires patches to build; can you get in touch with the maintainer and ask them to cut a new release that includes the right headers on OS X? (I know the software hasn't been updated in a while; this is partly a test to see if they're still willing to maintain it at all.)

If that doesn't prove fruitful, one way you could use Homebrew to install this software is to [maintain your own tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap). The linked document has some advice; please let us know if you have any questions and we'd be happy to help you set things up.

DomT4:
Any update on Tim's comments?

pdex:
Holiday week, I'll get to this next week.

pdex:
Chip (the upstream author) is going to setup a github repo for jdep. Once that's done I'll submit the patches to him as a PR.

pdex:
patches submitted: https://github.com/FUDCo/jdep/pull/1

Formula/jdep.rb
```diff
@@ -0,0 +1,104 @@
+class Jdep < Formula
+  desc "Analyze Java .class files and produce Makefile dependencies."
+  homepage "http://www.fudco.com/software/jdep.html"
+  url "http://www.fudco.com/software/jdep-1.3.tar.gz"
+  sha256 "3dd925ae2523cca59e4aaa7cbb8daa28927b40520f0f8dccead91d9edea1dc65"
+
+  patch :DATA
```
dunn: Could you replace this with a patch do block, along with a comment linking to the upstream PR?
```ruby
patch do
  url "https://github.com/FUDCo/jdep/commit/452bd87c69a1c0f03f381b32e995b0b9a813fc25.diff"
  sha256 "55c7be39f34ed974e4ed2213626ded6205df07dc9e8099e2f49f63118540a993"
end
```
(Inline patches can get mangled easily.)

Formula/jdep.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def install
+    system "make"
```
DomT4: Is there no `make install` at all?

MikeMcQuaid: Please request one from in an upstream issue if there's not one.

Formula/jdep.rb
MikeMcQuaid: Also we may not accept this until upstream confirms its merged.

pdex: @dunn will do

@MikeMcQuaid upstream merged 10 days before your comment:
https://github.com/FUDCo/jdep/pull/1

Formula/jdep.rb
pdex: @DomT4 @MikeMcQuaid Do you want it to function like automake/configure and install to ${PREFIX} ?

You guys should really update the new formula cookbook with all of these requirements.

MikeMcQuaid: > @DomT4 @MikeMcQuaid Do you want it to function like automake/configure and install to ${PREFIX} ?

Yes.

> You guys should really update the new formula cookbook with all of these requirements.

We cannot update with every corner case as, as in this case, it's a conversation rather than a hard rule.

DomT4: We'll accept formulae without a `make install`, but it's not particularly difficult to write it into a `Makefile` that already exists & makes life easier for users even outside of package management, so is a "nice thing to have" when upstreams are willing.

MikeMcQuaid:
Passing on this until we have a release that doesn't require patching and an upstream report requesting a `make install` (ideally).

pdex:
Sorry I didn't have time to go back and forth on this, I would really recommend that maybe you update the wiki to suggest people create taps instead of submitting PRs to core. That way you could groom someone's tap to a core recipe.

MikeMcQuaid:
No worries, thanks @pdex.

PR #2537, opened 2016-07-01
etcd 3.0.1

ilovezfs:


DomT4:
Test needs to be using the full path to the executable, but otherwise 👍.

PR #2541, opened 2016-07-01
cryptopp: add CXXFLAGS argument to `make` enabling C++11

fish2000:
`ENV.cxx11` does not quite cut it in this case – this change builds CryptoPP in C++11 mode correctly

### DIFF
--- a/Formula/cryptopp.rb
+++ b/Formula/cryptopp.rb
@@ -13,8 +13,12 @@ class Cryptopp < Formula
   option :cxx11
 
   def install
-    ENV.cxx11 if build.cxx11?
-    system "make", "CXX=#{ENV.cxx}"
+    if build.cxx11?
+      ENV.cxx11
+      system "make", "CXX=#{ENV.cxx}", "CXXFLAGS=-std=c++11"
+    else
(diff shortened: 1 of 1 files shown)


tdsmith:
Can you describe what happened when this wasn't provided?

fish2000:
When building with the current formula, as written, CryptoCPP does not build in C++11 mode.

Here are my logs from building with the current formula: https://gist.github.com/fish2000/4e59b0bff9df1dc837ba3d947a74cade

… The PR I submitted adds the make argument as specified in the CryptoCPP distributions’ `Install.txt` file (q.v. https://github.com/weidai11/cryptopp/blob/master/Install.txt#L50)

tdsmith:
It does, actually! Superenv adds that argument silently; you can see that in the `.cc` files in ~/Library/Homebrew/Logs/cryptopp. Are you running into any trouble otherwise?

fish2000:
Oh my good lord, you are correct – I had conflated a local link-time error with the issue as reported, based on a topical log grep. I reran the build commands in a cleanroom-ish environment, and so yeah: I jumped the gun with the PR. My bad. Thanks for pointing me in the right direction, yes.

PR #2547, opened 2016-07-01
minio 20160603193205 (new formula)

nl5887:

---

### DIFF
--- a/Formula/minio.rb
+++ b/Formula/minio.rb
@@ -8,7 +8,7 @@ require "language/go"
 class Minio < Formula
   desc "object storage server compatible with Amazon S3"
   homepage "https://github.com/minio/minio"
-  url "https://github.com/minio/minio/archive/RELEASE.2016-06-03T19-32-05Z.tar.gz", :using => :nounzip
+  url "https://github.com/minio/minio/archive/RELEASE.2016-06-03T19-32-05Z.tar.gz"
   version "20160603193205"
   sha256 "8ecf52cc344c99f84bef632b8dfecf8fdc6a8493a541b2fe42f5b22d8bbfa802"
   head "https://github.com/minio/minio.git"
@@ -18,9 +18,10 @@ class Minio < Formula
   def install
     ENV["GOPATH"] = buildpath
 
-    mkdir_p buildpath/"src/github.com/minio/minio/"
+    clipath = buildpath/"src/github.com/minio/minio"
+    clipath.install Dir["*"]
+
     cd buildpath/"src/github.com/minio/minio/" do
-      safe_system "tar", "vxf", buildpath/"RELEASE.2016-06-03T19-32-05Z.tar.gz", "--strip-components=1"
       system "go", "build", "-o", buildpath/"minio"
     end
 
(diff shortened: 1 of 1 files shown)


Formula/minio.rb
```diff
@@ -0,0 +1,33 @@
...
+  head "https://github.com/minio/minio.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    mkdir_p buildpath/"src/github.com/minio/minio/"
+    cd buildpath/"src/github.com/minio/minio/" do
+      safe_system "tar", "vxf", buildpath/"RELEASE.2016-06-03T19-32-05Z.tar.gz", "--strip-components=1"
```
dunn: Why not just `ln -s` it?  Here's how we do it for Hugo (and a lot of other Go formulae): https://github.com/Homebrew/homebrew-core/blob/master/Formula/hugo.rb#L200

nl5887: I had that before, but Fish shell was causing issues, referring back to the original folder. This works with Hugo because it has main.go in the root. Minio needs to be compiled in the `buildpath/"src/github.com/minio/minio/`itself.

dunn: Maybe `mv Dir["buildpath/*"], buildpath/"src/github.com/minio/minio/"`?

nl5887: Moving failed and is still failing with me, I assume it is failing because it is moving its destination folder (/src/github.com/minio/minio) as well to its destination, causing an error. These issues where the reason to choose the approach I used.

nl5887: Any other suggestions how to solve this? Or will my approach do as well?

dunn: Do all formula using the `ln_s` trick fail in Fish?  If so that's a larger problem we need to address.

nl5887: I have to do some testing, but the issue is that fish will resolve to the linked to folder. So when you ln -s buildpath target, changing directory to target will resolve to buildpath again. More info about this issue can be found here: https://github.com/fish-shell/fish-shell/issues/1957.

nl5887:
I've found an other solution while checking the other go formulas. Seems that the combination ln_s, cd and fish are failing only in specific cases for go: 

* when having a vendored package (./vendor/) in buildpath
* when having all source in the buildpath root

The other formulas I checked don't have this issue and seems to install fine using fish. But every formula depending to cd into the linked target instead of the original source has the probability to fail.

Formula/minio.rb
```diff
@@ -0,0 +1,34 @@
+require "language/go"
+
+# we're using an ugly workaround here, because the tar is being extracted in builddir,
+# and we need to build into the $GOPATH/src/github.com/minio/minio/. Softlinking isn't working
+# with every shell. So we're using the go_resource to go get the correct revision, while
```
dunn: Does this comment still apply?

nl5887: you're right, comment has been removed.

MikeMcQuaid:

PR #2548, opened 2016-07-01
libosinfo: formula tweaks

ilovezfs:
- sort build dependencies to the top
- add linked library dependency gettext
- make gobject-introspection recommended instead of optional since it will be installed anyway
- remove the patch for "ld: unknown option" since it's been upstreamed
- fix indentation of the heredoc in the test

### DIFF
--- a/Formula/libosinfo.rb
+++ b/Formula/libosinfo.rb
@@ -10,22 +10,19 @@ class Libosinfo < Formula
     sha256 "f39179801faa67e260672bcaa07406bad710283894d50e26838bb57c5a58b12a" => :mavericks
   end
 
-  depends_on "check"
-  depends_on "glib"
   depends_on "intltool" => :build
-  depends_on "libsoup"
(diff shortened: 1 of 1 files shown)


DomT4:
Seems reasonable.

ilovezfs:
CC @nijikon RE: #2688

nijikon:
Based on the `gobject-introspection` change I think that this needs a revision bump.

ilovezfs:
@nijikon Are you sure? The primary reason for the change is that configure treats it as opt-out not opt-in if gobject-instrospection is installed during the build, which it is:
```
checking for gobject-introspection... yes
```

ilovezfs:
In other words, I think the bottles you shipped actually have introspection turned on, despite the fact that it's supposedly optional not recommended.

nijikon:
@ilovezfs interesting, let's merge this 👍

ilovezfs:
okie dokie

Issue #2517, opened 2016-06-30
Question: Correct way to add ArangoDB 3.0?


fceller:
The new major version of ArangoDB 3.0 is not (datafile) compatible with ArangoDB 2.8. The correct way to upgrade is, to dump the data using ArangoDB 2.8 and then import the data into 3.0. It also requires some changes to the install Foxx apps.

How should the new ArangoDB formula look like? 

(a) create a new "arangodb3" formula and keep the old "arangodb"
(b) rename the "arangodb" formula to "arangodb2" and create a "arangodb" for ArangoDB 3.0
(c) something else

Thanks for any hints

ilovezfs:
@fceller I'm happy to defer to you on this and will close the PR, but this is how I started: https://github.com/Homebrew/homebrew-core/pull/2515

There does seem to be a Mavericks build failure.

ilovezfs:
@fceller Here's the full log for 10.9: https://gist.githubusercontent.com/ilovezfs/4a63c98e41eb54364295a162f4322f44/raw/97b2af849e0fb198ff8624d4cbb56c4fbbab353f/gistfile1.txt

dothebart:
hm, ```
n file included from /tmp/arangodb-20160630-46463-37lwyp/ArangoDB-3.0.0/3rdParty/rocksdb/rocksdb/db/compaction_job.cc:46:0:
/tmp/arangodb-20160630-46463-37lwyp/ArangoDB-3.0.0/3rdParty/rocksdb/rocksdb/table/block.h:14:20: fatal error: malloc.h: No such file or directory
 #include <malloc.h>
                    ^
compilation terminated.
```
Strange, that doesn't happen during regular compiles on macs over here...

ilovezfs:
It's specifically OS X 10.9

dothebart:
whats it supposed to do? is an include path missing? or is the filename wrong?

fceller:
Anyone any thoughts on the original question? If not, I will try to 

- fix the issue @ilovezfs has encountered
- try to push the old formula to homebrew-versions

Is this sensible? @DomT4?

ilovezfs:
[...]

PR #2576, opened 2016-07-02
arangodb 3.0.2

fceller:

---

Following the discussion at https://github.com/Homebrew/homebrew-core/issues/2517, I have renamed the formula and create an version for ArangoDB 2 at https://github.com/Homebrew/homebrew-versions/pull/1304

### DIFF
--- a/Formula/arangodb.rb
+++ b/Formula/arangodb.rb
@@ -1,8 +1,8 @@
 class Arangodb < Formula
   desc "Universal open-source database with a flexible data model"
   homepage "https://www.arangodb.com/"
-  url "https://www.arangodb.com/repositories/Source/ArangoDB-2.8.9.tar.gz"
-  sha256 "cff21ca654056bed08781c5e462966f5f15acec7b6522191d286dee3339e327e"
+  url "https://www.arangodb.com/repositories/Source/ArangoDB-2.8.10.tar.gz"
+  sha256 "3f2adba7bb4fe5a93959c6929b984b053e66d3653144ed6bb1cdedd017c65050"
 
   bottle do
     revision 1
--- a/Formula/arangodb.rb
+++ b/Formula/arangodb.rb
@@ -1,8 +1,10 @@
 class Arangodb < Formula
-  desc "Universal open-source database with a flexible data model"
+  desc "The Multi-Model NoSQL Database."
   homepage "https://www.arangodb.com/"
-  url "https://www.arangodb.com/repositories/Source/ArangoDB-2.8.10.tar.gz"
-  sha256 "3f2adba7bb4fe5a93959c6929b984b053e66d3653144ed6bb1cdedd017c65050"
+  url "https://www.arangodb.com/repositories/Source/ArangoDB-3.0.2.tar.gz"
+  sha256 "04c00d58d7e63137ccb7d0a73112aa01e34e08c98b3c82c62ef7987f0d214ac2"
+
+  head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
 
   bottle do
     revision 1
@@ -11,11 +13,7 @@ class Arangodb < Formula
     sha256 "60d59e09f4a440cffeb077f1de14784519223e6f25ccaaf6df692dc63fd6cac8" => :mavericks
   end
 
-  head do
-    url "https://github.com/arangodb/arangodb.git", :branch => "unstable"
-    depends_on "cmake" => :build
-  end
-
+  depends_on "cmake" => :build
   depends_on "go" => :build
   depends_on "openssl"
 
@@ -26,66 +24,82 @@ class Arangodb < Formula
     cause "Fails with compile errors"
   end
 
+  resource "arangodb2" do
+    url "https://www.arangodb.com/repositories/Source/ArangoDB-2.8.10.tar.gz"
+    sha256 "3a455e9d6093739660ad79bd3369652db79f3dabd9ae02faca1b014c9aa220f4"
+  end
+
+  resource "upgrade" do
+    url "https://www.arangodb.com/repositories/Source/upgrade3-1.0.0.tar.gz"
+    sha256 "965f899685e420530bb3c68ada903c815ebd0aa55e477d6949abba9506574011"
+  end
+
   def install
-    ENV.libcxx
+    ENV.cxx11
 
-    if build.head?
-      mkdir "arangodb-build" do
-        system "cmake", "..", "-DHOMEBREW=On", "-DUSE_OPTIMIZE_FOR_ARCHITECTURE=Off", "-DASM_OPTIMIZATIONS=Off", "-DETCDIR=#{prefix}/etc", "-DVARDIR=#{var}", *std_cmake_args
-        system "make", "install"
-      end
+    (libexec/"arangodb2/bin").install resource("upgrade")
 
-    else
-      # clang on 10.8 will still try to build against libstdc++,
-      # which fails because it doesn't have the C++0x features
-      # arangodb requires.
+    resource("arangodb2").stage do
+      ENV.cxx11
+
       args = %W[
         --disable-dependency-tracking
-        --prefix=#{prefix}
+        --prefix=#{libexec}/arangodb2
         --disable-relative
-        --datadir=#{share}
         --localstatedir=#{var}
+        --program-suffix=-2.8
       ]
 
-      args << "--program-suffix=-unstable" if build.head?
-
-      if ENV.compiler != :clang
-        ENV.append "LDFLAGS", "-static-libgcc -static-libstdc++"
+      if ENV.compiler == "gcc-6"
+        ENV.append "CXXFLAGS", "-O2 -g -fno-delete-null-pointer-checks"
+        inreplace "3rdParty/Makefile.v8", "CXXFLAGS=\"", "CXXFLAGS=\"-fno-delete-null-pointer-checks "
       end
 
       system "./configure", *args
       system "make", "install"
     end
+
+    mkdir "build" do
+      args = std_cmake_args + %W[
+        -DHOMEBREW=ON
+        -DUSE_OPTIMIZE_FOR_ARCHITECTURE=OFF
+        -DASM_OPTIMIZATIONS=OFF
+        -DETCDIR=#{etc}
+        -DVARDIR=#{var}
+      ]
+
+      if ENV.compiler == "gcc-6"
+        ENV.append "V8_CXXFLAGS", "-O3 -g -fno-delete-null-pointer-checks"
+      end
+
+      system "cmake", "..", *args
+      system "make", "V=1", "Verbose=1", "VERBOSE=1", "install"
+    end
   end
 
-  # moving the "if" inside post_install does not work
-  if build.head?
-    def post_install
-      (var/"lib/arangodb3").mkpath
-      (var/"log/arangodb3").mkpath
-    end
-  else
-    def post_install
-      (var/"arangodb").mkpath
-      (var/"log/arangodb").mkpath
+  def post_install
+    (var/"lib/arangodb3").mkpath
+    (var/"log/arangodb3").mkpath
 
-      system "#{sbin}/arangod", "--upgrade"
-    end
+    args = %W[
+      #{libexec}/arangodb2
+      #{var}/lib/arangodb
+      #{opt_prefix}
+      #{var}/lib/arangodb3
+    ]
+
+    system libexec/"arangodb2/bin/upgrade.sh", *args
   end
 
   def caveats
     s = <<-EOS.undent
-      Please note that clang and/or its standard library 7.0.0 has a severe
-      performance issue. Please consider using '--cc=gcc-5' when installing
-      if you are running on such a system.
-    EOS
+      The database format between ArangoDB 2.x and ArangoDB 3.x has
+      been changed, please checkout
+      https://docs.arangodb.com/3.0/Manual/Administration/Upgrading/index.html
 
-    if build.head?
-      s += <<-EOS.undent
-        A default password has been set. You can change it by executing
-          #{sbin}/arango-secure-installation
-      EOS
-    end
+      An empty password has been set. Please change it by executing
+        #{opt_sbin}/arango-secure-installation
+    EOS
 
     s
   end
@@ -101,12 +115,8 @@ class Arangodb < Formula
         <true/>
         <key>Label</key>
         <string>#{plist_name}</string>
-        <key>ProgramArguments</key>
-        <array>
-          <string>#{opt_sbin}/arangod</string>
-          <string>-c</string>
-          <string>#{etc}/arangodb/arangod.conf</string>
-        </array>
+        <key>Program</key>
+        <string>#{opt_sbin}/arangod</string>
         <key>RunAtLoad</key>
         <true/>
       </dict>
@@ -115,6 +125,8 @@ class Arangodb < Formula
   end
 
   test do
-    assert_equal "it works!\n", shell_output("#{bin}/arangosh --javascript.execute-string \"require('@arangodb').print('it works!')\"")
+    testcase = "require('@arangodb').print('it works!')"
+    output = shell_output("#{bin}/arangosh --server.password \"\" --javascript.execute-string \"#{testcase}\"")
+    assert_equal "it works!\n", output
   end
 end


Formula/arangodb3.rb
```diff
@@ -0,0 +1,85 @@
...
+    revision 1
+    sha256 "f9699b28c53f744b7b4701890739d7a9369a6e0b144d486d7fee27ed4b4a9486" => :el_capitan
+    sha256 "7490e2e4026fca6442fce8c6f56d654deb45ba864ca6b7a73872d670b4475d49" => :yosemite
+    sha256 "60d59e09f4a440cffeb077f1de14784519223e6f25ccaaf6df692dc63fd6cac8" => :mavericks
+  end
+
+  head do
+    url "https://github.com/arangodb/arangodb.git", :branch => "unstable"
+    depends_on "cmake" => :build
+  end
```
ilovezfs: The depends_on "cmake" does not need to be limited to any particular spec. So this simplifies to
```
head do
  url "https://github.com/arangodb/arangodb.git", :branch => "unstable"
end
```
But when the only element in a head block is the url, it should be simplified to the non-block form
```
head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
```
Audit will require that to be placed below the checksum like this:
```
sha256 "f2e9a44bbd73608cace6a3bfe357d2e58aed1eb8b0a4675c36c6a9b15620576a"
head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
```

Formula/arangodb3.rb
```diff
@@ -0,0 +1,85 @@
...
+  def install
+    ENV.libcxx
+
+    mkdir "arangodb-build" do
+      system "cmake", "..", "-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DHOMEBREW=On", "-DUSE_OPTIMIZE_FOR_ARCHITECTURE=Off", "-DASM_OPTIMIZATIONS=Off", "-DETCDIR=#{prefix}/etc", "-DVARDIR=#{var}", *std_cmake_args
+      system "make", "install"
+    end
+  end
+
+  # moving the "if" inside post_install does not work
```
ilovezfs: This comment should be removed.

Formula/arangodb3.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  fails_with :clang do
+    build 600
+    cause "Fails with compile errors"
+  end
+
+  def install
+    ENV.libcxx
+
+    mkdir "arangodb-build" do
```
ilovezfs: Let's make this
```
    mkdir "build" do
      args = std_cmake_args + %W[
        -DHOMEBREW=ON
        -DUSE_OPTIMIZE_FOR_ARCHITECTURE=OFF
        -DASM_OPTIMIZATIONS=OFF
        -DETCDIR=#{etc}
        -DVARDIR=#{var}
      ]
      system "cmake", "..", *args
      system "make", "install"
    end
```

Formula/arangodb3.rb
```diff
@@ -0,0 +1,85 @@
...
+      system "cmake", "..", "-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DHOMEBREW=On", "-DUSE_OPTIMIZE_FOR_ARCHITECTURE=Off", "-DASM_OPTIMIZATIONS=Off", "-DETCDIR=#{prefix}/etc", "-DVARDIR=#{var}", *std_cmake_args
```
ilovezfs: I believe the `-DETCDIR=#{prefix}/etc` is a bug and should be `-DETCDIR=#{etc}` as I have it above, but I could be mistaken if there's some explanation for the non-standard location?

Formula/arangodb3.rb
```diff
@@ -0,0 +1,85 @@
...
+  def post_install
+    (var/"lib/arangodb3").mkpath
+    (var/"log/arangodb3").mkpath
+  end
+
+  def caveats
+    s = <<-EOS.undent
+      The database format between ArangoDB 2.x and ArangoDB 3.x has
+      been changed. Please export the old database using arangodump
+      and import it again using arangorestore.
```
ilovezfs: This seems like it's duplicating upstream documentation.

Formula/arangodb3.rb
ilovezfs: Or perhaps it should be `-DETCDIR=#{etc}/arangodb3` as you have it in the plist.

Formula/arangodb3.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      An empty password has been set. Please change it by executing
+        #{sbin}/arango-secure-installation
+    EOS
+
+    s
+  end
+
+  plist_options :manual => "#{HOMEBREW_PREFIX}/opt/arangodb3/sbin/arangod"
```
ilovezfs: Does starting it manually require any args?

ilovezfs:
Formula rename typically looks like this: https://github.com/Homebrew/homebrew-core/commit/d257db5ab697e34bd393118c7127fd8f0ccbcf9f

Presumably you want current users of the preexisting arangodb.rb to be `brew upgrade`'d to this "new" formula?

Formula/arangodb3.rb
```diff
@@ -0,0 +1,85 @@
...
+        </array>
+        <key>RunAtLoad</key>
+        <true/>
+      </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    assert_equal "it works!\n", shell_output("#{bin}/arangosh --javascript.execute-string \"require('@arangodb').print('it works!')\"")
```
ilovezfs: would be good if this could be refactored into multiple lines so it doesn't exceed the 80 character column limit

Formula/arangodb3.rb
fceller: Fixed

Formula/arangodb3.rb
fceller: Fixed

Formula/arangodb3.rb
fceller: done

Formula/arangodb3.rb
fceller: The "arangodb3" is added by the CMakeFile.txt, will try.

Formula/arangodb3.rb
fceller: It might be a good idea to include a hint, because only all 2.x release, there is no automatic update. I will add a link to the upstream doc.

Formula/arangodb3.rb
fceller: No, there were some args in 2.8, but the are no longer needed.

ilovezfs: It looks like the plist itself is still passing the args.

Formula/arangodb3.rb
fceller: done

Formula/arangodb3.rb
fceller: I will check

Formula/arangodb3.rb
fceller: also added a rename

Formula/arangodb3.rb
fceller: @ilovezfs removed the args

Formula/arangodb3.rb
```diff
@@ -0,0 +1,88 @@
...
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>KeepAlive</key>
+        <true/>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>ProgramArguments</key>
```
ilovezfs: If there's only one argument (i.e., just the program name itself), you should use `<key>Program</key>` instead of `<key>ProgramArguments</key>`. It takes a `<string>` instead of an `<array>`.

Formula/arangodb3.rb
```diff
@@ -0,0 +1,88 @@
...
+        <key>RunAtLoad</key>
+        <true/>
+      </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    testcase = "require('@arangodb').print('it works!')"
+    output = shell_output("#{bin}/arangosh --javascript.execute-string \"${testcase}\"")
```
ilovezfs: The test still fails without a blank `--password`, right?

ilovezfs: Actually, I think it's ` --server.password`, but same question.

Formula/arangodb3.rb
fceller: I see. Fixed.

Formula/arangodb3.rb
fceller: yes, right. Will add.

MikeMcQuaid:
I don't think this formula should be renamed. Instead the `install` process should automatically handle the migration of existing data in a `post_install` method.

ilovezfs:
@MikeMcQuaid The intent of the rename had nothing to do with the database migration, but rather was intended to reflect the upstream changes in the paths, which now use "arangodb3" subfolders instead of "arangodb" subfolders across the board. For example, "etc/arangodb3" instead of "etc/arangodb", "share/arangodb3" instead of "share/arangodb", etc.

That said, @fceller you should do as Mike said and undo the rename. Also, the post_install method is going to require some work if it's going to handle the migration automatically. Do you know the sequence of commands necessary to accomplish that task?

fceller:
@ilovezfs there is no automatic migration. The reason is, that you need to dump the database BEFORE upgrading and then import the dump AFTER upgrading. So there is no easy procedure unless you can execute something before installation using the ArangoDB 2 version.

ilovezfs:
That should be possible if you use a bootstrapping technique and vendor arrangodb 2 (e.g., installing arrangodb 2 into the buildpath).

fceller:
@ilovezfs if the arangodb2 from is available, can I use "brew arangodb2" or how does this bootstrapping work?

ilovezfs:
Even if arrangodb2 is accepted into homebrew/versions, that doesn't help because cross-tap dependencies are only permissible if required by non-default formula options. If you mean the user's preexisting install that's being upgraded, you won't know the path to the binaries because it won't be linked during brew upgrade, and you won't know exactly which version the user is coming from, so the full path to the binaries within the Cellar would be variable. Also, there could be existing arrangodb2 databases that need migration even if the formula isn't currently installed because the user e.g., does brew uninstall + brew install instead of brew upgrade.

For example, during an upgrade from wget 1.17.1 to 1.18, here's what you see:
```
bash-3.2$ which wget
bash-3.2$ ls /usr/local/bin/wget
ls: /usr/local/bin/wget: No such file or directory
bash-3.2$ ls /usr/local/Cellar/wget
1.17.1	1.18
bash-3.2$
```
So you'd have to "figure out" that the preexisting wget binary's path is `/usr/local/Cellar/wget/1.17.1/bin/wget`, and not, say, `/usr/local/Cellar/wget/1.17/bin/wget`.

So it's probably simpler to put arrangodb2 into a resource block, and then `make install` it into the buildpath, since then you'll know exactly where to find it and have a guarantee it's available since you put it there.

fceller:
@ilovezfs thanks for the explanation. Will try.

ilovezfs:
If you want to do all of the migration steps during post_install, then you'd need to install arrangodb2 into arrangodb3's prefix in `libexec/"arrangodb2"`, which might be the most flexible way to handle this situation, especially since the postinstall step would be re-runnable on demand, whereas if you put it in the buildpath, then it's evanescent.

ilovezfs:
Note that the post install step can be isolated from the rest of brew install and re-run on demand for your testing purposes by running the command `brew postinstall arrangodb`.

ilovezfs:
@fceller Here's an example that may help: https://github.com/Homebrew/homebrew-core/blob/master/Formula/compcert.rb#L38-L55

You'd substitute `libexec` for `buildpath` if you want to keep it around to be available during post install.

fceller:
@ilovezfs cool, that should help a lot

fceller:
@ilovezfs I'm a little puzzled. Do you know why mavericks failed? Can only see a timeout in the logs.

Formula/arangodb.rb
```diff
@@ -11,81 +12,92 @@ class Arangodb < Formula
...
 
-    if build.head?
-      mkdir "arangodb-build" do
-        system "cmake", "..", "-DHOMEBREW=On", "-DUSE_OPTIMIZE_FOR_ARCHITECTURE=Off", "-DASM_OPTIMIZATIONS=Off", "-DETCDIR=#{prefix}/etc", "-DVARDIR=#{var}", *std_cmake_args
-        system "make", "install"
-      end
+    resource("upgrade").stage do
+      (libexec/"arangodb2/bin").mkpath
+      cp "upgrade.sh", libexec/"arangodb2/bin"
+    end
```
ilovezfs:
```
resource("upgrade").stage { (libexec/"arangodb2/bin").install "upgrade.sh" }
```

ilovezfs: Actually, since the only thing in the tarball is `upgrade.sh`:
```
(libexec/"arangodb2/bin").install resource("upgrade")
```

ilovezfs:
@fceller `brew install --bottle ... ; brew bottle ...` creates the bottle but doesn't run post_install. The subsequent `brew install ./foobottle ...` pours the bottle and then runs post install. So most likely the time out is during post install since the initial `brew install --bottle` didn't time out.

fceller:
@ilovezfs is it possible to see the output of the `brew install`. The post install seems to work on the other two and I do not have access to a mavericks system.

ilovezfs:
@fceller I'll enable `HOMEBREW_VERBOSE=1` on CI and let's see if that helps. If not, we can also try `HOMEBREW_DEBUG=1`

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@fceller So here's what the verbose log shows for Mavericks:
```
x arangodb/3.0.1/.bottle/etc/arangodb3/arango-dfdb.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arango-init-database.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arango-secure-installation.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arangobench.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arangod.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arangodump.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arangoimp.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arangorestore.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/arangosh.conf
x arangodb/3.0.1/.bottle/etc/arangodb3/foxx-manager.conf
==> Finishing up
ln -s ../Cellar/arangodb/3.0.1/bin/foxx-manager foxx-manager
ln -s ../Cellar/arangodb/3.0.1/bin/arangosh arangosh
ln -s ../Cellar/arangodb/3.0.1/bin/arangorestore arangorestore
ln -s ../Cellar/arangodb/3.0.1/bin/arangoimp arangoimp
ln -s ../Cellar/arangodb/3.0.1/bin/arangodump arangodump
ln -s ../Cellar/arangodb/3.0.1/bin/arangobench arangobench
ln -s ../Cellar/arangodb/3.0.1/sbin/arangod arangod
ln -s ../Cellar/arangodb/3.0.1/sbin/arango-secure-installation arango-secure-installation
ln -s ../Cellar/arangodb/3.0.1/sbin/arango-init-database arango-init-database
ln -s ../Cellar/arangodb/3.0.1/sbin/arango-dfdb arango-dfdb
mkdir -p /usr/local/share/man/man8
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man8/rcarangod.8 rcarangod.8
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man8/foxx-manager.8 foxx-manager.8
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man8/arangod.8 arangod.8
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man8/arango-dfdb.8 arango-dfdb.8
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangosh.1 arangosh.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangorestore.1 arangorestore.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangoimp.1 arangoimp.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangodump.1 arangodump.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangobench.1 arangobench.1
ln -s ../../Cellar/arangodb/3.0.1/share/doc/arangodb3 arangodb3
ln -s ../Cellar/arangodb/3.0.1/share/arangodb3 arangodb3
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160705-62181-1ioeynk-0.sb nice /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/postinstall.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/arangodb.rb
==> /usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2/bin/upgrade.sh /usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2 /usr/local/var/lib/arangodb /usr/local/opt/arangodb /usr/local/var/lib/arangodb3
Upgrade in progress...
Waiting for ArangoDB 2 to boot...
/usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2/bin/upgrade.sh: line 43: 62184 Abort trap: 6           ${ARANGOD2} -c ${ARANGOD2_CONF} --log.tty ""
Waiting for ArangoDB 2 to boot...
Waiting for ArangoDB 2 to boot...
Waiting for ArangoDB 2 to boot...
Waiting for ArangoDB 2 to boot...
```
with the "Waiting for ArangoDB 2 to boot..." repeating forever until CI times out.

fceller:
@ilovezfs I tried to put in some output when executing the upgrade, to understand, why it fails under mavericks. Now I'm getting:

`Error: private method `ruby_has_encoding?' called for Homebrew:Module`

Which I'm not using at all. Do you have an hint, what is causing this?

ilovezfs:
Yes: https://github.com/Homebrew/brew/pull/279#issuecomment-230473188

fceller:
@BrewTestBot test this please

apjanke:
Should be fixed now, in https://github.com/Homebrew/brew/commit/61ffa47fd9dc179ddff792db1dc6f55464f6c16b.

@BrewTestBot test this please

ilovezfs:
@apjanke thanks!

fceller:
@ilovezfs I think I found the reason it is not working under Mavericks: with gcc6 there is an issue with V8.

fceller:
@BrewTestBot test this please

ilovezfs:
@fceller Mavericks still seems to get stuck.

fceller:
@ilovezfs yes, very frustrating. I found that compiling with GCC requires special parameters for V8 - which I have added.

Without a mavericks machine it is nearly impossible to debug. Is there a way to see the actual output of the run?

fceller:
@ilovezfs I have downloaded and install the mavericks bottle on my system without problem. So, I'm a little stuck.

ilovezfs:
It's running with verbose and debug right now, so we'll see if that turns anything useful up.

ilovezfs:
@fceller looks like it's the same as above:
```
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangosh.1 arangosh.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangorestore.1 arangorestore.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangoimp.1 arangoimp.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangodump.1 arangodump.1
ln -s ../../../Cellar/arangodb/3.0.1/share/man/man1/arangobench.1 arangobench.1
ln -s ../../Cellar/arangodb/3.0.1/share/doc/arangodb3 arangodb3
ln -s ../Cellar/arangodb/3.0.1/share/arangodb3 arangodb3

==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160708-97612-1wlst7c-0.sb nice /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/postinstall.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/arangodb.rb
/usr/local/Library/Homebrew/postinstall.rb (Formulary::FromPathLoader): loading /usr/local/Library/Taps/homebrew/homebrew-core/Formula/arangodb.rb
==> /usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2/bin/upgrade.sh /usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2 /usr/local/var/lib/arangodb /usr/local/opt/arangodb /usr/local/var/lib/arangodb3
Upgrade in progress...
Waiting for ArangoDB 2 to boot...

/usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2/bin/upgrade.sh: line 43: 97615 Abort trap: 6           ${ARANGOD2} -c ${ARANGOD2_CONF} --log.tty ""

Waiting for ArangoDB 2 to boot...

Waiting for ArangoDB 2 to boot...

Waiting for ArangoDB 2 to boot...

Waiting for ArangoDB 2 to boot...
...
```

fceller:
@ilovezfs that still looks that there is a problem with GCC6. I'll try GCC5

ilovezfs:
@fceller ok I built a bottle with GCC6 on El Capitan and the issue is reproducible. Here's what I'm seeing.
```
iMac-TMP:local joe$ /usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2/sbin/arangod-2.8 -c /usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2/etc/arangodb/arangod-2.8.conf 
2016-07-08T19:03:31Z [49511] INFO file-descriptors (nofiles) new hard limit is -1, new soft limit is 1024
2016-07-08T19:03:31Z [49511] INFO ArangoDB 2.8.10 [darwin] 64bit, using ICU 54.1, V8 4.3.61, OpenSSL 1.0.2h  3 May 2016
2016-07-08T19:03:31Z [49511] INFO using default language 'en'
2016-07-08T19:03:31Z [49511] INFO loaded database '_system' from '/usr/local/var/lib/arangodb/databases/database-76099'
2016-07-08T19:03:31Z [49511] INFO running WAL recovery (1 logfiles)
2016-07-08T19:03:31Z [49511] INFO replaying WAL logfile '/usr/local/var/lib/arangodb/journals/logfile-7972160.db' (1 of 1)
2016-07-08T19:03:31Z [49511] INFO WAL recovery finished successfully
2016-07-08T19:03:31Z [49511] INFO using endpoint 'tcp://127.0.0.1:8529' for non-encrypted requests
2016-07-08T19:03:31Z [49511] INFO using default API compatibility: 20800
2016-07-08T19:03:31Z [49511] INFO JavaScript using startup '/usr/local/Cellar/arangodb/3.0.1/libexec/arangodb2/share/arangodb/js', application '/usr/local/var/lib/arangodb-apps'
2016-07-08T19:03:32Z [49511] INFO In database '_system': Database is up-to-date (20810/prod/standalone/existing)
Abort trap: 6
```

And lldb shows https://gist.github.com/ilovezfs/b9494adc623722628b7cb0bae65c8412

fceller:
@ilovezfs yes, that seems to the crash in V8 that happens with GCC6. The comments suggested that "-fno-delete-null-pointer-checks" should fix the problem. I have added `fails_with :gcc => "6.0"`.

fceller:
At least that is what node it doing (which also uses V8).

ilovezfs:
```
fails_with :gcc => "6.0"
```
I don't think that will actually work. That won't actually match anything. It would need to be `fails_with :gcc => "6"` but I don't believe we allow that in core, since it will just cause it to skip building Mavericks anyway given that gcc5 is now in homebrew/versions.

Most likely the gcc6 breakage needs to be directly fixed or the formula set to be only >= yosemite.

fceller:
@ilovezfs I see. I'll try to find the correct GCC6 flags.

ilovezfs:
@fceller superenv may filter them out behind your back, so make sure you're testing outside of brew first to get an accurate assessment of what works and what doesn't. Then we can worry about how to communicate that to brew if possible.

fceller:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

fceller:
@ilovezfs There is a second problem with GCC and Darwin: Exception do not work properly when linking statically the gcc_s. So I removed that option. But now I get

```
Error: undefined method `relocate_dynamic_files' for #<Keg:/usr/local/Cellar/arangodb/3.0.1>
Please report this bug:
/usr/local/Library/Homebrew/cmd/bottle.rb:193:in `bottle_formula'
/usr/local/Library/Homebrew/keg.rb:245:in `lock'
/usr/local/Library/Homebrew/formula_lock.rb:29:in `with_lock'
/usr/local/Library/Homebrew/keg.rb:241:in `lock'
/usr/local/Library/Homebrew/cmd/bottle.rb:186:in `bottle_formula'
/usr/local/Library/Homebrew/cmd/bottle.rb:452:in `bottle'
/usr/local/Library/Homebrew/cmd/bottle.rb:451:in `each'
/usr/local/Library/Homebrew/cmd/bottle.rb:451:in `bottle'
/usr/local/Library/brew.rb:87:in `send'
/usr/local/Library/brew.rb:87

```

ilovezfs:
@fceller Yeah that was a regression in brew itself which is now fixed. Nothing specific to this formula.

fceller:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@fceller It's a miracle: :green_apple:

fceller:
@ilovezfs it now compiles and runs the tests under mavericks. I cannot test anything else, because I have no mavericks machine available.

ilovezfs:
@fceller excellent job getting that fixed with GCC6. What exactly was the problem? Is there an upstream issue report and/or pull request fixing it?

fceller:
@ilovezfs there are two problems:

ilovezfs:
(1) v8 is vendored by arangodb, right? So if v8 has already fixed this, then presumably the vendored version in arangodb needs to be updated accordingly, or am I misunderstanding the situation?

(2) sounds like was an issue the formula itself was creating by setting static, yes?

fceller:
(1) Basically, it is an interpretation of the standard of C++ and how to deal with delete. V8 and GCC 6 disagree on a certain aspect. And you need to add -fno-delete-null-pointer-checks to force GCC 6 to use what V8 expects.

ArangoDB 3.0 automatically adds the correct flags to a GCC compile. ArangoDB 2.8 does not - but compiled with clang normally. It was the combination, that ArangoDB 3.0 needs a C++ 11 compiler and CLANG is too old on mavericks.

(2) I removed the static linkage from the formula. As far as I can see it should not case any harm linking dynamically.

It is still incorrect behavior of GCC under Darwin, that exceptions are not caught when linking statically. But is the bug report is now 6 years old, I assume it will not be fixed.

ilovezfs:
Ah OK, so it's only a tweak for the vendored arangodb version 2 that is in there for the sake of the migration. Is upstream still going to be doing additional version 2 maintenance releases?

fceller:
@ilovezfs Yes, ArangoDB 2.8 will be support for some months to come.

fceller:
I have created a version formula here: https://github.com/Homebrew/homebrew-versions/pull/1304

ilovezfs:
It would be great if you could open an issue report suggesting they backport the GCC6 flag fixes to 2.8

ilovezfs:
@fceller Also, what's the deal with the :DATA patch for malloc_usable_size? Has that been reported somewhere?

fceller:
@ilovezfs yes, that will be fixed in ArangoDB 3.0.2.

fceller:
@ilovezfs I've already backported this in Homebrew/homebrew-versions#1304

ilovezfs:
@fceller should this PR be refreshed to point to 3.0.2 with the patch removed instead since you already tagged it?

Formula/arangodb.rb
```diff
@@ -11,81 +12,98 @@ class Arangodb < Formula
...
-  end
-
+  depends_on "cmake" => :build
   depends_on "go" => :build
   depends_on "openssl"
 
   needs :cxx11
 
+  # malloc_usable_size is not supported
+  patch :DATA
```
DomT4: If this has already been upstreamed you'll need to move it inside a `stable do` block, otherwise it'll presumably fail to apply for `head`.

Formula/arangodb.rb
```diff
@@ -11,81 +12,98 @@ class Arangodb < Formula
...
 
-    if build.head?
-      mkdir "arangodb-build" do
-        system "cmake", "..", "-DHOMEBREW=On", "-DUSE_OPTIMIZE_FOR_ARCHITECTURE=Off", "-DASM_OPTIMIZATIONS=Off", "-DETCDIR=#{prefix}/etc", "-DVARDIR=#{var}", *std_cmake_args
-        system "make", "install"
-      end
+    (libexec/"arangodb2/bin").install resource("upgrade")
+
+    resource("arangodb2").stage do
+      ENV.libcxx
```
DomT4: Do we not push down the `ENV` from `def install` into resource stages automatically? That could produce some interesting quirks.

Formula/arangodb.rb
```diff
@@ -11,81 +12,98 @@ class Arangodb < Formula
...
   end
 
+  def post_install
+    (var/"lib/arangodb3").mkpath
+    (var/"log/arangodb3").mkpath
+
+    args = %W[
+      #{libexec}/arangodb2
+      #{var}/lib/arangodb
+      #{HOMEBREW_PREFIX}/opt/arangodb
```
DomT4: Does a simple `#{opt_prefix}` not work here?

Formula/arangodb.rb
```diff
@@ -11,81 +12,98 @@ class Arangodb < Formula
...
+      https://docs.arangodb.com/3.0/Manual/Administration/Upgrading/index.html
 
-    if build.head?
-      s += <<-EOS.undent
-        A default password has been set. You can change it by executing
-          #{sbin}/arango-secure-installation
-      EOS
-    end
+      An empty password has been set. Please change it by executing
+        #{sbin}/arango-secure-installation
```
DomT4: No change in functionality here but `opt_sbin` is a bit nicer to look at/memorise in printed caveats.

Formula/arangodb.rb
fceller: done

Formula/arangodb.rb
fceller: I'll try

ilovezfs: Does it even make sense to be specifying both ENV.libcxx and ENV.cxx11 simultaneously?

Formula/arangodb.rb
fceller: Should this also be #{opt_prefix] in plist_options?

Formula/arangodb.rb
ilovezfs: Would prefer to just switch this over to 3.0.2 if it's ready ...

Formula/arangodb.rb
fceller: done

Formula/arangodb.rb
DomT4: Sadly not. IIRC the magic doesn't work in `plist_options`.

Formula/arangodb.rb
DomT4: Possibly not, from vague recollection. I'm fairly sure the latter does the former.

Formula/arangodb.rb
DomT4: That's certainly desirable if a possibility.

Formula/arangodb.rb
ilovezfs: Yup.

```
  def cxxflags
    args = cflags
    args << "-std=c++11" if cxx11?
    args << "-stdlib=libc++" if libcxx?
    args << "-stdlib=libstdc++" if libstdcxx?
    args
  end
```
```
  def cxx11?
    config.include?("x")
  end

  def libcxx?
    config.include?("g")
  end

  def libstdcxx?
    config.include?("h")
  end
```

```
  def cxx11
    if homebrew_cc == "clang"
      append "HOMEBREW_CCCFG", "x", ""
      append "HOMEBREW_CCCFG", "g", ""
    elsif gcc_with_cxx11_support?(homebrew_cc)
      append "HOMEBREW_CCCFG", "x", ""
    else
      raise "The selected compiler doesn't support C++11: #{homebrew_cc}"
    end
  end

  def libcxx
    append "HOMEBREW_CCCFG", "g", "" if compiler == :clang
  end

  def libstdcxx
    append "HOMEBREW_CCCFG", "h", "" if compiler == :clang
  end
```

So ENV.cxx11 -> x and g -> cxx11 and libcxx -> -std=c++11 and -stdlib=libc++

and

ENV.libcxx -> g -> libcxx -> -stdlib=libc++

and

ENV.libstdcxx -> h -> libstdcxx -> -stdlib=libstdc++

ilovezfs: >Do we not push down the ENV from def install into resource stages automatically? That could produce some interesting quirks.

Yes, it gets pushed down, so in this particular case an additional ENV.libcxx would be harmless but redundant in that a "g" was already appended by the ENV.cx11, and appending yet another "g" wouldn't affect the include?("g") or the include?("x").

MikeMcQuaid:
I think if you still can't get Mavericks working it'd be reasonable to just make this depend on Yosemite and above.

ilovezfs:
@MikeMcQuaid He did get it working! https://github.com/Homebrew/homebrew-core/pull/2576#issuecomment-231549928

It turned out to be GCC6 issues as opposed to Mavericks per se.

Formula/arangodb.rb
```diff
@@ -26,66 +29,80 @@ class Arangodb < Formula
...
   end
 
+  def post_install
+    (var/"lib/arangodb3").mkpath
+    (var/"log/arangodb3").mkpath
+
+    args = %W[
+      #{libexec}/arangodb2
+      #{var}/lib/arangodb
+      #{opt_prefix}/arangodb
```
ilovezfs: This is the cause of the CI failure. It should be `#{opt_prefix}` not `#{opt_prefix}/arangodb`, which is redundant since `opt_prefix` means `/usr/local/opt/arangodb`:
```
Josephs-MacBook-Pro:~ joe$ brew ruby -e "puts Formula[\"arangodb\"].opt_prefix"
/usr/local/opt/arangodb
```

Formula/arangodb.rb
DomT4: > Yes, it gets pushed down...

Cool 👍. I thought it would otherwise we'd have seen some fairly bizarre behaviour between main build & resource build interaction by now.

ilovezfs: On the other hand, it looks like changes to the environment made during the resource staging block also make it back into def install, which some might consider a bug.

Formula/arangodb.rb
fceller: Fixed


Am 10.07.16 um 11:54 schrieb ilovezfs:

ilovezfs:
@fceller so is that 3.0.2 tag shippable?

fceller:
@ilovezfs it should be today. I'll update the pull-request as soon as it becomes available.

fceller:
@BrewTestBot test this please

fceller:
@BrewTestBot test this please

ilovezfs:
@fceller was 2.8.10 retagged? I guess https://github.com/Homebrew/homebrew-versions/pull/1304 needs the same update?

fceller:
@ilovezfs retag done

ilovezfs:
@fceller I think it's back to hanging on mavericks :(

fceller:
@ilovezfs interesting, will add the ENV.cxx11 in the resource again.

ilovezfs:
Not sure what that will do since it's just appending the same thing to the same string again, but we'll see I guess...

Did anything else change? What's different about the new tarball?

ilovezfs:
weird. @fceller can you point out what you changed and where to fix it?

fceller:
I did unfortunately change two things:

- add ENV.libcxx in install
- add ENV.libcxx and ENV.cxx11in resource

I will remove the ENV.libcc to see which of the two is the root cause

fceller:
@ilovezfs it seems that the ENV.cxx11 is needed in the resource block as well.

Can you merge now?

ilovezfs:
>Can you merge now?

As soon as we understand why that's the case.

@UniqMartin @MikeMcQuaid @apjanke @xu-cheng Is it the case that ENV.cxx11 has to be specified *again* inside a `resource.stage do ... end` even if it was already specified at the beginning of the beginning of install (see formula body in this PR)?

ilovezfs:
Specifically, https://github.com/fceller/homebrew/blob/678ee64812dc3ed2a277fcdea03541917915a2c7/Formula/arangodb.rb#L38-L43

MikeMcQuaid:
@ilovezfs I don't believe it should be if it is: that's fine.

ilovezfs:
@MikeMcQuaid I'm concerned that if it's not, then there's something else going on here causing the Mavericks hangs to happen intermittently.

As far as I can tell the HOMEBREW_CCCFG makes into the resource block just fine, but apparently setting ENV.cxx11 a second time is somehow preventing the hang (???), which doesn't seem like a plausible explanation.

Also, I can find no precedent for setting it twice in core.

MikeMcQuaid:
I'm fine to just merge this as-is and investigate more later.

ilovezfs:
Okie dokie. Seems like we're merging voodoo spells, but so be it.

ilovezfs:
@fceller Nice work on this, especially getting the upgrade working in post_install. Thanks for your contribution to Homebrew! :tada:

fceller:
@ilovezfs I'm afraid all the testing confused the brew bot: The bottle is now broken, because it contains "3.0.1" and "3.0.2":

> wget "https://homebrew.bintray.com/bottles/arangodb-3.0.2.el_capitan.bottle.tar.gz"
> tar xzvf arangodb-3.0.2.el_capitan.bottle.tar.gz
> ls -l arangodb
total 0
drwxr-xr-x@ 13 fceller  wheel  442  8 Jul 19:58 3.0.1
drwxr-xr-x  13 fceller  wheel  442  9 Jul 19:36 3.0.2

Can regenrated the bottle from scratch?

fceller:
@ilovezfs @MikeMcQuaid the mavericks & yosemite bottles are fine. It is just the el_captain one that is broken.

ilovezfs:
@fceller thanks for the heads up. Jenkins is down right now for some reason, but we can fix it once it's back up.

fceller:
@ilovezfs thanks a lot. It is strange, that the bottle somehow contains old data. I assume all the testing confused the brewer.

ilovezfs:
Yeah I haven't diagnosed the situation yet but it is indeed very weird.

ilovezfs:
@fceller it looks like a false alarm to me:
```
iMac-TMP:~ joe$ brew install arangodb
==> Downloading https://homebrew.bintray.com/bottles/arangodb-3.0.2.el_capitan.bottle.tar.gz
==> Pouring arangodb-3.0.2.el_capitan.bottle.tar.gz
==> /usr/local/Cellar/arangodb/3.0.2/libexec/arangodb2/bin/upgrade.sh /usr/local/Cellar/arangodb/3.0.
==> Caveats
The database format between ArangoDB 2.x and ArangoDB 3.x has
been changed, please checkout
https://docs.arangodb.com/3.0/Manual/Administration/Upgrading/index.html

An empty password has been set. Please change it by executing
  /usr/local/opt/arangodb/sbin/arango-secure-installation

To have launchd start arangodb now and restart at login:
  brew services start arangodb
Or, if you don't want/need a background service you can just run:
  /usr/local/opt/arangodb/sbin/arangod
==> Summary
🍺  /usr/local/Cellar/arangodb/3.0.2: 8,952 files, 303.3M
iMac-TMP:~ joe$ ls /usr/local/Cellar/arangodb
3.0.2
Mac-TMP:~ joe$ arangodump --version
3.0.2

architecture: 64bit
asan: false
asm-crc32: false
boost-version: 1.61.0b1
build-date: 2016-07-10 20:12:20
compiler: clang
cplusplus: 201103
endianness: little
fd-client-event-handler: poll
fd-setsize: 1024
icu-version: 54.1
jemalloc: false
libev-version: 4.22
maintainer-mode: false
openssl-version: OpenSSL 1.0.2h  3 May 2016
rocksdb-version: 4.8.0
server-version: 3.0.2
sizeof int: 4
sizeof void*: 8
sse42: false
tcmalloc: false
v8-version: 5.0.71.39
vpack-version: 0.1.30
zlib-version: 1.2.8

iMac-TMP:~ joe$
```
What's making you think it has 3.0.1?

fceller:
@ilovezfs the problem lies in the file "/usr/local/etc/arangodb3/arangodb.conf". It contains the path "3.0.1" which no longer exists causing errors later.

fceller:
@ilovezfs if I check `fgrep 3.0.1 /usr/local/etc/arangodb3/arangod.conf` it shows an entry (which is wrong).

MikeMcQuaid:
You cannot rely on `etc` files to be overwritten by default on bottle installation. That could be the issue?

ilovezfs:
Yeah, it needs to be fixed in the post_install method.

@fceller correct that an upgrade will leave you with 3.0.1 config file right now. But if you delete the config file before brew install, or if it's a fresh install, the 3.0.2 file is what's installed.

fceller:
@ilovezfs @MikeMcQuaid would not in this case be much better to use the "/usr/local/share/arangodb3" instead of the full qualified "/usr/local/Cellar/arangodb/3.0.2/share/arangodb3" within the config file?

This way there is no need to patch the config file on every update.

MikeMcQuaid:
@fceller Yeh, seems like it 👍

fceller:
@MikeMcQuaid will change this. Shall I create a new pull request or amend to this?

ilovezfs:
@fceller sometimes people do this: https://github.com/Homebrew/homebrew-core/blob/master/Formula/couchdb.rb#L130-L134

But I agree that a separate directory is probably more user friendly.

ilovezfs:
@fceller you should simply open a new PR.

DomT4:
> would not be much better to use the "/usr/local/share/arangodb3" instead of the full qualified "/usr/local/Cellar/arangodb/3.0.2/share/arangodb3" within the config file?

`opt_pkgshare` would work as well. That links to `/usr/local/opt/arangodb3/share/arangodb3`, which is "safe" in terms of not breaking across updates.

DomT4:
[This](https://github.com/Homebrew/homebrew-core/blob/fb81645562f7e495a49561065912788af2d1f14f/Formula/unbound.rb#L26-L30) is generally my favourite way to edit files that the bottle doesn't necessarily overwrite or can't be fixed in bottles (Such as things that set a specific `$USER` during compile).

PR #2577, opened 2016-07-02
hg-git 0.8.5 (new formula)

catap:

---

### DIFF
--- a/Formula/hg-git.rb
+++ b/Formula/hg-git.rb
@@ -1,0 +1,22 @@
+class HgGit < Formula
+  desc "Git Mercurial Plugin"
+  homepage "http://hg-git.github.io"
+  url "https://bitbucket.org/durin42/hg-git/get/0.8.5.tar.bz2"
+  sha256 "60bf47386fc0c9eef809344ff2be98eaa5add822f70dd25670470d06673fe2c4"
+  head "https://bitbucket.org/durin42/hg-git", :using => :hg
+
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on :hg
+
+  def install
+    system "python", *Language::Python.setup_install_args(prefix)
+  end
+
+  test do
+    (testpath/".hgrc").write <<-EOS.undent
+      [extensions]
+      hggit=
+    EOS
+    system "hg", "clone", "git://github.com/schacon/hg-git.git"
+  end
+end


chdiza:
This package has been proposed a few times in the past, and has always been rejected.  (Not by me.)

It's just a python package, installable with `pip` or `setuptools`, and HB maintainers typically don't accept those kinds of things.

(That doesn't mean there's anything wrong with your formula.)

catap:
@chdiza but brew has [hg-flow](https://github.com/Homebrew/homebrew-core/blob/master/Formula/hg-flow.rb).

Meanwhile I agree that hg-flow has installed as single py file.

MikeMcQuaid:
> It's just a python package, installable with pip or setuptools, and HB maintainers typically don't accept those kinds of things.

We accept them not as long as they are runnable and not just e.g. a library.

Formula/hg-git.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    EOS
+  end
+
+  test do
+    (testpath/".hgrc").write <<-EOS.undent
+      [extensions]
+      hggit=
+    EOS
+    system "hg", "clone", "git://github.com/schacon/hg-git"
```
MikeMcQuaid: Test is failing: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4110/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_hg_git/

Formula/hg-git.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def caveats; <<-EOS.undent
+    1. Put following lines into your ~/.hgrc
+    2. For more information go to http://hg-git.github.io
+
+        [extensions]
+        hggit=
+
+    EOS
+  end
```
MikeMcQuaid: These caveats aren't Homebrew-specific so can be removed.

chdiza:
It's not a runnable CLI program.  It's a plugin for Mercurial.

Formula/hg-git.rb
catap: But without this line it doesn't work.

dunn: Yes, but we shouldn't be replicating upstream documentation since it'll end up neither as helpful nor as up-to-date.  We can't take on the role of explaining how to use every piece of software we install.

MikeMcQuaid:
@chdiza Given we allow installation of `hg` from Homebrew this makes sense.

Formula/hg-git.rb
catap: Got it.

dunn:
The test is failing, though: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4653/

chdiza:
It's failing because `hg-git` depends on `dulwich`, which is a python library.

DomT4:
Status update here please.

MikeMcQuaid:

PR #2579, opened 2016-07-02
consul-backinator 1.0 (new formula)

aaronhurt:
New formula for my consul cluster backup utility.


DomT4:
I'm afraid this doesn't currently meet our eligibility requirements.
```
Error: 1 problem in 1 formula
consul-backinator:
```
It's not massively far off, so it may be a case of we can look at this afresh in a few weeks once it does, but until then you may want to host the formula [in your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md). Thank you nonetheless for your submission to Homebrew ❤️.

PR #2581, opened 2016-07-02
Enable static libraries for imagemagick

jeroen:
This enables the the static libraries for `imagemagick` which we need to build self-contained applications. Tested to build without warnings or side effects.

Some [precedents](https://github.com/Homebrew/legacy-homebrew/commits/master?author=jeroenooms) of enabling static libraries. 

Yes we _could_ make static libraries optional, but it would be much nicer if they get bottled by default so that my users don't have to build from source. :blue_heart: :beer:

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -64,7 +64,7 @@ class Imagemagick < Formula
       --disable-dependency-tracking
       --disable-silent-rules
       --enable-shared
-      --disable-static
+      --enable-static
       --with-modules
     ]
 


DomT4:
Seems fine. We only really object to static libraries if they cause some obscure problem _or_ more commonly where enabling static disables dynamic.

Merged in https://github.com/Homebrew/homebrew-core/commit/91cb1a5c11221b80c69c52510668b0193ea910fb. Thanks again @jeroenooms!

PR #2595, opened 2016-07-02
glyr 1.0.9

ilovezfs:
Also, test cleanups:
- keep lines from exceeding the eightieth column
- avoid an intermediate file by using "-w stdout" and pipe_output

### DIFF
--- a/Formula/glyr.rb
+++ b/Formula/glyr.rb
@@ -1,8 +1,8 @@
 class Glyr < Formula
   desc "Music related metadata search engine with command-line interface and C API"
   homepage "https://github.com/sahib/glyr"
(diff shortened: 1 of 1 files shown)


DomT4:
If you're feeling particularly kind you could wrap that long test line to 80 but otherwise 👍.

ilovezfs:
@DomT4 How about
```
    search = "--artist Beatles --album \"Please Please Me\""
    cmd = "#{bin}/glyrc cover --no-download #{search} -w stdout"
    assert_match %r{^https?://}, pipe_output(cmd, nil, 0)
```

DomT4:
If it works, that seems fine.

PR #2597, opened 2016-07-02
gvp 0.3.0

ilovezfs:
- apply upstream patch for syntax error in switch operator
- subcommand used in the test changed from "init" to "in"


ilovezfs:
@DomT4 :ship:?

DomT4:
👍

PR #2605, opened 2016-07-02
libsoundio 1.1.0

ilovezfs:
needs to be restricted to >= Yosemite due to stdatomic.h, which
Mavericks and below don't have


ilovezfs:
@BrewTestBot test this please

PR #2607, opened 2016-07-02
micropython 1.8.2

ilovezfs:
- switch to using the tag since micropython now has git submodules
- use inreplace to work around an #include bug in the axTLS submodule
- use inreplace to work around undefined symbol "mp_thread_get_state"
- add "make axtls" to the install method


ilovezfs:
@DomT4 PR refreshed.

Also, any idea what's up with Jenkins?

MikeMcQuaid:
@ilovezfs Looking.

ilovezfs:
@DomT4 :ship:?

DomT4:
Seems reasonable 👍

PR #2613, opened 2016-07-02
pgrouting 2.2.3

ilovezfs:
- add head spec
- apply upstream commits adding two missing includes
- add linked-library dependency gmp


ilovezfs:
@DomT4 PR refreshed.

DomT4:
👍

PR #2615, opened 2016-07-02
pmd 5.5.0

ilovezfs:
- add Java requirement
- build now requires a toolchains.xml file

### DIFF
--- a/Formula/pmd.rb
+++ b/Formula/pmd.rb
@@ -1,8 +1,8 @@
 class Pmd < Formula
   desc "Source code analyzer for Java, JavaScript, and more"
   homepage "https://pmd.github.io"
-  url "https://github.com/pmd/pmd/releases/download/pmd_releases/5.3.3/pmd-src-5.3.3.zip"
-  sha256 "38004ea3274d2f71701f438606a4c4095b7a86110e8d7f2a8940170c9bd5ddbb"
+  url "https://github.com/pmd/pmd/releases/download/pmd_releases/5.5.0/pmd-src-5.5.0.zip"
+  sha256 "825d9061754cef30fc1b4584f509c7dd8a7b023a2c6ec264bf7b82bb0615c0b4"
 
   bottle do
     cellar :any_skip_relocation
@@ -11,10 +11,39 @@ class Pmd < Formula
     sha256 "e5ed61e35b5c980b0e1bcf616bc2615f1c5310ad66ce8f4901ca225764e7e78c" => :mavericks
   end
 
+  depends_on :java => "1.8+"
   depends_on "maven" => :build
 
   def install
-    ENV.java_cache
+    java_user_home = buildpath/"java_user_home"
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{buildpath}/java_user_home"
+    java_cache_repo = HOMEBREW_CACHE/"java_cache/.m2/repository"
+    java_cache_repo.mkpath
+    (java_user_home/".m2").install_symlink java_cache_repo
+
+    (java_user_home/".m2/toolchains.xml").write <<-EOS.undent
+      <?xml version="1.0" encoding="UTF8"?>
+      <toolchains>
+        <toolchain>
+          <type>jdk</type>
+          <provides>
+            <version>#{ENV["JAVA_HOME"][/((\d\.?)+)\.\d/, 1]}</version>
+          </provides>
+          <configuration>
+            <jdkHome>#{ENV["JAVA_HOME"]}</jdkHome>
+          </configuration>
+        </toolchain>
+        <toolchain>
+          <type>jdk</type>
+          <provides>
+            <version>1.7</version>
+          </provides>
+          <configuration>
+            <jdkHome>#{ENV["JAVA_HOME"]}</jdkHome>
+          </configuration>
+        </toolchain>
+      </toolchains>
+    EOS
 
     system "mvn", "clean", "package"
 


ilovezfs:
needs a java requirement, I guess, but not sure what version

DomT4:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-toolchains-plugin:1.1:toolchain (default) on project pmd: Cannot find matching toolchain definitions for the following toolchain types:
[ERROR] jdk [ version='1.7' ]
[ERROR] Please make sure you define the required toolchains in your ~/.m2/toolchains.xml file.
```
The upstream documentation isn't [much clearer](https://github.com/pmd/pmd#how-to-build-pmd).

ilovezfs:
@DomT4 PR refreshed.

ilovezfs:
:green_apple:

Formula/pmd.rb
```diff
@@ -11,11 +11,41 @@ class Pmd < Formula
...
   end
 
+  depends_on :java => "1.8+"
   depends_on "maven" => :build
 
   def install
     ENV.java_cache
 
+    java_version = ENV["JAVA_HOME"][/((\d\.?)+)\.\d/, 1]
+    java_user_home = ENV["_JAVA_OPTIONS"][/-Duser.home=(.*) ?/, 1]
```
DomT4: `ENV.java_cache` sets `-Duser.home` so it should always be `#{HOMEBREW_CACHE}/java_cache` here no?

ilovezfs: Right. The point is to avoid hard coding a string in two places if it's only necessary to hard code it in one place.

DomT4: Alright, fair enough 👍.

ilovezfs: @DomT4 simplified it to
```
toolchains = HOMEBREW_CACHE/"java_cache/.m2/toolchains.xml"
```

If we decide to change "java_cache", I'm sure we'll be grepping for the problem anyway.

Formula/pmd.rb
```diff
@@ -11,11 +11,40 @@ class Pmd < Formula
...
 
+  depends_on :java => "1.8+"
   depends_on "maven" => :build
 
   def install
     ENV.java_cache
 
+    java_version = ENV["JAVA_HOME"][/((\d\.?)+)\.\d/, 1]
+    toolchains = HOMEBREW_CACHE/"java_cache/.m2/toolchains.xml"
+    rm toolchains if File.exist? toolchains
```
ilovezfs: @UniqMartin @apjanke @DomT4 ideally I guess this would use some kind of locking but I'm not sure if we really need to worry about it. Any thoughts on whether the lack of safety here is a concern?

MikeMcQuaid: Why does this have to edit the cache?

ilovezfs: It's adding a file to the -Duser.home set by ENV.java_cache, which happens to be the folder HOMEBREW_CACHE/"java_cache".

ilovezfs: @MikeMcQuaid If we're worried about the lack of safefty, one possibility would be NOT to use ENV.java_cache and set a different -Duser.home that's in the buildpath (e.g., buildpath/".brew_home/java_user_home"). Frankly the fact that ENV.java_cache sets a user_home in the cache is quite problematic since it is totally unsafe for multiple brew prefixes to access the same user_home when it doesn't respect formula locking at all.

ilovezfs: Not to mention the fact that formula locking alone isn't sufficient since multiple formulae can use the same java cache (it's not segregated per formula) so it's already totally unsafe even if you assume a single prefix and it respected formula locking.

ilovezfs: For this formula, I think we should probably do this:
```
diff --git a/Formula/pmd.rb b/Formula/pmd.rb
index 45a8c01..80b39fd 100644
--- a/Formula/pmd.rb
+++ b/Formula/pmd.rb
@@ -15,13 +15,11 @@ class Pmd < Formula
   depends_on "maven" => :build
 
   def install
-    ENV.java_cache
+    ENV.append "_JAVA_OPTIONS", "-Duser.home=#{buildpath}/java_user_home"
 
     java_version = ENV["JAVA_HOME"][/((\d\.?)+)\.\d/, 1]
-    toolchains = HOMEBREW_CACHE/"java_cache/.m2/toolchains.xml"
-    rm toolchains if File.exist? toolchains
 
-    toolchains.write <<-EOS.undent
+    (buildpath/"java_user_home/.m2/toolchains.xml").write <<-EOS.undent
       <?xml version="1.0" encoding="UTF8"?>
       <toolchains>
         <toolchain>
```

ilovezfs: Except, I guess it needs to be `ENV["_JAVA_OPTIONS"] =` since there was talk of adding `ENV.java_cache` to everything.

ilovezfs:
PR refreshed.

ilovezfs:
@DomT4 :ship:?

DomT4:
We're overriding `ENV.java_cache` here but I think the positives in not needing to rm files in the cache outweighs the negative, so tentatively 👍.

ilovezfs:
@DomT4 what do you think of allowing ENV.java_cache to take an optional argument setting it to a custom directory, so we're not directly poking at _JAVA_OPTIONS to manually set user.home?

DomT4:
Would have to review the implementation, but in theory I think it's reasonable enough. Obvious problem would be people trying to set it to somewhere they're not allowed to write due to the sandbox.

ilovezfs:
It seems to generally crop up as an issue during test-do:
```
homebrew/homebrew-core/Formula/apache-flink.rb
18:    ENV["_JAVA_OPTIONS"] = "-Djava.io.tmpdir=#{testpath} -Duser.home=#{testpath}"

homebrew/homebrew-core/Formula/civl.rb
23:    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{testpath}"

homebrew/homebrew-core/Formula/jboss-forge.rb
17:    ENV["FORGE_OPTS"] = "-Duser.home=#{testpath}"

homebrew/homebrew-core/Formula/maven.rb
52:    system "#{bin}/mvn", "compile", "-Duser.home=#{testpath}"

homebrew/homebrew-versions/kettle54.rb
59:    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{testpath}"
```

MikeMcQuaid:
Ideally the home should point to the testpath and we set a more specific location for Maven's cache but I have no idea if that's possible.

ilovezfs:
I'm certainly confused by the naming "ENV.java_cache" when the variable being set is "user.home"

Are there also separate variables for temp and for cache?

ilovezfs:
@MikeMcQuaid Yeah what you're saying is possible

```
   | Default: ${user.home}/.m2/repository
  <localRepository>/path/to/local/repo</localRepository>
```

ilovezfs:
@DomT4 the whole thing is sufficiently confusing to me that I have no idea if it's a good idea, but this works if it's desirable:
```
diff --git a/Library/Homebrew/extend/ENV/shared.rb b/Library/Homebrew/extend/ENV/shared.rb
index 2debbae..bd756b6 100644
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -246,8 +246,9 @@ module SharedEnvExtension
     set_cpu_flags(flags)
   end
 
-  def java_cache
-    append "_JAVA_OPTIONS", "-Duser.home=#{HOMEBREW_CACHE}/java_cache"
+  def java_cache(path = nil)
+    path ||= HOMEBREW_CACHE/"java_cache"
+    append "_JAVA_OPTIONS", "-Duser.home=#{path}"
   end
 
   # ld64 is a newer linker provided for Xcode 2.5
```

and
```
-    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{buildpath}/java_user_home"
+    ENV.java_cache buildpath/"java_user_home"
```

ilovezfs:
Then I think this formula is fine as it currently is in the PR until `#{person}` cleans the Java situation up.

Formula/pmd.rb
```diff
@@ -11,10 +11,35 @@ class Pmd < Formula
     sha256 "e5ed61e35b5c980b0e1bcf616bc2615f1c5310ad66ce8f4901ca225764e7e78c" => :mavericks
   end
 
+  depends_on :java => "1.8+"
   depends_on "maven" => :build
 
   def install
-    ENV.java_cache
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{buildpath}/java_user_home"
```
MikeMcQuaid: Could we maybe make a symlink back to the `.m2/repository` cache directory so that it's used? For bonus points: maybe this is the only thing we should be putting in the cache with `ENV.java_cache` at all CC @DomT4 ?

DomT4:
> It's confusing as-is, i.e. that ENV.java_cache actually manipulates user.home and the effect on caching is only a side effect of that.

Yeah, in hindsight I should have named the variable `ENV.java_sandbox` or something. The purpose of `ENV.java_home` is dual:
* Stop Java things constantly violating the sandbox. Even simple Java executions can violate the sandbox because `.oracle_jre_usage` gets tracked. That isn't normally fatal but not infrequently if it can't write a configuration file it'll abort.

* Restore Java/Maven caching, which we smashed by implementing the sandbox. This is particularly useful because you don't have to repeatedly fetch hundreds of MBs of data for repeated `brew install` or `brew test` runs. This is a problem more for Maven stuff than Java.

ilovezfs:
@MikeMcQuaid regarding le symlink, how's that?

Formula/pmd.rb
```diff
@@ -11,10 +11,39 @@ class Pmd < Formula
...
+  depends_on :java => "1.8+"
   depends_on "maven" => :build
 
   def install
-    ENV.java_cache
+    java_user_home = buildpath/"java_user_home"
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{buildpath}/java_user_home"
+    java_cache_repo = HOMEBREW_CACHE/"java_cache/.m2/repository"
+    java_cache_repo.mkpath
+    (java_user_home/".m2").install_symlink java_cache_repo => "repository"
```
UniqMartin: You can omit the `=> "repository"` as the base names of the source and target of the symlink are identical.

ilovezfs:
@UniqMartin fixed.

UniqMartin:
> Yeah, in hindsight I should have named the variable `ENV.java_sandbox` or something. The purpose of `ENV.java_cache` is dual:

Yeah, in hindsight its always easier to spot the flaws. I'm not blaming you for the status quo! I'm just trying to avoid tucking more functionality onto it and making it even less comprehensible. How about we name the feature `ENV.java_home` once that day comes and `#{person}` implements a good replacement?

> regarding le symlink, how's that?

I think this is a wonderful solution. If we can get this cleaned up and generalized a bit so we can make this the implementation of `ENV.java_home` (just an idea for naming the replacement of `ENV.java_cache`) we could hide all those implementation details from unsuspecting contributors and would be able to start using it in all kinds of (Maven-using) Java formulae.

ilovezfs:
>If we can get this cleaned up

@UniqMartin Yeah this in particular is really gross:
```
ENV["_JAVA_OPTIONS"] = "-Duser.home=#{buildpath}/java_user_home"
```
It's a defensive measure in case user.home starts globally getting set to `HOMEBREW_CACHE/"java_cache`, but it would kill as collateral damage any other options superenv had set (we don't right now, but still ...).

DomT4:
> I'm not blaming you for the status quo! I'm just trying to avoid tucking more functionality onto it and making it even less comprehensible. How about we name the feature ENV.java_home once that day comes and #{person} implements a good replacement?

No worries, no offence taken. Just wanted to re-explain the dual purpose behind `ENV.java_cache` because when it's discussed one of the purposes tends to get forgotten discussing the other. I'm as guilty as anyone else on that.

Renaming it at some point in future makes sense to me; `ENV.java_home` is a good name, most people who know Java also know Java/Java-stuff does its caching primarily in the user's home as well, so that should be clear enough.

ilovezfs:
Actually I think "ENV.java_home" is quite problematic because we already define `ENV["JAVA_HOME"]` which means something else altogether.

UniqMartin:
After giving this some more thought, I agree. I think we need to make this change good enough that we don't have to call `ENV.<something>` at all. Let's just always set up a more useful Java home for every build and test run. This would also have the advantage that we won't expose yet another method to formula authors that we then have to support more or less indefinitely …

ilovezfs:
By the way, having the home directory during install be a separate directory .brew_home that is a subidrectory of the buildpath (rather than perhaps having home be the buildpath itself) is causing some problems because it means relocating all the contents of buildpath with buildpath.children, which I've done in some formulae, is actually unsafe because it drags the home directory along with it.

ilovezfs:
```
cayley.rb
24:    (buildpath/"src/github.com/google/cayley").install buildpath.children

consul-template.rb
36:    dir.install buildpath.children

gauge.rb
32:    dir.install buildpath.children

godep.rb
19:    (buildpath/"src/github.com/tools/godep").install buildpath.children

ipfs.rb
23:    (buildpath/"src/github.com/ipfs/go-ipfs").install buildpath.children

nomad.rb
19:    (buildpath/"src/github.com/hashicorp/nomad").install buildpath.children

syncthing-inotify.rb
21:    dir.install buildpath.children

vault.rb
38:    contents = buildpath.children - [buildpath/".brew_home"]

```

DomT4:
> Let's just always set up a more useful Java home for every build and test run.

Eventually it's supposed to be invisible as it is today for borrowing `ENV["HOME"]`.

Part of me is tempted to set it & see what breaks, but ho hum. The reason we hadn't so far is there are things known to break with it set, we're just unsure how deep that runs & whether there's ever _actually_ going to be progress on getting that fixed upstream.

> which I've done in some formulae, is actually unsafe because it drags the home directory along with it.

I'm pretty sure `.install` is a Pathname extension on our part? Could just teach it to not drag `.brew_home` with any `.install` call?

ilovezfs:
>I'm pretty sure .install is a Pathname extension on our part? Could just teach it to not drag .brew_home with any .install call

That feels like a hack, but it would indeed be better than the current situation.

UniqMartin:
> That feels like a hack, but it would indeed be better than the current situation.

Definitely feels like a hack. If we touch this code, we could as well change the code that sets up the temporary home directory and make it a sibling of the build path instead of making it a child. I don't think anything should be broken by this and nothing should be assuming the home directory is a subdirectory of `buildpath`, or am I missing something?

DomT4:
> nothing should be assuming the home directory is a subdirectory of buildpath, or am I missing something?

The important things are:

* It doesn't interfere with the user's own home directory.
* We toss it out at the end of the install/test process.

That's about it.

ilovezfs:
Do we even have a "magic pathname variable" for home or is it hard-coded `buildpath/".brew_home"` only right now?

DomT4:
[Install](https://github.com/Homebrew/brew/blob/4aedeea96d4c9d9c20bc822d520e453ac8964c56/Library/Homebrew/formula.rb#L1550-L1567) & [test](https://github.com/Homebrew/brew/blob/4aedeea96d4c9d9c20bc822d520e453ac8964c56/Library/Homebrew/formula.rb#L1325-L1344).

ilovezfs:
```
Josephs-MacBook-Pro:Formula joe$ ag .brew_home
groovyserv.rb
26:    ENV["GRADLE_USER_HOME"] = buildpath/".brew_home"

openshift-cli.rb
21:    (buildpath/".git/info/exclude").atomic_write "/.brew_home"

vault.rb
38:    contents = buildpath.children - [buildpath/".brew_home"]
```

UniqMartin:
So two of those instances you found are actually workarounds for the fact that `.brew_home` is in the build path instead of being somewhere else. Nice! More arguments that this is something worth improving.

ilovezfs:
@UniqMartin I think it would also allow us to relatively "safely" replace a lot of the `Dir["*"]` in formulae with `.children`

DomT4:
> So two of those instances you found are actually workarounds for the fact that .brew_home is in the build path instead of being somewhere else.

There's another in the main repo with our Node setup, FWIW.

ilovezfs:
I wonder if we can set the java_user_home = brew home, and always insert the symlink (or always insert the symlink when there's a :java requirement).

DomT4:
I'm not following.

Currently cooking, and cleaning, and following this thread, so bear with me, but we always want to dispose of whatever `ENV["HOME"]` is at the end of install/test, but if Java stuff is downloading things we _want_ to retain those until manual user intervention. Did you discover unique variables for Java's user home & where things are cached by Maven/Gradle/whatever?

ilovezfs:
The symlink I was referring to points from the transient java_user_home into the permanent-ish HOMEBREW_CACHE/java_cache/.m2/repository but I guess that only covers the maven case?

> discover unique variables for Java's user home & where things are cached 

At least for maven they are entirely separable, but it's specified in a configuration file not an environment variable. It can even be configured on a "profile" basis (so different locations for "dev", "test", etc.).

DomT4:
I'm reluctant to dive into theoreticals too much given we've looped over various versions in the last 4/5 months without actually moving anywhere _(Not a dig at you or anyone else there, I add!)_.

Essentially, I agree as long as we dispose of the temporary home after install/test & it exists somewhere we're happy for sandboxed processes to write to it could be useful to relocate `ENV["HOME"]` outside of the buildpath/testpath. Same with the Java situation as long as caching is preserved where appropriate.

Happy to review any proposed PRs in relation to that or the Java situation.

UniqMartin:
Yep, I think we basically agree and have come up with two completely orthogonal tasks:

- [ ] **(New)** Move `.brew_home` into a sibling directory of `buildpath` and `testpath` instead of making it a subdirectory. Make sure this still plays nice with the sandbox, it is still wiped when done, and that any formulae that rely on the current location are updated appropriately.
- [ ] **(Previously discussed but again refined here)** Deprecate `ENV.java_cache` in favor of always setting up `user.home` via `ENV["_JAVA_OPTIONS"]` and providing the necessary directory scaffold in the home directory (with symlinks pointing to appropriate locations in `HOMEBREW_CACHE`) so that cached downloads by Maven et al. can persist across build runs. (Possibly limit this setup to formulae that have a `:java` dependency.)

I'm equally happy to review PRs that address any of these, though I might end up implementing the first item myself sooner than later.

MikeMcQuaid:
@UniqMartin Agreed.

ilovezfs:
Shipping as-is. This can of course be updated if the way brew handles java user home changes.

DomT4:
👍

PR #2618, opened 2016-07-02
rainbarf 1.3

ilovezfs:
fix the test
set bottle unneeded
add head spec


ilovezfs:
@BrewTestBot test this please

ilovezfs:
Current version in core has the same problem https://bot.brew.sh/job/Homebrew%20Testing/1057/

ilovezfs:
OK, this only works on laptops since it's looking for battery info. The error only occurs on desktops where it's not useful.

ilovezfs:
Reported upstream https://github.com/creaktive/rainbarf/issues/30

PR #2629, opened 2016-07-02
tcpreplay 4.1.1

ilovezfs:
- remove tbd fix since it's been upstreamed
- require Xcode because it insists on using an SDK path
- set SDK to the one brew is actually using rather than letting it guess
- depend on libdnet unconditionally since configure offers no way to
  honor --without-libdnet if libdnet is installed

### DIFF
--- a/Formula/tcpreplay.rb
+++ b/Formula/tcpreplay.rb
@@ -1,8 +1,8 @@
 class Tcpreplay < Formula
   desc "Replay saved tcpdump files at arbitrary speeds"
   homepage "http://tcpreplay.appneta.com"
-  url "https://github.com/appneta/tcpreplay/releases/download/v4.1.0/tcpreplay-4.1.0.tar.gz"
-  sha256 "ad285b08d7a61ed88799713c4c5d657a7a503eee832304d3a767f67efe5d1a20"
+  url "https://github.com/appneta/tcpreplay/releases/download/v4.1.1/tcpreplay-4.1.1.tar.gz"
+  sha256 "61b916ef91049cad2a9ddc8de6f5e3e3cc5d9998dbb644dc91cf3a798497ffe4"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


DomT4:
The `inreplace` shouldn't be necessary any more? Think upstream merged a fix a while ago, which should be in this release... hopefully.

ilovezfs:
@DomT4 PR refreshed.

Formula/tcpreplay.rb
```diff
@@ -12,23 +12,19 @@ class Tcpreplay < Formula
     sha256 "ab0379428462fbdc2653feae2bcc404cf6b6d2129ddf0461019c53794ce87e4f" => :mavericks
   end
 
-  depends_on "libdnet" => :recommended
+  depends_on :xcode
```
DomT4: This needs tagging with either `=> :run` or `=> :build`. Is this new? Can't remember the need for a full Xcode in the past.

ilovezfs: I doubt it's new, but it hard fails with CLI only.

ilovezfs: It's always effectively been assumed by the formula because `"--with-libpcap=#{MacOS.sdk_path}/usr"` crashes if you don't have Xcode.

DomT4: Really? I thought `MacOS.sdk_path` was just `nil` when Xcode wasn't around, so it became effectively `--with-libpcap=/usr`. Was I sniffing something funny the day I thought that?

ilovezfs: No you're correct. It's MacOS.sdk.version that will die.

ilovezfs: The problem is that in 4.1.1, without Xcode, configure dies with
```
checking what Mac OS X SDK to use... configure: error: "Could not figure out the location of a Mac OS X SDK"
```

ilovezfs: @DomT4 OK, looks like this works fine since it's trying to set isysroot:
```
    if MacOS::Xcode.installed?
      args << "--with-macosx-sdk=#{MacOS.sdk.version}"
    else
      inreplace "configure",
        /^.*Could not figure out the location of a Mac OS X SDK.*$/,
        "MACOSX_SDK_PATH=/"
    end
```

ilovezfs:
@DomT4 PR refreshed.

DomT4:
👍 whenever the CI catches up to this PR & declares itself happy.

PR #2632, opened 2016-07-02
vegeta 6.1.0

ilovezfs:
- new go_resource "golang.org/x/net"
- avoid lines exceeding the eightieth column
- require successful exit statuses and a match in the test


ilovezfs:
@DomT4 :ship:?

DomT4:
Seems reasonable.

PR #2633, opened 2016-07-02
voldemort 1.10.18

ilovezfs:
build system now uses gradle instead of ant

### DIFF
--- a/Formula/voldemort.rb
+++ b/Formula/voldemort.rb
@@ -1,8 +1,8 @@
 class Voldemort < Formula
   desc "Distributed key-value storage system"
   homepage "http://www.project-voldemort.com/"
-  url "https://github.com/voldemort/voldemort/archive/release-1.9.5-cutoff.tar.gz"
-  sha256 "d521943873665d863a207c93557dc322c091e22a4293df4e3b86eb298ebbda7c"
+  url "https://github.com/voldemort/voldemort/archive/release-1.10.18-cutoff.tar.gz"
+  sha256 "a51fd1a53adb36a3fc70617ae5cd05ed31892a64074dfb612bb4a78aae28ecf6"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Maybe worth env scripting `VOLDEMORT_HOME` to the executables & ditching the caveat?

ilovezfs:
@DomT4 how's that?

DomT4:
If it works, and it seems to, 👍. Nice work.

ilovezfs:
@DomT4 Thanks. Also, working on the env scripts caused me to find a bug: the name of the environment variable for the config directory had changed, but I'd not noticed (not sure in which version that change actually occurred - the version in core right now may be broken).

DomT4:
Heh, if it turns out that change was made prior to this version it's quite telling in terms of how few people bother to read caveats that nobody has reported that to us.

ilovezfs:
Appears to have been VOLDEMORT_CONFIG_DIR since 2011 https://github.com/voldemort/voldemort/commit/903943ff3137ceb95566c2b5a5dfba0f52005aa4

Our VOL_CONF_DIR goes all the way back to when the formula was originally added https://github.com/Homebrew/homebrew-core/blob/00b63b132c7b4ae15528b303180e315449680b50/Formula/voldemort.rb but I can't find any reference to it in the current sources, so who knows when that was last a thing that worked!

It also makes me wonder if anyone actually uses this formula. lol

DomT4:
According to GA nobody is using this formula. According to Bintray, 24 downloads in the last year, including 8 today.

ilovezfs:
So 1/3 of a download a week, not including today. Well, that's not zero, I suppose. Ahem.

DomT4:
Not our most popular formula it's fair to say.

ilovezfs:
Perhaps we can have homebrew/core and homebrew/not-so-core.

UniqMartin:
I want to have `homebrew/turtles` and `homebrew/rabbits` for different release frequencies. :joy:

PR #2643, opened 2016-07-02
libimobiledevice, libplist: Add support for Python 3 bindings

bjoernhaeuser:

---

Current problem with `brew audit  --strict --online <formula>`:

```
libimobiledevice:
      /usr/local/Cellar/libimobiledevice/1.2.0/lib/python3.5/site-packages/imobiledevice.so
Error: 1 problem in 1 formula
```

Can anyone tell me how to fix this?

The corresponding PR for fixing the inreplace and patch is here: libimobiledevice/libimobiledevice#335

### DIFF
--- a/Formula/libimobiledevice.rb
+++ b/Formula/libimobiledevice.rb
@@ -20,21 +20,71 @@ class Libimobiledevice < Formula
     depends_on "libxml2"
   end
 
+  option "with-python", "Enable Cython Python bindings"
+  option "with-python3", "Enable Cython Python3 bindings"
+
+  libplist_depends = []
+  libplist_depends << "with-python" if build.with? "python"
+  libplist_depends << "with-python3" if build.with? "python3"
+
   depends_on "pkg-config" => :build
   depends_on "libtasn1"
-  depends_on "libplist"
   depends_on "usbmuxd"
   depends_on "openssl"
+  depends_on "libplist" => libplist_depends
+  depends_on :python => :optional
+  depends_on :python3 => :optional
+
+  resource "cython" do
+    url "https://github.com/cython/cython/archive/0.24.tar.gz"
+    sha256 "b60b91f1ec88921a423d5f0a5e2a7c232cdff12d9130088014bf89d542ce137b"
+  end
+
+  if build.with?("python") && build.with?("python3")
+    raise "Either python or python3 bindings can be installed at the same time."
+  end
+
+  patch do
+    # ensure python3 compatibility
+    url "https://github.com/libimobiledevice/libimobiledevice/pull/335.patch"
+    sha256 "ceb6076131ea204f411bf2d6eb584b83c300f2caf8def1c34a781920039e0063"
+  end
 
   def install
+    args = %W[
+      --disable-dependency-tracking
+      --disable-silent-rules
+      --prefix=#{prefix}
+    ]
+
+    if build.with? "python3"
+      version = Language::Python.major_minor_version "python3"
+      resource("cython").stage do
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python#{version}/site-packages"
+        system "python3", *Language::Python.setup_install_args(buildpath)
+      end
+      ENV.prepend_path "PATH", buildpath/"bin"
+
+      args << "PYTHON=python3"
+      args << "PYTHON_LDFLAGS=-undefined dynamic_lookup"
+    elsif build.with? "python"
+      resource("cython").stage do
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python2.7/site-packages"
+        system "python", *Language::Python.setup_install_args(buildpath)
+      end
+      ENV.prepend_path "PATH", buildpath/"bin"
+    else
+      args << "--without-cython"
+    end
+
     system "./autogen.sh" if build.head?
-    system "./configure", "--disable-dependency-tracking",
-                          "--disable-silent-rules",
-                          "--prefix=#{prefix}",
-                          # As long as libplist builds without Cython
-                          # bindings, libimobiledevice must as well.
-                          "--without-cython"
-    system "make", "install"
+    inreplace "cython/Makefile.in", "-no-undefined", ""
+    system "./configure", *args
+    system "make", "install", "PYTHON_LDFLAGS=-undefined dynamic_lookup"
+  end
+
+  test do
+    system "#{bin}/idevice_id"
   end
 
   test do
--- a/Formula/libplist.rb
+++ b/Formula/libplist.rb
@@ -13,39 +13,54 @@ class Libplist < Formula
   end
 
   head do
-    url "http://git.sukimashita.com/libplist.git"
-
+    url "https://git.libimobiledevice.org/libplist.git"
     depends_on "automake" => :build
     depends_on "autoconf" => :build
     depends_on "libtool" => :build
   end
 
   option "with-python", "Enable Cython Python bindings"
+  option "with-python3", "Enable Cython Python3 bindings"
 
   depends_on "pkg-config" => :build
   depends_on "libxml2"
   depends_on :python => :optional
+  depends_on :python3 => :optional
 
   resource "cython" do
     url "https://pypi.python.org/packages/c6/fe/97319581905de40f1be7015a0ea1bd336a756f6249914b148a17eefa75dc/Cython-0.24.1.tar.gz#md5=890b494a12951f1d6228c416a5789554"
     sha256 "84808fda00508757928e1feadcf41c9f78e9a9b7167b6649ab0933b76f75e7b9"
   end
 
+  if build.with?("python") && build.with?("python3")
+    raise "Either python or python3 bindings can be installed at the same time."
+  end
+
   def install
-    ENV.deparallelize
     args = %W[
       --disable-dependency-tracking
       --disable-silent-rules
       --prefix=#{prefix}
     ]
 
-    if build.with? "python"
+    if build.with? "python3"
+      version = Language::Python.major_minor_version "python3"
       resource("cython").stage do
-        ENV.prepend_create_path "PYTHONPATH", buildpath+"lib/python2.7/site-packages"
-        system "python", "setup.py", "build", "install", "--prefix=#{buildpath}",
-                 "--single-version-externally-managed", "--record=installed.txt"
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python#{version}/site-packages"
+        system "python3", *Language::Python.setup_install_args(buildpath)
       end
-      ENV.prepend_path "PATH", "#{buildpath}/bin"
+      ENV.prepend_path "PATH", buildpath/"bin"
+
+      args << "PYTHON=python3"
+      args << "PYTHON_LDFLAGS=-undefined dynamic_lookup"
+    elsif build.with? "python"
+      version = Language::Python.major_minor_version "python3"
+
+      resource("cython").stage do
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python#{version}/site-packages"
+        system "python", *Language::Python.setup_install_args(buildpath)
+      end
+      ENV.prepend_path "PATH", buildpath/"bin"
     else
       args << "--without-cython"
     end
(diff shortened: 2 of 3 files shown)


Formula/libimobiledevice.rb
```diff
@@ -20,20 +20,76 @@ class Libimobiledevice < Formula
...
   depends_on "libtasn1"
-  depends_on "libplist"
   depends_on "usbmuxd"
   depends_on "openssl"
+  depends_on "libplist" => (build.with?("python") ? "with-python" : []) || (build.with?("python3") ? "with-python3" : [])
+  depends_on :python => :optional
+  depends_on :python3 => :optional
+
+  resource "cython" do
+    url "http://cython.org/release/Cython-0.24.tar.gz"
```
dunn: This is failing: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4177/
And locally:
```
🌲  curl -IL http://cython.org/release/Cython-0.24.tar.gz
HTTP/1.1 404 Not Found
Server: GitHub.com
Date: Sun, 03 Jul 2016 19:43:00 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 9340
ETag: "5745e9c5-247c"
Access-Control-Allow-Origin: *
Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; img-src data:; connect-src 'self'
X-GitHub-Request-Id: 688317F6:270F:4678B70:57796AC4

**Comment by DomT4** [comment]
@tdsmith Can you do some sanity checking here? ❤️

File: Formula/libimobiledevice.rb [hunk 2]
```diff
@@ -20,20 +20,76 @@ class Libimobiledevice < Formula
...
+    sha256 "6de44d8c482128efc12334641347a9c3e5098d807dd3c69e867fa8f84ec2a3f1"
+  end
+
+  patch :DATA
```
DomT4: We'll need this reporting upstream, and the link to that bug/patch submission linked here.

Formula/libimobiledevice.rb
```diff
@@ -20,20 +20,76 @@ class Libimobiledevice < Formula
...
 
   def install
+    ENV.deparallelize
```
DomT4: This should also be reported upstream as a bug. Things _should_ build in parallel.

Formula/libimobiledevice.rb
bjoernhaeuser: Changed it to use the github url

Formula/libimobiledevice.rb
bjoernhaeuser: Added the existing PR as a comment

Formula/libimobiledevice.rb
bjoernhaeuser: Removed the line and it still builds.. :-)

Formula/libimobiledevice.rb
```diff
@@ -20,20 +20,77 @@ class Libimobiledevice < Formula
...
+  depends_on :python => :optional
+  depends_on :python3 => :optional
+
+  resource "cython" do
+    url "https://github.com/cython/cython/archive/0.24.tar.gz"
+    sha256 "b60b91f1ec88921a423d5f0a5e2a7c232cdff12d9130088014bf89d542ce137b"
+  end
+
+  # for reference https://github.com/libimobiledevice/libimobiledevice/pull/335
+  patch :DATA
```
DomT4: Now you've submitted upstream does it work if we use https://github.com/libimobiledevice/libimobiledevice/pull/335.patch in a `patch do` format?

Formula/libplist.rb
```diff
@@ -21,29 +21,41 @@ class Libplist < Formula
...
     ]
 
-    if build.with? "python"
+    if build.with? "python3"
+      version = Language::Python.major_minor_version "python3"
       resource("cython").stage do
-        ENV.prepend_create_path "PYTHONPATH", buildpath+"lib/python2.7/site-packages"
-        system "python", "setup.py", "build", "install", "--prefix=#{buildpath}",
-                 "--single-version-externally-managed", "--record=installed.txt"
+        ENV.prepend_create_path "PYTHONPATH", buildpath+"lib/python#{version}/site-packages"
```
DomT4: Nit but we're trying to stamp out unnecessary `+` usage in favour of `/`. Would you mind switching? 🙈

Formula/libplist.rb
```diff
@@ -21,29 +21,41 @@ class Libplist < Formula
...
+      end
+      ENV.prepend_path "PATH", "#{buildpath}/bin"
+
+      args << "PYTHON=python3"
+      args << "LDFLAGS=" + `python3-config --ldflags`.chomp
+    elsif build.with? "python"
+      version = Language::Python.major_minor_version "python3"
+
+      resource("cython").stage do
+        ENV.prepend_create_path "PYTHONPATH", buildpath+"lib/python#{version}/site-packages"
```
DomT4: Same comment here re `+` Vs `/`

Formula/libplist.rb
MikeMcQuaid: @DomT4 Are we? If so: `brew audit` calls you 😉

DomT4: Yeah, there was a mini discussion about it after the `audit` issue where the `pkgshare` audit was only bothering to check `/` rather than `+` and none of us noticed because most of our path creation is done using `/` now. So it makes sense to stop mixing & matching. `+` has a valid use case, but simple path creation should be adopting `/` ideally.

Formula/libimobiledevice.rb
bjoernhaeuser: Do you have an example which I could copy?

Formula/libplist.rb
bjoernhaeuser: I am not so much into ruby, can you give me an example which I can copy here?

bjoernhaeuser:
Changed all the path to use "/" and also using the PR as a patch source.

Formula/libimobiledevice.rb
```diff
@@ -13,27 +13,73 @@ class Libimobiledevice < Formula
...
+  depends_on :python3 => :optional
+
+  resource "cython" do
+    url "https://github.com/cython/cython/archive/0.24.tar.gz"
+    sha256 "b60b91f1ec88921a423d5f0a5e2a7c232cdff12d9130088014bf89d542ce137b"
+  end
+
+  patch do
+    # ensure python3 compatibility
+    url "https://github.com/libimobiledevice/libimobiledevice/pull/335.patch"
```
MikeMcQuaid: I'd like to see some movement on this before we include it.

Formula/libimobiledevice.rb
```diff
@@ -13,27 +13,73 @@ class Libimobiledevice < Formula
...
+      version = Language::Python.major_minor_version "python3"
+      resource("cython").stage do
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python#{version}/site-packages"
+        system "python3", *Language::Python.setup_install_args(buildpath)
+      end
+      ENV.prepend_path "PATH", buildpath/"bin"
+
+      args << "PYTHON=python3"
+      args << "LDFLAGS=" + `python3-config --ldflags`.chomp
+    elsif build.with? "python"
```
MikeMcQuaid: What if both are specified? Should probably handle that case.

MikeMcQuaid: Could `raise`

Formula/libplist.rb
```diff
@@ -13,45 +13,62 @@ class Libplist < Formula
...
+      version = Language::Python.major_minor_version "python3"
+      resource("cython").stage do
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python#{version}/site-packages"
+        system "python3", *Language::Python.setup_install_args(buildpath)
+      end
+      ENV.prepend_path "PATH", buildpath/"bin"
+
+      args << "PYTHON=python3"
+      args << "LDFLAGS=" + `python3-config --ldflags`.chomp
+    elsif build.with? "python"
```
MikeMcQuaid: Need to handle e.g. `raise` if both are passed.

Formula/libimobiledevice.rb
bjoernhaeuser: Do you have an example where I can copy this from? Do not want to come up with my own implementation, which does not fulfill all the requirements

Formula/libimobiledevice.rb
bjoernhaeuser: I do not expect anything happening there, the maintainers are not very active on github. Should we close this PR here because of this?

Formula/libimobiledevice.rb
ilovezfs: See the following formulae: boost, bullet, kibana, poppler

bjoernhaeuser: Added a raise when python and python3 bindings are both provided

Formula/libplist.rb
bjoernhaeuser: Also added this here!

bjoernhaeuser:
Rebased onto current master.

Formula/libimobiledevice.rb
tdsmith: There've been a couple of recent commits; maybe you can ask them about it on IRC.

Formula/libimobiledevice.rb
```diff
@@ -20,21 +20,71 @@ class Libimobiledevice < Formula
...
+    if build.with? "python3"
+      version = Language::Python.major_minor_version "python3"
+      resource("cython").stage do
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python#{version}/site-packages"
+        system "python3", *Language::Python.setup_install_args(buildpath)
+      end
+      ENV.prepend_path "PATH", buildpath/"bin"
+
+      args << "PYTHON=python3"
+      args << "LDFLAGS=" + `python3-config --ldflags`.chomp
```
tdsmith: This is where the explicit linkage is coming from; you probably don't need or want this line. If it fails without it, try `-undefined dynamic_lookup` instead of shelling out to python3-config.

bjoernhaeuser: That gives me:

```
configure: error: unrecognized option: `-undefined dynamic_lookup'
Try `./configure --help' for more information
```

tdsmith: Er, sorry, `LDFLAGS=-undefined dynamic_lookup`.

bjoernhaeuser: Hm, I thought I tried that 😄 nonetheless, i added them right here and getting this error:

```
./configure
--disable-dependency-tracking
--disable-silent-rules
--prefix=/usr/local/Cellar/libimobiledevice/1.2.0
PYTHON=python3
LDFLAGS=-undefined dynamic_lookup
```

```
Last 15 lines from /Users/bjoernhaeuser/Library/Logs/Homebrew/libimobiledevice/02.configure:
checking for Python site-packages path... /usr/local/Cellar/python3/3.5.2_1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages
checking python extra libraries...  -ldl  -framework CoreFoundation
checking python extra linking flags... -u _PyMac_Error /usr/local/opt/python3/Frameworks/Python.framework/Versions/3.5/Python
checking consistency of all components of python development environment... no
configure: error:
  Could not link test program to Python. Maybe the main Python library has been
  installed in some non-standard library path. If so, pass it to configure,
  via the LDFLAGS environment variable.
  Example: ./configure LDFLAGS="-L/usr/non-standard-path/python/lib"
  ============================================================================
   ERROR!
   You probably have to install the development version of the Python package
   for your distribution.  The exact name of this package varies among them.
  ============================================================================
```

tdsmith: The failing check is here: https://github.com/libimobiledevice/libimobiledevice/blob/master/m4/ac_python_devel.m4#L231

Hmm. Maybe if you set PYTHON_LDFLAGS?

bjoernhaeuser: That did the trick! Changed and the audit runes fine!

bjoernhaeuser:
Rebased onto current master.

MikeMcQuaid:
Closing but please reopen when/if upstream accepts the patch. Thanks!

PR #2646, opened 2016-07-02
amazon-ecs-cli 0.3.1

ilovezfs:
- remove unnecessary Language::Go.stage_deps
- run make test
- clean up the formula

### DIFF
--- a/Formula/amazon-ecs-cli.rb
+++ b/Formula/amazon-ecs-cli.rb
@@ -1,10 +1,8 @@
-require "language/go"
-
 class AmazonEcsCli < Formula
   desc "CLI for Amazon ECS to manage clusters and tasks for development."
   homepage "https://aws.amazon.com/ecs"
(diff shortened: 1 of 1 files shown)


DomT4:
Not sure what's going on with this formula.

I can't see a `vendor` or `godep` folder upstream, but this formula has no `go_resource`s but does have `Language::Go.stage_deps resources, buildpath/"src"` and the consequent `require`.

ilovezfs:
It has always been that way apparently: https://github.com/Homebrew/legacy-homebrew/pull/45053

@DomT4 I wonder if we should make this an error not a warning: 
```
Warning: tried to stage empty resources array
```
When is that ever acceptable?

DomT4:
It should probably be a hard `raise` when `HOMEBREW_DEVELOPER` is set & a warning (as currently) otherwise.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Yeah, the vendoring system is there, I just missed it because it's nested. There's a new new version out, above this one, but that aside 👍.

PR #2659, opened 2016-07-02
no-more-secrets 0.2.0

ilovezfs:
- fix -lncursesw regression with inreplace
- apply upstream commit to fix the version
- use make install
- match version in the test using new upstream -v CLI option


DomT4:
Sigh, this old chestnut:
```
gcc -Wextra -Wall -o bin/nms obj/nms.o obj/main.o -lncursesw
gcc -Wextra -Wall -o bin/sneakers obj/nms.o obj/sneakers.o -lncursesw
ld: library not found for -lncursesw
ld: library not found for -lncursesw
```

You can `inreplace` it out to `-lncurses` & see if upstream will make it configurable generally, if they haven't already.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
👍

PR #2665, opened 2016-07-03
luaradio 0.1.2 (new formula)

muellermartin:

---

`brew audit --strict --online luaradio` complains about "GitHub repository not notable enough" but I guess that might change ;)

### DIFF
--- a/Formula/luaradio.rb
+++ b/Formula/luaradio.rb
@@ -40,7 +40,6 @@ class Luaradio < Formula
       top:run()
     EOS
 
-    assert_equal "Hello, world!",
-      shell_output("#{bin}/luaradio test.lua")
+    assert_equal "Hello, world!", shell_output("#{bin}/luaradio test.lua")
   end
 end
--- a/Formula/luaradio.rb
+++ b/Formula/luaradio.rb
@@ -3,6 +3,7 @@ class Luaradio < Formula
   homepage "http://luaradio.io/"
   url "https://github.com/vsergeev/luaradio/archive/v0.1.0.tar.gz"
   sha256 "4423f94c0bb259952eae7db11e4db4509dfa7c6e205798c35a095050c530212b"
+  head "https://github.com/vsergeev/luaradio.git"
 
   depends_on "pkg-config" => :build
   depends_on "luajit"
--- a/Formula/luaradio.rb
+++ b/Formula/luaradio.rb
@@ -1,8 +1,8 @@
 class Luaradio < Formula
   desc "lightweight, embeddable flow graph signal processing framework for SDR"
   homepage "http://luaradio.io/"
-  url "https://github.com/vsergeev/luaradio/archive/v0.1.0.tar.gz"
-  sha256 "4423f94c0bb259952eae7db11e4db4509dfa7c6e205798c35a095050c530212b"
+  url "https://github.com/vsergeev/luaradio/archive/v0.1.2.tar.gz"
+  sha256 "344cf0cd3799ca2dfbf8206952797f304e74c9456fe8b4061ab0377a44e9e2e5"
   head "https://github.com/vsergeev/luaradio.git"
 
   depends_on "pkg-config" => :build
--- a/Formula/luaradio.rb
+++ b/Formula/luaradio.rb
@@ -1,0 +1,47 @@
+class Luaradio < Formula
+  desc "lightweight, embeddable flow graph signal processing framework for SDR"
+  homepage "http://luaradio.io/"
+  url "https://github.com/vsergeev/luaradio/archive/v0.1.0.tar.gz"
+  sha256 "4423f94c0bb259952eae7db11e4db4509dfa7c6e205798c35a095050c530212b"
+
+  depends_on "pkg-config" => :build
+  depends_on "luajit"
+  depends_on "fftw" => :recommended
+
+  def install
+    cd "embed" do
+      system "make", "install", "PREFIX=#{prefix}"
+    end
+  end
+
+  test do
+    # Create text file which will be read and processed by LuaRadio
+    (testpath/"hello").write("Hello, world!")
+    # Read from file, convert byte stream to chars and output them
+    (testpath/"test.lua").write <<-EOS.undent
+      local radio = require('radio')
+
+      local PrintBytes = radio.block.factory("PrintBytes")
+
+      function PrintBytes:instantiate()
+          self:add_type_signature({radio.block.Input("in", radio.types.Byte)}, {})
+      end
+
+      function PrintBytes:process(c)
+          for i = 0, c.length - 1 do
+              io.write(string.char(c.data[i].value))
+          end
+      end
+
+      local source = radio.RawFileSource("hello", radio.types.Byte, 1e6)
+      local sink = PrintBytes()
+      local top = radio.CompositeBlock()
+
+      top:connect(source, sink)
+      top:run()
+    EOS
+
+    assert_equal "Hello, world!",
+      shell_output("#{bin}/luaradio test.lua")
+  end
+end
--- a/Formula/luaradio.rb
+++ b/Formula/luaradio.rb
@@ -16,9 +16,7 @@ class Luaradio < Formula
   end
 
   test do
-    # Create text file which will be read and processed by LuaRadio
     (testpath/"hello").write("Hello, world!")
-    # Read from file, convert byte stream to chars and output them
     (testpath/"test.lua").write <<-EOS.undent
       local radio = require('radio')
 
--- a/Formula/luaradio.rb
+++ b/Formula/luaradio.rb
@@ -11,6 +11,13 @@ class Luaradio < Formula
 
   def install
     cd "embed" do
+      # Ensure file placement is compatible with HOMEBREW_SANDBOX.
+      inreplace "Makefile" do |s|
+        s.gsub! "install -d $(DESTDIR)$(INSTALL_CMOD)",
+                "install -d $(PREFIX)/lib/lua/5.1"
+        s.gsub! "$(DESTDIR)$(INSTALL_CMOD)/radio.so",
+                "$(PREFIX)/lib/lua/5.1/radio.so"
+      end
       system "make", "install", "PREFIX=#{prefix}"
     end
   end


muellermartin:
Sadly the checks of @BrewTestBot only failed because of `GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)` (see [logs e.g. for El Capitan](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4202/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_luaradio___strict___online/)) but funnily enough meanwhile it got 149 stars (propably because it was [featured on Hacker News](https://news.ycombinator.com/item?id=12023407)).

muellermartin:
@BrewTestBot Please test this again.

DomT4:
Looks like there's a new upstream release already.

Formula/luaradio.rb
```diff
@@ -0,0 +1,47 @@
...
+  url "https://github.com/vsergeev/luaradio/archive/v0.1.0.tar.gz"
+  sha256 "4423f94c0bb259952eae7db11e4db4509dfa7c6e205798c35a095050c530212b"
+
+  depends_on "pkg-config" => :build
+  depends_on "luajit"
+  depends_on "fftw" => :recommended
+
+  def install
+    cd "embed" do
+      system "make", "install", "PREFIX=#{prefix}"
```
DomT4: Does this need to be built from a git checkout to get [accurate version output](https://github.com/vsergeev/luaradio/blob/3fcc8056ef03ce6b5292edeff43b92d6aa3d0e85/embed/Makefile#L40-L46)?

Formula/luaradio.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  test do
+    # Create text file which will be read and processed by LuaRadio
```
DomT4: Don't need this comment.

Formula/luaradio.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      local source = radio.RawFileSource("hello", radio.types.Byte, 1e6)
+      local sink = PrintBytes()
+      local top = radio.CompositeBlock()
+
+      top:connect(source, sink)
+      top:run()
+    EOS
+
+    assert_equal "Hello, world!",
```
DomT4: Unless this exceeds 80 characters you can plonk it on the same line rather than splitting it across two.

Formula/luaradio.rb
```diff
@@ -0,0 +1,47 @@
+class Luaradio < Formula
+  desc "lightweight, embeddable flow graph signal processing framework for SDR"
+  homepage "http://luaradio.io/"
+  url "https://github.com/vsergeev/luaradio/archive/v0.1.0.tar.gz"
+  sha256 "4423f94c0bb259952eae7db11e4db4509dfa7c6e205798c35a095050c530212b"
```
rxseger: To allow easily building the latest version (with brew install luaradio --HEAD), can you add:

    head "https://github.com/vsergeev/luaradio.git"

muellermartin:
@DomT4 I know, but I was not able to successfully build it.

Formula/luaradio.rb
muellermartin: Not really, because if the git command fails it falls back to `0.0.0` and for the Mac OS X build it only relies on the `VERSION_MAJOR` part which by chance is still `0` for this release. For releases greater than major version 0 this will break therefore this might be a bug (I guess the fallback should contain the current release version).
Fortunately if I run `luaradio --version` I get the right version (`0.1.0`). But this version number is retrieved from `radio/init.lua`.

Formula/luaradio.rb
muellermartin: Why?

MikeMcQuaid: Because it's self-evident.

Formula/luaradio.rb
```diff
@@ -0,0 +1,47 @@
...
+  def install
+    cd "embed" do
+      system "make", "install", "PREFIX=#{prefix}"
+    end
+  end
+
+  test do
+    # Create text file which will be read and processed by LuaRadio
+    (testpath/"hello").write("Hello, world!")
+    # Read from file, convert byte stream to chars and output them
```
MikeMcQuaid: Similarly this comment

Formula/luaradio.rb
muellermartin: The `brew test` command did not complain even in strict mode about that line(s), and I did that because a few other packages (e.g. cdiff, ispell, mksh, zsh) did the same thing (I'm no Ruby programmer therefore I had to look for examples).

Formula/luaradio.rb
muellermartin: Feel free to remove it, if it bothers you.

Formula/luaradio.rb
muellermartin: Similarly feel to remove it.

Formula/luaradio.rb
MikeMcQuaid: This is a pull request so we make comments and ask you to remove it for us.

muellermartin: I know and I kindly refuse to do so because I disagree.

MikeMcQuaid: Please close this PR if you don't want to address our comments.

Formula/luaradio.rb
MikeMcQuaid: It's a stylistic preference; `brew test` won't complain.

Formula/luaradio.rb
muellermartin: Are you serious? Only because of a comment?

MikeMcQuaid: I'm serious. Not because of a comment but because of the process. The way Homebrew PRs work is we request you make changes and you either make our changes as requested or close the PR.

muellermartin: > The way Homebrew PRs work is we request you make changes and you either make our changes as requested or close the PR.

Where's that written? I did not find that in the contribution guideline or the code of conduct.

I included that comment because this was the first time I used LuaRadio (which might result in poor code quality) and processing text files is rather unusual for a software defined radio library/tool but I felt it might be a good simple test. The comments where placed to make that somewhat clear and therefore an extra file is needed.

Honestly I feel a bit upset that I have to justify myself for my opinion on such a trivial issue and I'm again disappointed that I'm getting encountered in such a rough way as to me it seems like your opinion the only one that matters.

I like to enable new tools to be available through Homebrew but this nitpicking on trivial, non-blocking issues bothers me almost every time in a way that I want to quit doing that.

MikeMcQuaid: > Where's that written? I did not find that in the contribution guideline or the code of conduct.

It's not written; it goes without saying on pretty much any project on GitHub (or open-source in general) that if changes are requested in a PR and you aren't willing to make them: your PR will not be merged.



> The comments where placed to make that somewhat clear and therefore an extra file is needed.

Sure but for those who maintain Homebrew these comments are self-evident from the code. Comments should describe why something is done, not describe what the code does.



> Honestly I feel a bit upset that I have to justify myself for my opinion on such a trivial issue and I'm again disappointed that I'm getting encountered in such a rough way as to me it seems like your opinion the only one that matters.

Well, two maintainers have asked you to do this and, when it comes to maintaining software in Homebrew I'm afraid our opinions do matter more than yours do. We've been running this project (some would say successfully) for a long time and you're not being treated differently to any other PR.



> I like to enable new tools to be available through Homebrew but this nitpicking on trivial, non-blocking issues bothers me almost every time in a way that I want to quit doing that.

I also like new tools being available. It is a blocking issue to this PR being merged otherwise we wouldn't have asked you to do it. Our review process is thorough because when things are merged we're responsible for maintaining them and when we allow bad code into Homebrew it lowers the overall standard.

I hope that adds some explanation for this. If that's sufficient, please remove the comment. If it's not, please close this PR or indicate you'd like us to close it. Thanks!

muellermartin: Thanks for your explanation. I've removed the comments and included the other suggestions.

Formula/luaradio.rb
DomT4: > But this version number is retrieved from radio/init.lua.

As long as the version is accurate, that's fine.

DomT4:
> @DomT4 I know, but I was not able to successfully build it.

Can you elaborate?

muellermartin:
I just tried to update the version, but the install failed due to some incorrect paths (not available etc.). I had no time to take a closer look, but I suspect [this commit](https://github.com/vsergeev/luaradio/commit/3fcc8056ef03ce6b5292edeff43b92d6aa3d0e85) to be the cause: I think pkg-config was added and because it's not yet a build dependency it fails to get the correct path for LuaJit.

vsergeev:
@muellermartin, it may have to do with the extra `$(PREFIX)` on line [117](https://github.com/vsergeev/luaradio/commit/3fcc8056ef03ce6b5292edeff43b92d6aa3d0e85#diff-dc3974ca67d9912303e94d0d11d19f2eR117) that didn't get updated in the commit -- that path should just be `$(DESTDIR)$(INSTALL_CMOD)`. Could you confirm that this fixes the install?

If we can get this squared away on OS X,  I can release 0.1.2 with the fix, and the install should not have to change much moving forward.

muellermartin:
@vsergeev Thanks for your cool project! And thanks for your suggestion which seems to work for me. Did you follow the [conversation on the version number](https://github.com/Homebrew/homebrew-core/pull/2665#discussion_r69501292) above? There seems to be a small "bug" in the Makefile which falls back to version "0.0.0" - currently this is no issue on Mac OS as long the major version is smaller than 1, but on other systems this might result in a wrong version suffix on the library files. Should I file a bug report?

vsergeev:
@muellermartin, thanks and I see. The 0.0.0 fallback was to support building from out of the git tree, as currently the C API derives its version from the git tag. It really shouldn't be packaged or released as 0.0.0, as the C API would report the wrong version.

It sounds like it may be best change the strategy for determining the version number to pulling it from the Lua `radio` package instead of git.

vsergeev:
@muellermartin, I've committed the two fixes -- the line 117 `$(PREFIX)` fix, as well as version scraping from the radio package. Would you mind giving the `next` branch a try?

muellermartin:
@vsergeev Thanks for these fixes! I've cecked them out from the `next` branch and they seem to work.

vsergeev:
@muellermartin Great! I've released [v0.1.2](https://github.com/vsergeev/luaradio/blob/master/ChangeLog.md), which has those fixes merged in. This release should cooperate better with the brew formula.

muellermartin:
Thanks again! I've updated the formula to version 0.1.2 and for me everything seems to work so far.

muellermartin:
Why does the current commit (d6c7529) fail with this error on all systems (El Capitan):

    install: mkdir /usr/local/Cellar/luajit/2.0.4_1/lib/lua: Operation not permitted

See: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4660/

Shouldn't this directory be read-/writeable by Homebrew?

DomT4:
> Shouldn't this directory be read-/writeable by Homebrew?

It is, we've a build sandbox in place that prevents formula from writing wherever they want to though.

Partly for security, but also partly because Homebrew isn't really designed with the intention of unique software sharing the same keg - It conflicts with our symlink-based design, as well as generally causing user confusion when they `brew uninstall luajit` and it removes something placed there by another piece of software.

It's more or less a philosophical clash. Happens sometimes, naturally, given the scope of a package manager.

The problem is that in upstream's embed `Makefile` it calls:
```make
INSTALL_CMOD = $(shell pkg-config luajit --variable INSTALL_CMOD)
```
Which presuming Homebrew is in the usual place returns:
```
/usr/local/Cellar/luajit/2.0.4_1/lib/lua/5.1
```
Which isn't helpful to either of us, really. I'd have looked into fixing that before if I'd known it existed; to my knowledge this is the first time `INSTALL_CMOD` has been encountered being used by _another formula_ in Homebrew.

vsergeev:
@DomT4, @muellermartin, if the brew luajit installation searches `/usr/local/lib/lua/5.1` for C modules, then I could modify the upstream Makefile to simply install the symlink in`$(PREFIX)/lib/lua/5.1`, which I believe would be compatible with both Linux and brew installs. Unfortunately, I'm not aware of any convention for the Lua C module install path (and was hoping the pkg-config flag could solve that), but perhaps using `$(PREFIX)/` is the closest thing.

DomT4:
> if the brew luajit installation searches /usr/local/lib/lua/5.1 for C modules

It doesn't yet, but I'm planning to make it do so. It's fairly obvious the `lua` family, inc `luajit`, expects both `CMOD` & `LMOD` to point to a permanent location rather than a transient one. I'm only surprised it's taken this long for us to discover really.

IRT upstream fixes, I don't necessary expect you to bend much for us on this. We're making a habit out of rewriting elements to install within the Cellar & then symlink elsewhere to be compatible with both our build sandbox & upstream placement expectations; we more or less accepted doing that when we decided to implement the sandbox.

```ruby
  def install
    cd "embed" do
      # Ensure file placement is compatible with HOMEBREW_SANDBOX.
      inreplace "Makefile" do |s|
        s.gsub! "install -d $(DESTDIR)$(INSTALL_CMOD)",
                "install -d $(PREFIX)/lib/lua/5.1"
        s.gsub! "$(DESTDIR)$(INSTALL_CMOD)/radio.so",
                "$(PREFIX)/lib/lua/5.1/radio.so"
      end
      system "make", "install", "PREFIX=#{prefix}"
    end
  end
```
That sort of thing would not be the end of the world for us to keep.

muellermartin:
Wow, thanks everybody for putting so much effort into making Homebrew better!

@DomT4 I've tested your changes to lua/luajit by reinstalling the luaradio formula and I can confirm that the lua module is now installed to `/usr/local/lib/lua/5.1`. Could you trigger a new test for @BrewTestBot to validate this on other systems in your sandbox? If I understand this correctly and the tests succeed this means the package is ready to be shipped, right?

DomT4:
@muellermartin You'll need to adopt my proposed block of ruby for `def install` above to be sandbox-compatible _(It's not on by default for users yet)_, but IMO after that if the CI is happy we should be 💯.

muellermartin:
@DomT4 Added your proposed addition in latest commit 6dcf074.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/03ae50075180cd98f20363b1e1ca3347d1b89a33. Thank you for your contribution to Homebrew @muellermartin; we appreciate it! :smiley_cat:

muellermartin:
Awesome! Thank you guys from Homebrew for all of your efforts, too! And of course @vsergeev for creating luaradio and helping here :)

PR #2670, opened 2016-07-03
valgrind: fix assertion failure on 10.11

tdsmith:


Formula/valgrind.rb
```diff
@@ -5,6 +5,11 @@ class Valgrind < Formula
   stable do
     url "http://valgrind.org/downloads/valgrind-3.11.0.tar.bz2"
     sha256 "6c396271a8c1ddd5a6fb9abe714ea1e8a86fce85b30ab26b4266aeb4c2413b42"
+
+    # Fix tst->os_state.pthread - magic_delta assertion failure on OSX 10.11
+    # https://bugs.kde.org/show_bug.cgi?id=354883
+    # https://github.com/liquid-mirror/valgrind/commit/8f0b10fdc795f6011c17a7d80a0d65c36fcb8619.diff
```
DomT4: Does this not apply cleanly directly?

tdsmith: Dunno; it's not an official mirror, though, so I didn't want to link it directly in case it vanishes.

DomT4: Fair enough. If you're feeling particularly kind could dump this in `formula-patches` but otherwise 👍. People keep accidentally changing inline patch whitespace & breaking things.

DomT4:
What's the situation here? PR hasn't been updated in > a month.

tdsmith:
@BrewTestBot test this please

tdsmith:
Looks like this fix is still recommended upstream; thanks for the poke. Note that the `needs response` label is not removed after comments by Homebrew team members.

Formula/valgrind.rb
MikeMcQuaid: 👍 to `formula-patches`

zmwangx:
@tdsmith This patch doesn't seem to apply?

zmwangx:
The problem has been resolved.

PR #2672, opened 2016-07-03
i2pd 2.8.0 (new formula)

magicpriest:
- [+] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [+] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [+] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [+] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

Adding formula for i2p network client, written in C++.

### DIFF
--- a/Formula/i2pd.rb
+++ b/Formula/i2pd.rb
@@ -1,0 +1,23 @@
+class I2pd < Formula
+  desc "Full-featured C++ implementation of I2P client."
+  homepage "http://i2pd.website/"
+  head "https://github.com/PurpleI2P/i2pd.git"
+
+  stable do
+    url "https://github.com/PurpleI2P/i2pd/archive/2.8.0.tar.gz"
+    sha256 "3f3f538b3c5b3095021311b434456c42d6f3b5e43dc7972c263b10ef179ba063"
+  end
+
+  depends_on "libressl"
+  depends_on "boost"
+
+  def install
+    system "make", "HOMEBREW=1"
+
+    bin.install "i2pd"
+  end
+
+  test do
+    system "#{bin}/i2pd", "--help"
+  end
+end


Formula/i2pd.rb
```diff
@@ -0,0 +1,23 @@
+class I2pd < Formula
+  desc "Full-featured C++ implementation of I2P client."
+  homepage "http://i2pd.website/"
+  head "https://github.com/PurpleI2P/i2pd.git"
+
+  stable do
+    url "https://github.com/PurpleI2P/i2pd/archive/2.8.0.tar.gz"
```
dunn: If there aren't any stable-specific patches or requirements, these lines can be outside

Formula/i2pd.rb
```diff
@@ -0,0 +1,23 @@
...
+    sha256 "3f3f538b3c5b3095021311b434456c42d6f3b5e43dc7972c263b10ef179ba063"
+  end
+
+  depends_on "libressl"
+  depends_on "boost"
+
+  def install
+    system "make", "HOMEBREW=1"
+
```
dunn: This empty line can be removed.

Formula/i2pd.rb
```diff
@@ -0,0 +1,23 @@
...
+    bin.install "i2pd"
```
dunn: No `make install`?

Formula/i2pd.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  test do
+    system "#{bin}/i2pd", "--help"
```
dunn: Is it possible for a better test?  Good examples are [`cmake.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/cmake.rb) and [`pandoc.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pandoc.rb).

magicpriest:
@dunn 
I'll resolve all issues you mentioned during next day and write a new detailed test from scratch based on you suggestions.

Formula/i2pd.rb
magicpriest: @dunn 
I've run through Make file and the only option to build sources on OSX is set `make HOMEBREW=1`. On this scenario sources builds for OSX and binary file works perfectly. 
Although I've trying to build sources with `make install` and compilation fails with message "No target for 'install'", I suppose that author want to separate source building process and create this rule. So that is really a problem? Because if it is a problem I'll try to suggest for author another scenario.

dunn:
Thanks!

DomT4:
Perhaps worth noting the upstream `Makefile` for Homebrew isn't portable. It [presumes](https://github.com/PurpleI2P/i2pd/blob/b668c4c302a8ba69234e067c061f9607353c5fc5/Makefile.homebrew#L2) Homebrew will always exist in `/usr/local`, which whilst is the case most of the time isn't mandatory.

Formula/i2pd.rb
dunn: It's not a big problem, just something that would be nice to have.

DomT4:
Closing as dead here since there's been no update after July 5th. Happy to take/review a fresh PR with the various issues raised fixed as desired. Thank you nonetheless for your submission to Homebrew @magicpriest ❤️.

gwerbin:
What was actually needed here?

It looks like it just needs something like

```ruby
system("make", "HOMEBREW=1", "BREWROOT=#{HOMEBREW_PREFIX}",
       "BOOSTROOT=#{Formula['boost'].opt_prefix}",
       "SSLROOT=#{Formula['libressl'].opt_prefix}")
```

Plus a better `test do` and removing the unnecessary `stable do`.

PR #2706, opened 2016-07-04
cayley 0.5.0

ilovezfs:
- use godep instead of go_resource blocks
- set the version in ldflags
- match version in the test

### DIFF
--- a/Formula/cayley.rb
+++ b/Formula/cayley.rb
@@ -1,12 +1,8 @@
-require "language/go"
-
 class Cayley < Formula
   desc "Graph database inspired by Freebase and Knowledge Graph"
   homepage "https://github.com/google/cayley"
-  url "https://github.com/google/cayley/archive/v0.4.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/cayley.rb
```diff
@@ -16,124 +12,39 @@ class Cayley < Formula
...
-    # Create sample configuration that uses the Homebrew-based directories
-    inreplace "cayley.cfg.example", "/tmp/cayley_test", "#{var}/cayley"
-
-    # Install binary and configuration
-    bin.install "cayley"
-    etc.install "cayley.cfg.example" => "cayley.conf"
+    (buildpath/"src/github.com/google/cayley").install buildpath.children
+    cd "src/github.com/google/cayley" do
+      system "godep", "restore"
+      system "go", "build", "-ldflags=-X main.Version=0.5.0", "cmd/cayley/cayley.go"
```
DomT4: Maybe worth making this `main.Version=#{version}`?

DomT4:
One nit but 👍.

Formula/cayley.rb
ilovezfs: weird. thought that's what I had there.

PR #1663, opened 2016-06-02
alot 0.3.7 (new formula)

dlh:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

PR #2718, opened 2016-07-04
notmuch: enable python bindings by default

dlh:

---

See [here](https://github.com/Homebrew/homebrew-core/pull/1663#discussion_r68493402) for the rationale.

### DIFF
--- a/Formula/notmuch.rb
+++ b/Formula/notmuch.rb
@@ -3,6 +3,7 @@ class Notmuch < Formula
   homepage "https://notmuchmail.org"
   url "https://notmuchmail.org/releases/notmuch-0.22.tar.gz"
   sha256 "d64118ef926ba06fba814a89a75d20b0c8c8ec07dd65e41bb9f1e9db0dcfb99a"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Formula/notmuch.rb
```diff
@@ -11,12 +12,13 @@ class Notmuch < Formula
...
   end
 
+  option "without-python", "Build without python support"
+
   depends_on "pkg-config" => :build
   depends_on "gmime"
   depends_on "talloc"
   depends_on "xapian"
   depends_on :emacs => ["21.1", :optional]
-  depends_on :python => :optional
```
dunn: You can just flip this from `:optional` to `:recommended`.

dlh: Doesn't that pull in `python` as a dependency? The bindings work with the system python.

dunn: The `:python` special dependency uses whatever Python is installed as long as it's >= 2.7

dlh: I pushed the change, but `brew install` wants to pull in Homebrew's python:

    $ brew install notmuch
    ==> Installing dependencies for notmuch: python
    ^C

Oddly, `brew install --build-from-source` doesn't pull in python. Is this a known issue or caveat?

tdsmith: Actually you can delete that line altogether and add `option "without-python", "Build without Python bindings"`. In the absence of any declared Python dependency, it will build against system Python. Those bindings should work with any Python 2.7 (though they may not, which is a bug that can be fixed).

It's expected that bottling against `:python` always resolves to Homebrew's python in practice; we can avoid that.

tdsmith: (all of this is terribly counterintuitive and i'm sorry about it)

dlh: Updated

tdsmith:
Thanks!

Formula/notmuch.rb
dunn: > It's expected that bottling against :python always resolves to Homebrew's python in practice; we can avoid that.

Ah, right, I forgot about that.

PR #2726, opened 2016-07-05
espeak: fix build

danielbair:
Patched to build static and dynamic libraries with needed changes 'per brew audit' now.  
Please note: upstream development is dead on the espeak project from my research.  
A new project named espeak-ng is progressing with new developers, but this is needed for the older espeak project and other projects that link against the old espeak.


MikeMcQuaid:
@danielbair Just FYI: you don't need to resubmit PRs on changes: you can just push commits to them. It's also worth using a non-`master` branch for this reason.

> Please note: upstream development is dead on the espeak project from my research.

In this case I think it's probably worth just removing this project/formula file from Homebrew instead of patching it.

pettarin:
@MikeMcQuaid Hi, I would like to suggest to keep the eSpeak formula, as no official version of eSpeak-ng has been released yet, and eSpeak is still a popular choice for people looking for a free software TTS, plus third party projects (including a couple of mine) depend on it. Personally, I am planning to switch to eSpeak-ng only after the new lead developer of that project releases an official/packaged version --- right now one has to get and compile the dev sources.

MikeMcQuaid:
Sorry, I don't think we'll merge this as-is if it needs this much patching.

> Hi, I would like to suggest to keep the eSpeak formula, as no official version of eSpeak-ng has been released yet, and eSpeak is still a popular choice for people looking for a free software TTS, plus third party projects (including a couple of mine) depend on it. Personally, I am planning to switch to eSpeak-ng only after the new lead developer of that project releases an official/packaged version --- right now one has to get and compile the dev sources.

@pettarin That seems reasonable.

danielbair:
@MikeMcQuaid will you re-consider merging with the patches only being four lines now?
This provides libespeak.dylib which greatly accelerates @pettarin aeneas application on osx.

Formula/espeak.rb
```diff
@@ -13,16 +14,53 @@ class Espeak < Formula
...
++++ new/src/speech.h	2016-07-10 18:34:17.000000000 +0700
+@@ -46,9 +46,9 @@
+ #define __cdecl
+ //#define ESPEAK_API  extern "C"
+ 
+-#ifdef LIBRARY
+-#define USE_ASYNC
+-#endif
++//#ifdef LIBRARY
++//#define USE_ASYNC
```
MikeMcQuaid: Could just comment out this single line.

MikeMcQuaid:
@danielbair It's looking better now. Could you try to use `inreplace` to make a modification using regex? Should be able to only modify two lines that way. Also, please add a comment for each change explaining why it's needed and report it to upstream (even if you consider them dead) with a link to the upstream report.

danielbair:
@MikeMcQuaid I have modified to use `inreplace` and provided comments. I have also submitted an upstream patch with these changes to support OS X here... https://sourceforge.net/p/espeak/patches/17/
@dunn Would you be willing to remove the `marked for removal/rejection` flag per the discussion on this pull request?

PR #2729, opened 2016-07-05
genext2fs 1.4.1 (new formula)

a0s:
Generates an ext2 filesystem as a normal (non-root) user

### DIFF
--- a/Formula/genext2fs.rb
+++ b/Formula/genext2fs.rb
@@ -1,19 +1,15 @@
 class Genext2fs < Formula
   desc "Generates an ext2 filesystem as a normal (non-root) user"
-  homepage "http://freecode.com/projects/genext2fs"
-  url "https://downloads.sourceforge.net/projects/genext2fs/files/genext2fs/1.4.1/genext2fs-1.4.1.tar.gz"
+  homepage "http://genext2fs.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/genext2fs/genext2fs/1.4.1/genext2fs-1.4.1.tar.gz"
   sha256 "404dbbfa7a86a6c3de8225c8da254d026b17fd288e05cec4df2cc7e1f4feecfc"
 
   def install
-    system "./configure", "--disable-debug",
-                          "--disable-dependency-tracking",
-                          "--disable-silent-rules",
-                          "--prefix=#{prefix}"
-
+    system "./configure", "--disable-debug", "--disable-dependency-tracking", "--disable-silent-rules", "--prefix=#{prefix}"
     system "make", "install"
   end
 
   test do
-    system "#{bin}/genext2fs", "--help"
+    system "#{bin}/genext2fs", "--root", testpath, "--size-in-blocks", "20", File.join(testpath, 'test.img')
   end
 end
--- a/Formula/genext2fs.rb
+++ b/Formula/genext2fs.rb
@@ -1,0 +1,19 @@
+class Genext2fs < Formula
+  desc "Generates an ext2 filesystem as a normal (non-root) user"
+  homepage "http://freecode.com/projects/genext2fs"
+  url "https://downloads.sourceforge.net/projects/genext2fs/files/genext2fs/1.4.1/genext2fs-1.4.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/genext2fs.rb
```diff
@@ -0,0 +1,19 @@
+class Genext2fs < Formula
+  desc "Generates an ext2 filesystem as a normal (non-root) user"
+  homepage "http://freecode.com/projects/genext2fs"
```
DomT4: Freecode is dead. Might as well use the SF homepage at http://genext2fs.sourceforge.net/

Formula/genext2fs.rb
```diff
@@ -0,0 +1,19 @@
...
+  url "https://downloads.sourceforge.net/projects/genext2fs/files/genext2fs/1.4.1/genext2fs-1.4.1.tar.gz"
+  sha256 "404dbbfa7a86a6c3de8225c8da254d026b17fd288e05cec4df2cc7e1f4feecfc"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+
```
DomT4: Don't need this empty line. The spacing is far enough apart that leaving no empty line doesn't impact readability here.

Formula/genext2fs.rb
```diff
@@ -0,0 +1,19 @@
...
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/genext2fs", "--help"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

orangeudav:
Fixed

Formula/genext2fs.rb
```diff
@@ -0,0 +1,15 @@
+class Genext2fs < Formula
+  desc "Generates an ext2 filesystem as a normal (non-root) user"
+  homepage "http://genext2fs.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/genext2fs/genext2fs/1.4.1/genext2fs-1.4.1.tar.gz"
+  sha256 "404dbbfa7a86a6c3de8225c8da254d026b17fd288e05cec4df2cc7e1f4feecfc"
+
+  def install
+    system "./configure", "--disable-debug", "--disable-dependency-tracking", "--disable-silent-rules", "--prefix=#{prefix}"
```
DomT4: Sorry, the way you had this styled before was fine. My point about not needing the empty line applied only to the empty line between the end of the `./configure` list and the `system "make", "install"` call.

Formula/genext2fs.rb
```diff
@@ -0,0 +1,15 @@
...
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/genext2fs", "--root", testpath, "--size-in-blocks", "20", File.join(testpath, 'test.img')
```
DomT4: Should be able to use `testpath/"test.img"` without the File dancing? Please wrap this to 80 characters as well, you can fall onto multiple lines if necessary.

orangeudav: File.join - is true way (in Ruby) to build path. "#{testpath}/test.img" - is dirty way. 
It is not dancing, please, read this: 
http://stackoverflow.com/questions/4114480/when-is-file-join-useful
http://stackoverflow.com/questions/597488/how-to-do-a-safe-join-pathname-in-ruby

Yep, line is too long anyway

MikeMcQuaid: @orangeudav `"#{testpath}/test.img"` is the way we do it, thanks.

DomT4: @orangeudav I'm aware, thanks. Nonetheless, Homebrew has style guidelines. I didn't make the request because enforcing style every day is glorious fun; it helps contributors if we're consistent about the way we write things.

orangeudav:
Fixed

MikeMcQuaid:

PR #2736, opened 2016-07-05
cromwell: substitute corrected URL

ilovezfs:
- URL now has the expected file name "cromwell-0.19.3.jar"
- restore the simpler way of deriving the jar file name in install


ilovezfs:
@DomT4 upstream woke up

DomT4:
🎉. Feel free to ship when you're happy.

PR #2762, opened 2016-07-06
corectl 0.7.0

AntonioMeireles:
Hi all, 

please bump and, once again, thanks for your time, efforts and patience. 

regards

### DIFF
--- a/Formula/corectl.rb
+++ b/Formula/corectl.rb
@@ -1,9 +1,18 @@
 class Corectl < Formula
   desc "CoreOS over OS X made very simple"
   homepage "https://github.com/TheNewNormal/corectl"
-  url "https://github.com/TheNewNormal/corectl/archive/v0.5.10.tar.gz"
-  sha256 "e7b86914316d00f161c92f2b0bc0beb2ce26486199057e0536b16e6d47e2b47a"
   head "https://github.com/TheNewNormal/corectl.git", :branch => "golang"
+
+  stable do
+    url "https://github.com/TheNewNormal/corectl/archive/v0.7.0.tar.gz"
+    sha256 "bdbb7d453232995c93862b7e97edb43d312f4976c84c3781e902a140c88ebb45"
+
+    # until 0.7.1 is out
+    patch do
+      url "https://github.com/TheNewNormal/corectl/commit/e519a6f0d1d3c141baef631bdfc65b4e130ff60d.patch"
+      sha256 "966caa8a8f0703fffb5befcc3bec546699c06ca9512d6b1739f28029fe27ed89"
+    end
+  end
 
   bottle do
     cellar :any_skip_relocation
@@ -13,10 +22,17 @@ class Corectl < Formula
 
   depends_on "go" => :build
   depends_on "godep" => :build
+  depends_on "ocaml" => :build
+  depends_on "opam" => :build
   depends_on :macos => :yosemite
 
   def install
     ENV["GOPATH"] = buildpath
+
+    opamroot = buildpath/"opamroot"
+    opamroot.mkpath
+    ENV["OPAMROOT"] = opamroot
+    ENV["OPAMYES"] = "1"
 
     path = buildpath/"src/github.com/TheNewNormal/#{name}"
     path.install Dir["*"]
@@ -25,15 +41,31 @@ class Corectl < Formula
     args << "VERSION=#{version}" if build.stable?
 
     cd path do
-      system "make", "corectl", *args
-      system "make", "documentation/man"
-      bin.install "corectl"
+      system "opam", "init", "--no-setup"
+      qcow_format_revision = build.head? ? "master" : "96db516d97b1c3ef2c7bccdac8fb6cfdcb667a04"
+      system "opam", "pin", "add", "qcow-format",
+        "https://github.com/mirage/ocaml-qcow.git##{qcow_format_revision}"
+      system "opam", "install", "uri", "qcow-format", "ocamlfind"
+
+      system "make", "tarball", *args
+
+      bin.install Dir["bin/*"]
+
       man1.install Dir["documentation/man/*.1"]
-      share.install "cloud-init", "profiles"
+      pkgshare.install "examples"
     end
   end
 
+  def caveats; <<-EOS.undent
+    Starting with 0.7 "corectl" has a client/server architecture. So before you
+    can use the "corectl" cli, you have to start the server daemon:
+
+    $ corectld start
+
+    EOS
+  end
+
   test do
-    assert_match(/#{version}/, shell_output("#{bin}/corectl version"))
+    assert_match version.to_s, shell_output("#{bin}/corectl version")
   end
 end


AntonioMeireles:
hold down... working on this getting built on your sandboxes...

ilovezfs:
@AntonioMeireles probably needs to be `system "opam", "init", "--no-setup"`

AntonioMeireles:
@ilovezfs yes, thanks for the catch!

ilovezfs:
@AntonioMeireles You're welcome. In terms of sandbox compliance, you can pass `--sandbox` or `export HOMEBREW_SANDBOX=1` in your environment. Also, I'd suggest look at these to see if you're doing basically the same things in the same order.
```
coccinelle.rb:23:    system "opam", "init", "--no-setup"
fstar.rb:48:    system "opam", "init", "--no-setup"
infer.rb:32:    system "opam", "init", "--no-setup"
zero-install.rb:94:    system "opam", "init", "--no-setup"
```

AntonioMeireles:
hi again, is there still anything holding this back ?

Formula/corectl.rb
```diff
@@ -13,26 +13,64 @@ class Corectl < Formula
 
   depends_on "go" => :build
   depends_on "godep" => :build
+  depends_on "ocaml" => :build
+  depends_on "opam" => :build
   depends_on :macos => :yosemite
 
+  # until 0.7.1 is out
+  if build.stable?
```
DomT4: Please lob this into a `stable do` block with the stable url/sha256. We don't really do the `if build.stable?` syntax outside `def install` now.

Formula/corectl.rb
```diff
@@ -13,26 +13,64 @@ class Corectl < Formula
...
 
     cd path do
-      system "make", "corectl", *args
-      system "make", "documentation/man"
-      bin.install "corectl"
+      system "make", "default", *args
+
+      system "opam", "init", "--no-setup"
+      system "opam", "pin", "add", "qcow-format",
+        "git://github.com/mirage/ocaml-qcow#master"
```
DomT4: Any chance this can use `https://` rather than `git://`. The latter may well be firewalled to death in some places.

Formula/corectl.rb
```diff
@@ -13,26 +13,64 @@ class Corectl < Formula
...
+      system "opam", "install", "uri", "qcow-format", "ocamlfind"
+      system "make", "hyperkit"
+
+      bin.install "bin/corectl"
+      bin.install "bin/corectld"
+      bin.install "bin/corectld.runner"
+
       man1.install Dir["documentation/man/*.1"]
-      share.install "cloud-init", "profiles"
+      share.install "examples/cloud-init", "examples/profiles"
```
DomT4: In hindsight, this should possibly be using `pkgshare` rather than `share`.

Formula/corectl.rb
```diff
@@ -13,26 +13,64 @@ class Corectl < Formula
...
     end
   end
 
+  def caveats; <<-EOS.undent
+      Starting with 0.7 "corectl" has a client/server architecture. Accordingly before start
```
DomT4: This is indented a bit too much. Should be two spaces instead of four spaces/1 tab. We're fairly strict on not carrying caveats that aren't related to Homebrew specifically. We don't really want caveats to become an extension of the upstream documentation.

AntonioMeireles:
@DomT4 thanks for the feedback! at sight all your points are addressed in latest iteration. 
(fwiw _caveats_ thing has been downsized but its's really usefull as there is fundamental change in behaviour regarding previous versions (now deamon needs to be have been launched) and somehow we need to inform users that are updating about that...))

ilovezfs:
@AntonioMeireles should this be bumped to 0.7.4?

AntonioMeireles:
@ilovezfs no, not yet. we want wider feedback for 0.7.0 first before exposing to an wider audience an even richer feature surface in later versions. in any case present 0.7.4 still needs a _few more things_ before being in a state where it makes sense to be in Homebrew.

ilovezfs:
@AntonioMeireles maybe you'd like to add 0.7.4 in a `devel do` block?

AntonioMeireles:
@ilovezfs Formula updated to build present HEAD (0.7.4 right now).

Formula/corectl.rb
```diff
@@ -13,26 +20,51 @@ class Corectl < Formula
...
+      system "opam", "install", "uri", "qcow-format", "ocamlfind"
+      system "make", "hyperkit"
+
+      bin.install "bin/corectl"
+      bin.install "bin/corectld"
+      bin.install "bin/corectld.runner"
+
       man1.install Dir["documentation/man/*.1"]
-      share.install "cloud-init", "profiles"
+      pkgshare.install "examples/cloud-init", "examples/profiles"
```
ilovezfs: Wouldn't be simpler to `pkgshare.install "examples"`? Are there ones you're deliberately not installing for some reason?

AntonioMeireles:
@ilovezfs just done

Formula/corectl.rb
```diff
@@ -13,26 +20,48 @@ class Corectl < Formula
...
+        "https://github.com/mirage/ocaml-qcow.git#master"
+      system "opam", "install", "uri", "qcow-format", "ocamlfind"
+
+      system "make", "tarball", *args
+
+      bin.install Dir["bin/*"]
+
       man1.install Dir["documentation/man/*.1"]
-      share.install "cloud-init", "profiles"
+      pkgshare.install Dir["examples/"]
```
ilovezfs: it should just be "examples" no need for the Dir[] part

AntonioMeireles: updated

Formula/corectl.rb
```diff
@@ -13,26 +20,48 @@ class Corectl < Formula
...
+        "https://github.com/mirage/ocaml-qcow.git#master"
+      system "opam", "install", "uri", "qcow-format", "ocamlfind"
+
+      system "make", "tarball", *args
+
+      bin.install Dir["bin/*"]
+
       man1.install Dir["documentation/man/*.1"]
-      share.install "cloud-init", "profiles"
+      pkgshare.install "examples/"
```
ilovezfs: hehe also drop the "/"

AntonioMeireles: dropped

Formula/corectl.rb
```diff
@@ -13,26 +20,48 @@ class Corectl < Formula
...
+    Starting with 0.7 "corectl" has a client/server architecture. Accordingly before start
+    using the "corectl" cli the server daemon needs to have been started already:
+
+    $ corectld start
+
+    EOS
+  end
+
   test do
     assert_match(/#{version}/, shell_output("#{bin}/corectl version"))
```
ilovezfs: This should be
```
assert_match version.to_s, shell_output("#{bin}/corectl version")
```

Formula/corectl.rb
```diff
@@ -1,9 +1,16 @@
 class Corectl < Formula
   desc "CoreOS over OS X made very simple"
   homepage "https://github.com/TheNewNormal/corectl"
-  url "https://github.com/TheNewNormal/corectl/archive/v0.5.10.tar.gz"
-  sha256 "e7b86914316d00f161c92f2b0bc0beb2ce26486199057e0536b16e6d47e2b47a"
   head "https://github.com/TheNewNormal/corectl.git", :branch => "golang"
+  stable do
+    url "https://github.com/TheNewNormal/corectl/archive/v0.7.0.tar.gz"
+    sha256 "bdbb7d453232995c93862b7e97edb43d312f4976c84c3781e902a140c88ebb45"
```
ilovezfs: should be a blank line here

Formula/corectl.rb
```diff
@@ -1,9 +1,16 @@
 class Corectl < Formula
   desc "CoreOS over OS X made very simple"
   homepage "https://github.com/TheNewNormal/corectl"
-  url "https://github.com/TheNewNormal/corectl/archive/v0.5.10.tar.gz"
-  sha256 "e7b86914316d00f161c92f2b0bc0beb2ce26486199057e0536b16e6d47e2b47a"
   head "https://github.com/TheNewNormal/corectl.git", :branch => "golang"
```
ilovezfs: Should be a blank line here

AntonioMeireles:
update yet again

Formula/corectl.rb
```diff
@@ -1,10 +1,19 @@
...
   desc "CoreOS over OS X made very simple"
   homepage "https://github.com/TheNewNormal/corectl"
-  url "https://github.com/TheNewNormal/corectl/archive/v0.5.10.tar.gz"
-  sha256 "e7b86914316d00f161c92f2b0bc0beb2ce26486199057e0536b16e6d47e2b47a"
+
   head "https://github.com/TheNewNormal/corectl.git", :branch => "golang"
 
+  stable do
+    url "https://github.com/TheNewNormal/corectl/archive/v0.7.0.tar.gz"
+    sha256 "bdbb7d453232995c93862b7e97edb43d312f4976c84c3781e902a140c88ebb45"
```
ilovezfs: blank line here

Formula/corectl.rb
```diff
@@ -13,27 +23,49 @@ class Corectl < Formula
...
+
       man1.install Dir["documentation/man/*.1"]
-      share.install "cloud-init", "profiles"
+      pkgshare.install "examples"
     end
   end
 
+  def caveats; <<-EOS.undent
+    Starting with 0.7 "corectl" has a client/server architecture. Accordingly before start
+    using the "corectl" cli the server daemon needs to have been started already:
```
ilovezfs: There's a grammar problem with this sentence. Maybe make it

>So before you can use the "corectl" cli, you have to start the server daemon:

Formula/corectl.rb
```diff
@@ -1,10 +1,20 @@
 class Corectl < Formula
   desc "CoreOS over OS X made very simple"
   homepage "https://github.com/TheNewNormal/corectl"
-  url "https://github.com/TheNewNormal/corectl/archive/v0.5.10.tar.gz"
-  sha256 "e7b86914316d00f161c92f2b0bc0beb2ce26486199057e0536b16e6d47e2b47a"
+
```
ilovezfs: kill this blank line

ilovezfs:
:+1:

@DomT4 ?

ilovezfs:
@AntonioMeireles Thanks for your patience. I know I'm picky :)

Formula/corectl.rb
```diff
@@ -13,27 +22,49 @@ class Corectl < Formula
...
     args = []
     args << "VERSION=#{version}" if build.stable?
 
     cd path do
-      system "make", "corectl", *args
-      system "make", "documentation/man"
-      bin.install "corectl"
+      system "opam", "init", "--no-setup"
+      system "opam", "pin", "add", "qcow-format",
+        "https://github.com/mirage/ocaml-qcow.git#master"
```
DomT4: Any chance we can pin this to a specific commit rather than `master`? Note: I only use `opam` minimally, so if this comment makes no sense, feel free to say 😉.

AntonioMeireles: i'm _just_ following hyperkit's upstream _modus operandi_ here - see [this](https://github.com/docker/hyperkit/blob/master/circle.yml).

ilovezfs: @AntonioMeireles yes, we're pickier than upstreams when comes to pinning things to specific versions
CC @tdsmith

AntonioMeireles: @ilovezfs sorry here, but regarding opam et. al at least me, won't pretend to know more than _the_ upstreams...

ilovezfs: @AntonioMeireles what you'll need to do is pin it to a specific #commit `if build.stable?`, and since you have a head url at the top too, use #master `if build.head?`

AntonioMeireles: @ilovezfs i can _still_ get that part - what i also _get_ is that if/when it happens that `master` and `stable` are at same step (sometimes happens) the resulting builds will be different. and they shouldn't. Also, in this (very) specific case the issue is somewhat moot - because what is actually happening is that the docker/mirage guys are doing the pinning themselves on their forked repo. 

Of course it takes me 2m to do what you want - but - again - is it that better that when the `head` and `stable` are the same two different users (one that picks `stable` and another `head`) get potentially different binaries ?

ilovezfs: Yes.

This should work for us:
```
qcow_format_revision = build.head? ? "master" : "96db516d97b1c3ef2c7bccdac8fb6cfdcb667a04"
system "opam", "pin", "add", "qcow-format",
     "https://github.com/mirage/ocaml-qcow.git##{qcow_format_revision}"
```

AntonioMeireles: done.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/ad1bb2909278c78d898e15ff8d4ab4379f939940. Thank you for another contribution to Homebrew @AntonioMeireles; we appreciate it! :smiley_cat: 

And thank you for maintaining your patience whilst dealing with ILZ & myself, we're not the easiest combination in the world 😄.

PR #2763, opened 2016-07-06
kubernetes-cli - Add zsh completion

antoineco:

---

Follow up to the previous PR from another contributor, this time for zsh completion.

I also added a note in the _caveats_ to let people know they have the enable zsh completion explicitly. That's because the resultion completion script is missing a `#compdef` header.

### DIFF
--- a/Formula/kubernetes-cli.rb
+++ b/Formula/kubernetes-cli.rb
@@ -27,7 +27,15 @@ class KubernetesCli < Formula
 
     dir = "_output/local/bin/darwin/#{arch}"
     bin.install "#{dir}/kubectl"
+
     (bash_completion/"kubectl").write Utils.popen_read("#{bin}/kubectl completion bash")
+    (zsh_completion/"_kubectl").write Utils.popen_read("#{bin}/kubectl completion zsh")
+  end
+
+  def caveats; <<-EOS.undent
+    Add the following to ~/.zshrc to enable zsh completion:
+      source #{HOMEBREW_PREFIX}/share/zsh/site-functions/_kubectl
+    EOS
   end
 
   test do


Formula/kubernetes-cli.rb
```diff
@@ -27,7 +27,21 @@ def install
...
     dir = "_output/local/bin/darwin/#{arch}"
     bin.install "#{dir}/kubectl"
+
     (bash_completion/"kubectl").write Utils.popen_read("#{bin}/kubectl completion bash")
+    (zsh_completion/"_kubectl").write Utils.popen_read("#{bin}/kubectl completion zsh")
+  end
+
+  def caveats; <<-EOS.undent
+    Add the following to ~/.zshrc to enable zsh completion:
+      source #{HOMEBREW_PREFIX}/share/zsh/site-functions/_kubectl
```
dunn: If this is a bug in the completion script it should be reported to the developers.

Formula/kubernetes-cli.rb
```diff
@@ -27,7 +27,21 @@ def install
...
+
+    You can configure remote access to Kubernetes clusters with:
+      kubectl config
```
dunn: This (and the next paragraph) aren't Homebrew-specific, are they?  If not they should be removed.

Formula/kubernetes-cli.rb
antoineco: This is not necessarily a bug, but I could suggest this addition upstream.

Formula/kubernetes-cli.rb
antoineco: This was inspired by the [awscli formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli.rb). It applies to your previous comment as well.

Now that this is clarified, I can remove the whole caveats block if you wish. Could you confirm?

Formula/kubernetes-cli.rb
dunn: Thanks! You can leave this line until we hear back from upstream.

DomT4:
Possibly worth cross-referencing this comment: https://github.com/Homebrew/homebrew-core/pull/2777#issuecomment-230892499

Formula/kubernetes-cli.rb
elyscape: This isn't really necessary; putting `_kubectl` in the `site-functions` directory makes it get automatically loaded so long as zsh's completion is enabled.

elyscape:
Indeed, this will cause issues until kubernetes/kubernetes#28049 (on zsh < 5.1) and (much more critically) kubernetes/kubernetes#27538 are fixed.

Formula/kubernetes-cli.rb
elyscape: To elaborate, zsh will automatically load files in `site-functions` that start with an underscore.

antoineco: Right, but only if the script starts with the `#compdef` header.

Ref: http://zsh.sourceforge.net/Doc/Release/Completion-System.html

> When compinit is run, it searches all such files accessible via fpath/FPATH and reads the first line of each of them. This line should contain one of the tags described below. Files whose first line does not start with one of these tags are not considered to be part of the completion system and will not be treated specially.

apjanke: This is a well-established convention for `zsh` completion files. A completion definition that doesn't get picked up by `compinit` is effectively broken. We should just add the `#compdef kubectl` line at the start of the file when we generate it, and remove the caveat.

MikeMcQuaid:
Closing until mentioned issues are fixed.

jedahan:
It looks like all the relevant kubernetes bugs have been solved or closed, wanna take another look at this?

MikeMcQuaid:
If someone opens a PR: sure.

antoineco:
@jedahan good point
@MikeMcQuaid I updated my PR

MikeMcQuaid:
@antoineco You'll need to open a new one as this is closed.

PR #2770, opened 2016-07-06
owamp: run as daemon using plist [help required]

candlerb:
- [*] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [*] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [*] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [*] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

This is an update to owamp to:
- Fix the `-Z` (foreground) flag when running under launchd: [patch submitted upstream](https://github.com/perfsonar/owamp/pull/11)
- Includes a plist file
- Creates default config files and working directories

However there is a problem which I don't know how best to fix in the homebrew framework: the daemon starts as user "nobody" and then tries to create a subdirectory but does not have permissions to do so.

```
Jul  6 14:24:44 bri12-client-1networkcityfibrecom.local owampd[6830] <Error>: FILE=policy.c, LINE=811, Unable to mkdir(/usr/local/var/lib/owamp/catalog): Permission denied
```

This can be fixed by manually doing `chown nobody:nogroup /usr/local/var/lib/owamp/catalog` after installing the package, at which point the daemon runs happily.

I don't know how best to approach this. Should the daemon run as the homebrew owner rather than 'nobody'? Or is there a hook to create or chown a directory when the daemon starts up?

### DIFF
--- a/Formula/owamp.rb
+++ b/Formula/owamp.rb
@@ -17,6 +17,8 @@ class Owamp < Formula
   # Fix to prevent tests hanging under certain circumstances.
   # Provided by Aaron Brown via perfsonar-user mailing list:
   # https://lists.internet2.edu/sympa/arc/perfsonar-user/2014-11/msg00131.html
+  # Also fix to allow owampd -Z to start under launchd.
+  # https://github.com/perfsonar/owamp/pull/11
   patch :DATA
 
   def install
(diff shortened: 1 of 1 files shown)


Formula/owamp.rb
```diff
@@ -17,6 +17,8 @@ class Owamp < Formula
   # Fix to prevent tests hanging under certain circumstances.
   # Provided by Aaron Brown via perfsonar-user mailing list:
   # https://lists.internet2.edu/sympa/arc/perfsonar-user/2014-11/msg00131.html
```
MikeMcQuaid: Could you see if this was ever submitted upstream and, if not, consider doing so? Thanks!

Formula/owamp.rb
```diff
@@ -25,6 +27,83 @@ def install
...
                           "--mandir=#{man}"
     system "make", "install"
+
+    (buildpath+"owampd.conf").write owampd_conf
+    (buildpath+"owampd.limits").write owampd_limits
+    etc.install "owampd.conf"
+    etc.install "owampd.limits"
+
+    (var+"lib/owamp").mkpath
+    (var+"log/owamp").mkpath
```
MikeMcQuaid: Put these in `make install` and use `/` instead of `+` both here and above.

Formula/owamp.rb
```diff
@@ -25,6 +27,83 @@ def install
...
+    # FIXME: daemon fails with
+    # Jul  6 14:24:44 bri12-client-1networkcityfibrecom.local owampd[6830] <Error>: FILE=policy.c, LINE=811, Unable to mkdir(/usr/local/var/lib/owamp/catalog): Permission denied
+    # until you do: chown nobody:nogroup /usr/local/var/lib/owamp
+    (var+"run").mkpath
+  end
+
+  def owampd_conf; <<-EOS.undent
```
MikeMcQuaid: Is this provided with the software, at all?

MikeMcQuaid:
> I don't know how best to approach this. Should the daemon run as the homebrew owner rather than 'nobody'? Or is there a hook to create or chown a directory when the daemon starts up?

It should run as the Homebrew user and be `inreplace`d in `post_install` to `ENV["USER"]`

Formula/owamp.rb
candlerb: No, it's not in the master branch of perfsonar/owamp. I will submit upstream.

(Thanks for all other comments - I won't have time to re-integrate for another week or two, but I will do so)

MikeMcQuaid:

PR #2773, opened 2016-07-06
launch_socket_server: add default plist.

MikeMcQuaid:

---

Add a default plist to launch_socket_server which does the most useful typical configuration for a Homebrew user: proxy everything on localhost:8080 to localhost:80 so you can use e.g. `nginx` on port 80 as a non-root user.

Also, while we're here, simplify installation to install all the same files using `make install` and do some general cleanup.

CC @josh @mistydemeo


MikeMcQuaid:
@apjanke Can you remember if the test was ever passing here on Mavericks? It looks like the CI was never green for this.

mistydemeo:
According to the documentation it should work on 10.9, but maybe we should just set a minimum OS X tag of 10.10.

mistydemeo:
Let's bump the revision, it'd be useful to be able to know that people with the latest version installed have the plist.

apjanke:
> Can you remember if the test was ever passing here on Mavericks?

Sorry, don't remember.

PR #2774, opened 2016-07-06
grafana 3.1.1 (new formula)

ctdk:

---


Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+  url "https://github.com/grafana/grafana/archive/v3.0.4.tar.gz"
+  sha256 "f26a374326e64a8f83c57fdf916ba1c8524dd55002dfe628b4854d05fed3715a"
+
+  head "https://github.com/grafana/grafana.git"
+
+  depends_on "go" => :build
+  depends_on "node" => :build
+
+  def install
+    ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
```
DomT4: Don't need this any longer. It gets done by the language extension now.

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
+require "language/go"
+require "language/node"
+require "open3"
```
DomT4: If this is only used for testing it should be at the top of `test do`.

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+  depends_on "node" => :build
+
+  def install
+    ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
+    ENV["GOPATH"] = buildpath
+    grafana_path = buildpath/"src/github.com/grafana/grafana"
+    grafana_path.install Dir["*"]
+    grafana_path.install ".jscs.json", ".jsfmtrc", ".jshintrc", ".bowerrc"
+
+    Language::Go.stage_deps resources, buildpath/"src"
```
DomT4: You haven't added any `go_resource`s to stage.

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
+require "language/go"
```
DomT4: Don't need this unless you're also using `go_resource`s.

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+    ENV["GOPATH"] = buildpath
+    grafana_path = buildpath/"src/github.com/grafana/grafana"
+    grafana_path.install Dir["*"]
+    grafana_path.install ".jscs.json", ".jsfmtrc", ".jshintrc", ".bowerrc"
+
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd grafana_path do
+      # Might do it differently for head vs. release
+      system %q(sed -i.bak 's/"sass-lint": "^1.6.0",/"sass-lint": "1.7.0",/;' package.json)
```
DomT4: This should be using an `inreplace`, but if we can drop `head` and drop this I'm game for that.

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+      system "npm", "install", "grunt-cli", *Language::Node.local_npm_install_args
+      system "node_modules/grunt-cli/bin/grunt", "build"
+    end
+
+    cp(grafana_path/"conf/sample.ini", grafana_path/"conf/grafana.ini")
+
+    bin.install grafana_path/"bin/grafana-cli"
+    bin.install grafana_path/"bin/grafana-server"
+    (etc/"grafana").mkpath
+    (var/"log/grafana").mkpath
```
DomT4: Anything related to `var` being created should be done in `post_install`.

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+    (var/"lib/grafana").mkpath
+    (var/"lib/grafana/plugins").mkpath
```
DomT4: `mkpath` is an alias for `mkdir_p` so if you use `(var/"lib/grafana/plugins").mkpath` you don't need the line above creating `(var/"lib/grafana")`

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+    etc.install grafana_path/"conf/grafana.ini" => "grafana/grafana.ini"
+    pkgshare.install Dir[grafana_path/"conf", grafana_path/"public_gen", grafana_path/"vendor"]
```
DomT4: Why would you need to install the `vendor` Go resources permanently? Go does static builds, they shouldn't be necessary?

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+    mv pkgshare/"public_gen", pkgshare/"public"
+  end
+
+  plist_options :manual => "grafana-server --config=#{HOMEBREW_PREFIX}/etc/grafana/grafana.ini --homepath #{HOMEBREW_PREFIX}/share/grafana cfg:default.paths.logs=#{HOMEBREW_PREFIX}/var/log/grafana cfg:default.paths.data=#{HOMEBREW_PREFIX}/var/lib/grafana cfg:default.paths.plugins=#{HOMEBREW_PREFIX}/var/lib/grafana/plugins"
```
DomT4: It feels like we should/could be setting a lot of this with an env script.

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+          <integer>10240</integer>
+        </dict>
+      </dict>
+    </plist>
+   EOS
+  end
+
+  test do
+    system bin/"grafana-server", "-v"
+    RUBY_VERSION =~ /^(\d+\.\d+)/
```
DomT4: Where/why is this being used?

Formula/grafana.rb
```diff
@@ -0,0 +1,123 @@
...
+    v = $1.to_f
+    Dir.mktmpdir("grafana", HOMEBREW_TEMP) do |tdir|
```
DomT4: You're already in a temporary directory in `test do`. Don't need to create another.

Formula/grafana.rb
ctdk: A commit with these changes is coming, but I'm dropping some notes here before sending that up. Here, I've changed it to `inreplace`, but the release version still has that sass-lint dependency that needs changed on OS X (Linux, frustratingly, doesn't have the problem with sass-lint 1.8.0 that OS X does). I'm going to try and get that changed upstream, but until then that small replacement is necessary.

Formula/grafana.rb
ctdk: The `vendor` name is unfortunate here. This `vendor` directory is vendored javascript (phantomjs specifically, last I checked). The vendored Go resources are in `Godeps`.

Formula/grafana.rb
ctdk: I added a temp directory because the test was blowing up without it, but I shifted the steps around in the coming commit & was able to remove that. Thanks for pointing it out.

Formula/grafana.rb
ctdk: I was using it to run the listening test differently depending on the version of ruby Homebrew is using -- on Yosemite and El Capitan (and on Mavericks on my desktop here) using `spawn` works fine because the system ruby is 2.something, but the usual Mavericks installation still uses 1.8.7, so the test was breaking because `Process.spawn` wasn't added until somewhere in 1.9. (This is why this PR ended up having 4 commits instead of 1, because I didn't figure out right away why the test was passing locally on Mavericks but blowing up with brew-test-bot.)

I've changed the test to use the older style `PTY.spawn` for all versions and removed that test. When Homebrew drops Mavericks support, that can be changed back to `Process.spawn`.

Formula/grafana.rb
ctdk: I can craft something up for this. Upstream has init scripts and such for various Linuxes, but not upstream. There's a plist in here, of course, for running it as a service, but for running manually I just put the usual command line for running it manually.

ctdk:
I've just pushed up a commit that hopefully addresses the issues previously identified.

ctdk:
(If this has gotten too messy, especially because of the thing I mentioned in a line note where a test was passing in Mavericks locally but failing on brew-test-bot I can pull this down, make a new branch, and do a squash commit.)

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  def install
+    ENV["GOPATH"] = buildpath
+    grafana_path = buildpath/"src/github.com/grafana/grafana"
+    grafana_path.install Dir["*"]
+    grafana_path.install ".jscs.json", ".jsfmtrc", ".jshintrc", ".bowerrc"
+
+    cd grafana_path do
+      inreplace "package.json", '"sass-lint": "^1.6.0"', '"sass-lint": "1.7.0"'
+      system "go", "run", "build.go", "setup"
```
DomT4: Are there any external dependencies for the `go` stuff here?

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+  depends_on "node" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    grafana_path = buildpath/"src/github.com/grafana/grafana"
+    grafana_path.install Dir["*"]
+    grafana_path.install ".jscs.json", ".jsfmtrc", ".jshintrc", ".bowerrc"
+
+    cd grafana_path do
+      inreplace "package.json", '"sass-lint": "^1.6.0"', '"sass-lint": "1.7.0"'
```
DomT4: Can you add a `#` comment explaining this inreplace? Thanks!

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+      system "go", "run", "build.go", "setup"
+      system "go", "run", "build.go", "build"
+      system "npm", "install", *Language::Node.local_npm_install_args
+      system "npm", "install", "grunt-cli", *Language::Node.local_npm_install_args
+      system "node_modules/grunt-cli/bin/grunt", "build"
+    end
+
+    cp(grafana_path/"conf/sample.ini", grafana_path/"conf/grafana.ini")
```
DomT4: Why are we copying this here? Is it just to handle the rename or?

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    bin.install grafana_path/"bin/grafana-cli"
```
DomT4: Rather than having to append `grafana_path` to all of this, maybe it's saner to just execute these steps inside the `cd grafana_path do` block?

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+    bin.install grafana_path/"bin/grafana-server"
+    (grafana_path/"grafana").write(env_script)
+    chmod 0755, grafana_path/"grafana"
+    bin.install grafana_path/"grafana"
+    (etc/"grafana").mkpath
```
DomT4: Shouldn't need this line. `etc.install` will create it if it doesn't exist.

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+    etc.install grafana_path/"conf/grafana.ini" => "grafana/grafana.ini"
+    pkgshare.install Dir[grafana_path/"conf", grafana_path/"public_gen", grafana_path/"vendor"]
+    mv pkgshare/"public_gen", pkgshare/"public"
```
DomT4: Might be a bit neater to just drop `public_gen` from the above line & then do:
```ruby
pkgshare.install "public_gen" => "public"
```
here instead.

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+    (var/"log/grafana").mkpath
+    (var/"lib/grafana/plugins").mkpath
+  end
+
+  def env_script
+    <<-EOS.undent
+      #!/usr/bin/env bash
+      DAEMON=grafana-server
+      EXECUTABLE=#{bin/"grafana-server"}
+      CONFIG=#{HOMEBREW_PREFIX}/etc/grafana/grafana.ini
```
DomT4: `#{etc}/grafana/grafana.ini`

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+      HOMEPATH=#{HOMEBREW_PREFIX}/share/grafana
```
DomT4: `pkgshare`

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+      LOGPATH=#{HOMEBREW_PREFIX}/var/log/grafana
```
DomT4: `#{var}/log/grafana`

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+      DATAPATH=#{HOMEBREW_PREFIX}/var/lib/grafana
```
DomT4: `#{var}/lib/grafana`

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+      PLUGINPATH=#{HOMEBREW_PREFIX}/var/lib/grafana/plugins
```
DomT4: `#{var}/log/grafana/plugins`

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+          <key>SuccessfulExit</key>
+          <false/>
+        </dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>ProgramArguments</key>
+        <array>
+          <string>#{opt_bin}/grafana-server</string>
+          <string>--config</string>
+          <string>#{HOMEBREW_PREFIX}/etc/grafana/grafana.ini</string>
```
DomT4: `#{etc}/grafana/grafana.ini`

Formula/grafana.rb
```diff
@@ -0,0 +1,149 @@
...
+          <string>--homepath</string>
+          <string>#{HOMEBREW_PREFIX}/share/grafana</string>
```
DomT4: `opt_pkgshare`

Formula/grafana.rb
ctdk: No. It does use the go sqlite library, but that ships with OS X.

Formula/grafana.rb
ctdk: Yeah. I'm changing the installation line to skip the rename here.

Formula/grafana.rb
ctdk: (It does download external go libs, of course, but those are all vendored with godeps.)

Formula/grafana.rb
```diff
@@ -0,0 +1,154 @@
...
+    grafana_path = buildpath/"src/github.com/grafana/grafana"
+    grafana_path.install Dir["*"]
+    grafana_path.install ".jscs.json", ".jsfmtrc", ".jshintrc", ".bowerrc"
+
+    cd grafana_path do
+      # The sass-lint npm package dependencey in package.json specifies any
+      # version greater than 1.6.0 for grafana, but on OS X versions 1.8.0+
+      # break. This replaces the specification for any version >= 1.6.0 and
+      # sets it to 1.7.0 (which works).
+      inreplace "package.json", '"sass-lint": "^1.6.0"', '"sass-lint": "1.7.0"'
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/grafana.rb
```diff
@@ -0,0 +1,154 @@
...
+      pkgshare.install "public_gen" => "public"
+    end
+  end
+
+  def post_install
+    (var/"log/grafana").mkpath
+    (var/"lib/grafana/plugins").mkpath
+  end
+
+  def env_script
```
MikeMcQuaid: This is a bunch of logic to live in the formula; is there a reason we can't just use upstream scripts as-is?

Formula/grafana.rb
```diff
@@ -0,0 +1,154 @@
...
+      system "npm", "install", "grunt-cli", *Language::Node.local_npm_install_args
+      system "node_modules/grunt-cli/bin/grunt", "build"
+
+      bin.install "bin/grafana-cli"
+      bin.install "bin/grafana-server"
+      (bin/"grafana").write(env_script)
+      chmod 0755, bin/"grafana"
+      (etc/"grafana").mkpath
+      cp("conf/sample.ini", "conf/grafana.ini.example")
+      unless (etc/"grafana/grafana.ini").exist?
```
MikeMcQuaid: `etc.install` will check if the file exists first so this isn't needed.

Formula/grafana.rb
```diff
@@ -0,0 +1,154 @@
...
+        etc.install "conf/sample.ini" => "grafana/grafana.ini"
+      end
+      etc.install "conf/grafana.ini.example" => "grafana/grafana.ini.example"
+      pkgshare.install Dir["conf", "vendor"]
```
MikeMcQuaid: Don't need the `Dir["` here.

Formula/grafana.rb
```diff
@@ -0,0 +1,154 @@
...
+          <key>NumberOfFiles</key>
+          <integer>10240</integer>
+        </dict>
+      </dict>
+    </plist>
+   EOS
+  end
+
+  test do
+    require "pty"
```
MikeMcQuaid: Should be able to do this with `fork` instead.

Formula/grafana.rb
ctdk: That's actually happened with the next version, so this isn't necessary anymore. I wasn't sure about changing the version on this in the middle of getting it accepted, but I'll updated all that accordingly (and take this chunk out entirely now that it's not relevant).

Formula/grafana.rb
ctdk: There is no comparable script upstream. There's a SysV init script and a systemd service definition, but that won't work here. I added that script there per the request in this comment: https://github.com/Homebrew/homebrew-core/pull/2774#discussion_r70016409.

Formula/grafana.rb
ctdk: Is it particularly worthwhile to do so here? PTY is in the ruby stdlib, so it's not a gem dependency that's unlikely to be installed. I had considered `fork` when I was writing this test, but using `spawn` seemed a lot more straightforward (but see below) as far as being able to run grafana long enough to see if it successfully started, kill the process, and read its output. With `fork` it still needs to get the standard out from the child and kill the running process (because grafana won't stop on its own). I can do it that way, but I'm not sure it's actually better.

Originally here I used `Process.spawn`, which Worked On My Machine™ running Mavericks, but failed when running on the homebrew jenkins server when it was building it for Mavericks (but worked fine on Yosemite and El Capitan). It turned out that my system ruby on this Mavericks box had been upgraded to Ruby 2.0, while the brew jenkins Maverics was using 1.8.7 as its system ruby. Unfortunately, `Process.spawn` isn't available under 1.8.7.

I had changed it to use `Process.spawn` for Yosemite and El Capitan and `PTY.spawn` for Mavericks, but decided that it was less confusing to just use `PTY.spawn` for all three of them.

That said, once El Capitan's successor comes out & homebrew drops Mavericks then `Process.spawn` would work just fine.

ctdk: Tried it out with `fork`, incidentally, but it didn't work particularly well.

Formula/grafana.rb
ilovezfs: Then to me at least, it sounds like this software has not actually been properly ported to macOS by upstream, which is typically disqualifying for a new formula.

ilovezfs:
```
Error: 10 problems in 1 formula
grafana:
  * Use spaces instead of tabs for indentation
  * Use spaces instead of tabs for indentation
  * Use spaces instead of tabs for indentation
  * Use spaces instead of tabs for indentation
  * C: 142: col 24: Do not use `::` for method calls.
  * C: 145: col 1: Tab detected.
  * C: 145: col 2: Use 2 (not -5) spaces for indentation.
  * C: 146: col 1: Tab detected.
  * C: 147: col 1: Tab detected.
  * C: 148: col 1: Tab detected.
```

Formula/grafana.rb
ctdk: Before I get it fixed, which thing is the problem -- the presence of the wrapper script created by env_script, or the copying around the various conf files and such? (There's no wrapper script for any platform, but I could submit something upstream to get it added.)

ctdk:
Problem with audit fixed. My editor (at least on this computer) keeps converting spaces to tabs when there's 8 of them, and the :: thing was per the ruby docs' suggestion.

AlekSi:
Would like to see this accepted soon.

Formula/grafana.rb
```diff
@@ -0,0 +1,158 @@
...
+  head "https://github.com/grafana/grafana.git"
+
+  depends_on "go" => :build
+  depends_on "node" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    grafana_path = buildpath/"src/github.com/grafana/grafana"
+    grafana_path.install Dir["*"]
+    grafana_path.install ".jscs.json", ".jsfmtrc", ".jshintrc", ".bowerrc"
```
ilovezfs: You can replace
```
    grafana_path.install Dir["*"]
    grafana_path.install ".jscs.json", ".jsfmtrc", ".jshintrc", ".bowerrc"
```
with
```
    grafana_path.install buildpath.children
```

Formula/grafana.rb
```diff
@@ -0,0 +1,158 @@
...
+      system "go", "run", "build.go", "setup"
+      system "go", "run", "build.go", "build"
+      system "npm", "install", *Language::Node.local_npm_install_args
+      system "npm", "install", "grunt-cli", *Language::Node.local_npm_install_args
+      system "node_modules/grunt-cli/bin/grunt", "build"
+
+      bin.install "bin/grafana-cli"
+      bin.install "bin/grafana-server"
+      (bin/"grafana").write(env_script)
+      chmod 0755, bin/"grafana"
```
ilovezfs: This line shouldn't be necessary since permissions are set/"corrected" automatically by brew during installation.

ctdk:
Addressed those most recent issues @ilovezfs brought up, and bumped the version again (3.1.1 came out a few days ago).

Formula/grafana.rb
MikeMcQuaid: Can you remove the `env_script` in favour of letting people call `grafana-server` directly and see if you can get upstream to accept and bundle this script with the application?

Formula/grafana.rb
MikeMcQuaid: If it works: I'm fine with the current test as-is 👍

ctdk: Oh yes, the test works fine.

Formula/grafana.rb
ctdk: I certainly can. I only added it because of a request earlier. I'll take that out, and submit that script to upstream.

MikeMcQuaid:
Thanks for your contribution to Homebrew, @ctdk, and your patience! Without people like you submitting PRs we couldn't run this project. You rock!

PR #2777, opened 2016-07-06
kubernetes-cli: have kubectl generate completions

elyscape:

---

There's no real need to pre-compile the bash completion at installation time when we can have it generate it at runtime. ~~This also adds zsh completion using the same trick.~~

Signed-off-by: Eli Young elyscape@gmail.com

### DIFF
--- a/Formula/kubernetes-cli.rb
+++ b/Formula/kubernetes-cli.rb
@@ -27,7 +27,7 @@ class KubernetesCli < Formula
 
     dir = "_output/local/bin/darwin/#{arch}"
     bin.install "#{dir}/kubectl"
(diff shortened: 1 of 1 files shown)


elyscape:
Actually, I'm going to remove the zsh completion due to kubernetes/kubernetes#28049 and (more critically) kubernetes/kubernetes#27538.

elyscape:
zsh completion removed.

dunn:
> There's no real need to pre-compile the bash completion at installation time when we can have it generate it at runtime.

What is the benefit of re-generating it for each shell session?

DomT4:
Yeah, I'm not convinced this is particularly performance friendly. Also curious as to an answer on @dunn's query.

elyscape:
So I don't really remember why I thought this was a good idea. Woo!

PR #2792, opened 2016-07-07
cryptol 2.4.0

ilovezfs:
- use the hackage tarball for stable
- remove the patch, which is already included in this version
- use `install_cabal_package` instead of `make install`
- z3 is a run-time dependency only
- style cleanups in the test


DomT4:
👍 when you're happy.

PR #2795, opened 2016-07-07
jmxtrans 20160706-259

ilovezfs:
Also,
- style tweaks
- move some items to `doc` and `pkgshare`


ilovezfs:
@DomT4 :ship:?

DomT4:
If you're happy, :shipit:

ilovezfs:
@DomT4 I wouldn't say I'm remotely happy with the install method in this one, but at least it's sightly cleaner than I found it.

DomT4:
Yeah, it's not glorious, but it'll do 😄.

PR #2799, opened 2016-07-07
libcapn 2.0.0

ilovezfs:
also some style tweaks

### DIFF
--- a/Formula/libcapn.rb
+++ b/Formula/libcapn.rb
@@ -4,12 +4,12 @@ class Libcapn < Formula
   head "https://github.com/adobkin/libcapn.git"
 
   stable do
-    url "https://github.com/adobkin/libcapn/archive/1.1.0.tar.gz"
-    sha256 "fbe93f8fad4247b898d518ea0f38484eb062cd830a41bb26a6ac4b95cbf076e4"
+    url "https://github.com/adobkin/libcapn/archive/v2.0.0.tar.gz"
+    sha256 "6a0d786a431864178f19300aa5e1208c6c0cbd2d54fadcd27f032b4f3dd3539e"
 
     resource "jansson" do
-      url "https://github.com/akheron/jansson/archive/v2.5.tar.gz"
-      sha256 "f328f25fc74a14b6a636245ad28d4cd3affd792f7ffaab4c021a99b3694e4287"
+      url "https://github.com/akheron/jansson.git",
+        :revision => "8f067962f6442bda65f0a8909f589f2616a42c5a"
     end
   end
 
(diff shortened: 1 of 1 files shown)


Formula/libcapn.rb
```diff
@@ -20,58 +20,32 @@ class Libcapn < Formula
...
-    spec = Tab.for_name("libcapn").source["spec"]
-    example = if spec == "stable"
-      "send_push.c"
-    elsif spec == "head" || spec == "devel"
-      "send_push_message.c"
-    end
-    cp "#{doc}/examples/#{example}", testpath
+    cp pkgshare/"examples/send_push_message.c", testpath
     flags = [
       "-I#{Formula["openssl"].opt_prefix}/include",
```
DomT4: Not your fault, but this should be `opt_include` rather than `opt_prefix/include`.

Formula/libcapn.rb
```diff
@@ -20,58 +20,32 @@ class Libcapn < Formula
...
-      url "https://github.com/akheron/jansson.git", :revision => "e44b2231b50aea5de78b7ea2debec0d5327cd711"
-    end
-  end
-
   depends_on "cmake" => :build
   depends_on "pkg-config" => :build
   depends_on "openssl"
 
   def install
     if build.stable?
```
DomT4: Do we still need to check this with `if build.stable?`

ilovezfs: Yes. jansson is actually a submodule, but upstream hasn't set a specific revision. For stable, I arbitrarily pegged it to the current HEAD revision of jansson, but we could let it float for both stable and head and drop the resource block, if that's preferred.

ilovezfs: @DomT4 do we have a general policy on git submodules that point to their respective repository's head in a stable spec?

DomT4: Not a hard one, no. It makes sense to peg it to a specific commit for stable/devel & roam around freely for `head`, IMO. This _seems_ reasonable to me as-is.

Formula/libcapn.rb
ilovezfs: Audit checks for that would be nice, like the pkgshare one

ilovezfs:
@DomT4 PR refreshed.

Formula/libcapn.rb
```diff
@@ -20,58 +20,28 @@ class Libcapn < Formula
...
-      ENV.cflags.to_s.split,
-    ].flatten
-    system ENV.cc, "-o", "example", example, *flags
-    if spec == "stable"
-      assert_match /unable to use specified SSL certificate \(12\)/, shell_output("./example", 1)
-    elsif spec == "head" || spec == "devel"
-      assert_match /unable to use specified PKCS12 file \(errno: 9012\)/, shell_output("./example", 255)
-    end
+    cp pkgshare/"examples/send_push_message.c", testpath
+    system ENV.cc, pkgshare/"examples/send_push_message.c",
```
DomT4: Should this be `testpath/"send_push_message.c"`? Bit confused that it was copied to the `testpath` & then seems to be unused, but may be missing something obvious.

DomT4: 👍 on deleting that line.

PR #2813, opened 2016-07-07
fish: Version independent config dirs for fish. Fixes completion support.

### LINKED ISSUE #45124: Fish completion supporting is broken
`fish_completion.install` could install script successfully, but the completion function doesn't work. I think brew installed script into `/usr/local/share/fish/vendor_completions.d/` but fish only using `/usr/share/fish/vendor-completions.d`, and `pkg-config --variable completionsdir fish` which is `/usr/local/Cellar/fish/2.2.0/share/fish/vendor_completions.d` (see fish-shell/fish-shell#1485), both of two directories not exist to cause homebrew's fish completion supporting broken.

gyakovlev:

---

This fixes https://github.com/Homebrew/legacy-homebrew/issues/45124 by using 3 extra dirs.

```
 #{HOMEBREW_PREFIX}/share/fish/vendor_functions.d
 #{HOMEBREW_PREFIX}/share/fish/vendor_completions.d
 #{HOMEBREW_PREFIX}/share/fish/vendor_conf.d
```

all Formulae using `fish_completion.install` will just work now as it installes files to 
_#{HOMEBREW_PREFIX}/share/fish/vendor_completions.d_
(there are 8 in core tap, I tested on youtube-dl formula)

other 2 dirs are for 3-rd party fish shell extensions like this one https://github.com/fisherman/homebrew-tap

It's my first contribution, please let me know if something is wrong.

### DIFF
--- a/Formula/fish.rb
+++ b/Formula/fish.rb
@@ -1,9 +1,17 @@
 class Fish < Formula
   desc "User-friendly command-line shell for UNIX-like operating systems"
   homepage "https://fishshell.com"
-  url "https://fishshell.com/files/2.3.1/fish-2.3.1.tar.gz"
-  mirror "https://github.com/fish-shell/fish-shell/releases/download/2.3.1/fish-2.3.1.tar.gz"
-  sha256 "328acad35d131c94118c1e187ff3689300ba757c4469c8cc1eaa994789b98664"
+
+  stable do
+    url "https://fishshell.com/files/2.3.1/fish-2.3.1.tar.gz"
+    mirror "https://github.com/fish-shell/fish-shell/releases/download/2.3.1/fish-2.3.1.tar.gz"
+    sha256 "328acad35d131c94118c1e187ff3689300ba757c4469c8cc1eaa994789b98664"
+
+    # Skip extra dirs creation during install phase patch for current stable. To be removed on the next fish release.
+    # As discussed in https://github.com/Homebrew/homebrew-core/pull/2813
+    # Use part of https://github.com/fish-shell/fish-shell/commit/9abbc5f upstream commit and patch the makefile.
+    patch :DATA
+  end
 
   bottle do
     sha256 "99462c8b9fc844882b8877f2b016823ce7c9e54dd89d532e13ce9e3af90558d4" => :el_capitan
@@ -22,9 +30,17 @@ class Fish < Formula
 
   def install
     system "autoconf" if build.head? || build.devel?
+
     # In Homebrew's 'superenv' sed's path will be incompatible, so
     # the correct path is passed into configure here.
-    system "./configure", "--prefix=#{prefix}", "SED=/usr/bin/sed"
+    args = %W[
+      --prefix=#{prefix}
+      --with-extra-functionsdir=#{HOMEBREW_PREFIX}/share/fish/vendor_functions.d
+      --with-extra-completionsdir=#{HOMEBREW_PREFIX}/share/fish/vendor_completions.d
+      --with-extra-confdir=#{HOMEBREW_PREFIX}/share/fish/vendor_conf.d
+      SED=/usr/bin/sed
+    ]
+    system "./configure", *args
     system "make", "install"
   end
 
@@ -39,7 +55,30 @@ class Fish < Formula
     EOS
   end
 
+  def post_install
+    (pkgshare/"vendor_functions.d").mkpath
+    (pkgshare/"vendor_completions.d").mkpath
+    (pkgshare/"vendor_conf.d").mkpath
+  end
+
   test do
     system "#{bin}/fish", "-c", "echo"
   end
 end
+
+__END__
+As discussed in https://github.com/Homebrew/homebrew-core/pull/2813
+Grab part of https://github.com/fish-shell/fish-shell/commit/9abbc5f upstream commit and patch the makefile.
+--- a/Makefile.in
++++ b/Makefile.in@@ -664,5 +664,5 @@ install-force: all install-translations
+@@ -664,5 +664,5 @@ install-force: all install-translations
+ 	$(INSTALL) -m 755 -d $(DESTDIR)$(datadir)/fish/completions
+-	$(INSTALL) -m 755 -d $(DESTDIR)$(extra_completionsdir)
+-	$(INSTALL) -m 755 -d $(DESTDIR)$(extra_functionsdir)
+-	$(INSTALL) -m 755 -d $(DESTDIR)$(extra_confdir)
++	$(INSTALL) -m 755 -d $(DESTDIR)$(extra_completionsdir); true
++	$(INSTALL) -m 755 -d $(DESTDIR)$(extra_functionsdir); true
++	$(INSTALL) -m 755 -d $(DESTDIR)$(extra_confdir); true
+ 	$(INSTALL) -m 755 -d $(DESTDIR)$(datadir)/fish/functions
+
+


gyakovlev:
Jenkins build failed with
`install: mkdir /usr/local/share/fish: Operation not permitted
make: *** [install-force] Error 71`

not sure why it failed to create directory.

Formula/fish.rb
```diff
@@ -22,9 +22,16 @@ class Fish < Formula
 
   def install
     system "autoconf" if build.head? || build.devel?
+    args = %W[
+      --prefix=#{prefix}
+      --with-extra-functionsdir=#{HOMEBREW_PREFIX}/share/fish/vendor_functions.d
```
dunn: `#{HOMEBREW_PREFIX}/share/fish` on each line should be `#{pkgshare}` instead; it's failing because Homebrew doesn't allow writing outside the formula's Cellar during installation.

gyakovlev: `#{pkgshare}` will not work. The patch is about having version independent dirs outside of the Cellar so other Formulae can install completions and functions into that dir and fish can use those.
By default fish looking only into own Cellar share dirs for scripts. My PR is trying to make fish to look into #{HOMEBREW_PREFIX}/share/fish for completions and functions.

zsh has similar thing
`-enable-site-fndir=#{HOMEBREW_PREFIX}/share/zsh/site-functions`
and `zsh_completion.install` will install files into that dir.


I guess ZSH does not create site-functions dir during install and does not trigger the failure.

dunn: Right, my mistake.

Yeah, looks like for this to work Fish will need to change their install process.

gyakovlev: is it acceptable to modify Makefile (using `inreplace` or something else) inside Formula?
Im not sure if upstream will remove directory installation from Makefile.

dunn: If zsh does it, then it doesn't seem like it's totally impractical.  Could you file an issue with Fish and see if they're open to it?  Thanks!

zanchey: I think we'd be willing to change (though it won't be in until the next release; you could carry a patch until then?) but I'm not sure whether fish should just never create the directory (probably breaks some other packaging systems) or use `install || true` to soft-fail if the prefix is unwriteable; probably the latter.

DomT4: > #{pkgshare} will not work. The patch is about having version independent dirs outside of the Cellar so other Formulae can install completions and functions into that dir and fish can use those.
By default fish looking only into own Cellar share dirs for scripts. My PR is trying to make fish to look into #{HOMEBREW_PREFIX}/share/fish for completions and functions.

Yeah. By definition it wouldn't be sensible to use `pkgshare` here. It does make sense to use a real directory in `HOMEBREW_PREFIX`. We're doing something similar enough [already](https://github.com/Homebrew/brew/blob/c6022104f9df7fa1fc7bae596a770cdc073c8552/Library/Homebrew/keg.rb#L315-L324).

Does `fish` actually put anything in that directory at compile time, or is it purely used to teach `fish` to look for functions there as well as the default location?

gyakovlev: @zanchey Great to hear. soft-failing does sound sane.
it works on my system.
```
/usr/bin/install -c -m 755 -d /usr/local/share/fish/vendor_completions.d || true
install: chmod 755 /usr/local/share/fish/vendor_completions.d: Operation not permitted
/usr/bin/install -c -m 755 -d /usr/local/share/fish/vendor_functions.d || true
install: chmod 755 /usr/local/share/fish/vendor_functions.d: Operation not permitted
/usr/bin/install -c -m 755 -d /usr/local/share/fish/vendor_conf.d || true
install: chmod 755 /usr/local/share/fish/vendor_conf.d: Operation not permitted
```

@DomT4 It does not put anything, just creates empty dirs  and teaches fish to look there
 ```
$(INSTALL) -m 755 -d $(DESTDIR)$(extra_completionsdir)
$(INSTALL) -m 755 -d $(DESTDIR)$(extra_functionsdir)
$(INSTALL) -m 755 -d $(DESTDIR)$(extra_confdir)
```



can I just stop makefile from installing the dirs and mkpath instead? seems to be the cleanest solution. fish upstream can definitely soft fail extra dir creation and we'll create in our desired location ourselves. I've pushed another commit with above changes, not sure how to ask Jenkins to re-run the check.

Comment (gyakovlev) [inline Formula/fish.rb]: Fixed whitespace.

now formula is failing audit
```
ish:
  * Use pkgshare instead of (share/"fish")
  * Use pkgshare instead of (share/"fish")
  * Use pkgshare instead of (share/"fish")
```

How can I bypass this check? We have to create directly.

Comment (zanchey) [inline Formula/fish.rb]: It doesn't, and if the directory doesn't exist then there's no real problems, but other packaging systems would generally expect the directory to be owned by `fish`

> Does fish actually put anything in that directory at compile time, or is it purely used to teach fish to look for functions there as well as the default location?

It doesn't put anything there; it doesn't even need to exist, but the approach for other packaging systems would generally be for fish to "own" the directory. Homebrew and Nix (among others) do challenge the standard approach a bit, though.

**Comment by zanchey** [comment]
I've patched this in the fish master as fish-shell/fish-shell@9abbc5f - the rest I will leave to you :-)

**Comment by gyakovlev** [comment]
@zanchey great! thanks for your co-operation.

so in current state I just modify the Makefile to stop creating those dirs on install and create dirs with mkpath.

After fish releases https://github.com/fish-shell/fish-shell/commit/9abbc5f formula can stop patching Makefile and just create those dirs wherever we need them to be.

@dunn What can I else do to have this PR merged? Maybe squash it to be single commit?

File: Formula/fish.rb [hunk 2]
```diff
@@ -20,12 +20,34 @@ class Fish < Formula
...
+      SED=/usr/bin/sed
+    ]
     # In Homebrew's 'superenv' sed's path will be incompatible, so
     # the correct path is passed into configure here.
-    system "./configure", "--prefix=#{prefix}", "SED=/usr/bin/sed"
+
+    system "./configure", *args
+
+    # Skip extra dirs creation during install phase, we'll create them later with mkpath.
+    inreplace "Makefile" do |c|
```
dunn: Is this equivalent to the upstream patch?  Can you add a comment linking to that commit so whoever updates Fish to the next version knows to remove it?

Formula/fish.rb
```diff
@@ -20,12 +20,34 @@ class Fish < Formula
 
   depends_on "pcre2"
 
+  skip_clean "share/fish"
```
DomT4: We shouldn't need this. `skip_clean` applies to inside the Cellar, rather than out of it.

Formula/fish.rb
DomT4: > now formula is failing audit

Sigh. You've found another edge case in our audit code around `pkgshare`. _Throws self out of the window_. Ignore it, for now.

Formula/fish.rb
DomT4: Let's switch to using the upstream patch directly in a `stable do` block as per normal if we can.

Formula/fish.rb
```diff
@@ -20,12 +20,34 @@ class Fish < Formula
...
+    # Skip extra dirs creation during install phase, we'll create them later with mkpath.
+    inreplace "Makefile" do |c|
+      c.gsub! "$(INSTALL) -m 755 -d $(DESTDIR)$(extra_functionsdir)", ""
+      c.gsub! "$(INSTALL) -m 755 -d $(DESTDIR)$(extra_completionsdir)", ""
+      c.gsub! "$(INSTALL) -m 755 -d $(DESTDIR)$(extra_confdir)", ""
+    end
+
     system "make", "install"
+
+    (share/"fish/vendor_functions.d").mkpath
```
DomT4: You'll probably need to shunt these to `post_install`. Fairly sure `brew` won't bother including them in the bottle if they're empty.

gyakovlev:
finally found some time to fix fix and test everything. Also squashed into single commit.

works ok for me, but still failing audit

```
fish:
  * Use pkgshare instead of (share/"fish")
  * Use pkgshare instead of (share/"fish")
  * Use pkgshare instead of (share/"fish")
```

Formula/fish.rb
```diff
@@ -39,7 +54,30 @@ def caveats; <<-EOS.undent
     EOS
   end
 
+  def post_install
+    (share/"fish/vendor_functions.d").mkpath
```
dunn: These are what the audit is complaining about; they can be `pkgshare/"vendor_functions.d"` etc.

gyakovlev: @dunn We already discussed that, check your comments earlier.
using pkgshare defeats purpose of this PR.
I use share to have system-wide directory so other formulas can put files in that dir .

dunn: Yeah we can't use `pkgshare` instead of `"#{HOMEBREW_PREFIX}/share/fish"` but this `post_install` method is installing to `share/"fish"` which _is_ `pkgshare`.

gyakovlev: sorry, my bad. I thought pkgshare will go into cellar.
Pushed an update, now passing audit.

Formula/fish.rb
```diff
@@ -1,9 +1,16 @@
...
-  sha256 "328acad35d131c94118c1e187ff3689300ba757c4469c8cc1eaa994789b98664"
+
+  stable do
+    url "https://fishshell.com/files/2.3.1/fish-2.3.1.tar.gz"
+    mirror "https://github.com/fish-shell/fish-shell/releases/download/2.3.1/fish-2.3.1.tar.gz"
+    sha256 "328acad35d131c94118c1e187ff3689300ba757c4469c8cc1eaa994789b98664"
+
+    # Skip extra dirs creation during install phase patch for current stable. To be removed on the next fish release.
+    # As discussed in https://github.com/Homebrew/homebrew-core/pull/2813
+    patch :DATA
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

MikeMcQuaid: Ah, I see it's done in https://github.com/fish-shell/fish-shell/commit/9abbc5f. You could use https://github.com/fish-shell/fish-shell/commit/9abbc5f.patch if that'll apply otherwise just note https://github.com/fish-shell/fish-shell/commit/9abbc5f in a comment above. Thanks!

gyakovlev:
added comment as @MikeMcQuaid suggested [here](https://github.com/Homebrew/homebrew-core/pull/2813#discussion_r73847093) 
I cannot use upstream patch fish-shell/fish-shell@9abbc5f directly, so took only the lines we need.

MikeMcQuaid:

For future reference the preferred commit message format for simple version updates is `fish: use version independent config dirs.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md). Thanks!

PR #2830, opened 2016-07-08
pdnsd: add plist file

rakkesh:

---

pdnsd: added a plist file to start the service on system boot.

### DIFF
--- a/Formula/pdnsd.rb
+++ b/Formula/pdnsd.rb
@@ -24,6 +24,50 @@ class Pdnsd < Formula
     system "make", "install"
   end
 
+  def caveats; <<-EOS.undent
+    This install of "pdnsd" expects config files to be in #{etc}
+    All state files (status and cache) are stored in #{var}/cache/pdnsd.
+
+    pdnsd needs to run as root since it listens on privileged ports.
+
+    Sample config file can be found at #{etc}/pdnsd.conf.sample.
+
+    Note that you must create the config file before starting the service,
+    and change ownership to "root" or pdnsd will refuse to run:
+      sudo chown root #{etc}/pdnsd.conf
+
+    For other related utilities, e.g. pdnsd-ctl, to run, change the ownership
+    to the user (default: nobody) running the service:
+      sudo chown -R nobody #{var}/log/pdnsd.log #{var}/cache/pdnsd
+    EOS
+  end
+
+  plist_options :startup => true, :manual => "sudo pdnsd"
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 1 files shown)


Formula/pdnsd.rb
```diff
@@ -24,6 +24,43 @@ def install
     system "make", "install"
   end
 
+  def caveats; <<-EOS.undent
+    Before starting "pdnsd" using "brew services" or manually, make sure that,
+    #{etc}/pdnsd.conf, #{var}/log/pdnsd.log and directory #{var}/cache/pdnsd
```
MikeMcQuaid: This is going to break future upgrades when some of these files/directories can no longer be overwritten. Can we work on a better solution here?

rakkesh: `chown` `nobody` would not break an upgrade in this case. FWIW, I can still edit the configuration file with my account without the need for `sudo`, after which I have to `chown` to `nobody` again. So, a `brew upgrade` can overwrite the configuration file; `log` and `cache` need not be dealt with during any of `brew` operations.

I did not find another way to deal with `pdnsd` which needs access to a privileged port(53). Also, running `pdnsd` with the current login user instead of user `nobody` seemed risky. If this seems like trouble for maintenance, I would rather close this PR.

rakkesh: > FWIW, I can still edit the configuration file with my account without the need for sudo, after which I have to chown

`#{etc}/pdnsd.conf` needs to be owned by `root` not `nobody`. But the above statement is still true.

MikeMcQuaid: So nothing installed by a non-`nobody` user into those directories, ok, sees fine then 👍

Formula/pdnsd.rb
```diff
@@ -24,6 +24,43 @@ def install
...
+    are all owned by "nobody".
+
+    sudo chown nobody #{etc}/pdnsd.conf
+    sudo chown nobody #{var}/log/pdnsd.log
+    sudo chown -R nobody #{var}/cache/pdnsd
```
MikeMcQuaid: Could do all these three in a single command. Also, indent these by two spaces.

Formula/pdnsd.rb
```diff
@@ -24,6 +24,43 @@ def install
     system "make", "install"
   end
 
+  def caveats; <<-EOS.undent
+    Before starting "pdnsd" using "brew services" or manually, make sure that,
+    #{etc}/pdnsd.conf, #{var}/log/pdnsd.log and directory #{var}/cache/pdnsd
+    are all owned by "nobody".
```
MikeMcQuaid: `owned by "nobody".` -> `owned by "nobody":` and indent the lines below by two spaces.

ilovezfs:
```
==> brew audit pdnsd
==> FAILED
Error: 1 problem in 1 formula
pdnsd:
  * 29: Trailing whitespace was found
```

MikeMcQuaid:
Thanks again @rakkesh!

PR #2840, opened 2016-07-08
duti: add 10.12 support

zmwangx:

---

Another year, another Darwin version, [another patch](https://github.com/moretension/duti/pull/20). Upstream has abandoned the project — last year's patch hasn't been merged yet; but after applying the patch in this PR, `duti` still compiles and runs perfectly fine on 10.12 PB1.

It is also worth mentioning that `duti` is reasonably popular — [286 bottle downloads within the last month alone](https://bintray.com/homebrew/bottles/duti/view#statistics) — so I don't think it should be boneyarded.

### DIFF
--- a/Formula/duti.rb
+++ b/Formula/duti.rb
@@ -15,11 +15,12 @@ class Duti < Formula
 
   depends_on "autoconf" => :build
 
-  # Add hardcoded SDK path for El Capitan. See https://github.com/moretension/duti/pull/13
-  if MacOS.version == :el_capitan
+  # Add hardcoded SDK path for El Capitan or later.
+  # See https://github.com/moretension/duti/pull/20.
+  if MacOS.version >= :el_capitan
     patch do
-      url "https://github.com/moretension/duti/pull/13.patch"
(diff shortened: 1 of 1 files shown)


DomT4:
So, given the package's semi-decent usage & that the patch is only really overcoming a build system quirk upstream rather than runtime functionality breakage (As far as we're aware) I'll take this, but if we're back in this seat in a year's time this formula will probably end up in the boneyard.

It's perhaps worth hunting around for an active fork that has some community backing, if people want it to definitely outlive 10.12.

Merged in https://github.com/Homebrew/homebrew-core/commit/feb59d71496b8c830ad624c484e77ea485bcbfd8. Thanks @zmwangx!

zmwangx:
> It's perhaps worth hunting around for an active fork that has some community backing

Looking at the [network](https://github.com/moretension/duti/network) I don't see any. Neither does a Google search yield anything interesting, given my google-fu.

I have no problem keeping this updated in my personal fork when the time comes, but people who are less familiar with how to brew stuff themselves will hate it.

PR #2845, opened 2016-07-08
direvent 5.1 (new formula)

fornwall:

### DIFF
--- a/Formula/direvent.rb
+++ b/Formula/direvent.rb
@@ -2,6 +2,7 @@ class Direvent < Formula
   desc "Monitors events in the file system directories"
   homepage "http://www.gnu.org.ua/software/direvent/direvent.html"
   url "https://ftpmirror.gnu.org/direvent/direvent-5.1.tar.gz"
+  mirror "https://ftp.gnu.org/gnu/direvent/direvent-5.1.tar.gz"
   sha256 "c461600d24183563a4ea47c2fd806037a43354ea68014646b424ac797a959bdb"
 
   def install
@@ -14,6 +15,6 @@ class Direvent < Formula
   end
 
   test do
-    system "#{bin}/direvent", "--version"
+    assert_match version.to_s, shell_output("#{bin}/direvent --version")
   end
 end
--- a/Formula/direvent.rb
+++ b/Formula/direvent.rb
@@ -1,0 +1,19 @@
+class Direvent < Formula
+  desc "Monitors events in the file system directories"
+  homepage "http://www.gnu.org.ua/software/direvent/direvent.html"
+  url "https://ftpmirror.gnu.org/direvent/direvent-5.1.tar.gz"
+  sha256 "c461600d24183563a4ea47c2fd806037a43354ea68014646b424ac797a959bdb"
+
+  def install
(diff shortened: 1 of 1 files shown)


Formula/direvent.rb
```diff
@@ -0,0 +1,19 @@
+class Direvent < Formula
+  desc "Monitors events in the file system directories"
+  homepage "http://www.gnu.org.ua/software/direvent/direvent.html"
+  url "https://ftpmirror.gnu.org/direvent/direvent-5.1.tar.gz"
```
DomT4: Can you add the usual mirror as well? Thanks!
```ruby
mirror "https://ftp.gnu.org/direvent/direvent-5.1.tar.gz"
```

Formula/direvent.rb
```diff
@@ -0,0 +1,19 @@
...
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "-C", "tests", "check-local"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/direvent", "--version"
```
DomT4: Let's assert on the version here. It's not much stronger, but it's something:
```ruby
assert_match version.to_s, shell_output("#{bin}/direvent")
```

Even better though if you could think up a functionality test? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test.

ilovezfs: Yeah, I'd like to see something that demonstrates this actually works with macOS file systems since we know how much testing gnu stuff gets on darwin.

Formula/direvent.rb
fornwall: Done!

Formula/direvent.rb
fornwall: There is a test suite run by the `make -C tests check-local` command during install - hopefully that should be enough?

Formula/direvent.rb
```diff
@@ -0,0 +1,20 @@
...
+  url "https://ftpmirror.gnu.org/direvent/direvent-5.1.tar.gz"
+  mirror "https://ftp.gnu.org/gnu/direvent/direvent-5.1.tar.gz"
+  sha256 "c461600d24183563a4ea47c2fd806037a43354ea68014646b424ac797a959bdb"
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "-C", "tests", "check-local"
```
DomT4: What's the difference between `make check` & `make -C tests check-local`? Is the latter preferable in some way?

DomT4: Ping on this question?

apjanke: `make check` is a lot more extensive, and it passes for me locally. Gonna try switching to that.

dunn:
This needs to be rebased on master.

apjanke:
Superseded by #4151 (with author credit preserved).

Thanks for your contribution to Homebrew!

PR #2846, opened 2016-07-08
minio-mc 20160603184837 (new formula)

nl5887:
- [ x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

### DIFF
--- a/Formula/minio-mc.rb
+++ b/Formula/minio-mc.rb
@@ -23,6 +23,6 @@ class MinioMc < Formula
 
   test do
     system bin/"mc", "mb", testpath/"test"
-    assert File.exist?(testpath/"test")
+    assert_predicate testpath/"test", :exist?
   end
 end
--- a/Formula/minio-mc.rb
+++ b/Formula/minio-mc.rb
@@ -5,7 +5,8 @@ class MinioMc < Formula
   version "20160713214605"
   sha256 "611444a66ea3d2cc8fbb821147bcd3995c1956ef2aa9d62cd35f6d07f3f972cf"
   head "https://github.com/minio/mc.git"
-  conflicts_with "midnight-commander"
+
+  conflicts_with "midnight-commander", :because => "Both install a `mc` binary"
 
   depends_on "go" => :build
 
@@ -21,7 +22,7 @@ class MinioMc < Formula
   end
 
   test do
-    system bin/"mc", "mb", (testpath"/test")
+    system bin/"mc", "mb", testpath/"test"
     assert File.exist?(testpath/"test")
   end
 end
--- a/Formula/minio-mc.rb
+++ b/Formula/minio-mc.rb
@@ -1,10 +1,11 @@
 class MinioMc < Formula
   desc "ls, cp, mkdir, diff and rsync for filesystems and object storage"
   homepage "https://github.com/minio/mc"
-  url "https://github.com/minio/mc/archive/RELEASE.2016-06-03T18-48-37Z.tar.gz"
-  version "20160603184837"
-  sha256 "ac27ce934d5d71a1150eed69641a7ba801f3ae90c064ee944c154e311a5e6dbc"
+  url "https://github.com/minio/mc/archive/RELEASE.2016-07-13T21-46-05Z.tar.gz"
+  version "20160713214605"
+  sha256 "611444a66ea3d2cc8fbb821147bcd3995c1956ef2aa9d62cd35f6d07f3f972cf"
   head "https://github.com/minio/mc.git"
+  conflicts_with "midnight-commander"
 
   depends_on "go" => :build
 
@@ -14,12 +15,13 @@ class MinioMc < Formula
     clipath = buildpath/"src/github.com/minio/mc"
     clipath.install Dir["*"]
 
-    cd buildpath/"src/github.com/minio/mc/" do
+    cd clipath do
       system "go", "build", "-o", bin/"mc"
     end
   end
 
   test do
-    system bin/"mc", "version"
+    system bin/"mc", "mb", (testpath"/test")
+    assert File.exist?(testpath/"test")
   end
 end
--- a/Formula/minio-mc.rb
+++ b/Formula/minio-mc.rb
@@ -1,0 +1,30 @@
+require "language/go"
+
+class MinioMc < Formula
+  desc "ls, cp, mkdir, diff and rsync for filesystems and object storage"
+  homepage "https://github.com/minio/mc"
+  url "https://github.com/minio/mc/archive/RELEASE.2016-06-03T18-48-37Z.tar.gz"
+  version "20160603184837"
+  sha256 "ac27ce934d5d71a1150eed69641a7ba801f3ae90c064ee944c154e311a5e6dbc"
+  desc "object storage server compatible with Amazon S3"
+  head "https://github.com/minio/mc.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    clipath = buildpath/"src/github.com/minio/mc"
+    clipath.install Dir["*"]
+
+    cd buildpath/"src/github.com/minio/mc/" do
+      system "go", "build", "-o", buildpath/"mc"
+    end
+
+    bin.install buildpath/"mc"
+  end
+
+  test do
+    system "#{bin}/mc"
+  end
+end
--- a/Formula/minio-mc.rb
+++ b/Formula/minio-mc.rb
@@ -20,6 +20,6 @@ class MinioMc < Formula
   end
 
   test do
-    system bin/"mc","version"
+    system bin/"mc", "version"
   end
 end
--- a/Formula/minio-mc.rb
+++ b/Formula/minio-mc.rb
@@ -1,12 +1,9 @@
-require "language/go"
-
 class MinioMc < Formula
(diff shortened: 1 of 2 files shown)


Formula/minio-mc.rb
```diff
@@ -0,0 +1,30 @@
+require "language/go"
+
+class MinioMc < Formula
+  desc "ls, cp, mkdir, diff and rsync for filesystems and object storage"
+  homepage "https://github.com/minio/mc"
+  url "https://github.com/minio/mc/archive/RELEASE.2016-06-03T18-48-37Z.tar.gz"
+  version "20160603184837"
+  sha256 "ac27ce934d5d71a1150eed69641a7ba801f3ae90c064ee944c154e311a5e6dbc"
+  desc "object storage server compatible with Amazon S3"
```
DomT4: You've added two `desc`s?

Formula/minio-mc.rb
```diff
@@ -0,0 +1,30 @@
+require "language/go"
```
DomT4: Don't need this unless you're using `go_resource`s.

Formula/minio-mc.rb
```diff
@@ -0,0 +1,30 @@
...
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    clipath = buildpath/"src/github.com/minio/mc"
+    clipath.install Dir["*"]
+
+    cd buildpath/"src/github.com/minio/mc/" do
+      system "go", "build", "-o", buildpath/"mc"
```
DomT4: Rather than installing it to the `buildpath` and then moving it to the `bin` you can just do `"-o", bin/"mc"` here.

Formula/minio-mc.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    bin.install buildpath/"mc"
+  end
+
+  test do
+    system "#{bin}/mc"
```
DomT4: Can this test be modified to do something more substantial? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

nl5887:
Regarding the tests, to really run some tests we're depending on a Minio server, don't think we want that. Currently changed to test `mc version`.

Formula/minio-mc.rb
```diff
@@ -0,0 +1,25 @@
...
+    clipath = buildpath/"src/github.com/minio/mc"
+    clipath.install Dir["*"]
+
+    cd buildpath/"src/github.com/minio/mc/" do
+      system "go", "build", "-o", bin/"mc"
+    end
+  end
+
+  test do
+    system bin/"mc", "version"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. In this case could you do a `cp` or `mkdir` and check it worked? Thanks!

MikeMcQuaid:
> Regarding the tests, to really run some tests we're depending on a Minio server, don't think we want that. Currently changed to test mc version.

Sorry, just saw this, it'll be fine as-is.

Formula/minio-mc.rb
```diff
@@ -0,0 +1,25 @@
...
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    clipath = buildpath/"src/github.com/minio/mc"
+    clipath.install Dir["*"]
+
+    cd buildpath/"src/github.com/minio/mc/" do
+      system "go", "build", "-o", bin/"mc"
```
UniqMartin: This is going to conflict with the `mc` binary from the `midnight-commander` formula. We'll need to add a `conflicts_with` line to both formulae or (if this can be safely done) rename this binary.

Formula/minio-mc.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    clipath = buildpath/"src/github.com/minio/mc"
+    clipath.install Dir["*"]
+
+    cd buildpath/"src/github.com/minio/mc/" do
```
UniqMartin: Please use `cd clipath do` to avoid some unnecessary repetition.

Formula/minio-mc.rb
```diff
@@ -0,0 +1,27 @@
+class MinioMc < Formula
+  desc "ls, cp, mkdir, diff and rsync for filesystems and object storage"
+  homepage "https://github.com/minio/mc"
+  url "https://github.com/minio/mc/archive/RELEASE.2016-07-13T21-46-05Z.tar.gz"
+  version "20160713214605"
+  sha256 "611444a66ea3d2cc8fbb821147bcd3995c1956ef2aa9d62cd35f6d07f3f972cf"
+  head "https://github.com/minio/mc.git"
+  conflicts_with "midnight-commander"
```
UniqMartin: Please leave an empty line before this line and it's nicer if you provide a reason as a second argument (see also other formula that use this), e.g.:

```ruby
conflicts_with "midnight-commander", :because => "Both install a `mc` binary"
```

Please also add a similar `conflicts_with` line to the `midnight-commander` formula and include it in your PR. (The conflicts need to be declared in all involved formulae.)

Formula/minio-mc.rb
```diff
@@ -0,0 +1,27 @@
...
+    clipath = buildpath/"src/github.com/minio/mc"
+    clipath.install Dir["*"]
+
+    cd clipath do
+      system "go", "build", "-o", bin/"mc"
+    end
+  end
+
+  test do
+    system bin/"mc", "mb", (testpath"/test")
```
UniqMartin: You have a syntax error in this line and this is most likely the reason for the test failure (the `/` should be outside of the string). You can also omit the parentheses, i.e.:

```ruby
system bin/"mc", "mb", testpath/"test"
```

Formula/minio-mc.rb
```diff
@@ -0,0 +1,27 @@
...
+    assert File.exist?(testpath/"test")
```
UniqMartin: Please use the following instead, which will result in a clearer message if the assertion fails:

```ruby
assert_predicate testpath/"test", :exist?
```

Formula/minio-mc.rb
```diff
@@ -0,0 +1,28 @@
...
+    clipath.install Dir["*"]
+
+    cd clipath do
+      system "go", "build", "-o", bin/"mc"
+    end
+  end
+
+  test do
+    system bin/"mc", "mb", testpath/"test"
+    assert File.exist?(testpath/"test")
```
UniqMartin: I guess you might have overlooked this, because I commented after you had already pushed an update, hence I'm reposting this comment:

Please use the following instead, which will result in a clearer message if the assertion fails:

```ruby
assert_predicate testpath/"test", :exist?
```

UniqMartin:
The test is now passing and the build only “fails” because of an issue with Midnight Commander (but that's an issue you can safely ignore, as it is not your fault and not related to this formula).

dunn:
🔔

Formula/minio-mc.rb
DomT4: I'd kind of prefer that we don't start rewriting these unless we add it to the strict audit or run a giant `sed` through the core. Code fragmentation is already a nightmare when it comes to formulae.

MikeMcQuaid: Agreed to an audit check but I still think it's worth making this change for now. Go ahead with @UniqMartin's suggestion, thanks @nl5887.

nl5887:
Ok, using assert_predicate now.

Formula/minio-mc.rb
UniqMartin: Apologies for the bad advice! I've been influenced by work on the package manager itself and failed to realize that `assert_predicate` can be unavailable in formulae. This assert is part of `MiniTest` that doesn't ship with older Rubies (like Ruby 1.8.7 that we use on our Mavericks bot). This is also the reason why the formula test is failing on 10.9 currently.

MikeMcQuaid: Sorry from me too!

UniqMartin:
Sorry again for the bad advice and for the delay in handling this! I dropped the commit that switched to `assert_predicate`, made some last-minute adjustments to your commits as our audit has been tightened up a bit meanwhile, and then squashed everything to two commits.

Merged in badeeeb5aced1cf9ef0cc9332453ac56a0db9df8 and f424abaa5b36d56dffb5eddbbe4c36adcefe5626. Thank you for this contribution to Homebrew, @nl5887! :tada:

nl5887:
No worries, can happen. Maybe its me and my unfamiliarity with Ruby, but there are a lot of best practices (use chomp, use backticks in install) I noticed that are not really documented somewhere (or I don't know where to look). Maybe it will be good to extract those from the comments on the PR or something.

PR #2856, opened 2016-07-09
bacula-fd 7.4.2

ilovezfs:
Also,
- don't set user name in the plist
- add a test

### DIFF
--- a/Formula/bacula-fd.rb
+++ b/Formula/bacula-fd.rb
@@ -1,8 +1,8 @@
 class BaculaFd < Formula
   desc "Network backup solution"
   homepage "http://www.bacula.org/"
-  url "https://downloads.sourceforge.net/project/bacula/bacula/7.4.1/bacula-7.4.1.tar.gz"
-  sha256 "d0874e94c0534274455efd91d6f9477c57c7f14882138f03b3572eff2660e861"
+  url "https://downloads.sourceforge.net/project/bacula/bacula/7.4.2/bacula-7.4.2.tar.gz"
+  sha256 "0c5f7fe1d35fc2aa4ffba73e977b710b955b1fe3e45ab90c67b47b279152ad67"
 
   bottle do
     sha256 "57d67afab43a0bbfb8c47f1db23467feaebd08a9da8510bcc95a6c9646fd8371" => :el_capitan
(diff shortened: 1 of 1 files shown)


Formula/bacula-fd.rb
```diff
@@ -18,24 +18,19 @@ def install
...
 
     system "make"
     system "make", "install"
 
-    # Ensure var/run exists:
-    (var + "run").mkpath
-
-    # Create the working directory:
-    (var + "lib/bacula").mkpath
+    (var/"run").mkpath
```
DomT4: These should be shunted into `post_install`.

ilovezfs: Can you explain the logic for those given they're in the .bottle hidden dir?

ilovezfs: Or is it because it's var/run not var/run/bacula so it doesn't make it in there?

Formula/bacula-fd.rb
```diff
@@ -18,24 +18,19 @@ def install
     #   (conio support not tested)
     # * working directory in /var/lib/bacula, reasonable place that
     #   matches Debian's location.
-    readline = Formula["readline"].opt_prefix
     system "./configure", "--prefix=#{prefix}",
-                          "--sbindir=#{bin}",
                           "--with-working-dir=#{var}/lib/bacula",
-                          "--with-pid-dir=#{HOMEBREW_PREFIX}/var/run",
+                          "--with-pid-dir=#{var}",
```
ilovezfs: I suppose I shouldn't have deleted the "run" there.

Formula/bacula-fd.rb
DomT4: At this point, no. I remembered it once, but I didn't retain the reasoning. You'll have to ask @MikeMcQuaid.

ilovezfs: OK, thanks.

ilovezfs:
PR refreshed.

Formula/bacula-fd.rb
```diff
@@ -18,24 +18,22 @@ def install
     #   (conio support not tested)
     # * working directory in /var/lib/bacula, reasonable place that
     #   matches Debian's location.
-    readline = Formula["readline"].opt_prefix
     system "./configure", "--prefix=#{prefix}",
-                          "--sbindir=#{bin}",
```
MikeMcQuaid: I'd rather we left this as-is for now rather than janking around current users.

MikeMcQuaid:
One comment otherwise 👍

Formula/bacula-fd.rb
ilovezfs: Not that discussion again!
*ilovezfs curls up into the fetal position*

MikeMcQuaid: Sorry! I mean until we've decided on a resolution to that discussion: let's leave formulae that are using `bin`/`sbin` as-is.

ilovezfs: OK.

ilovezfs:
PR refreshed.

DomT4:
:shipit:

MikeMcQuaid:
👍

PR #2861, opened 2016-07-09
doxygen: use Qt 5

jakepetroules:

---

Qt 5 works well and unlike Qt 4, properly supports new macOS versions.

### DIFF
--- a/Formula/doxygen.rb
+++ b/Formula/doxygen.rb
@@ -14,19 +14,20 @@ class Doxygen < Formula
   end
 
   option "with-graphviz", "Build with dot command support from Graphviz."
-  option "with-doxywizard", "Build GUI frontend with qt support."
+  option "with-qt5", "Build GUI frontend with Qt support."
   option "with-libclang", "Build with libclang support."
 
   deprecated_option "with-dot" => "with-graphviz"
+  deprecated_option "with-doxywizard" => "with-qt5"
 
   depends_on "cmake" => :build
   depends_on "graphviz" => :optional
-  depends_on "qt" if build.with? "doxywizard"
+  depends_on "qt5" => :optional
   depends_on "llvm" => "with-clang" if build.with? "libclang"
 
   def install
     args = std_cmake_args
-    args << "-Dbuild_wizard=ON" if build.with? "doxywizard"
+    args << "-Dbuild_wizard=ON" if build.with? "qt5"
     args << "-Duse_libclang=ON -DLLVM_CONFIG=#{Formula["llvm"].opt_bin}/llvm-config" if build.with? "libclang"
 
     mkdir "build" do


ilovezfs:
:+1:

ilovezfs:
```
option "with-doxywizard", "Build GUI frontend with qt support."
```

is non-default so the build needs to be confirmed locally.

UniqMartin:
It would be nice to deprecate the `--with-doxywizard` option in favor of a `--with-qt5` option, as this plays more nicely with our dependency resolution code. Here's the suggested diff:

```diff
diff --git i/Formula/doxygen.rb w/Formula/doxygen.rb
index d50ff85f..da2f5c8f 100644
--- i/Formula/doxygen.rb
+++ w/Formula/doxygen.rb
@@ -14,19 +14,20 @@ class Doxygen < Formula
   end

   option "with-graphviz", "Build with dot command support from Graphviz."
-  option "with-doxywizard", "Build GUI frontend with qt support."
+  option "with-qt5", "Build GUI frontend with Qt support."
   option "with-libclang", "Build with libclang support."

   deprecated_option "with-dot" => "with-graphviz"
+  deprecated_option "with-doxywizard" => "with-qt5"

   depends_on "cmake" => :build
   depends_on "graphviz" => :optional
-  depends_on "qt" if build.with? "doxywizard"
+  depends_on "qt5" => :optional
   depends_on "llvm" => "with-clang" if build.with? "libclang"

   def install
     args = std_cmake_args
-    args << "-Dbuild_wizard=ON" if build.with? "doxywizard"
+    args << "-Dbuild_wizard=ON" if build.with? "qt5"
     args << "-Duse_libclang=ON -DLLVM_CONFIG=#{Formula["llvm"].opt_bin}/llvm-config" if build.with? "libclang"

     mkdir "build" do
```

DomT4:
Agree with @UniqMartin. More broad notation: If anything has an existing mandatory or recommended `qt` dependency the formula `revision` should be bumped as well. That's not the case here, but it will be elsewhere I expect.

ilovezfs:
Possibly optional too since it's such a big change, no?

DomT4:
Unless we have hard analytics data pointing to a lot of people activating the `qt` option of `<formula>`, I'd prefer to avoid the optional revision bumps. Revisions are a very blunt instrument, and we try to use them fairly sparingly _(In theory)_.

ilovezfs:
>Unless we have hard analytics data

Yes, I agree with that, and it's probably worth checking to see for which formulae, if any, that's the case. I don't know if that's searchable on an option -> formulae basis or if it's only visible one-by-one formula -> option?

DomT4:
You can search the formula name & you'll get a list of everything executed. _As far as I'm aware_ we don't keep tabs on what is dragged in from other formulae with `depends_on` statements, so these should all be direct installations.

screen shot 2016-07-09 at 17 15 51

ilovezfs:
It is amazing how few people use options. Makes me want to get rid of every option everywhere lol

DomT4:
There are some cases where options are wildly popular, emacs for example is tilted _heavily_ in favour of option-using installations. But yeah, overall, they serve a niche and we pour a lot of time into making that niche happy.

You get to number 81 in our most installed formulae before you hit a single option invocation, which is `libpng --universal`. The most popular non-universal formula-with-option is `vim --with-lua` down in 189th position.

DomT4:
First one done & shipped. Merged in https://github.com/Homebrew/homebrew-core/commit/03040ab77566fcdbf672cea330fff8ec8ce7a162. Thanks for getting the ball rolling on these @jakepetroules!

PR #2863, opened 2016-07-09
quassel: use Qt 5

jakepetroules:

---

Qt 5 works well and unlike Qt 4, properly supports new macOS versions.

### DIFF
--- a/Formula/quassel.rb
+++ b/Formula/quassel.rb
@@ -23,11 +23,7 @@ class Quassel < Formula
 
   depends_on "cmake" => :build
   depends_on "pkg-config" => :build
-
-  # Official binary packages upstream now built against qt5 by default. But source
-  # packages default to qt4 *for now*, and Homebrew prioritises qt5 in PATH due to keg_only.
-  depends_on "qt5" => :optional
-  depends_on "qt" => :recommended
+  depends_on "qt5" => :recommended
 
   needs :cxx11
 


ilovezfs:
@jakepetroules looks like it now has a 0 exit code for `-v`. So the test needs to be updated accordingly:
```
 assert_match "Quassel IRC", shell_output("#{bin}/quasselcore")
```
Note that since it's a substring match we prefer a string for the first arg instead of a regex which is why I also made that change above.

ilovezfs:
@jakepetroules Also looks like there's a new version: http://www.quassel-irc.org/ (0.12.4)

So the Qt5 patch can be removed too since it looks like it came from upstream so presumably it's already in the new version now.

UniqMartin:
```console
$ brew install quassel --without-qt5
[…snip…]
==> cmake -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFIX=/opt/brewery/qfive/
Last 15 lines from /opt/brewery/qfive/.brew-logs/quassel/01.cmake:
-- Check for working CXX compiler: /opt/brewery/qfive/Library/ENV/4.3/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Using CMake 3.6.0
-- Building for Qt4...
CMake Error at /opt/brewery/qfive/Cellar/cmake/3.6.0/share/cmake/Modules/FindQt4.cmake:1328 (message):
  Found unsuitable Qt version "" from NOTFOUND, this code requires Qt 4.x
  CMakeLists.txt:324 (find_package)


-- Configuring incomplete, errors occurred!
See also "/tmp/quassel-20160709-25786-1kwq81i/quassel-0.12.2/CMakeFiles/CMakeOutput.log".
```

Qt 5 is no longer just a recommended dependency but in fact a required dependency.

----

Even after fixing the test as suggested by @ilovezfs:

```console
$ brew install quassel
$ brew test quassel
Testing quassel
==> Using the sandbox
==> /opt/brewery/qfive/Cellar/quassel/0.12.2/bin/quasselcore -v
Error: quassel: failed
Expected /Quassel\ IRC/ to match "quassel v0.12.2 (dist-9c5e6c6)\n".
```

Not only did the return value change, but also the output when switching from Qt 4 to Qt 5.

MikeMcQuaid:
Closing this as we've moved it to Homebrew Cask.

PR #2882, opened 2016-07-10
gx-go 1.2.1

ilovezfs:
- switch to tarball
- indent :revision in go_resource blocks two spaces beyond the url
- remove godep and gx dependencies


DomT4:
So many `go_resource`s 😢. LGTM.

PR #2895, opened 2016-07-11
pdnsrec 4.0.1

Habbie:

---
- update to 4.0.0
- add HEAD support
- remove example config as upstream ships one by default now

### DIFF
--- a/Formula/pdnsrec.rb
+++ b/Formula/pdnsrec.rb
@@ -1,8 +1,8 @@
 class Pdnsrec < Formula
   desc "Non-authoritative/recursing DNS server"
   homepage "https://www.powerdns.com/recursor.html"
-  url "https://downloads.powerdns.com/releases/pdns-recursor-3.7.3.tar.bz2"
-  sha256 "859ca6071147dd2e2ac1b2a5c3d5c2cbff0f5cbc501660db4259e7cbf27fea11"
+  url "https://downloads.powerdns.com/releases/pdns-recursor-4.0.1.tar.bz2"
+  sha256 "472db541307c8ca83a846d260ecfc854fd8e879c1bb2ce5683a8df5d21e860b0"
 
   bottle do
     cellar :any_skip_relocation
@@ -12,40 +12,36 @@ class Pdnsrec < Formula
     sha256 "a9faf9edf0e71de5e9ec9fa70d27811f353a8451b41746208fbdc0d592aa5910" => :mountain_lion
   end
 
-  depends_on :macos => :lion
+  depends_on "pkg-config" => :build
   depends_on "boost"
-  depends_on "lua" => :optional
+  depends_on "openssl"
+  depends_on "lua"
+
+  needs :cxx11
+
+  fails_with :clang do
+    build 600
+    cause "incomplete C++11 support"
+  end
 
   def install
-    # Set overrides using environment variables
-    ENV["DESTDIR"] = "#{prefix}"
-    ENV["OPTFLAGS"] = "-O0"
-    ENV.O0
+    ENV.cxx11
 
-    # Include Lua if requested
-    if build.with? "lua"
-      ENV["LUA"] = "1"
-      ENV["LUA_CPPFLAGS_CONFIG"] = "-I#{Formula["lua"].opt_include}"
-      ENV["LUA_LIBS_CONFIG"] = "-llua"
-    end
+    args = %W[
+      --prefix=#{prefix}
+      --sysconfdir=#{etc}/powerdns
+      --disable-silent-rules
+      --with-boost=#{Formula["boost"].opt_prefix}
+      --with-openssl=#{Formula["openssl"].opt_prefix}
+      --with-lua
+    ]
 
-    # Adjust hard coded paths in Makefile
-    inreplace "Makefile.in", "/usr/sbin/", "#{sbin}/"
-    inreplace "Makefile.in", "/usr/bin/", "#{bin}/"
-    inreplace "Makefile.in", "/etc/powerdns/", "#{etc}/powerdns/"
-    inreplace "Makefile.in", "/var/run/", "#{var}/run/"
+    system "./configure", *args
+    system "make", "install"
+  end
 
-    # Compile
-    system "./configure"
-    system "make"
-
-    # Do the install manually
-    bin.install "rec_control"
-    sbin.install "pdns_recursor"
-    man1.install "pdns_recursor.1", "rec_control.1"
-
-    # Generate a default configuration file
-    (prefix/"etc/powerdns").mkpath
-    system "#{sbin}/pdns_recursor --config > #{prefix}/etc/powerdns/recursor.conf"
+  test do
+    output = shell_output("#{sbin}/pdns_recursor --version 2>&1")
+    assert_match "PowerDNS Recursor #{version}", output
   end
 end


DomT4:
Apparently it wants OpenSSL now:
```
checking for openssl/crypto.h in /usr/local/opt/openssl... no
checking whether compiling and linking against OpenSSL works... no
configure: error: OpenSSL not found
```

If that's assuming Homebrew lives in `/usr/local` we'll probably want to feed it a `--with-openssl` style argument, if possible, here.

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
     sha256 "a9faf9edf0e71de5e9ec9fa70d27811f353a8451b41746208fbdc0d592aa5910" => :mountain_lion
   end
 
+  head do
+    url "https://github.com/powerdns/pdns.git"
+
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+    depends_on "libtool"  => :build
```
DomT4: Please don't bother lining up the `:build` tags. I don't think it makes things that much more readable & some linters/etc groan about it.

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
+  depends_on "pkg-config" => :build if build.with? "lua"
+
   def install
-    # Set overrides using environment variables
-    ENV["DESTDIR"] = "#{prefix}"
-    ENV["OPTFLAGS"] = "-O0"
-    ENV.O0
+    args = %W[
+      --prefix=#{prefix}
+      --sysconfdir=#{prefix}/etc/powerdns
```
DomT4: Should be `#{etc}/powerdns` unless we want to smash people's configuration files every revision/update.

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
+      --with-openssl=#{Formula["openssl"].opt_prefix}
+    ]
 
     # Include Lua if requested
     if build.with? "lua"
-      ENV["LUA"] = "1"
-      ENV["LUA_CPPFLAGS_CONFIG"] = "-I#{Formula["lua"].opt_include}"
-      ENV["LUA_LIBS_CONFIG"] = "-llua"
+      args << "--with-lua"
```
DomT4: Can one-line this now it's not a blob:
```ruby
args << "--with-lua" if build.with? "lua"
```

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
-    # Adjust hard coded paths in Makefile
-    inreplace "Makefile.in", "/usr/sbin/", "#{sbin}/"
-    inreplace "Makefile.in", "/usr/bin/", "#{bin}/"
-    inreplace "Makefile.in", "/etc/powerdns/", "#{etc}/powerdns/"
-    inreplace "Makefile.in", "/var/run/", "#{var}/run/"
+    inreplace "pdns_recursor.cc", "SYSCONFDIR", "\"#{etc}/powerdns/\""
+    inreplace "rec_control.cc", "SYSCONFDIR", "\"#{etc}/powerdns/\""
+
+    # Bootstrap?
+    Dir.chdir "pdns/recursordist" if build.head?
```
DomT4: Don't need to call `Dir` directly here.

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
+    system "./bootstrap" if build.head?
```
DomT4: If you've got two `if build.head?` lines here blob 'em:

```ruby
if build.head?
  blah
  blah
end
```

Formula/pdnsrec.rb
Habbie: Ah - I copied this from another formula. Will fix.

Formula/pdnsrec.rb
Habbie: just `chdir` instead of `Dir.chdir`?

Formula/pdnsrec.rb
Habbie: pdnsrec's `make install` does not write recursor.conf to etcdir, only recursor.conf-dist. Below, at line 47, I actually removed the clobbering that this formula did. If I use #{etc} here, users will have to edit their config in /usr/local/Cellar/... instead of /usr/local/etc/powerdns, as far as I can see.

Habbie:
I'll fix the openssl issue and the nits; awaiting response on the `etc/recursor.conf` stuff.

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
+      args << "--with-lua"
     end
 
-    # Adjust hard coded paths in Makefile
-    inreplace "Makefile.in", "/usr/sbin/", "#{sbin}/"
-    inreplace "Makefile.in", "/usr/bin/", "#{bin}/"
-    inreplace "Makefile.in", "/etc/powerdns/", "#{etc}/powerdns/"
-    inreplace "Makefile.in", "/var/run/", "#{var}/run/"
+    inreplace "pdns_recursor.cc", "SYSCONFDIR", "\"#{etc}/powerdns/\""
+    inreplace "rec_control.cc", "SYSCONFDIR", "\"#{etc}/powerdns/\""
```
ilovezfs: These replacements shouldn't be necessary. You should be able to set SYSCONFDIR=... in the make/make install.

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
-    # Do the install manually
-    bin.install "rec_control"
-    sbin.install "pdns_recursor"
-    man1.install "pdns_recursor.1", "rec_control.1"
-
-    # Generate a default configuration file
-    (prefix/"etc/powerdns").mkpath
-    system "#{sbin}/pdns_recursor --config > #{prefix}/etc/powerdns/recursor.conf"
+  test do
+    output = shell_output("#{sbin}/pdns_recursor --version 2>&1", 0)
```
ilovezfs: 0 is already the default

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
+
+    # Bootstrap?
+    Dir.chdir "pdns/recursordist" if build.head?
+    system "./bootstrap" if build.head?
 
     # Compile
-    system "./configure"
+    system "./configure", *args
     system "make"
+    system "make", "install"
```
ilovezfs: Is separating make and make install necessary?

Formula/pdnsrec.rb
ilovezfs: Also, you're setting sysconfdir in configure, so why is this needed at all? Is there a bug?

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,49 @@ class Pdnsrec < Formula
...
+
   def install
-    # Set overrides using environment variables
-    ENV["DESTDIR"] = "#{prefix}"
-    ENV["OPTFLAGS"] = "-O0"
-    ENV.O0
+    args = %W[
+      --prefix=#{prefix}
+      --sysconfdir=#{prefix}/etc/powerdns
+      --with-openssl=#{Formula["openssl"].opt_prefix}
```
ilovezfs: Configure can't find this on its own?

Formula/pdnsrec.rb
Habbie: just `cd` according to @ilovezfs

Habbie:
I'll do another iteration to re-figure out what the story with SYSCONFDIR etc. was so you get a clear picture.

Formula/pdnsrec.rb
ilovezfs: I think we can avoid that. See suggestion above.

Formula/pdnsrec.rb
ilovezfs: I'm not thrilled with two aspects of this
a) the only-if-head directory change
b) the directory change is permanent not in a block

I'd be tempted to do this instead:
```
  def install
    (buildpath/"pdns/recursordist").install buildpath.children if build.stable?

    ...

    cd "pdns/recursordist" do
      inreplace %w[pdns_recursor.cc rec_control.cc],
        "SYSCONFDIR",
        "\"#{etc}/powerdns/\""                        
      system "./bootstrap" if build.head?
      system "./configure", *args
      system "make"
      system "make", "install"
    end
```

Formula/pdnsrec.rb
ilovezfs: Note that for HEAD these would need to be after the directory change

Formula/pdnsrec.rb
ilovezfs: And here's an even better idea: we could just remove the heads spec for this thing.

DomT4: ILZ is correct on our `cd` preference, but I didn't see a way to use `cd` here in our preferred `do` format without making this formula silly complex to visually parse.

ilovezfs: I think we should just kill the head spec. It's a subfolder within the pdns repository, which is what's causing the madness, and we already have a separate pdns formula that has its own head spec.

Habbie: The head spec in the pdns formula installs a different product, to be clear.

DomT4: Yeah, I don't necessarily disagree with you.

It seems like there's a lot we'd need to do to accommodate `head` here. We could look at doing so, but perhaps it's worth waiting to see if we get a sudden influx of demand for `head` on this formula.

Habbie: I'll take out the HEAD stuff for now, it will simplify things. Prior to this 4.0.0 release there was some demand in fact, but now that 4.0.0 is finally out (after 16 months) demand should be limited.

ilovezfs: >sudden influx of demand for head on this formula.

LOL LOL LOL

DomT4: > LOL LOL LOL

I added "on this formula" in the hope people wouldn't burst out laughing at the phrasing 😅.

ilovezfs: @Habbie are you saying `make install` doesn't install recursor at all if you're doing `brew install --HEAD pdns`?

Habbie: Yes, I am. pdns_server, pdns-recursor and dnsdist are distinct products that live in one git tree because they share a lot of code. The non-HEAD versions are distributed as tarballs that are each individually complete (and thus overlap a bit), but when building from git, getting dnsdist or the recursor requires a chdir first.

ilovezfs: I see.

@DomT4 The other option would be to require Autotools for stable, and build from the source archive tarball, which would bring them both into alignment without further fuss.

ilovezfs: Anyway, I can live with the random directory change for only head but it's pretty yucky.

DomT4: > The other option would be to require Autotools for stable

Heavily prefer not to do this as much as possible.

Habbie: I'll just take HEAD out, it's fine. Expect a version with all your comments honoured within a day or two.

Habbie:
So, if I do `--sysconfdir=#{etc}/powerdns`, recursor correctly compiles with this default `# config-dir=/usr/local/etc/powerdns` but it installs `recursor.conf-dist` directly into `/usr/local/etc/powerdns/` which I think is a violation of homebrew policy?

If I do `--sysconfdir=#{prefix}/etc/powerdns`, I get `/usr/local/etc/powerdns//recursor.conf-dist -> ../../Cellar/pdnsrec/4.0.0/etc/powerdns/recursor.conf-dist` which is good, but the recursor gets this configuration default `config-dir=/usr/local/Cellar/pdnsrec/4.0.0/etc/powerdns` which is bad, as it requires the user to put their config there - or we need a `recursor.conf` symlink in `/usr/local/etc/powerdns` which feels like the kind of thing an editor might accidentally destroy. It would also not survive upgrades.

So, I feel that my only choice here is to do the latter (`--sysconfdir=#{prefix}/etc/powerdns`) plus patch those two binaries so they get the right `config-dir`. Thoughts?

Habbie:
For your review, I have pushed a new version. HEAD is gone, openssl and boost are detected properly. The config path issue is not fully fixed (see previous comment for options). Some commented lines are still in there as this is unfinished.

Formula/pdnsrec.rb
ilovezfs: Setting it to #{etc}/powerdns will not clobber existing files. Then you should handle the upgrade/replace/merge/whatever logic in a `post_install` method.

ilovezfs:
>recursor.conf-dist directly into /usr/local/etc/powerdns/ which I think is a violation of homebrew policy?

It's the reverse. The `etc` and `var` pathnames are exceptional in that they deliberately point into the `HOMEBREW_PREFIX` instead of into the formula `prefix`. Then in `post_install` (no sandbox and post bottle pouring) you make sure that the relevant directories are created in `var` (e.g., `(var/"run/foo").mkpath`) if needed, and that the config files are upgraded/merged/replaced/renamed/whatever in `etc` or `etc/"foo"` if the files were installed with the incorrect name or if they weren't installed at all because brew refused to overwrite a preexisting file.

Habbie:
@ilovezfs so `--sysconfdir=#{etc}/powerdns` and no patching would be fine?

Formula/pdnsrec.rb
ilovezfs: ping.

Formula/pdnsrec.rb
```diff
@@ -14,38 +14,34 @@ class Pdnsrec < Formula
...
-    inreplace "Makefile.in", "/usr/sbin/", "#{sbin}/"
-    inreplace "Makefile.in", "/usr/bin/", "#{bin}/"
-    inreplace "Makefile.in", "/etc/powerdns/", "#{etc}/powerdns/"
-    inreplace "Makefile.in", "/var/run/", "#{var}/run/"
+    args << "--with-lua" if build.with? "lua"
 
-    # Compile
-    system "./configure"
-    system "make"
+    # inreplace "pdns_recursor.cc", "SYSCONFDIR", "\"#{etc}/powerdns/\""
```
ilovezfs: delete

Formula/pdnsrec.rb
```diff
@@ -14,38 +14,34 @@ class Pdnsrec < Formula
...
+    # inreplace "rec_control.cc", "SYSCONFDIR", "\"#{etc}/powerdns/\""
```
ilovezfs: delete

Formula/pdnsrec.rb
```diff
@@ -14,38 +14,34 @@ class Pdnsrec < Formula
...
 
-    # Do the install manually
-    bin.install "rec_control"
-    sbin.install "pdns_recursor"
-    man1.install "pdns_recursor.1", "rec_control.1"
+    # Compile
```
ilovezfs: delete

Formula/pdnsrec.rb
```diff
@@ -14,38 +14,34 @@ class Pdnsrec < Formula
...
+    system "./configure", *args
+    # system "make"
```
ilovezfs: delete

Formula/pdnsrec.rb
```diff
@@ -14,38 +14,34 @@ class Pdnsrec < Formula
 
   depends_on :macos => :lion
   depends_on "boost"
+  depends_on "openssl"
   depends_on "lua" => :optional
```
ilovezfs: I wonder if we should just require this and kill the option.

Habbie: Works for me. Doing that.

Formula/pdnsrec.rb
Habbie: sorry, missed this comment.

Habbie:
Updated. Lua is no longer optional. Removed all the comments. Went back to `--sysconfdir=#{etc}/powerdns`. I hope this is done now!

ilovezfs:
@Habbie I don't think your commit made it to the mothership.

Habbie:
It kind of did but not with the changes. Fixed now. Apologies.

MikeMcQuaid:
Fails to build on 10.9: http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4947/version=mavericks/testReport/junit/brew-test-bot/mavericks/install_pdnsrec/

Habbie:
Bumped version requirement to 10.10.

ilovezfs:
@Habbie I'd suggest trying this first:
```
diff --git a/Formula/pdnsrec.rb b/Formula/pdnsrec.rb
index 0b076a4..7b7183b 100644
--- a/Formula/pdnsrec.rb
+++ b/Formula/pdnsrec.rb
@@ -12,14 +12,17 @@ class Pdnsrec < Formula
     sha256 "a9faf9edf0e71de5e9ec9fa70d27811f353a8451b41746208fbdc0d592aa5910" => :mountain_lion
   end
 
-  depends_on :macos => :yosemite
   depends_on "boost"
   depends_on "openssl"
   depends_on "lua"
+  depends_on "gcc" => :linked if MacOS.version <= :mavericks
 
+  needs :cxx11
 
   def install
+    ENV.cxx11
+
     args = %W[
       --prefix=#{prefix}
       --sysconfdir=#{etc}/powerdns
```

Habbie:
Did that, mavericks still unhappy. Looks like it's using clang though - and you tried to make it use gcc it looks like.

Habbie:
Added `fails_with :clang` block to force gcc usage on Mavericks. Bumped to 4.0.1. Please review again!

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,37 @@ class Pdnsrec < Formula
...
+  depends_on "pkg-config" => :build
   depends_on "boost"
-  depends_on "lua" => :optional
+  depends_on "openssl"
+  depends_on "lua"
+  depends_on "gcc" => :linked if MacOS.version <= :mavericks
+
+  needs :cxx11
+
+  fails_with :clang do
```
ilovezfs: is this needed anywhere but Mavericks?

Formula/pdnsrec.rb
```diff
@@ -12,40 +12,37 @@ class Pdnsrec < Formula
     sha256 "a9faf9edf0e71de5e9ec9fa70d27811f353a8451b41746208fbdc0d592aa5910" => :mountain_lion
   end
 
-  depends_on :macos => :lion
+  depends_on "pkg-config" => :build
   depends_on "boost"
-  depends_on "lua" => :optional
+  depends_on "openssl"
+  depends_on "lua"
+  depends_on "gcc" => :linked if MacOS.version <= :mavericks
```
ilovezfs: Let's kill this entire line.

Formula/pdnsrec.rb
ilovezfs: Ah, I see that's the point of the 600.

Formula/pdnsrec.rb
Habbie: If I kill that line, gcc does not get installed and pdnsrec does not build on mavericks

ilovezfs: That's handled by the fails_with.

Habbie: Hmm, pretty sure it wasn't for me. Will give that another shot.

Habbie: If I do
```diff
-  depends_on "gcc" => :linked if MacOS.version <= :mavericks
+  #depends_on "gcc" => :linked if MacOS.version <= :mavericks
```
I get
```
Error: pdnsrec cannot be built with any available compilers.
Install GNU's GCC
  brew install gcc
```

ilovezfs: Run test-bot instead of brew install :)

Habbie: (1) how do I do that? (2) won't that still leave Mavericks users with a confusing error (if we ignore the presence of bottles)?

ilovezfs:
```
brew pull 2895
export HOMEBREW_DEVELOPER=1
export HOMEBREW_SANDBOX=1
brew test-bot pdnsrec
```
No, it just gives them opportunity to decide whether they want to install gcc.

DomT4: > Let's kill this entire line.

It actually links to `gcc` in the bottle, so I'm not sure ditching the dep was wise. It may cause some of the `linkage` issues we'd discussed before:

```
@@HOMEBREW_PREFIX@@/opt/gcc/lib/gcc/6/libstdc++.6.dylib
```

ilovezfs: I know it does, but I do not consider that a sufficient reason for a hard dependency on gcc in the formula for Mavericks, unless it could be scoped to the bottle, which it can't be. Fails_with is supposed to provide flexibility not lock in a specific compiler dependency.

DomT4: > Fails_with is supposed to provide flexibility not lock in a specific compiler dependency.

Sure, ideally. It doesn't though, really, due to the way we handle this for bottles. It won't be any issue once we stop shipping Mavericks bottles in the near future anyway, so meh.

ilovezfs: >Sure, ideally.

It's test-bot that's rescuing `CompilerSelectionError` and then making a choice to install gcc as a remedy. If that behavior causes us to add `depends_on "gcc"` in the formula, it's the tail wagging the dog.

DomT4: Ship it as-is. We'll see how we go.

ilovezfs: Sounds good. :dog:

ilovezfs:
@Habbie Nice work on this. I like how it turned out! Thanks for another contribution to Homebrew and your patience with the tag teaming.

Habbie:
Thanks for the work, to both of you!

PR #2908, opened 2016-07-11
corectl 0.7.7

AntonioMeireles:
Hi again!

please bump 😄 

Thanks in advance

### DIFF
--- a/Formula/corectl.rb
+++ b/Formula/corectl.rb
@@ -1,18 +1,9 @@
 class Corectl < Formula
   desc "CoreOS over OS X made very simple"
   homepage "https://github.com/TheNewNormal/corectl"
+  url "https://github.com/TheNewNormal/corectl/archive/v0.7.7.tar.gz"
+  sha256 "9c71b2d0f5bd0ce22446fd66f85498b0fd748348d9d90eb353e9cb3767e6c7bb"
   head "https://github.com/TheNewNormal/corectl.git", :branch => "golang"
-
-  stable do
-    url "https://github.com/TheNewNormal/corectl/archive/v0.7.0.tar.gz"
-    sha256 "bdbb7d453232995c93862b7e97edb43d312f4976c84c3781e902a140c88ebb45"
-
-    # until 0.7.1 is out
-    patch do
-      url "https://github.com/TheNewNormal/corectl/commit/e519a6f0d1d3c141baef631bdfc65b4e130ff60d.patch"
-      sha256 "966caa8a8f0703fffb5befcc3bec546699c06ca9512d6b1739f28029fe27ed89"
-    end
-  end
 
   bottle do
     cellar :any_skip_relocation


Formula/corectl.rb
```diff
@@ -4,13 +4,12 @@ class Corectl < Formula
   head "https://github.com/TheNewNormal/corectl.git", :branch => "golang"
 
   stable do
-    url "https://github.com/TheNewNormal/corectl/archive/v0.7.0.tar.gz"
-    sha256 "bdbb7d453232995c93862b7e97edb43d312f4976c84c3781e902a140c88ebb45"
-
-    # until 0.7.1 is out
+    url "https://github.com/TheNewNormal/corectl/archive/v0.7.6.tar.gz"
+    sha256 "a2c60b3dff3fa28f0829453c59d27b5cff82e76f7b994f10d35d1e4f6ecabdd4"
+    # until 0.7.6 is out
```
DomT4: The version you've bumped to is `0.7.6` 😄

AntonioMeireles:
@DomT4 addressed.

AntonioMeireles:
meanwhile i got 0.7.7 out...

Formula/corectl.rb
```diff
@@ -4,14 +4,8 @@ class Corectl < Formula
   head "https://github.com/TheNewNormal/corectl.git", :branch => "golang"
 
   stable do
```
ilovezfs: the `stable do ... end` block also needs to be removed, and the url and sha256 moved all the way up in the formula above the head url.

AntonioMeireles:
@ilovezfs addressed, hopefully

ilovezfs:
@AntonioMeireles Yup, thank you. Another question while we're at it again! Is "godep" currently actually used by the formula? I notice you now have a vendor folder in your repository.

AntonioMeireles:
yes it is. see https://github.com/tools/godep#migrating-to-vendor for the gory details.

ilovezfs:
@AntonioMeireles Hmmm. I thought that when you used that procedure to migrate to the vendor folder, that godep could be dropped entirely.

AntonioMeireles:
@ilovezfs no. no, no. `Godeps/Godeps.json` is still where is has been always. this is what `godep` consumes. it's just that the actual payload moved from `Godeps/_src/workspace/` to `vendor/`

Also do note that `godep` is used in the Makefile (L14:15) to help assemble the GOPATH

ilovezfs:
I see. Usually, we would do this (I left out the ldflags):
```
ENV["GOPATH"] = buildpath
(buildpath/("src/github.com/TheNewNormal/corectl").install buildpath.children
cd "src/github.com/TheNewNormal/corectl" do
  go build -o server cmd/server/main.go
  go build -o client cmd/client/main.go
end
```

AntonioMeireles:
ok, thanxs. will look at that once not in vacation .. anyway 'fatness' atm only at build time, not runtime.

ilovezfs:
@AntonioMeireles yup, no problem. Just thought I'd point out that `go build` actually does work out of the box for your software (I just tried) once that vendor folder is there.

ilovezfs:
@AntonioMeireles Thanks for keeping this formula up to date with your software! You're awesome :100:

DomT4:
> I thought that when you used that procedure to migrate to the vendor folder, that godep could be dropped entirely.

Depends how you do it. Godep does support the Vendor experiment. It's not my preferred way of Go vendoring, but it is valid 🙈.

ilovezfs:
@DomT4 My point is that he already did it! https://github.com/TheNewNormal/corectl/tree/golang/vendor

So we might as well reap the benefits ...

PR #2910, opened 2016-07-11
fabio 1.1.5

magiconair:
- [N] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

It passes `brew audit --strict <formula>` as described in https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#create-your-pull-request-from-a-new-branch

With `brew audit --strict --online <formula>` I get an `Error: broken pipe` without further explanation. The binary is fine though.

---


Formula/fabio.rb
```diff
@@ -21,15 +19,14 @@ class Fabio < Formula
...
-    mkdir_p buildpath/"src/github.com/ebay"
-    ln_s buildpath, buildpath/"src/github.com/ebay/fabio"
-
-    ENV["GOPATH"] = "#{buildpath}/_third_party:#{buildpath}"
+    mkdir_p buildpath/"src/github.com/eBay"
+    ln_s buildpath, buildpath/"src/github.com/eBay/fabio"
 
-    Language::Go.stage_deps resources, buildpath/"src"
+    ENV["GOPATH"] = buildpath.to_s
+    ENV["GO15VENDOREXPERIMENT"] = "1"
```
DomT4: Don't need to set this. Has been on by default since Go 1.6, which we ship.

Formula/fabio.rb
```diff
@@ -21,15 +19,14 @@ class Fabio < Formula
...
 
-    system "go", "build", "-o", "fabio"
-    bin.install "fabio"
+    system "go", "install", "-tags", "netgo", "github.com/eBay/fabio"
```
DomT4: This is trying to do the step of opening static libraries in the Go prefix. Can we disable that behaviour here?

```
==> go install -tags netgo github.com/eBay/fabio
go install net: open /usr/local/Cellar/go/1.6.2/libexec/pkg/darwin_amd64/net.a: operation not permitted
```

magiconair:
@DomT4 Disclaimer: I'm the maintainer of `fabio` and just updated the existing formula that someone else put there. This is my first contribution to HomeBrew which I love BTW.

As for `GO15VENDOREXPERIMENT=1`: Yes, I know that this is the default for go 1.6 but since I couldn't find a way to specify the minimum required Go version in the formula and don't have enough Ruby foo to figure it out I've put it in. Also, the build was failing on my machine but that could have been because of my setup. 

As for `-tags netgo`: This is how I ship the binary. I can remove that since it isn't essential. 

However, I can also take another approach altogether. Since I put binary packages of `fabio` on github I could just download the version I've built for OSX and package that instead of building it from source. Would that be acceptable?

magiconair:
@DomT4 I think that fixed all the issues. The `--online` test still fails with `Error: broken pipe` Please let me know whether this is good enough for merging.

DomT4:
> This is my first contribution to HomeBrew which I love BTW.

Thanks! We're glad you find it useful, and welcome to contributing 🍺.

> As for GO15VENDOREXPERIMENT=1: Yes, I know that this is the default for go 1.6 but since I couldn't find a way to specify the minimum required Go version in the formula

`depends_on "go"` == Use the `go` installed by `brew install go`, which barring some hold-up will generally always be the latest stable release. For reasons relating to keeping Homebrew maintainable whilst preserving maintainer sanity we only permit building against the latest formula the `depends_on` statement applies to.

> As for -tags netgo: This is how I ship the binary. I can remove that since it isn't essential.

Thanks! Can you elaborate at all on why some Go projects `open` the static libraries in Go? Is that an optimisation thing? We've seen it a few times but to be honest I don't write enough Go to quite grasp what's happening there.

> Since I put binary packages of fabio on github I could just download the version I've built for OSX and package that instead of building it from source. Would that be acceptable?

We don't generally permit non-from-source builds in the `core` beside things like `.jar`s.

We prefer to build from source, partially because it means we can decide how something is best packaged to suit Homebrew's setup & we've some options if critical security issues appear upstream but a new upstream release won't be available for a week/month/etc. _(Pure example there, nothing to do with your project specifically!)_.

> The --online test still fails with Error: broken pipe

We're aware of this issue but haven't yet successfully narrowed it down all that much. Sorry you're encountering that.

Merged in https://github.com/Homebrew/homebrew-core/commit/f78d7d611b59d31d471072243e3d8da326a4801a. Thank you for your contribution to Homebrew @magiconair; we appreciate it! :smiley_cat:

magiconair:
`-tags netgo` builds a static binary which does not depend on libc since it includes its own DNS resolver. I think this triggers that `net` package is recompiled but I'm not sure. The problem you're having is that your setup does not allow that. I usually have go installed from source in my home directory under `~/go1.x.x`

PR #2920, opened 2016-07-12
lua/luajit: ensure module persistence

DomT4:

---

A loose comparison would be to say LMOD/CMOD are representative of Python's site-packages setup & how we handle that. It isn't expected to point to a transient location.

Note that `lua51` isn't impacted because we've been rewriting the `pkg-config` file there to point to `HOMEBREW_PREFIX` as the `prefix` variable, which means the rest of the paths are automatically correct.

Ref: https://github.com/Homebrew/homebrew-core/pull/2665.

### DIFF
--- a/Formula/lua.rb
+++ b/Formula/lua.rb
@@ -60,10 +60,9 @@ class Lua < Formula
     ENV.universal_binary if build.universal?
 
     # Subtitute formula prefix in `src/Makefile` for install name (dylib ID).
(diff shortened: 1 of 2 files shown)


DomT4:
Self-reminder: `lua53` in Versions probably needs fixing as well.

DomT4:
^ It didn't. `lua53` is also linked to `HOMEBREW_PREFIX` in its `pkg-config` file and therefore is fine.

PR #2926, opened 2016-07-12
gimp moved to cask

AnastasiaSulyagina:

---

In case somebody is opposed, cc @vitorgalvao for discussion


MikeMcQuaid:
👍 but let's give it a few days to make sure the `osxfuse` and/or `tuntap` changes don't 💥

DomT4:
Fine with this one, but this is likely to be noticed. `osxfuse` & `tuntap` together don't match 25% of `gimp` installations via us.

vitorgalvao:
Thumbs up. Nothing more to add.

cc @jawshooah @adidalal

adidalal:
Sure, LGTM

ilovezfs:
What's the goal with killing the gimp source build?

MikeMcQuaid:
Less maintenance, consistency with original Homebrew principle that this isn't the place for GUI apps (and the resulting deletion of e.g. `brew linkapps` that can eventually result).

ilovezfs:
```
 -  depends_on "ghostscript" => :optional
-  depends_on "poppler" => :optional
-  depends_on "libexif" => :optional
```

Does the upstream version provide all three of those?

ilovezfs:
```
-  depends_on "libpng" => :recommended
-  depends_on "libwmf" => :recommended
-  depends_on "libtiff" => :recommended
```
Also, these. My assumption that the answer would be yes for sure I realized is unwarranted.

In addition, there may be items we're unconditionally requiring that are actually options in configure and not included by upstream's version.

In any case, `brew linkapps` is :no_good:, so even if upstream's version is a severely incomplete toy version compared to ours or aborts on launch in a ball of :fire:, good riddance from my own perspective.

Issue #2501, opened 2016-06-29
mailhog is not truly building 0.2.0 version


swrobel:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

_You can erase any parts of this template not applicable to your Issue._

### Bug reports:

The [mailhog formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mailhog.rb) downloads the correct 0.2.0 version of the archive from github, but then downloading outdated versions of the deps, such as [Mailhog](https://github.com/mailhog/MailHog/commit/91a9d8afe119dc96a6c9acfa94c997765c824908), [Mailhog-UI](https://github.com/mailhog/MailHog-UI/commit/ebdfa0f91fac40169ca376953d7c13f9d83cd01a) and [Mailhog-Server](https://github.com/mailhog/MailHog-Server/commit/1dfb973b724a309c18474fe032b1458c0d8654fc).

I would submit a PR to update these, but:

1. I'm not very familiar with the Go build process
2. I'm not sure what @maijs's rationale was for locking to these specific versions, especially when it comes to external deps (it seems obvious to me that the Mailhog ones should be locked to their respective version tags)

dunn:
Usually we use the latest revision of a dependency since Go often just pulls `HEAD`; but it looks like the dependency revisions weren't updated with the 2.0 release.  So a PR to bump the revisions to either their latest or to a corresponding 0.2.0 tag would be great!

DomT4:
I looked into this the other night & fell down a deep, deep rabbithole. It turns out we've added some dependencies as `go_resource`s, but not all, but if you add them all it doesn't compile. It'd be _really_ nice if upstream moved over to the vendor system here really.

maijs:
@swrobel The specific revisions were locked because with the most recent versions of external dependencies MailHog would not build, at least at the time the formula was created.

swrobel:
@maijs I was afraid of that. I've filed a [related issue](https://github.com/mailhog/MailHog/issues/100) with Mailhog about the complicated build process

In the interim, how does everyone feel about using the [prebuilt binaries](https://github.com/mailhog/MailHog/releases) that mailhog offers?

maijs:
[...]

PR #2931, opened 2016-07-12
mailhog: references to specific dependency revisions removed.

maijs:

---

This PR removes references to specific dependency revisions and makes sure that, if built from source, MailHog is using the latest versions of its dependencies. This issue was originally mentioned by @swrobel in #2501.

### DIFF
--- a/Formula/mailhog.rb
+++ b/Formula/mailhog.rb
@@ -18,113 +18,92 @@ class Mailhog < Formula
   depends_on "go" => :build
 
   go_resource "github.com/gorilla/context" do
-    url "https://github.com/gorilla/context.git",
-      :revision => "1c83b3eabd45b6d76072b66b746c20815fb2872d"
+    url "https://github.com/gorilla/context.git"
   end
 
   go_resource "github.com/gorilla/mux" do
-    url "https://github.com/gorilla/mux.git",
-      :revision => "ad4d7a5882b961e07e2626045eb995c022ac6664"
+    url "https://github.com/gorilla/mux.git"
   end
 
   go_resource "github.com/gorilla/pat" do
-    url "https://github.com/gorilla/pat.git",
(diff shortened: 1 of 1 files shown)


dunn:
My comment at https://github.com/Homebrew/homebrew-core/issues/2501#issuecomment-230158489 might have been poorly worded.  Since Go usually pulls the latest code, when we add a `go_resource` to a formula we generally use the latest commit, but we still want to pin the dependency to that commit so things don't shift around underneath us and cause breakages.

DomT4:
Yeah, we're not going to accept this as-is.

99% of the reason `go_resource`s became a thing is because Go's old dependency system of simply fetching the latest commit from the master branch of a dependency was a disaster for consistency. No desire to go back there.

swrobel:
@DomT4 would this be acceptable for building the HEAD version? Maybe we could have the versioned released in homebrew-binary and this formula would just build from HEAD...

dunn:
The HEAD build can use unpinned resources in its `head do` block (see e.g., [`telegraf.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/telegraf.rb)), but we don't allow HEAD-only formula.

swrobel:
but, but...

Monty Python

(that was my attempt at a head-only joke if that wasn't obvious)

maijs:
I will pin down the commits that correspond to the 0.2.0 release. Stay tuned.

MikeMcQuaid:

swrobel:
Honestly I don't see how that's a reasonable solution: the current formula is broken

DomT4:
Did something a bit naughty: https://github.com/Homebrew/homebrew-core/pull/4488. Not completely ideal, but increasingly it's either this or the boneyard, so let's try one form of drastic before trying the other form of drastic 😅.

Issue #2932, opened 2016-07-12
imagemagick 6.9.5-1 fails to compile on OS X 10.11.5


jonnybarnes:
https://gist.github.com/jonnybarnes/79387baeac3da1fed59e24af79d77b49

dunn:
It's failing to find libltdl, which is part of libtool: https://gist.github.com/jonnybarnes/79387baeac3da1fed59e24af79d77b49#file-01-configure-L1-L585

OSX ships with an older libtool, which apparently doesn't include it:
```
🔥  locate ltdl
/Applications/Oracle Calendar.app/Contents/Frameworks/libltdl.7.dylib
/Applications/Oracle Calendar.app/Contents/Frameworks/libltdl.dylib
/Applications/Paint.app/Contents/Frameworks/libltdl.7.dylib
/Applications/Paint.app/Contents/Frameworks/libltdl.dylib
/usr/local/Cellar/libtool/2.4.6/include/libltdl
/usr/local/Cellar/libtool/2.4.6/include/libltdl/lt_dlloader.h
/usr/local/Cellar/libtool/2.4.6/include/libltdl/lt_error.h
/usr/local/Cellar/libtool/2.4.6/include/libltdl/lt_system.h
/usr/local/Cellar/libtool/2.4.6/include/ltdl.h
/usr/local/Cellar/libtool/2.4.6/lib/libltdl.7.dylib
/usr/local/Cellar/libtool/2.4.6/lib/libltdl.a
/usr/local/Cellar/libtool/2.4.6/lib/libltdl.dylib
/usr/local/Cellar/libtool/2.4.6/share/aclocal/ltdl.m4
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl/lt__alloc.h
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl/lt__argz_.h
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl/lt__dirent.h
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl/lt__glibc.h
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl/lt__private.h
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl/lt__strl.h
/usr/local/Cellar/libtool/2.4.6/share/libtool/libltdl/lt_dlloader.h
[...]

PR #2935, opened 2016-07-12
imagemagick 6.9.5-2

zmwangx:


RandomDSdevel:
@zmwangx:  Hmm, looks like the bottle build errored out on us.  From what I see in [the top-level CI console output](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4766/console), the [Mavericks build](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4766/version=mavericks/console) succeeded but, strangely enough, both the [Yosemite build](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4766/version=yosemite/console) and the [El Capitan](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/4766/version=el_capitan/console) one crashed.  Scrutinizing the 10.11.5 build console output I linked previously further reveals that the problem reared its ugly head when ImageMagick's configure phase was trying to look for a little something called libltdl, which, upon further research, appears to be part of libtool.  I wonder what in the blazes could be going on here…

zmwangx:
See #2932. I haven't looked into the issue, but my local build is fine.

RandomDSdevel:
Ah, whoops:  since I wasn't expecting that bug to persist for builds of later versions of ImageMagick, I overlooked it.  Thanks for pointing me back to it, @zmwangx!

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/0f16c1c3a18ee5eeceda06a33765f1ddea19bb4c. Thanks again @zmwangx, and appreciate the patience whilst I ran around the Homebrew repos screaming _"Why is everything broken, what did we do"_ until I spotted & fixed it, heh.

PR #2940, opened 2016-07-12
ArangoDB 3.0.3: do not use versioned path in config

fceller:

---

### DIFF
--- a/Formula/arangodb.rb
+++ b/Formula/arangodb.rb
@@ -1,16 +1,16 @@
 class Arangodb < Formula
   desc "The Multi-Model NoSQL Database."
   homepage "https://www.arangodb.com/"
-  url "https://www.arangodb.com/repositories/Source/ArangoDB-3.0.2.tar.gz"
-  sha256 "04c00d58d7e63137ccb7d0a73112aa01e34e08c98b3c82c62ef7987f0d214ac2"
+  url "https://www.arangodb.com/repositories/Source/ArangoDB-3.0.3.tar.gz"
+  sha256 "09fb1161afcade627b1e022c2c15b6f26c17eac9d07b9636829212dca01272bb"
   head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
 
   bottle do
     sha256 "358382598b83897515c32c3182b24fea99842d721d626ccef9cf169a8dbb1639" => :el_capitan
     sha256 "db166984be2d166206ee348b3d6741093b8bec5b45ac9f456cbb300d4d6c0e02" => :yosemite
-    sha256 "57139fadef3855ef14bed71c2f9c5f37a2f991f35163cf6b810020fa0e077d13" => :mavericks
   end
 
+  depends_on :macos => :yosemite
   depends_on "cmake" => :build
   depends_on "go" => :build
   depends_on "openssl"
@@ -62,6 +62,7 @@ class Arangodb < Formula
         -DHOMEBREW=ON
         -DUSE_OPTIMIZE_FOR_ARCHITECTURE=OFF
         -DASM_OPTIMIZATIONS=OFF
+        -DCMAKE_INSTALL_DATADIR=#{share}
         -DETCDIR=#{etc}
         -DVARDIR=#{var}
       ]
@@ -71,11 +72,22 @@ class Arangodb < Formula
       end
 
       system "cmake", "..", *args
-      system "make", "V=1", "Verbose=1", "VERBOSE=1", "install"
+      system "make", "install"
+
+      %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
+        inreplace etc/"arangodb3/#{f}.conf", pkgshare, opt_pkgshare
+      end
     end
   end
 
   def post_install
+    oldpath_prefix = "#{HOMEBREW_PREFIX}/Cellar/arangodb/3.0."
+    oldpath_regexp = /#{Regexp.escape(oldpath_prefix)}[12]/
+
+    %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
+      inreplace etc/"arangodb3/#{f}.conf", oldpath_regexp, opt_prefix, false
+    end
+
     (var/"lib/arangodb3").mkpath
     (var/"log/arangodb3").mkpath
 


DomT4:
Somehow this timed out at 90 minutes on Mavericks, again. Took ~30 minutes on Yosemite/El Cap 😕.

fceller:
@DomT4 yes, something is definitely weird with Mavericks. Unfortunate, I do not have access to such a machine.

fceller:
@DomT4 @ilovezfs there is something weird with the Mavericks & GCC 6. If downloaded and installed the "Mavericks" bottle on my ElCaptain. The executable contains the following references:

```
> otool -L /usr/local/sbin/arangod 
/usr/local/sbin/arangod:
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/gcc/lib/gcc/6/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.22.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
	/usr/local/lib/gcc/6/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)

```

The problem is `/usr/local/opt/gcc/lib/gcc/6/libstdc++.6.dylib`. I do have gcc-6 installed

```
> brew search gcc6
homebrew/versions/gcc6 ✔
```

But it seems that there is no such directory

```
> ls -l /usr/local/opt/gcc/lib/gcc/
total 0
drwxr-xr-x  54 fceller  admin  1836 10 Feb 14:38 5
```

Which package should install the file `libstdc++.6.dylib`? Maybe this is missing as `depends_on`  in the formula.

fceller:
The situation seems to be:

If GCC 6 is installed, the bottle will work (at least under ElCaptain).
If GCC 6 is not installed, the bottle will fail, because the (correct) libstdc++.6.dylib is missing.

fceller:
@BrewTestBot please test this

ilovezfs:
@fceller you should be able to fix that with
```
depends_on "gcc" if MacOS.version <= :mavericks
```

Formula/arangodb.rb
```diff
@@ -71,7 +71,13 @@ def install
       end
 
       system "cmake", "..", *args
-      system "make", "V=1", "Verbose=1", "VERBOSE=1", "install"
+      system "make", "install"
+
+      configs = ["arangod", "arango-dfdb", "arangosh", "foxx-manager"]
```
ilovezfs: Might be nicer to do this:
```
    %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
      inreplace etc/"arangodb3/#{f}.conf", pkgshare, opt_pkgshare
    end
```

fceller: Done

DomT4:
> @BobTestBot please test this

Just a note this isn't supposed to work for anyone who isn't a maintainer, so if you're being ignored by CI that'll be why.

ilovezfs:
@MikeMcQuaid It seems the voodoo spell may have lost its efficacy.

(RE: https://github.com/Homebrew/homebrew-core/pull/2576#issuecomment-231667194)

fceller:
@ilovezfs a colleague of mine still had a mavericks ISO image. Hope that it is possible to create this behavior locally.

ilovezfs:
@fceller that would be very nice of her/him if possible!

ilovezfs:
@fceller here's the log:
```
ln -s ../Cellar/arangodb/3.0.2/share/arangodb3 arangodb3
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160713-74299-1s22gt2-0.sb nice /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/postinstall.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/arangodb.rb
/usr/local/Library/Homebrew/postinstall.rb (Formulary::FromPathLoader): loading /usr/local/Library/Taps/homebrew/homebrew-core/Formula/arangodb.rb
==> /usr/local/Cellar/arangodb/3.0.2/libexec/arangodb2/bin/upgrade.sh /usr/local/Cellar/arangodb/3.0.2/libexec/arangodb2 /usr/local/var/lib/arangodb /usr/local/opt/arangodb /usr/local/var/lib/arangodb3
Upgrade in progress...
Waiting for ArangoDB 2 to boot...
2016-07-13T21:49:58Z [74302] INFO ArangoDB 2.8.10 [darwin] 64bit, using ICU 54.1, V8 4.3.61, OpenSSL 1.0.2h  3 May 2016
2016-07-13T21:49:58Z [74302] INFO using default language 'en'
2016-07-13T21:49:58Z [74302] INFO WAL directory '/usr/local/var/lib/arangodb/journals' does not exist. creating it...
2016-07-13T21:49:58Z [74302] INFO created base application directory '/usr/local/var/lib/arangodb-apps/_db'
2016-07-13T21:49:58Z [74302] INFO loaded database '_system' from '/usr/local/var/lib/arangodb/databases/database-91159'
2016-07-13T21:49:58Z [74302] INFO collection '_users' does not exist, no authentication available
2016-07-13T21:49:58Z [74302] INFO using endpoint 'tcp://127.0.0.1:8529' for non-encrypted requests
2016-07-13T21:49:58Z [74302] INFO using default API compatibility: 20800
2016-07-13T21:49:58Z [74302] INFO JavaScript using startup '/usr/local/Cellar/arangodb/3.0.2/libexec/arangodb2/share/arangodb/js', application '/usr/local/var/lib/arangodb-apps'
2016-07-13T21:49:59Z [74302] INFO In database '_system': No version information file found in database directory.
2016-07-13T21:49:59Z [74302] INFO In database '_system': Found 22 defined task(s), 17 task(s) to run
2016-07-13T21:49:59Z [74302] INFO In database '_system': state prod/standalone/init, tasks setupUsers, createUsersIndex, addDefaultUser, updateUserModel, setupGraphs, createModules, createRouting, createKickstarterConfiguration, insertRedirectionsAll, setupAqlFunctions, createStatistics, createConfiguration, systemAppEndpoints, setupQueues, setupJobs, moveFoxxApps, dropAal
2016-07-13T21:49:59Z [74302] INFO In database '_system': init successfully finished
2016-07-13T21:50:00Z [74302] INFO file-descriptors (nofiles) hard limit is -1, soft limit is 10240
2016-07-13T21:50:00Z [74302] INFO Authentication is turned off
2016-07-13T21:50:00Z [74302] INFO ArangoDB (version 2.8.10 [darwin]) is ready for business. Have fun!
ArangoDB 2 has started, beginning dump process...
Server version: 2.8.10
Connected to ArangoDB 'tcp://127.0.0.1:8529', database: '_system', username: 'root'
Writing dump to output directory '/usr/local/var/lib/arangodb/DUMP'
Last tick provided by server is: 20210711
# Dumping document collection '_aqlfunctions'...
# Dumping document collection '_graphs'...
# Dumping document collection '_modules'...
# Dumping document collection '_routing'...
# Dumping document collection '_system_users_users'...
# Dumping document collection '_users'...
Processed 6 collection(s), wrote 1142 byte(s) into datafiles, sent 6 batch(es)
Dump has finished, shutting down ArangoDB 2...
2016-07-13T21:50:09Z [74302] INFO control-c received, beginning shut down sequence
2016-07-13T21:50:13Z [74302] INFO ArangoDB has been shut down
Waiting for ArangoDB 3 to boot...
2016-07-13T21:50:13Z [74311] INFO using SSL options: SSL_OP_CIPHER_SERVER_PREFERENCE, SSL_OP_TLS_ROLLBACK_BUG
2016-07-13T21:50:13Z [74311] INFO file-descriptors (nofiles) hard limit is unlimited, soft limit is 10240
2016-07-13T21:50:13Z [74311] INFO WAL directory '/usr/local/var/lib/arangodb3/journals' does not exist. creating it...
2016-07-13T21:50:13Z [74311] INFO ArangoDB 3.0.2 [darwin] 64bit, using VPack 0.1.30, ICU 54.1, V8 5.0.71.39, OpenSSL 1.0.2h  3 May 2016
2016-07-13T21:50:13Z [74311] INFO created base application directory '/usr/local/var/lib/arangodb3-apps/_db'
2016-07-13T21:50:13Z [74311] INFO loaded database '_system' from '/usr/local/var/lib/arangodb3/databases/database-1'
2016-07-13T21:50:13Z [74311] INFO the server has 4 (hyper) cores, using 1 scheduler thread(s), 4 dispatcher thread(s)
2016-07-13T21:50:13Z [74311] INFO JavaScript using startup '/usr/local/opt/arangodb/share/arangodb3/js', application '/usr/local/var/lib/arangodb3-apps'
2016-07-13T21:50:14Z [74311] INFO In database '_system': No version information file found in database directory.
2016-07-13T21:50:14Z [74311] INFO In database '_system': Found 14 defined task(s), 12 task(s) to run
2016-07-13T21:50:14Z [74311] INFO In database '_system': state standalone/init, tasks setupGraphs, setupUsers, createUsersIndex, addDefaultUserSystem, createModules, createRouting, insertRedirectionsAll, setupAqlFunctions, createStatistics, createFrontend, setupQueues, setupJobs
2016-07-13T21:50:14Z [74311] INFO In database '_system': init successfully finished
2016-07-13T21:50:14Z [74311] INFO using endpoint 'http+tcp://127.0.0.1:8529' for non-encrypted requests
2016-07-13T21:50:14Z [74311] INFO Authentication is turned on
2016-07-13T21:50:14Z [74311] INFO Authentication system only
2016-07-13T21:50:14Z [74311] INFO Authentication for unix sockets is turned on
arangod(74311,0x116a32000) malloc: *** error for object 0xe00007fc60211b56: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
/usr/local/Cellar/arangodb/3.0.2/libexec/arangodb2/bin/upgrade.sh: line 67: 74311 Illegal instruction: 4  ${ARANGOD3} --log.foreground-tty false --log.file "-"
Waiting for ArangoDB 3 to boot...
Waiting for ArangoDB 3 to boot...
Waiting for ArangoDB 3 to boot...
etc.
```

fceller:
@ilovezfs thanks a lot, I'll try to reproduce this on my colleagues computer.

ilovezfs:
@fceller Do you mind trying `ENV.O0` at the top of install as an experiment?

fceller:
@ilovezfs that is a good idea

fceller:
Also:

ilovezfs:
That warning comes up pretty much every time gcc is used around here. Sometimes an issue, sometimes not.

fceller:
@ilovezfs I could reproduce the crash on my colleague's machine. It is inside "libc++" :-(

```
Warning: arangodb dependency gmp was built with a different C++ standard
library (libc++ from clang). This may cause problems at runtime.

```

I'll check if there is any way to build it with libc++ from clang

ilovezfs:
>I'll check if there is any way to build it with libc++ from clang

Yeah, if you could get it building on Mavericks with clang that would be super!

MikeMcQuaid:
I think at this point it's worth depending on Yosemite and leave it to users who are interested to make this work on Mavericks (there may not be any!) to fix this up if they need it.

ilovezfs:
@MikeMcQuaid Also, we're not leaving them out in the cold if we do that, since there's a brand new Mavericks 2.8.10 bottle in homebrew/versions.

fceller:
@MikeMcQuaid @ilovezfs how do I specify "depending on at least yosemite"?

ilovezfs:

fceller:
@ilovezfs thanks & done

ilovezfs:
@fceller you're welcome. Is the 3.0.1/3.0.2 config issue thing sorted out by this commit too?

fceller:
@ilovezfs I hope so. That was the initial reason for the commit. I've added:

```
      %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
         inreplace etc/"arangodb3/#{f}.conf", pkgshare, opt_pkgshare
      end
```

which should switch to the "/usr/local/opt" path.

ilovezfs:
OK, and that works because there's actually no one that has 3.0.1?

ilovezfs:
I think there's a problem though in that those replacements are going to fail when you bump the version to 3.0.3, since the 3.0.3 in pkgshare is not going to match the opt_pkgshare in the preexisting 3.0.2 config files, which won't be clobbered, and inreplace dies if the replacement isn't successful. Unless I've misunderstood how this works?

fceller:
I would assume so. We did not publish the 3.0.1 formula, but went straight to 3.0.2.

The replace is not done in the existing files in /usr/local/etc. I'm replacing the path during the "make install" in the "def install".

ilovezfs:
`etc` refers to `/usr/local/etc` even during `def install`, not `#{prefix}/etc` (i.e., `not /usr/local/Cellar/arangodb/3.0.2/etc`).

fceller:
I see. will change to "#{prefix}/etc".

ilovezfs:
I don't think that will work since you've correctly set `-DETCDIR=#{etc}`, so those files won't exist in #{prefix}/etc, will they? Unless `-DETCDIR=#{etc}` were actually being ignored, which it's not, otherwise the inreplace would already be failing during CI.

This is why we usually handle this in post_install  :)

fceller:
@ilovezfs you lost be there, will try it out.

fceller:
@ilovezfs I tried with the original "etc" version and AFAICS it does the correct replace:

ilovezfs:
@fceller right, but pouring the bottle won't cause that to happen in a user's preexisting /usr/local/etc/arangodb3/arangod.conf unless it's done in `post_install`.

fceller:
@ilovezfs understood. You talk about existing 3.0.2 installation. Yes, you are right, they need to be patched once in post_install. The correct change will only create the correct paths within the bottle.

Will add.

ilovezfs:
@fceller you can test it locally if you generate a bottle with `brew test-bot arangodb` and then install that bottle over an existing 3.0.2 installation using brew upgrade.

ilovezfs:
@fceller also you probably will want to add `revision 1` to the formula so that all of your users will end up with the same configuration file now. Otherwise, that won't happen until 3.0.3 since existing installs don't get an upgrade unless you add a revision or the version number actually increases.

fceller:
@ilovezfs done

ilovezfs:
@fceller there may be some forthcoming style quibbles, but that looks much better!

Did you have a chance to test it?

fceller:
@ilovezfs it did work localky. Needed to get the regex right. But finally it worked.

Formula/arangodb.rb
```diff
@@ -6,6 +6,7 @@ class Arangodb < Formula
   head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
 
   bottle do
+    revision 1
```
apjanke: The `revision 1` line should go outside the `bottle do` block. It needs to apply to the entire formula, not just the bottle build.

fceller: ok, moved it outside

Formula/arangodb.rb
```diff
@@ -71,11 +74,28 @@ def install
...
+      system "make", "install"
+
+      %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
+        inreplace etc/"arangodb3/#{f}.conf", pkgshare, opt_pkgshare
+      end
     end
   end
 
   def post_install
+    oldpath = Regexp.escape("#{HOMEBREW_PREFIX}/Cellar/arangodb/3.0.") + "[12]"
```
ilovezfs: @UniqMartin I'm guessing this is not our standard way of writing this in brew?

MikeMcQuaid: `Regexp.escape("#{HOMEBREW_PREFIX}/Cellar/arangodb/3.0.[12]")`

Formula/arangodb.rb
```diff
@@ -71,11 +74,28 @@ def install
...
+
+    %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
+      file = etc/"arangodb3/#{f}.conf"
+      oldconfig = File.read(file)
+      newconfig = oldconfig.gsub(Regexp.new(oldpath), opt_prefix)
+
+      if oldconfig != newconfig
+        mv file, "#{file}.old-3.0.2"
```
MikeMcQuaid: You can use `inreplace` to modify the file directly; don't think you need to create a backup if that's the only change.

Formula/arangodb.rb
UniqMartin: Mike's suggestion unfortunately won't quite work, as it's escaping too much. My suggestion would be to split this into two steps and then use `oldpath_regexp` further below (instead of repeatedly re-creating the regular expression on use):

```ruby
oldpath_prefix = "#{HOMEBREW_PREFIX}/Cellar/arangodb/3.0."
oldpath_regexp = /#{Regexp.escape(oldpath_prefix)}[12]/
```

Formula/arangodb.rb
```diff
@@ -71,11 +74,28 @@ def install
...
+      end
     end
   end
 
   def post_install
+    oldpath = Regexp.escape("#{HOMEBREW_PREFIX}/Cellar/arangodb/3.0.") + "[12]"
+
+    %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
+      file = etc/"arangodb3/#{f}.conf"
+      oldconfig = File.read(file)
```
UniqMartin: `file` is going to be a `Pathname` instance here. We thus prefer to use `file.read`.

Formula/arangodb.rb
```diff
@@ -5,6 +5,8 @@ class Arangodb < Formula
   sha256 "04c00d58d7e63137ccb7d0a73112aa01e34e08c98b3c82c62ef7987f0d214ac2"
   head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
 
+  revision 1
```
UniqMartin: This should go between the `sha256` and `head` line above, with an empty line between the `revision` and the `head` line.

DomT4:
> and leave it to users who are interested to make this work on Mavericks (there may not be any!)

According to our analytics there isn't. Exclusively 10.10 & 10.11 already, apparently.

MikeMcQuaid:
Thanks @DomT4. 👍 x 💯 to shipping without Mavericks.

Formula/arangodb.rb
```diff
@@ -71,11 +72,26 @@ def install
...
   end
 
   def post_install
+    oldpath_prefix = "#{HOMEBREW_PREFIX}/Cellar/arangodb/3.0."
+    oldpath_regexp = /#{Regexp.escape(oldpath_prefix)}[12]/
+
+    %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
+      file = etc/"arangodb3/#{f}.conf"
+
+      if oldpath_regexp.match(File.read(file))
```
ilovezfs: you can do
```
inreplace etc/"arangodb3/#{f}.conf" do |s|
  s.gsub! oldpath_regexp, opt_prefix, false
end
```

The `false` sets audit = false so that you don't have to redundantly pre-verify that the inreplace will be successful.

fceller:
@ilovezfs done

fceller:
@ilovezfs the brew bot does not like the change:

```
arangodb:
  * 'inreplace ... do' was used for a single substitution (use the non-block form instead).

```

ilovezfs:
Hmm, @UniqMartin it seems https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils/inreplace.rb#L18 should support the third argument if audit is going to be actively blocking a useful feature of the block form just because it did only one replacement.

UniqMartin:
@ilovezfs I agree, but I guess the reason for the omission is that this useful feature is also a very rarely used feature and the advice to refrain from using the block form for a single substitution is relatively new. I think the lack of feature parity between these two forms is surprising in this case, but it should be easy to fix that. Do you want to take a stab at it?

ilovezfs:
@UniqMartin Sure, just wanted to verify this wasn't a deliberate omission before spending time on it. Thanks!

ilovezfs:
@fceller @UniqMartin voilà https://github.com/Homebrew/brew/pull/552.

fceller:
@ilovezfs cool. Can you retest the formula?

ilovezfs:
@fceller You need to update it to be
```
    %w[arangod arango-dfdb arangosh foxx-manager].each do |f|
      inreplace etc/"arangodb3/#{f}.conf", oldpath_regexp, opt_prefix, false
    end
```

fceller:
@ilovezfs done

ilovezfs:
@DomT4 :ship:?

UniqMartin:
I'm not the one being asked, but LGTM!

ilovezfs:
@UniqMartin I'm about 93.8% sure @DomT4 will not mind your substitution.

ilovezfs:
@fceller you have the patience of a saint. Thanks for your hard work on this formula and the one in homebrew/versions. We really appreciate it. :100:

DomT4:
@ilovezfs I spend a few hours offline & you've already replaced me with Martin. I'm shocked & hurt ILZ 😉. In all seriousness, this looks great. Nice work everyone.

fceller:
No worries. 3.0.4 will be much easier :-)

Issue #2782, opened 2016-07-07
cfitsio examples bad SHA256


parejkoj:
When I try to install the cfitsio examples via the "--with-examples" argument--`brew install cfitsio --with-examples`--I get a bad SHA256 error, as shown below.

I suspect that the cexamples.zip files changed since the brew was last touched. Should we just update the SHA256 sum?

A gist-log is here, though I'm doubt it is necessary: https://gist.github.com/anonymous/7cead437915e03961e43838fc8dbb3cc

```
$ brew install cfitsio --with-examples
==> Downloading https://heasarc.gsfc.nasa.gov/FTP/software/fitsio/c/cfitsio3370.
Already downloaded: /Users/parejkoj/Library/Caches/Homebrew/cfitsio-3.370.tar.gz
==> ./configure --prefix=/usr/local/Cellar/cfitsio/3.370 --enable-reentrant
==> make shared
==> make install
==> make fpack funpack
==> Downloading https://heasarc.gsfc.nasa.gov/docs/software/fitsio/cexamples/cex
######################################################################## 100.0%
Error: SHA256 mismatch
Expected: 85b2deecbd40dc2d4311124758784b1ff11db1dd93ac8e7a29f3d6cda5f8aa3d
Actual: ed17b6d0f2a3d9858d6b19a073b2479823b37e501b7fd0ef9fa08988ad7ab8fc
Archive: /Users/parejkoj/Library/Caches/Homebrew/cfitsio--examples-2014.01.23.zip
To retry an incomplete download, remove the file above.
```

DomT4:
Looks like it's been updated, aye:
```
Last-Modified: Wed, 06 Apr 2016 15:24:26 GMT
```
There's actually a `cfitsio` update to `3.390` available as well. We'd review a PR to update both, if you'd like to submit one?

DomT4:
Fixed by https://github.com/Homebrew/homebrew-core/commit/dfc3f215e57b2a988ccd3d997e41d7b2fe7412bf.

parejkoj:
Thank you @DomT4 !

PR #2944, opened 2016-07-13
cfitsio 3.390

zmwangx:

---

I saw #2782, so I'm doing a drive-by update.
- Updated the examples resource;
- Added a test block (now installing `testprog*` to `share/cfitsio/testprog/` so that we can test it later).

Another thought: maybe this formula could be migrated to homebrew/science, because `cfitsio` deals with FITS images used only in astronomy (AFAIK)? Plus `ccfits`, its C++ wrapper, is in homebrew/science.

### DIFF
--- a/Formula/cfitsio.rb
+++ b/Formula/cfitsio.rb
@@ -1,10 +1,10 @@
 class Cfitsio < Formula
   desc "C access to FITS data files with optional Fortran wrappers"
   homepage "https://heasarc.gsfc.nasa.gov/docs/software/fitsio/fitsio.html"
-  url "https://heasarc.gsfc.nasa.gov/FTP/software/fitsio/c/cfitsio3370.tar.gz"
-  mirror "ftp://heasarc.gsfc.nasa.gov/software/fitsio/c/cfitsio3370.tar.gz"
-  version "3.370"
-  sha256 "092897c6dae4dfe42d91d35a738e45e8236aa3d8f9b3ffc7f0e6545b8319c63a"
+  url "https://heasarc.gsfc.nasa.gov/FTP/software/fitsio/c/cfitsio3390.tar.gz"
+  mirror "ftp://heasarc.gsfc.nasa.gov/software/fitsio/c/cfitsio3390.tar.gz"
+  version "3.390"
+  sha256 "62d3d8f38890275cc7a78f5e9a4b85d7053e75ae43e988f1e2390e539ba7f409"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Formula/cfitsio.rb
```diff
@@ -39,4 +40,18 @@ def install
       end
     end
   end
+
+  test do
+    cp Dir["#{pkgshare}/testprog/testprog*"], testpath
+    flags = (ENV.cflags || "").split + (ENV.cppflags || "").split + (ENV.ldflags || "").split
```
apjanke: Is this flag propagation part necessary? Won't the compiler pick these up from the existing environment variables? I'm wondering because I think the flags variables, especially the LDFLAGS, might have embedded spaces & quotes which `split` may not handle correctly.

apjanke:
:+1:, with one question about flag handling inside the test code.

Formula/cfitsio.rb
zmwangx: @apjanke I'm not sure, I've been treating this line as a boilerplate when compiling C/C++ programs inside a test block 😂

MikeMcQuaid: Could just do `.to_s.split` if you really needed this.

zmwangx: @MikeMcQuaid There are 66 formulae with exactly this line (try `grep -rI 'ENV.cflags || ""' "$(brew --repository homebrew/core)"`); that's why I've been using it as a boilerplate. Any reason to change it to something else?

MikeMcQuaid: It's messy. Feel free to bulk change the others, though, if we find something cleaner.

apjanke: Mike - do you know why this propagation is necessary in the first place? It seems like this is sort of bypassing environment sanitization. Should we even be picking up user-defined compilation flags when doing a test of a build tool package? Or is this doing something else related to how `brew test` sets up the environment?

apjanke: Okay, I think I see what it's doing. It's not picking up user-defined variables. The tests run under StdEnv, which first sanitizes these vars and [may then set up new values for them](https://github.com/Homebrew/brew/blob/4b2c4ef25835a87b5bac4fb26eae91dbfed93863/Library/Homebrew/extend/ENV/std.rb#L31-L47).

So, this line pulls the StdEnv-supplied values out of the env vars and puts them in the `cc` command line. That sounds like a way of adding support for compilers which don't respect those environment variables. Are there any? Or maybe it's doing it so all those flags show up in the test execution output?

MikeMcQuaid: I don't think it is necessary; I think it was some sort of early "handle superenv if we ever port tests to that" or something. I don't think there's any need for it, really.

zmwangx: So I removed this line. What about a mass removal from the other 60+ formulae to prevent its further spread?

MikeMcQuaid: @zmwangx want to just create a mass PR?

zmwangx: @MikeMcQuaid Sure. Should I put everything in one commit or use one commit per formula?

MikeMcQuaid: The latter, thanks!

zmwangx: #3104.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/dfc3f215e57b2a988ccd3d997e41d7b2fe7412bf. Thank you once again @zmwangx. You've made some very solid & valuable contributions lately, we appreciate that.

PR #2954, opened 2016-07-13
corectl 0.7.9

AntonioMeireles:
please bump. 

thanks!

### DIFF
--- a/Formula/corectl.rb
+++ b/Formula/corectl.rb
@@ -1,8 +1,8 @@
 class Corectl < Formula
   desc "CoreOS over OS X made very simple"
   homepage "https://github.com/TheNewNormal/corectl"
-  url "https://github.com/TheNewNormal/corectl/archive/v0.7.8.tar.gz"
-  sha256 "4b1db8aac23394e508c721c56fe077a357627ad35c332eb2b7129758a5a7c8ed"
+  url "https://github.com/TheNewNormal/corectl/archive/v0.7.9.tar.gz"
+  sha256 "51d3949694347dc5146471225404f7b9f892acd8760575da3461192f28af8434"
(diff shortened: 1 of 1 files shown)


AntonioMeireles:
@ilovezfs if awke ;-) && thanks in advance!

ilovezfs:
>&& thanks in advance!

Does that mean that I *get no thanks* if I'm not awake?

Formula/corectl.rb
```diff
@@ -42,6 +42,7 @@ def install
 
       bin.install Dir["bin/*"]
 
+      doc.install "README.md", "LICENSE"
```
ilovezfs: @AntonioMeireles This should not be here. Those go in the prefix and are handled automatically: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/metafiles.rb

AntonioMeireles:
Of course not, You all deserve many thanks for keeping this alive... It's
just that you seem to be in a timezone similar to mine so being you to push
would reduce latency 😀

Formula/corectl.rb
ilovezfs: @AntonioMeireles ping

AntonioMeireles: missed this earlier. done. && THANKS!

AntonioMeireles:
@ilovezfs (pong). build passed ok, anything i've still missed or you just busy. 

thanks in advance

AntonioMeireles:
@ilovezfs  you 🚀 !

ilovezfs:
@AntonioMeireles Keep up the great work! By the way have you seen our new formula? https://github.com/Homebrew/homebrew-core/blob/master/Formula/govendor.rb

PR #2957, opened 2016-07-13
ooniprobe 1.6.0 & add plist

hellais:

---

### DIFF
--- a/Formula/ooniprobe.rb
+++ b/Formula/ooniprobe.rb
@@ -167,14 +167,55 @@ class Ooniprobe < Formula
   end
 
   def post_install
+    require "open3"
     system bin/"ooniresources"
-    system bin/"oonideckgen", "-o", "#{HOMEBREW_PREFIX}/share/ooni/decks/"
+    Open3.popen3("#{bin}/oonideckgen", "-o",
+                 "#{HOMEBREW_PREFIX}/share/ooni/decks/") do |_, stdout, _|
+      current_deck = stdout.read.split("\n")[-1].split(" ")[-1]
+      ln_s current_deck, "#{HOMEBREW_PREFIX}/share/ooni/decks/current.deck"
+    end
   end
 
   def caveats; <<-EOS.undent
-    Decks are installed to #{HOMEBREW_PREFIX}/share/ooni.
+    Decks are installed to #{HOMEBREW_PREFIX}/share/ooni/decks/.
     EOS
   end
+
+  def plist; <<-EOS.undent
+   <?xml version="1.0" encoding="UTF-8"?>
+   <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+   <plist version="1.0">
+   <dict>
+     <key>Label</key>
+       <string>#{plist_name}</string>
+     <key>Program</key>
+       <string>#{opt_bin}/ooniprobe</string>
+     <key>ProgramArguments</key>
+     <array>
+       <string>-i</string>
+       <string>#{HOMEBREW_PREFIX}/share/ooni/decks/current.deck</string>
+     </array>
+     <key>RunAtLoad</key>
+       <false/>
+     <key>KeepAlive</key>
+       <false/>
+     <key>StandardErrorPath</key>
+       <string>/dev/null</string>
+     <key>StandardOutPath</key>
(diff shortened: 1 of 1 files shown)


Formula/ooniprobe.rb
```diff
@@ -166,16 +166,57 @@ def install
     (var/"lib/ooni").mkpath
   end
 
-  def post_install
-    system bin/"ooniresources"
-    system bin/"oonideckgen", "-o", "#{HOMEBREW_PREFIX}/share/ooni/decks/"
```
DomT4: Do these no longer need regenerating at install, regardless of bottle creation date?

Formula/ooniprobe.rb
```diff
@@ -166,16 +166,57 @@ def install
...
+   <plist version="1.0">
+   <dict>
+     <key>Label</key>
+       <string>#{plist_name}</string>
+     <key>Program</key>
+       <string>#{opt_bin}/ooniprobe</string>
+     <key>ProgramArguments</key>
+     <array>
+       <string>-i</string>
+       <string>#{HOMEBREW_PREFIX}/share/ooni/decks/current.deck</string>
```
DomT4: Might as well make this `opt_share/ooni/decks/current.deck` since that's what it's actually pointing to.

Formula/ooniprobe.rb
DomT4: Oh, I see, you've just moved it further below. Apologies, that lost me momentarily. We'd prefer for it to be up here. `def plist` is supposed to be the second to last thing in a formula, before `test do` but after everything else.

hellais: Yes that makes sense. In fact `brew audit --strict --online ooniprobe` hinted at this :

```
  * `caveats method` (line 215) should be put before `plist block` (line 169)
```

and that was why I moved the `caveats`, but didn't move the `post_install`. Anyways will fix.

DomT4: I presume you got Git'd in the way it display its diffs. Sometimes when you move an entire block it shows the wrong one as having been moved in a diff. Can be frustrating on things like this.

Formula/ooniprobe.rb
hellais: I don't understand what you mean by this. I see `#{HOMEBREW_PREFIX}/share/ooni/` is actually pointing to `/usr/local/Cellar/ooniprobe/1.6.0/share/ooni/`.

hellais:
Squashed and force pushed the re-ordering of the methods.

Formula/ooniprobe.rb
DomT4: It's a symlink, not an actual directory. For example:
```
~> ls -l /usr/local/share/autogen
/usr/local/share/autogen -> ../Cellar/autogen/5.18.7/share/autogen
```
& `opt_prefix` is a symlink to `/Cellar/formula/current_linked_ver`.

Very few things Homebrew puts in the top-level `HOMEBREW_PREFIX` are _real_ directories, outside of `etc`, `var` and a few special cases here & there elsewhere.

DomT4: `opt_prefix` will always == `HOMEBREW_PREFIX/Cellar/formula/current_linked_ver` and `opt_share` will always == `HOMEBREW_PREFIX/Cellar/formula/current_linked_ver/share`, essentially.

The opt helpers are just a way to reference a path that doesn't break every revision/update.

hellais: Ok I understand, though wouldn't this mean that if on the next update the the plist will be referencing a path that is that of the previously installed version of the package instead of the new one?

Since plists need to be installed manually I would prefer that when an update happens it will automatically point to the newer version of the deck.

DomT4: I'm evidently doing a terrible job of explaining here.

`opt_xyz` will be safe across updates. It's merely a symlink to the linked version in the Cellar, which barring user fiddling is the latest available.

hellais: Ah ok I see what you mean. Will use `opt_prefix` there.

hellais:
I changed the `HOMEBREW_PREFIX` to `opt_share` if you want to look again.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/074b3fad9382ca08e4b59ba937da1b6043510ce9. Thank you for another contribution to Homebrew @hellais; we appreciate it! :smiley_cat:

PR #2963, opened 2016-07-13
steem 0.12.1

TigerND:


Formula/steem.rb
```diff
@@ -0,0 +1,32 @@
+class Steem < Formula
+  desc "Proof of Work blockchain with an unproven consensus algorithm."
+  homepage "https://steemit.com"
+  url "https://github.com/steemit/steem", :using => :git, :tag => "v0.8.5b", :revision => "0ea6049e760e0452ecb22b98a2ad808493bd3760"
```
DomT4: Don't need `:using => :git` here. Simply make the URL:
```ruby
url "https://github.com/steemit/steem.git"
```

DomT4: Also, please split this out a bit more so it doesn't rampage past 80 characters quite so much.
```ruby
url "https://github.com/steemit/steem.git",
    :tag => "v0.8.5b"
    :revision => "0ea6049e760e0452ecb22b98a2ad808493bd3760"
```
etc. Does this need building from a git checkout rather than archive tarball for a specific reason?

Formula/steem.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  depends_on :macos => :el_capitan
+
+  depends_on "cmake" => :build
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "python3" => :build
```
DomT4: `:python3`, not `"python3"`

Formula/steem.rb
```diff
@@ -0,0 +1,32 @@
+class Steem < Formula
+  desc "Proof of Work blockchain with an unproven consensus algorithm."
+  homepage "https://steemit.com"
+  url "https://github.com/steemit/steem", :using => :git, :tag => "v0.8.5b", :revision => "0ea6049e760e0452ecb22b98a2ad808493bd3760"
+
+  depends_on :macos => :el_capitan
+
+  depends_on "cmake" => :build
+  depends_on "autoconf" => :build
```
DomT4: It seems unlikely you require the autotools _and_ `cmake`?

Formula/steem.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on "doxygen" => :build
+  depends_on "berkeley-db" => :optional
+  depends_on "gperftools" => :optional
+  depends_on "icu4c"
+  depends_on "boost"
+  depends_on "openssl"
+  depends_on "readline"
+
+  def install
+    system "cmake", "-DENABLE_CONTENT_PATCHING=OFF", "-DLOW_MEMORY_NODE=ON", "CMakeLists.txt"
```
DomT4: You'll want a `std_cmake_args` in here somewhere. See most of our other `cmake`-using formulae for an example.

Formula/steem.rb
TigerND: because of the fact that submodules synchronization is required

Formula/steem.rb
TigerND: some submodules do use autotools

MikeMcQuaid:
@BrewTestBot test this please

TigerND:
when I specify
```ruby
depends_on :macos => :el_capitan
```
it compiles the project only on `El Captain`.

When I specify
```ruby
```
it compiles one on `Mavericks`, `Yosemite` and `El Captain`....
It's wrong because Mavericks is older than Yosemite:
```
Version 10.9: "Mavericks"
Version 10.10: "Yosemite"
Version 10.11: "El Capitan"
Version 10.12: "Sierra"
```

ilovezfs:
@MikeMcQuaid @UniqMartin currently somehow this formula is breaking CI on everything else due to the placement of the *std_cmake_args not being acceptable to Mavericks. I'm not sure how the "contamination" is occurring but it's pretty cute exploit if you wanted to bring our CI to its knees.

MikeMcQuaid:
@ilovezfs It was actually me breaking CI earlier rather than this PR.

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
...
+  depends_on "doxygen" => :build
+  depends_on "berkeley-db" => :optional
+  depends_on "gperftools" => :optional
+  depends_on "icu4c"
+  depends_on "boost"
+  depends_on "openssl"
+  depends_on "readline"
+
+  def install
+    system "cmake", *std_cmake_args, "-DENABLE_CONTENT_PATCHING=OFF", "-DLOW_MEMORY_NODE=ON", "CMakeLists.txt"
```
MikeMcQuaid: `*std_cmake_args` needs to be the last argument.

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
...
+    system "make"
+    bin.install "programs/steemd/steemd"
```
MikeMcQuaid: Is there no `make install` we can use? If not, could you request one.

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
+class Steem < Formula
+  desc "Proof of Work blockchain with an unproven consensus algorithm."
+  homepage "https://steemit.com/"
+  url "https://github.com/steemit/steem.git",
+      :tag => "v0.8.5b",
```
MikeMcQuaid: Is this a beta?

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
+class Steem < Formula
+  desc "Proof of Work blockchain with an unproven consensus algorithm."
+  homepage "https://steemit.com/"
+  url "https://github.com/steemit/steem.git",
```
MikeMcQuaid: Why build from Git rather than a tarball?

Formula/steem.rb
MikeMcQuaid: Looks like it; we'll probably need to pass until it's a stable version.

ilovezfs:
@MikeMcQuaid odd. All the failures seem to be
```
==> brew readall --aliases homebrew/core
==> FAILED
etc.
```

but I guess you're saying you caused "contamination"?

Formula/steem.rb
ilovezfs: Looks like it's designated latest release, no?
https://github.com/steemit/steem/releases

Formula/steem.rb
ilovezfs: Probably due to the submodule: https://github.com/steemit/steem/blob/master/.gitmodules

which unfortunately isn't pegged to a particular revision, so ideally it would be installed as a resource instead. However, I believe @DomT4 has said there's no firm rule regarding submodules that just point to a master HEAD unconditionally being unacceptable.

Formula/steem.rb
MikeMcQuaid: README says "Introducing Steem (beta)"

MikeMcQuaid:
@ilovezfs Yeh, that could be contamination, weird.

Formula/steem.rb
ilovezfs: Ah, OK. But that's not in contradistinction to some stable version of the same software, just the developer being humble. I suppose this could go in devel only if the "beta" label is to be taken seriously?

ilovezfs:
```
On branch master
Your branch is ahead of 'origin/master' by 23 commits.
  (use "git push" to publish your local commits)

nothing to commit, working directory clean
```

```
rtv 1.10.0
aha 0.4.9
libiodbc 3.52.12
Update autojump.rb
autojump 22.3.3
no-more-secrets 0.2.1
nethogs 0.8.5
nethogs 0.8.2
steem 0.8.5b
Jenkins thinks that Mavericks is older than Yosemite, but it's not. https://en.wikipedia.org/wiki/OS_X
Yosemite support has been added
python => python3
Minor fixes
steem 0.8.5
steem 0.8.5
Add plist to run ooniprobe measurements daily
Update ooniprobe to 1.6.0
vim 7.4.2033
ArangoDB 3.0.2: do not use versioned path in config
Artifactory 4.9.1
mbedtls: added a patch to fix "error: unknown type name 'mbedtls_time_t'"
mbedtls: the main dev branch is named "development" instead of "master"
strongswan 5.5.0
mkvtoolnix: update 9.3.1 bottle.
mkvtoolnix 9.3.1
```

Formula/steem.rb
MikeMcQuaid: I don't have strong feelings either way, honestly. Happy for you to overrule if you prefer.

ilovezfs: Might be worth having @TigerND ask the developer his preference. It makes me think of how Gmail called itself beta for five million years though :)

ilovezfs:
@MikeMcQuaid 
```
 ARGV << "--cleanup" if ENV["JENKINS_HOME"]
```
So maybe that's not getting set.

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def install
+    system "cmake", "-DENABLE_CONTENT_PATCHING=OFF", "-DLOW_MEMORY_NODE=ON", "CMakeLists.txt", *std_cmake_args
+    system "make"
+    bin.install "programs/steemd/steemd"
+    bin.install "programs/cli_wallet/cli_wallet"
+  end
+
+  test do
+    system "#{bin}/steemd", "-h"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
...
+  depends_on "gperftools" => :optional
+  depends_on "icu4c"
+  depends_on "boost"
+  depends_on "openssl"
+  depends_on "readline"
+
+  def install
+    system "cmake", "-DENABLE_CONTENT_PATCHING=OFF", "-DLOW_MEMORY_NODE=ON", "CMakeLists.txt", *std_cmake_args
+    system "make"
+    bin.install "programs/steemd/steemd"
```
MikeMcQuaid: Is there a `make install`? If not, could you request one upstream? Thanks!

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
...
+  depends_on "doxygen" => :build
+  depends_on "berkeley-db" => :optional
+  depends_on "gperftools" => :optional
+  depends_on "icu4c"
+  depends_on "boost"
+  depends_on "openssl"
+  depends_on "readline"
+
+  def install
+    system "cmake", "-DENABLE_CONTENT_PATCHING=OFF", "-DLOW_MEMORY_NODE=ON", "CMakeLists.txt", *std_cmake_args
```
MikeMcQuaid: Don't need to specify `CMakeLists.txt`

Formula/steem.rb
```diff
@@ -0,0 +1,33 @@
+class Steem < Formula
+  desc "Proof of Work blockchain with an unproven consensus algorithm."
+  homepage "https://steemit.com/"
+  url "https://github.com/steemit/steem.git",
+      :tag => "v0.9.0rc2",
+      :revision => "54337e4eae639aaf1150611923ea1e04693905e6"
+
+  depends_on :macos => [:yosemite, :el_capitan, :sierra]
```
MikeMcQuaid: Just do `depends_on :macos => :yosemite`.

Formula/steem.rb
ilovezfs: Yup, there's a make install.

Formula/steem.rb
```diff
@@ -0,0 +1,40 @@
...
+  depends_on "readline"
+
+
+  def install
+    cmake_args = %W[
+      CMakeLists.txt
+      -DLOW_MEMORY_NODE=ON
+    ]
+
+    cmake_args << "-DENABLE_CONTENT_PATCHING=OFF" if build.without? "qt5"
```
ilovezfs: Does this actually do anything? As far as I can tell the build system doesn't interact with qt at all and that option isn't used for anything.

TigerND: https://github.com/steemit/steem/blob/master/BUILD_UBUNTU.md

ilovezfs: Do you mind asking upstream if that works yet on OS X and what the requirements are?

Formula/steem.rb
```diff
@@ -0,0 +1,39 @@
+class Steem < Formula
+  desc "Proof of Work blockchain with an unproven consensus algorithm."
+  homepage "https://steemit.com/"
+  url "https://github.com/steemit/steem.git",
+      :tag => "v0.12.1",
+      :revision => "0cca4bcd8212aed662a5aba4e35a5b75be3dcb1a"
+
+  depends_on :macos => [:yosemite, :el_capitan, :sierra]
```
ilovezfs: this should just be `depends_on :macos => :yosemite`

Formula/steem.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on "berkeley-db" => :optional
+  depends_on "gperftools" => :optional
+  depends_on "qt5" => :optional
+  depends_on "icu4c"
+  depends_on "boost"
+  depends_on "openssl"
+  depends_on "readline"
+
+  def install
+    cmake_args = %W[
```
ilovezfs: This should be `= std_cmake_args + %W[` to avoid the `*(cmake_args + std_cmake_args)` at the bottotm.

Formula/steem.rb
```diff
@@ -0,0 +1,39 @@
...
+      :tag => "v0.12.1",
+      :revision => "0cca4bcd8212aed662a5aba4e35a5b75be3dcb1a"
+
+  depends_on :macos => [:yosemite, :el_capitan, :sierra]
+
+  depends_on "cmake" => :build
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on :python3 => :build
```
ilovezfs: it doesn't use python at run time?

Formula/steem.rb
DomT4: Unpegged submodules are irritating because it's yet another thing that can cause mysterious breakages that we then don't have much control over fixing. That's largely my concern about it.

Formula/steem.rb
DomT4: It's not impossible but it's uncommon.

MikeMcQuaid: Does `depends_on "python3" => :build` work? If so: that's preferable.

Formula/steem.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on "openssl"
+  depends_on "readline"
+
+  def install
+    cmake_args = %W[
+      CMakeLists.txt
+      -DCMAKE_BUILD_TYPE=Release
+      -DLOW_MEMORY_NODE=OFF
+      -DENABLE_CONTENT_PATCHING=ON
+      -DBOOST_ROOT=/usr/local/opt/boost160
```
MikeMcQuaid: `-DBOOST_ROOT=#{Formula["boost"].opt_prefix}`

Formula/steem.rb
```diff
@@ -0,0 +1,39 @@
+class Steem < Formula
+  desc "Proof of Work blockchain with an unproven consensus algorithm."
+  homepage "https://steemit.com/"
+  url "https://github.com/steemit/steem.git",
+      :tag => "0.13.0-rc3",
+      :revision => "6958b0c6cce693684a24a7594b0b056f9dbcd234"
+
+  depends_on :macos => [:yosemite, :el_capitan, :sierra]
```
MikeMcQuaid: Just do `depends_on :macos => :yosemite`

Formula/steem.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on "cmake" => :build
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on :python3 => :build
+  depends_on "doxygen" => :build
+  depends_on "berkeley-db" => :optional
+  depends_on "gperftools" => :optional
+  depends_on "icu4c"
+  depends_on "boost160"
```
MikeMcQuaid: This needs to be `depends_on "boost"`

PR #2968, opened 2016-07-13
cmake: fix breakage with hdf5

chuckatkins:
Apply upstream fix from
https://gitlab.kitware.com/cmake/cmake/issues/16190

### DIFF
--- a/Formula/cmake.rb
+++ b/Formula/cmake.rb
@@ -1,10 +1,21 @@
 class Cmake < Formula
   desc "Cross-platform make"
   homepage "https://www.cmake.org/"
-  url "https://cmake.org/files/v3.6/cmake-3.6.0.tar.gz"
-  sha256 "fd05ed40cc40ef9ef99fac7b0ece2e0b871858a82feade48546f5d2940147670"
+  revision 1
 
   head "https://cmake.org/cmake.git"
+
+  stable do
+    url "https://cmake.org/files/v3.6/cmake-3.6.0.tar.gz"
+    sha256 "fd05ed40cc40ef9ef99fac7b0ece2e0b871858a82feade48546f5d2940147670"
+
+    # This patch fixes an incompatibility with hdf5
+    # See https://gitlab.kitware.com/cmake/cmake/issues/16190
+    patch do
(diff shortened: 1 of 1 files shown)


chuckatkins:
@UniqMartin @billhoffman This should address the HDF5 issue discussed earlier.  This is also my first dealings with Homebrew so please do let me know if the formula needs adjusting and I'll fix it asap.

UniqMartin:
Merged in dd6660dc1d0cdba2783b060adbdf762590f1a6e3. Thank you for this first contribution to Homebrew, @chuckatkins! :tada:

> This should address the HDF5 issue discussed earlier.

In case anyone reading this wonders what that is referring to. I was contacted by e-mail about this because I happened to be the last one to touch the formula, thus was assumed to be the maintainer of the formula and the one to apply a fix. I replied that everyone is welcome to contribute. And that's what just happened …

> This is also my first dealings with Homebrew so please do let me know if the formula needs adjusting and I'll fix it asap.

Congratulations! For a first-time contribution this is pretty spot on (including using a `stable do` block, bumping the `revision`, and the other formatting tweaks we expect). :+1: The only thing missing from this submission is the check list from [our PR template](https://github.com/Homebrew/homebrew-core/blob/master/.github/PULL_REQUEST_TEMPLATE.md)—but maybe you never saw it because you didn't create the PR via the GitHub website?

chuckatkins:
I was the one who actually introduced the bug in cmake and just fixed it
upstream yesterday.

I did see the checklist but I didn't know it was to be included so I went
through the list, fixed a few things that came up from it, an deleted it
from the pr message.   I'll be sure to keep the checklist in the message in
the future.

Once 3.6.1 is released the patch can be removed since it's already included.

Thanks for merging so quickly!

UniqMartin:
No worries! The checklist is both for you (so hopefully many typical problems get addressed before the PR lands and we don't have to point them out in review, which speeds up the entire process, including less wasted CI time) and also for us (so that we can see whether you omitted any of the items, which is perfectly okay if you're hitting problem you cannot resolve yourself).

PR #2993, opened 2016-07-14
Mbedtls: fixed --HEAD and added a patch for current stable

andyli:
2 changes to the mbedtls formula:
- head branch is "development"
- added a patch from upstream that fixes `brew install neko --HEAD`

### DIFF
--- a/Formula/mbedtls.rb
+++ b/Formula/mbedtls.rb
@@ -1,16 +1,29 @@
 class Mbedtls < Formula
   desc "Cryptographic & SSL/TLS library"
   homepage "https://tls.mbed.org/"
-  url "https://tls.mbed.org/download/mbedtls-2.3.0-apache.tgz"
-  sha256 "590734c8bc8b3ac48e9123d44bf03562e91f8dce0d1ac2615c318c077f3215b2"
   head "https://github.com/ARMmbed/mbedtls.git", :branch => "development"
 
   stable do
+    url "https://tls.mbed.org/download/mbedtls-2.3.0-apache.tgz"
+    sha256 "590734c8bc8b3ac48e9123d44bf03562e91f8dce0d1ac2615c318c077f3215b2"
+
+    # https://github.com/ARMmbed/mbedtls/issues/522
+    # They are commits already applied to the upstream.
     patch do
-      # https://github.com/ARMmbed/mbedtls/issues/522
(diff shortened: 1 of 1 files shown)


dunn:
> added a patch from upstream that fixes brew install neko --HEAD

But you're applying the patch to the stable build only?

Why does this particular fix need to be backported?  (And if it should be, it looks like there are related commits that need to come with it: https://github.com/ARMmbed/mbedtls/issues/522#issuecomment-232680774)

Formula/mbedtls.rb
```diff
@@ -3,7 +3,16 @@ class Mbedtls < Formula
   homepage "https://tls.mbed.org/"
   url "https://tls.mbed.org/download/mbedtls-2.3.0-apache.tgz"
   sha256 "590734c8bc8b3ac48e9123d44bf03562e91f8dce0d1ac2615c318c077f3215b2"
-  head "https://github.com/ARMmbed/mbedtls.git"
+  head "https://github.com/ARMmbed/mbedtls.git", :branch => "development"
+
+  stable do
```
dunn: If you add a `stable do` block, please move all the stable lines into it (i.e., the `url` and `sha256`).

andyli:
> But you're applying the patch to the stable build only?

Yes, because the change has already be applied to upstream, thus `brew install mbedtls --HEAD` wouldn't need it.

> Why does this particular fix need to be backported? (And if it should be, it looks like there are related commits that need to come with it: ARMmbed/mbedtls#522 (comment))

Because without this fix, the other formula (neko head) will fail to build.
The other commits are added after I made this PR, and I've just added them as well.

Formula/mbedtls.rb
andyli: Revised as suggested :)

dunn:
> Because without this fix, another formula (neko) will fail to build.

Got it, I misread your original comment.

DomT4:
> Because without this fix, another formula (neko) will fail to build.

This seems like reason to `revision` bump this formula if there's active breakage?

andyli:
Bumped revision as suggested :)

dunn:
Thanks! Merged as 6b78a7498ecf33faa12ddc4fdf322b522d63a565 and. 881a58c322098df9b31215771b5fd24c1c367ffb.

PR #3001, opened 2016-07-14
ffmpeg: fix rtmpdump nonfree

JCount:

---

--with-rtmpdump causes explicit linking against openssl ;
solution: same behavior as if openssl enabled directly

This may be me being overly picky, however compiling ffmpeg with rtmpdump enabled causes ffmpeg to link to openssl directly. Currently, building with openssl enabled triggers the --enable-nonfree arg as a work-around for openssl's non-GPLv3 compatible license. At least in my mind, since rtmpdump causes similar linkage, the formula should act as if openssl had been configured, because for all intents and purposes it has.

Specifically, on my machine, ffmpeg is linked to /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib and /usr/local/opt/openssl/lib/libssl.1.0.0.dylib, in addition to /usr/local/opt/rtmpdump/lib/librtmp.1.dylib

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -146,7 +146,7 @@ class Ffmpeg < Formula
 
     # These librares are GPL-incompatible, and require ffmpeg be built with
     # the "--enable-nonfree" flag, which produces unredistributable libraries
-    if %w[faac fdk-aac openssl].any? { |f| build.with? f }
+    if %w[faac fdk-aac openssl rtmpdump].any? { |f| build.with? f }
       args << "--enable-nonfree"
     end
 


tdsmith:
background: https://people.gnome.org/~markmc/openssl-and-the-gpl.html

On its face this seems reasonable to me; what are the implications of building a non-redistributable library? Will an end user ever regret this?

ilovezfs:
CC @zmwangx

JCount:
Food for thought:
* GnuTLS and PolarSSL are offered as alternatives to OpenSSL on rtmpdump's page.
* That in mind, I modified rtmpdump's formula to build with GnuTLS
* It built successfully, and linkage showed no vestigial ties to OpenSSL
* Next, I successfully built FFmpeg with the GnuTLS version, and the OpenSSL remnants vanished.
* **However**; I have not been able to test that librtmp is actually functional with FFmpeg

zmwangx:
IMHO, this level of paranoia is unnecessary. I'm no maintainer, but from my understanding Homebrew doesn't seem to care too much about licenses as long as things are open source, and users who would like to distribute artifacts should investigate the licensing repercussions themselves (through the included LICENSE/COPYING/etc. [metafiles](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/metafiles.rb), for instance). That's why unlike, say, Debian source packages, Homebrew formulae don't include any sort of license field.

If you look at the FFmpeg formula, the licensing options `--enable-gpl`, `--enable-version3` and `--enable-nonfree` are used only to ensure `configure` will proceed. `--enable-gpl` and `--enable-version3`, in particular, are passed unconditionally for convenience. In order to be paranoid about licenses, you have to check each option against the [entire license checking block](https://github.com/FFmpeg/FFmpeg/blob/fd6dbc5/configure#L5015-L5047):

```sh
die_license_disabled() {
    enabled $1 || { enabled $2 && die "$2 is $1 and --enable-$1 is not specified."; }
}

die_license_disabled_gpl() {
    enabled $1 || { enabled $2 && die "$2 is incompatible with the gpl and --enable-$1 is not specified."; }
}

die_license_disabled gpl frei0r
die_license_disabled gpl libcdio
die_license_disabled gpl librubberband
die_license_disabled gpl libsmbclient
die_license_disabled gpl libvidstab
die_license_disabled gpl libx264
die_license_disabled gpl libx265
die_license_disabled gpl libxavs
die_license_disabled gpl libxvid
die_license_disabled gpl x11grab

die_license_disabled nonfree cuda
die_license_disabled nonfree cuvid
die_license_disabled nonfree libfaac
die_license_disabled nonfree libnpp
enabled gpl && die_license_disabled_gpl nonfree libfdk_aac
enabled gpl && die_license_disabled_gpl nonfree openssl

die_license_disabled version3 gmp
die_license_disabled version3 libopencore_amrnb
die_license_disabled version3 libopencore_amrwb
die_license_disabled version3 libsmbclient
die_license_disabled version3 libvo_amrwbenc

enabled version3 && { enabled gpl && enable gplv3 || enable lgplv3; }
```

and even then it may not be enough, which is why you're raising this issue. That's way too much trouble for us and doesn't affect 99.99% of the user base. Therefore, I'd say let's stick to what compiles, and no more.

---

This and the fact that `rtmpdump` doesn't have to use OpenSSL, as you said.

zmwangx:
I want to add one more point: FFmpeg binaries built with `--enable-nonfree` are unredistributable, but that doesn't imply the inverse, i.e., ones built without aren't necessarily redistributable. Once again, if one wants to redistribute the binaries he needs to investigate licensing repercussions himself. Homebrew doesn't help much in this regard.

JCount:
I am fine with closing this, since in practicality this really is a non-issue given the scope and complexity of the real issues this touches on. Thank you for the clarity @zmwangx 

That said, if there is interest in changing the rtmpdump formula to build with GnuTLS, I do have one I could put into a separate PR. However, it just may not be worth it.

DomT4:
I think everything is fine as-is, at the moment, and agree we're safe to close this.

As mentioned above beside making sure what we're distributing from `homebrew-core` is FOSS we don't do a lot of license policing. It's on upstream developers to check licensing compatibility if they want to integrate it into their products/projects/etc.

FWIW, OpenSSL is in the process of relicensing themselves. I think they're going about it in a [horribly complicated way](https://www.openssl.org/blog/blog/2015/08/01/cla/) but I understand the need for them to do that at this point.

Thank you nonetheless for the submission to Homebrew @JCount; appreciate you being willing to help out ❤️.

PR #3004, opened 2016-07-14
libgcrypt 1.7.2

DomT4:

---

### DIFF
--- a/Formula/libgcrypt.rb
+++ b/Formula/libgcrypt.rb
@@ -1,9 +1,9 @@
 class Libgcrypt < Formula
   desc "Cryptographic library based on the code from GnuPG"
   homepage "https://directory.fsf.org/wiki/Libgcrypt"
-  url "https://gnupg.org/ftp/gcrypt/libgcrypt/libgcrypt-1.7.1.tar.bz2"
-  mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/libgcrypt/libgcrypt-1.7.1.tar.bz2"
-  sha256 "450d9cfcbf1611c64dbe3bd04b627b83379ef89f11406d94c8bba305e36d7a95"
+  url "https://gnupg.org/ftp/gcrypt/libgcrypt/libgcrypt-1.7.2.tar.bz2"
+  mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/libgcrypt/libgcrypt-1.7.2.tar.bz2"
+  sha256 "3d35df906d6eab354504c05d749a9b021944cb29ff5f65c8ef9c3dd5f7b6689f"
 
   bottle do
     cellar :any
--- a/Formula/libgcrypt.rb
+++ b/Formula/libgcrypt.rb
@@ -38,12 +38,16 @@ class Libgcrypt < Formula
 
     # Parallel builds work, but only when run as separate steps
     system "make"
-    system "make", "install"
-    # Make check currently dies on El Capitan
-    # https://github.com/Homebrew/homebrew/issues/41599
+    # Slightly hideous hack to help `make check` work in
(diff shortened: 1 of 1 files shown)


chdiza:
@DomT4 Did you have a problem with the `make check` phase of this?  I have a weird failing test in my local setup.

chdiza:
Funny, I seem to be running into the known `make check` problem in the official formula even though I never had that problem before and all I did was bump the version.

BTW @DomT4, do we actually know that upstream is going to resolve that issue?  I followed the link to their bug tracker to [here](https://bugs.gnupg.org/gnupg/issue2056), where it looks like they have not done anything about it *and* closed the ticket.

DomT4:
> Did you have a problem with the make check phase of this? I have a weird failing test in my local setup.

No problem locally. What did you encounter on which OS X version?

> BTW @DomT4, do we actually know that upstream is going to resolve that issue? I followed the link to their bug tracker to here

Yeah. Upstream's answer seems to be by implication _"If you care, fix this yourself & submit a PR"_.

chdiza:
> No problem locally. What did you encounter on which OS X version?

I'm running a very custom set of gpg2-related formulas on El Cap.  I have a hunch about what's happening, and if I'm right, it might help upstream fix the problem.  Got to go confirm this now.

chdiza:
My hunch was right.  The mystery is why, despite the fact the official formula needs to put `make check` *after* `make install`, my custom formula does not need to do that.  It's got to do with ... sigh ... System Integrity Protection.  Mine is disabled.  So the test failure that the offical formula works around is due to the ball of fun that is SIP.

This seems like a useful data point for anyone with the expertise to finally resolve that issue.

DomT4:
> So the test failure that the offical formula works around is due to the ball of fun that is SIP.

Yeah. More specifically not being able to pass `DYLD_*` variables down to child processes. We saw the same thing with `perl`, amongst other things. It's a complete PITA to resolve, depending how deep the prior logic runs in the codebase.

chdiza:
I wonder if it's worth trying to get upstream to disable that one test on OS X.

DomT4:
Ha. This is nasty, but it works:
```ruby
system "install_name_tool", "-change", lib/"libgcrypt.20.dylib", buildpath/"src/.libs/libgcrypt.20.dylib", buildpath/"tests/.libs/random"

```

DomT4:
It's interesting, because all the compiled test executables have this sort of blob of code:
```c
  if test -f "$progdir/$program"; then
    # Add our own library path to DYLD_LIBRARY_PATH
    DYLD_LIBRARY_PATH="/private/tmp/libgcrypt-20160714-67958-12k9o8o/libgcrypt-1.7.2/src/.libs:$DYLD_LIBRARY_PATH"

    # Some systems cannot cope with colon-terminated DYLD_LIBRARY_PATH
    # The second colon is a workaround for a bug in BeOS R4 sed
    DYLD_LIBRARY_PATH=`$ECHO "$DYLD_LIBRARY_PATH" | /usr/local/Library/ENV/4.3/sed 's/::*$//'`

    export DYLD_LIBRARY_PATH

    if test "$libtool_execute_magic" != "%%%MAGIC variable%%%"; then
      # Run the actual program with our arguments.
      func_exec_program ${1+"$@"}
    fi
```

But only `random` is failing. So _something_ must be going wrong about `random`.

chdiza:
Will that also work when SIP is disabled?

DomT4:
Seems to, looking at the pre-SIP OS X versions on CI. You're welcome to test locally if you want to double check?

chdiza:
The CI is good enough for me.

I have SIP turned on again in the last couple of days, which is why I was suddenly bit today even though HB was bit a while ago.  I turned it off to test my hunch, and then turned it back on again.

PR #3006, opened 2016-07-14
flow 0.29.0

madjam002:


gabelevi:
If #3006 works, I'd rather have that merged, since it addressed the TODO in flow.rb. If that doesn't apply, then this is cool tool.

Thanks for updating this so fast, though! You beat me :)

DomT4:
Let's go for https://github.com/Homebrew/homebrew-core/pull/3007 since it resolves the tarball issue, but thank you for your submission to Homebrew @madjam002. Appreciate the willingness to help out ❤️.

PR #3014, opened 2016-07-15
pandoc-crossref 0.2.2.1

ilovezfs:
now builds with GHC 8 out of the box, so remove the upstreamed fixes


ilovezfs:
@MikeMcQuaid bye bye grotesque string replacements!

MikeMcQuaid:
@ilovezfs Nice!

PR #3015, opened 2016-07-15
mypy 0.4.3

ilovezfs:
use sphinx-doc from homebrew/core instead of vendoring it in the buildpath.

### DIFF
--- a/Formula/mypy.rb
+++ b/Formula/mypy.rb
@@ -2,8 +2,8 @@ class Mypy < Formula
   desc "Experimental optional static type checker for Python"
   homepage "http://www.mypy-lang.org/"
   url "https://github.com/JukkaL/mypy.git",
-    :tag => "v0.4.2",
-    :revision => "b8a2df91707bc3d5c9c3677c8017b614e41d017a"
+      :tag => "v0.4.3",
+      :revision => "e5f27adf8f143604af241533cdd69df3ddb8d1c9"
   head "https://github.com/JukkaL/mypy.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@MikeMcQuaid CI still seems to be a bit broken. homebrew/core isn't getting set back to origin/master at the beginning for some reason

MikeMcQuaid:
@ilovezfs Taking a look.

ilovezfs:
@MikeMcQuaid cool, thanks: https://github.com/Homebrew/brew/issues/515

MikeMcQuaid:
Sorted.

ilovezfs:
@tdsmith :ship:?

Formula/mypy.rb
```diff
@@ -13,97 +13,32 @@ class Mypy < Formula
     sha256 "bed90ec74eb884802e768059c10d435a5d7f7be2b6b990a1df6bc92291aed4d3" => :mavericks
   end
 
-  option "without-docs", "Don't build documentation"
+  option "without-sphinx-doc", "Don't build documentation"
```
DomT4: Seems worthy of a `deprecated_option`?

ilovezfs: fixed.

Formula/mypy.rb
```diff
@@ -2,8 +2,8 @@ class Mypy < Formula
   desc "Experimental optional static type checker for Python"
   homepage "http://www.mypy-lang.org/"
   url "https://github.com/JukkaL/mypy.git",
-    :tag => "v0.4.2",
-    :revision => "b8a2df91707bc3d5c9c3677c8017b614e41d017a"
+      :tag => "v0.4.3",
```
MikeMcQuaid: Why the increased indent?

ilovezfs: I'm happy either way but this seems to be @DomT4's preferred indentation for these based on inspection of other formulae. (Please correct me if I'm wrong @DomT4!)

We also have this exact same discrepancy with `go_resource` blocks right now where some of them are like this
(a)
```
  go_resource "github.com/BurntSushi/toml" do
    url "https://github.com/BurntSushi/toml.git",
    :revision => "f0aeabca5a127c4078abb8c8d64298b147264b55"
  end
```
(b)

Some are like this
```
  go_resource "github.com/BurntSushi/toml" do
    url "https://github.com/BurntSushi/toml.git",
      :revision => "f0aeabca5a127c4078abb8c8d64298b147264b55"
  end
```
and some are like this
(c)
```
  go_resource "github.com/BurntSushi/toml" do
    url "https://github.com/BurntSushi/toml.git",
        :revision => "f0aeabca5a127c4078abb8c8d64298b147264b55"
  end
```
With tags there's also a fourth option, where people stick both on the same line, again with the indentation starting at random locations.

From what I can gather (a) is preferred by no one but is being introduced by the gdm tool, which is doing them wrong, but that there is a pervasive lack of consistency between (b) and (c), including with tags, as above.

It would be nice IMO to have a rule or rules for this that we follow consistently, whatever those rule(s) might be :)

DomT4: I do prefer the equal indent, whether the `:` lines up with `url` or lines up with the start of the `"<url>` itself, but I'm not particularly passionate about it.

The `go_resource`s tend to differ depending on whether they were written by a human or by `gdm`, heh.

MikeMcQuaid: ah, I see why now. I don't mind either way 😀

ilovezfs: >whether the : lines up with url

@UniqMartin expressed a clear preference against (a).

DomT4: Let's go for an indent like this then:
```ruby
url "https://shiny-example.github.com.git",
    :revision => "f0aeabca5a127c4078abb8c8d64298b147264b55"
```

ilovezfs: There's also a discrepancy in the indentation of multi-line system command invocations. For configure, `brew create` produces
```
                          "--disable-dependency-tracking",
                          "--disable-silent-rules",
                          "--prefix=#{prefix}"
```

However, for `go` formulae, when a system command invocation wraps, the "correct" way is apparently:
(d)
```
    system "go", "build", "-o", bin/"slackcat",
           "-ldflags", "-X main.version=#{version}"
```
or
(e)
```
 system "go", "build",
        "-o", bin/"slackcat",
        "-ldflags", "-X main.version=#{version}"
```
but not
(f)
```
    system "go", "build", "-o", bin/"slackcat",
                          "-ldflags", "-X main.version=#{version}"
```

I'm not sure what accounts for this discrepancy and which is actually our preferred style. If it is indeed (d), it seems the `brew create` template should be changed accordingly, and non-compliant formulae updated.

UniqMartin: For invocations of `./configure` we clearly prefer to align the options. For other tools and binaries where there's maybe less structure and an intermediate subcommand, this rule doesn't work as well. (I haven't seen a style that is both consistent and looks beautiful in all circumstances.)

The `./configure` style is also responsible for why [none of our method invocations are checked for proper argument alignment](https://github.com/Homebrew/brew/blob/b6772c1fde5878fdb492aad8b4f9c0bc10284a29/Library/.rubocop.yml#L102-L104) and why some of those `url` examples got away with no indentation at all. The only modes RuboCop supports is either:

- align with the first argument in the first line or
- indent one step (two spaces)

for any arguments that don't fit on the first line. I don't have a strong preference, should we consider enabling this check, as long as we indent method arguments at all.

DomT4: Given all the ways we use indent in Homebrew, even within formulae, I'm not sure there's ever going to be a simple, programmatic way to enforce it to be honest... happy to be proven wrong in time though.

ilovezfs:
```
-   align with the first argument in the first line or
-   indent one step (two spaces)
```
@DomT4 Perhaps we should choose one of the supported modes and stick to it.

DomT4: I'm not sure it's worth trying to enforce an indent style onto formulae at this point, where those are the only two options. It's semi-reasonable to say that different usages merit different indents & enforcing consistency without more active consideration may actually make things look even stranger.

ilovezfs: I disagree unless our "custom" requirements are being enforced programatically. It would be better to just go with whatever RuboCop natively supports (unless someone is willing to do the work to customize it) rather than chasing people around to fix this or that indentation in every review where they've stepped outside our nebulous lines.

DomT4: >  (unless someone is willing to do the work to customize it) 

Honestly, I think this is where we need to end up, similar to [cask](https://github.com/caskroom/rubocop-cask). A custom Rubocop that works for Homebrew, that we can write our own desires into. It's just finding the time for any of us to take the steps towards getting that done that's a pain.

ilovezfs: The first step toward making such a thing happen is clarifying exactly what our custom requirements are if we had our heart's desire.

PR #3016, opened 2016-07-15
scala 2.12.0-M5 (devel)

cheleb:
- [x ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [ ?] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

### DIFF
--- a/Formula/scala.rb
+++ b/Formula/scala.rb
@@ -27,22 +27,22 @@ class Scala < Formula
   end
 
   devel do
-    url "http://www.scala-lang.org/files/archive/scala-2.12.0-M4.tgz"
-    sha256 "2ea78303f1d1ba53ed4a1b905e22ea7a14556da0b31eb081f21a1b723bf672b7"
-    version "2.12.0-M4"
+    url "http://www.scala-lang.org/files/archive/scala-2.12.0-M5.tgz"
+    sha256 "fd260bd1d2fb7ba1e8003d40a463959a4111a3e4a25194e8d9e54e23cb39dd08"
+    version "2.12.0-M5"
 
     depends_on :java => "1.8+"
 
     resource "docs" do
-      url "http://www.scala-lang.org/files/archive/scala-docs-2.12.0-M4.zip"
-      sha256 "ccfbc72c5bfd3743b08f4ecfb3dea46cb2b982eaaba5ada24378a93c351f8c77"
-      version "2.12.0-M4"
+      url "http://www.scala-lang.org/files/archive/scala-docs-2.12.0-M5.zip"
+      sha256 "025d4db9755ffe0afc7254a105f95ca0ea6a77385d5c9b51a441aa8f34aefa5e"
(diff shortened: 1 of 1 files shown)


Formula/scala.rb
```diff
@@ -27,22 +27,22 @@ class Scala < Formula
...
 
     depends_on :java => "1.8+"
 
     resource "docs" do
-      url "http://www.scala-lang.org/files/archive/scala-docs-2.12.0-M4.zip"
-      sha256 "ccfbc72c5bfd3743b08f4ecfb3dea46cb2b982eaaba5ada24378a93c351f8c77"
-      version "2.12.0-M4"
+      url "http://www.scala-lang.org/files/archive/scala-docs-2.12.0-M5.zip"
+      sha256 "025d4db9755ffe0afc7254a105f95ca0ea6a77385d5c9b51a441aa8f34aefa5e"
+      version "2.12.0-M5"
```
ilovezfs: @DomT4 resource blocks can have versions? I'm confused :)

DomT4: Yeah, they can. Unnecessary in most cases, but yeah, they don't automatically inherit their parent's version.

ilovezfs: Is that ever used anywhere?

DomT4: Resource versioning? Mostly for cache sanity, I believe.

ilovezfs: I'm still not convinced this isn't an outright bug.

DomT4: Most of our versioning system is an outright bug :trollface:. Let's get this merged so we don't spam OP with brew implementation details, but agree at some point we need to handle `-` separated versions more sanely.

ilovezfs: What I'm saying is that I think OP could delete those versions from the formula and nothing will break.

DomT4: The version changes to `scala--docs-4.zip`, etc, which means we have resources in the cache that don't align with the expected devel version. That has the potential to cause issues when trying to cleanup the cache, etc.

ilovezfs: Wow, we don't propagate the explicit version from the spec? That's a pretty awesome cache bug. I'll open an issue in brew repo.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/f73de34c442fa6c599453877f349d6907461f69c. Thank you for your contribution to Homebrew @cheleb; we appreciate it! :smiley_cat:

PR #3017, opened 2016-07-15
sqlite: add regexp extension support

megahall:

---

@plumbojumbo @vszakats @twexler @DomT4 @bfontaine

Hello recent sqlite maintainers,

I am working on this PR for sqlite regex support. However I ran into some
strange problems and I could use your advice.
1. The build uses `sqlite-autoconf-3130000` instead of `sqlite-src-3100200`.
   This combined / mashed-up copy of the code does not include
   `ext/misc/regexp.c`. So I tried to follow the pattern of the `functions`
   support which downloads the extra file.
2. The Fossil SCM repository seems to prevent direct-linking to the latest
   version of the file... the interface it uses for file downloads is totally
   bizarre:
   
   Default (Hardcoded) URL:
   `http://www.sqlite.org/src/raw/ext/misc/regexp.c?name=a68d25c659bd2d893cd1215667bbf75ecb9dc7d4`
   
   Other Failed Attempts:
   `http://www.sqlite.org/src/raw/ext/misc/regexp.c`: returns repo homepage
   `http://www.sqlite.org/src/raw/ext/misc/regexp.c?name=trunk`: returns some bogus internal SCM file
   
   Can I fix this issue, or should I host the file elsewhere? Right now I
   cheated and used a local webserver as a POC to get feedback first.
3. Something kind of bad has changed about the extension library and extension
   function naming. Loading failed when the extensions were named
   `libsqlitefunctions.dylib` and `libsqliteregexp.dylib`. Loading succeeds after
   I rename them to `libfunctions.dylib` and `libregexp.dylib`. However I was
   concerned this could cause some kind of unexpected breakage if these `SONAMES`
   are too common. Note this also means the upstream `functions` feature is
   broken as-is. I fixed it here temporarily but I am not sure if I did it right.

Sincerely,
@megahall

### DIFF
--- a/Formula/sqlite.rb
+++ b/Formula/sqlite.rb
@@ -84,9 +84,9 @@ class Sqlite < Formula
       system ENV.cc, "-fno-common",
                      "-dynamiclib",
                      "extension-functions.c",
-                     "-o", "libsqlitefunctions.dylib",
+                     "-o", "libfunctions.dylib",
                      *ENV.cflags.to_s.split
-      lib.install "libsqlitefunctions.dylib"
+      lib.install "libfunctions.dylib"
     end
 
     if build.with? "regexp"
@@ -117,7 +117,7 @@ class Sqlite < Formula
 
         If the program is built so that loading extensions is permitted,
         the following will work:
-         sqlite> SELECT load_extension('#{lib}/libsqlitefunctions.dylib');
+         sqlite> SELECT load_extension('#{lib}/libfunctions.dylib');
          sqlite> select cos(radians(45));
          0.707106781186548
          sqlite> SELECT load_extension('#{lib}/libregexp.dylib');
--- a/Formula/sqlite.rb
+++ b/Formula/sqlite.rb
@@ -23,6 +23,7 @@ class Sqlite < Formula
   option "with-unlock-notify", "Enable the unlock notification feature"
   option "with-icu4c", "Enable the ICU module"
   option "with-functions", "Enable more math and string functions for SQL queries"
+  option "with-regexp", "Enable regexp function and operator for SQL queries"
   option "with-dbstat", "Enable the 'dbstat' virtual table"
   option "with-json1", "Enable the JSON1 extension"
   option "with-session", "Enable the session extension"
@@ -34,6 +35,12 @@ class Sqlite < Formula
     url "https://sqlite.org/contrib/download/extension-functions.c?get=25", :using => :nounzip
     version "2010-01-06"
     sha256 "991b40fe8b2799edc215f7260b890f14a833512c9d9896aa080891330ffe4052"
+  end
+
+  resource "regexp" do
+    url "http://localhost/regexp.c", :using => :nounzip
+    version "2016-04-12"
+    sha256 "a079722fe557d8a510021990af061e9f16026344e05b9f407bee60de712d968e"
   end
 
   resource "docs" do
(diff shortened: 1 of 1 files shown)


Formula/sqlite.rb
```diff
@@ -36,6 +37,12 @@ class Sqlite < Formula
     sha256 "991b40fe8b2799edc215f7260b890f14a833512c9d9896aa080891330ffe4052"
   end
 
+  resource "regexp" do
+    url "http://localhost/regexp.c", :using => :nounzip
```
ilovezfs: why the localhost url?

vszakats:
Using URL `https://www.sqlite.org/src/raw/ext/misc/regexp.c?name=a68d25c659bd2d893cd1215667bbf75ecb9dc7d4` should be fine I think. Linking to the latest version dynamically would cause the checksum to fail as soon as the source file is updated in its repository.

DomT4:
> The build uses sqlite-autoconf-3130000 instead of sqlite-src-3100200. This combined / mashed-up copy of the code does not include ext/misc/regexp.c.

Is this deliberate upstream? It may well be but that smells like a packaging bug.

Formula/sqlite.rb
megahall: See item 2 in the PR body. This is not fully ready to merge until I get full feedback and a plan for dealing with item 2.

MikeMcQuaid: We don't want to link to the latest version anyway.

MikeMcQuaid: So just use a particular revision. Out of interest, though, why isn't this included with SQLite itself?

plumbojumbo:
>> The build uses sqlite-autoconf-3130000 instead of sqlite-src-3100200.
This combined / mashed-up copy of the code does not include
ext/misc/regexp.c. So I tried to follow the pattern of the functions
support which downloads the extra file.

> Is this deliberate upstream? It may well be but that smells like a packaging bug.

It is deliberate. Not all the extensions from `ext/` are equal. Some of them (e. g. `fts*`, `icu`, `json1`, `rbu`, `rtree`, `session`) are supported "officially". They are contained in the amalgation/autoconf sources and have `./configure` switches, but are not included by default.

Other extensions in `ext/` (like `regexp`) are not officially supported and only present in the full snapshot. Most of these are considered sophisticated and/or useful demonstrations of the extension and virtual table APIs.

> Something kind of bad has changed about the extension library and extension
function naming. Loading failed when the extensions were named
libsqlitefunctions.dylib and libsqliteregexp.dylib. Loading succeeds after
I rename them to libfunctions.dylib and libregexp.dylib.

[See](http://sqlite.org/c3ref/load_extension.html). In case of the `regexp` extension, if you name the library `libsqliteregexp.dylib`, you need to provide the correct entry point (`sqlite_regexp_init`) when loading it, otherwise SQLite will look for `sqlite_sqliteregexp_init` and fail. The CLI `.load` command, the `load_extension` SQL function, and the `sqlite_load_extension` C API all have appropriate pramaters for that. Naming the library `libregexp.dylib` allows it to be loaded without having to specify the entry point.

Loading `libsqlitefunctions.dylib` has always worked for me. Actually, this extension uses the generic `sqlite_extension_init` entry point and should load fine whatever you name it, unless something else goes wrong.

> However I was concerned this could cause some kind of unexpected breakage if these SONAMES are too common.

They are keg-only anyway, only symlinked to `/usr/local/opt/sqlite/lib/`. You wouldn't compile anything with them as a dependency, and when loading an extension you specify it's path, not it's install_name, so this shouldn't be a problem.

> Note this also means the upstream functions feature is broken as-is. I fixed it here temporarily but I am not sure if I did it right.

How exactly are you trying to load it? Note also that the `functions` extension is not an upstream extension but a [contributed one](https://sqlite.org/contrib).

tdsmith:
A tap might be the best place to maintain a sqlite build with experimental or demonstration features.

DomT4:
Yeah, if I'm reading @plumbojumbo's comment correctly I think we're best passing on this at least until a point in time where it's officially supported & included in the autoconf release.

PR #3034, opened 2016-07-15
c-kermit: fix test, fix code style issues

UniqMartin:

---

Testing with the `PWD` command turns out to be problematic because it prints the real path, but `testpath` is the logical path, resulting in:

```
==> /usr/local/Cellar/c-kermit/9.0.302/bin/kermit -C PWD,exit
Error: c-kermit: failed
<"/tmp/c-kermit-test-20160711-81517-1v46fdo"> expected but was
<"/private/tmp/c-kermit-test-20160711-81517-1v46fdo">.
```

### DIFF
--- a/Formula/c-kermit.rb
+++ b/Formula/c-kermit.rb
@@ -24,6 +24,7 @@ class CKermit < Formula
   end
 
   test do
-    assert_equal "#{testpath}", shell_output("#{bin}/kermit -C PWD,exit").chomp
+    assert_match "C-Kermit #{version}",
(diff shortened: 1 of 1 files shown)


DomT4:
Adopting `.realpath` would work too, I suspect? But change seems reasonable enough as-is.

UniqMartin:
Yes, sticking `.realpath` in the right place would work, too. But I figured checking just the output of the `VERSION` command is basically as good as testing with the `PWD` command.

PR #3044, opened 2016-07-16
kubernetes-helm: update url and checksum

ilovezfs:
The repository was renamed, which necessarily changes the checksum.

### DIFF
--- a/Formula/kubernetes-helm.rb
+++ b/Formula/kubernetes-helm.rb
@@ -1,8 +1,8 @@
 class KubernetesHelm < Formula
   desc "The package manager for Kubernetes"
   homepage "https://helm.sh"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Technically it shouldn't, actually but if it's changed 👍

ilovezfs:
@MikeMcQuaid it does because the top level internal folder in the archive has the name of the repository, though I suppose GitHub could preserve the old tarballs if that's desirable. This same thing happened when the let's encrypt repo was renamed certbot.

ilovezfs:
This one is probably not worth fixing because there's a new version and some of the dependencies have moved as well, though I usually prefer to leave things in a working state before upgrading them.

PR #3059, opened 2016-07-16
vifm 0.8.2

ilovezfs:
- use solely the GitHub tarball
- prevent opportunistic linking against libmagic, GTK+, and X11
- run "make check"
- use a string for the substring match in the test

### DIFF
--- a/Formula/vifm.rb
+++ b/Formula/vifm.rb
@@ -1,9 +1,8 @@
 class Vifm < Formula
   desc "Ncurses based file manager with vi like keybindings"
   homepage "https://vifm.info/"
-  url "https://downloads.sourceforge.net/project/vifm/vifm/vifm-0.8.1a.tar.bz2"
-  mirror "https://github.com/vifm/vifm/releases/download/v0.8.1a/vifm-0.8.1a.tar.bz2"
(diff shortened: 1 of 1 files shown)


DomT4:
Tempting to just make the GitHub release the default to avoid SF, but 👍.

ilovezfs:
@DomT4 the checksums match so that sounds like a good idea.

ilovezfs:
@UniqMartin interestingly this one has the same parallelization issue as xorriso (https://github.com/Homebrew/homebrew-core/pull/1941#discussion_r67043692) where copying our own install-sh from our automake formula fixes the make install issue. Not good.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Can we consider adding less options until we get users breaking down the door asking for 'em? That `:linked` system for the GTK tree looks pretty much as bad as I feared it would.

ilovezfs:
@DomT4 Let's kill them all and always pass "--without-libmagic" to prevent the ridiculous opportunistic linking, and probably also "--without-gtk+" and "--without-X11" as a hedge against future opportunistic linking.

ilovezfs:
@DomT4 Better?

DomT4:
👍. We can always consider adding options if people come asking for 'em.

PR #3060, opened 2016-07-16
imagemagick: add options to disable modules and threads

jeroen:
Add options to build without threads and without modules.

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -28,6 +28,8 @@ class Imagemagick < Formula
   option "with-quantum-depth-32", "Compile with a quantum depth of 32 bit"
   option "without-opencl", "Disable OpenCL"
   option "without-magick-plus-plus", "disable build/install of Magick++"
+  option "without-modules", "Disable support for dynamically loadable modules"
+  option "without-threads", "Disable threads support"
 
   depends_on "xz"
   depends_on "libtool" => :run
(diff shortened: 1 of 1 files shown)


Formula/imagemagick.rb
```diff
@@ -65,9 +65,14 @@ def install
       --disable-silent-rules
       --enable-shared
       --enable-static
-      --with-modules
     ]
 
+    if build.without? "modules"
```
DomT4: You'll need to actually add the related `option "without-modules", "Build without modules"` and same with threads.

Can you share a little more on why it would be desirable to ship without modules/threads?

jeroenooms:
OK I documented the options.

The reason is that we need build a completely self contained application which statically links to magick++. I found that `--enable-static --with-modules` builds the actual output modules (png, pdf, etc) only as shared objects. Because these in our application these are unavailable at the user machine, nothing works. 

Only when building with `--enable-static --without-modules` the entire magick++ library will be contained in the static libraries and we can create a standalone application.

jeroenooms:
squashed the commits.

jeroenooms:
By the way `--without-threads` is useful if you want to link magick into an environment which is not thread safe. But this option is less critical to me than `--without-modules`.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/4f8ca679a0a01bb4d02e7cd8674c33494ec848e1. Thanks @jeroenooms!

Issue #2946, opened 2016-07-13
ffmpeg failed to build on 10.11.5


hughbiquitous:
ffmpeg build failed during `make install` when installing ffmpeg with options:

```
  --with-faac
  --with-fdk-aac
  --with-ffplay
  --with-fontconfig
  --with-freetype
  --with-libass
  --with-libbluray
  --with-libvorbis
  --with-libvpx
  --with-openh264
  --with-openjpeg
  --with-openssl
  --with-opus
  --with-rtmpdump
  --with-tools
  --with-webp
  --with-x265
  --with-xz
```

Console output during `brew install`:

```
==> make install
Last 15 lines from /Users/hugh/Library/Logs/Homebrew/ffmpeg/02.make:
libavcodec/libopenh264enc.c:198:29: error: no member named 'sSliceCfg' in 'SSpatialLayerConfig'
    param.sSpatialLayers[0].sSliceCfg.uiSliceMode               = s->slice_mode;
    ~~~~~~~~~~~~~~~~~~~~~~~ ^
libavcodec/libopenh264enc.c:199:29: error: no member named 'sSliceCfg' in 'SSpatialLayerConfig'
    param.sSpatialLayers[0].sSliceCfg.sSliceArgument.uiSliceNum = avctx->slices;
    ~~~~~~~~~~~~~~~~~~~~~~~ ^
libavcodec/libopenh264enc.c:201:26: error: use of undeclared identifier 'SM_DYN_SLICE'
    if (s->slice_mode == SM_DYN_SLICE) {
                         ^
libavcodec/libopenh264enc.c:204:37: error: no member named 'sSliceCfg' in 'SSpatialLayerConfig'
            param.sSpatialLayers[0].sSliceCfg.sSliceArgument.uiSliceSizeConstraint = s->max_nal_size;
            ~~~~~~~~~~~~~~~~~~~~~~~ ^
9 errors generated.
```

See https://gist.github.com/hughbiquitous/55ec160b5e739686f0256ef777bed594.

MikeMcQuaid:
Could you `brew update` and try again? We may have fixed this. Thanks!

hughbiquitous:
Thanks Mike, but sadly, the same result. Did `brew update` x2 and `brew doctor` before attempting. New gist: https://gist.github.com/096fbdb8faf9f93a17690c05ceb40a09

henryborchers:
I also had the same problem so it's not just you.

JCount:
[...]

PR #2971, opened 2016-07-13
OpenH264 1.6.0 retag fix

JCount:
- [X] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [X] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [X] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

Fix sha256 mismatch likely due to release retagging

PR #3070, opened 2016-07-17
ffmpeg: vendor openh264 1.5.0

ilovezfs:
This can be reversed when upstream ffmpeg supports openh264 1.6.0.

Closes https://github.com/Homebrew/homebrew-core/issues/2946.

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -77,9 +77,17 @@ class Ffmpeg < Formula
   depends_on "libbs2b" => :optional
   depends_on "rubberband" => :optional
   depends_on "zimg" => :optional
-  depends_on "openh264" => :optional
   depends_on "xz" => :optional
   depends_on "libebur128" => :optional
+
+  depends_on "nasm" => :build if build.with? "openh264"
+
+  # Remove when ffmpeg has support for openh264 1.6.0
+  # See https://github.com/cisco/openh264/issues/2505
+  resource "openh264-1.5.0" do
+    url "https://github.com/cisco/openh264/archive/v1.5.0.tar.gz"
+    sha256 "98077bd5d113c183ce02b678733b0cada2cf36750370579534c4d70f0b6c27b5"
+  end
 
   def install
     args = %W[
@@ -94,6 +102,15 @@ class Ffmpeg < Formula
       --host-cflags=#{ENV.cflags}
       --host-ldflags=#{ENV.ldflags}
     ]
+
+    if build.with? "openh264"
+      resource("openh264-1.5.0").stage do
+        system "make", "install-shared", "PREFIX=#{libexec}/openh264-1.5.0"
+        chmod 0444, libexec/"openh264-1.5.0/lib/libopenh264.dylib"
+      end
+      ENV.prepend_path "PKG_CONFIG_PATH", libexec/"openh264-1.5.0/lib/pkgconfig"
+      args << "--enable-libopenh264"
+    end
 
     args << "--enable-opencl" if MacOS.version > :lion
 
@@ -129,7 +146,6 @@ class Ffmpeg < Formula
     args << "--enable-librubberband" if build.with? "rubberband"
     args << "--enable-libzimg" if build.with? "zimg"
     args << "--disable-indev=qtkit" if build.without? "qtkit"
-    args << "--enable-libopenh264" if build.with? "openh264"
     args << "--enable-libebur128" if build.with? "libebur128"
 
     if build.with? "xz"


ilovezfs:
CC @MikeMcQuaid @zmwangx

zmwangx:
LGTM but as a x264 user I really can't say much. (One thing that comes to mind: will this cause user confusion? Looking at the [release notes of openh264 1.6.0](https://github.com/cisco/openh264/releases/tag/v1.6.0) I can't tell if it's big for users or not.) But I'll make sure to test FFmpeg against the latest openh264 when the next release comes (if others don't get ahead of me).

ilovezfs:
@zmwangx Another option is this patch, which does at least fix the build: http://ffmpeg.org/pipermail/ffmpeg-devel/2016-May/194492.html

There was a bit of discussion on the ffmpeg mailing list about this, which seemed to revolve around two things: a) people not liking other people's patches b) hostility to patches for an unreleased version of openh264. Clearly the latter concern has been resolved. Not so sure about the former or how many different patches have actually been proposed to date.

zmwangx:
My vote on shipping this PR and wait for user feedback (if any).

ilovezfs:
>My vote on shipping this PR and wait for user feedback (if any).

Yeah, it basically restores the situation to the status quo at the time you originally shipped 3.1.1, and fixes the regression I caused when I shipped openh264 1.6.0 without checking whether anything with an optional dependency on it was impacted.

DomT4:
The checksum for the latest release in `openh264` changed recently & we're not sure why yet: https://github.com/Homebrew/homebrew-core/pull/2971. Probably a retag but they haven't responded upstream. Possibly worth checking the SHA for `1.5.0` isn't different as well, just in case.

zmwangx:
@DomT4 I just checked and the checksum is right, and it's the same as what we used to have for openh264 1.5.0.

DomT4:
Cool, thanks for checking!

@ilovezfs Maybe worth a short in-formula notation as to why this needs vendoring & when it can be removed.

ilovezfs:
Isn't it there already?

ilovezfs:
line 85-86

DomT4:
Ha, yes. Yeah, I was looking at the `stage` block rather than the `resource` block. Excuse the brain 💨.

ilovezfs:
No, it's something I wonder about: whether it's best to put such things by the resource-do, or by resource.stage, or both. And when a head and stable both have the same issue being patched but in two different ways, what's the official way to cross reference, etc. Rules! There must be rules!

MikeMcQuaid:
👍

Formula/ffmpeg.rb
```diff
@@ -77,10 +77,18 @@ class Ffmpeg < Formula
   depends_on "libbs2b" => :optional
   depends_on "rubberband" => :optional
   depends_on "zimg" => :optional
-  depends_on "openh264" => :optional
   depends_on "xz" => :optional
   depends_on "libebur128" => :optional
 
+  depends_on "nasm" if build.with? "openh264"
```
DomT4: Is this actually runtime? `nasm` almost never is.

Formula/ffmpeg.rb
```diff
@@ -95,6 +103,15 @@ def install
       --host-ldflags=#{ENV.ldflags}
     ]
 
+    if build.with? "openh264"
+      resource("openh264-1.5.0").stage do
+        system "make", "install-shared", "PREFIX=#{libexec}/openh264-1.5.0"
+        chmod 0444, "#{libexec}/openh264-1.5.0/lib/libopenh264.dylib"
```
DomT4: Very unlike you to use `#{libexec}` instead of `libexec` ILZ, are you feeling okay? :trollface:

DomT4:
>  Rules! There must be rules!

This is fine how it is, IMO. It's how we handle patches, which seems to be easily understood by people.

Formula/ffmpeg.rb
ilovezfs: I had copy-pasted from here https://github.com/Homebrew/homebrew-core/blob/master/Formula/openh264.rb#L19 and then fixed up the prefixes. But this was a gross oversight, I admit :)

Formula/ffmpeg.rb
ilovezfs: Yup you're correct https://github.com/Homebrew/homebrew-core/blob/master/Formula/openh264.rb#L15

ilovezfs:
>This is fine how it is, IMO. It's how we handle patches, which seems to be easily understood by people.

Yes, that's a good point. Since there's no analogue to `.stage` for patches, it makes sense to put the comments for resources above their blocks just as we do with patches to be consistent. Also, it has the benefit of having the comment be placed where the item first occurs in the formula.

PR #3080, opened 2016-07-18
unbound 1.5.9

fnevesbr:

---

Updated new version source and signature
Included dependency on expat

### DIFF
--- a/Formula/unbound.rb
+++ b/Formula/unbound.rb
@@ -1,8 +1,8 @@
 class Unbound < Formula
   desc "Validating, recursive, caching DNS resolver"
   homepage "https://www.unbound.net"
-  url "https://unbound.net/downloads/unbound-1.5.8.tar.gz"
-  sha256 "33567a20f73e288f8daa4ec021fbb30fe1824b346b34f12677ad77899ecd09be"
+  url "https://unbound.net/downloads/unbound-1.5.9.tar.gz"
+  sha256 "01328cfac99ab5b8c47115151896a244979e442e284eb962c0ea84b7782b6990"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Formula/unbound.rb
```diff
@@ -13,11 +13,13 @@ class Unbound < Formula
 
   depends_on "openssl"
   depends_on "libevent"
+  depends_on "expat"
```
DomT4: Does it need a newer version than the system provided one now?

fnevesbr: Strangely on my machine I could not build without including the expart
module from homebrew

$ uname -a

Darwin ... 15.5.0 Darwin Kernel Version 15.5.0: Tue Apr 19 18:36:36 PDT 2016

DomT4: Do you have a `gist-logs` for that build at all, by any chance? It's entirely possible the OS X `expat` is too old now, just nice to check we're not adding dependencies unnecessarily.

fnevesbr: Could be a problema on my Xcode installation. Even 1.5.8 formula doesn't
build with the same error, my previous installation was done using a
"bottled" version.

Does this help?

% brew gist-logs unbound

https://gist.github.com/cfdbc58665104f9cb38cb61da7305f6a

DomT4: Ah, I'm almost certain it's because you don't have the Command Line Tools package installed, which is perfectly valid & supported by both us & Apple, but some upstreams struggle with it. Can you remove the `expat` dependency & try this locally:

```ruby
def install
  args = %W[
    --prefix=#{prefix}
    --sysconfdir=#{etc}
    --with-libevent=#{Formula["libevent"].opt_prefix}
    --with-ssl=#{Formula["openssl"].opt_prefix}
  ]
  args << "--with-libexpat=#{MacOS.sdk_path}/usr/include" unless MacOS::CLT.installed?

  system "./configure", *args
  inreplace "doc/example.conf", 'username: "unbound"', 'username: "@@HOMEBREW-UNBOUND-USER@@"'
  system "make", "install"
end
```

fnevesbr: Just did the xcode-select --install

And now 1.5.8 builds I'll test the 1.5.9 without the extrea dependency

DomT4: I think if we adopt the code block I proposed that'll get it working for everyone without making 'em install the CLT to get there.

fnevesbr: If this works for you apparently both versions works correctly. I mean the
ones build with the system and brew versions of expat.

fnevesbr: 1.5.9 without the expat dependency builds when you do have CLT installed.
Sorry but I could not replicate and test your patch as I've already
installed CLT.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/e2babc0f860aec6d257514ea5d50182d93824bfb. Thank you for your contribution to Homebrew @fnevesbr; we appreciate it! :smiley_cat:

PR #3081, opened 2016-07-18
pkgconfig: copy instead of hard coding the path to Homebrew's pc files.

xu-cheng:
cc @DomT4

### DIFF
--- a/Formula/pkg-config.rb
+++ b/Formula/pkg-config.rb
@@ -4,7 +4,7 @@ class PkgConfig < Formula
   url "https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.1.tar.gz"
   mirror "https://fossies.org/linux/misc/pkg-config-0.29.1.tar.gz"
   sha256 "beb43c9e064555469bd4390dcfd8030b1536e0aa103f08d7abf7ae8cac0cb001"
-  revision 1
+  revision 2
 
   bottle do
     sha256 "7354d2e9f337af437a22433926dd38a68413a8d8bef3aac52f203ae581d7669d" => :el_capitan
@@ -13,12 +13,14 @@ class PkgConfig < Formula
   end
 
   def install
+    mkdir_p lib/"pkgconfig"
+    cp Dir["#{HOMEBREW_LIBRARY}/Homebrew/os/mac/pkgconfig/#{MacOS.version}/*"], lib/"pkgconfig"
     pc_path = %W[
       #{HOMEBREW_PREFIX}/lib/pkgconfig
       #{HOMEBREW_PREFIX}/share/pkgconfig
       /usr/local/lib/pkgconfig
       /usr/lib/pkgconfig
-      #{HOMEBREW_LIBRARY}/Homebrew/os/mac/pkgconfig/#{MacOS.version}
+      #{lib}/pkgconfig
     ].uniq.join(File::PATH_SEPARATOR)
 
     ENV.append "LDFLAGS", "-framework Foundation -framework Cocoa"


DomT4:
I'm not sure this isn't just as problematic in some ways, sadly.

We do actually occasionally update the `pkg-config` files, particularly for newer systems, so if you installed Sierra at the start of developer preview & then didn't reinstall `pkg-config` for 6 months, which isn't an impossibility given the `pkg-config` release timetable, something like:

```bash
pkg-config --exists --print-errors "libxml-2.0 > 2.9.3"
```
Would fail, even though the version on that OS is now 2.9.4.

I think the current system works fairly sanely, except when we do mass refactoring, which is hopefully closer to finishing than beginning, at least when it comes to moving around `pkg-config` files 😕.

xu-cheng:
> We do actually occasionally update the pkg-config files, particularly for newer systems, so if you installed Sierra at the start of developer preview & then didn't reinstall pkg-config for 6 months,

We could or should do revision bump for this.

My thought is that lots pc files are internal files for Homebrew, therefore they shouldn't be used by pkgconfig directly.

DomT4:
> We could or should do revision bump for this.

The problem largely here is that `revision` bumps are incredibly blunt. You could revision bump for Sierra in a couple of months, for example, but that forces everyone on <10.12 to rebuild for nothing. There's no fine-grained control on that mechanism.

> My thought is that lots pc files are internal files for Homebrew, therefore they shouldn't be used by pkgconfig directly.

Theoretically, if we meant this completely in terms of not making our vendored `pkg-config` files available to users outside of Homebrew's `brew install` mechanism, we could cut the line out of the formula & rely completely on `superenv` & `stdenv` to insert the path, I suspect.

MikeMcQuaid:
What problem is this solving?

DomT4:
Embedding our internally vendored `.pc` files in the default search path. At this point, I think we've exposed them for so long people have come to rely on them & it'd be unfair to take them away by relying on std/superenv entirely.

I think ultimately the least worst system is the one we already use, presuming we're not going to make a habit out of moving around our vendored `pkg-config` files in future.

xu-cheng:
I was thinking to avoid mixing formula runtime files with Homebrew's internal files. But I don't have too strong preference. 

> we could cut the line out of the formula & rely completely on superenv & stdenv to insert the path, I suspect.

Noted this change has nothing to do superenv as we overwrite the search path inside superenv. As result, this only affects the usage of pkgconfig for normal user's project.

>  it'd be unfair to take them away by relying on std/superenv entirely.

That's why I send this PR as superenv/stdenv should be homebrew's internal mechanism. Therefore, it shouldn't be used by formula outside Homebrew.

UniqMartin:
I think the minor advantage of not exposing the Homebrew-internal path comes at the more severe cost of having to revision bump `pkg-config` every time any of those files change. It could be a net win if this made the bottle relocatable, but this doesn't seem to be the case.

This also doesn't solve a more fundamental problem, namely that `pkg-config` usually doesn't get updated with a major macOS upgrade, meaning it will continue to point at the `.pc` files of the previously installed OS until `pkg-config` is reinstalled.

DomT4:
> It could be a net win if this made the bottle relocatable, but this doesn't seem to be the case.

Yeah, it won't be because it actively remembers `HOMEBREW_PREFIX` in the path, which on CI always == `/usr/local`.

I think this causes more trouble than the existing system as-is sadly, although I appreciate the intention. I'm a 👎 on it unless we can hurdle the challenges laid out above.

PR #3093, opened 2016-07-18
dvm 0.6.3

carolynvs:

---

### DIFF
--- a/Formula/dvm.rb
+++ b/Formula/dvm.rb
@@ -1,8 +1,8 @@
 class Dvm < Formula
   desc "Docker Version Manager"
   homepage "https://github.com/getcarina/dvm"
-  url "https://github.com/getcarina/dvm/archive/0.6.2.tar.gz"
-  sha256 "57a14fc85c37d30c3c2265eb0f06d2fdc18273306052a5ccc748b54a5c38660c"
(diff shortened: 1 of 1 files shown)


Formula/dvm.rb
```diff
@@ -15,14 +15,16 @@ class Dvm < Formula
 
   def install
     ENV["GOPATH"] = buildpath
+    ENV["PATH"] = "#{ENV["PATH"]}:#{buildpath}/bin"
```
DomT4: Perhaps worth doing:
```ruby
ENV.append_path "PATH", buildpath/"bin"
```
here if it works, for the sake of looking a bit neater 🙈.

carolynvs: @DomT4 I'm not quite sure what that append_path function is doing. 😊  I tried replacing the entire line with the line above and while it executes without error, the current PATH isn't updated with the right value and the build fails.

DomT4:
Not quite sure why `ENV.append_path "PATH", buildpath/"bin"` wouldn't work, but if we wait for me to find the time to look into it properly it could be a while, so I'll take this as-is 😄. Thanks for checking nonetheless!

Merged in https://github.com/Homebrew/homebrew-core/commit/2961935039d5ce544fbdc3af08e9f67a80014817. Thank you for your contribution to Homebrew @carolynvs; we appreciate it! :smiley_cat:

PR #3101, opened 2016-07-18
libgpg-error 1.24

ilovezfs:


DomT4:
You'll need to revision bump `libgcrypt` as well. It "remembers" where `libgpg-error` lives.

ilovezfs:
um
```
==> brew install vpnc
==> FAILED
vpnc: TuntapRequirement unsatisfied!

You can install with Homebrew Cask:
  brew cask install tuntap
```

DomT4:
That's gunk. You can ignore that one.

ilovezfs:
>That's gunk. You can ignore that one.

would prefer to fix it ;)

DomT4:
It's only happening here because the `vpnc` formula still has a Mavericks bottle that predates our removing the `tuntap` formula. I'm not sure we've ever properly fixed this, it's just been pretty hidden.

ilovezfs:
ctunnel, fastd, iodine, openvpn, peervpn, shrewsoft-vpn-client, vpnc all have this issue. Should we just remove the mavericks bottles from their bottle blocks?

DomT4:
It's tempting. I think bottles effectively say "We've tested this on this platform, we know it works, we'll keep testing that it works". We can't make that promise now.

`openvpn` at least is only impacted purely on the Mavericks bottle. The Yosemite/El Cap ones don't use `:tuntap` anymore.

ilovezfs:
even though it will mean rebuilds for some people, I think maybe we should just bump the revision on all of them and let bot handle the bottle block fixups?

ilovezfs:
@DomT4 https://github.com/Homebrew/homebrew-core/pull/3103

DomT4:
I think it's worth moving ahead with this rather than waiting for that PR to be discussed out. Only need the `libgcrypt` revision bump here & you're as 👍 as you're going to get.

ilovezfs:
@DomT4 no need to bump libksba or gpgme?

DomT4:
Historically, nope.

DomT4:
Lol, okay, apparently only historically because nobody reported it to us. They do indeed embed the path in the config output 😒 🔫

ilovezfs:
hehe no worries :see_no_evil:

DomT4:
IMO, we could/should just be inreplacing a lot of these to point to `opt_prefix` rather than Cellar paths in the config executables.

DomT4:
I can't remember the last time a `libgpg-error` _actually_ broke something but we have to treat every update like it does currently because of that config output.

ilovezfs:
I can't say I'm opposed to regularly rebuilding security tools even if it feels unnecessary.

DomT4:
It runs too deep to be all that desirable. That config path gets passed down through a whole bunch of things, including usage outside of Homebrew. We've contributing to an endless cycle of pointless rebuilds because the configuration output takes the prefix too literally, heh.

ilovezfs:
mind checking dirmngr, gpa, npth, and pinentry too?

DomT4:
Promising start:

```
~> dirmngr
[1]    19716 segmentation fault  dirmngr
```
Rebuilding from source sorted that.

Don't have `gpa` installed but no reason for it to store the Cellar path IMO.  `pinentry` shows no direct references to the Cellar.

ilovezfs:
hehe libassuan does too. oopsies

ilovezfs:
@DomT4 looks like all we need to do is change the value of "prefix" in libgpg-error/1.23/bin/gpg-error-config from `/usr/local/Cellar/libgpg-error/1.23` to `/usr/local/opt/libgpg-error` and then rebuild the others, and it takes care of itself.

ilovezfs:
Same problem for libassuan and gpgme, though
```
gpgme/1.6.0/bin/gpgme-config:25:assuan_cflags="-I/usr/local/Cellar/libassuan/2.4.2/include -I/usr/local/opt/libgpg-error/include"
```

DomT4:
> looks like all we need to do is change the value of "prefix" in libgpg-error/1.23/bin/gpg-error-config from /usr/local/Cellar/libgpg-error/1.23 to /usr/local/opt/libgpg-error and then rebuild the others, and it takes care of itself.

This sounds reasonable 👍. Same with `gpgme`.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@DomT4 seems there need to be a few more revisions hehe. Welcome to library hell :fire:

ilovezfs:
@DomT4 Is the `mcabber` installation known to ever get stuck?

DomT4:
Not as far as I recall, but has been a little while since it last trundled through CI. Feel free to do it in post if this one goes green, but would be good to add a small notation to each `inreplace` to explain what we're doing & why.

ilovezfs:
@DomT4 I wonder if it's good ol' SDKROOT again.

ilovezfs:
>==> brew install mcabber

Though I think that just means installing the bottle.

DomT4:
Yeah. It shouldn't be building from source in this situation, unless there's either no bottle or you've changed the version/revision, neither of which are true here.

ilovezfs:
Ick:
```
./lib/mcabber/libbeep.so
./lib/mcabber/libfifo.so
./lib/mcabber/liburlregex.so
./lib/mcabber/libxttitle.so
```

DomT4:
Always nice to see a package well-ported to the OS it's operating on 🙈.

ilovezfs:
@DomT4 I think the `HOMEBREW_DEBUG=1` sword may need to be unsheathed for the next round.

DomT4:
Hurrah. It installs very quickly locally, so there's no reason for bottle pouring to take all year:
```
==> Summary
🍺  /usr/local/Cellar/mcabber/1.0.1: 411 files, 796.0K, built in 33 seconds
```

ilovezfs:
@DomT4 Yeah, I'm not sure what's going on. El Capitan seemed stuck too, but on `gpgme`.

ilovezfs:
@DomT4 (:green_apple: - `:mavericks`). The problem was coming from gpgme's `make check` which was getting stuck, probably because CI isn't a tty but not sure at this point.

DomT4:
Beside needing the `inreplace` notations which can be done in post, I think this is alright to 🚢.

ilovezfs:
@DomT4 :ship:'d

DomT4:
Nice, looks good locally 🎉

Issue #2260, opened 2016-06-21
Formulas of GUI apps that can be safely removed in favour of Casks.


vitorgalvao:
This post is a continuation of https://github.com/Homebrew/brew-evolution/pull/3#issuecomment-225597589.

An effort of deduplication between Homebrew (HB) and Homebrew-Cask (HBC) is being discussed on https://github.com/caskroom/homebrew-cask/issues/15603. It’s been agreed between both projects this is desirable, but [the rules of what should remain/be dropped are a bit harder](https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-223563011). In addition, for https://github.com/Homebrew/brew-evolution/pull/3 to continue, having those rules defined is starting to become a necessity.

This post is mostly to first gather with help form the community which formulas can be safely removed in favour of the casks. Please keep in mind we’re all aware [a deduplicaton effort has some small inconveniences](https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-225932164). Maintaining a piece of software in both repos managed separately with few differences because a few users prefer to install one way rather than the other is undesirable.

Here’s the preliminary list of apps that produce app bundles (homebrew maintainers, please feel free to edit the post directly to add/remove):

+ [apcupsd](https://github.com/Homebrew/homebrew-core/blob/master/Formula/apcupsd.rb)
+ [avidemux](https://github.com/Homebrew/homebrew-core/blob/master/Formula/avidemux.rb)
+ [cppcheck](https://github.com/Homebrew/homebrew-core/blob/master/Formula/cppcheck.rb)
+ [djview4](https://github.com/Homebrew/homebrew-core/blob/master/Formula/djview4.rb)
+ [emacs](https://github.com/Homebrew/homebrew-core/blob/master/Formula/emacs.rb)
+ [gammaray](https://github.com/Homebrew/homebrew-core/blob/master/Formula/gammaray.rb)
+ [graphviz](https://github.com/Homebrew/homebrew-core/blob/master/Formula/graphviz.rb)
+ [kdiff3](https://github.com/Homebrew/homebrew-core/blob/master/Formula/kdiff3.rb)
+ [macvim](https://github.com/Homebrew/homebrew-core/blob/master/Formula/macvim.rb)
+ [mpv](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mpv.rb)
+ [pinentry-mac](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pinentry-mac.rb)
[...]

PR #3103, opened 2016-07-18
revisions for tuntap removal

ilovezfs:
See discussion in https://github.com/Homebrew/homebrew-core/pull/3101


ilovezfs:
CC @AnastasiaSulyagina @DomT4 @MikeMcQuaid

MikeMcQuaid:
👍 but let's hold off merging this until we figured out why the migration didn't happen on these.

DomT4:
Migration didn't happen? I thought the point of this PR was largely that we can't claim to actually have tested these formulae using Mavericks/An updated `tuntap` now, so keeping bottles around was disingenuous, per the conversation in https://github.com/Homebrew/homebrew-core/pull/3101. Particularly since CI will keep attempting to pour previously built bottles but now has no way of testing them.

Am I missing a puzzle piece about the migration here?

ilovezfs:
@DomT4 well I'd think in general we'd want these revisions to be part of the PR moving something from homebrew/pqr caskroom/xyz, no?

DomT4:
Sure, in general, except we've already killed off `tuntap` so we'd be waiting forever here?

ilovezfs:
@DomT4 Right. I think @MikeMcQuaid just wants to clarify exactly what the procedure here is. For example, should this PR be recursive? It's missing anything that uses one of these formulae but didn't declare a tuntap dependency explicitly.

Also it should be mandatory to have CI support fulfillment of Cask dependencies and requirements that are satisfied by a default in Cask. For formulae where that's not actually possible at this time,  I think any of them that have a non-optional dependency on something in Cask that cannot be installed no longer belong in homebrew/core, just as we do not support any formulae with non-optional x11 requirements or non-optional cross-tap dependencies in homebrew/core.

So I'm not convinced this PR is even correct.  I think these formulae (as well as any that have a non-optional dependency on them recursively) need to be moved to the boneyard, to a homebrew/cask-dependent-limbo, or to caskroom/xyz tap unless their tuntap requirement can be reasonably made optional without compromising the usefulness of the software or until we find a way to install the tuntap cask during CI.

DomT4:
> Also it should be mandatory to have CI support fulfillment of Cask dependencies and requirements that are satisfied by a default in Cask. 

At least in `tuntap`'s case this requires root privileges to install, which is frankly dangerous to make programmatic on boxes which aren't disposable & are used for build after build after build. The user our CI runs on is deliberately privilege-limited to somewhat curtail that issue.

At least in `fuse` we've been using Travis to handle this problem, with the understanding whilst we can't build bottles for the formulae they can be completely built & tested by Travis CI. It also doesn't matter about executing `sudo` programatically there because if we accidentally trash the box, not a big deal due to how Travis is set up.

We could keep a `tuntap` permanently installed on the boxes like we do an up-to-date Java or X11, but Mike has previously made objection noises when people have suggested as much; I can understand where Mike is coming from on the objection.

> I think any of them that have a non-optional dependency on something in Cask that cannot be installed no longer belong in homebrew/core, just as we do not support any formulae with non-optional x11 requirements 

I somewhat agree, but if someone suggest a `homebrew/tuntap` tap & doesn't want to volunteer to maintain it forever themselves I'm going to throw them in a dungeon at this point. Our tap structure has mostly allowed maintainers to cherry pick the work they want to do, which is fine, but severely degrades UX.

ilovezfs:
>We could keep a tuntap permanently installed on the boxes like we do an up-to-date Java or X11, but Mike has previously made objection noises when people have suggested as much; I can understand where Mike is coming from on the objection.

I would pitch a rather unpleasant fit if we did that as it's a totally unacceptable idea unless it were a separate physical box or it was a vm, container, etc.

ilovezfs:
>I somewhat agree, but if someone suggest a homebrew/tuntap tap & doesn't want to volunteer to maintain it forever themselves I'm going to throw them in a dungeon at this point.

I agree. I think these all belong in the boneyard. If someone is able to get any or all of them admitted into a cask at some later date, that's lovely but not our concern.

DomT4:
> I think these all belong in the boneyard.

Well, `openvpn` doesn't actually use `:tuntap` on Yosemite or above, which soon will be all of Homebrew's bottled Mac versions 😉.

ilovezfs:
By these all, I mean whichever fit the criteria described above. For openvpn setting a minimum 10.10 requirement is fine.

ilovezfs:
@MikeMcQuaid @DomT4 PR refreshed.

ilovezfs:
@MikeMcQuaid reading over https://github.com/caskroom/homebrew-cask/issues/15603 and https://github.com/Homebrew/homebrew-core/pull/2075 it seems reverse dependencies have just not been discussed at all.

For instance, graphviz is currently slated for cask (according to the table), and has the following first-order reverse dependencies: bokken, eralchemy, fobis, ford, gammaray, gprof2dot, makefile2graph, plantuml, simgrid, xdot, homebrew/php/mondrian, homebrew/science/pnapi. But whether or not any/all/none of them are compatible with the Cask version has not entered the conversation.

DomT4:
@ilovezfs We've (Well, I've) already shot down `graphviz` becoming a Cask only in https://github.com/Homebrew/homebrew-core/issues/2260 for pretty much that reason. I'm sure it's perfectly possible to develop against the upstream `.pkg` but I think doing so for us is asking for chaos given we'd never know if something needed rebuilding for compatibility, etc.

ilovezfs:
@DomT4 I see. I suppose the table should be updated? It still says
```
graphviz.rb 	cask 	GUI
```

DomT4:
The discussion has become kind of fragmented so I'm not sure what is getting updated where at this point, but aye, ideally.

AnastasiaSulyagina:
I haven't seen #2260 till now. Will update

DomT4:
Thanks @AnastasiaSulyagina.

ilovezfs:
@MikeMcQuaid any thoughts on shipping this?

MikeMcQuaid:
@ilovezfs Seems fine. Ideally we'll figure out a way to autoinstall casks in these cases but I think we're probably a bit off that point now.

kim0:
Hi folks, 

As a user of some of the VPN apps that are being now moved into "boneyard" .. Can someone explain what is going on? Why this sudden move, and can I help get those VPN utils back?

Thanks

ilovezfs:
@kim0 you can `brew tap homebrew/boneyard`.

MikeMcQuaid:
@kim0 In short: we can no longer build tuntap from source and use it on modern versions of OS X so it doesn't make sense to keep around formulae that are unbuildable and untestable.

PR #3104, opened 2016-07-19
Remove preprocessor/compiler/linker flags from environment in test blocks

zmwangx:
- [ ] **Too many of them.** Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [ ] **Too many of them.** Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

This PR attempts to remove

``` ruby
flags = (ENV.cflags || "").split + (ENV.cppflags || "").split + (ENV.ldflags || "").split
```

from test blocks en masse, which is useless as discussed in https://github.com/Homebrew/homebrew-core/pull/2944#discussion_r70846681. [According to Mike](https://github.com/Homebrew/homebrew-core/pull/2944#discussion_r71087380)

> I don't think it is necessary; I think it was some sort of early "handle superenv if we ever port tests to that" or something. I don't think there's any need for it, really.

CC @apjanke @MikeMcQuaid.

### DIFF
--- a/Formula/atk.rb
+++ b/Formula/atk.rb
@@ -36,8 +36,7 @@ class Atk < Formula
     EOS
     gettext = Formula["gettext"]
     glib = Formula["glib"]
-    flags = (ENV.cflags || "").split + (ENV.cppflags || "").split + (ENV.ldflags || "").split
-    flags += %W[
+    flags = %W[
       -I#{gettext.opt_include}
       -I#{glib.opt_include}/glib-2.0
       -I#{glib.opt_lib}/glib-2.0/include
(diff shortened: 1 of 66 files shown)


ilovezfs:
>flags = (ENV.cflags || "").split + (ENV.cppflags || "").split + (ENV.ldflags || "").split

Wow, that is indeed grotesque.

apjanke:
👍 Thanks for doing the legwork on this.

I'm fine with merging this without running the build & tests, since this affects only `test do` blocks and the individual changes are small.

MikeMcQuaid:

PR #3124, opened 2016-07-19
imagemagick: build with fontconfig by default

jeroen:
I would like to suggest we build `imagemagick` with `fontconfig` by default. Currently it does not support any text/annotation/labeling features which is very limiting.

``` sh
curl -O https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/helloworld.svg
convert helloworld.svg out.png 
#>> convert: unable to read font `serif' @ warning/annotate.c/RenderType/927
```

I verified that all of these issues disappear when we build `--with-fontconfig` so perhaps this would be a better default for most users.

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -30,6 +30,8 @@ class Imagemagick < Formula
   option "without-magick-plus-plus", "disable build/install of Magick++"
   option "without-modules", "Disable support for dynamically loadable modules"
   option "without-threads", "Disable threads support"
+  option "without-freetype", "Disable TrueType support"
+  option "without-fontconfig", "Disable fontconfig support"
 
   depends_on "xz"
   depends_on "libtool" => :run
@@ -39,9 +41,9 @@ class Imagemagick < Formula
   depends_on "libpng" => :recommended
   depends_on "libtiff" => :recommended
   depends_on "freetype" => :recommended
+  depends_on "fontconfig" => :recommended
 
   depends_on :x11 => :optional
-  depends_on "fontconfig" => :optional
   depends_on "little-cms" => :optional
   depends_on "little-cms2" => :optional
   depends_on "libwmf" => :optional
@@ -68,6 +70,9 @@ class Imagemagick < Formula
       --enable-shared
       --enable-static
     ]
+
+    args << (build.without? "fontconfig") ? "--without-fontconfig" : "--with-fontconfig"
+    args << (build.without? "freetype") ? "--without-freetype" : "--with-freetype"
 
     if build.without? "modules"
       args << "--without-modules"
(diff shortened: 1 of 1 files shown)


zmwangx:
I have no particular objection to recommending `--with-fontconfig` (I always build with it after all), but your changes are problematic. You shouldn't touch

```ruby
    args << "--with-fontconfig=yes" if build.with? "fontconfig"
    args << "--with-freetype=yes" if build.with? "freetype"
```

and your adding 

```
      --with-fontconfig=yes
      --with-freetype=yes
```

just completely masks `--without-fontconfig` and `--without-freetype`.

jeroenooms:
I changed the script a bit. How do you suggest we make `--with-fontconfig` the default leaving the `--with-fontconfig` option untouched?

Formula/imagemagick.rb
```diff
@@ -69,6 +71,9 @@ def install
       --enable-static
     ]
 
+    args << (build.without? "fontconfig") ? "--without-fontconfig" : "--with-fontconfig"
+    args << (build.without? "freetype") ? "--without-freetype" : "--with-freetype"
+
     if build.without? "modules"
```
zmwangx: Remove these three lines.

Formula/imagemagick.rb
```diff
@@ -107,8 +112,6 @@ def install
     args << "--with-quantum-depth=#{quantum_depth}" if quantum_depth
     args << "--with-rsvg" if build.with? "librsvg"
     args << "--without-x" if build.without? "x11"
-    args << "--with-fontconfig=yes" if build.with? "fontconfig"
-    args << "--with-freetype=yes" if build.with? "freetype"
```
zmwangx: Restore these two lines.

Formula/imagemagick.rb
jeroenooms: But then it won't be enabled by default. I tried building with the default configure options on a machine that has fontconfig and it won't build with fontconfig support.

zmwangx: That's why I ask you to restore `args << "--with-fontconfig=yes" if build.with? "fontconfig"`.

jeroenooms: I don't get it. This way it will only work if the user installs with `brew install imagemagick --with-fontconfig`?

zmwangx: `depends_on "fontconfig" => :recommended` implies `build.with? "fontconfig"`. Just make my recommended changes locally and see for yourself.

jeroenooms: OK I was not aware of that. Let me fix it. I accidentally deleted the branch from my repository which closed the pull request :(

PR #3129, opened 2016-07-19
imagemagick: enable fontconfig by default

jeroen:
I would like to suggest we build `imagemagick` with `fontconfig` by default. Currently it does not support any text/annotation/labeling features which is very limiting.

``` sh
curl -O https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/helloworld.svg
convert helloworld.svg out.png 
#>> convert: unable to read font `serif' @ warning/annotate.c/RenderType/927
```

I verified that all of these issues disappear when we build `--with-fontconfig` so perhaps this would be a better default for most users.

FYI there was some constructive comments by @zmwangx in https://github.com/Homebrew/homebrew-core/pull/3124 which I accidentally closed. Let's blame the heat.

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -39,9 +39,9 @@ class Imagemagick < Formula
   depends_on "libpng" => :recommended
   depends_on "libtiff" => :recommended
   depends_on "freetype" => :recommended
+  depends_on "fontconfig" => :recommended
 
   depends_on :x11 => :optional
-  depends_on "fontconfig" => :optional
   depends_on "little-cms" => :optional
   depends_on "little-cms2" => :optional
   depends_on "libwmf" => :optional


DomT4:
> Currently it does not support any text/annotation/labeling features which is very limiting.

To be honest, I'm a little unconvinced people are bothered looking at our analytics for this formula. Only 431 installations, or 0.39% of the total "known" installations, bothered to install `--with-fontconfig`. That doesn't smell like enormous demand, and `fontconfig` is quite a hefty dependency given the annoyingly slow `post_install` step necessary.

I'll leave this open for some debate though, I can be convinced either way personally, but we are trying to make decisions like this based on hard data these days.

jeroenooms:
Users might not be aware that the reason features are not working is due to missing fontconfig. It took me a while to figure this out myself: Initially I assumed imagemagick was just buggy. Hence the number of `--with-fontconfig` installations is perhaps not the best metric. 

I fully agree to keep things light were possible and minimize installation time but most people many people that install `imagemagick` might already have `fontconfig`?

I can see both sides, but having used `imagemagick` pretty intensely for the past few weeks I found that a lot of stuff doesn't work well without access to fonts. What makes it worse is that this is very difficult to debug for the user. At best there is an error message about missing fonts (not mentioning `fontconfig` at all). But in other cases there is no error and you just get an image with no text or super ugly text. I suspect most users will then conclude imagemagick just sucks.

MikeMcQuaid:
> To be honest, I'm a little unconvinced people are bothered looking at our analytics for this formula. Only 431 installations, or 0.39% of the total "known" installations, bothered to install --with-fontconfig. That doesn't smell like enormous demand, and fontconfig is quite a hefty dependency given the annoyingly slow post_install step necessary.
> 
> I'll leave this open for some debate though, I can be convinced either way personally, but we are trying to make decisions like this based on hard data these days.

I agree, that's enough for a reason for me to pass on this PR (sorry @jeroenooms!).

zmwangx:
@DomT4 

> To be honest, I'm a little unconvinced people are bothered looking at our analytics for this formula. 

I believe GA data is not available to anyone other than Mike and maintainers who have requested access (per Analytics.md), right?

DomT4:
Correct. We're trying to work out a way to offer people a data dump, but GA doesn't make it particularly easy to export en mass without me manually clicking through hundreds of pages of data, exporting it into unique files & then squashing the files together, at least in so far as I've discovered to date.

DomT4:
> To be honest, I'm a little unconvinced people are bothered looking at our analytics for this formula.

I possibly worded this unclearly. By this I meant _"Looking at our analytics data for `imagemagick`, I'm unconvinced people are bothered by missing the features offered by building in `fontconfig` support"._

DomT4:
[Imagemagick-Export.pdf](https://github.com/Homebrew/homebrew-core/files/374896/Analytics.All.Mobile.App.Data.Top.Events.20160619-20160719.pdf)

That's all the "known" `imagemagick` installations, if people are curious. The level of data below that (not exported automatically) is split into install & build failures, each of which is split into numbers per OS X version. That's where the data on `imagemagick` ends.

ilovezfs:
> fontconfig is quite a hefty dependency given the annoyingly slow post_install step necessary.

Is that fixable?

jeroenooms:
Are there stats on how many people that install `imagemagick` do not have `fontconfig` yet?

ilovezfs:
Also, it looks like upstream distributes binaries. Do their binaries have the fontconfig option on or off?

zmwangx:
@ilovezfs

> Also, it looks like upstream distributes binaries. Do their binaries have the fontconfig option on or off?

That's a good question.

```sh
$ wget http://www.imagemagick.org/download/binaries/ImageMagick-x86_64-apple-darwin15.4.0.tar.gz
$ tar xf ImageMagick-x86_64-apple-darwin15.4.0.tar.gz
$ DYLD_LIBRARY_PATH=$PWD/ImageMagick-7.0.1/lib ImageMagick-7.0.1/bin/convert -version
Version: ImageMagick 7.0.1-4 Q16 x86_64 2016-05-15 http://www.imagemagick.org
Copyright: Copyright (C) 1999-2016 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Features: Cipher DPC HDRI
Delegates (built-in): bzlib freetype jng jpeg lcms png
```

So no, no fontconfig out of box.

For reference, here's the output of `convert -version` with my IM which was brewed with `--with-fontconfig --with-librsvg --with-webp`:

```sh
$ convert -version
Version: ImageMagick 6.9.5-2 Q16 x86_64 2016-07-13 http://www.imagemagick.org
Copyright: Copyright (C) 1999-2016 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Features: Cipher DPC Modules
Delegates (built-in): bzlib cairo fontconfig freetype jng jpeg ltdl lzma png rsvg tiff webp xml zlib
```

Also, 

> > fontconfig is quite a hefty dependency given the annoyingly slow post_install step necessary.
>
> Is that fixable?

If you use X, you know how annoyingly slow `fc-cache` can be, and I don't think there's a way around it, unless you want to defer `fc-cache` to the first time the user launches something that uses fc, which means whatever the user wants to use will sit idly for up to a few minutes...

ilovezfs:
>having used imagemagick pretty intensely for the past few weeks I found that a lot of stuff doesn't work well without access to fonts. What makes it worse is that this is very difficult to debug for the user. At best there is an error message about missing fonts (not mentioning fontconfig at all). But in other cases there is no error and you just get an image with no text or super ugly text. I suspect most users will then conclude imagemagick just sucks.

OK, so this remark is probably predictable, but I'm finding it hard to grok why the binaries upstream distributes would lack this feature if the software is as broken as described without it.

DomT4:
> If you use X, you know how annoyingly slow fc-cache can be, and I don't think there's a way around it, unless you want to defer fc-cache to the first time the user launches something that uses fc, which means whatever the user wants to use will sit idly for up to a few minutes...

Yeah. This is more or less what I was going to say in more detail. You only defer it till the user runs something, which can lead to mysterious hangs. I'd rather the mysterious hang came from us & was announced sadly, even if it is frustrating 😢.

zmwangx:
> OK, so this remark is probably predictable, but I'm finding it hard to grok why the binaries upstream distributes would lack this feature if the software is as broken as described without it.

I guess most people don't use IM to deal with text (I occasionally do, by the way). Most of the things you can do with IM don't involve text at all. The canonical tutorial/reference for IMv6 is https://www.imagemagick.org/Usage/, and you can evaluate for yourself the percentage of examples that would require fontconfig.

zmwangx:
Forgot to say we already recommend `freetype` which should cover some of the font stuff. Not sure how much it covers though.

DomT4:
> Are there stats on how many people that install imagemagick do not have fontconfig yet?

No. That would require knowing more about users' complete installations than I'm comfortable with.

ilovezfs:
Users with any of these https://gist.github.com/ilovezfs/6ef69e88c152ea5ec23b95eb035986d3 already have fontconfig if that helps.

DomT4:
Not necessarily the `--include-build` ones, unless the bottle isn't `cellar :any`, and even then that'd presume users won't bother to remove `:build` dependencies later 😄.

ilovezfs:
Yes, I know. Here's the relevance of the imprecision:
```
Josephs-MacBook-Pro:Formula joe$ diff AA BB
47d46
< gexiv2
54d52
< glib-networking
56d53
< gnome-icon-theme
70d66
< gsettings-desktop-schemas
166d161
< homebrew/science/insighttoolkit
255d249
< megatools
288d281
< radare2
```

DomT4:
My comment wasn't necessarily for your benefit 😉

MikeMcQuaid:
Passing on this, sorry!

PR #3131, opened 2016-07-19
reposurgeon 3.37

ilovezfs:
remove upstreamed `inreplace`.


ilovezfs:
@DomT4 :ship:?

DomT4:
If you feel like it:
```ruby
(share/"emacs/site-lisp/reposurgeon").install "reposurgeon-mode.el"
```
Can become:
```ruby
elisp.install "reposurgeon-mode.el"
```
But can be done in post. One is simply shorthand for the other, albeit shorthand we've done kind of a bad job enforcing.

👍 though.

PR #3134, opened 2016-07-19
python & python3: fix sqlite3 on sierra

mateusdelbianco:

---

Steps to reproduce: On a MacOS sierra system without this patch, open the `python` or `python3` binary compiled by homebrew and run `import _sqlite3`.

Expected output (no errors):

```
$ python
Python 2.7.12 (default, Jul 19 2016, 15:02:58)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.33.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import _sqlite3
>>>

$ python3
Python 3.5.2 (default, Jul 19 2016, 15:09:52)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.33.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import _sqlite3
>>>
```

### DIFF
--- a/Formula/python.rb
+++ b/Formula/python.rb
@@ -140,7 +140,7 @@ class Python < Formula
 
     # Allow sqlite3 module to load extensions:
     # https://docs.python.org/library/sqlite3.html#f1
-    if build.with? "sqlite"
+    if MacOS.version < :sierra && build.with?("sqlite")
       inreplace("setup.py", 'sqlite_defines.append(("SQLITE_OMIT_LOAD_EXTENSION", "1"))', "")
     end
 
--- a/Formula/python3.rb
+++ b/Formula/python3.rb
@@ -150,7 +150,9 @@ class Python3 < Formula
     end
 
     # Allow sqlite3 module to load extensions: https://docs.python.org/library/sqlite3.html#f1
-    inreplace("setup.py", 'sqlite_defines.append(("SQLITE_OMIT_LOAD_EXTENSION", "1"))', "pass") if build.with? "sqlite"
+    if MacOS.version < :sierra && build.with?("sqlite")
+      inreplace("setup.py", 'sqlite_defines.append(("SQLITE_OMIT_LOAD_EXTENSION", "1"))', "pass")
+    end
 
     # Allow python modules to use ctypes.find_library to find homebrew's stuff
     # even if homebrew is not a /usr/local/lib. Try this with:


DomT4:
Will leave this for Tim to sanity review. I'm still unable to reproduce this on my 10.12 box personally, with either `python` or `python3` 😕.

mateusdelbianco:
It's weird, but I tried compiling python/python3 on another 10.12 box and it worked. I also recompiled everything (xz, pkg-config, openssl, readline, sqlite, gdbm, python, python3) on my first box, without this patch, and it failed again.

The only difference is that that I'm using the new Xcode 8 beta 3 (clang-800.0.33.1) on the failed box, and Xcode 8 beta 2 (clang-800.0.31) on the working one.

I'll keep investigating: I'll install Xcode 8 beta 3 on the box that's working, recompile everything again and see if it breaks.

tdsmith:
I suspect the root cause is that the wrong libsqlite dylib is being located; I'd like to make sure we understand that first.

mateusdelbianco:
It's definitely not the Xcode version. Right now I have two boxes, one that only compiles python with sqlite support using the patch, and another that doesn't need the patch. Any ideas what I can do to help figure out why?

notDavid:
I'm not sure if this is relevant, just posting fyi; on my 10.12 Beta with Xcode 8.0 beta 2 it fails also:

```
root$ find / -iname '_sqlite3.so' 2>/dev/null
/System/Library/Frameworks/Python.framework/Versions/2.6/lib/python2.6/lib-dynload/_sqlite3.so
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_sqlite3.so
```

shouldn't there be an _sqlite3.so somewhere in `
'/usr/local/Cellar/python3/3.5.2_1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/lib-dynload/` ?

(in my case there isn't one.)

notDavid:
fyi, the problem is solved when i change this line https://github.com/Homebrew/homebrew-core/blob/master/Formula/python3.rb#L133 to :

```
ldflags  = ["-L/usr/local/opt/sqlite/lib "]
cppflags = ["-I/usr/local/opt/sqlite/include "]
```

tdsmith:
Please open a new issue if this problem persists; this is probably not how we'll choose to solve it.

XxUnkn0wnxX:
will this be merged or not as I am having the same issue: ImportError: No module named _sqlite3

tdsmith:
This will not be merged; I can't reproduce.

XxUnkn0wnxX:
well I run the script: nzbToSickBeard.py

from https://github.com/clinton-hall/nzbToMedia

after I have reinstalled python & sqlite via brew command several times already and keeps giving me same error. so your say I can never have sickrage automated scripts running ever again?

XxUnkn0wnxX:
python fails to build sqlite:

*** WARNING: renaming "_sqlite3" since importing it failed: dlopen(build/lib.macosx-10.12-x86_64-2.7/_sqlite3.so, 2): Symbol not found: _sqlite3_enable_load_extension
  Referenced from: build/lib.macosx-10.12-x86_64-2.7/_sqlite3.so
  Expected in: flat namespace
 in build/lib.macosx-10.12-x86_64-2.7/_sqlite3.so

Python build finished, but the necessary bits to build these modules were not found:
bsddb185           dl                 imageop         
linuxaudiodev      ossaudiodev        spwd            
sunaudiodev                                           
To find the necessary bits, look in setup.py in detect_modules() for the module's name.


Failed to build these modules:
_sqlite3

XxUnkn0wnxX:
Fixed all issues solution was very simple

XxUnkn0wnxX:
1st remove sqlite, python & python3 if installed via homebrew

brew uninstall --force python
brew uninstall --force python3
brew uninstall --force sqlite

this removes all copies

then make sure Xcode 8 or later is installed

then re install command line tools via xcode-select --install then sudo xcode-select --reset

then finally install command line tools via pkg file found here: https://developer.apple.com/download/more/

look for command line tools in search

I downloaded "Command Line Tools (macOS 10.12) for Xcode 8.dmg"

the open DMG and install the pkg found there.

now install sqlite 1st then python & python 3 if needed

do brew install python --verbose and make sure there are no warnings, if none issues should be resolved.

mateusdelbianco:
@XxUnkn0wnxX yes, installing CLI works.

@tdsmith wonder if it should be more clear that we need CLI for python and other packages now...

ilovezfs:
I find it dubious that CLT-made-it-work implies CLT-is-mandatory.

XxUnkn0wnxX:
@mateusdelbianco I had it installed since El Capitan but since updating it didn't find the tools. without a full reinstall of all them + reset.

vinniejames:
Thanks @XxUnkn0wnxX, fixed it for me too

tdsmith:
This was fixed by #5399; sorry for the delay.

PR #3139, opened 2016-07-19
wxmac: fix custom print paper size bug

sarim:
Fix Custom Print Paper Size bug. Upstream #16959

Patch from upstream wxWidgets bug tracker.
http://trac.wxwidgets.org/ticket/16959

### DIFF
--- a/Formula/wxmac.rb
+++ b/Formula/wxmac.rb
@@ -9,6 +9,13 @@ class Wxmac < Formula
     url "https://github.com/wxWidgets/wxWidgets/releases/download/v3.0.2/wxWidgets-3.0.2.tar.bz2"
     sha256 "346879dc554f3ab8d6da2704f651ecb504a22e9d31c17ef5449b129ed711585d"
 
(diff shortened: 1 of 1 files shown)


DomT4:
Does it work if we use the actual upstream patch rather than inlining it here? We're trying to gently phase out inline patches beyond a certain length because people keep accidentally breaking 'em.

sarim:
Umm, What other way is there to add patches other than inline?

DomT4:
```ruby
patch do
  url "blah"
  sha256 "blah"
end
```

sarim:
@DomT4 did that. check now.

sarim:
At last all checks passed :)

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/dbc17c1f6eeb6ab185154acdd61653a3db3bf113. Thank you for your contribution to Homebrew @sarim; we appreciate it! :smiley_cat:

PR #3154, opened 2016-07-20
compcert 2.7.1

ilovezfs:
Coq 8.5 support

### DIFF
--- a/Formula/compcert.rb
+++ b/Formula/compcert.rb
@@ -1,20 +1,8 @@
-class Camlp5TransitionalModeRequirement < Requirement
-  fatal true
-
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 Finally!

DomT4:
Nice! So much nicer a formula to look at this way 😄. LGTM, Thanks 🙇.

ilovezfs:
@DomT4 So what you're saying is, you don't think every formula should vendor its own compiler? :confused:

:)

DomT4:
> So what you're saying is, you don't think every formula should vendor its own compiler?

I am very, very much saying this, yes 😉

PR #3158, opened 2016-07-20
elixirscript 0.21.0

ilovezfs:
- set MIX_ENV to prod per upstream build instructions
- use mix to install build-time requirement rebar in the buildpath
- run "mix test" and "npm test"
- install the docs
- use the full path in the test
- clean up the test


DomT4:
👍

PR #3166, opened 2016-07-20
grip 4.3.0 (new formula)

zmwangx:

---

`grip` is a Markdown previewer that uses the GitHub Markdown API to render Markdown files exactly as they would be rendered on GitHub. I've been using it for years and absolutely love it. Being a command line utility that has little to do with Python functionality-wise, it would be nice to decouple it from Python installations (which are constantly in flux for some people, me included) and manage it with brew.

The only thing I'm not sure about here is `desc`. `grip` markets itself as a GitHub Readme previewer, but
1. It can be used with any Markdown (GFM) file;
2. It can't be used with other markup languages that are allowed for README on GitHub, e.g., reST.

Therefore, I took the liberty to give it an IMO more fitting description, "GitHub Markdown previewer".

### DIFF
--- a/Formula/grip.rb
+++ b/Formula/grip.rb
@@ -1,0 +1,83 @@
+class Grip < Formula
+  desc "GitHub Markdown previewer"
+  homepage "https://github.com/joeyespo/grip"
+  url "https://github.com/joeyespo/grip/archive/v4.3.2.tar.gz"
+  sha256 "c39c91764e1674718e185f7dec12b70347ca5ba845eeb67e996a9e9fd8710bbe"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  resource "click" do
+    url "https://files.pythonhosted.org/packages/7a/00/c14926d8232b36b08218067bcd5853caefb4737cda3f0a47437151344792/click-6.6.tar.gz"
+    sha256 "cc6a19da8ebff6e7074f731447ef7e112bd23adf3de5c597cf9989f2fd8defe9"
+  end
+
+  resource "docopt" do
+    url "https://files.pythonhosted.org/packages/a2/55/8f8cab2afd404cf578136ef2cc5dfb50baa1761b68c9da1fb1e4eed343c9/docopt-0.6.2.tar.gz"
+    sha256 "49b3a825280bd66b3aa83585ef59c4a8c82f2c8a522dbe754a8bc8d08c85c491"
+  end
+
+  resource "Flask" do
+    url "https://files.pythonhosted.org/packages/55/8a/78e165d30f0c8bb5d57c429a30ee5749825ed461ad6c959688872643ffb3/Flask-0.11.1.tar.gz"
+    sha256 "b4713f2bfb9ebc2966b8a49903ae0d3984781d5c878591cf2f7b484d28756b0e"
+  end
+
+  resource "itsdangerous" do
+    url "https://files.pythonhosted.org/packages/dc/b4/a60bcdba945c00f6d608d8975131ab3f25b22f2bcfe1dab221165194b2d4/itsdangerous-0.24.tar.gz"
+    sha256 "cbb3fcf8d3e33df861709ecaf89d9e6629cff0a217bc2848f1b41cd30d360519"
+  end
+
+  resource "Jinja2" do
+    url "https://files.pythonhosted.org/packages/f2/2f/0b98b06a345a761bec91a079ccae392d282690c2d8272e708f4d10829e22/Jinja2-2.8.tar.gz"
+    sha256 "bc1ff2ff88dbfacefde4ddde471d1417d3b304e8df103a7a9437d47269201bf4"
+  end
+
+  resource "Markdown" do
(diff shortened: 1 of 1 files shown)


zmwangx:
Installation on the CI [was failing](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5143/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_grip/) due to setuptools oddities: `grip`'s `setup.py` has [this line](https://github.com/joeyespo/grip/blob/919eabd3bf74f1ac56808d070362eddc61aa7f3e/setup.py#L45)

```python
package_data={'grip': ['static/*', 'templates/*', 'vendor/*']},
```

where the glob `static/*` includes a directory `static/octicon` which older setuptools can't handle.  See https://github.com/pypa/setuptools/issues/261. Newer setuptools automatically ignores such directories. As OS X's system Python ships with very old setuptools — e.g., 1.1.6 on OS X 10.11.6 (`cat /System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/setuptools/version.py`), the installation fails where you typically won't experience with an up-to-date pip which comes with up-to-date setuptools as well.

My solution to this is to use brewed Python 3 (although `grip` is supported on Python 2.6 and 2.7). Brewed Python 3 comes with fairly recent setuptools, and Python 3 is the future (and present) anyway.

DomT4:
> My solution to this is to use brewed Python 3 (although grip is supported on Python 2.6 and 2.7). Brewed Python 3 comes with fairly recent setuptools, and Python 3 is the future (and present) anyway.

You can always vendor `setuptools` in the formula whilst using the system Python 2.7. I don't have a problem using `python3` here, but that would be the solution using the vanilla Python.

Formula/grip.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  resource "Werkzeug" do
+    url "https://files.pythonhosted.org/packages/b7/7f/44d3cfe5a12ba002b253f6985a4477edfa66da53787a2a838a40f6415263/Werkzeug-0.11.10.tar.gz"
+    sha256 "cc64dafbacc716cdd42503cf6c44cb5a35576443d82f29f6829e5c49264aeeee"
+  end
+
+  def install
+    xy = Language::Python.major_minor_version "python3"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python#{xy}/site-packages"
+    %w[click docopt Flask itsdangerous Jinja2 Markdown MarkupSafe path-and-address Pygments requests Werkzeug].each do |r|
```
DomT4: Unless we're actually manipulating the array I can live without one.

Formula/grip.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python#{xy}/site-packages"
+    system "python3", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    assert_match "Homebrew", shell_output("#{bin}/grip #{HOMEBREW_PREFIX}/README.md --export -")
```
DomT4: Please don't assume `HOMEBREW_PREFIX/README.md` will always exist. There are discussions ongoing in regards to it potentially moving. Would prefer if you wrote a little markdown to the `testpath` directly.

Formula/grip.rb
zmwangx: @DomT4 You mean just `resources.each`? That will do. (This part was auto-generated by homebrew-pypi-poet (`poet -s grip`) and I didn't look into it 😂 )

DomT4: Yeah. Tim & I don't necessarily see perfectly eye to eye on array usage & he wrote the poet tool 😄.

DomT4: (Not a diss of Tim that, I add. Just a difference in styles & opinions on formulae readability)

zmwangx:
> You can always vendor setuptools in the formula whilst using the system Python 2.7.

Right. I'll vendor setuptools if anyone objects to depending on python3 which is quite hefty. PYTHONPATH should take care of finding the new version, but it will require a little bit of testing.

zmwangx:
@DomT4 Addressed resources and test.

DomT4:
I'm happy enough here, but will give it till later today to let people be excited about one Python or the other if they wish to.

zmwangx:
Bumped to 4.3.2.

zmwangx:
grip 4.3.1 restored compatibility with older setuptools, so I switched back to Python 2.7. (Apparently Debian users complained because their setuptools is also outdated and worse still don't have much of an option to get up-to-date versions).

DomT4:
Works for me. At some point in some distant future we'll have to shunt as much over to `python3` as possible but I understand from the upstream timetable there's no massive rush at this point.

Merged in https://github.com/Homebrew/homebrew-core/commit/7960a9bfc92eacd1f67e8d78893d3594500652b5. Thanks again @zmwangx! 🙇

PR #3183, opened 2016-07-21
wireguard-tools 0.0.20160722 (new formula)

zx2c4:

---

Hello,

This is a formula for the `wg(8)` tool, for [WireGuard](https://www.wireguard.io/). I'm the author of `pass`, which is already in here as `pass.rb`, so I'm somewhat familiar with brew maintenance. As part of the effort to make userspace implementations of WireGuard for OS X, getting the wireguard-tools package running smoothly on Homebrew is quite important. I expect for this package to become popular overtime. It has packages currently for Debian, Arch, OpenWRT, Gentoo, and OpenSUSE, with more on their way. The package ships with a man page as well, accessible at `man wg`.

Here's a screenshot after a successful brew installation:
wireguard-osx

Looking forward,
Jason

### DIFF
--- a/Formula/wireguard-tools.rb
+++ b/Formula/wireguard-tools.rb
@@ -1,0 +1,17 @@
+class WireguardTools < Formula
+  desc "Tools for the WireGuard secure network tunnel"
+  homepage "https://www.wireguard.io/"
+  # Despite the experimental tag the tools themselves are stable.
+  # Please only update version when the tools have been modified/updated.
+  url "https://git.zx2c4.com/WireGuard/snapshot/WireGuard-experimental-0.0.20160722.tar.xz"
+  sha256 "0dcda97b6bb4e962f731a863df9b4291c1c453b01f4faba78be4aaa13a594242"
+  head "https://git.zx2c4.com/WireGuard", :using => :git
+
+  def install
+    system "make", "PREFIX=#{prefix}", "-C", "src/tools", "install"
+  end
+
+  test do
+    system "#{bin}/wg", "help"
+  end
+end


DomT4:
Hey @zx2c4. If there's no stable release for this at all at the present time this will be a better fit for our https://github.com/Homebrew/homebrew-devel-only tap. We _try_ not to package purely experimental software in `homebrew/core`.

Besides that, this looks 👍. If you want to drop a PR to https://github.com/Homebrew/homebrew-devel-only happy to re-review & likely merge over there. Thanks!

zx2c4:
Hi @DomT4 -- Sorry for the confusion in version numbering, but in fact wireguard-tools is quite stable. It simply lives in the same repository and has the same snapshot release cycles of the currently experimental module for the Linux kernel. But the tools component, which is being packaged here, is very stable. Upstream could separate it into its own repository, but it's easiest just to keep the releases synchronized, at least for now.

So, this package _is_ a good candidate for `homebrew/core`, especially as various implementations start to rely on it. And even if nothing else in `homebrew/core` marks this as a dependency, as a sysadmin tool subcomands like `wg pubkey` are very useful.

So, please reconsider!

Thanks,
Jason

DomT4:
Thanks, that's useful information. You did name the formula `wireguard-tools` so fair shout on me having missed that hint.

Is there some upstream confirmation they consider the tools stable? It'd be useful if they would cut a stable release for the tools, but I understand the lower maintenance burden around not doing so.

Maybe it's worth adding some notation to the formula that we don't have to update this _incredibly_ regularly despite the `experimental` tag because the tools themselves are stable.

zx2c4:
Hey @DomT4 -- I've just pushed a new package that adds a descriptive comment. Fortunately, in this case, I actually am upstream. I'll only be submitting pull requests to bump the version when it's relevant, and I'll include the tool-related changelog in the pull request messages.

DomT4:
> I'll only be submitting pull requests to bump the version when it's relevant, and I'll include the tool-related changelog in the pull request messages.

Sadly I can't promise you you'll be the only one updating it. We don't really "do" the concept of individual formula maintainers, although there are a number of formula that require enough delicate dancing inevitably only a few people feel passionate enough to do so. That's more or less why the notation is a desirable thing.

Formula/wireguard-tools.rb
```diff
@@ -0,0 +1,19 @@
+class WireguardTools < Formula
+  desc "Tools for the WireGuard secure network tunnel"
+  homepage "https://www.wireguard.io/"
+  # Though this tools package is from an experimental tag, these tools themselves are
```
DomT4: At the risk of being annoying _(Which I specialise in)_, can we trim this a little bit? Something like:
```ruby
# Despite the experimental tag the tools themselves are stable.
# Please only update version when the tools have been modified/updated.
```

zx2c4:
Hi @DomT4 -- I trimmed the line note and pushed a new revision. I think we should be all set now!

apjanke:
LGTM. 👍

DomT4:
Yeah, this is fine. I've just been drowned by life lately & hadn't had time to pull yet.

Merged in https://github.com/Homebrew/homebrew-core/commit/38e4e0bfd2d1f12131afc0d397354df130599f23. Thank you for your contribution to Homebrew @zx2c4; we appreciate it! :smiley_cat:

Issue #3168, opened 2016-07-20
Agda failed to build on 10.10


madgen:
- [x] Ran `brew update` and retried your prior step?

Also tried upgrading everything.
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

Yup, it reports no issues.
### Bug report:

Agda fails to install from source on Yosemite (10.10) ie. `brew install -s agda`. It succeeds using bottles.

Note that it also fails if I do "--without-stdlib".

`brew gist-logs agda`

https://gist.github.com/anonymous/a0e86b5ea77f0bda8ee0059e61ead793

When I run it for a second time (right after it failed without changing anything else), it fails with a different error. Here are the logs for that:

https://gist.github.com/anonymous/94f3376530e8e2631ff354c10d4a7d2c

ilovezfs:
Neat. I'll take a look.

ilovezfs:
@madgen  The error is
```
[20 of 24] Compiling Language.Haskell.Exts.InternalParser ( dist/dist-sandbox-a9dc4f2d/build/Language/Haskell/Exts/InternalParser.hs, dist/dist-sandbox-a9dc4f2d/build/Language/Haskell/Exts/InternalParser.o )
dist/dist-sandbox-a9dc4f2d/build/Language/Haskell/Exts/InternalParser.o: getModificationTime: does not exist (No such file or directory)
cabal: Leaving directory '/var/folders/k3/2mjf5njs0_36dhrx60rzqcfr0000gn/T/cabal-tmp-21283/haskell-src-exts-1.17.1'
Failed to install haskell-src-exts-1.17.1
cabal: Error: some packages failed to install:
haskell-src-exts-1.17.1 failed during the building phase. The exception was:
ExitFailure 1
```

Since I cannot reproduce build failure for haskell-src-exts, I think this may be a transient timing issue. Do you mind trying the build again?

ilovezfs:
@madgen actually that second build was successful. Your emacs is just too old.
```
brew install emacs
```
before doing the agda installation.

ilovezfs:
@DomT4 obviously this one should have a minimum emacs requirement as discussed here https://github.com/agda/agda/issues/1945#issuecomment-212995311

DomT4:
@ilovezfs If you can find out roughly which version it needs, feel free to tack that into the formula 👍.

madgen:
[...]

PR #3193, opened 2016-07-21
agda: raise emacs requirement to 23.4

ilovezfs:
Fixes "agda-mode compile" so that it doesn't error out with
  Symbol's function definition is void: byte-compile-disable-warning
  Unable to compile the following Emacs Lisp files:
    etc.

See
  https://github.com/Homebrew/homebrew-core/issues/3168
  https://github.com/agda/agda/issues/1945
  https://github.com/Homebrew/legacy-homebrew/pull/45327

### DIFF
--- a/Formula/agda.rb
+++ b/Formula/agda.rb
@@ -44,7 +44,7 @@ class Agda < Formula
   end
 
   depends_on "gmp"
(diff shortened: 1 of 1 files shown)


ilovezfs:
CI is pointless here since bottling will just use brew emacs before and after this change.

PR #3194, opened 2016-07-21
agda: remove gmp dependency

ilovezfs:
See Homebrew/legacy-homebrew#39134 and Homebrew/legacy-homebrew#39709.


ilovezfs:
CC @mietek @zmwangx

MikeMcQuaid:
👍 as it's green

PR #3200, opened 2016-07-21
libopenmpt 0.2.6611-beta18 (new formula)

jdek:


DomT4:
Should live in https://github.com/Homebrew/homebrew-devel-only until there's a stable release.

PR #3217, opened 2016-07-22
groovyserv 1.1.0

ilovezfs:
add groovy as a run-time dep
add gradle build-time dep instead of re-downloading it during the build
add a test

### DIFF
--- a/Formula/groovyserv.rb
+++ b/Formula/groovyserv.rb
@@ -1,8 +1,8 @@
 class Groovyserv < Formula
   desc "Speed up Groovy startup time"
   homepage "https://kobo.github.io/groovyserv/"
-  url "https://bitbucket.org/kobo/groovyserv-mirror/downloads/groovyserv-1.0.0-src.zip"
-  sha256 "2e684f589792aad638ef5391fbad7a6b3c2e7e1d7f35e098cb1d3d60a8b1882b"
+  url "https://bitbucket.org/kobo/groovyserv-mirror/downloads/groovyserv-1.1.0-src.zip"
+  sha256 "4d425c606fba54c6a9e3babc79f178de470fc3d5479a1561d947ca7045b40544"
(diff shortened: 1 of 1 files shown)


Formula/groovyserv.rb
```diff
@@ -14,17 +14,13 @@ class Groovyserv < Formula
...
 
   depends_on "go" => :build
-
-  # This fix is upstream and can be removed in the next released version.
-  patch do
-    url "https://github.com/kobo/groovyserv/commit/4ea88fbfe940b50801be5e0b0bc84cd0ce627530.diff"
-    sha256 "b7de43a030f97d8368ad9a9b895bd1242abc53c4ce0f12e7ab2acfc82e97da65"
-  end
+  depends_on "gradle" => :build
+  depends_on "groovy" => :run
```
DomT4: IMO, reasonable to not bother with the `=> :run` in this case.

ilovezfs: what happened to pedantry über alles?

DomT4: Unless it actually means something actionable I can't be bothered with `:run`.

We use it where deps are almost always `:build` to stop people forgetting to tag 'em properly, but otherwise the practical impact on users for non-special dependencies is nout.

ilovezfs: naught?

DomT4: Nout? It's British slang 🙈. Just means "nothing".

"How much of the money James owes did he give back?"
"Nout 😒"

etc.

ilovezfs: Looks like a simple misspelling of "naught" to me!

DomT4: 😭

ilovezfs:
@DomT4 :ship:?

DomT4:
👍 🙇

PR #3228, opened 2016-07-22
git-annex 6.20160619

ilovezfs:
patch adds cabal constraints for the sake of the aws and esqueleto deps


ilovezfs:
yes the tag name is "wrong" but upstream has decided to just go with it: http://git-annex.branchable.com/bugs/It__39__s_July_not_June/

PR #3231, opened 2016-07-22
gitlab-ci-multi-runner 1.4.0

ilovezfs:
drop godep dependency

### DIFF
--- a/Formula/gitlab-ci-multi-runner.rb
+++ b/Formula/gitlab-ci-multi-runner.rb
@@ -4,8 +4,8 @@ class GitlabCiMultiRunner < Formula
   desc "The official GitLab CI runner written in Go"
   homepage "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner"
   url "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner.git",
-    :tag => "v1.3.3",
-    :revision => "6220bd5f0cdb5008ba9488fabcd556959e80dce5"
+    :tag => "v1.4.0",
+    :revision => "5dd9b2fdfe118bd146b06f905dfa0f3346e2a575"
   head "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner.git"
 
   bottle do
@@ -16,7 +16,6 @@ class GitlabCiMultiRunner < Formula
   end
 
   depends_on "go" => :build
-  depends_on "godep" => :build
   depends_on "docker" => :recommended
 
   go_resource "github.com/jteeuwen/go-bindata" do
@@ -25,18 +24,17 @@ class GitlabCiMultiRunner < Formula
   end
 
   resource "prebuilt-x86_64.tar.gz" do
-    url "https://gitlab-ci-multi-runner-downloads.s3.amazonaws.com/v1.3.3/docker/prebuilt-x86_64.tar.gz",
+    url "https://gitlab-ci-multi-runner-downloads.s3.amazonaws.com/v1.4.0/docker/prebuilt-x86_64.tar.gz",
       :using => :nounzip
-    version "1.3.3"
-    sha256 "5ae536cb7598607f1a0d4c0cb7f31d83d6c19e386106ea50192933a726f33ca9"
+    version "1.4.0"
+    sha256 "01246d4b97ff91e3ce7de258bf0987bee265ed71d647e5dbf10b104b7073a645"
   end
 
   def install
-    mkdir_p buildpath/"src/gitlab.com/gitlab-org"
-    ln_sf buildpath, buildpath/"src/gitlab.com/gitlab-org/gitlab-ci-multi-runner"
-
     ENV["GOPATH"] = buildpath
-
+    proj = "gitlab.com/gitlab-org/gitlab-ci-multi-runner"
+    dir = buildpath/"src/#{proj}"
+    dir.install buildpath.children
     ENV.prepend_create_path "PATH", buildpath/"bin"
     Language::Go.stage_deps resources, buildpath/"src"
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
This smells like it's headed for the boneyard.

ilovezfs:
@DomT4 PR refreshed.

Formula/gitlab-ci-multi-runner.rb
```diff
@@ -25,39 +24,40 @@ class GitlabCiMultiRunner < Formula
...
+        system "go-bindata", "-pkg", "docker",
+               "-nocompress", "-nomemcopy", "-nometadata",
+               "-o", dir/"executors/docker/bindata.go",
+               "prebuilt-x86_64.tar.gz"
+      end
 
-      # Copy from Makefile
-      system "godep", "go", "build", "-o", "gitlab-ci-multi-runner", "-ldflags", "-X main.NAME=gitlab-ci-multi-runner -X main.VERSION=#{version} -X main.REVISION=#{commit_sha}"
+      commit_sha = Utils.popen_read("git", "rev-parse", "--short", "HEAD")
+      system "go", "build", "--ldflags="\
```
DomT4: Does it need the `\` rather than a simple `,`?

DomT4:
One question but otherwise 👍.

Formula/gitlab-ci-multi-runner.rb
ilovezfs: needs to be a single string, so the alternative is `+` not `,`:
```
      system "go", "build", "--ldflags=" +
             "-X #{proj}/common.NAME=gitlab-ci-multi-runner " +
             "-X #{proj}/common.VERSION=#{version} " +
             "-X #{proj}/common.REVISION=#{commit_sha}"
```

ilovezfs: And of course, there's also open rebellion:
```
      ldflags = %W[
        --ldflags=
        -X #{proj}/common.NAME=gitlab-ci-multi-runner
        -X #{proj}/common.VERSION=#{version}
        -X #{proj}/common.REVISION=#{commit_sha}
      ].join(" ")
      system "go", "build", ldflags
```
or
```
      ldflags = %W[
        --ldflags=
        -X #{proj}/common.NAME=gitlab-ci-multi-runner
        -X #{proj}/common.VERSION=#{version}
        -X #{proj}/common.REVISION=#{commit_sha}
      ]
      system "go", "build", ldflags.join(" ")
```

DomT4: > needs to be a single string

Unusual.

I like rebellion option two. That seems at least pleasant to look at, even if it doesn't quite conform to our usual array usage.

ilovezfs: @DomT4 I believe that's the case for the `--ldflags` for all of our go formulae that have that arg.

DomT4: Really? Maybe we just don't have many Go formulae that use multiple `--ldflags` so I hadn't seen it much before. Interesting.

ilovezfs: Right, it's usually only "version" that's being set.

DomT4: Huh. Alright, let's go for one of the rebellion options then. They're cleaner to look at at least.

DomT4: (Subjectively, I appreciate)

ilovezfs: @DomT4 take a look at `doctl.rb`, `gwenhywfar.rb`, `pike.rb`, `rakudo-star.rb`, `ruby.rb`, and `timidity.rb`

So a bit all over the map in terms of precedent.

DomT4: Yeah, we've never really tried to enforce consistency on this sort of thing because it hasn't come up much 😢

ilovezfs: For this at least, I'm inclined to copy `doctl.rb`, since it's a `go` formula, it's close to rebellion option 1, and it's slightly less repetitive than that since it factors out the `base_flag`.

ilovezfs: That would be
```
      commit_sha = Utils.popen_read("git", "rev-parse", "--short", "HEAD")
      base_flag = "-X #{proj}/common"
      ldflags = %W[
        #{base_flag}.NAME=gitlab-ci-multi-runner
        #{base_flag}.VERSION=#{version}
        #{base_flag}.REVISION=#{commit_sha}
      ].join(" ")
      system "go", "build", "--ldflags", ldflags
```
which of course looks a bit over engineered.

ilovezfs: More choices:
```
      commit_sha = Utils.popen_read("git", "rev-parse", "--short", "HEAD")
      ldflags = %W[
        -X #{proj}/common.NAME=gitlab-ci-multi-runner
        -X #{proj}/common.VERSION=#{version}
        -X #{proj}/common.REVISION=#{commit_sha}
      ]
      system "go", "build", "--ldflags", ldflags.join(" ")
or
      system "go", "build", "--ldflags=" + ldflags.join(" ")
or
      system "go", "build", "--ldflags=#{ldflags.join(" ")}"
```

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Pretty grim but may be the best we can get it here. :shipit:

ilovezfs:
Yeah, maybe we can up with something that makes these types of single-string args less awkward, but this is at least less outside the norm than the `\` continuations were.

PR #3234, opened 2016-07-22
openjpeg 2.1.1

JCount:

---

Update and replace v1.5.2 with up-to-date major version 2
NOTE: not ABI/API backwards compatible

### DIFF
--- a/Formula/openjpeg.rb
+++ b/Formula/openjpeg.rb
@@ -1,12 +1,10 @@
 class Openjpeg < Formula
   desc "Library for JPEG-2000 image manipulation"
-  homepage "http://www.openjpeg.org"
-  url "https://mirrors.ocf.berkeley.edu/debian/pool/main/o/openjpeg/openjpeg_1.5.2.orig.tar.gz"
-  mirror "https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/o/openjpeg/openjpeg_1.5.2.orig.tar.gz"
-  sha256 "aef498a293b4e75fa1ca8e367c3f32ed08e028d3557b069bf8584d0c1346026d"
-  revision 1
+  homepage "http://www.openjpeg.org/"
+  url "https://github.com/uclouvain/openjpeg/archive/v2.1.1.tar.gz"
+  sha256 "82c27f47fc7219e2ed5537ac69545bf15ed8c6ba8e6e1e529f89f7356506dbaa"
 
-  head "https://github.com/uclouvain/openjpeg.git", :branch => "openjpeg-1.5"
+  head "https://github.com/uclouvain/openjpeg.git"
 
   bottle do
     cellar :any
@@ -16,9 +14,11 @@ class Openjpeg < Formula
     sha256 "a30aa5b0a7ebcc1daba910671183084d69afb1d30cb85bfeb8b213f8e7a617d7" => :mountain_lion
   end
 
+  option "without-doxygen", "Do not build HTML documentation."
   option "with-static", "Build a static library."
 
   depends_on "cmake" => :build
+  depends_on "doxygen" => [:build, :recommended]
   depends_on "little-cms2"
   depends_on "libtiff"
   depends_on "libpng"
@@ -26,11 +26,9 @@ class Openjpeg < Formula
   def install
     args = std_cmake_args
     args << "-DBUILD_SHARED_LIBS=OFF" if build.with? "static"
+    args << "-DBUILD_DOC=ON" if build.with? "doxygen"
     system "cmake", ".", *args
     system "make", "install"
-
-    # https://github.com/uclouvain/openjpeg/issues/562
-    (lib/"pkgconfig").install_symlink lib/"pkgconfig/libopenjpeg1.pc" => "libopenjpeg.pc"
   end
 
   test do
@@ -39,7 +37,7 @@ class Openjpeg < Formula
 
       int main () {
         opj_image_cmptparm_t cmptparm;
-        const OPJ_COLOR_SPACE color_space = CLRSPC_GRAY;
+        const OPJ_COLOR_SPACE color_space = OPJ_CLRSPC_GRAY;
 
         opj_image_t *image;
         image = opj_image_create(1, &cmptparm, color_space);
@@ -48,7 +46,7 @@ class Openjpeg < Formula
         return 0;
       }
     EOS
-    system ENV.cc, "-I#{include}/openjpeg-1.5", "-L#{lib}", "-lopenjpeg",
+    system ENV.cc, "-I#{include}/openjpeg-2.1", "-L#{lib}", "-lopenjp2",
            testpath/"test.c", "-o", "test"
     system "./test"
   end
--- a/Formula/gpac.rb
+++ b/Formula/gpac.rb
@@ -31,7 +31,6 @@ class Gpac < Formula
   depends_on "sdl" => :optional
   depends_on "theora" => :optional
   depends_on "ffmpeg" => :optional
-  depends_on "openjpeg" => :optional
 
   def install
     args = ["--disable-wx",
--- a/Formula/grib-api.rb
+++ b/Formula/grib-api.rb
@@ -15,7 +15,6 @@ class GribApi < Formula
   depends_on :fortran
   depends_on "cmake" => :build
   depends_on "jasper" => :recommended
-  depends_on "openjpeg" => :optional
   depends_on "libpng" => :optional
 
   # Fixes build errors in Lion
--- a/Formula/libav.rb
+++ b/Formula/libav.rb
@@ -21,7 +21,6 @@ class Libav < Formula
 
   option "with-opencore-amr", "Enable AMR-NB de/encoding and AMR-WB decoding " \
     "via libopencore-amrnb and libopencore-amrwb"
-  option "with-openjpeg", "Enable JPEG 2000 de/encoding via OpenJPEG"
   option "with-openssl", "Enable SSL support"
   option "with-rtmpdump", "Enable RTMP protocol support"
   option "with-schroedinger", "Enable Dirac video format"
@@ -52,7 +51,6 @@ class Libav < Formula
   depends_on "libvorbis" => :optional
   depends_on "libvpx" => :optional
   depends_on "opencore-amr" => :optional
-  depends_on "openjpeg" => :optional
   depends_on "opus" => :optional
   depends_on "rtmpdump" => :optional
   depends_on "schroedinger" => :optional
@@ -91,7 +89,6 @@ class Libav < Formula
     args << "--enable-libmp3lame" if build.with? "lame"
     args << "--enable-libopencore-amrnb" if build.with? "opencore-amr"
     args << "--enable-libopencore-amrwb" if build.with? "opencore-amr"
-    args << "--enable-libopenjpeg" if build.with? "openjpeg"
     args << "--enable-libopus" if build.with? "opus"
     args << "--enable-librtmp" if build.with? "rtmpdump"
     args << "--enable-libschroedinger" if build.with? "schroedinger"
--- a/Formula/openslide.rb
+++ b/Formula/openslide.rb
@@ -3,7 +3,7 @@ class Openslide < Formula
   homepage "http://openslide.org/"
   url "https://github.com/openslide/openslide/releases/download/v3.4.1/openslide-3.4.1.tar.xz"
   sha256 "9938034dba7f48fadc90a2cdf8cfe94c5613b04098d1348a5ff19da95b990564"
-  revision 1
+  revision 2
 
   bottle do
     cellar :any
--- a/Formula/poppler.rb
+++ b/Formula/poppler.rb
@@ -3,6 +3,7 @@ class Poppler < Formula
   homepage "https://poppler.freedesktop.org/"
   url "https://poppler.freedesktop.org/poppler-0.46.0.tar.xz"
   sha256 "967d35d13d61dee2fee656b80efef9e388a9e752bc79b7123f15b49c7769e487"
+  revision 1
 
   bottle do
     sha256 "b646131621704dc93d7a7af4d79fd939a1858dd00457eb5a4cacb92af022ed93" => :el_capitan
@@ -41,7 +42,7 @@ class Poppler < Formula
 
   def install
     ENV.cxx11 if MacOS.version < :mavericks
-    ENV["LIBOPENJPEG_CFLAGS"] = "-I#{Formula["openjpeg"].opt_include}/openjpeg-1.5"
+    ENV["LIBOPENJPEG_CFLAGS"] = "-I#{Formula["openjpeg"].opt_include}/openjpeg-2.1"
 
     args = %W[
       --disable-dependency-tracking


DomT4:
As well as pulling the `poppler` CFLAGS change into here you'll need to add formula revision bumps to both `openslide` and `poppler`.

Formula/openjpeg.rb
```diff
@@ -1,36 +1,31 @@
...
+  url "https://github.com/uclouvain/openjpeg/archive/v2.1.1.tar.gz"
+  sha256 "82c27f47fc7219e2ed5537ac69545bf15ed8c6ba8e6e1e529f89f7356506dbaa"
 
   bottle do
     cellar :any
-    sha256 "50be982ed3846844ac93a34ebd159c680bed484f96b6c0ea274c2e88cc5501de" => :el_capitan
-    sha256 "c9cc5b4f37fdf8b8fc1b043597ab6ac4aa30bfa8fde94b7b0d67a30f03a3cb1b" => :yosemite
-    sha256 "92badbf4968bfda26318855b28940f564e60c701fe3b4e29bc4f173748da7476" => :mavericks
-    sha256 "a30aa5b0a7ebcc1daba910671183084d69afb1d30cb85bfeb8b213f8e7a617d7" => :mountain_lion
+    sha256 "f40e41ca1b1ee59149411ca5054bb7aded1323f64d5b84a30ff88377d946610d" => :yosemite
```
DomT4: Appreciate you pulled this from Versions, but please don't modify the bottle block.

Formula/openjpeg.rb
```diff
@@ -1,36 +1,31 @@
...
+    sha256 "9004d4427b26fb5d5178c5c8e942cc225d6e5e9a2439d63fb74c4022ce87544a" => :mavericks
+    sha256 "f1c0a8432b57e712d0281dc0093a1d072f6f51f3c9e81eb533faa182956d46e3" => :mountain_lion
   end
 
+  option "with-doc", "Build HTML documentation."
```
DomT4: Make this:
```ruby
option "without-doxygen", "Do not build HTML documentation."
```
And then:
```ruby
depends_on "doxygen" => [:build, :recommended]
```
If the HTML docs are huge you can stick to `with-` and `:optional`, but if they're a reasonable size we like to include 'em.

JCount:
whoops, sorry @DomT4 I forgot about the revision bump on poppler. I will add one to that and openslide.

Formula/openjpeg.rb
```diff
@@ -1,36 +1,31 @@
...
 
   def install
     args = std_cmake_args
     args << "-DBUILD_SHARED_LIBS=OFF" if build.with? "static"
+    args << "-DBUILD_DOC=ON" if build.with? "doc"
     system "cmake", ".", *args
     system "make", "install"
-
-    # https://github.com/uclouvain/openjpeg/issues/562
-    (lib/"pkgconfig").install_symlink lib/"pkgconfig/libopenjpeg1.pc" => "libopenjpeg.pc"
```
DomT4: We may need to keep the non-versioned `pkg-config` symlink. Not everything is going to bother to check the versioned one.

Formula/openjpeg.rb
JCount: oh, sorry, i must have accidentally copy-pasted it somehow. I will fix it.

Formula/openjpeg.rb
JCount: I am not quite certain what you mean right here. The ABI/API compatibility is completely broken between the two major versions. http://www.openjpeg.org/abi-check/timeline/openjpeg/ 

The library is now actually openjp2 ; however, most projects have seemed to been phasing in support for the next major version for quite some time.

DomT4: I was just pondering if it was worth keeping a vanilla `libopenjp.pc` symlink. Probably safe to not bother. Let's leave it out for now & see if anything breaks (At least because of that alone).

Formula/openjpeg.rb
JCount: ok. it's also worth mentioning this also builds man pages for the binaries. Both man1 and man3. Total size the share folder is 12.4 MB (not MiB).

JCount:
@DomT4 in this case how much should the commits be broken out? Is one per formula the general goal?

DomT4:
Aye. More or less:

```
openjpeg 2.1.1
poppler: revision for openjpeg
openslide: revision for openjpeg
```
Where each is an individual commit.

Formula/openjpeg.rb
JCount: If that alone caused something to break, it probably would have be due to something that had been hard coded for compatibility, a bit like poppler. Otherwise in my mind, it might actually be an indication of a bit of an issue upstream with how their `pkg-config` related makefiles are constructed. (It would mean they were relying on downstream to use misconfigured build environments from a 'pure' perspective.

JCount:
DomT4, there are some formulae where openjpeg is an optional dep. but actually trying to build with either version of openjpeg fails silently (unless you are doing verbose logging). Namely gpac and grib-api. So, at some point it might make sense to remove the flag because unless you run `linkage` the build flags homebrew shows make it seem as if it was built in.

This does however break `libav --with-openjpeg`. Therefore, should I create a openjpeg15 versions formula and then revise the libav dependencies, vendor it for libav (seems less preferable), or excise it from libav? I really have no sense of the best course here.

DomT4:
> Namely gpac and grib-api. So, at some point it might make sense to remove the flag because unless you run linkage the build flags homebrew shows make it seem as if it was built in.

If they can be fixed, ideally we'd fix them. If they remain broken despite that, we can remove them.

> This does however break libav --with-openjpeg. Therefore, should I create a openjpeg15 versions formula and then revise the libav dependencies, vendor it for libav (seems less preferable), or excise it from libav?

Our analytics data shows less than 40 `libav` installations called with the `--with-openjpeg` flag. I'm fine with getting rid of that flag completely given that.

JCount:
- For `gpac` see: 
[INSTALLME#L26-L41](https://github.com/gpac/gpac/blob/v0.6.1/INSTALLME#L26-L41)
(the list is up to date with the .zip ; and 1.3.0 is ancient, from 2011)
My feeling here is just remove it.

- For `grib-api` see:
[GRIB API CMake installation](https://software.ecmwf.int/wiki/display/GRIB/GRIB+API+CMake+installation)
"Enable JPEG2000 support. This option should look for Jasper or OpenJPG
(OpenJPEG only supported up to version 1.5.2)"
So, it would require a versions formula or vendor of 1.5.2 ;
and a fix because currently cmake cannot find openjpeg (1.5.2).
It would also mean having to modify the formula because `--with-openjpeg` would need to imply `--without-jasper`.

JCount:
For fixing the actual build of `grib-api`:

`--with-openjpeg` results in:
```
-- Could NOT find OpenJPEG (missing:  OPENJPEG_LIBRARY OPENJPEG_INCLUDE_DIR)
```
The problem seems to be that `find_package( OpenJPEG )` is failing in `CMakeLists.txt`

The relevent cmake module is:
[FindOpenJPEG.cmake gist](https://gist.github.com/JCount/1296e489ddeed5251422944bf7acd6d0)

It seems that either `OPENJPEG_PATH` is not set, or the logic used to resolve `OPENJPEG_LIBRARY` & `OPENJPEG_INCLUDE_DIR` from it is not working.

Beyond that, I get stuck in terms of the best remedy here.

JCount:
Also, @DomT4, any idea why the build bot is choking on `poppler` as a dependency of `frescobaldi`? Is this merely a case of letting good old Jenkins digest for a while, or is there something I actively broke somehow and need to fix?

JCount:
This creates issues with `homebrew/science/openimageio` since it relies on `1.5.2`, but there already is a debian [patch](https://anonscm.debian.org/cgit/pkg-phototools/openimageio.git/tree/debian/patches/0003-Fix_OpenJPEG2_build.patch) in testing to add support for `openjp2`.  **_However_**, `openjpeg` is currently a required dependency, so something will have to be done for it in `homebrew/science`.

This pushes me in the direction of creating an `openjpeg15` formula, at least for the time being. 

Debian is strongly pushing projects to move to `openjp2`, or at least add in support, so that could mean a lot going forward.

DomT4:
> For fixing the actual build of grib-api:

Feel free to get rid of it. Looking at our analytic data for `grib-api` nobody has actually built with the option since analytics were switched on. If it turns out anyone cares that much hopefully they'll work with upstream to update the `openjpeg` support.

> However, openjpeg is currently a required dependency, so something will have to be done for it in homebrew/science.

I'd probably just vendor in the older version for that formula. It seems like `openjpeg`'s usage in other formulae is low enough that creating a complete versioned formula for it immediately would be overkill.

Can always reconsider if people complain we've broken things for them, but in all honesty I've been trying to avoid our collection of versioned formulae expanding because we find contributors are less interested in helping maintain them & consequently it becomes an unreasonably heavy weight on us compared to how few people use versioned formulae beside the few obvious big ones.

> Debian is strongly pushing projects to move to openjp2, or at least add in support, so that could mean a lot going forward.

Yes, indeed. We follow Debian on breaking changes from time to time because they tend to make the path a bit easier.

DomT4:
> Also, DomT4, any idea why the build bot is choking on poppler as a dependency of frescobaldi? Is this merely a case of letting good old Jenkins digest for a while, or is there something I actively broke somehow and need to fix?

This is a stupid CI bug on our side. It can be more or less ignored.

JCount:
Ok, so for the PR to vendor `openjpeg` in `homebrew/science/openimageio`, what would be the best way to tie it back to the change here in `homebrew/core`. Should I mention this PR in the commit message for the vendor, or is just referencing this PR in the new PR enough?

JCount:
With a correction to the cmake flags, openimageio's HEAD builds with openjpeg-2.1.1

DomT4:
> Should I mention this PR in the commit message for the vendor, or is just referencing this PR in the new PR enough?

I'd probably stick it in the commit message, yeah.

> Also, for this I would need to bump the revision as well, correct?

Yes.

> This would cause depends_on "cmake" => :build in openimageio to become depends_on "cmake" However, should I add depends_on "little-cms2" to the list of required dependencies for openimageio, since the vendor of openjpeg has it as a required dependency? 

Not sure why `cmake` wouldn't still be `:build`? It's build for `openimageio` & `openjpeg` so it should be fine to remain `:build`? Maybe I was missing the point you were making there.

Adding the `depends_on "little-cms2"` line is 👍, along with a short comment alongside like:
```ruby
depends_on "little-cms2" # Needed for vendored openjpeg
```

Your proposed changes for HEAD in https://github.com/JCount/homebrew-science/commit/dca2236bc7f3b469d5ff23df3aa155706018016f look good.

JCount:
DomT4, yeah it still should be `depends_on "cmake" => :build`. Unfortunately, my brain decided to go on holiday yesterday and I forgot that the vendor, like the other resources, would be bottled.

Also, while I am doing this should I also update the formula to the most recent stable release, 1.6.15? It is still within RB-1.6 so it shouldn't break backwards compatibility.

DomT4:
> Also, while I am doing this, should I also update the formula to the most recent stable release, 1.6.15?

This would be fantastic if you could 👍.

JCount:
ilovezfs actually did that yesterday, because the `glew` update broke its compatibility with `openimageio`. So, I merged those changes into my local work on the formula. I still have to fix some bugs in my vendor of `openjpeg-1.5.2`, mainly with `pkgconfig`, but I should have it done soon; or else completely run out of ideas and relent to seeking real assistance. :confounded:

JCount:
@DomT4: I opened a PR for `openimageio`

Formula/openjpeg.rb
```diff
@@ -1,12 +1,10 @@
...
-  url "https://mirrors.ocf.berkeley.edu/debian/pool/main/o/openjpeg/openjpeg_1.5.2.orig.tar.gz"
-  mirror "https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/o/openjpeg/openjpeg_1.5.2.orig.tar.gz"
-  sha256 "aef498a293b4e75fa1ca8e367c3f32ed08e028d3557b069bf8584d0c1346026d"
-  revision 1
+  homepage "http://www.openjpeg.org/"
+  url "https://github.com/uclouvain/openjpeg/archive/v2.1.1.tar.gz"
+  sha256 "82c27f47fc7219e2ed5537ac69545bf15ed8c6ba8e6e1e529f89f7356506dbaa"
 
-  head "https://github.com/uclouvain/openjpeg.git", :branch => "openjpeg-1.5"
+  head "https://github.com/uclouvain/openjpeg.git"
```
JCount: I inadvertently removed the `head` from the formula. Pointing it to `master` makes the most sense bacause `:branch => "openjpeg-2.1"` would merely build the most recent RB-2.1 point release.

dunn:
FYI homebrew-versions already has the 2.0.x and 2.1.x formulae:
- https://github.com/Homebrew/homebrew-versions/blob/master/openjpeg20.rb
- https://github.com/Homebrew/homebrew-versions/blob/master/openjpeg21.rb

Is the 1.x deprecated, or now less popular than the 2.x branches?

JCount:
https://github.com/uclouvain/openjpeg/releases/tag/version.1.5.2 
- The release was on Mar 28, 2014, and since then it has only received commits to the `openjpeg-1.5` branch, with no release later releases.
- The `CMakeLists.txt` in the head of this branch was tagged for `1.5.3` on April 1, 2014 but since then development on the branch appears to have been sporadic and eventually stalling; the branch is currently `stale` and development is focused on the `openjpeg-2.1` branch, with updates to it being cut from `master`.
- Most of the formulae than currently provide an option to use `openjpeg` actually either silently fail to detect or link it, can easily be patched on homebrew's side to build with `openjpeg-2.1`, already support it, or in the extreme case of `leptonica`, enabling  `--with-openjpeg` breaks the build currently, but with `openjpeg 2.1.1` as a drop in, it builds and links fine.
- The only exceptions that I have found are `openimageio`; see Homebrew/homebrew-science#3865 ; where there are already commits in HEAD for supporting `openjpeg-2.1`, as well as `libav` which currently has no support for anything beyond `1.5.2` that I can find.
- In terms of an external gauge for whether it pseudo-deprecated, Debian has flagged relying on it as a bug that has now been elevated to critical because they plan on removing support for it.
- I can't find any official declaration by `OpenJPEG` that `1.5.2` is deprecated. However as the [OpenJPEG ABI Tracker](http://www.openjpeg.org/abi-check/timeline/openjpeg/) shows, current development is on "v2" of the code which broke backwards compatibility in a quite complete fashion.

@dunn If you have any suggestions or feel that the above information is inaccurate, or that I am interpreting it incorrectly, feel free to give me your thoughts. (You are a member of this project after all) I think I probably was a little over enthusiastic with my response to your comment. However, I felt that putting my reasoning for this change out in the open might be useful. Especially since it could potentially cause some minor issues.

JCount

DomT4:
Per `openimageio`:

> I'm planning to cut releases on Aug 1, and it will build fine against OpenJpeg 2.1. Is that soon enough, or do I need to try to do it sooner? 

I think it'd be reasonable enough to wait 6 days to get this all wrapped up. I'm wary about rushing upstreams unless completely necessary.

> Is the 1.x deprecated, or now less popular than the 2.x branches?

I think we've sat on this for long enough & it's causing relatively minimal grief in comparison to these big updates that it's worth getting done. It's easier to do it at this point whilst only a handful of things need fixing, otherwise we end up in the `jpeg` situation 🙈.

dunn:
I'm totally fine with promoting 2.1 to core!  But yeah, let's wait for OpenImageIO, and kill off those Versions formulae at the same time (or at least the 2.1.x one).

JCount:
That sounds great to me, especially since `1.5.2` resisted being vendored in `openimageio`. (You can tell when I have too much energy, as demonstrated above)

Since `openimageio` will be adding support in the next release, do you want me to simply close the PR in `homebrew/science`, or should I just transform it into a version update when available so that it remains in step with this PR?

DomT4:
Yeah, vendoring is definitely worth avoiding when possible. Even with relatively simple vendors it can bulk out the formula quite a bit.

> should I just transform it into a version update when available so that it remains in step with this PR?

I think this is 👍 if you're happy with that.

DomT4:
@BrewTestBot test this please

JCount:
Only red because of the `frescobaldi` error:
> This is a stupid CI bug on our side. It can be more or less ignored.
~ the wise DomT4

😤

ilovezfs:
@JCount just change
```
depends_on "poppler" => "with-qt"
```
to
```
depends_on "poppler" => ["with-qt", "without-qt5"]
```
in `frescobaldi` and it will be fixed.

DomT4:
@ilovezfs Did you test that? I loosely thought because it was already installed this sort of thing would just happen:
```
~> brew install poppler --with-qt
Warning: poppler-0.46.0 already installed
```

ilovezfs:
@DomT4 You're right. It's caused by the `dspdfviewer` install higher up, so that won't work. It could be fixed by changing the `dspdfviewer` to use a default poppler installation since `dspdfviewer` has Qt 4 support if we chose to take advantage of it https://github.com/dannyedel/dspdfviewer/blob/8aadd17b35c9adaa88e87d2b23d68eac64c5a768/poppler-qt.h#L1-L4

ilovezfs:
@DomT4 since frescobaldi has Qt5 support on the way https://github.com/wbsoft/frescobaldi/issues/842#issuecomment-233535533, I think maybe in the mean time dspdfviewer should have qt5 as a non-default option to avoid the conflict.

ilovezfs:
@DomT4 won't setting bottle unneeded on dspdfviewer avoid the problem too?
```
iMac-TMP:~ joe$ time brew install -s dspdfviewer
==> Installing dependencies for dspdfviewer: poppler
==> Installing dspdfviewer dependency: poppler
==> Downloading https://poppler.freedesktop.org/poppler-0.46.0.tar.xz
Already downloaded: /Users/joe/Library/Caches/Homebrew/poppler-0.46.0.tar.xz
==> ./configure --prefix=/usr/local/Cellar/poppler/0.46.0 --enable-xpdf-headers --enable-poppler-glib --disable-gtk-test --enable-introspection=yes --enable-poppler-qt5
==> make install
==> Downloading https://poppler.freedesktop.org/poppler-data-0.4.7.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/poppler--font-data-0.4.7.tar.gz
==> make install prefix=/usr/local/Cellar/poppler/0.46.0
🍺  /usr/local/Cellar/poppler/0.46.0: 489 files, 23.6M, built in 48 seconds
==> Installing dspdfviewer
==> Downloading https://github.com/dannyedel/dspdfviewer/archive/v1.15.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/dspdfviewer-1.15.tar.gz
==> cmake .. -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/dspdfviewer/1.15 -DCMAKE_BUILD_TYPE=Release -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_VE
==> make install
🍺  /usr/local/Cellar/dspdfviewer/1.15: 7 files, 429.9K, built in 9 seconds

real	0m58.627s
user	3m36.110s
sys	0m30.066s
iMac-TMP:~ joe$ 
```
It doesn't look terribly burdensome in terms of build time for users even if you multiply those times by four to be conservative.

ilovezfs:
Also, since both `qt5` and `qt` are non-default options for poppler, the poppler build-time doesn't "count" since the user will have to build poppler regardless of whether dspdfviewer has a bottle since the qt deps aren't build-time only.

DomT4:
> changing the dspdfviewer to use a default poppler installation since dspdfviewer has Qt 4 support if we chose to take advantage of it

Not really useful to anyone given we're very likely cutting off support for Qt4 with Sierra. We want to be going the other way really, reducing the `qt4` elements around.

> won't setting bottle unneeded on dspdfviewer avoid the problem too?

Yes, but doing so is a regression on the overall desire to bottle _everything_ that uses a development tool native to OS X/macOS. The eventual desire for Homebrew is the ability to use Homebrew without _needing_ to install Xcode/CLT at all.

I think we can continue to ignore the error on `frescobaldi` in the medium term. To date doing so hasn't caused a single reported problem.

ilovezfs:
I think the dspdfviewer should have its bottle disabled to avoid perpetual CI conflicts.

ilovezfs:
Hence: https://github.com/Homebrew/homebrew-core/pull/3556

DomT4:
Left a comment there. Other maintainers are welcome to do so. I don't think it needs to hold this up.

DomT4:
Let's get this over with. Let the chaos begin!

Merged in https://github.com/Homebrew/homebrew-core/commit/544b30511f3b7a1dd899a39d604b966c7b78bb85, https://github.com/Homebrew/homebrew-core/commit/c4cf09f5d488c49e6e934c80ed4a2c673b5e9b8b, https://github.com/Homebrew/homebrew-core/commit/aa32c6c45f50e2a32c8ad068c8f9387707416f5f, https://github.com/Homebrew/homebrew-core/commit/6d76d79f7ba6079e209dd3c3f7b2dc3ae39a86ab, https://github.com/Homebrew/homebrew-core/commit/77da839deed56d9674f4e2b39c8e6943003da4f6 & https://github.com/Homebrew/homebrew-core/commit/099c2976d30b93001e18df7baace00612aefdedc. Thank you for your patience on this, and your contributions to Homebrew @JCount 😺.

ilovezfs:
@JCount you're awesome! Great work on this. :fireworks:

JCount:
Thanks @DomT4 and @ilovezfs! 🎊 
This would have _**no**_ semblance of a mergeable PR without you! 🙇

PR #3241, opened 2016-07-23
python and python3: add --with-sphinx-doc option for building HTML docs

zmwangx:

---

Having offline HTML docs is good for air travel, crappy hotel Wi-Fi or other situations where docs.python.org is not accessible or very slow.

### DIFF
--- a/Formula/python.rb
+++ b/Formula/python.rb
@@ -18,6 +18,13 @@ class Python < Formula
   option "with-tcl-tk", "Use Homebrew's Tk instead of OS X Tk (has optional Cocoa and threads support)"
   option "with-poll", "Enable select.poll, which is not fully implemented on OS X (https://bugs.python.org/issue5154)"
 
+  # sphinx-doc depends on python, but on 10.6 or earlier python is fulfilled by
+  # brew, which would lead to circular dependency.
+  if MacOS.version > :snow_leopard
+    option "with-sphinx-doc", "Build HTML documentation"
+    depends_on "sphinx-doc" => [:build, :optional]
+  end
+
   deprecated_option "quicktest" => "with-quicktest"
   deprecated_option "with-brewed-tk" => "with-tcl-tk"
 
@@ -197,6 +204,13 @@ class Python < Formula
     (libexec/"setuptools").install resource("setuptools")
     (libexec/"pip").install resource("pip")
     (libexec/"wheel").install resource("wheel")
+
+    if MacOS.version > :snow_leopard && build.with?("sphinx-doc")
+      cd "Doc" do
+        system "make", "html"
+        doc.install Dir["build/html/*"]
+      end
+    end
   end
 
   def post_install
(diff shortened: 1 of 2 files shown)


DomT4:
It's a small concern that using `sphinx-doc` for `python` introduces a circular dependency on OS X 10.6 because `sphinx-doc` needs Python 2.7 & the most common way to satisfy that Python 2.7 requirement will be our `python`.

That said, frankly I'm not sure we've got any 10.6 users left at this point. Any thoughts on this PR in general @tdsmith?

zmwangx:
Right, I wasn't thinking about 10.6... Maybe something like

```ruby
if MacOS.version > :snow_leopard
  option "with-sphinx-doc", "Build HTML documentation"
  depends_on "sphinx-doc" => [:build, :optional]
end
```

for `python.rb`? It looks horrible though and I've never seen anything like this in core.

DomT4:
Personally I'd just like to completely stop supporting 10.6 for `homebrew/core`, but that's my opinion. Let's see what Tim has to say.

zmwangx:
This proposed changed is optional rather than recommended (because the uncompressed docsets are pretty huge — 34M for Python 2.7.12 and 41M for Python 3.5.2), so I'd assume it doesn't have to await a heavy-handed decision like dropping support for 10.6 completely even if the option doesn't work on 10.6, right?

Of course there might be other concerns.

tdsmith:
lgtm

DomT4:
> Of course there might be other concerns.

Yeah, this isn't the primary driver of my problems with 10.6. It's more driven by us being basically unable to fix most problems that come up without significant user help, which isn't always forthcoming, and there's an unreasonable number of blobs in formulae for 10.6/10.7 to support what is less than 0.2% of Homebrew's entire userbase.

Still, not strictly relevant to this PR itself.

zmwangx:
I meant to say there might be other concerns about this PR itself... But points taken 😺. I'm not bothered at all by dropping support for 10.6 simply because I'm not in that demographic...

MikeMcQuaid:
Lets try and support 10.6 if we can and it's no extra grief.

zmwangx:
So any thoughts on

```ruby
if MacOS.version > :snow_leopard
  option "with-sphinx-doc", "Build HTML documentation"
  depends_on "sphinx-doc" => [:build, :optional]
end
```

proposed above?

MikeMcQuaid:
Looks good 👍

DomT4:
You'll need to tweak the `if build.with? "sphinx-doc"` install blob in `python` as well. Something like:
```ruby
if MacOS.version > :snow_leopard && build.with?("sphinx-doc")
```
Seems fine.

zmwangx:
I made the change (with a comment on why), but I'm not sure where to place the block.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/259d62b76f1eda2e63d7c88651689e3629da93df & https://github.com/Homebrew/homebrew-core/commit/1e8c627cc584add11e0266253b2429bfb5419bcc. Thanks @zmwangx!

PR #3206, opened 2016-07-22
nghttp2 1.13.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #3243, opened 2016-07-23
spdylay: remove libxml2 dependency

ilovezfs:


ilovezfs:
CC @DomT4 ludicrous, eh?

ilovezfs:
folded this PR into #3206

PR #3247, opened 2016-07-23
pup 0.4.0

ilovezfs:
- upstream now uses the vendor directory
- clean up the formula


DomT4:
Looks like all the resources except perhaps `gox` & its dependencies are safe to snip. Upstream introduced `/vendor`.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Test should be using `#{bin}/pup` but otherwise seems reasonable 👍.

ilovezfs:
@DomT4 fixed.

DomT4:
LGTM

PR #3255, opened 2016-07-23
offlineimap 7.0.0

ilovezfs:
`six` is required but not provided by the system until 10.10


ilovezfs:
@DomT4 or we can drop Mavericks support.

DomT4:
Na, that's overkill here. It's alright as-is.

PR #3258, opened 2016-07-23
libmatheval 1.1.11 (new formula)

dholm:


Formula/libmatheval.rb
```diff
@@ -0,0 +1,40 @@
...
+  homepage "https://www.gnu.org/software/libmatheval/"
+  url "https://ftpmirror.gnu.org/libmatheval/libmatheval-1.1.11.tar.gz"
+  sha256 "474852d6715ddc3b6969e28de5e1a5fbaff9e8ece6aebb9dc1cc63e9e88e89ab"
+
+  depends_on "bdw-gc" => :linked
+  depends_on "bison" => :build
+  depends_on "guile"
+
+  def install
+    # The configure script adds the Guile link directives in the wrong order.
```
ilovezfs: This is a brand new formula, so it would be nice if we weren't already having to patch it. Has this been reported upstream?

Formula/libmatheval.rb
```diff
@@ -0,0 +1,40 @@
...
+    ENV.append "LDFLAGS", `guile-config link`
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    # "tests" depends on deprecated Guile features.
```
ilovezfs: This also seems problematic. A vanilla "make install" should work, and it would be nice to be able to run the tests. Has the fact that the tests rely on deprecated features been reported upstream?

Formula/libmatheval.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      int main(void) {
+        void* evaluator = evaluator_create("x^2");
+        printf("%s\\n", evaluator_get_string(evaluator));
+        evaluator_destroy(evaluator);
+        return 0;
+      }
+    EOS
+    system ENV.cc, "-o", testpath/"test",
+           testpath/"test.c", "-lmatheval"
```
ilovezfs: should also have `-L#{lib}`

Formula/libmatheval.rb
```diff
@@ -0,0 +1,40 @@
...
+    assert_equal "(x^2)", pipe_output("#{testpath}/test").strip
```
ilovezfs: `pipe_output` does not assert its exit status by default, so `shell_output` should be used unless you're actually passing something in as well. Also, this should be `./test`

Formula/libmatheval.rb
```diff
@@ -0,0 +1,40 @@
+class Libmatheval < Formula
+  desc "Library to parse and evaluate symbolic expressions input as text."
+  homepage "https://www.gnu.org/software/libmatheval/"
+  url "https://ftpmirror.gnu.org/libmatheval/libmatheval-1.1.11.tar.gz"
+  sha256 "474852d6715ddc3b6969e28de5e1a5fbaff9e8ece6aebb9dc1cc63e9e88e89ab"
+
+  depends_on "bdw-gc" => :linked
```
ilovezfs: does the documentation specify this as a dependency?

dunn:
🔔

Formula/libmatheval.rb
dholm: No, upstream appears to be dead. :/

Formula/libmatheval.rb
dholm: Support requests on the mailing list have mostly been unanswered for the past two years.

Formula/libmatheval.rb
dholm: Fixed.

Formula/libmatheval.rb
dholm: Fixed.

Formula/libmatheval.rb
dholm: No, this was suggested by `brew audit --strict --online libmatheval`.

MikeMcQuaid: You can remove the `:linked` dependencies now.

Formula/libmatheval.rb
MikeMcQuaid: We're almost certainly not going to accept this then, sorry.

MikeMcQuaid:
Upstream is dead and needs patched to work so passing on this, sorry!

PR #3259, opened 2016-07-23
uwsgi 2.0.13.1

dholm:
Depends on #3258.

### DIFF
--- a/Formula/uwsgi.rb
+++ b/Formula/uwsgi.rb
@@ -1,9 +1,8 @@
 class Uwsgi < Formula
   desc "Full stack for building hosting services"
   homepage "https://uwsgi-docs.readthedocs.org/en/latest/"
-  url "https://projects.unbit.it/downloads/uwsgi-2.0.11.2.tar.gz"
-  sha256 "0b889b0b4d2dd3f6625df28cb0b86ec44a68d074ede2d0dfad0b91e88914885c"
-  revision 2
+  url "https://projects.unbit.it/downloads/uwsgi-2.0.13.1.tar.gz"
+  sha256 "2eca0c2f12ab76f032154cd147f4d5957d3195a022678d59cb507f4995a48d7f"
 
   head "https://github.com/unbit/uwsgi.git"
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
I'm not eager to add more options to this formula. Also, I'm not thrilled with adding more untestable `:tuntap` even if it's optional.

DomT4:
Agree it'd be nice to limit the options a little until we see more demand for them. The _only_ options our analytics know about being used are `--with-python` and `--with-python --with-python3`. That's it, out of all the options included before today.

apjanke:
While we want to limit options, these plugins could be added unconditionally for the ones that don't introduce new dependencies, right? (e.g. greenlet, stackless, coro, libffi)

What do folks think about changing that big `plugins = ["airbrake", "alarm_curl", ` to be one item per line, for more readable diffs? I can see that this PR added (and maybe removed) some items there and re-wrapped them, but I can't tell what changed without running my fingers along the lines and comparing them item by item.

ilovezfs:
I wonder if there's an "all plugins" option to avoid the crazy.

ilovezfs:
I guess we couldn't do that anyway because of `:tuntap`. Yeah one per line sounds fine.

DomT4:
Yeah, still worth keeping the array non-empty but:
```ruby
plugins = %w[
  airbrake
  alarm_curl
  alarm_speech
  asyncio
  cache
]
```
And so on is a lot more readable.

dholm:
Will fix the array and options.

I don't think there is an all plugin option, the annoying thing is that if you try to build a plugin that is unsupported the entire build fails. :(

dholm:
@apjanke `greenlet` and `stackless` require Python modules to be installed separately.  The same goes for `coro` except that it's a Perl module.

apjanke:
I like the new formatting a lot better. Can't remember why the last build failed, so...

@BrewTestBot test this please

MikeMcQuaid:
Install is failing: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/6671/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_uwsgi/

MikeMcQuaid:
Closing due to no response.

PR #3262, opened 2016-07-23
xcbuild 0.1.0 (new formula)

modocache:

---

xcbuild is an Xcode-compatible build tool developed by Facebook.

### DIFF
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -1,0 +1,24 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git",
+    :tag => "0.1.0",
+    :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa",
+    :shallow => false
+  head "https://github.com/facebook/xcbuild.git",
+    :shallow => false
+
+  depends_on "cmake" => :build
+  depends_on "ninja" => :build
+  depends_on "libpng"
+
+  def install
+    system "make"
+    system "make", "install", "PREFIX=#{prefix}"
+    bin.install_symlink "#{prefix}/usr/bin/xcbuild"
+  end
+
+  test do
+    system "#{bin}/xcbuild", "-showsdks"
+  end
+end


ilovezfs:
@modocache This needs
```
  depends_on "libpng"
```
and then it's fine.

modocache:
Yup, looks like it. I'll consider submitting an update to the xcbuild documentation as well.

ilovezfs:
@modocache `libpng` cannot be `=> :build` since it's actually linked at run time. It doesn't need any decorators in this case.

Formula/xcbuild.rb
```diff
@@ -0,0 +1,22 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0",
+                                                 :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa",
+                                                 :shallow => false
+  head "https://github.com/facebook/xcbuild.git", :shallow => false
+
+  depends_on "cmake" => :build
+  depends_on "libpng" => :build
```
ilovezfs: should go last as well, since we prefer =>:build deps first

modocache:
@ilovezfs Thanks for all the reviews! Updated.

Formula/xcbuild.rb
```diff
@@ -0,0 +1,22 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0",
```
DomT4: Please split each `:` onto its own line. Having everything indented over to the far right is both pretty grim looking & blasts past our desired ~80 character width.

DomT4: i.e.
```ruby
url "https://github.com/facebook/xcbuild.git",
    :tag => "0.1.0"
```
etc.

modocache: Can do!

Formula/xcbuild.rb
```diff
@@ -0,0 +1,24 @@
...
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git",
+    :tag => "0.1.0",
+    :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa",
+    :shallow => false
+  head "https://github.com/facebook/xcbuild.git",
+    :shallow => false
+
+  depends_on "cmake" => :build
+  depends_on "ninja" => :build
```
grp: Ninja is useful, but not technically required, for using xcbuild. I don't know Homebrew well enough, but something like Apt's `Recommends:` would make sense here? If that's not possible, I'd probably say drop the `:build`.

DomT4: I read from the above comment that you need/would likely want `ninja` at runtime as well as buildtime?

We have a similar syntax to Apt here. Just:
```ruby
depends_on "ninja" => :recommended
```

Formula/xcbuild.rb
```diff
@@ -0,0 +1,24 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git",
+    :tag => "0.1.0",
+    :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa",
+    :shallow => false
+  head "https://github.com/facebook/xcbuild.git",
+    :shallow => false
```
DomT4: Just to be confusing:
```ruby
head "https://github.com/facebook/xcbuild.git", :shallow => false
```
Was actually fine. We tend to drop `:tag` and `:revision` onto their own lines but it doesn't make as much sense for the `:shallow` modifier.

ilovezfs: more of those nebulous and esoteric white space rules whose secrets are never truly known in all their splendor :)

DomT4: Yeah. We used to have a full example formula with detailed line references for people to glance through, but we scrapped it in favour of better split documentation & more real-world formulae usage examples. Sometimes I wonder if it would have still be useful 😓.

ilovezfs: are these shallow false actually needed?

ilovezfs: they look highly suspicious to me :)

willneumob:
Definitely need this.

modocache:
Sorry to let this languish, will update in a few hours.

lembacon:
Still no updates after 12 days 🤔️

MikeMcQuaid:
> Still no updates after 12 days 🤔️

@lembacon It's a weekend and we're all unpaid volunteers working in our spare time 🤔

MikeMcQuaid:
(including our contributors like @modocache)

MikeMcQuaid:
Closing but will accept a new PR with comments addressed.

modocache:
Thanks! Yup, happy to submit another one, hopefully soon. It'd be awesome if someone beat me to it! :)

PR #3263, opened 2016-07-24
Berkeley db dbm

horta:
Adding the option to enable the legacy dbm support, required for Linux-PAM for example.


DomT4:
You've pulled in hundreds of commits, almost all of them impacting Linux & not OS X/macOS. Please rebase against an up-to-date master.

Issue #3252, opened 2016-07-23
Update open-zwave


irundaia:
Hi guys,

I'm trying to update the open-zwave formula. However I'm running into an issue here. 
First I tried to install openzwave directly from the source. Using `make install` the library and headers were automatically installed in `/usr/local/`.

So I thought that updating the formula shouldn't be too hard... I've updated the formula to:

```ruby
class Openzwave < Formula
  desc "Library for selected Z-Wave PC controllers"
  homepage "http://www.openzwave.com"
  url "http://old.openzwave.com/downloads/openzwave-1.4.164.tar.gz"
  sha256 "4ecf39787aaf278c203764069b581dbc26094ce57cafeab4a0c1f012d2c0ac69"

  def install
    ENV["BUILD"] = "release"
    ENV["PREFIX"] = prefix

    system "make", "install"
  end

  test do
    (testpath/"test.cpp").write <<-EOS.undent
      #include <iostream>
      #include "Options.h"
      int main() {
        OpenZWave::Options::Create("", "", "");
        return 0;
      }
    EOS
    system ENV.cxx, "-o", "test", "-I", prefix/"include/openzwave", "-lopenzwave", "test.cpp"
    system "./test"
  end
end
```

However, when installing openzwave using `brew install open-zwave` fails. Below the log can be found:

```
2016-07-23 13:33:21 +0200

make
install

/Applications/Xcode.app/Contents/Developer/usr/bin/make -C /private/tmp/openzwave-20160723-32952-1d52u9p/openzwave-1.4.164/cpp/build/ - --jobserver-fds=3,4 -j install
Makefile:193: Documentation not being built
Making libopenzwave pkg-config file
Installing Shared Library
Installing Headers
Installing Config Database
Installing Documentation
Installing Pkg-config Files
install: chmod 755 //usr/lib/pkgconfig: Operation not permitted
cp: //usr/lib/pkgconfig/libopenzwave.pc: Operation not permitted
make[1]: *** [install] Error 1
make: *** [install] Error 2

HOMEBREW_VERSION: 0.9.9
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 2b9a2833bc3c6bc8e7b7344e8178ce98e29ebe4b
Last commit: 2 days ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 926d76796fd24c48a71e46f9fd671cb51e8b6e20
Core tap last commit: 9 hours ago
[...]

PR #3268, opened 2016-07-24
Update open-zwave.rb

irundaia:
Hi, 

I was asking for [help](https://github.com/homebrew/homebrew-core/issues/3252) while updating open-zwave. @DomT4 asked me to submit a pull request, so here it is. 

Please do not merge this as-is.

---

Update open-zwave to version 1.4.164.

### DIFF
--- a/Formula/open-zwave.rb
+++ b/Formula/open-zwave.rb
@@ -1,26 +1,22 @@
 class OpenZwave < Formula
   desc "Library for selected Z-Wave PC controllers"
   homepage "http://www.openzwave.com"
-  url "http://old.openzwave.com/downloads/openzwave-1.2.919.tar.gz"
-  sha256 "473229f3dd3d6b260e6584b17e5c5f2e09e61805f89763f486a9f7aa2b4181ba"
+  url "http://old.openzwave.com/downloads/openzwave-1.4.164.tar.gz"
+  sha256 "4ecf39787aaf278c203764069b581dbc26094ce57cafeab4a0c1f012d2c0ac69"
 
-  bottle do
-    sha256 "4bf1d8a8ba3fcf4ee39df9bfe09017d3a432047e5e6bf0dce6f6e612ad174b95" => :el_capitan
-    sha256 "9ceb267d8fb564daefb535f65ac71f426535f7552b9feffcb3638793f0c40810" => :yosemite
-    sha256 "753eb6cb76dd0c170c1e84285a702fb9fb49c76aa8d59970ecea2160938f0bba" => :mavericks
-    sha256 "be2b4c51703ee93865a6c755a4394b6271eb599ea98a7431b54d090276a0e9f7" => :mountain_lion
-  end
-
-  # Patch to build a .dylib instead of a .so
-  # This patch has been merged upstream and can be removed after the next release:
-  # https://code.google.com/p/open-zwave/source/detail?r=954
-  patch :DATA
+  depends_on "pkg-config" => :build
 
   def install
     ENV["BUILD"] = "release"
     ENV["PREFIX"] = prefix
 
-    system "make", "install"
+    # Make sure to set the correct install name otherwise the dylib would be incorrectly linked.
+    # An issue was opened upstream: https://github.com/OpenZWave/open-zwave/issues/947
+    inreplace "cpp/build/Makefile",
+      "LDFLAGS += -dynamiclib",
+      "LDFLAGS += -dynamiclib -install_name #{lib}/$(SHARED_LIB_NAME)"
+
+    system "make", "install", "pkgconfigdir=#{lib}/pkgconfig", "sysconfdir=#{etc}/open-wave"
   end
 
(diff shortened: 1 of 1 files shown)


DomT4:
```
cpp/build/support.mk
94:pkgconfigdir ?= $(shell if [ -d "/usr/lib64/pkgconfig" ]; then echo "/usr/lib64/pkgconfig"; else echo "/usr/lib/pkgconfig"; fi)

ozw_config
19:pcfile=/usr/lib/pkgconfig/libopenzwave.pc
```

That's probably why. It's ended up hardcoded to `/usr/lib`, regardless of the `prefix` being passed, partly because there's no `pkg-config` dependency in this formula so it assumes what upstream presumably think is a safe default _(Very much isn't on OS X)_. Adding a `pkg-config => :build` dep doesn't entirely resolve the problem though:

```
install: chmod 755 //usr/local/lib/pkgconfig: Operation not permitted
cp: //usr/local/lib/pkgconfig/libopenzwave.pc: Operation not permitted
```

This'll work:
```diff
diff --git a/Formula/open-zwave.rb b/Formula/open-zwave.rb
index 718accf..cc641f7 100644
--- a/Formula/open-zwave.rb
+++ b/Formula/open-zwave.rb
@@ -4,11 +4,13 @@ class OpenZwave < Formula
   url "http://old.openzwave.com/downloads/openzwave-1.4.164.tar.gz"
   sha256 "4ecf39787aaf278c203764069b581dbc26094ce57cafeab4a0c1f012d2c0ac69"
 
+  depends_on "pkg-config" => :build
+
   def install
     ENV["BUILD"] = "release"
     ENV["PREFIX"] = prefix
 
-    system "make", "install"
+    system "make", "install", "pkgconfigdir=#{lib}/pkgconfig"
   end
 
   test do
```

irundaia:
Great thanks, I can confirm that this works now. But the files still aren't installed in /usr/local. I'll have another look at it soon. 

Thanks for the effort!

DomT4:
Are you sure they aren't there in symlink form? Homebrew doesn't place anything, well, except perhaps `npm`, into `/usr/local` _directly_.

We install into, for example, `/usr/local/Cellar/open-zwave/1.4.164` and then anything in that directory + `bin`, `include`, `lib` or `share` gets linked into `/usr/local` via symlink.

irundaia:
I couldn't find them with an `ls -l` in `/usr/local/lib` I'll make sure to have another look.

irundaia:
You are correct! Sorry for the mixup :(

DomT4:
Yes, indeed. It's a very odd one:
```
Warning: The installation was broken.
Broken dylib links found:
  /private/tmp/open-zwave-20160728-22096-xjz6t2/openzwave-1.4.164/libopenzwave-1.4.dylib
==> Summary
🍺  /usr/local/Cellar/open-zwave/1.4.164: 386 files, 3.0M, built in 15 seconds
```

@ilovezfs: You're good at mysteries. Any thoughts here?

ilovezfs:
Yeah, the problem is that install_name is not set so the example binary MinOZW gets linked with the path to the library at link time, which happens to be the buildpath in this case.

```
+    inreplace "cpp/build/Makefile",
+      "LDFLAGS += -dynamiclib",
+      "LDFLAGS += -dynamiclib -install_name #{lib}/$(SHARED_LIB_NAME)"
```
fixes it.

irundaia:
Indeed that works :+1:  Thanks a bunch!

ilovezfs:
@irundaia you're welcome. It would be helpful if you could report that upstream and include a comment above the inreplace with the URL to the bug report and brief note (can be one line) about what bug it fixes.

Also, the sysconfdir is currently set incorrectly because it's defaulting to #{prefix}/etc (i.e., /usr/local/Cellar/open-zwave/1.4.164/etc, which is incorrect), not #{etc} or #{etc}/open-wave (i.e., /usr/local/etc or /usr/local/etc/open-wave, which is correct). Since it's dumping quite a few files in etc, it would best to make it #{etc}/open-wave in this case.

```
system "make", "install", "pkgconfigdir=#{lib}/pkgconfig", "sysconfdir=#{etc}/open-wave"
```
will fix that.

DomT4:
> Yeah, the problem is that install_name is not set so the example binary MinOZW gets linked with the path to the library at link time, which happens to be the buildpath in this case.

That is, er, quite something 👀. Nice spot, thanks.

DomT4:
Test is failing:
```
==> ./test
�[31m2016-07-28 15:55:02.477 Error, Cannot find a path to the configuration files at . Exiting...
�[39m�[31m2016-07-28 15:55:02.478 Error, Exception: Options.cpp:91 - 1 - Cannot Find Configuration Files
�[39mError: Options.cpp:91 - Cannot Find Configuration Files
Error: open-zwave: failed
Failed executing: ./test 
```

MikeMcQuaid:
Closing due to no response.

PR #3269, opened 2016-07-24
glew 2.0.0

ilovezfs:


ilovezfs:
@DomT4 :green_apple:

DomT4:
👍. There's a small cluster of `glew`-using formulae in Science as well, don't know if you want to sort those out with revision bumps after. Up to you.

ilovezfs:
@DomT4 Yeah, I saw those were there and was dreading them, since CI on science virtually never works unless twenty other formulae are corrected at the same time.

DomT4:
Yeah, I try to avoid Science 😕, not least because they also support Linux for formulae & I have no desire to do that dance at all.

ilovezfs:
@DomT4 what's with the permissions error? https://travis-ci.org/Homebrew/homebrew-core/builds/146997018

tdsmith:
> don't know if you want to sort those out with revision bumps after. Up to you.

I'm glad ILZ is handling the -science fallout but if you'd prefer not to do it yourself please do tag some homebrew-scientists and give them/us a few days before pulling breaking PRs.

DomT4:
I've been ignored for days tagging the entire team before. Become somewhat exasperated with trying. Will try to have even more patience with them in future.

PR #3276, opened 2016-07-24
imagemagick: add zero-configuration option

jeroen:
Apologies but I need one more feature from `imagemagick`. It turns out the default build hardcodes locations of xml configuration files. The `--enable-zero-configuration` option disables this and always uses default settings, which is needed for standalone, statically linked applications.

Upstream issue: https://github.com/ImageMagick/ImageMagick/issues/232

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -30,6 +30,7 @@ class Imagemagick < Formula
   option "without-magick-plus-plus", "disable build/install of Magick++"
   option "without-modules", "Disable support for dynamically loadable modules"
   option "without-threads", "Disable threads support"
+  option "with-zero-configuration", "Disables depending on XML configuration files"
 
   depends_on "xz"
   depends_on "libtool" => :run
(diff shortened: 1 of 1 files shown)


Formula/imagemagick.rb
```diff
@@ -30,6 +30,7 @@ class Imagemagick < Formula
   option "without-magick-plus-plus", "disable build/install of Magick++"
   option "without-modules", "Disable support for dynamically loadable modules"
   option "without-threads", "Disable threads support"
+  option "with-zero-configuragion", "Disables depending on XML configuration files"
```
rakkesh: Did you mean `with-zero-configuration` here?

Formula/imagemagick.rb
```diff
@@ -110,6 +111,7 @@ def install
     args << "--with-fontconfig=yes" if build.with? "fontconfig"
     args << "--with-freetype=yes" if build.with? "freetype"
     args << "--with-webp=yes" if build.with? "webp"
+    args << "--enable-zero-configuration" if build.with? "zero-configuragion"
```
rakkesh: And here, `zero-configuration`.

jeroenooms:
@rakkesh sorry for the typo, thanks for catching that! Corrected now.

apjanke:
No worries. This looks like a reasonable option, and the build works when I tested locally with it.

PR #3279, opened 2016-07-24
Set `gdk-pixbuf` environment variables in non-/usr/local installs

staticfloat:

---

When running `gdk-pixbuf` binaries in a non-`/usr/local` cellar location, we must set the proper environment variables, or the commands will fail, regardless of whether `with-relocations` is enabled or not.

Note that this doesn't quite pass `brew audit --strict --online gdk-pixbuf`, as it complains about `gettext` not being declared as a `:linked` dependency.

I have tested this on both a `/usr/local`-based `brew` installation, and a non-`/usr/local`-based `brew` installation, and the `postinstall` step completes successfully for both.

### DIFF
--- a/Formula/gdk-pixbuf.rb
+++ b/Formula/gdk-pixbuf.rb
@@ -32,6 +32,18 @@ class GdkPixbuf < Formula
     sha256 "c38cbf14bee68a15a12edb55a5fa39e36a8dc3d82b4160e9cefea921eda6a13d"
   end
 
+  # gdk-pixbuf has an internal version number separate from the overall
+  # version number that specifies the location of its module and cache
+  # files, this will need to be updated if that internal version number
+  # is ever changed (as evidenced by the location no longer existing)
+  def gdk_so_ver
+    "2.0"
+  end
+
+  def gdk_module_ver
+    "2.10.0"
+  end
+
   def install
     ENV.universal_binary if build.universal?
     ENV.append_to_cflags "-DGDK_PIXBUF_LIBDIR=\\\"#{HOMEBREW_PREFIX}/lib\\\""
@@ -54,29 +66,29 @@ class GdkPixbuf < Formula
 
     # Other packages should use the top-level modules directory
     # rather than dumping their files into the gdk-pixbuf keg.
-    inreplace lib/"pkgconfig/gdk-pixbuf-2.0.pc" do |s|
+    inreplace lib/"pkgconfig/gdk-pixbuf-#{gdk_so_ver}.pc" do |s|
       libv = s.get_make_var "gdk_pixbuf_binary_version"
       s.change_make_var! "gdk_pixbuf_binarydir",
-        HOMEBREW_PREFIX/"lib/gdk-pixbuf-2.0"/libv
+        HOMEBREW_PREFIX/"lib/gdk-pixbuf-#{gdk_so_ver}"/libv
     end
 
     # Remove the cache. We will regenerate it in post_install
-    (lib/"gdk-pixbuf-2.0/2.10.0/loaders.cache").unlink
+    (lib/"gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache").unlink
   end
 
   def post_install
     # Change the version directory below with any future update
     if build.with?("relocations") || HOMEBREW_PREFIX.to_s != "/usr/local"
-      ENV["GDK_PIXBUF_MODULE_FILE"]="#{lib}/gdk-pixbuf-2.0/2.10.0/loaders.cache"
-      ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
+      ENV["GDK_PIXBUF_MODULE_FILE"]="#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
+      ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders"
     end
     system "#{bin}/gdk-pixbuf-query-loaders", "--update-cache"
   end
 
   def caveats; <<-EOS.undent
     Programs that require this module need to set the environment variable
-      export GDK_PIXBUF_MODULE_FILE="#{lib}/gdk-pixbuf-2.0/2.10.0/loaders.cache"
-      export GDK_PIXBUF_MODULEDIR="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
+      export GDK_PIXBUF_MODULE_FILE="#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
+      export GDK_PIXBUF_MODULEDIR="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders"
     If you need to manually update the query loader cache, set these variables then run
       #{bin}/gdk-pixbuf-query-loaders --update-cache
     EOS
--- a/Formula/gdk-pixbuf.rb
+++ b/Formula/gdk-pixbuf.rb
@@ -66,7 +66,7 @@ class GdkPixbuf < Formula
 
   def post_install
     # Change the version directory below with any future update
-    if build.with?("relocations")
+    if build.with?("relocations") || HOMEBREW_PREFIX.to_s != "/usr/local"
       ENV["GDK_PIXBUF_MODULE_FILE"]="#{lib}/gdk-pixbuf-2.0/2.10.0/loaders.cache"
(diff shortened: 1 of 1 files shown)


Formula/gdk-pixbuf.rb
```diff
@@ -80,7 +80,7 @@ def caveats; <<-EOS.undent
     If you need to manually update the query loader cache, set these variables then run
       #{bin}/gdk-pixbuf-query-loaders --update-cache
     EOS
-  end if build.with?("relocations")
+  end if build.with?("relocations") || HOMEBREW_PREFIX.to_s != "/usr/local"
```
apjanke: This "`if build.with?("relocations")...`" looks like it's in the wrong place. This `end` appears to belong to `def caveats`. The `end` corresponding to the `if build.with?("relocations") ...` looks like it's up on line 72. Let's just remove the notations on the `end` if they're not going to be validated by `ruby`.

staticfloat: I believe that line 83 is meant to suppress defining the `caveats` function at all if the `build.with?` conditional is not met, whereas line 69 is meant to suppress the actual ruby code between lines 69 and 72.

apjanke:
Makes sense to me.

The previous comment in `post_install` says:
```
# Change the version directory below with any future update
```
but the paths are still referencing `#{lib}/gdk-pixbuf-2.0/2.10.0`, even though this is gdk-pixbuf 2.34. There's a 2.10.0 reference a few lines up in `(lib/"gdk-pixbuf-2.0/2.10.0/loaders.cache").unlink`, too. Should these be updated with gdk-pixbuf 2.34? (The numbers all seem internally consistent, but I'm wondering if that could cause problems interacting with other packages.)

Even if not, maybe those "2.0" and "2.10.0" module versions could be collected in variables further up near the main version number, to avoid the repetition make it more obvious to maintainers that they may need to be bumped. 

> Note that this doesn't quite pass brew audit --strict --online gdk-pixbuf, as it complains about gettext not being declared as a :linked dependency.

This is a new audit check that we've added recently, to catch undeclared dependencies. Go ahead and add a `depends_on "gettext"` in this PR to get the audit to pass.

Formula/gdk-pixbuf.rb
apjanke: Oh. That makes sense.

staticfloat:
Unfortunately, the `2.10.0` references are still correct; `gdk-pixbuf` has an internal version number that it keeps at `2.10.0` for the modules, and this will need to be updated whenever they decide to change that internal API number.

What is the proper rubyism for defining this kind of variable?  As someone who doesn't do much ruby, do I just defined a method that returns a constant string.

> Go ahead and add a depends_on "gettext" in this PR to get the audit to pass.

Roger roger.

apjanke:
> What is the proper rubyism for defining this kind of variable?

Uh, good question. These are class-level properties, so I'd say class constants would be appropriate. Functionally equivalent to little methods that just return a value, but you can write them more concisely so they could go in a block with the URL and other version info.

```
class GdkPixbuf < Formula
  ...
  url "https://download.gnome.org/sources/gdk-pixbuf/2.34/gdk-pixbuf-2.34.0.tar.xz"
  sha256 "d55e5b383ee219bd0e23bf6ed4427d56a7db5379729a6e3e0a0e0eba9a8d8879"
  # GDK dylib and module versions are independent of gdk-pixbuf version
  GDK_SO_VERSION = "2.0"
  GDK_MODULE_VERSION = "2.10.0"
  ...

   skip_clean "lib/gdk-pixbuf-#{GDK_SO_VERSION}"
```

But since this is in Homebrew DSL, it's probably a more Homebrew-specific style issue. I'll see if I can find some similar occurrences. And any other maintainers want to chime in here?

For the names, I think something like "SO version" or "dylib version" would be a better term than "major version" for the thing that is "2.0". That looks like a dynamic library compatibility version, like is embedded in `.so` or `.dylib` filenames. Which makes sense, as modules are typically runtime-loaded dylibs.

```
$ ls /usr/local/lib/libgdk*
/usr/local/lib/libgdk_pixbuf-2.0.0.dylib  /usr/local/lib/libgdk_pixbuf-2.0.dylib
/usr/local/lib/libgdk_pixbuf-2.0.a
```

We use SemVer terminology here, ("major.minor.patch") so "major version" would refer to just the "2" component of either the package or dylib version.

Or maybe gdk-pixbuf has some other internal term for those versions?

apjanke:
Tests pass now. Will merge this once I can follow up on the appropriate way to do those factored-out versions.

Formula/gdk-pixbuf.rb
```diff
@@ -19,6 +19,7 @@ class GdkPixbuf < Formula
   depends_on "libtiff"
   depends_on "libpng"
   depends_on "gobject-introspection"
+  depends_on "gettext" => :linked
```
DomT4: Let's cut this. We're not adding any more of these for now.

DomT4:
Seems mildly clunky, but I'm not sure there's any less of a way to make it so if we have to do this. It's fine as-is, beside the one nit highlighted.

staticfloat:
All comments addressed.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/fc05135bffd6419d550d6d9f83afbb15832c7bf5. Thanks again @staticfloat! 😺

staticfloat:
So it turns out that the `loaders.cache` file that is explicitly [unlinked here](https://github.com/staticfloat/homebrew-core/blob/1149a2ffac13d3d092f9a7734c388ed2e0a99f45/Formula/gdk-pixbuf.rb#L75-L76), and then [regenerated here](https://github.com/staticfloat/homebrew-core/blob/1149a2ffac13d3d092f9a7734c388ed2e0a99f45/Formula/gdk-pixbuf.rb#L85), is not getting relinked after the `post_install` step.  What is the proper way to request a relink of the keg?  Or should we just manually add a `ln` step at the end of the `post_install`?

DomT4:
>  is not getting relinked after the post_install step.

Can you elaborate on this? Isn't getting linked into the Cellar or into the prefix? It shouldn't appear in the Cellar, but should be present in:

```
HOMEBREW_PREFIX/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache
```

staticfloat:
Notice that `GDK_PIXBUF_MODULE_FILE` is [being set to `"#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/gdk-pixbuf.rb#L82), which DOES live in the `Cellar`.  This is done on purpose because having non-symlinked files laying around in `HOMEBREW_PREFIX/lib/` isn't a good idea, because `brew` doesn't know what those files belong to, and therefore doesn't remove them when `gdk-pixbuf` gets unlinked.  This doesn't necessarily break anything, (I believe the next `post_install` invocation of `gdk-pixbuf-query-loaders` will overwrite the pre-existing file no problem) but doesn't seem very homebrew-like.

staticfloat:
On second thought, this may be a difference between non-`/usr/local` brew and the typical configuration:  It's entirely possible that `gdk-pixbuf`, in the absence of those environment variables IS placing its file in `/usr/local/lib/gdk-pixbuf/2.10.0`, and therefore this issue doesn't manifest for the vast majority of users (and, subsequently, we have non-symlink files laying around in `/usr/local/lib`).  If true, I personally am in favor of always setting the environment variables in the `post_install` step to normalize behavior.

bjarthur:
ping.  not sure i understand the last two comments.  what's the path forward?

staticfloat:
The path forward is to remove [this conditional](https://github.com/Homebrew/homebrew-core/blob/530190f23c95eaddd9c7e7844dbfe39d6ba9d060/Formula/gdk-pixbuf.rb#L81), instead always defining the environment variables, and then link the created file (`#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache`) into `#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache`.

This is because as it stands right now, in a "normal" `/usr/local` Homebrew installation, the `loaders.cache` file is not a symlink to the currently-installed `gdk-pixbuf` keg, it is a normal file.  This (theoretically) doesn't allow for things like quick switching from version to version through unlinking different kegs of the same formula, just plain doesn't work properly on non-`/usr/local` Homebrew installations, and just in general breaks Homebrew encapsulation.

On a non-`/usr/local` Homebrew installation, the environment variables are defined properly and the `loaders.cache` file is created in the proper place, but no symlink is created because [`post_install` is invoked after `link`](https://github.com/Homebrew/brew/blob/7395ff20018785bed7d008337b7cf98781190e0a/Library/Homebrew/formula_installer.rb#L494-L505).  This could be as simple as adding an `ln -sf` step to the `post_install` function, but I was asking the Homebrew devs what the most idiomatic way of doing this is.

mkborregaard:
I am just reporting that I had this issue on 0.6, and the lines of code here: https://github.com/JuliaPackaging/Homebrew.jl/issues/121#issuecomment-236888166 fixed it. Gtk version 0.13.0.

JCount:
@mkborregaard 
In the future, please don't comment on long resolved/closed issues/PRs, especially when it pertains solely to an independent, outside project. You should make a comment within that project itself and merely reference the external thread.

PR #3281, opened 2016-07-25
qjackctl: fix build errors

### LINKED ISSUE #3278: Build error for qjackctl
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

Running `brew install qjackctl` fails for me with basically this error in the configure step:

    configure: error: qmake-qt5 not found in current PATH. Maybe QT development environment isn't available. (qt5-devel)

Full log files as requested in the issue template: https://gist.github.com/muellermartin/cb2815e53d5243022bdf921dd9eb24b3

This issue could be obsoleted by #2260 if there would be a readily available build.

muellermartin:

---

qmake-qt5 cannot be found by the configure script. This is fixed adding
the prefix of the Qt5 installation as a configure option.
Building with LLVM/Clang raises an error related to a variable sized
array. This is fixed by patching a source file to use a dynamically
allocated array instead.

Fixes #3278.

### DIFF
--- a/Formula/qjackctl.rb
+++ b/Formula/qjackctl.rb
@@ -19,7 +19,7 @@ class Qjackctl < Formula
   # Fixes varaible length array error with LLVM/Clang
   # Reported upstream: https://github.com/rncbc/qjackctl/issues/17
   patch do
-    url "https://github.com/rncbc/qjackctl/pulls/18.patch"
+    url "https://github.com/rncbc/qjackctl/pull/18.patch"
     sha256 "cbb7cc72e0086b8e6df24c8c3a462dc0e9297f095573adb7a3cf98502a1902d4"
   end
 
@@ -30,7 +30,7 @@ class Qjackctl < Formula
                           "--disable-dbus",
                           "--disable-xunique",
                           "--prefix=#{prefix}",
-                          "--with-qt5=#{Formula["qt5"].prefix}"
+                          "--with-qt5=#{Formula["qt5"].opt_prefix}"
 
     system "make", "install"
     prefix.install bin/"qjackctl.app"
--- a/Formula/qjackctl.rb
+++ b/Formula/qjackctl.rb
@@ -19,7 +19,7 @@ class Qjackctl < Formula
   # Fixes varaible length array error with LLVM/Clang
   # Reported upstream: https://github.com/rncbc/qjackctl/issues/17
   patch do
-    url "https://patch-diff.githubusercontent.com/raw/rncbc/qjackctl/pull/18.patch"
+    url "https://github.com/rncbc/qjackctl/pulls/18.patch"
     sha256 "cbb7cc72e0086b8e6df24c8c3a462dc0e9297f095573adb7a3cf98502a1902d4"
   end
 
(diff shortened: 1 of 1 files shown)


Formula/qjackctl.rb
```diff
@@ -14,14 +14,23 @@ class Qjackctl < Formula
   depends_on "autoconf" => :build
   depends_on "qt5"
   depends_on "jack"
+  depends_on "portaudio" => :linked
+
+  # Fixes varaible length array error with LLVM/Clang
+  # Reported upstream: https://github.com/rncbc/qjackctl/issues/17
+  patch do
+    url "https://github.com/rncbc/qjackctl/pulls/18.patch"
```
DomT4: Should just be `pull` right?

Formula/qjackctl.rb
```diff
@@ -14,14 +14,23 @@ class Qjackctl < Formula
...
 
   def install
     system "./autogen.sh"
     system "./configure", "--disable-debug",
                           "--enable-qt5",
                           "--disable-dbus",
                           "--disable-xunique",
-                          "--prefix=#{prefix}"
+                          "--prefix=#{prefix}",
+                          "--with-qt5=#{Formula["qt5"].prefix}"
```
DomT4: This shouldn't really be necessary due to the way Homebrew sets up the environment, but if it is use `opt_prefix` rather than `prefix`. Thanks.

Formula/qjackctl.rb
muellermartin: Oh, you're absolutely right, that must have been happened when changing to the short version suggested by `brew audit`. I'll fix it!

Formula/qjackctl.rb
muellermartin: For me it's necessary (see #3278). I guess this is because `qt5` is not liked due to conflicts with `qt`...
I'll update the formula to use `opt_prefix` instead of `prefix`.

Formula/qjackctl.rb
```diff
@@ -14,14 +14,23 @@ class Qjackctl < Formula
   depends_on "autoconf" => :build
   depends_on "qt5"
   depends_on "jack"
+  depends_on "portaudio" => :linked
+
+  # Fixes varaible length array error with LLVM/Clang
+  # Reported upstream: https://github.com/rncbc/qjackctl/issues/17
+  patch do
+    url "https://github.com/rncbc/qjackctl/pull/18.patch"
```
DomT4: Would you mind switching this over to the commit patch now upstream have merged the PR? Thanks!

muellermartin: I had to look up how to do that, but I guess this is now accomplished in the latest commit of the formula (d9a396e).

DomT4: Apologies, yeah, you just take the upstream commit reference & prod `.diff` or `.patch` on the end of it.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/07ec6e6c02e8da9980a774ab2c3821f4c3cda9f7. Thank you for once again @muellermartin! 😺

PR #3284, opened 2016-07-25
newt: always build --with-python

tdsmith:
Newt is almost never installed without python (cf http://rpubs.com/tdsmith/198233, bottom figure, second row.)


DomT4:
👍. ❤️ if you can think of a test to add to `newt`.

PR #3287, opened 2016-07-25
Golo 3.2.0-M3 (devel)

rlespinasse:
http://golo-lang.org/news/2016/07/25/golo-3.2.0-M3


ilovezfs:
Thanks for your contribution to Homebrew! We appreciate it.

PR #3293, opened 2016-07-25
little-cms2: fix checksum

ilovezfs:
upstream didn't gzip the tar file in the original upload


ilovezfs:
Reported by user to us here: https://github.com/Homebrew/homebrew-versions/issues/1321
Reported to upstream: https://github.com/mm2/Little-CMS/issues/88

PR #2691, opened 2016-07-03
libvirt 2.0.0

nijikon:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

PR #3301, opened 2016-07-25
libvirt 2.0.0

justinclift:


justinclift:
Note, this is an updated replacement for PR #2691.

Formula/libvirt.rb
```diff
@@ -14,6 +14,7 @@ class Libvirt < Formula
 
   depends_on "pkg-config" => :build
   depends_on "gnutls"
+  depends_on "curl"
```
DomT4: Which version of `curl` does it require? I'd be surprised if the version of cURL on El Cap for example wasn't new enough.

Formula/libvirt.rb
```diff
@@ -62,4 +67,26 @@ def install
       end
     end
   end
+
+  test do
+    output = shell_output("#{bin}/virsh -v", 0)
```
DomT4: The `, 0` bit is assumed with `shell_output` unless it's told otherwise. Doesn't need stating explicitly.

DomT4:
CI failure was a false fail. Jenkins had a panic. I picked up a couple of issue outside of that but we look pretty good otherwise.

Formula/libvirt.rb
justinclift: Any version seems to work.  It's just that the default curl doesn't seem to be found for some unknown reason (I didn't investigate), whereas a Homebrew provided one does.

Formula/libvirt.rb
justinclift: No worries.  Will push a revised commit in a sec.

Formula/libvirt.rb
DomT4: Interesting. Can you provide the failure logs? Wondering if that's something we can fix fairly easily.

justinclift: Sure, give me a few minutes to run through it again. :smile:

justinclift: ... and now the failure isn't occurring.

I have no idea what was causing it the other day, but curl doesn't seem to be needed after all. 😵 

Will push a new commit in a sec.

justinclift:
@DomT4 Sorry for the noise with curl.  No idea what happened there. 😦

DomT4:
No worries at all! These things happen. Appreciate you double checking to spare us the unnecessary dependency.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/55faf289b26bf124f6a3a700a39b60b0a1ccf7a1. Thank you for another contribution to Homebrew @justinclift; we appreciate it! :smiley_cat:

justinclift:
Thanks @DomT4. :smile:

PR #3330, opened 2016-07-26
emacs migrated to cask

AnastasiaSulyagina:

---

Deleting emacs from homebrew-core as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


DomT4:
Crossposting myself here: https://github.com/Homebrew/homebrew-core/pull/3333#issuecomment-235429626

MikeMcQuaid:
I think in this case we could consider removing X11 and/or Cocoa options in favour of the Cask for them as we still build a CLI by default (I believe?)

MikeMcQuaid:
To differentiate these we could consider renaming formulae like this to `emacs-cli` (and/or maybe rename the cask to `emacs-gui`)?

zmwangx:
Again I find this unacceptable. To reiterate (on https://github.com/caskroom/homebrew-cask/issues/15603#issuecomment-223133740),

1. CLI-only Emacs (as bottled) works perfectly fine and is used by many.
2. emacsforosx.com is NOT the official build, and the GNU Emacs project does NOT distribute official binaries. You can't remove the official build option in favor of a third-party binary distribution.

I think renaming the cask to `emacs-gui` might be a good idea.

zmwangx:
> I think renaming the cask to emacs-gui might be a good idea.

Or `emacs-app` maybe, taking a page from MacPorts (see https://www.gnu.org/software/emacs/download.html).

zmwangx:
Sorry for not keeping my comments as one...

> I think in this case we could consider removing X11 and/or Cocoa options in favour of the Cask for them as we still build a CLI by default (I believe?)

That of course comes with the (big) catch that people can no longer install devel and HEAD versions of Emacs.app, because cask only provides `emacs` and `emacs-nightly` at the moment (nightly != HEAD).

Also, expect friction as many people are probably used to `brew install emacs --with-cocoa` by now (it's even on https://www.gnu.org/software/emacs/download.html). (Already covered by @DomT4 before I published this post)

Does `--with-cocoa` generate a lot of support burden?

MikeMcQuaid:
> emacsforosx.com is NOT the official build, and the GNU Emacs project does NOT distribute official binaries. You can't remove the official build option in favor of a third-party binary distribution.

I'm confused about "official build option" considering the GNU Emacs site links to both `brew install` and the emacsforosx.com site.

dunn:
The GUI options are used by at least half of the install base, right?  If we're dead-set on removing GUI apps from core, I'd rather the entire formula be migrated to the Emacs tap.

MikeMcQuaid:
> We don't really handle that sort of divorce well. We remove the options, the next time people rebuild they get an unexpected vanilla CLI emacs and come moan at us asking what changed...

Yes, we may need some DSL that expresses than an option has gone away and tells people what to do instead.

> I really want to avoid the situation where we make people brew install emacs for the CLI & brew cask install emacs for the GUI or some combination thereof.

If the goal of bringing Homebrew and Cask closer together is to reduce or remove the number of .apps we have in Homebrew: isn't this somewhat inevitable? What would you propose instead? I think https://github.com/Homebrew/brew-evolution/pull/3 helps highlight some of the various problems we have with the status quo.

MikeMcQuaid:
> The GUI options are used by at least half of the install base, right? If we're dead-set on removing GUI apps from core, I'd rather the entire formula be migrated to the Emacs tap.

Another option we could consider is migrating only users of particular options to the casks. I think the (my) long-goal is to have basically everything that builds or handles an app be a Cask rather than a Formula so we can remove all code like e.g. `linkapps` from Homebrew.

zmwangx:
> > emacsforosx.com is NOT the official build, and the GNU Emacs project does NOT distribute official binaries. You can't remove the official build option in favor of a third-party binary distribution.
>
> I'm confused about "official build option" considering the GNU Emacs site links to both brew install and the emacsforosx.com site.

In case I wasn't clear, emacsforosx.com and `brew install emacs --with-cocoa` are both third party solutions — emacsforosx.com listed alongside (and below) `brew install` reinforces the fact that it is unofficial. For an Emacs binary to be official it has be hosted by gnu.org on ftp.gnu.org.

However, downloading the source tarball from ftp.gnu.org and run `./configure --with-whatever-options`, `make`, `make install` is the official build option, and `brew install --build-from-source emacs --with-whatever-options` is just a wrapper around the official option.

MikeMcQuaid:
> In case I wasn't clear, emacsforosx.com and brew install emacs --with-cocoa are both third party solutions — emacsforosx.com listed alongside (and below) brew install reinforces the fact that it is unofficial. For an Emacs binary to be official it has be hosted by gnu.org on ftp.gnu.org.

I see 👍 

> However, downloading the source tarball from ftp.gnu.org and run ./configure --with-whatever-options, make, make install is the official build option, and brew install --build-from-source emacs --with-whatever-options is just a wrapper around the official option.

I partially agree here but I do think there's a difference when we're running this for people and distributing binaries (not for this option, I'll admit).

DomT4:
>  If we're dead-set on removing GUI apps from core, I'd rather the entire formula be migrated to the Emacs tap.

This tosses out our policy of not accepting cross-tap dependencies in `core` though, unless you also wanted every formula with a recursive `emacs` dep there as well. That would be a serious policy change to Homebrew.

> If the goal of bringing Homebrew and Cask closer together is to reduce or remove the number of .apps we have in Homebrew: isn't this somewhat inevitable? What would you propose instead? I think Homebrew/brew-evolution#3 helps highlight some of the various problems we have with the status quo

I think this is too black-and-white, at least right out of the gate. I think primarily CLI applications with an _optional_ GUI element is reasonable, and I think it's reasonable to say Homebrew isn't the place for pure `.apps`. I think we're going to seriously degrade UX if we immediately cut an arbitrary line in the sand here.

I think, generally speaking, expecting users to install both CLI & GUI separately is really crappy UX where we already offer a formula _option_ to install that GUI.

If you want to go down that route particularly I'd like to see us implement a mechanism where `brew install emacs --with-cocoa` triggers a vanilla `brew install emacs` & then triggers `brew cask install emacs`. Essentially `depends_on :cask => "emacs" if build.with? "cocoa"` or such. 

Although in this case that's not even a good example because the Cask `emacs` isn't an official one.

MikeMcQuaid:
> I think this is too black-and-white, at least right out of the gate. I think primarily CLI applications with an optional GUI element is reasonable, and I think it's reasonable to say Homebrew isn't the place for pure .apps. I think we're going to seriously degrade UX if we immediately cut an arbitrary line in the sand here.

Sorry, to be clear: let's try and figure out what perfect looks like and then work our way back and figure out how to get there.

> I think, generally speaking, expecting users to install both CLI & GUI separately is really crappy UX where we already offer a formula option to install that GUI.

If the existing GUI doesn't provide some sort of CLI: I'd be surprised if the same users are using both.

MikeMcQuaid:
To be clear on my comments though: more contentious formulae like this will be the last to be migrated. I'm in no rush, here.

dunn:
> This tosses out our policy of not accepting cross-tap dependencies in core though, unless you also wanted every formula with a recursive emacs dep there as well. That would be a serious policy change to Homebrew.

Ah, right.

ilovezfs:
I think it's fine to bar .app/GUI options from ever being the default in core, but categorically ruling out non-default .app/GUI options seems like a rather obvious mistake.

ilovezfs:
>Another option we could consider is migrating only users of particular options to the casks. I think the (my) long-goal is to have basically everything that builds or handles an app be a Cask rather than a Formula so we can remove all code like e.g. linkapps from Homebrew.

That would be acceptable if Cask gets a build-from-source/build-bottle feature, which doesn't seem insane to me if there's truly going to be more integration.

zmwangx:
Another thought about `--with-cocoa`, might have been brought up before: what about forgoing `linkapps` but leaving the option as is, and mention the app bundle in the caveat, like

> If you built emacs --with-cocoa, you can find Emacs.app in $(brew --prefix emacs), and you might want to symlink it to /Applications by running
>
>     ln -s "$(brew --prefix emacs)/Emacs.app" /Applications

or, if people are concerned about the issues discussed in https://github.com/Homebrew/brew-evolution/pull/3,

> ..., you might want to move it to /Applications by running
> 
>     rm -rf /Applications/Emacs.app && mv "$(brew --prefix emacs)/Emacs.app" /Applications

ilovezfs:
You should be aware that another one of the Laws of Robotics is that we're trying to kill off as many caveats as possible, but that could arguably be treated as one of the "valid" uses. I think that leads naturally to being replaced with an automated version of the same thing, but that was already proposed and rejected in https://github.com/Homebrew/brew-evolution/pull/3, so I'm really not sure where that leaves us logically other than launching apps from the Cellar, which I'm fairly sure no one will be OK with.

zmwangx:
The resolution of Homebrew/brew-evolution#3 was

> I think we have some sort of agreement here that we will not implement this proposal but instead start to gradually deprecate linkapps (by moving app bundles out of HB), right?

(Parenthesized part by me.)

And out of all these [/^\w+ migrated to cask$/ PRs](https://github.com/Homebrew/homebrew-core/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aopen%20label%3A%22maintainer%20feedback%22%20%22migrated%20to%20cask%22), only the Emacs thread has (rightfully) generated this much attention, so I think it's fair to not replace with an automated version (hard to do satisfactorily anyway) and leave a caveat for this special case.

ilovezfs:
@zmwangx I think for the sake of fairness we might have to include macvim too :D

DomT4:
I suspect it's quite likely that if we bar `.app`s completely, regardless of mandatory/recommended/optional status, we _will_ end up launching GUI apps out of the Cellar in much the same way we already have done previously.

Before it was CLI only executing a vanilla `fontforge` launched the GUI, rather than using the `.app` which was broken _(Not due to Homebrew in particular, I add)_. With `wireshark` before the latest release you used to only be able to access the GUI by running for example a vanilla `wireshark`; there's an `.app` now but that method of launching is still supported. Same with `emacs --with-cocoa` - Vanilla execution launches GUI.

Ultimately if we shoot down `.app`s 100% with no exceptions or optional inclusion, that's where we're heading again unless we completely bar GUI in Homebrew, in which case a big chunk of the GTK/GTK+3/Qt/Qt5 dependency tree is off to die.

zmwangx:
@ilovezfs Well, vim is going nowhere, so I don't think the Church of Emacs owe the Cult of vi anything here... But free feel to appeal https://github.com/Homebrew/homebrew-core/pull/3337, I'm always happy to see people's wished granted, even if people refer to vi users 😃 

> Same with emacs --with-cocoa - Vanilla execution launches GUI.

Ah, I forgot that. Yes, that's quite annoying.

ilovezfs:
@zmwangx The reason that it's probably fine that MacVim has been moved is that the upstream build has nearly every option enabled. Is there no analogue for emacs that isn't awful?

zmwangx:
> Is there no analogue for emacs that isn't awful?

Not that I know of. emacsforosx.com (built from https://github.com/caldwell/build-emacs) is sort of semi-official, but it doesn't come close to having every option enabled (https://github.com/caldwell/build-emacs/issues/1 might be the reason; see also https://github.com/caldwell/build-emacs/issues/25 and https://github.com/caldwell/build-emacs/issues/37). I don't know any other trusted binary distribution.

There are probably Emacs experts around here who can tell you more.

MikeMcQuaid:
> Not that I know of. emacsforosx.com (built from https://github.com/caldwell/build-emacs) is sort of semi-official

If it's recommended on the official site and `emacs` in Cask: I think that's official enough for us. Like a lot of things: if it doesn't include Your Favourite Option: I think that's a discussion to be had with them. It could be no-one ever asked and it could be our analytics help convince them.

> Ultimately if we shoot down .apps 100% with no exceptions or optional inclusion, that's where we're heading again unless we completely bar GUI in Homebrew, in which case a big chunk of the GTK/GTK+3/Qt/Qt5 dependency tree is off to die.

GTK and Qt stuff is a somewhat special case as, most of the time, it doesn't even try to build `.app`s  and mostly just are targeting Linux desktop environments and just happens to support OS X due to the underlying toolkit. I would perhaps be in favour of getting them into their own taps, eventually, bar literally the GTK and Qt formulae themselves.

What I want to do longer-term with this move is convince people that to port a GUI application to OS X you need to provide a binary, self-enclosed .app bundle that can be moved by `cask install` and then not care about any Cask or Homebrew dependencies again. Perhaps if very popular upstreams provide buildsystems that do this for us but cannot distribute the binaries for whatever reason we could consider building the .apps ourselves somehow and distributing them as a Cask.

The main problems we have (which seem to bear repeating):
1. `brew linkapps` provides a shoddy experience whether using symlinks or aliases
2. almost no application bundles are actual application bundles (i.e. self-contained, relocatable) so we cannot improve 1.
3. large GUI applications often take a long time to build, use a huge number of options that we can't CI test and have large dependency trees which takes much longer to install
4. our work improving the situation benefits only Homebrew and not other package managers or OS X GUI application users

We're done similar moves before with gently pushing out (most) things that only work in 32-bit mode, things that don't build with Clang, things that require X11 and things that require manual user intervention before installing. These changes have resulted in a better user experience for Homebrew users and a better user experience for everyone using OS X with or without a package manager. I think this is probably a harder move (and therefore may take longer) than the previous ones but I think it's arguably more rather than less important.

zmwangx:
I still don't see what's so wrong about having a directory named Emacs.app in the keg. You can use the GUI without the "app": if you built Emacs --with-cocoa, `bin/emacs` launches the GUI by default, and `bin/emacs -nw` launches the TUI. You don't have to know anything about or do anything with the app bundle, so `linkapps` is not necessary here. AFAIK things only get messy if you want to launch Emacs from, say, Spotlight, and even then you'll only run into issues if you keep multiple versions around. People have been living with that for years. How many people have complained? Just because there were certain issues with symlinking apps over at HBC (which deals with thousands of apps) doesn't mean these issues affect the user experience of every single app bundle.

> If it's recommended on the official site and emacs in Cask: I think that's official enough for us. Like a lot of things: if it doesn't include Your Favourite Option: I think that's a discussion to be had with them. It could be no-one ever asked and it could be our analytics help convince them.

People asked (see issues linked above). The problem seems to be that Emacs was never supposed to be a self-contained app bundle in the first place (not if you want to link against various dependencies). This is problem 2 in your list, but as I said above, problem 1 may not need to be tackled, so problem 2 may not be a problem either. Problem 4 also seems to resolve itself if you just stop trying to solve problem 1. As for problem 3,

> large GUI applications often take a long time to build, use a huge number of options that we can't CI test and have large dependency trees which takes much longer to install

Emacs build time is pretty short, and `--with-cocoa` doesn't lead to a huge dep tree. Testing, well... If the test bot could test hugely popular options in the future (mark certain options as `:popular` or something?), then I suppose it would not only benefit Emacs. But I understand that my armchair comment may not be welcome.

---

The main problem I see with removing `--with-cocoa` is that, as other folks and I have mentioned above, the GUI option will be crippled on multiple levels (optional dependencies, version availability, etc.), and that's not good for the ecosystem around Emacs where people do need the flexibility. Not to mention for most people it's kind of weird to have two independent Emacs installations serving different purposes.

> What I want to do longer-term with this move is convince people that to port a GUI application to OS X you need to provide a binary, self-enclosed .app bundle

It's a respectable goal, but good luck convincing FSF... For one, I don't think FSF loves OS X / macOS (they apparently don't) to the extent that they will invest lots of time to improve the user experience of Emacs.app on OS X which isn't necessary after all; also, I think people are happy enough with the status quo.

Overall, (I think) I fully understand why the policy change is necessary, but what about keeping one (or two, or three) exceptions, for historical reasons maybe?

MikeMcQuaid:
> I still don't see what's so wrong about having a directory named Emacs.app in the keg.

OS X GUI applications should be relocatable, self-contained (and arguably self-updating) .apps. There's various issues that you run into when they are not.

> Overall, (I think) I fully understand why the policy change is necessary, but what about keeping one (or two, or three) exceptions, for historical reasons maybe?

Exceptions mean we can't kill code and they mean having to explain them ad infinitum, unfortunately.

ilovezfs:
>OS X GUI applications should be relocatable, self-contained (and arguably self-updating) .apps

Putting aside "self-updating" since that's contrary to policy, if a .app built by brew were indeed "relocatable and self-contained," would it then be acceptable to have in core, and indeed even as a default option?

zmwangx:
> Exceptions mean we can't kill code and they mean having to explain them ad infinitum, unfortunately.

I still think `linkapps` can be killed without `--with-cocoa` being killed. As for explanations, there might be more explanations to do when people find `--with-cocoa` missing.

> OS X GUI applications should be relocatable, self-contained (and arguably self-updating) .apps. There's various issues that you run into when they are not.

That's why I said "a directory named Emacs.app in the keg". I think we can accept the fact that it's not relocatable and self-contained (we have `--disable-ns-self-contained`), thus not a standard "app". It's just something with a GUI component that's masquerading under the `.app` extension (which is unfortunately necessary for the functionality of part of the GUI, including the dock icon and menu bar).

---

I do realize I'm not going to convince anyone, so I'm backing down and instead trying to improve the cask Emacs experience a bit:

- Add `emacs-pretest` to fill the gap: https://github.com/caskroom/homebrew-versions/pull/2362;
- Add appcasts to `emacs` (https://github.com/caskroom/homebrew-cask/pull/23214) and `emacs-nightly` (https://github.com/caskroom/homebrew-versions/pull/2363).

In particular, up to this point `emacs-nightly` was seldom updated, which completely defeats the purpose of a nightly. I'm not sure if the non-Sparkle, non-GitHub, non-SourceForge, non-HockeyApp appcast here will actually help (I haven't been following the development of HBC for a while). But at the very least we can keep the pretest version up to date.

zmwangx:
Another thought on removing `--with-cocoa`: if we're going to remove it, it's probably better to do it sooner than later, along with Emacs 25 major release. We're already on 25.1-rc1, which means Emacs 25 is not too far ahead. (Looking at the release history, there were only two weeks between 24.5-rc1 and 24.5.)

ilovezfs:
If the decision has already been made that the Cocoa and X11 options are going away but the formula is staying, should this PR be updated to reflect that intent?

AnastasiaSulyagina:
So should I completely strip emacs of x11 and cocoa?

DomT4:
> I still think linkapps can be killed without --with-cocoa being killed.

Agree with this, per the discussion further up.

> If the decision has already been made that the Cocoa and X11 options are going away but the formula is staying, should this PR be updated to reflect that intent?

I don't think there's a hard decision made here on this yet.

You might get a consensus on X11, but I think there's merit to discussing the `cocoa` option, not least because it is what upstream recommend & it seems unlikely to ever go anywhere productive upstream due to the political situation between GNU & non-free operating systems.

There's also the situation with GNOME where some stuff that should be an `.app` hasn't been updated as an `.app` for years, and that doesn't seem likely to change, so we've built more recent versions as a non-app GUI in Homebrew.

I agree with Mike entirely that Homebrew provides a non-optimal experience around `.app`s, but I think we're waving a turquoise flag when the politics around this are tangerine or mauve more often than not.

All of the alternatives proposed here _so far_ feel like a bit of a regression on the status quo for users, and I'm not looking at this _particular discussion_ and thinking _"Wow, it is going to save me a bunch of time now `emacs` no longer has that option"_.

Agree overall migrating appropriate things to the Cask is a big 👍, but here I'm rather ❓.

zmwangx:
> All of the alternatives proposed here so far feel like a bit of a regression on the status quo for users

Speaking of regressions, moving any part of Emacs to Cask will definitely be a regression in terms of upgradeability (AFAICT cask upgrade is getting nowhere). Let's agree on the fact that Emacs.app will never upgrade itself.

MikeMcQuaid:
There's no hard decision yet; let's hold off merging this for a while and modifying the emacs formula. This is popular enough of a package that it'll probably be one of the last we handle. Just leave this open for now, thanks @AnastasiaSulyagina!

---

A happy middle-ground it seems may be to kill linkapps and related code once we've done enough cask migrations while leaving stuff like Emacs (and perhaps all "cross-platform" apps that will never have a Cask-friendly self-update process)? How's that sound?

dunn:
> A happy middle-ground it seems may be to kill linkapps and related code once we've done enough cask migrations while leaving stuff like Emacs (and perhaps all "cross-platform" apps that will never have a Cask-friendly self-update process)? How's that sound?

So there'd still be `Emacs.app`, just no support for linking/moving it to `/Applications`? Fine with me.

ilovezfs:
>You might get a consensus on X11

@DomT4 Given the long-standing rule that X11 options are fine as long as they're non-default, I'd see no reason to kill the X11 option if the Cocoa is staying. If someone finds it useful, that's awesome even if other people think X11 on OS X is :no_good:.

dunn:
I'd vote to make Cocoa default and kill X11.  Mike objected to making the GUI on by default in the past, but if we're killing off the `linkapps` stuff would it be less of an issue?

ilovezfs:
>A happy middle-ground it seems may be to kill linkapps and related code once we've done enough cask migrations while leaving stuff like Emacs (and perhaps all "cross-platform" apps that will never have a Cask-friendly self-update process)? How's that sound?

@MikeMcQuaid This sounds very reasonable. What would the UX be for these app bundles?  Say nothing, and let the users figure out their own workflow for these app bundles? Caveat suggesting copy-pasting them into /Applications? Homebrew mechanism to copy-paste them into /Applications? Something else?

MikeMcQuaid:
> Say nothing, and let the users figure out their own workflow for these app bundles?

I'd say this. Copy-pasting won't work because they aren't self-contained. I'd maybe even go as far as saying we extract the executable from the app bundle and stick it in `bin`.

> I'd vote to make Cocoa default and kill X11. Mike objected to making the GUI on by default in the past

I'd still probably like to avoid the default being a GUI, particularly if we're making the choose to support .apps less than we do now.

DomT4:
> I'd maybe even go as far as saying we extract the executable from the app bundle and stick it in bin.

Symlinks/exec scripts for this already work great for `emacs` and other formulae & I believe it's pretty much what Cask do with executables they want to be public from inside `.app`s, so I'd be 👍 on that sort of thing.

Still allows us to move away from things like `linkapps` as well, which is essentially broken code.

> Given the long-standing rule that X11 options are fine as long as they're non-default, I'd see no reason to kill the X11 option if the Cocoa option is staying. If someone finds it useful, that's awesome even if other people think X11 on OS X is 🙅.

I was working off analytics. 68 known installations are using `--with-x11` on this formula, either alone or with other arguments, compared to... over 4000 using `--with-cocoa`.

ilovezfs:
>over 4000 using --with-cocoa

Sounds like cocoa should probably be the default despite the heterodoxy.

DomT4:
Bearing in mind the total known number of completely vanilla `emacs` installations is 5.7k since analytics went live, so an option like `--with-cocoa` doing 4k+ is huge.

zmwangx:
> I'd maybe even go as far as saying we extract the executable from the app bundle and stick it in bin.

That doesn't work. I went down that path before hoping it would be the ideal solution, but although there are just a few files in `Emacs.app`:

```
$ tree /usr/local/opt/emacs/Emacs.app
/usr/local/opt/emacs/Emacs.app
└── Contents
    ├── Info.plist
    ├── MacOS
    │   └── Emacs
    ├── PkgInfo
    └── Resources
        ├── Credits.html
        ├── Emacs.icns
        ├── English.lproj
        │   └── InfoPlist.strings
        └── document.icns

4 directories, 7 files
```

they are essential for the dock icon and menu bar to function.

ilovezfs:
>Bearing in mind the total known number of completely vanilla emacs installations is 5.7k since analytics went live, so an option like --with-cocoa doing 4k+ is huge.

Yeah, that basically means our packaging is "wrong."

MikeMcQuaid:
> Yeah, that basically means our packaging is "wrong."

Or that at least some of those people would have a better time with the Cask but just aren't aware of it.

ilovezfs:
>Or that at least some of those people would have a better time with the Cask but just aren't aware of it.

Yeah I think that's two ways of saying the same thing. 4000/5700 is ... insane given people barely use options *at all* in general.

ilovezfs:
@DomT4 or is it 4000/9700? Still completely crazy.

DomT4:
@ilovezfs Yeah, the second set of numbers is more accurate. 5.7k known vanilla `emacs` installations + at least 4k known `emacs --with-cocoa` installations.

MikeMcQuaid:
I think pointing people towards the `cask` in the `caveats` for `with-cocoa` would be a good idea.

> This makes me want to build them statically and/or bury an entire mini brew installation inside them.

@ilovezfs Maybe some site like emacsforosx.com could do that and have a binary that could be distributed through Homebrew Cask 😉

ilovezfs:
@MikeMcQuaid I shall leave that to someone who actually knows what magical incantation is needed to exit emacs :)

I meant more generally though that it seems like self-contained app bundles could be generated for /Applications if brew itself was "vendored" inside them.

MikeMcQuaid:
> @MikeMcQuaid I shall leave that to someone who actually knows what magical incantation is needed to exit emacs :)

(while investigating this thread I ran the system one and had to kill it by closing the Terminal. It may still be running there in the background somewhere 😆)

> I meant more generally though that it seems like self-contained app bundles could be generated for /Applications if brew itself was "vendored" inside them.

I guess I'm confused as to why Homebrew would be involved/required here at all. Anyway, vaguely off topic but interesting nonetheless.

ilovezfs:
>(while investigating this thread I ran the system one and had to kill it by closing the Terminal. It may still be running there in the background somewhere :laughing:)

I'm glad I'm not alone.

>I guess I'm confused as to why Homebrew would be involved/required here at all. Anyway, vaguely off topic but interesting nonetheless.

Only as an upgrade mechanism, if that wasn't clear.

To return to the topic, now that I understand what you mean by not "self-contained" (i.e., cannot be copy-pasted into /Applications and live to tell the tale), I think your position that "say nothing" is the right choice sounds reasonable.

DomT4:
> someone who actually knows what magical incantation is needed to exit emacs :)

I execute `kill emacs` using Alfred 😆.

ilovezfs:
>@ilovezfs Maybe some site like emacsforosx.com could do that and have a binary that could be distributed through Homebrew Cask :wink: 

I totally think @zmwangx should volunteer to do this.

zmwangx:
> This makes me want to build them statically and/or bury an entire mini brew installation inside them.

If you build and install Emacs `--with-ns` but without `--disable-ns-self-contained`, the resulting structure would be

```
$ tree -L3 Emacs.app
Emacs.app
└── Contents
    ├── MacOS
    │   ├── bin
    │   └── libexec
    └── Resources
        ├── etc
        ├── info
        ├── lisp
        ├── man
        ├── share
        └── var

11 directories, 0 files
```

So it looks like merely a custom set of `--bindir`, `--libexecdir`, `--infodir`, `--mandir`, `--datadir` and such, and it makes me think that you may be able to embed other libs in there. But changes to the build process are apparently necessary which is probably undesirable, and if David Caldwell (maintainer of emacsforosx.com) can't figure it out (see https://github.com/caldwell/build-emacs/issues/1), I probably won't either. Moreover, the binary will likely be huge if you include all the optional dependencies.

```
$ brew deps --recursive --include-optional emacs | wc -l
63
```

> I shall leave that to someone who actually knows what magical incantation is needed to exit emacs :)

Can't believe a vi user shooting emacs like this... You know, novices quit vi but powering down their computers.

ilovezfs:
>Moreover, the binary will likely be huge if you include all the optional dependencies.

If Cask supported delta updates, that wouldn't in itself be a problem, would it?

zmwangx:
> If Cask supported delta updates, that wouldn't in itself be a problem, would it?

Not sure I understand. How do you delta update a dmg downloaded from the Internet?

ilovezfs:
A pkg can either be a full installer or an upgrade installer, where the latter only needs the diff.

zmwangx:
> A pkg can either be a full installer or an upgrade installer, where the latter only needs the diff.

Right, but upstream needs to provide incremental pkg installers (probably won't happen), and AFAIK `installer` can only install to the root of a volume, meaning you can't place the app anywhere, right? (Unless maybe you install to an empty volume and then manually relocate the files.)

ilovezfs:
>Right, but upstream needs to provide incremental pkg installers

Right.

>AFAIK installer can only install to the root of a volume, meaning you can't place the app anywhere

It can be restricted to system volume, or any target volume, or to a user directory. Typically, user directory is a horribly broken approach, and any target volume (as opposed to just the live system volume) is fine.

In practice, hard coded /Applications/foo.app is usually fine. I believe there's also some support for designating whether a bundle is relocatable.

zmwangx:
> It can be restricted to system volume, or any target volume, or to a user directory.

I know it can be any target volume, but I didn't know `-target` can be a user directory. `man 1 installer` doesn't seem to say that, and I just tried that with MacTeX which resulted in an error:

```
$ installer -verbose -package /Volumes/ext/software/misc/mactex/mactex-20160603.pkg -target /tmp
installer: Error trying to locate volume at /tmp
```

> In practice, hard coded /Applications/foo.app is usually fine.

But you just violated the "relocatable" principle 😉 And emacsforosx.com is supposed to be a "To install, drag this icon (to anywhere)..." app.

ilovezfs:
Using the user directory has to specifically be enabled in the plist file embedded in the pkg. You can't just do it with anything.

As for relocatibility, I believe that can work with upgrades and delta upgrades if properly configured with rpath, etc. and relocatibility is toggled on in the pkg plist. I've just not played around with that feature personally.

I don't even think it would be totally crazy to hard code the paths for an app bundle that only works if located in /Applications or go a step further and have hard-coded internal paths pointing into /usr/local/Cellar assuming it worked well.

From an ease of upgrade perspective I think being able to use Homebrew bottles including non-relocatable ones would be nice. It would be interesting to know how many of the full set of optional and recommended formula deps for emacs have relocatable bottles vs non-relocatable bottles.

DomT4:
> In practice, hard coded /Applications/foo.app is usually fine. 

I believe from macOS 10.12 applications aren't actually going to live in `/Applications`, even though they'll still, er, exist in `/Applications`. I think Apple is randomising the path or something, so it's not entirely predictable. I was only half paying attention when this was discussed on the WWDC stream.

zmwangx:
> Using the user directory has to specifically be enabled in the plist file embedded in the pkg. You can't just do it with anything.

I see.

> I think Apple is randomising the path or something, so it's not entirely predictable. I was only half paying attention when this was discussed on the WWDC stream.

Which session is it, security? (Now that you mention it, I realized I haven't watched the security session yet.)

DomT4:
Yeah, "Gatekeeper Path Randomization". From [30:22 in this video](https://developer.apple.com/videos/play/wwdc2016/706). There's a blog post [here](http://weblog.rogueamoeba.com/2016/06/29/sierra-and-gatekeeper-path-randomization/).

DomT4:
This bit is interesting though. I hadn't read this from any official Apple source before, although it could well be in the linked video & I'd not picked it up half-listening previously:

> Apple may not view this as an issue, given that GPR will be disabled once the user moves the application out of the Downloads folder. However, many users run applications from the Downloads folder, never moving them.

DomT4:
The video possibly raises more questions than it answers about this security feature _(true to form from Apple)_ but _"If the user explicitly moves the single app by itself, possibly to applications, this mechanism will be turned off"_ is the big line probably interesting to us.

I hope by that they mean there doesn't have to be _user involvement_ in physically moving the `.app` & programmatic moving of `.app`s isn't going to suddenly start being problematic 😕. Signed disk images, MAS apps & signed Apple installer packages are exempt as well.

zmwangx:
While this GPR is interesting and confusing, I doubt it will affect any of our command line workflows. Gatekeeper relies on `com.apple.quarantine` and optionally `com.apple.metadata:kMDItemWhereFroms` to determine if an app is downloaded, unzipped, etc. `curl`, `tar`, `unzip` etc. don't set those xattrs, so there's no way Gatekeeper could tell that the app was downloaded or built from source or whatever.

As an example, I just tried to unzip `AppCleaner-3.3.zip`. When I unzipped it with The Unarchiver (GUI app) and tried to launch the app, I got a prompt "'AppCleaner.app' is an application created by The Unarchiver. Are you sure you want to open it?" When I `unzip` on the command line instead, the app just launched with no Gatekeeper BS.

And just for fun, I also tarred up `/usr/local/opt/emacs/Emacs.app` and untarred with The Unarchiver. This time it couldn't be opened because it wasn't signed... Interestingly enough, Archive Utility doesn't set `com.apple.quarantine` so the resulting app wasn't subject to Gatekeeper, and it opened up just fine.

DomT4:
Potentially worth tying into this discussion is the Cask discussion on setting Gatekeeper on downloads to close a potential security hole: https://github.com/caskroom/homebrew-cask/issues/22388

> EDIT. I'll add that the experiments were done on 10.11.6.

I've got a 10.12 VM. I'll try duplicating your experiments later there, but I agree with your summary that if nothing changes with us or Cask it should be alright, at least for now; Apple could well expand or pull back on the changes before 10.12 goes stable.

zmwangx:
I find manually enabling Gatekeeper (which from my experience isn't so helpful to begin with), but I'll read the Cask discussion later to see where people are coming from.

I removed my 10.12 VM yesterday to save some ~30G of space so I can't test anything now...

DomT4:
I sympathise with the Cask team, although they haven't set out their argument as far as I can see, in that users are likely to invest a certain amount of trust in them & may presume they offer no less protection* than Apple, and consequently may adopt less cautious behaviour assuming that protection is there.

It's not entirely fair, but I wouldn't be surprised if that's how they've been left feeling at times. The weight of knowing so many people trust you isn't something that is particularly easy to carry.

*_Protection is subjective._

ilovezfs:
Let's just collectively agree that whenever a Homebrew user uses Cask in a way that requires `sudo` that he or she is a Cask user, not a Homebrew user, with respect to that action and its consequences, regardless of whether `brew` was part of the original command line invocation or not.

CamJN:
Since no one has mentioned it, I'd like to point out that `--with-ns` actually enables better colour support when run in the terminal (ie run `emacs` with the `-nw` flag), there was (might still be) a patch that you only get with that flag that prevents emacs hanging in some cases, and in general it is used as a bit of a "mac stuff here" flag in addition to "build a mac gui". 

Believe me, I wouldn't build the gui if it wasn't worth it, I have to go to great lengths to make sure I don't accidentally launch the gui by mistake. But the improvements that come with it aren't something I'm willing to give up.

Also to those whinging about quitting Emacs: `Ctrl-x Ctrl-c` is way easier than hacking through my power cable with an axe and lighting my computer on fire which is I believe the official way to quit vim 😛.

ilovezfs:
> --with-ns actually enables better colour support when run in the terminal

That alone sounds like a sufficient reason to make it the default.

ilovezfs:
If GUI should not be default, it sounds like we should make the `cocoa` option default enabled, but have an env script wrapper that sets `-tw`.

zmwangx:
@ilovezfs What's `-tw`? You mean `-nw`? There's a problem with this approach: `-nw` is `--no-window-system`, and there's no `--no-no-window-system` option.

ilovezfs:
>@ilovezfs What's -tw? You mean -nw? There's a problem with this approach: -nw is --no-window-system, and there's no --no-no-window-system option.

Yes, `-nw`, sorry.

>There's a problem with this approach: -nw is --no-window-system, and there's no --no-no-window-system option.

We could teach the wrapper script to accept `--window-system` and `--no-no-window-system`, right? But probably what would be more straightforward is to have a separate script for the `exec #{prefix}/Emacs.app/Contents/MacOS/Emacs "$@"` behavior.

zmwangx:
> We could teach the wrapper script to accept --window-system and --no-no-window-system, right? But probably what would be more straightforward is to have a separate script for the exec #{prefix}/Emacs.app/Contents/MacOS/Emacs "$@" behavior.

Both sounds wrong to me, because I don't think we are in the position to offer additional binaries or otherwise non-existent options. Also, the situation is a tad more complicated for `emacsclient`.

Overall I don't think defaulting `--with-cocoa` is a good idea.

ilovezfs:
>Overall I don't think defaulting --with-cocoa is a good idea.

I'll leave that between you and @CamJN, then ;)

MikeMcQuaid:
If we leave GUIs in here: we're defaulting to non-GUI options where possible.

ilovezfs:
It seems like this is one where we're going to end up with both the formula and the cask in perpetuity, which may be fine, I guess.

DomT4:
I think _not_ defaulting to a GUI is sane, personally. I think we should keep the GUI option for `emacs`, but I think keeping it as an option is a reasonable compromise that everyone will be a little annoyed by but nobody will be completely outraged by.

> Also to those whinging about quitting Emacs: Ctrl-x Ctrl-c

Oh, how intuitive 😈 😉

ilovezfs:
@DomT4 do you agree that the Cask should continue to exist as well in this case?

zmwangx:
> do you agree that the Cask should continue to exist as well in this case?

Cask does provide a better macOS app experience for people who don't need the optional deps (ignoring the upgrade issue, which isn't much of a problem for stable users since the stable channel is seldom updated, just like how Apple updates Macs these days).

DomT4:
> @DomT4 do you agree that the Cask should continue to exist as well in this case?

I don't have major beef with the Cask continuing to live. I think it & our `--with-cocoa` option basically do the same thing in slightly different ways, but both are just as valid. I don't expect that to be the case with most things, but it seems to be so here.

MikeMcQuaid:
Passing on this for now as per discussion. Thanks again for opening @AnastasiaSulyagina. Have opened a PR with various tweaks discussed here in https://github.com/Homebrew/homebrew-core/pull/3531.

julienchastang:
If anyone is missing the X11 formula, [here it is](https://github.com/julienchastang/homebrew-core/tree/x11emacs).

PR #3331, opened 2016-07-26
gedit migrated to cask

AnastasiaSulyagina:

---

Deleting gedit from homebrew-core as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


DomT4:
I'm not massive fan of `gedit` so if it goes, so be it, but if we get rid of it I think it should be boneyarded rather than migrated. Our version is `3.20.2` and the last upstream `.dmg` release for OS X was... `gedit-3.2.6-3.dmg` over 4 years ago now.

Might as well call a duck a duck & accept `gedit` aren't really providing OS X binaries now.

MikeMcQuaid:
> Might as well call a duck a duck & accept gedit aren't really providing OS X binaries now.

Agreed; makes sense to make this migration the other way and migrate the Cask into Homebrew (despite my 😭 about supporting a GUI package). @vitorgalvao, thoughts?

vitorgalvao:
Agreed with removing from HBC and handing it over to HB. Another case like [`mpv`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mpv.rb). A shame, but it’s better than having users download a perpetually outdated package.

Removing in https://github.com/caskroom/homebrew-cask/pull/23360.

PR #3334, opened 2016-07-26
idris 0.12.1

ilovezfs:
- remove upstreamed GHC 8 patch
- remove the global cabal.config file since the new version of the
  trifecta dependency already has the relevant constraints bumped


DomT4:
👍

PR #3337, opened 2016-07-26
macvim migrated to cask

AnastasiaSulyagina:

---

Deleting macvim from homebrew-core as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


tdsmith:
This will make it harder to obtain macvim built against a non-system Python. macvim has proven to be a troublesome but popular package for us.

MikeMcQuaid:
@tdsmith It's maybe worth us contacting upstream about the desire for this?

chdiza:
Not sure upstream will be sympathetic.  MacVim has never consistently played nice with non-system pythons at any point in its entire existence (2007, IIRC).  If memory serves, this isn't the fault of the upstream devs, either current or former.

MikeMcQuaid:
Passing on this for now but thanks for the PR @AnastasiaSulyagina!

PR #3339, opened 2016-07-26
quassel migrated to cask

AnastasiaSulyagina:

---

Deleting quassel from homebrew-core as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


DomT4:
There might be a legitimate reason to want to use CLI instead of GUI here, but according to our analytics nobody has actually installed this formula since we went live with those, so 👍 from me.

MikeMcQuaid:
Thanks again @AnastasiaSulyagina! Check out the new commit message format I used for any future ones (`quassel: migrate to Homebrew Cask.`).

PR #3341, opened 2016-07-26
wireshark migrated to cask

AnastasiaSulyagina:

---

Deleting wireshark from homebrew-core as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 


DomT4:
Ref: https://github.com/Homebrew/homebrew-core/issues/2260#issuecomment-227589174

tdsmith:
I understand that wireshark provides useful CLI tools and I think Homebrew can usefully continue to provide those; I don't have strong feelings about whether we retain the GUI.

MikeMcQuaid:
Agreed, this is another case where I think we could deprecate the GUI and leave the CLI around.

PR #3351, opened 2016-07-27
jrnl: convert to virtualenv installation

tdsmith:


ilovezfs:
@MikeMcQuaid any thoughts on an acceptable standard commit title for these? Both Tim's "foo: convert to virtualenv installation" and my "foo: use Language::Python::Virtualenv" seem a tad verbose, but I'd be happy with either or some alternative. It would be nice if they were all the same though (except of course when the switch is combined in one commit with a formula version bump whose bump subject takes precedence).

tdsmith:
> acceptable standard commit title

It's okay if we don't standardize this, right?

ilovezfs:
No rebellious commit subjects for you!

MikeMcQuaid:
`jrml: use new virtualenv feature.'`seems good. Standardising seems good if we're going to do a bunch of these so we can easily search.

PR #3352, opened 2016-07-27
awscli: use new virtualenv feature

ilovezfs:


tdsmith:
@BrewTestBot test this please

CamJN:
Does this mean having awscli installed won't conflict with having boto3 installed in the system python anymore?

tdsmith:
Dunno how they conflicted before but they should definitely be isolated now.

PR #3365, opened 2016-07-27
nfdump 1.6.15

Pristavkin:
[Yes ] Have you followed the guidelines in our Contributing document?
[Yes ] Have you checked to ensure there aren't other open Pull Requests for the same formula update/change?
[Yes ] Have you built your formula locally prior to submission with brew install (where is the name of the formula you're submitting)?
[Yes ] Does your submission pass brew audit --strict --online (after doing brew install )?

As I know developing of nfdump moved to github, so I've updated formula to new version and new source.

### DIFF
--- a/Formula/nfdump.rb
+++ b/Formula/nfdump.rb
@@ -1,9 +1,8 @@
 class Nfdump < Formula
   desc "Tools to collect and process netflow data on the command-line"
   homepage "http://nfdump.sourceforge.net"
-  url "https://github.com/phaag/nfdump.git",
-    :tag => "v1.6.15",
-    :revision => "68d660e984e8b6ce099fc7309f61f3bfa460598f"
+  url "https://github.com/phaag/nfdump/archive/v1.6.15.tar.gz"
+  sha256 "9505c0511d273b9aa3f87a5e664425689a3c7370c6ae3bbc05ff4bdb41bfd457"
 
   bottle do
     cellar :any_skip_relocation
@@ -17,7 +16,7 @@ class Nfdump < Formula
 
   def install
     system "./configure", "--prefix=#{prefix}", "--enable-readpcap"
-    system "make", "-j1", "install"
+    ENV.deparallelize { system "make", "install" }
   end
 
   test do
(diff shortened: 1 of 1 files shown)


Formula/nfdump.rb
```diff
@@ -1,8 +1,9 @@
 class Nfdump < Formula
-  desc "Tools to collect and process netflow data on the command line"
+  desc "Tools to collect and process netflow data on the command-line"
   homepage "http://nfdump.sourceforge.net"
-  url "https://downloads.sourceforge.net/project/nfdump/stable/nfdump-1.6.13/nfdump-1.6.13.tar.gz"
-  sha256 "251533c316c9fe595312f477cdb051e9c667517f49fb7ac5b432495730e45693"
+  url "https://github.com/phaag/nfdump.git",
```
DomT4: Can we not use the tarball here?
```ruby
url "https://github.com/phaag/nfdump/archive/v1.6.15.tar.gz"
sha256 ""
```
etc. Does that break?

Formula/nfdump.rb
```diff
@@ -12,9 +13,11 @@ class Nfdump < Formula
     sha256 "09a51a71d30f8045075f11f8511a4ec0d6bfd6f9d3cbc3bbb2927c911fce1ce1" => :mavericks
   end
 
+  depends_on "automake" => :build
+
   def install
     system "./configure", "--prefix=#{prefix}", "--enable-readpcap"
-    system "make", "install"
+    system "make", "-j1", "install"
```
DomT4: Don't hardcode `-j1`. If it can't build parallelised do:
```ruby
ENV.deparallelize { system "make", "install" }
```
That would ideally be reported upstream as well. Things should build in parallel.

Formula/nfdump.rb
Pristavkin: I'm not sure I understand you clearly. I didn't use tarball here, and in my opinion that's good idea - when new release will be pushed to github, the only thing we'll need to do is to change release tag. Are you ok with it?

DomT4: We prefer to use the tarball. It's a slightly quicker operation than a git checkout, and it can be protected with a stronger checksum than git's native SHA1.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/b358ee9c3db207ced93860a4023f29e8ee7b282f. Thank you for your contribution to Homebrew @Pristavkin; we appreciate it! :smiley_cat:

PR #3367, opened 2016-07-27
ngspice: add option to build shared library.

jmaibaum:

---

Add --with-shared option to ngspice.  It might look a little hacky, but
it only works this way, i.e. build two times in a row, with different
configure args.  Arch Linux does it the same way.

### DIFF
--- a/Formula/ngspice.rb
+++ b/Formula/ngspice.rb
@@ -11,6 +11,7 @@ class Ngspice < Formula
   end
 
   option "without-xspice", "Build without x-spice extensions"
+  option "with-shared", "Build shared library"
 
   deprecated_option "with-x" => "with-x11"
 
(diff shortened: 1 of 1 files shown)


jmaibaum:
As there has not been any reaction to this PR yet, I just wanted to ask if there is anything I could and should do to get it merged?

dunn:
Most (or at least much) software doesn't need a second compile for shared support, so this seems like a bug.  Could you ask the developers about it?

DomT4:
If this is a wontfix upstream we're unlikely to take it here, FWIW 😕. It's a bit of a hack & something that really should be supported upstream.

jmaibaum:
There is a ticket for this upstream (https://sourceforge.net/p/ngspice/support-requests/19), but they say it can't be fixed and one should not call it a bug, rather a wished feature that they won't support.
At least they say, that the two-builds-in-a-row way is supported, i.e. they do not consider it a hack.
But if it is considered a hack here and that you won't merge it, it's OK for me.

MikeMcQuaid:
Thanks for the PR anyway, @jmaibaum!

jmaibaum:
Just a short question: would in this case a separate formula that just builds the library, i.e. something like "libngspice", be acceptable for inclusion into Homebrew?

MikeMcQuaid:
@jmaibaum I'd be 🆗 with that but you might want to see if other maintainers 👍/ 👎 before putting time in.

DomT4:
> But if it is considered a hack here and that you won't merge it, it's OK for me.

Yeah, there's evidently a difference of opinion between upstream & us. It happens 😄.

> would in this case a separate formula that just builds the library, i.e. something like "libngspice", be acceptable for inclusion into Homebrew?

I'd probably be fine with this. Wouldn't necessarily want it to become a regular thing but I agree it could well make sense in this case.

PR #3387, opened 2016-07-28
ibex 2.3.1

ilovezfs:
remove upstreamed patch
defer building the examples until the test to fix linkage audit error


apjanke:
@BrewTestBot test this please

ilovezfs:
@apjanke the problem is that some of the installed test files have linkage to files the build path temporary directory. So brew audit will fail since the linked files do not exist.

ilovezfs:
@apjanke it should be fixed now :)

PR #3427, opened 2016-07-29
bazel 0.3.1

damienmg:
- [ X ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [ X ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ X ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [ X ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

Small improvements along version bumping:
- Use BAZEL_WRKDIR to put all generated file in the Homebrew cache
- Install the wrapper script along with Bazel

### DIFF
--- a/Formula/bazel.rb
+++ b/Formula/bazel.rb
@@ -1,8 +1,8 @@
 class Bazel < Formula
   desc "Google's own build tool"
   homepage "http://www.bazel.io/"
-  url "https://github.com/bazelbuild/bazel/archive/0.3.0.tar.gz"
-  sha256 "d2309c29781dc4ede79eb652776517ca3c1dff29b58849dc80bdb86031cce3ad"
+  url "https://github.com/bazelbuild/bazel/archive/0.3.1.tar.gz"
+  sha256 "52beafc9d78fc315115226f31425e21df1714d96c7dfcdeeb02306e2fe028dd8"
   head "https://github.com/bazelbuild/bazel.git"
 
   bottle do
@@ -16,11 +16,14 @@ class Bazel < Formula
 
   def install
     ENV["EMBED_LABEL"] = "#{version}-homebrew"
+    # Force Bazel to put all temporary file in the homebrew cache
+    ENV["BAZEL_WRKDIR"] = "#{HOMEBREW_CACHE}/bazel-workdir"
 
     system "./compile.sh"
     system "./output/bazel", "build", "scripts:bash_completion"
 
-    bin.install "output/bazel" => "bazel"
+    bin.install "scripts/packages/bazel.sh" => "bazel"
+    bin.install "output/bazel" => "bazel-real"
     bash_completion.install "bazel-bin/scripts/bazel-complete.bash"
     zsh_completion.install "scripts/zsh_completion/_bazel"
   end
--- a/Formula/bazel.rb
+++ b/Formula/bazel.rb
@@ -16,8 +16,8 @@ class Bazel < Formula
 
   def install
     ENV["EMBED_LABEL"] = "#{version}-homebrew"
-    # Force Bazel to put all temporary file in the homebrew cache
-    ENV["BAZEL_WRKDIR"] = "#{buildpath}"
+    # Force Bazel ./compile.sh to put its temporary file in the homebrew cache
+    ENV["BAZEL_WRKDIR"] = "#{buildpath}/work"
 
     system "./compile.sh"
     system "./output/bazel", "--output_user_root", "#{buildpath}/output_user_root", "build", "scripts:bash_completion"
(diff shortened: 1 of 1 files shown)


Formula/bazel.rb
```diff
@@ -16,11 +16,14 @@ class Bazel < Formula
 
   def install
     ENV["EMBED_LABEL"] = "#{version}-homebrew"
+    # Force Bazel to put all temporary file in the homebrew cache
+    ENV["BAZEL_WRKDIR"] = "#{HOMEBREW_CACHE}/bazel-workdir"
```
ilovezfs: Are these files reusable from one build of bazel to the next, or should these files be in the Homebrew temporary directory?

UniqMartin: No matter what the answer is, we don't expect formulae to write directly to the cache and we don't have code that would be capable of cleaning up arbitrary directories in the Homebrew cache. Thus this will need to be changed to something else. IMO, this should be always somewhere in `buildpath` where it will be cleaned up once the build concludes.

UniqMartin: I'm also not sure whether this actually works. I just checked our El Capitan build bot and there are still 1.2 GB that have been dumped to `/var/tmp/_bazel_brew` (I don't think all of it can be attributed to #3426). `~/Library/Caches/Homebrew/bazel-workdir` has been created, but it only contains two completely empty subdirectories …

damienmg: The directory should be cleaned after execution yes.

Let me retest

damienmg: Ok that's because of the bazel run underneath for generating the bazel-complete.bash. We need to fix it upstream.

damienmg:
I  added another commit.

There is one remaining issue: each bazel run use a cache folder and the last command actually call bazel from bazel so the inner call write that cache directory, we should fix that upstream.

Formula/bazel.rb
```diff
@@ -16,11 +16,14 @@ class Bazel < Formula
 
   def install
     ENV["EMBED_LABEL"] = "#{version}-homebrew"
+    # Force Bazel to put all temporary file in the homebrew cache
```
UniqMartin: I guess you could tweak this comment a bit in the next iteration, as it no longer exactly matches what is happening.

Formula/bazel.rb
```diff
@@ -16,11 +16,14 @@ class Bazel < Formula
...
+    ENV["BAZEL_WRKDIR"] = "#{buildpath}"
```
UniqMartin: I'm not entirely sure how the contents of `BAZEL_WRKDIR` are structured, but it could make sense to confine them all to a subdirectory, e.g. `#{buildpath}/work`, as `#{buildpath}` is also the location where the source tarball gets extracted to.

UniqMartin:
@damienmg Thanks on working to rectify this! I hope I'm not being too difficult “complaining” about this, but hopefully you agree that `bazel` is a bit atypical in where it puts its temporary build data and it's somewhat unhelpful that the default on macOS happens to be a location that's neither `/tmp` nor `$TMPDIR`, thus isn't periodically cleaned by the OS.

The latest changes make the build finish with only 200 MB being dumped to `/var/tmp/_bazel_${USER}`. We're getting closer to the goal of not dumping anything at all in there. :+1:

Formula/bazel.rb
damienmg: Done. Technically ./compile.sh create 2 subdir in BAZEL_WRKDIR but it is indeed cleaner to put it in a subdirectory.

damienmg:
No problem, leading an OSS project myself and I know what it is like :)

I will fix the remaining issue for the /var/tmp/_bazel_${USER} files upstream.

damienmg:
Last upstream fix is at https://bazel-review.googlesource.com/4221

dfabulich:
@damienmg That upstream fix merged, ready to go out in 0.3.2 or 0.4.0 or whatever, but that leaves this PR in limbo. Can we merge this 0.3.1 PR today?

damienmg:
Yes I think this PR should be merged I was waiting for it to go in.

damienmg:
@UniqMartin: can this PR be merge and the remaining of the tmp issue will be fixed on 0.3.2 release of Bazel (so next update).

ilovezfs:
@damienmg that sounds fine. I've triggered a new bottle run and then this can ship.

ilovezfs:
@damienmg :ship:'d Thanks for your work on this and your contribution to Homebrew :tada: 

Looking forward to 0.3.2!

PR #3357, opened 2016-07-27
luyten 0.4.8

ilovezfs:
Created with `brew bump-formula-pr`.

PR #3428, opened 2016-07-29
luyten: migrate to gui

ilovezfs:
See discussion here: https://github.com/Homebrew/homebrew-core/pull/3357#discussion_r72778372

CC @UniqMartin @DomT4


UniqMartin:
:+1:

> CC @UniqMartin @DomT4

We don't get a notification if you retroactively add this. :wink: (But maybe this was intentional?)

DomT4:
I kind of hate migrating things to taps other than the 💀-yard these days given how badly looked after taps have become, but if this is untestable in a sane way I guess this is fine.

ilovezfs:
>We don't get a notification if you retroactively add this. :wink: (But maybe this was intentional?)

@UniqMartin Ug. No, not intentional. That's somewhat annoying.

>I kind of hate migrating things to taps other than the :skull:-yard these days given how badly looked after taps have become, but if this is untestable in a sane way I guess this is fine.

@DomT4 The best we can do, I think, is verifying either there are no exceptions or there is a HeadlessException and an exit status of 0. I'm happy to throw that in if you'd prefer to keep this in core.

DomT4:
It's entirely GUI based, I think it's fair enough to relocate it, just anxious about us seemingly using taps more often than not to ditch software we don't love very much & it then gets universally ignored. Not a criticism of you or this PR; a broader concern I've voiced quite a few times by now.

ilovezfs:
@DomT4 Have you seen https://en.wikipedia.org/wiki/It's_a_Good_Life_%28The_Twilight_Zone%29

It definitely does feel like we're wishing them into a cornfield.

DomT4:
😄 It's alright, some of our formulae are already _"grotesque, walking horror"s_ 😉

ilovezfs:
>some

understatement of the year?

PR #3433, opened 2016-07-29
mpd: test for crash on HEAD builds

jacobvosmaer:

---

As of homebrew-core 69acd35e30dbf14ec798f41c879d426e1f4ea014 and
mpd 2777a23672dcea12f1b9abf2c137eab987660ece I have a reproducable
crashing bug that occurs only when I `brew install mpd --HEAD` but
not if I do `brew install mpd`. This commit adds a test that makes
MPD crash after being compiled from source at HEAD on my system.

### DIFF
--- a/Formula/mpd.rb
+++ b/Formula/mpd.rb
@@ -132,6 +132,7 @@ class Mpd < Formula
 
     begin
       assert_match "OK MPD", shell_output("curl localhost:6600")
+      assert_match "ACK", shell_output("(sleep 1; echo playid foo) | nc localhost 6600")
     ensure
       Process.kill "SIGINT", pid
       Process.wait pid


jacobvosmaer:
On my machine the test I added in this PR fails:

```
$ brew test mpd
Testing jacobvosmaer/core/mpd
==> Using the sandbox
server_socket: bind to '0.0.0.0:6600' failed: Address already in use (continuing anyway, because binding to '[::]:6600' succeeded)
output: No 'AudioOutput' defined in config file
output: Attempt to detect audio output device
output: Attempting to detect a osx audio device
output: Successfully detected a osx audio device
==> curl localhost:6600
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0client: [0] opened from [::1]:51323
client: [0] closed
100    99    0    99    0     0  16242      0 --:--:-- --:--:-- --:--:-- 19800
==> (sleep 1; echo playid foo) | nc localhost 6600
client: [1] opened from [::1]:51324
libc++abi.dylib: terminating with uncaught exception of type ProtocolError: Integer expected: foo
Error: jacobvosmaer/core/mpd: failed
Expected /ACK/ to match "OK MPD 0.20.0\n".
```

So there must be something wrong with my machine. As long as I can install a bottle, instead of compiling locally, I am fine. But this test might be worth adding to the formula.

jacobvosmaer:
I originally raised this issue with MPD itself: https://bugs.musicpd.org/view.php?id=4556 but that got closed as unreproducable.

DomT4:
> So there must be something wrong with my machine. 

Not necessarily, unless I'm misunderstanding your original post. CI doesn't test `HEAD` because `HEAD` by its very nature is unstable, unpredictable & is something we have minimal ability to fix when it breaks.

It's also entirely possible that the error hasn't yet made it into the stable build, which is what you'll get when you execute a vanilla `brew install mpd`. Unless you're using `mpd` incorrectly, which I can't judge because I'm not a user of the software myself, your upstream bug report seems valid - I can reproduce it with the `--HEAD` build:

```
~> mpd --stdout --no-daemon --no-config
~> sleep 1; echo playid foo | nc localhost 6600
OK MPD 0.20.0
libc++abi.dylib: terminating with uncaught exception of type ProtocolError: Integer expected: foo
[1]    6683 abort      mpd --stdout --no-daemon --no-config
```

This is the full `lldb` backtrace:

```
libc++abi.dylib: terminating with uncaught exception of type ProtocolError: Integer expected: foo
Process 6901 stopped
* thread #3: tid = 0x71d45e, 0x00007fff92328f06 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.root.default-qos', stop reason = signal SIGABRT
    frame #0: 0x00007fff92328f06 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff92328f06 <+10>: jae    0x7fff92328f10            ; <+20>
    0x7fff92328f08 <+12>: movq   %rax, %rdi
    0x7fff92328f0b <+15>: jmp    0x7fff923237cd            ; cerror_nocancel
    0x7fff92328f10 <+20>: retq
(lldb) bt
* thread #3: tid = 0x71d45e, 0x00007fff92328f06 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.root.default-qos', stop reason = signal SIGABRT
  * frame #0: 0x00007fff92328f06 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff9aea24ec libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fff914216df libsystem_c.dylib`abort + 129
    frame #3: 0x00007fff8e8d7c11 libc++abi.dylib`abort_message + 257
    frame #4: 0x00007fff8e8fddff libc++abi.dylib`default_terminate_handler() + 243
    frame #5: 0x00007fff9e8706c3 libobjc.A.dylib`_objc_terminate() + 124
    frame #6: 0x00007fff8e8fb00e libc++abi.dylib`std::__terminate(void (*)()) + 8
    frame #7: 0x00007fff8e8fb8df libc++abi.dylib`__cxxabiv1::call_terminate(bool, _Unwind_Exception*) + 35
    frame #8: 0x00007fff8e8fb60d libc++abi.dylib`__cxxabiv1::scan_eh_tab(__cxxabiv1::(anonymous namespace)::scan_results&, _Unwind_Action, bool, _Unwind_Exception*, _Unwind_Context*) + 582
    frame #9: 0x00007fff8e8fb32e libc++abi.dylib`__gxx_personality_v0 + 178
    frame #10: 0x00007fff917119ba libunwind.dylib`_Unwind_RaiseException + 162
    frame #11: 0x00007fff8e8faa69 libc++abi.dylib`__cxa_throw + 104
    frame #12: 0x000000010000b867 mpd`ParseCommandArgInt(char const*, int, int) + 279
    frame #13: 0x000000010000d20d mpd`handle_playid(Client&, Request, Response&) + 38
    frame #14: 0x000000010000c46d mpd`command_process(Client&, unsigned int, char*) + 926
    frame #15: 0x0000000100012c63 mpd`client_process_line(Client&, char*) + 327
    frame #16: 0x0000000100012ee7 mpd`Client::OnSocketInput(void*, unsigned long) + 105
    frame #17: 0x000000010005137d mpd`BufferedSocket::ResumeInput() + 43
    frame #18: 0x0000000100051455 mpd`BufferedSocket::OnSocketReady(unsigned int) + 45
    frame #19: 0x000000010005176b mpd`FullyBufferedSocket::OnSocketReady(unsigned int) + 39
    frame #20: 0x0000000100051106 mpd`SocketMonitor::Dispatch(unsigned int) + 22
    frame #21: 0x0000000100052d31 mpd`EventLoop::Run() + 405
    frame #22: 0x00000001000071e2 mpd`___Z8mpd_mainiPPc_block_invoke + 1563
    frame #23: 0x00007fff922d493d libdispatch.dylib`_dispatch_call_block_and_release + 12
    frame #24: 0x00007fff922c940b libdispatch.dylib`_dispatch_client_callout + 8
    frame #25: 0x00007fff922cd29b libdispatch.dylib`_dispatch_root_queue_drain + 1890
    frame #26: 0x00007fff922ccb00 libdispatch.dylib`_dispatch_worker_thread3 + 91
    frame #27: 0x00007fff9ae9f4de libsystem_pthread.dylib`_pthread_wqthread + 1129
    frame #28: 0x00007fff9ae9d341 libsystem_pthread.dylib`start_wqthread + 13
```

jacobvosmaer:
Thanks for taking the time to look into this @DomT4 ! Good to know that it is not just my machine that breaks things.

> Unless you're using mpd incorrectly

No this is not the case. The test makes an invalid request to MPD but that should lead to an error response (a line starting with `ACK`), never to a crash.

As I wrote the [upstream bug](https://bugs.musicpd.org/view.php?id=4556) got closed but I believe the maintainer does not have access to an OS X build environment. So this may be something OS X-specific?

jacobvosmaer:
I wonder if this is a clang vs gcc issue.

DomT4:
It could be. It's not limited to Apple's `clang` because building `mpd` against Homebrew's `llvm` formula, which includes `clang`, doesn't produce any better an outcome:
```
~> otool -L `which mpd`
...
/usr/local/opt/llvm/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
```

```
~> brew test -v mpd
server_socket: bind to '0.0.0.0:6600' failed: Address already in use (continuing anyway, because binding to '[::]:6600' succeeded)
output: No 'AudioOutput' defined in config file
output: Attempt to detect audio output device
output: Attempting to detect a osx audio device
output: Successfully detected a osx audio device
==> curl localhost:6600
  0     0   client: [0] opened from [::1]:64517
 0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0client: [0] closed
100    99    0    99    0     0  38003      0 --:--:-- --:--:-- --:--:-- 49500
==> (sleep 1; echo playid foo) | nc localhost 6600
client: [1] opened from 127.0.0.1:64518
libc++abi.dylib: terminating with uncaught exception of type ProtocolError: Integer expected: foo
```

Unhelpfully, I can't actually get it to build against GCC-6 or GCC-5 on OS X, so hard to check for sure.

jacobvosmaer:
One thing that stood out in the diff from stable MPD to master is that it now requires GCC 4.7 (or later) instead of 4.6.

I know too little about C/C++ development to know if that is significant. And I know too little about Homebrew to know how to tweak compiler settings / choices. But I have not given up yet, I can compile MPD locally without having to go through Formula/mpd.rb and it reproduces the crash. So if I can just find the right `./configure` option or environment variable...

jacobvosmaer:
OK, gcc-6 from homebrew gives strange compile errors when compiling MPD.

```
/usr/local/bin/g++-6 -DHAVE_CONFIG_H -I.  -DNDEBUG -I./src -I/usr/local/include -DSYSTEM_CONFIG_FILE_LOCATION='"/usr/local/etc/mpd.conf"' -I/usr/local/Cellar/libao/1.2.0/include        -D_THREAD_SAFE -pthread -g -O2 -fvisibility=hidden -fno-threadsafe-statics -fmerge-all-constants -ffast-math -ftree-vectorize -ffunction-sections -fdata-sections -Wall -Wextra -Wmissing-declarations -Wshadow -Wpointer-arith -Wcast-qual -Wwrite-strings -Wsign-compare -MT src/output/plugins/liboutput_plugins_a-OSXOutputPlugin.o -MD -MP -MF src/output/plugins/.deps/liboutput_plugins_a-OSXOutputPlugin.Tpo -c -o src/output/plugins/liboutput_plugins_a-OSXOutputPlugin.o `test -f 'src/output/plugins/OSXOutputPlugin.cxx' || echo './'`src/output/plugins/OSXOutputPlugin.cxx
In file included from /System/Library/Frameworks/CoreAudio.framework/Headers/CoreAudio.h:19:0,
                 from src/output/plugins/OSXOutputPlugin.cxx:31:
/System/Library/Frameworks/CoreAudio.framework/Headers/AudioHardware.h:162:2: error: expected unqualified-id before '^' token
 (^AudioObjectPropertyListenerBlock)(    UInt32                              inNumberAddresses,
  ^
/System/Library/Frameworks/CoreAudio.framework/Headers/AudioHardware.h:162:2: error: expected ')' before '^' token
```

Is that what you were talking about @DomT4 ? I will try gcc-5 and gcc-47 next.

jacobvosmaer:
I might be misreading the errors but it feels like GCC 5/6 are choking on OS X framework headers.

jacobvosmaer:
Looks like we really want clang just to get the OS X code to compile.

This looks interesting: http://stackoverflow.com/questions/11913151/catching-derived-exceptions-types-fails-on-clang-macos-x

Something about clang not 'seeing' one error derives from another while GCC does. Wish I actually knew C++...

jacobvosmaer:
I am starting to understand what `ENV.libcxx` in the formula does. It tells clang to use libc++, which is the C++ standard library implementation native to LLVM. It seems we need this combination (clang++ and libc++) to get MPD to compile at all on OS X. But now on MPD master something changed that makes `catch (const std::exception &e)` not catch ProtocolError, even though ProtocolError derives from std::exception via std::runtime_exception.

jacobvosmaer:
Hey, I got something. Using llvm38 from Homebrew seems to fix it.

```
#!/bin/sh
CC=clang-3.8 \
  CXX="clang++-3.8 -stdlib=libc++"\
  CXXFLAGS="$CXXFLAGS -nostdinc++ -I/usr/local/opt/llvm38/lib/llvm-3.8/include/c++/v1"\
  LDFLAGS="$LDFLAGS -L/usr/local/opt/llvm38/lib/llvm-3.8/lib"\
  ./configure --enable-osx
```

```
% 	{sleep 1 ; echo playid foo } | nc localhost 6600
client: [0] opened from [::1]:49739
OK MPD 0.20.0
exception: Integer expected: foo
ACK [2@0] {} Integer expected: foo
client: [0] closed
```

Now the question is (and I don't know the answer!) how to tell homebrew to compile MPD using llvm38.

DomT4:
> Is that what you were talking about @DomT4 ?

Yeah, that's the failure I was running into.

> I am starting to understand what ENV.libcxx in the formula does. It tells clang to use libc++, which is the C++ standard library implementation native to LLVM.

It dumps `-stdlib=libc++` into the flags to force building against it, yup. We added this back in 2014 in https://github.com/Homebrew/homebrew-core/commit/52089d59, so has been hanging around for a while.

> Now the question is (and I don't know the answer!) how to tell homebrew to compile MPD using llvm38.

There's no great way of enforcing this, to be honest. There are ways, but not great ways. It's more problematic that this is broken against the system compiler. Even if we fix it in Homebrew, that doesn't cover over the broader problem that will start hitting non-brew OS X users of `mpd` eventually.

I'd recommend nudging the information you've gathered upstream, because this seems entirely legitimate & something that need to address if intend to support OS X/macOS beyond the current release.

jacobvosmaer:
@DomT4 maybe if I can get this bug to reproduce on Linux upstream can do something with it. It looks like I am currently the most active MPD contributor who uses OS X... :(

By the way I managed to get a non-crashing build using core/llvm instead of versions/llvm38.

```
CC=/usr/local/opt/llvm/bin/clang \
  CXX="/usr/local/opt/llvm/bin/clang++ -stdlib=libc++"\
  CXXFLAGS="$CXXFLAGS -nostdinc++ -I/usr/local/opt/llvm/include/c++/v1"\
  LDFLAGS="$LDFLAGS -L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib"\
  ./configure --enable-osx
```

Is that (a build that needs core/llvm) something we can work with on the Homebrew side?

jacobvosmaer:
Ran a bisect with:

```
#!/bin/sh
Run() {
  echo "Run: $*"
  if ! "$@" ; then
    exit 1
  fi
}

Run ./autogen.sh
Run /usr/bin/env CC=clang CXX=clang++  ./configure --enable-osx --disable-icu
Run make clean
make -j4 || exit 125

pkill mpd
./src/mpd --stderr
sleep 5
result=1
if (sleep 1 ; echo playid foo) | nc localhost 6600 | grep ACK ; then
  result=0
fi

pkill mpd
exit $result
```

Started with: `git bisect start master ffe689363f1350e86255ee8c03585a2a744a8819 --`

Which points to:

```
fb547260d1b992b64c0ffd78e3bc55db86f90bda is the first bad commit
```

jacobvosmaer:
This whole thing is baffling. As I wrote above, in MPD commit fb547260d1b992b64c0ffd78e3bc55db86f90bda the application command `playid foo` becomes a crasher. I can undo this by patching the handle_playid function as follows:

```
diff --git a/src/command/PlayerCommands.cxx b/src/command/PlayerCommands.cxx
index 7a71aa0..edd66ca 100644
--- a/src/command/PlayerCommands.cxx
+++ b/src/command/PlayerCommands.cxx
@@ -31,6 +31,7 @@
 #include "AudioFormat.hxx"
 #include "ReplayGainConfig.hxx"
 #include "util/ConstBuffer.hxx"
+#include "Log.hxx"
 
 #ifdef ENABLE_DATABASE
 #include "db/update/Service.hxx"
@@ -70,6 +71,7 @@ handle_play(Client &client, Request args, gcc_unused Response &r)
 CommandResult
 handle_playid(Client &client, Request args, gcc_unused Response &r)
 {
+	FormatDebug(exception_domain, "oh, hello");
     int id = args.ParseOptional(0, -1);
 
     Error error;

```

If I print a debug statement before going deeper into the call stack towards the `throw`, the `try {} catch` higher up works. If I leave out that debug statement, the exception is uncaught. How does this make any sense?

jacobvosmaer:
The upstream ticket got closed. I propose we do the following:

- merge this test so at least we know when we have a problem
- hope that LLVM in the next OS X release (10.12) fixes the crash, and that 10.12 gets out before the crash scenario makes it into a stable release of mpd

DomT4:
I think upstream closing the ticket is incorrect on their part & this is something they'll see reported from OS X users whenever the current head becomes stable unless resolved by then, but not much I can do to get them to change their mind here, so yeah, this PR seems reasonable as a check for that issue becoming a wider problem.

Merged in https://github.com/Homebrew/homebrew-core/commit/e69dc2a3d8134f2dedc6ba9607b02737d4dd8b0e. Thank you for your contribution to Homebrew @jacobvosmaer; we appreciate it! 😺

PR #3434, opened 2016-07-29
plantuml v8046

greenantdotcom:
Per http://plantuml.com/changes.html, changes are:
## V8046
- Salt fails when there is an empty line between @startuml and salt keywords
- Marking class diagram if is public, private, or protected
- Problem with a while inside a fork
- Apply skinparam defaultTextAlignment to other items
- Need a way to color and number different routes on an activity diagram
## V8045
- Set background color for one note
- Database modeling
- Does PlantUML support references between class/component members
- Ability to hide packages
- Tooltip issue on sequence diagram
- Param to center text in activity
- Why do strange symbols appear on an eps printout, but not on the screen
- Package style erase Partition style
- Notes on conditionnal elements
## V8043
- Arrow's thickness in sequence diagrams and skins
- Arrow marker`s background color is incorrect in v. 8042
- Leak in TranscoderUtil

### DIFF
--- a/Formula/plantuml.rb
+++ b/Formula/plantuml.rb
@@ -1,8 +1,8 @@
 class Plantuml < Formula
   desc "Draw UML diagrams"
   homepage "http://plantuml.sourceforge.net/"
-  url "https://downloads.sourceforge.net/project/plantuml/plantuml.8041.jar"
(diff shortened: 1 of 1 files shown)


DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/7867ba154326887f9f2a32d7a61ef9dd1fb212a5. Thank you for your contribution to Homebrew @greenantdotcom; we appreciate it! :smiley_cat:

PR #3445, opened 2016-07-30
pdns 4.0.1

ilovezfs:
re-parallelize the build as the issue was fixed upstream


DomT4:
👍

ilovezfs:
Actually looks like upstream fixed the parallelization issue https://github.com/PowerDNS/pdns/pull/2039 so let's see how it does.

PR #3462, opened 2016-07-30
Added easy-rsa from the openvpn project

offbyone:
- [X ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [X ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [X ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [X ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

(caveat, the online checks don't work right now because my brew install hangs when trying to read some cask data. I manually looked for duplicates)

### DIFF
--- a/Formula/easy-rsa.rb
+++ b/Formula/easy-rsa.rb
@@ -9,17 +9,15 @@ class EasyRsa < Formula
   def install
     # We don't use the installer from the package, as it doesn't organize things
     # well for brew install. It puts the binary in the prefix root, and defaults
-    # to a pki path that will be overwritten on installation. Instead, I rewrite
-    # its defaults to find its SSL and PKI in brew-appropriate locations.
+    # to a pki path that will be overwritten on installation.
     inreplace "easyrsa3/easyrsa" do |s|
       s.gsub! /set_var EASYRSA		"\$PWD"/,
               "set_var EASYRSA		\"#{libexec}\""
       s.gsub! /set_var EASYRSA_PKI	.*/,
               "set_var EASYRSA_PKI	\"#{etc}/pki\""
       openssl_path = Formula["openssl"].opt_bin/"openssl"
-      ohai "Using OpenSSL from #{openssl_path}"
       s.gsub! /set_var EASYRSA_OPENSSL	openssl/,
-              "set_var EASYRSA_OPENSSL		\"#{openssl_path}\""
+              "set_var EASYRSA_OPENSSL	\"#{openssl_path}\""
     end
     bin.install "easyrsa3/easyrsa"
     libexec.install "easyrsa3/openssl-1.0.cnf", "easyrsa3/x509-types"
--- a/Formula/easy-rsa.rb
+++ b/Formula/easy-rsa.rb
@@ -1,0 +1,34 @@
+class EasyRsa < Formula
+  desc "CLI utility to build and manage a PKI CA."
+  homepage "https://github.com/OpenVPN/easy-rsa"
+  url "https://github.com/OpenVPN/easy-rsa/archive/3.0.1.tar.gz"
+  sha256 "a1fff75a27ea7da3f37fbfed715633f55b9ca25f5b14cac38e525c5c995e68ae"
+
+  depends_on "openssl"
+
+  def install
+    # We don't use the installer from the package, as it doesn't organize things
+    # well for brew install. It puts the binary in the prefix root, and defaults
+    # to a pki path that will be overwritten on installation. Instead, I rewrite
+    # its defaults to find its SSL and PKI in brew-appropriate locations.
+    inreplace "easyrsa3/easyrsa" do |s|
+      s.gsub! /set_var EASYRSA		"\$PWD"/,
+              "set_var EASYRSA		\"#{libexec}\""
+      s.gsub! /set_var EASYRSA_PKI	.*/,
+              "set_var EASYRSA_PKI	\"#{etc}/pki\""
+      openssl_path = Formula["openssl"].opt_bin/"openssl"
+      ohai "Using OpenSSL from #{openssl_path}"
+      s.gsub! /set_var EASYRSA_OPENSSL	openssl/,
+              "set_var EASYRSA_OPENSSL		\"#{openssl_path}\""
+    end
+    bin.install "easyrsa3/easyrsa"
+    libexec.install "easyrsa3/openssl-1.0.cnf", "easyrsa3/x509-types"
+  end
+
+  test do
+    # Create a stub PKI directory in the test sandbox.
+    ENV["EASYRSA_PKI"] = testpath/"pki"
+    system bin/"easyrsa", "--help"
+    system bin/"easyrsa", "init-pki"
+  end
+end


Formula/easy-rsa.rb
```diff
@@ -0,0 +1,34 @@
+class EasyRsa < Formula
+  desc "CLI utility to build and manage a PKI CA."
+  homepage "https://github.com/OpenVPN/easy-rsa"
+  url "https://github.com/OpenVPN/easy-rsa/archive/3.0.1.tar.gz"
+  sha256 "a1fff75a27ea7da3f37fbfed715633f55b9ca25f5b14cac38e525c5c995e68ae"
+
+  option "with-openssl", "Use the openSSL provided by homebrew"
```
DomT4: If this would use the system OpenSSL otherwise, make our's a complete default.

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,34 @@
...
+  depends_on "openssl" => :optional
+
+  def install
+    # We don't use the installer from the package, as it doesn't organize things
```
DomT4: Can you elaborate on this comment?

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,34 @@
...
+    # well for brew install.
+    inreplace "easyrsa3/easyrsa" do |s|
+      s.gsub! /set_var EASYRSA		"\$PWD"/, \
```
DomT4: You shouldn't need the `\` here or on two lines below. A comma should suffice?

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,34 @@
...
+                "set_var EASYRSA_OPENSSL		\"#{openssl_path}\""
+      end
+    end
+    bin.install "easyrsa3/easyrsa"
+    libexec.install "easyrsa3/openssl-1.0.cnf", "easyrsa3/x509-types"
+  end
+
+  test do
+    # Create a stub PKI directory in the test sandbox.
+    Kernel.system({ "EASYRSA_PKI" => "./pki" }, "#{bin}/easyrsa --help")
```
DomT4: This won't work on < Ruby 2.0. You'll want to do something like:
```ruby
test do
  ENV["EASYRSA_PKI"] = testpath/"pki"
  (testpath/"pki").mkpath
  system bin/"easyrsa", "--help"
end
```
I suspect?

offbyone: Is it not safe to assume ruby 2+ for Homebrew?

DomT4: No, not yet. We still support the older Rubies, at least until https://github.com/Homebrew/brew/pull/513 lands & we get broader user feedback on that working as-expected.

Formula/easy-rsa.rb
offbyone: The package installer leaves the easyrsa command in the prefix, not in a bin/ subdirectory. I re-laid it out in here so that the tool can find all of its data properly, and end up on the homebrew PATH

Formula/easy-rsa.rb
offbyone: Okay, can do.

Formula/easy-rsa.rb
offbyone: Okay, I'll change that too.

offbyone: I've up​dated accordingly.

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,34 @@
...
+    # well for brew install. It puts the binary in the prefix root, and defaults
+    # to a pki path that will be overwritten on installation. Instead, I rewrite
+    # its defaults to find its SSL and PKI in brew-appropriate locations.
+    inreplace "easyrsa3/easyrsa" do |s|
+      s.gsub! /set_var EASYRSA		"\$PWD"/,
+              "set_var EASYRSA		\"#{libexec}\""
+      s.gsub! /set_var EASYRSA_PKI	.*/,
+              "set_var EASYRSA_PKI	\"#{etc}/pki\""
+      openssl_path = Formula["openssl"].opt_bin/"openssl"
+      ohai "Using OpenSSL from #{openssl_path}"
```
DomT4: Let's not `ohai`. It's not really useful knowledge given the dependency makes it fairly obvious.

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,34 @@
...
+  homepage "https://github.com/OpenVPN/easy-rsa"
+  url "https://github.com/OpenVPN/easy-rsa/archive/3.0.1.tar.gz"
+  sha256 "a1fff75a27ea7da3f37fbfed715633f55b9ca25f5b14cac38e525c5c995e68ae"
+
+  depends_on "openssl"
+
+  def install
+    # We don't use the installer from the package, as it doesn't organize things
+    # well for brew install. It puts the binary in the prefix root, and defaults
+    # to a pki path that will be overwritten on installation. Instead, I rewrite
```
DomT4: Let's compact this a bit. Can probably live without the:
```
Instead, I rewrite its defaults to find its SSL and PKI in brew-appropriate locations.
```
part, because again it should be obvious enough.

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,34 @@
...
+    # its defaults to find its SSL and PKI in brew-appropriate locations.
+    inreplace "easyrsa3/easyrsa" do |s|
+      s.gsub! /set_var EASYRSA		"\$PWD"/,
+              "set_var EASYRSA		\"#{libexec}\""
+      s.gsub! /set_var EASYRSA_PKI	.*/,
+              "set_var EASYRSA_PKI	\"#{etc}/pki\""
+      openssl_path = Formula["openssl"].opt_bin/"openssl"
+      ohai "Using OpenSSL from #{openssl_path}"
+      s.gsub! /set_var EASYRSA_OPENSSL	openssl/,
+              "set_var EASYRSA_OPENSSL		\"#{openssl_path}\""
```
DomT4: Why the significant difference in indent here between what you're `inreplacing` & the blob you're replacing it with?

offbyone:
Sorry about the commit noise, something went loopy with my rebase. All comments addressed.

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,32 @@
...
+      s.gsub! /set_var EASYRSA		"\$PWD"/,
+              "set_var EASYRSA		\"#{libexec}\""
+      s.gsub! /set_var EASYRSA_PKI	.*/,
+              "set_var EASYRSA_PKI	\"#{etc}/pki\""
+      openssl_path = Formula["openssl"].opt_bin/"openssl"
+      s.gsub! /set_var EASYRSA_OPENSSL	openssl/,
+              "set_var EASYRSA_OPENSSL	\"#{openssl_path}\""
+    end
+    bin.install "easyrsa3/easyrsa"
+    libexec.install "easyrsa3/openssl-1.0.cnf", "easyrsa3/x509-types"
```
DomT4: Apologies, last question as far as I can see. Why aren't these living in `#{etc}/pki`? _(May well be a perfectly valid place to put them in this context, just a question)_.

dunn: ping on this

offbyone: Mostly because that's closer to where the installer itself put them.

dunn: Where does the installer put them?

offbyone: In a subdirectory of the path where the executable is. The "installer" is just "unzip the release".

dunn: Gotcha. Well, since we're moving things around we may as well put them in a more conventional location.

offbyone: Rrrgh. I'm not sure I agree; those are executable instructions in the context of the code in question.

And honestly, at this point? meh. It's 2 week turnarounds and path twiddling. My level of "care" is falling fast to zero. It's installed for me, I'm good, and ... meh.

dunn: There's only so many of us, and we're all working in our spare time.  

if `x509-types` is an executable, then I agree `libexec` makes sense, but is `openssl-1.0.cnf`?

If @DomT4 is happy with it as is, I'll defer to him.  Otherwise if you're still interested in getting this merged, let's try to do it right.

MikeMcQuaid: @offbyone Please make the requested changed or close this out, thanks! Maybe consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap). Thanks!

offbyone: Hey, sorry about being punchy; rough week.

So, I think the right thing to do is to install `openssl-1.0.cnf` and `x509-types` in `#{etc}/pki`, but I'm not sure how to do installs into subdirectories using the .install on folders. How would that work?

dunn: No worries!

You can call `#install` on any `Pathname`, so `(etc/"pki").install` would work here.

Formula/easy-rsa.rb
```diff
@@ -0,0 +1,32 @@
...
+    # well for brew install. It puts the binary in the prefix root, and defaults
+    # to a pki path that will be overwritten on installation.
+    inreplace "easyrsa3/easyrsa" do |s|
+      s.gsub! /set_var EASYRSA		"\$PWD"/,
+              "set_var EASYRSA		\"#{libexec}\""
+      s.gsub! /set_var EASYRSA_PKI	.*/,
+              "set_var EASYRSA_PKI	\"#{etc}/pki\""
+      openssl_path = Formula["openssl"].opt_bin/"openssl"
+      s.gsub! /set_var EASYRSA_OPENSSL	openssl/,
+              "set_var EASYRSA_OPENSSL	\"#{openssl_path}\""
```
MikeMcQuaid: Can you please submit these patches to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

MikeMcQuaid: Or alternatively we figure out how to not have this being a long-running patch.

ilovezfs:
Closing for now but totally feel free to reopen with the requested changes.

PR #3465, opened 2016-07-31
pygobject: ensure pygtk lands in sys.path

tdsmith:
cf Homebrew/brew#608


Formula/pygobject.rb
```diff
@@ -29,5 +30,11 @@ def install
                           "--prefix=#{prefix}",
                           "--disable-introspection"
     system "make", "install"
+    (lib/"python2.7/site-packages/pygtk.pth").append_lines (HOMEBREW_PREFIX/
```
DomT4: It's not strictly necessary, but if you want to make me particularly happy:
```ruby
(lib/"python2.7/site-packages/pygtk.pth").append_lines <<-EOS.undent
  (HOMEBREW_PREFIX/"lib/python2.7/site-packages/gtk-2.0")
EOS
```
Looks a fair bit nicer. Wouldn't have bothered pointing it out but the test failed & gave me a window 😅.

DomT4:
👍 whenever you're 😃

Issue #3347, opened 2016-07-26
clang-format failed to build


ploxiln:
brew gist-logs: https://gist.github.com/anonymous/fd188c3f572069dae561ed54029df44e

the gist of the gist:

```
CMake Error at /usr/local/Cellar/cmake/3.6.1/share/cmake/Modules/Platform/Darwin.cmake:76 (message):
  CMAKE_OSX_DEPLOYMENT_TARGET is '10.11' but CMAKE_OSX_SYSROOT:

   ""

  is not set to a MacOSX SDK with a recognized version.  Either set
  CMAKE_OSX_SYSROOT to a valid SDK or set CMAKE_OSX_DEPLOYMENT_TARGET to
  empty.
```

This seems very similar to https://github.com/Homebrew/legacy-homebrew/issues/42221 where the fix was to make a `/Applications/Xcode.app` symlink. I don't have any Xcode installed, but of course I do have the command-line build tools installed (and built many homebrew formulae from source over the years). Does this formula really need the full Xcode.app installed? Should the description say so?

DomT4:
Sigh, this again. Can you try sticking `ENV.delete("SDKROOT")` on the first line inside the `def install` block and see if `brew install clang-format` works for you then? Thanks.

ploxiln:
That didn't seem to have the desired effect.

```
[pierce@plo-pro Formula]$ git diff
diff --git a/Formula/clang-format.rb b/Formula/clang-format.rb
index bbc3bde..0366a87 100644
--- a/Formula/clang-format.rb
+++ b/Formula/clang-format.rb
@@ -41,6 +41,7 @@ class ClangFormat < Formula
   depends_on "subversion" => :build
 
   def install
+    ENV.delete("SDKROOT")
     (buildpath/"projects/libcxx").install resource("libcxx")
     (buildpath/"tools/clang").install resource("clang")
 
[pierce@plo-pro Formula]$ brew install clang-format
==> Cloning http://llvm.org/svn/llvm-project/llvm/tags/google/stable/2016-06-27/
(... same failure)
```

new gist-logs just in case: https://gist.github.com/anonymous/308da5da1e34997734801131a77fc61a

ilovezfs:
[...]

PR #3479, opened 2016-07-31
clang-format: set CMAKE_OSX_SYSROOT=/ for CLT only

ilovezfs:
Closes https://github.com/Homebrew/homebrew-core/issues/3347.

### DIFF
--- a/Formula/clang-format.rb
+++ b/Formula/clang-format.rb
@@ -46,6 +46,7 @@ class ClangFormat < Formula
 
     mkdir "build" do
       args = std_cmake_args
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @DomT4

DomT4:
👍. Hopefully this is one of the things we can kill off once `scons` plays nicely under `superenv`.

ilovezfs:
> Hopefully this is one of the things we can kill off once scons plays nicely under superenv.

Do we have an open issue or PR for that currently?

DomT4:
No. Think we used to have one in `legacy`. It's something @mistydemeo knows the most about, but Misty isn't around much for now.

PR #3483, opened 2016-07-31
libbdplus 0.1.2 (new formula)

will-in-wi:

---

libbdplus implements the Blu-Ray BD+ specification for VLC and other open source players.

This is useful when using VLC to play Blu-Ray discs.

### DIFF
--- a/Formula/libbdplus.rb
+++ b/Formula/libbdplus.rb
@@ -13,7 +13,6 @@ class Libbdplus < Formula
     depends_on "libtool" => :build
   end
 
-  depends_on "bison" => :build
   depends_on "libgcrypt"
 
   def install
--- a/Formula/libbdplus.rb
+++ b/Formula/libbdplus.rb
@@ -23,5 +23,17 @@ class Libbdplus < Formula
   end
 
   test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <libbdplus/bdplus.h>
+      int main() {
+        int major = -1;
+        int minor = -1;
+        int micro = -1;
+        bdplus_get_version(&major, &minor, &micro);
+        return 0;
+      }
+    EOS
+    system ENV.cc, "test.c", "-L#{lib}", "-I#{include}", "-lbdplus", "-o", "test"
+    system "./test"
   end
 end
--- a/Formula/libbdplus.rb
+++ b/Formula/libbdplus.rb
@@ -1,0 +1,28 @@
+class Libbdplus < Formula
+  desc "Implements the BD+ System Specifications"
+  homepage "https://www.videolan.org/developers/libbdplus.html"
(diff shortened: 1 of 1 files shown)


Formula/libbdplus.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  head do
+    url "https://git.videolan.org/git/libbdplus.git"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+  end
+
+  depends_on "bison" => :build
```
DomT4: Is this necessary on all platforms? A little surprised if at least the newer OS X/macOS versions don't have a new enough `bison`.

Formula/libbdplus.rb
```diff
@@ -0,0 +1,28 @@
...
+  depends_on "libgcrypt"
+
+  def install
+    system "./bootstrap" if build.head?
+    system "./configure", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Formula/libbdplus.rb
will-in-wi: I don't know, and I'm not sure how to tell. Bison is required by libaacs, and I assumed that since they are very similar libraries coded by roughly the same people, that we probably need that here too. I have no clue how to figure out whether this is the case.

will-in-wi: As best I can tell, the bison requirement is based on the existence of `*.y` files. They exist in libaacs and they do not in libbdplus, so I'm removing this line.

MikeMcQuaid: Try to remove it from here and CI will tell us.

Formula/libbdplus.rb
will-in-wi: This is a library, so there is no executable. I don't know enough C to build an example file and compile it against the library to test. There is a test executable, but it requires a blu-ray disk to be somewhere in order to run the test.

Not sure how to even start on this.

MikeMcQuaid: You should be able to copy-paste something like `tinyxml2.rb`'s test to do something useful.

will-in-wi:
I've successfully added a test.

Formula/libbdplus.rb
will-in-wi: Thanks for the pointer. I was able to make a test using that as a template.

dunn:
Thanks! Merged as 6242ab6f30b6b7c1b8f2a28ea60404118d070074.

will-in-wi:
Thanks!

PR #2600, opened 2016-07-02
icmake 8.01.00

ilovezfs:
Created with `brew bump-formula-pr`.

PR #3487, opened 2016-07-31
icmake: boneyard

apjanke:
Newer versions (8.0+) of `icmake` are broken on OS X in new ways. Let's go ahead and boneyard it now, to avoid further wasted effort on updates, or making users think it's well supported.

Icmake isn't used by anything in Homebrew core; it's just there as a prerequisite for supporting `yodl`, which is used for building `zsh` docs in the development versions. But none of the current GitHub-tagged `yodl` releases build under the 7.x version of `icmake` anyway, and current development `zsh` versions require newer versions of `yodl`, so this 7.x `icmake` isn't doing us any good.

```
$ brew uses icmake
$
```

`icmake` and `yodl` are brittle, and are written against GNU-specific versions of utilities like `sed`, `find`, and `grep`, but without a mechanism for finding them at nonstandard locations (e.g. the `g`-prefixed ones we install). It'd be a lot of ongoing work to get them running on BSD. IMHO, we should yank them from Homebrew to avoid further maintenance work, and if it comes down to it, try to get Zsh to migrate to a different document format. (Seriously).

Note: the GNU-specific dependencies of `icmake` and `yodl` mean that even if they build and install, that doesn't mean they're working correctly. E.g. there are behavior differences between BSD and GNU `sed` that mean `icmake` on OS X may produce different (incorrect) output at run time but still return a success code. I don't think we can support `icmake` or `yodl` unless someone produces an actual BSD port of them, not just a packaging of the main releases.

Closes #2600
Goes with https://github.com/Homebrew/homebrew-boneyard/pull/136


DomT4:
Seems wholly reasonable 👍. You'll want to update `tap_migrations.json` as well, though.

apjanke:
Thanks for the reminder. Couldn't use the `dev-cmd/boneyard-formula-pr.rb` to get it right automatically due to current readall-related issues.

Amended to fix `tap_migrations.json`.

DomT4:
👍

apjanke:
@ilovezfs, what do you think?

ilovezfs:
> Couldn't use the dev-cmd/boneyard-formula-pr.rb to get it right automatically due to current readall-related issues.

What's this referring to?

>@ilovezfs, what do you think?

Yeah, let's just kill it.

ilovezfs:
Commit subject should be "migrate to boneyard"

apjanke:
> Couldn't use the dev-cmd/boneyard-formula-pr.rb to get it right automatically due to current readall-related issues.
>
> What's this referring to?

Homebrew/brew#615

apjanke:
Amended commit message.

ilovezfs:
:+1:

PR #3501, opened 2016-08-01
update tap_migrations

ilovezfs:
- removals from boneyard
- note that lmutil had previously been migrated from binary to boneyard

Goes together with https://github.com/Homebrew/homebrew-boneyard/pull/137.


MikeMcQuaid:
👍 🚢

PR #3509, opened 2016-08-01
curl: enable HTTP/2 support by default

vszakats:

---

`nghttp2` is a mature/stable component by now, and it makes it for a better experience having HTTP/2 enabled in curl by default. It also means that curl will be built against OpenSSL instead of Secure Transport by default. OpenSSL uses its own root certificate bundle, instead the OS-wide one used by Secure Transport &mdash; which might be a reason why this step wasn't made so far. 

Anyhow, please feel free to discuss/reject this. [Couldn't find past discussions about this. In case there was any and this one is redundant, my apologies.]

Related read: https://daniel.haxx.se/blog/2016/08/01/curl-and-h2-on-mac/

Even though I've been installing curl with this option enabled since a good while, I haven't tested this mod locally &ndash; besides the usual audit check. I'll be watching the bot for reporting any breakage.

### DIFF
--- a/Formula/curl.rb
+++ b/Formula/curl.rb
@@ -20,7 +20,7 @@ class Curl < Formula
   option "with-gssapi", "Build with GSSAPI/Kerberos authentication support."
   option "with-libmetalink", "Build with libmetalink support."
   option "with-libressl", "Build with LibreSSL instead of Secure Transport or OpenSSL"
-  option "with-nghttp2", "Build with HTTP/2 support (requires OpenSSL or LibreSSL)"
+  option "without-nghttp2", "Build without HTTP/2 support (for optional Secure Transport support)"
 
   deprecated_option "with-idn" => "with-libidn"
   deprecated_option "with-rtmp" => "with-rtmpdump"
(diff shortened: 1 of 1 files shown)


vszakats:
Timed-out downloading the curl source package (on two out of three platforms).

DomT4:
> OpenSSL uses its own root certificate bundle, instead the OS-wide one used by Secure Transport — which might be a reason why this step wasn't made so far.

We generate the OpenSSL cert bundle using the system keychains, so there's no wild difference except:

* If built against Secure Transport `curl` will adhere to live changes to the Keychain. If a cert is marked untrusted `curl` will fail, whereas `openssl` will happily keep using that now untrusted cert if it was exported at the time of `.pem` creation.

This has typically been rejected in the past because our `curl` package was primarily intended to achieve a couple of things, one of which was significant closeness to how Apple build the system `curl` whilst offering the newest `curl` available.

The other was to avoid making everyone install `openssl` via `brew`, although this argument is more redundant these days because I doubt there are many Homebrew installations out there lacking a brewed `openssl`.

I'm reasonably willing to take input from upstream on this, but looking at our analytic data over the last few months only 613 "known" builds out of 25k+ are specifically going for `--with-nghttp2`. That doesn't make a massively compelling case _by itself_ to ignore how awfully OpenSSL handles interaction with PEM files in more or less blindly trusting them.

MikeMcQuaid:
> This has typically been rejected in the past because our curl package was primarily intended to achieve a couple of things, one of which was significant closeness to how Apple build the system curl whilst offering the newest curl available.

Agreed. Also, we should use the most OS X native APIs (i.e. SecureTransport) by default so passing on this.

PR #3511, opened 2016-08-02
imagemagick: improve option and configure argument handling

zmwangx:

---

This is a continuation of the work in 8d6f16e. In this commit we fix and/or improve the configure arguments related to the recommended and optional features, and make them consistent. We follow these rules:
1. Only specify `--disable-*` or `--without-*` for Homebrew-recommended, default-on-in-configure features.
2. Specify `--enable-*` or `--disable-*` (`--with-*` or `--without-*`) when building with or without Homebrew-recommended yet default-off-in-configure features. Currently affected features: opencl and modules.
3. Only specify `--enable-*` or `--with-*` for Homebrew-optional features.

Features in the respective groups (by rule) are now sorted to ease management. The recommended and optional dependencies are also sorted respectively.

Note that I made X an exception of rule 3 (by explicitly specifying `--without-x` when building without x11) because in my experience, autodetection of X could cause buildtime issues. Interested parties are recommended to investigate this issue. (Specifically, I wasn't specifying `--with-x11`, but X was autodetected anyway, and as a result I got `X11/Xos.h` missing or something like that during build — which of course shouldn't be missing. I think `--with-x11` might be broken too, but I'm not really interested in that.)

To accompany the configure argument fixes and/or improvements, I have added tests to make sure the Homebrew-recommended features and delegates are present.

Notable functional changes, including but not limited to
1. We now explicitly specify `--enable-opencl` when building with OpenCL support (recommended). Previous bottles don't actually include OpenCL support, so I bumped the revision accordingly.
2. Previously we were specifying `--with-gs-font-dir` when building _without_ ghostscript, which presumably is a mistake. Now it goes together with `--with-gslib`. (Please correct me if I'm wrong.)

By the way, let me know one-liners are preferred over those `if..else..end` blocks.

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -7,6 +7,7 @@ class Imagemagick < Formula
   url "https://dl.bintray.com/homebrew/mirror/imagemagick-6.9.5-4.tar.xz"
   mirror "https://www.imagemagick.org/download/ImageMagick-6.9.5-4.tar.xz"
   sha256 "00549fb8588673bb510e9dc952d9937b2cd1186e9322f5265dd96803e592eed8"
+  revision 1
   head "http://git.imagemagick.org/repos/ImageMagick.git"
 
   bottle do
@@ -35,25 +36,25 @@ class Imagemagick < Formula
   depends_on "libtool" => :run
   depends_on "xz"
 
+  depends_on "freetype" => :recommended
   depends_on "jpeg" => :recommended
   depends_on "libpng" => :recommended
   depends_on "libtiff" => :recommended
-  depends_on "freetype" => :recommended
 
-  depends_on :x11 => :optional
+  depends_on "fftw" => :optional
   depends_on "fontconfig" => :optional
+  depends_on "ghostscript" => :optional
+  depends_on "liblqr" => :optional
+  depends_on "librsvg" => :optional
+  depends_on "libwmf" => :optional
   depends_on "little-cms" => :optional
   depends_on "little-cms2" => :optional
-  depends_on "libwmf" => :optional
-  depends_on "librsvg" => :optional
-  depends_on "liblqr" => :optional
   depends_on "openexr" => :optional
-  depends_on "ghostscript" => :optional
-  depends_on "webp" => :optional
   depends_on "homebrew/versions/openjpeg21" if build.with? "jp2"
-  depends_on "fftw" => :optional
   depends_on "pango" => :optional
+  depends_on "webp" => :optional
   depends_on :perl => ["5.5", :optional]
+  depends_on :x11 => :optional
 
   needs :openmp if build.with? "openmp"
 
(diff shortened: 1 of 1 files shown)


DomT4:
> We follow these rules:

I kind of rewrote part of the formula the way it is now intentionally like this in https://github.com/Homebrew/homebrew-core/commit/cc1a80bf7aea43f50bb51c7d63937551c2e28980 & https://github.com/Homebrew/homebrew-core/commit/8d6f16ed3c192873c8a96694b2c1b1a706c46b64. I believe the latter because `webp` had started to be scooped out of the environment even though I hadn't requested for it to be so.

The idea is to be more guarded against defaults changing, which isn't unheard of to say the least. I know it makes the formula chunkier but I'm a little happier playing it safer with `imagemagick` given how many people use the thing.

Formula/imagemagick.rb
```diff
@@ -132,6 +136,14 @@ def caveats
   end
 
   test do
-    system "#{bin}/identify", test_fixtures("test.png")
+    system bin/"identify", test_fixtures("test.png")
+    # Test recommended features and delegates.
+    features = shell_output("#{bin}/convert -version")
+    assert_match "freetype", features if build.with? "freetype"
```
MikeMcQuaid: May be worth inverting the match if it's `build.without?` too.

zmwangx:
>  I believe the latter because webp had started to be scooped out of the environment even though I hadn't requested for it to be so.

Here's my take on this issue: for an optional features that users need a `--with` to enable, but will hardly ever explicitly declare `--without`, it makes sense to just let it be. Even if it's automatically picked up by `configure`, it's just one more feature (e.g., being able to handle WebP images) that the user may or may not ever need. There's no harm unless the user later uninstalls `webp`, which is pretty unlikely.

Alternatively, we should do

```ruby
if build.with? "foo" then args << "--with-foo" else args << "--without-foo" end
```

for all options, because the problem is presumably not limited to webp. What about always using this oneliner? Or maybe more concisely,

```ruby
args << (build.with?("foo") ? "--with-foo" : "--without-foo")
```

Formula/imagemagick.rb
zmwangx: Do we care about testing builds with options? Quite a while ago I was told that we don't.

MikeMcQuaid: We do not, generally, no but if you're adding `build.with?` it's probably worth adding the inverse.

MikeMcQuaid: Or just skip the `build.with?`

zmwangx: Sure, I'll remove the `build.with?`. (Otherwise there will be six lines that are going to be executed.)

zmwangx: *never going to be executed.

DomT4:
> Even if it's automatically picked up by configure, it's just one more feature (e.g., being able to handle WebP images) that the user may or may not ever need. There's no harm unless the user later uninstalls webp, which is pretty unlikely.

Yeah, this isn't a feature of Homebrew, it's a bug. At some point in the future, think Homebrew 1.5 rather than 1.1, the existing plan is for `superenv` to filter out `HOMEBREW_PREFIX/lib` and `HOMEBREW_PREFIX/include` & reinsert those paths relative to _specific dependencies_.

The behaviour where formulae are free to sniff out whatever they like from `HOMEBREW_PREFIX` as long as it isn't from `bin` isn't something we welcome, really, and it causes more problems than it solves. Time and time again we get bug reports of something failing to compile & the cause ends up being traced to something it has sniffed out & wants to link against despite not being requested to do so.

If people want a feature, it should be an option with a hard off when not selected or built in by default. Accidental linking is never an upside from a package maintainability point of view.

> Alternatively, we should do

The current syntax is used almost universally. Changing things like this piecemeal isn't great UX for us or contributors. I agree it's a bit chunky, but the majority of maintainers are comfortable with the choice & some have very vocally railed against the alternatives, including:
```ruby
(build.with?("foo") ? "--with-foo" : "--without-foo")
```
which I'm relatively fond of.

zmwangx:
> The behaviour where formulae are free to sniff out whatever they like from HOMEBREW_PREFIX as long as it isn't from bin isn't something we welcome, really, and it causes more problems than it solves. 

I can sympathize with this. Even in this case, letting `configure` sniff out unsolicited X wasn't conducive to success of the build.

> The current syntax is used almost universally.

Sure, if folks are fine with six lines per option instead of one and aren't bothered by the resulting length, I'm gonna live with it. I have extended each option to be explicit about what to enable/disable and build with/without. One perk is that we aren't affected by `configure` defaults anymore.

FWIW the old more concise version is preserved in https://github.com/zmwangx/homebrew-core/blob/imagemagick-fix-options-concise/Formula/imagemagick.rb, in case anyone wants to compare.

MikeMcQuaid:
I appreciate the work here but given it adds a lot of complexity but doesn't fix reported user issues we'll pass for now, sorry!

PR #3512, opened 2016-08-02
libxml2: fix --with-python option

vladimir-mencl-eresearch:

---

Just a minor fix: even when the formula receives the `--with-python` option, it passes `--without-python` to `configure`.

Fix this by using a conditional expression that yields the `--with-python` option for `configure` when the `--with-python` option is passed to the formula.

### DIFF
--- a/Formula/libxml2.rb
+++ b/Formula/libxml2.rb
@@ -40,7 +40,7 @@ class Libxml2 < Formula
 
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
-                          "--without-python",
+                          ((build.with? "python") ? "--with-python" : "--without-python"),
                           "--without-lzma"
     system "make"
     ENV.deparallelize


DomT4:
Appreciate this is slightly confusing, but this is intentional.

If you check down the formula there's a `if build.with? "python"` block where the Python elements are installed if requested. We just tell `./configure` to skip building it because it doesn't always build in the desired way, hence the custom block.

Appreciate the submission to Homebrew nonetheless ❤️.

vladimir-mencl-eresearch:
Thanks for the fast reply!

I did find that conditional block (and that's where I took the condition from :-) ).

However, when I run `brew install --with-python libxml2`, despite brew also running `python setup.py install --prefix=/usr/local/Cellar/libxml2/2.9.4`, I do not get libxml2 installed in `/usr/local/lib/python2.7//site-packages` - only if I change the configure options to build the python code as done in this pull request.

I have just tried running the build steps in a separate build tree manually to see the detailed output.  And I found the python `setup.py` installs the libraries into `/usr/local/Cellar/libxml2/2.9.4/lib/python2.7/site-packages/`.

That explains why I could not find the libraries in the standard python install.  I might be missing something - is this intended?  Are there any guidelines to make brew-installed python pick up python libraries installed with brew like this?

I may not be the only one confused - when I was searching for `python libxml2 brew` earlier today, I found /usr/local/Cellar/libxml2/2.9.4/lib/python2.7/site-packages/ which instructs to manually edit the Formula - so that's why I decided to contribute a fix instead.

Sorry about opening the discussion in a closed PR - but it looks like it may not be clear how to use python interface to libraries installed with brew...

DomT4:
>  I do not get libxml2 installed in /usr/local/lib/python2.7//site-packages

Yeah, this is because the formula is `keg_only` to avoid clashing with the system `libxml2` installation, which can cause all kinds of fun. That isn't the case with the vast, vast majority of formulae with Python bindings in Homebrew, but it is here. `libxml2` is a little special in how it gets handled because of the system overlap.

> Are there any guidelines to make brew-installed python pick up python libraries installed with brew like this?

There's some automagically generated [caveats](https://github.com/Homebrew/brew/blob/8284fb3fef9860a67f3303f17cb6fd99a2ebc168/Library/Homebrew/caveats.rb#L85-L108) that get printed if you build `libxml2 --with-python`. They'll appear at the end of the build, and remain accessible via `brew info libxml2`:
```
If you need Python to find bindings for this keg-only formula, run:
  echo /usr/local/opt/libxml2/lib/python2.7/site-packages >> /usr/local/lib/python2.7/site-packages/libxml2.pth
```

vladimir-mencl-eresearch:
Sorry, that was on my screen and I missed it.  All makes sense now.  Thanks for the extra explanation!

DomT4:
No worries, easily done. Do appreciate it's a bit confusing to request `--with-python` and almost immediately see `--without-python` being passed.

PR #3526, opened 2016-08-02
pdnsrec: revision for boost

ilovezfs:
boost 1.61.0 patch

### DIFF
--- a/Formula/pdnsrec.rb
+++ b/Formula/pdnsrec.rb
@@ -3,6 +3,7 @@ class Pdnsrec < Formula
   homepage "https://www.powerdns.com/recursor.html"
   url "https://downloads.powerdns.com/releases/pdns-recursor-4.0.1.tar.bz2"
   sha256 "472db541307c8ca83a846d260ecfc854fd8e879c1bb2ce5683a8df5d21e860b0"
+  revision 1
 
   bottle do
     sha256 "31428673a49cfd7d9499e98154daefc493da5e4457adb3ff58ea5f60fdd08ba2" => :el_capitan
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 I threw in the goofy `depends_on "gcc"` until there's something better since there's no point in causing CI failures elsewhere.

DomT4:
👍. That "something better" may well be when we stop shipping bottles for Mavericks, which will defer the issue until whenever the Yosemite `clang` starts being problematic, sigh.

ilovezfs:
At some point, maybe we'll acknowledge that Apple really doesn't support anything other than the latest version of macOS and only bottle for that version.

DomT4:
To be fair, they do at least support two OS X/macOS versions. Yosemite still gets fairly frequent security updates, and they got Xcode updates until Xcode 7.3. We've discussed cutting back to two major OS X/macOS versions before though.

ilovezfs:
I think with the unitary SDK approach, they really only seriously support the latest regardless of what they say. All lower versions have a screwball experience with development now.

ilovezfs:
CC @Habbie

Habbie:
Looks like `configure: MTasker will use the Boost context library for context switching` turned into `configure: MTasker will use System V ucontexts for context switching` for all OSX versions. I'll investigate.

ilovezfs:
@Habbie thank you!!!

justinclift:
@DomT4 Btw, Mavericks is still receiving security updates too.  Installed one here just a few days ago. :wink:

DomT4:
> Btw, Mavericks is still receiving security updates too. Installed one here just a few days ago. 😉

I suspect the most recent one will be the last one, unless something big happens. Yosemite should be eligible for another year of support yet, and El Cap another two years, which was my thinking.

Habbie:
From the Boost 1.61 notes "fcontext_t: no longer part of public API". This is going to hurt :)

Habbie:
@ilovezfs fixed by https://github.com/PowerDNS/pdns/pull/4271 which will be merged onto master and released in 4.0.2 soon. Can you add it to the brew as a patch?

ilovezfs:
@DomT4 @Habbie PR refreshed with @Habbie's patch.

DomT4:
Thanks @ilovezfs, and thanks @Habbie for the upstream fix ❤️.

PR #3547, opened 2016-08-02
terraform 0.7.0

hrak:

---

Upgrade terraform to version 0.7.0

### DIFF
--- a/Formula/terraform.rb
+++ b/Formula/terraform.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Terraform < Formula
   desc "Tool to build, change, and version infrastructure"
   homepage "https://www.terraform.io/"
-  url "https://github.com/hashicorp/terraform/archive/v0.6.16.tar.gz"
-  sha256 "c84bae32a170d993982de9c537eac74f70601e7a667dc2ea9803b86e04b1221d"
+  url "https://github.com/hashicorp/terraform/archive/v0.7.0.tar.gz"
+  sha256 "15f8f0589978096152876f5b9ae0853700f52666ff7911e4dad5505e80b12d3b"
   head "https://github.com/hashicorp/terraform.git"
 
   bottle do
@@ -30,11 +30,6 @@ class Terraform < Formula
   go_resource "golang.org/x/tools" do
     url "https://go.googlesource.com/tools.git", :revision => "977844c7af2aa555048a19d28e9fe6c392e7b8e9"
   end
-
-  # This patch works around a known test failure, which will be fixed in terraform version 0.6.17.
-  # https://github.com/hashicorp/terraform/issues/6709
-  # https://github.com/hashicorp/terraform/commit/6a20e8
-  patch :DATA
 
   def install
     ENV["GOPATH"] = buildpath
@@ -71,7 +66,6 @@ class Terraform < Formula
         "-os", "darwin",
         "-output", "bin/terraform-{{.Dir}}", *terraform_files
       bin.install "bin/terraform-terraform" => "terraform"
-      bin.install Dir["bin/*"]
       zsh_completion.install "contrib/zsh-completion/_terraform"
     end
   end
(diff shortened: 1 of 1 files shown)


erickt:
@hrak: With this commit, terraform is now emitting these warnings:

```
[WARN] /usr/local/Cellar/terraform/0.7.0/bin/terraform-provider-tls overrides an internal plugin for tls-provider.
  If you did not expect to see this message you will need to remove the old plugin.
  See https://www.terraform.io/docs/internals/internal-plugins.html
[WARN] /usr/local/Cellar/terraform/0.7.0/bin/terraform-provider-cobbler overrides an internal plugin for cobbler-provider.
  If you did not expect to see this message you will need to remove the old plugin.
  See https://www.terraform.io/docs/internals/internal-plugins.html
[WARN] /usr/local/Cellar/terraform/0.7.0/bin/terraform-provider-clc overrides an internal plugin for clc-provider.
  If you did not expect to see this message you will need to remove the old plugin.
  See https://www.terraform.io/docs/internals/internal-plugins.html
...
```

Would it be possible to remove this? It might be as simple as removing https://github.com/hrak/homebrew-core/blob/aa800fb2fddbedf1b615fc9f952961fdc5086210/Formula/terraform.rb#L64.

DomT4:
>  It might be as simple as removing https://github.com/hrak/homebrew-core/blob/aa800fb2fddbedf1b615fc9f952961fdc5086210/Formula/terraform.rb#L64.

I doubt it. That's the build process for `terraform`. Wouldn't be much of an installation to install without that 😃.

Can you elaborate on where you're seeing those warnings? Are they build or runtime?

RichardKnop:
👍 I would like this merged soonish. Terraform 0.7 has some really good new features that could simplify our manifests.

ruudk:
🙏

hrak:
@erickt This is a documented change in Terraform 0.7.0. The plugins are no longer provided as seperate binaries:

> In versions of Terraform prior to 0.7, each plugin shipped as a separate binary. In versions of Terraform >= 0.7, all of the official plugins are shipped as a single binary. This saves a lot of disk space and makes downloads faster for you!
> 
> However, when you upgrade you will need to manually delete old plugins from disk.
> 

See https://www.terraform.io/docs/internals/internal-plugins.html for more info

olly:
I just installed this formula after uninstalling all previous versions of terraform. It looks as if the separate provider binaries are still built. I assume to keep backwards compatibility with other tooling. Once I removed all binaries except `terraform` from `/usr/local/Cellar/terraform/0.7.0/bin` the warnings were silenced.

MikeMcQuaid:
> 👍 I would like this merged soonish. Terraform 0.7 has some really good new features that could simplify our manifests.

@RichardKnop Please be patient, we're volunteers working on our free time.

DomT4:
> I just installed this formula after uninstalling all previous versions of terraform. It looks as if the separate provider binaries are still built. I assume to keep backwards compatibility with other tooling. Once I removed all binaries except terraform from /usr/local/Cellar/terraform/0.7.0/bin the warnings were silenced.

Interesting. Thanks for that, and to @hrak for the answer linking the upstream documentation. That's the sort of thing that's useful on deciding how to move ahead.

I'm unsure it's great UX to install executables that are essentially there to nag users to remove said executables, but it seems to be upstream's intention. The question largely is whether or not _we_ should remove those executables so users don't see a flood of warnings.

kpumuk:
@DomT4 Another option would be to add a flag like `--with-deprecated-binaries`, and not install them by default

DomT4:
That'd be acceptable to me, at least. We could just execute:
```ruby
rm_f Dir[bin/"*"].select { |f| f.to_s.include?("terraform-") }
```
Or some variation thereof.

If people feel an option is a better fit than an outright removal we could wrap that in a `if build.without? "deprecated-executables"` flag & respective `option "with-deprecated-executables", "Explanation"`.

DomT4:
The official zip upstream for OS X/macOS only includes the `terraform` executable, so I think we'd be alright to get rid of the rest unconditionally, unless that sounds unacceptable to those of you who actually use `terraform`.

mwarkentin:
Works for me (light TF user).

kpumuk:
Works for me as well

DomT4:
Great. Thanks for the feedback both. @hrak If you've no objection as well, would you mind tacking that into the formula here:

```diff
diff --git a/Formula/terraform.rb b/Formula/terraform.rb
index 3e1d5fe..fc67cae 100644
--- a/Formula/terraform.rb
+++ b/Formula/terraform.rb
@@ -74,6 +74,7 @@ class Terraform < Formula
       bin.install Dir["bin/*"]
       zsh_completion.install "contrib/zsh-completion/_terraform"
     end
+    rm_f Dir[bin/"*"].select { |f| f.to_s.include?("terraform-") }
   end
 
   test do
```

I think we'll be in a good place to ship this with that tweak.

kpumuk:
Or just remove `bin.install Dir["bin/*"]` ?

DomT4:
Heh, I missed the `bin.install "bin/terraform-terraform" => "terraform"` line above. Removing the extra `bin.install Dir["bin/*"]` is certainly simpler 🙈.

hrak:
PR updated with requested changes (removed ```bin.install Dir["bin/*"]```)

MikeMcQuaid:
Sorry for the delay. Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

ruudk:
Thanks all! Very awesome :D

Issue #3541, opened 2016-08-02
Mercurial should use Homebrew python so it can support TLS 1.1+


indygreg:
### Bug reports:

The mercurial package currently installs a `hg` script with a shebang of `#!/usr/bin/python`.

Apple's Python (even on the latest beta of MacOS) doesn't support TLS 1.1+ and other modern crypto features. Homebrew's Python does support modern crypto because it is built/linked with a modern OpenSSL.

Mercurial 3.9 takes a much more strict approach to connection security and issues warnings when TLS 1.1+ cannot be used. See https://www.mercurial-scm.org/wiki/SecureConnections for more.

What this all means is that Mercurial 3.9 out of the box on Homebrew issues warnings when connecting to https:// URLs.

The issue can be fixed by having Mercurial use Homebrew's Python in its shebang. According to Mercurial's Makefile, this can be resolved by ensuring `python` resolves to the Homebrew Python 2.7 during installation.

DomT4:
What has changed from the last time we had this discussion back in https://github.com/Homebrew/legacy-homebrew/issues/42416?

indygreg:
I forgot about that discussion!

What has changed is that Mercurial 3.9 now issues warnings when it can't use TLS 1.1+.

I also recently learned that /usr/bin/python on OS X doesn't support TLS 1.1+. So basically any [Homebrew] Python process using /usr/bin/python can't practice modern and safe security. This is a glaring security deficiency in Apple's Python. Fortunately Homebrew's Python can use TLS 1.1+. Not only do I think the Mercurial package should use Homebrew's Python, but I also think *all* Homebrew Python packages should be using Homebrew's Python so they can use modern security technology. Failure to do so leaves Homebrew user's more susceptible to security risks. e.g. TLS 1.0 has known vulnerabilities such as CRIME and BEAST. While there are mitigations available, the best mitigation is to negotiate TLS 1.1 or 1.2 and avoid the TLS 1.0 attack surface altogether. https://wiki.mozilla.org/Security/Server_Side_TLS has a nice overview of the state of TLS.

indygreg:
If Homebrew switches away from Apple's Python, it will also need to install a global hgrc pointing to Homebrew's CA certificates file. This is because Mercurial will no longer be able to use the special hacks in Apple's Python to load the OS X CA store.

The requirement for configuring the CA store in Mercurial packages is documented at the bottom of https://www.mercurial-scm.org/wiki/SecureConnections

[...]

PR #3548, opened 2016-08-02
mercurial: bottle against Homebrew python

tdsmith:
Hg uses the python ssl module, which is linked to OpenSSL 0.9.8 on OS X ≤ 10.11, which lacks TLS 1.1 support.

`--install-from-source` will use the Python in PATH (which may be system python!); `--without-python` will use system python.

### DIFF
--- a/Formula/mercurial.rb
+++ b/Formula/mercurial.rb
@@ -13,6 +13,13 @@ class Mercurial < Formula
     sha256 "8934a8fc26c36e56f017d99ac43a0ebe9d6922f8276c1dfccd9ccb980ae05e45" => :mavericks
   end
 
+  if MacOS.version <= :snow_leopard
+    depends_on :python
+  else
+    option "without-python", "Use system Python (will not support TLS 1.1)"
+    depends_on :python => :recommended
+  end
+
   def install
     ENV.minimal_optimization if MacOS.version <= :snow_leopard
 
--- a/Formula/mercurial.rb
+++ b/Formula/mercurial.rb
@@ -36,8 +36,8 @@ class Mercurial < Formula
   end
 
   def caveats
-    return unless (bin/"hg").exist?
-    cacerts_configured = `#{bin}/hg config web.cacerts`.strip
+    return unless (opt_bin/"hg").exist?
+    cacerts_configured = `#{opt_bin}/hg config web.cacerts`.strip
     return if cacerts_configured == ""
     <<-EOS.undent
       Mercurial is configured to use a certificate bundle file as its trust
--- a/Formula/mercurial.rb
+++ b/Formula/mercurial.rb
@@ -14,8 +14,7 @@ class Mercurial < Formula
     sha256 "8934a8fc26c36e56f017d99ac43a0ebe9d6922f8276c1dfccd9ccb980ae05e45" => :mavericks
   end
 
-  option "with-custom-python", "Install against the python in PATH instead of "\
-                               "Homebrew's python"
+  option "with-custom-python", "Install against the python in PATH instead of Homebrew's python"
   if build.with? "custom-python"
     depends_on :python
   else
@@ -38,7 +37,7 @@ class Mercurial < Formula
   def caveats
     return unless (opt_bin/"hg").exist?
     cacerts_configured = `#{opt_bin}/hg config web.cacerts`.strip
-    return if cacerts_configured == ""
+    return if cacerts_configured.empty?
     <<-EOS.undent
       Mercurial is configured to use a certificate bundle file as its trust
       store for TLS connections instead of using the default OpenSSL store.
--- a/Formula/mercurial.rb
+++ b/Formula/mercurial.rb
@@ -39,11 +39,11 @@ class Mercurial < Formula
     cacerts_configured = `#{opt_bin}/hg config web.cacerts`.strip
     return if cacerts_configured.empty?
     <<-EOS.undent
-      Mercurial is configured to use a certificate bundle file as its trust
-      store for TLS connections instead of using the default OpenSSL store.
-      If you have trouble connecting to remote repositories, consider unsetting
-      the `web.cacerts` property. You can determine where the property is being
-      set by running:
+      Homebrew has detected that Mercurial is configured to use a certificate
+      bundle file as its trust store for TLS connections instead of using the
+      default OpenSSL store. If you have trouble connecting to remote
+      repositories, consider unsetting the `web.cacerts` property. You can
+      determine where the property is being set by running:
         hg config --debug web.cacerts
     EOS
   end
--- a/Formula/mercurial.rb
+++ b/Formula/mercurial.rb
@@ -5,6 +5,7 @@ class Mercurial < Formula
   homepage "https://mercurial-scm.org/"
   url "https://mercurial-scm.org/release/mercurial-3.9.tar.gz"
   sha256 "834f25dcff44994198fb8a7ba161a6e24204dbd63c8e6270577e06e6cedbdabc"
+  revision 1
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


tdsmith:
@BrewTestBot test this please

ngoldbaum:
@indygreg does this also need some additional configuration to point mercurial at homebrew's CA certificates file?

DomT4:
Probably. I've got:
```
[web]
cacerts = /usr/local/etc/openssl/cert.pem
```
in my `.hgrc`.

I left a comment asking about whether we could tweak this in a `hgrc` file in `/usr/local/etc` rather than having to modify one in `$HOME` which violates our sandbox in https://github.com/Homebrew/homebrew-core/issues/3541#issuecomment-236964224, but I don't think I got a response.

ngoldbaum:
The resolution order for configuration files is documented here: https://selenic.com/hg/help/config

The relevant bit:

On Unix, the following files are consulted:

* "<repo>/.hg/hgrc" (per-repository)
* "$HOME/.hgrc" (per-user)
* "<install-root>/etc/mercurial/hgrc" (per-installation)
* "<install-root>/etc/mercurial/hgrc.d/*.rc" (per-installation)
* "/etc/mercurial/hgrc" (per-system)
* "/etc/mercurial/hgrc.d/*.rc" (per-system)
* "<internal>/default.d/*.rc" (defaults)

globau:
using homebrew's python works as long as you don't have ```cacerts``` set anywhere.

my ~/.hgrc file had (not sure where it came from):
```
[web]
cacerts = /etc/hg-dummy-cert.pem
```
this resulted in:
```
abort: error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)
```

creating an ```.rc``` file in ```/usr/local/etc/mercurial/hgrc.d``` that set ```cacerts = /usr/local/etc/openssl/cert.pem``` didn't work due to ```~/.hgrc``` being read first.

removing all config entries that set ```cacerts``` fixed the issue (this included deleting the file in ```hgrc.d```).

perhaps this needs a warning on deploy, instructing users to remove ```[web]cacerts``` from their ```~/.hgrc``` if present?

tdsmith:
is there a mercurial subcommand we can run to test for that?

globau:
```
$ hg config web.cacerts
/etc/hg-dummy-cert.pem
$
```
and with it commented out:
```
$ hg config web.cacerts
$
```

indygreg:
```
$ hg init empty
$ cd empty
$ hg pull https://hg.mozilla.org/mozilla-unified
(you can cancel this command before it finishes running)
```

If that works without any warnings about TLS or CA certificates, that's OK'ish. Although installing a hgrc in `<install-root>/etc/mercurial/hgrc` is highly recommended to avoid ambiguity.

@globau: you likely had that hg-dummy-cert.pem set by some old instructions you read and followed. Apple's Python/OpenSSL has this hack where the loading of a specifically crafted x509 certificate triggers the loading of CA certificates from the system keychain. Mercurial started using this hack itself a number of releases ago. But before that people were instructed to perform this hack manually.

Regarding loading of configs, Mercurial assembles a list of config locations then loads them from most global to most local. Last write wins. So `/etc/mercurial/hgrc` ->`<install-root>/etc/mercurial/hgrc` -> `~/.hgrc` -> `.hg/hgrc`. See `mercurial.scmutil.osrcpath()`. The system and user paths are platform dependent. `hg config` will print the active config.

tdsmith:
"(you can cancel this command before it finishes running)"

Heh. I'd hope.

indygreg:
FWIW, I intentionally instructed you to pull from that (very large) repo because the repo is using Mercurial's "clone bundles" feature to redirect clones to a separate server to serve a pre-generated bundle file. And, the preferred server (https://hg.cdn.mozilla.net/) requires SNI. So the output should be something like:

```
pulling from https://hg.mozilla.org/mozilla-unified
applying clone bundle from https://hg.cdn.mozilla.net/mozilla-unified/5ce69082c2fb2468505ef4aa7a1ae6c14b7baae5.gzip-v2.hg
adding changesets
changesets [                   <=>                                                                              ] 46366
```
If Python/OpenSSL doesn't support SNI, Mercurial will fall back to an Amazon S3 URL... and there is another bug to be filed because SNI should work with modern Python/OpenSSL :)

Formula/mercurial.rb
```diff
@@ -13,6 +13,13 @@ class Mercurial < Formula
     sha256 "8934a8fc26c36e56f017d99ac43a0ebe9d6922f8276c1dfccd9ccb980ae05e45" => :mavericks
   end
 
+  if MacOS.version <= :snow_leopard
+    depends_on :python
```
MikeMcQuaid: Could we follow this pattern instead: https://github.com/Homebrew/homebrew-core/blob/c91fecf77a056011802d81582952d2e9623838d4/Formula/macvim.rb#L27-L31 for MacVim where we're doing a similar thing?

ilovezfs:
Should this `depend_on "python"` with an escape hatch specifically for system python, as opposed to "punishing" `--build-from-source` users with system python via `:python`? I don't think most users know they can do `--build-bottle` + postinstall etc.

MikeMcQuaid:
@ilovezfs That was my thoughts with the inline comment; make it match `MacVim` in that regard.

MikeMcQuaid:
@ilovezfs Actually, no, it's not quite the same and I agree with your way 👍

DomT4:
This is probably worth a revision bump, as well, IMO.

tdsmith:
Feedback addressed; should now be less mysterious.

Formula/mercurial.rb
```diff
@@ -13,6 +14,14 @@ class Mercurial < Formula
     sha256 "8934a8fc26c36e56f017d99ac43a0ebe9d6922f8276c1dfccd9ccb980ae05e45" => :mavericks
   end
 
+  option "with-custom-python", "Install against the python in PATH instead of "\
```
DomT4: I'd really prefer we didn't do this with options. Can we either EOS them, i.e.:
```ruby
  option "with-custom-python", <<-EOS.undent
    Install against the python in PATH instead of Homebrew's python
  EOS
```
or keep them on one line. It's possibly just me who finds `\` grim, but I'm unconvinced it's all that readable at a glance.

Formula/mercurial.rb
```diff
@@ -26,6 +35,20 @@ def install
     zsh_completion.install "contrib/zsh_completion" => "_hg"
   end
 
+  def caveats
+    return unless (opt_bin/"hg").exist?
+    cacerts_configured = `#{opt_bin}/hg config web.cacerts`.strip
+    return if cacerts_configured == ""
```
DomT4: `return if cacerts_configured.empty?`

?

Formula/mercurial.rb
```diff
@@ -13,6 +14,14 @@ class Mercurial < Formula
     sha256 "8934a8fc26c36e56f017d99ac43a0ebe9d6922f8276c1dfccd9ccb980ae05e45" => :mavericks
   end
 
+  option "with-custom-python", "Install against the python in PATH instead of "\
+                               "Homebrew's python"
+  if build.with? "custom-python"
```
MikeMcQuaid: If we can figure out a way to avoid needing the `build.with?` here that'd be good as we all know it causes breakage.

Formula/mercurial.rb
ilovezfs: I'd prefer we just make the text more concise so it fits on a single line always.

tdsmith: i like the continuation better than a heredoc :p but this basically fits on a line; if it runs longer than 80 characters will anyone be mad at me

ilovezfs: You can keep it under 80. I have confidence in your optionhaikufu.

Formula/mercurial.rb
tdsmith: We want different dependencies depending on an option that doesn't fit the :recommended form so I think that's going to be hard to swing; open to ideas.

Formula/mercurial.rb
DomT4: I don't have a massive objection to letting lines stretch to 83/84 characters if necessary. I'll permit 5% 😈

ilovezfs: I don't believe you for even a second.

DomT4: Ye of little faith 😇

PR #3552, opened 2016-08-03
gogs 0.9.48

brewingcode:

---

### DIFF
--- a/Formula/gogs.rb
+++ b/Formula/gogs.rb
@@ -1,0 +1,23 @@
+class Gogs < Formula
+  desc "Go Git Service: A painless self-hosted Git service"
+  homepage "https://gogs.io"
+  url "https://cdn.gogs.io/gogs_v0.9.48_darwin_amd64.zip"
+  version "0.9.48"
+  sha256 "8afbd71fde279e77ed5b16c8a80ef409661ccb3953bf6cc867fbede532686475"
+
+  def startup_script; <<-EOS.undent
+    #!/bin/sh
+    cd #{var/"gogs"} && ./gogs "$@"
+    EOS
+  end
+
+  def install
+    (var/"gogs").mkpath
+    (var/"gogs").install Dir["*"]
+    (bin/"gogs").write startup_script
+  end
+
+  test do
+    system "gogs", "--version"
+  end
+end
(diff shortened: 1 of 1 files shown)


Formula/gogs.rb
```diff
@@ -0,0 +1,41 @@
+class Gogs < Formula
+  desc "Go Git Service: A painless self-hosted Git service"
+  homepage "https://gogs.io"
+  url "https://cdn.gogs.io/gogs_v0.9.48_darwin_amd64.zip"
```
DomT4: We don't take binaries here. This will need to be built from source.

Formula/gogs.rb
```diff
@@ -0,0 +1,41 @@
...
+    [ -d gogs.bak ] && { echo "#{var}/gogs.bak already exists, quitting" >&2; exit 1; }
+    mv gogs gogs.bak
+    brew reinstall gogs
+    for d in custom data log; do [ -d gogs.bak/$d ] && cp -R gogs.bak/$d gogs; done
+    EOS
+  end
+
+  def install
+    (var/"gogs").mkpath
+    (var/"gogs").install Dir["*"]
```
DomT4: We won't accept a formula that installs the "true" components completely outside the keg. That nullifies a lot of the upsides of package management.

Formula/gogs.rb
brewingcode: No problem, I'll just cancel the PR. Building from source involves a lot of gymnastic with go and its build system, the binary install is *far* simpler.

Formula/gogs.rb
brewingcode: Yeah, the behavior of the binary requires that its state and binaries are in the same directory, which makes upgrading difficult with `brew`, since it completely wipes the components for each version. Guess `gogs` is just shit outta luck for a binary `brew`.

brewingcode:
Closing, not a good fit for `brew install`.

Formula/gogs.rb
DomT4: I see you've already filed a Cask [here](https://github.com/caskroom/homebrew-cask/pull/23388), which I agree is probably the right thing to do 👍.

PR #3556, opened 2016-08-03
dspdfviewer: use bottle :disabled

ilovezfs:


DomT4:
I disagree with this PR for the reasons referenced in https://github.com/Homebrew/homebrew-core/pull/3234#issuecomment-237106014, and I think it's generally bad UX to disable bottles simply to resolve CI issues, although there's precedent.

I think equally valid would just be removing `qt` support from `poppler`:

screen shot 2016-08-03 at 02 36 50

ilovezfs:
removing qt support from poppler will kill this formula entirely, no?

ilovezfs:
@DomT4 or do you mean kill the Qt support but keep the Qt 5 support?

DomT4:
No. This formula uses the `qt5` option of `poppler`, not the `qt` option. You may be thinking of `frescobaldi`, which would likely temporarily end up in the 💀 until the Qt5 releases are available.

ilovezfs:
Right, you have to pick one to kill. Given that this formula took 9 seconds to build for me, I think it's worse UX to ship frescobaldi off for a temporary tour in the boneyard.

ilovezfs:
@DomT4 what are the download numbers for this bottle?

DomT4:
Given we're less than 2 months away from Sierra going live for everyone, at which point killing off `qt` options in favour of `qt5` ones is going to _have_ to go faster, I don't think we're particularly jumping the shark there unless `frescobaldi` is planning a Qt5 release in the next couple months.

> what are the download numbers for this bottle?

110 installations here, 73 for `frescobaldi`.

ilovezfs:
Since the part of the build that takes time is poppler, which has to be built from source unless we're going to make it `depends_on "qt5" => :recommended`, this seems like a very small additional burden given that this bottle isn't doing anything to improve the total install time or reduce the need for a source build.

ilovezfs:
Would you support making poppler `depends_on "qt5" => :recommended`?

DomT4:
> this seems like a very small additional burden given that this bottle isn't doing anything to improve the total install time or reduce the need for a source build.

My point isn't particularly burden. It's more that it's bad UX & runs counter to the push to bottle everything that uses a OS X/macOS development tool desire.

> Would you support making poppler depends_on "qt5" => :recommended?

I don't think that's merited on the current analytics numbers. Frankly on the current analytics numbers you could make a case `poppler` shouldn't have any options at all, but there we go 😄.

ilovezfs:
> Frankly on the current analytics numbers you could make a case poppler shouldn't have any options at all, but there we go

Well that's where I was going with it. I'd make poppler `depends_on "qt5"`, no options, and ship frescobaldi off to the boneyard. Qt 5 über alles.

ilovezfs:
@DomT4 Actually, it looks like frescobaldi might be a good candidate for Cask: 

http://www.frescobaldi.org/download
https://github.com/wbsoft/frescobaldi/releases

DomT4:
> Actually, it looks like frescobaldi might be a good candidate for Cask:

That'd seem reasonable, to be fair, yeah. Nothing seems to use it in Homebrew, so that'd be safe. Maybe @AnastasiaSulyagina has some thoughts here.

AnastasiaSulyagina:
It seems to be more cask than formula, yeh.
I could make a cask and migrate it.

ilovezfs:
>I could make a cask and migrate it. 

Sounds good to me! @MikeMcQuaid thoughts?

MikeMcQuaid:
👍 from me but you probably need to make sure it installs Lilypond too.

I'm also 👎 on disabling bottles to work around a CI bug. Can you elaborate a bit more on the problem here and we can think of a smarter solution to this case? Also agreed to removing `qt` support much more aggressively now, even if it means widespread boneyarding.

ilovezfs:
@MikeMcQuaid You can see the CI failure here:

http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5725/version=el_capitan/console
http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5725/version=yosemite/console
http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5725/version=mavericks/console

The error is
```
Error: poppler: --with-qt and --with-qt5 cannot be used at the same time
```
which is triggered by
```
brew install frescobaldi
```

It dies here: https://github.com/Homebrew/homebrew-core/blob/master/Formula/poppler.rb#L57

The cause is that higher up, CI has already done
```
==> brew install dspdfviewer
```
and the `dspdfviewer` formula has
```
  depends_on "poppler" => "with-qt5"
```
So poppler has already been installed by the time we get to the `brew install frescobaldi`.

Because `brew test-bot` does all uninstalls at the end, rather than uninstalling everything installed after each test, the `brew install frescobaldi` fails because it's unable to install its dependency which is
```
depends_on "poppler" => "with-qt"
```

If test-bot ran them in the reverse order, frescobaldi, then dspdfviewer, it would be frescobaldi that succeeded and dspdfviewer that failed since they're just competing to go first.

MikeMcQuaid:
@ilovezfs Ok, cool, so it's an issue with the bot not uninstalling/reinstalling with the right options. I'm confused as to why it's trying to use both options, though?

ilovezfs:
@MikeMcQuaid the problem is that the formula is calling `raise` in the middle of install, which prevents the user from accidentally breaking the pre-existing installation of dspdfviewer, since it will break if poppler is reinstalled with-qt4 instead. However, in the context of CI, that raise is totally bogus, and if the poppler formula would just get out of its own way, test-bot would have no trouble proceeding with successfully reinstalling poppler with the appropriate option. The immediate "bug" here IMO is that formulae should *never* have `raise` in their install method and the qt4/qt5 (or fooX/fooY) conflict should be handled explicitly with some kind of conflicts_with or whatever in a non-adhoc fashion.

ilovezfs:
@MikeMcQuaid if you run `brew test-bot poppler` you can see the failure. Here's what happens:
```
iMac-TMP:local joe$ brew test-bot poppler
==> brew doctor
==> brew --env
==> brew config
==> brew tests --official-cmd-taps
==> brew tests --generic --official-cmd-taps
==> brew tests --no-compat
==> brew readall --syntax
==> brew uses poppler
==> brew fetch --retry git
==> brew fetch --retry poppler --build-bottle
==> brew install --only-dependencies --build-bottle --verbose poppler
==> brew install --build-bottle --verbose poppler
==> brew audit poppler
==> brew bottle --verbose --json poppler
==> brew bottle --merge --write --no-commit ./poppler-0.46.0_1.el_capitan.bottle.json
==> brew uninstall --force poppler
==> brew uninstall --force git
==> brew install ./poppler-0.46.0_1.el_capitan.bottle.1.tar.gz
==> brew test poppler --verbose
==> brew fetch --retry diff-pdf
==> brew install diff-pdf
==> brew linkage --test diff-pdf
==> brew test --verbose diff-pdf
==> brew fetch --retry dspdfviewer
==> brew install dspdfviewer
==> brew linkage --test dspdfviewer
==> brew test --verbose dspdfviewer
==> brew fetch --retry evince
==> brew install evince
==> brew linkage --test evince
==> brew test --verbose evince
==> brew fetch --retry frescobaldi
==> brew install frescobaldi
==> FAILED
==> Installing dependencies for frescobaldi: poppler
==> Installing frescobaldi dependency: poppler
==> Using the sandbox
==> Downloading https://poppler.freedesktop.org/poppler-0.46.0.tar.xz
Already downloaded: /Users/joe/Library/Caches/Homebrew/poppler-0.46.0.tar.xz
Error: poppler: --with-qt and --with-qt5 cannot be used at the same time
==> brew fetch --retry mat
==> brew install mat
==> brew linkage --test mat
==> brew test --verbose mat
==> brew fetch --retry pdf2htmlex
==> brew install pdf2htmlex
==> brew linkage --test pdf2htmlex
==> brew test --verbose pdf2htmlex
==> brew fetch --retry pdf2svg
==> brew install pdf2svg
==> brew linkage --test pdf2svg
==> brew test --verbose pdf2svg
==> brew fetch --retry pdfgrep
==> brew install pdfgrep
==> brew linkage --test pdfgrep
==> brew test --verbose pdfgrep
==> brew fetch --retry xournal
==> brew install xournal
==> brew linkage --test xournal
==> brew test --verbose xournal
==> brew fetch --retry vips
==> brew install vips
==> brew linkage --test vips
==> brew test --verbose vips
==> brew uninstall --force poppler
iMac-TMP:local joe$
```

ilovezfs:
Given that the tests are supposedly testing the just built bottle, it is pretty "strange behavior" that after
```
==> brew install dspdfviewer
```
installs a non-default version of poppler, basically doing`brew install --with-qt5 poppler`, that test-bot does not unwind that and reinstall the bottle before proceeding. At that point it has effectively stopped testing the bottle! In this case we're butting heads with another formula that wants to do its own non-default `brew install --with-qt poppler`, but the even more concerning thing is that all the tests after `brew install dspdfviewer` have stopped testing the bottled version.

ilovezfs:
@MikeMcQuaid to answer your question about why both options, the reason it's trying to use both at once is this line: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula_installer.rb#L370

MikeMcQuaid:
@ilovezfs I think you can maybe override the tab there somehow in the case of these conflicting. Agreed a DSL is The Right Way.

ilovezfs:
Yeah, "honoring the tab" during CI at all seems weird.

ilovezfs:
@MikeMcQuaid it's also "too late" to handle the ignore-the-tab issue in the poppler formula itself because superenv has already set up a bogus environment including both Qt versions.

MikeMcQuaid:
> Yeah, "honoring the tab" during CI at all seems weird.

Optionally disabling that would be handy 👍 

Otherwise your suggested current workaround where you provide both `with-qt5` and `without-qt` seems to make sense.

PR #3571, opened 2016-08-03
supervisor 3.3.1

ilovezfs:
use new virtualenv feature

### DIFF
--- a/Formula/supervisor.rb
+++ b/Formula/supervisor.rb
@@ -1,8 +1,10 @@
 class Supervisor < Formula
+  include Language::Python::Virtualenv
+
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @tdsmith @DomT4

DomT4:
Would be nice to standardise where we place `Language::Python::Virtualenv`. We put it underneath the formula name for GHC stuff, perhaps it's worth doing the same thing here.

That nit beside, everything else seems good to me.

ilovezfs:
@DomT4 sounds like it would be worth having an audit check to enforce that ordering.

In any case, fixed.

PR #3573, opened 2016-08-03
git-remote-hg 0.3 (new formula)

DoomHammer:

---

Nice bridge allowing use of familiar `git` commands with Mercurial repositories.

### DIFF
--- a/Formula/git-remote-hg.rb
+++ b/Formula/git-remote-hg.rb
@@ -1,0 +1,18 @@
+class GitRemoteHg < Formula
+  desc "Transparent bidirectional bridge between Git and Mercurial"
+  homepage "https://github.com/felipec/git-remote-hg"
+  url "https://github.com/felipec/git-remote-hg/archive/v0.3.tar.gz"
+  sha256 "2dc889b641d72f5a73c4c7d5df3b8ea788e75a7ce80f5884a7a8d2e099287dce"
+
+  depends_on :hg
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    inreplace "git-remote-hg", "python2", "python"
+    system "make", "install", "prefix=#{prefix}"
+  end
+
+  test do
+    system "git", "clone", "hg::https://www.mercurial-scm.org/repo/hello"
+  end
+end


MikeMcQuaid:
> Does your submission pass brew audit --strict --online <formula> (after doing brew install <formula>)?

It does not.

DoomHammer:
Might be discrepancies between Linuxbrew that I use and Homebrew. Will try to fix any problems.

DoomHammer:
It does, now. Sorry about that.

Formula/git-remote-hg.rb
```diff
@@ -0,0 +1,17 @@
+class GitRemoteHg < Formula
+  desc "Transparent bidirectional bridge between Git and Mercurial"
+  homepage "https://github.com/felipec/git-remote-hg"
+  url "https://github.com/felipec/git-remote-hg/archive/v0.3.tar.gz"
+  sha256 "2dc889b641d72f5a73c4c7d5df3b8ea788e75a7ce80f5884a7a8d2e099287dce"
+
+  depends_on :hg
+  depends_on :python
```
DomT4: Should be:
```
```
I believe, unless https://github.com/Homebrew/homebrew-core/pull/3548 has an impact on other things that tie `hg` and Python together.

DoomHammer: Tried this, but `git-remote-hg` needs `python2` binary so this `if` unfortunately breaks things as seen in http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5779/ . What's the best approach in such case?

DomT4: Ah, that old chestnut. We `inreplace` out the `python2` for `python` in that case. We do so already for `src`, `duplicity`, probably a fair few others.

@tdsmith Just to check, is there any risk mixing Python interpreters here, if we switch our `mercurial` to use our Python & this uses the system Python instead?

DoomHammer: Here.

Formula/git-remote-hg.rb
```diff
@@ -0,0 +1,18 @@
...
+  desc "Transparent bidirectional bridge between Git and Mercurial"
+  homepage "https://github.com/felipec/git-remote-hg"
+  url "https://github.com/felipec/git-remote-hg/archive/v0.3.tar.gz"
+  sha256 "2dc889b641d72f5a73c4c7d5df3b8ea788e75a7ce80f5884a7a8d2e099287dce"
+
+  depends_on :hg
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    inreplace "git-remote-hg", "python2", "python"
```
MikeMcQuaid: Please consider reporting this to the upstream project.

MikeMcQuaid:
Thanks again @DoomHammer!

chulito1:
Unistall chulito1

Formula/git-remote-hg.rb
DoomHammer: I can report it, but I believe this is done on purpose rather than as a mistake.

MikeMcQuaid: It may well be but if they want to support OS X they should probably know that `python2` isn't provided there.

DoomHammer: Here it is: https://github.com/felipec/git-remote-hg/pull/63

DomT4: I'd be surprised if they accept this to be honest. I believe on some Linux distros `#!/usr/bin/env python` points to `python3` now whereas `python2` is always "safe", so that's likely the reasoning.

OS X isn't particularly ahead of the curve. I tend to consider if a Apple bug rather than an upstream one.

MikeMcQuaid: It's still worth asking I think; even if it's an Apple bug it's silly for all OS X package managers to have to patch this indefinitely.

tdsmith: FWIW Apple does provide a `python2.7` which is the only 2.x version to use anyway.

PR #3579, opened 2016-08-03
m-cli 0.1.1

rgcr:

---

This update also disables some options like `update`, `uninstall` that must be handled by `brew` if the package is installed from there.

### DIFF
--- a/Formula/m-cli.rb
+++ b/Formula/m-cli.rb
@@ -1,17 +1,20 @@
 class MCli < Formula
   desc "Swiss Army Knife for Mac OS X"
   homepage "https://github.com/rgcr/m-cli"
-  url "https://github.com/rgcr/m-cli/archive/v0.1.0.tar.gz"
-  sha256 "d28f9b4abdb783554eee01432ffb55dbd206f018b6825018f00fe722af799963"
+  url "https://github.com/rgcr/m-cli/archive/v0.1.1.tar.gz"
+  sha256 "66c4b158c1fdfa976a33c97069ae96c2428ca92f5a01b27430ef36930a267d53"
 
   bottle :unneeded
 
   def install
     prefix.install Dir["*"]
-    # Use absolute rather than relative path to plugins.
     inreplace prefix/"m" do |s|
+      # Use absolute rather than relative path to plugins.
       s.gsub! /^\[ -L.*|^\s+\|\| pushd.*|^popd.*/, ""
       s.gsub! /MPATH=.*/, "MPATH=#{prefix}"
+      # Disable options "update" && "uninstall", they must be handled by brew
+      s.gsub! /update_mcli \&\&.*/, "printf \"Try: brew update && brew upgrade m-cli \\n\" && exit 0"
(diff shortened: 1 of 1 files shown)


Formula/m-cli.rb
```diff
@@ -1,17 +1,20 @@
...
   def install
     prefix.install Dir["*"]
-    # Use absolute rather than relative path to plugins.
     inreplace prefix/"m" do |s|
+      # Use absolute rather than relative path to plugins.
       s.gsub! /^\[ -L.*|^\s+\|\| pushd.*|^popd.*/, ""
       s.gsub! /MPATH=.*/, "MPATH=#{prefix}"
+      # Disable options "update" && "uninstall", they must be handled by brew
+      s.gsub! /update_mcli \&\&.*/, "printf \"Try: brew update && brew upgrade m-cli \\n\" && exit 0"
+      s.gsub! /uninstall_mcli \&\&.*/, "printf \"Try: brew uninstall m-cli \\n\" && exit 0"
```
MikeMcQuaid: Can you please submit these patches (or at least issues requesting this functionality) to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

rgcr:
This is the issue rgcr/m-cli/issues/66

Basically I need to change this, just to show a little message  because the update process does
`git fetch --depth=1 && git reset --hard origin/master` and the uninstall process does a simple `rm -rf` to the folder, so I think it's a good idea to disable those options if it's installed via `brew`.

```diff
diff --git a/m b/m
index 523ef3b..d1bc615 100755
--- a/m
+++ b/m
@@ -56,10 +56,10 @@ __EOF__

 case $1 in
     --update)
-        update_mcli && exit 0
+        printf "Try: brew update && brew upgrade m-cli \n"  && exit 0
         ;;
     --uninstall)
-        uninstall_mcli && exit 0
+        printf "Try: brew uninstall m-cli \n" && exit 0
         ;;
 esac
```

MikeMcQuaid:
@rgcr Oh, sorry, I see, you are upstream! Yeh, ideally all the current `inreplace`s would be replaced with something that allows configuration at install-time beyond just patching the source code.

rgcr:
@MikeMcQuaid, I know but the main script is just a wrapper to call other scripts(plugins) and It will never be changed again once I fix this, also I think changing a few lines in the main script for brew it's easier than create or modify the entire process that depends of git. Thank you for your attention :)  

Here is a previous discussion about inreplace in the formula
Homebrew/homebrew-core/pull/3137

DomT4:
@MikeMcQuaid There's precedent for this sort of thing with `heroku`, FWIW. Generally we permit these inreplaces where upstream doesn't have a `Makefile` or build script that handles it.

I can probably put up with it unless upstream want to adopt a flexible `Makefile`, unless you feel more strongly than I do.

MikeMcQuaid:
Fine with merging as-is for now as @rgcr is submitting PRs himself anyway. Thanks for the PR ✨!

PR #3581, opened 2016-08-03
go: bootstrap rewrite

DomT4:

---

**go: limit to 10.8 & above:**
Upstream don't test or support anything less than 10.8 at this point. There are a "known" 141 installations below that out of 45k, so doesn't seem to be huge demand.

This isn't _absolutely_ necessary; we could keep carrying it for 10.6 & below till it completely breaks, or we could cap at 10.7 instead of 10.8. I'd like to cap it at least to 10.7 to avoid this formula getting wildly complicated around which bootstrap to use. Open to thoughts here.

**go: rewrite bootstrap steps:**
- Use more recent versions to handle bootstrapping.
- Drop the unnecessary compile step in the bootstrap versions of Go, which are
  already binaries. Ref: golang/go#16579 (comment).
- Cap stable to 10.8-10.11 due to broken build on macOS Sierra/10.12.
- Add explicit versions to the bootstrap to stop them being cached incorrectly
  as `go--gobootstrap-64.tar.gz`.

### DIFF
--- a/Formula/go.rb
+++ b/Formula/go.rb
@@ -8,6 +8,17 @@ class Go < Formula
     version "1.6.3"
     sha256 "6326aeed5f86cf18f16d6dc831405614f855e2d416a91fd3fdc334f772345b00"
 
+    # 1.6.3 does not build on macOS Sierra. Users should use devel instead
+    # until 1.7 is stable (due soon).
+    depends_on MaximumMacOSRequirement => :el_capitan
+
+    # Should use the last stable binary release to bootstrap.
+    resource "gobootstrap" do
+      url "https://storage.googleapis.com/golang/go1.6.2.darwin-amd64.tar.gz"
+      version "1.6.2"
+      sha256 "6ebbafcac53bbbf8c4105fa84b63cca3d6ce04370f5a04ac2ac065782397fc26"
+    end
+
     go_version = "1.6"
     resource "gotools" do
       url "https://go.googlesource.com/tools.git",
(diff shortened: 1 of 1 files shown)


Formula/go.rb
```diff
@@ -8,6 +8,17 @@ class Go < Formula
...
     sha256 "6326aeed5f86cf18f16d6dc831405614f855e2d416a91fd3fdc334f772345b00"
 
+    # 1.6.3 does not build on macOS Sierra. Users should use devel instead
+    # until 1.7 is stable (due soon).
+    depends_on MaximumMacOSRequirement => :el_capitan
+
+    # Should use the last stable binary release to bootstrap.
+    resource "gobootstrap" do
+      url "https://storage.googleapis.com/golang/go1.6.2.darwin-amd64.tar.gz"
+      version "1.6.2"
```
dunn: Any reason not to use the current release for bootstrapping?

Formula/go.rb
```diff
@@ -8,6 +8,17 @@ class Go < Formula
     version "1.6.3"
     sha256 "6326aeed5f86cf18f16d6dc831405614f855e2d416a91fd3fdc334f772345b00"
 
+    # 1.6.3 does not build on macOS Sierra. Users should use devel instead
+    # until 1.7 is stable (due soon).
+    depends_on MaximumMacOSRequirement => :el_capitan
+
+    # Should use the last stable binary release to bootstrap.
+    resource "gobootstrap" do
+      url "https://storage.googleapis.com/golang/go1.6.2.darwin-amd64.tar.gz"
```
zackehh: Should this be `1.6.3`?

Formula/go.rb
DomT4: Technically speaking, I would presume there's no reason why you couldn't bootstrap with the current release, but I haven't tested this, and it's useful to bootstrap with a version we _know_ works as expected and didn't cause any compile bugs. That way we should always know any compile bugs were introduced with the latest release.

Another minor consideration is political, to an extent. I wanted to avoid the _"If you can use the binary for the current release to bootstrap, why can't you just use the binary completely rather than wasting my life building from source"_ kind of discussion that pops up every now & again.

Formula/go.rb
DomT4: See the comments just below between Alex & myself.

Formula/go.rb
dunn: 👍

MikeMcQuaid:
Makes sense 👍

cwonrails:
Great work! When will this be merged?

DomT4:
Sorry for the delay here, I've been away from the laptop all day till now. Thanks to @MikeMcQuaid for merging.

PR #3584, opened 2016-08-03
yara 3.5.0

ilovezfs:
- remove upstreamed patch
- gnu89 no longer needed
- yara-python was moved out of the yara repository

### DIFF
--- a/Formula/yara.rb
+++ b/Formula/yara.rb
@@ -1,18 +1,9 @@
 class Yara < Formula
   desc "Malware identification and classification tool"
   homepage "https://github.com/VirusTotal/yara/"
+  url "https://github.com/VirusTotal/yara/archive/v3.5.0.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 yara-python was moved to a separate repo: https://github.com/VirusTotal/yara-python. That separate repo that vendors yara itself as a submodule, or can optionally use an independently built yara. yara-python is also installable via `pip install yara-python`, so there are several ways this could be handled. My initial inclination is just to drop yara-python and see if anyone shows up asking for it back, but it could be vendored as a resource or transformed into the main piece of the formula pulling in yara as a submodule or moved to a separate formula with a yara dependency or with the vendored yara submodule or or or...

tdsmith:
>  My initial inclination is just to drop yara-python

:+1:

DomT4:
👍. Tim's agreement on dropping the Python element unless we get people banging on the door asking for it back is good enough for me.

PR #3589, opened 2016-08-04
ucspi-tools: migrate to boneyard

DomT4:

---

Upstream hasn't cut a release in over 18 months. In the mean time this has gone through 10 formula revisions & has a whole vendored LibreSSL because upstream isn't keeping up with LibreSSL development, sadly, which is rapid & regularly breaking.

I'm uncomfortable keeping an unsupported LibreSSL buried inside this formula when it looks increasingly like that's going to become a permanent fixture rather than the temporary hack it was intended to be when inserted initially.

We wouldn't accept this formula today due to notability issues in our audit and our usage data shows only 18 "known" installations since analytics were switched on a few months ago.

Let's remove this from Homebrew until the situation improves in terms of mandating LibreSSL but not matching its release cadence.


MikeMcQuaid:
Kill it.

PR #3597, opened 2016-08-04
s-search 0.5.5

ilovezfs:
gvt vendoring

### DIFF
--- a/Formula/s-search.rb
+++ b/Formula/s-search.rb
@@ -3,9 +3,18 @@ require "language/go"
 class SSearch < Formula
   desc "Web search from the terminal"
   homepage "https://github.com/zquestz/s"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 vendor!

DomT4:
They would be using one of the few popular vendoring systems we don't have a formula for 😅. Nice work here 👍

ilovezfs:
@DomT4 yeah I was considering asking for a tag for gvt to create a formula, but it didn't look like anything else in core is using it (yet).

DomT4:
Aye. It's a useful tool, and possibly the last _big_ Go vendoring helper I know a lot of people recommend that we don't ship a formula for. Maybe worthwhile.

PR #3603, opened 2016-08-04
lego 0.3.0 (new formula)

tdsmith:
Let's Encrypt client written in go.

 Before you ask: no, I think we can't have a better test here; actually doing any work means the test would have to register a user account on a Let's Encrypt endpoint.

### DIFF
--- a/Formula/lego.rb
+++ b/Formula/lego.rb
@@ -1,0 +1,94 @@
+require "language/go"
+
+class Lego < Formula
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Decided to pull this as it's a documented requirement now for our Ops workflow.

tdsmith:
:+1:, thanks!

PR #3607, opened 2016-08-04
efl: fix array brace position

miguelarauj1o:

---

### DIFF
--- a/Formula/efl.rb
+++ b/Formula/efl.rb
@@ -40,8 +40,7 @@ class Efl < Formula
             "--disable-silent-rules",
             "--enable-cocoa",
             "--enable-i-really-know-what-i-am-doing-and-that-this-will-probably-break-things-and-i-will-fix-them-myself-and-send-patches-aba", # There's currently (1.14) no clean profile for Mac OS, so we need to force passing configure
-            "--prefix=#{prefix}",
-           ]
+            "--prefix=#{prefix}"]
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@UniqMartin Out of interest: do you agree with this Rubocop check? I'm not sure if I do 😒

UniqMartin:
@MikeMcQuaid I do, in the sense that it enforces a more consistent brace placement. But I don't agree with the proposed fix, though that's the one with a minimal diff. I think in almost all cases we want to have either:

```ruby
args = [
  "--foo",
  "--bar",
  "--baz",
]
```

Or even better:

```ruby
args = %W[
  --foo
  --bar
  --baz
]
```

(The latter is what we tend to use when changing these kinds of arrays or in new formulae, as far as I can tell.)

DomT4:
> (The latter is what we tend to use when changing these kinds of arrays or in new formulae, as far as I can tell.)

👍 x 💯 x 💯

DomT4:
Made some changes separately afterwards because there were other things annoying me in the formula, which you can see in https://github.com/Homebrew/homebrew-core/commit/6f2678fa1e03fb8f724aae0c2ffc18145c945927. 

Merged this PR in https://github.com/Homebrew/homebrew-core/commit/944f300b2931ec84077fe45e96a1d3a8ee03d4c8. Thanks for helping with the style cleanup @miguelarauj1o! 😺

PR #3613, opened 2016-08-04
sonar 6.0

Godin:
Also fix Java version - Java 1.8 is required starting from SonarQube 5.6.

### DIFF
--- a/Formula/sonar.rb
+++ b/Formula/sonar.rb
@@ -1,10 +1,10 @@
 class Sonar < Formula
   desc "Manage code quality"
   homepage "http://www.sonarqube.org/"
-  url "https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-5.6.1.zip"
-  sha256 "9cb74cd00904e7c804deb3b31158dc8651a41cbe07e9253c53c9b11c9903c9b1"
+  url "https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-6.0.zip"
+  sha256 "2256e0b44629632d9528911c8d5916ce246c7abe2be65a6a59a94dee504d62a6"
 
-  depends_on :java => "1.7+"
+  depends_on :java => "1.8+"
 
   bottle :unneeded
 


MikeMcQuaid:
CC @xu-cheng for thoughts on the failure here; could it be updated to state which file has that linkage?

DomT4:
I ignored this the other day; I think it's an upstream packaging problem which isn't actionable by us, although potentially should be reported upstream. I'll hunt through the upstream download in a bit & try to find where the link is coming from.

Godin:
@DomT4 if by upstream here you mean SonarQube, then my best guess that this link might come from its usage of Java Service Wrapper. I could check this deeper tomorrow.

Godin:
I confirm that broken link comes from Java Service Wrapper:
```
$ otool -L -arch ppc64 sonarqube-6.0/bin/macosx-universal-64/wrapper
sonarqube-6.0/bin/macosx-universal-64/wrapper:
	/usr/lib/libgcc_s_ppc64.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 111.1.4)

$ otool -L -arch ppc64 sonarqube-6.0/bin/macosx-universal-64/lib/libwrapper.jnilib
sonarqube-6.0/bin/macosx-universal-64/lib/libwrapper.jnilib:
	/usr/lib/libgcc_s_ppc64.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 111.1.4)
```

Not sure that we (SonarQube) can fix this in short/mid term, because we use Java Service Wrapper provided only as a binary distribution - https://github.com/sonatype/jsw-binaries , and it doesn't receive any updates/changes due to the change of JSW license - this is the last version, which is compatible with LGPL.

However given that those files are in universal binary formal and contain multiple versions for different architectures:
```
$ otool -L -arch x86_64 sonarqube-6.0/bin/macosx-universal-64/lib/libwrapper.jnilib
sonarqube-6.0/bin/macosx-universal-64/lib/libwrapper.jnilib:
	/usr/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 111.1.4)

$ otool -L -arch x86_64 sonarqube-6.0/bin/macosx-universal-64/wrapper
sonarqube-6.0/bin/macosx-universal-64/wrapper:
	/usr/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 111.1.4)
```
don't think that this creates any problem - PPC arch never will be selected for Mac OS X >= Lion.

MikeMcQuaid:
> I'll hunt through the upstream download in a bit & try to find where the link is coming from.

Bonus points if you can fix the warning to make that obvious 😉

Godin:
@MikeMcQuaid well, I quickly looked on how this check implemented in brew, so might try, but can't promise 😉

MikeMcQuaid:
@Godin Sorry, that was directed at @DomT4 but you're welcome to do so too!

Godin:
@MikeMcQuaid FWIW I think that this goes down to
https://github.com/Homebrew/brew/blob/29bfc4caa3f7dde97ad4fd62d7da0adededbeede/Library/Homebrew/vendor/macho/macho/fat_file.rb#L159
called from (my naive and just visual backtrace) :
https://github.com/Homebrew/brew/blob/29bfc4caa3f7dde97ad4fd62d7da0adededbeede/Library/Homebrew/os/mac/ruby_mach.rb#L55
https://github.com/Homebrew/brew/blob/91a47a9ed6d85c1bf8f507cdfef4462a07f61831/Library/Homebrew/os/mac/linkage_checker.rb#L26
maybe the first one might simply use
https://github.com/Homebrew/brew/blob/29bfc4caa3f7dde97ad4fd62d7da0adededbeede/Library/Homebrew/vendor/macho/macho/fat_file.rb#L184
instead.

And at this point I'm wondering if relaxation of this warning blocker for this PR ?

MikeMcQuaid:
@Godin Not a blocker, thanks for digging in!

UniqMartin:
Sorry for the late input; I just found this. Assuming you don't actually support PPC64 with the rest of your code, I think the best solution to avoid the audit failure would be to remove the PPC64-specific Mach-O slice from the fat (universal) binaries by using `lipo(1)`. Ideally, this would be already done by upstream if my assessment is correct. We could also do this in the formula, but that would mean it would lose its `bottle :unneeded` status and this would be a pity.

MikeMcQuaid:
@UniqMartin If it's `cellar :any` I think it's fine to lose.

Godin:
@MikeMcQuaid thank you for merge!

@UniqMartin You are right - PPC64 part can be removed. How come I forgot about `lipo`?
I doubt that it was done in https://github.com/sonatype/jsw-binaries , which is used by SQ, because it already uses its latest release - https://github.com/SonarSource/sonarqube/blob/6.0/sonar-application/pom.xml#L127
It might be tricky to promote such change up to sonatype/jsw-binaries since it is probably not maintained having latest release 6 years ago.
But I could try to do this change in SonarQube. This seems more correct than change in formula.

MikeMcQuaid:
> But I could try to do this change in SonarQube. This seems more correct than change in formula.

hqdefault

Godin:
For the record - ticket on SQ side: https://jira.sonarsource.com/browse/SONAR-7959

PR #3629, opened 2016-08-05
you-get 0.4.523

ilovezfs:
use new virtualenv feature

### DIFF
--- a/Formula/you-get.rb
+++ b/Formula/you-get.rb
@@ -1,8 +1,10 @@
 class YouGet < Formula
+  include Language::Python::Virtualenv
+
   desc "Dumb downloader that scrapes the web"
   homepage "https://you-get.org/"
-  url "https://github.com/soimort/you-get/archive/v0.4.486.tar.gz"
-  sha256 "22002542c079ba049a8c2797700d07d5da2e22eb38eeb46ead19c28783855ca1"
+  url "https://files.pythonhosted.org/packages/93/bd/f7a81c7b73d3a7bac9f254ec11854c70ab2f5c22dcdbefbe01573faee014/you-get-0.4.523.tar.gz"
+  sha256 "01e51b48665a7b3de97394c9ad8e34e510664ffdada6b4f1fcf1651977e8347f"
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @UniqMartin @MikeMcQuaid @tdsmith 

Is it just me, or is this really awkward?
```
    virtualenv_create(libexec, "python3")
```

If the formula declares `depends_on :python3`, the `virtualenv_create(libexec, "python3")` seems redundant, but if I don't add that, install fails.

tdsmith:
This is the syntax I anticipated.

ilovezfs:
Why would you anticipate `depends_on :python3` + `virtualenv_install_with_resources` not being sufficient for the python3 formulae? That makes no sense to me.

tdsmith:
Spooky action at a distance, man.

ilovezfs:
Sorry but just defaulting to python2 unconditionally even though the formula explicitly depends on python3 is goofy, to put it gently.

tdsmith:
PRs welcome?

ilovezfs:
I'm not in the habit of opening spooky PRs.

tdsmith:
_That_ is demonstrably false :P

ilovezfs:
@tdsmith https://github.com/Homebrew/brew/pull/636.

UniqMartin:
I have to agree with @ilovezfs that

```ruby
virtualenv_create(libexec, "python3")
```

seems fairly inelegant and hard-coding a preference for Python 2 seems weird (but maybe is warranted given that this is still the system Python on macOS).

Making `virtualenv_install_with_resources` detect the desired Python for unambiguous cases would be great, but I wonder if we have such unambiguous cases. A `depends_on :python3` doesn't seem to be a sufficient indicator as that could still mean that Python 2 is equally fine (as we don't specify `depends_on :python` unless a Python 2 newer than the system Python is required, if I recollect correctly how the current Python handling is supposed to work).

ilovezfs:
pulling this as-is until Homebrew/brew#636 is completed and merged.

PR #3643, opened 2016-08-05
xapian 1.4.0

jaylett:

However:
- [ ] Is this actually ready to merge?

because it contains a hard-coded path based on where my copy of brew puts stuff. This is covered further down.

---
- there isn't a development version of Xapian at the moment (perhaps adding a `--head` would be useful in future)
- 1.4.0 now falls back to checking python if it can't find python2, so we don't need to cater for that in the formula
- the python caveat doesn't seem to show, so I've probably done something silly here
- brew test isn't testing the bindings, which it probably should do; or perhaps the bindings should be in their own formula(e)
- I'm unconvinced `--with-java` does anything useful, because of how Java libraries work; but I'm also unconvinced that brew should bother supporting it anyway. If brew has a standard for where to put Java libraries, I could probably make this usable.
- the python bindings require sphinx-build (to build their documentation), but because the `PATH` during builds is sanitised, I don't know how to use the one from brew `sphinx-doc`. I've included a change which shows how to specify the path to it (you set `SPHINX_BUILD` to the bindings configure run), but my ruby isn't sufficient to make that dynamic, even if I knew how to get the path to the brew one. I also don't know how this is supposed to work for system python, where you'd have to `pip install` sphinx (but it then would work).
- is making `php56` a dependency the right thing to do here? While it's possible that you could install with the system PHP, that way almost certainly lies terror. (Also, because it's `php56` I haven't made it `:optional` because the option should remain `--with-php`. This may not be the neatest approach.)

### DIFF
--- a/Formula/xapian.rb
+++ b/Formula/xapian.rb
@@ -28,8 +28,7 @@ class Xapian < Formula
   deprecated_option "ruby" => "with-ruby"
 
   depends_on :python => :optional
-  depends_on "sphinx-doc" if build.with? "python" => :build
-  depends_on "php56" if build.with? "php"
+  depends_on "sphinx-doc" => :build if build.with? "python"
   depends_on :ruby => "2.1"
 
   skip_clean :la
@@ -49,7 +48,7 @@ class Xapian < Formula
           XAPIAN_CONFIG=#{bin}/xapian-config
           --without-csharp
           --without-tcl
-          SPHINX_BUILD=/usr/local/bin/sphinx-build
+          SPHINX_BUILD=#{Formula["sphinx-doc"].opt_bin}/sphinx-build
         ]
 
         if build.with? "java"
--- a/Formula/xapian.rb
+++ b/Formula/xapian.rb
@@ -3,12 +3,12 @@ class Xapian < Formula
   homepage "https://xapian.org/"
 
   stable do
-    url "http://oligarchy.co.uk/xapian/1.2.23/xapian-core-1.2.23.tar.xz"
-    sha256 "9783aeae4e1a6d06e5636b270db4b458a7d0804a31da158269f57fa5dc86347d"
+    url "http://oligarchy.co.uk/xapian/1.4.0/xapian-core-1.4.0.tar.xz"
+    sha256 "10584f57112aa5e9c0e8a89e251aecbf7c582097638bfee79c1fe39a8b6a6477"
 
     resource "bindings" do
-      url "http://oligarchy.co.uk/xapian/1.2.23/xapian-bindings-1.2.23.tar.xz"
-      sha256 "19b4b56c74863c51733d8c2567272ef7f004b898cf44016711ae25bc524b2215"
+      url "http://oligarchy.co.uk/xapian/1.4.0/xapian-bindings-1.4.0.tar.xz"
+      sha256 "3aec7a009d0bf0b95968420bf68683176c05d63140eaf1cf265d39afe8fa6253"
     end
   end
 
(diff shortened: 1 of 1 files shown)


Formula/xapian.rb
```diff
@@ -39,6 +28,9 @@ class Xapian < Formula
   deprecated_option "ruby" => "with-ruby"
 
   depends_on :python => :optional
+  depends_on "sphinx-doc" if build.with? "python" => :build
```
DomT4: This is ordered slightly wrong. You want:
```ruby
depends_on "sphinx-doc"  => :build if build.with? "python"
```
Is that a hard requirement though? Is there no documentation unless you build with bindings?

Formula/xapian.rb
```diff
@@ -39,6 +28,9 @@ class Xapian < Formula
...
+  depends_on "php56" if build.with? "php"
```
DomT4: This would have to be `homebrew/php/php56` but I think we'd prefer to leave it built against the system.

Formula/xapian.rb
DomT4: More specifically, is there no valid use case where you'd want documentation but not the bindings?

Formula/xapian.rb
```diff
@@ -57,6 +49,7 @@ def install
           XAPIAN_CONFIG=#{bin}/xapian-config
           --without-csharp
           --without-tcl
+          SPHINX_BUILD=/usr/local/bin/sphinx-build
```
DomT4: The dynamic way of finding this as you asked is:
```
SPHINX_BUILD=#{Formula["sphinx-doc"].opt_bin}/sphinx-build
```

Formula/xapian.rb
```diff
@@ -101,6 +92,13 @@ def install
   end
 
   def caveats
+    if build.with? "python"
```
DomT4: It's been a while but I'm _fairly_ sure we have core logic to nag people to add Homebrew's site-packages directory to the system Python if necessary, which isn't the case for Ruby, hence the caveats there.

DomT4:
> there isn't a development version of Xapian at the moment 

That's fine.

> brew test isn't testing the bindings, which it probably should do

We don't test non-default options, which is why.

>  If brew has a standard for where to put Java libraries, I could probably make this usable.

Where are they installed currently when building with the option?

Formula/xapian.rb
jaylett: There's currently no way to build the bindings without documentation (for those bindings; it's totally different to the documentation for Xapian itself, which doesn't get built by this formula), so you'll get a build error if you try.

Formula/xapian.rb
jaylett: This isn't about that; there's a change we're still tracking down in how the latest versions of SWIG & some stuff we do interact which means we need to augment the `PYTHONPATH`. Since this is grotty (and I'm only just becoming aware of the detail of the problem), I'll understand if you'd rather wait until we figure out a proper fix for it and release 1.4.1.

jaylett:
Java libraries aren't really installed; they're made usable outside the build tree, but they aren't copied anywhere (because there isn't a standard place to put them AIUI). (1.2 did this as well.) We could copy the contents of that out somewhere at the end of the formula, even if it's just within the cellar and then show a caveat saying where they are.

jaylett:
Hmm…our build system needs to call `sphinx-build` directly, rather than the homebrew wrapper, so I suspect we're not going to be able to move this forward until we can disable python docs building. (We have to cope with other platforms where the same sphinx-build is shared between python2 and python3.)

Formula/xapian.rb
DomT4: Ding @tdsmith for thoughts.

DomT4:
> Java libraries aren't really installed; they're made usable outside the build tree, but they aren't copied anywhere (because there isn't a standard place to put them AIUI). (1.2 did this as well.) 

I'd probably just remove the option & see if anyone complains. It's quite likely nobody's using the `with-java` option if it's been essentially broken & nobody has complained to us.

> Hmm…our build system needs to call sphinx-build directly, rather than the homebrew wrapper

You'd have to adjust `PYTHONPATH` by the direct executable sans environment helpers is in `Formula["sphinx-doc"].opt_libexec/bin`.

jaylett:
Are the tests because of bottles linked against the previous version? I can update omega (which I didn't know was available in homebrew), but the other two should just work when rebuilt from source I think…

Formula/xapian.rb
tdsmith: Is there a link to a discussion?

jaylett: FYI: we are narrowing in on a good solution to this problem upstream.

tdsmith:
@jaylett Can you summarize what the blocking issues are now?

jaylett:
@tdsmith we've fixed the problem with python upstream, but I'm waiting for a 1.4.1 release rather than using a snapshot. If you'd prefer to close this until then (I don't have a timeline, but now GSoC is over it's high on the priority list), I can open a new one later.

MikeMcQuaid:
That'd be great, thanks @jaylett.

PR #3647, opened 2016-08-05
Wireshark: Remove the Qt4 option.

geraldcombs:
Make --with-qt build with Qt5. Remove support for building with Qt4.
Qt5 has much better support for current versions of macOS than Qt4,
particularly retina displays. We (the Wireshark project) don't build or
test Wireshark+Qt4+macOS and haven't for quite some time. We would very
much appreciate not getting bug reports from Homebrew users who are
having trouble with Qt4.

### DIFF
--- a/Formula/wireshark.rb
+++ b/Formula/wireshark.rb
@@ -13,10 +13,9 @@ class Wireshark < Formula
     sha256 "54e3add2e9631aa400bca0f06a9952dc16d620a2db3a21f1bb450e94c3f8f493" => :mavericks
   end
 
-  option "with-gtk+3", "Build the wireshark command with gtk+3"
-  option "with-gtk+", "Build the wireshark command with gtk+"
-  option "with-qt", "Build the wireshark-qt command (can be used with or without either GTK option)"
-  option "with-qt5", "Build the wireshark-qt command with qt5 (can be used with or without either GTK option)"
+  option "with-qt", "Build the wireshark command with Qt5 (can be used with or without either GTK option)"
+  option "with-gtk+3", "Build the wireshark-gtk command with gtk+3"
+  option "with-gtk+", "Build the wireshark-gtk command with gtk+"
   option "with-headers", "Install Wireshark library headers for plug-in development"
 
   depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


Formula/wireshark.rb
```diff
@@ -13,10 +13,9 @@ class Wireshark < Formula
     sha256 "54e3add2e9631aa400bca0f06a9952dc16d620a2db3a21f1bb450e94c3f8f493" => :mavericks
   end
 
-  option "with-gtk+3", "Build the wireshark command with gtk+3"
-  option "with-gtk+", "Build the wireshark command with gtk+"
-  option "with-qt", "Build the wireshark-qt command (can be used with or without either GTK option)"
-  option "with-qt5", "Build the wireshark-qt command with qt5 (can be used with or without either GTK option)"
+  option "with-qt", "Build the wireshark command with Qt5 (can be used with or without either GTK option)"
```
DomT4: It'll still be `qt5` with us, so you'll want this option name to remain `with-qt5`.

Can you also add a `deprecated_option` which will nudge existing `wireshark --with-qt` users over to `wireshark --with-qt5` more seamlessly next time they upgrade/reinstall than simply yanking the option. The syntax for that is:

```ruby
deprecated_option "with-qt" => "with-qt5"
```

That should be placed one empty line beneath the last option.

Formula/wireshark.rb
```diff
@@ -57,10 +55,10 @@ def install
     args = std_cmake_args
     args << "-DENABLE_GNUTLS=ON" << "-DENABLE_GCRYPT=ON"
 
-    if build.with?("qt") || build.with?("qt5")
+    if build.with?("qt")
```
DomT4: Same notation here more or less, this should remain `if build.with?("qt5")`

Formula/wireshark.rb
```diff
@@ -104,7 +102,7 @@ def install
     ENV.deparallelize # parallel install fails
     system "make", "install"
 
-    if build.with?("qt") || build.with?("qt5")
+    if build.with?("qt")
```
DomT4: And here as well.

DomT4:
Hey, Thanks for creating this.

We were planning to start deprecating Qt4 in general with macOS Sierra because we're unconvinced Qt4 will build on that platform & we have no desire to carry around a new hacky patch upstream will never be interested in for another year, but I'm happy to shunt the timetable up a little bit here if it's causing you folks grief.

I've left a few code comments that need to be addressed, but on the idea, 👍.

DomT4:
Wanted to get this done so pushed the requested changes in https://github.com/Homebrew/homebrew-core/commit/a05941384bcd8e4ef846237aba79d14dc47d757f. Should lessen the burden on both of us as people start migrating across, hopefully. Thanks for raising this.

geraldcombs:
Thanks!

PR #3648, opened 2016-08-06
wxpython: Add comma to install array

dj-m:

---

Modified install array to include comma after the last item

### DIFF
--- a/Formula/wxpython.rb
+++ b/Formula/wxpython.rb
@@ -25,12 +25,12 @@ class Wxpython < Formula
     sha256 "41ec8003758d804b8c426ce654a87f2d9f4be3be40fcbcb3d5686e3ecabaddbc" => :mavericks
   end
 
+  option :universal
+
   if MacOS.version <= :snow_leopard
     depends_on :python
     depends_on FrameworkPythonRequirement
   end
-
-  option :universal
   depends_on "wxmac"
 
   def install
@@ -69,6 +69,7 @@ class Wxpython < Formula
   end
 
   test do
-    system "python -c 'import wx ; print wx.version()'"
+    output = shell_output("python -c 'import wx ; print wx.version()'")
+    assert_match version.to_s, output
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/wxpython.rb
```diff
@@ -29,9 +29,9 @@ class Wxpython < Formula
     depends_on :python
     depends_on FrameworkPythonRequirement
   end
-  depends_on "wxmac"
 
   option :universal
+  depends_on "wxmac"
```
DomT4: Need an empty line between the last option & first `depends_on`. You can move the other block with `depends_on` within as well:

```ruby
option :universal

if MacOS.version <= :snow_leopard
  depends_on :python
  depends_on FrameworkPythonRequirement
end
depends_on "wxmac"
```

Formula/wxpython.rb
```diff
@@ -60,11 +60,15 @@ def install
...
+      "BUILD_STC=1",
     ]
 
     cd "wxPython" do
       system "python", "setup.py", "install", "--prefix=#{prefix}", *args
     end
   end
+
+  test do
+    system "python -c 'import wx ; print wx.version()'"
```
DomT4: Let's assert on the version here as well.
```ruby
output = shell_output("python -c 'import wx ; print wx.version()'")
assert_match version.to_s, output
```

dj-m: Is it acceptable to have it as one line? e.g. 

`assert_match version.to_s, shell_output("python -c 'import wx ; print wx.version()'")`

DomT4: We prefer to wrap to 80 characters as much as possible. Haven't checked but I'd be surprised if that wasn't over. There are various ways to wrap to the 80, but the block I proposed is the most common one in use, at least till we change our minds 😄.

dj-m: Great point! Update's on its way... thanks so much! 👏🏾

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/b9082a9b0daad7e8bda2fde60ab943c3d3186ece. Thank you @dj-m! 😺

MikeMcQuaid:
Nice work @dj-m 😍

PR #3664, opened 2016-08-06
ejabberd 16.08

ilovezfs:
use upstream vendoring

### DIFF
--- a/Formula/ejabberd.rb
+++ b/Formula/ejabberd.rb
@@ -1,119 +1,8 @@
 class Ejabberd < Formula
   desc "XMPP application server"
   homepage "https://www.ejabberd.im"
-
-  stable do
-    url "https://www.process-one.net/downloads/ejabberd/16.04/ejabberd-16.04.tgz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 the upstream vendoring is revision pegged, so I think these resource blocks should be taken out back.

ilovezfs:
@dunn basically this reverts https://github.com/Homebrew/legacy-homebrew/pull/46462. Do you still see a need to use resource blocks here? We've been using non-cached resources pretty extensively now with the go formulae wherever possible, and it's been fine.

DomT4:
I'm happy if @dunn is happy.

dunn:
Yeah, if upstream isn't just pulling the master branch anymore, 👍

ilovezfs:
@dunn I actually had at first updated all of the resource blocks with the latest release tags in each repo, and then the build failed because upstream had specifically pegged a lower tag for some of them (so I could have gone back in and lowered the tags accordingly, of course).

PR #3670, opened 2016-08-06
sshuttle 0.78.1

ilovezfs:
use new virtualenv feature
use multi-argument form of `system` in the test

### DIFF
--- a/Formula/sshuttle.rb
+++ b/Formula/sshuttle.rb
@@ -1,11 +1,11 @@
 class Sshuttle < Formula
+  include Language::Python::Virtualenv
+
   desc "Proxy server that works as a poor man's VPN"
   homepage "https://github.com/sshuttle/sshuttle"
-  url "https://pypi.python.org/packages/source/s/sshuttle/sshuttle-0.78.0.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 don't we bottle all of these now?

DomT4:
Apparently Python stuff doesn't use system development tools that require Xcode/CLT installed, so they are fine to not bottle, but my preference is _heavily_ towards bottling. I prefer to not expose users to any sort of build process, however tiny, where possible.

ilovezfs:
@DomT4 yeah I'm very skeptical of bottle unneededs generally. They seem to also expose users to 404 issues unnecessarily even when they're "appropriate."

ilovezfs:
@DomT4 :ship:?

DomT4:
Bit weird the syntax is still `virtualenv_install_with_resources` even when there's no resources, but 👍.

ilovezfs:
@DomT4 I think it should just be `virtualenv_install` and if you don't know that means resources too, if there are any, then it's a learning opportunity. CC @tdsmith

ilovezfs:
@DomT4 alternatively we could introduce new pedantry and have two versions, `virtualenv_install_with_resources` and `virtualenv_install` and raise if the first is used when there are no resources, but that seems tedious without an obvious benefit other than precision for the sake of precision.

PR #3685, opened 2016-08-07
dxpy 0.191.0

ilovezfs:
conflicts with android-sdk due to `dx`

### DIFF
--- a/Formula/dxpy.rb
+++ b/Formula/dxpy.rb
@@ -1,8 +1,8 @@
 class Dxpy < Formula
   desc "DNAnexus toolkit utilities and platform API bindings for Python"
   homepage "https://github.com/dnanexus/dx-toolkit"
-  url "https://pypi.python.org/packages/source/d/dxpy/dxpy-0.182.1.tar.gz"
-  sha256 "0022f5f89e64b994b8c4cbcf596b0e514d9958569e7fc000ae36d5a49cdf1e9a"
+  url "https://files.pythonhosted.org/packages/04/16/1d1e1ff25d4f694f04f88417afbd72703d850aebe9f84fdbb8ed836089d7/dxpy-0.191.0.tar.gz"
+  sha256 "b98a55fc766a9e7044f6f299d148c5dde3967f22669add915a1a1921e4006820"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


ilovezfs:
@tdsmith @DomT4 I've deliberately not converted this one to use the new virtualenv stuff because it doesn't work properly with the gnureadline module. Using the old way, everything is fine. If instead I install with virtualenv, the test puts out the following warning:
```
Warning: incompatible readline module detected (libedit), tab completion disabled
```
This is because when it's installed with the virtualenv way, `import readline` is still picking up the system readline. A hard dependency on "python" works around it (so presumably not an issue for the bottles), but for a local source build the hard "python" dependency is not necessary using the old way, and it picks up the correct readline since the PYTHONPATH has been set to make that happen.

@tdsmith Is there some way way to get it to actually use the intended version of readline using the new virtualenv feature without having to also patch the software? This seems like a problem that's going to end up being more widespread than just this formula.

ilovezfs:
@MikeMcQuaid based on what I'm seeing here, I think we may need to halt conversion to virtualenv temporarily because it's unclear how much breakage may be occurring under the radar for local source builds that isn't being detected by us given that the problem is masked when the :python requirement is fulfilled with Homebrew python during bottling.

tdsmith:
Did you try adding `depends_on "readline"`? Readline stuff is special-cased everywhere. I think it's not likely that there's a more generic problem but I'm not sure what you're suggesting it would be or where you think it's interacting with :python, which isn't being applied on the bots for this formula.

Also either the `desc` or the installation method is wrong here; the bindings aren't being exposed.

ilovezfs:
@tdsmith yes I tried adding readline and it made no difference. In this case, the bottles will be just as broken as the source builds since it doesn't use `:python`, so the problem won't be masked by that. But in cases where `:python` is used, problems like this will be masked during bottling since building against Homebrew python prevented the issue.

Why is the virtualenv not using the gnureadline installed in the virtualenv, whereas the old way setting PYTHONPATH works perfectly? Is there some workaround that will make the virtualenv behave correctly or do we have to break out PYTHONPATH with an env script here in combination with virtualenv?

Bottom line here is converting this one breaks it.

tdsmith:
It should be rare that we use :python when installing apps so I'm not worried that we're missing things.

Why is gnureadline being resourced? I don't see it in dxpy's dependency tree.

tdsmith:
> Why is gnureadline being resourced? I don't see it in dxpy's dependency tree.

Ah, but dxpy complains without it, got it.

The proximal cause of failure is that the `gnureadline` package tries to override Python's readline module but the readline.so in lib/python2.7/lib-dynload appears in sys.path before the readline.py in the virtualenv site-packages. I'm not really sure how this ever works. In any case it's specific to readline.

ilovezfs:
>It should be rare that we use :python when installing apps so I'm not worried that we're missing things.

I think the safe thing to do would be to ditch :python and :python3 in favor of "python" and "python3" in anything that uses virtualenv until there's some mechanism for testing during CI whether not setting PYTHONPATH is breaking that formula when the requirement is satisfied by the system version.

>Why is gnureadline being resourced? I don't see it in dxpy's dependency tree.

It gets added just-in-time for OS X from a separate text file.

https://github.com/dnanexus/dx-toolkit/blob/master/src/python/requirements_readline.txt

https://github.com/dnanexus/dx-toolkit/blob/364f3f02522d6cf41a632ed82752fa8e8c5907df/src/python/setup.py#L55-L61

ilovezfs:
>appears in sys.path before the readline.py in the virtualenv site-packages

Right, and what else do we have that's working solely because of the environment script wrappers that set PYTHONPATH?

tdsmith:
> I think the safe thing to do would be to ditch :python and :python3 in favor of "python" and "python3" in anything that uses virtualenv until there's some mechanism for testing during CI whether not setting PYTHONPATH is breaking that formula when the requirement is satisfied by the system version.

There's no system version of python3 and we shouldn't use :python for apps so I don't think this is actionable.

> Right, and what else do we have that's working solely because of the environment script wrappers that set PYTHONPATH?

I believe that this only affects modules that try to override stdlib packages, which is bad and unusual behavior.

A workaround is to write a .pth file containing `import sys; sys.path.insert(0, #{libexec}/lib/python2.7/site-packages/readline.py")` into site-packages.

tdsmith:
FWIW gnureadline's readme says it doesn't support masking the stdlib module when it's installed with pip, because pip correctly balks:

> The module is called gnureadline so as not to clash with the readline module in the standard library. This keeps polite installers such as pip happy and is sufficient for shells such as IPython. In order to use this module in the standard Python shell it has to be installed with the more impolite easy_install from setuptools. 

This is plausibly a dx bug; they should import gnureadline instead of readline if they're gonna care.

ilovezfs:
>There's no system version of python3

Right in :python3 case it's user's alternative :python3 which may have worked fine previously due to the wrappers, but potentially won't anymore without them

>This is plausibly a dx bug; they should import gnureadline instead of readline if they're gonna care.

Yes, that's what I meant by wanting a fix that doesn't require new patching we didn't previosuly have to do. Indeed string replacing gnureadline did fix it in my testing before.

>A workaround is to write a .pth file containing import sys; sys.path.insert(0, #{libexec}/lib/python2.7/site-packages/readline.py") into site-packages.

Maybe the backend virtualenv stuff should do that in all cases where there's a :python or :python3 requirement instead of "python" or "python3" to make sure we're not accidentally and gratuitously breaking things that currently are not broken?

tdsmith:
> Maybe the backend virtualenv stuff should do that in all cases where there's a :python or :python3 requirement instead of "python" or "python3" to make sure we're not accidentally and gratuitously breaking things that currently are not broken?

It's safe to do this in all cases to facilitate packages which use gnureadline lazily.

ilovezfs:
>It's safe to do this in all cases to facilitate packages which use gnureadline lazily.

Is there a way to do it for everything so that the site packages always get picked up first as they have heretofore?

tdsmith:
It's not the case that the sys.path sorting order is different; it's the case that gnureadline only has magical powers when it's installed with obsolete tooling. I really don't think there's a generic problem here.

ilovezfs:
>I really don't think there's a generic problem here.

And yet there's nothing *testing* no longer setting PYTHONPATH isn't breaking things that previously worked. I'm not encouraged by the fact that the only reason this formula's breakage came to anyone's attention was because the software was kind enough to print a warning that I proceeded to chase even though it didn't cause the test-do block to fail. But I'll defer to your judgment, of course.

tdsmith:
https://github.com/dnanexus/dx-toolkit/issues/169

ilovezfs:
Thanks Tim!

MikeMcQuaid:
> There's no system version of python3 and we shouldn't use :python for apps so I don't think this is actionable.

It'll always be used for apps on e.g. 10.6. Would be good to have an `audit` for this case.

ilovezfs:
Hmm that fix doesn't actually seem to work: https://github.com/Homebrew/brew/pull/657#issuecomment-238206705

ilovezfs:
I am pulling this as-is. We can convert it when there's a viable fix that works.

PR #3688, opened 2016-08-07
owfs 3.1p2

ilovezfs:
- pluck missing header files from upstream git repository
- D_POSIX_C_SOURCE inreplace isn't needed with current configure
  options, though the underlying bug is still not fixed
- switch dependency to libusb since that's what it currently links to
- avoid opportunistically enabling ftdi


ilovezfs:
OK looks like the complication can be killed because upstream responded and rolled a new release for us: https://sourceforge.net/p/owfs/bugs/70/#2427.

PR #3691, opened 2016-08-07
WIP: Initial dotnet formula

haf:

---

Ref https://github.com/dotnet/cli/issues/533

Unfortunately MSFT is a bit new on the OSS scene so they [release 1.0 RTM](https://blogs.msdn.microsoft.com/dotnet/2016/06/27/announcing-net-core-1-0/) tagged as `preview2` on github, which is why this formula is **not in fact** a pre-release formula.

The original pkg file installs like this:

screen shot 2016-08-08 at 00 13 54

In `/usr/local/share/dotnet/` that is. TODO: not sure how to copy the `stage2` folder to this location, or the Cellar location of the formula.

### DIFF
--- a/Formula/dotnet.rb
+++ b/Formula/dotnet.rb
@@ -1,0 +1,75 @@
+class Dotnet < Formula
+  desc ".Net Core"
+  homepage "https://github.com/dotnet/cli/releases"
+  url "https://github.com/dotnet/cli/archive/v1.0.0-preview2.tar.gz"
+  sha256 "a0708ed3eedb10e95dcea96c62f85be1c38b91f2be9afa89d8fc655b7e0c4706"
+
+  patch :DATA
+  head "https://github.com/dotnet/cli.git"
+
+  depends_on "cmake" => :build
+  depends_on "openssl"
+
+  def install
+    system "./build.sh", "--targets", "Prepare,Compile"
+
+    # Relink openssl to the homebrew version
+    # Would be smoother to tell clang to use this openssl prefix, but alas...
+    openssl_path = Formula["openssl"].opt_prefix
+
+    stage2 = buildpath/"artifacts/osx.#{MacOS.version}-x64/stage2"
+    system "install_name_tool -add_rpath #{openssl_path}/lib #{stage2}/shared/Microsoft.NETCore.App/1.0.0/System.Security.Cryptography.Native.dylib"
+
+    prefix.install Dir["#{stage2}/*"]
+  end
+
+  test do
+    system prefix/"dotnet", "version"
+  end
+end
+__END__
+---
+ build_projects/dotnet-cli-build/PrepareTargets.cs           | 9 ---------
+ build_projects/shared-build-targets-utils/Utils/GitUtils.cs | 4 ++--
+ 2 files changed, 2 insertions(+), 11 deletions(-)
+
+diff --git a/build_projects/dotnet-cli-build/PrepareTargets.cs b/build_projects/dotnet-cli-build/PrepareTargets.cs
+index 5e6b948..3af853b 100644
+--- a/build_projects/dotnet-cli-build/PrepareTargets.cs
++++ b/build_projects/dotnet-cli-build/PrepareTargets.cs
+@@ -464,15 +464,6 @@ public static BuildTargetResult CheckPrereqCmakePresent(BuildTargetContext c)
+         [Target]
+         public static BuildTargetResult SetTelemetryProfile(BuildTargetContext c)
+         {
+-            var gitResult = Cmd("git", "rev-parse", "HEAD")
+-                .CaptureStdOut()
+-                .Execute();
+-            gitResult.EnsureSuccessful();
+-
+-            var commitHash = gitResult.StdOut.Trim();
+-
+-            Environment.SetEnvironmentVariable("DOTNET_CLI_TELEMETRY_PROFILE", $"https://github.com/dotnet/cli;{commitHash}");
+-
+             return c.Success();
+         }
+
+diff --git a/build_projects/shared-build-targets-utils/Utils/GitUtils.cs b/build_projects/shared-build-targets-utils/Utils/GitUtils.cs
+index bae5c67..522958f 100644
+--- a/build_projects/shared-build-targets-utils/Utils/GitUtils.cs
++++ b/build_projects/shared-build-targets-utils/Utils/GitUtils.cs
+@@ -6,12 +6,12 @@ public static class GitUtils
+     {
+         public static int GetCommitCount()
+         {
+-            return int.Parse(ExecuteGitCommand("rev-list", "--count", "HEAD"));
++            return 3121;
+         }
+
+         public static string GetCommitHash()
+         {
+-            return ExecuteGitCommand("rev-parse", "HEAD");
++            return "1e9d529bc54ed49f33102199e109526ea9c6b3c4";
+         }
+
+         private static string ExecuteGitCommand(params string[] args)
+--


haf:
Currently failing to build unless the latest openssl binaries are used with `brew link --force openssl` with:

```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'Crypto' threw an exception. ---> System.TypeInitializationException: The type initializer for 'CryptoInitializer' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'System.Security.Cryptography.Native': The specified module could not be found.
 (Exception from HRESULT: 0x8007007E)
   at Interop.CryptoInitializer.EnsureOpenSslInitialized()
   at Interop.CryptoInitializer..cctor()
   --- End of inner exception stack trace ---
   at Interop.Crypto..cctor()
   --- End of inner exception stack trace ---
   at Interop.Crypto.GetRandomBytes(Byte* buf, Int32 num)
   at System.IO.Path.GetCryptoRandomBytes(Byte* bytes, Int32 byteCount)
   at System.IO.Path.GetRandomFileName()
   at Microsoft.DotNet.InternalAbstractions.TemporaryDirectory..ctor()
   at Microsoft.Extensions.EnvironmentAbstractions.DirectoryWrapper.CreateTemporaryDirectory()
   at Microsoft.DotNet.Configurer.NuGetPackagesArchiver..ctor()
   at Microsoft.DotNet.Cli.Program.ConfigureDotNetForFirstTimeUse(INuGetCacheSentinel nugetCacheSentinel)
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, ITelemetry telemetryClient)
   at Microsoft.DotNet.Cli.Program.Main(String[] args)
/private/tmp/dotnet-20160807-11838-1cf1ilq/cli-1.0.0-preview2/build_projects/dotnet-cli-build/build.sh: line 117: 11982 Abort trap: 6           dotnet restore
```

Then failing with:

```
                        Error: Empty installation
```

Which I suppose is because it places stuff in a special folder.

haf:
Ref https://github.com/dotnet/cli/issues/3964#issuecomment-236832511

Added two TODOs I need help with.

Formula/dotnet.rb
```diff
@@ -0,0 +1,92 @@
...
+  def install
+    system "./build.sh", "--targets", "Prepare,Compile"
+
+    stage2 = buildpath/"artifacts/osx.10.11-x64/stage2"
+    # TODO: how do I tell homebrew that the folders are under stage2?
+
+    # Relink openssl to the homebrew version
+    # Would be smoother to tell clang to use this openssl prefix, but alas...
+    openssl_path = Formula["openssl"].opt_prefix
+    system 'sudo', %W|install_name_tool
```
tdsmith: this doesn't need to run as root

Formula/dotnet.rb
```diff
@@ -0,0 +1,92 @@
...
+    # `test do` will create, run in and delete a temporary directory.
+    #
+    # This test will fail and we won't accept that! It's enough to just replace
+    # "false" with the main program this formula installs, but it'd be nice if you
+    # were more thorough. Run the test with `brew test cli`. Options passed
+    # to `brew install` such as `--HEAD` also need to be provided to `brew test`.
+    #
+    # The installed folder is not in the path, so use the entire path to any
+    # executables being tested: `system "#{bin}/program", "do", "something"`.
+    system "artifacts/darwin-x64/stage2/dotnet", "version"
```
tdsmith: The test runs against the installed binaries and not from `buildroot`; you probably want `system bin/"dotnet", ...`.

Formula/dotnet.rb
```diff
@@ -0,0 +1,92 @@
...
+  patch :DATA
+  head "https://github.com/dotnet/cli.git"
+
+  depends_on "cmake" => :build
+  depends_on "openssl"
+
+  def install
+    system "./build.sh", "--targets", "Prepare,Compile"
+
+    stage2 = buildpath/"artifacts/osx.10.11-x64/stage2"
```
tdsmith: I assume the `10.11` changes with OS X version?

haf: Yes, was having a hard time finding the listing of what I can compose the formula using.

Formula/dotnet.rb
```diff
@@ -0,0 +1,93 @@
+class Dotnet < Formula
+  desc ".Net Core"
+  homepage "https://github.com/dotnet/cli/releases"
+  url "https://github.com/dotnet/cli/archive/v1.0.0-preview2.tar.gz"
+  version "1"
+  sha256 "a0708ed3eedb10e95dcea96c62f85be1c38b91f2be9afa89d8fc655b7e0c4706"
+
+  patch :DATA
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/dotnet.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    # Relink openssl to the homebrew version
+    # Would be smoother to tell clang to use this openssl prefix, but alas...
+    openssl_path = Formula["openssl"].opt_prefix
+    system "install_name_tool",
+      %W|-add_rpath
+         #{openssl_path}/lib
+         #{prefix}/shared/Microsoft.NETCore.App/1.0.0/System.Security.Cryptography.Native.dylib|
+
+    # Have a binary
```
MikeMcQuaid: Can remove this comment.

Formula/dotnet.rb
```diff
@@ -0,0 +1,93 @@
+class Dotnet < Formula
+  desc ".Net Core"
+  homepage "https://github.com/dotnet/cli/releases"
+  url "https://github.com/dotnet/cli/archive/v1.0.0-preview2.tar.gz"
+  version "1"
```
MikeMcQuaid: Can remove this line. If it can't detect the version go for `1.0.0-preview2`

Formula/dotnet.rb
haf: It's because they require git to build... How do you normally deal with that?

MikeMcQuaid: Just use a git checkout.

Formula/dotnet.rb
dunn: `buildpath/"artifacts/osx.#{MacOS.version}-x64/stage2"` should work.

haf:
@dunn Excellent, thanks for that hint. So then it's only this TODO left before this may even work in some manner: `# TODO: how do I tell homebrew that the folders are under stage2?`

dunn:
> how do I tell homebrew that the folders are under stage2?

Sorry, I'm not sure what you mean?  Which folders, and what needs to be done to them/

haf:
@dunn Normally it would output stuff to `prefix`, but not this project. Instead it outputs to `buildpath/"artifacts/osx.#{MacOS.version}-x64/stage2"`; so I want that stuff into the folder (prefix plus something?) where the 'final product' ends up.

dunn:
Compile steps should always leave the build product in the temporary directory `buildpath`.  For projects with `configure` scripts we pass `--prefix=#{prefix}` so that `make install` knows where to move things to; our `std_cmake_args` does something similar for CMake projects.

TL;DR nothing is ever automatically installed into `prefix`, so you're doing the right thing by doing it manually.

haf:
@dunn is this what you're thinking of?

Formula/dotnet.rb
```diff
@@ -0,0 +1,88 @@
...
+    stage2 = buildpath/"artifacts/osx.#{MacOS.version}-x64/stage2"
+    # TODO: how do I tell homebrew that the folders are under stage2?
+
+    # Relink openssl to the homebrew version
+    # Would be smoother to tell clang to use this openssl prefix, but alas...
+    openssl_path = Formula["openssl"].opt_prefix
+    sh "install_name_tool -add_rpath #{openssl_path}/lib #{stage2}/shared/Microsoft.NETCore.App/1.0.0/System.Security.Cryptography.Native.dylib"
+
+    bin.install stage2/"dotnet"
+    prefix.install Dir["*"]
```
dunn: This might need to be `prefix.install Dir["#{buildpath}/artifacts/osx.#{MacOS.version}-x64/stage2/*"]`, but if everything other than the `dotnet` binary belongs in the root prefix (as opposed to `lib`, `libexec`, etc.) then yeah, this is the right approach.

Formula/dotnet.rb
```diff
@@ -0,0 +1,88 @@
+class Dotnet < Formula
+  desc ".Net Core"
+  homepage "https://github.com/dotnet/cli/releases"
+  url "https://github.com/dotnet/cli/archive/v1.0.0-preview2.tar.gz"
```
MikeMcQuaid: Still better to use a GitHub checkout here and it'll remove the need for the patch.

haf:
The current state, with moving `dotnet` to `bin` causes this exception:

```
dotnet
Failed to load , error: dlopen(, 1): no suitable image found.  Did find:
	/usr/local/lib/: not a file
	/usr/lib/: not a file
	/usr/local/lib/: not a file
	/usr/lib/: not a file
	/usr/local/lib/: not a file
	/usr/lib/: not a file
```

Because it's looking to a sibling folder, but not finding it.

Instead

```ruby
prefix.install Dir["#{stage2}/*"]
```

And nothing else makes it work as intended.

I'm going to fix the git checkout like @MikeMcQuaid wanted, but this PR is starting to get ready besides this particular change request.

haf:
In summary: how do I make the dylibs/loading work when the `dotnet` binary (perhaps) need to be in the root?

dunn:
I'll look at this more closely when I can, but if installing directly into `prefix` is required for it to work, that's OK.

MikeMcQuaid:
> I'll look at this more closely when I can, but if installing the libraries directly into prefix is required for it to work, that's OK.

Agreed as long as `audit` is passing.



> In summary: how do I make the dylibs/loading work when the dotnet binary (perhaps) need to be in the root?

You probably need to ask upstream that question as it's not really a Homebrew one.

haf:
@MikeMcQuaid I'm really just asking for a link to where I can understand how the loading works on OS X with binary vs dependent library; not particularly asking for netcore.

Formula/dotnet.rb
tdsmith: ping for this change

haf:
Please tell me what you want changed by pointing me in the direction of the documentation that documents the APIs you want to use. You say git reference; where are those docs?

MikeMcQuaid:
Passing on this until we have a stable version that builds without patching and passes `brew audit`.

haf:
Wow. What a dude you are Mike.

MikeMcQuaid:
@haf We're doing our best to run this project effectively.

haf:
@MikeMcQuaid I ask for help, you close the issue. And the lint has been passing many times over and its version is RTM, so you should not close it. You should focus on running the project productively instead, and that means taking into account lead times on change requests you make of me. If you think I've dropped this PR, then you're mistaken.

MikeMcQuaid:
A list of what needs addressed before we can include this:

- The version may be RTM but it's mislabeled. Upstream should fix this before we include it.
- "I'm really just asking for a link to where I can understand how the loading works on OS X with binary vs dependent library; not particularly asking for netcore." this isn't something we can really help you with as I don't understand the question and doubt if I did I'd be able to provide you with the information you need to fix this.
- We usually don't accept new formulae that need patching to build
- This is how you use a `git` checkout: http://www.rubydoc.info/github/Homebrew/brew/master/Formula.url
- Our CI was still failing
- This should pass all the things you checked in the original checkbox

tdsmith:
wrt dynamic linkage discovery, you're probably looking the bottom of `man dyld`, the top of `man ld`, and `man install_name_tool`. The punchline is that OS X's linker remembers the install_name of the dylibs to which it is linked and expects those to be absolute paths, possibly including special prefixes like @rpath.

haf:
1. Upstream has segregated the runtime called CoreCLR from the cli tools, and the cli tools compose your app in-place with the runtime. However, it does sound strange, I agree that Microsoft has labelled preview2 of the cli tools as the RTM release of .Net Core, but there it is.
 2. @tdsmith thanks, that was the info I was looking for before and the link to the example means I can update this PR to match it (if you'll please reopen it)
 3. Well, we discussed that and agreed we could use git checkouts higher up in the thread, as the only reason for patching was that it builds the git checkout number into the version number.
 4. -
 5. CI has been green, but since you prefer rebased PRs it's been turned red again. However, now it's gone so I can't see what it failed with any longer.
 6. And it did, and this PR is a living thing that changes as you come with feedback.

I'm looking forward to continuing this PR when it's open again.

ilovezfs:
@haf did you find a workaround for the openssl-formula-must-be-linked problem?

haf:
@ilovezfs Yes, that's now part of the PR and works really nicely!

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
I reopened this 10 days ago and there's been no movement since then and CI is failing in a way that's trivially reproducible by running `brew install dotnet` or `brew audit dotnet --new-formula` on any of the 3 most recent OS X versions. Please resubmit this PR when both of those are passing, you are using a tag/release that doesn't have `preview` in the name and it does not require any patching to function and, ideally, no `install_name_tool` calls. This will very likely require working with upstream and may require another stable release from them.

To preempt some responses:
- you are not being held to a higher standard than any other pull-request. If anything we've provided considerably more help than we normally would
- it's disappointing to be personally criticised for closing this pull-request, reopen it and see no progress on e.g. using a git checkout which was a problem before this PR was closing and remains unaddressed. 
- these PR and formula quality standards come from running this package manager for a fairly long time. They are not personal and not a criticism of anyone but they set a bar that we need to keep maintenance time at sustainable levels and quality at an acceptable level for our users.

PR #3694, opened 2016-08-08
afl-fuzz 2.30b

alex:

---

### DIFF
--- a/Formula/afl-fuzz.rb
+++ b/Formula/afl-fuzz.rb
@@ -1,8 +1,8 @@
 class AflFuzz < Formula
   desc "American fuzzy lop: Security-oriented fuzzer"
   homepage "http://lcamtuf.coredump.cx/afl/"
-  url "http://lcamtuf.coredump.cx/afl/releases/afl-2.29b.tgz"
-  sha256 "adb302d3ad740233760f0612856ea15509512e7c32edb7cb4aad2e73f69a2773"
+  url "http://lcamtuf.coredump.cx/afl/releases/afl-2.30b.tgz"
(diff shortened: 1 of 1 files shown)


Formula/afl-fuzz.rb
```diff
@@ -15,7 +15,7 @@ def install
     # behaving correctly!" in a nested login shell.
     # Reported to lcamtuf@coredump.cx 6th Apr 2016.
```
MikeMcQuaid: Could you try to submit a patch to them? It doesn't look like they are working on fixing this.

alex: Sent an email to the mailing list asking about it, I'm not up for drafting a patch myself, as I don't understand the root cause. (Unless you just mean a patch do the same thing as this find/replace)

MikeMcQuaid: Let's just remove the patch for now, then, thanks!

alex: If this `inreplace` is removed then `afl-fuzz` does not build, as far as I understand.

MikeMcQuaid: Can you try it and see if it builds for you? If it does, we'll see if it builds on CI and if it does: we can remove the patch 😀

alex: Confirmed that it built locally, trying on CI now.

alex:
@MikeMcQuaid Ok, looks to be clean after all! I guess it was fixed and no one noticed.

MikeMcQuaid:
Thanks again @alex! This is why I hate patching 😀

PR #3703, opened 2016-08-08
go: set 1.7rc5 as stable version for sierra

### LINKED ISSUE #3665: Go 1.7rc5 and 10.12
Many thanks for the Go bootstrap rewrite that makes it easy to install 1.7rc5 with `brew install go --devel`. But all formulas that depend on Go (docker, syncthing...) still require stable and fail to install like this:

```
→ brew install syncthing
Warning: You are using OS X 10.12.
We do not provide support for this pre-release version.
You may encounter build failures or other breakages.
Please create pull-requests instead of filing issues.
go: This formula either does not compile or function as expected on OS X
versions newer than El Capitan due to an upstream incompatibility.
```

Even when Go 1.7rc5 is successfully installed. 

I can understand that this is minor edge case that will soon fix itself when stable versions are released, so this should probably not get fixed. ~~I am perfectly ok to live with my quick, stupid hack to make things work on 10.12: Switch "stable" and "devel" labels in the `go.rb` formula.~~ (better solution further down)

To illustrate how stupid this hack is, why nobody in their right mind should ever do this or run beta software at all: "godep" complained on one 10.12 machine about not finding dependencies, but worked nicely on another, under exact the same circumstances/versions/etc. Reinstalling godep fixed this problem.

ilovezfs:
Also,
- go back to stable bootstrap binary for devel since devel spec no
  longer needs to work on sierra

Closes #3665.

### DIFF
--- a/Formula/go.rb
+++ b/Formula/go.rb
@@ -3,27 +3,41 @@ class Go < Formula
   homepage "https://golang.org"
 
   stable do
-    url "https://storage.googleapis.com/golang/go1.6.3.src.tar.gz"
-    mirror "https://fossies.org/linux/misc/go1.6.3.src.tar.gz"
-    version "1.6.3"
-    sha256 "6326aeed5f86cf18f16d6dc831405614f855e2d416a91fd3fdc334f772345b00"
+    if MacOS.version < :sierra
+      url "https://storage.googleapis.com/golang/go1.6.3.src.tar.gz"
+      mirror "https://fossies.org/linux/misc/go1.6.3.src.tar.gz"
+      version "1.6.3"
+      sha256 "6326aeed5f86cf18f16d6dc831405614f855e2d416a91fd3fdc334f772345b00"
 
-    # 1.6.3 does not build on macOS Sierra. Users should use devel instead
-    # until 1.7 is stable (due soon).
-    depends_on MaximumMacOSRequirement => :el_capitan
+      # 1.6.3 does not build on macOS Sierra.
+      depends_on MaximumMacOSRequirement => :el_capitan
 
-    # Should use the last stable binary release to bootstrap.
-    resource "gobootstrap" do
-      url "https://storage.googleapis.com/golang/go1.6.2.darwin-amd64.tar.gz"
-      version "1.6.2"
-      sha256 "6ebbafcac53bbbf8c4105fa84b63cca3d6ce04370f5a04ac2ac065782397fc26"
-    end
+      go_version = "1.6"
+      resource "gotools" do
+        url "https://go.googlesource.com/tools.git",
+            :branch => "release-branch.go#{go_version}",
+            :revision => "c887be1b2ebd11663d4bf2fbca508c449172339e"
+      end
+    else
+      url "https://storage.googleapis.com/golang/go1.7rc5.src.tar.gz"
+      version "1.7rc5"
+      sha256 "206c90e797e66335fe134052568f63a493f27b86f765087add390d5fb4c596c4"
 
-    go_version = "1.6"
-    resource "gotools" do
-      url "https://go.googlesource.com/tools.git",
-          :branch => "release-branch.go#{go_version}",
-          :revision => "c887be1b2ebd11663d4bf2fbca508c449172339e"
+      # Should use the last stable binary release to bootstrap.
+      # Not the case here because 1.6.3 is lacking a fix for an issue which breaks
+      # compile on macOS Sierra; in future this should share bootstrap with stable.
+      resource "gobootstrap" do
+        url "https://storage.googleapis.com/golang/go1.7rc5.darwin-amd64.tar.gz"
+        version "1.7rc5"
+        sha256 "d9991c5e60464f75334368fa6831484f5c577de9dadfb6e799aab43e95ef5894"
+      end
+
+      go_version = "1.7"
+      resource "gotools" do
+        url "https://go.googlesource.com/tools.git",
+             :branch => "release-branch.go#{go_version}",
+             :revision => "a84e830bb0d2811304f6e66498eb3123ca97b68e"
+      end
     end
   end
 
@@ -34,38 +48,23 @@ class Go < Formula
     sha256 "5f570b6c7aa2d7caa6c715af6dce6fa30d7fbd5acc46fac8fbc3232270956f9e" => :mavericks
   end
 
-  devel do
-    url "https://storage.googleapis.com/golang/go1.7rc5.src.tar.gz"
-    version "1.7rc5"
-    sha256 "206c90e797e66335fe134052568f63a493f27b86f765087add390d5fb4c596c4"
-
-    # Should use the last stable binary release to bootstrap.
-    # Not the case here because 1.6.3 is lacking a fix for an issue which breaks
-    # compile on macOS Sierra; in future this should share bootstrap with stable.
-    resource "gobootstrap" do
-      url "https://storage.googleapis.com/golang/go1.7rc5.darwin-amd64.tar.gz"
+  if MacOS.version < :sierra
+    devel do
+      url "https://storage.googleapis.com/golang/go1.7rc5.src.tar.gz"
       version "1.7rc5"
-      sha256 "d9991c5e60464f75334368fa6831484f5c577de9dadfb6e799aab43e95ef5894"
-    end
+      sha256 "206c90e797e66335fe134052568f63a493f27b86f765087add390d5fb4c596c4"
 
-    go_version = "1.7"
-    resource "gotools" do
-      url "https://go.googlesource.com/tools.git",
-          :branch => "release-branch.go#{go_version}",
-          :revision => "a84e830bb0d2811304f6e66498eb3123ca97b68e"
+      go_version = "1.7"
+      resource "gotools" do
+        url "https://go.googlesource.com/tools.git",
+            :branch => "release-branch.go#{go_version}",
+            :revision => "a84e830bb0d2811304f6e66498eb3123ca97b68e"
+      end
     end
   end
 
   head do
     url "https://github.com/golang/go.git"
-
-    # Should use the last stable binary release to bootstrap.
-    # See devel for notes as to why not the case here, for now.
-    resource "gobootstrap" do
-      url "https://storage.googleapis.com/golang/go1.7rc5.darwin-amd64.tar.gz"
-      version "1.7rc5"
-      sha256 "d9991c5e60464f75334368fa6831484f5c577de9dadfb6e799aab43e95ef5894"
-    end
 
     resource "gotools" do
       url "https://go.googlesource.com/tools.git"
@@ -78,6 +77,19 @@ class Go < Formula
   option "without-race", "Build without race detector"
 
   depends_on :macos => :mountain_lion
+
+  resource "gobootstrap" do
+    if MacOS.version < :sierra
+      # Should use the last stable binary release to bootstrap.
+      url "https://storage.googleapis.com/golang/go1.6.2.darwin-amd64.tar.gz"
+      version "1.6.2"
+      sha256 "6ebbafcac53bbbf8c4105fa84b63cca3d6ce04370f5a04ac2ac065782397fc26"
+    else
+      url "https://storage.googleapis.com/golang/go1.7rc5.darwin-amd64.tar.gz"
+      version "1.7rc5"
+      sha256 "d9991c5e60464f75334368fa6831484f5c577de9dadfb6e799aab43e95ef5894"
+    end
+  end
 
   def install
     (buildpath/"gobootstrap").install resource("gobootstrap")


ilovezfs:
CC @MikeMcQuaid

Formula/go.rb
```diff
@@ -2,28 +2,44 @@ class Go < Formula
...
-    # 1.6.3 does not build on macOS Sierra. Users should use devel instead
-    # until 1.7 is stable (due soon).
-    depends_on MaximumMacOSRequirement => :el_capitan
-
-    # Should use the last stable binary release to bootstrap.
-    resource "gobootstrap" do
-      url "https://storage.googleapis.com/golang/go1.6.2.darwin-amd64.tar.gz"
-      version "1.6.2"
-      sha256 "6ebbafcac53bbbf8c4105fa84b63cca3d6ce04370f5a04ac2ac065782397fc26"
+  if MacOS.version < :sierra
```
MikeMcQuaid: Can you put this inside the `stable do` instead; does that work?

Formula/go.rb
```diff
@@ -34,38 +50,38 @@ class Go < Formula
...
-    url "https://storage.googleapis.com/golang/go1.7rc5.src.tar.gz"
-    version "1.7rc5"
-    sha256 "206c90e797e66335fe134052568f63a493f27b86f765087add390d5fb4c596c4"
-
-    # Should use the last stable binary release to bootstrap.
-    # Not the case here because 1.6.3 is lacking a fix for an issue which breaks
-    # compile on macOS Sierra; in future this should share bootstrap with stable.
-    resource "gobootstrap" do
-      url "https://storage.googleapis.com/golang/go1.7rc5.darwin-amd64.tar.gz"
+  if MacOS.version < :sierra
```
MikeMcQuaid: Similarly here.

MikeMcQuaid: Although I guess in this case it shouldn't exist without so actually: this is fine.

Formula/go.rb
```diff
@@ -34,38 +50,38 @@ class Go < Formula
...
+            :branch => "release-branch.go#{go_version}",
+            :revision => "a84e830bb0d2811304f6e66498eb3123ca97b68e"
+      end
     end
   end
 
-  head do
-    url "https://github.com/golang/go.git"
-
+  if MacOS.version < :sierra
```
MikeMcQuaid: Similarly can you put this inside the `resource` block.

MikeMcQuaid:
A few suggestions and would like to see @DomT4 🆗 this.

ilovezfs:
@MikeMcQuaid PR refreshed.

MikeMcQuaid:
LGTM but will wait on the @DomT4 magic seal of approval

monkseal_byturasphoto_fixed

ilovezfs:
May look closer to 

vckdckc

ilovezfs:
@MikeMcQuaid since stable's bootstrap binary is upgraded here from 1.6.2 -> 1.6.3, should this be getting a revision bump too?

DomT4:
To be honest, I'm really not a fan of this 😕.

The problem isn't formula here, it's our core code & that's where we should be fixing this otherwise all we're doing is committing to do this for every formula impacted between here & stable Sierra; this problem is currently presumably already impacting formulae like `wxmac`.

If you comment out the `depends_on MaximumMacOSRequirement => :el_capitan` blob from `stable do` here and then _install the devel version_ you can do things like `brew install syncthing` no problem:

```
==> Summary
🍺  /usr/local/Cellar/go/1.7rc5: 6,434 files, 250.6M, built in 1 minute 26 seconds
==> Summary
🍺  /usr/local/Cellar/syncthing/0.14.3: 6 files, 12.1M, built in 19 seconds

$ go version
go version go1.7rc5 darwin/amd64
$ syncthing --version
syncthing v0.14.3 "Dysprosium Dragonfly" (go1.7rc5 darwin-amd64) 2016-07-28 11:15:14 UTC
```

So the problem is that for whatever reason Homebrew is reading `MaximumMacOSRequirement` tucked inside a block as applying to the whole formula. The `devel` element itself is not the issue. That's a core bug, not a formula bug.

You can also do things like this:
```ruby
def install
  raise "Please install --devel" if MacOS.version == :sierra && build.stable?
end
```
and then proceed to `brew install go --devel && brew install syncthing` and everything is 🎉.

> update stable bootstrap binary to 1.6.3 for < sierra

The last PR _specifically_ laid out my reasoning for using the _last_ rather than the current stable release, FWIW.

ilovezfs:
@DomT4 I'll put the 1.6.2 back, as I thought the reason was just unavailability of that download. As for the rest of what you said, I refer you to https://github.com/Homebrew/brew/pull/625#issuecomment-238185178

If we're not going to "fix" core because that's not a "bug," then I think this PR is the only option we have that isn't awful for the users.

DomT4:
I'm not completely sure Mike's comment is correct there, unless I'm misunderstanding him, because either the core isn't doing what he says or he's addressing a slightly different issue.

The only blocker here is our usage of `MaximumMacOSRequirement` inside `stable do` - If you remove that `brew` doesn't care that you've installed `--devel` on a parent dependency & contently builds against it as demonstrated. There's nothing else in Homebrew apparently correctly preventing me from doing that.

Interpreted as correct & literal, Mike's comment kind of makes me want to say we're not supporting Sierra _at all_ for formulae that don't build stable on it & they'll just have to wait. This is already a problem for `wxmac` _(I went and checked this time)_:

```
$ brew install -v wxmac --devel
==> Summary
🍺  /usr/local/Cellar/wxmac/3.1.0_2: 811 files, 24M, built in 8 minutes 24 seconds
$ brew install -v wxpython
Warning: You are using OS X 10.12.
We do not provide support for this pre-release version.
You may encounter build failures or other breakages.
Please create pull-requests instead of filing issues.
wxmac: This formula either does not compile or function as expected on OS X
versions newer than El Capitan due to an upstream incompatibility.
```
Sure, we could also update `wxmac` to devel for everyone or just Sierra, but are we going to adopt this logic for _every_ formula appropriate between here & Sierra, bearing in mind we have a whole two maintainers in `core` here every single day & we're barely on top of things as it is 😕.

ilovezfs:
IMO devel and head should be allowed to be used to satisfy dependencies and get added into the paths by superenv, etc. since otherwise devel and head specs are basically a waste of everyone's time, but I'll leave that for you and Mike to sort out.

MikeMcQuaid:
> The only blocker here is our usage of MaximumMacOSRequirement inside stable do - If you remove that brew doesn't care that you've installed --devel on a parent dependency & contently builds against it as demonstrated. There's nothing else in Homebrew apparently correctly preventing me from doing that.

Can we remove that requirement for now? Is it maybe just that last release we didn't use this requirement in a stable block?

> Sure, we could also update wxmac to devel for everyone or just Sierra, but are we going to adopt this logic for every formula appropriate between here & Sierra, bearing in mind we have a whole two maintainers in core here every single day & we're barely on top of things as it is

What are the alternatives?

DomT4:
> What are the alternatives?

My point is simply that if there's nothing stopping me on El Capitan from doing:
```
brew install go --devel && brew install syncthing
```
the addition of `MaximumMacOSRequirement` inside a `stable do` block shouldn't change that behaviour.

> Can we remove that requirement for now? Is it maybe just that last release we didn't use this requirement in a stable block?

I'm 99% sure we used it in the same way when El Capitan was in beta with no issues. I'm sure this is a new issue between then & now, intentional or otherwise. We could remove the requirement & either let users fail or add something like:
```ruby
def install
  raise "Please install --devel" if MacOS.version == :sierra && build.stable?
end
```
but I'm not sure why `MaximumMacOSRequirement` is infecting the entire `brew uses` tree if locked within a `stable do` block that isn't used for compile.

ilovezfs:
Note that this exact same issue just came up again, with user confusion being the outcome of the status quo: https://github.com/Homebrew/homebrew-core/pull/3708#issuecomment-238276952

DomT4:
I'm working on something. 🐻 with me for a bit.

ilovezfs:
@DomT4 if it's circumventing the max os requirement in some way, I don't think that's the right approach.

DomT4:
Well, _this_ was circumventing the requirement in some way really, just on a more local level. I'm all ears on a solution to the broader problem.

ilovezfs:
>all we're doing is committing to do this for every formula impacted between here & stable Sierra

I don't have a problem with that if it's for important stuff like compilers. I don't think it needs to be all or nothing.

If you want to get super specific about just solving the go issue, we could mass replace all `depends_on "go"` with `depends_on :go` and designate in the requirement that "go" from core and a "homebrew/versions/go17" both satisfy the requirement :)

Alternatively, I think some gangrene might be called for, and we could just ship 1.7rc5 as stable for all versions of macOS since the 1.7 release isn't far off and it seems to be working fine for people.

DomT4:
> f you want to get super specific about just solving the go issue

The point is I don't. I don't want us to hack it in every formula this year, and then do it again next year, and that becomes the system. That system is a PoS when it comes to maintainability and maintainers are the biggest thing the core is lacking.

We introduced this problem in core code & frankly it feels like a farce that we're talking about dancing around it on tiptoes instead of accepting a requirement _scoped_ to `stable`, `devel` or `head` should not manage to infect the entire `brew uses` tree when that specific block is unused by an installation.

ilovezfs:
The issue is that by definition `depends_on "go"` is only satisfied by the current, stable version, the current stable version does not work on Sierra, therefore error. That is correct behavior, unless you want to make it acceptable for `depends_on "go"` to be satisfied by some version other than the current stable version.

DomT4:
> The issue is that by definition depends_on "go" is only satisfied by the current, stable version

Except it isn't on El Capitan, as I pointed out earlier. On _any system_ beside Sierra I can `brew install go --devel` and follow up with `brew install syncthing` and `brew` is a happy happy tyrant. 

_Even on_ Sierra if I remove the `MaximumMacOSRequirement` line I can use devel as a replacement for stable fine. The problems only start when `MaximumMacOSRequirement` is introduced anywhere in the deps/reqs tree.

So no, not really, `depends_on "go"` is not only satisfied by the current, stable version. That may be the design intention but it's not reflected in practical application. That limitation is true on Sierra, and only true because `MaximumMacOSRequirement` worms its way into the recursive deps/reqs tree.

ilovezfs:
I see, superenv is using opt bin, and dependency is considered fulfilled. Cool. So maybe we can teach max os requirement not to fail if the formula is already installed (in the sense of current stable, devel, or head)?

DomT4:
Indeed. What I was playing with before this conversation blew up a bit was:
```diff
diff --git a/Library/Homebrew/formula_installer.rb b/Library/Homebrew/formula_installer.rb
index d8d1a8f..d3bb276 100644
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -308,6 +308,7 @@ class FormulaInstaller
 
     req_map.each_pair do |dependent, reqs|
       reqs.each do |req|
+        next if dependent.installed?
         puts "#{dependent}: #{req.message}"
         fatals << req if req.fatal?
       end
```

ilovezfs:
I was about to suggest this:
```
iMac-TMP:local joe$ git diff
diff --git a/Library/Homebrew/requirement.rb b/Library/Homebrew/requirement.rb
index 7a40d41..c90e977 100644
--- a/Library/Homebrew/requirement.rb
+++ b/Library/Homebrew/requirement.rb
@@ -199,6 +199,7 @@ class Requirement
 
       formulae.each do |f|
         f.requirements.each do |req|
+          next if f.installed?
           if prune?(f, req, &block)
             next
           else
```

DomT4:
Both seem to work fine, checking my 10.12 box. Don't have a strong preference. Might be worth waiting to see if Mike is content with either.

ilovezfs:
I suspect both are :-1:.

ilovezfs:
@DomT4 for example, suppose you uninstall python3 but it's actually required by one of the dependencies. Currently the requirement would trigger an install of python3 and all would be well. With either of the above hacks, it would fail, and we'd have a new issue on our doorstep.

DomT4:
Can you elaborate on that a bit?
```
vagrant$ brew rm python3
vagrant$ brew install -s -v you-get
Warning: You are using OS X 10.12.
We do not provide support for this pre-release version.
You may encounter build failures or other breakages.
Please create pull-requests instead of filing issues.
==> Installing dependencies for you-get: python3
==> Installing you-get dependency: python3
```

That's tested against both my & your proposed tweaks. Maybe I'm misunderstanding.

ilovezfs:
that's a requirement in you-get itself, not in a dependency of you-get, so of course it wouldn't be affected ...

DomT4:
Okay, so care to provide a more explicit reproducible example at all? I also tested `gedit`, which uses `iso-codes`, which requires `python3`, if that's any better.

ilovezfs:
The change I suggested in `requirement.rb` does break with the `gedit` example, but your `fatals` specific check in `formula_installer.rb` does not break, so your change is looking good. This PR will still provide a significantly better user experience since `brew install foo` will work even if go isn't already installed, whereas your proposal requires the user to guess or research or divine that maybe the devel spec will work and thereby stumble into a working configuration, but I'm not necessarily opposed to a relatively terrible user experience if you're opposed to the extra work a better experience entails ;)

gabriel:
FYI, rc6 was just released: https://github.com/golang/go/compare/go1.7rc5...go1.7rc6

ilovezfs:
In this particular case, I think the most reasonable thing to do is declare 1.7 stable now and move on with life, since it's a better experience for the users and no extra work for us.

ilovezfs:
@DomT4 I think what you're suggesting may still be problematic though
```
requirements/apr_requirement.rb
4:  fatal true

requirements/emacs_requirement.rb
2:  fatal true

requirements/cctools_requirement.rb
7:  fatal true

requirements/fortran_requirement.rb
4:  fatal true

requirements/java_requirement.rb
4:  fatal true

requirements/language_module_requirement.rb
4:  fatal true

requirements/maximum_macos_requirement.rb
4:  fatal true

requirements/minimum_macos_requirement.rb
4:  fatal true

requirements/mpi_requirement.rb
9:  fatal true

requirements/osxfuse_requirement.rb
4:  fatal true
21:  fatal true

requirements/perl_requirement.rb
2:  fatal true

requirements/python_requirement.rb
4:  fatal true
57:  fatal true

requirements/ruby_requirement.rb
2:  fatal true

requirements/tuntap_requirement.rb
4:  fatal true

requirements/unsigned_kext_requirement.rb
4:  fatal true

requirements/x11_requirement.rb
7:  fatal true

```

Are you sure all of those should not be fatal if the formula is already installed?

ilovezfs:
@DomT4 It might be safer to provide a method like `req.waive_if_installed?` which would be true for only MaximumMacOSRequirement and MinimumMacOSRequirement, and any others for which it makes sense.
```
     req_map.each_pair do |dependent, reqs|
       reqs.each do |req|
+        next if req.waive_if_installed? && dependent.installed?
         puts "#{dependent}: #{req.message}"
         fatals << req if req.fatal?
       end
```

ilovezfs:
Another thing that might help in addition to your suggested workaround would be allowing requirements to take an optional custom message argument to add to the generic message, since then we could actually suggest installing devel rather than creating a research project for everyone.

DomT4:
> whereas your proposal requires the user to guess or research or divine that maybe the devel spec will work and thereby stumble into a working configuration

I discussed with Mike a long, long time ago now nudging users towards `devel` or `--HEAD` with the MMOSR message but there was enough faith users would find it. Might be able to tweak the requirement to take a block with an additional explanation/comment when necessary, but smells unnecessary.

There's a valid argument that users running a prerelease system should be prepared to have to use prerelease software, and shouldn't need hand-holding to get there.

> Are you sure all of those should not be fatal if the formula is already installed?

To be fair, I'm pretty sure all of those are fulfilled by having a formula or cask installed?

> It might be safer to provide a method like

You could also just scope it to this specific scenario if we really wanted to drill specific:
```diff
diff --git a/Library/Homebrew/formula_installer.rb b/Library/Homebrew/formula_installer.rb
index d8d1a8f..111efb9 100644
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -308,6 +308,7 @@ class FormulaInstaller
 
     req_map.each_pair do |dependent, reqs|
       reqs.each do |req|
+        next if dependent.installed? && req.name == "maximummacos"
         puts "#{dependent}: #{req.message}"
         fatals << req if req.fatal?
       end
```

ilovezfs:
>To be fair, I'm pretty sure all of those are fulfilled by having a formula or cask installed beside the obvious two?

As an example, an obvious problem is the version constraints. It's useless to specify a minimum Java or emacs requirement if the check is skipped when it's already installed. That defeats the purpose which is to check if the min version is satisfied for that specific dependent. Since the requirements can take options that vatu dependent by dependent you can't just assume installed means all requirements should now be non-fatal for that formula all of the time.

I don't find the "you should know to test beta and head versions" a reasonable expectation when we could easily choose not to waste people's time when they fail to live up to that standard.

>smells unnecessary

Of course it's unnecessary since we shouldn't be supporting Sierra at all, but if we are going to, it might as well attempt not to waste everyone's time doing redundant research. I'm serious about not supporting Sierra at all though. Maybe we should be putting all Sierra specific stuff in a branch since it feels right now like we're saying one thing (we support Sierra) and doing another (supporting Sierra in an embarrassing, half-baked manner).

>You could also just scope it ...

Of course, but using a method for that keeps the metadata about the requirement with the requirement instead of scattering hard-coded information about a particular subset of the requirements in some other file.

ilovezfs:
Anyway, after reflecting on this, I think the best answers are what's in this PR or setting 1.7.5 as stable for all macOS versions.

ilovezfs:
@MikeMcQuaid given that @DomT4's objections seem to be related to work load and distribution of labor burdens, not technical or stylistic qualms with the changes, I think this should ship and the issue can be revisited if the predictions of an unreasonable amount of extra work turn out to match reality.

DomT4:
I filed https://github.com/Homebrew/brew/pull/662.

> If you have a better idea, please file a competing PR. I'm sick to death of discussion.

Take this to heart, particularly. If people want to decide we should fix this in every single formula it comes up in forevermore, fine, but I'm dumping the maintenance burden squarely on your shoulders.

ilovezfs:
As that's obviously a totally unacceptable state of affairs, this is closed.

PR #3706, opened 2016-08-08
smartmontools 6.5

txaj:
--enable-savestates and --enable-attributelog were renamed respectively
to --with-savestates and --with-attributelog :
https://www.smartmontools.org/browser/tags/RELEASE_6_5/smartmontools/ChangeLog#L652

explicit --enable-drivedb flag removed because the defaults is yes.


dunn:
Thanks! Merged as 2dab01f1eed18dbeabb4b68ee58024deb68a8948.

PR #3707, opened 2016-08-08
owfs 3.1p3

ilovezfs:
- D_POSIX_C_SOURCE inreplace isn't needed with current configure options, though the underlying bug is still not fixed
- drop dependency on libusb since it's not used with current configure options


ilovezfs:
@DomT4 :ship:?

MikeMcQuaid:
@ilovezfs you're a big boy now, you can feel free to 🚢 anything like this at your discretion without needing an 🆗.

ilovezfs:
LOL, you crack me up sometimes.

PR #3710, opened 2016-08-08
owfs: enable debug, usb, ftdi

ilovezfs:
upstream requested not to pass --disable-debug:

>  Don't do --disable-debug. There is the runtime switch --debug which
>  enables debug. Only if the small time for testing that switch bothers
>  you (unlikely), you can --disable-debug. Don't do that! Reason is we
>  have no chance to debug any user problems if you --disable-debug.

Also, enable usb and ftdi support as upstream confirmed this is working
on macOS.


MikeMcQuaid:
👍

PR #3713, opened 2016-08-08
voltdb: Fix install block

dj-m:

---

Added %r around the regular expression in the install block & after-comma spacing

### DIFF
--- a/Formula/voltdb.rb
+++ b/Formula/voltdb.rb
@@ -1,9 +1,9 @@
 class Voltdb < Formula
   desc "Horizontally-scalable, in-memory SQL RDBMS"
   homepage "https://github.com/VoltDB/voltdb"
-  head "https://github.com/VoltDB/voltdb.git"
   url "https://github.com/VoltDB/voltdb/archive/voltdb-5.6.tar.gz"
   sha256 "9ea24d8cacdf2e19ba60487f3e9dfefa83c18cb3987571abc44b858ce0db7c3e"
+  head "https://github.com/VoltDB/voltdb.git"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Nice work @dj-m ✨

Formula/voltdb.rb
```diff
@@ -17,11 +17,15 @@ class Voltdb < Formula
...
+    inreplace Dir["bin/*"] - ["bin/voltadmin", "bin/voltdb", "bin/rabbitmqloader"],
+      %r{VOLTDB_LIB=\$VOLTDB_HOME\/lib}, "VOLTDB_LIB=$VOLTDB_HOME/lib/voltdb"
 
     (lib/"voltdb").install Dir["lib/*"]
     lib.install_symlink lib/"voltdb/python"
     prefix.install "bin", "tools", "voltdb", "version.txt", "doc"
   end
+
+  test do
+    assert_match version.to_s, shell_output("voltdb --version")
```
DomT4: Just FYI, in future, calling an executable that lives inside the Cellar should _always_ be done with a full path, i.e. `#{bin}/voltdb --version`

MikeMcQuaid: @DomT4 Missed that, will fix up.

DomT4: Beat ya to it 😃.

MikeMcQuaid: You already did in e934d03 I see 😀

dj-m:
Should it be:

`system bin/"voltdb", "--version"`

in the test block?

DomT4:
Heh, this is where it gets a bit silly. The preferred style for non-asserted tests is:
```ruby
system bin/"echo", "horse"
```
for example, but if you're asserting we prefer:
```ruby
assert_match "horse", shell_output("#{bin}/echo horse")
```
Or such. I wasn't too fussed about mixing & matching `#{bin}` and `bin` in the test blocks but ILZ has been on a kick against that so trying to save him stuff to fix later 😄.

dj-m:
@MikeMcQuaid Thanks for the info! I know the string interpolation can be tricky... 🤔

Formula/voltdb.rb
DomT4: Yeah, if I leave it it only sits on my mind until I do something about it 😅.

PR #3733, opened 2016-08-09
qscintilla2 with qt5 support

schoeps:

---

This pull request addresses https://github.com/Homebrew/legacy-homebrew/issues/37591 and https://github.com/Homebrew/legacy-homebrew/issues/41464. This changset als sets python as the default switch for pyqt5 (i.e., not python3) since this seems more consistent with all other formulae.

### DIFF
--- a/Formula/qscintilla2.rb
+++ b/Formula/qscintilla2.rb
@@ -14,7 +14,8 @@ class Qscintilla2 < Formula
   end
 
   option "with-plugin", "Build the Qt Designer plugin"
-  option "without-python", "Skip building the Python bindings"
+  option "with-python", "Build Python bindings"
+  option "without-python3", "Do not build Python3 bindings"
 
   depends_on "qt5"
   depends_on :python3 => :recommended
--- a/Formula/qscintilla2.rb
+++ b/Formula/qscintilla2.rb
@@ -32,9 +32,6 @@ class Qscintilla2 < Formula
   end
 
   def install
-    # Make sure Qt5 is found even if not linked
-    ENV.prepend_path "PATH", Formula["qt5"].bin
-
     spec = ENV.compiler == :clang && MacOS.version >= :mavericks ? "macx-clang" : "macx-g++"
     args = %W[-config release -spec #{spec}]
 
@@ -66,7 +63,6 @@ class Qscintilla2 < Formula
           system python, "configure.py", "-o", lib, "-n", include,
                            "--apidir=#{prefix}/qsci",
                            "--destdir=#{lib}/python#{version}/site-packages/PyQt5",
-                           "--stubsdir=#{lib}/python#{version}/site-packages/PyQt5",
                            "--qsci-sipdir=#{share}/sip",
                            "--pyqt=PyQt5",
                            "--pyqt-sipdir=#{Formula["pyqt5"].opt_share}/sip/Qt5",
--- a/Formula/qscintilla2.rb
+++ b/Formula/qscintilla2.rb
@@ -3,6 +3,7 @@ class Qscintilla2 < Formula
   homepage "https://www.riverbankcomputing.com/software/qscintilla/intro"
   url "https://downloads.sf.net/project/pyqt/QScintilla2/QScintilla-2.9.3/QScintilla_gpl-2.9.3.tar.gz"
   sha256 "98aab93d73b05635867c2fc757acb383b5856a0b416e3fd7659f1879996ddb7e"
+  revision 1
 
   bottle do
     cellar :any
@@ -12,28 +13,29 @@ class Qscintilla2 < Formula
     sha256 "5a265016e1a21e110bb9c344e7db8fd41b5690df5ffc359394c6e62aad8af494" => :mavericks
   end
 
-  option "without-plugin", "Skip building the Qt Designer plugin"
+  option "with-plugin", "Build the Qt Designer plugin"
   option "without-python", "Skip building the Python bindings"
 
-  depends_on :python => :recommended
-  depends_on :python3 => :optional
+  depends_on "qt5"
+  depends_on :python3 => :recommended
+  depends_on :python => :optional
 
-  if build.with? "python3"
-    depends_on "pyqt" => "with-python3"
-  elsif build.with? "python"
-    depends_on "pyqt"
-  else
-    depends_on "qt"
+  if build.with?("python") && build.with?("python3")
+    depends_on "sip" => "with-python3"
+    depends_on "pyqt5" => "with-python"
+  elsif build.with?("python")
+    depends_on "sip"
+    depends_on "pyqt5" => "with-python"
+  elsif build.with?("python3")
+    depends_on "sip" => "with-python3"
+    depends_on "pyqt5"
   end
 
   def install
-    # On Mavericks we want to target libc++, this requires an
-    # unsupported/macx-clang-libc++ flag.
-    if ENV.compiler == :clang && MacOS.version >= :mavericks
-      spec = "unsupported/macx-clang-libc++"
-    else
-      spec = "macx-g++"
-    end
+    # Make sure Qt5 is found even if not linked
+    ENV.prepend_path "PATH", Formula["qt5"].bin
+
+    spec = ENV.compiler == :clang && MacOS.version >= :mavericks ? "macx-clang" : "macx-g++"
     args = %W[-config release -spec #{spec}]
 
     cd "Qt4Qt5" do
@@ -63,10 +65,12 @@ class Qscintilla2 < Formula
           (share/"sip").mkpath
           system python, "configure.py", "-o", lib, "-n", include,
                            "--apidir=#{prefix}/qsci",
-                           "--destdir=#{lib}/python#{version}/site-packages/PyQt4",
-                           "--stubsdir=#{lib}/python#{version}/site-packages/PyQt4",
+                           "--destdir=#{lib}/python#{version}/site-packages/PyQt5",
+                           "--stubsdir=#{lib}/python#{version}/site-packages/PyQt5",
                            "--qsci-sipdir=#{share}/sip",
-                           "--pyqt-sipdir=#{HOMEBREW_PREFIX}/share/sip",
+                           "--pyqt=PyQt5",
+                           "--pyqt-sipdir=#{Formula["pyqt5"].opt_share}/sip/Qt5",
+                           "--sip-incdir=#{Formula["sip"].opt_include}",
                            "--spec=#{spec}"
           system "make"
           system "make", "install"
@@ -79,7 +83,7 @@ class Qscintilla2 < Formula
       mkpath prefix/"plugins/designer"
       cd "designer-Qt4Qt5" do
         inreplace "designer.pro" do |s|
-          s.sub! "$$[QT_INSTALL_PLUGINS]", "#{lib}/qt4/plugins"
+          s.sub! "$$[QT_INSTALL_PLUGINS]", "#{lib}/qt5/plugins"
           s.sub! "$$[QT_INSTALL_LIBS]", lib
         end
         system "qmake", "designer.pro", *args
@@ -90,9 +94,9 @@ class Qscintilla2 < Formula
   end
 
   test do
-    Pathname("test.py").write <<-EOS.undent
-      import PyQt4.Qsci
-      assert("QsciLexer" in dir(PyQt4.Qsci))
+    (testpath/"test.py").write <<-EOS.undent
+      import PyQt5.Qsci
+      assert("QsciLexer" in dir(PyQt5.Qsci))
     EOS
     Language::Python.each_python(build) do |python, _version|
       system python, "test.py"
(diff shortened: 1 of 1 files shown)


Formula/pyqt5.rb
```diff
@@ -3,6 +3,7 @@ class Pyqt5 < Formula
   homepage "https://www.riverbankcomputing.com/software/pyqt/download5"
   url "https://downloads.sourceforge.net/project/pyqt/PyQt5/PyQt-5.6/PyQt5_gpl-5.6.tar.gz"
   sha256 "2e481a6c4c41b96ed3b33449e5f9599987c63a5c8db93313bd57a6acbf20f0e1"
+  revision 2
```
MikeMcQuaid: Should be `1`

MikeMcQuaid:
Installation is failing on all platforms: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/6046/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_qscintilla2/



> Have you built your formula locally prior to submission with brew install <formula> (where <formula> is the name of the formula you're submitting)?

Did this work for you?

schoeps:
Yes it did, that's why I made a cross in the box...

I looked at the results. Error is the following
```
mkdir: /usr/local/Cellar/qt/4.8.7_2/qsci/api/python/: Operation not permitted
mkdir: /usr/local/opt/pyqt/share/sip/Qsci/: Operation not permitted
```
There seems to be an issue  with the installation paths of qscintilla. My local installation did not care about this.

MikeMcQuaid:
You can replicate this behaviour with `export HOMEBREW_SANDBOX=1`; it's a security feature that we need fixed before we can merge.

schoeps:
is the sandbox rule new? I believe the original qscintilla2 was already broken.

schoeps:
...and I am afraid I am not able to clean this up.

schoeps:
I am surprised but if it's ok for Jenkins then it's ok for me :)

Formula/pyqt5.rb
```diff
@@ -15,8 +16,8 @@ class Pyqt5 < Formula
 
   deprecated_option "enable-debug" => "with-debug"
 
-  depends_on :python3 => :recommended
-  depends_on :python => :optional
+  depends_on :python => :recommended
+  depends_on :python3 => :optional
```
MikeMcQuaid: I remembered why they were the way round they were: otherwise the bottles will always require installing Homebrew's Python.

schoeps: Now the bottle is always asking for Homebrew's python3, right? This is not too great, either?

I could do it as in https://github.com/Homebrew/homebrew-core/blob/master/Formula/pyqt.rb:
```
option "without-python", "Build without python 2 support"
depends_on :python3 => :optional
```
but is the loop `Language::Python.each_python(build) do |python, version|` then still taking care of python 2?

MikeMcQuaid: We pretty much always mandate our own Python 3 so we've not had complaints about that.

I think it's worth just leaving this file as-is for now (or at least pulling into another PR)

schoeps: I dropped my changeset to pyqt5. However, things are still inconsistent among pyqt,pyqt5 and qscintilla2.

Therefore I still ask in `qscintilla2.rb` for explicit `pyqt --with-python3` and `pyqt5 --with-python3` (albeit the latter is default, the former not).

Formula/qscintilla2.rb
```diff
@@ -11,26 +12,30 @@ class Qscintilla2 < Formula
...
-    depends_on "pyqt" => "with-python3"
-  elsif build.with? "python"
-    depends_on "pyqt"
+  if build.with?("python") || build.with?("python3")
+    depends_on "sip" => ("with-python3" if build.with?("python3"))
+    depends_on "pyqt5" => ("with-python3" if build.with?("python3")) if build.with?("qt5")
+    depends_on "pyqt" => ("with-python3" if build.with?("python3")) if build.without?("qt5")
   else
-    depends_on "qt"
+    depends_on "qt5" if build.with?("qt5")
```
MikeMcQuaid: `depends_on "qt5" => :optional`

Formula/qscintilla2.rb
```diff
@@ -11,26 +12,30 @@ class Qscintilla2 < Formula
...
+    depends_on "qt" if build.without?("qt5")
```
MikeMcQuaid: `depends_on "qt" => :recommended`

Formula/qscintilla2.rb
```diff
@@ -11,26 +12,30 @@ class Qscintilla2 < Formula
...
   option "without-python", "Skip building the Python bindings"
 
   depends_on :python => :recommended
   depends_on :python3 => :optional
 
-  if build.with? "python3"
-    depends_on "pyqt" => "with-python3"
-  elsif build.with? "python"
-    depends_on "pyqt"
+  if build.with?("python") || build.with?("python3")
```
MikeMcQuaid: Can you split the `build.with?("python3")` into another `if`; it's a bit hard to follow this as-is.

MikeMcQuaid:
We're dropping Qt4 support very soon so another option to simplify this would be to always build against Qt5.

Formula/qscintilla2.rb
```diff
@@ -59,25 +64,33 @@ def install
     if build.with?("python") || build.with?("python3")
       cd "Python" do
         Language::Python.each_python(build) do |python, version|
+          pyargs = ["--apidir=#{prefix}/qsci"]
+          pyargs << "--spec=#{spec}"
+          pyargs << "--qsci-sipdir=#{share}/sip"
+          if build.with?("qt5")
+            pyargs << "--pyqt=PyQt5"
+            pyargs << "--destdir=#{lib}/python#{version}/site-packages/PyQt5"
+            pyargs << "--pyqt-sipdir=#{Formula["pyqt5"].opt_prefix}/share/sip/Qt5"
```
DomT4: You're looking for `opt_share` here rather than `opt_prefix/share`. Is it actually installing anything there or is this purely a reference?

Formula/qscintilla2.rb
```diff
@@ -59,25 +64,33 @@ def install
...
+          else
+            pyargs << "--pyqt=PyQt4"
+            pyargs << "--destdir=#{lib}/python#{version}/site-packages/PyQt4"
+            pyargs << "--pyqt-sipdir=#{Formula["pyqt"].opt_prefix}/share/sip"
```
DomT4: Same thing re `opt_share`

Formula/qscintilla2.rb
```diff
@@ -88,10 +101,17 @@ def install
   end
 
   test do
-    Pathname("test.py").write <<-EOS.undent
-      import PyQt4.Qsci
-      assert("QsciLexer" in dir(PyQt4.Qsci))
-    EOS
+    if build.with?("qt5")
+      Pathname("test.py").write <<-EOS.undent
```
DomT4: Don't need the `Pathname`. Just `(testpath/"test.py").write <<-EOS.undent` etc.

schoeps:
some recommendation seem to have broken qt5 :(

schoeps:
ok, the problem is that `qscintilla2 --with-python3` also tries to compile for python2 which is not necessarily included in pyqt5. What to do? I would need to distinguish four cases: `--with-python`, `--with-python3`, `--with-python` & `--with-python3`, `--without-python`....

Formula/qscintilla2.rb
```diff
@@ -11,26 +12,38 @@ class Qscintilla2 < Formula
...
   depends_on :python => :recommended
   depends_on :python3 => :optional
 
-  if build.with? "python3"
-    depends_on "pyqt" => "with-python3"
-  elsif build.with? "python"
-    depends_on "pyqt"
+  if build.with?("python") && build.with?("python3")
+    depends_on "sip" => "with-python3" # python2 is recommended
+    depends_on "pyqt5" => "with-python" if build.with?("qt5") # python3 is recommended
```
MikeMcQuaid:
```ruby
if build.with?("qt5")
  depends_on "pyqt5" => "with-python"
else
  depends_on "pyqt" => "with-python3"
end
```

MikeMcQuaid: (and same below)

Formula/qscintilla2.rb
```diff
@@ -11,26 +12,38 @@ class Qscintilla2 < Formula
...
 
   def install
-    # On Mavericks we want to target libc++, this requires an
-    # unsupported/macx-clang-libc++ flag.
     if ENV.compiler == :clang && MacOS.version >= :mavericks
-      spec = "unsupported/macx-clang-libc++"
-    else
+      if build.with?("qt5")
+        spec = "macx-clang"
+      else # if Qt4
```
MikeMcQuaid: Can drop this comment.

Formula/qscintilla2.rb
```diff
@@ -11,26 +12,38 @@ class Qscintilla2 < Formula
...
+        spec = "unsupported/macx-clang-libc++"
+      end
+    else # if gcc and (Qt4 or Qt5)
```
MikeMcQuaid: Can drop this comment.

Formula/qscintilla2.rb
```diff
@@ -11,26 +12,38 @@ class Qscintilla2 < Formula
...
   else
-    depends_on "qt"
+    depends_on "qt5" => :optional
+    depends_on "qt" => :recommended
   end
 
   def install
-    # On Mavericks we want to target libc++, this requires an
-    # unsupported/macx-clang-libc++ flag.
     if ENV.compiler == :clang && MacOS.version >= :mavericks
```
MikeMcQuaid: Could do `spec = if ENV.compiler == :clang && MacOS.version >= :mavericks` and drop the `spec =` below.

Formula/qscintilla2.rb
```diff
@@ -59,25 +72,34 @@ def install
...
+          system python, "configure.py", "-o", lib, "-n", include, *pyargs
           system "make"
           system "make", "install"
           system "make", "clean"
         end
       end
     end
 
-    if build.with? "plugin"
+    if build.with?("plugin")
```
MikeMcQuaid: No need to change stuff like this.

MikeMcQuaid:
Needs rebased on `master` as well as my tweaks. Cases look good so far 👍.

Formula/qscintilla2.rb
schoeps: Sorry, I don't get this one.

MikeMcQuaid: Instead if `if foo` with a `spec = ` below` you can do `spec if foo` in Ruby and rely on the implicit return from statements.

schoeps:
@MikeMcQuaid: I think everything is rather nice now but do you happen to know why python does not find Qsci anymore?
```
==> python test.py
  File "test.py", line 1, in <module>
    import PyQt5.Qsci
ImportError: No module named PyQt5.Qsci
Error: qscintilla2: failed
Failed executing: python test.py
```

Formula/qscintilla2.rb
```diff
@@ -42,11 +42,14 @@ def install
...
         s.gsub! "$$[QT_INSTALL_TRANSLATIONS]", prefix/"trans"
         s.gsub! "$$[QT_INSTALL_DATA]", prefix/"data"
+        s.gsub! "$$[QT_HOST_DATA]", prefix/"data"
+        #s.gsub! "TARGET = qscintilla2", "TARGET = qscintilla2-qt5"
       end
 
       inreplace "features/qscintilla2.prf" do |s|
         s.gsub! "$$[QT_INSTALL_LIBS]", lib
         s.gsub! "$$[QT_INSTALL_HEADERS]", include
+        #s.gsub! "LIBS += -lqscintilla2", "LIBS += -lqscintilla2-qt5"
```
MikeMcQuaid: Can remove this commented-out line.

Formula/qscintilla2.rb
```diff
@@ -42,11 +42,14 @@ def install
         s.gsub! "$$[QT_INSTALL_HEADERS]", include
         s.gsub! "$$[QT_INSTALL_TRANSLATIONS]", prefix/"trans"
         s.gsub! "$$[QT_INSTALL_DATA]", prefix/"data"
+        s.gsub! "$$[QT_HOST_DATA]", prefix/"data"
+        #s.gsub! "TARGET = qscintilla2", "TARGET = qscintilla2-qt5"
```
MikeMcQuaid: Can remove this commented-out line.

MikeMcQuaid:
Tests are 💚. A couple of final nits then happy to 🚢 this.

MikeMcQuaid:
Thanks @schoeps!

DomT4:
Was this tested against `sqliteman` & `codequery` to ensure having them linked recursively to two different Qt branches didn't cause anything odd? The latter should be using `qt5` already, but IIRC development on the former is pretty 💀.

MikeMcQuaid:
@DomT4 It was not. At this point I figured something without broken CI that crops up later was going to become either fixed or boneyarded shortly anyway.

ruben2020:
Shouldn't you have another version of qscintilla2 for qt4? You can call it `qscintilla2-qt4`

ilovezfs:
@ruben2020 it's pretty unlikely we'll ship something like that even in homebrew/versions. Is there a reverse dependency you have in mind that you think requires qscintilla2 and qt4?

ruben2020:
I'm the owner of `codequery`. But it's no problem for me to switch to qt5. I've been building it for qt5 experimentally before anyway.

ilovezfs:
@ruben2020 Cool! I assume you've seen https://github.com/Homebrew/homebrew-core/pull/4258

ruben2020:
Thanks @ilovezfs . I didn't know about it. I will check it out.

PR #2279, opened 2016-06-22
Erlang 19.0

erichkist:

PR #3761, opened 2016-08-10
wrangler 1.2

ilovezfs:
upstream fix for erlang 19
add a head spec
reparallelize the build
add a test

### DIFF
--- a/Formula/wrangler.rb
+++ b/Formula/wrangler.rb
@@ -1,8 +1,18 @@
 class Wrangler < Formula
   desc "Refactoring tool for Erlang with emacs and Eclipse integration"
   homepage "https://www.cs.kent.ac.uk/projects/wrangler/Wrangler/"
-  url "https://github.com/RefactoringTools/wrangler/archive/wrangler1.1.01.tar.gz"
-  sha256 "e61c13ec772e137efdcf5aa8f21210ef424eac3ee2b918efe5e456985bb19626"
+  head "https://github.com/RefactoringTools/wrangler.git"
+
+  stable do
+    url "https://github.com/RefactoringTools/wrangler/archive/wrangler1.2.tar.gz"
+    sha256 "7b8700a3f8c3ef6a91d4c6ddd71e8c6cfc0816ee799a1d860a9e2955456e66f3"
+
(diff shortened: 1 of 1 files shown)


ilovezfs:
Version in core is currently broken due to erlang 19 (and would need same patch as here), which was not noticed during https://github.com/Homebrew/homebrew-core/pull/2279 apparently.

DomT4:
Yeah, this had no test so would have been completely ignored by CI. AFAIK `erlang` also doesn't ship any dynamic libraries, so `linkage` wouldn't have spotted anything off either.

Formula/wrangler.rb
```diff
@@ -13,9 +23,13 @@ class Wrangler < Formula
...
   def install
-    ENV.deparallelize
     system "./configure", "--prefix=#{prefix}"
     system "make"
     system "make", "install"
   end
+
+  test do
+    suffixtree = Dir.glob("#{lib}/erlang/*/*/*/suffixtree").shift
+    assert File.executable?(suffixtree), "suffixtree must be executable"
```
MikeMcQuaid: Do we have anything we can run (even for a `--version`)? I fear this won't detect breakage.

ilovezfs: none that I could find, or it would be here!

ilovezfs:
I think erlang stuff may need to be revision bumped always until the new version stuff is in place because my general understanding is the builds are highly likely to break unless upstream has already deliberately adapted the build to the latest. In this I think this thing is mainly emacs bindings? But I wouldn't be surprised if there are other broken erlang builds right now.

ilovezfs:
Yeah, erlang 19 also broke disco, and yaws has a checksum mismatch but their current tarball builds.

ilovezfs:
Time for disco to die I think: https://github.com/Homebrew/homebrew-boneyard/pull/141.

ilovezfs:
Yaws fix: https://github.com/Homebrew/homebrew-core/pull/3794.

apjanke:
I think this is ready to merge?

ilovezfs:
@apjanke assuming objections to awful test have been dropped ...

ilovezfs:
@apjanke :ship:'d.

PR #3764, opened 2016-08-10
libjson-rpc-cpp 0.7.0 (devel)

ilovezfs:
also cherry-pick upstream fix for parallelization issue into stable


cinemast:
Great, thanks!

PR #3163, opened 2016-07-20
nvm 0.31.4

ilovezfs:
Created with `brew bump-formula-pr`.

PR #3771, opened 2016-08-10
nvm v0.31.4

sscotth:

---

Adjust test to match new pattern. Fixes #3163.

### DIFF
--- a/Formula/nvm.rb
+++ b/Formula/nvm.rb
@@ -1,8 +1,8 @@
 class Nvm < Formula
   desc "Manage multiple Node.js versions"
   homepage "https://github.com/creationix/nvm"
-  url "https://github.com/creationix/nvm/archive/v0.31.2.tar.gz"
-  sha256 "d2a3253d3ac992cd26c56cded4e014896e1d3d1a6394bac0e719c46fa90a8bb9"
+  url "https://github.com/creationix/nvm/archive/v0.31.4.tar.gz"
+  sha256 "e88a804a83c9a1b3363b83d20ac37a893e7207a8b8e8c34880765c4f90d38dd2"
   head "https://github.com/creationix/nvm.git"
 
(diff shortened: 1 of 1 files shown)


Formula/nvm.rb
```diff
@@ -39,6 +39,6 @@ def caveats; <<-EOS.undent
     output = pipe_output("NODE_VERSION=homebrewtest #{prefix}/nvm-exec 2>&1")
     assert_no_match /No such file or directory/, output
     assert_no_match /nvm: command not found/, output
-    assert_match "N/A: version \"homebrewtest\" is not yet installed", output
+    assert_match /N\/A: version "N\/A" is not yet installed/, output
```
DomT4: This should be fine to remain a `""` assert rather than a regex one, no?

sscotth: There are additional statements after the "is not yet installed". The full output is:

```
N/A: version "N/A" is not yet installed.

You need to run "nvm install N/A" to install it before using it.
```

DomT4: `assert_match` is happy doing a partial assert though, even with `""`. `assert_equal` is the one that wants the whole cake, I believe.

`assert_no_match` won't take a string, but `assert_match` will. We prefer to use a string when possible because it means we don't have to escape slashes or full stops, which helps with readability at a glance.

sscotth: I see. I found other examples of assert_match with a regex so I just assumed.

https://github.com/Homebrew/homebrew-core/search?utf8=%E2%9C%93&q=assert_match

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/da051def8b8ee65f4b606948d87c083438a0353b. Thank you for your contribution to Homebrew @sscotth; we appreciate it! :smiley_cat:

PR #3781, opened 2016-08-11
gitlab-ci-multi-runner 1.4.2

ilovezfs:
style cleanups

### DIFF
--- a/Formula/gitlab-ci-multi-runner.rb
+++ b/Formula/gitlab-ci-multi-runner.rb
@@ -4,8 +4,8 @@ class GitlabCiMultiRunner < Formula
   desc "The official GitLab CI runner written in Go"
   homepage "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner"
   url "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner.git",
-    :tag => "v1.4.1",
-    :revision => "fae8f189cd367d870c3d41471ba569070acee2e1"
+    :tag => "v1.4.2",
+    :revision => "bcc1794f6dd7edf8e3984bfbf37c2baf9887f03a"
   head "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


DomT4:
👍

ilovezfs:
@DomT4 hopefully slightly less grim than before?

Also, based on my testing of various ldflags approaches in this formula, I believe https://github.com/Homebrew/homebrew-core/blob/master/Formula/docker-machine-driver-xhyve.rb#L34 may be broken because as far as I can tell using the -ldflags option multiple times never works.

Formula/gitlab-ci-multi-runner.rb
```diff
@@ -24,51 +24,46 @@ class GitlabCiMultiRunner < Formula
...
 
     cd dir do
       resource("prebuilt-x86_64.tar.gz").stage do
-        system "go-bindata", "-pkg", "docker",
-               "-nocompress", "-nomemcopy", "-nometadata",
-               "-o", dir/"executors/docker/bindata.go",
-               "prebuilt-x86_64.tar.gz"
+        system *%W[go-bindata -pkg docker -nocompress -nomemcopy -nometadata
+                   -o #{dir}/executors/docker/bindata.go prebuilt-x86_64.tar.gz]
       end
```
ilovezfs: @DomT4 also I'm not sure if this is idiomatic or if we prefer what was there already?

DomT4:
Yeah, looks nicer. The spacing on the version check in the test always annoyed me, so 😺.

I don't spend a lot of time in go passing `-ldflags` so I'm not sure, but I'd be surprised if simply passing the `-ldflags` with new arguments repeatedly wasn't in some way problematic, so `docker-machine-driver-xhyve` might be building differently than expected.

Formula/gitlab-ci-multi-runner.rb
DomT4: I think it can be either really. I don't have a problem with it being done this way, at least.

ilovezfs:
I've noticed upstreams tend to set more of these parameters than we do and that we usually only set version to be the correct value. What would be more consistent here with other formulae would be to only set version and allow the other values to be "wrong" (i.e. imprecise defaults such as "dev" for the branch) since they generally are irrelevant to the software's functionality. In fact, most of the setting of the version value via ldflags that we do all over the place is useless except that we then go use it in the test for matching. But just running foo --version and check for 0 exit status would probably be just as good at detecting breakage without verifying the version value we ourselves set.

Formula/gitlab-ci-multi-runner.rb
ilovezfs: Mostly I prefer the other way, but here this seemed to look slightly less staccato.

DomT4: Whichever way you're happy with here, 👍 from me.

DomT4:
I think it's useful for us to set the version `-ldflags` when necessary because the information can be useful to upstream as well, in terms of problems getting reported to them, them going "What version is this" and people not thinking to check `brew info` for that information.

In terms of other values, in an ideal world we'd mimic upstream more but it's unrealistic to expect people to go hunting for all the flags, so as long as it doesn't break things to leave them out I'm not fussed.

ilovezfs:
Yup version is always a nice-to-have for that reason, and this formula probably doesn't deserve special treatment in terms of dolling out additional ldflags scraps of gruel.

PR #3789, opened 2016-08-11
exercism 2.3.0

ilovezfs:
use upstream vendor directory
fix head url

### DIFF
--- a/Formula/exercism.rb
+++ b/Formula/exercism.rb
@@ -1,11 +1,9 @@
-require "language/go"
-
 class Exercism < Formula
   desc "command-line tool to interact with exercism.io"
   homepage "http://cli.exercism.io"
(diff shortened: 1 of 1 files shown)


DomT4:
Yeeeeeeeesssss, more `go_resource`s bite the dust.

👍

ilovezfs:
@DomT4 at some point we might want to consider shipping any `go` formulae that don't use vendoring off to the boneyard as a matter of policy.

DomT4:
Go 1.7 removes support for unsetting `GO15VENDOREXPERIMENT` so that's already likely to kill off `fugu`. Come Go 1.8 `godep` workspaces are being abolished as well. They're pushing pretty hard to get people onto the vendor system, to their credit.

PR #3793, opened 2016-08-11
disco: migrate to boneyard

ilovezfs:


DomT4:
👍. You'll want to remove the `conflicts_with "disco"` line from `mono` as well, or that'll misbehave.

ilovezfs:
@apjanke any thoughts on ditching this? I recall you'd done some work on this formula at one point.

apjanke:
No thoughts. I'm not an actual `disco` user; I just came across some audit problems/upgrade needed for it.

Why the boneyarding?

apjanke:
Oh, is this because of Erlang 19 breakage? https://github.com/Homebrew/homebrew-core/pull/3761#issuecomment-239105784

Yeah, probably boneyard time until they update for the current Erlang.

Given that Erlang has ABI/compatibility changes with each major version, that might be a candidate for maintaining multiple major versions (`erlang18`, `erlang19`, `erlang20`).

ilovezfs:
>is this because of Erlang 19 breakage?

yup that's the only reason

seanjensengrey:
@ilovezfs to reference what @apjanke said, expecting Erlang apps to all use the same "current version" for some value of current (r18, r19, r20, etc) is not an expectation in the Erlang world. This will cause boneyard churn. Homebrew should most definitely have multiple major versions of Erlang.

ilovezfs:
@seanjensengrey feel free to track https://github.com/Homebrew/brew/issues/620 ;)

MikeMcQuaid:
We'll consider including multiple versions that are still actively receiving security updates and bugfixes. If upstream Erlang does not provide that: we'll still end up boneyarding projects, just not as quickly.

seanjensengrey:
These are the current Erlang versions, 19 is going to require folks to modify their code, so many systems will be on 18 for awhile. 

* https://github.com/erlang/otp/tree/OTP-19.0.5
* https://github.com/erlang/otp/tree/OTP-18.3.4.4
* https://github.com/erlang/otp/tree/OTP-17.5.6.9

ilovezfs:
I don't think this situation is that different from gcc, boost, go, ghc, ... we really only support the latest in homebrew/core.

apjanke:
Just looking at the release date history on GitHub, there are Erlang 17 and 18 releases interleaved for quite some time. And the latest Erlang 18 release, OTP-18.3.4, is only a week older than the Erlang 19 golden release. So it seems like patch releases keep coming out for quite a while.

I'll try to find some formal documentation of the Erlang support lifecycle (and [we're not the only ones wondering](http://erlang.org/pipermail/erlang-questions/2016-July/089764.html)), but it does seem to me like a good candidate for multiple major versions, more like Python 2/3 or GTK than gcc or go.

MikeMcQuaid:
Even if that were the case it'd still be a build-time dependency so wouldn't solve the issue.

PR #3801, opened 2016-08-11
asciidoc: include docs in installation

apjanke:

---

Includes the documentation as part of the asciidoc installation. The docs contain useful reference files, and the AsciiDoc user's guide refers to them.


DomT4:
Unusual to have another `make` step after `make install` but if that's the way upstream wrote it, 👍.

apjanke:
Yep, and also unusual to name it `make docs` instead off `make install-doc`, but that's indeed the way upstream did it, which is probably why we missed it up til now.

apjanke:
Merged.

PR #3820, opened 2016-08-12
miller 4.4.0

johnkerl:
Release following same procedure as always


ilovezfs:
@johnkerl thanks for opening this so promptly! By the way you may be interested in using

https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb

PR #3827, opened 2016-08-12
shairport-sync 2.8.4, 2.8.4.4 (devel)

ilovezfs:
sysconfdir instead of no config files
put pid file in var/run not prefix
add --with-metadata to avoid https://github.com/mikebrady/shairport-sync/issues/356

### DIFF
--- a/Formula/shairport-sync.rb
+++ b/Formula/shairport-sync.rb
@@ -1,15 +1,20 @@
 class ShairportSync < Formula
   desc "AirTunes emulator. Shairport Sync adds multi-room capability."
   homepage "https://github.com/mikebrady/shairport-sync"
-  url "https://github.com/mikebrady/shairport-sync/archive/2.8.3.6.tar.gz"
-  sha256 "f59c3bfc2b95af60b822a9729abd94f33330ff81090b00194cf0c999c59133b6"
-  head "https://github.com/mikebrady/shairport-sync.git"
+  url "https://github.com/mikebrady/shairport-sync/archive/2.8.4.tar.gz"
+  sha256 "80dd94c5f37b43e9b157dd5335f8edaf11109859d0144e0046c7b86fe88f6547"
+  head "https://github.com/mikebrady/shairport-sync.git", :branch => "development"
 
   bottle do
     revision 1
     sha256 "db6badb38ad81e3b8a884fb6997d45c9895134f530ba30fb0e2a1a3a979044ce" => :el_capitan
     sha256 "7b5b30ab5741b5833f3a170efee2724aa77df233741223895f107b70528d3079" => :yosemite
     sha256 "4da33ef7fb709b02bc147db5484422cf2cb7f1f40f31a11a1cb58c0be7f190d1" => :mavericks
+  end
+
+  devel do
+    url "https://github.com/mikebrady/shairport-sync/archive/2.8.4.4.tar.gz"
+    sha256 "a25e85386b3c0e32de1e01350d835e11414a42b87a136953d2d09e4ed45e1209"
   end
 
   depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


Formula/shairport-sync.rb
```diff
@@ -1,9 +1,10 @@
 class ShairportSync < Formula
   desc "AirTunes emulator. Shairport Sync adds multi-room capability."
   homepage "https://github.com/mikebrady/shairport-sync"
-  url "https://github.com/mikebrady/shairport-sync/archive/2.8.3.6.tar.gz"
-  sha256 "f59c3bfc2b95af60b822a9729abd94f33330ff81090b00194cf0c999c59133b6"
-  head "https://github.com/mikebrady/shairport-sync.git"
+  url "https://github.com/mikebrady/shairport-sync/archive/2.8.4.tar.gz"
+  sha256 "80dd94c5f37b43e9b157dd5335f8edaf11109859d0144e0046c7b86fe88f6547"
+  head "https://github.com/mikebrady/shairport-sync.git",
+    :branch => "development"
```
DomT4: Nooooooooooooooooooooooo, please don't split this onto a new line 😰

ilovezfs: my :crystal_ball: shows two-space continuations in your future

DomT4: We've already started adopting 4 spaces for this sort of thing so it lines up with the opening `"`. Generally though, I loathe `:branch` falling onto its own line 🙈.

ilovezfs: >We've already started adopting 4 spaces for this sort of thing so it lines up with the opening "

ew.

> I loathe :branch falling onto its own line

Not sure I follow the logic there. I find it much easier to miss the non-default branch when it's all the way on the right hand side, even possibly passed the eightieth column (:scream:) than when it's indented on the following line, where it's harder not to notice.

DomT4: > ew.

You agreed this wasn't an awful option in a discussion thread not two weeks ago 😆.

> Not sure I follow the logic there.

I consider it part of the URL almost, and we don't police URLs for length. I don't mind that much, I'd just like us to sort the darn spacing issue out 🙈.

Formula/shairport-sync.rb
```diff
@@ -1,9 +1,9 @@
 class ShairportSync < Formula
   desc "AirTunes emulator. Shairport Sync adds multi-room capability."
   homepage "https://github.com/mikebrady/shairport-sync"
-  url "https://github.com/mikebrady/shairport-sync/archive/2.8.3.6.tar.gz"
-  sha256 "f59c3bfc2b95af60b822a9729abd94f33330ff81090b00194cf0c999c59133b6"
-  head "https://github.com/mikebrady/shairport-sync.git"
+  url "https://github.com/mikebrady/shairport-sync/archive/2.8.4.tar.gz"
+  sha256 "80dd94c5f37b43e9b157dd5335f8edaf11109859d0144e0046c7b86fe88f6547"
+  head "https://github.com/mikebrady/shairport-sync.git", :branch => "development"
```
DomT4: Thanks! ❤️. I could probably learn to live with it if you really, really, really want to make the change. The lack of agreement amongst the team on how many spaces is an appropriate solution to multiple line URLs troubles me more than that 😄.

Formula/shairport-sync.rb
```diff
@@ -22,6 +27,12 @@ class ShairportSync < Formula
   depends_on "libdaemon"
   depends_on "libconfig"
 
+  # opened PR 12 Aug 2016: "use sysconfdir not $(DESTDIR)/etc for config files"
```
DomT4: Ooft. Thanks for filing this upstream!

ilovezfs: Our prior `--with-configfiles=no` was a bit of a red flag that something was amiss.

Surprisingly, upstream feels the need to contemplate the matter: https://github.com/mikebrady/shairport-sync/pull/355#issuecomment-239451224

Formula/shairport-sync.rb
ilovezfs: You're welcome.

>lack of agreement amongst the team on how many spaces is an appropriate solution

Might be more concerning if we all agreed since there's actually no "correct" answer here.

PR #3838, opened 2016-08-12
radare2: Added signing option

alexkornitzer:

This is nothing major and you may not accept trying to automate signing but as there are a fair few things to sign its nice to have the option to automate it.

---

Added the option to sign the build, it uses the built in commands from
the radare2 Makefile. In order for it to succeed the org.radare.radare2
certificate must exist in the keychain. The process is defined in
https://github.com/radare/radare2/blob/master/doc/osx.md

### DIFF
--- a/Formula/radare2.rb
+++ b/Formula/radare2.rb
@@ -1,3 +1,22 @@
+class CodesignRequirement < Requirement
+  include FileUtils
+  fatal true
+
+  satisfy(:build_env => false) do
+    mktemp do
+      cp "/usr/bin/false", "radare2_check"
+      quiet_system "/usr/bin/codesign", "-f", "-s", "org.radare.radare2", "--dryrun", "radare2_check"
+    end
+  end
+
+  def message
+    <<-EOS.undent
+      org.radare.radare2 identity must be available to build with automated signing.
+      See: https://github.com/radare/radare2/blob/master/doc/osx.md
+    EOS
+  end
+end
+
 class Radare2 < Formula
   desc "Reverse engineering framework"
   homepage "http://radare.org"
@@ -35,6 +54,8 @@ class Radare2 < Formula
     end
   end
 
+  option "with-code-signing", "Codesign executables to provide unprivileged process attachment"
+
   depends_on "pkg-config" => :build
   depends_on "valabind" => :build
   depends_on "swig" => :build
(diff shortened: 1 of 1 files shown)


DomT4:
What's the gain in allowing signing of `radare2`?

If we're going to do it we should do it like `llvm` does, but I'd like a strong justification here really.

AlexKornitzer:
`radare2` has debugger features similar to that of `gdb` and `lldb`.
Thus it hits the same limitations as these programs when trying to attach to processes when not run as root. So in order to attach to a process running with my user privileges I have to run `radare2` as root to succeed. If signed then this elevation of privilege is not required.

For what I grasp `radare2` does handle signing the same way as llvm and the osx.md links to https://llvm.org/svn/llvm-project/lldb/trunk/docs/code-signing.txt. After the certificate has been created once then signing can be automated.

I know that the `gdb` formula just lists this as a caveat but `radare2` has a fair few additional libraries that require signing.

AlexKornitzer:
@DomT4 any further thoughts on this?

DomT4:
I guess I could live with it if written in the same way the `llvm` formula has this sort of thing written, i.e. with the Requirement rather than inline user hacks.

AlexKornitzer:
I have just read through the `llvm` formula again and I see what you mean, but I am still not sure forcing users to sign is the correct approach as not all people that use `radare2` will need codesigning. I think that if `brew` had the ability to set a brew-wide signing certificate that is then used for all signing of formulas that require it, then it would make sense. Yet this is not a feature that exists and the inability to automate certificate creation breaks the fluid automation of installing a formula.

The guys over at radare2 have decided that signing is not mandatory and for that reason have optional `make` commands that can be called for people who wish attach without root permissions. I think that giving users the option to sign is still better/flexible (is that not the point of options?). The only inline hack  is setting the path to the login.leychain. Could I set the satisfy check up for when the signing option is set?

The only reason I am reluctant to force it is that I feel it may annoy users who now have to faff about setting up a signing certificate. I wonder if there is a simple efficient way we can do this? Ina perfect world I would like to sign the formulas that need signing with my developer cert.

DomT4:
The `llvm` mechanism isn't forced. You can discover this yourself by running `brew install llvm -s`. The only way the Requirement is triggered is if you `brew install llvm --with-lldb`. In this case you'd borrow the requirement, then do something like:

```ruby
option "with-code-signing", "Codesign executables to provide unprivileged process attachment"

depends_on CodesignRequirement if build.with? "code-signing"
```

The point of the Requirement is that if you don't have the certificate in place already it'll die before the build even starts. Otherwise `radare2` gets downloaded, then the build starts, everything seems fine, then it blows up, often with not great help messages. Requirements are better UX for this particular issue.

AlexKornitzer:
Right I see what you mean now. I have redone the commit, is it ok now?

apjanke:
Looks good to me now. And the justification for doing code signing sounds sufficient.

Any concerns, Dom?

DomT4:
It seems fine. I don't think we want to expand this sort of functionality too broadly I can live with it in this formula for the reasons given.

apjanke:

PR #3855, opened 2016-08-13
diff-so-fancy 0.11.1

DomT4:

---

### DIFF
--- a/Formula/diff-so-fancy.rb
+++ b/Formula/diff-so-fancy.rb
@@ -1,8 +1,8 @@
 class DiffSoFancy < Formula
   desc "Good-lookin' diffs with diff-highlight and more"
   homepage "https://github.com/so-fancy/diff-so-fancy"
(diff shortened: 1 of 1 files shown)


landaire:
Can Node installation be optional, and if the "manual" installation steps are taken then whatever needs to be done by the user can be printed out to the console? I don't really want to install node just for this one program that doesn't even truly depend on it.

MikeMcQuaid:
@landaire It's a built-time only dependency. If you can figure out how to make the formula build without having Node installed: we'll accept a pull request.

DomT4:
Yeah, it's also `cellar :any`, so unless you choose to build from source or you're running <10.9 _(which is less than 3 or 4% of Homebrew's user base as far as we know)_ there's no way you'll ever see the `node` dep in reality.

We switched over the installation script because it means contributors, who more often than not here are the upstream developers, aren't having to work out what goes where for Homebrew & why. It's a simpler process less likely to randomly break.

PR #3874, opened 2016-08-14
vim: don't assume all pythons are framework builds

### LINKED ISSUE #1165: vim: installation fails when using pyenv

- [x] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

### Bug reports:

This issue was originally reported to pyenv, here: https://github.com/yyuu/pyenv/issues/589

When running `brew install/upgrade vim` and pyenv, an error occurs when attempting to upgrade the package. This appears to be due to incompatible unicode formats. 

https://gist.github.com/d9f660ae9f77b8ae449335df283247be

### Steps to reproduce
Requirements:
- pyenv
- vim uninstalled

Install
```
pyenv global 2.7.11
brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/e0de5d3e7bf9976ea9232b083add47fc9106e09e/Formula/vim.rb --with-python --with-cscope --with-lua --with-ruby --with-luajit --with-perl
```

Upgrade
```
pyenv global system
brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/e0de5d3e7bf9976ea9232b083add47fc9106e09e/Formula/vim.rb --with-python --with-cscope --with-lua --with-ruby --with-luajit --with-perl
pyenv global 2.7.11
brew upgrade vim
```

Not sure what the best approach here is. Checking to see if pyenv is available seems hacky, but workable.

tdsmith:
Fixes #1165.

### DIFF
--- a/Formula/vim.rb
+++ b/Formula/vim.rb
@@ -2,8 +2,8 @@ class Vim < Formula
   desc "Vi \"workalike\" with many additional features"
   homepage "http://www.vim.org/"
   # *** Vim should be updated no more than once every 7 days ***
-  url "https://github.com/vim/vim/archive/v7.4.2196.tar.gz"
-  sha256 "8a7d38fed55e622a52a6f866d65397ae0f07dcf2aeaf53d006fc4dc4da6bb877"
+  url "https://github.com/vim/vim/archive/v7.4.2210.tar.gz"
+  sha256 "5ac86cdb4e7f0c7cdcd4a62858d07724dccb61ad727d769ee21fcc6ef7187b7b"
   head "https://github.com/vim/vim.git"
 
   bottle do


tdsmith:
Applies https://github.com/vim/vim/pull/980.

Formula/vim.rb
```diff
@@ -61,6 +61,13 @@ def install
       ENV.append "LDFLAGS", "-F#{buildpath}/Frameworks"
     end
 
+    # Fix build with flat Python builds
+    # https://github.com/Homebrew/homebrew-core/issues/1165
+    # https://github.com/vim/vim/pull/980
+    inreplace "src/auto/configure",
```
MikeMcQuaid: Presumably we can't apply https://github.com/vim/vim/pull/980.patch?

Formula/vim.rb
```diff
@@ -61,6 +61,13 @@ def install
       ENV.append "LDFLAGS", "-F#{buildpath}/Frameworks"
     end
 
+    # Fix build with flat Python builds
```
MikeMcQuaid: `flat (non-framework)` just to be more obvious?

MikeMcQuaid:
One question and a nit then 👍

tdsmith:
The PR was pulled and vim does continuous releases so we can just bump the version 🎉

MikeMcQuaid:
Oh nice 🎉

PR #3880, opened 2016-08-14
mobile-shell: improve the test

ilovezfs:
Change recommended by upstream:

>  For the test method, you could now do mosh --local 127.0.0.1 true or
> mosh --local ::1 true. Those will start a session without using ssh
> and test much more functionality.

### DIFF
--- a/Formula/mobile-shell.rb
+++ b/Formula/mobile-shell.rb
@@ -41,6 +41,7 @@ class MobileShell < Formula
 
   test do
     ENV["TERM"] = "xterm"
-    system "#{bin}/mosh-client", "-c"
+    ENV["LC_ALL"] = "en_US.UTF-8"
+    system bin/"mosh", "--local", "127.0.0.1", "true"
   end
 end


ilovezfs:
@cgull your suggested test worked fine for me locally, however in the CI environment, this happens:
```
==> /usr/local/Cellar/mobile-shell/1.2.6/bin/mosh --local 127.0.0.1 true
Server started without pseudo-terminal. Opening 80x24 terminal.

tcgetattr: Inappropriate ioctl for device
Error: mobile-shell: failed
```

cgull:
Oops, that wasn't my best most accurate suggestion ever, I forgot that this doesn't work.  If Homebrew has any methods for running smoke tests on a pty of any sort (virtual terminal logging capture/view, expect-type interactions), that'll get around this problem.

The Mosh test suite uses tmux in "control mode", but that's almost certainly more dependencies and complexity than you want for a smoke test.

PR #3888, opened 2016-08-14
dar 2.5.6

ilovezfs:
drop support for Yosemite and below due to thread_local requirement

### DIFF
--- a/Formula/dar.rb
+++ b/Formula/dar.rb
@@ -1,8 +1,8 @@
 class Dar < Formula
   desc "Backup directory tree and files"
   homepage "http://dar.linux.free.fr/doc/index.html"
-  url "https://downloads.sourceforge.net/project/dar/dar/2.4.22/dar-2.4.22.tar.gz"
-  sha256 "19a977e3ca7ed15e4406bb3e08ff2b8225d291a5c4189a6d5b7b3d75eea50f2b"
+  url "https://downloads.sourceforge.net/project/dar/dar/2.5.6/dar-2.5.6.tar.gz"
+  sha256 "47db04275d9f10f3a074e226d6fe760184e9694f1811a98f0b41e4f1dee30815"
 
   bottle do
     sha256 "7ce41e06f2f8a0d4dd65a52904250c6cd7f44223b3c438931646c4ad2ff6f97c" => :sierra
@@ -11,38 +11,37 @@ class Dar < Formula
     sha256 "a74696553c86f94dfd1ebdad29ad9c5d22c79d78fc023e43bb86f272fbfc4d87" => :mavericks
   end
 
-  option "with-docs", "build programming documentation (in particular libdar API documentation) and html man page"
+  option "with-doxygen", "build libdar API documentation and html man page"
   option "with-libgcrypt", "enable strong encryption support"
   option "with-lzo", "enable lzo compression support"
   option "with-upx", "make executables compressed at installation time"
 
-  if build.with? "docs"
-    depends_on "coreutils" => :build
-    depends_on "doxygen" => :build
-  end
+  deprecated_option "with-docs" => "with-doxygen"
 
-  depends_on "gettext" => :optional
-  depends_on "gnu-sed" => :build
+  depends_on :macos => :el_capitan # needs thread-local storage
+  depends_on "doxygen" => [:build, :optional]
+  depends_on "upx" => [:build, :optional]
   depends_on "libgcrypt" => :optional
   depends_on "lzo" => :optional
-  depends_on "upx" => [:build, :optional]
+  depends_on "xz" => :optional
+
+  needs :cxx11
 
   def install
-    ENV.prepend_path "PATH", "#{Formula["gnu-sed"].opt_libexec}/gnubin"
-    ENV.prepend_path "PATH", "#{Formula["coreutils"].opt_libexec}/gnubin" if build.with? "docs"
-    ENV.libstdcxx if ENV.compiler == :clang && MacOS.version >= :mavericks
+    ENV.cxx11
 
     args = %W[
+      --enable-mode=64
       --disable-debug
       --disable-dependency-tracking
       --disable-dar-static
       --prefix=#{prefix}
     ]
-
-    args << "--disable-build-html" if build.without? "docs"
+    args << "--disable-build-html" if build.without? "doxygen"
+    args << "--disable-upx" if build.without? "upx"
     args << "--disable-libgcrypt-linking" if build.without? "libgcrypt"
     args << "--disable-liblzo2-linking" if build.without? "lzo"
-    args << "--disable-upx" if build.without? "upx"
+    args << "--disable-libxz-linking" if build.without? "xz"
 
     system "./configure", *args
     system "make", "install"


Formula/dar.rb
```diff
@@ -10,6 +10,19 @@ class Dar < Formula
     sha256 "a74696553c86f94dfd1ebdad29ad9c5d22c79d78fc023e43bb86f272fbfc4d87" => :mavericks
   end
 
+  devel do
+    url "https://downloads.sourceforge.net/project/dar/dar/2.5.5/dar-2.5.5.tar.gz"
+    sha256 "aa77d6778206b00ad7b37d5f615aafff3ea770b35d96625429c14ff67b78466d"
+
+    depends_on "gcc" if MacOS.version <= :el_capitan
```
MikeMcQuaid: We usually rely on the `fails_with` for this instead.

Formula/dar.rb
```diff
@@ -27,11 +40,18 @@ class Dar < Formula
   depends_on "upx" => [:build, :optional]
 
   def install
+    if build.stable? && ENV.compiler == :clang && MacOS.version >= :mavericks
+      ENV.libstdcxx
+    elsif build.devel?
+      ENV.O0
+      ENV.deparallelize
```
MikeMcQuaid: Yuck. Can you report upstream?

Formula/dar.rb
ilovezfs: I agree with you but this is due to the linkage, which can cause CI issues elsewhere. CC @DomT4

Formula/dar.rb
ilovezfs: I'd also like the facts on Xcode 8's clang first (does it work? does it also need the ENV.O0?), but yeah this is quite gross and definitely merits reporting.

Also, the deparallelize may be appropriate for stable too. I haven't tried to break it.

Formula/dar.rb
DomT4: Yeah, it's a PITA with the linkage checks when removed. For example:
```
~> brew pull https://github.com/Homebrew/homebrew-core/pull/3888
<remove the depends_on line>
~> brew install dar --devel
~> brew rm gcc
~> brew missing
<blank>
~> brew linkage --test dar
Missing libraries:
  /usr/local/lib/gcc/6/libgcc_s.1.dylib
  /usr/local/opt/gcc/lib/gcc/6/libstdc++.6.dylib
```

DomT4: As far as I know that a legitimate linkage issue as well, because things compiled against the system link to `/usr/lib/libstdc++.6.dylib` for example.

Formula/dar.rb
DomT4: > does it work?

Yup!

> does it also need the ENV.O0?

Nope!

ilovezfs: Weird. Mind checking gcc and gcc5 (I need O0 for both). I'm wondering if it's because the machine is Skylake, but that didn't seem likely since setting -march core2 instead of -march native for optimization didn't help, but that doesn't rule it out definitively.

ilovezfs:
Since it worked without O0 with Xcode 8's clang, I think we should consider a minimum OS requirement of El Capitan for this formula once Xcode 8 ships, and drop both Mavericks and Yosemite at the same time.

Formula/dar.rb
DomT4: I just closed the VM down to work on something else 🙈. Unless something changed in today/yesterday's (depending on timezone) Xcode, I know `gcc` builds fine on Sierra at least.

ilovezfs: Not building gcc itself! hehe. I mean building this formula *with* gcc. If it's just a Skylake thing, we can keep support for Mavericks and Yosemite via gcc which has the requisite thread_local feature. Even if O0 is required until upstream has a fix, that should be good enough not to outright drop them, but I'm assuming it's not a big deal either way, since the Mavericks downloads were nil and I'm guessing the Yosemite numbers weren't much better.

Formula/dar.rb
ilovezfs: The linkage is only a consequence of test-bot, which  behavior which rescues the fails_with exception and chooses to `brew install gcc` as its workaround. So this isn't really legitimate unless we know we're dealing with the bottled version, which certainly isn't true (yet) for devel specs and is also not really true for stable specs with bottles either.

So this is mainly a workaround for the cases where test-bot fails to install gcc before running brew test on a gcc linked formula. That might be fixable by checking brew linkage for gcc and brew installing gcc before running brew test if gcc isn't already installed. Alternatively we could rescue the test failure and install gcc at that point if the error was a failure to run due to a missing gcc library,  then call `retry`, else call `raise` for other any other exception (which is basically the same thing test-bot is already doing for installs) but the end result should be the same either way, where the chosen medicine is preventive or curative.

MikeMcQuaid: I think it's fine as-is given this. Note if this was a library we wouldn't accept this behaviour but it's (begrudingly) fine for applications. Ideally we'd make this `if` based on the Clang version rather than the OS X version, though.

ilovezfs: this?
```
depends_on "gcc" if DevelopmentTools.clang_build_version < 800
```

MikeMcQuaid: If that doesn't 💥: yes.

ilovezfs: I think what would :boom: here would be ENV.compiler

MikeMcQuaid: :shipit:

ilovezfs: yup, DevelopmentTools.clang_build_version < 800 is fine.

ilovezfs: Interesting ... no need for ENV.O0 with gcc if I use `--env=std` or `xcrun make`. So I guess I'm reporting a superenv bug to myself not an issue to upstream at this point lol.

ilovezfs:
OK, findings are that reason is it fails with superenv is that make refurbishing slips in the `-Os`, which is what causes the breakage for gcc6. The problem is not specific to Skylake, and occurred during CI in the same way. In detail,

gcc6 + Os fails
gcc6 + O2 works
gcc5 + O2 fails
gcc5 + O0 works

so unless we're going to check the specific gcc version during install, I think setting ENV.O2 and probably adding an explicit fails_with for gcc5, plus reporting the above findings upstream, is probably the way to go.

MikeMcQuaid:
> so unless we're going to check the specific gcc version during install, I think setting ENV.O2 and probably adding an explicit fails_with for gcc5, plus reporting the above findings upstream, is probably the way to go.

👍 with comments :trollface:

ilovezfs:
bizarrely, running autoreconf prevents the need to set O2. :scream_cat:

MikeMcQuaid:
@ilovezfs :shipit:?

ilovezfs:
@MikeMcQuaid no, I think now that Sierra is out, this should be recast as a full replacement for the 2.4.x stable spec, and support dropped for Yosemite and below. 2.4 and 2.5 are both considered stable by upstream.

MikeMcQuaid:
@ilovezfs 👍

ilovezfs:
@MikeMcQuaid PR refreshed.

MikeMcQuaid:
👍 :shipit:

PR #3890, opened 2016-08-14
gpg: additional changes

DomT4:

---

A swing at some things raised in https://github.com/Homebrew/brew/issues/708.

### DIFF
--- a/Formula/gnupg.rb
+++ b/Formula/gnupg.rb
@@ -42,15 +42,20 @@ class Gnupg < Formula
   end
 
   def caveats; <<-EOS.undent
-    All commands have been installed with the suffix '1'.
+    This formula does not install either `gpg` or `gpgv` executables into
+    into the PATH.
 
-    If you really need to use these commands with their normal names, you
-    can add a "gpgbin" directory to your PATH from your #{shell_profile} like:
+    If you really need to use these tools without the "1" suffix you can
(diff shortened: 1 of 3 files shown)


Formula/gnupg.rb
```diff
@@ -42,7 +42,8 @@ def install
   end
 
   def caveats; <<-EOS.undent
-    All commands have been installed with the suffix '1'.
+    This formula does not install either `gpg` or `gpgv` vanilla executables
+    into the PATH. All executables have been installed with a "1" suffix.
```
tdsmith: I might avoid "vanilla" since it's kinda idiomatic. Maybe something like:

> This formula installs tools named gpg1 and gpgv1. If you need to use these tools without the "1" suffix, you may...

Exciting changes afoot!

DomT4: 👍 on vanilla, will reword this blob a bit.

> Exciting changes afoot!

Hopefully. At this point we're diving into parts that weren't working entirely as expected when I was trying to blob everything together in one big PR, so I'm kind of frantically running around trying to find the root cause of why that was happening now there's less changes ongoing around it 🙈.

DomT4: Tweaked the wording. Let me know if it reads a little better to you.

MikeMcQuaid:
LGTM 👍 🚢

PR #3924, opened 2016-08-16
go 1.7

DomT4:

---


ilovezfs:
godep needs a revision bump. other go build tools may need one too.

DomT4:
`godep` seems to be working fine locally from the old bottle. Is there an issue I'm not remembering here? May well be, to be fair.

DomT4:
Oh, I see. Found the prior discussion. Let's bump anyway, they're tiny tools & it won't hurt.

Formula/go.rb
```diff
@@ -3,27 +3,16 @@ class Go < Formula
...
-      version "1.6.2"
-      sha256 "6ebbafcac53bbbf8c4105fa84b63cca3d6ce04370f5a04ac2ac065782397fc26"
-    end
+    url "https://storage.googleapis.com/golang/go1.7.src.tar.gz"
+    mirror "https://fossies.org/linux/misc/go1.7.src.tar.gz"
+    version "1.7"
+    sha256 "72680c16ba0891fcf2ccf46d0f809e4ecf47bbf889f5d884ccb54c5e9a17e1c0"
 
-    go_version = "1.6"
+    go_version = "1.7"
```
ilovezfs: would prefer go_version = version.to_s.split(".")[0..1].join(".") if the goal is to use "1.7" for "1.7", "1.7.1", "1.7.2", ...

Formula/go.rb
```diff
@@ -101,7 +67,7 @@ def install
       system "#{bin}/go", "install", "-race", "std"
     end
 
-    if build.with?("godoc") || build.with?("vet")
+    if build.with?("godoc")
```
ilovezfs: I think you can drop the parentheses now, no?

DomT4: Yeah. Not my change here, I overlooked it.

Formula/go.rb
DomT4: Meh, let's not sex it up unless it proves necessary.

DomT4:
Lord am I glad I was anal enough to run around fixing these the other week:

> /usr/local/Cellar/go/1.7/libexec/pkg/tool/darwin_amd64/link: -X flag requires argument of the form importpath.name=value

ilovezfs:
indeed.

ilovezfs:
was fugu fail reported upstream?

Formula/go.rb
ilovezfs: already use that elsewhere. would prefer not to update this every time.

DomT4:
Not yet, but have been watching it for a while. No activity in a year, and the last stable version of Go they tested against was Go _1.4_. We're not going to get a quick resolution there, but I will file it.

Formula/go.rb
DomT4: Alright, I'll do it just for you. Don't say I'm never nice.

ilovezfs: thanks. we can kill it next time if it proves to be wrong.

Formula/godep.rb
```diff
@@ -23,16 +23,20 @@ def install
...
+            "ImportPath": "go.googlesource.com/tools",
+            "Rev": "3fe2afc9e626f32e91aff6eddb78b14743446865"
+          }
+        ]
       }
     EOS
-    (testpath/"src/foo/bar/Godeps").install "Godeps.json"
-    cd("src/foo/bar") { system bin/"godep", "path" }
+    system bin/"godep", "restore"
+    assert File.exist?("src/go.googlesource.com/tools/README")
```
ilovezfs: I think we've been adding a message following these assert exists

Formula/go.rb
```diff
@@ -34,51 +23,28 @@ class Go < Formula
...
-    end
-
     resource "gotools" do
       url "https://go.googlesource.com/tools.git"
     end
   end
 
   option "without-cgo", "Build without cgo"
   option "without-godoc", "godoc will not be installed for you"
-  option "without-vet", "vet will not be installed for you"
```
ilovezfs: I think removing this is going to force source builds for everyone who currently has `without-vet` in their INSTALL_RECEIPT but that's not your fault.

Formula/go.rb
```diff
@@ -34,51 +23,28 @@ class Go < Formula
...
   option "without-race", "Build without race detector"
 
   depends_on :macos => :mountain_lion
 
+  # Should use the last stable binary release to bootstrap.
+  # More explicitly, leave this at 1.7 when 1.7.1 is released.
```
ilovezfs: Is there a citation for this? Insisting on stage 1 not stage 2 bootstraps sounds really weird to me.

ilovezfs:
Which formulae has this PR been tested with already?

Formula/go.rb
DomT4: Not sure, but yeah, can't be avoided here 😢

Formula/go.rb
DomT4: Can you elaborate on which part you want a citation for?

Formula/godep.rb
DomT4: Trying to get this done fairly quickly; also working on GPG stuff in the background which is more important. Just wanted to push this through ASAP so I didn't wake up to 450 thumbs up in the other thread later.

ilovezfs:
Does this break gox? I'm wondering if the formulae that bootstrap gox first from go_resource blocks may need updates to a newer gox revision.

Formula/go.rb
ilovezfs: why would using 1.7.1 for bootstrap be wrong (for next update of course not this one since 1.7.1 doesn't exist yet)

DomT4: We had this discussion recently 😉. It enables us to bootstrap with a build we know with a high degree of certainty works, so if there's build errors we know the errors are likely with the new release. Also gets people off our back around wasting their lives by building from source blah blah blah.

ilovezfs:
Anyway, regardless of comments, this looks great. I'd be in favor of shipping as soon as you have :green_apple:

DomT4:
Typically there have never been compatibility problems with `gox` on major updates. This seems to be the case again this time from a cursory check.

Formula/go.rb
ilovezfs: does upstream not do stage 2 builds?

ilovezfs:
Yeah, I'll probably trigger a rebuild of all go everything locally after this ships but I'm expecting very little breakage if any.

DomT4:
I did a partial rebuild locally & it didn't blow up. I think Homebrew's exposure to 1.7 breakage is fairly low, compared particularly to previous major releases. Always happy to be unpleasantly surprised though 🙈.

DomT4:
Shipped!

ilovezfs:
hurray! :tada:

PR #3933, opened 2016-08-16
mg 20160815

mkhl:


MikeMcQuaid:
Fails to build on 10.9: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/6365/version=mavericks/testReport/junit/brew-test-bot/mavericks/install_mg/

mkhl:
Ah, right. Current upstream uses the `fstatat` syscall which doesn’t seem to be available prior to 10.10, see also https://github.com/ibara/mg/pull/12#issuecomment-239393159

I don’t see an easy way to fix this so I’ll close this PR.

Sorry for the noise. Suggestions welcome.

MikeMcQuaid:
> Ah, right. Current upstream uses the fstatat syscall which doesn’t seem to be available prior to 10.10, see also ibara/mg#12 (comment)
> 
> I don’t see an easy way to fix this so I’ll close this PR.

Could add a `depends_on :macos => :yosemite` with a comment explaining this.

mkhl:
@BrewTestBot test this please

mkhl:
@MikeMcQuaid I see. Thanks!

MikeMcQuaid:
> @BrewTestBot test this please

As you're not a maintainer this will have no effect. When you force-push it'll rebuild, though.

mkhl:
Got it

MikeMcQuaid:
Thanks again @mkhl!

PR #3943, opened 2016-08-16
gradle: correct java dependency to 1.7

kunickiaj:
The last commit to update to Gradle 3.0 did not include an update to the required Java version as Java 6 is no longer supported for Gradle 3.x.


MikeMcQuaid:
Thanks again @kunickiaj!

PR #3947, opened 2016-08-16
adr-tools 1.2.0 (new formula)

jhulten:

---

Add adr-tools from https://github.com/npryce/adr-tools/ to support
managing architecture decision records.

### DIFF
--- a/Formula/adr-tools.rb
+++ b/Formula/adr-tools.rb
@@ -19,12 +19,13 @@ class AdrTools < Formula
       s.gsub! "cmd=$(dirname $0)/adr-$1", "cmd=$adr_bin_dir/adr-$1"
     end
 
-    prefix.install Dir["src/*.md"]
-    prefix.install Dir["src/config.sh"]
-    bin.install Dir["src/*"]
+    prefix.install Dir["src/*.md"]      # Places the base templates in `prefix`
+    prefix.install Dir["src/config.sh"] # Places the config in `prefix`
+    bin.install Dir["src/*"]            # Copies the rest of the commands to `prefix/bin`
   end
 
   test do
-    system "true"
+    assert_match(/Run 'adr help COMMAND' for help on a specific command./,
+                 pipe_output("#{bin}/adr help"))
   end
 end
--- a/Formula/adr-tools.rb
+++ b/Formula/adr-tools.rb
@@ -21,7 +21,7 @@ class AdrTools < Formula
 
     prefix.install Dir["src/*.md"]
     prefix.install Dir["src/config.sh"]
-    bin.install Dir["src/*"] - Dir["src/*.md"] - Dir["src/config.sh"]
+    bin.install Dir["src/*"]
   end
 
   test do
--- a/Formula/adr-tools.rb
+++ b/Formula/adr-tools.rb
@@ -19,9 +19,9 @@ class AdrTools < Formula
       s.gsub! "cmd=$(dirname $0)/adr-$1", "cmd=$adr_bin_dir/adr-$1"
     end
 
-    prefix.install Dir["src/*.md"]      # Places the base templates in `prefix`
-    prefix.install Dir["src/config.sh"] # Places the config in `prefix`
-    bin.install Dir["src/*"]            # Copies the rest of the commands to `prefix/bin`
+    prefix.install Dir["src/*.md"] # Places the base templates in `prefix`
+    prefix.install "src/config.sh" # Places the config in `prefix`
+    bin.install Dir["src/*"]       # Copies the rest of the commands to `prefix/bin`
   end
 
   test do
--- a/Formula/adr-tools.rb
+++ b/Formula/adr-tools.rb
@@ -1,0 +1,30 @@
+class AdrTools < Formula
+  desc "CLI tool for working with Architecture Decision Records."
+  homepage "https://github.com/npryce/adr-tools"
+  url "https://github.com/npryce/adr-tools/archive/1.2.0.tar.gz"
+  sha256 "3f45646c099ae996b97c855a9a27ed540e076580eba59c2e8613453b7a67e412"
+  head "https://github.com/npryce/adr-tools.git"
+
+  def install
+    inreplace "src/config.sh" do |s|
+      s.gsub! 'adr_bin_dir="$(dirname $0)"', "adr_bin_dir=\"#{bin}\""
+      s.gsub! 'adr_template_dir="$adr_bin_dir"', "adr_template_dir=\"#{prefix}\""
+    end
+
+    inreplace Dir["src/adr-*"], 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+    inreplace Dir["src/_adr_*"], 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+
+    inreplace "src/adr" do |s|
+      s.gsub! 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+      s.gsub! "cmd=$(dirname $0)/adr-$1", "cmd=$adr_bin_dir/adr-$1"
+    end
+
+    prefix.install Dir["src/*.md"]
+    prefix.install Dir["src/config.sh"]
+    bin.install Dir["src/*"] - Dir["src/*.md"] - Dir["src/config.sh"]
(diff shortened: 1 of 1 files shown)


Formula/adr-tools.rb
```diff
@@ -0,0 +1,30 @@
...
+      s.gsub! 'adr_template_dir="$adr_bin_dir"', "adr_template_dir=\"#{prefix}\""
+    end
+
+    inreplace Dir["src/adr-*"], 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+    inreplace Dir["src/_adr_*"], 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+
+    inreplace "src/adr" do |s|
+      s.gsub! 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+      s.gsub! "cmd=$(dirname $0)/adr-$1", "cmd=$adr_bin_dir/adr-$1"
+    end
```
MikeMcQuaid: Can you please submit these patches to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/adr-tools.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    prefix.install Dir["src/*.md"]
+    prefix.install Dir["src/config.sh"]
+    bin.install Dir["src/*"] - Dir["src/*.md"] - Dir["src/config.sh"]
```
MikeMcQuaid: If you've `.install`ed above those files will no longer exist.

Formula/adr-tools.rb
```diff
@@ -0,0 +1,30 @@
...
+    inreplace Dir["src/adr-*"], 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+    inreplace Dir["src/_adr_*"], 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+
+    inreplace "src/adr" do |s|
+      s.gsub! 'source "$(dirname $0)/config.sh"', "source '#{prefix}/config.sh'"
+      s.gsub! "cmd=$(dirname $0)/adr-$1", "cmd=$adr_bin_dir/adr-$1"
+    end
+
+    prefix.install Dir["src/*.md"]
+    prefix.install Dir["src/config.sh"]
```
MikeMcQuaid: Should this be in `etc`?

Formula/adr-tools.rb
```diff
@@ -0,0 +1,30 @@
...
+    bin.install Dir["src/*"] - Dir["src/*.md"] - Dir["src/config.sh"]
+  end
+
+  test do
+    system "true"
```
MikeMcQuaid: This needs a test.

Formula/adr-tools.rb
MikeMcQuaid: Also: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/6397/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_adr_tools___new_formula/

Formula/adr-tools.rb
jhulten: So line 19 is viably an upstream fix. The developer currently says 'check out the directory and add it to your path' but I can push for better layout and proper install if that is needed.

MikeMcQuaid: We need to figure out a long-term way to get this working without patching. Another option would be to use symlinks or `bin.write_exec_script`.

dunn:
Any update here?

Formula/adr-tools.rb
jhulten: Unless I `gsub!` the destinations as well, the code looks for `config.sh`. This isn't ideal, but again it comes down to how much change @npryce is willing to take on his end (assuming I ship some PRs his way).

Formula/adr-tools.rb
MikeMcQuaid: Nudge on this.

jhulten: RE: L18

How would you approach this, @MikeMcQuaid? While the original developer expected `config.sh` to exist in the same directory as the adr commands, this seems like a *terrible* idea unless I rename it and then I have a different set of gsubs.

MikeMcQuaid: > While the original developer expected config.sh to exist in the same directory as the adr commands

That's fine; install everything to `libexec` and use `bin.write_exec_script Dir["#{libexec}/bin/*"]`

MikeMcQuaid: @jhulten Any news?

npryce:
Hi. I'm the developer of adr-tools. Happy to change the way configuration works. I added config.sh to try and avoid the need to edit the source like this when packaging the app. But it's obviously not a good mechanism. Please raise issue(s) on the adr-tools project.

ilovezfs:
@BrewTestBot test this please

MikeMcQuaid:
Passing on this, I'm afraid, until upstream are able to provide a more flexible configuration mechanism or `bin.write_exec_script Dir["#{libexec}/bin/*"]` is working on a new PR. Sorry and thanks anyway!

npryce:
If someone tells me what you need, I can implement it in the upstream repo.  But so far I've only received a link to this thread, and no request for any changes.

MikeMcQuaid:
@npryce My apologies, I wasn't expecting you to do anything but @jhulten or another user to raise the issue. My take is removing our `inreplace` calls and providing similar support to them with a `configure` script or similar. Thanks!

jhulten:
And I have been hip deep in preparing a house to move into, so my non-work geek time is limited. 

@npryce The issue I see is that the config file is sourced and expected to be in the same directory as the executables. If that was converted to a command `adr-env` and loaded defaults, plus overrides and presented values as environment variables (ala `docker-machine env` or whatnot) that would allow the other components to `eval $(adr-env)` instead of `source "$(dirname $0)/config.sh"`.

I will revisit this when I have mental space on the brew packaging end. No need to reopen.

npryce:
I've added a command `adr-config` that outputs what was the contents of config.sh, and made the adr scripts eval the output.  This will, I hope, let packagers configure things by writing their own adr-config scripts that output variables that reflect how they have packaged the application.

Let me know if that works or if it needs any changes.

MikeMcQuaid:
@npryce Thanks. That's an improvement but ideally there'd be a `configure`-style script on your end which does the current replacements for us (even if you're literally just calling `sed` with the same stuff). Our main reasoning for that is that you know your own project better than we ever will and if you're making some sweeping changes in a future version that require changing or tweaking those `sed` lines then you'll be in a better position to do so than a random Homebrew contributor who just wants the latest version of your stuff. Thanks for being a fanasticly responsive upstream maintainer regardless, though! 🎉

npryce:
I'm not sure I understand what you want.  A configure script that changes the variable definitions in adr-config?  Or a configure script _instead_ of adr-config that rewrites all the source to set variables in them?  

The latter sounds too risky. 

 The former doesn't sound it would help -- if I had to change the structure of the program(s), I'd have to change the variables, which means a configure script would have different command-line parameters, and you'd have to change your packaging scripts.  Instead, I expect packagers to have their own adr-config script and replace the one in the source when building the package.

MikeMcQuaid:
@npryce I want to be able to do `./configure --prefix=#{prefix}` or a Makefile or CMake equivalent and then `make install`. I don't want us to have to supply our own scripts or (as currently) perform replacements of other files, I'd like us to be able to specify an installation prefix and that is all. Note that that's the most standard way of installing most built-from-source software. How that's actually implemented under the hood I have no real preference.

kevinjalbert:
Looking at the recent [ADR](https://github.com/npryce/adr-tools/blob/bd5c858c8bf33dde90d41d30d2e36528881d2762/doc/adr/0007-invoke-adr-config-executable-to-get-configuration.md) it seems like the config issue _should_ be resolved as of [2.0.0](https://github.com/npryce/adr-tools/releases/tag/2.0.0)?

> Added a mechanism for packagers to configure the application to match the installation conventions of the target platform

Is that the case? Thus allowing movement on the inclusion of it in homebrew?

npryce:
It could be packaged for Homebrew (and other software repositories) already. The change you refer to made it easier to do so.

PR #3963, opened 2016-08-17
libjson-rpc-cpp 0.7.0

ilovezfs:


ilovezfs:
@MikeMcQuaid audit insists on the revision 1, but I guess that's correct since the devel spec is being promoted without change? Should I make it like 7.0.0.1 instead or something?

MikeMcQuaid:
Either leave `revision 1` or ignore the `audit` in this case.

ilovezfs:
OK, I'll kill it and ignore the audit.

DomT4:
Can you file an Issue/PR to highlight/fix the audit issue so we don't forget this is a problem as well? 🙈. Otherwise, 👍 to ignoring this for now.

ilovezfs:
Is it actually a bug, though? It's getting named `formula_revision` not `spec_revision`.

DomT4:
It wasn't an issue in the past, and it's not desired behaviour, so I'd consider it a bug personally.

PR #3969, opened 2016-08-17
flake8 3.0.4

sigmavirus24:

---

### DIFF
--- a/Formula/flake8.rb
+++ b/Formula/flake8.rb
@@ -1,8 +1,10 @@
 class Flake8 < Formula
+  include Language::Python::Virtualenv
+
   desc "Lint your Python code for style and logical errors"
   homepage "http://flake8.pycqa.org/"
-  url "https://files.pythonhosted.org/packages/53/0a/b2c28a77dfc508ed9f7334252311e1aaf8f0ceaaeb1a8f15fa4ba3e2d847/flake8-2.6.2.tar.gz"
-  sha256 "231cd86194aaec4bdfaa553ae1a1cd9b7b4558332fbc10136c044940d587a778"
+  url "https://files.pythonhosted.org/packages/b0/56/48727b2a6c92b7e632180cf2c1411a0de7cf4f636b4f844c6c46f7edc86b/flake8-3.0.4.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/flake8.rb
```diff
@@ -30,15 +30,33 @@ class Flake8 < Formula
...
-    %w[pycodestyle pyflakes mccabe].each do |dependency|
+    %w[pycodestyle pyflakes mccabe enum34].each do |dependency|
       resource(dependency).stage do
         system "python", *Language::Python.setup_install_args(libexec/"vendor")
       end
     end
 
+    resource("configparser").stage do
+      # NOTE(sigmavirus24): configparser uses namespace packages which do not
+      # play well with --single-version-externally-managed
```
MikeMcQuaid: CC @tdsmith for thoughts on this.

sigmavirus24: So to add a bit more context, I tried this out at first by having configparser as part of the array above and kept seeing import errors. Only by breaking it out did it work. I think this is because brew (correctly) installs dependencies in a subdirectory by convention. Removing `--single-version-externally-managed` made this work better. I have no good way of explaining it =/ I barely understand this myself.

tdsmith: I actually don't know what easy_install is doing that allows this to work but it's probably bad. :D The way namespace packages are handled with --single-version-externally-managed is to write a special .pth file into site-packages, which unfortunately doesn't help when the installation prefix isn't a blessed sitedir from which .pth files are read. The workaround we've had is to touch an __init__.py in the namespace directory.

But! It's probably better to convert flake8 to use the new virtualenv encapsulation where namespace packages just work; basically you should add `include Language::Python::Virtualenv` at the very top of the class definition and replace the install method with

```ruby
def install
end
```

sigmavirus24: Awesome @tdsmith! I'll tackle that later this week probably. Thanks for the pointer!

sigmavirus24: Tested this out and it works like a charm. ✨ Thank you both ✨ I've updated the PR with this approach.

PR #3971, opened 2016-08-17
tbb: add optional python library

quantumsteve:

---

This adds an option to build TBB's new python library that monkey-patches multiprocessing.Pool.

https://software.intel.com/en-us/blogs/2016/04/04/unleash-parallel-performance-of-python-programs

### DIFF
--- a/Formula/tbb.rb
+++ b/Formula/tbb.rb
@@ -17,6 +17,8 @@ class Tbb < Formula
   # requires malloc features first introduced in Lion
   # https://github.com/Homebrew/homebrew/issues/32274
   depends_on :macos => :lion
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "swig" => :build
 
   def install
     # Intel sets varying O levels on each compile command.
@@ -32,6 +34,11 @@ class Tbb < Formula
     system "make", *args
     lib.install Dir["build/BUILDPREFIX_release/*.dylib"]
     include.install "include/tbb"
+
+    cd "python" do
+      ENV["TBBROOT"] = prefix
+      system "python", *Language::Python.setup_install_args(prefix)
+    end
   end
 
   test do


Formula/tbb.rb
```diff
@@ -16,7 +16,12 @@ class Tbb < Formula
 
   # requires malloc features first introduced in Lion
   # https://github.com/Homebrew/homebrew/issues/32274
-  depends_on :macos => :lion
+  depends_on macos: :lion
```
MikeMcQuaid: Just leave this as it was for now, thanks.

Formula/tbb.rb
```diff
@@ -16,7 +16,12 @@ class Tbb < Formula
...
+  depends_on python: :optional
+
+  if build.with? "python"
+    depends_on "swig" => :build
```
MikeMcQuaid: `depends_on "swig" => :build if build.with? "python"`

Formula/tbb.rb
```diff
@@ -17,6 +17,8 @@ class Tbb < Formula
   # requires malloc features first introduced in Lion
   # https://github.com/Homebrew/homebrew/issues/32274
   depends_on :macos => :lion
+  depends_on python: :optional
```
DomT4: I'd prefer we keep using the hashrocket until we make this a broader change.

MikeMcQuaid: @DomT4 Reinstate the Rubocop (that I suspect is suggesting people change this)?

DomT4: Ah, yes, I suspect so. Will double check that locally.

MikeMcQuaid: @DomT4 https://github.com/Homebrew/brew/pull/736.

quantumsteve:
After updating I noticed that `--with-python` is now generating a [sandbox error](https://gist.github.com/quantumsteve/bf19364cf35201f78c388f787ecfaa1d). How do you recommend addressing this?

tdsmith:
This is a tbb bug; it shouldn't write files outside of its install prefix. I would probably:
- always build this instead of adding an option
- not provide the `python` target to `make`
- install the python module manually

The latter should look something like:
```ruby
cd "python" do
  system "python", *Language::Python.setup_install_args(prefix)
end
```

Formula/tbb.rb
```diff
@@ -17,6 +17,8 @@ class Tbb < Formula
   # requires malloc features first introduced in Lion
   # https://github.com/Homebrew/homebrew/issues/32274
   depends_on :macos => :lion
+  depends_on :python
```
DomT4: This should probably be `depends_on :python if MacOS.version <= :snow_leopard`

Formula/tbb.rb
```diff
@@ -32,6 +34,11 @@ def install
     system "make", *args
     lib.install Dir["build/BUILDPREFIX_release/*.dylib"]
     include.install "include/tbb"
+
+    cd "python" do
+      ENV["TBBROOT"]=prefix
```
DomT4: Nit but please leave spacing in formulae usage of this sort of thing:
```ruby
ENV["TBBROOT"] = prefix
```

tdsmith:
Awesome :tada: Thanks for this!

PR #3972, opened 2016-08-17
Added symlink to plantuml.jar for compatibility

SommerEngineering:

---

I added a symlink for this existing formula for plantuml. Because some tools needed a plantuml.jar file at the path.

### DIFF
--- a/Formula/plantuml.rb
+++ b/Formula/plantuml.rb
@@ -15,9 +15,12 @@ class Plantuml < Formula
       #!/bin/bash
       GRAPHVIZ_DOT="#{Formula["graphviz"].opt_bin}/dot" exec java -jar #{prefix}/#{jar} "$@"
     EOS
+    system "ln" " -s" " #{prefix}/#{jar}" " #{prefix}/bin/plantuml.jar"
+    system "chmod" " +x" " #{prefix}/bin/plantuml.jar"
   end
 
   test do
     system bin/"plantuml", "-testdot"
+    system "java" "-jar" "bin/plantuml.jar" "-testdot"
   end
 end


MikeMcQuaid:
> Because some tools needed a plantuml.jar file at the path.

Which tools?

Formula/plantuml.rb
```diff
@@ -15,9 +15,12 @@ def install
       #!/bin/bash
       GRAPHVIZ_DOT="#{Formula["graphviz"].opt_bin}/dot" exec java -jar #{prefix}/#{jar} "$@"
     EOS
+    system "ln" " -s" " #{prefix}/#{jar}" " #{prefix}/bin/plantuml.jar"
+    system "chmod" " +x" " #{prefix}/bin/plantuml.jar"
```
MikeMcQuaid: We don't support `.jar` files in `bin`.

SommerEngineering:
For example the PlantUMLBuilder: https://github.com/SommerEngineering/PlantUMLBuilder

SommerEngineering:
@MikeMcQuaid: Thanks for your review. Physical, there is no `jar` file at the `bin` directory -- it is just the link. Is this also not allowed? How could I modify the install script in order to add the link to the path? Because if I create the link at the prefix directory, it is no longer available through the path...

MikeMcQuaid:
> Is this also not allowed?

It is not. Instead, consider modifying PlantUMLBuilder to use the correct executable name. Thanks!

SommerEngineering:
Thanks @MikeMcQuaid for the fast feedback. Okay, I have done the change within my tool. I do not know if any other tool need a `plantuml.jar` file. Therefore, I close this request.

PR #3976, opened 2016-08-17
node-build 2.3.5

jasonkarns:


DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/a3da2d10f6e13cf56e5904f0f0229c14e78acc5e. Thank you for another contribution to Homebrew @jasonkarns! 😺

jasonkarns:
@DomT4 quick question... I use a script for auto-bumping the formula for nodenv and node-build (and a few other nodenv related tools in our nodenv tap).

The script just broke this morning because the core formula were moved from `Library/Formula` into `Library/Taps/homebrew/homebrew-core`. I've fixed the script, but I noticed that of all the taps, homebrew-core is the _only_ one that doesn't put the formula directly in the root of the tap repo. (they're under `Formula/` in the tap)

Are there plans to expect taps to follow this pattern and begin placing their formula under a `Formula/` directory in the root of the tap? Or is this only for homebrew-core (and is there any chance I could persuade the homebrew team to make homebrew-core follow the same directory layout as regular taps?)

Sorry for the off-topic question. Wasn't sure if I should open an issue about this.

MikeMcQuaid:
> I use a script for auto-bumping the formula for nodenv and node-build (and a few other nodenv related tools in our nodenv tap

I'd advise you to use https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb

> Are there plans to expect taps to follow this pattern and begin placing their formula under a Formula/ directory in the root of the tap? Or is this only for homebrew-core (and is there any chance I could persuade the homebrew team to make homebrew-core follow the same directory layout as regular taps?)

Putting large number of formulae into the tap root is a bad idea for GitHub's interface to remain usable. I'd recommend everyone uses the `Formula` layout but we can't mandate it, unfortunately. I'd defer on such things to scripts like those above to tell you where files are.

jasonkarns:
@MikeMcQuaid great, thanks!!

DomT4:
Eventually we're planning to move everything under the Homebrew umbrella to the formulae live inside a `Formula` folder, but at the moment it's known to cause some edge-case issues so that's why there's the difference.

MikeMcQuaid:
> Eventually we're planning to move everything under the Homebrew org umbrella to the formulae live inside a Formula folder layout, but at the moment switching layouts is known to cause some edge-case issues so that's why there's the difference.

Although we might hold off on that in e.g. moving homebrew-core to having subdirectories by e.g. first letter to avoid GitHub/Git getting upset with large numbers of files in a single dir.

DomT4:
👍, no immediate plans, but certainly we could improve our handling here overall.

jasonkarns:
awesome. last question, as i'm looking into the bump-formula-pr script... 

What do I need to do to be able to use it? Doesn't seem to be a formula of its own to install, nor is it already in my PATH from homebrew. And it failed as a brew subcommand as well:

```
$ brew info bump-formula-pr
Error: No available formula with the name "bump-formula-pr" 

$ brew bump-formula-pr
Error: Unknown command: bump-formula-pr

$ brew-bump-formula-pr
-bash: brew-bump-formula-pr: command not found
```

MikeMcQuaid:
`HOMEBREW_DEVELOPER=1 brew bump-formula-pr --help` should be useful. We could/should perhaps consider making it a non-developer command CC @ilovezfs for thoughts.

jasonkarns:
ah, magic env var. works! thanks again.

ilovezfs:
@MikeMcQuaid I'm not sure people will know what to do if we request *any* changes, but I guess we could let them make the changes in the GitHub GUI and then have whoever does the `brew pull --bottle` for the PR also do the rebase/squash on the contributor's behalf.

Also, I know @DomT4 has expressed concerns about automated PRs and people not being responsive to requests for changes, questions, etc. but some percent of the time the stars should align and the PR will be :green_apple: upon submission.

An alternative might be to revert https://github.com/Homebrew/legacy-homebrew/pull/48322 so that setting HOMEBREW_DEVELOPER only temporarily is less cumbersome, since there now seem to be quite a few use cases for that option that didn't exist at the time of the original PR.

DomT4:
> then have whoever does the brew pull --bottle for the PR also do the rebase/squash on the contributor's behalf.

I'll do it for new contributors, but at this point I more or less expect regular contributors to be used to Homebrew's squash-as-much-as-possible workflow.

> Also, I know @DomT4 has expressed concerns about automated PRs and people not being responsive to requests for changes, questions, etc.

To be fair, this hasn't come to fruition yet. I'd be fine with allowing `brew bump-formula-pr` to become more visible & seeing how it goes.

My concerns linger around people switching over to that mechanism & us missing valuable opportunities to fix other things, whether style issues, missing resource updates or migrating to new systems like the `virtualenv` one for example, which you've seen from your own usage, but some of those issues can be resolved over time.

Eventually we're likely to employ `brew test-bot` to fix style problems automatically on the fly.

jasonkarns:
Only speaking for myself, I used my own script for automated PRs (originally written by @mislav). I maintain nodenv/node-build (and related plugins) and we do version bumps at least as frequently as node itself drops releases (which is nearly weekly some months). So automating the PR itself is crucial. Though I would like to say that I would be responsive to any requests on the PRs, despite the PR itself being automated. Though, I expect to feel more confident in my homebrew PRs once we're using bump-formula-pr instead of our homegrown script

And I don't personally consider `HOMEBREW_DEVELOPER` any more or less cumbersome than `--homebrew-developer`. Only less discoverable.

jasonkarns:
Another benefit of seeing this bump-formula-pr script was that I am now (newly) aware of a formula's ability to point to a git tag+sha instead of url+checksum. I expect this to be preferable for github-hosted tools and certainly simplify our release process.

ilovezfs:
>I expect this to be preferable for github-hosted tools and certainly simplify our release process.

We generally only allow it when it's *required*, which is typically only when the repo has submodules or the build system itself has a bug that causes it to fail with the source archive tarball, in which case that issue must be reported upstream so that it can be reversed as soon as that happens.

DomT4:
> I expect this to be preferable for github-hosted tools and certainly simplify our release process.

We extremely strongly prefer using the url + checksum method unless a git clone is needed for an explicit reason, i.e. git submodules. URL/Checksum is less resource intensive, on GitHub as well as us IIRC.

ilovezfs:
@DomT4 jinx.

DomT4:
@ilovezfs Darn. We're going to have to appear in the same place at some point so people don't think we're the same person you know 😄.

ilovezfs:
>I'll do it for new contributors, but at this point I more or less expect regular contributors to be used to Homebrew's squash-as-much-as-possible workflow.

We could add `--squash` to `brew pull` ...

jasonkarns:
> We extremely strongly prefer using the url + checksum method

ACK. i'll leave ours as-is then

DomT4:
> We could add --squash to brew pull ...

I don't know, I'm not particularly keen on introducing a system where maintainers are expected to essentially finish up PRs for users. I don't think squashing is hostile request, certainly any more than our mountain of audit checks.

MikeMcQuaid:
@ilovezfs Thoughts on making this a non-developer command? e.g. `brew edit` and friends already are. Maybe we should make `HOMEBREW_DEVELOPER` be set for all of them? It feels like we should decide whether the formula update workflow should require `HOMEBREW_DEVELOPER` to be set or not. Arguably we should print a nice message like `HOMEBREW_DEVELOPER required for this command` rather than pretending it doesn't exist, too...

ilovezfs:
@MikeMcQuaid my gut reaction is to open up virtually all the commands to everyone and not hide them behind the HOMEBREW_DEVELOPER wall and to change the default update behavior back to always popping the stash, possibly with an opt-in for not popping the stash, or just drop that code path entirely since I think it does more harm than good in terms of support burden and user experience.

MikeMcQuaid:
@ilovezfs It has eliminated an annoying class of issues, though, where people who don't understand Git end up in weird states that they don't know how to resolve. Interesting in @DomT4's perception of this too, though.

MikeMcQuaid:
@ilovezfs Or alternatively we set `HOMEBREW_DEVELOPER` implicitly at the Git level where if you've ever run e.g. `brew edit` or similar commands it's something we toggle on for you.

ilovezfs:
>It has eliminated an annoying class of issues

Yeah, I'm not denying it does some good. I just think it's outweighed by the bad it's doing.

I guess my stance is that I think HOMEBREW_DEVELOPER should be reserved for the truly developery things, and not foisted onto the typical contributor workflow, which seems like using it for a purpose other than it was intended to put it mildly.

ilovezfs:
>I don't think squashing is hostile request, certainly any more than our mountain of audit checks.

@DomT4 the point is that there is a significant gap between the skills required to run brew bump-formula-pr initially and what's required to go fix the PR subsequently if we ask for a change.

As for the current mountain of audit checks, I think hostile is an understatement.

PR #3981, opened 2016-08-17
s6: remove extra blank line

miguelarauj1o:

---

### DIFF
--- a/Formula/s6.rb
+++ b/Formula/s6.rb
@@ -22,7 +22,6 @@ class S6 < Formula
     sha256 "cb4adf9531421b2321c3b1d0a50e388a21f1c1e7bdb38caaa4a7e8158fc0ce72" => :mavericks
   end
 
-
   head do
     url "git://git.skarnet.org/s6"
 


ilovezfs:
@DomT4 I thought as a matter of policy we didn't allow PRs for audit --strict changes like this. I think it would be better to change things like this in bulk rather than cluttering the commit history and unraveling that policy.

DomT4:
We've been accepting PRs for style fixes ever since we stopped displaying non-version/revision changes in `brew update`'s output.

DomT4:
In fact we actively encourage it as a way to get into contributing to `brew` now, heh: https://github.com/Homebrew/brew#contributing

ilovezfs:
I don't think that makes much sense but :see_no_evil:

DomT4:
I think it's fine here, more or less. Almost nobody is watching `homebrew/core` since the split _(which is something of a good thing)_, and `update` hides the cosmetic changes, so we can get away with being noisy.

ilovezfs:
We *can* but this can be scripted across all formulae and done in a single commit, so Can does not imply Ought, as Kant would say.

DomT4:
If you want to reopen the discussion about the contributing block in `homebrew/brew` feel free to do so there, but whilst it says what it currently says I'm not going to shoot down contributors who are relatively new to the ecosystem but want to help out, essentially.

DomT4:
@miguelarauj1o As a minor notation here, quite often empty lines in this area are caused by our bottle code panicking about placement.

The `bottle do` block should essentially never be squished against another block, so in this case ideally you'd have wanted to remove the empty line where you did & add an empty line to avoid the squishing, such as https://github.com/Homebrew/homebrew-core/commit/d7789fd7a1151010411e5b9738f2190030484713.

MikeMcQuaid:
> We can but this can be scripted across all formulae and done in a single commit, so Can does not imply Ought, as Kant would say.

@ilovezfs My hope is that if we can get Rubocop doing our audit in future that `brew audit --fix` can fix all this stuff automatically. In the mean time: please feel free to script mass changes to such things and commit them, though. I want to do the same thing in the core at some point too.

DomT4:
@ilovezfs Merry Christmas: https://github.com/Homebrew/homebrew-core/commit/431e7b2fd379b3a604dd8a1964603ef022d9761d

DomT4:
@miguelarauj1o There's a still a ton of low-hanging fruit to pick even with the automated run. Obvious ones being `share/name` which should be `pkgshare`, args arrays using `[""]` format instead of `%W[]` format unless there's inline comments, etc.

Our magical fixing script isn't completely magical, at least compared to a human eye.

ilovezfs:
@miguelarauj1o https://github.com/Homebrew/homebrew-core/issues/4029 you might like tackling some of those.

PR #3992, opened 2016-08-17
ammonite-repl 0.7.4

emanresusername:


emanresusername:
hold off on this one `0.7.4` is on the way soon

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/a923e6b80d42230046796172056068d881095880. Thank you for your contribution to Homebrew @emanresusername; we appreciate it! :smiley_cat:

ilovezfs:
There is a bug in the brew version detection code:
```
Summary
/usr/local/Cellar/ammonite-repl/0.7: 2 files, 30.7M, built in 0 seconds
```

Until that bug is fixed this one has to have its version set explicitly in the formula:
```
    version "0.7.4"
```

ilovezfs:
@DomT4 I assume you'll fix this since I'm on my iPhone ;)

emanresusername:
@DomT4 lmk if i should push an addendum for that

DomT4:
I have zero idea how this is returning a bad version 😕. Fixed in https://github.com/Homebrew/homebrew-core/commit/0f1e16f64adcc4b7fb0b0661108684429f5c541f for now. Thanks for noticing @ilovezfs.

ilovezfs:
@DomT4 I already chased it down locally and figured out where it breaks. That's the only reason I hadn't already done this one because I wanted to fix that first in the brew repo :)

DomT4:
It's such a simple version, it shouldn't be problematic 😅. Shall await your `brew` PR with curiosity. We can pluck this line out of the formula once the `brew` PR is merged, but might as well not break the update for now.

ilovezfs:
Yup that's why I piped up just now despite being AFK :)

PR #4009, opened 2016-08-18
bash 4.4-beta2 (devel)

zgracem:

---

Hi! This is my first Homebrew pull request; I hope I did everything properly. (I removed the two babylon.network mirrors because I couldn't get the domains to resolve.)

### DIFF
--- a/Formula/bash.rb
+++ b/Formula/bash.rb
@@ -85,15 +85,13 @@ class Bash < Formula
   end
 
   devel do
-    url "https://ftpmirror.gnu.org/bash/bash-4.4-rc1.tar.gz"
-    mirror "https://mirrors.ocf.berkeley.edu/gnu/bash/bash-4.4-rc1.tar.gz"
-    mirror "https://mirrors.kernel.org/gnu/bash/bash-4.4-rc1.tar.gz"
-    mirror "https://ftp.gnu.org/gnu/bash/bash-4.4-rc1.tar.gz"
-    mirror "https://gnu.cu.be/bash/bash-4.4-rc1.tar.gz"
-    mirror "https://mirror0.babylon.network/gnu/bash/bash-4.4-rc1.tar.gz"
-    mirror "https://mirror1.babylon.network/gnu/bash/bash-4.4-rc1.tar.gz"
-    mirror "https://mirror.unicorncloud.org/gnu/bash/bash-4.4-rc1.tar.gz"
-    sha256 "54838ce0e9db6a8920d4c9f6563fd74dac45d3d3c14c8df4cf7ceb68a91e244b"
+    url "https://ftpmirror.gnu.org/bash/bash-4.4-beta2.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
Here are some fresh mirrors to replace the ones that died:

https://mirror.clarkson.edu/gnu/bash/bash-4.4-beta2.tar.gz
https://fr.mirror.babylon.network/gnu/bash/bash-4.4-beta2.tar.gz
https://mirror1.unicorncloud.org/gnu/bash/bash-4.4-beta2.tar.gz
https://gnu.cu.be/bash/bash-4.4-beta2.tar.gz
https://nl.mirror.babylon.network/gnu/bash/bash-4.4-beta2.tar.gz
https://www.localmsp.org/gnu/bash/bash-4.4-beta2.tar.gz

Bonus points for updating the mirrors in the stable block too so they match :)

DomT4:
We're not exactly running a mirrors deficit here 😄. Not hugely fussed about adding more mirrors, but removing any dead ones from stable is ❤️ if you wouldn't mind.

ilovezfs:
@DomT4 it's for the sake of the patches in stable. The reason the patches had ended up vendor was due to issues with the mirrors being unreliable is this is intended to prevent that from ever being an issue again. But it may be that the mirror system is now far more reliable and this can be pared all the way down to the nub with just the standard 2 or the standard 2 plus Berkeley.

DomT4:
Generally adding the two GNU URL types, `ftpmirror` and `ftp.gnu.org`, and then adding Berkeley & `mirrorservice` is bulletproof. If all of those are down the internet is probably broken 😄.

I'm mostly worried about when we add the check that mirrors are still functional the formulae that have lots of mirrors are going to pop big in the audit.

ilovezfs:
Im concerned that check will be a problem. Sometimes the mirrors take a few days to be updated so removing them because they don't have the latest version will be a mistake unless it's done asynchronously with a lag of several days at least.

ilovezfs:
If the mirror itself is dead that's another story :)

DomT4:
Theoretically, and I'm not promising anything here, we could check if the _previous version_ hadn't yet appeared on the mirror. If that was the case, 99.9% of the time that mirror is 💀.

ilovezfs:
I think I'd just check for the mirror itself being dead before worrying about anything fancier but not having the current version or the previous is definitely bad. I'm just not sure it's worth the extra work since the mirror server itself 404ing is almost always the problem.

DomT4:
Not a bad idea. Remember this discussion when I file the `brew` PR at some undetermined future date, likely within the next 2 weeks unless Mike or you beat me to it 😃.

ilovezfs:
@cryptostrophe there are two more of the same in the patch block :)

cryptostrophe:
Oops! Thanks for the catch.

ilovezfs:
@cryptostrophe thanks for putting up with us!

ilovezfs:
@cryptostrophe Merged! Thank you for your contribution to Homebrew! :tada:

PR #4020, opened 2016-08-18
Fix redundant-with-version-scanned-from-URL errors

ilovezfs:
Reaps the rewards of Homebrew/brew#744.

CC @DomT4


DomT4:
Nice work 👍

PR #4040, opened 2016-08-19
peco 0.4.0

ilovezfs:
`go_install` prototype

### DIFF
--- a/Formula/peco.rb
+++ b/Formula/peco.rb
@@ -1,10 +1,9 @@
-require "language/go"
-
 class Peco < Formula
   desc "Simplistic interactive filtering tool"
   homepage "https://github.com/peco/peco"
-  url "https://github.com/peco/peco/archive/v0.3.6.tar.gz"
-  sha256 "edc1ec186a0f439ae84071c9e00f68fec6f8fe49efc9b6bb10462e72f7286b23"
+  url "https://github.com/peco/peco/archive/v0.4.0.tar.gz"
+  sha256 "d04559c04aaf7aab829186e8e977765d8d800ed21a96e0e99c091b90cb7c0f50"
+  head "https://github.com/peco/peco.git"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @DomT4 any thoughts on the prototype?

Formula/peco.rb
```diff
@@ -13,55 +12,28 @@ class Peco < Formula
...
-
-  go_resource "github.com/mattn/go-runewidth" do
-    url "https://github.com/mattn/go-runewidth.git",
-    :revision => "d6bea18f789704b5f83375793155289da36a3c7f"
-  end
-
-  go_resource "github.com/nsf/termbox-go" do
-    url "https://github.com/nsf/termbox-go.git",
-    :revision => "362329b0aa6447eadd52edd8d660ec1dff470295"
+  def go_install(rel = "", options = {})
```
MikeMcQuaid: Dunno what `rel` is

Formula/peco.rb
```diff
@@ -13,55 +12,28 @@ class Peco < Formula
...
+    ENV["GOPATH"] = buildpath
+    using = options[:using]
+    rel = Pathname.new(rel)
+    proj = Pathname.new(active_spec.url.split("/")[2..4].join("/"))
```
MikeMcQuaid: Would be good to make the `2..4` bit more obvious what it's doing and why?

Formula/peco.rb
```diff
@@ -13,55 +12,28 @@ class Peco < Formula
...
+    proj = proj.dirname/proj.basename(proj.extname)
+    dir = buildpath/"src/#{proj}"
```
MikeMcQuaid: Similarly a better variable would be nice.

MikeMcQuaid:
Code seems 👍 just needs more readability

Formula/peco.rb
ilovezfs: relative path

Formula/peco.rb
ilovezfs: hehe that's become something of a standard:
```
consul-template.rb:42:    dir = buildpath/"src/github.com/hashicorp/consul-template"
gauge.rb:31:    dir = buildpath/"src/github.com/getgauge/gauge"
gitlab-ci-multi-runner.rb:35:    dir = buildpath/"src/gitlab.com/gitlab-org/gitlab-ci-multi-runner"
ironcli.rb:29:    dir = buildpath/"src/github.com/iron-io/ironcli"
openssl.rb:69:        dir = "build-#{arch}"
pup.rb:31:    dir = buildpath/"src/github.com/ericchiang/pup"
syncthing-inotify.rb:20:    dir = buildpath/"src/github.com/syncthing/syncthing-inotify"
the_platinum_searcher.rb:20:    dir = buildpath/"src/github.com/monochromegane/the_platinum_searcher"
```
but sure :)

MikeMcQuaid: Obvious in context, not obvious outside of context 😉

Formula/peco.rb
MikeMcQuaid: `def go_install(relative_path = "", options = {})` 😉

ilovezfs: Also it can likely be detected most of the time, so probably belongs in the options hash for override only.

Formula/peco.rb
ilovezfs: I think it may also need to be downcased.

Formula/peco.rb
ilovezfs: There's probably a proper `go` term of art for that directory.

PR #4043, opened 2016-08-19
Ansible: fix add gnu-tar dependency

joemaller:

---

Ansible's unarchive module requires GNU tar. ref: ansible/ansible-modules-core#4390

### DIFF
--- a/Formula/ansible.rb
+++ b/Formula/ansible.rb
@@ -17,6 +17,7 @@ class Ansible < Formula
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "libyaml"
   depends_on "openssl"
+  depends_on "gnu-tar"
 
   #
   # ansible (core dependencies)


MikeMcQuaid:
I don't think we want to add this for a single module that only applies when targeting OS X. It would be good to request BSD tar support from upstream, too/

joemaller:
@MikeMcQuaid There's been discussion about other tar flavors, but the Ansible team is set on GNU tar for now.

- https://github.com/ansible/ansible-modules-core/pull/4352
- https://github.com/ansible/ansible-modules-core/issues/3952#issuecomment-238409367

Instead of installing a dependency, should there be some sort of installation note about the gnu-tar requirement for unarchiving?

MikeMcQuaid:
> Instead of installing a dependency, should there be some sort of installation note about the gnu-tar requirement for unarchiving?

Perhaps. Paging @DomT4 and @ilovezfs for thoughts.

ilovezfs:
Since it's run-time, not build-time, I think it needs to be a flat-out dependency, but if this were a formula other than ansible, I'd be talking about :skull: at this point.

DomT4:
I'd probably make it `:recommended` to give people an escape if they want one. Agree this is tantamount to upstream declaring this module isn't portable for OS X/macOS though.

Does upstream check for both `gtar` and `tar` for the GNU flavour? The former is what we encourage & is the default, but the latter is an option some people favour.

joemaller:
@DomT4 From what I can tell, upstream (Ansible) is never going to support any tar flavor besides GNU. Their goal is a Python-native solution that doesn't use any external decompression tools. Whenever they release that capability, Macs won't need gnu-tar anymore. Until then, we're stuck with gnu-tar or a runtime error.

DomT4:
Yeah, my question was more-or-less if `gnu-tar` is installed as simply `tar`, which is an option, does `ansible` bother checking `tar` could be `gnu-tar` or does it simply go _"Nope, no `gtar` here, panic time_"?

joemaller:
@DomT4 "panic time": 

    Unexpected error when accessing exploded file: [Errno 2] No such file or directory

If you check Anisble issue I linked at the top (4390) you can see exactly where Ansible assumes GNU tar and fails. The discussion at the end of that issue was what brought me here to suggest this.

joemaller:
Save you the link. here's what Ansible sent and what it got back:

```
"cmd": "/usr/bin/tar -C \"/tmp/\" -dz -f \"/Users/joe/.ansible/tmp/ansible-tmp-1471608951.6-71472984367082/source\"", 
"err": "Usage:\n  List:    tar -tf <archive-filename>\n  Extract: tar -xf <archive-filename>\n  Create:  tar -cf <archive-filename> [filenames...]\n  Help:    tar --help\n", 
```

ilovezfs:
@DomT4 I commented here: https://github.com/ansible/ansible-modules-core/pull/4352#issuecomment-241084903

Feel free to chime in :)

DomT4:
IMO:
* Make the dep `:recommended`.
* Leave a comment next to the dep linking to the upstream discussion ILZ posted in.
* Reconsider our position in 6 months.

ilovezfs:
It's worth noting that the PR has not been merged yet, so there may be some merit in additional whining.

MikeMcQuaid:
I don't think this is a case that's affecting by any means the majority of Ansible users as it's only when you're targeting OS X (most cases are targeting Linux). Arguably Ansible or the code targeting OS X should be the ones installing GNU's tar and not us.

ilovezfs:
@MikeMcQuaid I'm leaning toward the position that we shouldn't include it at all.

MikeMcQuaid:
@ilovezfs "it" being `gnu-tar"? How about we make it `:optional` and make the `option` text explain what you need it for?

ilovezfs:
>"it" being gnu-tar

Yes.

>How about we make it :optional

I'd rather we leave this problem between users and upstream.

I don't like the precedent of throwing in duplicative GNU utilities as dependencies – whether required, recommended, or optional – whenever an upstream doesn't actually properly port their software to macOS.

It's one thing when we deign to use a duplicative GNU utility to work around some minor build bug and report the problem upstream, but quite another when the software doesn't function properly or completely at run time without the duplicative GNU utility, especially when that's a deliberate upstream decision.

MikeMcQuaid:
> It's one thing when we deign to use a duplicative GNU utility to work around some minor build bug and report the problem upstream, but quite another when the software doesn't function properly or completely at run time without the duplicative GNU utility, especially when that's a deliberate upstream decision.

That seems fair. `pip install ansible` is not installing `gnu-tar`, after all.

@joemaller Passing on this for now, sorry!

joemaller:
@MikeMcQuaid no problem. Great discussion here, thanks all!

PR #4052, opened 2016-08-19
cppzmq (new formula)

jmuncaster:
Add C++ interface to ZeroMQ.

### DIFF
--- a/Formula/cppzmq.rb
+++ b/Formula/cppzmq.rb
@@ -1,0 +1,27 @@
+class Cppzmq < Formula
+  desc "C++ bindings for ZeroMQ"
+  homepage "https://github.com/zeromq/cppzmq"
+  url "https://github.com/zeromq/cppzmq/archive/92d2af6def80a01b76d5e73f073c439ad00ab757.tar.gz"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
We've generally passed on header-only C++ projects in the past as our feeling is they are better fits for a language-specific package manager or vendoring in your projet.

jmuncaster:
I am not aware of a comparable C++ specific package alternative. We use homebrew to avoid vendoring and to simplify new project setup. Is there really a difference between a header-only C++ library and a compiled one? Indeed, some calability (template libraries) cannot be distributed in compiled form.

MikeMcQuaid:
> Is there really a difference between a header-only C++ library and a compiled one? Indeed, some capability (template libraries) cannot be distributed in compiled form.

Yes. In this case all Homebrew is really doing for you is moving files into a particular location.

Passing on this for now, sorry. Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

jmuncaster:
Thanks for the tip

For posterity: `brew tap jmuncaster/homebrew-header-only`

PR #4073, opened 2016-08-20
minisat: update the homepage URL

ghost:
http://minisat.se links to the Buildbot page, from which it is not
clear how to get to the main information page.


dtrebbien:
Nevermind. It seems that the problem was temporary.

PR #4075, opened 2016-08-20
rfcdiff 1.45

ilovezfs:


Formula/rfcdiff.rb
```diff
@@ -38,12 +43,11 @@ def install
                    "sources=rfcdiff.pyht",
                    "prefix=#{prefix}",
                    "install"
+    bin.env_script_all_files(libexec/"bin", :PATH => "#{HOMEBREW_PREFIX}/opt/gnu-sed/libexec/gnubin:$PATH")
```
DomT4: You've only tagged this a `:build` dep but you're env scripting to it?

ilovezfs: fixed.

PR #4076, opened 2016-08-20
rsdns 4.0 (new formula)

adregner:
Rackspace Cloud DNS command line utilities. Minor patching required to
get this web of useful shell scripts to work inside homebrew, but it all
works out.

### DIFF
--- a/Formula/rsdns.rb
+++ b/Formula/rsdns.rb
@@ -5,8 +5,8 @@ class Rsdns < Formula
   sha256 "060bea37a85c9debe59bb23c8a4326e97ced04fe3de15689e34123c978dffb99"
   head "https://github.com/linickx/rsdns.git"
 
-  depends_on "jq"
   bottle :unneeded
+  depends_on "jq"
 
   # This package is essentially a web of shell script that all expect their
   # friends to be in some particular location. Fixing that upstream to be more
--- a/Formula/rsdns.rb
+++ b/Formula/rsdns.rb
@@ -1,0 +1,53 @@
+class Rsdns < Formula
+  desc "Rackspace Cloud DNS CLI"
+  homepage "https://github.com/linickx/rsdns"
+  url "https://github.com/linickx/rsdns/archive/v4.0.tar.gz"
+  sha256 "060bea37a85c9debe59bb23c8a4326e97ced04fe3de15689e34123c978dffb99"
+  head "https://github.com/linickx/rsdns.git"
+
+  depends_on "jq"
+  bottle :unneeded
+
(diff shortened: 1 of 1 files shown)


Formula/rsdns.rb
```diff
@@ -0,0 +1,53 @@
...
+  head "https://github.com/linickx/rsdns.git"
+
+  bottle :unneeded
+  depends_on "jq"
+
+  # This package is essentially a web of shell script that all expect their
+  # friends to be in some particular location. Fixing that upstream to be more
+  # homebrew friendly is a lot of work compared to the simple patch that can
+  # workaround this.
+  patch :DATA
```
MikeMcQuaid: It's a lot more work but it's the right thing to do. You could consider using symlinks or wrapper scripts instead of patching.

Formula/rsdns.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def install
+    libexec.install Dir["rsdns-*.sh"]
+    lib.install "lib/auth.sh", "lib/func.sh"
+    bin.install "rsdns"
+  end
+
+  def caveats
+    <<-EOS.undent
+      Rackspace Cloud credentials can be saved in ~/.rsdns_config. See #{prefix}/README.md for info.
```
MikeMcQuaid: Please remove these caveats; they aren't Homebrew-specific.

Formula/rsdns.rb
```diff
@@ -0,0 +1,53 @@
+class Rsdns < Formula
+  desc "Rackspace Cloud DNS CLI"
+  homepage "https://github.com/linickx/rsdns"
+  url "https://github.com/linickx/rsdns/archive/v4.0.tar.gz"
+  sha256 "060bea37a85c9debe59bb23c8a4326e97ced04fe3de15689e34123c978dffb99"
+  head "https://github.com/linickx/rsdns.git"
+
+  bottle :unneeded
+  depends_on "jq"
```
DomT4: Please leave an empty line between `bottle :unneeded` and the first `depends_on`. Thanks!

dunn:
🔔 !

ilovezfs:
Closing for now, but please feel free to reopen if you can get the patches adopted upstream. Thanks!

Formula/rsdns.rb
adregner: I played with that a little before settling on the patch and it seemed like more work to get all the scripts to work with the libraries and other scripts they expect to be in particular places.  This little patch felt like the simplest solution while still putting the files where they actually belong inside the packages directory tree following unix traditions.

PR #4080, opened 2016-08-20
elinks 0.13-current (devel)

jeromedalbert:

---

Previous devel version was quite old (2012). This new devel version is slightly more recent (2014), has support for `move-half-page-down` and `move-half-page-up` "vi-like" key bindings, and on my machine server-mode works out of the box compared to before.

### DIFF
--- a/Formula/elinks.rb
+++ b/Formula/elinks.rb
@@ -13,9 +13,9 @@ class Elinks < Formula
   end
 
   devel do
-    url "http://elinks.cz/download/elinks-0.12pre6.tar.bz2"
-    version "0.12pre6"
-    sha256 "383646375b8a325bef5a132c8300caab90eb0b842c5f8eff68febc00e29acada"
+    url "http://elinks.or.cz/download/elinks-current-0.13.tar.bz2"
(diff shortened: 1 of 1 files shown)


Formula/elinks.rb
```diff
@@ -13,9 +13,9 @@ class Elinks < Formula
   end
 
   devel do
-    url "http://elinks.cz/download/elinks-0.12pre6.tar.bz2"
-    version "0.12pre6"
-    sha256 "383646375b8a325bef5a132c8300caab90eb0b842c5f8eff68febc00e29acada"
+    url "http://elinks.or.cz/download/elinks-current-0.13.tar.bz2"
```
MikeMcQuaid: Is this a tagged version? Looks like a branchname which makes me concerns the SHA256 will change.

jeromedalbert: Can't verify if it is a branch since I can't `git clone` on their git server, but you're right; I just downloaded the file again today and it has a different sha256 sum.

Oh well, closing the PR then. Too bad their official dev tags are completely outdated.

PR #4087, opened 2016-08-21
peco 0.4.1

ilovezfs:
switch to upstream's glide vendoring

### DIFF
--- a/Formula/peco.rb
+++ b/Formula/peco.rb
@@ -1,10 +1,9 @@
-require "language/go"
-
 class Peco < Formula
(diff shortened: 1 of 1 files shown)


DomT4:
👍. Not intentionally ignoring the other `peco` thread you pinged me into. Promise I'll look that over once I ship out the PR I seem to have spent far too much time working on & disagreeing with myself about for the last 5 or so days.

ilovezfs:
@DomT4 no problem. I didn't think you were. I had no intention of actually shipping that prototype function in an actual formula, so the new version from upstream made it easy to dispense with.

PR #4088, opened 2016-08-21
pwntools 3.0.1

ilovezfs:
use new virtualenv feature

### DIFF
--- a/Formula/pwntools.rb
+++ b/Formula/pwntools.rb
@@ -1,8 +1,10 @@
 class Pwntools < Formula
+  include Language::Python::Virtualenv
+
   desc "CTF framework used by Gallopsled in every CTF"
   homepage "https://pwntools.com/"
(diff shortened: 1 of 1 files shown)


DomT4:
Resources need a bump/rejiggle here, apparently:
```
pkg_resources.DistributionNotFound: pysocks
Error: pwntools: failed
<0> expected but was
<1>.
```

ilovezfs:
@DomT4 PR refreshed. Also, any idea when `binutils` was there? Would you add it back?

DomT4:
Pretty sure it needs `binutils` for `objdump`, so probably required. Rest of the PR looks 💯.

ilovezfs:
worth adding a conflicts_with moreutils since they're both installing an errno binary?

DomT4:
Is that a new thing? There's no conflict currently. If it's a new thing, I guess no other choice.

PR #4107, opened 2016-08-21
qt5: optionally provide webkit module

CamJN:

---

When Qt dropped the qt-webkit module, they broke the capybara-webkit gem. This option allows you to use the latest Qt (at least up to 5.6.1) and still use capybara-webkit.

https://github.com/Homebrew/homebrew-core/pull/2087 exists but doesn't cover this change, and seems stalled.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -53,7 +53,7 @@ class Qt5 < Formula
   option "with-docs", "Build documentation"
   option "with-examples", "Build examples"
   option "with-oci", "Build with Oracle OCI plugin"
-
+  option "with-qtwebkit", "Build with QtWebkit module"
   option "without-webengine", "Build without QtWebEngine module"
 
   deprecated_option "qtdbus" => "with-dbus"
@@ -70,6 +70,12 @@ class Qt5 < Formula
   depends_on :xcode => :build
 
   depends_on OracleHomeVarRequirement if build.with? "oci"
+
+  resource "qt-webkit" do
+    # http://lists.qt-project.org/pipermail/development/2016-March/025358.html
+    url "https://download.qt.io/community_releases/5.6/5.6.1/qtwebkit-opensource-src-5.6.1.tar.gz"
+    sha256 "f5ba5afc5846fc755575dd04081a90a9536f920e312f18f6fb1f5a0c33f477b0"
+  end
 
   def install
     args = %W[
@@ -109,6 +115,11 @@ class Qt5 < Formula
     end
 
     args << "-skip" << "qtwebengine" if build.without? "webengine"
+
+    if build.with? "qtwebkit"
+      (buildpath/"qtwebkit").install resource("qt-webkit")
+      inreplace ".gitmodules", /.*status = obsolete\n((\s*)project = WebKit\.pro)/, "\\1\n\\2initrepo = true"
+    end
 
     system "./configure", *args
     system "make"


MikeMcQuaid:
Is this module still actively maintained?



> When Qt dropped the qt-webkit module, they broke the capybara-webkit gem. This option allows you to use the latest Qt (at least up to 5.6.1) and still use capybara-webkit.

It feels like a better solution is to rely on an old version of Qt 5.

CamJN:
I don't want to be too negative but I was doing that but homebrew makes it a real pain.

Specifically I need the latest `qt5` for `wireshark`, but and older one for `capybara-webkit`, and I need to link whichever I'm using at the time (can't use both obviously). Huge headache.

MikeMcQuaid:
> homebrew makes it a real pain.

Or: Qt has made this a real pain 😉.

> Is this module still actively maintained?

This is an important question we'd need answered.

CamJN:
Looks like it is: 
https://lists.webkit.org/pipermail/webkit-qt/2016-August/004068.html
http://qtwebkit.blogspot.ca/2016/08/qtwebkit-im-back.html

MikeMcQuaid:
I'm not sure what I think about this because 5.7.0 has been released and there's no accordingly stable version for that. Does this version work with 5.7.0? @DomT4 @ilovezfs: thoughts?

DomT4:
Yeah, this really depends for me on how quickly it is going to get updated. If it's going to become the only thing blocking Qt5 updates in future I'm 👎 on the idea. If we can build the existing resource against 5.7.0 that's probably fine but does need stating.

It's perhaps worth noting that QtWebKit always was an absolute security mess because of how infrequently Qt (and particularly QtWebKit) updates were pushed to release. Anything Webkit that doesn't have a rapid release schedule quickly runs into these problems. I'm wary that going out of our way to include it again, even optionally, is more or less facilitating terrible security practices.

I guess if people know what they're getting into it's fine, but I tend to worry most people don't really care & just want something that "works".

MikeMcQuaid:
> If we can build the existing resource against 5.7.0 that's probably fine but does need stating.

Yes, agreed this is the big question.

tdsmith:
I poked #qtwebkit on Freenode; the mailing list could also be a good source of people who could help us clarify some of the outstanding questions here.

annulen:
Tarball of legacy version, adjusted to Qt 5.7 (roughly the same as 5.6.1 otherwise, so there was no changelog): http://download.qt.io/community_releases/5.7/5.7.0/qtwebkit-opensource-src-5.7.0.tar.xz

Technology preview 3 can be built against Qt 5.2 - 5.7: https://github.com/annulen/webkit/releases/tag/qtwebkit-tp3, but at the moment it needs to be built separately from other Qt modules. We are working on fix at https://codereview.qt-project.org/169257.

annulen:
Staying with Qt 5.5 is a complete non-sense, it has lots of bugs fixed in 5.6 and later versions, and 5.6 is LTS release while 5.5 is already EOL.

Formula/qt5.rb
```diff
@@ -110,6 +116,11 @@ def install
 
     args << "-skip" << "qtwebengine" if build.without? "webengine"
 
+    if build.with? "qtwebkit"
+      (buildpath/"qtwebkit").install resource("qt-webkit")
+      inreplace ".gitmodules", /.*status = obsolete\n((\s*)project = WebKit\.pro)/, "\\1\n\\2initrepo = true"
```
MikeMcQuaid: Why is this needed?

MikeMcQuaid:
Would be good to use http://download.qt.io/community_releases/5.7/5.7.0/qtwebkit-opensource-src-5.7.0.tar.xz with Qt 5.7.

Formula/qt5.rb
CamJN: Because nowadays Qt chooses which modules to build based on the `.gitmodules` file.

MikeMcQuaid: Thanks!

MikeMcQuaid:
I see no harm in this for now but if it ends up blocking Qt updates in future we'll need to drop it.

PR #4117, opened 2016-08-22
metabase: migrate to Cask

AnastasiaSulyagina:

---

Deleting metabase from homebrew-core as part of deduplication process . See context [here](https://github.com/caskroom/homebrew-cask/issues/15603). 

cc @MikeMcQuaid


DomT4:
I thought `metabase` the formula & `metabase` the Cask were different things? Am I misremembering?

MikeMcQuaid:
Homebrew's is a jar: http://www.metabase.com/start/other.html and Cask's is a DMG/app: http://www.metabase.com/start/mac.html (and outdated).

ilovezfs:
"The Mac Application is the easiest way to setup Metabase locally on Mac OS X for personal use. Note that currently the Mac Application is not setup for shared use, so some Metabase features which involve the rest of your team won’t be possible."

tdsmith:
ping @kdoh; what do you think about the relative utility of metabase's Mac app and the .jar Homebrew installs?

DomT4:
I'm absolutely fine with this being migrated if that's the correct thing to do FWIW; my question was just driven by the belief they were different but both equally valid installation methods.

MikeMcQuaid:
Let's pass on this for now but thanks for the PR @AnastasiaSulyagina

PR #4144, opened 2016-08-22
zbackup: work around CHECK macro collision

apjanke:
This locally renames the `CHECK` macro to `ZBCHECK` to avoid collisions with other `CHECK` macros. This is required for compatibility with protobuf 3.x

Blocks #3415


apjanke:
Merged. That should be the last blocker for protobuf 3.

PR #3803, opened 2016-08-11
protobuf 3.0.0

jeroen:
Protobuf 3.0.0 is now final. Perhaps it should become the default at some point. I'll make that a seperate PR (once this one is merged).

PR #4145, opened 2016-08-22
Protobuf 3.0

apjanke:
Protobuf major version upgrade to 3.x.

This is an updated version of #3415, preserving original author credit.
- Reorders and squashes a couple commits for a tidy commit history
- Incorporates a compatibility fix for zbackup and removed the bogus version-specific dependency

Closes #3415 
Closes #3803 
## Stuff still broken
- [X] nanopb-generator is broken
- [ ] upscaledb is failing in the test; looks like it's crashing

### DIFF
--- a/Formula/supersonic.rb
+++ b/Formula/supersonic.rb
@@ -5,8 +5,6 @@ class Supersonic < Formula
   sha256 "1592dfd2dc73f0b97298e0d25e51528dc9a94e9e7f4ab525569f63db0442d769"
   revision 1
 
-  depends_on "gflags" => :linked
-
   if MacOS.version < :mavericks
     depends_on "protobuf" => "c++11"
     depends_on "boost" => "c++11"
@@ -17,6 +15,7 @@ class Supersonic < Formula
 
   depends_on "pkg-config" => :build
   depends_on "glog"
+  depends_on "gflags"
 
   needs :cxx11
 
--- a/Formula/supersonic.rb
+++ b/Formula/supersonic.rb
@@ -3,14 +3,9 @@ class Supersonic < Formula
   homepage "https://code.google.com/archive/p/supersonic/"
   url "https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/supersonic/supersonic-0.9.4.tar.gz"
   sha256 "1592dfd2dc73f0b97298e0d25e51528dc9a94e9e7f4ab525569f63db0442d769"
+  revision 1
 
-  bottle do
-    revision 1
-    sha256 "cd76e23c748ec6cfb520800a1890e691680921efa4e6e1bf1dd991b0557e6d33" => :el_capitan
-    sha256 "75c8903d7d637aa495c83eb3a3569b627fbdf9799e907a7d0fff536cf3ddb155" => :yosemite
-    sha256 "c72cb21c7f1efc5a790be376e55dcc4a1edc76cda6686c2ba7d3d7f8c2937321" => :mavericks
-    sha256 "2869dba7bb685f7f6e4df504b87e01a7dda685afc3f27fa1c6010c150982317f" => :mountain_lion
-  end
+  depends_on "gflags" => :linked
 
   if MacOS.version < :mavericks
     depends_on "protobuf" => "c++11"
--- a/Formula/ddar.rb
+++ b/Formula/ddar.rb
@@ -3,7 +3,7 @@ class Ddar < Formula
   homepage "https://github.com/basak/ddar"
   url "https://github.com/basak/ddar/archive/v1.0.tar.gz"
   sha256 "b95a11f73aa872a75a6c2cb29d91b542233afa73a8eb00e8826633b8323c9b22"
-  revision 1
+  revision 2
 
   head "https://github.com/basak/ddar.git"
 
--- a/Formula/gjstest.rb
+++ b/Formula/gjstest.rb
@@ -3,7 +3,7 @@ class Gjstest < Formula
   homepage "https://github.com/google/gjstest"
   url "https://github.com/google/gjstest/archive/v1.0.2.tar.gz"
   sha256 "7bf0de1c4b880b771a733c9a5ce07c71b93f073e6acda09bec7e400c91c2057c"
-  revision 4
+  revision 5
 
   head "https://github.com/google/gjstest.git"
 
--- a/Formula/libphonenumber.rb
+++ b/Formula/libphonenumber.rb
@@ -3,6 +3,7 @@ class Libphonenumber < Formula
   homepage "https://github.com/googlei18n/libphonenumber"
   url "https://github.com/googlei18n/libphonenumber/archive/libphonenumber-7.6.1.tar.gz"
   sha256 "2ed406c2b535f496a1ce9ced2ceefaa40c8473ad6dc53bb7e5fe0279f107333f"
+  revision 1
 
   bottle do
     cellar :any
--- a/Formula/mobile-shell.rb
+++ b/Formula/mobile-shell.rb
@@ -3,6 +3,7 @@ class MobileShell < Formula
   homepage "https://mosh.mit.edu/"
   url "https://mosh.mit.edu/mosh-1.2.6.tar.gz"
   sha256 "7e82b7fbfcc698c70f5843bb960dadb8e7bd7ac1d4d2151c9d979372ea850e85"
+  revision 1
 
   bottle do
     sha256 "ce0722d44de6a01bb2eeb65e57913ece2e67608220f88ce88313433774b848bd" => :el_capitan
--- a/Formula/nanopb-generator.rb
+++ b/Formula/nanopb-generator.rb
@@ -3,6 +3,7 @@ class NanopbGenerator < Formula
   homepage "https://koti.kapsi.fi/jpa/nanopb/docs/index.html"
   url "https://koti.kapsi.fi/~jpa/nanopb/download/nanopb-0.3.6.tar.gz"
   sha256 "3e6d5d4971dc11845261ddca7e1c67b96eabf95e839327c7d8ed6f07412edab7"
+  revision 1
 
   bottle do
     cellar :any_skip_relocation
@@ -15,8 +16,13 @@ class NanopbGenerator < Formula
   depends_on "protobuf"
 
   resource "protobuf-python" do
-    url "https://pypi.python.org/packages/source/p/protobuf/protobuf-2.6.0.tar.gz"
-    sha256 "b1556c5e9cca9069143b41312fd45d0d4785ca0cab682b2624195a6bc4ec296f"
+    url "https://pypi.python.org/packages/14/3e/56da1ecfa58f6da0053a523444dff9dfb8a18928c186ad529a24b0e82dec/protobuf-3.0.0.tar.gz"
+    sha256 "ecc40bc30f1183b418fe0ec0c90bc3b53fa1707c4205ee278c6b90479e5b6ff5"
+  end
+
+  resource "six" do
+    url "https://pypi.python.org/packages/source/s/six/six-1.10.0.tar.gz"
+    sha256 "105f8d68616f8248e24bf0e9372ef04d3cc10104f1980f54d57b2ce73a5ad56a"
   end
 
   def install
--- a/Formula/ola.rb
+++ b/Formula/ola.rb
@@ -3,6 +3,7 @@ class Ola < Formula
   homepage "https://www.openlighting.org/ola/"
   url "https://github.com/OpenLightingProject/ola/releases/download/0.10.2/ola-0.10.2.tar.gz"
   sha256 "986e61874bc80db3b23cf201af2dafa39e3412cc50cddf1cd449c869110bfd27"
+  revision 1
 
   bottle do
     sha256 "37dabc62d0ed531c06bdf4bc153858d0a8f30f895b4f3765d6f2b9a69b03e840" => :el_capitan
--- a/Formula/osm-pbf.rb
+++ b/Formula/osm-pbf.rb
@@ -10,6 +10,7 @@ class OsmPbf < Formula
     sha256 "c102e66477bbbb75dfd91e0349659f59781002246a57775f2d7cab5fbf7cae27" => :yosemite
     sha256 "6a9a5fb19889732dbe9d9837e38d417be53277a44df6b8fb2191c8f9f4872b70" => :mavericks
   end
+  revision 1
 
   depends_on "protobuf"
 
--- a/Formula/protobuf-c.rb
+++ b/Formula/protobuf-c.rb
@@ -9,6 +9,7 @@ class ProtobufC < Formula
     sha256 "178f84dfb66628862e08a3bf58c5a0bdb1a6624f4de63d1670832ea394b5c04d" => :yosemite
     sha256 "50c184193ace913ed581c66669418f4755e4ac44428a9383e10d6178022664aa" => :mavericks
   end
+  revision 1
 
   option :universal
 
--- a/Formula/protobuf-swift.rb
+++ b/Formula/protobuf-swift.rb
@@ -3,18 +3,12 @@ class ProtobufSwift < Formula
   homepage "https://github.com/alexeyxo/protobuf-swift"
   url "https://github.com/alexeyxo/protobuf-swift/archive/2.4.4.tar.gz"
   sha256 "c5960def4f9d48d4933f1a1ff1ac403ca278c0502ee048c6c8704d769b0ae7c5"
-
-  bottle do
-    cellar :any
-    sha256 "17cf44adc252e062bc9e63cdf2c110066c6b0ad0c994be7feaa1f35fe7b56831" => :el_capitan
-    sha256 "6346f4a7adda25f013c60e3d9bc0e158d571a6007d0dca6d1467525248b41345" => :yosemite
-    sha256 "a7eddb2d6c52a560242b8cb8ed819d9e028f52029df6fdff2f77e50eea1f449a" => :mavericks
-  end
+  revision 1
 
   depends_on "autoconf" => :build
   depends_on "automake" => :build
   depends_on "libtool" => :build
-  depends_on "protobuf"
+  depends_on "protobuf" => "2.6.1"
 
   def install
     system "./autogen.sh"
--- a/Formula/protobuf-swift.rb
+++ b/Formula/protobuf-swift.rb
@@ -1,14 +1,13 @@
 class ProtobufSwift < Formula
   desc "Implementation of Protocol Buffers in Apple Swift."
   homepage "https://github.com/alexeyxo/protobuf-swift"
-  url "https://github.com/alexeyxo/protobuf-swift/archive/2.4.4.tar.gz"
-  sha256 "c5960def4f9d48d4933f1a1ff1ac403ca278c0502ee048c6c8704d769b0ae7c5"
-  revision 1
+  url "https://github.com/alexeyxo/protobuf-swift/archive/3.0.0.tar.gz"
+  sha256 "3e93f410844049673a164698589f1cb2c8f8ee1e4169b6cee3c6c32f8f5c4edb"
 
   depends_on "autoconf" => :build
   depends_on "automake" => :build
   depends_on "libtool" => :build
-  depends_on "protobuf" => "2.6.1"
+  depends_on "protobuf"
 
   def install
     system "./autogen.sh"
@@ -19,14 +18,15 @@ class ProtobufSwift < Formula
 
   test do
     testdata = <<-EOS.undent
-       enum Flavor{
-         CHOCOLATE = 1;
-          VANILLA = 2;
-        }
-        message IceCreamCone {
-          optional int32 scoops = 1;
-          optional Flavor flavor = 2;
-        }
+      syntax = "proto3";
+      enum Flavor {
+        CHOCOLATE = 0;
+        VANILLA = 1;
+      }
+      message IceCreamCone {
+        int32 scoops = 1;
+        Flavor flavor = 2;
+      }
     EOS
     (testpath/"test.proto").write(testdata)
     system "protoc", "test.proto", "--swift_out=."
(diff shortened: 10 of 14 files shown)


apjanke:
Failed build:

https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/6660/

* nanopb-generator broken
* ola needed revision bump (now done)
* upscaledb breaking in test
* protobuf-swift audit failure (now fixed)

apjanke:
@agatti Any chance you know enough about `nanopb-generator` to get it working with `protobuf` 3.0?

Formula/supersonic.rb
```diff
@@ -3,14 +3,9 @@ class Supersonic < Formula
...
+  revision 1
 
-  bottle do
-    revision 1
-    sha256 "cd76e23c748ec6cfb520800a1890e691680921efa4e6e1bf1dd991b0557e6d33" => :el_capitan
-    sha256 "75c8903d7d637aa495c83eb3a3569b627fbdf9799e907a7d0fff536cf3ddb155" => :yosemite
-    sha256 "c72cb21c7f1efc5a790be376e55dcc4a1edc76cda6686c2ba7d3d7f8c2937321" => :mavericks
-    sha256 "2869dba7bb685f7f6e4df504b87e01a7dda685afc3f27fa1c6010c150982317f" => :mountain_lion
-  end
+  depends_on "gflags" => :linked
```
DomT4: This sort of syntax is dead now. Will need to remove it.

apjanke: Fixed.

agatti:
@apjanke I can take a look later this evening, it probably is a matter of infrastructure setup.  From a quick look at the log, it seems to be about https://pypi.python.org/pypi/six not being installed in the environment, and the test being ran without specifying the protobuf format version - so it'll default to v2.

ilovezfs:
@BrewTestBot test this please

apjanke:
Thanks @agatti.

FYI, in #4148 the `nanopb-generator` test was modified to explicitly specify the protocol version as `proto2`. This PR has since been rebased on master, so it includes that format version specification.

Maybe `six` needs to be added as a vendored resource dependency inside `nanopb-generator`?

ilovezfs:
@apjanke I just bumped libphonenumber so this PR will need a little kick.

apjanke:
Rebased on master to kick it.

DomT4:
@BrewTestBot test this please

DomT4:
The fetch failures should be resolved now, at least.

Noctem:
You're welcome to cherry-pick or merge Noctem/homebrew-core@87f632ce327e13500c16e5a40e4b613f33d4c770 if you'd like updated Python modules.

DomT4:
Probably not a bad idea. Looks like we're missing a `six` resource as well, which is causing the `nanopb-generator` test to fail, apparently.

apjanke:
Thanks @Noctem! That looks like it will solve some of our problems here. Gonna cherry-pick that in to this PR. The `six` there looks like it may work for `nanopb-generator` as well.

apjanke:
Okay, that fixed `nanopb-generator` for me locally, leaving only `upscaledb` broken.

Upscaledb vendors an older series of `libuv`, version 0.10.36. Even though libuv is [up to 1.9.x now](https://github.com/libuv/libuv/releases), they just released a 0.10.37 patch version back in June, about the time Protobuf 3.0 came out. Bumping the vendored `libuv` to 0.10.37 fixed the `brew test upscaledb` crash for me locally.

Amended the PR with these. Think it might work now.

apjanke:
@BrewTestBot test this please

apjanke:
Darnit. The upscaledb test is [still failing on the test bot](https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/6988/), even though it works for me locally.

Created a separate upscaledb libuv bump at https://github.com/Homebrew/homebrew-core/pull/4389 to see if this is actually a regression with protobuf 3.0, or if it's just exposing an existing error.

jmhodges:
Anything I can do to help with this?

ilovezfs:
@apjanke do you mind rebasing this? I think we should merge it regardless of the upscaledb issue and then a separate PR can deal with that.

apjanke:
Rebased on `master`.

I'm also inclined to merge this as is, though that may just be my eagerness for new releases talking. It seems likely that merging this will break our `upscaledb`, and I don't know how many users that will affect.

@jmhodges The only blocker on this is the failing `upscaledb` test. If you know why upscaledb is crashing with the new `protobuf`, that could help clear this up.

ilovezfs:
@apjanke I see this as a fairly straightforward case of "perfect is the enemy of the good" and think in general these PRs with several reverse dependencies should not be blocked by the few outliers (say like < 20% of the reverse dependencies) that will be broken, even if we let those outliers lie broken for a bit rather than immediately exiling them to the boneyard as part of the breaking PR, so long as there are PRs opened subsequently that show what broke, and we report the problems to the respective upstreams as soon as it's clear we're not going to fix it ourselves.

The alternative is being stuck with an old version in core, which is almost, though not exactly, the same as having a broken version in core, since it begins to impede people's work in just about the same way something broken would.

jmhodges:
I would debug from the previously failing build, but it looks like its 404ing! Working out how to test this locally. Don't see docs immediately. If I don't figure it out, I'll wait for this current build to fail.

jmhodges:
Ah, okay, I somehow need to get the `brew linkage` command going. I see its in dev-cmd but no docs on how to get it or the rest of those installed.

ilovezfs:
it's installed. Library/Homebrew/dev-cmd ... just export HOMEBREW_DEVELOPER=1 and voilà

jmhodges:
Thanks!

jmhodges:
Okay, yep, works locally. Great.

ilovezfs:
@jmhodges also for more toys, `brew tap homebrew/dev-tools`

jmhodges:
Oh, could it be the created bottle doesn't have it? We uninstall protobuf, and then try the new bottle and then tests start failing.

==> brew uninstall --force protobuf
==> brew uninstall --force autoconf automake libtool xz
==> brew install ./protobuf-3.0.0.el_capitan.bottle.tar.gz
==> brew test protobuf --verbose

jmhodges:
Boom! Reproduced by installing from the bottle!

Running

==> brew install --build-bottle --verbose protobuf
==> brew audit protobuf
==> brew bottle --verbose --json protobuf
==> brew bottle --merge --write --no-commit ./protobuf-3.0.0.el_capitan.bottle.json

and then

    brew linkage upscaledb --test
    Missing libraries:
      /usr/local/opt/protobuf/lib/libprotobuf.9.dylib

jmhodges:
I'm at my limits tonight. Have jury duty early in the morning. Perhaps someone with some knowledge of bottling can understand why that lib might not be copied into the bottle.

ilovezfs:
it segfaults though ...

jmhodges:
What segfault? I don't see a segfault in the current build failure.

ilovezfs:
https://gist.github.com/ilovezfs/6c899fa5e344b92e165d191d73867356

jmhodges:
How are you getting that? It seems plausible a test could segfault if it was expecting to dlopen a dylib or some such and didn't check its errors.

The dylib no longer exists on my local machine when I install from the newly made protobuf bottle. That seems to be the first failure. We can look at the upscaledb test failure after that.

jmhodges:
Oh interesting! The bottle tar.gz has a `libprotobuf.10.dylib`. Wonder how it got that!

ilovezfs:
```
iMac-TMP:~ joe$ brew linkage --test protobuf
No broken dylib links
iMac-TMP:~ joe$ brew linkage --test upscaledb
No broken dylib links
iMac-TMP:~ joe$ brew linkage upscaledb
System libraries:
  /System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices
  /usr/lib/libSystem.B.dylib
  /usr/lib/libc++.1.dylib
  /usr/lib/libz.1.dylib
Homebrew libraries:
  /usr/local/opt/boost/lib/libboost_chrono-mt.dylib (boost)
  /usr/local/opt/boost/lib/libboost_filesystem-mt.dylib (boost)
  /usr/local/opt/boost/lib/libboost_system-mt.dylib (boost)
  /usr/local/opt/boost/lib/libboost_thread-mt.dylib (boost)
  /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (openssl)
  /usr/local/opt/protobuf/lib/libprotobuf.10.dylib (protobuf)
  /usr/local/Cellar/upscaledb/2.2.0_1/lib/libupscaledb.0.dylib (upscaledb)
  /usr/local/Cellar/upscaledb/2.2.0_1/lib/libupsserver.0.dylib (upscaledb)
  /usr/local/Cellar/upscaledb/2.2.0_1/libexec/libuv/lib/libuv.dylib (upscaledb)
iMac-TMP:~ joe$ mdutil -t /usr/local/opt/protobuf/lib/libprotobuf.10.dylib
/usr/local/Cellar/protobuf/3.0.0/lib/libprotobuf.10.dylib
iMac-TMP:~ joe$ ls -l /usr/local/Cellar/protobuf/3.0.0/lib/libprotobuf.10.dylib
-r--r--r--  1 joe  admin  2143968 Sep  7 00:43 /usr/local/Cellar/protobuf/3.0.0/lib/libprotobuf.10.dylib
iMac-TMP:~ joe$
```
I'm just pulling this PR `brew pull 4145` and then brew install -s protobuf ; brew install -s upscaledb ; brew test -vd upscaledb ; scp test joe@10.0.1.15: ; ctrl D ; cd ~ ; sudo lldb ./test ; r`

jmhodges:
Right, it seems that `install -s protobuf` isn't the right step. You've got to create and then install from the bottle. Otherwise, everything works.

MikeMcQuaid:
@ilovezfs If we add a `revision` to `upscaledb` I'm almost 100% sure this will Just Work.

jmhodges:
Well, the `linkage upscaledb --test`works, I mean.

MikeMcQuaid:
Similarly with `gjstest`

ilovezfs:
@MikeMcQuaid this is protobuf installed from source then upscaledb installed from source then segfault. but I can try again with a super clean environment if you all think something is amiss for me.

MikeMcQuaid:
@ilovezfs I guess you've found another issue but those two definitely need a revision bump before pulling this PR.

jmhodges:
Yeah, @MikeMcQuaid is on to something. This seems very much because the build-bot is not installing upscaledb from source when it runs its test, but using the new bottle that makes the `libprotobuf.10.dylib` instead of `libprotobuf.9.dylib`.

ilovezfs:
@MikeMcQuaid yes, of course. I'm assuming apjanke left it out due to the segfault.

jmhodges:
Oh, I expected to be able to use the build output to guide what needed debugging. Would love to have those revision bumps in to make sure we're all on the same page.

ilovezfs:
the revision bumps are irrelevant if both are installed from source locally.

jmhodges:
I think you misunderstand why I'm asking for it! When you're trying to help out or just coming back to a problem after a bit away from it, it's nice to be able to look at documented test failures and know they are the actual test failures people are trying to debug elsewhere.

It's a communication and clarity thing to others or future you. We just wasted a bunch of people's time because we didn't have those revision bumps in showing the exact test failure that was the problem. Plus, the PR wouldn't have ever made it in without them, so why not put them in and be that much closer to having this shipping?

ilovezfs:
because I'm not apjanke so I can't do that otherwise I would for the reasons you stated

jmhodges:
Ah, I've not been asking you to do it. I thought it was clear from context that I was hoping apjanke would.

ilovezfs:
He will, no doubt, but given the hour in New York that's unlikely to happen imminently, I suspect.

ilovezfs:
yup same segfault with a clean environment.

jmhodges:
Hrm, tossing "-g" into the CFLAGS doesn't seem to be enough to get upstreamdb to build with symbols. Dropping the `--disable-debug` (which maybe is causing the issue) causes the build to break on a busted assert.

Really got to hit the bricks. But we're closer!

ilovezfs:
that's due to https://github.com/Homebrew/brew/blob/d39eeac0d5041dbbc41f939ef2dbc02823c82fcd/Library/Homebrew/shims/super/cc#L152

you have to remove the  /^-g\d?$/

ilovezfs:
@jmhodges with symbols:
```
iMac-TMP:~ joe$ sudo lldb ./test
Password:
(lldb) target create "./test"
Current executable set to './test' (x86_64).
(lldb) r
Process 66489 launched: './test' (x86_64)
Process 66489 stopped
* thread #1: tid = 0x1426125, 0x0000000100171d36 libupscaledb.0.dylib`::ups_env_create(henv=0x00007fff5fbffcb0, filename=<unavailable>, flags=0, mode=<unavailable>, param=0x0000000000000000) + 163 at upscaledb.cc:286, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000100171d36 libupscaledb.0.dylib`::ups_env_create(henv=0x00007fff5fbffcb0, filename=<unavailable>, flags=0, mode=<unavailable>, param=0x0000000000000000) + 163 at upscaledb.cc:286
   283 	  *henv = 0;
   284 	
   285 	  /* creating a file in READ_ONLY mode? doesn't make sense */
-> 286 	  if (unlikely(isset(flags, UPS_READ_ONLY))) {
   287 	    ups_trace(("cannot create a file in read-only mode"));
   288 	    return (UPS_INV_PARAMETER);
   289 	  }
```

jmhodges:
Shoot, I'm not able to get debug symbols. Added a  `ENV.prepend "CFLAGS", "-g"` to upscaledb's install method, commented out that entire `when` clause in `shims/super/cc` after removing the `-g` regex bit didn't do it, and even got a patch working to allow the `--disable-debug` to be dropped. But no luck.

Maybe there's a copy of cc somewhere I'm supposed to edit, instead?

ilovezfs:
@jmhodges try `--env=std` instead

jmhodges:
Nope, that didn't do it either.

    brew uninstall upscaledb; brew install --env=std -s --debug  upscaledb  && brew test -vd upscaledb

ilovezfs:
--env=std and these in the formula
```
      ENV.prepend "CFLAGS", "-g"
      ENV.prepend "CXXFLAGS", "-g"
```

jmhodges:
Ah! CXXFLAGS. Duh.

jmhodges:
Yeah, that didn't do it either. I've been running the `brew test -vd upscaledb`, letting it fail, then hopping into the tmp directory where its binary is and running it under `lldb test` and getting nada for symbols. After jury duty, I'll take another whack in a different direction.

jmhodges:
(Well, to be clear, I get symbols from the test binary, but not the upscaledb lib)

jmhodges:
Okay, yeah, I'm going to need some different help because I just cannot get the upstreamdb lib to build with debug symbols on my local machine. There's been a lot of undocumented stuff so far in this thread, and I'm assuming it's just another one of those little tweaks.

ilovezfs:
@jmhodges how do you usually build it? You should be able to just point CPPFLAGS and LDFLAGS at the installed protobuf3 and build however you usually do outside of brew.

jmhodges:
I've never built it before. I'm doing this to get a stable version of protobuf3 without breaking people's builds

jmhodges:
Oh, and using homebrew, the commands I'm using are listed above in exactness

ilovezfs:
@jmhodges Ah, OK. Thanks for clarifying.

ilovezfs:
@jmhodges I fixed it :) ...
```
perl -p -i -e 's/isset/myisset/g' `grep -ril isset *`
```

:smiling_imp:

jmhodges:
Great! I'll make a patch. In for the night, but soon, if @apjanke doesn't beat me to it.

ilovezfs:
@jmhodges @apjanke narrowed it down further and this does the job:
```
    inreplace "./src/5upscaledb/upscaledb.cc",
      "#  include \"2protobuf/protocol.h\"",
      "#  include \"2protobuf/protocol.h\"\n#define isset(f, b)       (((f) & (b)) == (b))"
```
or as a patch
```
diff --git a/src/5upscaledb/upscaledb.cc b/src/5upscaledb/upscaledb.cc
index 30496d0..8ada3aa 100644
--- a/src/5upscaledb/upscaledb.cc
+++ b/src/5upscaledb/upscaledb.cc
@@ -32,6 +32,7 @@
 #include "2page/page.h"
 #ifdef UPS_ENABLE_REMOTE
 #  include "2protobuf/protocol.h"
+#define isset(f, b)       (((f) & (b)) == (b))
 #endif
 #include "2compressor/compressor_factory.h"
 #include "2device/device.h"
```

Note that if you move that define above `#  include "2protobuf/protocol.h"` then it will still segfault. Putting back the original definition of `isset` from `0root/root.h` after that include fixes it. I'm not sure exactly what in protocol.h causes the redefinition but I think it's something in the generated header named `"2protobuf/messages.pb.h"` which itself includes some other files.

ilovezfs:
OK here's the offender: https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/port.h#L62

and macOS's `<sys/param.h>` defines
```
/* Bit map related macros. */
#define setbit(a,i)     (((char *)(a))[(i)/NBBY] |= 1<<((i)%NBBY))
#define clrbit(a,i)     (((char *)(a))[(i)/NBBY] &= ~(1<<((i)%NBBY)))
#define isset(a,i)      (((char *)(a))[(i)/NBBY] & (1<<((i)%NBBY)))
#define isclr(a,i)      ((((char *)(a))[(i)/NBBY] & (1<<((i)%NBBY))) == 0)
```
port.h is included by
https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/common.h#L40
which is included by
2protobuf/messages.pb.h
which is the generated header included by
2protobuf/protocol.h
which is included by
src/5upscaledb/upscaledb.cc after it has included 0root/root.h, where the upscaledb definition of isset occurs.

So renaming isset everywhere in upscaledb may indeed be the right answer here so that this collision never happens again.

apjanke:
Nice detective work!

Reported upstream: https://github.com/cruppstahl/upscaledb/issues/82

I think the simple `inreplace` or patch may be the way to go here: that'll take care of it for this version of upscaledb, and we can expect upstream to deal with it in a future release. Less work and code to read than renaming it everywhere. (Unless `inreplace` supports recursive globs?)

I've updated this PR with the workaround. If this build is green, I think we're good to merge.

apjanke:
@ilovezfs said:
> @apjanke I see this as a fairly straightforward case of "perfect is the enemy of the good" and think in general these PRs with several reverse dependencies should not be blocked by the few outliers (say like < 20% of the reverse dependencies) that will be broken, even if we let those outliers lie broken for a bit rather than immediately exiling them to the boneyard...

I see where you're coming from here, and I would also not like to block major version upgrades on a few less-used or more slowly developed formulae. But since we're talking about breaking a previously-shipped formula, I don't think "perfect" is the term for it; more like "not-broken". I'm real hesitant to push out stuff that breaks any formula.

I'd rather deal with situations like this by relaxing the "one version in core" guideline, and doing something like bumping protobuf to 3.0 while grandfathering in a `protobuf2` formula for back-compatibility for packages that need it (either in core, or by allowing a dep on `homebrew/versions`). Not every package has a lot of developer resources, and breaking formula that don't switch quickly to the latest major version of a library seems like punishing the developers and users of packages from smaller projects, or those that have a more conservative or quality-oriented development process.

ilovezfs:
>Unless inreplace supports recursive globs?

It does. The first argument can be an array.

ilovezfs:
@apjanke we can defer the must-not-break-anything/multiple-versions/who-gets-to-suffer conversation since it's now moot with respect to this PR.

apjanke:
Build failed due to `gjstest` linkage problem.

https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/7445/

Adding a revision bump to address.

jmhodges:
Thanks to both you!

ilovezfs:
@apjanke 
```
iMac-TMP:homebrew-core joe$ brew pull https://github.com/Homebrew/homebrew-core/pull/4145
==> Fetching patch 
Patch: https://github.com/Homebrew/homebrew-core/pull/4145.patch
==> Applying patch
Applying: protobuf 3.0.0
Applying: bump revisions for protobuf 3.0
Using index info to reconstruct a base tree...
M	Formula/libphonenumber.rb
Falling back to patching base and 3-way merge...
Auto-merging Formula/libphonenumber.rb
CONFLICT (content): Merge conflict in Formula/libphonenumber.rb
error: Failed to merge in the changes.
Patch failed at 0002 bump revisions for protobuf 3.0
The copy of the patch that failed is found in: .git/rebase-apply/patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
Error: Patch failed to apply: aborted.
```
lol

apjanke:
*sigh*; rebases again...

apjanke:
finally, a green light! Merging if no objections.

MikeMcQuaid:
@apjanke :shipit:

ilovezfs:
:boat: :rocket:

DomT4:
The rebase gunked up a bit 😢. Check out `libphonenumber`, and `nanopb-generator` seems to have been double revision bumped.

apjanke:
Thanks for catching that. I've amended to fix. libphonenumber is cleaned up. nanopb-generator had been revision bumped twice (once in the original PR, and once in a subsequent fix) I've reordered and cleaned up the commits so it has a single revision bump and its commit history is cleaner. Also tweaked the original commit to not remove bottle stanzas, for a cleaner commit history.

DomT4:
Some of the `revision`s landed in the wrong place that time, but let's get this shipped 👍.

The revision placement isn't problematic more than stylistically, so we can fix that in post rather than torturing the CI a bit more.

apjanke:
Yeah, I'll clean that up in post.

Merged! @ilya-medvedev, thank you for the original PR!

Let me know if there are any issues.

jmhodges:
Y'all are great! Thank you so much, @apjanke!

PR #4149, opened 2016-08-22
gcc 6.2.0

grandeit:

---

Update gcc to version 6.2.0

Removed patch [https://github.com/Homebrew/formula-patches/blob/master/gcc/gcc-6.1.0_infinite_memory_snacking.patch](https://github.com/Homebrew/formula-patches/blob/master/gcc/gcc-6.1.0_infinite_memory_snacking.patch) since the issue is fixed in 6.2.0, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70824

### DIFF
--- a/Formula/gcc.rb
+++ b/Formula/gcc.rb
@@ -21,25 +21,13 @@ class Gcc < Formula
 
   desc "GNU compiler collection"
   homepage "https://gcc.gnu.org"
-  revision 1
 
   head "svn://gcc.gnu.org/svn/gcc/trunk"
 
   stable do
-    url "https://ftpmirror.gnu.org/gcc/gcc-6.1.0/gcc-6.1.0.tar.bz2"
-    mirror "https://ftp.gnu.org/gnu/gcc/gcc-6.1.0/gcc-6.1.0.tar.bz2"
-    sha256 "09c4c85cabebb971b1de732a0219609f93fc0af5f86f6e437fd8d7f832f1a351"
-
-    # 6.1.0 contains a bug that in certain circumstances causes GCC to consume
-    # all available memory very quickly & effectively crash systems.
-    # This should be safe to remove on the next stable release.
-    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70977
-    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70824
-    # https://github.com/gcc-mirror/gcc/commit/75e7d6607
-    patch do
-      url "https://raw.githubusercontent.com/Homebrew/formula-patches/c963247c6b/gcc/gcc-6.1.0_infinite_memory_snacking.patch"
-      sha256 "636394ab2024ab026ead265b13b4c2a24e44c20ddfeab43a9be6e78e824de4f2"
-    end
+    url "https://ftpmirror.gnu.org/gcc/gcc-6.2.0/gcc-6.2.0.tar.bz2"
+    mirror "https://ftp.gnu.org/gnu/gcc/gcc-6.2.0/gcc-6.2.0.tar.bz2"
+    sha256 "9944589fc722d3e66308c0ce5257788ebd7872982a718aa2516123940671b7c5"
   end
 
   bottle do


apjanke:
Build failure; Jenkins log is complaining about an aborted build.

```

==> brew install --build-bottle --verbose gcc
Build was aborted
Recording test results
ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?
Archiving artifacts
==> git checkout master -f
```

I'm rerunning it.

apjanke:
@BrewTestBot test this please

DomT4:
> Jenkins log is complaining about an aborted build.

@ilovezfs killed the build. I presume he was impatiently awaiting another PR 😄.

ilovezfs:
@DomT4 just trying to figure out what's going on with this craziness:

https://github.com/Homebrew/homebrew-science/pull/4085#issuecomment-241563621
https://github.com/Homebrew/homebrew-core/pull/4150

DomT4:
Ugh. Once the current build has finished I'll go fix that (at least till it breaks again).

ilovezfs:
I'm about ready to add `touch libexec/".macOS-#{MacOS.version}"` to the formula to get unique checksums for each OS.

ilovezfs:
@DomT4 I already did a manual reset, which fixed it and then it immediately came back, but maybe you have some other magic in mind.

ilovezfs:
something very weird is going on with the kafka formula that's causing the yosemite and mavericks bottles to always have the exact same checksum.

DomT4:
In certain cases bottle checksums can actually be legitimately the same these days.

ilovezfs:
That seems to be the root of the conflict. I'm not kidding about the `touch libexec/".macOS-#{MacOS.version}"`, which I think will kill it.

DomT4:
@MikeMcQuaid's floated plans for reproducible bottles, and in certain cases serving one bottle for multiple OS versions where those bottles spit the same checksum. Adding a `touch` in that sense very much breaks at least the latter.

We've seen this thing quite a bit with CI over the last couple months and it hasn't always been a bottle issue FWIW. I do think it needs looking into more. If it really bothers you immediately, it's not obvious why `kafka` even needs a bottle.

ilovezfs:
Although actually that revision may be enough even though the checksums were the same as each other again, because they at least differ from the prior checksum.

ilovezfs:
>floated plans for reproducible bottles, and in certain cases serving one bottle for multiple OS versions where those bottles spit the same checksum.

is there a goal other than getting to throw the word "reproducible" around like it inherently adds value? :laughing:

DomT4:
Reproducible builds are a huge upside for security validation. Don't trust the CI? Fine, run `brew test-bot xyz` locally, get the same checksum, see? Right now we have no better mechanism than "_shrug_ trust us & trust the CI hasn't been compromised".

ilovezfs:
Sounds like pure fool's gold to me. :see_no_evil:

DomT4:
The best trust model is a model that doesn't rely on any single ultimate point of trust. Reproducible bottles are a step on that path. Anyway, CI is up and running again. So far it hasn't burst into testing flames, but will keep an eye on it.

ilovezfs:
@DomT4 that salmon PR is the one that seems to trigger it, so let's see how that does now.

DomT4:
Already re-queued 😉.

MikeMcQuaid:
Agreed with reproducible builds. It's not something I'm actively working on but it makes me happy that it's no longer impossible to have reproducible builds 😉. 

@ilovezfs what's the problem (if any) with identical checksums? As @DomT4 pointed out it may be a sign nothing has been compiled and (if there's no `resource`s or `patch`s i.e. a single download) it doesn't need a bottle. Otherwise something we may do longer-term is use a single bottle for multiple versions automatically if it's the same checksum 😍

DomT4:
This PR itself seems to be fine, so merged in https://github.com/Homebrew/homebrew-core/commit/270da3d865a541671b4e4b9e2b2c09af1157c3e6. Thank you for your contribution to Homebrew @gwaaf; we appreciate it! :smiley_cat:

If you feel like it, the `gcc6` in `homebrew/versions` needs updating as well. Absolutely fine if you prefer to not submit a PR there!

PR #4150, opened 2016-08-22
kafka: revision to fix bottles

ilovezfs:
sha256 of yosemite and mavericks bottle is the same, which may be
confusing CI.


ilovezfs:
hopefully the checksums will differ this time, though they were exactly the same here:

https://bot.brew.sh/job/Homebrew%20Testing/1050/

ilovezfs:
Exactly the same for Yosemite and Mavericks again. Odd.

PR #4154, opened 2016-08-23
open-mpi 2.0.0

goxberry:

---
- Update OpenMPI to version 2.0.0.
- Add --with-cxx-bindings option because MPI C++ bindings are
  deprecated by the MPI-3.0 standard and are no longer installed by
  default.
- Remove code that moves "vtsetup.jar" because that file no longer
  exists when OpenMPI 2.0.0 is built.

### DIFF
--- a/Formula/konoha.rb
+++ b/Formula/konoha.rb
@@ -3,6 +3,7 @@ class Konoha < Formula
   homepage "https://github.com/konoha-project/konoha3"
   url "https://github.com/konoha-project/konoha3/archive/v0.1.tar.gz"
   sha256 "e7d222808029515fe229b0ce1c4e84d0a35b59fce8603124a8df1aeba06114d3"
+  revision 1
 
   bottle do
     sha256 "85ae184ec6d34407588f31830e5669ee0a69be8d36ce2d9ef7074eb01b8831bf" => :el_capitan
--- a/Formula/open-mpi.rb
+++ b/Formula/open-mpi.rb
@@ -1,8 +1,8 @@
 class OpenMpi < Formula
   desc "High performance message passing library"
   homepage "https://www.open-mpi.org/"
-  url "https://www.open-mpi.org/software/ompi/v1.10/downloads/openmpi-1.10.3.tar.bz2"
-  sha256 "7484bb664312082fd12edc2445b42362089b53b17fb5fce12efd4fe452cc254d"
+  url "https://www.open-mpi.org/software/ompi/v2.0/downloads/openmpi-2.0.0.tar.bz2"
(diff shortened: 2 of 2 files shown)


Formula/open-mpi.rb
```diff
@@ -22,6 +22,7 @@ class OpenMpi < Formula
 
   option "with-mpi-thread-multiple", "Enable MPI_THREAD_MULTIPLE"
   option :cxx11
+  option "with-cxx-bindings", "Enable C++ MPI bindings"
```
MikeMcQuaid: Maybe worth just enabling these by default given they introduce no extra dependencies?

MikeMcQuaid: > Add --with-cxx-bindings option because MPI C++ bindings are
deprecated by the MPI-3.0 standard and are no longer installed by
default.

Never mind, ignore me! Can you add a `deprecated` to the option text?

goxberry: @MikeMcQuaid Done. Thanks for taking a look!

DomT4:
Looks like `konoha` needs a revision bump.

Formula/konoha.rb
```diff
@@ -3,6 +3,7 @@ class Konoha < Formula
   homepage "https://github.com/konoha-project/konoha3"
   url "https://github.com/konoha-project/konoha3/archive/v0.1.tar.gz"
   sha256 "e7d222808029515fe229b0ce1c4e84d0a35b59fce8603124a8df1aeba06114d3"
+  revision 1
```
goxberry: @DomT4 Is this change what you had in mind? I looked at the Jenkins errors, and they don't make sense to me because `brew install konoha` works for me, `libmpi.20.dylib` replaces 'libmpi.12.dylib`, and `konoha` doesn't use the C++ MPI bindings, so `libmpi_cxx_1.dylib` should not be needed.

DomT4: Yeah, that's what I had in mind.

For whatever reason, `konoha` links both `libmpi.12.dylib` and `libmpi_cxx.1.dylib` from:
```
lib/minikonoha/0.1/package/konoha.mpi/mpi_glue.dylib
```
So the formula needed a bump to move over to the updated links.

DomT4:
I let it be here to save the CI some pain, but generally prefer one-commit-per-formula-changed, so:
```
* open-mpi 2.0.0
* konoha: revision for open-mpi
```
etc. Just for future reference more than anything major.

Merged in https://github.com/Homebrew/homebrew-core/commit/3bbc07a7f2fa28d662905271d34d492112e81789. Thank you for your contribution to Homebrew @goxberry; we appreciate it! :smiley_cat:

PR #4160, opened 2016-08-23
mkvtoolnix 9.4.0

ilovezfs:
remove lzo option, which no longer exists
honor `--without-flac` and `--without-libmagic` in configure


DomT4:
👍

ilovezfs:
@DomT4 PR refreshed. Makes me wonder if we should just ditch all the options, except I guess qt5, and require libmagic, flac, and gettext unconditionally.

PR #4162, opened 2016-08-23
pcap_dnsproxy 0.4.7.0

ilovezfs:
remove the two xcode projects that were merged into the third
(thanks to Daniel Zeng for the fix)

case sensitivity fix in the test


DomT4:
Failure looks legit.

blackgear:
FileHash.xcodeproj and KeyPairGenerator.xcodeproj is merged into Pcap_DNSProxy.xcodeproj

Remove line 18,19,22,23 to fix the problems.

This formula will work: https://github.com/blackgear/homebrew-core/blob/de8dad6bcde36c8c5e497495eea7658648f1390c/Formula/pcap_dnsproxy.rb

cc @ilovezfs

ilovezfs:
@blackgear Thanks for that! It worked!

PR #4171, opened 2016-08-23
lean-cli 0.2.0 (new formula)

aisk:

---

### DIFF
--- a/Formula/lean-cli.rb
+++ b/Formula/lean-cli.rb
@@ -1,0 +1,24 @@
+class LeanCli < Formula
+  desc "Command-line tool to develop and manage LeanCloud apps."
+  homepage "https://github.com/leancloud/lean-cli"
+  url "https://github.com/leancloud/lean-cli/archive/v0.2.0.tar.gz"
+  sha256 "980980c764486aa933255fd64d391560e72bba78af8427253e1b4515ab6b68eb"
+  head "https://github.com/leancloud/lean-cli.git"
+
+  depends_on "go" => :build
+
+  def install
+    build_from = build.head? ? "homebrew-head" : "homebrew"
+    ENV["GOPATH"] = buildpath
+    mkdir_p buildpath/"src/github.com/leancloud/"
+    ln_s buildpath, buildpath/"src/github.com/leancloud/lean-cli"
+    system "go", "build", "-o", "bin/lean",
+                 "-ldflags", "-X main.pkgType=#{build_from}",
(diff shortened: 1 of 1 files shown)


Formula/lean-cli.rb
```diff
@@ -0,0 +1,22 @@
...
+  head "https://github.com/leancloud/lean-cli.git"
+
+  depends_on "go" => :build
+
+  def install
+    build_from = build.head? ? "homebrew-head" : "homebrew"
+    ENV["GOPATH"] = buildpath
+    mkdir_p buildpath/"src/github.com/leancloud/"
+    ln_s buildpath, buildpath/"src/github.com/leancloud/lean-cli"
+    system "go", "build", "-o", "output/lean", "-ldflags", "-X main.pkgType=#{build_from}", "github.com/leancloud/lean-cli/lean"
```
DomT4: Any reason we can't install direct here? `bin/"lean"` instead of `"output/lean"`. Would be good to break this into a couple of lines for the sake of not stretching out the formula as well:
```ruby
system "go", "build", "-o", bin/"lean",
             "-ldflags", "-X main.pkgType=#{build_from}",
             "github.com/leancloud/lean-cli/lean"
```
etc.

Formula/lean-cli.rb
```diff
@@ -0,0 +1,22 @@
...
+    bin.install "output/lean"
+  end
+
+  test do
+    system bin/"lean", "--version"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Otherwise, let's at least assert:
```ruby
assert_match version.to_s, shell_output("#{bin}/lean --version")
```

Formula/lean-cli.rb
aisk: We can't just run `go build github.com/leancloud/lean-cli/lean` here, because the go compiler will complain the output file `lean` is conflict with the `lean/` folder here.

Formula/lean-cli.rb
tdsmith: it looks like there isn't; playing with it, everything requires authentication

Formula/lean-cli.rb
tdsmith: Fixed.

tdsmith:
Thanks for the PR!

PR #4112, opened 2016-08-21
hg-fast-export 160415 (import from head-only)

bfontaine:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

This is a follow-up to #3913; upstream added a tag today :metal: 

https://github.com/Homebrew/homebrew-head-only/pull/215

PR #4175, opened 2016-08-23
hg-fast-export: fix test

apjanke:
The `hg-reset.sh` command will fail if run from outside a git repo, even if you're just doing 'hg-reset.sh --help'. This breaks the current test for `hg-fast-export`. We missed it when [initially adding the formula](https://github.com/Homebrew/homebrew-core/pull/4112) because the build skipped bottling for some reason, and apparently this also skipped the test.

This PR fixes the test by switching to using `hg-fast-export.sh --help`, which apparently is happy to run outside a git repo.

### DIFF
--- a/Formula/hg-fast-export.rb
+++ b/Formula/hg-fast-export.rb
@@ -15,6 +15,6 @@ class HgFastExport < Formula
   end
 
   test do
-    system bin/"hg-reset.sh", "--help"
+    system bin/"hg-fast-export.sh", "--help"
   end
 end


MikeMcQuaid:
👍

bfontaine:
My fault; I ran `hg-reset.sh --help` in the repo and checked that `$?` was `0` but forgot to run `brew test hg-fast-export`.

DomT4:
👍

MikeMcQuaid:
Thanks @apjanke!

PR #4180, opened 2016-08-23
zplug 2.2.1

ilovezfs:
fix the test

### DIFF
--- a/Formula/zplug.rb
+++ b/Formula/zplug.rb
@@ -1,8 +1,8 @@
 class Zplug < Formula
   desc "The next-generation plugin manager for zsh"
   homepage "https://zplug.sh/"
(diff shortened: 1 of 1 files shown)


DomT4:
Test failed.

ilovezfs:
@apjanke does the test fix look sane or should that not be needed?

apjanke:
Probably shouldn't be necessary, IMHO.

This looks like an issue with installing zplug to a non-default location. Looks to me like we should set `ZPLUG_HOME` inside the test (as described in the caveats) instead of creating new files.

ilovezfs:
@apjanke better?

apjanke:
Better! This looks right to me.

ilovezfs:
@apjanke great. Thanks for taking a look at this, since, as you know, I'm pretty zsh challenged.

PR #4182, opened 2016-08-23
sngrep 1.4.0

ilovezfs:
input "Q\n" not "Q" in the test so that it doesn't hang
assert exit code is 0 in the test

Setting `ENV["TERM"] = "xterm"` prevents "Error opening terminal: unknown."


DomT4:
👍

PR #4211, opened 2016-08-24
psqlodbc: add support for libiodbc

jimhester:

---

The previous formula supported only compiling with unixodbc, this adds
support for libiodbc as well. It ensures only one or the other can be
linked to the psqlodbc driver.

### DIFF
--- a/Formula/psqlodbc.rb
+++ b/Formula/psqlodbc.rb
@@ -19,13 +19,24 @@ class Psqlodbc < Formula
   end
 
   depends_on "openssl"
-  depends_on "unixodbc"
   depends_on :postgresql
+  depends_on "unixodbc" => :recommended
+  depends_on "libiodbc" => :optional
 
   def install
+    if build.with?("libiodbc") && build.with?("unixodbc")
+      odie "psqlodbc: --without-unixodbc must be specified when using --with-libiodbc"
+    end
+
+    args = %W[
+      --prefix=#{prefix}
+    ]
+
+    args << "--with-iodbc=#{Formula["libiodbc"].opt_prefix}" if build.with?("libiodbc")
+    args << "--with-unixodbc=#{Formula["unixodbc"].opt_prefix}" if build.with?("unixodbc")
+
     system "./bootstrap" if build.head?
-    system "./configure", "--prefix=#{prefix}",
-                          "--with-unixodbc=#{Formula["unixodbc"].opt_prefix}"
+    system "./configure", *args
     system "make"
     system "make", "install"
   end


Formula/psqlodbc.rb
```diff
@@ -19,13 +19,19 @@ class Psqlodbc < Formula
   end
 
   depends_on "openssl"
-  depends_on "unixodbc"
   depends_on :postgresql
+  depends_on "libiodbc" => :optional
+  depends_on "unixodbc" if build.without? "libiodbc"
```
DomT4: Just make this `depends_on "unixodbc" => :recommended`. It's not as smooth in terms of avoiding the dependency but currently Homebrew handles `if build.with?` style conditionals outside of `def install` really quite awfully.

Formula/psqlodbc.rb
```diff
@@ -19,13 +19,19 @@ class Psqlodbc < Formula
...
 
   def install
+    odbc_include = if build.with?("libiodbc")
```
DomT4: Use an `args` array instead for consistency.

Formula/psqlodbc.rb
jimhester: I did not do this originally because `unixodbc` and `libiodbc` are mutually exclusive. They cannot both be linked at the same time. So in order to use `libiodbc` you need to specify `--with-libiodbc --without-unixodbc`. But if you say the `if build.with?` conditionals don't work well I guess there is no choice.

jimhester:
https://github.com/Homebrew/homebrew-core/pull/4211/commits/80e29e0c5ea8f85827c0d010e9e645c73320dfbd makes the suggested changes, using `:recommended` and an args array.

MikeMcQuaid:
What happens if you specify both? If this `configure` fails, that's 🆒 but otherwise we should probably `raise` an exception.

jimhester:
@MikeMcQuaid If the libraries are both installed (but one is not linked) it would output
> Error: You must `brew link libiodbc` before psqlodbc can be installed
But it would happily install both dependencies (failing to link the second) and run `configure` if nothing was installed.

Because of this I added a check for both options being specified (https://github.com/Homebrew/homebrew-core/pull/4211/commits/6c15da0a16b6117c52382275b44260c8bcc13022), indicating you need to specify `--without-unixodbc` since that option is the default.

ilovezfs:
hmm if we make one of them keg only can it link both?

jimhester:
The configure script seems to only use `iodbc` if you pass it both options. The libraries are replacements for each other, it doesn't make sense to try and link with both of them.

ilovezfs:
Is it possible to build without both of them?

MikeMcQuaid:
@jimhester In that case it'd be good to `raise` if both are passed.

ilovezfs:
@MikeMcQuaid maybe we should just switch this over to libiodbc unconditionally and drop the options if that's the one that upstream goes with when both are specified.

jimhester:
@MikeMcQuaid I am using `ondie` (https://github.com/Homebrew/homebrew-core/pull/4211/files#diff-5d4ca727ca22eb8cb2087d7d5bb4f9eeR28) as that was what was suggested in Contributing.md. If you would like me to use `raise` I can change it however.

MikeMcQuaid:
@jimhester All good as is. Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #4214, opened 2016-08-24
swiftlint 0.12.0

jpsim:
This is somewhat of an urgent fix since it addresses an extreme performance regression shipped in 0.11.2 (https://github.com/realm/SwiftLint/issues/772).

Thanks!


ilovezfs:
:ship:'d

@jpsim Thanks for the quick fix!

PR #4219, opened 2016-08-24
libgcrypt: build static library too

wilhelmtell:
I'm not sure why this is not on by default in upstream libgcrypt. I think when you install a library, you come to expect getting a static library. I know I do, when I want to link against it.

### DIFF
--- a/Formula/libgcrypt.rb
+++ b/Formula/libgcrypt.rb
@@ -33,6 +33,7 @@ class Libgcrypt < Formula
 
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
+                          "--enable-static",
                           "--prefix=#{prefix}",
                           "--disable-asm",
                           "--with-libgpg-error-prefix=#{Formula["libgpg-error"].opt_prefix}"


ilovezfs:
No, the vast majority of things in homebrew are provided as dynamic not static libraries, and occasionally both dynamic and static are provided.

DomT4:
This flag provides both dynamic & stable, it's not replacing one with the other, so the PR is fine as-is.

ilovezfs:
Not sure we need it if OP's goal is simply to link the library...

ilovezfs:
Also, static is particularly sketchy with respect to security stuff.

DomT4:
We have a standing policy that if a static library can be built without:
* Disabling dynamic libraries.
* Removing `cellar :any`.
* Requiring an explicit additional build step.

We'll accept it.

ilovezfs:
@DomT4 why is test-bot testing things in homebrew/versions?

ilovezfs:
In that context of security/cryptographic libraries, I question the wisdom of said "policy" but do as you wish.

DomT4:
> why is test-bot testing things in homebrew/versions?

Who even knows at this point 😕.

> In the context of security/cryptographic libraries, I question the wisdom of said "policy" but do as you wish.

Ship has already sailed. At this point if we wanted to make a stand over static libraries & security-related formulae we'd have to revert/change/rewrite a hilariously significant bunch of formulae.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/c53d85beed9e6baaee34e57d7d61ef2c2ede0ab8. Thank you for your contribution to Homebrew @wilhelmtell; we appreciate it! :smiley_cat:

tdsmith:
> Ship has already sailed. At this point if we wanted to make a stand over static libraries & security-related formulae we'd have to revert/change/rewrite a hilariously significant bunch of formulae.

I think ilz raises a good point here; when a crypto library is updated and especially when a vuln is patched we'll want to bump any dependencies that might be linking statically and that'll require careful bookkeeping that I don't think anyone is maintaining.

DomT4:
More specifically: If anyone wants to go back and rewrite every security-related formulae in Homebrew that ships a static library & deal with the user blowback, have at it. Not saying I'll stop you, but this is the status quo right now.

tdsmith:
Would you allow me to put this behind an option? That would improve my confidence that we're doing the right thing by most of our users. If you're aware of other crypto libraries in Homebrew that offer static libraries, I'm happy to open PRs to do the same thing there.

DomT4:
I think you're going to annoy people, but if you're happy to deal with any annoyed people, by all means, I don't object to optioning these things.

As a standing request for all maintainers, my view is generally that people who want something changed should do the work & then be around if things go to hell so other maintainers don't get roasted over an open fire on their behalf.

My umbrage around this sort of thing is mostly that people change things, vanish & then I get to deal with upset/confused/angry/argumentative users, often having to make a positive case for something I wasn't asked about or didn't agree with in the first place.

As for other crypto libraries, we've been shipping static libraries with `openssl` for potentially forever. There's also `libressl`, `mbedtls`, `gsasl`, `nettle`, `libsodium`, `libssh2`, `botan`. Those are just the ones I can actively spot.

Using the security logic you can also make a case for anything that has frequent security issues losing its static library, particularly things like `libgd`, `miniupnpc`, etc.

MikeMcQuaid:
I'm with @DomT4 on this one. If we're going to roll-back this behaviour on a bunch of existing formulae I'd like you to produce a reproducible test-case that demonstrates the problem. Otherwise we're fixing speculative issues that may or may not exist and we have an issue tracker full of bugs that we know do exist.

tdsmith:
I think I've convinced myself that it would be difficult to accidentally link statically on OS X and if you're trying to link statically you probably deserve what you're getting.

The logic here ruffles my feathers, though. :p I'm not sure I could convince you that the theory of universal gravitation implies that users could hurt themselves falling from unprotected balconies.

MikeMcQuaid:
> The logic here ruffles my feathers, though

Which part of the logic? That we shouldn't be fixing "bugs" that might not exist or that we should be fixing bugs reported by users?

tdsmith:
That you want to see bodies before you conclude we should have installed a guardrail.

MikeMcQuaid:
Or, to speak about what we're actually talking about again, I want you to demonstrate an issue exists before speculatively fixing it. There's a reason developers care about reproducibility.

PR #4222, opened 2016-08-25
repo 1.12.34

ilovezfs:
bump `version_scheme` so that the tag and version can be the same again
install manifest-format.txt in `doc`
add a test

### DIFF
--- a/Formula/repo.rb
+++ b/Formula/repo.rb
@@ -2,13 +2,18 @@ class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
   url "https://gerrit.googlesource.com/git-repo.git",
-      :tag => "v1.12.32",
-      :revision => "745b4ad660f8050045b521c4e15b7d3ac0b3d70e"
-  version "1.22"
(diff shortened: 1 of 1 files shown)


Formula/repo.rb
```diff
@@ -2,13 +2,18 @@ class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
   url "https://gerrit.googlesource.com/git-repo.git",
-      :tag => "v1.12.32",
-      :revision => "745b4ad660f8050045b521c4e15b7d3ac0b3d70e"
-  version "1.22"
+      :tag => "v1.12.34",
+      :revision => "39252ba028012fd14e4e283217d842fb80206c52"
+  version_scheme 1
```
MikeMcQuaid: Does this figure `1.22` > `1.12.34` without this?

ilovezfs:
```
irb(main):001:0> Version.new("1.22") > Version.new("1.12.34")
=> true
```

MikeMcQuaid: HOORAY. Yeh 👍 then

PR #4225, opened 2016-08-25
gst-plugins-*: Make orc dependency recommended

aleb:

---

GStreamer developers recommend to hard-depend on "orc".
Removed "gtk+" because it's only used for examples.

### DIFF
--- a/Formula/gst-plugins-bad.rb
+++ b/Formula/gst-plugins-bad.rb
@@ -1,6 +1,7 @@
 class GstPluginsBad < Formula
   desc "GStreamer plugins less supported, not fully tested"
   homepage "https://gstreamer.freedesktop.org/"
+  revision 1
 
   stable do
     url "https://gstreamer.freedesktop.org/src/gst-plugins-bad/gst-plugins-bad-1.8.3.tar.xz"
@@ -26,6 +27,7 @@ class GstPluginsBad < Formula
   depends_on "gst-plugins-base"
   depends_on "openssl"
 
+  depends_on "orc" => :recommended
   depends_on "dirac" => :optional
   depends_on "faac" => :optional
   depends_on "faad2" => :optional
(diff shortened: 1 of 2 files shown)


DomT4:
> GStreamer developers recommend to hard-depend on "orc".

Can you elaborate on this? I couldn't see anything worded this strongly, but highly possible I missed it.

aleb:
See https://cgit.freedesktop.org/gstreamer/gst-plugins-good/tree/REQUIREMENTS - even though orc is listed in the Optional dependencies section, there is this note:

> Notes:		Used by many plugins for accelerating SIMD processing using
>                 run-time generated assembly. Not a hard requirement, but
>                 highly recommended. Packagers, please hard-depend on liborc.

I asked in #gstreamer and they confirmed it's the same for all the gst-plugins-* packages. The REQUIREMENTS files are outdated. The one for gst-plugins-bad does not even mention the orc dependency but it exists, see for example https://packages.debian.org/sid/gstreamer1.0-plugins-bad

DomT4:
I see. That seems fine then. We'll want to bump the formula revision of both the formulae here so default pre-existing installations & default new installations after this change don't differ.

aleb:
Done

Formula/gst-plugins-bad.rb
```diff
@@ -5,6 +5,7 @@ class GstPluginsBad < Formula
   stable do
     url "https://gstreamer.freedesktop.org/src/gst-plugins-bad/gst-plugins-bad-1.8.3.tar.xz"
     sha256 "7899fcb18e6a1af2888b19c90213af018a57d741c6e72ec56b133bc73ec8509b"
+    revision 1
```
DomT4: Sorry, wasn't clear enough. This needs to go underneath the `homepage` in this case. Revisions apply to the whole formula, so tucking them into a block fails.

aleb: Ok, I read "inserted underneath the homepage/url/sha block." in https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md

DomT4: Yeah. It's difficult to lump everything into documentation, which most people already skip over. I'll try and think of something clearer.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/021ec72eeb36761afb7f5c7ffdf883c0e722724c & https://github.com/Homebrew/homebrew-core/commit/fc2d0f09c34cdda27534b9b476d451e0bf95a428. Thank you for your contributions to Homebrew @aleb; we appreciate them! :smiley_cat:

PR #4240, opened 2016-08-26
bump git-fresh to 1.8.0

imsky:

---

### DIFF
--- a/Formula/git-fresh.rb
+++ b/Formula/git-fresh.rb
@@ -1,8 +1,8 @@
 class GitFresh < Formula
   desc "Utility to keep git repos fresh"
   homepage "https://github.com/imsky/git-fresh"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@imsky thanks for your contribution to Homebrew!

Two things for future reference:
1. you should write bump subjects like this: "git-fresh 1.8.0"
2. you added a blank line at the end of the file, which would have caused an audit --strict failure if it had not been removed before I actually merged the PR, which is exactly what would have happened if I'd merged this with the GitHub GUI (since the formula has no bottles), but in this case I used `brew pull` which fixed it automatically, so no harm, no foul.

imsky:
thanks @ilovezfs - will keep that in mind in the future!

ilovezfs:
@imsky cool thanks!

Issue #1705, opened 2016-06-04
Transition formulae to Qt 5


ilovezfs:
These depend on qt and don't already have a qt5 option.
- [x] :racehorse: avidemux - New releases have Qt5 support, OS X dmg ships with by default.
- [x] :racehorse: codequery - Has Qt5 support since 0.16.0.
- [x] :racehorse: djview4 - Should go and live in Homebrew/gui. Has Qt5 support.
- [x] :racehorse: gammaray - Has Qt5 support, should be the default.
- [x] :racehorse: gpsbabel - Has Qt5 support, is preferred in upstream configure script.
- [x] :racehorse: liblastfm - Has Qt5 support, preferred by upstream.
- [x] :racehorse: libqglviewer - Has Qt5 support, we're two versions behind upstream currently.
- [x] :racehorse: qcachegrind - Supports Qt5, has been recommending since last release in 2013.
- [x] :racehorse: qjackctl - Supports Qt5, upstream has been default that since last release.
- [x] :racehorse: quassel - One release behind upstream, has announced Qt4 support is being deprecated, switch to Qt5.
- [x] :racehorse: quazip - Has Qt5 support.
- [x] :racehorse: qxmpp - Has Qt5 support.
- [x] :turtle: *gnuplot - Has Qt5 support, is recommended by upstream over Qt4.
  (revdeps: libqalculate, homebrew/science/kat, homebrew/science/octave, tsung, homebrew/science/maxima, homebrew/science/unafold)
- [x] :turtle: *qwt - Has Qt5 support but gnuradio/pyqwt/qwtpolar currently require it and would likely need migrating too. (revdeps: gnuradio, pyqwt, qwtpolar, homebrew/science/qgis)
- [x] :turtle: *ttfautohint - Has Qt5 support. (revdeps: pdf2htmlex)
- [x] :snail: kdiff3 - Has Qt5 support, but Qt4 is default in upstream dmg.
- [x] :snail: libdbusmenu-qt - Has Qt5 support in upstream, but no new release in 4 years.
- [x] :snail: libechonest - Has Qt5 support, but using it renames headers which is quite breaky. Upside: Would kill qjson dep.
- [x] :snail: qca - Has optional Qt5 support but Qt4 is heavily preferred upstream.
- [x] :snail: *qjson - Has Qt5 support, but libdbusmenu-qt needs this and doesn't support Qt5 in releases yet. (revdeps: libdbusmenu-qt, libechonest)
[...]

PR #4258, opened 2016-08-27
codequery: transition to qt5

JCount:

---

transition codequery to build using Qt 5

---

In its current state this is a bit rough, with a somewhat hard-coded path for `qscintilla2`'s library in order for cmake to find it and use it. 

This is due to the fact that codequery's [FindQt5QScintilla.cmake](https://github.com/ruben2020/codequery/blob/v0.16.0/cmakefind/FindQt5QScintilla.cmake) looks for `NAMES qt5scintilla2 qscintilla2-qt5` whereas the dyld homebrew builds is `libqscintilla2.dylib`.

If anyone can think of a way to do this more neatly I would love any suggestions/guidance.

I suppose that this is part of the larger effort being chronicled in https://github.com/Homebrew/homebrew-core/issues/1705

_(I hope the build bot doesn't choke on this for some strange reason)_

**EDIT:**
this is just an idea I am throwing out there, and it may not be really practical, but `qscintilla2`'s formula could simply be modified to create a `qscintilla2-qt5.dyld` sym since it is now only being built with `qt5` after all.

### DIFF
--- a/Formula/codequery.rb
+++ b/Formula/codequery.rb
@@ -3,7 +3,7 @@ class Codequery < Formula
   homepage "https://github.com/ruben2020/codequery"
   url "https://github.com/ruben2020/codequery/archive/v0.16.0.tar.gz"
   sha256 "4896435a8aa35dbdca43cba769aece9731f647ac9422a92c3209c2955d2e7101"
-  revision 1
+  revision 2
 
   bottle do
     cellar :any
@@ -13,13 +13,17 @@ class Codequery < Formula
   end
 
   depends_on "cmake" => :build
-  depends_on "qt"
+  depends_on "qt5"
   depends_on "qscintilla2"
 
   def install
+    args = std_cmake_args
+    args << "-DBUILD_QT5=ON"
+    args << "-DQT5QSCINTILLA_LIBRARY=#{Formula["qscintilla2"].opt_lib}/libqscintilla2.dylib"
+
     share.install "test"
     mkdir "build" do
-      system "cmake", "..", "-G", "Unix Makefiles", *std_cmake_args
+      system "cmake", "..", "-G", "Unix Makefiles", *args
       system "make"
       system "make", "install"
     end


JCount:
ok... so it choked completely on yosemite's `sip`not being able to execute `python3` because the build system isn't resolving it's complete dependency tree. time to see what I may need to add as a build? dependency besides `python3`.

JCount:
Adding `depends_on :python3 => :build` seems to have allowed the brew bot to sidestep the the recursive dependency issue when building the bottles. Superficially the fact that `python3` isn't directly used by the formula is a bit of an issue, however the entire recursive dependency tree currently means that anyone installing this would most likely already have `python3` on their machine. (That is excluding the corner case where `pyqt5` and `qscintilla2` were built form source `without-python3`.)

There may be a better way of sidestepping the dependency issue, but I currently have zero ideas, not even bad ones.

Formula/codequery.rb
```diff
@@ -13,13 +13,18 @@ class Codequery < Formula
...
   depends_on "cmake" => :build
-  depends_on "qt"
+  depends_on :python3 => :build
+  depends_on "qt5"
   depends_on "qscintilla2"
 
   def install
+    args = std_cmake_args
+    args << "-DBUILD_QT5=ON"
+    args << "-DQT5QSCINTILLA_LIBRARY=#{Formula["qscintilla2"].opt_prefix}/lib/libqscintilla2.dylib"
```
tdsmith: this is good, if you were asking for feedback on this

dunn: `#{Formula["qscintilla2"].opt_prefix}/lib` can be `#{Formula["qscintilla2"].opt_lib}` though.

tdsmith:
What's the error look like without the :python3 dependency?

The `sip` executable is the same whichever Python it's built against; I'm not sure if we're adding unnecessary complexity by asking for sip to be assembled with one python or another, but maybe that's a red herring here.

JCount:
The current "fix" is definitely not the answer, but I put it in to see if everything else worked fine. As to the error, I think I will just force-push my original version so that the brew bot can show the error in its full glory, and therefore be more informative to you and anyone else.

Formula/codequery.rb
JCount: thank you @dunn, I completely overlooked that option

JCount:
The issue could be that [qscintilla2.rb#L28](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qscintilla2.rb#L28) is an `elsif` statement rather than an `else`, so perhaps the brew bot consequently is not treating this last sub-block as the default, the way it might if it were an `else`, and not processing the `depends_on :python3 => :recommended` correctly. Therefore then not having it in the sandbox when it does the more thorough testing.

tdsmith:
Okay, the proximal cause of the failure is that `brew install --only-dependencies --build-bottle --verbose codequery` is failing.

That command installs sip like:
```
==> Installing codequery dependency: sip
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160830-58433-y97sky.sb nice /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/build.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/sip.rb --verbose --with-python3
```

But python3 was never installed, so the sip build fails.

ruben2020:
I'm the owner of `codequery` and I will port it to qt5 for homebrew.

Even now, it can be built for qt5 using:
`cmake -G "Unix Makefiles" -DBUILD_QT5=ON ..`

However, the GUI executable is called `codequery-qt5`. I want to change that to `codequery`. So I need to release a new version.

JCount:
@ruben2020 Thank you very much for the upstream help! I can close this PR if you would like to submit your own, or I could simply adapt this PR to your new release for `qt5` support whenever it come out. Whichever one you would prefer is fine, and is what I will do. Thank you again!

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@JCount I merged https://github.com/Homebrew/brew/pull/835 so hopefully this will now pass CI.

JCount:
@ilovezfs Your fix to `formula_installer` seems to have worked, since `python3` is now correctly being installed/satisfied before the brew bot tries to satisfy `qscintilla2`'s `depends_on "sip" => "with-python3"`. Good insight and work in addressing the actual problem rather than chasing red herrings the way I was!

MikeMcQuaid:

ruben2020:
Thanks @JCount and @ilovezfs

ilovezfs:
@ruben2020 you're welcome! thanks for being such an engaged and accessible upstream

PR #4259, opened 2016-08-27
Install gdk-pixbuf `loaders.cache` file to proper location

staticfloat:

---

`brew audit --new-formula` gives the following errors:

```
$ brew audit --new-formula ./gdk-pixbuf.rb
gdk-pixbuf:
  * Use the `ln` Ruby method instead of `system "ln"`
  * Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "gettext" => :linked
```
- I am not using the `ln` ruby method because I didn't know how to make it overwrite existing files (which it must if you run `postinstall` twice, or if you have previously installed `gdk-pixbuf` through Homebrew).  I saw [this comment](https://github.com/Homebrew/brew/blob/f4fb655aaa907e8ec642469ae21ca03e6cef2e21/Library/Homebrew/cmd/linkapps.rb#L44-L46) in the Homebrew codebase and decided to follow suit.
- I was previously told not to add the `"gettext" -> :linked` dependency, so I am holding off on that for now.

The rationale behind this change is explored in detail in [this thread](https://github.com/Homebrew/homebrew-core/pull/3279), but I'll quote the most relevant parts here:

> as it stands right now, in a "normal" /usr/local Homebrew installation, the loaders.cache file is not a symlink to the currently-installed gdk-pixbuf keg, it is a normal file. This (theoretically) doesn't allow for things like quick switching from version to version through unlinking different kegs of the same formula, just plain doesn't work properly on non-/usr/local Homebrew installations, and in general breaks Homebrew encapsulation.
> 
> On a non-/usr/local Homebrew installation, the environment variables are defined properly and the loaders.cache file is created in the proper place, but no symlink is created because post_install is invoked after link.

### DIFF
--- a/Formula/gdk-pixbuf.rb
+++ b/Formula/gdk-pixbuf.rb
@@ -77,12 +77,14 @@ class GdkPixbuf < Formula
   end
 
   def post_install
-    # Change the version directory below with any future update
-    if build.with?("relocations") || HOMEBREW_PREFIX.to_s != "/usr/local"
-      ENV["GDK_PIXBUF_MODULE_FILE"]="#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
-      ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders"
-    end
+    # loaders.cache should be placed into Keg-specific lib directory
+    module_file = "#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
+    module_dir = "#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders"
+    ENV["GDK_PIXBUF_MODULE_FILE"] = module_file
+    ENV["GDK_PIXBUF_MODULEDIR"] = module_dir
     system "#{bin}/gdk-pixbuf-query-loaders", "--update-cache"
+    # Link newly created module_file into global gdk-pixbuf directory
+    system "ln", "-sf", module_file, "#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/"
   end
 
   def caveats; <<-EOS.undent
--- a/Formula/gdk-pixbuf.rb
+++ b/Formula/gdk-pixbuf.rb
@@ -84,7 +84,7 @@ class GdkPixbuf < Formula
     ENV["GDK_PIXBUF_MODULEDIR"] = module_dir
     system "#{bin}/gdk-pixbuf-query-loaders", "--update-cache"
     # Link newly created module_file into global gdk-pixbuf directory
-    system "ln", "-sf", module_file, "#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/"
+    ln_sf module_file, "#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/"
   end
 
   def caveats; <<-EOS.undent
--- a/Formula/gdk-pixbuf.rb
+++ b/Formula/gdk-pixbuf.rb
@@ -76,21 +76,30 @@ class GdkPixbuf < Formula
     (lib/"gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache").unlink
   end
 
+  # Where we want to store the loaders.cache file, which should be in a
+  # Keg-specific lib directory, not in the global Homebrew lib directory
+  def module_file
+    "#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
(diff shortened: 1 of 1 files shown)


dunn:
> I am not using the ln ruby method because I didn't know how to make it overwrite existing files

You can use `ln_sf`; see http://ruby-doc.org/stdlib-2.3.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-ln_sf

staticfloat:
To be explicit, does that mean that the comment in the Homebrew codebase that I linked to above warning me off of the `ln_sf` ruby method should be ignored?

dunn:
Oh, I didn't realize you linked to a comment about `ln_sf`, my apologies.  @UniqMartin, what are the issues with `ln_sf`?

staticfloat:
Well, I figured I'd add the commit changing it anyway.  If we decide we don't want it, we can just cherry-pick the first commit.

MikeMcQuaid:
Is it maybe worth adding some logic in here instead: https://github.com/Homebrew/brew/blob/d39eeac0d5041dbbc41f939ef2dbc02823c82fcd/Library/Homebrew/keg.rb#L316-L357?

staticfloat:
I'm not sure what we would put there, perhaps you can enlighten me.  The whole reason why we need to manually call `ln` is because [`link` gets called before `post_install`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula_installer.rb#L498-L511), so any files we create within `post_install` are not linked in properly.

Formula/gdk-pixbuf.rb
```diff
@@ -77,12 +77,14 @@ def install
...
 
   def post_install
-    # Change the version directory below with any future update
-    if build.with?("relocations") || HOMEBREW_PREFIX.to_s != "/usr/local"
-      ENV["GDK_PIXBUF_MODULE_FILE"]="#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
-      ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders"
-    end
+    # loaders.cache should be placed into Keg-specific lib directory
+    module_file = "#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
+    module_dir = "#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders"
```
MikeMcQuaid: Pull out this except the `loaders` and then add just the `loaders` onto the `ENV` declaration so you can reuse it with the `ln_sf` below.

Formula/gdk-pixbuf.rb
```diff
@@ -77,12 +77,14 @@ def install
...
+    ENV["GDK_PIXBUF_MODULE_FILE"] = module_file
+    ENV["GDK_PIXBUF_MODULEDIR"] = module_dir
     system "#{bin}/gdk-pixbuf-query-loaders", "--update-cache"
+    # Link newly created module_file into global gdk-pixbuf directory
+    ln_sf module_file, "#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/"
```
MikeMcQuaid: Will all the parent directories definitely already exist?

MikeMcQuaid:
@staticfloat Ignore me then!

Formula/gdk-pixbuf.rb
staticfloat: Yes, after linking the bottle will always have created these directories.  If the directories don't exist, you have a horribly broken version of `gdk-pixbuf`.

staticfloat:
I've cleaned some stuff up per Mike's suggestion, and rebased on top of current `master`.

Formula/gdk-pixbuf.rb
```diff
@@ -76,19 +76,30 @@ def install
     (lib/"gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache").unlink
   end
 
+  # Where we want to store the loaders.cache file, which should be in a
+  # Keg-specific lib directory, not in the global Homebrew lib directory
+  def module_file
+    "#{lib}/gdk-pixbuf-#{gdk_so_ver}/#{gdk_module_ver}/loaders.cache"
+  end
```
MikeMcQuaid: I normally hate methods like this but it does clear things up a lot 😆

MikeMcQuaid:
Thanks again @staticfloat!

For future reference the preferred commit message format is `gdk-pixbuf: install `loaders.cache` file to proper location`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md). Thanks!

staticfloat:
Thanks!

> For future reference the preferred commit message format is gdk-pixbuf: installloaders.cachefile to proper location.

Perhaps I'm just blind, but I don't see where I went wrong with my original commit.  I was adding these other commits onto the end so we could discuss them before I squashed them all together.

MikeMcQuaid:
> Perhaps I'm just blind, but I don't see where I went wrong with my original commit. I was adding these other commits onto the end so we could discuss them before I squashed them all together.

You're not blind, I didn't read your commit properly, my mistake!

PR #4275, opened 2016-08-27
pyenv: add completion script

robsonpeixoto:


zmwangx:
This is not how `pyenv`'s completion works. In particular, installing `pyenv.zsh` to `$(brew --prefix)/share/zsh/site-functions` should have no effect.

To enable completion, just `eval "$(pyenv init -)"` (which you need to do anyway to enable any shell-related features of pyenv — basically everything). To see how that enables completion, let me show you what it really does:

```zsh
$ pyenv init -
export PATH="/Users/zmwang/.pyenv/shims:${PATH}"
export PYENV_SHELL=zsh
source '/usr/local/Cellar/pyenv/1.0.0/libexec/../completions/pyenv.zsh'
command pyenv rehash 2>/dev/null
pyenv() {
  local command
  command="$1"
  if [ "$#" -gt 0 ]; then
    shift
  fi

  case "$command" in
  rehash|shell|virtualenvwrapper|virtualenvwrapper_lazy)
    eval "$(pyenv "sh-$command" "$@")";;
  *)
    command pyenv "$command" "$@";;
  esac
}
```

Note that this was done in Zsh, but it's similar in Bash and Fish. In all cases, the completion script is sourced automatically.

zmwangx:
See also https://github.com/yyuu/pyenv#advanced-configuration if you're opposed to `pyenv init`.

robsonpeixoto:
Thanks @zmwangx. I did just because I don't use the `pyenv init -` because I do not like the concept of shims. I only use it to install python and create virtualenvs.

I'll put it in my fish config file.

PR #4281, opened 2016-08-28
libpsl 0.14.0 (new formula)

SteelPangolin:

---

[libpsl](https://github.com/rockdaboot/libpsl) provides data for, and a C interface to, the [Mozilla Public Suffix List](https://publicsuffix.org/).

### DIFF
--- a/Formula/libpsl.rb
+++ b/Formula/libpsl.rb
@@ -1,0 +1,29 @@
+class Libpsl < Formula
+  desc "C library for the Publix Suffix List"
+  homepage "https://github.com/rockdaboot/libpsl"
+  url "https://github.com/rockdaboot/libpsl/releases/download/libpsl-0.14.0/libpsl-0.14.0.tar.gz"
+  sha256 "e8c794bcedf45c87dee4810570bc62ba7130217887e34e164e3eadc2259f2efb"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "md5sha1sum" => :build
+  # A build script (psl-make-dafsa) requires Python 2.7.
+  depends_on :python => :build if MacOS.version <= :snow_leopard
+  depends_on "icu4c"
+
+  def install
+    # OS X doesn't follow PEP-394, so change psl-make-dafsa's shebang line to use Python 2.
+    inreplace "src/psl-make-dafsa", "#!/usr/bin/env python2", "#!/usr/bin/env python"
+
+    system "autoconf"
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    test_output = shell_output("#{bin}/psl --print-unreg-domain www.example.com").strip
+    assert_equal "www.example.com: com", test_output
+  end
+end


SteelPangolin:
The build seems to be failing because of the upstream repo's Github star count. [Debian](https://packages.debian.org/jessie/libpsl0), [Fedora](https://admin.fedoraproject.org/pkgdb/package/rpms/libpsl/), and [FreeBSD](https://github.com/freebsd/freebsd-ports/tree/master/dns/libpsl) have packaged it, and [mainline `curl` has optional support for it](https://github.com/curl/curl/blob/master/configure.ac#L170).

Formula/libpsl.rb
```diff
@@ -0,0 +1,25 @@
+class Libpsl < Formula
+  desc "C library for the Publix Suffix List"
+  homepage "https://github.com/rockdaboot/libpsl"
+  url "https://github.com/rockdaboot/libpsl/releases/download/libpsl-0.14.0/libpsl-0.14.0.tar.gz"
+  sha256 "e8c794bcedf45c87dee4810570bc62ba7130217887e34e164e3eadc2259f2efb"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "md5sha1sum" => :build
+  depends_on :python2 => :build
```
dunn: We have special dependencies `:python` and `:python3`.  What version of Python 2 is required?  If it's 2.7, you can do
```ruby
```
If it's just whatever Python is on the system you shouldn't need it at all.

SteelPangolin: Thanks for the tip, @dunn. The build-time script `src/psl-make-dafsa` has the shebang line `#!/usr/bin/env python2`, uses `with` but doesn't need to import it from `__future__`, and doesn't use set literals or set/dict comprehensions, so I think it'll run on either 2.6 or 2.7.

dunn:
> the shebang line `#!/usr/bin/env python2`

That might need to be patched to read `#!/usr/bin/env python`, since the macOS convention is for `python` to be Python 2 and `python3` to be Python 3, while Linux usually does `python` => Python 3 and `python2` => Python 2.

Formula/libpsl.rb
```diff
@@ -0,0 +1,41 @@
...
+    assert_equal "www.example.com: com", test_output
+  end
+end
+
+__END__
+--- a/src/psl-make-dafsa
++++ b/src/psl-make-dafsa
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python2
++#!/usr/bin/env python
```
SteelPangolin: @dunn: this implements the shebang line change.

dunn: For a one-liner like that we've got a helper method you can use:
```ruby
inreplace "src/psl-make-dafsa", "#!/usr/bin/env python2", "#!/usr/bin/env python"

Comment (MikeMcQuaid) [inline Formula/libpsl.rb]: Ideally also report it to upstream so they are aware this is required to support OS X.

Comment (dunn) [inline Formula/libpsl.rb]: I thought we found that it wasn't possible to make the shebang cross-platform?  We're doing this already with 
- [`emscripten.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/emscripten.rb)
- [`git-remote-hg.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/git-remote-hg.rb)
- [`gst-validate.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/gst-validate.rb)
- [`src.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/src.rb)
- [`yle-dl.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/yle-dl.rb)

Comment (tdsmith) [inline Formula/libpsl.rb]: Apple doesn't ship `python2` but Apple and I think everyone else *does* ship `python2.7` which should be fine since 2.6 is long since EOL'd.

Comment (MikeMcQuaid) [inline Formula/libpsl.rb]: @dunn Idealy we'd report those all upstream if those upstreams wish to support OS X.

Comment (dunn) [inline Formula/libpsl.rb]: @tdsmith Oh, I didn't know that; cool.

**Comment by SteelPangolin** [comment]
Thanks for your help and patience. I've updated the PR to use the patch helper method, and opened an [upstream PR](https://github.com/rockdaboot/libpsl/pull/56) to fix the problem everywhere.

File: Formula/libpsl.rb [hunk 3]
```diff
@@ -0,0 +1,29 @@
+class Libpsl < Formula
+  desc "C library for the Publix Suffix List"
+  homepage "https://github.com/rockdaboot/libpsl"
+  url "https://github.com/rockdaboot/libpsl/releases/download/libpsl-0.14.0/libpsl-0.14.0.tar.gz"
+  sha256 "e8c794bcedf45c87dee4810570bc62ba7130217887e34e164e3eadc2259f2efb"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "md5sha1sum" => :build
+  # A build script (psl-make-dafsa) requires Python 2.7.
```
zmwangx: Remove this comment.

Formula/libpsl.rb
```diff
@@ -0,0 +1,29 @@
+class Libpsl < Formula
+  desc "C library for the Publix Suffix List"
+  homepage "https://github.com/rockdaboot/libpsl"
```
zmwangx: 18 stars, 5 watchers, no forks. Please explain why this package is notable, or we can't accept it in core. Ref: https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Acceptable-Formulae.md.

Formula/libpsl.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "md5sha1sum" => :build
+  # A build script (psl-make-dafsa) requires Python 2.7.
+  depends_on :python => :build if MacOS.version <= :snow_leopard
+  depends_on "icu4c"
+
+  def install
+    # OS X doesn't follow PEP-394, so change psl-make-dafsa's shebang line to use Python 2.
```
zmwangx: The inreplace is self-documenting enough, I don't think we need this comment.

zmwangx (changes requested):
Notability seems to be a major problem. Maybe you want to explore the idea of hosting this in your own tap, which is really simple. https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md

Formula/libpsl.rb
MikeMcQuaid: Would be good to use the upstream patch instead of this which also self-documents a bit more

MikeMcQuaid (changes requested):

SteelPangolin:
Will set up a tap for this.

PR #4300, opened 2016-08-28
swift: fix build script argument

Crazor:
build-script argument --build-jobs has been renamed to --jobs.

### DIFF
--- a/Formula/swift.rb
+++ b/Formula/swift.rb
@@ -59,37 +59,53 @@ class Swift < Formula
 
   def install
     workspace = buildpath.parent
-    build = workspace/"build"
+    build_dir = workspace/"build"
 
     ln_sf buildpath, "#{workspace}/swift"
     resources.each { |r| r.stage("#{workspace}/#{r.name}") }
 
-    mkdir build do
-      system "#{buildpath}/utils/build-script",
-        "-R",
-        "--build-subdir=",
-        "--no-llvm-assertions",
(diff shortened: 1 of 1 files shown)


DomT4:
Is the change just in `HEAD`, not stable yet?
```
==> /private/tmp/swift-20160828-8642-3tafsv/swift-swift-2.2.1-RELEASE/utils/build-script -R --build-subdir= --no-llvm-assertions --no-swift-assertions --no-swift-stdlib-assertions -- --workspace=/private/tmp/swift-20160828-8642-3tafsv --build-args=-j4 --lldb-use-system-debugserver --install-prefix=/usr/local/Cellar/swift/2.2.1 --darwin-deployment-version-osx=10.10 --jobs=4
Error: Unknown setting: jobs
```

Crazor:
You're absolutely right. I pushed a fix.

DomT4:
I think the fix here is a little overcomplicated for the situation. Keeping the block mostly how it was originally, but tweaking it to use an `args` array and then tacking on:
```ruby
if build.head?
  args << "--jobs=#{ENV.make_jobs}"
else
  args << "--build-jobs=#{ENV.make_jobs}"
end
```
Is probably the simplest way of getting this done.

ilovezfs:
Closing for now but please feel free to reopen with the above suggested change.

apjanke:
Following up with the suggested change in #6307.

PR #4306, opened 2016-08-28
Qu1j0t3 add vasm formula

qu1j0t3:
- [Y] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

Unfortunately I erroneously deleted my original fork of Homebrew for this old PR: https://github.com/Homebrew/homebrew-core/pull/1703

I'm reopening with advice from upstream on setting default "syntaxes" for the set of CPUs supported by this assembler. It's therefore not usually necessary specify _both_ CPU and syntax, although any combination can be built by specifying the options.

Re-reading the old PR, it seems that the concern was mainly CI coverage? In which case we'd need to build more CPUs by default? I have a problem with that in that most users would have a particular CPU in mind and don't need to install anything else. 

I'm reopening a PR to see if we can reach an accommodation. Frank Wille (upstream author) provided this info:

> Also some CPU backends are less mature and rarely used. I would concentrate
> on one or two CPUs, where you need a cross-assembler for.
> M68k, PPC and Z80 are the most mature and frequently used backends (in this
> order).

### DIFF
--- a/Formula/vasm.rb
+++ b/Formula/vasm.rb
@@ -36,7 +36,7 @@ class Vasm < Formula
     cpu_options.each do |cpu|
       syntax_options.each do |syntax|
         prog = "vasm#{cpu}_#{syntax}"
-        system "make", prog, "CPU="+cpu, "SYNTAX="+syntax
+        system "make", prog, "CPU=#{cpu}", "SYNTAX=#{syntax}"
         bin.install prog
       end
     end
--- a/Formula/vasm.rb
+++ b/Formula/vasm.rb
@@ -10,7 +10,7 @@ class Vasm < Formula
   option "with-arm", "Enable ARM CPU target (binary vasmarm_SYNTAX)"
   option "with-c16x", "Enable c16x/st10 microcontroller target (binary vasmc16x_SYNTAX)"
   option "with-jagrisc", "Enable Atari Jaguar GPU/DSP RISC target (binary vasmjagrisc_SYNTAX)"
-  option "with-m68k", "Enable Motorola 68K CPU target (binary vasmm68k_SYNTAX)"
+  option "with-m68k", "Enable Motorola 68K CPU target (binary vasmm68k_SYNTAX) (default)"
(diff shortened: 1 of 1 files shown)


Formula/vasm.rb
```diff
@@ -0,0 +1,84 @@
+class Vasm < Formula
+  desc "Portable and retargetable assembler"
+  homepage "http://sun.hasenbraten.de/vasm/"
+  url "http://server.owl.de/~frank/tags/vasm1_7e.tar.gz"
+  version "1.7e"
+  sha256 "2878c9c62bd7b33379111a66649f6de7f9267568946c097ffb7c08f0acd0df92"
+
+  option "with-6502", "Enable 6502 CPU target (binary vasm6502_SYNTAX)"
```
dunn: This is too many options.  Can most of them be included unconditionally? Are there any that should be `recommended` instead of `optional`?

dunn: Oh, does exactly one have to be specified?

Formula/vasm.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    system "make", "vobjdump"
+    bin.install "vobjdump"
+  end
+
+  def caveats
+    opts = %w[6502 6800 arm c16x jagrisc m68k ppc tr3200 vidcore x86 z80].select { |c| build.with? c }
+    "No target CPU was specified with --with-<cpu>, so #{cpu_options} has been chosen by default." if opts.empty?
+  end
```
MikeMcQuaid: Remove these caveats.

Formula/vasm.rb
MikeMcQuaid: Just build the default set and maybe add one other option `with-all` to build the rest too.

MikeMcQuaid:
@qu1j0t3 Any news on this?

ilovezfs:
@BrewTestBot test this please

MikeMcQuaid:
Closing due to no response. Will reconsider accepting when comments are addressed.

PR #4307, opened 2016-08-28
gst-plugins-good: disable osx_video sink on Sierra

amezin:
osx_video plugin fails to build on macOS Sierra.
Disable it until the issue is resolved.

https://bugzilla.gnome.org/show_bug.cgi?id=770526

### DIFF
--- a/Formula/gst-plugins-good.rb
+++ b/Formula/gst-plugins-good.rb
@@ -74,7 +74,11 @@ class GstPluginsGood < Formula
     # Upstream says it hasn't "been actively tested in a long time";
     # successor is glimagesink (in gst-plugins-bad).
     # https://bugzilla.gnome.org/show_bug.cgi?id=756918
-    args << "--disable-osx_video" if MacOS.version == :snow_leopard
+    # Also, it fails to build on macOS Sierra.
+    # https://bugzilla.gnome.org/show_bug.cgi?id=770526
+    if MacOS.version == :snow_leopard or MacOS.version == :sierra
+      args << "--disable-osx_video"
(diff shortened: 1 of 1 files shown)


Formula/gst-plugins-good.rb
```diff
@@ -53,6 +53,18 @@ class GstPluginsGood < Formula
 
   depends_on "libogg" if build.with? "flac"
 
+  # This plugin causes hangs on Snow Leopard (and possibly other versions?)
+  # Upstream says it hasn't "been actively tested in a long time";
+  # successor is glimagesink (in gst-plugins-bad).
+  # https://bugzilla.gnome.org/show_bug.cgi?id=756918
+  # Also, it fails to build on Sierra.
+  # https://bugzilla.gnome.org/show_bug.cgi?id=770526
+  if MacOS.version == :snow_leopard or MacOS.version == :sierra
```
DomT4: Eh. I'd rather we not bother with the option until we know this is fixed upstream, since that's essentially an invite to attempt to build broken software.

Let's stick to the below `args` block & just tweak that to:
```ruby
if MacOS.version == :snow_leopard || MacOS.version == :sierra
  args << "--disable-osx_video"
end
```

aw1621107:
Heh, beat me by a day. Thanks for making the fix!

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/1be05af488ce817652e0d08f2ee999b223ed0050. Thank you for your contribution to Homebrew @amezin; we appreciate it! :smiley_cat:

PR #4312, opened 2016-08-28
jing 20160828-alpha1

ilovezfs:


DomT4:
Were there legit problems with the latest actual release? Wasn't sure if that was upstream or me being silly with Java stuff 😄.

ilovezfs:
@DomT4 the most recent tag would still need lint-none stuff passed to it, so a new tag would be best if upstream can be sweet talked.

PR #4313, opened 2016-08-28
cfitsio: migrate to science

zmwangx:
### Justification for migration

`cfitsio` deals with the FITS file format, which AFAIK is mostly if not exclusively used for astronomical data. It is hosted by NASA and maintained by NASA astrophysicists.

```
$ brew uses cfitsio --include-build --include-optional
monetdb                                 homebrew/science/healpix                homebrew/science/wcslib
homebrew/science/astrometry-net         homebrew/science/openimageio            homebrew/x11/pyxplot
homebrew/science/ccfits                 homebrew/science/vips
```

Moreover, homebrew/science's description reads

> These formulæ provide software tailored to scientific endeavours. Need to store large amounts of data in a sparse matrix, solve a huge optimization problem or process the images from your space telescope? Maybe we can help!

Well, `cfitsio` is perfect for processing images from space telescopes 😂 (I happen to be a physicist — not in astrophysics, but had some exposure to it — and have indeed processed telescope images with `cfitsio` and its derivatives.)


dpo:
I'm in favor of this move. Note that `monetdb`'s dependency on `cfitsio` is `:optional`.

DomT4:
`gdal` would need fixing as well.

Will let this sit for a day or so in case any other maintainers have objections.

zmwangx:
> gdal would need fixing as well.

Done. Thanks for noticing that...

dunn:
Fine with me!

MikeMcQuaid:
👍 from me

DomT4:
That's a pretty conclusive 👍 from both core & science, so merged. Also merged the Science & X11 PRs.

Thank you for your contribution to Homebrew once again @zmwangx. As ever your work is a delight to deal with 🙇.

PR #4348, opened 2016-08-30
gst-plugins-bad: disable apple_media on Sierra

amezin:
qtkitvideosrc fails to build on macOS Sierra.

https://bugzilla.gnome.org/show_bug.cgi?id=770587

### DIFF
--- a/Formula/gst-plugins-bad.rb
+++ b/Formula/gst-plugins-bad.rb
@@ -55,7 +55,9 @@ class GstPluginsBad < Formula
     # upstream does not support Apple video for older SDKs
     # error: use of undeclared identifier 'AVQueuedSampleBufferRenderingStatusFailed'
     # https://github.com/Homebrew/legacy-homebrew/pull/35284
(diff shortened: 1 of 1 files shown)


Formula/gst-plugins-bad.rb
```diff
@@ -55,7 +55,9 @@ def install
     # upstream does not support Apple video for older SDKs
     # error: use of undeclared identifier 'AVQueuedSampleBufferRenderingStatusFailed'
     # https://github.com/Homebrew/legacy-homebrew/pull/35284
-    if MacOS.version <= :mavericks
+    # applemedia plugin fails to build on Sierra
+    # https://bugzilla.gnome.org/show_bug.cgi?id=770587
+    if MacOS.version <= :mavericks or MacOS.version == :sierra
```
DomT4: We prefer `||` over `or`.

amezin:
Fixed upstream.
Will make another PR with the patch.

PR #4355, opened 2016-08-30
dbus: Update HEAD build

ts826848:
./autogen took the place of ./configure. This updates that command and adds necessary depends_on.

### DIFF
--- a/Formula/dbus.rb
+++ b/Formula/dbus.rb
@@ -19,6 +19,12 @@ class Dbus < Formula
     sha256 "474de2afde8087adbd26b3fc5cbf6ec45559763c75b21981169a9a1fbac256c9"
   end
 
+  head do
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+  end
+
   # Patch applies the config templating fixed in https://bugs.freedesktop.org/show_bug.cgi?id=94494
   # Homebrew pr/issue: 50219
   patch do
@@ -30,6 +36,7 @@ class Dbus < Formula
     # Fix the TMPDIR to one D-Bus doesn't reject due to odd symbols
     ENV["TMPDIR"] = "/tmp"
 
+    system "./autogen.sh", "--no-configure" if build.head?
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
                           "--localstatedir=#{var}",


Formula/dbus.rb
```diff
@@ -30,16 +36,16 @@ def install
...
-                          "--prefix=#{prefix}",
-                          "--localstatedir=#{var}",
-                          "--sysconfdir=#{etc}",
-                          "--disable-xml-docs",
-                          "--disable-doxygen-docs",
-                          "--enable-launchd",
-                          "--with-launchd-agent-dir=#{prefix}",
-                          "--without-x",
-                          "--disable-tests"
+    system "./autogen.sh", "--disable-dependency-tracking",
```
MikeMcQuaid: This change needs to be only for `build.head?`

aw1621107: It appears so. If anything, it looks like the stable version (1.10.8 on my machine) made this change too, or at least it is an option for building.

ilovezfs: @aw1621107 configure is generated by autogen. Sometimes a "release tarball" will already have it pre-generated for you. It's a distinction between building from the raw source code repository and building from a packaged release of the sources.
```
Josephs-MacBook-Pro:~ joe$ tar -tf <(curl -sL https://dbus.freedesktop.org/releases/dbus/dbus-1.10.8.tar.gz)|grep configure
dbus-1.10.8/configure.ac
dbus-1.10.8/configure
Josephs-MacBook-Pro:~ joe$ tar -tf <(curl -sL https://dbus.freedesktop.org/releases/dbus/dbus-1.11.4.tar.gz)|grep configure
dbus-1.11.4/configure.ac
dbus-1.11.4/configure
```
Whereas in the git repository:
```
$ ls dbus|grep configure
configure.ac
```
After running autogen, you will have configure just like the tarball.

aw1621107: @MikeMcQuaid Just realized that I read your sentence with the question mark as part of the sentence and not the code bit. HEAD builds are the only ones that **require** `./autogen.sh`, but it seems to work for the stable and devel versions too.

@ilovezfs Yeah, that makes sense as to why only HEAD builds require it.

ilovezfs:
Passing on this for now, sorry. Thanks for the PR nonetheless!

aw1621107:
:(

ilovezfs:
@aw1621107 I don't mind reopening. Just make sure to limit the change to head only, and pass the arguments to configure not autogen.sh

Here's an example: https://github.com/Homebrew/homebrew-core/blob/a9e17ebc4754bb4f9e707eb76f1b7477044a6a24/Formula/pcre.rb

aw1621107:
@ilovezfs The reason I passed the arguments to autogen is because it told me to when I ran it as just `./autogen.sh`. Something like

```
==> ./autogen.sh
I am going to run ./configure with no arguments - if you wish
to pass any to it, please specify them on the ./autogen.sh command line.
```

And I didn't need to invoke `./configure` at all -- seems like autogen.sh does that automatically? -- so that's why ./configure is gone

ilovezfs:
@aw1621107 https://cgit.freedesktop.org/dbus/dbus/tree/autogen.sh

```
if test x"$NOCONFIGURE" = x; then
  run_configure=true
  for arg in $*; do
    case $arg in 
        --no-configure)
            run_configure=false
            ;;
        *)
            ;;
    esac
  done
else
  run_configure=false
fi
```
So pass `--no-configure` to autogen.sh

aw1621107:
Will do then

aw1621107:
And done!

ilovezfs:
@aw1621107 looks great! Thanks for your persistence here! :+1:

PR #4372, opened 2016-08-31
libngspice 26 (new formula)

jmaibaum:

---

ngspice can be build as a shared library, but their build system does
not allow building the executable and the library in one go without
interfering with Homebrew's formula guidelines.  A pull request [1](https://github.com/Homebrew/homebrew-core/pull/3367)
trying this in the existing ngspice formula has been disapproved.

During the discussion in [1](https://github.com/Homebrew/homebrew-core/pull/3367) it was suggested that, in this case, a
separate formula building _only_ the shared library might be a solution.
This is what this formula is for.

### DIFF
--- a/Formula/libngspice.rb
+++ b/Formula/libngspice.rb
@@ -19,22 +19,11 @@ class Libngspice < Formula
     patch :DATA
   end
 
-  option "without-cider", "Build without CIDER numerical device simulator"
-  option "without-xspice", "Build without XSPICE extensions"
-
   def install
     system "./autogen.sh" if build.head?
-
-    args = %W[
-      --disable-debug
-      --disable-dependency-tracking
-      --prefix=#{prefix}
-      --with-ngshared
-    ]
-    args << "--enable-cider" if build.with? "cider"
-    args << "--enable-xspice" if build.with? "xspice"
-
-    system "./configure", *args
+    system "./configure", "--prefix=#{prefix}", "--disable-debug",
+      "--disable-dependency-tracking", "--with-ngshared", "--enable-cider",
+      "--enable-xspice"
     system "make", "install"
 
     # To avoid rerunning autogen.sh for stable builds, work around the
--- a/Formula/libngspice.rb
+++ b/Formula/libngspice.rb
@@ -1,0 +1,79 @@
+class Libngspice < Formula
+  desc "Spice circuit simulator as shared library"
+  homepage "http://ngspice.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/ngspice/ng-spice-rework/26/ngspice-26.tar.gz"
+  sha256 "51e230c8b720802d93747bc580c0a29d1fb530f3dd06f213b6a700ca9a4d0108"
+
+  head do
+    url "git://git.code.sf.net/p/ngspice/ngspice"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "bison" => :build
+    depends_on "flex" => :build
+    depends_on "libtool" => :build
+
+    # Currently, headers don't get installed to include/*.
+    # There is a patch upstream that addresses this for HEAD.
+    # Upstream ticket: https://sourceforge.net/p/ngspice/bugs/327
+    patch :DATA
+  end
+
+  option "without-cider", "Build without CIDER numerical device simulator"
+  option "without-xspice", "Build without XSPICE extensions"
+
+  def install
+    system "./autogen.sh" if build.head?
+
+    args = %W[
+      --disable-debug
+      --disable-dependency-tracking
+      --prefix=#{prefix}
+      --with-ngshared
+    ]
+    args << "--enable-cider" if build.with? "cider"
+    args << "--enable-xspice" if build.with? "xspice"
+
+    system "./configure", *args
+    system "make", "install"
+
+    # To avoid rerunning autogen.sh for stable builds, work around the
+    # includedir bug by symlinking.  Upstream ticket:
+    # https://sourceforge.net/p/ngspice/bugs/327
+    include.install_symlink Dir[share/"ngspice/include/*"] if build.stable?
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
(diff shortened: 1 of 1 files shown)


Formula/libngspice.rb
```diff
@@ -0,0 +1,49 @@
...
+      #include <cstdlib>
+      #include <sharedspice.h>
+      int ng_exit(int status, bool immediate, bool quitexit, int ident, void *userdata) {
+        return status;
+      }
+      int main() {
+        return ngSpice_Init(NULL, NULL, ng_exit, NULL, NULL, NULL, NULL);
+      }
+    EOS
+    system ENV.cc, "test.cpp", "-I#{share}/ngspice/include/ngspice", "-L#{lib}", "-lngspice", "-o", "test"
```
bfontaine: Why are the headers under `share` instead of `include`?

jmaibaum: It's because that's where ngspice installs them by default. Should I move them manually?

jmaibaum:
@bfontaine The new commit includes your style suggestions from #4374 .

Formula/libngspice.rb
MikeMcQuaid: Yes, I think so, thanks.

Formula/libngspice.rb
```diff
@@ -0,0 +1,50 @@
...
+    url "git://ngspice.git.sourceforge.net/gitroot/ngspice/ngspice"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "bison" => :build
+    depends_on "libtool" => :build
+  end
+
+  option "without-cider", "Build without CIDER integration"
+  option "without-xspice", "Build without x-spice extensions"
```
MikeMcQuaid: What's the use-case for these?

jmaibaum: These are extenstions to the basic ngspice simulator, i.e. they offer more advanced simulation capabilities to the program/library. Ngspice can be built with or without them.

Formula/libngspice.rb
jmaibaum: After researching this issue a bit, I see several options to resolve it:

1. Symlink the folder `share/ngspice/include/ngspice` to `include/ngspice` by `include.install_symlink Dir[share/"ngspice/include/*"`
2. Use a patch file from the [upstream ticket] that modifies the makefiles to install the headers into the correct location.

What is the preferred way to do?

[upstream ticket]:https://sourceforge.net/p/ngspice/bugs/327/

Formula/libngspice.rb
bfontaine: Why would someone install `libngspice` without them? (i.e. are these options really needed?)

jmaibaum: I think this depends on the specific use case that people have in mind with using the library. I have no problem with hard-coding them in, but I thought that offering them as options would be most convenient, i.e. _if_ people want to disable them they can do so rather than force them to a specific set of options.

jmaibaum:
I have updated the Formula with a dual approach to solving the includedir bug:

There is a [patch upstream] that changes the dirs `Makefile.am` in order to put the files in the correct dir.  The new formula uses this patch for HEAD builds.

For stable builds, one could also use the aforementioned patch, but this would require changing a lot of stuff in the current formula (i.e. run autogen.sh as well for stable builds).  I worked around this by simply symlinking `share/include/*` to `include/*` in case of a stable build.

I have put two other Formulas on Gist that solve this issue by:

a. a "[symlink-only]" approach.
b. a "[patch-only]" approach.

If you prefer one of those to the current Formula in this PR, please tell me.

[patch upstream]:https://sourceforge.net/p/ngspice/bugs/327/
[symlink-only]:https://gist.github.com/jmaibaum/b84c4d239621d7bb4d4353b1fba28d91
[patch-only]:https://gist.github.com/jmaibaum/4bf859a3ebd3bb97f606539289b32e6e

Formula/libngspice.rb
```diff
@@ -0,0 +1,79 @@
...
+    depends_on "libtool" => :build
+
+    # Currently, headers don't get installed to include/*.
+    # There is a patch upstream that addresses this for HEAD.
+    # Upstream ticket: https://sourceforge.net/p/ngspice/bugs/327
+    patch :DATA
+  end
+
+  option "without-cider", "Build without CIDER numerical device simulator"
+  option "without-xspice", "Build without XSPICE extensions"
```
jmaibaum: I'm sorry for force-pushing here. Next time, I will just add another commit. I am still aware of the unresolved issue that was brought up with the two options.

The last comment on this by @bfontaine was:

> Why would someone install `libngspice` without them? (i.e. are these options really needed?)

and I answered:

> I think this depends on the specific use case that people have in mind with using the library. I have no problem with hard-coding them in, but I thought that offering them as options would be most convenient, i.e. if people want to disable them they can do so rather than being forced to a specific set of options.

MikeMcQuaid: Let's just omit the options for now, thanks.

Formula/libngspice.rb
```diff
@@ -0,0 +1,79 @@
...
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "bison" => :build
+    depends_on "flex" => :build
+    depends_on "libtool" => :build
+
+    # Currently, headers don't get installed to include/*.
+    # There is a patch upstream that addresses this for HEAD.
+    # Upstream ticket: https://sourceforge.net/p/ngspice/bugs/327
+    patch :DATA
```
MikeMcQuaid: Would `configure --incluedir=/foo/baz` as suggested there fix this without a patch for now?

jmaibaum: That was what I tried at first, before trying the patch and the symlinking. I think the commenter upstream meant to say that --includedir=/foo does allow relocating the headers _after_ the patch has been applied.

MikeMcQuaid: Gotcha 👍

MikeMcQuaid:

MikeMcQuaid:
Merged in 4f360cb.

PR #4386, opened 2016-08-31
zsh-syntax-highlighting: fix install with --git

alyssais:

---

See zsh-users/zsh-syntax-highlighting#357 for a description of the issue this fixes.

The maintainers of zsh-syntax-highlighting [decided](https://github.com/zsh-users/zsh-syntax-highlighting/issues/357#issuecomment-243836555) this would be better as a Homebrew patch than being implemented in zsh-syntax-highlighting itself.

### DIFF
--- a/Formula/zsh-syntax-highlighting.rb
+++ b/Formula/zsh-syntax-highlighting.rb
@@ -1,7 +1,9 @@
 class ZshSyntaxHighlighting < Formula
   desc "Fish shell like syntax highlighting for zsh"
   homepage "https://github.com/zsh-users/zsh-syntax-highlighting"
-  url "https://github.com/zsh-users/zsh-syntax-highlighting.git", tag: "0.4.1"
+  url "https://github.com/zsh-users/zsh-syntax-highlighting.git",
+    :tag => "0.4.1",
+    :revision => "c19ee583138ebab416b0d2efafbad7dc9f3f7c4f"
   sha256 "971b9c1e881a8d60442e40003c9e4c98bd1d5243a32fdf386a217c4cd7815197"
   head "https://github.com/zsh-users/zsh-syntax-highlighting.git"
 
--- a/Formula/zsh-syntax-highlighting.rb
+++ b/Formula/zsh-syntax-highlighting.rb
@@ -1,7 +1,7 @@
 class ZshSyntaxHighlighting < Formula
   desc "Fish shell like syntax highlighting for zsh"
   homepage "https://github.com/zsh-users/zsh-syntax-highlighting"
-  url "https://github.com/zsh-users/zsh-syntax-highlighting/archive/0.4.1.tar.gz"
+  url "https://github.com/zsh-users/zsh-syntax-highlighting.git", tag: "0.4.1"
   sha256 "971b9c1e881a8d60442e40003c9e4c98bd1d5243a32fdf386a217c4cd7815197"
   head "https://github.com/zsh-users/zsh-syntax-highlighting.git"
 
@@ -13,14 +13,6 @@ class ZshSyntaxHighlighting < Formula
   end
 
   def install
-    inreplace "Makefile" do |s|
-      # Fix for `brew install --git zsh-syntax-highlighting`.
-      # Without this, the git repository with no HEAD would cause
-      # `git rev-parse` to fail and abort the installation.
-      s.gsub! "git rev-parse HEAD",
-        "git rev-parse --verify HEAD 2> /dev/null || cat .revision-hash"
-    end
-
     system "make", "install", "PREFIX=#{prefix}"
   end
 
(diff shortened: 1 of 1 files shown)


Formula/zsh-syntax-highlighting.rb
```diff
@@ -13,6 +13,14 @@ class ZshSyntaxHighlighting < Formula
   end
 
   def install
+    inreplace "Makefile" do |s|
+      # Fix for `brew install --git zsh-syntax-highlighting`.
+      # Without this, the git repository with no HEAD would cause
+      # `git rev-parse` to fail and abort the installation.
+      s.gsub! "git rev-parse HEAD",
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

MikeMcQuaid:
> The maintainers of zsh-syntax-highlighting decided this would be better as a Homebrew patch than being implemented in zsh-syntax-highlighting itself.

Missed this, chiming in there.

tdsmith:
`install --git` seems like a Homebrew misfeature to me.

penman:
After trying to use it, that's my impression too. It doesn't seem to do anything useful. I had expected that it would set up some remotes or something, but if all it does is `git init; git add -A`, it's not really helpful.

ilovezfs:
Actual `brew install --git` is indeed a useful feature. For example, you can use it to transform an `inreplace` into a patch by setting `raise "foo"` just after the `inreplace`and then copying the output from `git diff`.

MikeMcQuaid:
Agreed on it being a useful feature. @penman What happens if we use a GitHub clone instead of a tarball? That may be a potential solution to avoid patching?

penman:
That would work, I think.

MikeMcQuaid:
Worth a shot 👍

penman:
Looks like that works @MikeMcQuaid

Formula/zsh-syntax-highlighting.rb
```diff
@@ -1,7 +1,7 @@
 class ZshSyntaxHighlighting < Formula
   desc "Fish shell like syntax highlighting for zsh"
   homepage "https://github.com/zsh-users/zsh-syntax-highlighting"
-  url "https://github.com/zsh-users/zsh-syntax-highlighting/archive/0.4.1.tar.gz"
+  url "https://github.com/zsh-users/zsh-syntax-highlighting.git", tag: "0.4.1"
```
MikeMcQuaid: Sorry, can you use hash-rockets (`=>`) and specify a `:revision` too? Thanks!

penman:
:+1: Will update later when I get home.

Why does it need a tag _and_ a revision, btw?

Formula/zsh-syntax-highlighting.rb
penman: :+1: Will update later when I get home.

Why does it need a tag _and_ a revision, btw?

MikeMcQuaid: To act like a checksum for the tag so it doesn't get changed by a malicious party.

danielshahaf: > To act like a checksum for the tag so it doesn't get changed by a malicious party.

Raised https://github.com/zsh-users/zsh-syntax-highlighting/issues/360 to track this.

danielshahaf:
For the record, I was wondering what would happen if somebody did `git commit` after the `git add -A`: that would have caused `$ZSH_HIGHLIGHT_REVISION` to report a meaningless value.  Cloning upstream's repository instead avoids this problem, too ☺

Formula/zsh-syntax-highlighting.rb
penman: Ah, clever!

penman: As an aside, I notice there are a handful of formulae that only specify a tag, with no revision. Would you be interested in a PR that fixed all of these at once by setting the revision to whatever commit the tag is currently pointed to?

Formula/zsh-syntax-highlighting.rb
```diff
@@ -1,7 +1,9 @@
 class ZshSyntaxHighlighting < Formula
   desc "Fish shell like syntax highlighting for zsh"
   homepage "https://github.com/zsh-users/zsh-syntax-highlighting"
-  url "https://github.com/zsh-users/zsh-syntax-highlighting/archive/0.4.1.tar.gz"
+  url "https://github.com/zsh-users/zsh-syntax-highlighting.git",
+    :tag => "0.4.1",
+    :revision => "c19ee583138ebab416b0d2efafbad7dc9f3f7c4f"
   sha256 "971b9c1e881a8d60442e40003c9e4c98bd1d5243a32fdf386a217c4cd7815197"
```
DomT4: `sha256` is redundant here, since you can't checksum a directory, which is what git clones create.

MikeMcQuaid:
Thanks gain @penman!

PR #4392, opened 2016-08-31
ironssh 0.9.1 (new formula)

zmre:
- [No - see below] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

Hello,

IronSSH is an alternative to sftp and scp that keeps files encrypted after they're uploaded and allows sharing of files with cryptographic enforcement. 

This fails the following two `--new-formula` tests:
- GitHub fork (not canonical repository)
- GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)

To the first: this is a fork of openssh, but it does not install anything that conflicts with openssh. At present it installs only `ironsftp` and the related man page. It will eventually also install `ironscp`. This is a complementary project to openssh that shares a code base. Please waive the fork requirement given the unusual reason for the fork.  https://github.com/ironcorelabs/ironssh is the base repo for this project.

To the second point: this project has just had its first release. Packages are available via rpm and deb for Linux platforms. We'd like to make it easy to install for Mac users as well.  Please consider an exception to the popularity requirement.

### DIFF
--- a/Formula/ironssh.rb
+++ b/Formula/ironssh.rb
@@ -1,0 +1,27 @@
+class Ironssh < Formula
+  desc "Fork of OpenSSH adding transparent E2E encryption to file transfers"
+  homepage "https://github.com/ironcorelabs/ironssh"
+  url "https://github.com/IronCoreLabs/ironssh/archive/0.9.1.tar.gz"
+  sha256 "6b2f78244dd836afdf6da561fda73d16fe42477b7f4034b1aeb09d75f23aff47"
+
+  depends_on "automake" => :build
+  depends_on "autoconf" => :build
+  depends_on "openssl"
+  depends_on "libsodium"
+
+  def install
+    system "autoreconf"
+
+    system "./configure",
+                          "--with-ssl-dir=#{Formula["openssl"].opt_prefix}",
+                          "--with-libedit=/usr",
+                          "--prefix=#{prefix}"
+
+    system "make", "SSH_PROGRAM=/usr/bin/ssh", "ironsftp"
+    system "make", "install"
+  end
+
+  test do
+    system "make", "iron-tests"
+  end
+end


apjanke:
Thanks for your submission! I don't know that we'd want to waive Homebrew Core's Acceptable Formula guidelines in this case, though.

For the "non-canonical fork" requirement: I'm not sure that a GitHub repo fork is the best mechanism for relating IronSSH with its openssh-portable parent. The term "fork" is kind of overloaded here. A GitHub repo fork is generally used for temporary work and changes that are expected to be contributed and merged back upstream. This sounds more like a fork at the project level, where the codebases are expected to diverge, or at least remain distinct, over time. You might want to de-fork the repo on GitHub to establish your project's distinct identity, and just use normal Git remote mechanisms to pull down upstream changes.

This could also be useful if you want people to contribute to your project on GitHub – as I understand it, a fork of a fork on GitHub is in the fork network of the parent repo. This makes it hard to contribute to both your project and the upstream. For example, as it stands now, I can only fork one of `IronCoreLabs/ironssh` or `openssh/openssh-portable`.

As for the popularity/maturity requirement, this does seem like too young a project for incorporation in to core Homebrew. You can still use Homebrew to install it by [creating your own Tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) and publishing the formula there. For example, an `IronCoreLabs/ironssh` tap (kept in the `IronCoreLabs/homebrew-ironssh` repo) would make sense here. That will allow you to iterate on your formula without going through the Homebrew maintainers, and users can still install via `brew` with `brew tap IronCoreLabs/ironssh; brew install ironssh`. Then once the project has more history, the formula could be migrated in to Homebrew Core.

The `brew test` step is failing, too.

```
failed: brew test ironssh --verbose

Stacktrace

        Testing ironssh
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160901-75112-p7edku.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/ironssh.rb --verbose
==> make iron-tests
make: *** No rule to make target `iron-tests'.  Stop.
Error: ironssh: failed
Failed executing: make iron-tests
/usr/local/Library/Homebrew/formula.rb:1595:in `block in system'
/usr/local/Library/Homebrew/formula.rb:1532:in `open'
```

The `brew test` step is a post-installation test intended to run against the package after it is installed, and does not have the source code available. Something more like `ironssh --version` would be appropriate in the `test do`. A `make tests` invocation generally goes inside the `install` process, possibly conditional on a `--with-test` option passed to `brew install`. See the core `audiofile` or `mysql` formulae for examples of this.

zmre:
That's good advice on the fork. We hadn't considered that others could not have their own forks of both ironssh and openssh.  Thanks for that.

I was somewhat hoping to avoid making and maintaining our own tap -- not from the effort perspective, but because we want to be available in homebrew-core in the future and I'm not sure how that transition works, but I suspect it is a bit awkward.

Thanks for the note on the brew test. I misunderstood when and where that got run and I thought the install step was causing it to run and succeed.  I'll get that fixed up.

Although the project is young, if we were to defork and to fix the test bit, would you reconsider adding ironssh to homebrew-core to save end users from some extra hassle?

Thanks,

..Patrick

MikeMcQuaid:
We'll consider including this project in core when it's a bit more established. It's less than a month old and, as mentioned above, we suggest you create your own tap for now. Sorry and thanks for the PR.

PR #4395, opened 2016-09-01
hashcat 3.10 (new formula)

cprecioso:

---

### DIFF
--- a/Formula/hashcat.rb
+++ b/Formula/hashcat.rb
@@ -1,0 +1,30 @@
+class Hashcat < Formula
+  desc "World's fastest and most advanced password recovery utility"
+  homepage "https://hashcat.net/hashcat/"
+  url "https://hashcat.net/files/hashcat-3.10.tar.gz"
+  sha256 "3b555e5f7b35ab6a4558bc460f28d80b32f5a211bf9e08d6a1ba1bad5203e3e9"
+
+  depends_on "gnu-sed" => :build
+
+  # Upstream could not fix OpenCL issue on Mavericks.
+  # https://github.com/hashcat/hashcat/issues/366
+  # https://github.com/Homebrew/homebrew-core/pull/4395
+  depends_on :macos => :yosemite
+
+  def install
+    system "make", "install", "CC=#{ENV.cc}", "PREFIX=#{prefix}"
+  end
+
+  test do
+    cp_r pkgshare.children, testpath
+    cp bin/"hashcat", testpath
+    (testpath/"my.dict").write <<-EOS.undent
+      foo
+      test
+      bar
+    EOS
+    hash = "098f6bcd4621d373cade4e832627b4f6"
+    cmd = "./hashcat -m 0 --potfile-disable --quiet #{hash} #{testpath}/my.dict"
+    assert_equal "#{hash}:test", shell_output(cmd).chomp
+  end
+end


Formula/hashcat.rb
```diff
@@ -0,0 +1,21 @@
+class Hashcat < Formula
+  desc "World's fastest and most advanced password recovery utility"
+  homepage "https://hashcat.net/hashcat/"
+  url "https://hashcat.net/files/hashcat-#{version}.tar.gz"
+  version "3.10"
```
apjanke: We prefer to omit the `version` line and let it be parsed from the `url`. An explicit `version` is typically only for weird filenames where the version can't be parsed automatically.

Formula/hashcat.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    args = %W[
+      CC=#{ENV.cc}
+      PREFIX=#{prefix}
+    ]
+    system "make", "install", *args
+  end
+
+  test do
+    system "hashcat", "--version"
```
apjanke: Could you use `assert_match ..., shell_output(...)` to do a basic check on the output here?

apjanke:
[Failed](https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/7032/) due to SHA mismatch.

```
        ==> Downloading https://hashcat.net/files/hashcat-.tar.gz
==> Downloading from https://hashcat.net/hashcat/
Downloaded to: /Users/brew/Library/Caches/Homebrew/hashcat-3.10.tar.gz
SHA256: 72f66db6df7602582240909ed9edcfa24395ba9f999864dbc97beaabd17b02ae
==> Retrying download
==> Downloading https://hashcat.net/files/hashcat-.tar.gz
==> Downloading from https://hashcat.net/hashcat/
Warning: Formula reports different SHA256: 3b555e5f7b35ab6a4558bc460f28d80b32f5a211bf9e08d6a1ba1bad5203e3e9
Downloaded to: /Users/brew/Library/Caches/Homebrew/hashcat-3.10.tar.gz
SHA256: 72f66db6df7602582240909ed9edcfa24395ba9f999864dbc97beaabd17b02ae
```

Note the file name here: `hashcat-.tar.gz`. The version is empty. That's because the `version` line is after the `url` line in the formula. We prefer to just write out the full URL, omit the `version` line, and let the version be parsed from the file name.

Formula/hashcat.rb
bfontaine: Plus the `version` is nil when `"https://hashcat.net/files/hashcat-#{version}.tar.gz"` is parsed; you’d need to put the `version "3.10"` line before it.

Formula/hashcat.rb
```diff
@@ -0,0 +1,21 @@
...
+  sha256 "3b555e5f7b35ab6a4558bc460f28d80b32f5a211bf9e08d6a1ba1bad5203e3e9"
+
+  depends_on "gnu-sed" => :build
+
+  def install
+    args = %W[
+      CC=#{ENV.cc}
+      PREFIX=#{prefix}
+    ]
+    system "make", "install", *args
```
bfontaine: You can inline the arguments here.

Formula/hashcat.rb
bfontaine: Also use `bin/"hashcat"` or `"#{bin}/hashcat"` instead of `"hashcat"`.

cprecioso:
I've fixed everything but the test, could you help me with that? I'm not literate in Ruby.

A simple example (in shell) would be:
```sh
echo "test" > example.dict
hashcat -m 0 --potfile-disable --quiet "098f6bcd4621d373cade4e832627b4f6" example.dict
```
Output should be: `098f6bcd4621d373cade4e832627b4f6:test`

How would I write that in Ruby?

ilovezfs:
@cprecioso here you go :)
```
  test do
    cp_r pkgshare.children, testpath
    cp bin/"hashcat", testpath
    (testpath/"my.dict").write <<-EOS.undent
      foo
      test
      bar
    EOS
    hash = "098f6bcd4621d373cade4e832627b4f6"
    cmd = "./hashcat -m 0 --potfile-disable --quiet #{hash} #{testpath}/my.dict"
    assert_equal "#{hash}:test", shell_output(cmd).chomp
  end
```

cprecioso:
Great, thanks, going to add it now. May I ask, why are you copying the binary and running from the current dir instead of directly `bin/"hashcat"`?

cprecioso:
Added the test

ilovezfs:
@cprecioso because running it directly will cause a sandbox violation. It checks whether it's being executed from the installed prefix, and if it is, it uses /Users/foo/.hashcat/blah which is not writable from within the sandbox. If it's executed from somewhere other than the installed prefix then it uses the path from which it's executed instead for the temp files, which prevents the sandbox violation.

MikeMcQuaid:
@ilovezfs Could we override $HOME for it somehow instead?

ilovezfs:
@MikeMcQuaid only by modifying the source code with a patch.

tdsmith:
Failing on 10.9:

```
==> ./hashcat -m 0 --potfile-disable --quiet 098f6bcd4621d373cade4e832627b4f6 /tmp/hashcat-test-20160901-92810-8zpesv/my.dict


ERROR: clBuildProgram(): CL_INVALID_BUILD_OPTIONS


Error: hashcat: failed
<0> expected but was
<nil>.
```

Does this support Mavericks?

zmwangx:
@tdsmith 

> Does this support Mavericks?

In an upstream bug report https://github.com/hashcat/hashcat/issues/366, a maintainer says the following:

> OK, to pack this together: This compiler error only occurs in case of OSX 10.9.5 and only for the Iris Pro, is that correct?

> OK, but since I can't fix this because I don't have that device with that OSX and since it's an error that occurs only for old OS I can't do anything to fix this. I'll close the issue, but for everyone stumbling on it, feel free to reopen and fix it.

Since it's pretty much a wontfix.

We can do either of the following:

1. `depends_on :macos => :yosemite`;
2. Wait until Sierra release, by which time we no longer need to test or bottle for Mavericks.

ilovezfs:
> depends_on :macos => :yosemite

this.

zmwangx:
@ilovezfs Did you notice the "only for Iris Pro" part?

ilovezfs:
@zmwangx yes, but ... I'm on 10.8, no iris, and ...
```
==> Cleaning
Fixing /usr/local/Cellar/hashcat/3.10/bin/hashcat permissions from 755 to 555
==> Finishing up
ln -s ../Cellar/hashcat/3.10/bin/hashcat hashcat
ln -s ../../Cellar/hashcat/3.10/share/doc/hashcat hashcat
ln -s ../Cellar/hashcat/3.10/share/hashcat hashcat
==> Summary
🍺  /usr/local/Cellar/hashcat/3.10: 542 files, 49.9M, built in 24 seconds
Josephs-MacBook-Pro:~ joe$ brew test -v hashcat
Testing hashcat
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160907-52880-btkv43.sb /usr/local/Cellar/ruby/2.3.1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/hashcat.rb -v
==> ./hashcat -m 0 --potfile-disable --quiet 098f6bcd4621d373cade4e832627b4f6 /tmp/hashcat-test-20160907-52881-1s6hsls/my.dict


ERROR: clBuildProgram(): CL_INVALID_BUILD_OPTIONS


Error: hashcat: failed
<0> expected but was
<nil>.
```
So, there's that.

zmwangx:
`depends_on :macos => :yosemite` it is.

zmwangx:
@cprecioso You haven't responded to our request for a while, but the formula is nice and I really want to get it merged, so I took the liberty to push the `depends_on :macos => :yosemite` change discussed above to this branch. (I have push access because you selected ["Allow edits from maintainers"](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) when submitting this PR, which is a [relatively new feature of GitHub](https://github.com/blog/2247-improving-collaboration-with-forks).) Rest assured, your contribution is still yours.

cprecioso:
Yes, don't worry. I've been stumped with coursework and totally forgot
about it. Thanks!

zmwangx:
Merged as 463f077. Thank you for your contribution to Homebrew!

PR #4402, opened 2016-09-01
gtk+3: add recommended dependency on librsvg

aleb:

---

Many GTK+3 apps use SVG icons, for example Builder, Pitivi, etc.

See for example https://packages.debian.org/sid/libgtk-3-0

### DIFF
--- a/Formula/gtk+3.rb
+++ b/Formula/gtk+3.rb
@@ -16,6 +16,10 @@ class Gtkx3 < Formula
 
   depends_on "pkg-config" => :build
   depends_on "gdk-pixbuf"
+  # Without this gdk-pixbuf cannot be used to load SVG icons, for example.
+  # Technically librsvg depends on gdk-pixbuf and not the other way around,
+  # because it declares its availability when it's installed.
+  depends_on "librsvg" => :recommended
   depends_on "atk"
   depends_on "gobject-introspection"
   depends_on "libepoxy"


DomT4:
You've introduced a potential circular dependency, since `librsvg` has an `:optional` dep on `gtk+3`...

MikeMcQuaid:
Passing on this for the reason that it introduces that dependency. We would consider switching the dependency around and/or adding a comment here to explain why.

aleb:
Good catch..
Librsvg depends on gtk+3 because it contains a rsvg-view-3 tool "to view a SVG file inside of a GtkWindow". I'd ignore that and remove the gtk+3 dependency from librsvg. What do you prefer?

tdsmith:
-1 if this doesn't change the gtk+3 build. This adds a little more effort to each GTK app that uses librsvg but that feels "honest" to me. Let us know if we're wrong. ;)

MikeMcQuaid:
> -1 if this doesn't change the gtk+3 build. This adds a little more effort to each GTK app that uses librsvg but that feels "honest" to me. Let us know if we're wrong. ;)

:+1: from me, good point @tdsmith

MikeMcQuaid:
(:+1: to your -1, that is 😆 )

aleb:
It might be "honest" but maybe less practical. Every GTK app will bang head against this.

For the record:
https://packages.debian.org/sid/libgtk-3-0
https://www.archlinux.org/packages/extra/x86_64/gtk3/
Not sure what Fedora does, but for sure http://pkgs.fedoraproject.org/cgit/rpms/pitivi.git/tree/pitivi.spec does not depend on librsvg. I can't find an easy way to dig here, I'll leave it at that.

PR #4422, opened 2016-09-01
llvm: tie exception handling to RTTI

### LINKED ISSUE #4400: llvm --without-rtti fails to build
`brew install llvm --without-rtti` errors out at the configuration stage.

```
$ brew install -s llvm --without-rtti
==> Using the sandbox
...
CMake Error at cmake/modules/HandleLLVMOptions.cmake:627 (message):
  Exception handling requires RTTI.  You must set LLVM_ENABLE_RTTI to ON
  CMakeLists.txt:415 (include)
```

https://gist.github.com/64db6e1cf71d0ecc09438f8b9809bc90

Looks like this is because we're enabling exception handling unconditionally, but it requires RTTI. So exception handling should probably be turned off as part of the `--without-rtti` option.

apjanke:

---

Prevents --without-rtti from breaking the build, since EH depends on
RTTI.

Fixes #4400.

### DIFF
--- a/Formula/llvm.rb
+++ b/Formula/llvm.rb
@@ -129,7 +129,7 @@ class Llvm < Formula
   option "without-lld", "Do not build LLD linker"
   option "with-lldb", "Build LLDB debugger"
   option "with-python", "Build bindings against custom Python"
-  option "without-rtti", "Build without C++ RTTI"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍 🚢

apjanke:
Merged. (No bottle update since this only affects non-default builds.)

PR #4439, opened 2016-09-02
Gecode fixes

liuq:

---

The SHA256 signature of the source tar.gz file was wrong and it has been fixed.
Moreover, a new option "with-qt" has been added to allow the compilation of
the Qt based graphical search inspector "gist".


DomT4:
> The SHA256 signature of the source tar.gz file was wrong and it has been fixed.

Can you find out why it changed? Thanks. It will have been correct whenever the change was merged.

> Moreover, a new option "with-qt" has been added to allow the compilation of
the Qt based graphical search inspector "gist".

Can it use Qt5? We'd prefer not to add any more `qt` dependencies because it's very likely `qt` won't work on Sierra & we aren't intending to carry another hacky patch for another year.

MikeMcQuaid:
@liuq Any news on this?

liuq:
I cannot deploy before testing with QT5 (as requested), but homebrew QT5 is currently broken.

ilovezfs:
@liuq https://github.com/Homebrew/homebrew-core/pull/5619 is sufficient to build if you have the patience for a qt5 build.

MikeMcQuaid:
We won't be adding any more `qt` dependencies so happy to make this PR Qt-less for now.

liuq:
@MikeMcQuaid In general I would agree, however the value of Gecode w.r.t the rsystems is in its (qt-based) debugging tools. I try to check with the #5619 if I can get a working qt5 environment.

MikeMcQuaid:
> I try to check with the #5619 if I can get a working qt5 environment.

Thanks!

liuq:
That's it, the current formula is qt5 compliant.

Moisan:
I wrote [upstream](http://www.gecode.org/bugs.html) to ask if the change to the source without version change was intentional.

liuq:
Yes, it is, the version is the very same but, probably, the source file on
the repository has changed (and the SHA signature accordingly).

Il giorno lun 10 ott 2016 alle 16:23 Thierry Moisan <
notifications@github.com> ha scritto:

MikeMcQuaid:
I'd rather we wait for a response first to ensure they've not been hacked.

liuq:
@MikeMcQuaid will you handle that? In case you can write to Christian Schulte cschulte@kth.se, who is one of the main gecode maintainers

ilovezfs:
@BrewTestBot test this please

Moisan:
I've got a [response from upstream](http://www.gecode.org/pipermail/bugs/2016-October/000336.html). 

> this was not intentional, but rather a side effect of updating one of the binary packages (which caused a new version of the source packages to be uploaded as well).  There has not been any change to the source code itself so it is safe to update the SHA256 to the new one.

MikeMcQuaid:
@Moisan Thanks!

Formula/gecode.rb
```diff
@@ -12,8 +12,21 @@ class Gecode < Formula
...
 
+  depends_on "qt5" => :optional
+
   def install
-    system "./configure", "--prefix=#{prefix}", "--disable-examples"
+    args = %W[
+      --prefix=#{prefix}
+      --disable-examples
+    ]
+    ENV.append "CXXFLAGS", "-std=c++11"
```
MikeMcQuaid: Instead use `ENV.cxx11` and see if that's sufficient.

Formula/gecode.rb
```diff
@@ -12,8 +12,21 @@ class Gecode < Formula
...
+    if build.with? "qt5"
+      args << "--enable-qt"
+      ENV.append_path "PKG_CONFIG_PATH", "#{HOMEBREW_PREFIX}/opt/qt5/lib/pkgconfig"
+    else
+      args << "--disable-qt"
+    end
+    system "./configure", *args
     system "make", "install"
   end
 end
```
MikeMcQuaid: Any chance you could add a simple test e.g. ideally something more substantial than e.g. `--version` or `--help` but just use them if necesssary? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

MikeMcQuaid (changes requested):

Formula/gecode.rb
liuq: Done!

Formula/gecode.rb
```diff
@@ -12,8 +12,78 @@ class Gecode < Formula
...
+        opt.parse(argc, argv);
+        Script::run<Test, DFS, Options>(opt);
+        return 0;
+      }
+    EOS
+  
+    args = %W[
+      -std=c++11
+      -I#{HOMEBREW_PREFIX}/opt/qt5/include
+      -I#{HOMEBREW_PREFIX}/include
```
MikeMcQuaid: `-I#{include}`

Formula/gecode.rb
```diff
@@ -12,8 +12,78 @@ class Gecode < Formula
...
+      -lgecodedriver -lgecodegist -lgecodesearch -lgecodeint -lgecodekernel -lgecodesupport
+      -L#{lib}
+      -o test
+    ]
```
MikeMcQuaid: Just hard-code these args (but split them over multiple lines to avoid exceeding 80 characters).

Formula/gecode.rb
```diff
@@ -12,8 +12,78 @@ class Gecode < Formula
...
+    system ENV.cxx, "test.cpp", *args
+    assert File.exist?("test")
+    system "./test | grep -q \"{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}\""
```
MikeMcQuaid: Replace the above two lines with `assert_match "{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}", shell_output("./test")`

MikeMcQuaid (changes requested):

Formula/gecode.rb
liuq: Ok, did it!

MikeMcQuaid (approved):

MikeMcQuaid:

liuq:
It's a pleasure to contribute to the community.

Il giorno mer 12 ott 2016 alle 18:16 Mike McQuaid <notifications@github.com>
ha scritto:

PR #4442, opened 2016-09-02
Collectd: remove openssl dependency

rubenk:


DomT4:
You'll need to elaborate here. Our audit shows this is still necessary.

tdsmith:
Ahahaha, what? The bottle binds to the system libressl and not our openssl.

```
$ brew linkage --reverse collectd
...
/usr/lib/libcrypto.35.dylib
  lib/collectd/snmp.so
```

tdsmith:
So this PR is one the one hand correct as provided, but on the other hand exposes that we need to wrangle collectd's build a little harder.

tdsmith:
This is probably inherited somehow from the system libsnmp, which similarly links the system libressl libcrypto.

tdsmith:
Indeed:
```
tim@rocketman:~$ /usr/bin/net-snmp-config --libs
-lnetsnmp -lcrypto.35
```

We probably can't rely on substituting our own openssl here since I think we want to match the libsnmp linkage.

I submit that either
a) we ignore the audit, or
b) we add a dependency on Homeberw net-snmp, or
c) we see if we can remove the potentially redundant? libressl link from snmp.so (how do linkers work), or
d) ?

b is probably the best answer for OS X < 10.11 since that libcrypto is old and busted™.

DomT4:
Oh, fun. We should probably decide whether we want to flag system LibreSSL linkage as well soon, I guess. It's a newer library but I don't have a lot of faith in Apple to transparently keep an external crypto library updated 🙈.

These kinds of things are going to creep up for as long as we allow system library usage though, which there's no desire to change in Homebrew, so potentially worth starting a commented whitelist for the check in `formula_cellar_checks`.

> b is probably the best answer for OS X < 10.11 since that libcrypto is old and busted™.

Agree that's the best immediate solution.

tdsmith:
Dropping -lcrypto.35 from net-snmp-config allows collectd to build without an explicit link because it uses `-flat_namespace -undefined suppress`[1] but I'm not sure the snmp functionality works. I think there's an implication that libressl symbols are leaking out of libnet-snmp?

[1] https://stackoverflow.com/questions/17225464/using-a-shared-library-linked-with-other-shared-libs

MikeMcQuaid:
> It's a newer library but I don't have a lot of faith in Apple to transparently keep an external crypto library updated 🙈.

Let's keep by the Homebrew ethos and avoid duplicating system dependencies unless we know it's going to definitely solve the problem. If they do update their system LibreSSL it's tempted to fall back to using that instead of OpenSSL where we can.

DomT4:
It can be quite difficult to say when/if/how much Apple updates their external crypto libraries. Theoretically, the base version of OpenSSL they're using is 0.9.8zh, but in reality they've backported several security fixes from the newer branches so it's not a pure 0.9.8zh.

The base version of LibreSSL they're using is 2.2.7, from back in May this year. 2.2.9 is the newest release on that branch, but there's also the 2.3.x branch, and the upstream-recommended 2.4.x stable branch. 2.2.x is no longer supported by OpenBSD at all. I presume they're still using a pure 2.2.7 based on the security release notes since.

Potentially worth noting is that Apple haven't shown any interest in shipping the LibreSSL headers, so it's a very incomplete development solution, and I think if we tried to use the system we'd end up:

* Not really knowing which version of which crypto library we were building against, and which security vulnerabilities, if any, we were entertaining by doing so.
* Having to use our OpenSSL/LibreSSL 3/4 of the time anyway because an awful lot wants the headers, but that mix-and-match confusing contributors.
* Upstream developers having picked up features from the newer LibreSSL branches that aren't available back to 2.2.7 & having to force our dependency there anyway.

I'm with you on using the system as much as possible, but I don't think it's a particularly sensible avenue when it comes to external crypto libraries that Apple has repeatedly demonstrated they're only mildly interested in keeping up-to-date _at some point_ on security fixes, let alone features.

MikeMcQuaid:
> Potentially worth noting is that Apple haven't shown any interest in shipping the LibreSSL headers, so it's a very incomplete development solution

Ok, that's fine/good to know 👍

tdsmith:
> We should probably decide whether we want to flag system LibreSSL linkage as well soon, I guess.

https://github.com/Homebrew/brew/pull/866; I think my regex was just accidentally too conservative.

rubenk:
Hi guys, thanks for looking into this. Indeed, net-snmp-config is leaking internal details and is next to useless. The flags it returns are how net-snmp was compiled/linked, not what you need to specify to link against it. I'll see if I can do something in upstream collectd to work around this as well.

PR #4448, opened 2016-09-02
crystal-lang 0.19.0

asterite:

---

Thank in you advance! 🙇

### DIFF
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -6,25 +6,8 @@ class CrystalLang < Formula
   head "https://github.com/crystal-lang/crystal.git"
 
   stable do
-    url "https://github.com/crystal-lang/crystal/archive/0.18.7.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/crystal-lang.rb
```diff
@@ -6,25 +6,8 @@ class CrystalLang < Formula
   head "https://github.com/crystal-lang/crystal.git"
 
   stable do
```
ilovezfs: the stable do block is no longer needed since the patches are being removed, so the url and sha256 can be moved right above the head url again

asterite: Ah, thank you. I had no idea what that did because I didn't put it there in the first place ( :-( ). I'll make the change right now and rebase this PR.

ilovezfs: Cool, thanks. It scopes whatever is inside it to the particular software spec (stable, devel, or head).

asterite: @ilovezfs Thank you for the explanation. I created a separate PR here: https://github.com/Homebrew/homebrew-core/pull/4449

PR #4449, opened 2016-09-02
crystal-lang 0.19.0

asterite:

---

I had to rebase or fetch or something and since I'm really bad at git I decided to start from scratch. Here's another PR, hopefully with everything in its right place :-)

### DIFF
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -1,31 +1,9 @@
 class CrystalLang < Formula
   desc "Fast and statically typed, compiled language with Ruby-like syntax"
   homepage "https://crystal-lang.org/"
-  revision 1
-
+  url "https://github.com/crystal-lang/crystal/archive/0.19.0.tar.gz"
+  sha256 "4d7d9770891bd8bd835251e2654316412a3f44074db0adeca28357c0993eb2d8"
   head "https://github.com/crystal-lang/crystal.git"
-
-  stable do
-    url "https://github.com/crystal-lang/crystal/archive/0.18.7.tar.gz"
-    sha256 "72b33fd4bd903a9f0957c74c2f95429e8c0f94c04df86396711b5560f540805d"
-
-    # LLVM 3.8 support patches begin.
-    patch do
-      url "https://github.com/crystal-lang/crystal/commit/53f1977c8c.patch"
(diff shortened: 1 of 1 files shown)


Formula/crystal-lang.rb
```diff
@@ -1,31 +1,10 @@
...
-      sha256 "f98b92d9003d2fa69fe36e218329cc616198e4a6283ba81b7c27e84d9d4afac0"
-    end
-
-    patch do
-      url "https://github.com/crystal-lang/crystal/commit/63ca4950e4.patch"
-      sha256 "bb0cbb466673f7f04996326c875c8c7f9a0335e31a3bfab1576358fdf8a697f1"
-    end
-    # LLVM 3.8 support patches end.
-  end
+  revision 1
```
ilovezfs: the revision line should be removed since this is a brand new version (revision counting starts from 0 but we never put `revision 0` explicitly)

asterite: Ah, OK.

Then why this PR was accepted, the one that introduced that revision? https://github.com/Homebrew/homebrew-core/commit/f1d68e0bf53e3be40be86575415652153c7e6862

Is there anything I can do so that only me can send PR to this formula?

asterite: Updated! :-)

ilovezfs: Thanks @asterite 

The revision forces an upgrade for users so that upgrades can be induced in between actual version bumps.

ilovezfs:
>Is there anything I can do so that only me can send PR to this formula?

@asterite we don't really have any mechanism like that per se, but we try to report any patching that is necessary to fix the build to upstreams, and we don't do any "feature patching."

For upstreams that want to tightly control distribution (I totally understand the desire), Homebrew Cask is usually a better fit since it distributes (often signed) binaries prebuilt by upstream (usually in the form of a app, dmg, or pkg).

asterite:
@ilovezfs I see, thanks. More than control, maybe I'd like to be able to review some changes, at least to understand them. I'll subscribe to this project notifications and try to use a mail filter.

ilovezfs:
>I'll subscribe to this project notifications and try to use a mail filter.

Be sure to make it case insensitive! hehe

tdsmith:

PR #3096, opened 2016-07-18
xcenv 1.0.0 (new formula)

maschall:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

PR #4457, opened 2016-09-03
xcenv 1.0.4

maschall:

---

Resubmitting after some gained attention (https://iosdevweekly.com/)

### DIFF
--- a/Formula/xcenv.rb
+++ b/Formula/xcenv.rb
@@ -1,0 +1,22 @@
+class Xcenv < Formula
+  desc "Xcode version manager"
+  homepage "https://github.com/xcenv/xcenv#readme"
+  url "https://github.com/xcenv/xcenv/archive/v1.0.4.tar.gz"
+  sha256 "cac6e5475b0dda35fe1c3771bc60b703e0df7b87b1bf607988c447768ca02122"
+  head "https://github.com/xcenv/xcenv.git"
+
+  def install
+    prefix.install ["bin", "libexec"]
+  end
+
+  def caveats; <<-EOS.undent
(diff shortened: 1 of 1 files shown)


Formula/xcenv.rb
```diff
@@ -0,0 +1,24 @@
...
+  def install
+    prefix.install ["bin", "libexec"]
+  end
+
+  def caveats; <<-EOS.undent
+    Xcenv stores data under ~/.xcenv by default. If you absolutely need to
+    store everything under Homebrew's prefix, include this in your profile:
+      export XCENV_ROOT=#{var}/xcenv
+    To enable shims and autocompletion, run this and follow the instructions:
+      xcenv init
```
dunn: This second part isn't Homebrew-specific information, is it?  That's all we allow in the caveats.

maschall: The tool is very much based off rbenv, so I followed the rbenv formula (https://github.com/Homebrew/homebrew-core/blob/master/Formula/rbenv.rb).

I am happy to remove it, if it's not allowed.

It is useful information for the user on how to further setup the tool, but I can see what you mean.

dunn: Yeah, let's remove it.  We should probably drop the same line from rbenv.

The rationale is that (1) upstream might change how it works and we'll be giving people old information and (2) the more text in the caveats, the less likely they'll be read.

maschall: Makes sense, removed

tdsmith:
previously https://github.com/Homebrew/homebrew-core/pull/3096

tdsmith:

PR #4465, opened 2016-09-03
minio: add launchd init script for running minio as a daemon

ghost:
add launchd init script for running minio as a daemon


Formula/minio.rb
```diff
@@ -15,6 +15,14 @@ class Minio < Formula
 
   depends_on "go" => :build
 
+  def configdir
+    etc/"minio"
+  end
+
+  def datadir
+    var/"minio"
+  end
```
MikeMcQuaid: Just inline these values rather than declaring functions.

Formula/minio.rb
```diff
@@ -41,6 +49,38 @@ def install
     bin.install buildpath/"minio"
   end
 
+  def post_install
+    # Make sure the datadir exists
```
MikeMcQuaid: Can remove this comment.

MikeMcQuaid:
Squash commits by running `brew update; git checkout $YOUR_BRANCH; git rebase --interactive origin/master` and then mark each commit other than the first as `squash` or `fixup` then `git push -f`. Thanks!

PR #4482, opened 2016-09-03
box2d: migrate to boneyard

MikeMcQuaid:

---

Checksum is broken and they aren't stating why which means they may have been e.g. hacked.

CC @lembacon https://github.com/Homebrew/homebrew-boneyard/pull/155

Closes #3834.


DomT4:
👍

Timmmm:
This is rather annoying. The source download is dynamically generated by github from git. Unless they have worked out how to break git's security model (extremely unlikely) they haven't been hacked, it is just that github has changed how they generate tar.gz files. Perhaps they switched to zopfli, or even just changed the compression settings.

A true solution is to either: a) download the data from git and verify the git commit hash, b) unzip the .tar.gz file and then calculate the checksum of the actual files, or c) get the box2d guys to upload a tar.gz as an actual attachement. That should give it a stable checksum.

It looks like brew doesn't support a) or b), so c) is probably the easiest option.

MikeMcQuaid:
I agree, it's annoying when upstream software changes their checksums and won't answer questions about it. I recommend you maintain this in your own tap if you'd like to use it.

Timmmm:
Upstream didn't change any checksums. I don't think you can blame them for
Brew's incorrent (but understandable) assumption that the checksum of a
tarball is not independent from the checksum of its content.

MikeMcQuaid:
@Timmmm you're being a little disrespectful at this point. We removed this package months ago due to an upstream unresponsive to legitimate security concerns. No-one is blaming anyone here but we can't package their software if they are unresponsive to these concerns. Nothing indicates to me that this has changed.

PR #4484, opened 2016-09-03
gst-plugins-*: recommend jpeg dependency.

MikeMcQuaid:

---

They picks this up opportunistically so makes sense to just enable.


DomT4:
Might be worth a revision bump. We usually do when adding new `:recommended` stuff.

You've changed some patch whitespacing as well. Probably worth reverting that as it usually manages to break something for someone 😢.

MikeMcQuaid:
@DomT4 Cheers. ⅔ 😉

Formula/gst-plugins-ugly.rb
```diff
@@ -3,6 +3,7 @@ class GstPluginsUgly < Formula
   homepage "https://gstreamer.freedesktop.org/"
   url "https://gstreamer.freedesktop.org/src/gst-plugins-ugly/gst-plugins-ugly-1.8.3.tar.xz"
   sha256 "6fa2599fdd072d31fbaf50c34af406e2be944a010b1f4eab67a5fe32a0310693"
+  revision 2
```
DomT4: Bumped the revision too hard here 🙈

PR #4498, opened 2016-09-04
zurl 1.6.0

jkarneges:

---

`brew audit` complained about linking to legacy openssl, so dependency on brew openssl was added. Also since zurl shares openssl data with libcurl, a dependency on brew curl was added too.

### DIFF
--- a/Formula/zurl.rb
+++ b/Formula/zurl.rb
@@ -1,8 +1,8 @@
 class Zurl < Formula
   desc "HTTP and WebSocket client worker with ZeroMQ interface"
   homepage "https://github.com/fanout/zurl"
-  url "https://dl.bintray.com/fanout/source/zurl-1.5.1.tar.bz2"
-  sha256 "945fccc160ea655249e7dd3706a5db096bab6d1d08fc6e290e53a9635f07c7d4"
+  url "https://dl.bintray.com/fanout/source/zurl-1.6.0.tar.bz2"
+  sha256 "eb30ff825f407311357b1ffb3ea220957fa932b3d61d2937f3e334e4ea346dd5"
 
   bottle do
     cellar :any
@@ -21,9 +21,17 @@ class Zurl < Formula
     sha256 "2dafa322670a94e20283aba2a44b92134d425bd326419b68ad4db8d0831a26ec"
   end
 
+  stable do
+    patch do
+      url "https://github.com/fanout/zurl/commit/de3c0f27cc30e9815f2f28160578a2928118ffd0.diff"
+      sha256 "a6930cdc5e399867a1abb53efd2857249f7ae7f801e7ab1106c63667e6f56fa6"
+    end
+  end
+
   def install
     system "./configure", "--prefix=#{prefix}", "--extraconf=QMAKE_MACOSX_DEPLOYMENT_TARGET=#{MacOS.version}"
     system "make"
+    system "make", "check"
     system "make", "install"
   end
 


ilovezfs:
This seems to build fine only adding the homebrew openssl dependency. I'm not sure I'm following why we also need to add a homebrew curl dependency.

jkarneges:
The problem is libcurl also links against openssl, and so if the system libcurl is used then zurl would end up linking to two different openssl versions at the same time.

ilovezfs:
```
iMac-TMP:~ joe$ otool -L  /usr/lib/libcurl.4.dylib
/usr/lib/libcurl.4.dylib:
	/usr/lib/libcurl.4.dylib (compatibility version 7.0.0, current version 8.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 57332.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1231.0.0)
	/System/Library/Frameworks/LDAP.framework/Versions/A/LDAP (compatibility version 1.0.0, current version 2.4.0)
	/System/Library/Frameworks/Kerberos.framework/Versions/A/Kerberos (compatibility version 5.0.0, current version 6.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1225.0.0)
```
It does?

jkarneges:
Interesting, I'd forgotten curl supported different ssl backends other than openssl. This means the custom host verification code won't work since it depends on `CURLOPT_SSL_CTX_FUNCTION`, but it shouldn't crash or anything. System libcurl should be fine then. And in fact there's not much point for zurl to depend on openssl if that code will never run anyway, so I'll see if I can remove the openssl dependency.

ilovezfs:
>custom host verification code won't work since it depends on CURLOPT_SSL_CTX_FUNCTION

Is that a valuable feature? I don't mind adding the curl dependency as long as we're all on the same page about why it's being added. Also you'd need to do
```
depends_on "curl" => "with-openssl"
```
since openssl is an optional (i.e., non-default), not recommended (i.e., default/what's-in-the-curl-bottle), option so it would need to build curl from scratch on the user's machine before installing the zurl bottle unless the user happened to already have a `with-openssl` curl in place for whatever preexisting reason. As you can see that's pretty rare:
```
Josephs-MacBook-Pro:Formula joe$ ag curl|grep openssl
shibboleth-sp.rb:15:  depends_on "curl" => "with-openssl"
xml-tooling-c.rb:20:  depends_on "curl" => "with-openssl"
```

jkarneges:
`zurl` needs direct access to openssl in order to implement the `trust-connect-host` option, which makes it possible to trust not only a certificate that matches the host of the request URI, but also a certificate that matches a specified connect host (set via the `connect-host` option). This is mainly useful when using `pushpin` (which depends on `zurl`) to proxy custom domain requests over the internet to a provider that uses a multi-tenant SSL cert (e.g. Heroku).

IMO this feature is not valuable enough to cause curl to need to be compiled for all users. If it were possible to have a `with-openssl` curl bottle available then maybe the dependency would be fine.

MikeMcQuaid:
> The problem is libcurl also links against openssl

The system curl doesn't, actually.

MikeMcQuaid:
> IMO this feature is not valuable enough to cause curl to need to be compiled for all users. If it were possible to have a with-openssl curl bottle available then maybe the dependency would be fine.

Let's disable this feature by default 👍

jkarneges:
Patched zurl so it doesn't require openssl, which in turn means it doesn't require a libcurl with openssl.

MikeMcQuaid:

PR #4505, opened 2016-09-04
mpv: replace docutils resource with a buildtime dep

zmwangx:

---

The mpv formula predates the docutils formula in core (the former was incorporated in December 2015 and the latter in May 2016). Now that we have a standalone formula for docutils, there's no reason to vendor docutils as a resource. Note that the currently vendored version is the same as the currently bottled version, so there shouldn't be any difference.

### DIFF
--- a/Formula/mpv.rb
+++ b/Formula/mpv.rb
@@ -14,6 +14,7 @@ class Mpv < Formula
   option "with-bundle", "Enable compilation of the .app bundle."
 
   depends_on "pkg-config" => :build
+  depends_on "docutils" => :build
   depends_on :python3
 
   depends_on "libass"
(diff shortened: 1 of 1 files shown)


zmwangx:
`varnish` could probably benefit from a similar switch. There are a few more poet-style formulae vendoring docutils, but apparently no one would bother converting those.

DomT4:
I presume there's no issue with `docutils` using Python 2.x & this formula Python3? I can't see why there would be in this case, but I don't really use Python at all myself.

zmwangx:
I believe all you need here from `docutils` is `rst2man.py`, which is a self-contained executable with the `/usr/bin/python` shebang, so you don't need to care about the Python version. See https://github.com/mpv-player/mpv/blob/f42e4374d55a3b68b0c4fcb342d9cdcd5d15c9c7/wscript_build.py#L26-L34.

zmwangx:
Submitted #4554 for `varnish`, too.

ilovezfs:
@zmwangx :rocket:

PR #4511, opened 2016-09-04
Courier Unicode 1.4 (new formula)

anthonygreen:
A dependency of maildrop http://www.courier-mta.org/maildrop/

### DIFF
--- a/Formula/courier-unicode.rb
+++ b/Formula/courier-unicode.rb
@@ -1,0 +1,43 @@
+class CourierUnicode < Formula
+  desc "Implements several algorithms related to the Unicode Standard"
+  homepage "http://www.courier-mta.org/unicode/"
+  url "https://downloads.sourceforge.net/project/courier/courier-unicode/1.4/courier-unicode-1.4.tar.bz2"
+  sha256 "2174f4cdd2cd3fe554d4cbbd9557abac0e54c0226084f368bcb2e66b0e78cf96"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+
+    # https://sourceforge.net/p/courier/mailman/message/34719502/
+    inreplace "courier-unicode.h", "extern const char *unicode_locale_chset_l(locale_t l);", ""
+    system "make"
+    system "make", "install"
+  end
+
+  # https://sourceforge.net/p/courier/mailman/message/34719502/
+  patch :p0, :DATA
+
+  test do
+    system "false"
+  end
+end
+
+__END__
+--- unicode.c
++++ unicode.c
+@@ -37,13 +37,6 @@
+ 	return fix_charset(c);
+ }
+ 
+-#if HAVE_LANGINFO_L
+-const char *unicode_locale_chset_l(locale_t l)
+-{
+-	return fix_charset(nl_langinfo_l(CODESET, l));
+-}
+-#endif
+-
+ static const char *fix_charset(const char *c)
+ {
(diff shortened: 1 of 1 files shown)


Formula/courier-unicode.rb
```diff
@@ -0,0 +1,43 @@
...
+    inreplace "courier-unicode.h", "extern const char *unicode_locale_chset_l(locale_t l);", ""
+    system "make"
+    system "make", "install"
+  end
+
+  # https://sourceforge.net/p/courier/mailman/message/34719502/
+  patch :p0, :DATA
+
+  test do
+    system "false"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Formula/courier-unicode.rb
```diff
@@ -0,0 +1,43 @@
...
+                          "--prefix=#{prefix}"
+
+    # https://sourceforge.net/p/courier/mailman/message/34719502/
+    inreplace "courier-unicode.h", "extern const char *unicode_locale_chset_l(locale_t l);", ""
+    system "make"
+    system "make", "install"
+  end
+
+  # https://sourceforge.net/p/courier/mailman/message/34719502/
+  patch :p0, :DATA
```
MikeMcQuaid: It's not clear to me from these message: was this patch (and the inreplace) merged?

anthonygreen: @MikeMcQuaid the patches never seem to have been merged. The last commit to SourceForge was in Sept '15 and the thread dates from Dec '15

Formula/courier-unicode.rb
anthonygreen: @svarshavchik can you suggest a suitable test?

svarshavchik: Anthony Green writes:

> « HTML content follows »
> In <URL:https://github.com/Homebrew/homebrew-core/pull/ 
> 4511#discussion_r77473991>Formula/courier-unicode.rb:
> > +    system "./configure", "--disable-debug",
> > +                          "--disable-dependency-tracking",
> > +                          "--disable-silent-rules",
> > +                          "--prefix=#{prefix}"
> > +
> > +    # https://sourceforge.net/p/courier/mailman/message/34719502/
> > +    inreplace "courier- 
> unicode.h", "extern const char *unicode_locale_chset_l(locale_t l);", ""
> > +    system "make"
> > +    system "make", "install"
> > +  end
> > +
> > +  # https://sourceforge.net/p/courier/mailman/message/34719502/
> > +  patch :p0, :DATA
> > +
> > +  test do
> > +    system "false"
>
> <URL:https://github.com/svarshavchik>@svarshavchik can you suggest a suitable  
> test?

I'm not sure I follow. To test unicode_local_chset_l(), use newlocale() to  
create a locale with a specific character set, which unicode_local_chset_l()  
should return.

Formula/courier-unicode.rb
MikeMcQuaid: I think we'll probably pass on this as it seems to be unmaintained.

anthonygreen: @MikeMcQuaid the project has moved to Github https://github.com/svarshavchik/courier-libs but I had issues attempting to use that source in a Homebrew formula. Additionally I couldn't get maildrop to compile on OSX and as reason for adding courier-libs was as a upstream dependency I figure we should pass until those issues are resolved

anthonygreen:
Closing pull request until such time as issues with the library and maildrop are resolved

PR #4517, opened 2016-09-04
update the sha256 for cli53

tomschlick:
Was getting this

```
Error: SHA256 mismatch
Expected: 1efd7b951a74ccc31b3392e19955eaa8c336b35815e24118103155c5b4a9a4a2
Actual: 0a87de05110d9bba851ba3522a1072494256e31fa653f59331313bff21d5c160
Archive: /Users/tom/Library/Caches/Homebrew/cli53-0.8.0.tar.gz
```

### DIFF
--- a/Formula/cli53.rb
+++ b/Formula/cli53.rb
@@ -2,7 +2,7 @@ class Cli53 < Formula
   desc "command-line tool for Amazon Route 53"
   homepage "https://github.com/barnybug/cli53"
   url "https://github.com/barnybug/cli53/archive/0.8.0.tar.gz"
-  sha256 "1efd7b951a74ccc31b3392e19955eaa8c336b35815e24118103155c5b4a9a4a2"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Can you open an issue upstream asking them why it changed? Want to make sure they weren't e.g. hacked.

tomschlick:
Yeah will do

tomschlick:
Nevermind, looks like 0.8.1 is out as of a few hours ago so I'll just create a PR for that one

MikeMcQuaid:
@tomschlick I think it's even more important given that in case they've been hacked and pushed out a non-legit version.

tomschlick:
@barnybug can you confirm that this was done intentionally?

barnybug:
yes

ilovezfs:
@barnybug thanks!!!

PR #4519, opened 2016-09-04
nginx: have head depends on openssl@1.1

JLHwung:
Add build with openssl@1.1 support, there will be an option parsed as `--with-openssl@1.1` on `brew info nginx`. It looks a little bit creepy but I don't know how to customize the parsed option.

Now it can only be built on `--HEAD`, so one can looking forward to build on 1.11.4.

---

### DIFF
--- a/Formula/nginx.rb
+++ b/Formula/nginx.rb
@@ -28,9 +28,8 @@ class Nginx < Formula
   deprecated_option "with-spdy" => "with-http2"
 
   depends_on "pcre"
+  depends_on "openssl"
   depends_on "passenger" => :optional
-  depends_on "openssl" => :recommended
-  depends_on "libressl" => :optional
 
   def install
     # Changes default port to 8080
@@ -41,19 +40,13 @@ class Nginx < Formula
 
     pcre = Formula["pcre"]
     openssl = Formula["openssl"]
-    libressl = Formula["libressl"]
 
     if build.head?
       openssl = Formula["openssl@1.1"]
     end
 
-    if build.with? "libressl"
-      cc_opt = "-I#{pcre.include} -I#{libressl.include}"
-      ld_opt = "-L#{pcre.lib} -L#{libressl.lib}"
-    else
-      cc_opt = "-I#{pcre.include} -I#{openssl.include}"
-      ld_opt = "-L#{pcre.lib} -L#{openssl.lib}"
-    end
+    cc_opt = "-I#{pcre.include} -I#{openssl.include}"
+    ld_opt = "-L#{pcre.lib} -L#{openssl.lib}"
 
     args = %W[
       --prefix=#{prefix}
--- a/Formula/nginx.rb
+++ b/Formula/nginx.rb
@@ -43,6 +43,10 @@ class Nginx < Formula
     openssl = Formula["openssl"]
     libressl = Formula["libressl"]
 
+    if build.head?
+      openssl = Formula["openssl@1.1"]
+    end
+
     if build.with? "libressl"
       cc_opt = "-I#{pcre.include} -I#{libressl.include}"
       ld_opt = "-L#{pcre.lib} -L#{libressl.lib}"
--- a/Formula/nginx.rb
+++ b/Formula/nginx.rb
@@ -1,9 +1,13 @@
 class Nginx < Formula
   desc "HTTP(S) server and reverse proxy, and IMAP/POP3 proxy server"
   homepage "https://nginx.org/"
-  url "https://nginx.org/download/nginx-1.10.1.tar.gz"
-  sha256 "1fd35846566485e03c0e318989561c135c598323ff349c503a6c14826487a801"
-  head "http://hg.nginx.org/nginx/", :using => :hg
+
+  stable do
+    url "https://nginx.org/download/nginx-1.10.1.tar.gz"
+    sha256 "1fd35846566485e03c0e318989561c135c598323ff349c503a6c14826487a801"
+
+    depends_on "openssl"
+  end
 
   bottle do
     sha256 "2b67c86454cc67bdeb2a637d9872ae471a810d8a2dae40b6a0c1dad7b253d30d" => :el_capitan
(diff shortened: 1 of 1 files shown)


tdsmith:
If I had my druthers, we would:
- remove the option to build with libressl
- have devel and stable depend on openssl 1.0
- have head depend on openssl 1.1

and get rid of all of the options entirely.

@DomT4 probably has opinions about that. :)

JLHwung:
I totally agree with @tdsmith. I will work on it if @DomT4 also votes up.

tdsmith:
> @tdsmith Would devel depend on openssl 1.1 once a new devel cut from head?

Yes, as soon as it's supported.

DomT4:
Since `nginx` isn't used by anything else, seems fine to start migrating it across. I don't want to make a complete habit of this, but I'm fine with dropping `libressl` here because it seems the majority who care about LibreSSL with `nginx` are using the `nginx-full` formula in the `nginx` tap.

> have devel and stable depend on openssl 1.0

This remains just `depends_on "openssl"` for the benefit of OP. I considered renaming it to `openssl-lts` but I thought that'd probably annoy the heck out of people.

> have head depend on openssl 1.1

Seems reasonable, yeah.

tdsmith:
Cool! Thanks Dom :)

JLHwung:
I have reworked on the branch, marking `openssl` as required because I haven't found any option like `--without-openssl`. I also removes building with libressl.

DomT4:
That isn't quite the way to go about the change. You'll want to do something like:

```diff
diff --git a/Formula/nginx.rb b/Formula/nginx.rb
index 93fc50c..298e204 100644
--- a/Formula/nginx.rb
+++ b/Formula/nginx.rb
@@ -1,9 +1,13 @@
 class Nginx < Formula
   desc "HTTP(S) server and reverse proxy, and IMAP/POP3 proxy server"
   homepage "https://nginx.org/"
-  url "https://nginx.org/download/nginx-1.10.1.tar.gz"
-  sha256 "1fd35846566485e03c0e318989561c135c598323ff349c503a6c14826487a801"
-  head "http://hg.nginx.org/nginx/", :using => :hg
+
+  stable do
+    url "https://nginx.org/download/nginx-1.10.1.tar.gz"
+    sha256 "1fd35846566485e03c0e318989561c135c598323ff349c503a6c14826487a801"
+
+    depends_on "openssl"
+  end
 
   bottle do
     sha256 "2b67c86454cc67bdeb2a637d9872ae471a810d8a2dae40b6a0c1dad7b253d30d" => :el_capitan
@@ -14,6 +18,14 @@ class Nginx < Formula
   devel do
     url "https://nginx.org/download/nginx-1.11.3.tar.gz"
     sha256 "4a667f40f9f3917069db1dea1f2d5baa612f1fa19378aadf71502e846a424610"
+
+    depends_on "openssl"
+  end
+
+  head do
+    url "http://hg.nginx.org/nginx/", :using => :hg
+
+    depends_on "openssl@1.1"
   end
 
   # Before submitting more options to this formula please check they aren't
@@ -29,8 +41,6 @@ class Nginx < Formula
 
   depends_on "pcre"
   depends_on "passenger" => :optional
-  depends_on "openssl" => :recommended
-  depends_on "libressl" => :optional
 
   def install
     # Changes default port to 8080
@@ -40,16 +50,14 @@ class Nginx < Formula
     end
 
     pcre = Formula["pcre"]
-    openssl = Formula["openssl"]
-    libressl = Formula["libressl"]
 
-    if build.with? "libressl"
-      cc_opt = "-I#{pcre.include} -I#{libressl.include}"
-      ld_opt = "-L#{pcre.lib} -L#{libressl.lib}"
+    if build.head?
+      openssl = Formula["openssl@1.1"]
     else
-      cc_opt = "-I#{pcre.include} -I#{openssl.include}"
-      ld_opt = "-L#{pcre.lib} -L#{openssl.lib}"
+      openssl = Formula["openssl"]
     end
+    cc_opt = "-I#{pcre.include} -I#{openssl.include}"
+    ld_opt = "-L#{pcre.lib} -L#{openssl.lib}"
 
     args = %W[
       --prefix=#{prefix}
```

JLHwung:
Thanks @DomT4. Reworked according to the diff.

Formula/nginx.rb
```diff
@@ -40,17 +50,16 @@ def install
...
 
     pcre = Formula["pcre"]
-    openssl = Formula["openssl"]
-    libressl = Formula["libressl"]
 
-    if build.with? "libressl"
-      cc_opt = "-I#{pcre.include} -I#{libressl.include}"
-      ld_opt = "-L#{pcre.lib} -L#{libressl.lib}"
+    if build.head?
+      openssl = Formula["openssl@1.1"]
```
MikeMcQuaid: Could simplify all this to
```ruby
openssl = build.head? ? Formula["openssl@1.1"] : Formula["openssl"]
```

JLHwung: I am good with if-else, as there is no other ternary operators in this file.

MikeMcQuaid: I prefer the ternary but have no strong preference. If you prefer if/else then:

```ruby
openssl = if build.head?
  Formula["openssl@1.1"]
else
  Formula["openssl"]
end
```

DomT4: @MikeMcQuaid I thought you hated ternary operators so I didn't suggest it 😄, but if you're happy I'm happy for OP to use that.

JLHwung: That even looks worse than ternary operator, I will go with your previous suggestion.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/3b58fe0254ae650d6428c4b7dc240100468047f9. Thank you for another contribution to Homebrew @JLHwung; we appreciate it! :smiley_cat:

JLHwung:
@DomT4 Thank you very much for your example. I should have read more about the cookbook. And thank @MikeMcQuaid for code style suggestion.

DomT4:
More than happy to help. It's a big part of what we're here for 😃.

PR #3872, opened 2016-08-14
Update Contributing documents

ma11hew28:
- Add `brew uninstall foo`
- Add `--build-from-source` to `brew install foo`
- Add `--strict --online` to `brew audit foo`

PR #4525, opened 2016-09-05
wego 2.0 (new formula)

zmwangx:

---

The test is currently written to detect the "No API key specified." error in `wego`'s output. In order to have it do anything useful one has to donate a forecast.io or Worldweatheronline API key, which is probably not a good idea to include as clear text in a public formula.

By the way, maybe the process for creating formulae in languages like Go, Haskell and <s>Python</s> (I forgot Python is [documented](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md)) could be documented somewhere? I personally have very limited experience with Go, and had to sift through the issue tracker to find out how the `go_resources` could be generated automatically (fortunately I've read some of the relevant discussions before and thus had some keywords in mind to search), then model the `install` block on existing `language/go`-using formulae (and they are quite inconsistent 😓 ).

### DIFF
--- a/Formula/wego.rb
+++ b/Formula/wego.rb
@@ -1,0 +1,39 @@
+require "language/go"
+
+class Wego < Formula
+  desc "Weather app for the terminal"
+  homepage "https://github.com/schachmat/wego"
+  url "https://github.com/schachmat/wego/archive/2.0.tar.gz"
+  sha256 "d63d79520b385c4ed921c7decc37a0b85c40af66600f8a5733514e04d3048075"
+  head "https://github.com/schachmat/wego.git"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/mattn/go-colorable" do
+    url "https://github.com/mattn/go-colorable.git",
+        :revision => "ed8eb9e318d7a84ce5915b495b7d35e0cfe7b5a8"
+  end
+
+  go_resource "github.com/mattn/go-runewidth" do
+    url "https://github.com/mattn/go-runewidth.git",
+        :revision => "d6bea18f789704b5f83375793155289da36a3c7f"
+  end
+
+  go_resource "github.com/schachmat/ingo" do
+    url "https://github.com/schachmat/ingo.git",
+        :revision => "b1887f863beaeb31b3924e839dfed3cf3a981ea8"
+  end
+
+  def install
+    ENV["GOPATH"] = buildpath
+    (buildpath/"src/github.com/schachmat").mkpath
+    ln_sf buildpath, buildpath/"src/github.com/schachmat/wego"
+    Language::Go.stage_deps resources, buildpath/"src"
+    system "go", "build", "-o", bin/"wego"
+  end
+
+  test do
+    ENV["WEGORC"] = testpath/".wegorc"
+    assert_match /No .*API key specified./, shell_output("#{bin}/wego 2>&1", 1)
+  end
+end


Formula/wego.rb
```diff
@@ -0,0 +1,38 @@
...
+  desc "Weather app for the terminal"
+  homepage "https://github.com/schachmat/wego"
+  url "https://github.com/schachmat/wego/archive/2.0.tar.gz"
+  sha256 "d63d79520b385c4ed921c7decc37a0b85c40af66600f8a5733514e04d3048075"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/mattn/go-colorable" do
+    url "https://github.com/mattn/go-colorable.git",
+    :revision => "ed8eb9e318d7a84ce5915b495b7d35e0cfe7b5a8"
```
ilovezfs: align the ':' under the '"'

ilovezfs:
@zmwangx you may find this of interest: https://github.com/ilovezfs/homebrew-core/blob/03820fe04b258621df6eb1dde499ec02e1818072/Formula/peco.rb#L18-L33

Formula/wego.rb
zmwangx: Is that the new norm? Note that before I submitted, I checked the existing golang formulae, ordered by last update. The three most recent ones — `acmetool`, `termshare` and `gitlab-ci-multi-runner` — all have different indentations... So I just went with what `gdm` emitted (which is consistent with `termshare`, last updated yesterday).

ilovezfs: yes, it's the new norm.

zmwangx: Cool with that (and Emacs is cool with that, too). Updated.

Formula/wego.rb
```diff
@@ -0,0 +1,38 @@
+require "language/go"
+
+class Wego < Formula
+  desc "Weather app for the terminal"
+  homepage "https://github.com/schachmat/wego"
+  url "https://github.com/schachmat/wego/archive/2.0.tar.gz"
+  sha256 "d63d79520b385c4ed921c7decc37a0b85c40af66600f8a5733514e04d3048075"
```
ilovezfs: head spec?

zmwangx:
>  https://github.com/ilovezfs/homebrew-core/blob/03820fe04b258621df6eb1dde499ec02e1818072/Formula/peco.rb#L18-L33

Like the abstraction. Is this a WIP somewhere?

ilovezfs:
@zmwangx I've not opened a `brew` PR for it yet, but some feedback from Mike on it here https://github.com/Homebrew/homebrew-core/pull/4040

ilovezfs:
@zmwangx the test fails for HEAD. Do you mind trying to tweak it so it passes for stable & head so it doesn't break predictably at next upgrade?
```
iMac-TMP:homebrew-core joe$ brew test -v wego
Testing wego
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160904-69435-1rp1gua.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/wego.rb -v --HEAD
==> /usr/local/Cellar/wego/HEAD-a8a0870/bin/wego 2>&1
Error: wego: failed
Expected /No\ API\ key\ specified\./ to match "2016/09/04 21:48:25 No forecast.io API key specified.\nYou have to register for one at https://developer.forecast.io/register\n".
```

zmwangx:
You know, relying on the stability error messages is never a good thing... Anyway, updated match target to a regex.

ilovezfs:
@zmwangx yeah, it tends to break if you look at it funny. Too bad this doesn't have some kind of minimal functionality without an API key.

zmwangx:
Or I could contribute an API key (from a new account) and hope no one abuses it.

ilovezfs:
@zmwangx I wouldn't worry about it in this case, but in general it would be interesting if we added something like a DSL for unlocking keys like this and put the password in an envvar on Jenkins hidden unless you have admin access.

zmwangx:
Sounds cool but then you can't run `brew test` locally.

ilovezfs:
Everything has tradeoffs! Of course "no one" actually runs `brew test` anyway ;)

zmwangx:
> I wouldn't worry about it in this case.

In this case, forecast.io's free tier is rate limited to 1000 API calls per day. Therefore, someone could DoS the key whenever there's a PR concerning `weco`.

> Of course "no one" actually runs brew test anyway ;)

At least PR authors have to run `brew test`. It's in the PR template:

- [ ] Have you built your formula locally prior to submission with brew install <formula> (where <formula> is the name of the formula you're submitting)?
- [ ] Does your submission pass brew audit --new-formula <formula> (after doing brew install <formula>)?

...

Wait, it's not there?

ilovezfs:
>In this case, forecast.io's free tier is rate limited to 1000 API calls per day. Therefore, someone could DoS the key whenever there's a PR concerning wego.

Right, I meant I wouldn't worry about adding a better test that used an API key.

>Wait, it's not there?

"Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?"

And that document references running `brew test` and fixing failing tests but, yes, somewhat ridiculous it's not directly in the template itself. :laughing:

ilovezfs:
@zmwangx you could attempt to delegate a fix to the template here https://github.com/Homebrew/homebrew-core/pull/3872

ilovezfs:
@zmwangx :shipit:

zmwangx:
Good?

zmwangx:
Assuming the answer is positive. 🚢 'ed.

ilovezfs:
yup ribbit = shipit

zmwangx:
Yeah knew that, just want to get a confirmation (kind of pointless).

ilovezfs:
ah, no problem

PR #4538, opened 2016-09-05
duplicity: migrate to openssl@1.1

tdsmith:
Bump resource versions; test that requests[security] is enabled.


tdsmith:
oops. https://github.com/pyca/pyopenssl/issues/525

PR #4554, opened 2016-09-06
varnish: replace docutils resource with a buildtime dep

zmwangx:

---

Just like in #4505 (mpv), the only things we need here from `docutils` are the `rst2man.py` and `rst2html.py` executables which can be found in `Formula["docutils"].opt_bin`.

While we're at it, fix audit issue and improve the test ever so slightly.

### DIFF
--- a/Formula/varnish.rb
+++ b/Formula/varnish.rb
@@ -11,24 +11,13 @@ class Varnish < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "docutils" => :build
   depends_on "pcre"
 
-  resource "docutils" do
-    url "https://pypi.python.org/packages/source/d/docutils/docutils-0.11.tar.gz"
-    sha256 "9af4166adf364447289c5c697bb83c52f1d6f57e77849abcccd6a4a18a5e7ec9"
-  end
-
   def install
-    ENV.prepend_create_path "PYTHONPATH", buildpath+"lib/python2.7/site-packages"
-    resource("docutils").stage do
-      system "python", "setup.py", "install", "--prefix=#{buildpath}"
-    end
-
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
-                          "--localstatedir=#{var}",
-                          "--with-rst2man=#{buildpath}/bin/rst2man.py",
-                          "--with-rst2html=#{buildpath}/bin/rst2html.py"
+                          "--localstatedir=#{var}"
     system "make", "install"
     (etc+"varnish").install "etc/example.vcl" => "default.vcl"
     (var+"varnish").mkpath
--- a/Formula/varnish.rb
+++ b/Formula/varnish.rb
@@ -23,10 +23,6 @@ class Varnish < Formula
     (var+"varnish").mkpath
   end
 
-  test do
-    system "#{opt_sbin}/varnishd", "-V"
-  end
-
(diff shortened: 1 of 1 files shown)


zmwangx:
Eww, forgot to edit the title.

ilovezfs:
one :star: forfeited

ilovezfs:
@zmwangx if possible try to keep commit subjects under 50 characters. So for this I'd probably do `varnish: depend on docutils` or if you feel the need to explain `varnish: depend on docutils in lieu of a resource`

zmwangx:
@ilovezfs Yes, I get magit (git interface within Emacs) warnings every day.

Since 50 isn't typically enforced (while something like 69 is more of a hard cutoff) and being able to understand what happened in each commit by glancing at a screenful of titles is really important in my workflow, I would usually choose to explain the change more clearly if I only exceed by a few characters, but I don't feel strongly about that.

ilovezfs:
data: https://gist.github.com/ilovezfs/20cdb1c71ae8da3b7837cb5803c35173

ilovezfs:
and perhaps more realistic, subjects containing a ':' and not containing the word "bottle":
https://gist.github.com/ilovezfs/7dc4c859dd4f3edf6ec551b081363b4f

zmwangx:
Ones longer than 150 chars:

```
$ git log --format='%h %s' master | awk '{ if (length($0) > 158) print; }'
ac27673 Upgrades the Java, Python and 32/64 bit Go Google App Engine SDKs to 1.9.5. SHAs taken from Google's download site at https://developers.google.com/appengine/downloads
6908809 libmemcached: fix build "___sync_fetch_and_add" compile error on OS X 10.5 and earlier. Based on patch-Portfile.diff in https://trac.macports.org/ticket/24021
b194683 updated cloudwatch tools to latest version. fixed issue where cloud watch tools wouldn't install without ec2-api-tools because there was no 'depends' statement
53d47f3 Fixed URL to tomcat tar.gz (yes, same checksum). It might be good to look out for unknown mirrors, as the one used here decided it no longer wanted to serve tomcat.
c1237f5 ogmtools now creates man pages in correct directory ogmtools now builds dvdxchap (to extract dvd chapter information) added libdvdread dependency for dvdxchap
```

Fascinating.

ilovezfs:
Also,
```
git log --format='%h %s' master | awk '{ if (length($0) > 158) print; }'|awk '{print $1}'|xargs -n1 git log -n1 --format="%ci : %ai"
2014-05-22 12:24:56 -0700 : 2014-05-22 18:20:27 +0930
2010-12-11 18:12:11 +0000 : 2010-12-09 15:19:06 -0800
2010-12-03 20:15:07 +0000 : 2010-12-03 12:26:34 -0500
2010-09-04 10:39:57 +0200 : 2010-09-03 14:07:54 -0700
2010-02-23 22:39:48 -0800 : 2010-02-21 04:45:50 -0500
```
:laughing:

zmwangx:
This year's data:

https://gist.github.com/anonymous/1f926900006247ecc349dc868b5e8e6d

`/:/ && !/bottle/`:

https://gist.github.com/anonymous/2ffb21cacd44cf99bf601fd6c1d1a6dc

Still quite some offenders.

Formula/varnish.rb
```diff
@@ -64,4 +60,8 @@ def plist; <<-EOS.undent
       </plist>
     EOS
   end
+
+  test do
+    assert_match version.to_s, shell_output("#{opt_sbin}/varnishd -V 2>&1")
```
ilovezfs: sbin not opt_sbin
```
Josephs-MacBook-Pro:Formula joe$ ag sbin|grep _output|wc -l
      17
Josephs-MacBook-Pro:Formula joe$ ag sbin|grep _output|grep opt|wc -l
       0
```

zmwangx: Fixed.

ilovezfs:
:+1: :ship:

PR #4556, opened 2016-09-06
Golo 3.2.0-M4 (devel)

rlespinasse:
http://golo-lang.org/news/2016/09/06/golo-3.2.0-M4


ilovezfs:
@rlespinasse thanks for the devel bump! :balloon:

PR #4558, opened 2016-09-06
Several Gnome package updates

tschoonj:


ilovezfs:
@BrewTestBot test this please

ilovezfs:
@tschoonj I bumped the timeout up just in case :)

ilovezfs:
@tschoonj not sure how the test for gxml used to be passing ... that's weird ...

I guess it'll need something like this
```
diff --git a/Formula/gxml.rb b/Formula/gxml.rb
index c69f9c6..f9fcce1 100644
--- a/Formula/gxml.rb
+++ b/Formula/gxml.rb
@@ -52,7 +52,7 @@ class Gxml < Formula
     gettext = Formula["gettext"]
     glib = Formula["glib"]
     libgee = Formula["libgee"]
-    flags = %W[
+    flags = ENV["CPPFLAGS"].split + %W[
       -I#{gettext.opt_include}
       -I#{glib.opt_include}/glib-2.0
       -I#{glib.opt_lib}/glib-2.0/include
```

tschoonj:
Just pushed a commit that fixes `gxml`'s test. It was failing locally too as it couldn't find `libxml2`'s headers.

The `gtkmm3` test failing is strange and I cannot reproduce it. The path to the `gdk/gdk.h` header is properly included as far as I can tell.

ilovezfs:
It was supposed to be finding the system ones due to the `ENV.libxml2`. I'm not sure why gxml even uses the libxml2 formula as a dependency, since removing it results in
```
iMac-TMP:homebrew-core joe$ brew linkage gxml
System libraries:
  /usr/lib/libSystem.B.dylib
  /usr/lib/libxml2.2.dylib
Homebrew libraries:
  /usr/local/opt/libgee/lib/libgee-0.8.2.dylib (libgee)
  gettext
```
using the system libxml2.

tschoonj:
I think the explicit `libxml2` dependency was added to ensure it runs on older versions of OS X, which ship with too old versions of `libxml2` to get `gxml` compiling.

ilovezfs:
bizarre.

ilovezfs:
@tschoonj the proximate cause is that gtk+3 is uninstalled by test-bot right after the zenity test.
```
==> brew test --verbose zenity
==> brew uninstall --force gtk+3
```
now why the bot decided to do that and then didn't re-install it ... not sure yet

ilovezfs:
@tschoonj can you put the gxml fix in a separate PR so I can merge that first? Also, the pangom commit needs to be moved above the gtk+3 commit (or to a separate PR to follow this one) because its reverse dependency gtkmm3 depends on gtk+3 but gtk+3 won't be installed at the point that gtkmm3 is tested.

ilovezfs:
@tschoonj Yeah, a separate PR for pangomm to follow is the way to go because otherwise I think you'll get this
```
==> brew install ./pangomm-2.40.1.el_capitan.bottle.tar.gz
==> brew test pangomm --verbose
==> brew fetch --retry gtkmm
==> FAILED
==> Downloading https://homebrew.bintray.com/bottles/gtkmm-2.24.5.el_capitan.bottle.tar.gz
==> Retrying download
==> Downloading https://homebrew.bintray.com/bottles/gtkmm-2.24.5.el_capitan.bottle.tar.gz
Error: Failed to download resource "gtkmm"
Download failed: https://homebrew.bintray.com/bottles/gtkmm-2.24.5.el_capitan.bottle.tar.gz
```

tschoonj:
Ok, I will open separate PRs then for all these commits. Do you happen to know if there's an easy way to do this? Via `brew` or `hub` perhaps?

ilovezfs:
the others should be fine together I think, but `git checkout -b foo-123 origin/master` + `git cherry-pick relevant-sha` + `hub pull-request --browse` is your friend if you want to separate them. But I think you just want to use `git rebase -i master` and delete the lines for gxml and pangomm. Then open those two in separate PRs with the above sequence.

ilovezfs:
@tschoonj voilà :green_heart:

ilovezfs:
@tschoonj :ship:'d! Thanks for this PR and especially for your patience.

PR #4561, opened 2016-09-06
pangomm 2.40.1

tschoonj:


ilovezfs:
cancelled CI on this one and we'll requeue it after 4558 is merged

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@tschoonj requeued now that #4558 has :boat:'d

ilovezfs:
and another :green_apple: !

ilovezfs:
@tschoonj thanks for your hard work on these PRs! Your contributions are much appreciated.

tschoonj:
You're most welcome, many thanks for guiding me through this.

Could you have a look at this issue with the build-bot uninstalling dependencies after a test? I have encountered this before and it is really annoying...

ilovezfs:
@tschoonj the problem is that once gtk+3 has passed it's no longer protected from being uninstalled, which is important in that if something that's tested subsequently only depends on it at build time, it needs to fail if it's actually linked to a build-time only dependency. But later when it checks brew uses it encounters a formula that depends on something already installed, which itself depended on gtk+3. However, since it's already installed, its dependencies aren't reconsidered to see whether or not they're installed. So we'd probably need to do a brew uses --installed --recursive on gtk+3 and uninstall all of those too, which would be rather slow, or override the bypass that skips checking dependencies for things already installed, and allow it to look for the just built bottle that it had uninstalled since it won't be able to fetch it by the normal mechanism. So I think we could fix it by always checking dependencies during the brew uses tests even something is already installed (or equivalently running brew missing before starting those tests and installing anything listed), and by copying the bottle to the homebrew cache after it's built so there's actually something found to install not a fetch failure.

PR #4570, opened 2016-09-06
imagemagick: unrecommend opencl and improve test

zmwangx:

---

I noticed the OpenCL issue back when I was working on #3511. As described in the commit message:

> OpenCL support is something to be --enable'd. We've only been specifying
> --disable-opencl when building --without, meaning that no build
> setting (bottle or source, with or without options) ever supported
> OpenCL since upstream made OpenCL opt-in. Miraculously though, no one
> ever complained, so it's probably safe to unrecommend it.
> 
> Also, from personal testing building with proper OpenCL support actually
> crashes IM. See details below. Might be related:
> https://www.imagemagick.org/discourse-server/viewtopic.php?t=24603.

I'm still in favor of more systematic treatment of build options (as in #3511), but if we only want to address user-reported issues (this PR itself contains a user — that's me — reported issue), this is it. CC @MikeMcQuaid.
## Details about the OpenCL problem (you shouldn't need to care)

More specifically, even when we use IM's default for everything else, building with `--enable-opencl` could lead to crashes in the most basic commands (like `convert <infile> -resize <width>x<height> <outfile>`). Last time I checked I got a segfault. Today the crash had more a helpful message (that's running inside Fusion though, I didn't bother to touch my main installation):

```
Assertion failed: (cache_info->opencl != (OpenCLCacheInfo *) NULL), function GetOpenCLEvents, file magick/cache.c, line 1699.
Abort trap: 6
```

Still. I don't care about OpenCL support (and no one does it seems), so I'm not gonna waste time debugging this.

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -139,6 +139,11 @@ class Imagemagick < Formula
   end
 
   test do
-    system "#{bin}/identify", test_fixtures("test.png")
+    assert_match "PNG", shell_output("#{bin}/identify #{test_fixtures("test.png")}")
+    # Check support for recommended features and delegates.
+    features = shell_output("#{bin}/convert -version")
+    %W[Modules freetype jpeg png tiff].each do |feature|
+      assert_match feature, features
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


DomT4:
👍 on the changes in this PR specifically, but if you change a default because said default is known to be breaking please bump the formula revision as well.

zmwangx:
Gonna give this another day before merging.

MikeMcQuaid:
👍

PR #4579, opened 2016-09-07
llvm 3.9.0

Godin:
- [x] `creduce`
- [x] `ponyc`
- [x] `castxml`
- [x] `ldc devel`
- [ ] `ldc`
- [x] `crystal-lang`

### DIFF
--- a/Formula/castxml.rb
+++ b/Formula/castxml.rb
@@ -15,6 +15,12 @@ class Castxml < Formula
     sha256 "df8baff500313f946a09a64bd1bb4a56d909ad4cce7d1c1eb1dcf637215b5183" => :yosemite
   end
 
+  # changes from upstream to fix compilation with LLVM 3.9
+  patch do
+    url "https://github.com/CastXML/CastXML/commit/e1ee6852c79eddafa2ce1f134c097decd27aaa69.patch"
+    sha256 "d47f4566bda6f8592c120052aeec404de371dc27b0ef15d5c337c34f87976901"
+  end
+
   depends_on "cmake" => :build
   depends_on "llvm"
 
--- a/Formula/castxml.rb
+++ b/Formula/castxml.rb
@@ -5,6 +5,7 @@ class Castxml < Formula
   mirror "https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/c/castxml/castxml_0.1+git20160706.orig.tar.xz"
   version "0.1+git20160706"
   sha256 "28e7df5f9cc4de6222339d135a7b1583ae0c20aa0d18e47fa202939b81a7dada"
+  revision 1
 
   head "https://github.com/CastXML/castxml.git"
 
--- a/Formula/creduce.rb
+++ b/Formula/creduce.rb
@@ -3,7 +3,7 @@ class Creduce < Formula
   homepage "https://embed.cs.utah.edu/creduce/"
   url "https://embed.cs.utah.edu/creduce/creduce-2.5.0.tar.gz"
   sha256 "2dcd784e1d27df60f4ea1d92c4c556c02da4152db353d544dce8b7813fa443e4"
-  revision 1
+  revision 2
 
   head "https://github.com/csmith-project/creduce.git"
 
--- a/Formula/creduce.rb
+++ b/Formula/creduce.rb
@@ -14,6 +14,16 @@ class Creduce < Formula
     sha256 "710460f6d639171d170f7ae3b835fa3ff708f56c1105c000086913571896659d" => :mavericks
   end
 
+  # changes from upstream to fix compilation with LLVM 3.9
+  patch do
+    url "https://github.com/csmith-project/creduce/commit/377e4f93790690edcd0cde057bb96300aca22858.diff"
+    sha256 "1bf6201132dd41953f4c7d454a8f383f432e5a3f2fac416d681bc5e8e8065e18"
+  end
+  patch do
+    url "https://github.com/csmith-project/creduce/commit/d09b8f69419d6c65c7b32675962c9aee813713b5.diff"
+    sha256 "995fcb3c5c7aef60ce34ece07a5d225f0bf7886539479988d8b8a6cc0c52de90"
+  end
+
   depends_on "astyle"
   depends_on "delta"
   depends_on "llvm"
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -12,6 +12,12 @@ class CrystalLang < Formula
     sha256 "0faca697844e5ef7bbcf9d750ed27c448e224363c7150ebebc77c73f01099e07" => :yosemite
   end
 
+  # changes from upstream to fix compilation with LLVM 3.9
+  patch do
+    url "https://github.com/crystal-lang/crystal/pull/3439.diff"
+    sha256 "ee55985315881461f1fd2fb7ccc36b547d5defb23a91b022c760d84abe328fd9"
+  end
+
   option "without-release", "Do not build the compiler in release mode"
   option "without-shards", "Do not include `shards` dependency manager"
 
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -4,6 +4,7 @@ class CrystalLang < Formula
   url "https://github.com/crystal-lang/crystal/archive/0.19.4.tar.gz"
   sha256 "e239afa449744e0381823531f6af66407ba1f4b78767bd67a9bb09d9fcc6b9e4"
   head "https://github.com/crystal-lang/crystal.git"
+  revision 1
 
   bottle do
     sha256 "b891e7a9c01906be5b6d51adc374c1882e036ae443b95abb1a8f27b072762fde" => :sierra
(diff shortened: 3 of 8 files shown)


DomT4:
Will also need to bump the revisions on `castxml`, `creduce`, `emacs-clang-complete-async`, `ponyc` and `rtags`. Some of those may well need new releases or compatibility patches backported.

Godin:
@DomT4 increased revision in all formulas which have `depends_on "llvm"`, i.e. `crystal-lang` and `ldc` in addition to your list. And indeed - according to Jenkins almost all of those projects should be updated/patched to compile with clang 3.9 :disappointed:

Shall we do the same for formulas with `depends_on "llvm" => :optional`, i.e. `doxygen`, `cctols` and `rust` ?

DomT4:
> And indeed - according to Jenkins almost all of those projects should be updated/patched to compile with clang 3.9

I feared as much, to be honest. LLVM updates are usually pretty hard to get through.

Let's see what the response is from upstreams as to whether this is something we can get done quickly through patch backports or new releases, but otherwise we might have to consider adding the 3.9.0 release as a new formula & migrating things over as possible.

> Shall we do the same for formulas with depends_on "llvm" => :optional

We don't bother; option usage is generally fairly low & forcing everyone to do a rebuild to cover the small minority that use those options isn't worthwhile.

Godin:
Strange, `ponyc` fails, while its current version 0.3.0 supposed to have support for LLVM 3.9 already - https://github.com/ponylang/ponyc/commit/f0fa3fc0c7c128c05b2493e5231eed94289d8f01

Godin:
According to release notes `ldc-1.1.0-beta2` (`devel` in homebrew) supposed to work too, but also fails.

Godin:
Seems that `crystal-lang` put update to LLVM 3.9 on hold - https://github.com/crystal-lang/crystal/issues/3237

Godin:
Patch from `castxml` upstream - https://github.com/CastXML/CastXML/commit/e1ee6852c79eddafa2ce1f134c097decd27aaa69 , but unfortunately due to change in `README.rst` it can't be applied cleanly on current version `0.1+git20160412` in homebrew.

Godin:
:yum: `creduce` successfully patched

Marqin:
* ponyc has updated to 0.3.2 and seems to have few more commits about fixing llvm39 since 0.3.0

Godin:
@Marqin thank you for notification. `master` branch has been merged into this to get update of `ponyc`, let's see what @BrewTestBot will tell us

Godin:
Looks like @BrewTestBot can't eat PRs containing merges from master with resolution of conflicts, so instead rebased this branch.

Godin:
😋  `ponyc` passed

Thanks again for notification @Marqin :thumbsup:

Marqin:
btw. could it be possible to add this as [standalone `llvm39` package](https://github.com/Homebrew/homebrew-versions/issues/1388) (like `llvm38`, `llvm37`, etc.)

Godin:
@Marqin I don't think the same question in multiple places gives any speedup.
Some work has been already started - https://github.com/Homebrew/homebrew-versions/pull/1384

And as stated in https://github.com/Homebrew/homebrew-versions/blob/19e6a94df3130a8038789aa5c1dcd5eccffc5850/.github/ISSUE_TEMPLATE.md
> Please note by far the quickest way to get a new formula into Homebrew is to file a Pull Request.

So to really speedup the process - IMO either provide a PR yourself for the opened issue, either help finalize an existing PR  😉

Godin:
:yum: `castxml` passed after update and patch

Godin:
`ldc devel` actually includes so called `ldc-lts`, which is `ldc-0.17.1`, and this one is failing, not `ldc-1.1.0-beta2`. I have no idea why `ldc-lts` was preserved here together with `stable ldc-1.0.0`. but this definitely doesn't look like an upstream issue. And so I'm wondering if it can be simply removed? This way we'll be able either do a patch just for `1.0.0` (hopefully by picking it from `1.1.0-beta2`), either wait `1.1.0` GA. Otherwise `0.17.1` should also be patched and guessing based on gap between version numbers it might be not that straightforward, i.e. probably without cherry-picking from upstream.

@DomT4 maybe you have an advice?

iMichka:
@Godin I updated castxml separately, because the old URL's were broken and there were no Sierra bottles. You can rebase your PR. Of course the patch for llvm 3.9 needs to stay the time being.

P.S. thanks for pushing llvm 3.9 forward 👍

JohanEngelen:
@Godin LDC 0.17.1 has already been patched for LLVM3.9, LDC's 'ltsmaster' branch, but has not been released yet unfortunately.
Hope we can release 0.17.2 soon, so that there is a good LDC bootstrap procedure with LLVM3.9. ("bootstrap" = building LDC without a pre-installed D compiler. 0.17.* can be built with just a C++ compiler, and then 0.17.* can build the latest version).

Godin:
@iMichka thanks for notification - rebased

JohanEngelen:
LDC 0.17.2 (can be built with LLVM3.9) has been released: https://github.com/ldc-developers/ldc/releases/tag/v0.17.2

John-Colvin:
The ldc formula should now work with llvm 3.9.0

Godin:
@JohanEngelen @John-Colvin thx for notification, rebased and `ldc-devel` passed. So just need to wait when it will go out of beta.

ilovezfs:
@Godin looks like ponyc needs a little conflict resolution

ilovezfs:
@Godin there's a crystal-lang PR adding 3.9 support now https://github.com/crystal-lang/crystal/pull/3439 if you'd like to test it out!

Godin:
@ilovezfs yup, `ponyc` got several updates in past days and gets conflict resolution with each rebase here. And thanks for notification about `crystal-lang`, I'll try to check it.

Godin:
😋  `crystal-lang` passed build after patch

However I did not checked whether it actually works or not.
I guess anyway would be better to wait when work in https://github.com/crystal-lang/crystal/pull/3439 will be finished and it will be closed.

JohanEngelen:
(Note that LDC has a chicken-egg problem with this:  we use homebrew in our CI tests to install LLVM.  Because LLVM3.9 is not available yet, we also can't test whether LDC works with LLVM3.9 on Mac...)

ilovezfs:
I don't see any reason to wait on crystal-lang/crystal#3439.

ilovezfs:
@Godin Thanks for your excellent work on this PR and your contribution to Homebrew! :fireworks:

ilovezfs:
@JohanEngelen chicken and egg problem solved :)

JohanEngelen:
@ilovezfs It would help me a lot if `brew install llvm39` would work (instead of `brew install llvm` which I read installs LLVM3.9 at the moment).

ilovezfs:
@JohanEngelen You can add a symlink in `Aliases`

```
ln -s ../Formula/llvm.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Aliases/llvm39
```

JohanEngelen:
The reason I want to type `llvm39` is that `llvm` or your proposed solution will require babysitting. If brew updates to llvm4.0, I don't want the tester to automatically update to it too.

ilovezfs:
I'd recommend using your own tap in that case.

jgavris:
@JohanEngelen @ilovezfs Hopefully someone will make an llvm39 formula https://github.com/Homebrew/homebrew-versions/blob/master/llvm38.rb

ilovezfs:
llvm.rb in core is the llvm39 formula.

JohanEngelen:
I'd much appreciate the copy-paste from homebrew-core/llvm.rb to homebrew-versions/llvm39.rb !

ilovezfs:
@JohanEngelen You can do that in your own tap ... it's not difficult to set up: https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md

JohanEngelen:
[sigh]  https://github.com/Homebrew/homebrew-versions/pull/1446

JohanEngelen:
This is very frustrating. Installing an arbitrary version of llvm (`brew install llvm`) is bound to give build problems for people. Not having a versioned `llvm` install available effectively means that brew cannot be used for it. Unfortunately, it appears TravisCI does not come with macports preinstalled, otherwise I would have switched already.

MikeMcQuaid:
> [sigh] Homebrew/homebrew-versions#1446

I'm not sure what reaction you expected with "I have no clue about homebrew contribution guidelines and do not wish to know."

> Installing an arbitrary version of llvm (brew install llvm) is bound to give build problems for people.

Which is why we allow you to install older versions in homebrew/versions, any version pinned to a version from your own tap and are working on improving our versions support (https://github.com/Homebrew/brew/issues/620).

> Not having a versioned llvm install available effectively means that brew cannot be used for it.

The large numbers of people doing `brew install llvm` disagree with you.

> Unfortunately, it appears TravisCI does not come with macports preinstalled, otherwise I would have switched already.

I'm not sure what this is meant to communicate to us. You can make your TravisCI job install MacPorts, checkout a specific version of Homebrew and disable auto-update, install from your own tap, etc. If you'd behaved a little differently I'd be willing to help you do that but you didn't so I'm not, sorry.

PR #4580, opened 2016-09-07
README and ISSUE_TEMPLATE: quote $(brew --repo)

zmwangx:
Always a good idea to quote command substitution.


MikeMcQuaid:
:shipit:

PR #4590, opened 2016-09-07
mercurial 3.9.1

ngoldbaum:
See https://www.mercurial-scm.org/wiki/WhatsNew#Mercurial_3.9.1_.282016-9-1.29

### DIFF
--- a/Formula/mercurial.rb
+++ b/Formula/mercurial.rb
@@ -3,9 +3,8 @@
 class Mercurial < Formula
   desc "Scalable distributed version control system"
   homepage "https://mercurial-scm.org/"
-  url "https://mercurial-scm.org/release/mercurial-3.9.tar.gz"
-  sha256 "834f25dcff44994198fb8a7ba161a6e24204dbd63c8e6270577e06e6cedbdabc"
-  revision 1
+  url "https://mercurial-scm.org/release/mercurial-3.9.1.tar.gz"
+  sha256 "625e4fc7e85ec2278c2828bdc547fce74091b3bbe4d9eeeba2d61af51195df74"
 
   bottle do
     cellar :any_skip_relocation


Formula/mercurial.rb
```diff
@@ -3,8 +3,8 @@
 class Mercurial < Formula
   desc "Scalable distributed version control system"
   homepage "https://mercurial-scm.org/"
-  url "https://mercurial-scm.org/release/mercurial-3.9.tar.gz"
-  sha256 "834f25dcff44994198fb8a7ba161a6e24204dbd63c8e6270577e06e6cedbdabc"
+  url "https://mercurial-scm.org/release/mercurial-3.9.1.tar.gz"
+  sha256 "625e4fc7e85ec2278c2828bdc547fce74091b3bbe4d9eeeba2d61af51195df74"
   revision 1
```
DomT4: Need to cut the `revision` line on stable version updates as well.

ngoldbaum: Oh, interesting, I never had to do that before.

ngoldbaum: Oh I see, because normally there isn't a rev bump

DomT4: Yeah, we _try_ to reserve them for exceptional circumstances because it forces absolutely everyone who has the formula installed to do a rebuild, which can be annoying.

Formula/mercurial.rb
```diff
@@ -3,9 +3,9 @@
 class Mercurial < Formula
   desc "Scalable distributed version control system"
   homepage "https://mercurial-scm.org/"
-  url "https://mercurial-scm.org/release/mercurial-3.9.tar.gz"
-  sha256 "834f25dcff44994198fb8a7ba161a6e24204dbd63c8e6270577e06e6cedbdabc"
-  revision 1
+  url "https://mercurial-scm.org/release/mercurial-3.9.1.tar.gz"
+  sha256 "625e4fc7e85ec2278c2828bdc547fce74091b3bbe4d9eeeba2d61af51195df74"
+  revision 0
```
DomT4: Sorry, did a bad job of being clear here. The entire line can be removed. `revision 0` is the default but it's implicit rather than explicit. Only needs to be explicit when the revision is increased beyond `0`.

zmwangx:
Merged as 4880067. Thank you for your contribution to Homebrew!

Issue #4543, opened 2016-09-06
OpenSSL@1.1 Migration [Tracking]


DomT4:
Borrowing my own comment from another thread:

> I don't have a lot of interest in carrying around what is essentially a feature patch (support for 1.1.0) given OpenSSL are supporting 1.0.2 as an LTS branch. Not planning to leap over upstream releases unless absolutely necessary for whatever reason.
> 
> Transition is also complicated by potentially overlapping linkage, so for example, if you `brew install curl --with-openssl` against OpenSSL 1.1.0 but you add `--with-libssh2` as well which remains linked against OpenSSL 1.0.2, we're not really sure how predictably that'll behave. Ideally, the plan is:
> - Anything standalone (i.e. nothing in `brew uses`) that supports OpenSSL 1.1.0 can use this new formula as the dependency.
> - Anything else needs to be upgraded by tree, so to update `curl` you'd want `libssh2` to be using OpenSSL 1.1.0 first, etc.
> 
> We may well change that plan as time passes; If I could I'd have the entire ecosystem using OpenSSL 1.1.0 tomorrow but it really depends on how quickly we can get this done without breaking significant things for users.

Immediate potential deps tree:

```
brew deps --include-build --include-optional --1 <formula>
```

Usage or potential usage list:

```
brew uses <formula> --include-build --include-optional
```

Generate a build report for addition to the list:

```
brew gist-logs <formula>
```

**Fair warning: This is going to be a messy list.**
- [ ] activemq-cpp - Not used by anything else - Fails to build - https://gist.github.com/anonymous/55392af913363942490f5df2f29fc6ce
- [ ] afflib - Used optionally by bulk_extractor, sleuthkit & autopsy - Fails to build, fixed upstream already, will be in 3.7.9 (?) release.
- [ ] aircrack-ng - Not used by anything else - Fails to build - https://gist.github.com/anonymous/241dee104589d7a37a4ed3861622f0a9
- [ ] alpine - Not used by anything else - Fails to build - https://gist.github.com/anonymous/d8fe6343d990cad18a988efa708fc666
- [ ] amap - Not used by anything else - Builds/runs fine, seemingly, once appdefs files are installed.
- [ ] ansible - Not used by anything else - Fails to build, pyOpenSSL - https://gist.github.com/anonymous/d691b6d13da3a8ee264b4f84c77ef3f0
[...]

PR #4591, opened 2016-09-07
curl: migrate to openssl@1.1

JLHwung:
Migrate curl openssl dependency to 1.1. I built locally and passed the brew test. If anything depends both on curl and OpenSSL, will there be any conflicts between curl's linked OpenSSL and the other?

By the way, `#{etc}/openssl@1.1/cert.pem` usage makes me a little bit awkward, we don't have anything like `Formula["openssl@1.1"].etc_prefix`?

I am also not sure whether I should add revision since OpenSSL is only an optional dependency. I could not find any related guidelines on [here](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#formulae-revisions).

It looks more complicated than I thought, see [here](https://github.com/Homebrew/homebrew-core/issues/4543), maybe we shall wait for other upstream to adapt to OpenSSL before upgrading curl's OpenSSL dependency?

---

---

### DIFF
--- a/Formula/curl.rb
+++ b/Formula/curl.rb
@@ -3,6 +3,7 @@ class Curl < Formula
   homepage "https://curl.haxx.se/"
   url "https://curl.haxx.se/download/curl-7.50.2.tar.bz2"
   sha256 "0c72105df4e9575d68bcf43aea1751056c1d29b1040df6194a49c5ac08f8e233"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


DomT4:
> It looks more complicated than I thought, see here, maybe we shall wait for other upstream to adapt to OpenSSL before upgrading curl's OpenSSL dependency?

Yeah. We need to wait for at least `libssh2` & `nghttp2` to be migrated over to `openssl@1.1` here.

JLHwung:
@DomT4 How do I know `curl` optionally depends on `libssh2` and `nghttp2` without manually going through `brew info curl`? Once I run `brew uses nghttp2` I got nothing, if I run `brew uses curl` there is nothing about `nghttp2`.

DomT4:
```
brew deps --include-build --include-optional --1 curl
brew uses nghttp2 --include-build --include-optional
```

JLHwung:
@DomT4 Awesome! I suggest updating these commands as well as `brew gist-logs` on the [main tracking list](https://github.com/Homebrew/homebrew-core/issues/4543) so people could help when they have spare time.

vszakats:
This libssh2 patch makes it possible to build/use it with final OpenSSL 1.1.0:
https://github.com/libssh2/libssh2/commit/64ebfd8182a9b6e637e65c3059e3798e199274b3

`nghttp2` works just fine with OpenSSL 1.1.0 at least as far it is required to build it for `curl`. (but may possibly need patches to build its executable tools - no info on that.)

MikeMcQuaid:
Passing on this for now until everything is sorted upstream. Sorry and thanks anyway!

vszakats:
Latest libssh2 (1.8.0) builds fine against OpenSSL 1.1, so the cited separate patch is no longer necessary.

PR #4593, opened 2016-09-07
pidcat: update to 2.1.0

JakeWharton:


zmwangx:
Merged as ea6d093. Thank you for your contribution to Homebrew! For future reference, our preferred commit subject is `foo 1.2.3`.

PR #4594, opened 2016-09-07
openssl@1.1: backport two fixes

DomT4:

---

### DIFF
--- a/Formula/openssl@1.1.rb
+++ b/Formula/openssl@1.1.rb
@@ -26,6 +26,17 @@ class OpensslAT11 < Formula
     depends_on :perl => "5.10"
   end
 
+  # Both these can be removed whenever the next 1.1.0 release is available.
+  patch do
+    url "https://github.com/openssl/openssl/commit/b7d2480c82.patch"
+    sha256 "74efa5161331ef8ecd95a15ebb7013017ec8ef734f40ff11ed06fdac2c808f18"
+  end
+
+  patch do
+    url "https://github.com/openssl/openssl/commit/92ed7fa57.patch"
+    sha256 "0ff0eb395accdc035cc1f6bfa072b309fff4f23e38d1768612966511d8a88faf"
+  end
+
   def arch_args
     {
       :x86_64 => %w[darwin64-x86_64-cc enable-ec_nistp_64_gcc_128],


tdsmith:
On principle, I worry about cherry-picking commits for openssl. :( I don't have the expertise to make sure they don't depend on other fixes applied to master, especially since openssl test coverage is dismal. My preference would be to wait for the next release. What's the advantage to patching these now? Projects like pyopenssl won't test correctly on other platforms until this gets fixed by an upstream release anyway.

DomT4:
We do it considerably less than a lot of package managers, but we are prone to backporting fixes for our current `openssl` formula because it saves us grief & I think that's "fair game" here too. 

Bearing in mind that all of the patches are picked off their respective branch rather than `master`, and go through two sets of individual review & two sets of CI upstream. We don't cherry pick from `master` for OpenSSL, pretty much ever.

> What's the advantage to patching these now?

Less breakage for users, is the obvious one. Beyond that, we can test stuff against 1.1.0 cleanly with the chance of it turning out to be some unrelated bug lessened. The next release could well be months unless 1.1.0 pretty drastically departs from OpenSSL's classic stable release timetable, or rather lack of release timetable.

In the mean time we block projects like `pyopenssl` from using Homebrew to test against, which reduces usage of `openssl@1.1` as a whole, which reduces the amount of real-world testing our new versioning system gets before we start using it more widely.

I'm happy to wait, but I think we'd be being overly-cautious here personally. Potentially worth a discussion point is that we don't regularly turn down patches for other crypto libraries, only ever OpenSSL.

tdsmith:
FWIW I checked in with the PyCA devs and they don't want us to patch.

If you're happy to wait, shall we close this?

> Less breakage for users, is the obvious one.

I guess I wonder what breakage means and who the users are. Holding off on these patches shouldn't affect `brew` users.

> Potentially worth a discussion point is that we don't regularly turn down patches for other crypto libraries, only ever OpenSSL.

Yes, I think we should always approach patches against security-critical libraries conservatively.

DomT4:
> Yes, I think we should always approach patches against security-critical libraries conservatively.

You're welcome to start enforcing this, as long as we do it consistently. A lot of my gripe with bothering to maintain the core any more is that we can't ever decide anything consistently & it makes everyone's lives harder.

tdsmith:
> A lot of my gripe with bothering to maintain the core any more is that we can't ever decide anything consistently & it makes everyone's lives harder.

Yes, it would be nice to have a conversation about policy-setting at some point.

DomT4:
As you said yourself in the past, too many cooks in the kitchen. I've basically given up on the team ever agreeing on anything consistently.

tdsmith:
Yeah, that's why I think defining a process for legislating policy or accepting changes would be helpful, especially as the team grows. I don't have a next action item, though. :/

DomT4:
No, me either to be honest.

It's a cache-22-esque situation; we need to expand the team so maintaining any single aspect of Homebrew doesn't fall overwhelmingly on one or two people, including things like you & Python, but every expansion of the team makes agreeing things as a team even more difficult than it already has become, which is massively demotivating & frustrating.

MikeMcQuaid:
> but every expansion of the team makes agreeing things as a team even more difficult than it already has become, which is massively demotivating & frustrating.

There is no agreement as a team and I don't think it's worth seeking to get it. As I've stated privately before: I think you should have a corresponding amount of say on a particular item based on how much you work on it and how responsible you've proved yourself to be in the past in cleaning up your mistakes. From that point of view, I actually agree with @tdsmith's viewpoint on being careful on OpenSSL patches but given @DomT4's almost solely working on this I think the final decision should be up to him on whether to merge this.

PR #4596, opened 2016-09-07
dynamodb-local 2016-05-17

pas256:

---

### DIFF
--- a/Formula/dynamodb-local.rb
+++ b/Formula/dynamodb-local.rb
@@ -69,6 +69,6 @@ class DynamodbLocal < Formula
   end
 
   test do
-    system "#{bin}/dynamodb-local", "-help"
+    system bin/"dynamodb-local", "-help"
   end
 end
--- a/Formula/dynamodb-local.rb
+++ b/Formula/dynamodb-local.rb
@@ -67,4 +67,8 @@ class DynamodbLocal < Formula
     </plist>
     EOS
   end
+
+  test do
+    system "#{bin}/dynamodb-local", "-help"
+  end
 end
--- a/Formula/dynamodb-local.rb
+++ b/Formula/dynamodb-local.rb
@@ -1,9 +1,9 @@
 class DynamodbLocal < Formula
   desc "Client-side database and server imitating DynamoDB"
   homepage "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Tools.DynamoDBLocal.html"
-  url "https://dynamodb-local.s3.amazonaws.com/dynamodb_local_2016-03-01.tar.gz"
-  version "2016-03-01"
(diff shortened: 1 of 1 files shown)


pas256:
This is my first brew contribution... be gentle

zmwangx:
Thank you for your contribution! Without people like you we couldn't possibly operate Homebrew at this scale.

That said, if you ran `brew audit --new-formula dynamodb-local` in the checklist above, you'll see an error:

```
$ brew audit --new-formula dynamodb-local
...
dynamodb-local:
Error: 1 problem in 1 formula
```

Could you please help us create a test block? There's a brief explanation of the test block [here](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula), and there are plenty of examples around.

zmwangx:
By the way, our preferred commit subject is `<formula name> <new version>`, which means `dynamodb-local 2016-05-17` in this case. I'd really appreciate it if you could fix that too.

pas256:
OK, changed. Thanks for the info.

How does that look?

Formula/dynamodb-local.rb
```diff
@@ -67,4 +67,8 @@ def plist; <<-EOS.undent
     </plist>
     EOS
   end
+
+  test do
+    system "#{bin}/dynamodb-local", "-help"
```
zmwangx: `bin/"dynamodb-local"` instead of `"#{bin}/dynamodb-local"`.

But do you think there's something more substantial than `-help` you can do here?

pas256: `-help` means Java is able to run the JAR. If the JAR had `-version` I would do that like the Elixir formula does, but it doesn't have that.

Open to suggestions if you have them

zmwangx: > -help means Java is able to run the JAR. If the JAR had -version I would do that like the Elixir formula does, but it doesn't have that.

Yes I know, and `-version` wouldn't be too thrilling a test either. Just wondering if you can squeeze some actual dynamodb usage into the test. Maybe insert something into the database then retrieve it? I'm not a user of this software and don't wish to learn to use every piece of software in Homebrew, so unfortunately I can't make suggestions.

If you tell me for sure that a more substantial test is impractical, I've no problem sticking with this.

pas256:
I have changed the test `bin` directory reference.

As for the test, that is really complex to do reliability. First, pick a language that happens to be installed on the computer, then install the AWS SDK so the code can write and read something.

I'd like to just go ahead with this test.

zmwangx:
Got it, thanks for the explanation!

zmwangx:
Merged as c9856e4. Thank you for your first contribution to Homebrew, and I hope we'll see you around.

PR #4598, opened 2016-09-07
go 1.7.1

cblecker:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

Modified description to pass `brew audit --new-formula Formula/go.rb`. Used title from https://golang.org/

PR #4608, opened 2016-09-08
timewarrior 1.0.0 (new formula)

pirj:

---

### DIFF
--- a/Formula/timewarrior.rb
+++ b/Formula/timewarrior.rb
@@ -1,0 +1,21 @@
+class Timewarrior < Formula
+  desc "Command-line time tracking application"
+  homepage "https://taskwarrior.org/docs/timewarrior/"
+  url "http://taskwarrior.org/download/timew-1.0.0.tar.gz"
+  sha256 "ac027910e1e8365bdd218a8b42389b26d017d38d3c96516c408db6d5a44e0bb5"
+  head "https://git.tasktools.org/scm/tm/timew.git"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    (testpath/".timewarrior/data").mkpath
+    (testpath/".timewarrior/extensions").mkpath
+    touch testpath/".timewarrior/timewarrior.cfg"
+    assert_match "Tracking foo", shell_output("#{bin}/timew start foo")
+  end
+end


Formula/timew.rb
```diff
@@ -0,0 +1,18 @@
+class Timew < Formula
```
zmwangx: Formula name should be `timewarrior` (i.e., source file should be `timewarrior.rb`), and class name should be `Timewarrior`.

Formula/timew.rb
```diff
@@ -0,0 +1,18 @@
...
+  desc "Timewarrior is a command-line time tracking application"
```
zmwangx: Description should be "Command-line time tracking application".

Formula/timew.rb
zmwangx: By the way, `timew` could be an alias if it's also commonly known as such.

Formula/timew.rb
```diff
@@ -0,0 +1,18 @@
...
+  homepage "https://taskwarrior.org/docs/timewarrior/"
+  url "http://taskwarrior.org/download/timew-1.0.0.tar.gz"
+  sha256 "ac027910e1e8365bdd218a8b42389b26d017d38d3c96516c408db6d5a44e0bb5"
```
zmwangx: Maybe also `head "https://git.tasktools.org/scm/tm/timew.git"` after this line.

Formula/timew.rb
```diff
@@ -0,0 +1,18 @@
...
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system "timew"
```
zmwangx: Test is failing, because it shows an interactive Y/N prompt.

zmwangx: Also, just calling the executable isn't a good test. It's best to ask it do something, then somehow validate the output.

zmwangx:
Thanks for submitting a new formula. A couple inline comments. By the way, the commit subject should be `timew 1.0.0 (new formula)` (formulae is the plural form of formula).

pirj:
Thanks for lightning fast reply and remarks. All seems to be fixed, however please take another look, @zmwangx.

Formula/timewarrior.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system "timew"
```
zmwangx: This is not addressed yet.

zmwangx:
Thanks for addressing the various problems, but please take a look at the test block.

zmwangx:
By the way, it is not in the checklist, but you should really run `brew test timewarrior` to see it break for yourself.

Re update: `which timew` is not really a good test. Please refer to https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula for guidelines.

pirj:
@zmwangx I did, but that file was already in place, and either the tool itself doesn't detect the shell as non-interactive, either the shell is interactive during tests.
I've addressed this checking against the version number.
Do you want me to add this as a preferable method to [Formula doc](http://www.rubydoc.info/github/Homebrew/brew/master/Formula#test-class_method) instead of current `We are fine if the executable does not error out`?

pirj:
@BrewStatusBot build and test

Formula/timewarrior.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    assert_equal version, shell_output("timew --version").strip
```
zmwangx: `version.to_s` actually.

zmwangx:
> I've addressed this checking against the version number.

Sorry, I saw the `system "which timew"` commit but not the latest.

> Do you want me to add this as a preferable method to Formula doc instead of current We are fine if the executable does not error out?

Is it really not possible to have a more substantial test? Testing `--version`, `--help` etc. is a fallback when it's hard to test anything else, which is usually reserved for daemon, stuff that require authentication, etc.

zmwangx:
> @BrewStatusBot build and test

The bot will automatically requeue your job when you push to this branch.

pirj:
> The bot will automatically requeue your job when you push to this branch.

@zmwangx I've seen other people [asking BrewTestBot to rebuild](https://github.com/Homebrew/homebrew-core/pull/4598#issuecomment-245463473)

Formula/timewarrior.rb
pirj: Redunant according to [docs ("The version string")](http://www.rubydoc.info/github/Homebrew/brew/master/Formula#version-class_method)

zmwangx:
> I've seen other people asking BrewTestBot to rebuild.

You might have noticed the "Homebrew member" badge there 😆 It only works for maintainers. And that's not necessary in most cases, because as I said the bot automatically requeues when you push. When occasionally manually trigger a rebuild when the earlier build failed for whatever reason that doesn't require any code change.

Formula/timewarrior.rb
zmwangx: It works, but it's the convention anyway so might as well respect it.

```
$ grep -rI 'assert_match version.to_s' /usr/local/Library/Taps/homebrew/homebrew-core | wc -l
150
```

pirj:
@zmwangx `version` usage changed according to conventions.
Thanks for ramp-up! Do you feel there's anything that needs to be done before merging?

1. Do you want me to address the example section for `test do` ?
2. I feel that interactive/non-interactive might be a common issue, do you want me to take a look if it's a Jenkins issue by some reason?
3. I see several build scheduled after mine, 7395-7401, unfortunately with no details. It seems some of them might be duplicates for the same PR, and make no sense in case of rebase and push --force (-with-lease).

Do you think those three issues deserve their own issues, or are well-known caveats already?

ilovezfs:
```
timewarrior 1.0.0 (new formula)
```
not
```
timewarrior 1.0.0 (new formulae)
```

Formula/timewarrior.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage "https://taskwarrior.org/docs/timewarrior/"
+  url "http://taskwarrior.org/download/timew-1.0.0.tar.gz"
+  sha256 "ac027910e1e8365bdd218a8b42389b26d017d38d3c96516c408db6d5a44e0bb5"
+  head "https://git.tasktools.org/scm/tm/timew.git"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make"
```
ilovezfs: don't need this. just "make", "install" works in this case

ilovezfs:
For the test, two things
1. it needs to use the full path
2. I'd suggest the following to get around the interactivity issues, though you could do something more meaningful if you'd like
```
test do
    (testpath/".timewarrior/data").mkpath
    (testpath/".timewarrior/extensions").mkpath
    touch ".timewarrior/timewarrior.cfg"
    assert_match "Tracking foo", shell_output("#{bin}/timew start foo")
end
```

zmwangx:
> Do you want me to address the example section for test do ?

What do you mean by "the example section"? If you mean a better test I'm all for it.

> I feel that interactive/non-interactive might be a common issue, do you want me to take a look if it's a Jenkins issue by some reason?

It's not a Jenkins issue, happened for me locally too. I didn't have time to investigate, but you might be able to just pipe a yes to it, or better yet, figure out some actual commands that work non-interactively.

> I see several build scheduled after mine.

Each push resulted in a build, no worries. Should be green now.

Heads up, I may not be available over the next few hours.

zmwangx:
Looks like ILZ already has a better test for you.

pirj:
@ilovezfs Thanks a lot for sane example (read: writing test for me).
Make+make install combined, commit message orpho error fixed. Looking forward for the build to pass.

zmwangx:
I took the liberty to edit the PR title so that it matches the final product better. There are lots of queued jobs right now so please be patient.

ilovezfs:
@pirj you're welcome.

zmwangx:
@ilovezfs Happy to :shipit:?

DomT4:
The URL should be `https://`. Feel free to do in post, though.

zmwangx:
@DomT4 Of course, thanks for noticing that. Easy to fix that when actually shipping.

zmwangx:
@DomT4 By the way, IIRC you were planning to audit HTTPS URLs, right?

DomT4:
No worries. It's something that I'm planning to audit sooner rather than later; I'm working on replacing myself robotically.

ilovezfs:
@DomT4 not sure auditing that is going to really be possible because you'll have to be careful to make sure the content actually resolves to the homepage not just that the server claims to accept https and then returns something which may or may not be the homepage. That's easier said than done.

zmwangx:
@ilovezfs Shouldn't be hard to let a bot crawl and update everything and have a human review all the changes?

ilovezfs:
@zmwangx yeah that's more likely to work unless we want to be dealing with bogus brew audit failures frequently.

zmwangx:
Oops I take back my word.

```
$ grep -rI http: | wc -l
2741
```

WTF.

ilovezfs:
I think the following have to be true to insist on https

1. https is not simply redirecting to http.

2. https and http return the same content modulo http/https string differences in the content. If they do not return the same content then https must resolve and http must not resolve.

3. The certificate chain is valid. If the certificate is not valid, then it is not an audit failure. (Lots of servers are configured in such a way that they appear to be claiming to accept https but don't actually accept it.)

zmwangx:
bot.brew.sh :trollface:

ilovezfs:
@zmwangx https://github.com/Homebrew/homebrew.github.io/issues/98

ilovezfs:
>Oops I take back my word.

What's the problem? We'll just have every maintainer do 1 per day for a year

zmwangx:
> We'll just have every maintainer do 1 per day for a year

I think vszakats will beat the said maintainer to it.

ilovezfs:
LOL

ilovezfs:
>@ilovezfs Happy to :shipit:?

@zmwangx :boat:

zmwangx:
Merged as 0fa7093. Thank you for your contribution to Homebrew!

pirj:
Thanks guys! That was an amazing experience ❤️

ilovezfs:
@pirj thanks for the cool new formula!

PR #4610, opened 2016-09-08
httpstat 1.1.2 (new formula)

reorx:
  
  Not exactly, according to the Brew Cookbook, `brew audit --strict --online $FORMULA` test passed, however `brew audit --new-formula <formula>` didn't pass with the following:
  
  ```
  httpstat:
  Error: 1 problem in 1 formula
  ```
  
  which seems not a problem for submitting this pull request.

---

### DIFF
--- a/Formula/httpstat.rb
+++ b/Formula/httpstat.rb
@@ -1,10 +1,11 @@
 class Httpstat < Formula
   desc "Curl statistics made simple"
   homepage "https://github.com/reorx/httpstat"
-  url "https://github.com/reorx/httpstat/archive/v1.1.2.tar.gz"
-  sha256 "0468abebfbae9ff7ea35d577b60ff9af9db3ff67ffd81b11ac005c4e9602919b"
+  url "https://github.com/reorx/httpstat/archive/v1.1.3.tar.gz"
+  sha256 "8ba3a8f0550468a634dce45d45dc03375e5dcdedea4c8edc8c26a326d929af04"
 
   depends_on :python if MacOS.version <= :snow_leopard
+  bottle :unneeded
 
   def install
     bin.install "httpstat.py" => "httpstat"
--- a/Formula/httpstat.rb
+++ b/Formula/httpstat.rb
@@ -2,14 +2,13 @@ class Httpstat < Formula
   desc "curl statistics made simple"
   homepage "https://github.com/reorx/httpstat"
   url "https://github.com/reorx/httpstat/archive/v1.1.2.tar.gz"
-  version "1.1.2"
   sha256 "0468abebfbae9ff7ea35d577b60ff9af9db3ff67ffd81b11ac005c4e9602919b"
 
   bottle :unneeded
 
   def install
-      bin.install "httpstat.py"
-      mv bin/"httpstat.py", bin/"httpstat"
+    bin.install "httpstat.py"
+    mv bin/"httpstat.py", bin/"httpstat"
   end
 
   test do
--- a/Formula/httpstat.rb
+++ b/Formula/httpstat.rb
@@ -11,6 +11,6 @@ class Httpstat < Formula
   end
 
   test do
-    assert_match "HTTP/1.1", shell_output("#{bin}/httpstat https://github.com")
+    assert_match "HTTP", shell_output("#{bin}/httpstat https://github.com")
   end
 end
--- a/Formula/httpstat.rb
+++ b/Formula/httpstat.rb
@@ -1,17 +1,16 @@
 class Httpstat < Formula
-  desc "curl statistics made simple"
+  desc "Curl statistics made simple"
   homepage "https://github.com/reorx/httpstat"
   url "https://github.com/reorx/httpstat/archive/v1.1.2.tar.gz"
   sha256 "0468abebfbae9ff7ea35d577b60ff9af9db3ff67ffd81b11ac005c4e9602919b"
 
-  bottle :unneeded
+  depends_on :python if MacOS.version <= :snow_leopard
 
   def install
-    bin.install "httpstat.py"
-    mv bin/"httpstat.py", bin/"httpstat"
+    bin.install "httpstat.py" => "httpstat"
   end
 
   test do
-    system "#{bin}/httpstat", "https://github.com"
+    assert_match "HTTP/1.1", shell_output("#{bin}/httpstat https://github.com")
   end
 end
--- a/Formula/httpstat.rb
+++ b/Formula/httpstat.rb
@@ -1,0 +1,18 @@
+class Httpstat < Formula
+  desc "curl statistics made simple"
+  homepage "https://github.com/reorx/httpstat"
+  url "https://github.com/reorx/httpstat/archive/v1.1.2.tar.gz"
+  version "1.1.2"
+  sha256 "0468abebfbae9ff7ea35d577b60ff9af9db3ff67ffd81b11ac005c4e9602919b"
+
+  bottle :unneeded
+
+  def install
+      bin.install "httpstat.py"
+      mv bin/"httpstat.py", bin/"httpstat"
+  end
+
+  test do
+    system "#{bin}/httpstat", "https://github.com"
+  end
+end


Formula/httpstat.rb
```diff
@@ -0,0 +1,17 @@
...
+  desc "curl statistics made simple"
+  homepage "https://github.com/reorx/httpstat"
+  url "https://github.com/reorx/httpstat/archive/v1.1.2.tar.gz"
+  sha256 "0468abebfbae9ff7ea35d577b60ff9af9db3ff67ffd81b11ac005c4e9602919b"
+
+  bottle :unneeded
+
+  def install
+    bin.install "httpstat.py"
+    mv bin/"httpstat.py", bin/"httpstat"
```
zmwangx:
```ruby
bin.install "httpstat.py" => "httpstat"
```

But `httpstat.py` is not executable, so installing it to `bin` seems pointless. Maybe you could add a shebang (`/usr/bin/env python`) to `httpstat.py`?

reorx: @zmwangx Thanks for pointing out, didn't know `install` could specify the destination name, I'll remove the `mv` later.

`httpstat.py` is now executable, just forgot to push the newest commit to master :)

Formula/httpstat.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  test do
+    system "#{bin}/httpstat", "https://github.com"
```
zmwangx: You should try to validate the output, e.g.

```ruby
assert_match "200 OK", shell_output("#{bin}/httpstat https://github.com")
```

Formula/httpstat.rb
zmwangx: Thanks, but you need to tag a new release and update to the new release here or it will still fail due to using the old tarball 😃

ilovezfs:
I'd suggest
```
class Httpstat < Formula
  include Language::Python::Virtualenv

  desc "Curl statistics made simple"
  homepage "https://github.com/reorx/httpstat"
  url "https://files.pythonhosted.org/packages/b0/27/80bdb353de887ad501cde707858ad8202c852bdea848cb55dd764dd19c0b/httpstat-1.1.2.tar.gz"
  sha256 "1295afdbcb485e2b29f7b264d9e320d0aeae8d4bd5bcd51f374b01c4e435d6c6"
  head "https://github.com/reorx/httpstat.git"

  def install
  end

  test do
    system "#{bin}/httpstat", "https://github.com"
  end
end
```

Formula/httpstat.rb
```diff
@@ -0,0 +1,17 @@
+class Httpstat < Formula
+  desc "curl statistics made simple"
```
zmwangx: Please use title case, i.e., capitalize `curl`.

zmwangx:
@ilovezfs When there's just a single file I'd rather not jump through PyPI and setuptools hoops.

zmwangx:
You do need `depends_on :python if MacOS.version <= :snow_leopard` as in @ilovezfs's version, by the way.

Formula/httpstat.rb
reorx: I think the output validation is not necessary, since I have covered all situation that may cause the program to fail, as long as it return 0, it means it can work properly. Even if github becomes 500 error doesn't got anything to do with httpstat's success or fail.

reorx:
Thanks @ilovezfs @zmwangx, I'll take you guys suggestions and update the PR when done

Formula/httpstat.rb
zmwangx: `brew test` is mostly used by our CI to ensure updates don't break things. When GitHub 500's we probably won't be discussing this anyway 😅 Having a minimum of validation helps us sleep better at night, and by adopting it I won't bother you with "test improvement" again later if I happen to forget what we agreed on :wink:

Formula/httpstat.rb
```diff
@@ -0,0 +1,16 @@
+class Httpstat < Formula
+  desc "Curl statistics made simple"
+  homepage "https://github.com/reorx/httpstat"
+  url "https://github.com/reorx/httpstat/archive/v1.1.2.tar.gz"
```
zmwangx: Still need new version for the shebang.

Formula/httpstat.rb
```diff
@@ -0,0 +1,16 @@
...
+  sha256 "0468abebfbae9ff7ea35d577b60ff9af9db3ff67ffd81b11ac005c4e9602919b"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    bin.install "httpstat.py" => "httpstat"
+  end
+
+  test do
+    assert_match "HTTP/1.1", shell_output("#{bin}/httpstat https://github.com")
```
zmwangx: I suggested matching against `200 OK` because it's safer. `HTTP/1.1` might change to `HTTP/2` at some point.

Formula/httpstat.rb
reorx: Actually the file in this version's tar ball does contain the `#!/usr/bin/env python` shebang, you can download and see :)

zmwangx: Oh sorry, did you remove the old version or retag? Anyway, please ignore this comment.

Formula/httpstat.rb
reorx: Well, then I think match `HTTP/` is better, but for `200 OK`,  I really think we shouldn't involved github's status into our contest of testing a program that has no relationship with github service, even though github is the fundamental dependency of brew service in practice, but it is not the requirement of httpstat program. So following the principle in testing concept that unrelated factor shouldn't be involved, we just need to see:

1. Whether httpstat exit with 0
2. Whether httpstat output a response of HTTP protocol

is enough to determine if httpstat is successfully installed.

zmwangx: >  then I think match HTTP/ is better

Sure, go ahead!

zmwangx:
This PR is in a mergeable state. The repo is fairly popular (>400 stars), but it's very new (~10 days). We could hold off merging to let the project mature for a while, or we could merge it in citing popularity. I'm vote for the latter. Thoughts?

ilovezfs:
memeware <- neologism

ilovezfs:
@reorx should this be refreshed to 1.1.3? Also I notice you mention python3 compat on the homepage. Should there be a formula option for using python3?

zmwangx:
@ilovezfs There are many Python formulae in core with Python support, but I don't recall a single one with a python3 option. It just doesn't provide any noticeable benefit in the great majority of cases.

zmwangx:
Somehow omitted the two most important numbers in the comment above... Added.

ilovezfs:
huh?
```
$ ag python3|grep '\:optional'|wc -l
      19
```

zmwangx:
Those are providing Python 3 bindings. Which is entirely different from a utility written in Python that be run with either Python 2 or Python 3, providing the same functionalities.

ilovezfs:
not interested in discussing this further. do what you want.

reorx:
@ilovezfs yes I released 1.1.3 version in pypi days ago, shortly after this pull request was sent, but I didn't update this pull request because I don't want other things to influence the review process of it, I had intended to refresh to 1.1.3 after PR is merged. Should I refresh it now?

zmwangx:
Go ahead, it doesn't make a difference.

reorx:
So is there anything else that I need to do? I'm sorry for the inexperience since it's the first time I submit a formula, I suppose by @zmwangx s saying

> This PR is in a mergeable state.

it's all depend on your guys to decide whether to merge or not?

ilovezfs:
The software looks useful and is very popular already. I have no issue with putting aside our ageism for this one.

zmwangx:
The only problem is

```
httpstat:
Error: 1 problem in 1 formula
```

which should resolve itself in two weeks and a half.

I'm happy to merge it right away though, this one is definitely better than many formulae we have in core that are either abandoned or used by no one or both.

zmwangx:
@reorx Could you please update to 1.1.3? Then either @ilovezfs or I will merge it.

reorx:
sure! I'll do it right away.

ilovezfs:
>which should resolve itself in two weeks and a half.

It actually resolves itself immediately because the check only occurs for a new formula, so immediately is lifted once it ships, such that audit will pass.

zmwangx:
I meant if we just sit for two weeks and a half and requeue this PR by then, the CI should be green.

ilovezfs:
>I meant if we just sit for two weeks and a half and requeue this PR by then, the CI should be green.

Indeed.

zmwangx:
@ilovezfs This could easily be a `bottle :unneeded`, right?

ilovezfs:
yup.

reorx:
Just get back from a weekly meeting... Submitted!

zmwangx:
Merged as 52a0107. Thank you for your contribution to Homebrew!

ilovezfs:
@reorx thanks for the cool new software

@zmwangx thanks for pulling this

zmwangx:
@ilovezfs What's the point of thanking me 😜

reorx:
Thanks for your patience and guidance @zmwangx @ilovezfs :)

ilovezfs:
@zmwangx it is the response to

>@reorx Could you please update to 1.1.3? Then either @ilovezfs or I will merge it.

PR #4627, opened 2016-09-08
freetype 2.7

lembacon:

---

### DIFF
--- a/Formula/fontconfig.rb
+++ b/Formula/fontconfig.rb
@@ -3,6 +3,7 @@ class Fontconfig < Formula
   homepage "https://wiki.freedesktop.org/www/Software/fontconfig/"
   url "https://www.freedesktop.org/software/fontconfig/release/fontconfig-2.12.1.tar.bz2"
   sha256 "b449a3e10c47e1d1c7a6ec6e2016cca73d3bd68fbbd4f0ae5cc6b573f7d6c7f3"
+  revision 1
 
   # The bottle tooling is too lenient and thinks fontconfig
   # is relocatable, but it has hardcoded paths in the executables.
--- a/Formula/freetype.rb
+++ b/Formula/freetype.rb
@@ -4,9 +4,9 @@ class Freetype < Formula
   # Note: when bumping freetype's version, you must also bump revisions of
   # formula with "full path" references to freetype in their pkgconfig.
   # See https://github.com/Homebrew/legacy-homebrew/pull/44587
-  url "https://downloads.sf.net/project/freetype/freetype2/2.6.5/freetype-2.6.5.tar.bz2"
-  mirror "https://download.savannah.gnu.org/releases/freetype/freetype-2.6.5.tar.bz2"
-  sha256 "e20a6e1400798fd5e3d831dd821b61c35b1f9a6465d6b18a53a9df4cf441acf0"
+  url "https://downloads.sf.net/project/freetype/freetype2/2.7/freetype-2.7.tar.bz2"
+  mirror "https://download.savannah.gnu.org/releases/freetype/freetype-2.7.tar.bz2"
+  sha256 "d6a451f5b754857d2aa3964fd4473f8bc5c64e879b24516d780fb26bec7f7d48"
 
   bottle do
     cellar :any
--- a/Formula/gd.rb
+++ b/Formula/gd.rb
@@ -3,6 +3,7 @@ class Gd < Formula
   homepage "https://libgd.github.io/"
   url "https://github.com/libgd/libgd/releases/download/gd-2.2.3/libgd-2.2.3.tar.xz"
   sha256 "746b6cbd6769a22ff3ba6f5756f3512a769bd4cdf4695dff17f4867f25fa7d3c"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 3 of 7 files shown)


zmwangx:
Why are you bumping all these dependents which don't seem to require bumps? I checked `fontconfig`, `imagemagick` and `libbluray`, all of them link against `/usr/local/opt/freetype/lib/libfreetype.6.dylib` and none of them stops working after the freetype upgrade. Did you get build errors perhaps?

DomT4:
@zmwangx See https://github.com/Homebrew/legacy-homebrew/pull/44587#issuecomment-145355908. We should probably just be fixing the `pkg-config` files though so this is the last one that needs everything revisioned to basically fix a text file.

zmwangx:
@DomT4 Ah thanks for the pointer, and regarding the situation, it's pretty gross. I certainly don't fancy bumping IM for this.

DomT4:
We'll need to sanity check it, but I'm in favour of rewriting the involved formulae's `.pc` files to use `opt_prefix` or `opt_lib` or whatever they're referencing rather than the Cellar path. That should mean unless there's API breakage in future we'll only need to do the `freetype` bump & nothing else each time.

zmwangx:
👍 Will look into this.

lembacon:
@DomT4 @zmwangx Actually, I have a fix in #4635. These formulae still needs one more revision bump afterwards, so their pkg-config files can pick up freetype's opt_prefix.

DomT4:
@BrewTestBot test this please

zmwangx:
@DomT4 Shouldn't we rebase this branch on 4304053 first?

lembacon:
@zmwangx Rebased. 😼

DomT4:
CI failure is CI being stupid about the order in which it tests & uninstalls stuff.

Merged! Thank you for your contributions to Homebrew @lembacon, and hurrah for this being the last set of revision bumps unless they actually break the API in future.

zmwangx:
Great, I almost got to the end of bottling top 100 for Sierra and now I have four more...

But really, getting rid of these painful bumps is priceless, thanks!

DomT4:
😄. This is what it's going to be like until we switch Mavericks off in favour of Sierra, constant rebottling of stuff we bottled yesterday 🙈.

zmwangx:
@DomT4 The IM El Cap bottle introduced in https://github.com/Homebrew/homebrew-core/commit/0efdb6a204be14bd464424ff4a62ad9c05a736a2 seems to have linkage problem? https://travis-ci.org/Homebrew/homebrew-core/jobs/158917006#L2075-L2080

zmwangx:
I ended up pushing a rebuild d3471fc which should fix the linkage problem. I think what happened before was that `libwmflite` lingering in the noisy environment got picked up by IM's `configure`.

ilovezfs:
@zmwangx I'd prefer we go through CI if there's a linkage problem in the bottles produced by test-bot, and figure out why that happened, and fix it. Otherwise, this will just happen again next time.

Also we know the first set of bottles were broken, then an actual formula revision is required so that people with the broken bottles get upgraded to the non-broken ones, which a rebuild alone will not trigger even if they brew update and brew upgrade.

The Homebrew Testing job way of slipping in a rebuild should only really be used if you forgot to pull --bottle or you generating a bottle for a new OS, but the existing bottles are fine for the other OSs and there's no independent reason to modify the formula file itself.

ilovezfs:
>I think what happened before was that libwmflite lingering in the noisy environment got picked up by IM's configure.

Opportunistic linking should typically be handled by finding the negative option in the configure script and passing it. If no such negative option exists, then an upstream bug report should request one, and we should use an inreplace, a patch, or an `ac_cv_foo` to achieve the equivalent until such an option is provided.

zmwangx:
> figure out why that happened, and fix it.

Yes, but I don't have that time now. Maybe you could investigate since you run a local test bot?

> if we know the first set of bottles were broken

Not exactly sure to what extent they're broken; at least I checked the common commands like `convert`, `identify` and such and they're all fine. I didn't spend time to track down which exact component had the linkage problem. Maybe `brew linkage` could be more helpful on that front.

> then an actual formula revision is required so that people with the broken bottles get upgraded to the non-broken ones

It was a short window so I'd rather instruct people to reinstall if they met any problems. However, now that you mentioned it, I'll submit a PR to set the wmf situation straight, then we can revision. Actually there was a PR #3511 to make everything as strict as possible (strict `--without` or `--disable` for anything not requested), but it was clunky and people didn't like the idea, so the alternative we're left with is whack-a-mole like dealing with this wmf thing.

zmwangx:
On second thought I don't want to play whack-a-mole after all. I'll clean up and revive #3511 tomorrow, and since there's a documented instance of screw-up, maybe it'll have a stronger case.

ilovezfs:
Yeah, with a short window, Homebrew Testing job should be fine, so it's not a huge deal one way or the other. But as you said your second post, the important thing is that an actual fix for the opportunistic linking be implemented regardless of whether that's before or after the fixed bottles are pushed.

zmwangx:
Will do it tomorrow 👍 Now I need to get some 💤

ilovezfs:
Yeah, no rush. Thanks for addressing the situation so quickly.

PR #4637, opened 2016-09-08
tldr 1.3.0 (new formula)

danielcompton:

---

The formula doesn't pass audit because the tldr-cpp-client doesn't have enough users.

```
$ brew audit --new-formula tldr
tldr:
Error: 1 problem in 1 formula
```

A similar PR was rejected in https://github.com/Homebrew/legacy-homebrew/pull/47867 because it doesn't have enough users. At first glance, this Formula would seem to have the same problem. 

However the main repo which has the users, contributions, and attention is https://github.com/tldr-pages/tldr. From that repo, you can install an NPM plugin which provides tldr results. However that has obvious issues, the biggest for me is that npm is not homebrew. That is the reasoning for adding this formula instead.

This formula was taken from https://github.com/tldr-pages/homebrew-tldr.

### DIFF
--- a/Formula/tldr.rb
+++ b/Formula/tldr.rb
@@ -1,0 +1,18 @@
+class Tldr < Formula
+  desc "Simplified and community-driven man pages"
+  homepage "https://tldr-pages.github.io"
+  url "https://github.com/tldr-pages/tldr-cpp-client/archive/v1.3.0.tar.gz"
+  sha256 "6210ece3f5d8f8e55b404e2f6c84be50bfdde9f0d194a271bce751a3ed6141be"
+  head "https://github.com/tldr-pages/tldr-cpp-client.git"
+
+  depends_on "pkg-config" => :build
+  depends_on "libzip"
+
+  def install
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    assert_match "brew", shell_output("#{bin}/tldr brew")
+  end
+end


apjanke:
The rationale for allowing this formula despite being "non-notable" makes sense to me.

Formula/tldr.rb
```diff
@@ -0,0 +1,23 @@
...
+  desc "Simplified and community-driven man pages"
+  homepage "https://tldr-pages.github.io"
+
+  url "https://github.com/tldr-pages/tldr-cpp-client/archive/v1.2.0.tar.gz"
+  sha256 "db2e70b3f4209c67b5b9896e052bf3d9e77570dcb04d0c9932776aa44cca9d3e"
+
+  head "https://github.com/tldr-pages/tldr-cpp-client.git"
+
+  depends_on "curl"
+  depends_on "libzip" => :build
```
Leandros: you have to set that as a build dependency, too.

apjanke: Do you mean as a run time dependency?

zmwangx:
```
$ brew linkage tldr
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/curl/lib/libcurl.4.dylib (curl)
  /usr/local/opt/libzip/lib/libzip.4.dylib (libzip)
```

So yes, should be a runtime dependency too.

danielcompton:
Thanks for the helpful feedback all, I've updated the formula to @zmwangx's suggestion.

zmwangx:
Thanks for fixing it!

@Leandros Do you feel strongly about the `"DEBUG=-march=native -O2"` flags to `make`?

Leandros:
@zmwangx No, but since it's currently a none-bottled formula, they won't do any harm. I'll look into creating a binary which can then be distributed via brew in the next days.

zmwangx:
> I'll look into creating a binary which can then be distributed via brew in the next days.

Our CI bottles formulae automatically so once we merge this we'll have binaries already.

Leandros:
Oh, than we should remove `-march=native`, it might created problems.

zmwangx:
@Leandros Currently the flags aren't included in this PR. If you don't think we should add them then we're golden.

Leandros:
They, unfortunately, are: https://github.com/tldr-pages/tldr-cpp-client/blob/master/Makefile#L7
However, we can simply change this in the formula, instead of the main repository (which might be a little bit dirty, though), by adding `make CFLAGS=` as build command.

zmwangx:
Oh I see. I'm personally fine with adding some flags, but will see if other maintainers have a problem with that.

Formula/tldr.rb
```diff
@@ -0,0 +1,20 @@
...
+  url "https://github.com/tldr-pages/tldr-cpp-client/archive/v1.2.0.tar.gz"
+  sha256 "db2e70b3f4209c67b5b9896e052bf3d9e77570dcb04d0c9932776aa44cca9d3e"
+  head "https://github.com/tldr-pages/tldr-cpp-client.git"
+
+  depends_on "curl"
+  depends_on "libzip"
+  depends_on "pkg-config" => :build
+
+  def install
+    system "make"
```
zmwangx: According to @Leandros, this should better be

```ruby
system "make", "CFLAGS="
```

@danielcompton Could you please make the change? Thanks.

danielcompton: Done

Leandros: Alternatively, or probably even better, would be 

    system "make", "DEBUG=-O2"

This will enable some, safe, optimizations, and disable the debugging flags.

Leandros:
Sorry for being a little late :/
I think it would be better to add:

     system "make", "DEBUG=-O2"

zmwangx:
@Leandros Which do you think is safer long term? People don't tend to change the build instructions in future updates unless they break.

From my initial interactions with `tldr` it doesn't seem to be a performance-critical or otherwise slow program, so I don't feel strongly about `-O2` either way.

ilovezfs:
The CFLAGS= shouldn't be necessary. If you call `make` specifically, superenv refurbishes the args

Leandros:
No, tldr is not performance critical. It's either I/O or network which will slow it down. However, `-O2` only enables safe optimizations, which shouldn't be critical or brake in a future update.

ilovezfs:
Bottling defaults to -march=core2 and -Os

zmwangx:
Sorry for the misinformation...

Leandros:
@ilovezfs That's sweet. If you call `make CFLAGS=-march="-march=core2 -Os` while bottling we're safe now.

Formula/tldr.rb
```diff
@@ -0,0 +1,20 @@
...
+  url "https://github.com/tldr-pages/tldr-cpp-client/archive/v1.2.0.tar.gz"
+  sha256 "db2e70b3f4209c67b5b9896e052bf3d9e77570dcb04d0c9932776aa44cca9d3e"
+  head "https://github.com/tldr-pages/tldr-cpp-client.git"
+
+  depends_on "curl"
+  depends_on "libzip"
+  depends_on "pkg-config" => :build
+
+  def install
+    system "make", "DEBUG=-O2"
```
zmwangx: `DBUEG=-O2` should be removed here because it won't actually do anything

Leandros: Sorry for the confusion. :worried:

zmwangx: I should be the one saying that to you...

danielcompton:
@Leandros I've given you push access to my fork, might be simplest for you to make the last changes :)

zmwangx:
Well I can update that for you too if you don't mind other people touching your code.

zmwangx:
Fixed. CI should be happy other than the "not notable enough" audit problem. Let's sit back and wait for a few 👍 from other maintainers.

Leandros:
Formula looks good. What is meant with "not stable enough"?

zmwangx:
```
$ brew audit --new-formula tldr
tldr:
Error: 1 problem in 1 formula
```

But as explained in the top post, the actual project is https://github.com/tldr-pages/tldr which is definitely notable enough.

Leandros:
Ahh, sorry, I read "stable" instead of "notable". 😊

Leandros:
I've also updated our tap, to reflect the latest changes, including those suggested by you! Thanks a lot! 👍 

https://github.com/tldr-pages/homebrew-tldr/commit/60102af4188e2ae4affaada8d946e64cf257946a

Do we want to include 1.3.0 in the update?

zmwangx:
> Do we want to include 1.3.0 in the update?

We certainly do. Shipping 1.2.0 only to immediately update to 1.3.0 would be pretty bad.

> I've also updated our tap

Once we merge this you can sunset the tap, point your users here, and submit any future changes here.

Leandros:
> Once we merge this you can sunset the tap, point your users to core, and submit any future changes here.

Yes, will do! 👍

Formula/tldr.rb
```diff
@@ -0,0 +1,20 @@
+class Tldr < Formula
+  desc "Simplified and community-driven man pages"
+  homepage "https://tldr-pages.github.io"
+  url "https://github.com/tldr-pages/tldr-cpp-client/archive/v1.3.0.tar.gz"
+  sha256 "6210ece3f5d8f8e55b404e2f6c84be50bfdde9f0d194a271bce751a3ed6141be"
+  head "https://github.com/tldr-pages/tldr-cpp-client.git"
+
+  depends_on "curl"
```
DomT4: Why can't this package use the system `curl`?

Formula/tldr.rb
```diff
@@ -0,0 +1,20 @@
...
+  depends_on "libzip"
+  depends_on "pkg-config" => :build
```
DomT4: Nit, but you'll make me especially happy if you put `:build` dependencies first in the deps list.

Formula/tldr.rb
Leandros: Is it guaranteed to be decently up to date? I remember there were incompatibilities.

zmwangx:
More thoughts folks, especially regarding eligibility? If not, I think we can merge this in a day or two.

Formula/tldr.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage "https://tldr-pages.github.io"
+  url "https://github.com/tldr-pages/tldr-cpp-client/archive/v1.3.0.tar.gz"
+  sha256 "6210ece3f5d8f8e55b404e2f6c84be50bfdde9f0d194a271bce751a3ed6141be"
+  head "https://github.com/tldr-pages/tldr-cpp-client.git"
+
+  depends_on "pkg-config" => :build
+  depends_on "libzip"
+
+  def install
+    system "make"
```
zmwangx: I took a look at `Makefile` and the `install` target already depends on `all`, so this line is redundant. I'll push a fix.

zmwangx:
I said "a day or two" in a comment made two days ago, so there we go.

zmwangx:
Merged as 3b3786d. Thank you @danielcompton for your contribution to Homebrew! And thank you @Leandros for your help. You can now point your users to core!

Leandros:
Awesome! Thank you very much.

PR #4638, opened 2016-09-08
trace2html 2016-09-08

ndfred:
Pulled from latest master, update to fix Chrome compatibility when viewing trace and pull in this patch that fixes Homebrew symlinks:
https://codereview.chromium.org/2134613003/

The repo moved from trace-viewer to catapult which does not have tags / releases, which is why I am pulling directly from the latest commit. The fixes upstream and in this formula should make it trivial to update to a newer version of catapult in the future.

### DIFF
--- a/Formula/trace2html.rb
+++ b/Formula/trace2html.rb
@@ -1,8 +1,9 @@
 class Trace2html < Formula
   desc "Utility from Google Trace Viewer to convert JSON traces to HTML"
-  homepage "https://github.com/google/trace-viewer"
-  url "https://github.com/google/trace-viewer/archive/2015-06-23.tar.gz"
-  sha256 "97ad0ca9c07a50c53a3d881b69b6b6f7424d6b0f4d9e9fb531c2d9273e413f19"
+  homepage "https://github.com/catapult-project/catapult"
+  url "https://github.com/catapult-project/catapult/archive/328391fbc1e3f931caf93d1e3f50d1643e624560.tar.gz"
+  version "2016-09-08"
+  sha256 "4439ba9037cd1328c7dcc4723537e02dfd99efad36363e025e06ab42c0b99c71"
 
   bottle :unneeded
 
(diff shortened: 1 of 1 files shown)


DomT4:
> The repo moved from trace-viewer to catapult which does not have tags / releases, which is why I am pulling directly from the latest commit. 

Please ask them to consider adding a tag here.

ndfred:
Someone already asked and got no response, I don't expect it to be any different if we ask again (but I pinged them on the issue):
https://github.com/catapult-project/catapult/issues/2323

ndfred:
Failed test is caused by the test infrastructure: `Warning: /usr/local/Cellar is not writable.`

I've been re-installing the formula locally and everything works fine:
`$ brew uninstall --force trace2html ; brew install --verbose --debug trace2html`

Brew audit also doesn't flag anything:
`$ brew audit --strict --online trace2html`

ilovezfs:
@ndfred fear not: https://github.com/Homebrew/brew/pull/896#issuecomment-245862919

ndfred:
No response from the catapult team yet, are tagged releases a hard requirement? This formula pulls a specific commit from master, not master itself (and a release would do the exact same thing). I also used the date as the version number which will allow a consistent migration path for future updates.

tdsmith:
> are tagged releases a hard requirement?

Yes.

One way you could use Homebrew to install this software is to [maintain your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md). The linked document has some advice; please let us know if you have any questions and we'll be happy to help you set things up.

MikeMcQuaid:
Passing on this for now, sorry.

ndfred:
Pinged them again on the issue, will re-open and update when they eventually create a tag on their repo. Until then, the package is broken (produces trace files that Chrome cannot open).

PR #4639, opened 2016-09-08
mapnik 3.0.12

springmeyer:

### DIFF
--- a/Formula/mapnik.rb
+++ b/Formula/mapnik.rb
@@ -1,9 +1,8 @@
 class Mapnik < Formula
   desc "Toolkit for developing mapping applications"
   homepage "http://www.mapnik.org/"
-  url "https://github.com/mapnik/mapnik/archive/v3.0.9.tar.gz"
-  sha256 "f0242606096e2c4ca2cd0caac1ff0fd5f8054a38b5f288ba38b0e397b5b311b2"
-  revision 1
+  url "https://github.com/mapnik/mapnik/releases/download/v3.0.12/mapnik-v3.0.12.tar.bz2"
+  sha256 "66a3d620c3ce543c91ea5b42a25079aca9a2a90f6079a2ce2a8714398fa57d6d"
 
   head "https://github.com/mapnik/mapnik.git"
 
@@ -84,4 +83,9 @@ class Mapnik < Formula
     system "make"
     system "make", "install"
   end
+
+  test do
+    assert_equal prefix.to_s, shell_output("#{bin}/mapnik-config --prefix").chomp
+  end
+
 end


Formula/mapnik.rb
```diff
@@ -84,4 +83,9 @@ def install
     system "make"
     system "make", "install"
   end
+
+  test do
+    assert_equal "#{prefix}".to_s, `#{bin}/mapnik-config --prefix`.chomp
```
zmwangx:
```ruby
assert_equal prefix.to_s, shell_output("#{bin}/mapnik-config --prefix").chomp
```

zmwangx:
One comment, and please squash to one commit with the subject `mapnik 3.0.12` (please keep in mind that our preferred commit subject is `<formula name> <new version number>`). Otherwise 👍

springmeyer:
@zmwangx thank you for the review. I've corrected the test syntax and squashed to a single commit. Is this acceptable now?

zmwangx:
Thanks, I'm happy and let's just wait to make sure the CI is happy too.

Formula/mapnik.rb
```diff
@@ -84,4 +83,9 @@ def install
     system "make"
     system "make", "install"
   end
+
+  test do
+    assert_equal prefix.to_s, shell_output("#{bin}/mapnik-config --prefix").chomp
+  end
+
```
ilovezfs: this line should not be here

Formula/mapnik.rb
```diff
@@ -1,9 +1,8 @@
 class Mapnik < Formula
   desc "Toolkit for developing mapping applications"
   homepage "http://www.mapnik.org/"
-  url "https://github.com/mapnik/mapnik/archive/v3.0.9.tar.gz"
-  sha256 "f0242606096e2c4ca2cd0caac1ff0fd5f8054a38b5f288ba38b0e397b5b311b2"
-  revision 1
+  url "https://github.com/mapnik/mapnik/releases/download/v3.0.12/mapnik-v3.0.12.tar.bz2"
```
zmwangx: I was just about to pull this before I noticed another thing. You should not change the structure of this download URL, and in particular, it's conventional to use gzip tarballs over bz2 around here.

zmwangx: So https://github.com/mapnik/mapnik/archive/v3.0.12.tar.gz

ilovezfs: @zmwangx actually this is correct as-is. We like to use the release tarballs if they're available. Usually the problem is the reverse. We have to switch to the archive tarball temporarily for some reason, and then when that reason goes away, people forget to restore the release tarball until one of the maintainers notices there is a release tarball available and that it works again, and puts it back.

zmwangx: Oh I see what you're saying 👍

Formula/mapnik.rb
```diff
@@ -1,9 +1,8 @@
...
+  sha256 "66a3d620c3ce543c91ea5b42a25079aca9a2a90f6079a2ce2a8714398fa57d6d"
```
ilovezfs: this blank line between the checksum and head should be removed. The blank line needs to be there if and only if there is a formula revision.

Formula/mapnik.rb
ilovezfs: @zmwangx by the reason is primarily, though not solely, that the release tarballs will not require automake, autoconf, and libtool, whereas the source archive tarball may.

zmwangx:
@BrewTestBot test this please

ilovezfs:
@zmwangx No rest for the wicked.

zmwangx:
@ilovezfs Converse is not true.

Formula/mapnik.rb
zmwangx: Fixed this.

Formula/mapnik.rb
zmwangx: Forgot to fix this... :trollface:

ilovezfs: tsk tsk.

zmwangx:
Merged as 85253b3. Thank you for your contribution to Homebrew!

Formula/mapnik.rb
ilovezfs: @zmwangx there I fixed it for you: https://github.com/Homebrew/homebrew-core/pull/4657

zmwangx: @ilovezfs  Kind of overkill but 👍 As a bonus, you might want to open another whitespace nazi PR to fix this :trollface: 

```
$ pcregrep -Ml '^  sha256.*\n\n^  head' *.rb | wc -l
238
```

ilovezfs: Be my guest.

Formula/mapnik.rb
springmeyer: @ilovezfs - thanks that is exactly the situation here. I am the maintainer of Mapnik and made specific changes upstream to enable homebrew to get back on the release tarball. The archive version automatically created by github is not viable - it does not include all the things needed to built easily and we don't want people to use it.

As far as .gz vs .bz2 - I changed the .bz2 because that is all we provide upstream. I can provide .gz if that is important for some reason but I would prefer not to maintain two compression formats unless absolutely necessary.

So, to recap, the new url is necessary.

zmwangx: Thanks for the clarification. gzip is not necessary. It's just that when there are several archives with the same content, the gzip one is usually chosen (from my anecdotal observation).

springmeyer: 👍

PR #4660, opened 2016-09-09
sonarqube: renamed from sonar

Godin:
Sonar has been renamed into SonarQube 3 years ago - http://www.sonarsource.com/company/news/sonarsource-news/sonar-becomes-the-sonarqube-platform/

### DIFF
--- a/Aliases/sonarqube
+++ b/Aliases/sonarqube
@@ -1,1 +1,0 @@
-../Formula/sonar.rb
--- a/Formula/sonarqube.rb
+++ b/Formula/sonarqube.rb
@@ -4,9 +4,9 @@ class Sonarqube < Formula
   url "https://sonarsource.bintray.com/Distribution/sonarqube/sonarqube-6.0.zip"
   sha256 "2256e0b44629632d9528911c8d5916ce246c7abe2be65a6a59a94dee504d62a6"
 
-  depends_on :java => "1.8+"
-
   bottle :unneeded
+
+  depends_on :java => "1.8+"
 
   def install
     # Delete native bin directories for other systems
--- a/Formula/sonarqube.rb
+++ b/Formula/sonarqube.rb
@@ -1,4 +1,4 @@
-class Sonar < Formula
+class Sonarqube < Formula
   desc "Manage code quality"
   homepage "http://www.sonarqube.org/"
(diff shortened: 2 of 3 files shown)


zmwangx:
`sonar` was last updated 2016-08-08, so apparently people have noticed but have been happy with the status quo. I wonder if it's more appropriate to add `sonarqube` as an alias.

zmwangx:
Actually there is already an alias: https://github.com/Homebrew/homebrew-core/blob/master/Aliases/sonarqube. I'll pass on this, but will let other maintainers chime in.

Godin:
@zmwangx well, I was updating formula since Jan 2015, but unfortunately wasn't aware of mechanism of renaming and aliases, so kept it as is to not break things for existing users.

What is preferable - keep an alias or perform rename?

zmwangx:
Oh it's you who's been updating the formula 😃 

On second thought, I don't feel strongly either way, you could rename it, but then you need to delete that alias.

Godin:
I also don't mind to keep an existing name 😄  But renaming seems cleaner than aliasing.
So added removal of existing alias to this PR, let someone with merge rights decide whether to accept or reject this 😉 

Also should be noted that failure of check in Jenkins has been already discussed in https://github.com/Homebrew/homebrew-core/pull/3613#issuecomment-238281556 but unfortunately I had no time to work on this in upstreams.

DomT4:
No objection to the rename. Would be nice to fix:
```
* `bottle modifier` (line 9) should be put before `depends_on` (line 7)
```
though 😉.

Godin:
@DomT4 sure, fixed

zmwangx:
I'm tempted to fix the test to remove useless use of regexp: `/SonarQube/` => `"SonarQube"` too.

DomT4:
Aye. It can be done in post if necessary, but killing unnecessary regexes in favour of strings is 👍.

zmwangx:
> It can be done in post if necessary

It's also a note to self 😃

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/8a1d0aea090a7e77cf4870c088eb759b9bad5072. Thank you for another contribution to Homebrew @Godin! 😺

Godin:
@DomT4 sorry, missed messages about regex. Thanks for merging.

PR #4672, opened 2016-09-10
nodeenv 1.0.0

ilovezfs:
use new virtualenv feature

### DIFF
--- a/Formula/nodeenv.rb
+++ b/Formula/nodeenv.rb
@@ -1,8 +1,10 @@
 class Nodeenv < Formula
+  include Language::Python::Virtualenv
+
   desc "Node.js virtual environment builder"
   homepage "https://github.com/ekalinin/nodeenv"
-  url "https://pypi.python.org/packages/source/n/nodeenv/nodeenv-0.13.6.tar.gz"
-  sha256 "feaafb0486d776360ef939bd85ba34cff9b623013b13280d1e3770d381ee2b7f"
+  url "https://files.pythonhosted.org/packages/fa/62/f3dc0d7b596f7187585520bca14c050909de88866e8f793338de907538cf/nodeenv-1.0.0.tar.gz"
+  sha256 "def2a6d927bef8d17c1776edbd5bbc8b7a5f0eee159af53b9924d559fc8d3202"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


zmwangx:
Regarding the test: why prepend to `PATH` and hoping there's a `node` in the prepended directory (may not be true, in which case the formula is broken but there might be another `node` on `PATH` masking the issue), when we can use the absolute path to said `node`?

ilovezfs:
because this is fine as-is?

ilovezfs:
@DomT4 you wrote the test in 93d74a07 so feel free to answer

zmwangx:
It may be fine on the CI, but for a developer testing this on a dev machine it'll most likely catch nothing even when the formula breaks. What about

```rb
node_in_env = testpath/"env-0.10.40-prebuilt/bin/node"
(testpath/"test.js").write "console.log('hello');"
assert_equal "hello", shell_output("#{node_in_env} test.js").chomp
assert_equal "v0.10.40", shell_output("#{node_in_env} -v").chomp
```

zmwangx:
> it'll most likely catch nothing even when the formula breaks

Oh, actually no, the chance of said developer happen to be running 0.10.40 is quite slim... So whatever.

ilovezfs:
As I've mentioned before, I really only care about what happens on test-bot with respect to test blocks.

zmwangx:
Sure.

DomT4:
Prepending is more in line with how the software actually behaves at run, which is what I was trying to emulate. As you point out, it's not a perfect solution.

I'd be amazed if anyone is running `0.10.40` now though, it's not even the latest 0.10.x series release, let alone any of the newer branches.

zmwangx:
By the way, I do care about running locally because having to throw random commits at CI trying to find out why the test works locally but fails on the CI is not fun.

zmwangx:
> I'd be amazed if anyone is running 0.10.40 now though, it's not even the latest 0.10.x series release, let alone any of the newer branches.

Yeah, see https://github.com/Homebrew/homebrew-core/pull/4672#issuecomment-246080794 where I realized and stepped down.

ilovezfs:
>By the way, I do care about running locally because having to throw random commits at CI trying to find out why the test works locally but fails on the CI is not fun.

This is why you should learn to run test-bot locally.

ilovezfs:
In any case, I'm happy to modify this test to be however @DomT4 would like it to be current since it's "his" test.

zmwangx:
The ops repo doesn't talk about how to set up though... So no, I'll continue to throw commits at bot.brew.sh 😉

DomT4:
I don't mind either way here.

If there's a desire to change it, the theoretical reasons for changing it are valid enough, but in reality if anyone runs into a real world issue with this test by now I'll at least consider eating a hat.

ilovezfs:
@zmwangx that's because there's nothing to explain. You run `brew test-bot foo` starting from an empty brew installation in a new macOS user without any local profile modifications.

zmwangx:
Let's just keep as is.

zmwangx:
> You run brew test-bot foo starting from an empty brew installation in a new macOS user without any local profile modifications.

Oh forgive me, I was thinking about setting up full fledged continuous integration...

ilovezfs:
>Oh forgive me, I was thinking about setting up full fledged continuous integration...

Ah yeah, that would require significantly more explanation and documentation, or at least an extensive poking around at the settings on Jenkins.

ilovezfs:
>I'll at least consider eating a hat.

@DomT4 I hear they pair well with crow meat and humble pie.

DomT4:
@ilovezfs We need to find you some better restaurants I fear 😄.

PR #4678, opened 2016-09-10
caddy: add launchd init

ghost:
make caddy a service, like nginx or apache


zmwangx:
"caddy 0.9.1" would be the title if you're upgrading caddy to 0.9.1. But you're not. So please revert the title which is too misleading right now.

PR #4681, opened 2016-09-10
protobuf 3.0.2

vhbit:


vhbit:
Oops, there is #4676 already. Closing in favor of it.

PR #4705, opened 2016-09-10
graphviz: Support building with ghostscript>=9.16

karanlyons:
Using preexisting patch from graphviz’s repo: https://github.com/ellson/graphviz/commit/f97c86e975aa472272d0a5b5a042b96eb77ebbc0

### DIFF
--- a/Formula/graphviz.rb
+++ b/Formula/graphviz.rb
@@ -1,9 +1,19 @@
 class Graphviz < Formula
   desc "Graph visualization software from AT&T and Bell Labs"
   homepage "http://graphviz.org/"
-  url "http://graphviz.org/pub/graphviz/stable/SOURCES/graphviz-2.38.0.tar.gz"
-  mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/g/graphviz/graphviz_2.38.0.orig.tar.gz"
-  sha256 "81aa238d9d4a010afa73a9d2a704fc3221c731e1e06577c2ab3496bdef67859e"
+
+  stable do
+    url "http://graphviz.org/pub/graphviz/stable/SOURCES/graphviz-2.38.0.tar.gz"
+    mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/g/graphviz/graphviz_2.38.0.orig.tar.gz"
+    sha256 "81aa238d9d4a010afa73a9d2a704fc3221c731e1e06577c2ab3496bdef67859e"
+
+    # https://github.com/ellson/graphviz/commit/f97c86e
+    # Support either version of ghostcript's error prefixes
+    patch do
+      url "https://raw.githubusercontent.com/Homebrew/formula-patches/6a6a1a3/graphviz/patch-gvloadimage_gs.c.diff"
+      sha256 "bcc0758dd9e0ac17bd1cde63a55a613124814002b470ac4a7a0c421c83a253ab"
+    end
+  end
 
   bottle do
     rebuild 1
--- a/Formula/graphviz.rb
+++ b/Formula/graphviz.rb
@@ -1,6 +1,7 @@
 class Graphviz < Formula
   desc "Graph visualization software from AT&T and Bell Labs"
   homepage "http://graphviz.org/"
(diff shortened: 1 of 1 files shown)


Formula/graphviz.rb
```diff
@@ -1,9 +1,14 @@
...
-  url "http://graphviz.org/pub/graphviz/stable/SOURCES/graphviz-2.38.0.tar.gz"
-  mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/g/graphviz/graphviz_2.38.0.orig.tar.gz"
-  sha256 "81aa238d9d4a010afa73a9d2a704fc3221c731e1e06577c2ab3496bdef67859e"
+
+  stable do
+    url "http://graphviz.org/pub/graphviz/stable/SOURCES/graphviz-2.38.0.tar.gz"
+    mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/g/graphviz/graphviz_2.38.0.orig.tar.gz"
+    sha256 "81aa238d9d4a010afa73a9d2a704fc3221c731e1e06577c2ab3496bdef67859e"
+
+    patch :p1, :DATA
```
DomT4: Can we not just use the `https://github.com/ellson/graphviz/commit/f97c86e975aa.patch` style of patch? Does that not apply cleanly?

karanlyons: Unfortunately not. They originally just replaced the error codes (`e_VMerror` -> `gs_error_VMerror`), but that broke compatibility with older versions of the plugin, so *then* they switched the codes back, but added the `#ifndef` checks.

If GitHub supports a way to generate a diff for one file based on two compared commits, then we’ve got a shot, otherwise I think this is the best method.

zmwangx: Do we need `:p1`? That's just a default.

By the way, I really hate inline patches, they can contain anything and screw up grepping and stuff. What about a gist or a hosted patch? https://github.com/Homebrew/formula-patches

karanlyons: Yeah, that's my Pythonista showing through.

I'm more than happy to do it without inline patches, I just wasn't sure of the procedure. Do I submit a seperate pull to that repo as well?

zmwangx: Yeah, if you CC me on a PR there I'll get it merged real quick 👍

karanlyons: Great, will do the next moment I'm at my machine.

zmwangx: Thanks, no rush.

Formula/graphviz.rb
```diff
@@ -1,9 +1,17 @@
...
-  url "http://graphviz.org/pub/graphviz/stable/SOURCES/graphviz-2.38.0.tar.gz"
-  mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/g/graphviz/graphviz_2.38.0.orig.tar.gz"
-  sha256 "81aa238d9d4a010afa73a9d2a704fc3221c731e1e06577c2ab3496bdef67859e"
+
+  stable do
+    url "http://graphviz.org/pub/graphviz/stable/SOURCES/graphviz-2.38.0.tar.gz"
+    mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/g/graphviz/graphviz_2.38.0.orig.tar.gz"
+    sha256 "81aa238d9d4a010afa73a9d2a704fc3221c731e1e06577c2ab3496bdef67859e"
+
+    patch do
```
zmwangx: Could you please also add a comment here explaining why the patch is needed, and link to upstream discussion or commit message if possible?

karanlyons: Can do. I shied away from it since I felt like I saw a lot of other commentless patches, but I’d honestly prefer having the explanations in general anyway.

dunn:
Would you mind squashing this down into one commit?  Thanks so much!

karanlyons:
These failures are I think due to some other issue, but I may be wrong.

dunn:
@brewtestbot test this please

MikeMcQuaid:

PR #4635, opened 2016-09-08
freetype: in `freetype-config` and `freetype2.pc`, replace `prefix` with `opt_prefix`

lembacon:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

This fixes https://github.com/Homebrew/legacy-homebrew/pull/44587#issuecomment-145355908.

PR #4709, opened 2016-09-10
imagemagick: fix option handling

zmwangx:

---

Specify `--enable`/`--disable` or `--with`/`--without` for every single feature and delegate to prevent opportunistic linking.

This is necessary for preventing screwups like the 6.9.5-9_2 rebuild #0 bottle, which picked up `libwmflite-0.2.7.dylib` from `libwmf`, which was then missing due to `libwmf` not being a recommended dep. That resulted in the post-launch audit failure in https://travis-ci.org/Homebrew/homebrew-core/jobs/158917006#L2075-L2080 and a subsequent rebuild.

### DIFF
--- a/Formula/imagemagick.rb
+++ b/Formula/imagemagick.rb
@@ -18,6 +18,7 @@ class Imagemagick < Formula
   option "with-fftw", "Compile with FFTW support"
   option "with-hdri", "Compile with HDRI support"
   option "with-opencl", "Compile with OpenCL support"
+  option "with-openjpeg", "Compile with JPEG 2000 support"
   option "with-openmp", "Compile with OpenMP support"
   option "with-perl", "Compile with PerlMagick"
   option "with-quantum-depth-8", "Compile with a quantum depth of 8 bit"
@@ -26,7 +27,7 @@ class Imagemagick < Formula
   option "without-magick-plus-plus", "disable build/install of Magick++"
   option "without-modules", "Disable support for dynamically loadable modules"
   option "without-threads", "Disable threads support"
-  option "with-zero-configuration", "Disables depending on XML configuration files"
+  option "with-zero-configuration", "Disable configuration files"
 
   deprecated_option "enable-hdri" => "with-hdri"
   deprecated_option "with-jp2" => "with-openjpeg"
@@ -40,20 +41,20 @@ class Imagemagick < Formula
   depends_on "libtiff" => :recommended
   depends_on "freetype" => :recommended
 
-  depends_on :x11 => :optional
+  depends_on "fftw" => :optional
   depends_on "fontconfig" => :optional
+  depends_on "ghostscript" => :optional
+  depends_on "liblqr" => :optional
+  depends_on "librsvg" => :optional
+  depends_on "libwmf" => :optional
   depends_on "little-cms" => :optional
   depends_on "little-cms2" => :optional
-  depends_on "libwmf" => :optional
-  depends_on "librsvg" => :optional
-  depends_on "liblqr" => :optional
   depends_on "openexr" => :optional
-  depends_on "ghostscript" => :optional
-  depends_on "webp" => :optional
   depends_on "openjpeg" => :optional
-  depends_on "fftw" => :optional
   depends_on "pango" => :optional
+  depends_on "webp" => :optional
   depends_on :perl => ["5.5", :optional]
+  depends_on :x11 => :optional
 
   needs :openmp if build.with? "openmp"
 
@@ -69,49 +70,44 @@ class Imagemagick < Formula
       --enable-static
     ]
 
-    if build.without? "modules"
-      args << "--without-modules"
-    else
-      args << "--with-modules"
+    %W[hdri opencl openmp zero-configuration].each do |feature|
+      if build.with? feature
+        args << "--enable-#{feature}"
+      else
+        args << "--disable-#{feature}"
+      end
     end
 
-    if build.with? "opencl"
-      args << "--enable-opencl"
-    else
-      args << "--disable-opencl"
+    %W[fftw fontconfig freetype jpeg magick-plus-plus modules openexr pango
+       perl threads webp].each do |feature|
+      if build.with? feature
+        args << "--with-#{feature}"
+      else
+        args << "--without-#{feature}"
+      end
     end
 
-    if build.with? "openmp"
-      args << "--enable-openmp"
-    else
-      args << "--disable-openmp"
+    %W[
+      ghostscript gslib
+      liblqr lqr
+      libpng png
+      librsvg rsvg
+      libtiff tiff
+      libwmf wmf
+      little-cms lcms
+      openjpeg openjp2
+      x11 x
+    ].each_slice(2) do |option, feature|
+      if build.with? option
+        args << "--with-#{feature}"
+      else
+        args << "--without-#{feature}"
+      end
     end
 
-    if build.with? "webp"
-      args << "--with-webp=yes"
-    else
-      args << "--without-webp"
-    end
-
-    if build.with? "openjpeg"
-      args << "--with-openjp2"
-    else
-      args << "--without-openjp2"
-    end
-
-    args << "--without-gslib" if build.without? "ghostscript"
-    args << "--with-perl" << "--with-perl-options='PREFIX=#{prefix}'" if build.with? "perl"
-    args << "--with-gs-font-dir=#{HOMEBREW_PREFIX}/share/ghostscript/fonts" if build.without? "ghostscript"
-    args << "--without-magick-plus-plus" if build.without? "magick-plus-plus"
-    args << "--enable-hdri=yes" if build.with? "hdri"
-    args << "--without-fftw" if build.without? "fftw"
-    args << "--without-pango" if build.without? "pango"
-    args << "--without-threads" if build.without? "threads"
-    args << "--with-rsvg" if build.with? "librsvg"
-    args << "--without-x" if build.without? "x11"
-    args << "--with-fontconfig=yes" if build.with? "fontconfig"
-    args << "--with-freetype=yes" if build.with? "freetype"
-    args << "--enable-zero-configuration" if build.with? "zero-configuration"
+    gs_font_dir = HOMEBREW_PREFIX/"share/ghostscript/fonts"
+    args << "--with-gs-font-dir=#{gs_font_dir}" if build.with? "ghostscript"
+    args << "--with-perl-options='PREFIX=#{prefix}'" if build.with? "perl"
 
     if build.with? "quantum-depth-32"
       quantum_depth = 32
@@ -123,7 +119,8 @@ class Imagemagick < Formula
     args << "--with-quantum-depth=#{quantum_depth}" if quantum_depth
 
     # versioned stuff in main tree is pointless for us
-    inreplace "configure", "${PACKAGE_NAME}-${PACKAGE_VERSION}", "${PACKAGE_NAME}"
+    inreplace "configure", "${PACKAGE_NAME}-${PACKAGE_VERSION}",
+                           "${PACKAGE_NAME}"
     system "./configure", *args
     system "make", "install"
   end
@@ -137,7 +134,8 @@ class Imagemagick < Formula
   end
 
   test do
-    assert_match "PNG", shell_output("#{bin}/identify #{test_fixtures("test.png")}")
+    test_image = test_fixtures("test.png")
+    assert_match "PNG", shell_output("#{bin}/identify #{test_image}")
     # Check support for recommended features and delegates.
     features = shell_output("#{bin}/convert -version")
     %W[Modules freetype jpeg png tiff].each do |feature|


zmwangx:
CC @DomT4 @MikeMcQuaid who objected to this in #3511. This should be reevaluated due to the aforementioned bottling failure.

Formula/imagemagick.rb
```diff
@@ -43,20 +44,20 @@ class Imagemagick < Formula
...
   depends_on "little-cms2" => :optional
-  depends_on "libwmf" => :optional
-  depends_on "librsvg" => :optional
-  depends_on "liblqr" => :optional
   depends_on "openexr" => :optional
-  depends_on "ghostscript" => :optional
-  depends_on "webp" => :optional
-  depends_on "openjpeg" => :optional
-  depends_on "fftw" => :optional
+  depends_on "homebrew/versions/openjpeg21" if build.with? "openjpeg"
```
DomT4: ??? This formula no longer exists. `openjpeg21` is now `openjpeg` in the core, which is why this was changed.

Formula/imagemagick.rb
```diff
@@ -72,49 +73,149 @@ def install
...
+    end
+
+    if build.with? "fontconfig"
+      args << "--with-fontconfig"
+    else
+      args << "--without-fontconfig"
+    end
+
+    if build.with? "ghostscript"
+      args << "--with-gslib" << "--with-gs-font-dir=#{HOMEBREW_PREFIX}/share/ghostscript/fonts"
```
DomT4: Suspect you'll want to split this onto two unique lines to avoid plowing past 80 characters.

Formula/imagemagick.rb
zmwangx: Oh that was a block copy/paste :) Fixed.

DomT4: You seem to have removed the dependency entirely now 😄

Formula/imagemagick.rb
```diff
@@ -72,49 +72,149 @@ def install
       --enable-static
     ]
 
+    if build.without? "freetype"
```
DomT4: It feels a bit weird to reverse the order here to `if build.without?` rather than `if build.with?`, particularly since we switch half way through the formula here. I'd maybe prefer we kept them using the same format to avoid contributors trying to change stuff quickly not realising they change layout half way though.

Formula/imagemagick.rb
```diff
@@ -72,49 +72,149 @@ def install
...
+      args << "--without-rsvg"
+    end
+
+    if build.with? "libwmf"
+      args << "--with-wmf"
+    else
+      args << "--without-wmf"
+    end
+
+    if build.with? "little-cms"
```
DomT4: Nothing for `little-cms2`? There isn't in the current formula either to be fair, just noticed that now I've looked at it exploded like this.

DomT4:
It's seriously tempting to reconsider our stance on:
```ruby
(build.with?("foo") ? "--with-foo" : "--without-foo")
```
where option handling add so much bulk to the formula. +100 lines here is a pretty significant change.

Formula/imagemagick.rb
zmwangx: ...

Formula/imagemagick.rb
zmwangx: Done.

Formula/imagemagick.rb
zmwangx: I was using `--without` for recommended options and `--with` otherwise. No problem with always using `--with`.

Formula/imagemagick.rb
zmwangx: No, I've checked multiple times already and there's no option for `little-cms2`. There are env vars `LCMS2_CFLAGS` and `LCMS2_LIBS` to override `pkg-config`, otherwise nothing.

zmwangx:
> ```ruby
> (build.with?("foo") ? "--with-foo" : "--without-foo")
> ```

Of course I'd love to use that.

MikeMcQuaid:
> It's seriously tempting to reconsider our stance on:

I don't think we have a stance really, it's to use it when it makes sense (and it does here) so 👍

ilovezfs:
How about something less repetitive like
```
%w[
  foo enable-foo disable-foo
  bar with-bar without-bar
  baz turn-on-baz turn-off-baz
].each_slice(3) do |option, on_value, off_value|
  args << (build.with?(option) ? "--#{on_value}" : "--#{off_value}")
end
```

zmwangx:
Thanks guys for suggestions. I adopted a much more compact style.

ilovezfs:
```
+41 −47
```
nice.

Formula/imagemagick.rb
```diff
@@ -72,49 +73,42 @@ def install
...
-    args << "--enable-hdri=yes" if build.with? "hdri"
-    args << "--without-fftw" if build.without? "fftw"
-    args << "--without-pango" if build.without? "pango"
-    args << "--without-threads" if build.without? "threads"
-    args << "--with-rsvg" if build.with? "librsvg"
-    args << "--without-x" if build.without? "x11"
-    args << "--with-fontconfig=yes" if build.with? "fontconfig"
-    args << "--with-freetype=yes" if build.with? "freetype"
-    args << "--enable-zero-configuration" if build.with? "zero-configuration"
+    args << "--with-gs-font-dir=#{HOMEBREW_PREFIX}/share/ghostscript/fonts" if build.with? "ghostscript"
```
ilovezfs: was the old version wrong which said
```
 -    args << "--with-gs-font-dir=#{HOMEBREW_PREFIX}/share/ghostscript/fonts" if build.without? "ghostscript" 
```

zmwangx: Yeah, I believe so.

ilovezfs: weird ...

zmwangx: That isn't the only error in the old version. It never mentioned `--with-openjp2`, for instance.

ilovezfs:
@zmwangx do you mind tweaking this in some way so that none of the lines go passed the eightieth column? I know they already did before you got to it, but you might as well "fix" it as part of this overhaul.

Formula/imagemagick.rb
zmwangx: @ilovezfs Got a cure for this line that doesn't look too gross?

zmwangx: (I would consider a heredoc for this pretty gross, for instance.)

ilovezfs: you can put the `HOMEBREW_PREFIX/"share/ghostscript/fonts"` in a useless variable. We do that all the time for this reason :D

ilovezfs: or you can always fall back on
```
if build.with? "ghostscript"
...
end
```

zmwangx: > or you can always fall back on

No thanks. I'll use a throwaway variable.

ilovezfs: See? I offer lots of awful choices!

zmwangx: At least you listed the desirable one as the first option. Couldn't ask for more.

zmwangx:
@ilovezfs Done.

Formula/imagemagick.rb
```diff
@@ -72,49 +73,51 @@ def install
...
+        args << "--disable-#{feature}"
+      end
     end
 
-    if build.with? "openjpeg"
-      args << "--with-openjp2"
-    else
-      args << "--without-openjp2"
+    %W[
+      fftw fftw
```
ilovezfs: let's factor out the ones where the pairs are the same https://gist.github.com/ilovezfs/8300cbda17d6bedeacbaab5d90494d74

also I prefer %w if there's no interpolation

Formula/imagemagick.rb
```diff
@@ -72,49 +73,51 @@ def install
...
-    else
-      args << "--disable-openmp"
-    end
-
-    if build.with? "webp"
-      args << "--with-webp=yes"
-    else
-      args << "--without-webp"
+    %W[
+      hdri hdri
```
ilovezfs: since these are also the same, you don't have to repeat them in pairs – just use each instead of each_slice(2) and go horizontal

Formula/imagemagick.rb
zmwangx: The enable/disable block ahead could certainly be reduced to one variable, but I don't see much benefit of refactoring here. But I don't feel strongly either way, so will do.

> also I prefer %w if there's no interpolation

To me that smells like a style violation (only in principle; in practice I don't think `audit` catches it). `%w` stands for single quotes whereas `%W` stands for double quotes, and apparently double quotes is the default quote throughout Homebrew's codebase.

ilovezfs: I think those guidelines are just coming from rubocop basically, which doesn't complain about %w, but you have quite some work to do if you'd like to analyze it that way:
```
Josephs-MacBook-Pro:Formula joe$ grep -r '%W' . |wc -l
     437
Josephs-MacBook-Pro:Formula joe$ grep -r '%w' . |wc -l
     313
```

zmwangx: Just leave them alone.

Formula/imagemagick.rb
```diff
@@ -126,7 +122,8 @@ def install
     args << "--with-quantum-depth=#{quantum_depth}" if quantum_depth
 
     # versioned stuff in main tree is pointless for us
-    inreplace "configure", "${PACKAGE_NAME}-${PACKAGE_VERSION}", "${PACKAGE_NAME}"
+    inreplace "configure",
```
ilovezfs: when it doesn't exceed the eightieth column to do so, it's usually
```
    # versioned stuff in main tree is pointless for us
    inreplace "configure", "${PACKAGE_NAME}-${PACKAGE_VERSION}",
                           "${PACKAGE_NAME}"
```

zmwangx: I was trying to guess your preferred style from https://github.com/Homebrew/homebrew-core/pull/4635#discussion_r78106419 😓 Although the current version does look quite hedious.

ilovezfs: I generally just try to ape @DomT4's practices except when he's wrong.

ilovezfs:
This now looks "correct" to me except for the `%W`s.

@DomT4 ?

Formula/imagemagick.rb
zmwangx: Apparently I aped [@DomT4's practices](https://github.com/Homebrew/homebrew-core/blob/bf8eb02904161e12a6317851c247df51d4f22a4b/Formula/termshare.rb) in #4525 and was [told](https://github.com/Homebrew/homebrew-core/pull/4525#discussion_r77465582) I was wrong. Nothing is reliable 👿

ilovezfs: Aligning under the ":" was definitely his idea https://github.com/Homebrew/homebrew-core/pull/3827#discussion_r74633980 :D

zmwangx: That's a good thing. I hope to see all `:`'s aligned to `"`s in the future. Never a fan of two space indents to begin with.

ilovezfs: Yeah my concern is primarily with keeping the `:branch => "foo"`, which is required whenever "foo" != "master", on the same line as the `url "https://github.com/this/that.git",` even when it spills off the page.

zmwangx: Not talking about the specific discussion thread there.

ilovezfs: That seems problematic to me since it makes it easier not to notice that a non-master branch has to be used if it's spilling off to the right, but ho hum :)

ilovezfs: >Not talking about the specific discussion thread there.

Yeah two space continuations in general are a bit weird, but I don't think rubocop understands alignment directly under the '"' unless we extend it, which hasn't happened yet.

MikeMcQuaid:
> How about something less repetitive like

Personally I'm not a fan of this as I think it makes it less readable to non-Rubyists (i.e. most formulae authors).

ilovezfs:
@MikeMcQuaid given the monstrosity in the bash.rb I think people can cope.

zmwangx:
Also want to point out that I've been having a monopoly on the IM formula for a while and that won't likely change in the near future. For folks who just want to add an option or two it should be relatively clear where and what to add.

ilovezfs:
@MikeMcQuaid did you have a chance to examine the actual product https://github.com/zmwangx/homebrew-core/blob/480de09987b20f95ff85708719150dea131dbdba/Formula/imagemagick.rb#L76-L109

It seems pretty readable ...

MikeMcQuaid:
I did. I think it's readable if you understand Ruby. If we want to start doing this format in more than just a couple of formula I'm going to need to point out there's a generalised (relatively) easy fix for this: make superenv filter /usr/local/lib out if that's the HOMEBREW_PREFIX and require linking only to `opt`

zmwangx:
`/usr/local/include` probably also needs to be filtered out, or an unsuspecting `configure` might decide something is available based on presence of the header, then fail when actually linking.

ilovezfs:
>I'm going to need to point out there's a generalised (relatively) easy fix for this: make superenv filter /usr/local/lib out if that's the HOMEBREW_PREFIX and require linking only to opt

That sounds like Pandora's Box but it's worth thinking about.

As for the present PR, I think it's overkill, and we should only pass the negative options that were causing a problem for CI, and then move on.

MikeMcQuaid:
> /usr/local/include probably also needs to be filtered out

Yep.



> That sounds like Pandora's Box but it's worth thinking about.

Like the sandbox: it'd be something we enable on CI until homebrew/core is 💚 with a flag.

zmwangx:
> there's a generalised (relatively) easy fix

IMO that shouldn't be mutually exclusive with changes proposed here. One side effect of this PR other than preventing opportunistic linking is that you no longer need to find out, for each option, whether `--without` or `--with` (`--enable` or `--disable`) is the default, which is subject to change from release to release. In the old version of the formula there are several options that are doing nothing due to picking the wrong way and they basically rely on opportunistic linking to pick up dependencies, and it fails when no linking is involved (remember #4570 where a recommended option has been doing nothing for a long time).

ilovezfs:
>Like the sandbox: it'd be something we enable on CI until homebrew/core is :green_heart: with a flag.

And here I thought your plan was

MikeMcQuaid:
> IMO that shouldn't be mutually exclusive with changes proposed here.

Perhaps not but with very invasive formula change like this it's good to think about how the package manager can make the formula author's life easier.

DomT4:
> make superenv filter /usr/local/lib out if that's the HOMEBREW_PREFIX and require linking only to opt

We've been discussing this for more or less my entire time around Homebrew 😄. It's definitely something that would be good to get done, but is going to need a ton of testing, perhaps even more than the sandbox did.

Formula/imagemagick.rb
DomT4: > Apparently I aped @DomT4's practices in #4525 and was told I was wrong.

Eh. Human error. I usually fix them when I find them, but evidently was too focused on fixing the build there to think about it much 😅.

> Aligning the ':' under the '"' was definitely his idea #3827 (comment) :D

@ilovezfs Do you have a book somewhere of everything I've publicly said on style or something? 😄.

zmwangx:
> perhaps even more than the sandbox did.

And sandbox issues are still popping up once in a while.

DomT4:
Yes. We only CI test default options to save the CI falling apart, so non-default options that have never been used much aren't absolutely certain to be sandbox-friendly.

zmwangx:
> Perhaps not but with very invasive formula change like this it's good to think about how the package manager can make the formula author's life easier.

Well, my point is this very invasive change will make future formula authors' lives easier.

zmwangx:
> non-default options that have never been used much aren't absolutely certain to be sandbox-friendly.

Not only that. People have different environments than the bot environment too, so even the default may have (and have had) issues.

ilovezfs:
Can we abandon this PR in favor of one that solely passes the negative option for libwmf? I get the sense that we're making a mountain out of a tiny, itsy bitsy mole hill here.

DomT4:
I do somewhat share Mike's concerns we've bumped up the Ruby knowledge on this formula, which we try to keep simple & accessible, but I don't mind trying it here and seeing how it goes. It's probably not something we want to roll out broadly, but given the vast majority of ImageMagick contributors only ever bump the version I think it's possibly safe enough here.

zmwangx:
> Can we abandon this PR in favor of one that solely passes the negative option for libwmf?

Actually no, there are other errors in the old version, so at least we need to fix those too.

zmwangx:
> we've bumped up the Ruby knowledge on this formula

Another way to *not* bump up the Ruby knowledge: just revert to [this damned commit](https://github.com/Homebrew/homebrew-core/pull/4709/commits/dc3977995aafb433f0f7ece33c26b46a2f34fa7a).

ilovezfs:
>I do somewhat share Mike's concerns we've bumped up the Ruby knowledge on this formula, which we try to keep simple & accessible, but I don't mind trying it here and seeing how it goes. It's probably not something we want to roll out broadly, but given the vast majority of ImageMagick contributors only ever bump the version I think it's possibly safe enough here.

:-1: on this. I think it's pretty clear this is not something we want to roll out elsewhere and I see nothing here in particular that justifies it.

>Actually no, there are other errors in the old version, so at least we need to fix those too.

Let's do that in a follow-up PR after one that just fixes the single linkage issue that caused you to push out a bottle rebuild, and bump the revision.

MikeMcQuaid:
> Well, my point is this very invasive change will make future formula authors' (for this particular formula) lives easier.

Sure. My point is we have thousands of formulae and many of them will require changes like this which, to me at least, points to a general solution. If we're not going for a general solution a simpler fix like @ilovezfs proposed may be a better call.

Formula/imagemagick.rb
ilovezfs: >@ilovezfs Do you have a book somewhere of everything I've publicly said on style or something?

Only the one between my ears.

zmwangx:
> points to a general solution

A general solution would be nice, but it only solves half of what is solved by this PR. *I* certainly don't want to read through IM's `configure --help` *again* to work out whether to use `--enable` or `--disable` or `--with` or `--without` for each single option.

Would you be more willing to take in something like dc39779?

MikeMcQuaid:
> A general solution would be nice, but even if implemented it would only solve half of what is solved by this PR. I certainly don't want to read through IM's configure --help again to work out whether to use --enable or --disable or --with or --without for each single option.

> whether --without or --with (--enable or --disable) is the default, which is subject to change from release to release

Do these contradict? I'm not sure I understand. My personal preference is ternary > if/else > current version. It may be helpful for @ilovezfs and @DomT4 to express their preferences too.

ilovezfs:
>I certainly don't want to read through IM's configure --help again

I don't think GitHub is planning on expunging this PR from their databases in the event that you ever feel a burning need for a cheat sheet.

zmwangx:
> Do these contradict? I'm not sure I understand.

Let me explain this better.

<details>
  <summary> `$ configure --help`
  </summary>
```
`configure' configures ImageMagick 6.9.5-9 to adapt to many kinds of systems.

Usage: ./configure [OPTION]... [VAR=VALUE]...

To assign environment variables (e.g., CC, CFLAGS...), specify them as
VAR=VALUE.  See below for descriptions of some of the useful variables.

Defaults for the options are specified in brackets.

Configuration:
  -h, --help              display this help and exit
      --help=short        display options specific to this package
      --help=recursive    display the short help of all the included packages
  -V, --version           display version information and exit
  -q, --quiet, --silent   do not print `checking ...' messages
      --cache-file=FILE   cache test results in FILE [disabled]
  -C, --config-cache      alias for `--cache-file=config.cache'
  -n, --no-create         do not create output files
      --srcdir=DIR        find the sources in DIR [configure dir or `..']

Installation directories:
  --prefix=PREFIX         install architecture-independent files in PREFIX
                          [/usr/local]
  --exec-prefix=EPREFIX   install architecture-dependent files in EPREFIX
                          [PREFIX]

By default, `make install' will install all the files in
`/usr/local/bin', `/usr/local/lib' etc.  You can specify
an installation prefix other than `/usr/local' using `--prefix',
for instance `--prefix=$HOME'.

For better control, use the options below.

Fine tuning of the installation directories:
  --bindir=DIR            user executables [EPREFIX/bin]
  --sbindir=DIR           system admin executables [EPREFIX/sbin]
  --libexecdir=DIR        program executables [EPREFIX/libexec]
  --sysconfdir=DIR        read-only single-machine data [PREFIX/etc]
  --sharedstatedir=DIR    modifiable architecture-independent data [PREFIX/com]
  --localstatedir=DIR     modifiable single-machine data [PREFIX/var]
  --libdir=DIR            object code libraries [EPREFIX/lib]
  --includedir=DIR        C header files [PREFIX/include]
  --oldincludedir=DIR     C header files for non-gcc [/usr/include]
  --datarootdir=DIR       read-only arch.-independent data root [PREFIX/share]
  --datadir=DIR           read-only architecture-independent data [DATAROOTDIR]
  --infodir=DIR           info documentation [DATAROOTDIR/info]
  --localedir=DIR         locale-dependent data [DATAROOTDIR/locale]
  --mandir=DIR            man documentation [DATAROOTDIR/man]
  --docdir=DIR            documentation root [DATAROOTDIR/doc/ImageMagick]
  --htmldir=DIR           html documentation [DOCDIR]
  --dvidir=DIR            dvi documentation [DOCDIR]
  --pdfdir=DIR            pdf documentation [DOCDIR]
  --psdir=DIR             ps documentation [DOCDIR]

Program names:
  --program-prefix=PREFIX            prepend PREFIX to installed program names
  --program-suffix=SUFFIX            append SUFFIX to installed program names
  --program-transform-name=PROGRAM   run sed PROGRAM on installed program names

X features:
  --x-includes=DIR    X include files are in DIR
  --x-libraries=DIR   X library files are in DIR

System types:
  --build=BUILD     configure for building on BUILD [guessed]
  --host=HOST       cross-compile to build programs to run on HOST [BUILD]
  --target=TARGET   configure for building compilers for TARGET [HOST]

Optional Features:
  --disable-option-checking  ignore unrecognized --enable/--with options
  --disable-FEATURE       do not include FEATURE (same as --enable-FEATURE=no)
  --enable-FEATURE[=ARG]  include FEATURE [ARG=yes]
  --enable-silent-rules   less verbose build output (undo: "make V=1")
  --disable-silent-rules  verbose build output (undo: "make V=0")
  --enable-reproducible-build
                          enable reproducible build
  --enable-dependency-tracking
                          do not reject slow dependency extractors
  --disable-dependency-tracking
                          speeds up one-time build
  --enable-compile-warnings=[no/yes/error]
                          Enable compiler warnings and errors
  --disable-Werror        Unconditionally make all compiler warnings non-fatal
  --enable-ld-version-script
                          enable linker version script (default is enabled
                          when possible)
  --enable-bounds-checking
                          enable run-time bounds-checking
  --enable-osx-universal-binary
                          build universal binary on OS X [[default=no]]
  --disable-openmp        do not use OpenMP
  --enable-opencl         enable OpenCL support
  --disable-largefile     omit support for large files
  --enable-shared[=PKGS]  build shared libraries [default=yes]
  --enable-static[=PKGS]  build static libraries [default=yes]
  --enable-fast-install[=PKGS]
                          optimize for fast installation [default=yes]
  --disable-libtool-lock  avoid locking (might break parallel builds)
  --enable-delegate-build look for delegate libraries in build directory
  --disable-deprecated    exclude deprecated methods in MagickCore and
                          MagickWand APIs
  --disable-installed     Formally install ImageMagick under PREFIX
  --disable-cipher        disable enciphering and deciphering image pixels
  --enable-zero-configuration
                          enable self-contained, embeddable,
                          zero-configuration ImageMagick
  --enable-hdri           accurately represent the wide range of intensity
                          levels found in real scenes
  --enable-pipes          enable pipes (|) in filenames
  --disable-assert        disable assert() statements in build
  --enable-maintainer-mode
                          enable make rules and dependencies not useful (and
                          sometimes confusing) to the casual installer
  --enable-hugepages      enable 'huge pages' support
  --enable-ccmalloc       enable 'ccmalloc' memory debug support
  --enable-efence         enable 'efence' memory debug support
  --enable-prof           enable 'prof' profiling support
  --enable-gprof          enable 'gprof' profiling support
  --enable-gcov           enable 'gcov' profiling support
  --enable-legacy-support  install legacy command-line utilities (default disabled)
  --disable-assert        turn off assertions
  --disable-docs          disable building of documentation

Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
  --with-gnu-ld           assume the C compiler uses GNU ld [default=no]
  --with-dmalloc          use dmalloc, as in http://www.dmalloc.com
  --with-gcc-arch=<arch>  use architecture <arch> for gcc -march/-mtune,
                          instead of guessing
  --includearch-dir=DIR   ARCH specific include directory
  --sharearch-dir=DIR     ARCH specific config directory
  --with-pkgconfigdir=DIR Path to the pkgconfig directory [LIBDIR/pkgconfig]
  --without-threads       disable threads support
  --with-pic[=PKGS]       try to use only PIC/non-PIC objects [default=use
                          both]
  --with-aix-soname=aix|svr4|both
                          shared library versioning (aka "SONAME") variant to
                          provide on AIX, [default=aix].
  --with-sysroot[=DIR]    Search for dependent libraries within DIR (or the
                          compiler's sysroot if not specified).
  --with-modules          enable building dynamically loadable modules
  --with-method-prefix=PREFIX
                          prefix MagickCore API methods
  --with-quantum-depth=DEPTH
                          number of bits in a pixel quantum (default 16)
  --with-cache=THRESHOLD  set pixel cache threshhold in MB (default available
                          memory)
  --with-frozenpaths      freeze delegate paths
  --without-magick-plus-plus
                          disable build/install of Magick++
  --with-package-release-name=NAME
                          encode this name into the shared library
  --with-perl             enable build/install of PerlMagick
  --with-perl-options=OPTIONS
                          options to pass on command-line when generating
                          PerlMagick build file
  --with-jemalloc         enable jemalloc memory allocation library support
  --with-umem             enable umem memory allocation library support
  --with-libstdc=DIR      use libstdc++ in DIR (for GNU C++)
  --without-bzlib         disable BZLIB support
  --with-x                use the X Window System
  --without-zlib          disable ZLIB support
  --with-apple-font-dir=DIR
                          Apple font directory
  --with-autotrace        enable autotrace support
  --without-dps           disable Display Postscript support
  --with-dejavu-font-dir=DIR
                          DejaVu font directory
  --without-fftw          disable FFTW support
  --without-flif          disable FLIF support
  --without-fpx           disable FlashPIX support
  --without-djvu          disable DjVu support
  --without-fontconfig    disable fontconfig support
  --without-freetype      disable Freetype support
  --without-raqm          disable Raqm support
  --with-gslib            enable Ghostscript library support
  --with-fontpath=DIR     prepend to default font search path
  --with-gs-font-dir=DIR  Ghostscript font directory
  --with-gvc              enable GVC support
  --without-jbig          disable JBIG support
  --without-jpeg          disable JPEG support
  --without-lcms          disable lcms (v1.1X) support
  --without-openjp2       disable OpenJP2 support
  --without-lqr           disable Liquid Rescale support
  --without-lzma          disable LZMA support
  --without-openexr       disable OpenEXR support
  --without-pango         disable PANGO support
  --without-png           disable PNG support
  --with-rsvg             enable RSVG support
  --without-tiff          disable TIFF support
  --without-webp          disable WEBP support
  --with-windows-font-dir=DIR
                          Windows font directory
  --with-wmf              enable WMF support
  --without-xml           disable XML support

Some influential environment variables:
  CC          C compiler command
  CFLAGS      C compiler flags
  LDFLAGS     linker flags, e.g. -L<lib dir> if you have libraries in a
              nonstandard directory <lib dir>
  LIBS        libraries to pass to the linker, e.g. -l<library>
  CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
              you have headers in a nonstandard directory <include dir>
  CPP         C preprocessor
  CXX         C++ compiler command
  CXXFLAGS    C++ compiler flags
  PKG_CONFIG  path to pkg-config utility
  PKG_CONFIG_PATH
              directories to add to pkg-config's search path
  PKG_CONFIG_LIBDIR
              path overriding pkg-config's built-in search path
  LT_SYS_LIBRARY_PATH
              User-defined run-time library search path.
  CXXCPP      C++ preprocessor
  XMKMF       Path to xmkmf, Makefile generator for X Window System
  ZLIB_CFLAGS C compiler flags for ZLIB, overriding pkg-config
  ZLIB_LIBS   linker flags for ZLIB, overriding pkg-config
  AUTOTRACE_CFLAGS
              C compiler flags for AUTOTRACE, overriding pkg-config
  AUTOTRACE_LIBS
              linker flags for AUTOTRACE, overriding pkg-config
  fftw3_CFLAGS
              C compiler flags for fftw3, overriding pkg-config
  fftw3_LIBS  linker flags for fftw3, overriding pkg-config
  ddjvuapi_CFLAGS
              C compiler flags for ddjvuapi, overriding pkg-config
  ddjvuapi_LIBS
              linker flags for ddjvuapi, overriding pkg-config
  FONTCONFIG_CFLAGS
              C compiler flags for FONTCONFIG, overriding pkg-config
  FONTCONFIG_LIBS
              linker flags for FONTCONFIG, overriding pkg-config
  FREETYPE_CFLAGS
              C compiler flags for FREETYPE, overriding pkg-config
  FREETYPE_LIBS
              linker flags for FREETYPE, overriding pkg-config
  RAQM_CFLAGS C compiler flags for RAQM, overriding pkg-config
  RAQM_LIBS   linker flags for RAQM, overriding pkg-config
  GVC_CFLAGS  C compiler flags for GVC, overriding pkg-config
  GVC_LIBS    linker flags for GVC, overriding pkg-config
  LCMS2_CFLAGS
              C compiler flags for LCMS2, overriding pkg-config
  LCMS2_LIBS  linker flags for LCMS2, overriding pkg-config
  LIBOPENJP2_CFLAGS
              C compiler flags for LIBOPENJP2, overriding pkg-config
  LIBOPENJP2_LIBS
              linker flags for LIBOPENJP2, overriding pkg-config
  LQR_CFLAGS  C compiler flags for LQR, overriding pkg-config
  LQR_LIBS    linker flags for LQR, overriding pkg-config
  LZMA_CFLAGS C compiler flags for LZMA, overriding pkg-config
  LZMA_LIBS   linker flags for LZMA, overriding pkg-config
  OPENEXR_CFLAGS
              C compiler flags for OPENEXR, overriding pkg-config
  OPENEXR_LIBS
              linker flags for OPENEXR, overriding pkg-config
  PANGO_CFLAGS
              C compiler flags for PANGO, overriding pkg-config
  PANGO_LIBS  linker flags for PANGO, overriding pkg-config
  PNG_CFLAGS  C compiler flags for PNG, overriding pkg-config
  PNG_LIBS    linker flags for PNG, overriding pkg-config
  RSVG_CFLAGS C compiler flags for RSVG, overriding pkg-config
  RSVG_LIBS   linker flags for RSVG, overriding pkg-config
  CAIRO_SVG_CFLAGS
              C compiler flags for CAIRO_SVG, overriding pkg-config
  CAIRO_SVG_LIBS
              linker flags for CAIRO_SVG, overriding pkg-config
  XML_CFLAGS  C compiler flags for XML, overriding pkg-config
  XML_LIBS    linker flags for XML, overriding pkg-config

Use these variables to override the choices made by `configure' or to help
it to find libraries and programs with nonstandard names/locations.

Report bugs to <https://github.com/ImageMagick/ImageMagick/issues>.
ImageMagick home page: <http://www.imagemagick.org>.
```
</details>

TL;DR: For IM, there are "optional features" and "optional packages". 

Each "optional feature" has `--enable-FEATURE` or `--disable-FEATURE`. If it is enabled by default, you need to specify `--disable-FEATURE` when building *without* the option; if it is disabled by default, you need to specify `--enable-FEATURE` when building *with* the option.

Each "optional package" has `--without-FEATURE` or `--with-FEATURE`. If it is included by default, you need to specify `--without-FEATURE` when building *without* the option; if it is not included by default, you need to specify `--with-FEATURE` when building *with* the option.

Basically, if you don't want to have two branches for each option, you need to determine whether to pass a `configure` flag when building `--with` or `--without`, and pick precisely one flag out of the four. The enable/with distinction is pretty clear; but the enable/disable and with/without distinctions are murky, and could be tweaked by upstream.

Hopefully it's clear why the old solution is bad now.

> ternary > if/else > current version

My vote: current version > ternary > if/else.

I'd be happy to write a ternary version and go with that.

zmwangx:
> I don't think GitHub is planning on expunging this PR from their databases in the event that you ever feel a burning need for a cheat sheet.

See comment above. This PR is only a cheat sheet for the easy half of the problem; the harder half can't be done without reading `configure --help` again and a lot more manual labor. The point is to avoid that manual labor by passing an appropriate flag in all configurations.

I do have the defaults figured out (locally) at one point, before submitting the previously rejected PR, but I realized how painful it was and discarded the work.

zmwangx:
> without reading configure --help again and a lot more manual labor

Oh, and re-checking `configure --help` once in a while, to make sure the defaults did not change.

MikeMcQuaid:
> I'd be happy to write a ternary version and go with that.

Let's see what the others think first. Thanks for the explanation ❤️

ilovezfs:
I don't mind ternary in a vacuum as long as it doesn't exceed the eightieth column, but it's inevitably not in a vacuum, exceeds the eightieth column, and then we're on to mixed style ternary plus if/else which is generally grotesque. The added formula length of five lines per option doesn't bother me at all when it's used specifically for dealing with a specific known problem case of opportunistic linking *during bottling*.

So nothing plus sparing use of if/else is my preference. I do not think we gain anything by vigorously policing the `--without/--disable` side of things when the consequence is only "correctness" (read: less functionality) of local source builds.

zmwangx:
> I don't mind ternary in a vacuum as long as it doesn't exceed the eightieth column, but it's inevitably not in a vacuum, exceeds the eightieth column, and then we're on to mixed style ternary plus if/else which is generally grotesque.

```
$ grep -rE 'args << \(.*\?.*:.*\)'
Formula/curl.rb:    args << (build.with?("libssh2") ? "--with-libssh2" : "--without-libssh2")
Formula/curl.rb:    args << (build.with?("libidn") ? "--with-libidn" : "--without-libidn")
Formula/curl.rb:    args << (build.with?("libmetalink") ? "--with-libmetalink" : "--without-libmetalink")
Formula/curl.rb:    args << (build.with?("gssapi") ? "--with-gssapi" : "--without-gssapi")
Formula/curl.rb:    args << (build.with?("rtmpdump") ? "--with-librtmp" : "--without-librtmp")
Formula/ecasound.rb:    args << ("--enable-rubyecasound=" + ((build.with? "ruby") ? "yes" : "no"))
Formula/ekg2.rb:    args << (build.with?("libgadu") ? "--with-libgadu" : "--without-libgadu")
Formula/gdal.rb:    args << (build.with?("postgresql") ? "--with-pg=#{HOMEBREW_PREFIX}/bin/pg_config" : "--without-pg")
Formula/gdal.rb:    args << (build.with?("mysql") ? "--with-mysql=#{HOMEBREW_PREFIX}/bin/mysql_config" : "--without-mysql")
Formula/gdal.rb:    args << (build.with?("opencl") ? "--with-opencl" : "--without-opencl")
Formula/gdal.rb:    args << (build.with?("armadillo") ? "--with-armadillo=#{Formula["armadillo"].opt_prefix}" : "--with-armadillo=no")
Formula/gearman.rb:    args << (build.with?("mysql") ? "--with-mysql=#{Formula["mysql"].opt_bin}/mysql_config" : "--without-mysql")
Formula/gearman.rb:    args << (build.with?("hiredis") ? "--enable-hiredis" : "--disable-hiredis")
Formula/gnuplot.rb:    args << ((build.with? "aquaterm") ? "--with-aquaterm" : "--without-aquaterm")
Formula/gnuplot.rb:    args << ((build.with? "x11") ? "--with-x" : "--without-x")
Formula/gtk-mac-integration.rb:    args << ((build.without? "gtk+3") ? "--without-gtk3" : "--with-gtk3")
Formula/libvpx.rb:    args << (build.with?("examples") ? "--enable-examples" : "--disable-examples")
Formula/lldpd.rb:    args << (build.with?("snmp") ? "--with-snmp" : "--without-snmp")
Formula/lldpd.rb:    args << (build.with?("json") ? "--with-json" : "--without-json")
Formula/nut.rb:    args << (build.with?("serial") ? "--with-serial" : "--without-serial")
Formula/nut.rb:    args << (build.with?("libusb-compat") ? "--with-usb" : "--without-usb")
Formula/nut.rb:    args << (build.with?("dev") ? "--with-dev" : "--without-dev")
Formula/nut.rb:    args << (build.with?("net-snmp") ? "--with-snmp" : "--without-snmp")
Formula/nut.rb:    args << (build.with?("neon") ? "--with-neon" : "--without-neon")
Formula/nut.rb:    args << (build.with?("powerman") ? "--with-powerman" : "--without-powerman")
Formula/nut.rb:    args << (build.with?("ipmi") ? "--with-ipmi" : "--without-ipmi")
Formula/nut.rb:    args << (build.with?("libltdl") ? "--with-libltdl" : "--without-libltdl")
Formula/nut.rb:    args << (build.with?("cgi") ? "--with-cgi" : "--without-cgi")
Formula/pdsh.rb:    args << ((build.without? "dshgroups") ? "--without-dshgroups" : "--with-dshgroups")
Formula/powerman.rb:    args << (build.with?("curl") ? "--with-httppower" : "--without-httppower")
Formula/powerman.rb:    args << (build.with?("net-snmp") ? "--with-snmppower" : "--without-snmppower")
Formula/powerman.rb:    args << (build.with?("genders") ? "--with-genders" : "--without-genders")
Formula/qemu.rb:    args << (build.with?("vde") ? "--enable-vde" : "--disable-vde")
Formula/qemu.rb:    args << (build.with?("gtk+") ? "--enable-gtk" : "--disable-gtk")
Formula/qemu.rb:    args << (build.with?("libssh2") ? "--enable-libssh2" : "--disable-libssh2")
```

> The added formula length of five lines per option doesn't bother me at all when it's used specifically for dealing with a specific known problem case of opportunistic linking during bottling.

It looks messy, inconsistent, and is not very logical.

> when the consequence is only "correctness" (read: less functionality) of local source builds.

It sure affects robustness: imagine when you uninstalled something and all of a sudden a non-dependent is broken. Which is precisely what happened for the failed bottle, except you can have random combinations and hence random breakage on users' machines.

zmwangx:
> imagine when you uninstall something and all of a sudden a non-dependent is broken.

By the way, you don't need to imagine. Certainly happened to me more than once before.

ilovezfs:
>It sure affects robustness: imagine when you uninstall something and all of a sudden a non-dependent is broken.

This is the `=> :linked` conversation all over again. I'm bowing out.

zmwangx:
> This is the => :linked conversation all over again.

I'm confused by `:linked` to this day and won't go down that road any further.

DomT4:
> My personal preference is ternary > if/else > current version. It may be helpful for @ilovezfs and @DomT4 to express their preferences too.

This would be fine with me. I don't have a major objection to ternary operators being used more broadly, but I agree with ILZ we really want to protect that 80 character boundary as much as possible, and mixing up style is so often a recipe for contributor confusion. The question there more or less becomes how often using a ternary would violate the 80.

> $ grep -rE 'args << \(.*\?.*:.*\)'

To be fair to ILZ, that's 10 formulae or something. Just happens those 10 formulae use the operator quite heavy, as far as things go. 10/3604 is more or less a statistical vacuum 😄.

zmwangx:
> 10/3604 is more or less a statistical vacuum

```
$ grep -rl 'args <<' | wc -l
430
```

So 10/n where n < 430 (you need to exclude those `args << "CC=#{ENV.cc}"` and stuff.)

jyaworski:
I ran into a problem that this could have helped with, today. I compiled with opencl, and when I went to recompile without it (because it's broken) it kept adding opencl. Using a --without-opencl would have been nice.

MikeMcQuaid:
> This is necessary for preventing screwups like the 6.9.5-9_2 rebuild #0 bottle, which picked up libwmflite-0.2.7.dylib from libwmf, which was then missing due to libwmf not being a recommended dep.

I'm happy to have a fix for this specific case (e.g. `libwmf`). I'm definitely 👎 on doing it for every option.

> make superenv filter /usr/local/lib out if that's the HOMEBREW_PREFIX and require linking only to opt

This is the correct fix, here. I'd rather we didn't paper over it on a formula-by-formula basis.

zmwangx:
> I'm happy to have a fix for this specific case (e.g. libwmf). I'm definitely 👎 on doing it for every option.

You're saying we should wait for opportunistic linking against some random dep, ship a bad bottle, find out later, then add a random patch for that particular dep. Rinse and repeat. Not sure this is the correct attitude when you can solve the problem once and for all elegantly.

Moreover, I can't tell you how many user source builds have opportunistic linking. Mine certainly did when I looked into this issue.

The attitude is certainly not in line with https://github.com/Homebrew/brew/issues/934 (which isn't my idea).

> I'd rather we didn't paper over it on a formula-by-formula basis.

Did I advocate this solution for every formula? No.

Is this approach especially beneficial to the `imagemagick` formula? Yes. `imagemagick` has so many options, it's hard to tell which ones are default-on and which ones are default-off, and defaults seem to change over time. Last time I checked there are wrong defaults in the formula (pretty sure there still are), the most prominent one being #4570. Always specifying with/without or enable/disable means you don't need to check defaults, which is a very tedious process.

> This is the correct fix, here.

There are a lot of talks, but is it going to be available tomorrow, next week, next month, or next year? Do we stick to a subtly broken formula in the mean time (especially the alternative solution also provides additional benefits)?

---

Overall, I've been maintaining the imagemagick formula [for a while now](https://github.com/Homebrew/homebrew-core/pulls?utf8=%E2%9C%93&q=is%3Apr%20imagemagick), and due to the futile debates I've wasted a lot of time. I wouldn't mind seeing fewer imagemagick issues and making my life easier by shipping this PR.

MikeMcQuaid:
> Not sure this is the correct attitude when you can solve the problem once and for all elegantly.

I think the same except the elegant solution is fixing this in the package manager rather than the formula.

If imagemagick adds other optional dependencies in future someone will need to notice and set them up similarly to how you have in this PR. That's not a "once and for all" solution.

> Moreover, I can't tell you how many user source builds have opportunistic linking. Mine certainly did when I looked into this issue.

Opportunistic linking doesn't matter when it's not for binary packages.

> The attitude is certainly not in line with Homebrew/brew#934 (which isn't my idea).

I have no idea why you've brought up that PR again. It's completely unrelated to this PR.

> There are a lot of talks, but is it going to be available tomorrow, next week, next month, or next year? Do we stick to a subtly broken formula in the mean time (especially the alternative solution also provides additional benefits)?

It will be available whenever it gets implemented. You are capable of doing that, too. I don't think this formula is subtly broken, I think there's the potential to ship binary packages that are broken for some people and we should work on the generalised solution rather than focusing on a specific package.

zmwangx:
> Opportunistic linking doesn't matter when it's not for binary packages.

When you uninstall an opportunistically linked dependency of `imagemagick`, `imagemagick` won't work anymore. It's certainly not only a problem for binary packages.

> I have no idea why you've brought up that PR again. It's completely unrelated to this PR.

1. I'm bringing up the issue "Prevent uninstalling in-use dependencies", not any specific implementation of it;
2. It's totally related, see explanation above.

> You are capable of doing that, too.

I'm not obligated to. No one is obligated to. More troublesome, no one has promised to do anything about it.

> rather than focusing on a specific package.

If I can make my life and users' lives easier with this specific package, I will.

And let me say this yet again, if I still wasn't clear enough: the approach in this PR has additional advantages that won't be provided by a generalized solution.

MikeMcQuaid:
> When you uninstall an opportunistically linked dependency of imagemagick, imagemagick won't work anymore. It's certainly not only a problem for binary packages.

This currently applies to `uninstall` as-is but `uninstall` could also be extended to avoid this.

> I'm not obligated to. No one is obligated to. More troublesome, no one has promised to do anything about it.
> If I can make my life and users' lives easier with this specific package, I will.

No-one is obligated to do anything, I agree. We can say that changes won't be included because they aren't a suitable fix, however, which is what I'm doing here. You did agree to be a Homebrew maintainer, however, and that means you've taken some small amount of responsibility to help maintain Homebrew itself and not just the packages you are personally interested in.

> And let me say this yet again, if I still wasn't clear enough: the approach in this PR has additional advantages that won't be provided by a generalized solution.

The underlying problem you cite in the PR body will still be there for other packages and can still occur with this package in future when new options are added and not noticed.

zmwangx:
> The underlying problem you cite in the PR body will still be there for other packages

So this package shouldn't be fixed?

> can still occur with this package in future when new options are added and not noticed.

It won't, and that's the point.

MikeMcQuaid:
> So this package shouldn't be fixed?

It should be fixed: in Homebrew/brew.

> It won't, and that's the point.

2017 rolls around and the new `flibble` library comes out. People install it with `brew install flibble` and a new version of `imagemagick` comes out while @zmwangx is on vacation which requires `--without-flibble` or otherwise opportunistically links against it. Fixing this in Homebrew/brew means `imagemagick` doesn't get to find `flibble`. This approach here means someone needs to notice `--without-flibble` is needed and add it to the list in the formula which further adds to the maintenance/update burden on this formula.

zmwangx:
Overlooked this part:

> You did agree to be a Homebrew maintainer, however, and that means you've taken some small amount of responsibility to help maintain Homebrew itself and not just the packages you are personally interested in.

Evidence I didn't fulfill my "small amount of responsibility"?

I proposed one localized solution here, someone proposes a more generalized solution, so it's suddenly my responsibility to implement the generalized solution? What's the reasoning?

> flibble blah blah blah

That's "new options added to ImageMagick", not "new options added to the `imagemagick` formula". They could also tweak their build system and detect libraries outside your carefully filtered paths (no need for anything fancy — just look inside `/usr/local/lib`). No one can promise this version of the formula will work with the next version of the software.

> the approach in this PR has additional advantages that won't be provided by a generalized solution.

Rephrase again: it is *not* mutually exclusive with the superenv fix.

zmwangx:
By the way, in case you need anything concrete against the "responsibility" accusation:

> not just the packages you are personally interested in.

```
$ git log --pretty=fuller -10000 | grep Commit: | sort | uniq -c | sort -bnr | head -5
   3271 Commit:     Zhiming Wang <zmwangx@gmail.com>
   2825 Commit:     ilovezfs <ilovezfs@icloud.com>
   1852 Commit:     Dominyk Tiller <dominyktiller@gmail.com>
    849 Commit:     Mike McQuaid <mike@mikemcquaid.com>
    234 Commit:     Dominyk Tiller <DomT4@users.noreply.github.com>
```

MikeMcQuaid:
> I proposed one localized solution here, someone proposes a more generalized solution, so it's suddenly my responsibility to implement the generalized solution? What's the reasoning?

You've pointed out a problem and offered a solution. I've said I don't believe this is a good solution to the problem you've pointed out for both this specific formula and in general for all the formulae this problem applies to. What generalised solution would you propose, instead? This approach cannot scale to all formulae.

> That's "new options added to ImageMagick", not "new options added to the imagemagick formula".

Ok, but you agree that that would be a problem that's not fixed by this approach?

> They could also tweak their build system and detect libraries outside your carefully filtered paths (no need for anything fancy — just look inside /usr/local/lib). 

With superenv if we filter this at the compiler/linked level: they can't, really. More realistically: there's no evidence that they would do that and, again, this fixes the issue specifically mentioned as the cause for this PR.

> By the way, in case you need anything concrete against the "responsibility" accusation:

Sorry, that was unfair. To rephrase and be clearer "You did agree to be a Homebrew maintainer, however, and that means you've taken some small amount of responsibility to help fix problems in  Homebrew/brew (the package manager) itself and not just inside the packages/formulae."

zmwangx:
> This approach cannot scale to all formulae.

I never said the general problem shouldn't be fixed, or this approach scales (if I did, scratch that). It's good for `imagemagick`, that's all I ever argued about (or all I insist now).

Also, you seem to be content with the current workaround, which is also a workaround, but far more sketchy. Picking one workaround above another seems random, it's basically "your opinion or mine".

> Ok, but you agree that that would be a problem that's not fixed by this approach?

Sure, and as I said there are also problems not fixed by the generalized approach, that is, the default-on vs default-off problem.

> they can't, really.

If you go that far it's probably more work than necessary, but I won't object.

> this fixes the issue specifically mentioned as the cause for this PR.

The primary cause. Not the secondary cause as I've pointed out time and again.

Also, please use the future tense.

> you've taken some small amount of responsibility to help fix problems in Homebrew/brew (the package manager) itself

I did. But I have the freedom to choose which problems I want to fix. This one will take a significant amount of thought and testing or it will very likely do a lot more harm than good, so I'd rather not be too involved.

MikeMcQuaid:
> I never said the general problem shouldn't be fixed, or this approach scales (if I did, scratch that). It's good for imagemagick, that's all I ever argued about (or all I insist now).
> 
> Also, you seem to be content with the current workaround, which is also a workaround, but far more sketchy. Picking one workaround above another seems random, it's basically "your opinion or mine".

I'd like to politely remind you that even if it is "your opinion or mine" one of us has been working on this project for significantly longer than the other.

> I did. But I have the freedom to choose which problems I want to fix. 

That's my point; this is not a good fix for this problem as it increases the maintenance burden of this formula (worth noting we don't have per-formula maintainers as a thing here; if you usually update it that's good but that's not something we can rely on indefinitely), doesn't scale to other formulae and doesn't stop the same problem being silently reintroduced in future.

> This one will take a significant amount of thought and testing or it will very likely do a lot more harm than good, so I'd rather not be too involved.

I think that's a disappointing attitude.

---

It's clear that you're unwilling to budge on this and my patience has been worn out so: feel free to merge it in whatever format you choose. To be clear, though, I'm extremely 👎 on taking this approach on any single other formula so please do not do so and if it produces additional problems for e.g. other contributors it may be reverted in future. We need to focus on generalised, package-manager solutions to problems like this rather than fixing them on a formula-by-formula basis (which doesn't scale to thousands of formulae).

PR #4711, opened 2016-09-11
mldonkey: upstream fixes for ocaml 4.03

ilovezfs:
add head spec
add test


ilovezfs:
I've requested a new release here: https://github.com/ygrek/mldonkey/issues/19

Issue #4707, opened 2016-09-10
QEMU build fails with "Undefined symbols for architecture x86_64” in sdl.o


karanlyons:
Specifically:

```
Undefined symbols for architecture x86_64:
  "_XFree", referenced from:
      _sdl_keyevent_to_keycode in sdl.o
  "_XGetAtomName", referenced from:
      _sdl_keyevent_to_keycode in sdl.o
  "_XkbFreeKeyboard", referenced from:
      _sdl_keyevent_to_keycode in sdl.o
  "_XkbGetKeyboard", referenced from:
      _sdl_keyevent_to_keycode in sdl.o
```

[Here’re the logs](https://gist.github.com/karanlyons/fbe4d6584717bfb31a2958f21206fc66)

It’s possible we’re just missing libX11?

dunn:
If there's an undeclared X11 dependency, that either needs to be disabled or the formula needs to be in our X11 tap, unfortunately.

zmwangx:
FWIW here's a grep of `X11/` on their cgit: http://git.qemu-project.org/?p=qemu.git&a=search&st=grep&s=X11%2F.

It seems that `XKBlib.h` is included if building with sdl and `SDL_VIDEO_DRIVER_X11` is defined.

karanlyons:
Yeah, so this X11 is only needed with the `--with-sdl` flag, but it still doesn’t build even despite having XQuartz installed (this is likely a search path issue, or at least I hope).

But we may first need to deal with the X11 tap portion of this issue. Given that this formula only has the X11 dependency when built with a flag (and not as a hard requirement for the default build) what’s the preferred way to handle it?

zmwangx:
Actually, if you follow my previous link, `--with-gdk+` might need X11 too.

Could you please try

```ruby
depends_on :x11 if build.with?("sdl") || build.with?("gdk+")
```

and see if it works?

karanlyons:
Thus far I know that `brew install qemu --with-libssh2 --with-vde` will build cleanly.

brew install qemu --with-libssh2 --with-vde --with-sdl` gives the same previously reported errors (concerning, since this time around I tried with `LD_LIBRARY_PATH=/usr/lib:/usr/local/lib:/opt/X11/lib`).

[...]

PR #4725, opened 2016-09-11
qemu: Disable cocoa UI when SDL/GTK+ are enabled.

karanlyons:
As discussed in https://github.com/Homebrew/homebrew-core/issues/4707, avoids build error found in http://git.qemu-project.org/?p=qemu.git;a=blob;f=configure;h=4b808f9d17f640347bb8ffd4033461464eb0b8bb;hb=1dc33ed90bf1fe1c2014dffa0d9e863c520d953a#l1702.

### DIFF
--- a/Formula/qemu.rb
+++ b/Formula/qemu.rb
@@ -49,14 +49,15 @@ class Qemu < Formula
       --disable-guest-agent
     ]
 
-    # Cocoa and SDL UIs cannot both be enabled at once.
-    if build.with? "sdl"
-      args << "--enable-sdl" << "--disable-cocoa"
+    # Cocoa and SDL/GTK+ UIs cannot both be enabled at once.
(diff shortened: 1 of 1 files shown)


zmwangx:
👍

zmwangx:
And this is why no one ever noticed (from our analytics):

screen shot 2016-09-11 at 2 25 50 pm

karanlyons:
The only reason I noticed is because I have a habit of enabling everything I can when doing installs.

zmwangx:
> this is why no one ever noticed

*why no one ever noticed `--with-gtk+` results in configure error.

Reason: `--with-gtk+` are always specified along with `--with-sdl`.

zmwangx:
Merged as 01486d1. Thank you for your contribution to Homebrew!

PR #4730, opened 2016-09-11
zsh-git-prompt 0.5 (new formula)

alyssais:

---

Note: this formula installs the Python implementation of zsh-git-prompt, not the alternative Haskell implementation. I thought about adding this as a build option, but getting it to build was rather difficult so I decided to drop it for now.

I based this formula on existing zsh library formulae, but I'm not sure if the `prefix.install` I saw them use is actually necessary. Seems like there's not really a need for them to be in the main prefix when they can just be loaded from the installation path itself?

### DIFF
--- a/Formula/zsh-git-prompt.rb
+++ b/Formula/zsh-git-prompt.rb
@@ -1,0 +1,25 @@
+class ZshGitPrompt < Formula
+  desc "Informative git prompt for zsh"
+  homepage "https://github.com/olivierverdier/zsh-git-prompt"
+  url "https://github.com/olivierverdier/zsh-git-prompt/archive/v0.5.tar.gz"
+  sha256 "87e5a908369f402e975426ffd61a8800f1c04c0a293f1d4015a6fb1f4408e77d"
+
+  bottle :unneeded
+
+  def install
+    inreplace "README.md", "path/to", opt_prefix
+    prefix.install Dir["*.{sh,py}"]
+  end
+
+  def caveats; <<-EOS.undent
+    For setup instructions:
+      less "#{opt_prefix}/README.md"
+    EOS
+  end
+
+  test do
+    system "git", "init"
+    assert_equal "(%{\e[01;35m%}master%{\e[00m%}|%{…%G%}%{\e[00m%}%{\e[00m%})",
+      shell_output("zsh -c '. #{opt_prefix}/zshrc.sh && git_super_status'").chomp
+  end
+end
--- a/Formula/zsh-git-prompt.rb
+++ b/Formula/zsh-git-prompt.rb
@@ -12,8 +12,11 @@ class ZshGitPrompt < Formula
   end
 
   def caveats; <<-EOS.undent
-    For setup instructions:
-      less "#{opt_prefix}/README.md"
+    First, make sure zsh-git-prompt is loaded from your zshrc:
+      source "#{opt_prefix}/zshrc.sh"
+
+    Then include $(git_super_status) in your PROMPT or RPROMPT, e.g.:
+      PROMPT='%B%m%~%b$(git_super_status) %# '
     EOS
   end
 
--- a/Formula/zsh-git-prompt.rb
+++ b/Formula/zsh-git-prompt.rb
@@ -11,7 +11,7 @@ class ZshGitPrompt < Formula
   end
 
   def caveats; <<-EOS.undent
-    First, make sure zsh-git-prompt is loaded from your zshrc:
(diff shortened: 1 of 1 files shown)


zmwangx:
```ruby
  def caveats; <<-EOS.undent
    For setup instructions:
      less "#{opt_prefix}/README.md"
    EOS
  end
```

I don't think this is a good caveat. A instructional caveat should summarize (rather than delegate) the most basic set up instructions and no more. Maybe something like

```ruby
  def caveats; <<-EOS.undent
    To get started, add the following to your .zshrc:

      source #{opt_prefix}/zshrc.sh

    Then include $(git_super_status) in your PROMPT, e.g.,

      PROMPT='%B%m%~%b$(git_super_status) %# '
    EOS
  end
```

More advanced setup should be left to users to discover on their own. After all, caveats are not supposed be a manual, or a pointer to a manual. It is a place for what's special about the Homebrew installation of the package (and in this case what's special is the path to `zshrc.sh`).

With the above considerations, the inreplace in `README.md` is also not needed.

zmwangx:
Regarding the test: the lines are too long, and the first line is not very readable. The project also doesn't promise stability of `git_super_status` output, so it's prone to breakage (although it could very well be stable, seeing that development has stalled for more than half a year).

I suggest just matching a substring:

```ruby
  test do
    system "git", "init"
    zsh_cmd = "source #{opt_prefix}/zshrc.sh && git_super_status"
    assert_match "master", shell_output("zsh -c '#{zsh_cmd}'")
  end
```

zmwangx:
Putting it together, here's my version:

```ruby
class ZshGitPrompt < Formula
  desc "Informative git prompt for zsh"
  homepage "https://github.com/olivierverdier/zsh-git-prompt"
  url "https://github.com/olivierverdier/zsh-git-prompt/archive/v0.5.tar.gz"
  sha256 "87e5a908369f402e975426ffd61a8800f1c04c0a293f1d4015a6fb1f4408e77d"
  head "https://github.com/olivierverdier/zsh-git-prompt.git"

  def install
    prefix.install Dir["*.{sh,py}"]
  end

  def caveats; <<-EOS.undent
    To get started, add the following to your .zshrc:

      source #{opt_prefix}/zshrc.sh

    Then include $(git_super_status) in your PROMPT, e.g.,

      PROMPT='%B%m%~%b$(git_super_status) %# '
    EOS
  end

  test do
    system "git", "init"
    zsh_cmd = "source #{opt_prefix}/zshrc.sh && git_super_status"
    assert_match "master", shell_output("zsh -c '#{zsh_cmd}'")
  end
end
```

Of course there's room for debate.

penman:
> I don't think this is a very good caveat

@zmwangx I actually rephrased that caveat to be very similar to yours, but I guess I forgot to push it! I'll get that up when I'm back at that computer.

zmwangx:
The caveat hasn't changed... (Test was updated and now it's nice.) Maybe you still haven't pushed it?

penman:
@zmwangx yeah, I did that change from a different computer to the one with the better caveat. :)

penman:
@zmwangx couldn't find the better caveats I wrote (did I dream it…?), but anyway, there are some better ones.

Formula/zsh-git-prompt.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  bottle :unneeded
+
+  def install
+    inreplace "README.md", "path/to", opt_prefix
+    prefix.install Dir["*.{sh,py}"]
+  end
+
+  def caveats; <<-EOS.undent
+    First, make sure zsh-git-prompt is loaded from your zshrc:
```
zmwangx: Use "your .zshrc" to be consistent with caveats of other zsh-related stuff. Other than that I'm happy with the caveats.

Formula/zsh-git-prompt.rb
```diff
@@ -0,0 +1,28 @@
+class ZshGitPrompt < Formula
+  desc "Informative git prompt for zsh"
+  homepage "https://github.com/olivierverdier/zsh-git-prompt"
+  url "https://github.com/olivierverdier/zsh-git-prompt/archive/v0.5.tar.gz"
+  sha256 "87e5a908369f402e975426ffd61a8800f1c04c0a293f1d4015a6fb1f4408e77d"
+
+  bottle :unneeded
+
+  def install
+    inreplace "README.md", "path/to", opt_prefix
```
zmwangx: Remove this.

zmwangx:
Merged as 39ce3f8. Thank you for another contribution to Homebrew!

PR #4734, opened 2016-09-12
libssh2: have HEAD depends on openssl@1.1

JLHwung:
Credits to @vszakats and @tholberton who bring about OpenSSL 1.1 Compatibility to libssh2.

---

### DIFF
--- a/Formula/libssh2.rb
+++ b/Formula/libssh2.rb
@@ -1,8 +1,13 @@
 class Libssh2 < Formula
   desc "C library implementing the SSH2 protocol"
   homepage "https://libssh2.org/"
-  url "https://libssh2.org/download/libssh2-1.7.0.tar.gz"
-  sha256 "e4561fd43a50539a8c2ceb37841691baf03ecb7daf043766da1b112e4280d584"
+
+  stable do
+    url "https://libssh2.org/download/libssh2-1.7.0.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
I'm afraid this is not as simple as the `nginx` PR, because `libssh2` is actually _used_ by other formulae which also have `openssl` dependencies.

That makes it possible to introduce a situation where someone builds `libssh2 --HEAD` against `OpenSSL@1.1` as this PR intends, which causes `cargo` (of `rust`) to then have a direct link to `openssl` and an indirect link to `OpenSSL@1.1`, which is the situation we're very much trying to avoid 😕.

JLHwung:
@DomT4 I'm sorry I thought `libssh2` is okay to migrate as I found it does not depends on any formula which depends on `openssl`, but I failed to check other formula depending on `libssh2` that also depends on `openssl`. Thank you so much for your reminder.

Would there be any case more complicated than having uncle node and nephew node depends on same node `openssl`? Since we can model the dependencies as a forest, would there be any case , for example, that formula A depends on formula B, which depends on an openssl-dependent formula C, B does not depend openssl, but A does depend on openssl. So that is a case when great uncle node and grandson node depends on `openssl`.

I will close the pull request now to wait until all the openssl-dependent downstream of `libssh2` migrating to openssl 1.1.

PR #4736, opened 2016-09-12
cattle 1.2.1

ilovezfs:
build from tag
install examples and tests
add a test


ilovezfs:
@BrewTestBot test this please

PR #3634, opened 2016-08-05
guile 2.0.12

sondr3:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

Also included is the latest release of the development version of Guile.

Issue #2674, opened 2016-07-03
xz broken on OSX 10.11.6 Beta (15G24b) with XCode 8 Beta


x8x:
xz installs fine, but when called it crashes with:
```
Dyld Error Message:
  Symbol not found: _clock_gettime
  Referenced from: /usr/local/opt/xz/bin/xz
  Expected in: /usr/lib/libSystem.B.dylib
```
commenting out this line:
```
#undef HAVE_CLOCK_GETTIME
```
in `xz-5.2.2/config.h.in` fixes it.

I see here https://github.com/Homebrew/homebrew-core/issues/1957 that there are known issues about `clock_gettime` related to OSX 10.12 but not sure why they also apply to OSX 10.11.6 Beta.

Unrelated (sorry!) , have some issue with `python3` but didn't dig in further:
```
==> Upgrading httpie
==> Installing dependencies for httpie: python3
==> Installing httpie dependency: python3
==> Downloading https://www.python.org/ftp/python/3.5.2/Python-3.5.2.tar.xz
Already downloaded: /Users/alpha/Library/Caches/Homebrew/python3-3.5.2.tar.xz
==> ./configure --prefix=/usr/local/Cellar/python3/3.5.2 --enable-ipv6 --datarootdir=/usr/local/Cellar/python3/3.5.2/share --datadir=/usr/local/Cellar
==> make
Last 15 lines from /Users/alpha/Library/Logs/Homebrew/python3/02.make:
  Expected in: /usr/lib/dyld: Symbol not found: _getentropy
  Referenced from: /private/tmp/python3-20160703-71169-isiobw/Python-3.5.2/./Programs/_freeze_importlib
  Expected in: /usr/lib/libSystem.B.dylib

libSystem.B.dylib

dyld: Symbol not found: _getentropy
  Referenced from: /private/tmp/python3-20160703-71169-isiobw/Python-3.5.2/./Programs/_freeze_importlib
  Expected in: /usr/lib/libSystem.B.dylib

/bin/sh: line 1: 78637 Trace/BPT trap: 5       ./Programs/_freeze_importlib ./Lib/importlib/_bootstrap.py Python/importlib.h
make: *** [Python/importlib.h] Error 133
make: *** Waiting for unfinished jobs....
/bin/sh: line 1: 78636 Trace/BPT trap: 5       ./Programs/_freeze_importlib ./Lib/importlib/_bootstrap_external.py Python/importlib_external.h
[...]

PR #4741, opened 2016-09-12
guile: add build hack for Sierra/Xcode 8

DomT4:

---

Also bump the revision to resolve some issues detailed in https://github.com/Homebrew/brew/pull/939.

### DIFF
--- a/Formula/guile.rb
+++ b/Formula/guile.rb
@@ -31,7 +31,7 @@ class Guile < Formula
     depends_on "gettext" => :build
   end
 
-  depends_on "pkg-config" => :build
+  depends_on "pkg-config" => :run # guile-config is a wrapper around pkg-config.
   depends_on "libtool" => :run
   depends_on "libffi"
   depends_on "libunistring"
--- a/Formula/guile.rb
+++ b/Formula/guile.rb
@@ -1,10 +1,10 @@
 class Guile < Formula
-  desc "GUILE: GNU Ubiquitous Intelligent Language for Extensions"
+  desc "GNU Ubiquitous Intelligent Language for Extensions"
   homepage "https://www.gnu.org/software/guile/"
   url "https://ftpmirror.gnu.org/guile/guile-2.0.11.tar.xz"
   mirror "https://ftp.gnu.org/pub/gnu/guile/guile-2.0.11.tar.xz"
   sha256 "aed0a4a6db4e310cbdfeb3613fa6f86fddc91ef624c1e3f8937a6304c69103e2"
-  revision 2
+  revision 3
 
   bottle do
     sha256 "d7e7ad8d491f84c1405b82ee8ef0da5b21f551b6a0f2795bae92e8bec2f19be2" => :el_capitan
@@ -17,6 +17,7 @@ class Guile < Formula
     url "http://git.savannah.gnu.org/r/guile.git",
         :tag => "v2.1.2",
         :revision => "d236022eb0d285af3d462de9e99a212eba459df2"
+
     depends_on "autoconf" => :build
     depends_on "automake" => :build
     depends_on "gettext" => :build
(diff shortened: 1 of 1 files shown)


zmwangx:
Do you want to combine this with #3634 2.0.12 bump?

DomT4:
I'm not really sure how to fix https://github.com/Homebrew/homebrew-core/pull/3634#issuecomment-238026059 to be honest.

The error message wasn't very helpful, and even less helpfully MacPorts merged 2.0.12 without issue, unless they're either not testing 10.9 or nobody has bothered reporting the build failure to them.

zmwangx:
Ah, okay.

> and even less helpfully MacPorts merged 2.0.12 without issue

What's MacPorts? 🙈

zmwangx:
In all seriousness though, indeed quite strange that no one reported anything. If we're desperate we can always install MacPorts on a Mavericks VM and see if it actually works.

DomT4:
Aye. I can always spin up a VM on Mavericks and take a peek later if necessary. I was hoping OP in the other thread would resolve the problem for us 😄.

Formula/guile.rb
```diff
@@ -48,6 +49,15 @@ class Guile < Formula
     cause "Segfaults during compilation"
   end
 
+  # http://debbugs.gnu.org/cgi/bugreport.cgi?bug=23870
+  # https://github.com/Homebrew/homebrew-core/issues/1957#issuecomment-229347476
+  if MacOS.version >= :el_capitan
```
mistydemeo: Is this needed on El Cap as well as Sierra?

zmwangx:
I do love wishful thinking 😉 That said, I fully respect what you've already done.

Formula/guile.rb
DomT4: Not completely sure, to be honest. Wasn't sure if impact would stretch to anything using Xcode 8 with those `clock_gettime` symbols around.

I'm generally a bit terrified of what's going to happen when people on El Cap switch over to Xcode 8 en mass, heh.

mistydemeo: I'll test, I have an El Cap/Xcode 8 VM.

DomT4: Thanks Misty, appreciate that.

mistydemeo: Does fail to build without:

```
  "_clock_getcpuclockid", referenced from:
      _scm_init_stime in libguile_2.0_la-stime.o
```

However, with the patch it instead fails at runtime because it tries to reference `clock_gettime` and it isn't available in 10.11's actual `libSystem`.

mistydemeo: (I say runtime, but it does fail the build since the built `guile` is called by a few targets during `make install`.)

DomT4: Thanks for poking at that. So much fun coming once Xcode 8 is in the wild. 

[This comment](https://github.com/Homebrew/homebrew-core/issues/2674#issuecomment-230395251) makes me wonder if we should mandate CLT installation with Xcode 8 on El Cap, presuming that does actually resolve the issue.

ilovezfs: yuck.

mistydemeo: > This comment makes me wonder if we should mandate CLT installation with Xcode 8 on El Cap, presuming that does actually resolve the issue.

It does, and I think that's a good idea. This is going to get ugly fast otherwise.

Given that this only shifts the problem around, let's avoid applying the patch on El Cap.

DomT4: @mistydemeo Do you have to set the active developer directory to the CLT or is it good enough to simply have the CLT installed?

> Given that this only shifts the problem around, let's avoid applying the patch on El Cap.

👍. I'll tweak it locally before pushing to save the CI another two hour job.

DomT4: Oh, wait, I can't. Was planning to bump the revision to deal with some sandboxing issues. Oh well, another 2 hour CI run it is 😄.

mistydemeo: > Do you have to set the active developer directory to the CLT or is it good enough to simply have the CLT installed?

I'll test this, then submit a PR with the warning.

DomT4: 🙇. Thanks Misty. Massively appreciate everything you've done around 10.12, especially as I've had less time to be here lately.

mistydemeo: Ugh, this did not fix the problem: https://gist.github.com/8ca5bad096d294d9626da1f1a51f61fb

Regardless of whether Xcode or the CLT is the active developer directory.

DomT4:
Heads up that I've queued a Sierra testing job to get a bottle for `guile` now. If it takes ~2 hours, that's normal, don't cancel it 😄. Anything over 2 hours is potentially a bit suspicious.

DomT4:
https://github.com/Homebrew/homebrew-core/commit/4eb35cf5d371cb592fa0803082b3e40bd8c7874b

PR #4376, opened 2016-08-31
jam: add test

bw-yuhogg:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

This adds a test for `jam`.

If this PR is merged, I can probably reuse this test for `ftjam` as well, but one thing at a time.

PR #4750, opened 2016-09-12
ftjam: add test

bw-yuhogg:

---

Pretty much identical to #4376, except that this is for `ftjam` rather than original `jam`.

### DIFF
--- a/Formula/ftjam.rb
+++ b/Formula/ftjam.rb
@@ -18,4 +18,23 @@ class Ftjam < Formula
     system "make"
     system "make", "install"
   end
+
+  test do
+    (testpath/"Jamfile").write <<-EOS.undent
+      Main ftjamtest : ftjamtest.c ;
+    EOS
+
+    (testpath/"ftjamtest.c").write <<-EOS.undent
+      #include <stdio.h>
+
+      int main(void)
+      {
+          printf("FtJam Test\\n");
+          return 0;
+      }
+    EOS
+
+    assert_match "Cc ftjamtest.o", shell_output(bin/"jam")
+    assert_equal "FtJam Test\n", shell_output("./ftjamtest")
+  end
 end


Formula/ftjam.rb
```diff
@@ -18,4 +18,23 @@ def install
...
+      #include <stdio.h>
+
+      int main(void)
+      {
+          printf("FtJam Test\\n");
+          return 0;
+      }
+    EOS
+
+    assert_match /Cc ftjamtest.o/, shell_output(bin/"jam").strip
```
DomT4: If this doesn't need to be a regex just use `""` rather than `//`.

rwhogg: It doesn't actually need to be a regex; about to force-push.

Want me to submit a PR for `jam` to change that one too once this is merged, or should we just leave it?

zmwangx: There are too many of them, so I'd prefer we either fix them en mass or fix only as we update.

DomT4: Don't mind either way, but fixing as things are updated seems cool.

Formula/ftjam.rb
```diff
@@ -18,4 +18,23 @@ def install
...
+      #include <stdio.h>
+
+      int main(void)
+      {
+          printf("FtJam Test\\n");
+          return 0;
+      }
+    EOS
+
+    assert_match "Cc ftjamtest.o", shell_output(bin/"jam").strip
```
zmwangx: Don't really need to `.strip`, we're justing matching a substring.

Formula/ftjam.rb
```diff
@@ -18,4 +18,23 @@ def install
...
+    assert_equal "FtJam Test", shell_output("./ftjamtest").strip
```
zmwangx: What about just matching the unstripped output

```rb
assert_equal "FtJam Test\n", shell_output("./ftjamtest")
```

Sorry for being picky.

rwhogg: Eh, no worries.

DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/057773b7a2ca430769f6dc0143fcdffa046a7a4a. Thank you @rwhogg 🙇

PR #4759, opened 2016-09-13
libtiff: remove now defunct mirror

zmwangx:

---

libtiff has been removed from remotesensing.org. No authoritative word on where it will be hosted next, so not updating the homepage yet.

http://www.asmail.be/msg0054883490.html

CC @mistydemeo @DomT4

### DIFF
--- a/Formula/libtiff.rb
+++ b/Formula/libtiff.rb
@@ -2,7 +2,6 @@ class Libtiff < Formula
   desc "TIFF library and utilities"
   homepage "http://www.remotesensing.org/libtiff/"
   url "http://download.osgeo.org/libtiff/tiff-4.0.6.tar.gz"
-  mirror "ftp://ftp.remotesensing.org/pub/libtiff/tiff-4.0.6.tar.gz"
   sha256 "4d57a50907b510e3049a4bba0d7888930fdfc16ce49f1bf693e5b6247370d68c"
   revision 2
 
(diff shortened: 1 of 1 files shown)


DomT4:
May be worth replacing it with https://mirrors.ocf.berkeley.edu/debian/pool/main/t/tiff/tiff_4.0.6.orig.tar.gz so there is a mirror, we try to tie together plaintext URLs with SSL/TLS mirrors, but anyway, 👍 on removing the current mirror one way or another.

zmwangx:
@DomT4 Sounds good. There are three Debian mirrors used in core: https://mirrors.ocf.berkeley.edu/debian/, https://mirrors.kernel.org/debian/, and https://mirrorservice.org/sites/ftp.debian.org/debian/. Which one do you prefer?

DomT4:
Heh. There's actually a story on this, believe it or not. For a while, universally we used https://mirrors.kernel.org but then Chrome pushed forwards with marking SSL/TLS with SHA1 certs in the chain as semi-problematic, so I started deprecating its usage in formulae... and then months later the cert chain got fixed and Chrome stopped whinging 😄.

Anyway, to actually answer the question, preference is Berkeley followed by mirrorservice.

zmwangx:
Not a fan of Berkeley (for reasons obvious to a certain demographic), but that's a personal issue 😜 Added.

PR #4783, opened 2016-09-13
gcc: Use stable version number + 1 as version number for --HEAD builds

ts826848:

---

Before, the version string would be empty, resulting in names like gcc-
and g++-, which tripped up brew when looking for a gcc install. This,
coupled with a tweak to brew to allow HEAD as a gcc version string,
should allow HEAD builds to be recognized

### DIFF
--- a/Formula/gcc.rb
+++ b/Formula/gcc.rb
@@ -73,7 +73,11 @@ class Gcc < Formula
   end
 
   def version_suffix
-    version.to_s.slice(/\d/)
+    if build.head?
+      ("#{stable.version}".slice(/\d/).to_i + 1).to_s
+    else
+      version.to_s.slice(/\d/)
+    end
   end
 
   # Fix for libgccjit.so linkage on Darwin
--- a/Formula/gcc.rb
+++ b/Formula/gcc.rb
@@ -74,7 +74,7 @@ class Gcc < Formula
 
   def version_suffix
     if build.head?
-      ("#{stable.version}".slice(/\d/).to_i + 1).to_s
+      (stable.version.to_s.slice(/\d/).to_i + 1).to_s
     else
       version.to_s.slice(/\d/)
     end


zmwangx:
Updated your title because it sounded generic and I was wondering from the outside (PR list) why a generic PR was made on homebrew-core instead of brew.

zmwangx:
I killed the build by the way because the change doesn't affect what is actually tested on the CI, and it's kind of pointless to reinstall all those dependents which could take a long long time.

aw1621107:
Oops, completely forgot about the titling scheme over here. Thanks for fixing it!

apjanke:
:+1: on this.

Would be nice if the commit message was changed to use the same format as the PR title, too.

aw1621107:
Alright, updated the commit message. Switched a few words around to keep it under the 50 character limit, but should be similar enough

aw1621107:
And pushed the updated version. `--HEAD` builds now result in `gcc-7`, `g++-7`, etc. being installed.

zmwangx:
Why the close?

aw1621107:
Oh, whoops. Thought this was done for some reason. Give me a second...

DomT4:
Where was this left? Were people happy or?

aw1621107:
Not entirely sure? The change to brew core got merged, so I think this should be ok?

MikeMcQuaid:
I'm good with this, thanks @aw1621107!

MikeMcQuaid:
Just want to see CI run once first.

MikeMcQuaid:
@BrewTestBot test this please

Formula/gcc.rb
```diff
@@ -73,7 +73,11 @@ def osmajor
   end
 
   def version_suffix
-    version.to_s.slice(/\d/)
+    if build.head?
+      ("#{stable.version}".slice(/\d/).to_i + 1).to_s
```
MikeMcQuaid: `stable.version.to_s` instead then this is good to 🚢

MikeMcQuaid (changes requested):

Formula/gcc.rb
aw1621107: Shouldn't it be `(stable.version + 1).to_s?`

aw1621107: Or `(stable.version.to_i + 1).to_s`. Still not familiar with how things work...

MikeMcQuaid: Still need the `slice` to get only the first digit.

aw1621107: Oh, I think I see what you're getting at. Fixed, I hope

MikeMcQuaid:
@aw1621107 Why close this?

aw1621107:
Dangit, thought it was done for some reason. My mistake.

DomT4:
Happy here unless Mike wants a clean CI run to check we haven't borked stable or such.

MikeMcQuaid:
@DomT4 Yeh, going to do that. @zmwangx please don't abort it this time, thanks.

MikeMcQuaid (approved):

MikeMcQuaid:

ilovezfs:
@aw1621107 @MikeMcQuaid This has broken HEAD builds of formulae that need gcc.

```
iMac-TMP:homebrew-science joe$ brew install --HEAD lightstringgraph
==> Installing lightstringgraph from homebrew/science
==> Installing dependencies for homebrew/science/lightstringgraph: beetl
==> Installing homebrew/science/lightstringgraph dependency: beetl
==> Downloading https://homebrew.bintray.com/bottles-science/beetl-0.10.0.el_capitan.bottle.tar.gz
==> Pouring beetl-0.10.0.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/beetl/0.10.0: 27 files, 4.7M
==> Installing homebrew/science/lightstringgraph
==> Using the sandbox
Error: No available formula with the name "gcc6" 
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
This formula was found in a tap:
homebrew/versions/gcc6
To install it, run:
  brew install homebrew/versions/gcc6
```

aw1621107:
@ilovezfs That's... interesting. If the stable version of gcc was installed, it should have exactly the same behavior as before this change. The name should also be `gcc-6`. Not sure where the hyphen went.

aw1621107:
@MikeMcQuaid @ilovezfs I think I see what's going on. I'll have a PR shortly.

PR #4813, opened 2016-09-14
rtags: Added plist support.

mpark:

---

This is based on:
https://github.com/Andersbakken/rtags#integration-with-launchd-mac-os-x

### DIFF
--- a/Formula/rtags.rb
+++ b/Formula/rtags.rb
@@ -36,6 +36,36 @@ class Rtags < Formula
     end
   end
 
+  plist_options :manual => "#{HOMEBREW_PREFIX}/bin/rdm --verbose --inactivity-timeout=300 --log-file=#{HOMEBREW_PREFIX}/var/log/rtags.log"
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+    <dict>
+      <key>Label</key>
+      <string>#{plist_name}</string>
+      <key>ProgramArguments</key>
+      <array>
+        <string>#{bin}/rdm</string>
+        <string>--verbose</string>
+        <string>--launchd</string>
+        <string>--inactivity-timeout=300</string>
+        <string>--log-file=#{var}/log/rtags.log</string>
+      </array>
+      <key>Sockets</key>
+      <dict>
+        <key>Listener</key>
+        <dict>
+          <key>SockPathName</key>
+          <string>#{ENV["HOME"]}/.rdm</string>
+        </dict>
+      </dict>
+    </dict>
+    </plist>
+    EOS
+  end
+
   test do
     mkpath testpath/"src"
     (testpath/"src/foo.c").write <<-EOS.undent


Formula/rtags.rb
```diff
@@ -28,6 +28,45 @@ def install
...
 
+  def caveats; <<-EOS.undent
+    The instance of rtags that run via brew services listens to
+    the socket file at #{var}/.rdm
+
+      rc --socket-file=#{var}/.rdm ...
+    EOS
+  end
+
+  plist_options :startup => false
```
DomT4: `false` doesn't need stating. Usually if not `plist_options :startup` you'd set something like:
```ruby
plist_options :manual => "rdm --log-file=#{var}/log/rtags.log --socket-file=#{var}/.rdm"
```
Or such.

Formula/rtags.rb
```diff
@@ -28,6 +28,45 @@ def install
     end
   end
 
+  def caveats; <<-EOS.undent
+    The instance of rtags that run via brew services listens to
+    the socket file at #{var}/.rdm
```
DomT4: Feels a bit questionable to have it be a hidden directory. Any particular reason for that?

mpark: Ah, I just kept the name from the default value:
```
--socket-file|-n [arg]                 Use this file for the server socket (default ~/.rdm).
```
Would it be better to just use `rdm` instead?

DomT4: Hidden directories are preferred in `HOME` by users, but a lot less common in places like `var`. `#{var}/rdm` sounds 👍 to me.

mpark: > Hidden directories are preferred in HOME by users
Using the default value of `~/.rdm` would actually improve usability I think, since the simple
invocation of `rc` will work without having to specify `rc --socket-file=#{var}/rdm`.

But I'm not sure if this is good practice. Is there a policy or a convention for home-brew
packages around this?

MikeMcQuaid: Just use the default value 👍

DomT4: Basically, as long as you're not relying on Homebrew being able to write to that prefix during install or test, that's fine to use the default.

mpark: Great!

I need some help with filling out this part:
```
<dict>
  <key>SockPathName</key>
  <string>#{var}/.rdm</string>
</dict>
```
What can I replace `#{var}` with, to get the user's `$HOME` directory?

MikeMcQuaid: `ENV["HOME"]` but what happens if you just don't specify this argument?

mpark: @MikeMcQuaid: As far as I understand, if this argument is not specified, `launchd` won't be listening to this socket for activity. The `--inactivity-timeout` flag in conjunction with this argument is what makes `rtags` run on-demand.

Formula/rtags.rb
mpark: Ah, I see. I got `plist_options :startup => false` from the formula for `kwm`.
For this plist configuration, we can just leave it out, since `launchd` listens to the socket at `~/.rdm` for activity and launches `rdm` automatically.

Formula/rtags.rb
MikeMcQuaid: Cool, seems fine then.

dunn:
@BrewTestBot test this please

dunn:
OK, just an audit failure:
```
        Error: 1 problem in 1 formula
rtags:
  * Please set plist_options when using a formula-defined plist.
```
Take a look at examples like [`bind.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bind.rb) and [`collectd.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/collectd.rb).

ilovezfs:
It's also run-time failure on El Capitan.
```
==> /usr/local/Cellar/rtags/2.3/bin/rc -c
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/rtags/2.3/bin/rc
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/rtags/2.3/bin/rc
  Expected in: /usr/lib/libSystem.B.dylib

```

ilovezfs:
@mpark run-time failure is now fixed by https://github.com/Homebrew/homebrew-core/pull/6118 so if you rebase your PR on master and fix the audit issue mentioned above it should be :green_apple:

MikeMcQuaid:

PR #4824, opened 2016-09-15
ponyc 0.3.1

SeanTAllen:

PR #4817, opened 2016-09-14
ponyc 0.3.1

SeanTAllen:

PR #4826, opened 2016-09-15
ponyc 0.3.1

SeanTAllen:


SeanTAllen:
We had a couple changes that make it difficult to build on Mavericks for homebrew (the mavericks clang is too old and doesn't have stdatomic.h installed) so I've set yosemite and up as a dependency. It took me a while to figure out how to get this to work (a few hours) because I'm not well versed in homebrew. If you would like an updated commit message to reflect this, let me know. Thanks.

tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this one. Sorry for the trouble you had with the version dependency.

UniqMartin:
For reference: #4817 and #4824 were the predecessors of this PR.

PR #4828, opened 2016-09-15
mfterm 1.0.7 (new formula)

zhovner:

---

Adding this to make complete toolset for working with Mifare cards.
Related to [mfoc](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mfoc.rb) and [mfcuk](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mfcuk.rb) that already in brew.

mfterm is packaged in [Kali Linux](http://tools.kali.org/wireless-attacks/mfterm). 

Maybe Dominyk @DomT4 can endorse it.

It has not enough stars for pass audit.

### DIFF
--- a/Formula/mfterm.rb
+++ b/Formula/mfterm.rb
@@ -1,9 +1,9 @@
 class Mfterm < Formula
   desc "Terminal for working with Mifare Classic 1-4k Tags"
   homepage "https://github.com/4ZM/mfterm"
-  url "https://github.com/4ZM/mfterm/archive/v1.0.6.tar.gz"
-  version "1.0.6"
-  sha256 "19adfe858cfa5c672aa543b1c6b439384597cb21f880ed74c9b26e5164006221"
+  url "https://github.com/4ZM/mfterm/releases/download/v1.0.7/mfterm-1.0.7.tar.gz"
+  version "1.0.7"
+  sha256 "b6bb74a7ec1f12314dee42973eb5f458055b66b1b41316ae0c5380292b86b248"
 
   head do
     url "https://github.com/4ZM/mfterm.git"
--- a/Formula/mfterm.rb
+++ b/Formula/mfterm.rb
@@ -2,7 +2,6 @@ class Mfterm < Formula
   desc "Terminal for working with Mifare Classic 1-4k Tags"
   homepage "https://github.com/4ZM/mfterm"
   url "https://github.com/4ZM/mfterm/releases/download/v1.0.7/mfterm-1.0.7.tar.gz"
-  version "1.0.7"
   sha256 "b6bb74a7ec1f12314dee42973eb5f458055b66b1b41316ae0c5380292b86b248"
 
   head do
--- a/Formula/mfterm.rb
+++ b/Formula/mfterm.rb
@@ -1,0 +1,34 @@
+class Mfterm < Formula
+  desc "Terminal for working with Mifare Classic 1-4k Tags"
+  homepage "https://github.com/4ZM/mfterm"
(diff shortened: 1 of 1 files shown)


Formula/mfterm.rb
```diff
@@ -0,0 +1,34 @@
+class Mfterm < Formula
+  desc "Terminal for working with Mifare Classic 1-4k Tags"
+  homepage "https://github.com/4ZM/mfterm"
+  url "https://github.com/4ZM/mfterm/releases/download/v1.0.7/mfterm-1.0.7.tar.gz"
+  version "1.0.7"
```
tdsmith: You should remove this line.

tdsmith:
I'll wait for @DomT4 to pronounce but one way you could use Homebrew to install this software is to [maintain your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md). The linked document has some advice; please let us know if you have any questions and we'll be happy to help you set things up.

Formula/mfterm.rb
zhovner: Done.

zmwangx:
> GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)

We have to cut a notability line somewhere and I think our current standard is pretty reasonable (50 stars really is very low on GitHub). If Dom would endorse it then sure; otherwise, I don't see how being "related" to mfoc and mfcuk (from nfc-tools) makes it worth inclusion.

Agree with Tim, might be better to maintain it in your own tap.

zhovner:
>I don't see how being "related" to mfoc and mfcuk (from nfc-tools) makes it worth inclusion

Because after you crack keys with  mfoc/mfcuk you need something to manipulate data on the card. I see this as logical continuation.

zmwangx:
> Because after you crack keys with mfoc/mfcuk you need something to manipulate data on the card. I see this as logical continuation.

But anyone can write a "related" utility. If this mfterm is from nfc-tools or otherwise endorsed by them then I would probably consider it a good argument.

DomT4:
I don't have a strong opinion on this, to be honest. It's one of those things that arguments can be made for either way fairly easily. I'm happy to go with any broader team decision here.

It's not a million miles off eligibility, but I'd perhaps lean towards saying this isn't quite in that buffer zone of being close enough to merit dismissing the warning entirely.

Kali is quite a specialist OS, so it doesn't particularly surprise me they'd carry this as it fits their use case perfectly. There's more merit to the packaged elsewhere argument if it's somewhere more generic, like Debian, MacPorts, etc.

zhovner:
>here's more merit to the packaged elsewhere argument if it's somewhere more generic, like Debian, MacPorts, etc.

No any other distro (except Kali Linux) package very specific hack tools like mfoc/mfcuk.
But brew does. So why not complete this toolset?

zmwangx:
Unfortunately they do meet our eligibility requirements, while this one doesn't. We didn't change our code to reject this formula, did we?

zhovner:
Okay, if the only problem is stars quantity, we can wait until it gets enough.
I think one post on ycombinator or slashdot will resolve this.

bfontaine:
It now has 51 stars.

bfontaine:
@BrewTestBot test this please.

MikeMcQuaid:

For future reference the preferred commit message format for simple version updates is `mfterm 1.0.7 (new formula)`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #4836, opened 2016-09-15
qjackctl 0.4.3

ilovezfs:
prevent opportunistic linking to portaudio
remove upstreamed patch
don't needlessly run autogen


dunn:
For posterity:
```
In file included from In file included from qjackctl.cpp:24:
./qjackctlMainForm.h:34:10: fatal error: 'jack/jack.h' file not found
#include <jack/jack.h>
         ^
qjackctlAlsaConnect.cpp:24:
./qjackctlMainForm.h:34:10: fatal error: 'jack/jack.h' file not found
#include <jack/jack.h>
         ^

ilovezfs:
@dunn thanks, fixed.

Issue #4841, opened 2016-09-15
Popular formulae with bottling trouble on 10.12


zmwangx:
Here's a list of popular formulae (among the 500 most installed according to our analytics) that had trouble being bottled for Sierra. The checklist is followed by a table with more details. Please help us fix your favorite formula(e)!

- [ ] xctool (https://github.com/facebook/xctool/issues/716, asked Sep 30, 2016)
- [ ] swiftlint (https://github.com/realm/SwiftLint/issues/815)
- [ ] pypy ([upstream issue](https://bitbucket.org/pypy/pypy/issues/2407/pypy-541-fails-to-build-on-macos-sierra))
- [x] sbcl (#5125 release planned)
- [x] erlang
- [x] elixir
- [x] ilmbase
- [x] rust
- [x] pyqt (no Sierra for you)
- [x] android-sdk (failure not related to 10.12)
- [x] gdal
- [x] libspatialite (caveat: `check_sql_stmt` check fails — happens on 10.10 and 10.11 too — and we ignored that)
- [x] libressl
- [x] haskell-stack (https://github.com/commercialhaskell/stack/issues/2577)
- [x] packer (failure not related to 10.12)
- [x] valgrind (not Sierra-ready at the moment)
- [x] postgis
- [x] bazel
- [x] boot2docker
- [x] cocoapods
- [x] wxpython
- [x] mpg123

Qt shitshow:

- [x] qt (no Sierra for you)
- [ ] qt5
- [ ] qscintilla2
- [ ] pyqt5
- [ ] qwt

## Details

Note:

1. Build numbers are for https://bot.brew.sh/job/Homebrew%20Sierra%20Testing/.

2. The table is also in [this gist](https://gist.github.com/zmwangx/ba639b206b88cdb1f09d1b446747792b) where you can search in the tsv.

| Rank | Formula       | Build # | Deps satisfied | Failure stage | Comment                                                                                                       | 
|------|---------------|---------|----------------|---------------|---------------------------------------------------------------------------------------------------------------| 
| 90   | qt5           | 744     | ✔︎             | install       |                                                                                                               | 
| 144  | xctool        | 744     | ✔︎             | install       |                                                                                                               | 
| 226  | swiftlint     | 750     | ✔︎             | install       |                                                                                                               | 
| 240  | qscintilla2   | -       | ✗              | -             | depends on qt5                                                                                                 | 
| 317  | packer        | 751     | ✔︎             | test          |                                                                                                               | 
[...]

PR #4850, opened 2016-09-15
pyqwt: migrate to boneyard

zmwangx:
- Latest release from 2009;
- No Qt5 support;
- 450 installs since the beginning of analytics;
- Blocking Qt5 migration of `qwt`, which is reasonably popular, and `qwt` can't be built on Sierra due to this reason.

Ref: #1705, #4841.


zmwangx:
CC @ilovezfs.

PR #4854, opened 2016-09-15
emacs: support switching to the Spacemacs icon

tchajed:
This obviates the need to tap d12frosted/homebrew-emacs-plus and install the emacs-plus formula, which is a copy of homebrew-core's emacs.rb.

### DIFF
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -32,6 +32,7 @@ class Emacs < Formula
   option "with-cocoa", "Build a Cocoa version of emacs"
   option "with-ctags", "Don't remove the ctags executable that emacs provides"
   option "without-libxml2", "Don't build with libxml2 support"
+  option "with-spacemacs-icon", "Use the Spacemacs Emacs icon"
 
   deprecated_option "cocoa" => "with-cocoa"
   deprecated_option "keep-ctags" => "with-ctags"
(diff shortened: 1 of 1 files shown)


dunn:
That tap was created because we've rejected this option before; see https://github.com/Homebrew/legacy-homebrew/pull/49371

We try to limit the number of options on formulae in order make them easier to maintain (both by humans and computers, since our CI doesn't test non-default options).

So I'm afraid we're going to pass on this again, sorry!

tchajed:
Ah, sorry, didn't find a reference to it because I didn't think to search legacy-homebrew.

The actual reason why I wanted this was because proof-general depends on emacs, which I don't have because I have emacs-plus. It seems like manually switching to the Spacemacs icon is the best way to go.

For future reference, after `brew linkapps emacs` you can do so with

```
curl -L "https://github.com/nashamri/spacemacs-logo/blob/master/spacemacs.icns?raw=true" -o /Applications/Emacs.app/Contents/Resources/Emacs.icns
```

dunn:
Cool, glad you found a solution.

The Emacs provided by the emacs-plus tap should satisfy the `depends_on :emacs` requirement, though.  Homebrew's Emacs ought to be installed only when there's no `emacs` in the PATH, or when the Emacs found is too old.  Does Spacemacs have a different version scheme?

tchajed:
Ah, that does seem to work. I'm not sure what state my $PATH was in.

`emacs-plus` is exactly the same binary as `emacs`. it doesn't even include Spacemacs, it only has a different icon. Spacemacs still has to be installed to `~/.emacs.d` manually.

PR #4869, opened 2016-09-15
readline 7.0

ilovezfs:


ilovezfs:
... expecting this to fail of course since revision bumps will also be needed, but one thing at a time ...

ilovezfs:
Locally all immediate dependents built except
- [ ] mdbtools
- [ ] postgresql
- [ ] sdcv
- [ ] sqsh
- [ ] yap

I have not checked the logs yet to see why.

ilovezfs:
sqsh is already broken having nothing to do with readline 7
```
cmd_connect.c:864:23: error: use of undeclared identifier 'CS_TDS_80'
            version = CS_TDS_80;
                      ^
cmd_connect.c:1261:22: error: use of undeclared identifier 'CS_TDS_80'
                case CS_TDS_80:
```

ilovezfs:
yap is already broken unless built with gcc4.x

ilovezfs:
OK, all looks good-to-go here as soon as CI can handle the load of the revisions.

ilovezfs:
@JCount feel free to test this PR :)

JCount:
@ilovezfs Haha. I I'll keep an eye on the CI as it runs and see if I can troubleshoot any of the breakage that may (will) occur that are rev deps.

PR #4870, opened 2016-09-16
x264 r2699

JCount:

---

update stable to r2699, devel to r2705;
remove option to build gpac mp4 output:
non-functional & would require static build of gpac (impractical);
formula reconfigured and cleaned up to reflect this

---

I felt it would simply be better to remove the `gpac` build option because x264 currently requires `libgpac_static` which seemingly needs to be built directly from a git repo source, otherwise it fails.

If it becomes/is possible to simply modify the `gpac` formula to build the static lib as a configuration option, then it might make sense to reenable the `gpac` mp4 output.

Also, the formula no longer automatically builds against an installation of `l-smash` unless the option is explicitly enabled when building from source.

### DIFF
--- a/Formula/x264.rb
+++ b/Formula/x264.rb
@@ -2,9 +2,8 @@ class X264 < Formula
   desc "H.264/AVC encoder"
   homepage "https://www.videolan.org/developers/x264.html"
   # the latest commit on the stable branch
-  url "https://git.videolan.org/git/x264.git", :revision => "fd2c324731c2199e502ded9eff723d29c6eafe0b"
-  version "r2668"
-
+  url "https://git.videolan.org/git/x264.git", :revision => "a5e06b9a435852f0125de4ecb198ad47340483fa"
+  version "r2699"
   head "https://git.videolan.org/git/x264.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


zmwangx:
Merged as 02e3e99. Thank you for another contribution to Homebrew!

For future reference, when you also bump devel, your commit subject should reflect that, so the correct subject should have been `x264 r2699, r2705 (devel)`.

JCount:
Ok, I will do that in the future. (It might be helpful to update/change 8. in [How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#create-your-pull-request-from-a-new-branch) since it currently states there to only show the `stable` version when updating both.) I did get rid of the line between the `version` tag and the `head` spec so I think I am slowly learning. 😄 

As to the installs with the bare `--with-mp4=`, I believe they are installs of `x264` from source with no additional configuration args. There are probably bare `--with-mp4=` args in any install of the formula which was built from source; the formula would be built with `--with-mp4=(gpac/l-smash) --with-mp4=` for example.

ilovezfs:
@JCount FWIW, you can run `brew pull 4870` and it will issue warnings about the subject line if it's "wrong"

JCount:
oh, thanks. Thats a handy tip!

zmwangx:
> --with-mp4=

I think we record complete options like `--with-mp4=gpac` for our analytics actions, so I still believe `--with-mp4=` is from confused users who didn't specify a value. But I could be wrong.

JCount:
I personally noticed instances where both `--with-mp4=gpac` and `--with-mp4=` were passed to `brew` when I ran `brew reinstall x264 --with-mp4=gpac` on a version built from source. I wonder if the way the formula and then the `brew` arguments were parsed created a pseudo "default" value for the argument of a `null` string in some weird edge cases, so homebrew would then end up tracking `--with-mp4=` as a specified argument in some instances.

zmwangx:
You mean the options tracked in `INSTALL_RECEIPT.json`? The values aren't included there. See https://github.com/Homebrew/brew/pull/566.

JCount:
Well, for what it's worth I did observe `Reinstalling x264 with --with-mp4=gpac --with-mp4=` when I know for a fact that I never explicitly passed `--with-mp4=`. The rest probably came about from the way the formula was written and the, for me, mysterious internals of homebrew. Any more, I know not.  😑

PR #4877, opened 2016-09-16
erlang: unrecommend wxmac

zmwangx:

---

`wxmac` stable does not work on Sierra yet, we cannot recommend it.

**Update.** We'll just unrecommend it across the board. See discussion below.

This should resolve our bottling problem for both `erlang` and `elixir`.

### DIFF
--- a/Formula/erlang.rb
+++ b/Formula/erlang.rb
@@ -4,6 +4,7 @@
 class Erlang < Formula
   desc "Programming language for highly scalable real-time systems"
   homepage "https://www.erlang.org/"
+  revision 1
   head "https://github.com/erlang/otp.git"
 
   stable do
@@ -34,7 +35,7 @@ class Erlang < Formula
   depends_on "openssl"
   depends_on "fop" => :optional # enables building PDF docs
   depends_on :java => :optional
-  depends_on "wxmac" => :recommended # for GUI apps like observer
+  depends_on "wxmac" => :optional # for GUI apps like observer
 
   fails_with :llvm
 


ilovezfs:
I'd prefer this be fixed or made optional for all the OSs.

zmwangx:
@ilovezfs "be fixed" seems to require *us* to make a release decision for wxWidgets/wxWidgets.

ilovezfs:
Please explain further.

Also note that erlang is broken on 10.11 + Xcode 8, which needs to be fixed too, preferably as part of this same PR.

zmwangx:
> Please explain further.

We need to move `wxmac` 3.1.0 from devel to stable.

> Also note that erlang is broken on 10.11 + Xcode 8, which needs to be fixed too, preferably as part of this same PR.

No idea on that.

ilovezfs:
>We need to move wxmac 3.1.0 from devel to stable.

Is there an upstream issue requesting a new stable release?

zmwangx:
This is a clumsy enough project with a Trac I doubt "requesting" works.

ilovezfs:
>This is a clumsy enough project with a Trac I doubt "requesting" works.

Then let's just make it optional across the board and monitor the analytics to see if anyone actually cares about the option anyway.

zmwangx:
> Then let's just make it optional across the board.

Sure.

zmwangx:
Do you think we should bump the revision? Technically it's the correct thing to do, but realistically I feel bad about forcing an upgrade that actually removes functionality.

ilovezfs:
It needs to be revision bumped if the contents are changing.

zmwangx:
Yeah of course, that's the "correct" thing. Done.

dch:
wx is part of a heavily used component called observer, a monitoring/management/debugging tool to connect to remote and local nodes. While apps like couchdb, rabbitmq etc won't care, this will be a surprise to many users & developers. Unless there's a simple patch to fix wx, can we add a notice that https://www.erlang-solutions.com/resources/download.html provides a prebuilt download that will presumably work correctly on Sierra? I can't test it as I've not upgraded yet.

zmwangx:
> this will be a surprise to many users & developers.

Said users and developers should unleash their anger on https://github.com/wxWidgets/wxWidgets (http://wxwidgets.org/). They haven't had a stable release in two years.

> can we add a notice that https://www.erlang-solutions.com/resources/download.html provides a prebuilt download that will presumably work correctly on Sierra?

This kind of notice is unheard of, but I'm not particularly against it.

Also, note that you can get erlang with wx still (I believe):

```
$ brew install wxmac --devel
$ brew install erlang --with-wxmac
```

zmwangx:
I think a more reasonable notice is to warn the lack of wx when `--with-wxmac` is not specified, and explain the situation.

ilovezfs:
@zmwangx I'm inclined to take @dch's concerns seriously. What's the minimal equired patch at this point to get wxmac working on Sierra?

zmwangx:
I haven't been able to get it to work.

ilovezfs:
http://trac.wxwidgets.org/changeset/f6a2d1caef5c6d412c84aa900cb0d3990b350938/git-wxWidgets is bogus?

zmwangx:
Last time I tried it failed, but I could have screwed up something because I was mostly paying attention to valgrind at that time. I'll try it again.

zmwangx:
@ilovezfs Hah, it actually seems to work, not sure what I did last time. I'm using Terminal.app on 10.12 so I don't have the logs of that session.

Hopefully there's a happy ending for everyone.

zmwangx:
I think I might have commented out the inline Yosemite patches or something last time, but it doesn't matter now.

ilovezfs:
@zmwangx excellent

zmwangx:
@dch We fixed building erlang on 10.12 without compromising. Thanks for the push on the right direction.

PR #4881, opened 2016-09-16
yap: migrate to boneyard

ilovezfs:


ilovezfs:
needs gcc 4. last stable was in 2010. devel doesn't build.

:skull:

Zialus:
All that is required to compile 6.2.2 successfully by any modern version of gcc is to add this [patch](https://aur.archlinux.org/cgit/aur.git/plain/friend-function.patch?h=yap).

There's also a new [_devel_](https://github.com/vscosta/yap-6.3/archive/YAP-6.3.4.tar.gz) version out, which I can compile successfully outside of homebrew but haven't been able to _formularize_ it.

I'll try to submit a pull request soon.

Zialus:
As promised, here it is https://github.com/Homebrew/homebrew-boneyard/pull/296 .

PR #4894, opened 2016-09-16
haskell-stack: use sdist

ilovezfs:
CC @borsboom

### DIFF
--- a/Formula/haskell-stack.rb
+++ b/Formula/haskell-stack.rb
@@ -5,8 +5,11 @@ class HaskellStack < Formula
 
   desc "The Haskell Tool Stack"
   homepage "https://haskellstack.org/"
-  url "https://github.com/commercialhaskell/stack/archive/v1.2.0.tar.gz"
-  sha256 "48f866168e81956807fac00f15dc0c9f38a390bba4de5ba53db220e3f25db983"
+  url "https://github.com/commercialhaskell/stack/releases/download/v1.2.0/stack-1.2.0-sdist-0.tar.gz"
+  version "1.2.0"
+  sha256 "872d29a37fe9d834c023911a4f59b3bee11e1f87b3cf741a0db89dd7f6e4ed64"
+  revision 1
+
   head "https://github.com/commercialhaskell/stack.git"
 
   bottle do


zmwangx:
`audit` doesn't allow revision, WTF:

```
$ brew audit haskell-stack
haskell-stack:
  * 'revision 1' should be removed
Error: 1 problem in 1 formula
```

dunn:
Probably it's scanning `stack-1.2.0-sdist-0.tar.gz` as a different version.

zmwangx:
Ah you're right. `haskell-stack-0_1.mavericks.bottle.tar.gz`.

ilovezfs:
ug thx

ilovezfs:
@BrewTestBot test this please

zmwangx:
Merged. Hopefully not many people need to immediate upgrade.

JCount:
Interestingly, and I think perhaps related to the fact that a explicit version tag had to be be added, since I have `haskell-stack-1.2.0.tar.gz` in my homebrew cache, trying to directly upgrade from source causes a `sha256` mismatch. The `sdist` version has a different hash but the updated source is not downloaded because the manually added version tag, to allow for the `revision`, matches.
```
haskell-stack 1.2.0_1
==> Upgrading haskell-stack
==> Using the sandbox
==> Downloading https://github.com/commercialhaskell/stack/releases/download/v1.2.0/stack-1.2.0-sdist-0.tar.gz
Already downloaded: /Users/James/Library/Caches/Homebrew/haskell-stack-1.2.0.tar.gz
Error: SHA256 mismatch
Expected: 872d29a37fe9d834c023911a4f59b3bee11e1f87b3cf741a0db89dd7f6e4ed64
Actual: 48f866168e81956807fac00f15dc0c9f38a390bba4de5ba53db220e3f25db983
Archive: /Users/James/Library/Caches/Homebrew/haskell-stack-1.2.0.tar.gz
To retry an incomplete download, remove the file above.
```
The only clear remedy is to remove the cached source, as it says, so the new source is downloaded.
Essentially, in this edge case, it behaved for me as if it were an issue with upstream retagging a release.

PR #4896, opened 2016-09-16
mpage 2.5.6 (new formula)

tchajed:

### DIFF
--- a/Formula/mpage.rb
+++ b/Formula/mpage.rb
@@ -1,0 +1,20 @@
+class Mpage < Formula
+  desc "Many to one page printing utility"
+  homepage "http://www.mesa.nl/pub/mpage/README"
+  url "http://www.mesa.nl/pub/mpage/mpage-2.5.6.tgz"
+  sha256 "4fe66dfc27f7c4bfbca60ef617f968aa2e6ee877e8921aa968c16a03aa4edc04"
+
+  def install
+    args = %W[
+      MANDIR=#{man1}
+      PREFIX=#{prefix}
+    ]
+    system "make", *args
+    system "make", "install", *args
+  end
+
+  test do
+    (testpath/"input.txt").write("Input text")
+    system bin/"mpage", "input.txt"
+  end
+end


Formula/mpage.rb
```diff
@@ -0,0 +1,21 @@
+class Mpage < Formula
+  desc "Mpage, many to one page printing utility"
+  homepage "http://www.mesa.nl/pub/mpage/"
```
zmwangx: We need a homepage, and a directory listing is hardly a homepage unfortunately.

Formula/mpage.rb
```diff
@@ -0,0 +1,21 @@
...
+  url "http://www.mesa.nl/pub/mpage/mpage-$version.tgz"
+  version "2.5.6"
```
zmwangx:
```
url "http://www.mesa.nl/pub/mpage/mpage-2.5.6.tgz"
```

No `version`. (This is Ruby by the way, not shell.)

Formula/mpage.rb
```diff
@@ -0,0 +1,21 @@
...
+      MANDIR=#{man1}
+      PREFIX=#{prefix}
+    ]
+    system "make", *args
+    system "make", "install", *args
+  end
+
+  test do
+    (testpath/"input.txt").write("Input text")
+    system bin/"mpage", "input.txt"
```
zmwangx: You sure this won't trigger a printer job?

Formula/mpage.rb
```diff
@@ -0,0 +1,21 @@
+class Mpage < Formula
+  desc "Mpage, many to one page printing utility"
+  homepage "http://www.mesa.nl/pub/mpage/"
+  url "http://www.mesa.nl/pub/mpage/mpage-$version.tgz"
+  version "2.5.6"
+  sha256 "4fe66dfc27f7c4bfbca60ef617f968aa2e6ee877e8921aa968c16a03aa4edc04"
+
+  def install
+    args = %W[
```
zmwangx: The Makefile is rather hilarious:

```Makefile
BINDIR=$(PREFIX)/bin
LIBDIR=$(PREFIX)/share
MANDIR=$(PREFIX)/man/man1
```

Just a comment. You should probably ask upstream to fix it.

Formula/mpage.rb
```diff
@@ -0,0 +1,21 @@
+class Mpage < Formula
+  desc "Mpage, many to one page printing utility"
```
zmwangx:
```ruby
ruby "Many to one page printing utility"
```

zmwangx (changes requested):
Please review out [acceptable formulae criterion](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Acceptable-Formulae.md); we won't package software without a homepage. I do realize this software is packaged by Debian, though.

Also, please do not just ignore and delete our PR template. You haven't even tried the formula yourself prior to submission, which is a waste of time for everyone.

Formula/mpage.rb
zmwangx: I wonder if http://www.mesa.nl/pub/mpage/README can serve as a minimal homepage.

Formula/mpage.rb
tchajed: Yes, it outputs postscript to stdout.

Formula/mpage.rb
tchajed: This is definitely a dead project (the last release is from 2008). Arguably this disqualifies it from Homebrew core.

Formula/mpage.rb
tchajed: I misinterpreted what `brew audit` wanted me to do, and I'm pretty sure I installed the formula with this syntax. In any case I've put the version in the URL and removed the call to url.

tchajed:
I actually didn't realize there was a PR template, since I created the pull request from hub, which [doesn't (yet) support GitHub pull request templates](https://github.com/Homebrew/homebrew-core/pull/4896). Rest assured the formula was tested and working locally.

Formula/mpage.rb
tchajed: done, thanks for catching

Formula/mpage.rb
tchajed: done

Formula/mpage.rb
zmwangx: > I'm pretty sure I installed the formula with this syntax.

You must have had a good URL to begin with, downloaded with that URL, then switched to the bad one. Later when you run `brew install` it just picks up the downloaded file from cache and no longer checked that URL. `brew audit --new-formula` can catch that, though.

zmwangx:
> I actually didn't realize there was a PR template, since I created the pull request from hub, which doesn't (yet) support GitHub pull request templates. Rest assured the formula was tested and working locally.

Okay, please remember to use the web interface for Homebrew-related matters in the future, and follow the template rigorously. This time you submitted a non-functioning formula which you could not detect due to caching which is excusable, but if you have followed the steps `brew audit --new-formula` would have caught that.

zmwangx (approved):

zmwangx:
The formula itself looks fine now, thanks for the fixes.

As for whether it qualifies for Homebrew, I'm personally reluctant to carry stuff that's abandoned a long time ago because the burden completely falls on our shoulders when stuff breaks, but in this case the code may be simple and ancient enough to not have much of a compatibility problem going forward.

Half a vote from me for letting it in. But if anything happens to it in the future it goes straight to boneyard, no questions asked.

Other maintainers: thoughts?

tchajed:
I will make sure I use the template in the future.

However, I did indeed run both `brew audit --new-formula mpage` and `brew audit --strict --online mpage` - even without the cached file, neither detects an invalid URL. Perhaps this is a useful new feature for `brew audit`? It's certainly surprising to me that this isn't checked, given that the homepage URL is checked.

Thanks for taking the time to go through this formula!

zmwangx:
> even without the cached file, neither detects an invalid URL.

Sorry, I thought it checks `url` too, but apparently it only checks `homepage`. Really need `brew fetch --force` to detect it.

zmwangx:
Related: https://github.com/Homebrew/brew/issues/935

zmwangx:
@BrewTestBot test this please

apjanke:
Eight years since the last change and it's building fine. Sounds like good old battle-tested Unix software to me, and a low maintenance burden. I say let it in.

zmwangx:
After two days, 1.5 👍 , 0 👎 . I doubt we'll get 2 more 👎 any time soon, so in it goes.

apjanke:
:shipit:

zmwangx:
Merged as 9b72166. Thank you for your contribution to Homebrew!

PR #4905, opened 2016-09-16
Update qt5 urls

ts826848:

---

This is probably _slightly_ premature...

At the very least I think the HEAD URL should be changed. Doesn't make sense for stable to be ahead of HEAD.


kanekotic:
even if repo is different the formulae fails to install with same problem as current branch on os x sierra

aw1621107:
Yeah... In any case, it seems that only the change to HEAD should be applied; stable shouldn't be changed until 5.7.1

DomT4:
Working on Qt5. It's being "fun". I'll take a fresh run at it later today, my CPU must think I hate it lately.

PR #4907, opened 2016-09-16
predictionio 0.9.6

gangodu:
- [Y] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

### DIFF
--- a/Formula/predictionio.rb
+++ b/Formula/predictionio.rb
@@ -1,8 +1,8 @@
 class Predictionio < Formula
   desc "Source machine learning server"
   homepage "https://prediction.io/"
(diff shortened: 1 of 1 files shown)


zmwangx:
I can confirm that `download.prediction.io` won't serve on port 443, so HTTP downgrade is necessary. http://predictionio.incubator.apache.org/install/install-linux/ also points to the HTTP download URL.

Note that even if HTTPS worked, they're just using SHA-1. I'm kinda tired working with these prematurely upgraded URLs. Maybe we should be more cautious with upgrading to undocumented URLs in the future.

(@gangodu This is a comment to our team, not you.)

zmwangx:
Merged as ee235a9. Thank you for your contribution to Homebrew!

PR #4920, opened 2016-09-17
erlang: fix the build on 10.11 with Xcode 8

ilovezfs:
Erlang uses some customized Autotools tests, so the general superenv
workaround in https://github.com/Homebrew/brew/pull/970 is insufficient

### DIFF
--- a/Formula/erlang.rb
+++ b/Formula/erlang.rb
@@ -4,6 +4,8 @@
 class Erlang < Formula
   desc "Programming language for highly scalable real-time systems"
   homepage "https://www.erlang.org/"
+  revision 1
+
   head "https://github.com/erlang/otp.git"
 
   stable do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

ilovezfs:
woo hoo :green_apple:

zmwangx (approved):
I approve the change 😎

zmwangx:
Actually, do we really need to revision? It's not like the runtime bug could have gone unnoticed.

ilovezfs:
@zmwangx I lean toward revision bumping given that the bottle has been rebuilt against a different SDK ;)

The only reason I asked about ffmpeg was because there are so many people who build that from source (probably needs more of the options enabled by default to fix that).

zmwangx:
👍

PR #4933, opened 2016-09-18
Remove manual `TERM` setting in test blocks

zmwangx:
Goes with https://github.com/Homebrew/brew/pull/982.

I haven't converted everything yet because I don't want to overload the CI in one job. I would do the rest in another PR or two.

Note that some tests do require `TERM` set to `xterm` to work, e.g. `ee`, `eg` and `git-test`, and that's fine — if `TERM` is not explicitly set to `xterm` then Homebrew/brew#982 makes them fail hard locally instead of only failing mysteriously on the CI.

Also, `asciiquarium` passes on CI but fails locally even when `TERM` is set to `xterm`, not sure why. But then, when I manually invoke `asciiquarium` (poured) it segfaults... I don't intend to investigate.

### DIFF
--- a/Formula/browser.rb
+++ b/Formula/browser.rb
@@ -14,7 +14,6 @@ class Browser < Formula
   end
 
   test do
-    ENV["TERM"] = "xterm"
     system "#{bin}/browser"
   end
 end
--- a/Formula/bvi.rb
+++ b/Formula/bvi.rb
@@ -17,7 +17,6 @@ class Bvi < Formula
   end
 
   test do
-    ENV["TERM"] = "xterm"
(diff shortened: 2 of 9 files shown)


apjanke:
`asciiquarium` is segfaulting for me whenever I run it on my 10.11 box, either poured or built from source. I'd say it's not a regression.

zmwangx:
Fit it or 💀

apjanke:
https://github.com/Homebrew/homebrew-core/issues/4936

apjanke:
Anyway, 👍 on this in conjunction with automatically setting `$TERM`.

PR #4938, opened 2016-09-18
openmesh version bump

autosquid:

It reports the following problem which I really do not known how to remove.

```
      /usr/local/Cellar/open-mesh/6.2/lib/python2.7/site-packages/openmesh.so
```

I checked the cmake file: it finds python by cmake ---- will that be troublesome?

Hope someone could fix the problem based on my P.R.

---
1. upgrade to version 6.2;
2. add python binding

### DIFF
--- a/Formula/open-mesh.rb
+++ b/Formula/open-mesh.rb
@@ -1,8 +1,8 @@
 class OpenMesh < Formula
   desc "Generic data structure to represent and manipulate polygonal meshes"
   homepage "https://openmesh.org/"
-  url "https://www.openmesh.org/media/Releases/5.1/OpenMesh-5.1.tar.gz"
-  sha256 "643262dec62d1c2527950286739613a5b8d450943c601ecc42a817738556e6f7"
+  url "https://www.openmesh.org/media/Releases/6.2/OpenMesh-6.2.tar.gz"
+  sha256 "570b5b2d3b949050d9628367268c495e87b3dc59f18a07c8037449356fe40374"
   head "http://openmesh.org/svnrepo/OpenMesh/trunk/", :using => :svn
 
   bottle do
@@ -15,6 +15,9 @@ class OpenMesh < Formula
   depends_on "cmake" => :build
   depends_on "qt" => :optional
 
+  option "without-python", "Build without python 2 support"
+  depends_on "boost-python"
+
   def install
     mkdir "build" do
       args = std_cmake_args
@@ -24,6 +27,14 @@ class OpenMesh < Formula
       else
         args << "-DBUILD_APPS=OFF"
       end
+
+      if build.without? "python"
+        args << "-DOPENMESH_BUILD_PYTHON_BINDINGS=OFF"
+      else
+        args << "-DOPENMESH_BUILD_PYTHON_BINDINGS=ON"
+      end
+
+      inreplace "#{buildpath}/src/Python/CMakeLists.txt", "${ACG_PROJECT_LIBDIR}/python", "${ACG_PROJECT_LIBDIR}/python2.7/site-packages"
 
       system "cmake", "..", *args
       system "make", "install"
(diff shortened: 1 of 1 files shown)


autosquid:
anyone can help?

MikeMcQuaid:
CC @tdsmith for help. Also, is there docs for what to do on this error? If so, let me know so I can be more helpful with this stuff in future.

Formula/open-mesh.rb
```diff
@@ -76,5 +87,6 @@ def install
     ]
     system ENV.cxx, "test.cpp", "-o", "test", *flags
     system "./test"
+    system "python -c 'import openmesh;'"
```
tdsmith: `system "python", "-c", "import openmesh"` please

Formula/open-mesh.rb
```diff
@@ -25,6 +28,14 @@ def install
         args << "-DBUILD_APPS=OFF"
       end
 
+      if build.without? "python"
+        args << "-DOPENMESH_BUILD_PYTHON_BINDINGS=OFF"
+      else
+        args << "-DOPENMESH_BUILD_PYTHON_BINDINGS=ON"
+      end
```
tdsmith: Why give an option to turn the bindings off? Let's leave them enabled.

tdsmith (changes requested):

tdsmith:
No, every solution is bespoke. :(

In general what needs to happen is:
- figure out where in the CMakelist something like PYTHON_LIBRARY is being added to LINK_LIBRARIES for the python binding target and make it stop
- add something like:

```cmake
if(APPLE)
  set_target_properties(whichever_target LINK_FLAGS "-undefined dynamic_lookup")
endif()
```

instead.

MikeMcQuaid:
@tdsmith Even that may be worth adding to some docs and linking from the audit warning? My CMake knowledge is enough that I may be able to help contributors beyond that.

Formula/open-mesh.rb
autosquid: maybe there are neat freaks who do not use the binding and want to keep their site-package dir clean.

tdsmith: Nonetheless, let's leave it always enabled; thanks!

tdsmith:
I sent the attached patch to upstream to fix the Python linking problem; it applies against master but it's diverged a little from the 3.6 branch.

https://gist.github.com/a806cabfdbf4ff9557850512061629d3

autosquid:
@tdsmith  I known little about week dynamic linking, so my question may be a little off the point: does the week dynamic linking of python.dylib means the openmesh library when imported will use whatever is in the surrounding python environment (instead of what is pointed out in openmesh.so if non-week dynamic linking is used) ?

tdsmith:
> does the week dynamic linking of python.dylib means the openmesh library when imported will use whatever is in the surrounding python environment

Yes.

> (instead of what is pointed out in openmesh.so if non-week dynamic linking is used) ?

If they don't match, the interpreter will segfault.

autosquid:
@tdsmith  thanks for your answering!


Also, what if the linked library is just some symbolic links ? For example:

A is non-weekly linked to /usr/local/lib/B.dylib which is a symbolic link to /usr/local/Cellar/B/lib/B.1.dylib. Will A be affected if I later upgrade B (and its symbolic) to /usr/local/Cellar/B/lib/B.2.dylib?

By the way, where can I learn things like these?

tdsmith:
It depends on the install_name of /usr/local/Cellar/B/lib/B.1.dylib; dylibs "remember" their name and the linker uses the remembered name stored in the dylib (called an install_name) instead of the current filesystem location of the dylib it finds.

If the install_name is /usr/local/Cellar/B/lib/B.1.dylib, A will break when you upgrade B. If the install_name is /usr/local/Cellar/B/lib/B.dylib, A will not break when you upgrade B (assuming B.1 and B.2 are compatible).

Dunno where to learn these things except by banging your head on them for long enough, sorry :)

MikeMcQuaid:
Passing on this due to no response. We'll review another PR with comments/conflicts/CI failures addressed. Sorry and thanks anyway!

PR #4941, opened 2016-09-18
Remove manual `TERM` setting in test blocks

zmwangx:
Follow up of #4933. The remaining formulae.

### DIFF
--- a/Formula/ne.rb
+++ b/Formula/ne.rb
@@ -18,9 +18,21 @@ class Ne < Formula
 
   test do
     ENV["TERM"] = "xterm"
-    (testpath/"test.txt").write("This is a test document.\n")
-    (testpath/"test_ne").write("GotoLine 2\nInsertString line 2\nExit\n")
-    system "script", "-q", "/dev/null", "#{bin}/ne", "--macro", ((testpath/"test_ne")).to_s, ((testpath/"test.txt")).to_s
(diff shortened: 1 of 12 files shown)


Formula/ne.rb
```diff
@@ -18,9 +18,21 @@ def install
 
   test do
     ENV["TERM"] = "xterm"
```
ilovezfs: if this isn't changing, maybe just drop it from the PR?

Formula/slrn.rb
```diff
@@ -34,6 +34,6 @@ def install
 
   test do
     ENV["TERM"] = "xterm"
```
ilovezfs: did you intend to remove this?

Formula/ne.rb
zmwangx: It's "related", so I'd rather just do it here than go through the trouble of opening and pulling another one...

Formula/newt.rb
```diff
@@ -54,7 +54,7 @@ def install
         newtFinished();
       }
     EOS
```
ilovezfs: no change to TERM here?

Formula/slrn.rb
zmwangx: No, it's one of the few cases where xterm is required.

Formula/newt.rb
zmwangx: Again, xterm required.

ilovezfs:
:+1:

PR #4942, opened 2016-09-18
Use Ruby 1.9+ symbol hash keys in all formulae.

MikeMcQuaid:

---

@DomT4 requested we do this en-masse after doing it in Homebrew/brew.

Skipping CI because it's lol too many.


ilovezfs:
I assume this will not ship until it's :green_apple: right?

MikeMcQuaid:
@ilovezfs It was cancelled because there was too many formulae. I trust Rubocop's `--auto-correct` and `brew readall` confirms these look good.

ilovezfs:
I was hoping for 977 new Sierra bottles. /s

UniqMartin:
I think stuff like `depends_on python3: :optional` (previously `depends_on :python3 => :optional`) reads horrible. It also breaks the symmetry with when it's not optional, where one still has to write `depends_on :python3`. The same problem arises with every special (symbol) dependency we have. I expect some confusion among contributors (particularly those not very familiar with Ruby) and the docs might need to be updated for this (if they haven't been yet).

Maybe there's are more such gems in this automated conversion, but I'm obviously not going to review 977 changed files for a change that has already been merged …

MikeMcQuaid:
@UniqMartin It may not read that nicely but I think the consistency trumps it, personally.

dunn:
> I think stuff like depends_on python3: :optional (previously depends_on :python3 => :optional) reads horrible.

Agree, but that's a problem with Ruby in general.

PR #4946, opened 2016-09-18
Fix `brew style` warnings.

MikeMcQuaid:

---

Fix those that Rubocop didn't do in https://github.com/Homebrew/homebrew-core/pull/4942.

### DIFF
--- a/Formula/par.rb
+++ b/Formula/par.rb
@@ -15,8 +15,8 @@ class Par < Formula
 
   conflicts_with "rancid", because: "both install `par` binaries"
 
-  # A patch by Jérôme Pouiller that adds support for multibyte
-  # charsets (like UTF-8), plus Debian packaging.
+  # Patch to add support for multibyte charsets (like UTF-8), plus Debian
+  # packaging.
   patch do
     url "http://sysmic.org/dl/par/par-1.52-i18n.4.patch"
     sha256 "2ab2d6039529aa3e7aff4920c1630003b8c97c722c8adc6d7762aa34e795861e"
--- a/Formula/gearman.rb
+++ b/Formula/gearman.rb
@@ -34,7 +34,7 @@ class Gearman < Formula
   def install
     # https://bugs.launchpad.net/gearmand/+bug/1368926
     Dir["tests/**/*.cc", "libtest/main.cc"].each do |test_file|
-      next unless /std::unique_ptr/ === File.read(test_file)
+      next unless /std::unique_ptr/ =~ File.read(test_file)
       inreplace test_file, "std::unique_ptr", "std::auto_ptr"
     end
 
--- a/Formula/ldid.rb
+++ b/Formula/ldid.rb
@@ -17,7 +17,7 @@ class Ldid < Formula
   depends_on "openssl"
 
   def install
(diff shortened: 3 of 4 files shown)


MikeMcQuaid:
@UniqMartin @bfontaine Struggling to get the `.rubocop_todo.yml` working as expected. Any thoughts/help would be welcome.

MikeMcQuaid:
@UniqMartin @bfontaine Ignore previous comment, wrong PR!

Formula/par.rb
```diff
@@ -15,7 +15,7 @@ class Par < Formula
 
   conflicts_with "rancid", because: "both install `par` binaries"
 
-  # A patch by Jérôme Pouiller that adds support for multibyte
+  # A patch by Jerome Pouiller that adds support for multibyte
```
bfontaine: I think we should leave this as-is; M. Pouiller’s first name is *Jérôme*, not *Jerome*.

zmwangx: I agree with ASCII source files.

zmwangx:
Is that all?

MikeMcQuaid:
@zmwangx I believe so after the hash fixes previously.

Formula/par.rb
MikeMcQuaid: Yeh, ASCII unfortunately 😭

chdiza:
Just out of curiosity (I have no views about this PR), what's wrong with UTF8 source files?

MikeMcQuaid:
@chdiza Rubocop complains because I guess some editors/ways of opening Ruby files do too.

chdiza:
> Rubocop complains because I guess some editors/ways of opening Ruby files do too.

Doesn't that just mean we should ignore that Rubocop complaint?  Why are broken editors Homebrew's problem?

MikeMcQuaid:
@chdiza I don't feel strongly either way but I'm fixing the existing "rules".

zmwangx:
> Just out of curiosity (I have no views about this PR), what's wrong with UTF8 source files?

7-bit ASCII code points are universally understood in most encodings. That's not true for UTF-8. I believe you need the magic comment to be safe, and even then, it may not be respected. Too much trouble just to include a non-ASCII name *in a comment*.

Formula/par.rb
UniqMartin: Very tragic, given the contents of the comment and the patch provided by this fine man.

I think we could bend the rules here and be more welcoming to people that use those “funny characters” in their names or their mother tongues. And I'm not aware of any user reports where this has caused any issues, so I don't see any harm in it.

UniqMartin: If ASCII-only is truly unavoidable, I think we should abbreviate the first name as “J.” instead of mutilating it by stripping those accents.

zmwangx: Oh just forget about the name. We carry many patches from many people, and we *don't* include names.

MikeMcQuaid: 👍 to just removing the name. This is definitely bikeshedding, for what it's worth so I'm going to just pick one then merge.

zmwangx: I amended this commit to remove the name. No more bikeshedding.

PR #1509, opened 2016-05-28
lftp 4.7.2

nijikon:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

PR #4956, opened 2016-09-18
lftp: migrate to boneyard

ilovezfs:


ilovezfs:
4.7.3 still doesn't build unless we apply @nijikon's patch, which does work.

zmwangx:
Quite a few installs actually, 3754 last month 7166 total.

ilovezfs:
if upstream can be convinced to absorb the patch, :skull: can be avoided

ilovezfs:
https://github.com/Homebrew/homebrew-core/pull/1509

PR #4960, opened 2016-09-18
Added Sierra bottle sha256 for rust

kevinastone:

---

### DIFF
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -19,6 +19,7 @@ class Rust < Formula
   end
 
   bottle do
+    sha256 "bd541d04581ba145baf54bbc98e014c968c9f447dc3380048cac9317ba95ed8b" => :sierra
     sha256 "e78b655e38815c01f0a366807d2ad6f57ded26c84d9b574b91d77072118b55c1" => :el_capitan
     sha256 "59516069f1bb877240fcd01edcd21d4b5c61636f5be3c10ccebc017d7f923420" => :yosemite
     sha256 "3667beb9555b8d6e93427c98a4b7bf3a461265e6c83e983da2607b07743a1a61" => :mavericks


MikeMcQuaid:
This won't work unfortunately; you need to wait for us to build and upload one and @zmwangx can provide more details on why this wasn't done already.

kevinastone:
@MikeMcQuaid ah, ok.  Was trying to figure out how to get a sierra bottle (it takes hours to build from source).

zmwangx:
See https://github.com/Homebrew/homebrew-core/issues/4841. I think upstream has a patch or something (probably need to dig up some discussions), but we haven't got to that yet because apparently there's no avid Rust programmer on our team. @ilovezfs knows more than I do on this issue, he actually did the research.

zmwangx:
> it takes hours to build from source.

Now it won't build from source. Which is why we don't have a bottle.

kevinastone:
Ah, maybe that's what up.  My source build just ran for 4-5 hours and I finally bailed.

MikeMcQuaid:
@zmwangx Maybe worth a Maximum OS X requirement?

zmwangx:
> My source build just ran for 4-5 hours

According to my experience it should get stuck early.

zmwangx:
Always good to set `--verbose`.

zmwangx:
> @zmwangx Maybe worth a Maximum OS X requirement?

If it saves people 4-5 hours, definitely, will do right away.

zmwangx:
See #4962 which temporarily disables it on Sierra.

kevinastone:
Just tested again.  It gets caught in a loop after that malloc error after about ~45 minutes.

zmwangx:
Yeah that's exactly the "expected" behavior.

alexcrichton:
Hello! I work over on the rust-lang/rust project, and I think I know what's going on here.

This bug was [first reported](https://github.com/rust-lang/rust/issues/34674) on July 5, 2016 and was [subsequently fixed](https://github.com/rust-lang/rust/issues/34719) on July 9, 2016. The fix was originally scheduled to go out in the 1.12 compiler but we [backported the fix](https://github.com/rust-lang/rust/pull/34927) to the 1.11 compiler. This means that versions of Rust 1.10 and below are all broken on OSX 10.12.

When building from source, however, we build from the *previous* stable version. That is, when building the 1.11 compiler, it uses the 1.10 compiler to kick off the compilation. I believe the problem you're seeing here is that when compiling the 1.11 compiler on OSX 10.12 you need to run the 1.10 compiler, which exhibits this bug.

We currently, and unfortunately, have no way to update the boostrap compiler for the 1.11 release. You *could* work around this by rebuilding 1.10 with an updated jemalloc and then using that to bootstrap 1.11. That being said, the 1.12 release is being released tomorrow (Sep 29, 2016). This compiler will bootstrap from the 1.11 release, which does indeed work on OSX 10.12, so the bootstrap should be able to proceed as usual.

Basically tl;dr; this'll fix itself in 24 hours!

alexcrichton:
Rust 1.12 was [just released](https://blog.rust-lang.org/2016/09/29/Rust-1.12.html)!

DomT4:
Thanks @alexcrichton. Appreciate the update yesterday & today ❤️. If you have a moment, do you have an opinion on the `cargo` issue raised in https://github.com/Homebrew/homebrew-core/issues/5185 & https://github.com/Homebrew/homebrew-core/pull/5392?

alexcrichton:
Certainly! Commented on them

PR #4967, opened 2016-09-19
direnv: add caveats

dashmug:

---

Updated formula to include the additional setup instructions required to integrate `direnv` to user's shell.

Reference: http://direnv.net/

### DIFF
--- a/Formula/direnv.rb
+++ b/Formula/direnv.rb
@@ -20,6 +20,30 @@ class Direnv < Formula
     system "make", "install", "DESTDIR=#{prefix}"
   end
 
+  def caveats; <<-EOS.undent
+    For direnv to work properly it needs to be hooked into the shell.
+    Each shell has its own extension mechanism:
+
+    Bash
+      Add the following line at the end of the "~/.bashrc" file:
+        eval "$(direnv hook bash)"
+      Make sure it appears even after rvm, git-prompt and other shell extensions
+      that manipulate the prompt.
+
+    Zsh
+      Add the following line at the end of the "~/.zshrc" file:
+        eval "$(direnv hook zsh)"
+
+    Fish
+      Add the following line at the end of the "~/.config/fish/config.fish" file:
+        eval (direnv hook fish)
+
+    Tcsh
+      Add the following line at the end of the "~/.cshrc" file:
+        eval `direnv hook tcsh`
+    EOS
+  end
+
   test do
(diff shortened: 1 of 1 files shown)


zmwangx:
Sorry, `caveats` is for brew-specific information rather than an instruction manual. It should *not* repeat the project's README or setup instructions.

dashmug:
I apologize.

I was only following what is already being used by the following existing formulas:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/rbenv.rb#L46
https://github.com/Homebrew/homebrew-core/blob/master/Formula/nodenv.rb#L22
https://github.com/Homebrew/homebrew-core/blob/master/Formula/pyenv.rb#L29

Those lines on those projects are not brew-specific instructions.

zmwangx:
The `#{var}/pyenv`, `#{var}/rbenv` and `#{var}/nodenv` parts are, so at least part of the caveats is justifiable. I reserve my opinion on the rest. In this case though, there's no brew-specific information at all, you were duplicating project README which users are expected to read.

Please do realize homebrew-core is a moving target, some of the code has been included for a long time (a few months could be "a long time" for style decisions) and no one revisited them since their inception, so not all code is exemplary. Also, team members don't necessarily agree on more murky issues like this.

dashmug:
Thanks for the clarification @zmwangx. I'll keep that in mind.

Keep up the good work.

zmwangx:
In case I sounded harsh, I want to assure you there's nothing personal here, and I'm afraid you'll have to bear with us before we have a fully automated bot to accept/reject changes and hence completely remove subjective judgements from cases like this 😉

dashmug:
No worries, man. I've previously thought that _caveats_ were for any additional post-install instructions, including non-homebrew stuff.

Can't we add this to the [contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document though?

MikeMcQuaid:
Not the contributing document but could in https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md

zmwangx:
That's not the right place. If you do want to add it you should find an appropriate place in https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md.

dashmug:
I see. Thanks.

PR #4968, opened 2016-09-19
pylucene 6.2.0 (resurrected formula)

coady:

---

### DIFF
--- a/Formula/pylucene.rb
+++ b/Formula/pylucene.rb
@@ -1,0 +1,36 @@
+class Pylucene < Formula
+  desc "Python extension for accessing Java Lucene"
+  homepage "https://lucene.apache.org/pylucene/index.html"
+  url "https://www.apache.org/dist/lucene/pylucene/pylucene-6.2.0-src.tar.gz"
+  sha256 "ae60cbb41a96ad22bcd03cb3af488abf76c6312466ec2ac3dc965ca08cd2ea71"
+
+  option "with-shared", "build jcc as a shared library"
+
+  depends_on ant: :build
+  depends_on java: "1.8+"
+  depends_on :python
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", lib/"python2.7/site-packages"
+    jcc = "JCC=python -m jcc --arch #{MacOS.preferred_arch}"
+    opt = "INSTALL_OPT=--prefix #{prefix}"
+    if build.with? "shared"
+      jcc << " --shared"
+      opoo "shared option requires python to be built with the same compiler: #{ENV.compiler}"
+    else
+      opt << " --use-distutils" # setuptools only required with shared
+      ENV["NO_SHARED"] = "1"
+    end
+
+    cd "jcc" do
+      system "python", "setup.py", "install", "--prefix=#{prefix}", "--single-version-externally-managed", "--record=install.txt"
+    end
+    ENV.deparallelize # the jars must be built serially
+    system "make", "all", "install", opt, jcc, "ANT=ant -Divy_install_path=#{buildpath} -Divy.default.ivy.user.dir=#{buildpath} -lib #{buildpath}", "PYTHON=python", "NUM_FILES=8"
+  end
+
+  test do
+    ENV.prepend_path "PYTHONPATH", HOMEBREW_PREFIX/"lib/python2.7/site-packages"
+    system "python", "-c", "import lucene; assert lucene.initVM()"
+  end
+end
--- a/tap_migrations.json
+++ b/tap_migrations.json
@@ -222,7 +222,6 @@
   "prooftree": "homebrew/gui",
   "pulse": "homebrew/boneyard",
   "pyenv-pip-rehash": "homebrew/boneyard",
-  "pylucene": "homebrew/boneyard",
   "pyqwt": "homebrew/boneyard",
   "pyxplot": "homebrew/x11",
   "qfits": "homebrew/boneyard",


MikeMcQuaid:
Install is failing: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/7930/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_pylucene/

coady:
> No Java runtime present, requesting install.
make: *** [compile] Error 97
...
Java: 1.8.0_101

Java is a dependency.
`depends_on java: "1.8+"`

Any suggestions?

MikeMcQuaid:
I'm not sure what's going on here as it's clearly using `javac` ok. CC @apjanke for Java thoughts.

MikeMcQuaid:
@apjanke Reping on this?

apjanke:
I can reproduce the build failure on my OS X 10.11.6 machine with Java 1.8.0_91 installed.

When I run it locally, I get this popup dialog in addition to the console error message.

screen shot 2016-11-11 at 2 51 10 pm

We've seen this happen with Java launcher programs or scripts which are looking specifically for Apple's old framework-based Java distribution, instead of the newer Oracle JDKs or using `/usr/libexec/java_home` to find the command line tools.

Excerpt from the console output from my local `brew install -s pylucene` attempt:

```
BUILD SUCCESSFUL
Total time: 6 seconds
ICU not installed
python -m jcc --arch x86_64 --jar lucene-java-6.2.0/lucene/build/core/lucene-core-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/analysis/common/lucene-analyzers-common-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/memory/lucene-memory-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/highlighter/lucene-highlighter-6.2.0.jar --jar build/jar/extensions.jar --jar lucene-java-6.2.0/lucene/build/queries/lucene-queries-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/queryparser/lucene-queryparser-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/sandbox/lucene-sandbox-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/analysis/stempel/lucene-analyzers-stempel-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/grouping/lucene-grouping-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/join/lucene-join-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/facet/lucene-facet-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/suggest/lucene-suggest-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/expressions/lucene-expressions-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/analysis/kuromoji/lucene-analyzers-kuromoji-6.2.0.jar --jar lucene-java-6.2.0/lucene/build/misc/lucene-misc-6.2.0.jar  --use_full_names --include lucene-java-6.2.0/lucene/expressions/lib/antlr4-runtime-4.5.1-1.jar --include lucene-java-6.2.0/lucene/expressions/lib/asm-5.1.jar --include lucene-java-6.2.0/lucene/expressions/lib/asm-commons-5.1.jar --package java.lang java.lang.System java.lang.Runtime --package java.util java.util.Arrays java.util.Collections java.util.HashMap java.util.HashSet java.util.TreeSet java.lang.IllegalStateException java.lang.IndexOutOfBoundsException java.util.NoSuchElementException java.text.SimpleDateFormat java.text.DecimalFormat java.text.Collator --package java.util.concurrent java.util.concurrent.Executors --package java.util.regex --package java.io java.io.StringReader --package java.nio.file java.nio.file.Path java.nio.file.Files java.nio.file.Paths --exclude org.apache.lucene.sandbox.queries.regex.JakartaRegexpCapabilities --exclude org.apache.regexp.RegexpTunnel --exclude org.apache.lucene.store.WindowsDirectory --exclude org.apache.lucene.store.NativePosixUtil --python lucene --mapping org.apache.lucene.document.Document 'get:(Ljava/lang/String;)Ljava/lang/String;' --mapping java.util.Properties 'getProperty:(Ljava/lang/String;)Ljava/lang/String;' --sequence java.util.AbstractList 'size:()I' 'get:(I)Ljava/lang/Object;' org.apache.lucene.index.IndexWriter:getReader --version 6.2.0 --module python/collections.py --module python/ICUNormalizer2Filter.py --module python/ICUFoldingFilter.py --module python/ICUTransformFilter.py  --files 8 --build
No Java runtime present, requesting install.
make: *** [compile] Error 97
```

That sounds like `python -m jcc` is using the old-style Apple JavaApplicationStub technique, which only works with the legacy Apple Java distribution.

The legacy Apple Java 6 is so old, that I think we should probably consider anything with a dependency on it to be unsupported. But Pylucene's web page says they require 1.7, and this formula is asking for 1.8. So this sounds like an issue in their build/packaging process.

apjanke:
Here's a guess. In the pylucene distribution, in `jcc/setup.py`, it has this:

```
elif platform in ("darwin",):
    try:
        from helpers.darwin import JAVAHOME, JAVAFRAMEWORKS
...
JDK = {
    'darwin': JAVAHOME or JAVAFRAMEWORKS,
    ...
 
LFLAGS = {
    'darwin/frameworks': ['-framework', 'JavaVM', '-mmacosx-version-min=10.5'],
    'darwin/home': ['-L%(darwin)s/jre/lib' %(JDK), '-ljava',
                    '-L%(darwin)s/jre/lib/server' %(JDK), '-ljvm',
                    '-Wl,-rpath', '-Wl,%(darwin)s/jre/lib' %(JDK),
                    '-Wl,-rpath', '-Wl,%(darwin)s/jre/lib/server' %(JDK), 
...

elif platform == 'darwin':
    if JAVAHOME is not None:
        INCLUDES['darwin'] = INCLUDES['darwin/home']
        LFLAGS['darwin'] = LFLAGS['darwin/home']
    elif JAVAFRAMEWORKS is not None:
        INCLUDES['darwin'] = INCLUDES['darwin/frameworks']
        LFLAGS['darwin'] = LFLAGS['darwin/frameworks']


```

That looks like it could pick up the framework-based legacy Apple Java 6 as a default if `helpers.darwin` does not report a JAVAHOME. I don't know enough Python to know how it does that.

There's a `JCC_JDK` environment variable that allows you to force the JDK location.
```
if 'JCC_JDK' in os.environ:
    JDK[platform] = os.environ['JCC_JDK']
```
However, setting it produced no change in behavior for me. I suspect this is a bug in the `setup.py` logic, in that `$JCC_JDK` sets `JDK[platform]`, but the later darwin-specific logic in the file looks directly at `JAVAHOME` and `JAVAFRAMEWORKS` when deciding which kind of JDK distribution to link to. 

This seems to me like it should be reported as an upstream bug. The inclusion of logic for the framework-based JDK linkage doesn't seem to make sense, as only Java 1.6 is available in that form, and pylucene now requires Java 1.7. It's unnecessary code, and the error messages it's producing here (when you take the pop-up dialog in to account) seem misleading. And the precedence of `JAVAHOME` over `$JCC_JDK` seems backwards.

Maybe someone who knows Python and its interaction with Java more could indicate whether `JAVAHOME` is being set up correctly here.

MikeMcQuaid:
> This seems to me like it should be reported as an upstream bug.

I agree. Passing on this PR as-is for now until it builds successfully.

PR #4976, opened 2016-09-19
gofabric8 0.4.72 (new formula)

msavy:

---

Submission of new formula for [gofabric8 CLI](http://github.com/fabric8io/gofabric8) of the [fabric8 project](https://github.com/fabric8io/fabric8).

### DIFF
--- a/Formula/gofabric8.rb
+++ b/Formula/gofabric8.rb
@@ -1,0 +1,36 @@
+class Gofabric8 < Formula
+  desc "CLI for fabric8 running on Kubernetes or OpenShift"
+  homepage "https://github.com/fabric8io/gofabric8/"
+  url "https://github.com/fabric8io/gofabric8/archive/v0.4.69.tar.gz"
+  sha256 "21a34018367d8aeb9446be23747ff918770d0c8e827e7d3e0b7a57542eb61274"
+
+  depends_on "go" => :build
+
+  GOVENDOR_PATH = "src/github.com/fabric8io/gofabric8".freeze
+
+  def install
+    ENV["GOPATH"] = buildpath
+    files = Dir["*"]
+    mkdir_p buildpath/GOVENDOR_PATH
+    mv files, buildpath/GOVENDOR_PATH
+    cd GOVENDOR_PATH
+    system "make", "install", "REV=homebrew"
+    cd buildpath
+    bin.install "bin/gofabric8"
+  end
+
+  def caveats; <<-EOS.undent
+    gofabric8 can use any Kubernetes or OpenShift installations
+    available on the path, or as specificed with `--server`.
+
+    If you don't have an installation, you can install Kubernetes:
+      brew cask install minikube
+    Or OpenShift:
+      brew cask install minishift
+    EOS
+  end
+
+  test do
+    system "#{bin}/gofabric8", "version"
+  end
+end
--- a/Formula/gofabric8.rb
+++ b/Formula/gofabric8.rb
@@ -1,22 +1,28 @@
 class Gofabric8 < Formula
   desc "CLI for fabric8 running on Kubernetes or OpenShift"
   homepage "https://github.com/fabric8io/gofabric8/"
-  url "https://github.com/fabric8io/gofabric8/archive/v0.4.71.tar.gz"
-  sha256 "4eacee14e73056933dab263918cc64e45c57a2c7c7575fd23dada2a6ff3c5fe1"
+  url "https://github.com/fabric8io/gofabric8/archive/v0.4.72.tar.gz"
+  sha256 "0c082072fc9731fc939b6d4b6719adf20e0e69ff780d613337e1832ab9e8a1db"
 
   depends_on "go" => :build
 
   def install
     ENV["GOPATH"] = buildpath
-    files = Dir["*"]
-    mkdir buildpath/"src/github.com/fabric8io/gofabric8" do
-      files.each { |f| mv buildpath/f, "." }
+    dir = buildpath/"src/github.com/fabric8io/gofabric8"
+    dir.install buildpath.children
+
+    cd dir do
       system "make", "install", "REV=homebrew"
     end
+
     bin.install "bin/gofabric8"
   end
 
   test do
-    system "#{bin}/gofabric8", "version"
+    Open3.popen3("#{bin}/gofabric8", "version") do |stdin, stdout, _|
+      stdin.puts "N" # Reject any auto-update prompts
+      stdin.close
+      assert_match(/gofabric8, version #{version} \(branch: 'unknown', revision: 'homebrew'\)/, stdout.read)
+    end
   end
 end
--- a/Formula/gofabric8.rb
+++ b/Formula/gofabric8.rb
@@ -1,19 +1,18 @@
 class Gofabric8 < Formula
   desc "CLI for fabric8 running on Kubernetes or OpenShift"
   homepage "https://github.com/fabric8io/gofabric8/"
-  url "https://github.com/fabric8io/gofabric8/archive/v0.4.69.tar.gz"
-  sha256 "21a34018367d8aeb9446be23747ff918770d0c8e827e7d3e0b7a57542eb61274"
+  url "https://github.com/fabric8io/gofabric8/archive/v0.4.71.tar.gz"
+  sha256 "4eacee14e73056933dab263918cc64e45c57a2c7c7575fd23dada2a6ff3c5fe1"
 
   depends_on "go" => :build
 
   def install
     ENV["GOPATH"] = buildpath
     files = Dir["*"]
-    mkdir_p buildpath/"src/github.com/fabric8io/gofabric8"
-    mv files, buildpath/"src/github.com/fabric8io/gofabric8"
-    cd "src/github.com/fabric8io/gofabric8"
-    system "make", "install", "REV=homebrew"
-    cd buildpath
+    mkdir buildpath/"src/github.com/fabric8io/gofabric8" do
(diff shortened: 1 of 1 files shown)


Formula/gofabric8.rb
```diff
@@ -0,0 +1,36 @@
...
+  def caveats; <<-EOS.undent
+    gofabric8 can use any Kubernetes or OpenShift installations
+    available on the path, or as specificed with `--server`.
+
+    If you don't have an installation, you can install Kubernetes:
+      brew cask install minikube
+    Or OpenShift:
+      brew cask install minishift
+    EOS
+  end
```
MikeMcQuaid: I think you can remove these; it's pretty self-explanatory.

Formula/gofabric8.rb
```diff
@@ -0,0 +1,36 @@
...
+  sha256 "21a34018367d8aeb9446be23747ff918770d0c8e827e7d3e0b7a57542eb61274"
+
+  depends_on "go" => :build
+
+  GOVENDOR_PATH = "src/github.com/fabric8io/gofabric8".freeze
+
+  def install
+    ENV["GOPATH"] = buildpath
+    files = Dir["*"]
+    mkdir_p buildpath/GOVENDOR_PATH
```
MikeMcQuaid: `mkdir do "src/github.com/fabric8io/gofabric8"` will create it and `cd` to it temporarily

Formula/gofabric8.rb
```diff
@@ -0,0 +1,36 @@
+class Gofabric8 < Formula
+  desc "CLI for fabric8 running on Kubernetes or OpenShift"
+  homepage "https://github.com/fabric8io/gofabric8/"
+  url "https://github.com/fabric8io/gofabric8/archive/v0.4.69.tar.gz"
+  sha256 "21a34018367d8aeb9446be23747ff918770d0c8e827e7d3e0b7a57542eb61274"
+
+  depends_on "go" => :build
+
+  GOVENDOR_PATH = "src/github.com/fabric8io/gofabric8".freeze
```
MikeMcQuaid: Don't make this a constant, just hard-code it throughout.

Formula/gofabric8.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def install
+    ENV["GOPATH"] = buildpath
```
MikeMcQuaid: I think this may be unnecessary? Worth trying without.

Formula/gofabric8.rb
```diff
@@ -0,0 +1,36 @@
...
+    files = Dir["*"]
+    mkdir_p buildpath/GOVENDOR_PATH
+    mv files, buildpath/GOVENDOR_PATH
```
MikeMcQuaid: May be worth trying a symlink instead of moving.

Formula/gofabric8.rb
```diff
@@ -0,0 +1,36 @@
...
+    cd GOVENDOR_PATH
+    system "make", "install", "REV=homebrew"
```
MikeMcQuaid: What's `REV` here? Also, need to specify the `prefix` somehow?

MikeMcQuaid (changes requested):

Formula/gofabric8.rb
msavy: This was my first approach - it doesn't work.

Formula/gofabric8.rb
msavy: If you look inside of the Makefile it's a custom variable which derives its information from git - and would be set to "unknown" when using a tarball. 

The information is shown when `gofabric8 --version` is invoked; it is handy for making it obvious that homebrew was the source of the installation.

Formula/gofabric8.rb
msavy: Okay

Formula/gofabric8.rb
MikeMcQuaid: No worries.

Formula/gofabric8.rb
MikeMcQuaid: Could instead use a Git checkout for this to get the right version.

Formula/gofabric8.rb
msavy: It is necessary - Go doesn't default to anything useful on this (and yes, I double checked!).

You'll see this is a common factor for most Go formula.

BTW this formula is using the Go Vendor Experiment, which may explain some of the peculiarities you see here.

Formula/gofabric8.rb
msavy: Aha, handy. Thank you!

msavy: @MikeMcQuaid mkdir_p doesn't seem to invoke an attached block, and mkdir naturally doesn't support the multiple levels of path.

Are you OK with me reverting to the previous more verbose method, or do you prefer to add support for using a block with mkdir_p to your DSL?

Formula/gofabric8.rb
msavy: I think these fields are mainly useful for those who have built off a specific commit hash or their own branch  - given this is an officially cut release, it should be OK to take this approach, unless you object.

msavy:
Thank you very much for your feedback Mike, much appreciated.

Having `mkdir_p` with a block could be handy, perhaps I should make a PR for that.

MikeMcQuaid:
> Having mkdir_p with a block could be handy, perhaps I should make a PR for that.

I think it's probably worth just making the existing `mkdir do` make the full path 👍

Formula/gofabric8.rb
MikeMcQuaid: Ok, cool.

Formula/gofabric8.rb
```diff
@@ -0,0 +1,22 @@
...
+  url "https://github.com/fabric8io/gofabric8/archive/v0.4.71.tar.gz"
+  sha256 "4eacee14e73056933dab263918cc64e45c57a2c7c7575fd23dada2a6ff3c5fe1"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    files = Dir["*"]
+    mkdir buildpath/"src/github.com/fabric8io/gofabric8" do
+      files.each { |f| mv buildpath/f, "." }
```
DomT4: Probably looks nicer to do:
```ruby
dir = buildpath/"src/github.com/fabric8io/gofabric8"
dir.install buildpath.children

cd dir do
  system "make", "install", "REV=homebrew"
end

bin.install "bin/gofabric8"
```

Formula/gofabric8.rb
```diff
@@ -0,0 +1,22 @@
...
+      system "make", "install", "REV=homebrew"
+    end
+    bin.install "bin/gofabric8"
+  end
+
+  test do
+    system "#{bin}/gofabric8", "version"
```
DomT4: Any chance of a stronger test? At least assert on the version otherwise so we know the output isn't unexpected failures instead, etc.

DomT4 (changes requested):

Formula/gofabric8.rb
msavy: I changed it to the current approach based upon previous feedback by Mike.

I'm perfectly happy to take the approach you suggest, as long as this seems like a solution we can stick with. 

It does seem substantially more compact, so I have no complaints and will change it unless someone suggests a motivation for not doing so (e.g. clarity or otherwise).

MikeMcQuaid: We can stick with this solution, @DomT4's just better at this than me 😁

Formula/gofabric8.rb
msavy: This is an interesting topic which I did give some thought to, before going with this (weak) approach.

> Any chance of a stronger test? 

Just to give some context: this is a CLI app which is entirely dependent upon the presence of other (cask, manually or remotely installed) apps to function, so it's difficult to test in a useful way without a completely controlled environment - including sudo access (think full-on integration testing).  

That said, perhaps this is a possibility to consider? I may not be aware of some brew functionality/assumptions to achieve this!

We'd need to ensure that no existing K8/OpenShift environment was running; install and run K8/OS (e.g. via Cask; with some `sudo` work required for xhyve drivers); ensure that the network config is satisfactory - then we could try to connect. Quite a few variables that could make the process squishier than you might prefer.   

> At least assert on the version otherwise so we know the output isn't unexpected failures instead, etc.

The reason I didn't assert any of the text in 'version' was that I was concerned it might break if they decide to change it around. It's a fairly non-standard layout. I'm perfectly happy to do add it, with that caveat mentioned.

BTW - I realise this is less than ideal. In a previous formula I submitted it was much easier to provide a safely self-contained test.

Formula/gofabric8.rb
msavy: Done 👍

MikeMcQuaid: Thanks!

Formula/gofabric8.rb
MikeMcQuaid: > BTW - I realise this is less than ideal. In a previous formula I submitted it was much easier to provide a safely self-contained test.

Thanks for the justification. Seems fine to me then 👍

MikeMcQuaid (approved):

msavy:
@DomT4 I've altered the test, let me know what you think.

@MikeMcQuaid BTW have you noticed that `brew audit --new-formula <formula>` has been broken at some point recently.

Thanks, all!

msavy:
Added PR to fix aforementioned issue - https://github.com/Homebrew/brew/pull/1137

MikeMcQuaid:

PR #4984, opened 2016-09-19
tarantool 1.7.1-401

ilovezfs:
update head URL to point to 1.8 branch
fix build on 10.11 with Xcode 8

### DIFF
--- a/Formula/tarantool.rb
+++ b/Formula/tarantool.rb
@@ -1,10 +1,10 @@
 class Tarantool < Formula
   desc "In-memory database and Lua application server."
   homepage "https://tarantool.org/"
-  url "https://tarantool.org/dist/1.6/tarantool-1.6.8.653.tar.gz"
-  version "1.6.8-653"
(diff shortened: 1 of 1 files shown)


bigbes:
I suppose it's better to fix this bug in the Tarantool's upstream and then, simply, update version here.

ilovezfs:
@bigbes if we wait on every upstream to fix Xcode 8 + El Capitan, nothing around here is going to work.

bigbes:
@ilovezfs Also, the reason I've not pushed 1.7.1 version to hombrew, is because it's in Alpha (vinyl storage engine is)

bigbes:
And 1.8 branch is not developed, yet :)

ilovezfs:
@bigbes I did not realize you were upstream for this. My apologies.

I was going based on https://tarantool.org/download.html which points to 1.7 as the default.

The patch in this PR is valid for getting this building successfully on 10.11 with Xcode 8, so my suggestion is that we bump the formula back down to 1.6.x but keep the patch until you have an upstream fix.

ilovezfs:
@bigbes https://github.com/Homebrew/homebrew-core/pull/5027.

PR #4991, opened 2016-09-19
dmd 2.071.2

ilovezfs:


John-Colvin:
I think the test failures are problems with the autotester, could someone try restarting them (or @ilovezfs just `git commit --amend && git push --force` to redo the commit and trigger the tester)

ilovezfs:
correct test-bot was broken at that time. happy to bump this one up in my queue

PR #5013, opened 2016-09-20
algernon 1

ilovezfs:
bump version_scheme
use vendor directory
clean up the test

### DIFF
--- a/Formula/algernon.rb
+++ b/Formula/algernon.rb
@@ -1,10 +1,9 @@
-require "language/go"
-
 class Algernon < Formula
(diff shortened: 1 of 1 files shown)


mistydemeo:
Did this really get updated from 1.0 to 1?

ilovezfs:
1.0 -> v1

we parse v1 as 1
we could override 1 to be v1

but both 1 < 1.0 and v1 < 1.0

I'd prefer not to be overriding our v removal every time, but even if we do choose to do that, it still needs the version_scheme bump.

PR #5016, opened 2016-09-20
gnu-smalltalk: fix the build on 10.11 with Xcode 8

ilovezfs:


ilovezfs:
@busla you can brew update now

PR #5022, opened 2016-09-20
Updated conflicts info with maven3x

pkuczynski:


zmwangx:
Thanks for the PR, but we do not record conflicts with external taps (including Homebrew org taps) in core.

pkuczynski:
I see. Understood...

PR #5027, opened 2016-09-20
tarantool 1.6.8-772

ilovezfs:
upstream says 1.7.x is not yet considered stable
point head spec at 1.7 branch since 1.8 is not developed yet

CC @bigbes


ilovezfs:
@bigbes shipped.

bigbes:
Thank you! Sorry, I've missed mentions.

PR #5028, opened 2016-09-20
pgtap: migrate to boneyard

ilovezfs:


ilovezfs:
conflicts with perl and has 39 events in analytics.

PR #5029, opened 2016-09-20
openconnect: make before make install

toonetown:

---

Due to the way the makefiles are set up, the `make` command has to be run prior to running `make install`

### DIFF
--- a/Formula/openconnect.rb
+++ b/Formula/openconnect.rb
@@ -49,6 +49,7 @@ class Openconnect < Formula
     ]
 
     system "./configure", *args
+    system "make"
     system "make", "install"
   end
 


zmwangx:
Please explain which components are not built when `make` isn't run first. It has been that way for years. I find it hard to believe any component is missing, when we've had thousands of install events in just a few months without a single complaint.

toonetown:
I believe that this is due to a change that has been made to the upstream code.  For example, I am able to run `brew install --build-from-source openconnect` and it will build and install version 7.07 just fine, however this is the error that I get when running `brew install --HEAD openconnect`:

```
2016-09-22 14:10:11 -0600

make
install

Making install in tests
make[1]: *** No rule to make target `../libopenconnect.la', needed by `serverhash'.  Stop.
  CC       lzstest.o
  CC       seqtest.o
  CC       serverhash.o
```

It looks like something might have changed in their makefiles where the tests are being built first, but they are depending on the library being built.

zmwangx:
That looks like a regression that should be reported to upstream first.

toonetown:
> That looks like a regression that should be reported to upstream first.

Done - I reported it to them.

dwmw2:
Fixed by http://git.infradead.org/users/dwmw2/openconnect.git/commitdiff/2c329c9b6445306d2135192ce42bf48698fc0a33

Thanks.

toonetown:
> Fixed by http://git.infradead.org/users/dwmw2/openconnect.git/commitdiff/2c329c9b6445306d2135192ce42bf48698fc0a33

Wow - super fast turnaround!  Closing this in favor of that fix.  (I have verified that it works).

PR #5041, opened 2016-09-20
bazel: fix build on macOS 10.12

ilovezfs:


ilovezfs:
Someone decided to take this PR and turn it into a functionally equivalent upstream PR, so I'll let them sort it out together.

PR #4539, opened 2016-09-05
ansible: migrate to openssl@1.1

tdsmith:

Issue #2767, opened 2016-07-06
brew upgrade should remember brew install options


claco:
While this issue is specifically about the vim formula, I suspect more of them have this same issue w/ dependencies.

Up until about 3 weeks ago, the vim formula would either install the vim bottle (and only the vim bottle), or compile from source. When using the Ansible `homebrew` or `package` module, this meant that vim used the system perl/ruby/python language versions (or whatever you had active in rbenv/plenv/pyenv.

Now, `brew install vim` AND `brew upgrade vim` now pull in the perl/ruby/python bottles as well, which aren't the system language versions or not active versions in pyenv/plenv/rbenv. This means any plugin perviously compiled against other versions of languages (think command-t in ruby as an example) now don't work at all, yielding a broken configuration. 

To combat this, one must `brew install --build-from-source vim` to use the active/system languages. Even worse, if you're upgrading vim, you now have to `brew unlink vim` and then `brew install --build-from-source` again, as `brew upgrade` will simply pull in the perl/ruby/python bottles again.

All gripes aside about vim pulling in these language dependencies, if seems like `brew upgrade x` should remember what `brew install x` had done, so it can repeat the process correctly.

Because it does not, and because of the dependency issues needing `unlink` and a `install`, one can't really use the Ansible `homebrew` module with `state=latest` for some formulas without jumping through hoops.

MikeMcQuaid:
It does remember installation options. In this case the issue is we've added a binary package for the first time and added new options you need to use: https://github.com/Homebrew/homebrew-core/blob/e40dc2e709ff5409c672e31c8f89ad31a3e9712e/Formula/vim.rb#L26-L28. We can't preserve the implied `--build-from-source` as that would mean we could never add binary packages and have them used.

> Because it does not, and because of the dependency issues needing unlink and a install, one can't really use the Ansible homebrew module with state=latest for some formulas without jumping through hoops.

We don't support this module, just FYI.

claco:
'We can't preserve the implied --build-from-source as that would mean we could never add binary packages and have them used.'

Is that because that option propagates down to dependencies?

MikeMcQuaid:
More just that, in this case, it'll imply `--build-from-source` for a package where no binary package was available so it would assume no-one wanted the binary package when it was first introduced.

claco:
[...]

PR #5042, opened 2016-09-20
ansible: fix openssl failures by using brewed python at all times

dch:

---

native Apple Python & OpenSSL combo does not handle gracefully situations
where SSLv3 and lower are disabled by the remote end. By using brewed
Python, we use the much newer brewed OpenSSL and we have support for
session negotation, without cryptic failures.

NB this is not related to https://github.com/Homebrew/homebrew-core/pull/4539 despite the surface similarity. One questions
the wisdom of having multiple versions of OpenSSL available to python.

### DIFF
--- a/Formula/ansible.rb
+++ b/Formula/ansible.rb
@@ -17,7 +17,7 @@ class Ansible < Formula
   end
 
   depends_on "pkg-config" => :build
-  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on :python
   depends_on "libyaml"
   depends_on "openssl"
 


tdsmith:
This is blocked on a final disposition of https://github.com/Homebrew/brew-evolution/pull/12.

halostatue:
@tdsmith Homebrew/brew-evolution#12 has been resolved as adopted. Should this move forward?

claco:
Sorry for the drive-by question on this. I found it while checking out the 2.2.0.0 Ansible PR.

Vim also went with brewed python/perl/ruby, which on the surface is nice, but on upgrade it broke existing installs that had extensions compiled against system perl/python/ruby.  Will this not also happen with ansible depending on brew python? Things like crytpography and pyrax, etc. Will one have to uninstall ansible, then re-brew install it, then recompile plugins/extensions used by modules?

In the case of the vim formula. there are options to keep on using the system perl/python/ruby, which I like (--with-custom-python, etc). Having yet another python installed via homebrew, and a system python, AND pyenv managed pythons seems likes a not fun time if I can't tell homebrew to please not add yet another one to the mix.

dch:
now that https://github.com/Homebrew/brew-evolution/pull/12 has been merged 🎱, can we get a 2nd look at this please? It would be *great* to have this land before https://github.com/Homebrew/homebrew-core/pull/6485 is merged in so that the new bottled ansible version has working TLS in place.

dunn:
@claco the `:python` special dependency means that if you build from source it will use the Python in your $PATH as long as it's new enough.

MikeMcQuaid:
Will let @tdsmith 👍 and merge this when happy.

tdsmith:
Fixed in #6485; thanks!

tdsmith:
Sorry I didn't follow up sooner, @claco:

> Vim also went with brewed python/perl/ruby, which on the surface is nice, but on upgrade it broke existing installs that had extensions compiled against system perl/python/ruby. Will this not also happen with ansible depending on brew python? Things like crytpography and pyrax, etc. Will one have to uninstall ansible, then re-brew install it, then recompile plugins/extensions used by modules?

This is better encapsulated than vim; right now we don't support using ansible with any modules that aren't installed by our ansible package. We may invent a mechanism for that in the future if we have demand (I think I see how to do it) but nobody's asked yet.

> In the case of the vim formula. there are options to keep on using the system perl/python/ruby, which I like (--with-custom-python, etc). Having yet another python installed via homebrew, and a system python, AND pyenv managed pythons seems likes a not fun time if I can't tell homebrew to please not add yet another one to the mix.

This is unfortunately hard to avoid. If you're using pyenv, it should be in PATH before Homebrew's prefix. In general, though, extension modules built for one python x.y _should_ work with all other python x.y installations, and it is a bug with that module if that isn't true. (Embedding a python interpreter in another piece of software, like vim does, is a separate case.)

claco:
@dunn 

```
the :python special dependency means that if you build from source it will use the Python in your $PATH as long as it's new enough.
```

Thanks. I guess then my minor gripe is that different formulas doing different things (--build-from-source and brew upgrade-isms) vs (--with-custom-python and brew upgrade-isms).

It could be late, and I could be totally wrong, but I though the latter option meant brew upgrade remembered my last set of options and did the same thing, while the former didn't. Maybe This is addressed elswhere in other homebrew improvements.

claco:
@dunn 

And this is exactly what is happening, which is the same issue that vim had until they added --with-custom-python

I have ansible 2.2.0.0 installed via brew install --build-from-source ansible. When I upgrade, homebrew does not remember that option, and instead foes to install the python dep, which is undesired. IF there was a config option, --with-custom-python, it would remember that.

```
 → ansible --version
ansible 2.2.0.0
  config file =
  configured module search path = Default w/o overrides
chri5797@MQR13TG8WL ~
 → brew upgrade ansible
Updating Homebrew...
ansible 2.2.0.0_1
==> Upgrading ansible
==> Installing dependencies for ansible: sqlite, python, openssl@1.1
==> Installing ansible dependency: sqlite
==> Downloading https://homebrew.bintray.com/bottles/sqlite-3.15.1.el_capitan.bottle.tar.gz
^
```

MikeMcQuaid:
Repeating myself from https://github.com/Homebrew/homebrew-core/issues/2767:

> More just that, in this case, it'll imply `--build-from-source` for a package where no binary package was available so it would assume no-one wanted the binary package when it was first introduced.

PR #5049, opened 2016-09-21
ruff 1.5.0 (new formula)

dreamhead:
Ruff is a IoT platform with JavaScript as its application language.

Visit https://ruff.io to know more about Ruff

### DIFF
--- a/Formula/ruff.rb
+++ b/Formula/ruff.rb
@@ -1,0 +1,17 @@
+class Ruff < Formula
+  desc "IoT platform with JavaScript as application language"
+  homepage "https://ruff.io"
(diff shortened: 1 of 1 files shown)


zmwangx:
You're installing from a binary distribution here, which isn't the right thing for Homebrew (just look at `bin/ruff` and `bin/rap`). I'm not even sure this software is open source. If it's indeed closed source, we probably can't include it in core.

dreamhead:
@zmwangx You're right. Ruff is not open source for now. 

What's your suggestion for cloused source software?

zmwangx:
I would recommend you host this formula in your own tap. It's pretty simple.

PR #5076, opened 2016-09-21
juju2 2.0-rc1

sinzui:

---

I propose that juju2 is now stable enough to enter homebrew-core. I want to remove it from homebrew-devel-only

Juju 2.0-rc1 guarantees API and command line compatibility with future release of the clients and the cloud agents. Users will be able to deploy applications today, and upgrade juju with the next release. The Juju team will support upgrades to other RC versions if they are released, upgrades to Juju 2.0.0 and greater are supported.

### DIFF
--- a/Formula/juju2.rb
+++ b/Formula/juju2.rb
@@ -1,0 +1,21 @@
+class Juju2 < Formula
+  desc "DevOps management tool"
+  homepage "https://jujucharms.com/"
+  url "https://launchpad.net/juju/2.0/2.0-rc1/+download/juju-core_2.0-rc1.tar.gz"
+  sha256 "bef47f4d4e929a71c0934a782043fecc409a76ca4e73fbe3406415226077f745"
+
+  depends_on "go" => :build
+  conflicts_with "juju", because: "juju and juju2 cannot be installed simultaneously, and juju environments aren't compatible with juju2"
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "go", "build", "github.com/juju/juju/cmd/juju"
+    system "go", "build", "github.com/juju/juju/cmd/plugins/juju-metadata"
+    bin.install "juju", "juju-metadata"
+    bash_completion.install "src/github.com/juju/juju/etc/bash_completion.d/juju-2.0"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
We'll consider including it when it's 2.0.0. Will juju 1 see any more releases? It's pretty likely this should replace that.

sinzui:
Juju 1.25.x has 2 more years of support. There will be releases to fix critical bugs, or to enable 1.x environments to upgrade to 2.x. That said, Juju 1 will not be getting enablement fixes to keep pace with cloud changes. Its features are already degrading because it cannot use new cloud regions or features announced within the last 6 months.

zmwangx:
The question is not development/maintenance cycle, but rather, is there any major reason to stick to juju1 instead of juju2? (For instance, Python 3.4 will continue to receive security fixes until 2019, but that doesn't mean we have python34, python35, and soon python36 in core.) From your comment above it seems that the answer is no, so most likely this should be an in-place upgrade of `juju.rb` rather than a new formula.

sinzui:
There is a major reason to *not* rename this formula to juju.rb. juju2 cannot manage a juju 1 state-server. Juju. Users who get an unintended upgrade to to juju2 will discover limited access to their running environments. They cannot redeploy identical environments if they need to build a replacement.

The juju developers plan to release a juju2 client in the future that will upgrade juju 1.x environments to juju 2.x. The juju formulas can be merged when upgrades between major versions are supported.

MikeMcQuaid:
This formula should become `juju.rb` but we have a better transition process for this now if you'd like to follow it.

0. the current `juju.rb` gets renamed to `juju@1.25.rb` (see https://github.com/Homebrew/brew/blob/master/docs/Rename-A-Formula.md)
0. in the same pull request: an alias is created from `juju@1.25.rb` to `juju.rb`
0. in this pull request you add this formula as `juju@2.0.rb` (when the stable version is released)
0. at some point in the future we change the alias so that it points to `juju@2.0.rb` (or a 2.x version)

MikeMcQuaid:
@sinzui In future feel free to just continue stuff like this in the same PR, thanks.

PR #5088, opened 2016-09-22
OpenCSG 1.4.2

kintel:

---

The build currently fails with this error:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/strip: symbols referenced by indirect symbol table entries that can't be stripped in: /usr/local/Cellar/opencsg/1.4.1/lib/libopencsg.1.4.1.dylib
```

Notes:
- The currently deployed version also fails with the same error, but the error is ignored. This is most probably due to the generated makefile (`src/Makefile`) prefixes the strip command with a '-' to ignore errors.
- I'm unable to reproduce this build error locally - my makefiles end up containing the '-' prefix and thus ignoring any strip errors. (@kintel)

Link to OpenCSG-1.4.0 built on the CI: https://bot.brew.sh/view/Testing/job/Homebrew%20Testing/1092/

-> the issue turned out to be triggered by building the OpenCSG examples.

### DIFF
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -1,8 +1,8 @@
 class Opencsg < Formula
   desc "The CSG rendering library"
   homepage "http://www.opencsg.org"
-  url "http://www.opencsg.org/OpenCSG-1.4.1.tar.gz"
-  sha256 "48182c8233e6f89cd6752679bde44ef6cc9eda4c06f4db845ec7de2cae2bb07a"
+  url "http://www.opencsg.org/OpenCSG-1.4.2.tar.gz"
+  sha256 "d952ec5d3a2e46a30019c210963fcddff66813efc9c29603b72f9553adff4afb"
 
   bottle do
     cellar :any
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -6,6 +6,9 @@ class Opencsg < Formula
 
   bottle do
     cellar :any
+    sha256 "577e6777db3c9ee1010577e9dd29f7d86cff106273ef650bf58b08b20020a751" => :el_capitan
+    sha256 "d6e5913457b310b32a3dd9673a248793fd53bc6d2863f55b3d3334be7665c544" => :yosemite
+    sha256 "26098d8c2d4e89f2a0389c470f8b094a805e97c959defa3381ab0cd8c8d8ec9e" => :mavericks
   end
 
   depends_on "qt5" => :build
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -18,6 +18,7 @@ class Opencsg < Formula
     system "qmake", "-r", "INSTALLDIR=#{prefix}",
       "INCLUDEPATH+=#{Formula["glew"].opt_include}",
       "LIBS+=-L#{Formula["glew"].opt_lib} -lGLEW"
+    system "cat", "src/Makefile"
     system "make", "install"
   end
 
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -20,7 +20,7 @@ class Opencsg < Formula
     sha256 "12cc799a6352eda4a18706eeefea059d14e23605a627dc12ed2a809f65328d69"
   end
 
-def install
+  def install
     system "qmake", "-r", "INSTALLDIR=#{prefix}",
       "INCLUDEPATH+=#{Formula["glew"].opt_include}",
       "LIBS+=-L#{Formula["glew"].opt_lib} -lGLEW"
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -24,7 +24,6 @@ class Opencsg < Formula
     system "qmake", "-r", "INSTALLDIR=#{prefix}",
       "INCLUDEPATH+=#{Formula["glew"].opt_include}",
       "LIBS+=-L#{Formula["glew"].opt_lib} -lGLEW"
-    system "cat", "src/Makefile"
     system "make", "install"
   end
 
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -16,7 +16,7 @@ class Opencsg < Formula
 
   # This patch disabling building examples
   patch do
-    url "https://raw.githubusercontent.com/openscad/formula-patches/d6fea2cf509a6feca48f498c167bce4970b16356/opencsg/patch-build.diff"
+    url "https://raw.githubusercontent.com/openscad/formula-patches/4e9b9aabcc89dff597f14ba64542e87d750bced6/opencsg/disable-examples.diff"
     sha256 "12cc799a6352eda4a18706eeefea059d14e23605a627dc12ed2a809f65328d69"
   end
 
--- a/Formula/opencsg.rb
+++ b/Formula/opencsg.rb
@@ -1,28 +1,12 @@
 class Opencsg < Formula
   desc "The CSG rendering library"
   homepage "http://www.opencsg.org"
-  url "http://www.opencsg.org/OpenCSG-1.4.0.tar.gz"
-  sha256 "ecb46be54cfb8a338d2a9b62dec90ec8da6c769078c076f58147d4a6ba1c878d"
+  url "http://www.opencsg.org/OpenCSG-1.4.1.tar.gz"
+  sha256 "48182c8233e6f89cd6752679bde44ef6cc9eda4c06f4db845ec7de2cae2bb07a"
   revision 1
-
-  bottle do
-    cellar :any
-    sha256 "577e6777db3c9ee1010577e9dd29f7d86cff106273ef650bf58b08b20020a751" => :el_capitan
-    sha256 "d6e5913457b310b32a3dd9673a248793fd53bc6d2863f55b3d3334be7665c544" => :yosemite
-    sha256 "26098d8c2d4e89f2a0389c470f8b094a805e97c959defa3381ab0cd8c8d8ec9e" => :mavericks
-  end
 
   depends_on "qt5" => :build
   depends_on "glew"
-
-  # This patch adds support for specifying INSTALLDIR
-  # It has been submitted upstream and accepted 20160709, through private email
-  # (as that's how submissions are done)
-  # Should be in the next release (> 1.4.0)
-  patch do
-    url "https://raw.githubusercontent.com/Homebrew/formula-patches/814a29d3ce4c6e7e919cd2fcd64bf45d421e821b/opencsg/patch-build.diff"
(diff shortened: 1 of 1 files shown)


kintel:
Question: Should I also submit a PR for removing the patch which is no longer needed? (https://github.com/Homebrew/formula-patches/blob/master/opencsg/patch-build.diff)

Formula/opencsg.rb
```diff
@@ -1,29 +1,13 @@
 class Opencsg < Formula
   desc "The CSG rendering library"
   homepage "http://www.opencsg.org"
-  url "http://www.opencsg.org/OpenCSG-1.4.0.tar.gz"
-  sha256 "ecb46be54cfb8a338d2a9b62dec90ec8da6c769078c076f58147d4a6ba1c878d"
+  url "http://www.opencsg.org/OpenCSG-1.4.1.tar.gz"
+  sha256 "48182c8233e6f89cd6752679bde44ef6cc9eda4c06f4db845ec7de2cae2bb07a"
   revision 1
```
zmwangx: Remove the revision.

Formula/opencsg.rb
```diff
@@ -1,29 +1,13 @@
...
 
-  bottle do
```
zmwangx: Keep the bottle block.

zmwangx (changes requested):

zmwangx:
> Should I also submit a PR for removing the patch which is no longer needed? 

That would be great.

zmwangx:
Build is failing by the way:

```
symbols referenced by indirect symbol table entries that can't be stripped in: /usr/local/Cellar/opencsg/1.4.1_1/lib/libopencsg.1.4.1.dylib
```

Formula/opencsg.rb
```diff
@@ -1,29 +1,16 @@
...
   homepage "http://www.opencsg.org"
-  url "http://www.opencsg.org/OpenCSG-1.4.0.tar.gz"
-  sha256 "ecb46be54cfb8a338d2a9b62dec90ec8da6c769078c076f58147d4a6ba1c878d"
-  revision 1
+  url "http://www.opencsg.org/OpenCSG-1.4.1.tar.gz"
+  sha256 "48182c8233e6f89cd6752679bde44ef6cc9eda4c06f4db845ec7de2cae2bb07a"
 
   bottle do
     cellar :any
-    sha256 "577e6777db3c9ee1010577e9dd29f7d86cff106273ef650bf58b08b20020a751" => :el_capitan
```
zmwangx: Keep the entire block, just don't touch it.

zmwangx (changes requested):

kintel:
@zmwangx Regarding your build error, what version of OS X and Xcode is this?

If this fails to strip, I would expect the current 1.4.0 version of OpenCSG to fail in exactly the same way.
I would also expect any qmake-built library to fail in a very similar way, unless homebrew has a strategy for dealing with this.

zmwangx:
It's not my machine. Look at the CI.

kintel:
The build failure looks like a regression in Qt-5.6.1: https://bugreports.qt.io/browse/QTBUG-54212

It should affect all qt5-qmake builds. To verify this, would it be possible to run the CI on the currently published version of the OpenCSG formula?

If this is indeed the case, is there a protocol for downgrading packages which trigger cascading issues like this, or do we just hang on to the previously built binaries and hope for the best?

zmwangx:
The current version works: https://bot.brew.sh/view/Testing/job/Homebrew%20Testing/1092/version=el_capitan/ and expectedly so because we test all dependents before bumping a formula (in this case qt5).

kintel:
@zmwangx That's odd. If you look at the succeeding test results, you'll see that the build fails in the exact same way, but the error is somehow being ignored.
Excerpts from the succeeding log:
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/strip: symbols referenced by indirect symbol table entries that can't be stripped in: /usr/local/Cellar/opencsg/1.4.0_1/lib/libopencsg.1.4.0.dylib
[...]
make[1]: [install_target] Error 1 (ignored)
```

..so my question is now: Who is ignoring this build error?

zmwangx:
Hmm that's weird but that's not on us because I we do is

```ruby
  def install
    system "qmake", "-r", "INSTALLDIR=#{prefix}",
      "INCLUDEPATH+=#{Formula["glew"].opt_include}",
      "LIBS+=-L#{Formula["glew"].opt_lib} -lGLEW"
    system "make", "install"
  end
```

Probably worth reporting to upstream to see what they say.

kintel:
I cannot reproduce the build error locally. Looks like the generated makefile differs slightly for no apparent reason. I'll look closer into it and see if I can get it reproduced.

zmwangx:
@kintel Before we figure out the problem, could you please hold pushing cosmetic changes? Each push queues up a job on our CI, and since qt5 is a dependency and isn't bottled on Sierra (because it won't build), a qt5 build is triggered and it blocks everything. Thanks.

kintel:
ok, I wasn't aware that qt5 isn't bottled on Sierra.
FYI: The indentation wasn't cosmetic - it's required for the build to run (I just forgot to run the audit locally).

kintel:
@zmwangx After my latest patch, the tests appear to run without errors. A few weird things though:

* On el capitan and yosemite, the tests succeed, but I don't get a verbose build log like you did in your manually triggered build (https://bot.brew.sh/view/Testing/job/Homebrew%20Testing/1092/)
* On sierra, Qt somehow fails to build before it gets to OpenCSG

ilovezfs:
>On sierra, Qt somehow fails to build before it gets to OpenCSG

correct Qt has no Sierra support yet except in pre-release branches

Formula/opencsg.rb
```diff
@@ -15,19 +14,17 @@ class Opencsg < Formula
   depends_on "qt5" => :build
   depends_on "glew"
 
-  # This patch adds support for specifying INSTALLDIR
-  # It has been submitted upstream and accepted 20160709, through private email
-  # (as that's how submissions are done)
-  # Should be in the next release (> 1.4.0)
+  # This patch disabling building examples
```
zmwangx: Need link to upstream bug report (are you the upstream?). Also, hosted patch should be submitted to Homebrew/formula-patches.

zmwangx (changes requested):

Formula/opencsg.rb
kintel: This isn't an upstream bug report. The patch simply disables building examples as we don't need the examples in homebrew. Also, the upstream doesn't have a public tracker; bugs are reported by email.

zmwangx: > The patch simply disables building examples as we don't need the examples in homebrew.

Yeah I did read the patch. But this is at least worth reporting upstream.

> bugs are reported by email.

Then we don't have to link it.

kintel: 1) This is now reported upstream
2) The patch is already submitted to formula-patches, as referenced earlier in this thread: https://github.com/Homebrew/formula-patches/issues/58

zmwangx: I must have been drowning in a sea of emails and didn't notice the formula-patches PR.

zmwangx: I merged that PR, so you need to update to a new URL. We use 7 character commit hashes, by the way.

zmwangx: That is to say, `990b9bb`.

Formula/opencsg.rb
```diff
@@ -15,19 +14,17 @@ class Opencsg < Formula
...
-    sha256 "9d710cf6c2d5495ca5ba51c0319785cefc21477c85fa3aacb9ccd3473fee54f3"
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/990b9bb/opencsg/disable-examples.diff"
+    sha256 "12cc799a6352eda4a18706eeefea059d14e23605a627dc12ed2a809f65328d69"
   end
 
   def install
     system "qmake", "-r", "INSTALLDIR=#{prefix}",
       "INCLUDEPATH+=#{Formula["glew"].opt_include}",
       "LIBS+=-L#{Formula["glew"].opt_lib} -lGLEW"
+    system "cat", "src/Makefile"
```
zmwangx: Please remove this debugging line, thanks.

zmwangx (changes requested):

zmwangx:
One more comment and hopefully we'll be good to go.

kintel:
@zmwangx In the meantime, OpenCSG-1.4.2 was made available upstream. Do you mind if I bump the version in this PR? Apart from that, this should be ready to merge.

zmwangx:
> Do you mind if I bump the version in this PR?

Yes please.

kintel:
ok, all done, and bumped to OpenCSG-1.4.2

Formula/opencsg.rb
MikeMcQuaid: I'd rather we didn't use a patch just to avoid building examples. If upstream want to provide an option to disable that, 🆒 and we can use it but I'm 👎 on patching just for that.

MikeMcQuaid (changes requested):

Formula/opencsg.rb
kintel: I've reported the ability to disable examples as a feature request upstream.
How would you go about building without patches? The build system simply doesn't properly build examples. We could of course patch how the examples are being built, but there isn't, IMO, any need to build examples for homebrew distribution.

MikeMcQuaid: I think ideally we'd patch how the examples are being built. In the medium-term our goal should be to remove patches from formulae so they can be built without. That said, it seems that's not happening in the short-term for this so happy to merge as-is.

MikeMcQuaid (approved):

MikeMcQuaid:
@BrewTestBot test this please

Formula/opencsg.rb
kintel: @zmwangx Is this good to go now?

MikeMcQuaid:

MikeMcQuaid:
Although if you get the chance to look at the Sierra failure it'd be ✨

zmwangx:
> if you get the chance to look at the Sierra failure it'd be ✨

There's nothing to look at at the moment; it depends on `qt5` which is still in stalemate.

MikeMcQuaid:
@zmwangx There's a Sierra bottle now so it's not that.

zmwangx:
Last time I checked this thread (yesterday maybe) it failed on the dependencies step due to qt5. Either I missed a newer build, or we need to rebuild this with the qt5 bottle.

zmwangx:
Yep, here's the failing build: https://bot.brew.sh/job/Homebrew%20Core/8972/

zmwangx:
Rebuilding did the trick. Shipped in 6dfa22f.

PR #5111, opened 2016-09-22
add sonar-scanner formula, replacement for sonar-runner

sv99:
sonar-runner deprecated

WARN: sonar-runner script is deprecated. Please use sonar-scanner instead.

### DIFF
--- a/Formula/sonar-scanner.rb
+++ b/Formula/sonar-scanner.rb
@@ -1,0 +1,30 @@
+class SonarScanner < Formula
+  desc "Launcher to analyze a project with SonarQube"
+  homepage "http://docs.sonarqube.org/display/SCAN/Analyzing+with+SonarQube+Scanner"
+  url "https://sonarsource.bintray.com/Distribution/sonar-scanner-cli/sonar-scanner-2.8.zip"
+  sha256 "0295365a7e5d4499ec6b46cb6c70f3fa127159b58b73930f675acd0897a6b350"
+  head "https://github.com/SonarSource/sonar-scanner-cli.git"
+
+  bottle :unneeded
+
+  def install
+    rm_rf Dir["bin/*.bat"]
+    libexec.install Dir["*"]
+    bin.install libexec/"bin/sonar-scanner"
+    etc.install libexec/"conf/sonar-scanner.properties"
+    ln_s etc/"sonar-scanner.properties", libexec/"conf/sonar-scanner.properties"
+
+    # sonar-scanner uses SONAR_SCANNER_HOME to work out where to install.
+    bin.env_script_all_files libexec/"bin/", :SONAR_SCANNER_HOME => libexec
+  end
+
+  def caveats; <<-EOS.undent
+    If this is your first install, you should adjust the default configuration:
+      #{libexec}/conf/sonar-scanner.properties
+    EOS
+  end
+
+  test do
+    assert_match version.to_s, shell_output("#{bin}/sonar-scanner --version")
+  end
+end


Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,29 @@
+class SonarScanner < Formula
+  desc "Launcher to analyze a project with SonarQube"
+  homepage "http://docs.sonarqube.org/display/SONAR/Analyzing+with+SonarQube+Scanner"
```
DomT4: Looks like this redirects to `http://docs.sonarqube.org/display/SCAN/Analyzing+with+SonarQube+Scanner` so we may as well use that.

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def install
+    # Remove windows files
+    rm_rf Dir["bin/*.bat"]
+    libexec.install Dir["*"]
+    bin.write_exec_script libexec/"bin/sonar-scanner"
+  end
+
+  def caveats; <<-EOS.undent
+      If this is your first install, you should adjust the default configuration:
```
DomT4: Use a two space indent rather than four, thanks!

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,29 @@
...
+        #{libexec}/conf/sonar-scanner.properties
```
DomT4: `opt_libexec` produces a slightly nicer output.

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      after that you should also add a new enviroment variable:
+        SONAR_SCANNER_HOME=#{libexec}
```
DomT4: We should env script this ideally. Something like [this](https://github.com/Homebrew/homebrew-core/blob/4fdd781d53d3fccb35a3c2be286b9849ff807adc/Formula/chapel.rb#L31).

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,29 @@
...
+      EOS
+  end
+
+  test do
+    system "#{bin}/sonar-scanner", "-h"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

DomT4 (changes requested):

Formula/sonar-scanner.rb
sv99: I simply rewrite existing formula. sonar-runner - deprecated by sonarcube.

sv99:
updated

Formula/sonar-scanner.rb
sv99: --version show full path to the configuration file:
```
sonar-scanner --version
...
INFO: Scanner configuration file: /usr/local/Cellar/sonar-scanner/2.8/libexec/conf/sonar-scanner.properties
...
```

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def caveats; <<-EOS.undent
+    If this is your first install, you should adjust the default configuration:
+      #{libexec}/conf/sonar-scanner.properties
+    EOS
+  end
+
+  test do
+    ENV["SONAR_SCANNER_HOME"] = libexec
```
dunn: This shouldn't be necessary, since the script in `bin` is an env wrapper.

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+  bottle :unneeded
+
+  def install
+    # Remove windows files
+    rm_rf Dir["bin/*.bat"]
+    libexec.install Dir["*"]
+    bin.install libexec/"bin/sonar-scanner"
+
+    # sonar-sanner uses this ENV to work out where to install.
+    ENV["SONAR_SCANNER_HOME"] = libexec
```
dunn: You've already installed, so is this necessary?

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+    bin.env_script_all_files libexec/"bin/", :SONAR_SCANNER_HOME => libexec
+  end
+
+  def caveats; <<-EOS.undent
+    If this is your first install, you should adjust the default configuration:
+      #{libexec}/conf/sonar-scanner.properties
```
dunn: Is this Homebrew-specific information? If not, please remove it!

Formula/sonar-scanner.rb
sv99: SONAR_SCANNER_HOME - without this variable - runtime error

Formula/sonar-scanner.rb
sv99: This is homebrew specific path of the global configuration file sonar-scanner.properties

Formula/sonar-scanner.rb
MikeMcQuaid: If it's a runtime error then this won't be set at runtime.

Formula/sonar-scanner.rb
MikeMcQuaid: Just remove the caveats then and install them into `etc.install` like other configuration files.

sv99: sonar-scanner --version show: 
INFO: Scanner configuration file: /usr/local/Cellar/sonar-scanner/2.8/libexec/conf/sonar-scanner.properties

?? not to the etc.install ??

sv99:
updated

Formula/sonar-scanner.rb
MikeMcQuaid: Ideally we'd convince it to put the file elsewhere. If not we can `etc.install` and use a symlink so both locations still work.

sv99: this path used in the script from SONAR_SCANNER_HOME. How can i add link -  etc.install copy script to the /usr/local/etc?

Formula/sonar-scanner.rb
dunn: The env wrapper script should take care of that, but it's not being run in the `def install` block, so it shouldn't be necessary there.

Formula/sonar-scanner.rb
dunn:
```ruby
etc.install "conf/sonar-scanner.properties"
ln_s etc/"conf/sonar-scanner.properties", libexec/"conf/sonar-scanner.properties"

File: Formula/sonar-scanner.rb [hunk 9]
```diff
@@ -0,0 +1,32 @@
...
+
+  def install
+    # Remove windows files
+    rm_rf Dir["bin/*.bat"]
+    libexec.install Dir["*"]
+    bin.install libexec/"bin/sonar-scanner"
+    etc.install libexec/"conf/sonar-scanner.properties"
+    ln_s etc/"sonar-scanner.properties", libexec/"conf/sonar-scanner.properties"
+
+    # sonar-sanner uses this ENV to work out where to install.
```
MikeMcQuaid: Can remove this variable.

Formula/sonar-scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+  desc "Launcher to analyze a project with SonarQube"
+  homepage "http://docs.sonarqube.org/display/SCAN/Analyzing+with+SonarQube+Scanner"
+  url "https://sonarsource.bintray.com/Distribution/sonar-scanner-cli/sonar-scanner-2.8.zip"
+  sha256 "0295365a7e5d4499ec6b46cb6c70f3fa127159b58b73930f675acd0897a6b350"
+  head "https://github.com/SonarSource/sonar-scanner-cli.git"
+
+  bottle :unneeded
+
+  def install
+    # Remove windows files
```
MikeMcQuaid: Can remove this comment.

Formula/sonar-scanner.rb
MikeMcQuaid: Comment, sorry, not variable.

Formula/sonar-scanner.rb
sv99: updated

MikeMcQuaid:

For future reference the preferred commit message format for simple version updates is `sonar-scanner 2.8 (new formula)`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #5120, opened 2016-09-22
lnav: fix build on 10.11 with Xcode 8

ilovezfs:
CC @DomT4 here we go again ...


DomT4:
Welp, not this dance again 🙈. I'm coming to strongly dislike new Xcode releases.

PR #5070, opened 2016-09-21
python: fix --universall on macOS 10.12 Sierra

rebelliard:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

Applies the patch on [Issue 27806: 2.7 32-bit builds fail on macOS 10.12 Sierra due to dependency on deleted header file QuickTime/QuickTime.h](https://bugs.python.org/issue27806) that is committed for release on `2.7.13`.

PR #5127, opened 2016-09-22
python: fix compiling python --universal on 10.11 with Xcode 8

JCount:

---

The fix in https://github.com/Homebrew/homebrew-core/pull/5070 is also needed when compiling `python --universal` on `10.11` with `Xcode 8`.

### DIFF
--- a/Formula/python.rb
+++ b/Formula/python.rb
@@ -68,9 +68,9 @@ class Python < Formula
     sha256 "c075353337f9ff3ccf8091693d278782fcdff62c113245d8de43c5c7acc57daf"
   end
 
-  # Patch for compiling universal binaries on macOS 10.12
+  # Patch for building universal binaries on macOS 10.12 and 10.11 with Xcode 8
   # https://bugs.python.org/issue27806
-  if MacOS.version >= :sierra && build.universal?
+  if build.universal? && DevelopmentTools.clang_build_version >= 800
     patch do
       url "https://hg.python.org/cpython/raw-rev/4030300fcb18"
       sha256 "e0625b20675d892abc3e0e9a58a4627b94e6ded017f352f55b7c91e214fbd248"


Formula/python.rb
```diff
@@ -68,9 +68,9 @@ class Python < Formula
     sha256 "c075353337f9ff3ccf8091693d278782fcdff62c113245d8de43c5c7acc57daf"
   end
 
-  # Patch for compiling universal binaries on macOS 10.12
+  # Patch for compiling universal binaries on macOS 10.12, as well on 10.11 with Xcode 8
   # https://bugs.python.org/issue27806
-  if MacOS.version >= :sierra && build.universal?
+  if (MacOS.version >= :sierra || MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0") && build.universal?
```
zmwangx: Too long a line.

zmwangx (changes requested):

Formula/python.rb
JCount: I was wondering, but RuboCop wasn't upset. Will fix.

zmwangx: Not everything is covered by RuboCop. We strive to stay within 80 columns (and I'm not the one that police this sort of thing the hardest), but in this case I think it's reasonable to just put build.universal? on the second line (I'm happy with 100 in a case like this, and others who police harder break the rule themselves too — see erlang.rb for instance — so it should be fine).

DomT4: You can probably just drop the `MacOS.version >= :sierra` part here, since that'll now be covered by the latter check.

DomT4 (changes requested):

Formula/python.rb
zmwangx: @DomT4 Why does MacOS.version >= :sierra imply MacOS::Xcode.installed??

JCount: Ok, thanks DomT4, that had occurred to me, but I then forgot when I was checking something else. Trying to use any other version of Xcode on Sierra is completely unsupported by `homebrew` after all.

DomT4: @zmwangx Oh, wait, you're right, I'm an idiot. Sorry, I thought this was strictly an Xcode problem for some reason.

DomT4: You could maybe do:
```ruby
if DevelopmentTools.clang_version => "8.0" && build.universal?
```
though.

DomT4: Or `DevelopmentTools.clang_build_version => "800"` to match the format we use elsewhere.

JCount: needs to be `>=` though. :stuck_out_tongue_closed_eyes:

DomT4: Ooft. Yup. I'm trying to do about 5 things at once at the moment & apparently my brain can't cope 😆

zmwangx: I guess you're quad core and need a better process scheduler.

Formula/python.rb
```diff
@@ -68,9 +68,9 @@ class Python < Formula
     sha256 "c075353337f9ff3ccf8091693d278782fcdff62c113245d8de43c5c7acc57daf"
   end
 
-  # Patch for compiling universal binaries on macOS 10.12
+  # Patch for compiling universal binaries on macOS 10.12, as well on 10.11 with Xcode 8
   # https://bugs.python.org/issue27806
-  if MacOS.version >= :sierra && build.universal?
+  if DevelopmentTools.clang_build_version >= 800 && build.universal?
```
ilovezfs: You should reverse the order of these two conditions since the second one is less common than the first.

JCount: Ahh, the optimization interpreted languages.

zmwangx:
Merged as 8ac27ee. Thank you for another contribution to Homebrew!

PR #5128, opened 2016-09-22
libvpx: Enable support for high bitdepth

kodawah:


zmwangx:
Any reason to not include this all the time?

kodabb:
@zmwangx updated

zmwangx:
No, I was asking a question. Let me rephrase it: does enabling this flag change the behavior of the program that would affect existing users of libvpx? Why isn't it enabled by default by the libvpx project?

zmwangx:
If there's performance impact for existing users then it's probably best to make it optional (and not enabled by default).

kodabb:
I originally implemented this patch as you described, but then I thought that this would harm the adoption of the 10 bit format which will be more beneficial for videos in the long run

Since the software is built as a library, and the related tools are skipped, I think that it should be as compatible as it can get, and let user optimize as they please

zmwangx:
> then I thought that this would harm the adoption of the 10 bit format which will be more beneficial for videos in the long run

Sorry, but I don't think we're in the position to push any new feature when they could be a setback to existing users. When upstream thinks it's ready to be pushed to the widest audience, they should make the feature an opt-out rather than an opt-in.

kodabb:
Don't forget that distributors are not as locked in backward compatibility as upstream is. As I said it's important that the library features as much functionality as possible.

Anyway since we're at a stalemate, I'll benchmark the slowdown introduced and report back.

MikeMcQuaid:
> Anyway since we're at a stalemate, I'll benchmark the slowdown introduced and report back.

Great, thanks!

MikeMcQuaid:
@kodabb Any news on that?

MikeMcQuaid:
Passing on this due to no response. We'll review a new PR when the above comments have been addressed. Sorry and thanks anyway!

PR #5146, opened 2016-09-23
AMQP-CPP 2.6.1

troian:
Integrate AMQP-CPP library

---

Signed-off-by: Artur Troian troian.ap@gmail.com

### DIFF
--- a/Formula/amqp-cpp.rb
+++ b/Formula/amqp-cpp.rb
@@ -1,0 +1,30 @@
+class AmqpCpp < Formula
+  desc "C++ library for communicating with a RabbitMQ message broker"
+  homepage "https://github.com/CopernicaMarketingSoftware/AMQP-CPP"
+
+  stable do
+    url "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git", revision: "ed2ffd3f03f2bba11f8af68c69ec96ab40a5c4b7"
+    version "2.6.1"
+  end
+
+  head do
+    url "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git", branch: "master"
+  end
+
+  needs :cxx11
+
+  depends_on "cmake" => :build
+  # Boost required by examples
+  depends_on "boost" => :build
+
+  def install
+    ENV.cxx11
+    system "cmake", "-DBUILD_SHARED=ON", "-DCMAKE_MACOSX_RPATH=1", "-DCMAKE_INSTALL_PREFIX=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    # TODO: use examples as test
+  end
+end
--- a/Formula/amqp-cpp.rb
+++ b/Formula/amqp-cpp.rb
@@ -1,30 +1,33 @@
 class AmqpCpp < Formula
   desc "C++ library for communicating with a RabbitMQ message broker"
   homepage "https://github.com/CopernicaMarketingSoftware/AMQP-CPP"
+  head "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git"
 
   stable do
-    url "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git", revision: "ed2ffd3f03f2bba11f8af68c69ec96ab40a5c4b7"
-    version "2.6.1"
-  end
-
-  head do
-    url "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git", branch: "master"
+    url "https://github.com/CopernicaMarketingSoftware/AMQP-CPP/archive/v2.6.2.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/amqp-cpp.rb
```diff
@@ -0,0 +1,30 @@
+class AmqpCpp < Formula
+  desc "C++ library for communicating with a RabbitMQ message broker"
+  homepage "https://github.com/CopernicaMarketingSoftware/AMQP-CPP"
+
+  stable do
```
DomT4: Don't need a `stable do` block unless you're doing something unique with stable, such as a dependency or resource not needed by `head`, etc.

Formula/amqp-cpp.rb
```diff
@@ -0,0 +1,30 @@
...
+    url "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git", revision: "ed2ffd3f03f2bba11f8af68c69ec96ab40a5c4b7"
```
DomT4: We'll need a stable, tagged release from upstream before we can consider including this.

Formula/amqp-cpp.rb
```diff
@@ -0,0 +1,30 @@
...
+    version "2.6.1"
+  end
+
+  head do
```
DomT4: Same notation on `head do` as `stable do`. Don't need it here. A simple:
```ruby
head "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git"
```
is fine. We assume the upstream branch is `master` unless told otherwise.

DomT4: We prefer hash rockets as well, instead of `:`. Won't be relevant here since the `branch` argument can be removed, but just for future reference.

Formula/amqp-cpp.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  needs :cxx11
+
+  depends_on "cmake" => :build
+  # Boost required by examples
+  depends_on "boost" => :build
+
+  def install
+    ENV.cxx11
+    system "cmake", "-DBUILD_SHARED=ON", "-DCMAKE_MACOSX_RPATH=1", "-DCMAKE_INSTALL_PREFIX=#{prefix}"
```
DomT4: use `*std_cmake_args` rather than passing ` "-DCMAKE_INSTALL_PREFIX=#{prefix}"` directly.

Formula/amqp-cpp.rb
```diff
@@ -0,0 +1,30 @@
...
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    # TODO: use examples as test
```
DomT4: Please do 😉

DomT4 (changes requested):

Formula/amqp-cpp.rb
troian: There is tag 2.6.1 so I'll put tag instead of revision

DomT4: Unless it needs a git checkout specifically you can do:
```ruby
url "https://github.com/CopernicaMarketingSoftware/AMQP-CPP/archive/v2.6.2.tar.gz"
sha256 ""
```

DomT4: (Fill in the sha256 though, if that wasn't obvious 😄)

MikeMcQuaid:
@DomT4 👍 ?

Formula/amqp-cpp.rb
```diff
@@ -0,0 +1,33 @@
+class AmqpCpp < Formula
+  desc "C++ library for communicating with a RabbitMQ message broker"
+  homepage "https://github.com/CopernicaMarketingSoftware/AMQP-CPP"
+  head "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git"
+
+  stable do
```
DomT4: We don't need this `stable do` block.

Formula/amqp-cpp.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <amqpcpp.h>
+      int main()
+      {
+        return 0;
+      }
+      EOS
+    system ENV.cxx, "test.cpp", "-std=c++11", "-L#{lib}", "-o", "test", "-lc++", "-lamqp-cpp"
```
DomT4: This goes quite a bit beyond 80 characters. Can you drop the ` "test", "-lc++", "-lamqp-cpp"` onto the line underneath, such as:
```ruby
system ENV.cxx, "test.cpp", "-std=c++11", "-L#{lib}", "-o"
                "test", "-lc++", "-lamqp-cpp"
```
Or such. Thanks!

MikeMcQuaid:

PR #5153, opened 2016-09-23
perl: fix build on 10.11 with Xcode 8

ilovezfs:


ilovezfs:
CC @mistydemeo seems upstream took your patch and ran with it, but thinks everything is fixed now https://rt.perl.org/Public/Bug/Display.html?id=128427#txn-1422265

Formula/perl.rb
```diff
@@ -1,12 +1,25 @@
 class Perl < Formula
   desc "Highly capable, feature-rich programming language"
   homepage "https://www.perl.org/"
-  url "http://www.cpan.org/src/5.0/perl-5.24.0.tar.xz"
-  mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/p/perl/perl_5.24.0.orig.tar.xz"
-  sha256 "a9a37c0860380ecd7b23aa06d61c20fc5bc6d95198029f3684c44a9d7e2952f2"
   revision 1
 
-  head "https://perl5.git.perl.org/perl.git", branch: "blead"
+  head "git://perl5.git.perl.org/perl.git", branch: "blead"
```
DomT4: Why the protocol downgrade?

ilovezfs: it wouldn't clone for me otherwise ... does it for u?

DomT4: Argh, no. Looks like they've lost the `https://`, only `http://` and `git://` are listed upstream now, and only the latter works. Nice.

ilovezfs: Yeah, pretty gross. Maybe there's a secure mirror that's actually kept in sync ... :(

DomT4: Kind of tempting to just drop `HEAD`, given we use it as a `default_formula` now and Perl does not remain completely compatible across major releases.

ilovezfs: I basically consider head specs to be for our own internal use hehe

Formula/perl.rb
```diff
@@ -20,17 +33,17 @@ class Perl < Formula
...
 
-  # Fixes Time::HiRes module bug related to the presence of clock_gettime
-  # https://rt.perl.org/Public/Bug/Display.html?id=128427
-  # Merged upstream, should be in the next release.
-  if MacOS.version >= :sierra
-    patch do
-      url "https://raw.githubusercontent.com/Homebrew/formula-patches/b18137128c4e0cb7e92e9ee007a9f78bc9d03b21/perl/clock_gettime.patch"
-      sha256 "612825c24ed19d6fa255bb42af59dff46ee65c16ea77abf4a59b754aa8ab05ac"
+  def install
+    if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
```
DomT4: Has this fix been reported upstream as well?

ilovezfs: most assuredly not

chdiza: I don't feel qualified to do it.  I don't have the ability to explain the problem or what the fix does.  Sorry! :smile:

DomT4: If we can at least report the issue upstream I'm a 👍 on carrying this.

DomT4: No preference on who does that!

ilovezfs:
original fix:
```
      inreplace "dist/Time-HiRes/HiRes.xs" do |s|
        s.gsub! "static int clock_gettime(int clock_id, struct timespec *ts) {",
                "int clock_gettime(clockid_t clock_id, struct timespec *ts) {"

        s.gsub! "static int clock_getres(int clock_id, struct timespec *ts) {",
                "int clock_getres(clockid_t clock_id, struct timespec *ts) {"
```

ilovezfs:
@chdiza do you mind verifying this new fix works for you? This should be a safer work around.

chdiza:
Seems to work over here.  I'll be back if I run into some weirdness.

ilovezfs:
Thanks @chdiza

PR #5168, opened 2016-09-24
devd 0.6

ilovezfs:
use upstream vendor directory and vendored rice-box.go


ilovezfs:
CC @cortesi does this look all right?

PR #5172, opened 2016-09-24
libpostal 0.3.1 (new formula)

ghost:

---

[libpostal](https://github.com/openvenues/libpostal) provides C APIs for parsing and normalising addresses.

First formula I've written so would greatly appreciate any help. Trying to audit the formula gives me the following error:

```
$ brew audit --new-formula libpostal
Error: type mismatch: String given
Please report this bug:
/usr/local/Homebrew/Library/Homebrew/cmd/search.rb:153:in `=~'
/usr/local/Homebrew/Library/Homebrew/cmd/search.rb:153:in `block in search_tap'
/usr/local/Homebrew/Library/Homebrew/cmd/search.rb:153:in `select'
/usr/local/Homebrew/Library/Homebrew/cmd/search.rb:153:in `search_tap'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:347:in `block (2 levels) in audit_formula_name'
```

### DIFF
--- a/Formula/libpostal.rb
+++ b/Formula/libpostal.rb
@@ -1,0 +1,44 @@
+class Libpostal < Formula
+  desc "C library for parsing or normalizing addresses"
+  homepage "https://github.com/openvenues/libpostal"
+  url "https://github.com/openvenues/libpostal/archive/v0.3.1.tar.gz"
+  head "https://github.com/openvenues/libpostal.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "snappy" => :run
+  depends_on "pkg-config" => :run
+
+  def install
+    system "autoreconf", "-fi", "--warning=no-portability"
+    system "./configure", "--prefix=#{prefix}", "--datadir=#{pkgshare}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <stdio.h>
+      #include <stdlib.h>
+      #include <libpostal/libpostal.h>
+      int main(int argc, char **argv) {
+          if (!libpostal_setup() || !libpostal_setup_parser()) {
+              exit(EXIT_FAILURE);
+          }
+
+          address_parser_options_t options = get_libpostal_address_parser_default_options();
+          address_parser_response_t *parsed = parse_address("781 Franklin Ave Crown Heights Brooklyn NYC NY 11216 USA", options);
+
+          address_parser_response_destroy(parsed);
+
+          libpostal_teardown();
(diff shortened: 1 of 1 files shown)


Soliah:
I just realised that this pulls in a lot of data during the build process which I'm not sure is required for the build step. Someone who has access to CI might want to delete the builds as they're taking up 4GB of space each.

I've raised an issue openvenues/libpostal#120 to determine if building without the data files can be done.

If not what is the best practice for creating a formula like this that has to pull in large amounts of data? Should I make my own tap?

DomT4:
The data amount it is fetching is fine; the build servers are fairly well provisioned but do clean up the immediate build and prune the cache after a set period.

However, if that data is being fetched we'd like a way to disable the build script doing that itself and add it into the formula as `resource`s instead, so that we can cache/checksum it. We can still stick it wherever upstream expect, but we like to handle external fetching ourselves as much as possible.

> Error: type mismatch: String given

This was a bug. It'll be fixed on the next tag of `brew`, which should be in the coming days.

Formula/libpostal.rb
```diff
@@ -0,0 +1,44 @@
+class Libpostal < Formula
+  desc "C library for parsing or normalizing addresses"
+  homepage "https://github.com/openvenues/libpostal"
+  url "https://github.com/openvenues/libpostal/archive/v0.3.1.tar.gz"
+  head "https://github.com/openvenues/libpostal.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "snappy" => :run
```
DomT4: Don't need to specify `:run` for most things. Only the ones the audit suggests should be build-only, or special dependencies (which none of these are). Just `depends_on "snappy"` is fine here.

Formula/libpostal.rb
```diff
@@ -0,0 +1,44 @@
...
+  depends_on "pkg-config" => :run
```
DomT4: Is this actually used at run or is this a dance to be able to use it in the test?

DomT4 (changes requested):

Formula/libpostal.rb
MikeMcQuaid: We don't add dependencies for tests alone. Just hard-code the output values in the test.

Formula/libpostal.rb
MikeMcQuaid: Agreed.

MikeMcQuaid (changes requested):

MikeMcQuaid:
Closing due to unaddressed comments. We'll review this formula again in a new PR when comments are addressed and CI is passing. Sorry and thanks anyway!

PR #5181, opened 2016-09-24
protobuf-swift: fix build with protobuf 3.1.0

ilovezfs:
There are generated files checked into upstream's SCM that need to be
regenerated due to the new version of protobuf. THe next release of
protobuf-swift will obviate the need to regerate them in the formula.


ilovezfs:
Reported upstream https://github.com/alexeyxo/protobuf-swift/issues/187

ilovezfs:
And now fixed by upstream: https://github.com/Homebrew/homebrew-core/pull/5183

PR #5189, opened 2016-09-25
sgfutils: migrate to boneyard

zmwangx:


zmwangx:
Zero installs according to analytics, 404'ing for more than four months (upstream move old releases into `old/`, 0.23 released in 2016-05-10, 0.23a released in 2016-07-09) and no one noticed. Prime candidate for boneyard apparently.

zmwangx:
We should probably tighten our grip on what non-GitHub projects are allowed here in core. They should provide proof-of-notability.

PR #4962, opened 2016-09-18
rust: disable stable on 10.12

zmwangx:
See #4960 for an example of an unfortunate soul who wasted 4-5 valuable
hours in his life just because we let him.

@ilovezfs I'd love to have your thoughts.

PR #5194, opened 2016-09-25
rust: fix cargo version & installation method

### LINKED ISSUE #5185: Stable Rust (1.11.0) formula contains much older `cargo` package than official release.
I noticed this because the version of `cargo` released with Rust 1.11.0 should support a new `cdylib` crate type, and I was getting errors. So I did some version comparison between the formula and the official installer.

Homebrew current El Capitan bottle:
```
rustc -V  => rustc 1.11.0
cargo -V => cargo 0.11.0 (259324c 2016-05-20)
```

Rust-lang.org installer "stable":
```
rustc 1.11.0 (9b21dcd6a 2016-08-15)
cargo 0.12.0-nightly (6b98d1f 2016-07-04)
```

It looks like the formula specifies https://github.com/Homebrew/homebrew-core/blob/master/Formula/rust.rb#L11 off the inaccurately-tagged https://github.com/rust-lang/cargo/releases/tag/0.11.0 which is far older (May 20) than the actual release version (July 4, for Rust v.11.10 released on August 15). The correct https://github.com/rust-lang/cargo commit appears to be rust-lang/cargo@6b98d1f8abf5b33c1ca2771d3f5f3bafc3407b93 and the code is also available via https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2016-07-05/cargo-nightly-x86_64-apple-darwin.tar.gz (which reports `0.12.0-nightly (6b98d1f 2016-07-04)`)

This issue means that Cargo does not actually possess all the advertised of the release listed on the Rust site. But if I'm not mistaken, a fix is as easy as updating the commit/dates in the formula.

nomoon:
- Fixes #5185 by installing the appropriate version of `cargo` released along with Rust 1.11.0. (EDIT: Now contains Rust 1.12 stable and alongside it cargo 0.13.0 stable)
- Cleans up the build/installation process for the cargo resource, since it didn't make sense to both check out the cargo git repo and then just download the pre-built cargo binaries anyway if choosing stable. Since building will only happen with --build-from-source and --HEAD, IMNSHO it makes sense to just retrieve the proper source commit for cargo and build as normal.

### DIFF
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -21,9 +21,9 @@ class Rust < Formula
   end
 
   bottle do
-    sha256 "e78b655e38815c01f0a366807d2ad6f57ded26c84d9b574b91d77072118b55c1" => :el_capitan
-    sha256 "59516069f1bb877240fcd01edcd21d4b5c61636f5be3c10ccebc017d7f923420" => :yosemite
-    sha256 "3667beb9555b8d6e93427c98a4b7bf3a461265e6c83e983da2607b07743a1a61" => :mavericks
+    sha256 "db162fd8fe0ad57a42bd597405f1290882d06543c285f45194010d3d215f139e" => :sierra
+    sha256 "78a222e3ea2c1d550b463f55582588287b7bcd5c84d65abe578b3d573f3c43f2" => :el_capitan
+    sha256 "e462dceb8bd2dd988ca9278c9a5839d9010674eee785d83c0f47e6efc7aa92a8" => :yosemite
   end
 
   option "with-llvm", "Build with brewed LLVM. By default, Rust's LLVM will be used."
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -7,7 +7,7 @@ class Rust < Formula
     sha256 "ac5907d6fa96c19bd5901d8d99383fb8755127571ead3d4070cce9c1fb5f337a"
 
     resource "cargo" do
-      url "https://github.com/rust-lang/cargo.git", :revision => "109cb7c33d426044d141457049bd0fffaca1327c"
+      url "https://github.com/rust-lang/cargo.git", :tag => "0.13.0", :revision => "109cb7c33d426044d141457049bd0fffaca1327c"
     end
   end
 
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -1,14 +1,13 @@
 class Rust < Formula
   desc "Safe, concurrent, practical language"
   homepage "https://www.rust-lang.org/"
-  revision 1
 
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.11.0-src.tar.gz"
-    sha256 "3685034a78e70637bdfa3117619f759f2481002fd9abbc78cc0f737c9974de6a"
+    url "https://static.rust-lang.org/dist/rustc-1.12.0-src.tar.gz"
+    sha256 "ac5907d6fa96c19bd5901d8d99383fb8755127571ead3d4070cce9c1fb5f337a"
 
     resource "cargo" do
-      url "https://github.com/rust-lang/cargo.git", :revision => "6b98d1f8abf5b33c1ca2771d3f5f3bafc3407b93"
+      url "https://github.com/rust-lang/cargo.git", :revision => "109cb7c33d426044d141457049bd0fffaca1327c"
     end
   end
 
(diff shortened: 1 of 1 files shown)


nomoon:
Maybe the revision has to be bumped in order to fix existing Rust 1.11 installations?

nomoon:
Also, Rust 1.12 is launching to stable on September 29th. Barring them changing much, updating this new formula for that release should be as simple as changing the rust source url/sha256 and the cargo git commit (once someone downloads 1.12 from the site to see which commit they picked for the release).

nomoon:
So there appears to be a problem compiling on Sierra, but I don't really understand what/how from the Jenkins log.

nomoon:
Oh, I see #4962 and that there are already Sierra problems.

ilovezfs:
The Sierra problems are fixed in HEAD. Feel free to encourage upstream to put out a new release :)

nomoon:
Is this going to be accepted soon? Because otherwise, we might as well wait until September 29th when Rust 1.12 comes out and the version will need to be updated again anyway.

nomoon:
New stable version was just released. Think I've got everything ready to go, just running a build to ensure that the compiled versions report the same as the today's distribution from rustup stable.

nomoon:
So it looks like `rustc -V` just reports `1.12.0` (no build commit/date) due to coming straight from the source tar and not git. `cargo -v` reports `cargo 0.13.0 (109cb7c 2016-08-19)` as expected.

DomT4:
Can folks decide whether https://github.com/Homebrew/homebrew-core/pull/5392 or this is the better PR, Thanks. I use Rust but not heavily enough to rule one way or another here.

nomoon:
@Domt4 I'm obviously biased here, but #5392 is just a quick update of the existing formula with the same commits/source link as I have in here. This PR also includes a clean-up of the weird `cargo-nightly-2015-09-17` resource code cruft in the existing formula. Both should result in the same install.

alexcrichton:
@nomoon this looks good to me, although I might recommend using Cargo's release tags rather than specific revisions. Although if it's easier to work with revs in Homebrew then no worries!

nomoon:
@alexcrichton I can pop that commit on. The original issue stemmed from the cargo v0.11.0 tag not tracking correctly, but it does actually look like cargo release 0.12.0 tracks rustc 1.11.0 and cargo 0.13.0 tracks cargo 1.12.0 correctly.

nomoon:
@alexcrichton Either way, the revision listed is equal to @rust-lang/cargo release tag for 0.13.0.

nomoon:
While we're at it, is it worthwhile or desirable (or possible?) to put some sort of check into the formula to make sure that we don't already have a rustup-installed version sitting in /usr/local/bin? Or is that just a function of homebrew throwing a "failed to link" error?

alexcrichton:
Ah the current recommended [version of rustup](https://rustup.rs/) doesn't actually install in `/usr/local`, although it's not *quite* official just yet (but hopefully soon!). We plan to deprecate the old `rustup.sh` which installs into `/usr/local` once rustup.rs is official, so maybe it's not worth the effort just yet?

Formula/rust.rb
```diff
@@ -3,18 +3,19 @@ class Rust < Formula
...
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.11.0-src.tar.gz"
-    sha256 "3685034a78e70637bdfa3117619f759f2481002fd9abbc78cc0f737c9974de6a"
+    url "https://static.rust-lang.org/dist/rustc-1.12.0-src.tar.gz"
+    sha256 "ac5907d6fa96c19bd5901d8d99383fb8755127571ead3d4070cce9c1fb5f337a"
 
     resource "cargo" do
-      # git required because of submodules
-      url "https://github.com/rust-lang/cargo.git", :tag => "0.11.0", :revision => "259324cd8f9bb6e1068a3a2b77685e90fda3e3b6"
+      url "https://github.com/rust-lang/cargo.git", :revision => "109cb7c33d426044d141457049bd0fffaca1327c"
```
dunn: For stable resources we prefer both a `:tag` and a `:revision`.

nomoon: Fixed in c372308.

nomoon:
Trying to figure out where the merge conflict came from. Probably had to do with the cargo bootstrap removal.

nomoon:
Ah well, looks like the other v1.12 PR got merged.

DomT4:
Sorry, was offline at the time and it was handled by another maintainer. If you're interested in removing the redundant resource please feel free to submit a fresh PR as desired, and thank you nonetheless for the submission to Homebrew.

nomoon:
No prob! :)

PR #5195, opened 2016-09-25
haskell-stack: make the bottles Sierra compatible

ilovezfs:
Remove the rpaths, which aren't actually needed here, so that the
El Capitan bottles can be deployed on Sierra too.

Add MaximumMacOSRequirement when bottling to prevent CI failure.

Raise an exception at the beginning of the `install` method to prevent
known-to-fail source builds on Sierra.

CC @zmwangx @borsboom

### DIFF
--- a/Formula/haskell-stack.rb
+++ b/Formula/haskell-stack.rb
@@ -8,7 +8,7 @@ class HaskellStack < Formula
   url "https://github.com/commercialhaskell/stack/releases/download/v1.2.0/stack-1.2.0-sdist-0.tar.gz"
   version "1.2.0"
   sha256 "872d29a37fe9d834c023911a4f59b3bee11e1f87b3cf741a0db89dd7f6e4ed64"
-  revision 1
+  revision 2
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 hmmm looks like max os requirement fails CI unconditionally ...

DomT4:
The CI passes `--build-bottle` though right? So `if build.bottle?` would be true here? Am I missing something? 🙈

ilovezfs:
@DomT4 right, so as a result, shouldn't bot skip it without failing it?

DomT4:
IIRC MMOSR is always a failure to CI, whether or not it should be is another question. I think certain other requirements may be as well, but don't quote me on that, I think about 5% of my brain is working this evening.

ilovezfs:
hmm I seem to recall it always be :green_apple:

ilovezfs:
@DomT4 I wonder if it's min os req I'm thinking of

DomT4:
Yeah, I think so. CI seems to handle that fine, off the top of my head.

PR #5198, opened 2016-09-25
aalib: migrate to boneyard

ilovezfs:


zmwangx:
Did you notice

```
$ brew uses --include-optional aalib
aview                                   gst-plugins-ugly                        homebrew/versions/gst-plugins-ugly010
gst-plugins-good                        homebrew/versions/gst-plugins-good010
```

ilovezfs:
nope. let's keep it I guess.

PR #5216, opened 2016-09-25
qt: fix building on Sierra and Xcode 8 by disabling phonon

cartr:

---

Qt 4 currently doesn't build on macOS Sierra. It turns out that this is due to changes in macOS's multimedia APIs, and by disabling Phonon, we can build Qt 4 successfully on Sierra. The formula builds and tests pass, and existing Qt 4 applications appear to work. (I tested Designer and a simple `pyqt` program.)

This should give the formulae in #1705 a little extra time to be ported, and it's also helpful for people using Qt 4 in projects outside of Homebrew.

### DIFF
--- a/Formula/qt.rb
+++ b/Formula/qt.rb
@@ -31,9 +31,9 @@ class Qt < Formula
   depends_on "postgresql" => :optional
 
   # Qt4 is dead upstream. We backported a build fix for 10.11 but do not
-  # intend to keep rescuing it forever, including for macOS 10.12. Homebrew will
-  # be migrating to Qt5 as widely as possible, which remains supported upstream.
-  depends_on MaximumMacOSRequirement => :el_capitan
+  # intend to keep rescuing it forever. Homebrew will be migrating to Qt5
+  # as widely as possible, which remains supported upstream.
+  depends_on MaximumMacOSRequirement => :sierra
 
   deprecated_option "qtdbus" => "with-dbus"
   deprecated_option "with-d-bus" => "with-dbus"
@@ -68,6 +68,10 @@ class Qt < Formula
         args << "unsupported/macx-clang"
       end
     end
+
+    # Phonon is broken on macOS 10.12+ and Xcode 8+ due to QTKit.framework
+    # being removed.
+    args << "-no-phonon" if MacOS.version >= :sierra || MacOS::Xcode.version >= "8.0"
 
     args << "-openssl-linked"
     args << "-I" << Formula["openssl"].opt_include
@@ -138,6 +142,8 @@ class Qt < Formula
     Qt Designer no longer picks up changes to the QT_PLUGIN_PATH environment
     variable as it was tweaked to search for plug-ins provided by formulae in
       #{HOMEBREW_PREFIX}/lib/qt4/plugins
+
+    Phonon is not supported on macOS Sierra.
     EOS
   end
 


Formula/qt.rb
```diff
@@ -69,6 +69,10 @@ def install
       end
     end
 
+    if MacOS.version >= :sierra
+      args << "-no-phonon"
+    end
```
bfontaine: You can inline this:
```rb
args << "-no-phonon" if MacOS.version >= :sierra
```

You can also add a short comment above the line to explain why it’s here.

cartr: I've made these changes.

DomT4:
I don't really want to do this, to be honest. The point of the MMOSR here and the notes above were almost explicitly it's not that this is impossible to fix but we're drawing a line somewhere and this is it. We already patched it in a really hacky manner for El Capitan, and I'm not really down for just lopping limbs off of Qt4 for years and years to come to buy projects more time to migrate to Qt5.

Projects have had ages to migrate to Qt5, Qt themselves have been pushing it, Qt4 isn't supported upstream, bug reports would be completely ignored, it behaves terribly on Retina displays, parts of it are laughably insecure.

The answer to https://github.com/Homebrew/homebrew-core/issues/1705 is to shunt as much to Qt5, which very few people have actually shown an interest in doing.

cartr:
@DomT4 I agree that the best solution to #1705 is to migrate all formulae to Qt 5. Relying on unsupported software isn't the best situation, and Qt 5 has a lot of improvements in security and on Retina displays.

But transitioning every project is a lot of work, and it will likely take a long time. Even if Homebrew one day manages to transition 100% of its own projects to Qt5, there will still be a lot of code out there that's written for Qt4 that will never be updated.

In my opinion, it's better to have these projects/code available with minor issues on Sierra rather than breaking them completely until we can make them perfect.

ilovezfs:
It seems to me that Qt4 would need to be forked to be its own, separate project, since the Qt Company itself has decided to retire it.

cartr:
According to BrewTestBot, Qt 4 currently fails under El Capitan with Xcode 8 because of the same issue. I've tweaked the formula to also check for Xcode 8, so this pull request should now also work on El Capitan.

kubark42:
Just to chime in here, the problem with removing Qt4 support is that it makes long-term support of  third-party packages challenging. For instance, ROS (Robotic Operating System) requires Qt4 until the latest version, Kinetic. Upgrading from ROS Indigo to ROS Kinetic might not be an option for many production environments, and so the lack of Qt4 forces developers to stay with MacOS.

In all fairness, it's not brew's fault that ROS requires installing Qt4 via brew. That's something to be fixed with ROS's maintainers. Fortunately, ROS is a very active project and so this issue is likely to be resolved. But it's easy to imagine there are other, less maintained projects out there which will see breakage because Qt4 isn't available.

ilovezfs:
@BrewTestBot test this please

MikeMcQuaid:
We're not going to do this, sorry. We have to draw a line somewhere and we can't keep indefinitely trying to support something that upstream refuses to. Please file your objections with them, not us.

cartr:
Understood. I'd like it if Homebrew continued to support Qt4, but drawing the line here is a perfectly reasonable decision to make given the lack of upstream support.

For people like @kubark42 who still need to use Qt4 to support external software, I've turned my suggested version of the `qt` formula into [a tap](https://github.com/cartr/homebrew-qt4), which you can install like this:

    brew install cartr/qt4/qt

At the moment, installation takes a long time since it has to build everything from source. I'll look into using bottles to speed it up. Once it's installed, formulae in core like `pyqt` that depend on Qt4 should be installable normally. Please feel free to file an issue at https://github.com/cartr/homebrew-qt4 if something doesn't work.

kubark42:
@MikeMcQuaid likewise understood the need to draw a hard line. That being said, projects like ROS have it tougher than many because almost all ROS modules are created by outside contributors. Many of these useful tools were made on research budgets or with student teams, so the original programmers have moved on and there's no one left to perform the Qt4-Qt5 migration. Again, totally not brew's responsibility to maintain this, and in some sense brew is a victim of its own success because projects point at brew as the solution to installing their  dependencies.

@cartr, thanks a lot. Installing now. I'll report back here with results, and move any issues to your project.

kubark42:
@cartr Success! This installed without error and ROS sees it just fine. Thanks a million.

MikeMcQuaid:
> That being said, projects like ROS have it tougher than many because almost all ROS modules are created by outside contributors. Many of these useful tools were made on research budgets or with student teams, so the original programmers have moved on and there's no one left to perform the Qt4-Qt5 migration. Again, totally not brew's responsibility to maintain this, and in some sense brew is a victim of its own success because projects point at brew as the solution to installing their dependencies.

@kubark42 It's worth mentioning: almost all Homebrew formulae are created by outside contributors and no-one is paid to work on Homebrew. It's perfectly possible (as @cartr is demonstrating) for ROS to maintain such things themselves.

MikeMcQuaid:
Cross-posting this because it's important: Qt is unsupported on 10.12 by the creators of Qt, not just Homebrew. If they continued to support it on 10.12 we'd do so too. It will likely never have another bugfix or security release and there's known security vulnerabilities in Qt 4's WebKit. One of the big reasons we're drawing a hard line is things will break and things are insecure for end-users and downstream packagers like Homebrew, ROS and @cartr are responsible for the security of the end-users who install our software.

M-Igashi:
@cartr you are a saint. until electrum migrate its dependency from 4 to 5, your tap is definitely needed for Sierra-bitcoin users.

scpeters:
@kubark42 ROS has already switched to using qt5 in its latest version (kinetic), as you noted. The gazebo simulator does not yet support qt5, but we are actively working on it. So ROS Indigo will not build on macOS Sierra, which is unfortunate, but there are other supported platforms for ROS Indigo.

This decision by the homebrew maintainers makes sense to me. It speeds up our timetable for migration, but that's a good thing, since qt4 is already unmaintained.

kubark42:
@scpeters thanks for the comment. I agree with your comments and really appreciate @cartr's efforts in that context. FWIW, I was actually successful in getting Indigo to work on macOS Sierra, check out https://github.com/mikepurvis/ros-install-osx/issues/63#issuecomment-251508063. 

Looking forward to ROS with Qt5 (although I would love if there were an option to use my existing Qt5 installation, or to install Qt5 locally like is done here: https://github.com/librepilot/LibrePilot/blob/next/make/tool_install/qt-install.qs. Installing Qt5 via brew can cause troublesome conflicts on systems with Qt development environments.)

PR #5219, opened 2016-09-25
libass: improve test

JCount:

---

fix failed test triggering a meaningless backtrace
through improved handling of return values in test block

---

**Note:** If this feels unnecessary, then I will close it no questions asked.

### DIFF
--- a/Formula/libass.rb
+++ b/Formula/libass.rb
@@ -39,7 +39,7 @@ class Libass < Formula
   end
 
   test do
-    (testpath/"test.cpp").write <<-EOS.undent
+    (testpath/"test.c").write <<-EOS.undent
       #include "ass/ass.h"
       int main() {
         ASS_Library *library;
@@ -62,7 +62,7 @@ class Libass < Formula
         }
       }
     EOS
-    system ENV.cc, "test.cpp", "-I#{include}", "-L#{lib}", "-lass", "-o", "test"
-    system "./test"
+    system ENV.cc, "test.c", "-I#{include}", "-L#{lib}", "-lass", "-o", "test"
+    shell_output("./test")
   end
 end


zmwangx:
Could you please elaborate on why the test could fail? And in case there's a traceback, it goes to stdout instead of stderr?

The last line feels too weird nevertheless so I'll probably pass on this.

JCount:
The test actually initializes first the (font) library, and then initializes the renderer with the library that was just initialized. The two `init` functions either return handles to initialized structs or if the initialization fails, `NULL`. If both structs are successfully initialized, then `main()` returns 0. If either of the two initializations fail, which would mean `libass` is not correctly installed/configured, it returns 1.

It turns out I screwed up the last line, so that's probably why it looks weird.
It should just be `shell_output("./test")` which asserts that the exit status is 0.

The traceback that I was referring was the backtrace brew itself was doing to the test program's main() loop. Which would be entirely meaningless trace. 

The two `libass` `init` functions should still generate other error messages written to stderr and return NULL, or throw an exception in severe cases. brew's test logic should still be able to catch this and do a trace of the calls to the actual `libass` library it is built against.

Currently, however;
The basic problem is that a non-zero return with `system "./test"` generates an exception that is meaningless. The two `init` functions are inherently meant to be 'safe' and would not necessarily generate exceptions on failures but rather return NULL, so the test needs to handle these. The revised commit is a bit cleaner.

**Disclaimer:**
  The above is hard to read, and at this point I honestly cannot tell if it makes sense.

_Concise version:_ The `init` functions return NULL on failure, so these need to be handle by the formula test. The revision I made may help address your concerns.

zmwangx:
> Concise version: The init functions return NULL on failure, so these need to be handle by the formula test. The revision I made may help address your concerns.

The only revision I can see here other than `.c` => `.ccp` is changing `system` to `shell_output`, which might prevent some backtrace, sure, but hardly justifies a change.

> these need to be handled

But they are already handled. You didn't change the C source file one bit. Or did I miss something?

JCount:
I feel that the primary issue is that there will always be a backtrace to main() no matter the error. I fear that this could obscure potentially helpful messages that `libass` prints to stderr by default indicating where things failed. For example, most of the ass_msg() calls in [ass_render.c#L39-#L115](https://github.com/libass/libass/blob/master/libass/ass_render.c#L39-#L115) (and related code) that give error information

JCount:
Closing this. 

On reflection, the current test does adequately ensure that the inits were successful, which really is the main, and only, goal. This PR is a meaningless attempt at 'perfection' on my part, and any improvements to the test could simply be part of the next version bump, and rightly belong there.

Thank you for humoring me @zmwangx, sorry for wasting your time here. 😓

zmwangx:
No problem.

PR #5225, opened 2016-09-25
aget: migrate to boneyard

zmwangx:


zmwangx:
- Both home page and download URLs 404;
- GitHub repo inactive since 2014 and has no tagged releases;
- Hardly anyone installs this (per analytics).

PR #5229, opened 2016-09-26
porg 0.10 (new formula)

cohei:

---

### DIFF
--- a/Formula/porg.rb
+++ b/Formula/porg.rb
@@ -10,7 +10,7 @@ class Porg < Formula
   depends_on "gtkmm3" => :recommended
 
   def install
-    args = %W[--disable-silent-rules --prefix=#{prefix}]
+    args = %W[--disable-silent-rules --prefix=#{prefix} --with-porg-logdir=#{var}/log/porg]
 
     args << "--disable-grop" if build.without? "gtkmm3"
 
--- a/Formula/porg.rb
+++ b/Formula/porg.rb
@@ -10,7 +10,7 @@ class Porg < Formula
   depends_on "gtkmm3" if build.with? "grop"
 
   def install
-    args = ["--disable-silent-rules", "--prefix=#{prefix}"]
+    args = %W[--disable-silent-rules --prefix=#{prefix}]
 
     if build.without? "grop"
       args << "--disable-grop"
--- a/Formula/porg.rb
+++ b/Formula/porg.rb
@@ -19,6 +19,8 @@ class Porg < Formula
   end
 
   test do
-    system "porg", "--version"
+    `porg -lp foo-1.0 'echo "hello world" > #{testpath}/foo.log'`
+
+    $?.success? && `porg -a` =~ /foo/
   end
 end
--- a/Formula/porg.rb
+++ b/Formula/porg.rb
@@ -12,9 +12,7 @@ class Porg < Formula
   def install
     args = %W[--disable-silent-rules --prefix=#{prefix}]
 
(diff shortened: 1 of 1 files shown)


Formula/porg.rb
```diff
@@ -0,0 +1,26 @@
+class Porg < Formula
+  desc "Source Code Package Organizer"
+  homepage "http://porg.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/porg/porg-0.10.tar.gz"
+  sha256 "48f8433193f92097824ed7a72c4babafb29dc2ffe60e7df3635664f59f09cedd"
+
+  option "without-grop", "Don't install Grop, which is the graphic interface of porg."
```
DomT4: Make this:
```ruby
option "without-gtkmm3", "Don't install Grop, which is the graphic interface of porg."
```
and then:
```ruby
depends_on "gtkmm3" => :recommended
```
Our dependency system doesn't entirely love `if build.with?` statements outside of `def install`. It works, but we prefer to avoid them where possible.

Formula/porg.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  depends_on "pkg-config" => :build
+  depends_on "gtkmm3" if build.with? "grop"
+
+  def install
+    args = ["--disable-silent-rules", "--prefix=#{prefix}"]
```
DomT4:
```ruby
args = %W[--disable-silent-rules --prefix=#{prefix}]
```
Looks a bit nicer.

Formula/porg.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    if build.without? "grop"
+      args << "--disable-grop"
```
DomT4: Unless this plows past 80 characters, which I don't think it would (?), just:
```ruby
args << "--disable-grop" if build.without? "gtkmm3"
```

Formula/porg.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  test do
+    system "porg", "--version"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

DomT4 (changes requested):

DomT4:
Install is being sandboxed to death:
```
mkdir: /var/log/porg: Operation not permitted
make[2]: *** [install-exec-local] Error 1
make[1]: *** [install-am] Error 2
```
You should be able to pass it a build option to point to `#{var}` instead.

Formula/porg.rb
cohei: OK.  I'll change it.

I thought `without-grop` would be easier for the users to understand that `Grop` won't be installed, but in `without-gtkmm3` case the description can still explain that to the users.

cohei:
I'll close this pulll request.

`porg` doesn't work at least on El Capitan according to the failing test. `porg` tracks installed packages but we can substitute homebrew for `porg` by writing formula or using `brew diy`. So `porg` is of little value.

`porg` doesn't work at least on El Capitan according to the failing test. In addition, `porg` is of little value because we can substitute homebrew for `porg`, which tracks installed packages, by writing formula or using `brew diy`.

MikeMcQuaid:
Thanks for explaining @cohei!

PR #5246, opened 2016-09-26
sqldiff 3.14.2 (new formula)

neutric:

---

### DIFF
--- a/Formula/sqldiff.rb
+++ b/Formula/sqldiff.rb
@@ -1,0 +1,21 @@
+class Sqldiff < Formula
+  desc "Displays the differences between SQLite databases."
+  homepage "https://www.sqlite.org/sqldiff.html"
+  url "https://www.sqlite.org/2016/sqlite-src-3140200.zip"
+  version "3.14.2"
+  sha256 "52507e20c2757b24b703b43ede77ce464c8106c1658a5b357974c435aa0677a6"
+
+  def install
+    system "./configure", "--disable-debug", "--prefix=#{prefix}"
+    system "make", "sqldiff"
+    bin.install "sqldiff"
+  end
+
+  test do
+    dbpath = testpath/"test.sqlite"
+    sqlpath = testpath/"test.sql"
+    sqlpath.write "create table test (name text);"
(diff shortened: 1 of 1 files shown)


Formula/sqldiff.rb
```diff
@@ -0,0 +1,21 @@
...
+  desc "Displays the differences between SQLite databases."
+  homepage "https://www.sqlite.org/sqldiff.html"
+  url "https://www.sqlite.org/2016/sqlite-src-3140200.zip"
+  version "3.14.2"
+  sha256 "52507e20c2757b24b703b43ede77ce464c8106c1658a5b357974c435aa0677a6"
+
+  def install
+    system "./configure", "--disable-debug", "--prefix=#{prefix}"
+    system "make", "sqldiff"
+    bin.install "sqldiff"
```
MikeMcQuaid: Is there no `make install`?

neutric: There isn't. This formula is basically [sqlite-analyzer.rb](https://github.com/Homebrew/homebrew-core/blob/master/Formula/sqlite-analyzer.rb) with some slight changes.

Both sqldiff and sqlite-analyzer are tools contained in the SQLite source tree. There is also dbhash, which could be useful. See the Tools section in the [SQLite documentation](https://sqlite.org/docs.html) for more information. It might make sense to make all three available in one formula.

MikeMcQuaid:
> It might make sense to make all three available in one formula.

Yeh, it feels like we could have them all be part of the `sqlite` formula and installed by default.

neutric:
Afaik, the three tools are not included in sqlite-autoconf-_version_.tar.gz which the `sqlite` formula is based on. No idea how painful it is to build SQLite from the raw source tree, but [How To Compile SQLite](https://sqlite.org/howtocompile.html) does not recommend it. Maybe roll them into a `sqlite-tools` formula?

MikeMcQuaid:
In that case I think separate formulae are fine.

MikeMcQuaid:

neutric:
Thanks for all the work you do, Mike!

PR #5248, opened 2016-09-26
purescript: fix optparse-applicative issue

ilovezfs:
Use upstream fix for building with optparse-applicative-0.13.0.0


ilovezfs:
CC @paf31 just FYI that a) 0.9.3 now needs the patch in this PR for a vanilla cabal build to succeed and b) 0.10.0-rc1 builds fine.

PR #5274, opened 2016-09-27
grok: fix build

zmwangx:

---

Deparallelize `make` due to race condition in generating `grok_capture_xdr.h`. Not reporting upstream because project is super dead.

While we're at it, add a test.


MikeMcQuaid:
> Deparallelize make due to race condition in generating grok_capture_xdr.h. Not reporting upstream because project is super dead.

Please do so anyway. If we package it: we should report it.

PR #5279, opened 2016-09-27
Pyinvoke

mdaniel:
- [X] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)? (well, yes, modulo issue 5249)

---

This formula started life at version 0.12.2 because that was the specific version required to build another piece of software. However, 0.12 is not the latest version; that is 0.13.0 and is the one at the head of this PR. But I thought it would be useful to leave 0.12.2 in the history in case someone also wishes to build using the older version of pyinvoke

As you might suspect, I purposefully did not call the formula "invoke", however, I had mixed feelings about leaving their `inv` and `invoke` on the `$PATH`, so if you have guidance around that I'd welcome it.

### DIFF
--- a/Formula/pyinvoke.rb
+++ b/Formula/pyinvoke.rb
@@ -5,7 +5,6 @@ class Pyinvoke < Formula
   homepage "http://pyinvoke.org/"
   url "https://github.com/pyinvoke/invoke/archive/0.13.0.tar.gz"
   sha256 "4942aa901618f99afa70271d7b15fd018da0f174178fbca4048354c785e7736a"
-
   head "https://github.com/pyinvoke/invoke.git"
 
   depends_on :python if MacOS.version <= :snow_leopard
@@ -26,8 +25,8 @@ class Pyinvoke < Formula
           for pattern in patterns:
               run("rm -rf {}".format(pattern))
     EOS
-    mkdir_p testpath/"foo"/"bar"
-    mkdir_p testpath/"baz"
+    (testpath/"foo"/"bar").mkpath
+    (testpath/"baz").mkpath
     system bin/"invoke", "clean"
     assert !File.exist?(testpath/"foo"), "\"pyinvoke clean\" should have deleted \"foo\""
     assert File.exist?(testpath/"baz"), "pyinvoke should have left \"baz\""
--- a/Formula/pyinvoke.rb
+++ b/Formula/pyinvoke.rb
@@ -1,0 +1,38 @@
+class Pyinvoke < Formula
+  include Language::Python::Virtualenv
+
+  desc "Pythonic task management & command execution"
+  homepage "http://pyinvoke.org/"
+  url "https://github.com/pyinvoke/invoke/archive/0.12.2.tar.gz"
+  sha256 "dfc9892a0731456c785bd9d515fe761767729d32155cfd77bd0f042bd6228187"
+
+  head "https://github.com/pyinvoke/invoke.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    virtualenv_install_with_resources
+  end
+
+  test do
+    (testpath/"tasks.py").write <<-EOS.undent
+      from invoke import run, task
+
+      @task
+      def clean(extra=''):
+          patterns = ['foo']
+          if extra:
+              patterns.append(extra)
+          for pattern in patterns:
+              run("rm -rf {}".format(pattern))
+    EOS
+    mkdir_p testpath/"foo"/"bar"
+    mkdir_p testpath/"baz"
+    system bin/"invoke", "clean"
+    assert !File.exist?(testpath/"foo"), "\"pyinvoke clean\" should have deleted \"foo\""
+    assert File.exist?(testpath/"baz"), "pyinvoke should have left \"baz\""
+    system bin/"invoke", "clean", "--extra=baz"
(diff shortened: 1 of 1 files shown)


Formula/pyinvoke.rb
```diff
@@ -0,0 +1,38 @@
...
+  homepage "http://pyinvoke.org/"
+  url "https://github.com/pyinvoke/invoke/archive/0.13.0.tar.gz"
+  sha256 "4942aa901618f99afa70271d7b15fd018da0f174178fbca4048354c785e7736a"
+
+  head "https://github.com/pyinvoke/invoke.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    virtualenv_install_with_resources
```
MikeMcQuaid: You're not using any `resource`s here? Does this have an external dependencies? CC @ilovezfs @tdsmith to check I haven't misunderstood this.

mdaniel: Correct, but compare that one line (which does everything one would wish for a `setup.py` based installation) versus:

```ruby
  def install
    ENV.prepend_create_path "PYTHONPATH", "#{libexec}/lib/python2.7/site-packages"

    system "python", "setup.py", "install", "--prefix=#{libexec}"
    bin.install Dir[libexec/"bin/*"]
    bin.env_script_all_files(libexec+"bin", :PYTHONPATH => ENV["PYTHONPATH"])
  end
```

It's not me being lazy; I can copy-and-paste with the best of them, but I think it's a vast legibility improvement. I was unable to find anything in [the Formula Cookbook](https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#specifying-gems-python-modules-go-projects-etc-as-dependencies) that implied one should favor the more verbose approach, so I didn't.

MikeMcQuaid: Yeh, not really your issue here just interested what @tdsmith thinks re: naming etc.

tdsmith:
It looks okay to me. It's installing all zero of the resources :)

MikeMcQuaid:
@tdsmith Last question: is it worth bottling things like this or better just have them `bottle :unneeded`?

tdsmith:
Doesn't hurt but pure-Python modules with no dependencies don't gain any appreciable benefits from bottling.

Formula/pyinvoke.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      @task
+      def clean(ctx, extra=''):
+          patterns = ['foo']
+          if extra:
+              patterns.append(extra)
+          for pattern in patterns:
+              run("rm -rf {}".format(pattern))
+    EOS
+    mkdir_p testpath/"foo"/"bar"
```
DomT4: `(testpath/"foo/bar").mkpath` is preferred.

Formula/pyinvoke.rb
```diff
@@ -0,0 +1,38 @@
...
+    mkdir_p testpath/"baz"
```
DomT4: And `(testpath/"baz").mkpath`

Formula/pyinvoke.rb
```diff
@@ -0,0 +1,38 @@
...
+    system bin/"invoke", "clean"
+    assert !File.exist?(testpath/"foo"), "\"pyinvoke clean\" should have deleted \"foo\""
```
DomT4: Doesn't need to be decided right here, but now we're enforcing Ruby 2.0 we can use things like `assert_predicate` if we like, which produces nicer error messages by default without necessarily having to explicitly write out one.

Formula/pyinvoke.rb
```diff
@@ -0,0 +1,38 @@
+class Pyinvoke < Formula
+  include Language::Python::Virtualenv
+
+  desc "Pythonic task management & command execution"
+  homepage "http://pyinvoke.org/"
+  url "https://github.com/pyinvoke/invoke/archive/0.13.0.tar.gz"
+  sha256 "4942aa901618f99afa70271d7b15fd018da0f174178fbca4048354c785e7736a"
+
```
DomT4: No empty line required here.

DomT4 (changes requested):

Formula/pyinvoke.rb
mdaniel: How does one `assert_predicate thing, *negated* operation`? I tried the simplistic `a_p thing, !:exist?` and got a horrible explosion and I didn't see any `:missing?` on `File`

I only found [one example](https://github.com/Homebrew/homebrew-core/blob/0379d0d/Formula/conan.rb#L141) across all the `homebrew-*` Taps I have locally, and [the documentation](http://ruby-doc.org/stdlib-2.0.0/libdoc/minitest/rdoc/MiniTest/Assertions.html#assert_predicate-method) wasn't helpful.

DomT4: That would be `refute_predicate`. Probably worth me polling @ilovezfs and @zmwangx on this change though, I'm not around as much as I used to be so they'll end up dealing with it day to day if we decide to change.

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:

For future reference the preferred commit message format for simple version updates is `pyinvoke 0.13.0 (new formula)`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #5286, opened 2016-09-27
NSD 4.1.13

aabdnn:
This PR updates NSD to version 4.1.13 (note that 4.1.12 was released, but had a bug, so it was quickly followed up by 4.1.13)


MikeMcQuaid:
Thanks again @aabdnn!

PR #5315, opened 2016-09-28
cayley 0.6.0

ilovezfs:
glide vendoring


dunn:
Guess they moved away from godep 
```
==> godep restore
godep: open Godeps/Godeps.json: no such file or directory

ilovezfs:
@dunn yup, now it belongs to the `glide` clan.

PR #5326, opened 2016-09-28
enable --with-debug for nghttp2

jeroen:
Debugging failing HTTP/2 connections requires that nghttp2 is build configured `--with-debug`, see here: https://github.com/curl/curl/issues/674#issuecomment-187181546


DomT4:
Does this actually work? Homebrew's superenv can kill debug stuff in the background, even if formulae have options set.

bfontaine:
@jeroenooms ping?

jeroenooms:
Hmm I'm not sure actually, I ended up building from source due to some other problem with the recipe. Thanks though.

PR #5327, opened 2016-09-28
telegraf 1.0.1

ilovezfs:
use gdm vendoring


ilovezfs:
@DomT4 +11 −365 lol

DomT4 (approved):
That's quite some diff 😸

PR #5332, opened 2016-09-28
aqbanking: migrate to boneyard

zmwangx:


zmwangx:
Homepage http://www.aqbanking.de/ as well as download URL are down. There's a GitHub mirror https://github.com/aqbanking/aqbanking but

1. There're no release tarballs there, so one has to run autoreconf;
2. Even after patching it still doesn't compile and insist on

    ```
git log -n 100 --pretty="format:------------------------------------------------------------------%n%ai %an%n%s%n%b" >ChangeLog
```

The formula is also not at all popular, with a total of ~50 install events. It appears to be more trouble than its worth.

I'll let it sit for a while in case upstream comes back up.

DomT4 (approved):

MikeMcQuaid (approved):

ilovezfs:
Looks like upstream came back.

MikeMcQuaid:
In that case I think it's worth holding off.

MikeMcQuaid:
> There're no release tarballs there, so one has to run autoreconf

That's fine.

> Even after patching the formula with autoconf, automake and libtool, it still doesn't compile and insist on
> 
> git log -n 100 --pretty="format:------------------------------------------------------------------%n%ai %an%n%s%n%b" >ChangeLog
> fatal: Not a git repository (or any of the parent directories): .git

Seems it's expecting a git checkout instead of the tarball which seems easy enough.

PR #5334, opened 2016-09-28
mg3a: Fix --with-all option

bw-yuhogg:

---

Previously, it failed with the message:
Error: no implicit conversion of String into Integer

### DIFF
--- a/Formula/mg3a.rb
+++ b/Formula/mg3a.rb
@@ -21,7 +21,7 @@ class Mg3a < Formula
 
   def install
     if build.with?("all")
-      mg3aopts = "-DALL" if build.with?("all")
+      mg3aopts = %w[-DALL]
     else
       mg3aopts = %w[-DDIRED -DPREFIXREGION -DUSER_MODES -DUSER_MACROS]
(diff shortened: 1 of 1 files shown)


Formula/mg3a.rb
```diff
@@ -21,7 +21,7 @@ class Mg3a < Formula
 
   def install
     if build.with?("all")
-      mg3aopts = "-DALL" if build.with?("all")
+      mg3aopts = ["-DALL"] if build.with?("all")
```
zmwangx: Maybe `%w[-DALL]` for style consistency?

zmwangx (changes requested):

Formula/mg3a.rb
zmwangx: And

```ruby
if build.with?("all")
  mg3aopts = ["-DALL"] if build.with?("all")
```

...

The latter `if build.with?("all")` is redundant.

rwhogg: Wow, yes it is. I'll fix that too while I'm here

zmwangx (approved):

zmwangx:
Thanks!

PR #5343, opened 2016-09-28
Require LLVM 3.8 for ponyc

SeanTAllen:
- [x ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x ] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

We are finding that users who install ponyc via homebrew without
having llvm 3.8 are installed because by default ponyc is built
with Link Time Optimizations on. This means they need to have the
same version of LLVM installed that was used when building ponyc.

This commit requires LLVM 3.8 specifically rather than the general
LLVM formula that could result in different versions. We believe
this should fix the issue but are unsure as building from source
on the user's machine always solves the problem and it is only
once this has been merged and a bottle created that we can test
whether the fix was successful.

If you have any advice on another way to test, we would appreciate
that.

### DIFF
--- a/Formula/ponyc.rb
+++ b/Formula/ponyc.rb
@@ -12,7 +12,7 @@ class Ponyc < Formula
   end
 
   depends_on :macos => :yosemite
-  depends_on "llvm"
+  depends_on "homebrew/versions/llvm38"
   depends_on "libressl"
   depends_on "pcre2"
   needs :cxx11


zmwangx:
Thanks for bringing this up, but unfortunately we can't take this in as is because default cross-tap dependency is not allowed in core.

More on topic, I think your diagnostics in https://github.com/ponylang/ponyc/issues/1274 is incorrect.

> your system has something like LLVM 3.7 (according to Apple versioning).

I could be wrong about `ponyc` locates LLVM at runtime, but in general that's not how Homebrew works. When you specify `depends_on "llvm"` in the formula, our packaged LLVM is used regardless of what's on the system. Which means LLVM 3.8.1. Even if the user has an older version of brewed LLVM beforehand, running `brew install ponyc` should update their LLVM first. Now it's possible to switch to an older version of LLVM within brew (the user's `0.4.0-8d8c38a [release] (llvm 3.6.2)` looks suspicious), but that's not typical. I would ask for more info from that user first, specifically `brew info llvm`.

SeanTAllen:
@zmwangx I get the same issue when installing via homebrew bottle.

here's my `brew info llvm` output:

```
brew info llvm
llvm: stable 3.8.1 (bottled), HEAD [keg-only]
Next-gen compiler infrastructure
http://llvm.org/
/usr/local/Cellar/llvm/3.6.2 (812 files, 275.7M)
  Poured from bottle on 2016-04-05 at 06:17:46
/usr/local/Cellar/llvm/3.8.1 (1,815 files, 612.6M)
  Poured from bottle on 2016-09-12 at 20:59:38
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/llvm.rb
==> Dependencies
Build: xz ✔, cmake ✔
Recommended: libffi ✔
Optional: graphviz ✘, ocaml ✘
==> Requirements
Optional: python ✔
==> Options
--universal
	Build a universal binary
--with-all-targets
	Build all targets. Default targets: AMDGPU, ARM, NVPTX, and X86
--with-graphviz
	Build with graphviz support
--with-lldb
	Build LLDB debugger
--with-ocaml
	Build with ocaml support
--with-python
	Build bindings against custom Python
--with-shared-libs
	Build shared instead of static libraries
--with-test
	Build LLVM unit tests
--with-toolchain
	Build with Toolchain to facilitate overriding system compiler
--without-compiler-rt
	Do not build Clang runtime support libraries for code sanitizers, builtins, and profiling
--without-libcxx
	Do not build libc++ standard library
--without-libffi
	Do not use libffi to call external functions
--without-libunwind
	Do not build libunwind library
--without-lld
	Do not build LLD linker
--without-polly
	Do not build Polly optimizer
--without-rtti
	Build without C++ RTTI or exception handling
--without-utils
	Do not install utility binaries
--HEAD
	Install HEAD version
==> Caveats
LLVM executables are installed in /usr/local/opt/llvm/bin.
Extra tools are installed in /usr/local/opt/llvm/share/llvm.
To use the bundled libc++ please add the following LDFLAGS:
  LDFLAGS="-L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib"

This formula is keg-only, which means it was not symlinked into /usr/local.

macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

Generally there are no consequences of this for you. If you build your
own software and it requires this formula, you'll need to add to your
build variables:

    LDFLAGS:  -L/usr/local/opt/llvm/lib
    CPPFLAGS: -I/usr/local/opt/llvm/include


If you need Python to find bindings for this keg-only formula, run:
  echo /usr/local/opt/llvm/lib/python2.7/site-packages >> /usr/local/lib/python2.7/site-packages/llvm.pth
```

zmwangx:
> So, from the error it appears that XCode 8 is a requirement if installing via Homebrew.

It could be, that's the supported configuration and what we use on CI.

But could you please try

```
brew link --force llvm
```

and see if that helps?

SeanTAllen:
@zmwangx sorry, i had already installed xcode 8 to test that and that fixes the issue.

zmwangx:
@SeanTAllen Great. I think the requirement may only apply to our bottle (built with Xcode-8) though. Depending on whether that's true, I think we have two solutions going forward:

1. If it only applies to our bottle, we can disable pouring bottle on systems without Xcode 8:

    ```ruby
    pour_bottle? do
      reason <<-EOS.undent
      The bottle requires Xcode 8 to work properly.
      EOS
      satisfy { DevelopmentTools.clang_build_version => "800" }
    end
    ```

    or maybe

    ```ruby
      satisfy { MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0" }
    ```

    (I'm not sure whether CLT-only config works here.)

    2. If even a source build

    ```
    brew install -s ponyc
    ```

    does not work, we can add Xcode 8 requirement in our formula:

    ```
    depends_on :xcode => "8.0"
    ```

Could you please ask your user to try `brew install -s ponyc` and report back the findings?

zmwangx:
By the way, I opened an issue #5346 for this.

PR #5349, opened 2016-09-28
libgda: revision for readline

ilovezfs:


zmwangx:
Um, you have two `revision 1` in this formula.

PR #5351, opened 2016-09-29
p11-kit 0.23.2 (new formula)

dunn:

---

The triumphant return.  Apparently all we needed was an env var: https://bugs.freedesktop.org/show_bug.cgi?id=91602#c1

https://github.com/Homebrew/homebrew-boneyard/pull/176

### DIFF
--- a/Formula/p11-kit.rb
+++ b/Formula/p11-kit.rb
@@ -1,0 +1,29 @@
+class P11Kit < Formula
+  desc "Library to load and enumerate PKCS#11 modules"
+  homepage "https://p11-glue.freedesktop.org"
+  url "https://p11-glue.freedesktop.org/releases/p11-kit-0.23.2.tar.gz"
+  sha256 "ba726ea8303c97467a33fca50ee79b7b35212964be808ecf9b145e9042fdfaf0"
+
+  depends_on "pkg-config" => :build
+  depends_on "libffi"
+
+  def install
+    # https://bugs.freedesktop.org/show_bug.cgi?id=91602#c1
+    ENV["FAKED_MODE"] = "1"
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--disable-trust-module",
+                          "--prefix=#{prefix}",
+                          "--sysconfdir=#{etc}",
+                          "--with-module-config=#{etc}/pkcs11/modules",
+                          "--without-libtasn1"
+    system "make"
+    system "make", "check"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/p11-kit", "list-modules"
+  end
+end
--- a/tap_migrations.json
+++ b/tap_migrations.json
@@ -204,7 +204,6 @@
   "ori": "homebrew/fuse",
   "orpie": "homebrew/boneyard",
   "osxfuse": "caskroom/cask",
-  "p11-kit": "homebrew/boneyard",
   "pan": "homebrew/boneyard",
   "paq8px": "homebrew/boneyard",
   "par2tbb": "homebrew/boneyard",


ilovezfs:
Not sure it actually matters one way or another at all but technically "shouldn't" the envvar be set only during make check?

dunn:
Hrm hrm, that's probably true.

ilovezfs:
If it makes no difference, you can probably just ignore that technicality. I assume if you grep the sources, the only reference to it is in make check?

dunn:
Appears in two test source files:
- https://cgit.freedesktop.org/p11-glue/p11-kit/tree/p11-kit/test-conf.c#n451
- https://cgit.freedesktop.org/p11-glue/p11-kit/tree/common/test-compat.c#n138

DomT4 (approved):

MikeMcQuaid (approved):

ilovezfs:
Is new formula the correct commit subject when something is returning from :skull: ?

DomT4:
Doesn't matter too much. I tend to use `(return from boneyard)` or something personally, but it's not a big deal.

Formula/p11-kit.rb
```diff
@@ -0,0 +1,27 @@
+class P11Kit < Formula
+  desc "Library to load and enumerate PKCS#11 modules"
+  homepage "https://p11-glue.freedesktop.org"
+  url "https://p11-glue.freedesktop.org/releases/p11-kit-0.23.2.tar.gz"
+  sha256 "ba726ea8303c97467a33fca50ee79b7b35212964be808ecf9b145e9042fdfaf0"
+
+  depends_on "pkg-config" => :build
+  depends_on "libffi"
```
toonetown: libffi and libtasn1 can both be `:recommended`, if we care about providing people with that flexibility.

Formula/p11-kit.rb
```diff
@@ -0,0 +1,27 @@
...
+  depends_on "libtasn1"
+
+  def install
+    # https://bugs.freedesktop.org/show_bug.cgi?id=91602#c1
+    ENV["FAKED_MODE"] = "1"
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--disable-trust-module",
+                          "--prefix=#{prefix}"
```
toonetown: I would suggest adding `--sysconfdir=#{etc}`.  Also, if you don't add `--with-module-config=#{etc}/pkcs11/modules`, then any modules that are created end up going into the `#{prefix}` path - which will change every time p11-kit is updated.

toonetown (commented):
I don't really have much say - but I would _strongly_ recommend at least adding the `--with-module-config` (and in my testing, I needed to add the `--sysconfdir` option too).  The optional providing of dependencies is less important IMO.

toonetown:
And BTW - way to be on it!  I was messing around with it yesterday and just left that comment on the upstream bug when I found out what the issue was.  Kind of bummed that you beat me to the punch... 😉

Formula/p11-kit.rb
dunn: Makes sense, thanks!

Formula/p11-kit.rb
dunn: Any idea how popular each option is?

toonetown: I'd actually go from the documentation at https://p11-glue.freedesktop.org/doc/p11-kit/devel-building.html:

> **libffi** for sharing of PKCS#11 modules between multiple callers in the same process. It is highly recommended that this dependency be treated as a required dependency.

I'd just set this as a dependency - don't make it optional.  It kind of defeats the purpose.

> **libtasn1** is required to build the trust module and code that interacts with certificates.

Since you are setting `--disable-trust-module`, you probably don't need to list `libtasn1` as a dependency...but you might need to also specify `--without-libtasn1` (but I'm not positive).

Alternatively, you could build `--disable-trust-module` when `libtasn1` is not included - and actually build the trust module when `libtasn1` is set as a dependency.  However, you may also need to specify `--with-trust-paths` because there really isn't a good "standard" trust path that I'm aware of on OS X.

toonetown:
I'm a 👍

DomT4 (approved):
I'm Dominyk, and I approve this PR. (And hope other non-Americans understand why I've phrased the approval like this 😅).

rprimus:
Fri 30 Sep 2016 08:10:16 BST

Just noticed this new formula being installed during a ```brew upgrade --all```:

```
==> Upgrading p11-kit
==> Downloading https://homebrew.bintray.com/bottles/p11-kit-0.23.2.el_capitan.bottle.tar.gz
==> Pouring p11-kit-0.23.2.el_capitan.bottle.tar.gz
🍺  /usr/local/Cellar/p11-kit/0.23.2: 58 files, 846.9K
```

```
 % brew info p11-kit
p11-kit: stable 0.23.2 (bottled)
Library to load and enumerate PKCS#11 modules
https://p11-glue.freedesktop.org
/usr/local/Cellar/p11-kit/0.18.4 (51 files, 829.2K) *
  Built from source on 2013-10-10 at 20:35:41
```

Just wondering why a newer bottle was not attempted to be built from ```CI```?

PR #5359, opened 2016-09-29
open-mpi 2.0.1 revision 1

ggouaillardet:

---

issue an user-friendly error message if the path to a Unix socket is too long.
it has been reported several times on the Open MPI mailing list that mpirun
fails to start (bind() error). the root cause is $TMPDIR can be quite long
and hence the path to a Unix socket is too long. unfortunatly, that was
not catched properly, leading to an un-understandable error message.

the patch is coming from https://github.com/open-mpi/ompi/pull/2135

Signed-off-by: ggouaillardet gilles@rist.or.jp

### DIFF
--- a/Formula/open-mpi.rb
+++ b/Formula/open-mpi.rb
@@ -4,6 +4,14 @@ class OpenMpi < Formula
   url "https://www.open-mpi.org/software/ompi/v2.0/downloads/openmpi-2.0.1.tar.bz2"
   sha256 "fed74f4ae619b7ebcc18150bb5bdb65e273e14a8c094e78a3fea0df59b9ff8ff"
 
+  # Patch Open MPI in order to fallback to oob/tcp when a Unix socket path
+  # is too long, and instead of silently truncating the path.
+  # This frequently occurs on OSX when both $TMPDIR and hostname are "long"
+  patch do
+    url "https://github.com/open-mpi/ompi/pull/2135.patch"
+    sha256 "35005a892ec1321fa7b6f427967e15e7a83cce0421043a3f42653675be9a24bd"
+  end
+
   bottle do
     sha256 "b2542fee9f047c5a559ebb30836ee50bacc11165848af346d15a6c5a80a669da" => :sierra
     sha256 "ae42aeebda77ce68836d690f6859f662da0a325236f539c6fb433475bfd56e8e" => :el_capitan


Formula/open-mpi.rb
```diff
@@ -3,6 +3,11 @@ class OpenMpi < Formula
   homepage "https://www.open-mpi.org/"
   url "https://www.open-mpi.org/software/ompi/v2.0/downloads/openmpi-2.0.1.tar.bz2"
   sha256 "fed74f4ae619b7ebcc18150bb5bdb65e273e14a8c094e78a3fea0df59b9ff8ff"
+  revision 1
+  patch do
```
MikeMcQuaid: Add a newline above this line (ideally fix so `brew audit --strict` is 👍).

Formula/open-mpi.rb
```diff
@@ -3,6 +3,11 @@ class OpenMpi < Formula
...
+    url "https://github.com/open-mpi/ompi/pull/2135.patch"
```
MikeMcQuaid: Can you add a brief comment describing what the patch is needed for? Thanks for submitting a PR to both projects, though!

MikeMcQuaid (changes requested):

ggouaillardet:
@MikeMcQuaid i made both changes and updated the PR, and am now double checking everything is fine

Formula/open-mpi.rb
```diff
@@ -3,6 +3,15 @@ class OpenMpi < Formula
   homepage "https://www.open-mpi.org/"
   url "https://www.open-mpi.org/software/ompi/v2.0/downloads/openmpi-2.0.1.tar.bz2"
   sha256 "fed74f4ae619b7ebcc18150bb5bdb65e273e14a8c094e78a3fea0df59b9ff8ff"
+  revision 1
```
MikeMcQuaid: I think you can probably omit this; I don't think this issue is severe enough to force a reinstall for anyone unaffected.

MikeMcQuaid (changes requested):

Formula/open-mpi.rb
ggouaillardet: @MikeMcQuaid done, we should be good for primetime

MikeMcQuaid:

For future reference the preferred commit message format is `open-mpi: fix handling long TMPDIRs.` and there's no need to sign off your own commits.

PR #5380, opened 2016-09-29
Add command-line instructions to switch bash shell

rgladwell:
Added more detailed instructions on how to switch to Homebrew bash shell.


MikeMcQuaid:
We intentionally don't add more detail here as users need to know how to do this themselves to they can recover from removal or corruption.

MikeMcQuaid:
But thanks for the PR anyway!

PR #5392, opened 2016-09-29
rust 1.12.0

alex:

---

### DIFF
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -3,12 +3,12 @@ class Rust < Formula
   homepage "https://www.rust-lang.org/"
 
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.11.0-src.tar.gz"
-    sha256 "3685034a78e70637bdfa3117619f759f2481002fd9abbc78cc0f737c9974de6a"
+    url "https://static.rust-lang.org/dist/rustc-1.12.0-src.tar.gz"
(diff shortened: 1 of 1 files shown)


zmwangx:
There's concern in #5185 that the `cargo` we ship is outdated. Maybe we should use the just released [0.13.0](https://github.com/rust-lang/cargo/releases/tag/0.13.0)?

zmwangx:
I also wonder if we should update the bootstrap `cargo`. It's still stuck at the 1.3.0 period.

alex:
@zmwangx I can upgrade the cargo if you'd like, that one hadn't been released at the time I sent this :-) Let me know if you prefer.

zmwangx:
I'm not a rust user so I'm just making wild guesses :)

alex:
@zmwangx since jenkins hadn't started yet, I went ahead and did the cargo 0.13.0 as well

zmwangx:
Cool, I'll make sure the first commit is removed from queue.

alexcrichton:
Ah yes Cargo doesn't track the same version number as rustc itself, so 0.13.0 used here is indeed correct!


Also it's ok to not worry too much about updating the Cargo bootstrap binary. If needed it can be updated, but if the compile succeeds then you should be A-OK

zmwangx:
Merged as 18a8389. Thank you for another contribution to Homebrew!

PR #5394, opened 2016-09-29
jless 481

zmwangx:

---

I noticed this due to broken download URL during bottling. First reaction was to boneyard since this hasn't been updated since 2012 (and Debian has hardly updated either except for removing `jless_382-iso262.orig.tar.gz` in favor of `jless_382-iso262-3.tar.gz`). However, after checking out its homepage I realized it's still going strong and has undergone dozens of beta and stable releases. It no longer needs patching. Also, there are still a few people installing this software...

### DIFF
--- a/Formula/jless.rb
+++ b/Formula/jless.rb
@@ -1,12 +1,11 @@
 # jless (Jam Less) is Japan-ized Less.
 # jless supports ISO 2022 code extension techniques and Japanese codes.
 class Jless < Formula
-  desc "Jless file pager supporting ISO2022"
+  desc "File pager supporting ISO2022"
   homepage "http://www.greenwoodsoftware.com/less/"
-  url "https://mirrors.ocf.berkeley.edu/debian/pool/main/j/jless/jless_382-iso262.orig.tar.gz"
-  mirror "https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/j/jless/jless_382-iso262.orig.tar.gz"
-  version "382+iso262+ext03"
-  sha256 "d33cc51f220647d015fd526b3782cde03c5d8b18c5233e6388a0047a70446fe7"
+  url "http://www.greenwoodsoftware.com/less/less-481.tar.gz"
+  sha256 "3fa38f2cf5e9e040bb44fffaa6c76a84506e379e47f5a04686ab78102090dda5"
+  version_scheme 1
 
   bottle do
     revision 1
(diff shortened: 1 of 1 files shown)


Formula/jless.rb
```diff
@@ -1,12 +1,11 @@
...
-  desc "Jless file pager supporting ISO2022"
+  desc "File pager supporting ISO2022"
   homepage "http://www.greenwoodsoftware.com/less/"
-  url "https://mirrors.ocf.berkeley.edu/debian/pool/main/j/jless/jless_382-iso262.orig.tar.gz"
-  mirror "https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/j/jless/jless_382-iso262.orig.tar.gz"
-  version "382+iso262+ext03"
-  sha256 "d33cc51f220647d015fd526b3782cde03c5d8b18c5233e6388a0047a70446fe7"
+  url "http://www.greenwoodsoftware.com/less/less-481.tar.gz"
+  sha256 "3fa38f2cf5e9e040bb44fffaa6c76a84506e379e47f5a04686ab78102090dda5"
+  version_scheme 1
```
MikeMcQuaid: Just to double-check: presumably upgrade fails without this?

Formula/jless.rb
```diff
@@ -15,26 +14,16 @@ class Jless < Formula
...
-    system "make install binprefix=j manprefix=j"
+    system "make", "install", "binprefix=j", "manprefix=j"
   end
 
   def caveats
     "You may need to set the environment variable 'JLESSCHARSET' to japanese-utf8"
   end
+
+  test do
+    assert_match version.to_s, shell_output("#{bin}/jless --version")
```
MikeMcQuaid: A better test would be awesome but this is definitely better than nothing so 👍 either way.

Formula/jless.rb
zmwangx: @MikeMcQuaid Sorry, I accidentally deleted your comment when I was trying to delete my own (I started a review instead of adding a single comment). To answer the question

> Just to double-check: presumably upgrade fails without this?

No, upgrade works without this, but `382+iso262+ext03` to `481` feels like a version scheme change to me, in principle. I don't mind removing it.

Formula/jless.rb
zmwangx: Since this is just an implementation of `less` it doesn't really work in a dumb term.

Formula/jless.rb
MikeMcQuaid: We only bump that value when we need to so 👍 to leaving it as-is.

Formula/jless.rb
MikeMcQuaid: Could try to `less` a one-line file and check it returns the same output (as presumably it'll exit immediately)? If not: no worries.

zmwangx:
I just realized that this has converged with `homebrew/dupes/less`. Exact same checksum, except greenwoodsoftware.com is used as a mirror there. Closing in favor of a boneyard PR. (Ideally this should be a tap migration, but we don't have https://github.com/Homebrew/brew-evolution/blob/master/004-better-tap-migrations.md yet.)

PR #5397, opened 2016-09-29
liquibase 3.5.1

jessedearing:
- [✅ ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [✅ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [✅ ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [✅ ] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

Author has published a [blog post](http://www.liquibase.org/2016/09/liquibase-3-5-2-pulled.html) advising that the 3.5.2 has been pulled and users use 3.5.1.

### DIFF
--- a/Formula/liquibase.rb
+++ b/Formula/liquibase.rb
@@ -1,8 +1,8 @@
 class Liquibase < Formula
   desc "Library for database change tracking"
   homepage "http://liquibase.org"
-  url "https://github.com/liquibase/liquibase/releases/download/liquibase-parent-3.5.2/liquibase-3.5.2-bin.tar.gz"
-  sha256 "15b0284e29a2661440d1c577478ec18eda2bafd3f90c583ce0a200ec3971beba"
+  url "https://github.com/liquibase/liquibase/releases/download/liquibase-parent-3.5.1/liquibase-3.5.1-bin.tar.gz"
+  sha256 "b99870efec8c779b4e279a8b0c730273f837d4b6dffd037176737ecc0641ab6b"
 
   bottle :unneeded
 


DomT4:
Any idea how bad the upstream issue for 3.5.2 is? Users won't get downgraded to 3.5.1 automatically. We could force it, but it's a pretty drastic step that'll stick with the formula forever if we do.

jessedearing:
@DomT4 I'm not sure how bad the issue is. I figured if it has been pulled then at least I would like to make sure new downloads are 3.5.1.

@nvoxland Can you shed some light on the severity of the issues that caused the need to pull the 3.5.2 release? Is it worth forcing a downgrade?

DomT4:
I'll take the URL/SHA256 fix for now so people trying to `brew install` today don't hit a 404, but am interested on the severity here still.

Merged in https://github.com/Homebrew/homebrew-core/commit/98912913f7785fccf6cfdf43d5f2f149b9351f9a. Thank you for your contribution to Homebrew @jessedearing! 😺

nvoxland:
The 3.5.2 bug turned out to be a false alarm so you'd be OK using 3.5.2 if you'd like. I'll push a new version out as 3.5.3 that is basically the same as 3.5.2 in the next day or two.

DomT4:
3.5.3 sounds good. Users will get upgraded to that once merged into Homebrew whether they've installed the 3.5.2 previously or the 3.5.1 just merged back in, so works either way.

PR #5399, opened 2016-09-30
pythons: fix sqlite module without CLT

### LINKED ISSUE #5347: Python problem on 10.12 systems without CLT
Multiple users have reported Python failing to `import _sqlite3` on 10.12, and reported success after installing CLT. `python.rb` also prevents users from pouring the bottle when CLT is not installed. (Homebrew/brew#1171, Homebrew/brew#1179, and https://github.com/Homebrew/homebrew-core/issues/1957#issuecomment-248053191 which I just remembered.) Unfortunately existing CLT is overwritten by macOS upgrade and people upgrade from El Capitan may not realize.

I wonder what's the best action we could take to prevent more users from wasting time on Python due to not having CLT.

tdsmith:
Closes #5347.

### DIFF
--- a/Formula/python.rb
+++ b/Formula/python.rb
@@ -158,10 +158,16 @@ class Python < Formula
       args << "--enable-universalsdk=/" << "--with-universal-archs=intel"
     end
 
-    # Allow sqlite3 module to load extensions:
-    # https://docs.python.org/library/sqlite3.html#f1
     if build.with? "sqlite"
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Nice work. Worth going upstream?

tdsmith:
Ideally but it would be a project. Doubt it would get picked up in the 2.7 branch, too.

MikeMcQuaid:
Ok, no worries. Filing an issue would probably be useful for both/either.

PR #5406, opened 2016-09-30
boost@1.61: rename from boost

ilovezfs:


MikeMcQuaid:
Let's see what happens 💥 🔥

ilovezfs:
@MikeMcQuaid thoughts on :boat:ing?

MikeMcQuaid:
@ilovezfs Gonna look at this today but I reckon we do whatever in core is require to not rerevision this e.g. https://github.com/Homebrew/brew/pull/1192 first.

ilovezfs:
@MikeMcQuaid Homebrew/brew#1192 isn't necessary if the first step in the procedure is creating boost@1.62 alongside boost, not renaming boost to boost@1.61.

MikeMcQuaid:
@ilovezfs I need to write up the steps better but I think this PRs rename should include an alias which in turn allows a pointer to `opt/boost`.

ilovezfs:
Boost versioning can wait until the version procedures are honed further.

PR #5416, opened 2016-09-30
glib-networking 2.50.0 and libsoup 2.56.0

tschoonj:
Enabled `p11-kit support`, which also required a modification to the `gnutls` formula.
`libsoup` depends on `glib-networking`, so I added that one too

### DIFF
--- a/Formula/glib-networking.rb
+++ b/Formula/glib-networking.rb
@@ -1,8 +1,8 @@
 class GlibNetworking < Formula
   desc "Network related modules for glib"
   homepage "https://launchpad.net/glib-networking"
-  url "https://download.gnome.org/sources/glib-networking/2.48/glib-networking-2.48.2.tar.xz"
-  sha256 "925c0c49d6b2b8b5695f2e33cd952d1dbb7d18d3f2f796413577719315bb3a84"
+  url "https://download.gnome.org/sources/glib-networking/2.50/glib-networking-2.50.0.tar.xz"
+  sha256 "3f1a442f3c2a734946983532ce59ed49120319fdb10c938447c373d5e5286bee"
 
   bottle do
     sha256 "8d9f25c6ad40d0461faf185e2d82a6b54d2b8df4ef485ebdc6f00129f9302c71" => :sierra
--- a/Formula/libsoup.rb
+++ b/Formula/libsoup.rb
@@ -1,8 +1,8 @@
 class Libsoup < Formula
   desc "HTTP client/server library for GNOME"
   homepage "https://live.gnome.org/LibSoup"
-  url "https://download.gnome.org/sources/libsoup/2.54/libsoup-2.54.1.tar.xz"
-  sha256 "47b42c232034734d66e5f093025843a5d8cc4b2357c011085a2fd04ef02dd633"
+  url "https://download.gnome.org/sources/libsoup/2.56/libsoup-2.56.0.tar.xz"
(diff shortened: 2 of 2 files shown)


tschoonj:
Looks like Jenkins is having a bad moment

Formula/gnutls.rb
```diff
@@ -20,6 +21,7 @@ class Gnutls < Formula
   depends_on "nettle"
   depends_on "guile" => :optional
   depends_on "unbound" => :optional
+  depends_on "p11-kit"
```
DomT4: Given we've only _just_ restored this from the boneyard, we might want to make this `:optional` for now.

tschoonj: In both `glib-networking` and `gnutls`?

DomT4: Aye, for now. We can review the situation later but if we introduce this by default and then end up removing `p11-kit` down the line because we discover further issues we'll have to force a rebuild of a wider group of formulae, which is a pain.

tschoonj: ok, fair enough

ilovezfs:
looks like there are some xcode 8 + gnutls issues to sort out

tschoonj:
Well I had to downgrade to Xcode 7.3 on my El Cap to get this working (`undefined reference to _getentropy`).

How do I fix this exactly, since it looks like you guys are using Xcode 8 on the El Cap bot?

DomT4:
Blocked on resolution of https://github.com/Homebrew/homebrew-core/pull/5434 and/or https://github.com/Homebrew/homebrew-core/commit/46be01e503df1e30dba395bffd87d2619e117748.

tschoonj:
@ilovezfs What are your thoughts on merging this one in?

ilovezfs:
Assuming the option works the PR looks benign.

RandomDSdevel:
@ilovezfs, @tschoonj:  The formulas edited by this PR might need an/another `revision` bump after I finally get around to filing another PR to resolve my comments regarding [commit `46be01e`](https://github.com/Homebrew/homebrew-core/commit/46be01e503df1e30dba395bffd87d2619e117748), though.  In the meantime, you guys will have to make this PR relative to the changes I committed in PR #5434.

ilovezfs:
@RandomDSdevel handled https://github.com/Homebrew/homebrew-core/pull/6174.

RandomDSdevel:
@ilovezfs:  Thanks for taking care of that for me!  I would have gotten around to it eventually, but I usually like to wait to make contributions until after I have everything managed by Homebrew on my system up to date so that I have a clean in-`brew` state to work with and have been held back from doing this by my current dependency on [[qt5] Qt v5.7.0 Fails to Build on OS X/macOS v10.11.6 'El Capitan'](https://github.com/Homebrew/homebrew-core/issues/6032).  I don't know if that should really be a concern of mine, but I thought it might be best if I made a conservative decision in that department.

ilovezfs:
@RandomDSdevel no worries. FYI: https://github.com/Homebrew/homebrew-core/pull/6175

RandomDSdevel:
@ilovezfs:  Thanks for the pointer!  I'll keep track of that, too, then.  Didn't remember that being an issue, but, as you might have guessed, I've been a bit frazzled lately, so who _knows_ whether, given my scatterbrained state, I might have recognized it as one or not?  😵😆.

tschoonj:
I removed all p11-kit related stuff from these commits this morning 😢 

didn't realize 0.23.2 would become stable 😄

tschoonj:
will merge this one in as is, feel free to add p11-kit support later

PR #5417, opened 2016-09-30
updated library from http://www.live555.com/liveMedia/public/ to lates…

michalsarna:
updated openrtsp formula with latest live library since last one used is no more available.


DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/71ab9db961eecf386067f74964c8bf997efac8a5. Thank you for your contribution to Homebrew @michalsarna; we appreciate it! :smiley_cat:

PR #5431, opened 2016-09-30
macvim: bump revision for Python 2.7.12_1

asbhat:

---

This has happened before (https://github.com/Homebrew/brew/issues/463)
Bumping revision so the embedded python framework path is updated. And a couple stylistic changes to appease `brew audit`

### DIFF
--- a/Formula/macvim.rb
+++ b/Formula/macvim.rb
@@ -5,6 +5,8 @@ class Macvim < Formula
   url "https://github.com/macvim-dev/macvim/archive/snapshot-110.tar.gz"
   version "8.0-110"
   sha256 "a3437a0edbe0d2229def37e342b746ce22028bd604a738f5ee0cc978c7996336"
+  revision 1
+
   head "https://github.com/macvim-dev/macvim.git"
 
   bottle do


Formula/macvim.rb
```diff
@@ -18,7 +20,7 @@ class Macvim < Formula
 
   deprecated_option "override-system-vim" => "with-override-system-vim"
 
-  depends_on :xcode => :build
+  depends_on xcode: :build
```
DomT4: Please don't change these for now. We're aware text editors nag to do so, but that's unintentional and we're working on it.

DomT4:
@tdsmith Any opinions on whether we can do this better so `macvim` and `vim` don't require `revision` bumps whenever Python gets one? Or even if it's actually a problem given we rewrite the linkage after install to point to `opt` rather than Cellar?

asbhat:
Removed the formatting changes

tdsmith:
The install_names oughta already be correct; I don't know what paths vim is remembering.

tdsmith:
Ideally #5562 will land first and then this should be rebuilt.

shadowhand:
Please merge this, it is causing any plugin that relies on Python to be inoperable, as seen in editorconfig/editorconfig-vim#70.

asbhat:
@shadowhand you can `brew rm vim && brew install -s vim` in the mean time. #5562 has failing tests so until that is fixed this probably won't be merged

shadowhand:
@asbhat yes that's what I did.

tdsmith:
@BrewTestBot test this please

tdsmith:
Should be fixed permanently. Thanks for your patience.

tdsmith:
No, looks like it was just fixed for now. :( Sorry, will have a better fix.

asbhat:
anything I can help with @tdsmith?

tdsmith:
I think I understand the failure mode now; Python's `sys.prefix` is being baked into the vim binary. My fix in #5562 wasn't enough because it didn't actually change `sys.prefix` (or `sys.execprefix`) and it should have. Simultaneously, I _thought_ it had worked because I tested with a patched version of vim!

PR #5075, opened 2016-09-21
shadowsocks-libev 2.5.3

ghost:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

PR #5458, opened 2016-10-01
Shadowsocks-libev: update 2.5.3 bottle.

gkl1368:


zmwangx:
Thanks, but this is a duplicate of #5075.

MikeMcQuaid:
@zmwangx This is 💚 and that is not.

zmwangx:
That's because we fixed clock stuff in superenv since then; the diff is exactly the same. I've requeued the other one. I'm okay with merging this instead, but that breaks the established rule of attribution.

zmwangx:
I pulled bottles from here but retained authorship of #5075, sorry!

PR #5459, opened 2016-10-01
glibmm 2.50.0 and libsigc++ 2.10.0

tschoonj:


MikeMcQuaid:
Thanks again @tschoonj!

PR #5468, opened 2016-10-01
jless: migrate to boneyard

zmwangx:


zmwangx:
http://www.greenwoodsoftware.com/less/ has converged with `homebrew/dupes/less`. Or maybe I'm just confused about the homepage, and in fact this `jless` fork doesn't have a homepage of its own (in which case it doesn't meet our acceptability criteria).

In any case, this should affect only a few users, and affected users could appeal the decision.

MikeMcQuaid:
This one seems OK but let's give at least 24h for boneyard PRs to sit.

zmwangx:
Okay.

Issue #3639, opened 2016-08-05
Build fail ACE (6.4.0)  MacOS 10.12 (16A270f)


Selenium:
### Bug reports:

Hello, 
Since the first beta os macOS Sierra the formula ACE 6.4.0 dosent build anymore.

`brew gist-logs <formula>`
https://gist.github.com/anonymous/9ec22f126918e6f93567d6ff9ebca1eb

mistydemeo:
We need to backport this commit: https://github.com/DOCGroup/ACE_TAO/commit/a2b3e76ab4407981816cfa903b9fc749b8c53886

However, even with that, there's a bug relating to a clock type: https://gist.github.com/0caa29006625fb87d6ff62c16096389a

```
clang++  -m64 -W -Wall -DACE_HAS_CUSTOM_EXPORT_MACROS=0  -DNDEBUG  -I/private/tmp/ace-20160810-
7106-hswow7/ACE_wrappers -DACE_NDEBUG -D__ACE_INLINE__ -I.. -DACE_BUILD_DLL  -c  -o .shobj/Name
_Proxy.o Name_Proxy.cpp
In file included from Name_Proxy.cpp:1:
In file included from /private/tmp/ace-20160810-7106-hswow7/ACE_wrappers/ace/Name_Proxy.h:21:
In file included from /private/tmp/ace-20160810-7106-hswow7/ACE_wrappers/ace/INET_Addr.h:15:
In file included from /private/tmp/ace-20160810-7106-hswow7/ACE_wrappers/ace/Sock_Connect.h:23:
In file included from /private/tmp/ace-20160810-7106-hswow7/ACE_wrappers/ace/Basic_Types.h:52:
In file included from /private/tmp/ace-20160810-7106-hswow7/ACE_wrappers/ace/os_include/os_limi
ts.h:25:
In file included from /private/tmp/ace-20160810-7106-hswow7/ACE_wrappers/ace/os_include/os_unis
td.h:25:
/private/tmp/ace-20160810-7106-hswow7/ACE_wrappers/ace/os_include/sys/os_types.h:49:16: error:
typedef redefinition with different types ('int' vs 'enum clockid_t')
   typedef int clockid_t;
               ^
/usr/include/time.h:171:3: note: previous definition is here
} clockid_t;
  ^
```

claudiofantacci:
Hello everyone, are there any news about this issue? :-)

jwillemsen:
[...]

PR #5476, opened 2016-10-01
Fixed Ace formula for macOS 10.12 SDK

francesco-romano:

---

See https://github.com/Homebrew/homebrew-core/issues/3639

This PR solves compilation issue on El Capitan with Xcode 8.0
This should also solve the same issue on macOS Sierra, but I am unable to test it.

This is a temporary workaround while waiting for a new release by ACE (https://github.com/DOCGroup/ACE_TAO) as they already fixed the problem upstream.

Unfortunately `brew audit` does not pass with the following output

```
$ brew audit ace
ace:
    Broken dylib links found:
      libExport_Lib.dylib
        libIPC_Tests_Server.dylib
        libTimer.dylib
        libtqtd.dylib
Error: 1 problem in 1 formula
```

which are built in the example phase of the formula.

Note that I had to add a build dependency on `openssl` as macOS 10.11 does not ship anymore the headers of the system openssl. Binary packages do not need this as the system version is taken.

### DIFF
--- a/Formula/ace.rb
+++ b/Formula/ace.rb
@@ -3,6 +3,7 @@ class Ace < Formula
   homepage "https://www.dre.vanderbilt.edu/~schmidt/ACE.html"
   url "http://download.dre.vanderbilt.edu/previous_versions/ACE-6.4.0.tar.bz2"
   sha256 "6a7fd5f23f6a004bb21978767df4e36adea078cd344e45bab50266213586c001"
+  revision 1
 
   bottle do
     cellar :any
@@ -10,6 +11,9 @@ class Ace < Formula
     sha256 "d3e7e45197f7d7643d675ff84057df769947f6c0a9185388d23b2a28375760f0" => :yosemite
     sha256 "72f21357797c350fe547a7978f402e564cf328c3662a0e321d1ec293f4899139" => :mavericks
   end
+
+  # this patch backports new files for macOS Sierra and possibility to compile with Xcode 8 on El Capitan
+  patch :DATA
 
   def install
     # Figure out the names of the header and makefile for this version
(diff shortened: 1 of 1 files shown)


Formula/ace.rb
```diff
@@ -11,6 +14,9 @@ class Ace < Formula
     sha256 "72f21357797c350fe547a7978f402e564cf328c3662a0e321d1ec293f4899139" => :mavericks
   end
 
+  # this patch backports new files for macOS Sierra and possibility to compile with Xcode 8 on El Capitan
+  patch :DATA
```
DomT4: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file? Thanks!

Formula/ace.rb
```diff
@@ -3,6 +3,9 @@ class Ace < Formula
   homepage "https://www.dre.vanderbilt.edu/~schmidt/ACE.html"
   url "http://download.dre.vanderbilt.edu/previous_versions/ACE-6.4.0.tar.bz2"
   sha256 "6a7fd5f23f6a004bb21978767df4e36adea078cd344e45bab50266213586c001"
+  revision 1
+
+  depends_on "openssl" => :build
```
DomT4: This is exceptionally rarely a `:build`-only dep. Are you sure?

Formula/ace.rb
francesco-romano: The problem is that ACE repository is not compilable because of their build system. We can only compile the release they do.
This unfortunately remove the possibility to add `HEAD` to the ACE formula.
This patch has already been merged into the master of ACE, but not in any released version.

Formula/ace.rb
francesco-romano: Well.. No, I'm not sure.
For what I understood, starting from macOS 10.11 Apple removed the headers of ssl from `/usr/include`. So, if a library/executable is compiled and linked against openssl, it will work anyway on macOS because of the presence of the lib in `/usr/lib/libssl.dylib`. 
But, if you have to compile an openssl-powered application, then we need to obtain the headers.

Now, I'm not more able to replicate the issue I had some days ago, so I'll delete this dependency.

francesco-romano:
It seems a new release (6.4.1) is scheduled in 1 to 2 weeks (see https://github.com/Homebrew/homebrew-core/issues/3639#issuecomment-250923676)

So, this PR will become useless in 2 weeks.
It is your choice to accept as a temporary workaround or not.

MikeMcQuaid:
Temporary workaround is 👍 for now. Thanks again for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

jwillemsen:
Added pull request #5752 for the new version, only no idea why it now fails on just one version

PR #5479, opened 2016-10-01
ffmpeg 3.1.4

zmwangx:

---

`openh264` vendor is here to stay. So is clock crap.

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -1,8 +1,8 @@
 class Ffmpeg < Formula
   desc "Play, record, convert, and stream audio and video"
   homepage "https://ffmpeg.org/"
-  url "https://ffmpeg.org/releases/ffmpeg-3.1.3.tar.bz2"
-  sha256 "58bc89c65dd114d874efbf76f76368d03b5e407f0a3f42d5b40801c280968a38"
+  url "https://ffmpeg.org/releases/ffmpeg-3.1.4.tar.bz2"
+  sha256 "7c99df75a4dc12d22c0f1ed11d0acf98cac1f8b5fe7a7434344b167f810bcbfa"
   head "https://github.com/FFmpeg/FFmpeg.git"
 
   bottle do


JCount:
@zmwangx It might be good to add in something roughly like:
```
depends_on "sdl2" if build.with? "ffplay" && build.head?
```
since the dependency has been switched to `sdl2`, with `sdl1` being completely removed from `head`(the `master` branch). This means that `--with-ffplay` is currently broken with the `head` spec.

If you would like, I can open a separate PR to fix this after this version bump gets merged.

zmwangx:
@JCount Thanks for the info. I don't use FFmpeg head, and head is usually pretty far from stable, so I wasn't aware of that.

I think it definitely makes sense to open a separate PR so that you could take credit for it.

zmwangx:
Btw, now that you mentioned I think `--with-openh264` might be broken in head too (AFAIK the master branch doesn't support openh264 1.5 or earlier, but I might be wrong). It would be great if that could be fixed in the same PR.

DomT4:
> If you would like, I can open a separate PR to fix this after this version bump gets merged.

Just a note, dependencies specific to one version (stable, devel or head) should go inside a `stable do` or `head do` block rather than using the `build.head?` statement on `depends_on` lines.

JCount:
Yeah, thank you for the reminder @DomT4. I figured it probably was alright to express it in an incorrect shorthand given the intended audience, but still, it is good to be consistent. Besides, `brew audit` catches this and gets mad if you actually try and use that syntax. `brew style` produces: ```Lint/RequireParentheses: Use parentheses in the method call to avoid confusion about precedence.``` 😸

DomT4:
> I figured it probably was alright to express it in an incorrect shorthand given the intended audience, but still, it is good to be consistent. 

Yeah. Basically presume anyone can read anything around here. We get some new contributors checking old PRs for guidance so these things can quickly be picked up on, even if not intended 🙈.

JCount:
🙀

PR #5483, opened 2016-10-01
oniguruma: Fix wrong uchar type in headers

felixbuenemann:
This backports two fixes from the kkos/oniguruma#develop branch that are
required to successfully compile hhvm from the hhvm/homebrew-hhvm tap.

The fixes are discussed in hhvm/homebrew-hhvm#50 and in the commit
comments of the broken code:
- kkos/oniguruma@c4be742
- kkos/oniguruma@f5fe5a3

The fixes are taken from:
- kkos/oniguruma@b38e7b0
- kkos/oniguruma@220d62a

The patch can be dropped when the next upstream version is released.


MikeMcQuaid:

jwatzman:
Thanks from the HHVM side as well :)

PR #5247, opened 2016-09-26
mobile-shell 1.2.6_3

yanokwa:
Fixes #5220 and updates homepage and url to new mosh website

Issue #5220, opened 2016-09-25
mobile-shell (mosh) fails with _clock_gettime error on 10.11.6 with Xcode 8


yanokwa:
mosh compiles and runs, but when I try to connect to a server with an SSH key, it fails with "Symbol not found: _clock_gettime". The binary from https://mosh.org/#getting or building from source does not have this problem.

```
$ brew config

HOMEBREW_VERSION: 1.0.4-7-g6e57355
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 6e5735565b9a31ec2b01e02a479a361189729c22
Last commit: 5 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 109939afe4af2dd8742cbc31af1bc2b0e04a6556
Core tap last commit: 27 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.10.0 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/local/var/pyenv/shims/python => /usr/local/var/pyenv/versions/2.7.12/bin/python2.7
Ruby: /usr/local/var/rbenv/shims/ruby => /usr/local/var/rbenv/versions/2.3.1/bin/ruby
Java: 1.8.0_66, 1.7.0_45
macOS: 10.11.6-x86_64
Xcode: 8.0
CLT: 7.3.1.0.1.1461711523
X11: 2.7.9 => /opt/X11
```

```
$ mosh --version

mosh 1.2.6 [build mosh 1.2.6]
Copyright 2012 Keith Winstein <mosh-devel@mit.edu>
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

```
$ mosh --ssh="ssh -i /Users/user/.ssh/id_rsa" root@1.2.3.4

dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/mobile-shell/1.2.6_2/bin/mosh-client
[...]

PR #5489, opened 2016-10-01
mobile-shell:  enable complete tests under 'make check'

cgull:

---

'make check' will use tmux to run complete tests if
available, including running the executables.

This likely would have caught #5220 and #5247.

This runs the tests at build time, rendering the smoke test in the `test` block irrelevant-- what's the proper way to handle this?
(We tried to improve that before in #3880)

### DIFF
--- a/Formula/mobile-shell.rb
+++ b/Formula/mobile-shell.rb
@@ -3,7 +3,7 @@ class MobileShell < Formula
   homepage "https://mosh.org"
   url "https://mosh.org/mosh-1.2.6.tar.gz"
   sha256 "7e82b7fbfcc698c70f5843bb960dadb8e7bd7ac1d4d2151c9d979372ea850e85"
-  revision 3
+  revision 4
 
   bottle do
     sha256 "bc1a1ce96af199e577ee7eecd75688f43aaa6bddec09a7973c487f8d4233e60f" => :sierra
@@ -25,6 +25,7 @@ class MobileShell < Formula
   depends_on "pkg-config" => :build
   depends_on "protobuf"
   depends_on :perl => "5.14" if MacOS.version <= :mountain_lion
+  depends_on "tmux" => :build if build.with?("test") || build.bottle?
 
   def install
     # Fix for 'dyld: lazy symbol binding failed: Symbol not found: _clock_gettime' issue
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@DomT4 thoughts on this?

Formula/mobile-shell.rb
```diff
@@ -3,7 +3,7 @@ class MobileShell < Formula
   homepage "https://mosh.org"
   url "https://mosh.org/mosh-1.2.6.tar.gz"
   sha256 "7e82b7fbfcc698c70f5843bb960dadb8e7bd7ac1d4d2151c9d979372ea850e85"
-  revision 3
+  revision 4
```
apjanke: Probably no need for a `revision` here. The installed binary will be the same, and that's the criteria for a `revision`.

Formula/mobile-shell.rb
```diff
@@ -25,6 +25,7 @@ class MobileShell < Formula
   depends_on "pkg-config" => :build
   depends_on "protobuf"
   depends_on :perl => "5.14" if MacOS.version <= :mountain_lion
+  depends_on "tmux" => :build if build.with?("test") || build.bottle?
```
DomT4: Don't need the parens or `|| build.bottle?` here.

DomT4:
> @DomT4 thoughts on this?

My only concern is that isn't `cellar :any`, so any users outside of `/usr/local` installing `mosh` in future are going to get a `tmux` installation dumped on them by default. There's an opt-out so not the end of the world. Overall, probably a 👍.

The existing test should remain though, smoke or otherwise they have their uses.

Formula/mobile-shell.rb
MikeMcQuaid: As it's using the test by default.

MikeMcQuaid (changes requested):

ilovezfs:
I don't see the need to run the test by default if we're running it during bottle. Is there some reason to believe it's going to fail for the source-build users if it passed when bottling?

dunn:
@ilovezfs do you mean we should change it to `depends_on "tmux" => :build if build.bottle?`

ilovezfs:
@dunn right, and changing without-test to with-test. That pattern is used in bup, gammaray, ghc, gnuplot, liblas, leveldb, python, and protobuf. mobile-shell is the only one that gets it wrong, as Mike pointed out, by using it when the test option is default on.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #5490, opened 2016-10-01
rust: doesn't use cargo to build cargo, so no need for extraneous resource

nomoon:

---

Originally part of #5194.

I also have a commit ready to go to add `devel`-level beta builds. There are usually around 6 per release cycle (6 weeks), so is this maybe too many intermediate formula updates to support?


zmwangx:
I suppose there was a reason for the cargo bootstrap, maybe faster builds?

Re devel: if you submit a PR I don't think we have a good reason to reject it, but rust builds are a nightmare for the CI already and adding devel would just double the testing time (yes, both stable *and* devel are tested), so I really want to see a sizeable demand for it...

DomT4:
> I suppose there was a reason for the cargo bootstrap, maybe faster builds?

Can you comment on this @alexcrichton, if you don't mind me pinging you again? 🙈. Is there a net gain to grabbing the already compiled `cargo` to build the new `cargo` with?

> rust builds are a nightmare for the CI already and adding devel would just double the testing time (yes, both stable and devel are tested), so I really want to see a sizeable demand for it...

Agree with this. It'll be easier for us soon but Rust builds are pretty heavy.

nomoon:
@DomT4 My workflow as I've taken up Rust has moved pretty decisively to rustup (for swapping toolchains to test things), and it's equally fast as homebrew bottles (since it distributes binaries whenever possible). I'm mostly just interested in having people having access to Rust in the easiest way possible so that they can make use of my things (Ruby gems with Rust-compiled native extensions). Definitely the beta channel is a stretch in that regard. Most people will just use the stable bottle, and that's great.

As far as I can tell, the cargo build time on my 2014 MBP was not noticeably different using the pre-built nightly that this PR removes. But it's depending on downloading more binaries to build sources, which seems to be sort-of counter to going through this whole build-from-source process anyway. Hence the PR.

alexcrichton:
@DomT4 

Sure! Feel free to ping me any time about Rust-y things!

> Is there a net gain to grabbing the already compiled cargo to build the new cargo with?

I think that in the context of Homebrew there shouldn't be any need to pull a custom already-compiled Cargo. Cargo's build system (via `./configure`) already does this by default, so you can probably just let Cargo handle that (like rustc handles downloading an older rustc)

If that network traffic isn't acceptable, though, then you'll still want to pull down a Cargo by hand.

nomoon:
@alexcrichton Is it at all likely that the `cargo-nightly-2015-09-17` bootstrap will become unable to compile cargo in the future? Because if that's the case, it'd be an especially good idea to let cargo handle the preferred bootstrap itself.

alexcrichton:
Yeah at some point Cargo will likely require some feature of a previous Cargo to compile itself, at which point that older Cargo will be unable to compile Cargo.

ilovezfs:
>Because if that's the case, it'd be an especially good idea to let cargo handle the preferred bootstrap itself.

Not sure I'm following this point. Wouldn't we just bump the resource's version?

MikeMcQuaid:
> If that network traffic isn't acceptable, though, then you'll still want to pull down a Cargo by hand.

This is probably why we did this. If the Cargo buildsystem is doing a download and strong (SHA256 or better) checksum validation afterwards I'm fine with dropping it.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

yurikoles:
It seems like this patch doesn't work with rust-1.14.

nomoon:
@yurikoles I just updated the Rust and Cargo sources in the formula to current stable ones.

ilovezfs:
This PR fails in the same way as https://github.com/Homebrew/homebrew-core/pull/8138.

https://gist.github.com/ilovezfs/6fe530d7a8e5fad8393a83541b90ce68

```
```

PR #5509, opened 2016-10-02
bullshtml 1.0.6 (new formula)

gjasny:
Bullshtml is an utility for the BullseyeCoverage C/C++ code
coverage analyzer. It renders much nicer HTML pages than the
BullseyeCoverage tool and also writes Clover compatible summaries
that can be picked up by CI tools like Atlassian Bamboo.


MikeMcQuaid:
I'm afraid this is not within our popularity guidelines so passing on it for now. Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

PR #5510, opened 2016-10-02
Added depends_on gmp to crystal-lang

DougEverly:
Adding gmp fixes this error when running crystal 0.19.3:

```
$ ./bin/crystal spec ./spec/std/json/**
ld: library not found for -lgmp
```

### DIFF
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -19,6 +19,7 @@ class CrystalLang < Formula
   depends_on "bdw-gc"
   depends_on "llvm"
   depends_on "pcre"
+  depends_on "gmp"
   depends_on "libyaml" if build.with? "shards"
 
   resource "boot" do


asterite:
@DougEverly Note that gmp, xml and yaml are all optional: not all programs use them, and if you need them you can just `brew install gmp`. But including all those libraries by default might be good too and will bring less frustration. Maybe you can change your PR to include all such libraries? (`gmp` and `libxml2` as far as I remember)

DougEverly:
Added `libxml2`. `yaml` too? Perhaps _all_ standard library dependencies should be included by default. This can be confusing to new (and current!) developers.

asterite:
@DougEverly I think those are the only dependencies for now :-)

ilovezfs:
libxml2 is provided by the operating system already ...

ilovezfs:
What's the status here?

asterite:
`libxml2` should be removed, because it does seem to be provided by the OS. So only adding `gmp` should be OK.

DougEverly:
I removed libxml2.

MikeMcQuaid:

For future reference the preferred commit message format is `crystal-lang: depend on gmp.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #5515, opened 2016-10-03
awscli: don't link deps binaries

ilovezfs:


tdsmith (approved):
thanks!

DomT4:
❤️ if we can get this setup written into the `brew` side of things rather than spreading it throughout formulae.

PR #5540, opened 2016-10-03
ffmpeg: fix head specific deps, refactor code into seperate specs

JCount:

---

`HEAD` has removed `sdl 1` support due to its EOL status, now requiring `sdl2` in order to build ffplay. Also, `HEAD` can now be built against brewed `openh264` (1.6.0), meaning it no longer has to rely on the vendor of `openh264 1.5.0`.

I tried to handle these two points of divergence between `stable` and `head` by creating minimal `'spec' do` blocks, leaving by far the bulk of the of the option and dependency declarations as shared code. I tested building both with `--with-ffplay --with-openh264` and both built without errors. However, I do not know if this is proper usage of the Hombrew DSL, or even the proper approach to try and achieve this dependency(/option) separation.

I am happy to do whatever revisions are necessary to make this work, just point me where I need to go.

Also, I did not attach a revision number because I honestly have no clue whether I should in this instance, so I decided to hold off and wait for feedback.

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -1,14 +1,42 @@
 class Ffmpeg < Formula
   desc "Play, record, convert, and stream audio and video"
   homepage "https://ffmpeg.org/"
-  url "https://ffmpeg.org/releases/ffmpeg-3.1.4.tar.bz2"
-  sha256 "7c99df75a4dc12d22c0f1ed11d0acf98cac1f8b5fe7a7434344b167f810bcbfa"
-  head "https://github.com/FFmpeg/FFmpeg.git"
+
+  stable do
+    url "https://ffmpeg.org/releases/ffmpeg-3.1.4.tar.bz2"
+    sha256 "7c99df75a4dc12d22c0f1ed11d0acf98cac1f8b5fe7a7434344b167f810bcbfa"
+
+    option "with-sdl", "Enable FFplay media player"
+    option "with-openh264", "Enable OpenH264 library"
+    deprecated_option "with-ffplay" => "with-sdl"
+
+    depends_on "sdl" => :optional
+    depends_on "nasm" => :build if build.with? "openh264"
+
+    # Remove when ffmpeg has support for openh264 1.6.0
+    # See https://github.com/cisco/openh264/issues/2505
+    # Master now has support, but not the 3.1.x branch
+    resource "openh264-1.5.0" do
+      url "https://github.com/cisco/openh264/archive/v1.5.0.tar.gz"
+      sha256 "98077bd5d113c183ce02b678733b0cada2cf36750370579534c4d70f0b6c27b5"
+    end
+  end
 
   bottle do
     sha256 "72394f281e42f49816f19a1fe842963ce710be992d9cdafb0b147782ea0a1af1" => :sierra
     sha256 "7c9200710f18df67ddb475436fc312ee4a6c2b0ab6f758c373c28873f84cc89b" => :el_capitan
     sha256 "cc0b73f7ba46f677e4e93e4c274c701895a28e14e97fa335dfea280285ad8df3" => :yosemite
+  end
+
+  head do
+    url "https://github.com/FFmpeg/FFmpeg.git"
+
+    # Support for SDL1 has been removed from master
+    option "with-sdl2", "Enable FFplay media player"
+    deprecated_option "with-ffplay" => "with-sdl2"
+
+    depends_on "sdl2" => :optional
+    depends_on "openh264" => :optional
   end
 
   option "without-x264", "Disable H.264 encoder"
@@ -23,7 +51,6 @@ class Ffmpeg < Formula
   option "with-openssl", "Enable SSL support"
   option "with-libssh", "Enable SFTP protocol via libssh"
   option "with-schroedinger", "Enable Dirac video format"
-  option "with-ffplay", "Enable FFplay media player"
   option "with-tools", "Enable additional FFmpeg tools"
   option "with-fdk-aac", "Enable the Fraunhofer FDK AAC library"
   option "with-libvidstab", "Enable vid.stab support for video stabilization"
@@ -34,7 +61,6 @@ class Ffmpeg < Formula
   option "with-snappy", "Enable Snappy library"
   option "with-rubberband", "Enable rubberband library"
   option "with-zimg", "Enable z.lib zimg library"
-  option "with-openh264", "Enable OpenH264 library"
   option "with-xz", "Enable decoding of LZMA-compressed TIFF files"
   option "with-libebur128", "Enable using libebur128 for EBU R128 loudness measurement"
 
@@ -58,7 +84,6 @@ class Ffmpeg < Formula
   depends_on "opencore-amr" => :optional
   depends_on "libass" => :optional
   depends_on "openjpeg" => :optional
-  depends_on "sdl" if build.with? "ffplay"
   depends_on "snappy" => :optional
   depends_on "speex" => :optional
   depends_on "schroedinger" => :optional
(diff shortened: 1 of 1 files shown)


Formula/ffmpeg.rb
```diff
@@ -1,16 +1,43 @@
...
-  head "https://github.com/FFmpeg/FFmpeg.git"
+
+  stable do
+    url "https://ffmpeg.org/releases/ffmpeg-3.1.4.tar.bz2"
+    sha256 "7c99df75a4dc12d22c0f1ed11d0acf98cac1f8b5fe7a7434344b167f810bcbfa"
+
+    option "with-ffplay", "Enable FFplay media player"
+    option "with-openh264", "Enable OpenH264 library (v1.5.0)"
+
+    depends_on "sdl" if build.with? "ffplay"
```
retokromer: `depends_on "sdl2" if build.with? "ffplay"`
as support for SDL1 has been removed

Formula/ffmpeg.rb
```diff
@@ -1,16 +1,43 @@
...
-  url "https://ffmpeg.org/releases/ffmpeg-3.1.4.tar.bz2"
-  sha256 "7c99df75a4dc12d22c0f1ed11d0acf98cac1f8b5fe7a7434344b167f810bcbfa"
-  head "https://github.com/FFmpeg/FFmpeg.git"
+
+  stable do
+    url "https://ffmpeg.org/releases/ffmpeg-3.1.4.tar.bz2"
+    sha256 "7c99df75a4dc12d22c0f1ed11d0acf98cac1f8b5fe7a7434344b167f810bcbfa"
+
+    option "with-ffplay", "Enable FFplay media player"
+    option "with-openh264", "Enable OpenH264 library (v1.5.0)"
```
MikeMcQuaid: Why are these duplicated, out of interest?

Formula/ffmpeg.rb
```diff
@@ -1,16 +1,43 @@
...
   end
 
+  head do
+    url "https://github.com/FFmpeg/FFmpeg.git"
+
+    option "with-ffplay", "Enable FFplay media player"
+    option "with-openh264", "Enable OpenH264 library"
+
+    # Support for SDL1 has been removed from master
+    depends_on "sdl2" if build.with? "ffplay"
```
MikeMcQuaid: Ideally use a `deprecated_option` and make this:
```
  deprecated_option "with-ffplay" => "with-sdl2"
  depends_on "sdl2" => :optional
  option "with-sdl2", "Enable FFplay media player"
```

Formula/ffmpeg.rb
MikeMcQuaid: Ideally use a `deprecated_option` and make this:
```
  deprecated_option "with-ffplay" => "with-sdl"
  depends_on "sdl" => :optional
  option "with-sdl", "Enable FFplay media player"
```

Formula/ffmpeg.rb
```diff
@@ -110,11 +125,13 @@ def install
     ]
 
     if build.with? "openh264"
-      resource("openh264-1.5.0").stage do
-        system "make", "install-shared", "PREFIX=#{libexec}/openh264-1.5.0"
-        chmod 0444, libexec/"openh264-1.5.0/lib/libopenh264.dylib"
+      if build.stable?
+        resource("openh264-1.5.0").stage do
+          system "make", "install-shared", "PREFIX=#{libexec}/openh264-1.5.0"
+          chmod 0444, libexec/"openh264-1.5.0/lib/libopenh264.dylib"
```
MikeMcQuaid: Why do you need to `chmod` this?

Formula/ffmpeg.rb
JCount: I could not seem to get the build options to be in scope within the blocks any other way, which I needed to be the case to correctly resolve the dependencies. I may be doing something wrong here, but the only solution I could come up with was this duplication.

Formula/ffmpeg.rb
JCount: This is referring to the `stable` branch, in this case `v3.1.4`, and it is in FFmpeg `version <next>` that had support for `SDL2` added and for `SDL1` removed. (which is the `head` version of this formula)

Formula/ffmpeg.rb
JCount: I honestly do not know, this is a vendor that I believe @ilovezfs did, but I just left it intact to avoid breaking something I did not understand. I can look into it, but as I remember it was pulled from the `openh264` formula's install block.

Formula/ffmpeg.rb
JCount: I thought it might be appropriate to do this, since it gets rid of the less than ideal `if` statements that could cause issues. In fact, I will make these changes and see if it perhaps fixes this issues I was running into, seemingly due to scope (or something akin to it).

Formula/ffmpeg.rb
retokromer: I had problems building on Ubuntu without the change. I will report this for Linuxbrew.
:+1: for fixing this in 3.1.5 and not «only» in the imminent 3.2.0.

Formula/ffmpeg.rb
```diff
@@ -1,16 +1,44 @@
...
     sha256 "72394f281e42f49816f19a1fe842963ce710be992d9cdafb0b147782ea0a1af1" => :sierra
     sha256 "7c9200710f18df67ddb475436fc312ee4a6c2b0ab6f758c373c28873f84cc89b" => :el_capitan
     sha256 "cc0b73f7ba46f677e4e93e4c274c701895a28e14e97fa335dfea280285ad8df3" => :yosemite
   end
 
+  head do
+    url "https://github.com/FFmpeg/FFmpeg.git"
+
+    # Support for SDL1 has been removed from master
+    option "with-sdl2", "Enable FFplay media player"
```
JCount: The `--with-sdl2` option is not as visible from the CLI as I might like, it seems that because of the way [info.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/info.rb#L40) parses args to allow for its three different command forms, passing `--HEAD` when running `brew info <formula>` currently has no effect on the display of `head` specific options, dependencies, requirements, etc. However, it is still shown when `brew options ffmpeg --HEAD` is run, and in the formula itself, so perhaps that is enough.

On a side note, I wonder if there is any merit in improving the `info` command's ability to take formula specific modifiers. It may be best to maintain `options` specialization there to avoid cumbersome code.... hmm...

Formula/ffmpeg.rb
MikeMcQuaid: I think you can remove it unless it breaks things locally.

Formula/ffmpeg.rb
MikeMcQuaid: Can you let me know what you did to verify? I may be able to help. Thanks!

Formula/ffmpeg.rb
MikeMcQuaid: > On a side note, I wonder if there is any merit in improving the info command's ability to take formula specific modifiers.

Yep, definitely.

JCount: Should I open an issue in `Homebrew/brew` suggesting this, or something somewhere else? I was looking at the code in Homebrew's library so I have a sense of what `cmd/options.rb` is doing differently than `cmd/info.rb`, but I am not certain I could create a PR to directly address it. However, I could try putting something together that could be either transformed through revisions or be used as a base for someone more knowledgeable to create a better PR.

Any specific thoughts about what I should (or shouldn't) do to pursue this @MikeMcQuaid?

Formula/ffmpeg.rb
JCount: Will do.

Formula/ffmpeg.rb
MikeMcQuaid: A issue or ideally PR (we can help you through that) for this would be great.

MikeMcQuaid:

JCount:
I'll get to work on putting together a PR for `homebrew/brew`. Thanks for your help with this! 😄

Issue #5339, opened 2016-09-28
shibboleth-sp no longer compiles on Sierra


FranklinChen:
If Homebrew was updated on Aug 10-11th 2016 and `brew update` always says `Already up-to-date.` you need to run: `cd "$(brew --repo)" && git fetch && git reset --hard origin/master && brew update`.
# Please follow the general troubleshooting steps first:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

_You can erase any parts of this template not applicable to your Issue._
### Bug reports:

shibboleth-sp no longer compiles on Sierra.

https://gist.github.com/19b239ddd3e5853140c463419a9779cc

`brew gist-logs <formula>`
(where `<formula>` is the name of the formula that failed to build).
### Formula Requests:

**Please note by far the quickest way to get a new formula into Homebrew is to file a [Pull Request](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md).**

We will consider your request but it may be closed if it's something we're not actively planning to work on.

FranklinChen:
Salient part of the error:

```
checking for apr-1-config... apxs:Error: /Applications/Xcode.app/Contents/Developer/Toolchains/OSX10.12.xctoolchain/usr/local/bin/apr-1-config not found!.
/usr/local/Homebrew/Library/Homebrew/shims/super/apr-1-config

Failed to execute apr-1-config --cflags

Xcode and/or the CLT appear to be misconfigured. Try one or both of the following:
  xcodebuild -license
  sudo xcode-select -switch /path/to/Xcode.app


Failed to execute apr-1-config --cppflags

Xcode and/or the CLT appear to be misconfigured. Try one or both of the following:
  xcodebuild -license
  sudo xcode-select -switch /path/to/Xcode.app

checking for apu-1-config... apxs:Error: /Applications/Xcode.app/Contents/Developer/Toolchains/OSX10.12.xctoolchain/usr/local/bin/apr-1-config not found!.
no
configure: error: Unable to locate apu-1-config, may need --with-apu1 option.
```

dunn:
Apple removed `apr-1-config` from Xcode 8.  Could you try adding `depends_on "apr"` to the formula, and see if that fixes it? Thanks!

fxcoudert:
This seems to have been fixed, closing.

PR #5546, opened 2016-10-03
shibboleth: depend on apr, apr-util for Sierra

FranklinChen:

---

Resolve https://github.com/Homebrew/homebrew-core/issues/5339

### DIFF
--- a/Formula/shibboleth-sp.rb
+++ b/Formula/shibboleth-sp.rb
@@ -11,7 +11,7 @@ class ShibbolethSp < Formula
 
   option "with-apache-22", "Build mod_shib_22.so instead of mod_shib_24.so"
 
-  depends_on macos: :yosemite
+  depends_on :macos => :yosemite
   depends_on "curl" => "with-openssl"
   depends_on "opensaml"
   depends_on "xml-tooling-c"
@@ -21,10 +21,8 @@ class ShibbolethSp < Formula
   depends_on "boost"
   depends_on "unixodbc"
 
-  if MacOS.version >= :sierra
-    depends_on "apr-util" => :build
-    depends_on "apr" => :build
-  end
+  depends_on "apr-util" => :build
+  depends_on "apr" => :build
 
   def install
     ENV.O2 # Os breaks the build
@@ -60,7 +58,7 @@ class ShibbolethSp < Formula
     EOS
   end
 
-  plist_options startup: true, manual: "shibd"
+  plist_options :startup => true, :manual => "shibd"
 
   def plist; <<-EOS.undent
     <?xml version="1.0" encoding="UTF-8"?>
--- a/Formula/shibboleth-sp.rb
+++ b/Formula/shibboleth-sp.rb
@@ -11,7 +11,9 @@ class ShibbolethSp < Formula
 
   option "with-apache-22", "Build mod_shib_22.so instead of mod_shib_24.so"
 
-  depends_on :macos => :yosemite
+  depends_on macos: :yosemite
+  depends_on "apr"
+  depends_on "apr-util"
   depends_on "curl" => "with-openssl"
   depends_on "opensaml"
   depends_on "xml-tooling-c"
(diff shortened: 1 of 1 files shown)


ilovezfs:
This is probably only needed at build time (`=> :build`). Also, it should probably be scoped to the systems that actually need it (e.g., `DevelopmentTools.clang_build_version >= 800).

FranklinChen:
I copied what `passenger` does for build scoping.

Formula/shibboleth-sp.rb
```diff
@@ -11,7 +11,7 @@ class ShibbolethSp < Formula
 
   option "with-apache-22", "Build mod_shib_22.so instead of mod_shib_24.so"
 
-  depends_on :macos => :yosemite
+  depends_on macos: :yosemite
```
MikeMcQuaid: Keep this as it was. Apologies that this has changed recently.

Formula/shibboleth-sp.rb
```diff
@@ -43,7 +48,19 @@ def install
...
-  plist_options :startup => true, :manual => "shibd"
+  def caveats
+    mod = (build.with? "apache-22")? "mod_shib_22.so" : "mod_shib_24.so"
+    <<-EOS.undent
+      You must manually edit httpd.conf to include
+      LoadModule mod_shib #{lib}/shibboleth/#{mod}
+      You must also manually configure
+        #{etc}/shibboleth/shibboleth2.xml
+      as per your own requirements. For more information please see
+        https://wiki.shibboleth.net/confluence/display/EDS10/3.1+Configuring+the+Service+Provider
```
MikeMcQuaid: Let's leave these as they contain Homebrew-specific paths but the instructions aren't Homebrew-only.

Formula/shibboleth-sp.rb
```diff
@@ -43,7 +48,19 @@ def install
     system "make", "install"
   end
 
-  plist_options :startup => true, :manual => "shibd"
```
MikeMcQuaid: Leave this as-is.

MikeMcQuaid (changes requested):

DomT4:
> This is probably only needed at build time (=> :build). Also, it should probably be scoped to the systems that actually need it (e.g., DevelopmentTools.clang_build_version >= 800).

It won't be build-time only. That's only really the case when using the system version of `apr`, and even then it's a hack around our dependency system more than anything else. Most stuff does dynamically link to the Apr library. Someone might just want to change the requirement in `brew` now this is expanding beyond `subversion` as a problem.

`DevelopmentTools.clang_build_version >= 800` isn't strictly true IIRC; they'll still be available via the 7.x CLT on El Cap, so even if you've Xcode 8 on 10.11 as long as the CLT is installed you'll be fine. Off the top of my head though, so feel free to check.

ilovezfs:
`DevelopmentTools.clang_build_version >= 800` is what's reported if both Xcode 8 and CLT 7 are installed if that's what you're asking.

DomT4:
Anything `:` there is fine as `:build`, kind of, that's a hack around our dep system IIRC. Anything `""` should be checked, but I'd be very surprised if they were build unless those formulae only produce static libraries/executables.

ilovezfs:
@DomT4 LOL https://github.com/Homebrew/homebrew-core/issues/5599

DomT4:
Yeah. Probably worth giving up doing this individually and just enforce it for Sierra via the requirement.

MikeMcQuaid:
Blocked on https://github.com/Homebrew/brew/pull/1227.

Formula/shibboleth-sp.rb
```diff
@@ -21,6 +21,11 @@ class ShibbolethSp < Formula
   depends_on "boost"
   depends_on "unixodbc"
 
+  if MacOS.version >= :sierra
+    depends_on "apr-util" => :build
+    depends_on "apr" => :build
```
MikeMcQuaid: Make these unconditional.

MikeMcQuaid: Also, are you sure they aren't linked against?

MikeMcQuaid (changes requested):

MikeMcQuaid:
@FranklinChen Any news on the comments above? Thanks!

FranklinChen:
I'd been busy, will address this.

FranklinChen:
I updated according to comments here.

Formula/shibboleth-sp.rb
MikeMcQuaid: Actually, they were there before so it's fine.

MikeMcQuaid (approved):

MikeMcQuaid:

FranklinChen:
Is this going to be merged?

MikeMcQuaid:
It was.

PR #5564, opened 2016-10-04
glib: test gio png sniffing

ilovezfs:
See #5482.


MikeMcQuaid:
👃

Formula/glib.rb
```diff
@@ -124,23 +124,42 @@ def post_install
...
-    (testpath/"test.c").write <<-EOS.undent
-      #include <string.h>
-      #include <glib.h>
-
-      int main(void)
+    (testpath/"png.c").write <<-'EOS'.undent
+      #include <gio/gio.h>
+      #include <stdio.h>
+      static const gsize data_size = 159;
+      static const guint8 data[] =
```
MikeMcQuaid: Could this be read from our fixture? If not: no worries.

ilovezfs:
I don't intend to merge this. This is the test program from the configure script for gdk-pixbuf, verbatim. I'm trying to confirm that there is indeed a discrepancy between what happens on Jenkins versus what the OP in #5482 and I are seeing locally.

MikeMcQuaid:
Gotcha.

ilovezfs:
@MikeMcQuaid unfortunately, the discrepancy is real: https://github.com/Homebrew/homebrew-core/issues/5482#issuecomment-251325853

:(

totally weird.

ilovezfs:
Underlying problem resolved by #5581 and #5582.

PR #5584, opened 2016-10-04
gnome-themes-standard 3.22.1

tschoonj:


DomT4:
Merged in https://github.com/Homebrew/homebrew-core/commit/1a830908c859c3a15a9aa5b0f475fe8488b633ea. Thanks again Tom! Almost called you Dom there, but that would be me 😅.

PR #5611, opened 2016-10-05
fontforge 20161004

jtanx:

---

This updates the formula to remove a number of hacks that are no longer necessary. I've also switched it to use the dist tarball, which no longer requires `./bootstrap`/autotools needed in every build, only HEAD.

I've also added back the optional `--with-x11` option to optionally build the GUI (this is very handy when testing HEAD builds).

### DIFF
--- a/Formula/fontforge.rb
+++ b/Formula/fontforge.rb
@@ -1,9 +1,8 @@
 class Fontforge < Formula
   desc "Command-line outline and bitmap font editor/converter"
   homepage "https://fontforge.github.io"
-  url "https://github.com/fontforge/fontforge/archive/20161001.tar.gz"
-  sha256 "103af2a6c8799390f20790e44064ea4f9ec6795255b7a065b3f9a352c2723c40"
-  head "https://github.com/fontforge/fontforge.git"
+  url "https://github.com/fontforge/fontforge/releases/download/20161005/fontforge-dist-20161004.tar.gz"
+  sha256 "ccba2d84cf009e2a51656af4e0191c6a95aa1a63e5dfdeb7423969c889a24a64"
 
   bottle do
     sha256 "aeeb0031067149f9795f9aac5bf3b623ab53557345509086de1124cd941b0ad8" => :sierra
@@ -11,33 +10,43 @@ class Fontforge < Formula
     sha256 "e5e8e6a5522841bccf8dc4fbfd4953b35e5375320785474ddd0dc7aa066b1fcb" => :yosemite
   end
 
+  head do
+    url "https://github.com/fontforge/fontforge.git"
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+
+    resource "gnulib" do
+      url "git://git.savannah.gnu.org/gnulib.git",
+        :revision => "29ea6d6fe2a699a32edbe29f44fe72e0c253fcee"
+    end
+  end
+
   option "with-giflib", "Build with GIF support"
   option "with-extra-tools", "Build with additional font tools"
 
   deprecated_option "with-gif" => "with-giflib"
 
-  # Autotools are required to build from source in all releases.
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
+  depends_on :x11 => :optional
+  if build.with? "x11"
+    depends_on "cairo" => "with-x11"
+    depends_on "pango" => "with-x11"
+  else
+    depends_on "cairo"
+    depends_on "pango"
+  end
   depends_on "pkg-config" => :build
   depends_on "libtool" => :run
   depends_on "gettext"
-  depends_on "pango"
   depends_on "zeromq"
   depends_on "czmq"
-  depends_on "cairo"
   depends_on "fontconfig"
   depends_on "libpng" => :recommended
   depends_on "jpeg" => :recommended
   depends_on "libtiff" => :recommended
+  depends_on "libspiro" => :recommended
+  depends_on "libuninameslist" => :recommended
   depends_on "giflib" => :optional
-  depends_on "libspiro" => :optional
   depends_on :python if MacOS.version <= :snow_leopard
-
-  resource "gnulib" do
-    url "git://git.savannah.gnu.org/gnulib.git",
-        :revision => "29ea6d6fe2a699a32edbe29f44fe72e0c253fcee"
-  end
 
   fails_with :llvm do
     build 2336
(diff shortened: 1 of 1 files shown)


jtanx:
I've also switched libspiro and libuninameslist to be recommended, because realistically, only FontForge uses these libraries.

Formula/fontforge.rb
```diff
@@ -16,71 +16,72 @@ class Fontforge < Formula
...
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+
+    resource "gnulib" do
+      url "git://git.savannah.gnu.org/gnulib.git",
+        :revision => "29ea6d6fe2a699a32edbe29f44fe72e0c253fcee"
+    end
+  end
+
+  depends_on :x11 => :optional
```
MikeMcQuaid: Please remove this, we'd rather not support it.

Formula/fontforge.rb
```diff
@@ -16,71 +16,72 @@ class Fontforge < Formula
 
   deprecated_option "with-gif" => "with-giflib"
 
-  # Autotools are required to build from source in all releases.
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
+  if build.head?
```
MikeMcQuaid: Make a `head do` block and put these in there along with the above `head` as a `url` instead.

Formula/fontforge.rb
MikeMcQuaid: Please remove this, we'd rather not support it.

Formula/fontforge.rb
```diff
@@ -16,71 +16,72 @@ class Fontforge < Formula
...
-  depends_on "pango"
   depends_on "zeromq"
   depends_on "czmq"
-  depends_on "cairo"
   depends_on "fontconfig"
   depends_on "libpng" => :recommended
   depends_on "jpeg" => :recommended
   depends_on "libtiff" => :recommended
+  depends_on "libspiro" => :recommended
+  depends_on "libuninameslist" => :recommended
```
MikeMcQuaid: If these are only used by FontForge then use vendor them as a `resource` in this package rather than having them separately.

Formula/fontforge.rb
```diff
@@ -16,71 +16,72 @@ class Fontforge < Formula
...
-    # Don't link libraries to libpython, but do link binaries that expect
-    # to embed a python interpreter
-    # https://github.com/fontforge/fontforge/issues/2353#issuecomment-121009759
     ENV["PYTHON_CFLAGS"] = `python-config --cflags`.chomp
-    ENV["PYTHON_LIBS"] = "-undefined dynamic_lookup"
-    python_libs = `python2.7-config --ldflags`.chomp
-    inreplace "fontforgeexe/Makefile.am" do |s|
-      oldflags = s.get_make_var "libfontforgeexe_la_LDFLAGS"
-      s.change_make_var! "libfontforgeexe_la_LDFLAGS", "#{python_libs} #{oldflags}"
-    end
```
MikeMcQuaid: These issues all fixed?

Formula/fontforge.rb
```diff
@@ -105,6 +106,9 @@ def caveats; <<-EOS.undent
 
     Alternatively, install with Homebrew-Cask:
       brew cask install fontforge
+
+    As a last resort, you may try the *unsupported* option:
+      brew install fontforge --with-x11
```
MikeMcQuaid: Please remove this, we'd rather not support it.

MikeMcQuaid (changes requested):

Formula/fontforge.rb
MikeMcQuaid: He's not active on Homebrew right now.

Formula/fontforge.rb
jtanx: Yes

Formula/fontforge.rb
jtanx: They're optional and they are separate libraries. No need to force them onto people who'd rather not build with these options, but at the same time, FontForge is the primary user.

Formula/fontforge.rb
jtanx: I think I originally had a `head do` block but rubocop complained because I also had an `if head?` further down in the compilation step.

Formula/fontforge.rb
jtanx: Uh, FontForge really does depend on X11. Without support for things like this, it's near impossible to get anyone on Mac to actually test new features with any reasonable sort of speed. X11 support has already been disabled by default, so there's not much in the way of having to support anything.

Formula/fontforge.rb
jtanx: Okay. He was the one who suggested that this way was possible.

Formula/fontforge.rb
jtanx: I'm hesitant to do this. They *are* distinct libraries and if I understand `resouce` correctly, it's just odd that you'd shift the creation of a *separate* entity/library into the formula for a dependant library - this goes against practically every package management system out there. 

IIRC, GEGL can use libspiro, although the current formula doesn't reference this.

Formula/fontforge.rb
jtanx: I've made the change.

Formula/fontforge.rb
DomT4: If you tag me in something it'll likely get a reply, I'm just not spending much time actively looking for stuff to do.

I have no objection to the X11 dep going in as optional personally, at least until https://github.com/fontforge/fontforge/pull/2782 is resolved positively (fingers crossed) and that's available in a release. There's a decent amount of optional X11 stuff in core.

It'd be a fair claim that most people doing `brew install fontforge` want at least a GUI optionally available. The alternative if upstream is planning to keep making `.app`s is to remove this entirely in favour of the Cask.

MikeMcQuaid: Given we're removed it previously I'd rather not re-add it until we have something that works without X11.

ilovezfs:
Newer version out if you'd like to refresh this
https://github.com/fontforge/fontforge/archive/20161012.tar.gz

jtanx:
No, I'd need to make the dist tarball first. But I have no time.

ilovezfs:
@jtanx no problem. Take your time.

MikeMcQuaid:
@jtanx Any news on this? Thanks!

jtanx:
Sorry - not much time to work on this, and I think there's an issue with compiling on Sierra too. I don't have easy access to a system to debug that. Someone else might need to work this one out, otherwise it may be a while before I get back on this.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@jtanx this PR is now conflicted because I just shipped https://github.com/Homebrew/homebrew-core/pull/6638.

I made two changes to the fontforge formula in that PR:
(1) I dropped the dependencies on zeromq and czmq because fontforge is no longer compatible with the current version of czmq due to fontforge's use of the czmq v2 API, which was retired. See https://github.com/fontforge/fontforge/issues/2950.
(2) I revision bumped the formula due to (1).

PR #5619, opened 2016-10-06
qt 5.7.0 update

steve3d:

---

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -14,9 +14,9 @@ end
 class Qt5 < Formula
   desc "Version 5 of the Qt framework"
   homepage "https://www.qt.io/"
-  url "https://download.qt.io/official_releases/qt/5.6/5.6.1-1/single/qt-everywhere-opensource-src-5.6.1-1.tar.xz"
-  mirror "https://www.mirrorservice.org/sites/download.qt-project.org/official_releases/qt/5.6/5.6.1-1/single/qt-everywhere-opensource-src-5.6.1-1.tar.xz"
-  sha256 "ce08a7eb54661705f55fb283d895a089b267c688fabe017062bd71b9231736db"
+  url "https://download.qt.io/official_releases/qt/5.7/5.7.0/single/qt-everywhere-opensource-src-5.7.0.tar.xz"
+  mirror "https://www.mirrorservice.org/sites/download.qt-project.org/official_releases/qt/5.7/5.7.0/single/qt-everywhere-opensource-src-5.7.0.tar.xz"
+  sha256 "a6a2632de7e44bbb790bc3b563f143702c610464a7f537d02036749041fd1800"
 
   head "https://code.qt.io/qt/qt5.git", :branch => "5.6", :shallow => false
 
@@ -48,6 +48,15 @@ class Qt5 < Formula
     sha256 "2cf77b820f46f0c404284882b4a4a97bf005b680062842cdc53e107a821deeda"
   end
 
+  # Fix build error under El Capitan/Sierra (10.11/10.12), under these version
+  # bluetooth class moved from IOBluetooth to CoreBluetooth, so anycode still
+  # using IOBluetooth will fail to build.
+  # This patch also fix the missing symbol in cups backend of qtwebengine
+  patch do
+    url "http://localhost/qt5-bluetooth-cups-mac-10.1x.patch"
+    sha256 "b5d219682ce33e5b5c1784689edd0f31879b107713b7060f69930f76fd11eb8d"
+  end
+
   keg_only "Qt 5 conflicts Qt 4"
 
   option "with-docs", "Build documentation"
@@ -73,7 +82,7 @@ class Qt5 < Formula
 
   resource "qt-webkit" do
     # http://lists.qt-project.org/pipermail/development/2016-March/025358.html
-    url "https://download.qt.io/community_releases/5.6/5.6.1/qtwebkit-opensource-src-5.6.1.tar.gz"
+    url "https://download.qt.io/community_releases/5.7/5.7.0/qtwebkit-opensource-src-5.7.0.tar.xz"
     sha256 "f5ba5afc5846fc755575dd04081a90a9536f920e312f18f6fb1f5a0c33f477b0"
   end
 
@@ -90,12 +99,21 @@ class Qt5 < Formula
       -qt-pcre
       -nomake tests
       -no-rpath
+      -pch
+      -system-proxies
+      -optimized-tools
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"
 
     args << "-plugin-sql-mysql" if build.with? "mysql"
     args << "-plugin-sql-psql" if build.with? "postgresql"
+
+    # Both mysql and postgresql require openssl libs to link
+    if build.with? "mysql" or build.with? "postgresql"
+      openssl_opt = Formula["openssl"].opt_prefix
+      args << "-L#{openssl_opt}/lib"
+    end
 
     if build.with? "dbus"
       dbus_opt = Formula["dbus"].opt_prefix
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -53,7 +53,7 @@ class Qt5 < Formula
   # using IOBluetooth will fail to build.
   # This patch also fix the missing symbol in cups backend of qtwebengine
   patch do
-    url "http://localhost/qt5-bluetooth-cups-mac-10.1x.patch"
+    url "https://gist.githubusercontent.com/steve3d/b246f2eb35e5b9f89d1dcaa3470da59d/raw/75ba606df18d2a4901ff12b5ddd78c41a1711509/qt5-bluetooth-cups-mac-10.1x.patch"
     sha256 "b5d219682ce33e5b5c1784689edd0f31879b107713b7060f69930f76fd11eb8d"
   end
 
@@ -83,7 +83,7 @@ class Qt5 < Formula
   resource "qt-webkit" do
     # http://lists.qt-project.org/pipermail/development/2016-March/025358.html
     url "https://download.qt.io/community_releases/5.7/5.7.0/qtwebkit-opensource-src-5.7.0.tar.xz"
-    sha256 "f5ba5afc5846fc755575dd04081a90a9536f920e312f18f6fb1f5a0c33f477b0"
+    sha256 "c7a3253cbf8e6035c54c3b08d8a9457bd82efbce71d4b363c8f753fd07bd34df"
   end
 
   def install
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -64,6 +64,7 @@ class Qt5 < Formula
   option "with-oci", "Build with Oracle OCI plugin"
   option "with-qtwebkit", "Build with QtWebkit module"
   option "without-webengine", "Build without QtWebEngine module"
+  option "without-plugins", "Build without using gif/jpeg/sql plugins"
 
   deprecated_option "qtdbus" => "with-dbus"
   deprecated_option "with-d-bus" => "with-dbus"
@@ -106,8 +107,19 @@ class Qt5 < Formula
 
     args << "-nomake" << "examples" if build.without? "examples"
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
-    args << "-plugin-sql-psql" if build.with? "postgresql"
+    if build.without? "plugins"
+      use_plugin = "qt"
+      args << "-qt-sql-sqlite"
+      inreplace "qtbase/configure" do |s|
+        s.gsub! "CFG_JPEG=plugin", "CFG_JPEG=yes"
+        s.gsub! "CFG_GIF=plugin", "CFG_GIF=yes"
+      end
+    else
+       use_plugin = "plugin"
+    end
+
+    args << "-#{use_plugin}-sql-mysql" if build.with? "mysql"
+    args << "-#{use_plugin}-sql-psql" if build.with? "postgresql"
 
     # Both mysql and postgresql require openssl libs to link
     if build.with? "mysql" or build.with? "postgresql"
@@ -129,7 +141,7 @@ class Qt5 < Formula
     if build.with? "oci"
       args << "-I#{ENV["ORACLE_HOME"]}/sdk/include"
       args << "-L#{ENV["ORACLE_HOME"]}"
-      args << "-plugin-sql-oci"
+      args << "-#{use_plugin}-sql-oci"
     end
 
     args << "-skip" << "qtwebengine" if build.without? "webengine"


ilovezfs:
I think we intend to wait until upstream puts out a Sierra compatible release. The workarounds have been known for a couple weeks now.

Formula/qt5.rb
```diff
@@ -48,13 +48,23 @@ class Qt5 < Formula
     sha256 "2cf77b820f46f0c404284882b4a4a97bf005b680062842cdc53e107a821deeda"
   end
 
+  # Fix build error under El Capitan/Sierra (10.11/10.12), under these version
+  # bluetooth class moved from IOBluetooth to CoreBluetooth, so anycode still
+  # using IOBluetooth will fail to build.
+  # This patch also fix the missing symbol in cups backend of qtwebengine
+  patch do
+    url "https://gist.githubusercontent.com/steve3d/b246f2eb35e5b9f89d1dcaa3470da59d/raw/75ba606df18d2a4901ff12b5ddd78c41a1711509/qt5-bluetooth-cups-mac-10.1x.patch"
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/qt5.rb
```diff
@@ -48,13 +48,23 @@ class Qt5 < Formula
...
+  end
+
   keg_only "Qt 5 conflicts Qt 4"
 
   option "with-docs", "Build documentation"
   option "with-examples", "Build examples"
   option "with-oci", "Build with Oracle OCI plugin"
   option "with-qtwebkit", "Build with QtWebkit module"
   option "without-webengine", "Build without QtWebEngine module"
+  option "without-plugins", "Build without using gif/jpeg/sql plugins"
```
MikeMcQuaid: Why would you use this option?

Formula/qt5.rb
```diff
@@ -90,12 +100,32 @@ def install
       -qt-pcre
       -nomake tests
       -no-rpath
+      -pch
+      -system-proxies
+      -optimized-tools
```
MikeMcQuaid: Can you explain what each of these does?

MikeMcQuaid:
> I think we intend to wait until upstream puts out a Sierra compatible release.

@ilovezfs I'm actually OK with us getting something in here before that as it's causing a lot of complaints and Qt's turnaround is getting a bit ridiculous.

ilovezfs:
Why wouldn't we patch 5.6.1 then?

MikeMcQuaid:
@ilovezfs I'm fine with patching 5.6.1 instead too. Whichever gets a patch submitted and ideally accepted upstream first.

Formula/qt5.rb
steve3d: -pch enable precompiled-headers to speed up the compile process, with this, the compile process can speed up to 30%-40%, so why not use it? unless you prefer the slow compile.
-system-proxies, use system proxies setting instead of qt's own proxy setting in qtnetwork module, if you don't have this, anyone what to use the proxy in qt-network programs will need to set another proxy settings within qt itself, so I doubt anyone would do that.
-optimized-tools acutally, this does nothing here, it just make sure the qmake/moc/uic/rcc are build without debug info, in release build, this is the default setting.

Formula/qt5.rb
steve3d: it's just a personal favor, I don't like to use mysql as a QtSQL plugin, also, the gif and jpeg, I'd like gif/jpeg built into the QtGUI, also I like the sqlite/mysql/postgresql built into QtSql, so I added this option here.

Formula/qt5.rb
steve3d: sorry for this, this problem exists from qt 5.3, I've tried to compile qt 5.3 under 10.10 a few years ago, the bluetooth problem exists, now it's qt 5.7, still bluetooth problem...

steve3d:
no, I've checked the qtconnective source, there is no such fix since 8 month ago.

also, qtwebengine files I've patched hasn't been modified for 3 years.

So, I'm pretty sure if no one is going to submit this to qt, then these will not exists in any branch. because I've waited almost 2 years.

The reason I want to make the patch is that I want to test something under mac with qt, but I'm not sure I'll use qt under mac for my purpose.

Formula/qt5.rb
MikeMcQuaid: That's fine with us if you're fine not getting the binary package 👍 Just wanted to check it wasn't a requirement on Sierra.

Formula/qt5.rb
```diff
@@ -90,12 +100,32 @@ def install
...
     args << "-nomake" << "examples" if build.without? "examples"
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
-    args << "-plugin-sql-psql" if build.with? "postgresql"
+    if build.without? "plugins"
+      use_plugin = "qt"
+      args << "-qt-sql-sqlite"
+      inreplace "qtbase/configure" do |s|
+        s.gsub! "CFG_JPEG=plugin", "CFG_JPEG=yes"
+        s.gsub! "CFG_GIF=plugin", "CFG_GIF=yes"
```
MikeMcQuaid: Can these not be passed as options instead of modifying the configure?

Formula/qt5.rb
MikeMcQuaid: `-pch` isn't on by default? If `-optimized-tools` is on by default we can omit it. `-system-proxies` seems sensible.

Formula/qt5.rb
MikeMcQuaid: That's cool, we just need the patch submitted upstream.

Formula/qt5.rb
steve3d: -pch is never default since qt 3.x,
-optimized-tools is only on when building release build, so, I think it's ok to omit.

Formula/qt5.rb
steve3d: sorry, seems build jpeg/gif into QtGui will cause problems, I'm going to test more, I'll let you know the results.

steve3d: confirmed, there is NO problem with this at all,

No, there is no way to pass these as options, so the only way to do this is by modifying the configure script.

schoeps:
There should be a recent update to qtconnectivity according to https://codereview.qt-project.org/#/c/172619/

Formula/qt5.rb
```diff
@@ -48,13 +48,23 @@ class Qt5 < Formula
...
+  # Fix build error under El Capitan/Sierra (10.11/10.12), under these version
+  # bluetooth class moved from IOBluetooth to CoreBluetooth, so anycode still
+  # using IOBluetooth will fail to build.
+  # This patch also fix the missing symbol in cups backend of qtwebengine
+  patch do
+    url "https://gist.githubusercontent.com/steve3d/b246f2eb35e5b9f89d1dcaa3470da59d/raw/75ba606df18d2a4901ff12b5ddd78c41a1711509/qt5-bluetooth-cups-mac-10.1x.patch"
+    sha256 "b5d219682ce33e5b5c1784689edd0f31879b107713b7060f69930f76fd11eb8d"
+  end
+
   keg_only "Qt 5 conflicts Qt 4"
```
schoeps: I guess `keg_only` can be dropped.

ilovezfs:
@schoeps yes, note that is the 5.6 branch, though.

Formula/qt5.rb
ilovezfs: not until `qt` is removed or made `keg_only` ...

schoeps: I thought dropping `qt` (i.e. 4) was the plan?

ilovezfs: @schoeps I think `qt` will likely first be made keg_only before going bye-bye.

ilovezfs: @schoeps here's the current situation for the reverse-dependencies: https://github.com/Homebrew/homebrew-core/issues/1705

Formula/qt5.rb
ilovezfs: gif: http://code.qt.io/cgit/qt/qtbase.git/tree/configure#n1547
jpeg: http://code.qt.io/cgit/qt/qtbase.git/tree/configure#n4815 controlled by http://code.qt.io/cgit/qt/qtbase.git/tree/configure#n1540

steve3d: @ilovezfs, of course I know that, but my question is have you EVER tried pass these by configure??

CFG_GIF will always be set to 'plugin' when you build a shared version.

steve3d:
BTW, how to add a warning to users want to build qt?
Building qt from source will eat at least 25G disk space, and even with pch enabled, the build process can still take about 40mins on my 4790k cpu.

Formula/qt5.rb
steve3d: like I said before, the system-proxies will only make qt use system proxies settings, without it users will need to use qt api to set proxy settings. I doubt any developer want their users to set proxy ONLY for qt.

and the pch, this settings is default on windows build, but never get to be default on any other platform, because, the msvc always support pch, but not all other compiler support it, I remember that gcc supports pch start with version 3.4. so the pch is never default setting except under windows with msvc.

MikeMcQuaid:
> BTW, how to add a warning to users want to build qt?

We don't need one, it's implied by Homebrew being a mostly/often build-from-source package manager for those who don't use `/usr/local`.

Formula/qt5.rb
yezezey: FYI
Enabling openssl requires something like

` $ OPENSSL_LIBS='-L/usr/local/opt/openssl/lib -lssl -lcrypto' ./configure -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib  -no-securetransport
`

Yup, you need to set both environment variable and command line options.

ilovezfs: @steve3d no, I've never passed any configure options to Qt5. However, Mike's question was whether what you're trying to change with string replacement can instead be handled by passing options. From reading the configure script, it appears that upstream at least intends for the jpeg and gif plugin settings to be configurable with configure options not direct modification of the file. If that sometimes, or always, fails under certain conditions, then it would be a bug, which would need to be reported with an upstream issue link or upstream pull request link. Since you're proposing this as a non-default option, the stakes are rather low here, but nonetheless if we follow our own "rules," this string replacement is not acceptable as-is since it's patching a bug you say you've identifiied in the configure script, and yet that bug has not been reported as far as I am aware.

ilovezfs: @yezezey I'm afraid that I may not understand what you're saying here. Are you requesting some sort of change to the formula file?

steve3d: it's not a bug, Like I said, there are options that build sql driver as plugin or built into QtSql. But for gif/jpeg, there is no such switch, the default behavior is just build as plugin when build shared link qt libs, or built into qt gui when build static link qt.

and this gif/jpeg plugin is purely my personal favorite.

as to the system-proxy, I doubt it will break abi compatibly.  also, If I'm developing a qt based applications, I will go to the system-proxy way.

yezezey: Adding -L/usr/local/opt/openssl/lib may compile successfully, but Qt itself is not linked to openssl. That sounds bad.

yezezey: I suggest something like this.
Diff to steve3d's 3rd commit. https://github.com/yezezey/homebrew-core/commit/28f46e6c4db62946dbebb2adceaa8964042890c9

I am not a Ruby person and it's not well tested because building Qt is time consuming...

yezezey: @ilovezfs
I intended to comment on the 3rd commit, my bad.
@steve3d
I appreciate your effort by the way.

ilovezfs:
@BrewTestBot test this please

Formula/qt5.rb
jakepetroules: No. I just wanted OpenSSL headers in /usr/local/include for unrelated reasons.

ilovezfs: @jakepetroules Ah, OK. Thanks for the quick reply.

ilovezfs:
@steve3d Build failed on Sierra after 37 minutes: https://bot.brew.sh/job/Homebrew%20Core/9051/version=any_sierra/consoleFull

schoeps:
I think that I compiled the formula on Sierra using `ENV.deparallelize`, as far as I recall it failed without it.

ilovezfs:
@schoeps any idea why that would be a Sierra specific issue ...

ilovezfs:
@BrewTestBot test this please

schoeps:
@ilovezfs No, sorry. It's just one of the things I did to get qt5 compiled. Possibly it's even unrelated since I might have changed something else back then. As people have mentioned, it's not fun debugging it since it takes so long :)

ilovezfs:
In the spirit of throwing spaghetti :spaghetti: at the wall, I hard-coded HOMEBREW_MAKE_JOBS=1 into CI, so we'll see if that helps :see_no_evil:

schoeps:
@ilovezfs Experimental computer science :)

steve3d:
@ilovezfs how to download the console output? It's almost not possible to search for the error in browser.

ilovezfs:
@steve3d right, it's a parallelization bug which means it may be difficult to reproduce on demand. Deparallelizing the build would be a workaround, but will also make the build take even longer than it already does.

schoeps:
@ilovezfs If my vote counts anything, I suggest to live with this workaround. Most people will not build qt from source and we really need a working qt.

ilovezfs:
In addition to the parallelization, these bottles don't work if you look at

https://bot.brew.sh/job/Homebrew%20Core/9053/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_qt5/

and

https://bot.brew.sh/job/Homebrew%20Core/9053/version=any_sierra/testReport/junit/brew-test-bot/sierra/test_qt5/

That's because http://code.qt.io/cgit/qt/qtbase.git/patch/?id=77a71c32c9d19b87f79b208929e71282e8d8b5d9 is needed for 5.7.0 as well.

ilovezfs:
I've now published the preexisting qt 5.6.1 El Capitan bottle as the Sierra bottle to reduce the sense of urgency.

@schoeps you should be able to `brew update` and get the Octave GUI to build successfully now on 10.12

ilovezfs:
@steve3d thanks for this PR!

Please open a new PR with your additional changes beyond what's in the `qt5` formula now in master if you `brew update`.

See https://github.com/Homebrew/homebrew-core/pull/5951.

Note that the following are now already addressed:

- qt3d parallelized build failure
- bluetooth header issue in qtconnectivity
- bluetooth header issue in qtwebengine
- cups issue
- xcrun can't run xcrun in Xcode 8

steve3d:
@ilovezfs do you mean that I need to create another PR for the plugins and sql with openssl of this PR?

for what version? 5.6.2 or 5.7.0? there is no 5.7.1 yet, so all those problems still exists in 5.7.0.

ilovezfs:
>@ilovezfs do you mean that I need to create another PR for the plugins and sql with openssl of this PR?

Yes, if you still want to see those changes made. (Also the PCH and proxy changes)

>for what version?

For 5.7.0 since that's now the current version.

> there is no 5.7.1

5.7.1 doesn't need to be out. It'll just be a 5.7.0 formula file change (if there's no change to the bottles), bottle rebuild (if there is a change to the bottles but the preexisting bottles aren't broken), or formula revision bump (if there is a change to the bottles but the preexisting bottles are broken or very different (e.g., linkage changed)).

PR #5657, opened 2016-10-07
weechat: update to 1.6 & offer --without-tcl

azrdev:

### DIFF
--- a/Formula/weechat.rb
+++ b/Formula/weechat.rb
@@ -15,6 +15,7 @@ class Weechat < Formula
 
   option "with-perl", "Build the perl module"
   option "with-ruby", "Build the ruby module"
+  option "with-tcl", "Build the tcl module"
   option "with-curl", "Build with brewed curl"
   option "with-debug", "Build with debug information"
 
@@ -39,6 +40,7 @@ class Weechat < Formula
     args << "-DENABLE_LUA=OFF" if build.without? "lua"
     args << "-DENABLE_PERL=OFF" if build.without? "perl"
     args << "-DENABLE_RUBY=OFF" if build.without? "ruby"
+    args << "-DENABLE_TCL=OFF" if build.without? "tcl"
     args << "-DENABLE_ASPELL=OFF" if build.without? "aspell"
     args << "-DENABLE_GUILE=OFF" if build.without? "guile"
     args << "-DENABLE_PYTHON=OFF" if build.without? "python"
--- a/Formula/weechat.rb
+++ b/Formula/weechat.rb
@@ -1,8 +1,8 @@
 class Weechat < Formula
   desc "Extensible IRC client"
   homepage "https://www.weechat.org"
-  url "https://weechat.org/files/src/weechat-1.5.tar.gz"
-  sha256 "3174558556a20ae8f9ee3abbf66b7d42b657d3370322555501a707e339e10771"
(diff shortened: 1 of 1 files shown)


mistydemeo:
Looks like the build is failing; are you able to look into it?

azrdev:
@mistydemeo thx for the headsup – I'm a linuxbrew user only, so I dunno why this fails here, but I reported upstream at https://github.com/weechat/weechat/issues/276#issuecomment-252416756

Formula/weechat.rb
```diff
@@ -39,6 +40,7 @@ def install
     args << "-DENABLE_LUA=OFF" if build.without? "lua"
     args << "-DENABLE_PERL=OFF" if build.without? "perl"
     args << "-DENABLE_RUBY=OFF" if build.without? "ruby"
+    args << "-DENABLE_TCL=OFF" if build.without? "tcl"
```
MikeMcQuaid: Is this not defaulting to off, then, and changing behavour?

MikeMcQuaid: Let's leave `tcl` on by default, thanks.

MikeMcQuaid:
@azrdev Any news on the comments? Thanks!

azrdev:
@MikeMcQuaid If I understand [`build.with(out)`](http://www.rubydoc.info/github/Homebrew/brew/master/BuildOptions#with%3F-instance_method) correctly, It might suffice to use `option "without-tcl"` on line 18 instead of `option "with-tcl"` -- correct?
Apart from that, [I see here](http://www.rubydoc.info/github/Homebrew/brew/master/Formula#depends_on-class_method) only the possibility to add tcl as a `:recommended` (i.e. default ON) dependency

MikeMcQuaid:
> It might suffice to use option "without-tcl" on line 18 instead of option "with-tcl" -- correct?

Yup 👍

dunn:
@BrewTestBot test this please

dunn:
https://bot.brew.sh/job/Homebrew%20Core/11685/
```
  "_res_9_init", referenced from:
      _irc_server_xfer_send_ready_cb in irc-server.o
```

dunn:
Also if you could change the test from
```ruby
system "weechat", "-r", "/quit"
```
to
```ruby
system bin/"weechat", "-r", "/quit"
```
that'd be great!

azrdev:
@dunn done, whatever this has to do with this PR

robbiet480:
@azrdev I was able to get this to build finally with the changes proposed in #4427.

neutric:
Upstream added resolv (see [here](https://github.com/weechat/weechat/commit/e098a0dc5414f67006c9e8f0ff02f22f715a0a55) and [here](https://github.com/weechat/weechat/commit/228040dc04648507d9a57dcc6324d436cd97a4bc)) to the build scripts but compilation continues to fail with undefined symbols (see dunn's comment above on Nov 18). `export LDFLAGS=-lresolv` as suggested in #4427 solves the problem. Upstream knows that this issue persists but doesn't have macOS to debug and thus [asks for help](https://github.com/weechat/weechat/issues/276#issuecomment-267806781). 

CMake properly detects resolv, so something is probably wrong about where in the CMake script it is added, but I know too little about CMake to be of help.

MikeMcQuaid:
@neutric Convincing upstream to set `LDFLAGS=-lresolv` in CMake for macOS is probably the best fix.

PR #5661, opened 2016-10-07
pacparser 1.3.7 (new formula)

aaronhurt:
Adding pacparser (github.com/pacparser/pacparser).  A library to parse proxy auto-config (PAC) files.

### DIFF
--- a/Formula/pacparser.rb
+++ b/Formula/pacparser.rb
@@ -1,0 +1,108 @@
+class Pacparser < Formula
+  desc "Library to parse proxy auto-config (PAC) files"
+  homepage "https://github.com/pacparser/pacparser"
+  url "https://github.com/pacparser/pacparser/archive/1.3.7.tar.gz"
+  sha256 "575c5d8096b4c842b2af852bbb8bcfde96170b28b49f33249dbe2057a8beea13"
+  head "https://github.com/pacparser/pacparser.git"
+
+  depends_on "python" => :optional
+
+  def install
+    # Disable parallel build due to upstream concurrency issue.
+    # https://github.com/pacparser/pacparser/issues/27
+    ENV.deparallelize
+    ENV["VERSION"] = version
+    Dir.chdir "src"
+    system "make", "install",
+           "PREFIX=#{prefix}"
+    if build.with? "python"
+      system "make", "install-pymod",
+             "EXTRA_ARGS=\"--prefix=#{prefix}\""
+    end
+  end
+
+  test do
+    # example pacfile taken from upstream sources
+    (testpath/"test.pac").write <<-'EOS'.undent
+      function FindProxyForURL(url, host) {
+
+        if ((isPlainHostName(host) ||
+            dnsDomainIs(host, ".manugarg.com")) &&
+            !localHostOrDomainIs(host, "www.manugarg.com"))
+          return "plainhost/.manugarg.com";
+
+        // Return externaldomain if host matches .*\.externaldomain\.com
+        if (/.*\.externaldomain\.com/.test(host))
+          return "externaldomain";
+
+        // Test if DNS resolving is working as intended
+        if (dnsDomainIs(host, ".google.com") &&
+            isResolvable(host))
+          return "isResolvable";
+
+        // Test if DNS resolving is working as intended
+        if (dnsDomainIs(host, ".notresolvabledomainXXX.com") &&
+            !isResolvable(host))
+          return "isNotResolvable";
+
+        if (/^https:\/\/.*$/.test(url))
+          return "secureUrl";
+
+        if (isInNet(myIpAddress(), '10.10.0.0', '255.255.0.0'))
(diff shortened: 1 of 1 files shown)


Formula/pacparser.rb
```diff
@@ -0,0 +1,25 @@
...
+  desc "Library to parse proxy auto-config (PAC) files"
+  homepage "https://github.com/pacparser/pacparser"
+  url "https://github.com/pacparser/pacparser/archive/1.3.6.tar.gz"
+  sha256 "c1f6f2ccd4fec329175f882102e47c57328e8e6e16aa9dbd4f8b859f9a028e83"
+  head "https://github.com/pacparser/pacparser.git"
+
+  depends_on "python" => :optional
+
+  def install
+    system "make", "-j1", "-C", "src", "install",
```
mistydemeo: Does this not work with a parallel build?

leprechau: No, it doesn't.  It's a known issue.  I disable that for my Alpine Linux build of this as well.

https://github.com/pacparser/pacparser/issues/27

MikeMcQuaid: Use `ENV.deparallelize` and add a link to the upstream issue and a comment in Ruby to clarify.

leprechau: Done, committed and squished.

Formula/pacparser.rb
```diff
@@ -0,0 +1,28 @@
...
+  sha256 "c1f6f2ccd4fec329175f882102e47c57328e8e6e16aa9dbd4f8b859f9a028e83"
+  head "https://github.com/pacparser/pacparser.git"
+
+  depends_on "python" => :optional
+
+  def install
+    # Disable parallel build due to upstream concurrency issue.
+    # https://github.com/pacparser/pacparser/issues/27
+    ENV.deparallelize
+    system "make", "-j1", "-C", "src", "install",
```
MikeMcQuaid: Can remove the `-j1` now. Additionally, a `Dir.chdir "src"` will save needing to repeat it below. You may also want to use a `ENV["VERSION"] = version` to avoid setting that twice, too.

MikeMcQuaid (changes requested):

Formula/pacparser.rb
```diff
@@ -0,0 +1,25 @@
...
+    if build.with? "python"
+      system "make", "-j1", "-C", "src", "install-pymod",
+             "EXTRA_ARGS=\"--prefix=#{prefix}\"",
+             "VERSION=#{version}"
+    end
+  end
+
+  test do
+    ## functional tests are run as part of the make
+    assert_equal version.to_s, shell_output("#{bin}/pactester -v").strip
```
leprechau: The upstream package has test cases that are run at the end of the make.  Those tests are -pure functional tests and do run through example PAC files.  If/when those tests fail the entire build will fail.

MikeMcQuaid: This is not a test of the software compilation but a test that it's been installed by Homebrew properly.

leprechau: Then I guess I'm confused as why a simple "it executes" isn't sufficient for this section if all it is testing is that the binary is in the proper location and executes?  Like I said there are a series of actual functional tests run during the make process.  I could replicate those in the test section of the package but it seems a bit redundant to run them twice.

Formula/pacparser.rb
leprechau: No problem, I can change that.

Formula/pacparser.rb
MikeMcQuaid: We've found `--version` is often not sufficient to identify packaging bugs compared to a functional/integration test. Duplicating a single functional test here would be 👍

Formula/pacparser.rb
```diff
@@ -0,0 +1,28 @@
+class Pacparser < Formula
+  desc "Library to parse proxy auto-config (PAC) files"
+  homepage "https://github.com/pacparser/pacparser"
+  url "https://github.com/pacparser/pacparser/archive/1.3.6.tar.gz"
```
leprechau: @manugarg np, I need to finish up this PR with the requested tests and I'll update to 1.3.7 at the same time.  We use pacparser in our go applications with my go-pacparser wrapper.  Thank you for the time and effort developing the library and test tool.

manugarg: @leprechau Sounds good. Thanks!

leprechau: Been a bit tied up with other projects the last few days but I will get this updated hopefully this week.  I haven't forgotten about it yet ;p

leprechau:
@MikeMcQuaid @manugarg updated to 1.3.7 and replicated the test cases in the test section of the formula

leprechau:
squished and pushed as a single commit

MikeMcQuaid (approved):

MikeMcQuaid:

PR #5690, opened 2016-10-08
csvtomd 0.1.1 (new formula)

zmwangx:

---

266 stars: not _too_ shabby. I found I needed this a couple times recently. Needless to say I'll be responsible for formulae I submitted, so no burden on anyone else (this one's just a ~100 line Python script, so no burden to begin with).

### DIFF
--- a/Formula/csvtomd.rb
+++ b/Formula/csvtomd.rb
@@ -1,0 +1,30 @@
+class Csvtomd < Formula
+  include Language::Python::Virtualenv
+
+  desc "CSV to Markdown table converter"
+  homepage "https://github.com/mplewis/csvtomd"
+  url "https://files.pythonhosted.org/packages/5a/d8/9dbce820243bb6db670cd1ddea80ea1890c6bfd5d122910fdd157d51d71f/csvtomd-0.1.1.tar.gz"
+  sha256 "f2cd1da71ca8ed823d1f02167039e07e58d15a73d672069dfddcda8639576490"
+
+  depends_on :python3
+
+  def install
+    virtualenv_create(libexec, "python3")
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):

Formula/csvtomd.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def install
+    virtualenv_create(libexec, "python3")
+    virtualenv_install_with_resources
+  end
+
+  test do
+    csv = testpath/"test.csv"
+    md = testpath/"test.md"
+    csv.write <<-EOS.undent
```
ilovezfs: you should be able to do `csv = <<-EOS.undent.strip` and avoid the `csv.read.strip` below.

Formula/csvtomd.rb
```diff
@@ -0,0 +1,30 @@
...
+      column 1,column 2
+      hello,world
+    EOS
+    md.write <<-EOS.undent
```
ilovezfs: you should be able to do md = <<-EOS.undent.strip and avoid the md.read.strip below.

zmwangx: Well, having a text file without LF as the last character is against my religion; it's not technically a text file per POSIX definitions.

ilovezfs: it's nice a file if you do `md =` ... instead of `md.write` and `md.read`.

just a string.

zmwangx: Oh I see, you mean I don't need a file here.

ilovezfs: yup :)

zmwangx: Addressed.

PR #5692, opened 2016-10-08
Add explicit Python support for collectd

jd:
The default configure script tries to guess if Python is present, but
links with the wrong one if the flag is not specified, which ends up not
working:

Unhandled python exception in importing module:
ImportError:
dlopen(/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_io.so,
2): Symbol not found: __PyCodecInfo_GetIncrementalDecoder

Adding an explicit dependency and flag at compile time make sure it
works properly.

### DIFF
--- a/Formula/collectd.rb
+++ b/Formula/collectd.rb
@@ -24,6 +24,7 @@ class Collectd < Formula
   end
 
   option "with-java", "Enable Java support"
+  option "with-python", "Enable Python support"
   option "with-protobuf-c", "Enable write_riemann via protobuf-c support"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:

For future reference the preferred commit message format is `collectd: add Python option.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #5710, opened 2016-10-09
zimg 2.3

ilovezfs:
set minimum macOS requirement of >= 10.11 and remove the fix in the
formula since upstream says

>  This won't be fixed. Upgrade your CLANG version to 3.8 (XCode 7.3?),
>  so that it will be compatible with ICC/ICL.


ilovezfs:
See
https://github.com/sekrit-twc/zimg/issues/52#issuecomment-244628970
https://github.com/NixOS/nixpkgs/pull/18029

PR #5711, opened 2016-10-09
ssh-agent-filter 0.4.2 (new formula)

kouk:

---

```
ssh-agent-filter:
```

However this [package is in debian stable](https://packages.debian.org/jessie/net/ssh-agent-filter)  since [2013](https://tracker.debian.org/pkg/ssh-agent-filter) and it's [popularity is growing](https://qa.debian.org/popcon.php?package=ssh-agent-filter).

### DIFF
--- a/Formula/ssh-agent-filter.rb
+++ b/Formula/ssh-agent-filter.rb
@@ -1,0 +1,62 @@
+class SshAgentFilter < Formula
+  desc "filtering proxy for ssh-agent"
+  homepage "https://github.com/tiwe-de/ssh-agent-filter"
+  url "https://github.com/tiwe-de/ssh-agent-filter/archive/0.4.2.tar.gz"
+  sha256 "5586e40da34f4afd873d7ff30f115c4a1b9455c1b81f3c4679c995db7615b080"
+
+  depends_on "help2man" => :build
+  depends_on "boost"
+  depends_on "nettle"
+
+  patch :DATA
+
+  def install
+    ENV.deparallelize
+    system "make", "ssh-agent-filter.1"
+    man1.install "ssh-agent-filter.1"
+    bin.install "ssh-agent-filter"
+    bin.install "afssh"
+  end
+
+  test do
+    # `test do` will create, run in and delete a temporary directory.
+    #
+    # This test will fail and we won't accept that! It's enough to just replace
+    # "false" with the main program this formula installs, but it'd be nice if you
+    # were more thorough. Run the test with `brew test ssh-agent-filter`. Options passed
+    # to `brew install` such as `--HEAD` also need to be provided to `brew test`.
+    #
+    # The installed folder is not in the path, so use the entire path to any
+    # executables being tested: `system "#{bin}/program", "do", "something"`.
+    system "#{bin}/ssh-agent-filter", "-V"
+  end
+end
+__END__
+Fix afssh https://github.com/tiwe-de/ssh-agent-filter/pull/5#issuecomment-242216945
+Fix man page generation due to missing perl Locale::gettext
+diff --git a/afssh b/afssh
+index c482aea..6c2fbdf 100755
+--- a/afssh
++++ b/afssh
+@@ -52,7 +52,7 @@ fi
+ declare -a agent_filter_args
+ 
+ if [ -x "${BASH_SOURCE%/*}/ssh-agent-filter" ]; then
+-	SAF=$(readlink -f "${BASH_SOURCE%/*}/ssh-agent-filter")
++	SAF=$(realpath "${BASH_SOURCE%/*}/ssh-agent-filter")
+ else
+ 	SAF=$(which ssh-agent-filter)
+ fi
+diff --git a/Makefile b/Makefile
+index b2e05ec..30b3e9f 100644
+--- a/Makefile
++++ b/Makefile
+@@ -28,7 +28,7 @@ all: ssh-agent-filter.1 afssh.1 ssh-askpass-noinput.1
+ 	pandoc -s -w man $< -o $@
+ 
+ %.1: %.help2man %
+-	help2man -i $< -o $@ -N -L C.UTF-8 $(*D)/$(*F)
++	help2man -i $< -o $@ -N $(*D)/$(*F)
+ 
+ ssh-agent-filter: ssh-agent-filter.o
+ 
--- a/Formula/ssh-agent-filter.rb
+++ b/Formula/ssh-agent-filter.rb
@@ -31,10 +31,9 @@ class SshAgentFilter < Formula
 
   test do
     (testpath/"test.sh").write <<-EOS.undent
-      mydir=$(pwd)
       cd $(mktemp -d)
-      ssh-keygen -q -N '' -f keyA
-      ssh-keygen -q -N '' -f keyB
+      ssh-keygen -q -N '' -t ed25519 -C keyA -f keyA
+      ssh-keygen -q -N '' -t ed25519 -C keyB -f keyB
       eval $(ssh-agent)
       pid=$SSH_AGENT_PID
       ssh-add keyA
@@ -43,7 +42,7 @@ class SshAgentFilter < Formula
       eval $(ssh-agent-filter -c keyA)
       sha2=$(ssh-add -l | awk '{print $2}')
       kill $SSH_AGENT_PID $pid 
-      test "$sha1" == "$sha2"
+      test "$sha1" = "$sha2"
     EOS
     system "sh", "test.sh"
   end
--- a/Formula/ssh-agent-filter.rb
+++ b/Formula/ssh-agent-filter.rb
@@ -18,15 +18,13 @@ class SshAgentFilter < Formula
   patch do
     # https://github.com/tiwe-de/ssh-agent-filter/pull/6
     url "https://patch-diff.githubusercontent.com/raw/tiwe-de/ssh-agent-filter/pull/6.diff"
-    sha256 "e64c93daed3a0b696dd5325a60a8a5d65a267c28607e29b886eb8901ec6e7da7"
+    sha256 "5bd81db7cbfaf04e681f343ef6742061eaf6c9ead7d30c3fe6bf98eb8690c32f"
   end
 
   patch :DATA
 
   def install
-    system "mkdir", "-p", "#{prefix}/bin"
-    system "mkdir", "-p", "#{prefix}/man/man1"
-    system "make", "install", "PREFIX=#{prefix}"
+   system "make", "install", "PREFIX=#{prefix}"
   end
 
   test do
--- a/Formula/ssh-agent-filter.rb
+++ b/Formula/ssh-agent-filter.rb
@@ -18,11 +18,9 @@ class SshAgentFilter < Formula
   patch :DATA
 
   def install
-    ENV.deparallelize
-    system "make", "ssh-agent-filter.1"
-    man1.install "ssh-agent-filter.1"
-    bin.install "ssh-agent-filter"
-    bin.install "afssh"
+    system "mkdir", "-p", "#{prefix}/bin"
+    system "mkdir", "-p", "#{prefix}/man/man1"
+    system "make", "install", "PREFIX=#{prefix}"
   end
 
   test do
--- a/Formula/ssh-agent-filter.rb
+++ b/Formula/ssh-agent-filter.rb
@@ -19,15 +19,6 @@ class SshAgentFilter < Formula
   end
 
   test do
-    # `test do` will create, run in and delete a temporary directory.
-    #
-    # This test will fail and we won't accept that! It's enough to just replace
-    # "false" with the main program this formula installs, but it'd be nice if you
-    # were more thorough. Run the test with `brew test ssh-agent-filter`. Options passed
-    # to `brew install` such as `--HEAD` also need to be provided to `brew test`.
-    #
-    # The installed folder is not in the path, so use the entire path to any
-    # executables being tested: `system "#{bin}/program", "do", "something"`.
     system "#{bin}/ssh-agent-filter", "-V"
   end
 end
--- a/Formula/ssh-agent-filter.rb
+++ b/Formula/ssh-agent-filter.rb
@@ -36,8 +36,7 @@ class SshAgentFilter < Formula
       ssh-keygen -q -N '' -t ed25519 -C keyB -f keyB
       eval $(ssh-agent)
       pid=$SSH_AGENT_PID
-      ssh-add keyA
-      ssh-add keyB
+      ssh-add keyA keyB
       sha1=$(ssh-add -l | awk '/keyA/{print $2}')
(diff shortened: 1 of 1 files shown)


Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,62 @@
...
+  desc "filtering proxy for ssh-agent"
+  homepage "https://github.com/tiwe-de/ssh-agent-filter"
+  url "https://github.com/tiwe-de/ssh-agent-filter/archive/0.4.2.tar.gz"
+  sha256 "5586e40da34f4afd873d7ff30f115c4a1b9455c1b81f3c4679c995db7615b080"
+
+  depends_on "help2man" => :build
+  depends_on "boost"
+  depends_on "nettle"
+
+  patch :DATA
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project as a pull request and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def install
+    ENV.deparallelize
```
MikeMcQuaid: Can you please submit this as an issue to the upstream developers of this project and add a link to the upstream issue submission in a comment in the formula file. Thanks!

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,62 @@
...
+    system "make", "ssh-agent-filter.1"
+    man1.install "ssh-agent-filter.1"
```
MikeMcQuaid: Is there no `make install` we can use?

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,62 @@
...
+  test do
+    # `test do` will create, run in and delete a temporary directory.
+    #
+    # This test will fail and we won't accept that! It's enough to just replace
+    # "false" with the main program this formula installs, but it'd be nice if you
+    # were more thorough. Run the test with `brew test ssh-agent-filter`. Options passed
+    # to `brew install` such as `--HEAD` also need to be provided to `brew test`.
+    #
+    # The installed folder is not in the path, so use the entire path to any
+    # executables being tested: `system "#{bin}/program", "do", "something"`.
```
MikeMcQuaid: Remove all these comments.

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,62 @@
...
+    system "#{bin}/ssh-agent-filter", "-V"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

MikeMcQuaid (changes requested):

MikeMcQuaid:
> ssh-agent-filter:
>   * GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)
> However this package is in debian stable since 2013 and it's popularity is growing.

I'm ok with accepting this project when it at least the `master` branch works without patching on macOS.

Formula/ssh-agent-filter.rb
MikeMcQuaid: Adding one upstream would be amazing if you felt like submitting a patch there?

Formula/ssh-agent-filter.rb
kouk: @MikeMcQuaid good idea, I pushed a better test.

Formula/ssh-agent-filter.rb
kouk: @MikeMcQuaid For the first chunk I added it as an external patch. For the second I will comment below.

Formula/ssh-agent-filter.rb
kouk: I had already submitted the first part of the patch here: https://github.com/tiwe-de/ssh-agent-filter/pull/5
I added it to the formula as an external patch.

kouk:
@MikeMcQuaid one thing I'm not sure about is a missing dependency of help2man, the perl library Locale::gettext, which I worked around by removing the relevant command line arguments from the Makefile. Is there a package in Homebrew that contains this perl module? Otherwise the master branch will not build the man page successfully, and I think this package is installed by default in Debian.

Formula/ssh-agent-filter.rb
kouk: done

Formula/ssh-agent-filter.rb
kouk: @MikeMcQuaid this wasn't actually necessary in the end so I removed it.

Formula/ssh-agent-filter.rb
```diff
@@ -19,7 +19,22 @@ def install
   end
 
   test do
-    system "#{bin}/ssh-agent-filter", "-V"
+    (testpath/"test.sh").write <<-EOS.undent
+      mydir=$(pwd)
+      cd $(mktemp -d)
+      ssh-keygen -q -N '' -f keyA
```
tiwe-de: This needs considerable processing time for finding primes. You might use `-t ed25519` to select a key type that is much faster to generate. In that case you should also use `-C keyA` to set the comment or ssh-add/ssh-agent might use the embedded comment instead of the filename.

Formula/ssh-agent-filter.rb
```diff
@@ -19,7 +19,22 @@ def install
   end
 
   test do
-    system "#{bin}/ssh-agent-filter", "-V"
+    (testpath/"test.sh").write <<-EOS.undent
+      mydir=$(pwd)
```
tiwe-de: `mydir` appears to be unused

Formula/ssh-agent-filter.rb
```diff
@@ -19,7 +19,22 @@ def install
...
+      ssh-keygen -q -N '' -f keyB
+      eval $(ssh-agent)
+      pid=$SSH_AGENT_PID
+      ssh-add keyA
+      ssh-add keyB
+      sha1=$(ssh-add -l | awk '/keyA/{print $2}')
+      eval $(ssh-agent-filter -c keyA)
+      sha2=$(ssh-add -l | awk '{print $2}')
+      kill $SSH_AGENT_PID $pid 
+      test "$sha1" == "$sha2"
```
tiwe-de: `==` is a bashism, use `=` instead

tiwe-de (changes requested):

Formula/ssh-agent-filter.rb
tiwe-de: It seems like you only want to install some of the files the Debian package contains. The missing parts are at least:
- afssh.1 (needs pandoc to generate it)
- ssh-agent-filter.bash-completion (which version of bash-completion is available on OS X?)

tiwe-de:
the removed option to `help2man` seems to be unnecessary  on OS X or with that version of `help2man`. On a GNU system with full-blown `help2man` it is needed for it to not mangle characters outside of ASCII (in my name).

kouk:
@tiwe-de Thanks for your suggestions, I pushed the changes.

Formula/ssh-agent-filter.rb
kouk: @tiwe-de sorry, I'm not sure I understand what that is. Could you give me a pointer?

tiwe-de: This should read "pre-generated", sorry.

kouk: @tiwe-de if you know of a nice way to get the pre-generated keys in the Formula I would do that. But having them verbatim inside the file seems a bit overkill.

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,61 @@
...
+    sha256 "307ccd6e872a3f3b999d8a4368d152dfbd3a7e43b4cff0cacbc0390c78b2976b"
+  end
+
+  patch do
+    # https://github.com/tiwe-de/ssh-agent-filter/pull/6
+    url "https://patch-diff.githubusercontent.com/raw/tiwe-de/ssh-agent-filter/pull/6.diff"
+    sha256 "5bd81db7cbfaf04e681f343ef6742061eaf6c9ead7d30c3fe6bf98eb8690c32f"
+  end
+
+  patch :DATA
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,61 @@
...
+  url "https://github.com/tiwe-de/ssh-agent-filter/archive/0.4.2.tar.gz"
+  sha256 "5586e40da34f4afd873d7ff30f115c4a1b9455c1b81f3c4679c995db7615b080"
+
+  depends_on "help2man" => :build
+  depends_on "boost"
+  depends_on "nettle"
+
+  patch do
+    # upstream didn't want this because realpath isn't included in coreutils for Debian wheezy.
+    # see https://github.com/tiwe-de/ssh-agent-filter/pull/5
```
MikeMcQuaid: I've commented there but we need to figure out how to get something merged there or can't accept this (as it shows upstream doesn't want to support macOS).

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  patch :DATA
+
+  def install
+   system "make", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    (testpath/"test.sh").write <<-EOS.undent
+      cd $(mktemp -d)
```
MikeMcQuaid: You probably want to add `set -e` and perhaps also `set -x` in here.

MikeMcQuaid (changes requested):

Formula/ssh-agent-filter.rb
kouk: @tiwe-de I hesitated to add a dependency on pandoc because that means installing GHC. If there's a way in homebrew to remove a build dependency after then that might not hurt.

tiwe-de: Isn't that what `=> :build` is for? So a user who installs the binary package ("bottle") would not need to install it.

Formula/ssh-agent-filter.rb
kouk: @MikeMcQuaid `set -e` for sure, but not sure why `set -x` is of benefit.

MikeMcQuaid: It will make it easier to debug a failure.

tiwe-de: if you use `-e` you might want to use a `trap` to clean up.

MikeMcQuaid: `test do` already runs in a (cleaned-up) temporary directory so the `mktemp -d` is probably unnecessary.

Formula/ssh-agent-filter.rb
tiwe-de: Actually the manpage contained in the bottles generated by jenkins have the correct umlaut.
Can you test if the output manpage changes with the Locale::gettext perl module installed?

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,61 @@
...
+      ssh-keygen -q -N '' -t ed25519 -C keyA -f keyA
+      ssh-keygen -q -N '' -t ed25519 -C keyB -f keyB
```
kouk: @MikeMcQuaid done.

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,47 @@
...
+  sha256 "5586e40da34f4afd873d7ff30f115c4a1b9455c1b81f3c4679c995db7615b080"
+
+  depends_on "help2man" => :build
+  depends_on "pandoc" => :build 
+  depends_on "boost"
+  depends_on "nettle"
+
+  patch do
+    # https://github.com/tiwe-de/ssh-agent-filter/pull/6
+    url "https://patch-diff.githubusercontent.com/raw/tiwe-de/ssh-agent-filter/pull/6.diff"
```
MikeMcQuaid: Use `https://github.com/tiwe-de/ssh-agent-filter/pull/6.diff` as the URL to remove the need for the comment.

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,47 @@
...
+   end
+   system "make", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    gencmd = ["ssh-keygen", "-q", "-N", "", "-t" "ed25519"]
+    system *gencmd, "-C", "keyA", "-f", "keyA"
+    system *gencmd, "-C", "keyB", "-f", "keyB"
+    (testpath/"test.sh").write <<-EOS.undent
+      eval $(ssh-agent)
```
MikeMcQuaid: Again, this probably wants `set -e` and `set -x`.

MikeMcQuaid (changes requested):

Formula/ssh-agent-filter.rb
```diff
@@ -0,0 +1,48 @@
...
+  sha256 "5586e40da34f4afd873d7ff30f115c4a1b9455c1b81f3c4679c995db7615b080"
+
+  depends_on "help2man" => :build
+  depends_on "pandoc" => :build 
+  depends_on "gettext" => :build 
+  depends_on "boost"
+  depends_on "nettle"
+
+  patch do
+    url "https://github.com/tiwe-de/ssh-agent-filter/pull/6.diff"
```
MikeMcQuaid: Please add a one-line description of what this pull request is on the line above. It also may be worth waiting until this is merged and confirmed to be correct.

MikeMcQuaid (changes requested):

tiwe-de:
it seems you are trying to enable locale/gettext support in help2man at runtime, but it is disabled at "compile" time; see [the source of help2man](https://sources.debian.net/src/help2man/1.47.4/help2man.PL/).

If help2man in homebrew is always "compiled" without gettext support, you can just leave out the locale setting because non-ascii characters will not be mangled then.

MikeMcQuaid:
@kouk Any news on the comments? Thanks!

kouk:
@MikeMcQuaid sorry, I didn't have time to spare the last two weeks. I will try and work on it in the coming days, but if I understand @tiwe-de 's comment at least one patch will be required in the formula due to help2man not having gettext support in MacOS.

MikeMcQuaid:
@kouk That's ok, just checking in. As long as the patches are submitted upstream (and given how responsive they are ideally merged) before we merge this. Thanks again!

fxcoudert:
Low-visibility github repo, still does not work unpatched on macOS, no commits since October 2016. Upstream needs to move before we can consider adding this formula.

PR #5712, opened 2016-10-09
upgrade docker-compose to v1.8.1

joaocc:


Formula/docker-compose.rb
```diff
@@ -1,8 +1,8 @@
 class DockerCompose < Formula
   desc "Isolated development environments using Docker"
   homepage "https://docs.docker.com/compose/"
-  url "https://github.com/docker/compose/archive/1.8.0.tar.gz"
-  sha256 "a93d702086fdc95f79d20462ddaf7cd174e3d8053b79fe3c565f0ace76a16750"
+  url "https://github.com/docker/compose/archive/1.8.1.tar.gz"
+  sha256 "ee6534e2e420bdcd9beb5884f5dabd9f6ee832e43b3ad226f8b74f547bdd7fcd"
```
MikeMcQuaid: Seems to be wrong: https://bot.brew.sh/job/Homebrew%20Core/8966/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/fetch_docker_compose/

MikeMcQuaid (changes requested):

Formula/docker-compose.rb
joaocc: You are right. I misread the url didn't notice this was tgz (I inserted the sha256sum of the dmg file for mac). I will remove the PR since I don't have time to make this work properly from source. Thx

PR #5713, opened 2016-10-09
voltdb 6.6

patope:

---

### DIFF
--- a/Formula/voltdb.rb
+++ b/Formula/voltdb.rb
@@ -1,8 +1,8 @@
 class Voltdb < Formula
   desc "Horizontally-scalable, in-memory SQL RDBMS"
   homepage "https://github.com/VoltDB/voltdb"
-  url "https://github.com/VoltDB/voltdb/archive/voltdb-5.6.tar.gz"
-  sha256 "9ea24d8cacdf2e19ba60487f3e9dfefa83c18cb3987571abc44b858ce0db7c3e"
+  url "https://github.com/VoltDB/voltdb/archive/voltdb-6.6.tar.gz"
+  sha256 "300d06c8f5a1cc4a3537f216ea105d5ccc8987b10bad2c676a40543c8903d72e"
   head "https://github.com/VoltDB/voltdb.git"
 
   bottle do
@@ -14,11 +14,14 @@ class Voltdb < Formula
   end
 
   depends_on :ant => :build
+  depends_on "cmake" => :build
+
+  patch :DATA
 
   def install
     system "ant"
 
-    inreplace Dir["bin/*"] - ["bin/voltadmin", "bin/voltdb", "bin/rabbitmqloader"],
+    inreplace Dir["bin/*"] - ["bin/voltadmin", "bin/voltdb", "bin/rabbitmqloader", "bin/voltdeploy", "bin/voltenv"],
       %r{VOLTDB_LIB=\$VOLTDB_HOME\/lib}, "VOLTDB_LIB=$VOLTDB_HOME/lib/voltdb"
 
     (lib/"voltdb").install Dir["lib/*"]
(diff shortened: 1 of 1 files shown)


Formula/voltdb.rb
```diff
@@ -14,11 +14,14 @@ class Voltdb < Formula
   end
 
   depends_on :ant => :build
+  depends_on "cmake" => :build
+
+  patch :DATA
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project as a pull-request and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

MikeMcQuaid (changes requested):

Formula/voltdb.rb
patope: I'm sure they wont take this patch in it's current form (it would break linux distributions). Patch resolves incompatibilities between homebrew and their supported install targets (linux and download/extract).

Patch contains two fixes:
 - First change is basically same as inreplace above, but voltenv script differs a bit from other scripts so it's here.
 - Second change removes explict check in voltenv is voltdb installed behind symbolic links. 

Without patch voltenv gives error:
```
Warning: VoltDB exists in your path at:
    /usr/local/bin
  But this script is using VoltDB at:
    /usr/local/Cellar/voltdb/6.6/bin
```
I can file an issue and ask why there is such check. I can replace patch with inreplace commands.

MikeMcQuaid: We could use an inreplace temporarily but long-term we need to get this working without any patches.

Formula/voltdb.rb
```diff
@@ -14,13 +14,21 @@ class Voltdb < Formula
   end
 
   depends_on :ant => :build
+  depends_on "cmake" => :build
 
   def install
     system "ant"
 
-    inreplace Dir["bin/*"] - ["bin/voltadmin", "bin/voltdb", "bin/rabbitmqloader"],
+    inreplace Dir["bin/*"] - ["bin/voltadmin", "bin/voltdb", "bin/rabbitmqloader", "bin/voltdeploy", "bin/voltenv"],
```
MikeMcQuaid: At this point I wonder if it's worth specifying the actual files rather than removing from a wildcard.

Formula/voltdb.rb
```diff
@@ -14,13 +14,21 @@ class Voltdb < Formula
...
       %r{VOLTDB_LIB=\$VOLTDB_HOME\/lib}, "VOLTDB_LIB=$VOLTDB_HOME/lib/voltdb"
 
+    inreplace "bin/voltenv" do |s|
+      s.gsub! /VOLTDB_VOLTDB="\$VOLTDB_LIB"/, "VOLTDB_VOLTDB=\"$VOLTDB_BASE/voltdb\""
+
+      # Remove voltenv installed as link check
+      s.gsub! /if \[ "\$\{0\}" = "\$SOURCE" \]; then/, "if [ \"${0}\" = \"${BASH_SOURCE[0]}\" ]; then"
```
MikeMcQuaid: Should definitely speak to upstream about this check to ensure there's not other things that will break here.

MikeMcQuaid (changes requested):

Formula/voltdb.rb
patope: I made pull request for this https://github.com/VoltDB/voltdb/pull/3973

MikeMcQuaid: Great, thanks! Could you add a link to it as part of the comment above?

Formula/voltdb.rb
```diff
@@ -14,13 +14,21 @@ class Voltdb < Formula
...
 
   depends_on :ant => :build
+  depends_on "cmake" => :build
 
   def install
     system "ant"
 
-    inreplace Dir["bin/*"] - ["bin/voltadmin", "bin/voltdb", "bin/rabbitmqloader"],
+    inreplace Dir["bin/*"] - ["bin/voltadmin", "bin/voltdb", "bin/rabbitmqloader", "bin/voltdeploy", "bin/voltenv"],
       %r{VOLTDB_LIB=\$VOLTDB_HOME\/lib}, "VOLTDB_LIB=$VOLTDB_HOME/lib/voltdb"
```
MikeMcQuaid: In general it'd be great to file an upstream issue asking if they could provide another configuration mechanism than "run a search and replace" through their scripts.

Formula/voltdb.rb
patope: This way list is a bit easier to maintain. Because `inreplace` gives an error if replace fails and error points to script that needs to be checked on upgrade.

Formula/voltdb.rb
patope: Agree. Some of scripts have been converted to python and those work without tweaks. Guess it's their longer term target here.

Formula/voltdb.rb
MikeMcQuaid: Is there no way to autopopulate this list?

Formula/voltdb.rb
MikeMcQuaid: Ok, cool. Could you add a comment stating that?

MikeMcQuaid (changes requested):

MikeMcQuaid (approved):

MikeMcQuaid:

PR #5714, opened 2016-10-09
pazpar2 1.12.6

ssp:

The brew audit command complains about a missing test for the formula (»A `test do` test block should be added«). As I added a test to this formula a while back, it’s gone now, and (probably due to the repository split) couldn’t figure out yet when/why it was removed, can you advise on what may have been the reasons for that and whether I try to re-add the previously existing test?

### DIFF
--- a/Formula/pazpar2.rb
+++ b/Formula/pazpar2.rb
@@ -1,9 +1,8 @@
 class Pazpar2 < Formula
   desc "Metasearching middleware webservice"
   homepage "https://www.indexdata.com/pazpar2"
-  url "http://ftp.indexdata.dk/pub/pazpar2/pazpar2-1.12.5.tar.gz"
-  sha256 "a526ac9a91f6ac4683af3e27d58adb70755d82da7ad909eb98edf68dd07062d3"
-  revision 1
+  url "http://ftp.indexdata.dk/pub/pazpar2/pazpar2-1.12.6.tar.gz"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> As I added a test to this formula a while back, it’s gone now, and (probably due to the repository split) couldn’t figure out yet when/why it was removed, can you advise on what may have been the reasons for that and whether I try to re-add the previously existing test?

Thanks for noticing. Please try to re-add it and we'll help you get it passing 👍

ssp:
Thanks @MikeMcQuaid 

I did a bit of testing and it looks like the test fails because current pazpar2 versions crash when running in test mode. Perhaps that’s why the test was removed in https://github.com/Homebrew/legacy-homebrew/commit/02b9d8905f4fe26149501bbad86b920eaa8df383 .

The problem also seems to happen on Linux, so I guess it’s not the homebrew setup triggering it. I opened an issue https://github.com/indexdata/pazpar2/issues/12 on the pazpar2 project. Perhaps their team understands the problem causing this.

So I guess I’ll wait until that issue is fixed before re-adding the test to the formula.

Sven

MikeMcQuaid:
@ssp Would be good to add back a crappy test, if nothing else e.g. `pazpar2 -v` or similar.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@ssp thanks for updating this and taking the time to report the test problem upstream! :tada:

PR #5725, opened 2016-10-09
pyside: migrate to boneyard

MikeMcQuaid:


MikeMcQuaid:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

robpieke:
Is there somewhere I can see the background/motivation for this? I still use PySide quite a bit and was caught by surprise by this migration. I'm sure there's a great reason why it happened, so I'm very curious.

ilovezfs:
needs to build with qt5, then it can come back :)

robpieke:
Even though the vanilla Homebrew version of Qt is 4.8? Okay.

Don't get me wrong, I definitely look forward to the hopefully-very-soon future where everything is Qt5 friendly ... feel like we've been waiting a while for it.

So, rookie question then: is there a way to "brew install" from the boneyard, or should I be looking to download & build the PySide source completely outside of Homebrew now?

MikeMcQuaid:
> Even though the vanilla Homebrew version of Qt is 4.8? Okay.

We're removing the reverse dependencies with a plan to actively remove Qt 4 when we've done that. It doesn't build on 10.12 and never will and we're not going to try and fix that as upstream will never support it.

> Don't get me wrong, I definitely look forward to the hopefully-very-soon future where everything is Qt5 friendly ... feel like we've been waiting a while for it.

I get it's annoying but stuff like this is what's going to force people to adopt it.

> So, rookie question then: is there a way to "brew install" from the boneyard, or should I be looking to download & build the PySide source completely outside of Homebrew now?

You can still `brew install` from the boneyard but the formula is unsupported and we won't fix any issues you have when using it. https://github.com/cartr/homebrew-qt4 may be interesting to you and they may be willing to accept and maintain formulae we boneyard.

ilovezfs:
@robpieke it looks like PySide for Qt 5 is already a thing: https://wiki.qt.io/PySide2 so maybe that will make its way over to homebrew/core at some point.

robpieke:
Thanks for all the replies, much appreciated! :)

halfnibble:
Ok. So I have to maintain a legacy PySide app. It runs on Windows XP. (I know, my life sucks). How do I get the old PySide back? It needs to be for Qt 4.8. Thanks.

tdsmith:
> How do I get the old PySide back?

That's up to you; sorry! `brew info pyside` will describe how to view the formula.

PR #5727, opened 2016-10-09
aqbanking: migrate to boneyard

MikeMcQuaid:


zmwangx:
May I ask why? This one isn't broken or anything.

zmwangx:
It seems you're removing qt revdeps. If that's the case: `qt` is just an optional dep of `gwenhywfar`.

MikeMcQuaid:
@zmwangx of course you can 😁 https://github.com/Homebrew/homebrew-core/issues/1705 as it depends on `gwenhywfar` (No Qt5 support, even in latest 2016 beta)

zmwangx:
Yeah, I'm saying you can simply remove the optional `qt` dep of `gwenhywfar`.

MikeMcQuaid:
@zmwangx Yes, you're right. Not sure how I missed that?

PR #5733, opened 2016-10-09
gnuradio: make python optional

ilovezfs:
to avoid qt4 being a default dependency


MikeMcQuaid (approved):

ilovezfs:
@BrewTestBot test this please

Issue #5732, opened 2016-10-09
fpc not installing on macOS 10.12


antonshiryaev:
### Bug reports:

FreePascal (formula is "fpc") does not compile on macOS 10.12, reporting 10.12 as an invalid build target parameter.

Gist-log: https://gist.github.com/4de22f53767976bcbdde4a328c500df5
<img width="1178" alt="2016-10-09 17 06 19" src="https://cloud.githubusercontent.com/assets/5638404/19221119/8c1d3c72-8e45-11e6-8aa4-8d3e24c5518d.png">

zmwangx:
Reproducible, see #5488. I believe it's an upstream issue. They need to acknowledge 10.12 as a valid target.

zmwangx:
Edited title so that people don't go `brew info freepascal`.

zmwangx:
I think we should put a `MaximumMacOSRequirement` on `fpc` at the moment, and wait for an upstream release.

@antonshiryaev I would appreciate it if you can contact the upstream about the 10.12 situation. Please drop a link here if you open an issue on a public issue tracker or post to a mailing list with an archive.

antonshiryaev:
I just checked and fpc installed from the official image works flawlessly on 10.12. No update to the actual code is necessary.

By upstream do you mean fpc? Because that's almost certainly abandonware.

zmwangx:
> fpc installed from the official image works flawlessly on 10.12

The dmg was built on an earlier macOS release. The build products could certainly work on 10.12; it's compiling on 10.12 that's a problem.

But I'll see if it's easy to patch. Give me some time.

> Because that's almost certainly abandonware.

Last (major) release in November 2015; why are you convinced it's abandonware?

antonshiryaev:
I see. 
Actually, I now believe I was wrong about the project's abandonment. I re-checked, and the latest snapshot release is dated 20.09.16. I'm trying to register at their bug tracker and will post a link here once I report the bug.

zmwangx:
Okay, I looked into it, it's the bootstrap binary that's causing the trouble. Unless there's another way to build it that doesn't involve the bootstrap (I bet there is), I doubt we can rescue it without upstream involvement.

> I'm trying to register at their bug tracker and will post a link here once I report the bug.

Cool, thanks.

antonshiryaev:
[...]

PR #5741, opened 2016-10-09
fpc: fix build on 10.12

ilovezfs:
bootstrap on Sierra with the El Capitan bottle

also, set deployment target to 10.9 for El Capitan and Yosemite, since
the upstream bootstrapper doesn't understand > 10.9

### DIFF
--- a/Formula/fpc.rb
+++ b/Formula/fpc.rb
@@ -13,15 +13,28 @@ class Fpc < Formula
   end
 
   resource "bootstrap" do
-    url "https://downloads.sourceforge.net/project/freepascal/Bootstrap/2.6.4/universal-macosx-10.5-ppcuniversal.tar.bz2"
-    sha256 "e7243e83e6a04de147ebab7530754ec92cd1fbabbc9b6b00a3f90a796312f3e9"
+    if MacOS.version < :sierra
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

antonshiryaev:
Just changing the deployment target appears to have been sufficient, fpc installed fine on my machine with the fix posted [here](https://github.com/Homebrew/homebrew-core/issues/5732#issuecomment-252497443).

ilovezfs:
There is no Homebrew DSL for requiring the CLT, nor should there be, so you should encourage them to fix it so that it can work with Xcode alone.

antonshiryaev:
This is not unique to Homebrew, if it's installed from the disk image it still requires the CLT.

There are a lot of oddities to the project. For instance, [here](http://bugs.freepascal.org/changelog_page.php?version_id=342) the reported latest version is 3.1.1, and [here](http://www.freepascal.org/download.var) it's 3.0.0, which is what's installed by Homebrew. I doubt I'll be able to encourage them to do much, but I'll try.

ilovezfs:
>if it's installed from the disk image it still requires the CLT.

that sounds like a bug too.

antonshiryaev:
It probably is. Unfortunately, it's a safe assumption that new releases are not coming. The maintainers apparently can't even be bothered to update the Downloads page.

Issue #5762, opened 2016-10-10
How to handle g_get_user_config_dir?


MikeLund:
Hi,

I am trying to update the bottle for apt-dater to the most recent version. The problem I am running into is that the code uses [g_get_user_config_dir](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-user-config-dir), which relates to XDG_CONFIG_HOME or something.

The problem is, it will check in ~/.config/apt-dater for program configs, because that's what g_get_user_config_dir returns. But the configs are already existing in /usr/local/etc/apt-dater/ (which links to ../../Cellar/apt-dater/1.0.3/etc/apt-dater/).

How is the best way to handle this? It seems like there must be a normal way to handle this (other than doing stuff with the code), because this is a semi-common issue where programs like to throw stuff in ~/.config, but still has config elsewhere (in the --prefix).

Thanks for any advise!

MikeMcQuaid:
We override `$HOME` during the build so it doesn't matter if it checks there during installation. At runtime it actively should check in `~/.config` too.

PR #5763, opened 2016-10-10
apt-dater 1.0.3

MikeLund:
All the checkmarks should be good. But please read https://github.com/Homebrew/homebrew-core/issues/5762 before accepting. If everything is fine in regards to that link, then it should be fine to accept. Otherwise putting this on hold. Thanks!

### DIFF
--- a/Formula/apt-dater.rb
+++ b/Formula/apt-dater.rb
@@ -1,8 +1,11 @@
 class AptDater < Formula
   desc "Manage package updates on remote hosts using SSH"
   homepage "https://www.ibh.de/apt-dater/"
-  url "https://downloads.sourceforge.net/project/apt-dater/apt-dater/0.9.0/apt-dater-0.9.0.tar.gz"
-  sha256 "10b8335c156dabae249aa071cf8689900fae325c2e9540e36a840a2a77d3eeb4"
+  url "https://github.com/DE-IBH/apt-dater.git",
+      :tag => "v1.0.3",
+      :revision => "fdad3b43e50bc3582e391d6feabae150e133b3b3"
+
+  head "https://github.com/DE-IBH/apt-dater.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

Formula/apt-dater.rb
```diff
@@ -1,8 +1,11 @@
 class AptDater < Formula
   desc "Manage package updates on remote hosts using SSH"
   homepage "https://www.ibh.de/apt-dater/"
-  url "https://downloads.sourceforge.net/project/apt-dater/apt-dater/0.9.0/apt-dater-0.9.0.tar.gz"
-  sha256 "10b8335c156dabae249aa071cf8689900fae325c2e9540e36a840a2a77d3eeb4"
+  url "https://github.com/DE-IBH/apt-dater.git",
+      :tag => "v1.0.3",
+      :revision => "fdad3b43e50bc3582e391d6feabae150e133b3b3"
```
MikeMcQuaid: Can this build from a archive tarball rather than a Git repository?

MikeLund: Maybe can use https://github.com/DE-IBH/apt-dater/releases, but why though? I thought git repo is better for Homebrew.

MikeLund:
Actually is probably best regardless to close without merging.
The config file syntax has completely changed, and it's quite confusing with the numerous-config-files thing. So maybe not very user-friendly to merge :/

MikeMcQuaid:
@MikeLund That's fine, we'd rather have the newest version.

MikeLund:
Sorry for time waste but I regret this and think it should not continue; current version is a bit too buggy, old version was fine and stable, this adds nothing. The developer also seems to be away, and his homepage doesn't even mention program anymore, so it's a bit dead in the water.
Will retry once everything is going good for them again.

MikeMcQuaid:
No worries, thanks for letting us know.

PR #5765, opened 2016-10-10
ice: 3.6.3

externl:

---

### DIFF
--- a/Formula/ice.rb
+++ b/Formula/ice.rb
@@ -1,8 +1,8 @@
 class Ice < Formula
   desc "Comprehensive RPC framework"
   homepage "https://zeroc.com"
-  url "https://github.com/zeroc-ice/ice/archive/v3.6.2.tar.gz"
-  sha256 "5e9305a5eb6081c8f128d63a5546158594e9f115174fc91208f645dbe2fc02fe"
+  url "https://github.com/zeroc-ice/ice/archive/v3.6.3.tar.gz"
+  sha256 "82ff74e6d24d9fa396dbb4d9697dc183b17bc9c3f6f076fecdc05632be80a2dc"
 
   bottle do
     sha256 "ef487cbf3e812e9c3f1fefd704ced9c946e3c64cfdb252e29b8d406dafb54336" => :sierra
@@ -58,6 +58,7 @@ class Ice < Formula
     args = %W[
       prefix=#{prefix}
       USR_DIR_INSTALL=yes
+      SLICE_DIR_SYMLINK=yes
       OPTIMIZE=yes
       DB_HOME=#{libexec}
       MCPP_HOME=#{Formula["mcpp"].opt_prefix}


ilovezfs:
@BrewTestBot test this please

Formula/ice.rb
```diff
@@ -58,6 +58,7 @@ def install
     args = %W[
       prefix=#{prefix}
       USR_DIR_INSTALL=yes
+      SLICE_DIR_SYMLINK=yes
```
ilovezfs: what is this?

externl: The build system changed how Ice's "slice" files are installed. We (I'm an Ice developer) changed the behaviour so that by default there is a now globally installed directory for all users to install their slice files (`/PREFIXPATH/share/slice`). Previously it was just a simlink to the slice folder (which is installed in a different location). 

Since Homebrew linking creates simlinks, and only special formula can create directories (bash, fish, etc.),  this uses the "old" logic and just creates a simlink.

ilovezfs: Hmmm.... if you need non-symlinked directories, usually people use `#{var}`, `#{var}/run`, `#{var}/foo`, `#{var}/run/foo`, `#{etc}`, or `#{etc}/foo`, since `#{var}` is `#{HOMEBREW_PREFIX}/var` not `#{prefix}/var` (i.e., it's `/usr/local/var` not `/usr/local/Cellar/foo/1.2.3/var`) and `#{etc}` is `#{HOMEBREW_PREFIX}/etc`, not `#{prefix}/etc` (i.e., it's `/usr/local/etc` not `/usr/local/Cellar/foo/1.2.3/etc`).

externl: Think more along the lines of what shells do. 

eg.
* bash completions go in `/usr/local/share/bash-completion`
* fish functions/completions go in `/usr/local/share/fish`
* zsh functions/completions go in `/usr/local/share/zsh`

Ice's slice files fall under the same idea. We don't "need" it to be a folder, its just a convenience on other platforms that allow this (rpm, apt, etc.). Since there are no other brew formula (that I know of) which install slice files, then just creating this as a symlink is perfectly sufficient.

ilovezfs: Would it be possible to put the `slice` symlink in `#{pkgshare}` (i.e., `#{share}/ice`) instead of `#{share}` since if we ever had a formula named `slice` you'd collide with its `pkgshare`?

externl: Unfortunately no, not for a patch release. User applications and user build systems would rely on this location. We would be happy to change this behaviour for the next major release, 3.7.

ilovezfs: @externl no problem. I installed locally to make sure I understood what you meant, and this looks like a reasonable solution for now.

ilovezfs:
@externl Thanks for your contribution to Homebrew, and being such an engaged upstream! We really appreciate it.

Just for future reference, the commit subjects for version bumps are standardized as follows:
```
==> Applying patch
Applying: ice: 3.6.3
==> Patch closes issue #5765
Warning: Nonstandard bump subject: ice: 3.6.3
Warning: Subject should be: ice 3.6.3
```

PR #5786, opened 2016-10-11
qscintilla2: fix build with Xcode 8

ilovezfs:
CC @schoeps @justinclift @dpo

### DIFF
--- a/Formula/qscintilla2.rb
+++ b/Formula/qscintilla2.rb
@@ -30,6 +30,15 @@ class Qscintilla2 < Formula
     depends_on "pyqt5"
   end
 
+  # Fix build with Xcode 8 "error: implicit instantiation of undefined template"
+  # Reported 7 Oct 2016 https://www.riverbankcomputing.com/pipermail/qscintilla/2016-October/001160.html
+  if DevelopmentTools.clang_build_version >= 800
+    patch do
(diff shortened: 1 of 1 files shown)


justinclift:
As a note, upstream has completely ignore the bug report about this on their mailing list:

&nbsp; &nbsp; https://www.riverbankcomputing.com/pipermail/qscintilla/2016-October/001160.html

Looking through their archives, they seem to ignore 95% of everything.  Would recommend steering clear of QScintilla2 for new projects. 😦

ilovezfs:
@justinclift is it possible there's some other mechanism for contacting/reporting other than the mailing list?

justinclift:
Their contact page directs people to the mailing list, and also says:

> Unless you are a commercial customer, do not email Riverbank directly.

So, I'm thinking no?

ilovezfs:
```
You are buying 1 PyQt commercial license at £350.00.
```

If they're going to ignore anyone who hasn't given them $429.00, then it sounds like the project needs to be forked or not used. I suspect it's worth emailing support@riverbankcomputing.com and hoping they don't look a gift horse in the mouth since macOS is one of their officially supported platforms for paying customers.

justinclift:
Hmmm, I'll give it a few more days before doing that.  Just in case.

ilovezfs:
Yup, no problem.

PR #5804, opened 2016-10-11
ironcli fix build under sandbox

rubenfonseca:
Force GLIDE_HOME to be outside home, so it builds inside the sandbox. Without this change the build fails every time because glide tries to write on the $USER home dir, which is forbidden inside the sandbox.

### DIFF
--- a/Formula/ironcli.rb
+++ b/Formula/ironcli.rb
@@ -26,6 +26,7 @@ class Ironcli < Formula
 
   def install
     ENV["GOPATH"] = buildpath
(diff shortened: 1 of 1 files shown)


ilovezfs:
We've actually been moving these to the cache.
```
ENV["GLIDE_HOME"] = HOMEBREW_CACHE/"glide_home/#{name}"
```

rubenfonseca:
@ilovezfs thank you! updated PR.

MikeMcQuaid:

PR #5870, opened 2016-10-13
QuestDB formula

bluestreak01:

---

New formula for brew installation of QuestDB

### DIFF
--- a/Formula/questdb.rb
+++ b/Formula/questdb.rb
@@ -1,0 +1,64 @@
+class Questdb < Formula
+  desc "Time Series Database"
+  homepage "https://www.questdb.org"
+  url "https://www.questdb.org/download/questdb-1.0.0-SNAPSHOT-20161013-1304-bin.tar.gz"
+  sha256 "66f4d43868e8d0eb61606712a25f397023c82e368bbcbfda7cc007611661984b"
+  bottle :unneeded
+  depends_on :java => "1.7+"
+
+  def datadir
+    var/"questdb"
+  end
+
+  def install
+    rm_rf "questdb.exe"
+    prefix.install Dir["*"]
+    bin.install_symlink "#{prefix}/questdb.sh" => "questdb"
+  end
+
+  plist_options :manual => "questdb start"
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>KeepAlive</key>
+        <dict>
+          <key>SuccessfulExit</key>
+          <false/>
+        </dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>ProgramArguments</key>
+        <array>
+          <string>#{opt_bin}/questdb</string>
+          <string>start</string>
+          <string>-d</string>
+          <string>#{datadir}</string>
+          <string>-n</string>
+        </array>
+        <key>RunAtLoad</key>
+        <true/>
+        <key>WorkingDirectory</key>
+        <string>#{var}</string>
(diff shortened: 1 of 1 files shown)


Formula/questdb.rb
```diff
@@ -0,0 +1,64 @@
+class Questdb < Formula
+  desc "Time Series Database"
+  homepage "https://www.questdb.org"
+  url "https://www.questdb.org/download/questdb-1.0.0-SNAPSHOT-20161013-1304-bin.tar.gz"
```
MikeMcQuaid: This doesn't look like a stable release: https://github.com/bluestreak01/questdb/releases. We need to use one of those.

Formula/questdb.rb
```diff
@@ -0,0 +1,64 @@
...
+  sha256 "66f4d43868e8d0eb61606712a25f397023c82e368bbcbfda7cc007611661984b"
+  bottle :unneeded
```
MikeMcQuaid: Add a newline above this line.

Formula/questdb.rb
```diff
@@ -0,0 +1,64 @@
...
+  depends_on :java => "1.7+"
```
MikeMcQuaid: Add a newline above this line.

Formula/questdb.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def datadir
+    var/"questdb"
```
MikeMcQuaid: This is only used in one place so hardcode it instead.

Formula/questdb.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def install
+    rm_rf "questdb.exe"
+    prefix.install Dir["*"]
```
MikeMcQuaid: Might be worth using `libexec` instead; that's what we generally do.

Formula/questdb.rb
```diff
@@ -0,0 +1,64 @@
...
+          <string>#{opt_bin}/questdb</string>
+          <string>start</string>
+          <string>-d</string>
+          <string>#{datadir}</string>
+          <string>-n</string>
+        </array>
+        <key>RunAtLoad</key>
+        <true/>
+        <key>WorkingDirectory</key>
+        <string>#{var}</string>
```
MikeMcQuaid: Feels like this should be a subdirectory or not specified.

Formula/questdb.rb
```diff
@@ -0,0 +1,64 @@
...
+          <integer>1024</integer>
+        </dict>
+      </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    system "#{bin}/questdb", "start"
+    sleep 2
```
MikeMcQuaid: Add a test to verify you can e.g. connect to the process and an `ensure` to kill it if anything fails.

MikeMcQuaid (changes requested):

bluestreak01:
Thanks a lot for the review Mike, I'll make suggested changes are re-submit.

Which directory can I use on osx to store data. I noticed that other projects use /usr/local/var and it worked on my osx, but jenkins failed with permission problems. Would you be able to advise?

Formula/questdb.rb
bluestreak01: Can we let this pull thru with snapshot version? I realise we need a release, but i'm not quite ready with all of the testing.

MikeMcQuaid: We cannot, sorry.

bluestreak01: included release

Formula/questdb.rb
bluestreak01: done

Formula/questdb.rb
bluestreak01: done

Formula/questdb.rb
bluestreak01: done

Formula/questdb.rb
bluestreak01: fixed with non-generic directory

PR #5877, opened 2016-10-13
qscintilla2: better Xcode 8 patch

ilovezfs:
patch is the work of Alan Garny

CC @agarny @justinclift @schoeps


schoeps:
I would do tests but currently the qt5-support of octave and gnuplot are broken due to the 5.6.2 update :(

ilovezfs:
>I would do tests but currently the qt5-support of octave and gnuplot are broken due to the 5.6.2 update :( 

That's probably my removing the pkg-config files hack, if I had to guess.
Here are two examples of how to fix the resulting problems:

https://github.com/Homebrew/homebrew-core/commit/ba7b72ba18bffa5251aa20241315935a9a4b889a
https://github.com/Homebrew/homebrew-core/commit/55329433e7ad37a7f38fb828559dd2d82bdeaa7a

We're considering switching to a `-no-framework` build, which will bring back the pkg-config files without needing a hack. But in the meantime mitigation like above should work fine.

PR #5878, opened 2016-10-13
Add comment about Yarn URL

Daniel15:
Hi!

The tarball for Yarn should be downloaded from its official website rather than from npm. Additionally, our current official release is 0.15.0... 0.15.1 was just a patch to the npm package and is not a "real" release (it's just specific to npm)

Unfortunately I don't have a Mac so I can't test this myself.

@rudineirk

### DIFF
--- a/Formula/yarn.rb
+++ b/Formula/yarn.rb
@@ -4,7 +4,7 @@ class Yarn < Formula
   desc "Javascript package manager"
   homepage "https://yarnpkg.com/"
   url "https://yarnpkg.com/downloads/0.15.0/yarn-v0.15.0.tar.gz"
-  sha256 "93A8083BD4989D3F7C05FCF57DFE232E00A35A09E48354B9316D2BC43F74B51B"
+  sha256 "93a8083bd4989d3f7c05fcf57dfe232e00a35a09e48354b9316d2bc43f74b51b"
   head "https://github.com/yarnpkg/yarn.git"
 
   bottle do
--- a/Formula/yarn.rb
+++ b/Formula/yarn.rb
@@ -3,8 +3,13 @@ require "language/node"
 class Yarn < Formula
   desc "Javascript package manager"
   homepage "https://yarnpkg.com/"
-  url "https://yarnpkg.com/downloads/0.15.0/yarn-v0.15.0.tar.gz"
-  sha256 "93a8083bd4989d3f7c05fcf57dfe232e00a35a09e48354b9316d2bc43f74b51b"
+  # Note: If updating this to a newer version, please change the URL to the official 
+  # yarnpkg.com download URL:
+  # url "https://yarnpkg.com/downloads/0.15.0/yarn-v0.15.0.tar.gz"
+  # This was not updated yet as Yarn 0.15.1 was an npm-specific release and is not available
+  # as a standalone tarball.
+  url "https://registry.npmjs.org/yarn/-/yarn-0.15.1.tgz"
+  sha256 "f99fd587e84987909d5f9e918b8fe524349fdc548e5bc5c380c8f8c0a70c6b87"
   head "https://github.com/yarnpkg/yarn.git"
 
   bottle do
--- a/Formula/yarn.rb
+++ b/Formula/yarn.rb
@@ -3,8 +3,8 @@ require "language/node"
 class Yarn < Formula
   desc "Javascript package manager"
   homepage "https://yarnpkg.com/"
-  url "https://registry.npmjs.org/yarn/-/yarn-0.15.1.tgz"
-  sha256 "f99fd587e84987909d5f9e918b8fe524349fdc548e5bc5c380c8f8c0a70c6b87"
+  url "https://yarnpkg.com/downloads/0.15.0/yarn-v0.15.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/yarn.rb
```diff
@@ -3,8 +3,8 @@
 class Yarn < Formula
   desc "Javascript package manager"
   homepage "https://yarnpkg.com/"
-  url "https://registry.npmjs.org/yarn/-/yarn-0.15.1.tgz"
-  sha256 "f99fd587e84987909d5f9e918b8fe524349fdc548e5bc5c380c8f8c0a70c6b87"
+  url "https://yarnpkg.com/downloads/0.15.0/yarn-v0.15.0.tar.gz"
+  sha256 "93A8083BD4989D3F7C05FCF57DFE232E00A35A09E48354B9316D2BC43F74B51B"
```
MikeMcQuaid: Needs to be lowercase: 93a8083bd4989d3f7c05fcf57dfe232e00a35a09e48354b9316d2bc43f74b51b

MikeMcQuaid (changes requested):

Formula/yarn.rb
Daniel15: Thanks, I adjusted it 😄

zmwangx:
I don't think we permit downgrades. Is there any actual harm to using npm 0.15.1 version? If not, it's probably better to correct it when it's time for the next release.

Daniel15:
> Is there any actual harm to using npm 0.15.1 version?

There's no harm, but any future releases should use the proper download links. The npm version of yarn is just a fallback and should not be used as the primary download source for cases where the regular tarball suffices (like here).

I assume the formulae allow comments? Should I just add a comment explaining to the next person that updates the formula that they should switch it to the proper URL?

Formula/yarn.rb
```diff
@@ -3,8 +3,8 @@
 class Yarn < Formula
   desc "Javascript package manager"
   homepage "https://yarnpkg.com/"
-  url "https://registry.npmjs.org/yarn/-/yarn-0.15.1.tgz"
-  sha256 "f99fd587e84987909d5f9e918b8fe524349fdc548e5bc5c380c8f8c0a70c6b87"
+  url "https://yarnpkg.com/downloads/0.15.0/yarn-v0.15.0.tar.gz"
```
MikeMcQuaid: Any objections to making this https://github.com/yarnpkg/yarn/releases/download/v0.15.0/yarn-v0.15.0.tar.gz? It'll make it easier for us to query for new versions. Also, as @zmwangx pointed out this needs to be 0.15.1. It may be worth using the auto-generated GitHub tarball from the tag if that's going to make that easier: https://github.com/yarnpkg/yarn/archive/v0.15.1.tar.gz

MikeMcQuaid (changes requested):

Formula/yarn.rb
Daniel15: @MikeMcQuaid What do you mean by "It'll make it easier for us to query for new versions"? The URL format is basically the same. We don't want to tightly-couple Yarn releases with GitHub's system, hence the yarnpkg.com URL.

You can get the latest Yarn version number from https://yarnpkg.com/latest-version, would that help in any way?

> It may be worth using the auto-generated GitHub tarball from the tag if that's going to make that easier

This won't work as it has the source code, not the compiled version.

Daniel15:
Okay, seeing as Homebrew needs to continue using 0.15.1, I replaced my change with a comment explaining that the URL should be changed in the future, once we release a new version of Yarn.

MikeMcQuaid:
> This won't work as it has the source code, not the compiled version.

@Daniel15 Can you elaborate a bit more on the compiled version stuff; I'm not familiar with the NodeJS ecosystem. Does that mean it provides e.g. compiled C-like code in binary form or that it provides e.g. minified JS?

Daniel15:
> Can you elaborate a bit more on the compiled version stuff

Sure! Sorry my previous reply was so terse. 

Yarn uses newer JavaScript features such as [async/await](https://www.twilio.com/blog/2015/10/asyncawait-the-hero-javascript-deserved.html). Node.js does not yet support these language features, so this code can't be executed in Node.js directly. Yarn (like many other JavaScript projects) uses a tool called [Babel](https://babeljs.io/) to transform the JavaScript to something that Node.js can understand, and the resulting code is what gets executed by Node.js.

The code in the Yarn repository is the original source code, while the code in the installer / Linux packages / tarball / npm is the compiled code that results from running the Yarn files through Babel (the original source code is not installed on end user's machines, just the compiled code). If you download the code from Github (either by cloning the repo or downloading a tarball of a tag), you'll just get the original source code.

MikeMcQuaid:
Gotcha, thanks @Daniel15. One last question: out of interest why is the latest version not provided on the homepage?

Daniel15:
>  out of interest why is the latest version not provided on the homepage?

Version 0.15.1 was just a patch to the npm release and the installer script, to fix a last minute issue. Every other release package (tarball, Debian/Ubuntu package, RPM package, Windows installer) is still at version number 0.15.0. Most likely next week we'll release 0.15.2 or 0.16.0 and the version numbers will all be consistent again.

MikeMcQuaid:
@Daniel15 Ok, thanks!

MikeMcQuaid (approved):

MikeMcQuaid:

PR #5886, opened 2016-10-13
jemalloc: revision, make Sierra patches unconditional

mistydemeo:

---

The Sierra patches only affects Sierra, but the bug is a particularly severe one that can be hard to diagnose. I've encountered a number of users who upgraded from previous versions of Sierra and then didn't have the ability to figure out what broke, or that jemalloc specifically was what they needed to reinstall. Since we usually tell users that they don't need to reinstall all their packages after upgrading to Sierra, a lot of users have broken jemallocs after the OS upgrade.

Since these patches are upstream already, I'd like to just include them unconditionally and bump the revision so that people on older OSs get them before they upgrade.


MikeMcQuaid (approved):
Makes sense to me 👍

MikeMcQuaid:
Cheers @mistydemeo!

PR #5888, opened 2016-10-14
mysql: fix runtime binding with Xcode 8 builds on macOS 10.11

JCount:

---

fixes lazy symbol binding runtime error on macOS 10.11 with the Xcode 8 SDK

```
Symbol not found: _clock_gettime
dyld::fastBindLazySymbol(ImageLoader**, unsigned long)
```

### DIFF
--- a/Formula/mysql.rb
+++ b/Formula/mysql.rb
@@ -60,10 +60,6 @@ class Mysql < Formula
 
     # -DINSTALL_* are relative to `CMAKE_INSTALL_PREFIX` (`prefix`)
     args = %W[
-      .
-      -DCMAKE_INSTALL_PREFIX=#{prefix}
-      -DCMAKE_FIND_FRAMEWORK=LAST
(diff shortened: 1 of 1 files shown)


JCount:
This should probably be reported upstream, if it hasn't been already (I could not find a report anywhere). However, I have no idea how to file a bug report with upstream in this case, so it should perhaps be someone other than me...

ilovezfs:
The correct fix here is to make it use cmake std args which always sets this https://github.com/Homebrew/brew/pull/1078

JCount:
Thank you, will do. I missed that addition, it's nice that there is now a neat, standard way to do this. (in most cases)

JCount:
This does not now need a revision bump, correct?

MikeMcQuaid:
It shouldn't do, correct. Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #5890, opened 2016-10-14
stackedit 4.3.14 (new formula)

bqbn:

---

### DIFF
--- a/Formula/stackedit.rb
+++ b/Formula/stackedit.rb
@@ -1,0 +1,61 @@
+require "language/node"
+
+class Stackedit < Formula
+  desc "In-browser markdown editor"
+  homepage "https://github.com/benweet/stackedit"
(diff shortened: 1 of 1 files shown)


Formula/stackedit.rb
```diff
@@ -0,0 +1,61 @@
...
+    EOS
+  end
+
+  test do
+    pid = fork do
+      exec "#{bin}/stackedit"
+    end
+    sleep 5
+    begin
+      assert_match %r{HTTP/1.1 200 OK}, shell_output("curl -I localhost:3000/")
```
bfontaine: We should choose a random unused port instead. This test will fail if one is already running something else on that port.

bqbn:
Thanks @bfontaine ! That makes sense. However, I'm going to withdraw this PR for now because some other bug has surfaced that requires a little more time to fix.

PR #5920, opened 2016-10-14
wine 1.9.21 (devel)

tobbi:

---

### DIFF
--- a/Formula/wine.rb
+++ b/Formula/wine.rb
@@ -27,9 +27,9 @@ class Wine < Formula
   end
 
   devel do
-    url "https://dl.winehq.org/wine/source/1.9/wine-1.9.20.tar.bz2"
-    mirror "https://downloads.sourceforge.net/project/wine/Source/wine-1.9.20.tar.bz2"
-    sha256 "2769f30642329916f12696f7be225644d28fc54d1b9181c979381d70b9a279df"
+    url "https://dl.winehq.org/wine/source/1.9/wine-1.9.21.tar.bz2"
+    mirror "https://downloads.sourceforge.net/project/wine/Source/wine-1.9.21.tar.bz2"
+    sha256 "8b6dd027021be9dd646d65bd19c0334c36dcf9403d8ed4a6e9460b6703efade1"
   end
 
   # note that all wine dependencies should declare a --universal option in their formula,


AlexanderNeilson:
Hi Team

This appears to be Jenkins terminating the build after 61 Minutes on Sierra.

I don't usually see wine take that long to build on my machines however I see El Cap also built in 60 minutes and succeeded so possibly the general spare cycles on the Jenkins test box.

One part of this may be to either find out a reason for slow compilation to try avoid hitting the timeout, that should help avoid a few delays.

The second thing seems to be that despite this change being to the devel part of the formula (not stable) Jenkins still built stable. Critically:
* Is there something we could code / change to allow Jenkins to build the changed part of the formula and not always stable?
* What changed between this and the other day when this formula (stable) built correctly and was accepted on your previous pull request?

Regards
Alexander

MikeMcQuaid:
> Is there something we could code / change to allow Jenkins to build the changed part of the formula and not always stable?

I think we always want to build both as development changes can often break those in stable.

> What changed between this and the other day when this formula (stable) built correctly and was accepted on your previous pull request?

Probably just entropy. I've requeued the CI.

tobbi:
Requeueing didn't seem to have worked. It's still aborting after roughly an hour.

AlexanderNeilson:
Can an attempt be tried either with a longer timeout in the system doing a hard kill in general? Or a single time override for this build.

Apologies Mike, I may have misunderstood the build process, just seeing the line / residual building stable but not seeing the devel step. (My worry was around not catching failures with the devel). But this could also explain the longer than I expected build time.

It does appear that the issue is a timeout of some overall time limit in the Jenkins build system killing it. Obviously there could be a risk with extending systemwide timeout for builds and would be great to have the builds complete faster but also there could be an issue with managing individual exceptions / overrides if there is even an option to extend the timeout on a job by job basis.

Regards
Alexander

MikeMcQuaid:
> overrides if there is even an option to extend the timeout on a job by job basis.

I've requeued this with a longer timeout CC @ilovezfs

tobbi:
This appears to have worked. Can this be merged?

MikeMcQuaid:
Thanks again for your contribution to Homebrew and sorry for the delay! Without people like you submitting PRs we couldn't run this project. You rock!

PR #5936, opened 2016-10-15
typescript 2.0.3 (new formula)

cglong:

---

### DIFF
--- a/Formula/typescript.rb
+++ b/Formula/typescript.rb
@@ -24,7 +24,7 @@ class Typescript < Formula
       var test = new Test();
       document.body.innerHTML = test.greet();
     EOS
-    
+
     system bin/"tsc", "test.ts"
     touch testpath/"test.js", :nocreate => true
   end
--- a/Formula/typescript.rb
+++ b/Formula/typescript.rb
@@ -1,0 +1,31 @@
+require "language/node"
+
+class Typescript < Formula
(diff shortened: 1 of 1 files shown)


Formula/typescript.rb
```diff
@@ -0,0 +1,31 @@
...
+        greet() {
+          return "Hello, world!";
+        }
+      };
+      var test = new Test();
+      document.body.innerHTML = test.greet();
+    EOS
+
+    system bin/"tsc", "test.ts"
+    touch testpath/"test.js", :nocreate => true
```
MikeMcQuaid: Why touch after the test was run?

MikeMcQuaid (changes requested):

Formula/typescript.rb
cglong: It's essentially verifying that `test.ts` was properly compiled and that `tsc` outputted `test.js` as expected. The `:nocreate` ensures that if `test.js` isn't present, an exception is thrown. This can easily be removed though if it's not considered a necessary validation 😄

MikeMcQuaid: Ah, ok, that wasn't clear. How about `assert File.exist?("test.js"), "test.js wasn't created"` or similar?

MikeMcQuaid (approved):

MikeMcQuaid:

PR #5940, opened 2016-10-15
uncrustify 0.64

ilovezfs:
use GitHub tarball
build system switched to CMake


dunn:
Neat:
```
==> ./configure --disable-dependency-tracking --prefix=/usr/local/Cellar/uncrustify/0.64
Failed to execute: ./configure
```
Maybe needs to be autoconf'd?

ilovezfs:
@dunn they switched to CMake.

dunn:
Oh, see, I can't read.

PR #5942, opened 2016-10-15
osm2pgsql: revision for boost

ilovezfs:
Build failure reported upstream here: https://github.com/openstreetmap/osm2pgsql/issues/628


ilovezfs:
caused by having lua51

PR #5950, opened 2016-10-15
ooniprobe 2.1.0

hellais:

### DIFF
--- a/Formula/ooniprobe.rb
+++ b/Formula/ooniprobe.rb
@@ -214,6 +214,10 @@ class Ooniprobe < Formula
        <true/>
      <key>KeepAlive</key>
        <true/>
+     <key>StandardErrorPath</key>
+       <string>/dev/null</string>
+     <key>StandardOutPath</key>
+       <string>/dev/null</string>
      <key>WorkingDirectory</key>
        <string>#{var}/log/</string>
    </dict>
--- a/Formula/ooniprobe.rb
+++ b/Formula/ooniprobe.rb
@@ -15,7 +15,7 @@ class Ooniprobe < Formula
   depends_on "geoip"
   depends_on "libdnet"
   depends_on "libyaml"
-  depends_on "openssl@1.1"
+  depends_on "openssl"
   depends_on "tor"
   depends_on :python
 
--- a/Formula/ooniprobe.rb
+++ b/Formula/ooniprobe.rb
@@ -219,7 +219,7 @@ class Ooniprobe < Formula
      <key>StandardOutPath</key>
        <string>/dev/null</string>
      <key>WorkingDirectory</key>
-       <string>#{var}/log/</string>
+       <string>#{prefix}</string>
    </dict>
    </plist>
    EOS
--- a/Formula/ooniprobe.rb
+++ b/Formula/ooniprobe.rb
@@ -3,8 +3,8 @@ class Ooniprobe < Formula
 
   desc "Network interference detection tool"
   homepage "https://ooni.torproject.org/"
-  url "https://pypi.python.org/packages/2e/5c/7fd90f852e0e4ed44e7740925326bb1798f908321b29394803755f5bef31/ooniprobe-2.0.1.tar.gz"
-  sha256 "cc4398716f7d7bb5806f8e0f25d3ec2ad36f7c611bd4506b21fb847da6120542"
+  url "https://pypi.python.org/packages/ea/31/f6ffff21b406b5c38b78f00b38897a48baae30eef0469089fb1330d86218/ooniprobe-2.1.0.tar.gz"
+  sha256 "867f51fcee8d84f68c42dea24c3384736a1e0cf153e1f18254e91682ca6e927a"
 
   bottle do
     sha256 "ab5aa52408ae04f8b8e546efe1cc0516c75ec1ca78a5199f996fe3822fe4acdb" => :sierra
@@ -200,41 +200,41 @@ class Ooniprobe < Formula
    <dict>
      <key>Label</key>
        <string>#{plist_name}</string>
-     <key>Program</key>
-       <string>#{opt_bin}/ooniprobe</string>
+     <key>EnvironmentVariables</key>
+     <dict>
+       <key>PATH</key>
+       <string>#{HOMEBREW_PREFIX}/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+     </dict>
      <key>ProgramArguments</key>
      <array>
-       <string>-i</string>
-       <string>#{pkgshare}/current.deck</string>
+       <string>#{opt_bin}/ooniprobe-agent</string>
+       <string>run</string>
      </array>
      <key>RunAtLoad</key>
-       <false/>
+       <true/>
      <key>KeepAlive</key>
-       <false/>
-     <key>StandardErrorPath</key>
-       <string>/dev/null</string>
-     <key>StandardOutPath</key>
-       <string>/dev/null</string>
-     <key>StartCalendarInterval</key>
-     <dict>
-       <key>Hour</key>
-       <integer>00</integer>
-       <key>Minute</key>
-       <integer>00</integer>
-     </dict>
+       <true/>
+     <key>WorkingDirectory</key>
+       <string>#{var}/log/</string>
    </dict>
    </plist>
    EOS
   end
 
   test do
-    (testpath/".ooni").mkpath
-    ooniconf = testpath/".ooni/ooniprobe.conf"
-    cp pkgshare/"ooniprobe.conf.sample", ooniconf
-    inreplace ooniconf, "/var/log", testpath
-    touch testpath/".ooni/initialized"
+    mkdir_p "#{testpath}/ooni/var_lib"
+    mkdir_p "#{testpath}/ooni/etc"
 
-    (testpath/"hosts.txt").write "github.com:443\n"
-    system bin/"ooniprobe", "-n", "blocking/tcp_connect", "-f", testpath/"hosts.txt"
+    (testpath/"ooni/settings.ini").atomic_write <<-EOS.undent
+      [directories]
+      usr_share = #{pkgshare}
+      var_lib = #{testpath}/ooni/var_lib
+      etc = #{testpath}/ooni/etc
+    EOS
+    (testpath/"ooni/var_lib/initialized").write ""
+
+    (testpath/"ooni/hosts.txt").write "github.com:443\n"
+    ENV["OONIPROBE_SETTINGS"] = "#{testpath}/ooni/settings.ini"
+    system bin/"ooniprobe", "-ng", "blocking/tcp_connect", "-f", testpath/"ooni/hosts.txt"
   end
 end


hellais:
Ok, now tests are all green 🍏

Formula/ooniprobe.rb
```diff
@@ -152,36 +189,32 @@ def install
...
-      ln_s current_deck, "#{HOMEBREW_PREFIX}/share/ooni/decks/current.deck"
-    end
-  end
+    If this is your first time installing ooniprobe run:
+       brew services start ooniprobe
 
-  def caveats; <<-EOS.undent
-    Decks are installed to #{HOMEBREW_PREFIX}/share/ooni/decks/.
+    The you will be able to access the ooniprobe web user interface by going to http://127.0.0.1:8842/
     EOS
```
MikeMcQuaid: A similar message will be automaticaly printed for `services` stuff and the rest of the caveats can be removed.

Formula/ooniprobe.rb
```diff
@@ -152,36 +189,32 @@ def install
...
-      var_lib = #{var}/lib/ooni
+      var_lib = #{var}/run/ooni
+      etc = #{etc}/run/ooni
     EOS
 
     ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
     system "python", *Language::Python.setup_install_args(libexec)
     bin.install Dir["#{libexec}/bin/*"]
-    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"], :PATH => ENV["PATH"])
```
MikeMcQuaid: Can you elaborate on why the PATH is needed?

Formula/ooniprobe.rb
```diff
@@ -135,12 +169,15 @@ class Ooniprobe < Formula
...
+
+    # We need to set the PATH prefix otherwise ooniprobe will not manage to
+    # find homebrew installed tor
+    ENV.prepend_create_path "PATH", "#{HOMEBREW_PREFIX}/bin"
     resources.each do |r|
       r.stage do
         system "python", *Language::Python.setup_install_args(libexec/"vendor")
       end
     end
-
```
MikeMcQuaid: Please re-add this whitespace.

Formula/ooniprobe.rb
```diff
@@ -135,12 +169,15 @@ class Ooniprobe < Formula
   def install
     ENV["PYTHONPATH"] = Formula["libdnet"].opt_lib/"python2.7/site-packages"
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+
+    # We need to set the PATH prefix otherwise ooniprobe will not manage to
+    # find homebrew installed tor
+    ENV.prepend_create_path "PATH", "#{HOMEBREW_PREFIX}/bin"
```
MikeMcQuaid: It seems like it should be configured to do this in the plist and use `"#{HOMEBREW_PREFIX}/opt/tor/bin"`.

Formula/ooniprobe.rb
```diff
@@ -190,35 +223,35 @@ def plist; <<-EOS.undent
...
    </dict>
    </plist>
    EOS
   end
 
   test do
+    # Initialize ooniprobe to skip informed consent when testing
+    puts "#{var}/run/ooni/initialized"
+    is_initialized = File.exist?("#{var}/run/ooni/initialized")
+    (var/"run/ooni/initialized").write "" unless is_initialized
```
MikeMcQuaid: Need to use a directory outside `var` for this testing i.e. in `testpath`.

MikeMcQuaid (changes requested):

Formula/ooniprobe.rb
hellais: It's explained in the comment above. Basically without setting the PATH inside of the homebrew generated script that wraps the execution of the python script, PATH (when running the plist) doesn't contain the homebrew path of commands.

Formula/ooniprobe.rb
hellais: OK I will think of some way to overcome this and generate a configuration file specific for testing.

Formula/ooniprobe.rb
MikeMcQuaid: We need to figure out some solution to that, then, because we can't embed the current user's and/or Homebrew PATH inside the env script.

Formula/ooniprobe.rb
MikeMcQuaid: Thanks!

Formula/ooniprobe.rb
hellais: Would it be ok if I were to create a wrapper script that is used just in the context of the plist that injects the correct PATH vars?

dunn:
Just two formatting errors; you can use `brew style --fix ooniprobe` to automatically correct them:
```
C:257:  9: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C:259:  1: Extra empty line detected at block body end.

**Comment by hellais** [comment]
> Just two formatting errors; you can use brew style --fix ooniprobe to automatically correct them.

'ight. Fixed, thanks!

File: Formula/ooniprobe.rb (hunk 1 above)
Comment (MikeMcQuaid) [inline Formula/ooniprobe.rb]: We'll consider a PR to Homebrew/brew to amend these messages but you need to remove them from here for now, thanks.

**Review by MikeMcQuaid — CHANGES_REQUESTED** [review]

File: Formula/ooniprobe.rb (hunk 1 above)
Comment (hellais) [inline Formula/ooniprobe.rb]: Can I leave the message with information on how to access the web UI, but just drop the ones regarding the services? Is there some better way to give users post installation instructions?

Comment (MikeMcQuaid) [inline Formula/ooniprobe.rb]: There instructions should be for things that every user needs to know and are Homebrew specific. Remember that in e.g. Debian no post-installation instructions are printed or deemed necessary.

**Comment by MikeMcQuaid** [comment]
> Conflicting files
> Formula/ooniprobe.rb

This will need to be resolved.

**Comment by ilovezfs** [comment]
@BrewTestBot test this please

**Comment by ilovezfs** [comment]
@BrewTestBot test this please

**Comment by hellais** [comment]
I am going to close this PR and submit another one for the recently released 2.1.0.

**Comment by MikeMcQuaid** [comment]
Just push to this PR instead, thanks.

**Comment by hellais** [comment]
Ah ok (sorry for the double closing).

File: Formula/ooniprobe.rb [hunk 6]
```diff
@@ -200,41 +200,41 @@ def plist; <<-EOS.undent
    <dict>
      <key>Label</key>
        <string>#{plist_name}</string>
-     <key>Program</key>
-       <string>#{opt_bin}/ooniprobe</string>
+     <key>EnvironmentVariables</key>
+     <dict>
+       <key>PATH</key>
+       <string>#{HOMEBREW_PREFIX}/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
```
MikeMcQuaid: Why is this change needed, out of interest?

Formula/ooniprobe.rb
```diff
@@ -200,41 +200,41 @@ def plist; <<-EOS.undent
...
      <key>ProgramArguments</key>
      <array>
-       <string>-i</string>
-       <string>#{pkgshare}/current.deck</string>
+       <string>#{opt_bin}/ooniprobe-agent</string>
+       <string>run</string>
      </array>
      <key>RunAtLoad</key>
-       <false/>
+       <true/>
```
MikeMcQuaid: Why is this change needed, out of interest?

Formula/ooniprobe.rb
```diff
@@ -200,41 +200,41 @@ def plist; <<-EOS.undent
...
      <key>KeepAlive</key>
-       <false/>
-     <key>StandardErrorPath</key>
-       <string>/dev/null</string>
-     <key>StandardOutPath</key>
-       <string>/dev/null</string>
```
MikeMcQuaid: Where are these logged now?

Formula/ooniprobe.rb
hellais: On my version of macOS (Sierra) it appears that the PATH does not include the `HOMEBREW_PREFIX` path when starting programs via launchd. This is to ensure the HOMEBREW_PREFIX is present in there.

MikeMcQuaid: Is `ooniprobe` calling other things in the prefix? If so: do you know what? Thanks!

Formula/ooniprobe.rb
MikeMcQuaid: Cool, thanks for explaining.

Formula/ooniprobe.rb
hellais: I was under the impression that not supplying a path would make the default be to not log, but I checked and it seems like this leads to them being logged.

I will add these back.

Formula/ooniprobe.rb
hellais: Off the top of my head I think the only other thing it's calling in the HOMEBREW_PREFIX is `tor`.

hellais:
Why is the build still pending?

hellais:
@BrewTestBot test this please

dunn:
@BrewTestBot test this please

dunn:
(That command only works for maintainers)

Formula/ooniprobe.rb
```diff
@@ -200,41 +200,45 @@ def plist; <<-EOS.undent
...
        <string>/dev/null</string>
-     <key>StartCalendarInterval</key>
-     <dict>
-       <key>Hour</key>
-       <integer>00</integer>
-       <key>Minute</key>
-       <integer>00</integer>
-     </dict>
+     <key>WorkingDirectory</key>
+       <string>#{var}/log/</string>
```
MikeMcQuaid: Can you make this something specific to this formula e.g. `prefix` so it doesn't put random files in there? Thanks!

MikeMcQuaid (changes requested):

Formula/ooniprobe.rb
MikeMcQuaid: It may be worth adding tor's `opt_bin` in here instead but I don't have strong feelings either way.

hellais: Ok I will leave as is for the moment as this has been tested and works as expected.

hellais:
Can the CI tests be restarted? I suspect the failure is only a temporary one and may not happen on a restart.

MikeMcQuaid:
@hellais done.

MikeMcQuaid (approved):

MikeMcQuaid:
@BrewTestBot test this please

dunn:
This is weird:
```
        Error: ooniprobe contains conflicting version dependencies (openssl@1.1 openssl) so cannot be installed
```
Does it just need to be rebased?

MikeMcQuaid:

For future reference the preferred commit message format is `ooniprobe 2.1.0`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #5983, opened 2016-10-16
nim: fix head build

ghost:

---

Disabled building the tools for the head build - causes error otherwise.

### DIFF
--- a/Formula/nim.rb
+++ b/Formula/nim.rb
@@ -16,6 +16,10 @@ class Nim < Formula
   def install
     if build.head?
       system "/bin/sh", "bootstrap.sh"
+
+      # Grab the tools source (only included in point releases)
+      system "git", "clone", "https://github.com/nim-lang/nimsuggest.git", "dist/nimsuggest"
+      system "git", "clone", "https://github.com/nim-lang/nimble.git", "dist/nimble"
     else
       system "/bin/sh", "build.sh"
     end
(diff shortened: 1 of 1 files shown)


Formula/nim.rb
```diff
@@ -21,19 +21,23 @@ def install
...
-    target.install "bin/nimgrep"
-    target.install "bin/nimsuggest"
-
     bin.install_symlink prefix/"nim/bin/nim"
     bin.install_symlink prefix/"nim/bin/nim" => "nimrod"
-    bin.install_symlink prefix/"nim/bin/nimble"
-    bin.install_symlink prefix/"nim/bin/nimgrep"
-    bin.install_symlink prefix/"nim/bin/nimsuggest"
+
+    # Building tools only for non-head branch
```
MikeMcQuaid: Have they been removed? If not, please work with upstream to fix this error with them rather than disabling them.

ysalmi:
The tools are actually part of two related repos that are included in the
point releases but are not present on the head branch. This PR is just a
temporary fix (so that the HEAD build works). I will work with upstream to
find a proper fix.

MikeMcQuaid:
> I will work with upstream to find a proper fix.

Let's wait until you've done that as it may inform the best fix here.

ysalmi:
I worked with the upstream devs to find a good solution. Now HEAD builds just like a normal release.

Formula/nim.rb
```diff
@@ -16,24 +16,28 @@ class Nim < Formula
   def install
     if build.head?
       system "/bin/sh", "bootstrap.sh"
+
+      # Grab the tools source (only included in point releases)
+      system "git", "clone", "https://github.com/nim-lang/nimsuggest.git", "dist/nimsuggest"
```
MikeMcQuaid: Please use `resource` blocks for these so they are cached.

MikeMcQuaid (changes requested):

Formula/nim.rb
```diff
@@ -16,24 +16,28 @@ class Nim < Formula
...
       system "/bin/sh", "build.sh"
     end
     system "/bin/sh", "install.sh", prefix
 
-    system "bin/nim e install_tools.nims"
+    bin.install_symlink prefix/"nim/bin/nim"
+    bin.install_symlink prefix/"nim/bin/nim" => "nimrod"
 
+    # build and install the tools
+    system "bin/nim", "e", "install_tools.nims"
```
MikeMcQuaid: Is there a longer option that can be used instead of `e`? Can remove the comment, it's self-explanatory from the code.

MikeMcQuaid (changes requested):

ysalmi:
OK, I'll try this. What's unclear to me from the docs is how do I specify
which directory the download gets unpacked to.

ysalmi:
Sadly not. I will remove the comment.

Formula/nim.rb
ysalmi: Done

Formula/nim.rb
ysalmi: Unfortunately not. And I have removed the comment now.

MikeMcQuaid (approved):

MikeMcQuaid:

ysalmi:
Mike, thank you for your guidance on this PR. And keep up the great work on
this amazing project.

PR #5986, opened 2016-10-16
Add xcbuild 0.1.0 (new formula)

Palleas:

---

This is a follow up of @modocache's #3262, it works locally on my computer, please let me know
if there are things that I need to fix. I'm not 100% sure of what I did with the dependencies.

### DIFF
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -4,7 +4,7 @@ class Xcbuild < Formula
   url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0", :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa"
   head "https://github.com/facebook/xcbuild.git", :shallow => false
 
-  depends_on "ninja" => :recommended
+  depends_on "ninja"
   depends_on "cmake"
   depends_on "libpng"
   depends_on "pkg-config"
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -12,6 +12,7 @@ class Xcbuild < Formula
   def install
     system "cmake", ".", *std_cmake_args
     system "make", "install"
+    bin.install_symlink "#{prefix}/usr/bin/xcbuild"
   end
 
   test do
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -7,7 +7,6 @@ class Xcbuild < Formula
   depends_on "ninja" => :recommended
   depends_on "cmake"
   depends_on "libpng"
-  depends_on "libxml2"
   depends_on "pkg-config"
 
   def install
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -5,7 +5,7 @@ class Xcbuild < Formula
   head "https://github.com/facebook/xcbuild.git", :shallow => false
 
   depends_on "ninja" => :recommended
-  depends_on "cmake" => :recommended
+  depends_on "cmake" 
   depends_on "libpng"
   depends_on "libxml2"
   depends_on "pkg-config"
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -10,7 +10,8 @@ class Xcbuild < Formula
   depends_on "pkg-config"
 
   def install
-    system "make", "install", "cmake_flags=-DCMAKE_INSTALL_PREFIX=#{bin}"
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
   end
 
   test do
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -1,10 +1,7 @@
 class Xcbuild < Formula
   desc "Xcode-compatible build tool"
   homepage "https://github.com/facebook/xcbuild"
-  url "https://github.com/facebook/xcbuild.git",
-      :tag => "0.1.0",
-      :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa",
-      :shallow => false
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0"
   head "https://github.com/facebook/xcbuild.git", :shallow => false
 
   depends_on "ninja" => :recommended
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -10,7 +10,6 @@ class Xcbuild < Formula
   depends_on "pkg-config"
 
   def install
-    system "make"
     system "make", "install", "PREFIX=#{prefix}"
     bin.install_symlink "#{prefix}/usr/bin/xcbuild"
   end
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -5,7 +5,7 @@ class Xcbuild < Formula
   head "https://github.com/facebook/xcbuild.git", :shallow => false
 
   depends_on "ninja" => :recommended
-  depends_on "cmake" 
+  depends_on "cmake"
   depends_on "libpng"
   depends_on "libxml2"
   depends_on "pkg-config"
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -1,0 +1,25 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git",
+      :tag => "0.1.0",
+      :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa",
+      :shallow => false
+  head "https://github.com/facebook/xcbuild.git", :shallow => false
+
+  depends_on "ninja" => :recommended
+  depends_on "cmake" => :recommended
+  depends_on "libpng"
+  depends_on "libxml2"
+  depends_on "pkg-config"
+
+  def install
+    system "make"
+    system "make", "install", "PREFIX=#{prefix}"
+    bin.install_symlink "#{prefix}/usr/bin/xcbuild"
+  end
+
+  test do
+    system "#{bin}/xcbuild", "-showsdks"
+  end
+end
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -1,7 +1,7 @@
 class Xcbuild < Formula
   desc "Xcode-compatible build tool"
   homepage "https://github.com/facebook/xcbuild"
-  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0"
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0", :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa"
   head "https://github.com/facebook/xcbuild.git", :shallow => false
 
   depends_on "ninja" => :recommended
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -10,8 +10,8 @@ class Xcbuild < Formula
   depends_on "pkg-config"
 
   def install
-    system "make", "install", "PREFIX=#{prefix}"
-    bin.install_symlink "#{prefix}/usr/bin/xcbuild"
+    system "make", "install", "cmake_flags=-DCMAKE_INSTALL_PREFIX=#{bin}"
+    #bin.install_symlink "#{bin}/usr/bin/xcbuild"
   end
 
   test do
--- a/Formula/xcbuild.rb
+++ b/Formula/xcbuild.rb
@@ -11,7 +11,6 @@ class Xcbuild < Formula
 
   def install
     system "make", "install", "cmake_flags=-DCMAKE_INSTALL_PREFIX=#{bin}"
-    #bin.install_symlink "#{bin}/usr/bin/xcbuild"
   end
 
   test do


Formula/xcbuild.rb
```diff
@@ -0,0 +1,22 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0"
```
MikeMcQuaid: Needs a `:revision` specified too.

Formula/xcbuild.rb
```diff
@@ -0,0 +1,22 @@
...
+  head "https://github.com/facebook/xcbuild.git", :shallow => false
+
+  depends_on "ninja" => :recommended
+  depends_on "cmake" 
+  depends_on "libpng"
+  depends_on "libxml2"
+  depends_on "pkg-config"
+
+  def install
+    system "make"
```
MikeMcQuaid: Worth seeing if this can be skipped in favour of just `make install`.

Formula/xcbuild.rb
```diff
@@ -0,0 +1,22 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0"
+  head "https://github.com/facebook/xcbuild.git", :shallow => false
+
+  depends_on "ninja" => :recommended
+  depends_on "cmake" 
+  depends_on "libpng"
+  depends_on "libxml2"
```
MikeMcQuaid: Is it not possible to use the macOS system version?

MikeMcQuaid (changes requested):

MikeMcQuaid:
Some whitespace needs addressed: https://bot.brew.sh/job/Homebrew%20Core/9567/version=any_sierra/testReport/junit/brew-test-bot/sierra/audit_xcbuild___new_formula/

Formula/xcbuild.rb
Palleas: I assumed Homebrew would use the system one magically... My bad! I've removed it.

Formula/xcbuild.rb
```diff
@@ -0,0 +1,20 @@
...
+  head "https://github.com/facebook/xcbuild.git", :shallow => false
+
+  depends_on "ninja" => :recommended
+  depends_on "cmake"
+  depends_on "libpng"
+  depends_on "pkg-config"
+
+  def install
+    system "make", "install", "PREFIX=#{prefix}"
+    bin.install_symlink "#{prefix}/usr/bin/xcbuild"
```
MikeMcQuaid: This is the wrong installation location; `bin` things should be installed directly into `#{bin}`.  Looks like you may be able to pass some arguments to fix this: https://github.com/facebook/xcbuild/blob/69af7fbf902c2fc36b7ab49a9bf95710a73818dd/Makefile#L24-L26

Palleas: So it should be something like `system "make", "install", "PREFIX=#{bin}"` directly?

MikeMcQuaid: I think you may need to set DESTDIR or cmake_flags instead and, yeh, remove the symlink.

Palleas: The problem with installing it to `bin` is the `xcbuild` bin file ends up in `/usr/local/Cellar/xcbuild/0.1.0/bin/usr/bin/xcbuild` and without a symlink, it can't be executed.
 From what I see in facebook/xcbuild#81, being able to split support files from bin is a work in progress..?

dunn:
The test is failing, and there are a few audit failures:
```
xcbuild:
  * Non-executables were installed to "/usr/local/Cellar/xcbuild/0.1.0/bin"
    The offending files are:
      /usr/local/Cellar/xcbuild/0.1.0/bin/Library
      /usr/local/Cellar/xcbuild/0.1.0/bin/usr
    Broken dylib links found:
      /usr/lib/libcompression.dylib
  * C: 14: col 5: Missing space after #.

**Comment by Palleas** [comment]
I've removed the commented code that fails the build but I'm not sure how to fix the other, I don't think I can tell the make command to install only the binaries to `bin`... Any idea? Should I invoke some people from facebook/xcbuild?

File: Formula/xcbuild.rb [hunk 5]
```diff
@@ -0,0 +1,19 @@
...
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0", :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa"
+  head "https://github.com/facebook/xcbuild.git", :shallow => false
+
+  depends_on "ninja" => :recommended
+  depends_on "cmake"
+  depends_on "libpng"
+  depends_on "pkg-config"
+
+  def install
+    system "make", "install", "cmake_flags=-DCMAKE_INSTALL_PREFIX=#{bin}"
```
dunn: We provide `std_cmake_args` which sets `DCMAKE_INSTALL_PREFIX` and other variables; try seeing if you can supply it to the ninja process.

dunn: (Also, looking at the Makefile it seems ninja is required, not just recommended?)

Palleas: I fixed the cmake thingie but in the end the bin still ends up in `/usr/local/Cellar/xcbuild/0.1.0/usr/bin/xcbuild`, I still need to symlink it, right?

dunn:
@BrewTestBot test this please

Palleas:
@dunn when do we get the feedback from `BrewTestBot` ? 😁

dunn:
Here they are, the test is still failing I'm afraid: https://bot.brew.sh/job/Homebrew%20Core/10450/

Palleas:
The only remaining issue is "Broken dylib links found", not sure what it means though 🤔

Formula/xcbuild.rb
```diff
@@ -0,0 +1,21 @@
+class Xcbuild < Formula
+  desc "Xcode-compatible build tool"
+  homepage "https://github.com/facebook/xcbuild"
+  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0", :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa"
```
MikeMcQuaid: Split this onto 2-3 lines.

Formula/xcbuild.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  depends_on "ninja"
+  depends_on "cmake"
+  depends_on "libpng"
+  depends_on "pkg-config"
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+    bin.install_symlink "#{prefix}/usr/bin/xcbuild"
```
MikeMcQuaid: This should be fixed so `cmake` installs this into the right location.

MikeMcQuaid (changes requested):

MikeMcQuaid:
> The only remaining issue is "Broken dylib links found", not sure what it means though 🤔

Do you know if this works on Yosemite? If not, you could add `depends_on :macos => :el_capitan` with a comment noting this error.

Formula/xcbuild.rb
Palleas: Fixed.

Formula/xcbuild.rb
Palleas: Unfortunately, like I mentioned earlier, I'm not sure how to do that since I'm not used to cmake...

Palleas:
> Do you know if this works on Yosemite? If not, you could add depends_on :macos => :el_capitan with a comment noting this error.

facebook/xcbuild#66 tells me it's complicated.

grp:
Awesome to see progress on this @Palleas!

 - Probably best to use `cmake` directly to build and install, rather than the `make` wrapper (that's meant more for humans). That should let the formula use Homebrew's standard CMake handling and lets the `Makefile` change as needed in the future without breaking the formula.
 - For `libcompression`, it should be optional: it'll only build and link it if the system has it. Is there something happening where it'd be compiled 10.11+ and then trying to run on 10.10 or below? Otherwise I'm not sure how it'd end up linking to it successfully then not able to find it at runtime.
 - For the install paths, it's a little complicated. There's two main issues:

     - The CMake install targets a `/Developer`-style root with `/Developer/Library` and `/Developer/usr/bin` rather than a Unix-style one with `/bin`.
     - The executables (and libraries) conflict in name with OS X system components. Putting them first in the `PATH` would break macOS or at least Xcode.

    My suggestion to handle these is to install outside the standard executable paths, and add a single symlink for the `xcbuild` tool to the standard `bin`.

    The installation directory should probably be called `Developer`, since it's useable as a developer root (i.e. `DEVELOPER_DIR`). That is, pass `-DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/xcbuild/0.1.0/Developer`.

    The symlink just for `xcbuild` prevents the other tools from conflicting (that one tool is intentionally named to not conflict). With the above install prefix, symlink `/usr/local/Cellar/xcbuild/0.1.0/bin/xcbuild` -> `/usr/local/Cellar/xcbuild/0.1.0/Developer/usr/bin/xcbuild`.
 - The `0.1.0` tag is a bit old at this point. I'll try and tag a new release ASAP once facebook/xcbuild#176 goes in.

MikeMcQuaid:
> Probably best to use cmake directly to build and install, rather than the make wrapper (that's meant more for humans). That should let the formula use Homebrew's standard CMake handling and lets the Makefile change as needed in the future without breaking the formula.

@grp What buildsystem does CMake generate by default? Normally it's a `Makefile`; what is it here?

> Is there something happening where it'd be compiled 10.11+ and then trying to run on 10.10 or below?

Nope, this is building and running on the same OS X version.

> My suggestion to handle these is to install outside the standard executable paths, and add a single symlink for the xcbuild tool to the standard bin.

Agreed. @Palleas Install everything anywhere inside `libexec` and use `bin.install_symlink "#{libexec}/usr/bin/xcbuild". That means we don't care about the current weird installation prefix.

@Palleas Is there anything else like manpages/docs we need to install correctly?

> The installation directory should probably be called Developer, since it's useable as a developer root (i.e. DEVELOPER_DIR). That is, pass -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/xcbuild/0.1.0/Developer.

This feels a bit weird and non-Homebrewy to me. How would you use this `DEVELOPER_DIR`?

> The CMake install targets a /Developer-style root with /Developer/Library and /Developer/usr/bin rather than a Unix-style one with /bin.

@grp As far as I recall the correct way in CMake to do this is to use `/Developer/usr` as your `DESTDIR` rather than `CMAKE_INSTALL_PREFIX`.

grp:
> @grp What buildsystem does CMake generate by default? Normally it's a Makefile; what is it here?

xcbuild has a `Makefile` that runs CMake to generate Ninja, then runs Ninja to build. It's meant for humans as a shortcut to the somewhat arcane CMake commands. For automated things like Homebrew it's fine to ignore it and just run CMake directly with whatever generator is easiest; all of them should work fine.

> Nope, this is building and running on the same OS X version.

Weird. Not sure what's going on there then. Any way to get a copy of the built binaries from Jenkins to see what's up? Or could try adding a debug `otool -L` to the formula...

> Agreed. @Palleas Install everything anywhere inside libexec and use `bin.install_symlink "#{libexec}/usr/bin/xcbuild". That means we don't care about the current weird installation prefix.

The only thing I'd note about this is that if you *do* want to use that directory, it's weird for it to be called `libexec` but passed to `DEVELOPER_DIR=` (see below). Maybe `libexec/Developer/usr/bin/xcbuild`?

> This feels a bit weird and non-Homebrewy to me. How would you use this DEVELOPER_DIR?

You could run (xcbuild's implementation of) `xcode-select` to pick it, or set `DEVELOPER_DIR=` before running xcbuild. Think of it as if Xcode itself were a Homebrew formula: it'd need to install it's `/Developer` directory somewhere; xcbuild is the same way.

Definitely is weird. I'm not sure that the current approach is a good way to set it up.

> @grp As far as I recall the correct way in CMake to do this is to use /Developer/usr as your DESTDIR rather than CMAKE_INSTALL_PREFIX.

Good point, better at build time rather than configure time. Although in this case the build is only being run once so would be the same either way.

MikeMcQuaid:
> xcbuild has a Makefile that runs CMake to generate Ninja, then runs Ninja to build. It's meant for humans as a shortcut to the somewhat arcane CMake commands. For automated things like Homebrew it's fine to ignore it and just run CMake directly with whatever generator is easiest; all of them should work fine.

Gotcha, thanks. In that case @Palleas this should be `ninja install`.

> You could run (xcbuild's implementation of) xcode-select to pick it, or set DEVELOPER_DIR= before running xcbuild. Think of it as if Xcode itself were a Homebrew formula: it'd need to install it's /Developer directory somewhere; xcbuild is the same way.

Makes sense. Installing to `#{prefix}/Developer` feels fine then 👍 

> Good point, better at build time rather than configure time. Although in this case the build is only being run once so would be the same either way, I think.

Sure, I guess I meant more for packagers that this is generally a more expected way to handle things; we expect `CMAKE_INSTALL_PREFIX` put things in there rather than with a nested `usr/bin` underneath (although given what you've said this should maybe be `Developer/usr/bin`?

My vague concern is how we ensure that e.g. if you add manpages in future that they are installed to the correct location?

grp:
That's a good point about adding things in the future. A script or install mode to put the symlinks in place for a normal Unix root seems like it work as part of xcbuild rather than the formula.

Would be interested if @strager has any thoughts on that — similar to facebook/xcbuild#81. (Which I now understand better after this discussion.)

MikeMcQuaid:
@grp From a packager's perspective: ideally a `CMAKE_INSTALL_PREFIX` of `usr/local/Cellar/xcbuild/0.1.0` would Do The Right Thing; even if it requires a `-DPACKAGER_MODE=1` or similar.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

Palleas:
Sorry I've been having a busy couple of weeks I'll try to tackle that this weekend!

steeve:
Note to all, I've been using the following formula, inspired by the one on this PR:
```ruby
class Xcbuild < Formula
  desc "Xcode-compatible build tool"
  homepage "https://github.com/facebook/xcbuild"
  url "https://github.com/facebook/xcbuild.git", :tag => "0.1.0", :revision => "08575f65f1c907e280ed4663b5461b7cfaf5cfaa"
  head "https://github.com/facebook/xcbuild.git", :shallow => false

  depends_on "ninja"
  depends_on "cmake"
  depends_on "libpng"
  depends_on "pkg-config"

  def install
    system "cmake", "-DCMAKE_INSTALL_PREFIX=#{prefix}", ".", *std_cmake_args
    system "make", "install"
    bin.install "xcbuild"
  end

  test do
    system "#{bin}/xcbuild", "-showsdks"
  end
end
```

steeve:
For what it's worth, I've just made a slightly modified version taking into the account to remarks here at https://github.com/znly/homebrew-znly/blob/master/Formula/xcbuild.rb

One can install it with
```
$ brew install znly/znly/xcbuild
```

MikeMcQuaid:
@steeve Could you submit a PR for this? Thanks!

Palleas:
Do you need me to close this one?

MikeMcQuaid:
No, we can leave it open and have a race 😉

steeve:
>> #8600

woodruffw:
Closing in favor of #8600, thanks @steeve.

PR #5999, opened 2016-10-17
xchelper

saltzmanjoelh:
Added xchelper formula

### DIFF
--- a/Formula/xchelper.rb
+++ b/Formula/xchelper.rb
@@ -1,7 +1,5 @@
 class Xchelper < Formula
-  desc "xchelper helps you stay in Xcode and off the command line. Build your Swift code on Linux with Xcode
-Archive your project and upload to S3 for automatic deployment with Xcode Server
-Keep the Xcode references to Swift dependency sources updated so that you don't have to keep change them when your dependencies update."
+  desc "Build your Swift code on Linux with Xcode. Archive and upload to S3."
   homepage "https://github.com/saltzmanjoelh/XcodeHelper"
   url "https://github.com/saltzmanjoelh/XcodeHelper/archive/master.tar.gz"
   version "1.0.10"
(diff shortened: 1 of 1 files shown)


Formula/xchelper.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def install
+    system "swift", "build", "--configuration", "release"
+    bin.install "./.build/release/xchelper"
+  end
+
+  def uninstall
+    system "swift", "build", "--configuration", "release"
+    bin.install "./.build/release/xchelper"
+  end
```
MikeMcQuaid: `def uninstall` does nothing so needs removed.

Formula/xchelper.rb
```diff
@@ -0,0 +1,21 @@
+class Xchelper < Formula
+  desc "Build your Swift code on Linux with Xcode. Archive and upload to S3."
+  homepage "https://github.com/saltzmanjoelh/XcodeHelper"
+  url "https://github.com/saltzmanjoelh/XcodeHelper/archive/master.tar.gz"
```
MikeMcQuaid: We needs tagged, stable releases.

Formula/xchelper.rb
```diff
@@ -0,0 +1,21 @@
...
+    bin.install "./.build/release/xchelper"
+  end
+
+  def uninstall
+    system "swift", "build", "--configuration", "release"
+    bin.install "./.build/release/xchelper"
+  end
+
+  test do
+    system "swift", "test"
```
MikeMcQuaid: This needs to run the tool rather than the unit tests as it's post-installation when the source is no longer available.

MikeMcQuaid (changes requested):

MikeMcQuaid:
This GitHub repository is not notable enough for us yet (https://bot.brew.sh/job/Homebrew%20Core/9583/version=any_sierra/testReport/junit/brew-test-bot/sierra/audit_xchelper___new_formula/) so passing on this for now. Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

PR #6005, opened 2016-10-17
dwarf: migrate to boneyard

ilovezfs:


MikeMcQuaid:
How come?

ilovezfs:
https://github.com/elboza/dwarf-ng/pull/4#issuecomment-254174472

MikeMcQuaid:
Seems fair given lack of response to all the issues.

MikeMcQuaid:
Although, I dunno; if it's not causing us problems is it work boneyarding?

ilovezfs:
I think so. If someone's actually using and cares they can do jumping jacks on upstream's front lawn.

MikeMcQuaid:
My view would be we leave it until it breaks and boneyard then rather than try to patch any further.

MikeMcQuaid:
> My view would be we leave it until it breaks and boneyard then rather than try to patch any further.

And furthermore it's got a few installs since we began analytics so I'm definitely 👎 on removing something that some people are using.

ilovezfs:
I think we need to be significantly more aggressive with boneyarding. I went out of my want to fix these two formulae and report the problems upstream, despite recommended boneyarding by other maintainers. I've subsequently been totally ignored by upstream and have no interest in staring at these open PRs in my list of open PRs indefinitely.

ilovezfs:
>And furthermore it's got a few installs since we began analytics so I'm definitely :-1: on removing something that some people are using.

Good then there will be someone to do the jumping jacks.

MikeMcQuaid:
> I think we need to be significantly more aggressive with boneyarding.

I agree when things are broken and/or causing us active support load and/or used by no-one. I disagree when they are used by people and upstream is just currently being shit.

> I've subsequently been totally ignored by upstream and have no interest in staring at these open PRs in my list of open PRs indefinitely.

That's reasonable, I'd do the same.

ilovezfs:
> agree when things are broken and/or causing us active support load and/or used by no-one.

Downloaded != used. If someone cares, they can do the jumping jacks to get it fixed. Maybe boneyarding this will motivate that.

In this case, the errors are so ridiculously trivial and easy to fix, and the code is of such questionable quality, that I see nothing worth preserving due to some blips in the analytics.

MikeMcQuaid:
> Downloaded != used.

That's "installed" rather than "downloaded" in this case. I'm going to continue to assume that at least a proportion of installs are used.

I'm not in disagreement about upstream quality. I'm just saying: let's wait until this breaks again rather than boneyard currently working software.

ilovezfs:
> I'm just saying: let's wait until this breaks again rather than boneyard currently working software.

So I take it all the Qt 4 rev-deps are coming back to Core then?

MikeMcQuaid:
Qt4 does not work on Sierra, is unmaintained and has known security vulnerabilities.

ilovezfs:
This software is unmaintained on macOS (it does not compile without my patch) and we are in no position to evaluate whether it does or doesn't have security issues. How do you expect upstream to suddenly start caring unless there's someone barking at them that Homebrew boneyarded it because it doesn't compile?

MikeMcQuaid:
On the flip side, we've already tried barking at them so not sure how it would help to have more people barking at them?

ilovezfs:
What we've been doing so far has had no impact. It's time to try something else.

MikeMcQuaid:
Ok, fair enough. On you go.

ilovezfs:
:bowing_man:

PR #6007, opened 2016-10-17
libmarisa: migrate to boneyard

ilovezfs:


MikeMcQuaid:
Similarly: if it's not causing us problems is it worth boneyarding?

MikeMcQuaid:
I'm also coming around to @chdiza's point about maybe having `brew boneyard-formula-pr` taking a mandatory "reason" argument that's inserted into the various commit messages. Thoughts?

ilovezfs:
No I don't think that's necessary.

ilovezfs:
>Similarly: if it's not causing us problems is it worth boneyarding?

jumping jacks rule again.

MikeMcQuaid:
@ilovezfs Would you actively object to it? Seems preferably to have this conversation every time (sometimes in hindsight). 

> jumping jacks rule again.

It's got no installs since records began so 👍 to boneyard this one.

ilovezfs:
>Would you actively object to it? Seems preferably to have this conversation every time (sometimes in hindsight). 

Yes.

MikeMcQuaid:
> Yes.

Can you elaborate a bit on that?

ilovezfs:
>Can you elaborate a bit on that? 

Most of the time the reason is self-evident if you look at the PR on GitHub not email due to cross links between issues and PRs. Mandating that information be duplicated when it's already there is just wasting time.

ilovezfs:
>a mandatory "reason" argument that's inserted into the various commit messages

@MikeMcQuaid I think it would be fine as long as it's not mandatory (i.e., has an opt-in or opt-out).

MikeMcQuaid:
@ilovezfs 🆒

Issue #4280, opened 2016-08-28
Google Code Migration to Archive


DomT4:
They've finally shunted (possibly) everything over to an archive link, with existing links hitting 404s. The following formulae will need checking/rewriting:

- [x] aoeui.rb
- [x] blazeblogger.rb
- [x] cconv.rb - https://github.com/Homebrew/homebrew-core/pull/4297
- [x] cifer.rb
- [x] cityhash.rb
- [x] clamz.rb
- [x] cmigemo.rb
- [x] cmockery.rb
- [x] csshx.rb
- [x] darkice.rb - https://github.com/Homebrew/homebrew-core/pull/4299
- [x] dircproxy.rb - *boneyard me.* - https://github.com/Homebrew/homebrew-core/pull/4304
- [ ] distcc.rb - *Python installation is a mess. Possible boneyard.*
- [x] dnsmap.rb
- [x] dsocks.rb
- [x] fex.rb
- [x] flow-tools.rb
- [x] game-music-emu.rb
- [x] garmintools.rb
- [x] git-multipush.rb
- [x] gmail-backup.rb
- [x] hostdb.rb
- [x] hqx.rb
- [x] htmlcompressor.rb
- [x] httperf.rb
- [ ] iphotoexport.rb - *boneyard, the replacement project is dead, let alone this one*.
- [ ] jing.rb - *Update version, I live on GitHub now*: https://github.com/relaxng/jing-trang
- [x] js-test-driver.rb
- [ ] jsdoc-toolkit.rb - *See if we can update to v3*, https://github.com/jsdoc3/jsdoc.
- [x] jslint4java.rb
- [x] jvmtop.rb
- [x] lastfmlib.rb - *Boneyard, doesn't build with Clang*. - https://github.com/Homebrew/homebrew-core/pull/4304
- [x] lib3ds.rb
- [x] libdnet.rb
- [x] libfreefare.rb
- [x] libkate.rb
- [x] libkml.rb
- [x] liblunar.rb
- [x] libmarisa.rb
- [x] libpgm.rb
- [x] libphonenumber.rb
- [x] libtess2.rb - https://github.com/Homebrew/homebrew-core/pull/4303
- [x] makepkg.rb
- [x] memcache-top.rb
- [x] memcacheq.rb
- [x] mfcuk.rb
- [x] mfoc.rb
- [x] mp3check.rb
- [x] mp4v2.rb
[...]

PR #6011, opened 2016-10-17
jsdoc-toolkit: migrate to boneyard

bw-yuhogg:

---

This was pointed out at #4280, where it was suggested that we move to the successor project `jsdoc3`. However, `jsdoc3` has already been available in Homebrew for a while now (they even conflict with each other) and the download link currently 404's.

Furthermore, the project homepage has the following notice:

```
NOTICE: As of 27 June 2010 the JsDoc Toolkit Version 2 project is no longer under active development or support.
```


MikeMcQuaid:
Thanks again @rwhogg!

ilovezfs:
@rwhogg just a note for future reference. This PR should also have removed
```
Formula/jsdoc3.rb
19:  conflicts_with "jsdoc-toolkit", :because => "both install jsdoc"

```
which is now causing https://github.com/Homebrew/homebrew-core/pull/6105 to fail.

Issue #5389, opened 2016-09-29
open-scene-graph - Looking for socket in socket - not found (10.12)


kolen:
macOS 10.12

Logs:
https://gist.github.com/anonymous/152c6db5a84192abac24359b6afb89fe

kolen:
Seems that bottle is successfully built: 965e3b8a858aa967f140a2cca7f9be88dfad4218

davenquinn:
I've also got this problem. Formula `--with-qt5` builds unsuccessfully. QT is installed (bottled). The bottled formula builds, but I need the QT links for working with OSGEarth.

ilovezfs:
This fixes it: https://github.com/Homebrew/homebrew-core/pull/6013

ilovezfs:
The current version in homebrew is now 3.5.5 which does not have this bug.

@kolen thank you for reporting this!

PR #6013, opened 2016-10-17
open-scene-graph: fix build on 10.12

ilovezfs:
CMakeLists.txt can't grapple with 10.12 when it expects 10.12.0

Fixes issue #5389.

### DIFF
--- a/Formula/open-scene-graph.rb
+++ b/Formula/open-scene-graph.rb
@@ -49,6 +49,10 @@ class OpenSceneGraph < Formula
   def install
     ENV.cxx11 if build.cxx11?
 
+    if MacOS.full_version.to_s.split(".").count < 3
+      inreplace "CMakeLists.txt", "    LIST(GET MACOS_VERSION_LIST 2 MACOS_VERSION_PATCH)", ""
+    end
+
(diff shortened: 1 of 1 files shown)


zmwangx:
Hmm, OS X never had 10.x.0 releases as far back as my memory goes. Wonder how this worked before.

ilovezfs:
it probably never did lol

MikeMcQuaid:
Looks like `head` has moved to https://github.com/openscenegraph/OpenSceneGraph and there's much newer releases there that don't seem to have this same text so probably worth trying an upgrade instead?

ilovezfs:
it looks like it worked
```
==> brew audit open-scene-graph
==> FAILED
Error: 1 problem in 1 formula
open-scene-graph:
  * Use MacOS.version instead of MACOS_VERSION
```
is the reason for CI failure. But yes an upgrade would be nice.

ilovezfs:
Also the audit issue is a false positive. "MACOS_VERSION" is part of the source code LOL

ilovezfs:
OK, new version compiles, so let's do that instead.

PR #6014, opened 2016-10-17
fdupes: bump version_scheme

zmwangx:

---

26c531a bumped fdupes from version 1.51 to 1.6.1, which requires a
version_scheme bump.


MikeMcQuaid (approved):
Nice spot, sorry for missing it. Feels like it may be worthy of an `audit` check; will have a think.

zmwangx:
@MikeMcQuaid Yes, since there's a "revision should not decrease" in place, we should add a "version should not decrease".

MikeMcQuaid:
> @MikeMcQuaid Yes, since there's a "revision should not decrease" in place, we should add a "version should not decrease".

@zmwangx Completely agree 👍

PR #6018, opened 2016-10-17
open-scene-graph 3.5.5

ilovezfs:


ilovezfs:
It seems the Qt 5 version now has its own separate repository https://github.com/openscenegraph/osgQt CC @MikeMcQuaid

MikeMcQuaid:
Think it's fine to merge this as-is 👍

MikeMcQuaid:
What happens if you use the `--with-qt5` option now? :boom: or just does nothing?

ilovezfs:
Does nothing. The Qt 5 version is now https://github.com/openscenegraph/osgQt

MikeMcQuaid:
Separate repo probably warrants a separate formula.

davenquinn:
So does somebody need to create an `open-scene-graph-qt5` formula, then? I'm not sure exactly how these were split apart or what would need to happen to recombine, but I might be game to try to publish something unless someone is already working on this...

MikeMcQuaid:
> open-scene-graph-qt5 formula, then

Probably `open-scene-graph-qt` but: yes. Could you try and open a pull request? [This document](http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html) should help and we're happy to walk you through anything else.

Thanks!

PR #6024, opened 2016-10-17
folly 2016.10.17.00

ilovezfs:
CC @yschimke

### DIFF
--- a/Formula/folly.rb
+++ b/Formula/folly.rb
@@ -1,10 +1,8 @@
 class Folly < Formula
   desc "Collection of reusable C++ library artifacts developed at Facebook"
   homepage "https://github.com/facebook/folly"
-  url "https://github.com/facebook/folly/archive/v2016.09.05.00.tar.gz"
-  sha256 "79e931af5d9610fee80ea81af82fcaed35b4d58a529e504c1326e62448e780ae"
-  revision 1
-
+  url "https://github.com/facebook/folly/archive/v2016.10.17.00.tar.gz"
+  sha256 "0f83685016d020111ba54ddc48c0cf33e1e0b9b35cee5ae82d5f2cbc5f6b0e82"
   head "https://github.com/facebook/folly.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Formula/folly.rb
```diff
@@ -41,21 +39,19 @@ def install
...
-      end
 
-      # Fix "candidate function not viable: no known conversion from
-      # 'folly::detail::Clock' to 'clockid_t' for 1st argument"
-      # See upstream PR mentioned above
-      if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+        # Fix "candidate function not viable: no known conversion from
+        # 'folly::detail::Clock' to 'clockid_t' for 1st argument"
+        # See upstream PR mentioned above
         inreplace "portability/Time.h", "typedef uint8_t clockid_t;", ""
```
Orvid: What is `clockid_t` being defined as on 10.11? `clock_gettime` shouldn't exist on 10.11, should it?

ilovezfs: hope this helps: https://bugs.erlang.org/browse/ERL-256?focusedCommentId=11633&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-11633

Orvid: I'm still not seeing why the addition of the casts is required. The casts aren't required on 10.12.

If I'm understanding everything correctly (I don't have easy access to a 10.11 machine to test with), Folly is currently defining its implementation of `clock_gettime` on OSX, but the declaration of `clockid_t` isn't compatible with the way OSX's headers define it. That means the only thing that needs to be done for OSX 10.11 to work is to make the declaration of `clockid_t` conditional on it not already being declared, correct?

As a side note, as-of https://github.com/facebook/folly/commit/823ecf8917d7f0534ed02b8d38bfc966cb0fd7b6, Folly can be compiled (properly) against Boost >= 1.61.

PR #6033, opened 2016-10-18
Go 1.7.2

fsouza:
The mirror URL doesn't work yet, should I remove it?


fsouza:
Actually, hold off. It has been tagged, but there might be an issue with the release (golang/go#17071).

PR #6060, opened 2016-10-18
yarn 0.16.0

chiedo:
Update yarn to 0.16.0

### DIFF
--- a/Formula/yarn.rb
+++ b/Formula/yarn.rb
@@ -3,13 +3,8 @@ require "language/node"
 class Yarn < Formula
   desc "Javascript package manager"
   homepage "https://yarnpkg.com/"
-  # Note: If updating this to a newer version, please change the URL to the official
-  # yarnpkg.com download URL:
-  # url "https://yarnpkg.com/downloads/0.15.0/yarn-v0.15.0.tar.gz"
-  # This was not updated yet as Yarn 0.15.1 was an npm-specific release and is not available
-  # as a standalone tarball.
-  url "https://registry.npmjs.org/yarn/-/yarn-0.15.1.tgz"
-  sha256 "f99fd587e84987909d5f9e918b8fe524349fdc548e5bc5c380c8f8c0a70c6b87"
+  url "https://yarnpkg.com/downloads/0.16.0/yarn-v0.16.0.tar.gz"
+  sha256 "cd1d7eeb8eb2518441d99c914e5fd18b68e2759743d212dfd8f00574a1de6da8"
   head "https://github.com/yarnpkg/yarn.git"
 
   bottle do


chiedo:
Not sure how long everything will take but if it's taking too long for them to merge it into Homebrew you can get 0.16.0 by doing the following.

**Run this:**

`brew edit yarn`

**Replace the content in that file with this:**

```
require "language/node"

class Yarn < Formula
  desc "Javascript package manager"
  url "https://yarnpkg.com/downloads/0.16.0/yarn-v0.16.0.tar.gz"
  sha256 "cd1d7eeb8eb2518441d99c914e5fd18b68e2759743d212dfd8f00574a1de6da8"
  head "https://github.com/yarnpkg/yarn.git"

  bottle do
    cellar :any_skip_relocation
    sha256 "04d5fe5e5667d77b40a882a1bdbe08625bd52e0d228af7aa925455760bafd58f" => :sierra
    sha256 "291ad03475ea4770f357559dc9116309796bf1af83c13644cafcd00163283bce" => :el_capitan
    sha256 "84a717ce964d4ec5cd69ae6336a79db27fa7580cee1c978e33e310e4eb8e36a3" => :yosemite
  end

  depends_on "node"

  def install
    system "npm", "install", *Language::Node.std_npm_install_args(libexec)
    bin.install_symlink Dir["#{libexec}/bin/*"]
  end

  test do
    (testpath/"package.json").write('{"name": "test"}')
    system bin/"yarn", "add", "jquery"
  end
end
```

**Run this**
`brew upgrade yarn`

Daniel15:
Thank you! 😄

Daniel15:
> Shouldn't this be: https://yarnpkg.com/downloads/#{VERSION}/yarn-v#{VERSION}.tar.gz?

@designorant - We do have an endpoint at https://yarnpkg.com/latest-version to get the latest version number, can Homebrew use that to determine the version to download? I think that'd be ideal.

tleunen:
Maybe we can get this merged and then find a way to always provide the latest version so manual PR like this one are not required.

designorant:
@Daniel15 Sorry, I've mistaken this PR for a Cask where versioned URLs are preferred. I don't know enough about Homebrew Core formulas to contribute, I'm afraid.

chrisfinazzo:
@tleunen, Do you know if they have a "latest" URL? Could pull from there - although even getting a location of the current tarball required some investigation....

Hmm.

Daniel15:
https://yarnpkg.com/latest.tar.gz will get you the latest version. However, how would Brew verify the checksum in that case? It'll change for every release.

chrisfinazzo:
@Daniel15 Yeah, I noticed that. Cask - which I'm a bit more familiar with - does have a mechanism to compute this using the `shasum` command line tool. In this case, do you think it's preferable to ask people to change the version number and the checksum for each update?

I don't think the documentation currently requires that.

Curious to hear what the Core team thinks...

designorant:
`brew audit` says that the version must be specified either via an attribute, or in the URL (Homebrew is smart enough to figure it out from there).

[Looks like the updating process will eventually be automated.](https://github.com/yarnpkg/yarn/issues/1202)

Daniel15:
>do you think it's preferable to ask people to change the version number and the checksum for each update?

Had a quick look and it looks like the Node.js formula also specifies an exact version, so I guess that's the correct way to do it with Homebrew: https://github.com/Homebrew/homebrew-core/blob/master/Formula/node.rb#L4

chiedo:
Yeah, yesterday was my first day updating a package on homebrew but I'm pretty sure it has to be a manual PR with the changes made.

Mainly because of the checksum. With that being said, Yarn does have the self-update option. I'm not sure how that works along side the Homebrew installation but that's another option...

tleunen:
self-update is broken in 0.15.1 :/ http://stackoverflow.com/questions/40030197/yarn-self-update-throws-oauth2-authentication-error

chiedo:
Yep @tleunen which is why this PR is definitely needed haha

chiedo:
If you can't wait for this to be merged, you can update manually by following this process:

https://github.com/yarnpkg/yarn/issues/1201#issuecomment-254595369

MikeMcQuaid:
@chiedo Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

Please be patient on version upgrades; Homebrew is a volunteer project and I don't think a ~24h turnaround is too unreasonable.

> Had a quick look and it looks like the Node.js formula also specifies an exact version, so I guess that's the correct way to do it with Homebrew: https://github.com/Homebrew/homebrew-core/blob/master/Formula/node.rb#L4

That's correct. We need to verify checksums, build packages and check tests.

> With that being said, Yarn does have the self-update option. I'm not sure how that works along side the Homebrew installation but that's another option...

Ideally it would be disabled when installed via Homebrew.

chiedo:
Thanks @MikeMcQuaid :) The turn-around was more than reasonably fast. I wasn't sure what to expect. Sometimes with open source projects the wait time is months, haha.

tleunen:
Thanks @MikaelSmith. No worries. Homebrew is even updated before npm so that's great ;)

PR #6088, opened 2016-10-19
Undo erroneous "(Mac) OS X" to "macOS" changes

ilovezfs:
Partial revert of 1e62c645b2fc2d82042d9f7c364c6a246f2e11ed

Some patches and string replacements were mistakenly modified as part of
the mass substitution of "macOS" for "OS X" and "Mac OS X" references.


MikeMcQuaid:
Apologies for missing these; thought I checked for patches.

ilovezfs:
That's OK. It'll only be nine floggings – one per file.

MikeMcQuaid:
Seems reasonable.

PR #6091, opened 2016-10-19
Artifactory 4.13.2

matankatz:


ilovezfs:
@matankatz thanks for your contribution to Homebrew! :tada: 

For future reference, note that the standardized commit subjects are all lower-case.

PR #6095, opened 2016-10-19
qscintilla2: add symlinks *-qt5.dylib

schoeps:

---

I suggest to add symlinks from `lib/libqsci*.dylib` to `lib/libqsci*-qt5.dylib` because some configure scripts are looking for `-qt5` (since most linux distributions ship qt4 and qt5 versions).

### DIFF
--- a/Formula/qscintilla2.rb
+++ b/Formula/qscintilla2.rb
@@ -64,6 +64,12 @@ class Qscintilla2 < Formula
       system "make", "install"
     end
 
+    # create symlinks to lib*-qt5.dylib because some configure scripts expect it
+    Pathname.glob("#{lib}/*.dylib") do |file|
+      newfile = File.dirname(file)+"/"+File.basename(file, ".dylib")+"-qt5.dylib"
+      ln_s file, newfile
+    end
+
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> because some configure scripts are looking for -qt5

Which ones?

schoeps:
I recently patched `octave` to look for `lib*.dylib` instead of `lib*-qt5.dylib` and I recalled that the problem was also mentioned in #4258. I have seen that also other packages do similar assumptions in their configure scripts. It seems to be the quasi standard among linux distributions. I don't like this but on the other hand doing things differently (more correctly if you like) creates more pain for future contributors. Upstream (riverbank) seems to be not very interested. 

Let me ask the other way around, is there any serious drawback in creating a few symlinks?

MikeMcQuaid:
> is there any serious drawback in creating a few symlinks?

We encourage upstream to rely on incorrect packaging behaviours. I'd rather avoid it as a result but will leave this open for a bit for other maintainers to disagree.

MikeMcQuaid:
Passing on this, sorry but thanks anyway.

PR #6106, opened 2016-10-20
mpv: unvendor waf

ilovezfs:
upstream fetches waf and verifies its checksum so we don't need to


fuzzyroddis:
I see `bootstrap.py` is using waf 1.8.12 yet `Formula/mpv.rb` was using `1.9.2`. Are you sure this won't cause problems?

ilovezfs:
This is the version upstream wants used. There are much more likely to be problems the other way using a random version unrelated to what upstream is using at any given time.

fuzzyroddis:
Fair point

PR #6129, opened 2016-10-21
ldc 1.0.1-alpha1

ilovezfs:
CC @JohanEngelen

### DIFF
--- a/Formula/ldc.rb
+++ b/Formula/ldc.rb
@@ -1,11 +1,13 @@
 class Ldc < Formula
   desc "Portable D programming language compiler"
   homepage "https://wiki.dlang.org/LDC"
-  revision 2
 
   stable do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

JohanEngelen:
Thanks.
I patched LLVM 1.0.0 so that it builds with LLVM3.9, there are some stupid test failures to be fixed and then we can do a subrelease 1.0.1.

ilovezfs:
I assume you mean you patched ldc 1.0.0

>stupid test failures to be fixed and then we can do a subrelease 1.0.1.

Take your time, but if the test failures are bogus, I'd probably defer their fixes to 1.0.2

JohanEngelen:
The failures were fixed in the commit hash that you chose here :-)  excellent

@klickverbot @redstar Let's release 1.0.1 ?

PR #6134, opened 2016-10-21
wxmac: reunite lost comment with its associated code

apjanke:
Looks like a comment got separated from the option it's annotating. Reunite them.


nijikon:
:rocket:

PR #6145, opened 2016-10-21
ffmbc: fix man pages not installed

ply:
ffmbc formula didn't install man files, so I fixed this.


nijikon:

PR #6147, opened 2016-10-21
dscanner 0.4.0-beta.3 (devel)

John-Colvin:

---

also switched to `make` instead of `dub` in order to avoid doing any downloading during the `install`. That means we need the submodules already there, hence using git instead of tarballs.

### DIFF
--- a/Formula/dscanner.rb
+++ b/Formula/dscanner.rb
@@ -1,8 +1,11 @@
 class Dscanner < Formula
   desc "Analyses e.g. the style and syntax of D code."
   homepage "https://github.com/Hackerpilot/Dscanner"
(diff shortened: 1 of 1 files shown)


Formula/dscanner.rb
```diff
@@ -11,12 +14,18 @@ class Dscanner < Formula
...
+  end
+
   depends_on "dmd" => :build
-  depends_on "dub" => :build
 
   def install
-    system "dub", "build"
-    bin.install "dscanner"
+    system "make", "dmdbuild"
+    bin.install "bin/dscanner"
```
John-Colvin: no, there isn't a make install

MikeMcQuaid:

PR #6149, opened 2016-10-21
terraform-provisioner-ansible 0.2 (new formula)

tboerger:
A new formula to provide an additional provisioner for the Terraform
CLI. Otherwise it's more complicated to provision machines created with
Terraform via Ansible.

### DIFF
--- a/Formula/terraform-provisioner-ansible.rb
+++ b/Formula/terraform-provisioner-ansible.rb
@@ -1,0 +1,56 @@
+require "language/go"
+
+class TerraformProvisionerAnsible < Formula
+  desc "Ansible provisioner for Terraform"
+  homepage "https://github.com/jonmorehouse/terraform-provisioner-ansible"
+  url "https://github.com/jonmorehouse/terraform-provisioner-ansible/archive/0.0.2.tar.gz"
+  sha256 "5ba93ea5d6f28e156d7e4eba90cd89032f7dc6aff396a8429c903da4efa2d8b6"
+  head "https://github.com/jonmorehouse/terraform-provisioner-ansible.git"
+
+  depends_on "go" => :build
+  depends_on "terraform" => :run
+
+  go_resource "github.com/hashicorp/terraform" do
+    url "https://github.com/hashicorp/terraform.git", :revision => "fa6a83ebdc323f2b415779786e102e69ddbf9a48"
+  end
+
+  go_resource "github.com/mitchellh/mapstructure" do
+    url "https://github.com/mitchellh/mapstructure.git", :revision => "f3009df150dadf309fdee4a54ed65c124afad715"
+  end
+
+  go_resource "github.com/mitchellh/go-homedir" do
+    url "https://github.com/mitchellh/go-homedir.git", :revision => "756f7b183b7ab78acdbbee5c7f392838ed459dda"
+  end
+
+  go_resource "github.com/mitchellh/go-linereader" do
+    url "https://github.com/mitchellh/go-linereader.git", :revision => "07bab5fdd9580500aea6ada0e09df4aa28e68abd"
+  end
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    terrapath = buildpath/"src/github.com/jonmorehouse/terraform-provisioner-ansible"
+    terrapath.install Dir["*"]
+
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd terrapath do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

Formula/terraform-provisioner-ansible.rb
```diff
@@ -0,0 +1,42 @@
...
+    github.com/hashicorp/terraform fa6a83ebdc323f2b415779786e102e69ddbf9a48
+    github.com/mitchellh/mapstructure f3009df150dadf309fdee4a54ed65c124afad715
+    github.com/mitchellh/go-homedir 756f7b183b7ab78acdbbee5c7f392838ed459dda
+    github.com/mitchellh/go-linereader 07bab5fdd9580500aea6ada0e09df4aa28e68abd
+  ]
+
+  terraform_deps.each_slice(2) do |x, y|
+    go_resource x do
+      url "https://#{x}.git", :revision => y
+    end
```
MikeMcQuaid: Please just hard-code these rather than making it dynamic.

MikeMcQuaid (changes requested):

Formula/terraform-provisioner-ansible.rb
tboerger: But this is an exact copy how this is also handled on the terraform formula itself?

tboerger: Ok, so you prefer the list of deps all with a separate call to go_resource instead of doing it more DRY?

Formula/terraform-provisioner-ansible.rb
```diff
@@ -0,0 +1,42 @@
+require "language/go"
+
+class TerraformProvisionerAnsible < Formula
+  desc "Ansible provisioner for Terraform"
+  homepage "https://github.com/jonmorehouse/terraform-provisioner-ansible"
+  url "https://github.com/jonmorehouse/terraform-provisioner-ansible/archive/0.0.2.tar.gz"
+  sha256 "5ba93ea5d6f28e156d7e4eba90cd89032f7dc6aff396a8429c903da4efa2d8b6"
+  head "https://github.com/jonmorehouse/terraform-provisioner-ansible.git"
+
+  depends_on "go" => :build
```
cblecker: Shouldn't this also have a dependency on terraform itself? (not for build like go, but for runtime)

Formula/terraform-provisioner-ansible.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd terrapath do
+      system "go", "build"
+      bin.install "terraform-provisioner-ansible"
+    end
+  end
+
+  test do
```
cblecker: Similar to the test for Terraform, could this test be a little more complex to ensure dependencies are met for both build and run?

cblecker (changes requested):

Formula/terraform-provisioner-ansible.rb
tboerger: Workout terraform hard to test :)

Formula/terraform-provisioner-ansible.rb
tboerger: Should make sense since it's useless without terraform. And maybe it should be possible to write usable tests than

tboerger: Added it as a dependency

Formula/terraform-provisioner-ansible.rb
tboerger: Done

Formula/terraform-provisioner-ansible.rb
tboerger: I have not been able to do that properly since it requires a file in the home directory to work properly and I don'T know another way than just calling `terraform` without any path to get it executed.

tboerger (commented):
I have also added a `caveats` section to mention the creation of the required file in the home directory.

cblecker:
Perhaps the test on the terraform formula has clues on how to write out a test file and then have terraform graph it offline to ensure that it can read and graph an example config with the dependencies.

https://github.com/Homebrew/homebrew-core/blob/master/Formula/terraform.rb#L82-L112

MikeMcQuaid (approved):

MikeMcQuaid:

PR #5267, opened 2016-09-26
kubernetes-cli 1.4.0

harsha-y:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

PR #6152, opened 2016-10-21
kubernetes-cli 1.4.4

antoineco:
Minor version update for the Kubernetes CLI tool, `kubectl`

I also re-added the `devel` block, which was removed in #5267 because the `stable` version was newer than the `devel` one. The correct way to handle this formula is to keep the devel version up-to-date, not to remove it.


nijikon:
@antoineco the correct way to do this is to remove `devel` version if it is the same as `stable` and no new `devel` is available.

antoineco:
Kubernetes **always** have a devel version newer than the stable one.

nijikon:

PR #6195, opened 2016-10-23
wxmac: remove --disable-debug #6117

### LINKED ISSUE #6117: why is wxmac built with --disable-debug?
The formula for wxmac specifies `--disable-debug` in the configure args.

According to: http://docs.wxwidgets.org/trunk/overview_debugging.html

> Starting with wxWidgets 2.9.1 debugging features are always available by default (and not only in a special "debug" build of the library) and you need to predefine wxDEBUG_LEVEL symbol as 0 when building both the library and your application to remove them completely from the generated object code. However the debugging features are disabled by default when the application itself is built with NDEBUG defined (i.e. in "release" or "production" mode) so there is no need to do this, unless the resources of the system your application will be running on are unusually constrained (notice that when asserts are disabled their condition is not even evaluated so the only run-time cost is a single condition check and the extra space taken by the asserts in the code).

In other words, minimal debug code is normally included in the built library and has practically no run-time cost.

When the library is built with `--disable-debug`, and you built your sources with `-DwxDEBUG_LEVEL=1`, the resulting objects cannot link.

Would it be possible to remove this flag?

rkitover:
According to: http://docs.wxwidgets.org/trunk/overview_debugging.html

wxwidgets debug code is included by default and has minimal runtime
cost, further discussion in the linked issue.

Closes #6117

### DIFF
--- a/Formula/wxmac.rb
+++ b/Formula/wxmac.rb
@@ -1,7 +1,7 @@
 class Wxmac < Formula
   desc "wxWidgets, a cross-platform C++ GUI toolkit (for macOS)"
   homepage "https://www.wxwidgets.org"
-  revision 3
+  revision 4
 
   head "https://github.com/wxWidgets/wxWidgets.git"
 
@@ -75,7 +75,6 @@ class Wxmac < Formula
 
   def install
     args = [
-      "--disable-debug",
       "--prefix=#{prefix}",
       "--enable-unicode",
       "--enable-std_string",


apjanke:
Thanks!

This change should include a revision bump, since it changes the binaries that will be installed. Just change that `revision 3` near the top to `revision 4`.

rkitover:
Fixed!

nijikon:

PR #6215, opened 2016-10-24
mitmproxy: work around distutils bug

### LINKED ISSUE #6020: mitmproxy fails to build on 10.12
- [x] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

Gist with the error report available at https://gist.github.com/agatti/29e54347dfab7741447d3cd6159607e9

Thank you!

ilovezfs:
fix copied from homebrew/python/pillow

Closes #6020

CC @mikemcquaid @tdsmith @agatti

### DIFF
--- a/Formula/mitmproxy.rb
+++ b/Formula/mitmproxy.rb
@@ -255,13 +255,17 @@ class Mitmproxy < Formula
     end
 
     resource("Pillow").stage do
-      inreplace "setup.py", "'brew', '--prefix'", "'#{HOMEBREW_PREFIX}/bin/brew', '--prefix'"
-      saved_sdkroot = ENV.delete "SDKROOT"
-      begin
-        venv.pip_install Pathname.pwd
-      ensure
-        ENV["SDKROOT"] = saved_sdkroot
+      inreplace "setup.py" do |s|
+        sdkprefix = MacOS::CLT.installed? ? "" : MacOS.sdk_path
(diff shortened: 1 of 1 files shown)


ilovezfs:
@tdsmith I'm wondering if we can/should move homebrew/python/pillow into core because copying this fix around every time pillow is used doesn't really make much sense.

I'm also starting think we should require the CLT because it seems like we're generally wasting an inordinate amount of time dealing with hacks to make things work when the user only has Xcode.

MikeMcQuaid:
> @tdsmith I'm wondering if we can/should move homebrew/python/pillow into core because copying this fix around every time pillow is used doesn't really make much sense.

That seems like a good idea.

> I'm also starting think we should require the CLT because it seems like we're generally wasting an inordinate amount of time dealing with hacks to make things work when the user only has Xcode.

Is that on all Xcode versions or just some? 

The one bonus I would say of supporting just Xcode is it's made us much more resilient to Apple moving things around.

tdsmith:
If we move pillow we should consider moving numpy. There's a long thread about this in -legacy somewhere.

Other alternatives include installing pillow's prebuilt wheels or allowing cross-tap dependencies.

MikeMcQuaid:
FYI: I'd like to move basically all of homebrew/python (and other taps that don't have their own communities/maintainers) into core.

ilovezfs:
>Is that on all Xcode versions or just some?

started deteriorating in El Capitan and has gotten worse in Sierra, so that corresponds to >= Xcode 7, though the issue is the OS not the Xcode per se

PR #6230, opened 2016-10-24
sdl2 2.0.5

jlnr:

---

I have also verified that the GPG signature of the tarball matches the one from GPG Keychain's default key server.

I have only tested the formula on macOS 10.11.6, so I have no idea if the pre-Lion patches are still functional.

### DIFF
--- a/Formula/mal4s.rb
+++ b/Formula/mal4s.rb
@@ -52,7 +52,7 @@ class Mal4s < Formula
       pid = fork do
         exec bin/"mal4s", "-t", "2", "-o", "out", pkgshare/"sample--newns.mal4s"
       end
-      sleep 1
+      sleep 2
       assert File.exist?("out"), "Failed to output PPM stream!"
     ensure
       Process.kill("TERM", pid)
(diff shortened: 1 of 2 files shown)


dunn:
This wasn't caused directly by your changes, but the test for a dependent package is now failing: https://bot.brew.sh/job/Homebrew%20Core/10060/version=any_sierra/testReport/junit/brew-test-bot/sierra/test_mal4s/

If you want to take a look at adding a commit to fix that, it'll speed this up.

jlnr:
I ran `brew test --verbose mal4s` as instructed by the build bot and the test passed (after a window briefly appeared). `mal4s` generally seems to work as well.

I've rebased this branch onto master and force-pushed it, maybe the problem will fix itself.

jlnr:
Here's the failing test block:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/mal4s.rb#L50-L60

It starts mal4s in a background thread and kills it after a second. I can get this test to fail sporadically if I reduce the `sleep 1` to `sleep 0.3`. So maybe Jenkins is just a little too slow to pass the test even with `sleep 1` – I'll bump it up to `sleep 3` and see if that fixes the issue.

jlnr:
🎉

MikeMcQuaid:

PR #6258, opened 2016-10-25
mikutter 3.4.6 (new formula)

midchildan:

---

mikutter is an extensible twitter client. It is also available in the official repos of Debian, Gentoo, and BSDs.

### DIFF
--- a/Formula/mikutter.rb
+++ b/Formula/mikutter.rb
@@ -1,0 +1,249 @@
+class Mikutter < Formula
+  desc "Extensible Twitter client."
+  homepage "http://mikutter.hachune.net"
+  url "http://mikutter.hachune.net/bin/mikutter.3.4.6.tar.gz"
+  sha256 "143318ad39ec60c9cf83e206f1a940e9464f0fda19c61e5e13a6cf327d1db076"
+  head "git://toshia.dip.jp/mikutter.git", :branch => "develop"
+
+  depends_on "gtk+"
+  depends_on "terminal-notifier" => :recommended
+  depends_on :ruby => "2.1"
+
+  resource "rake" do
+    url "https://rubygems.org/gems/rake-10.5.0.gem"
+    sha256 "2b55a1ad44b5c945719d8a97c302a316af770b835187d12143e83069df5a8a49"
+  end
+
+  resource "addressable" do
+    url "https://rubygems.org/gems/addressable-2.4.0.gem"
+    sha256 "7abfff765571b0a73549c9a9d2f7e143979cd0c252f7fa4c81e7102a973ef656"
+  end
+
+  resource "pkg-config" do
+    url "https://rubygems.org/gems/pkg-config-1.1.7.gem"
+    sha256 "1f3cd171432f4634805ebf7cd187d1d728d732bfead3837c349f2c502d8e9252"
+  end
+
+  resource "glib2" do
+    url "https://rubygems.org/gems/glib2-3.0.9.gem"
+    sha256 "5981cf42e73c13b4744b5ad80eeba12a2ac5b9114d81c000c5be11dc00401f24"
+  end
+
+  resource "atk" do
+    url "https://rubygems.org/gems/atk-3.0.9.gem"
+    sha256 "aba2b0201cfaecd9bd82437794aba9b4b3a6b6214fbd1c2264c2294d8ef4fa39"
+  end
+
+  resource "cairo" do
+    url "https://rubygems.org/gems/cairo-1.15.2.gem"
+    sha256 "954cb9bf7ac7cc79a5cc805be61ca81378454f20f621939fa6afb4513cab50d9"
+  end
+
+  resource "safe_yaml" do
+    url "https://rubygems.org/gems/safe_yaml-1.0.4.gem"
+    sha256 "248193992ef1730a0c9ec579999ef2256a2b3a32a9bd9d708a1e12544a489ec2"
+  end
+
+  resource "crack" do
+    url "https://rubygems.org/gems/crack-0.4.3.gem"
+    sha256 "5318ba8cd9cf7e0b5feb38948048503ba4b1fdc1b6ff30a39f0a00feb6036b29"
+  end
+
+  resource "delayer" do
+    url "https://rubygems.org/gems/delayer-0.0.2.gem"
+    sha256 "39ece17be3e4528d562a88aef7cb25143ef4ce77df2925a7534f8a163af1db94"
+  end
+
+  resource "delayer-deferred" do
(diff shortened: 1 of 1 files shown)


Formula/mikutter.rb
```diff
@@ -0,0 +1,378 @@
...
+
+  def install
+    mkpath libexec/"vendor"
+    resources.each do |r|
+      r.verify_download_integrity(r.fetch)
+      system("gem", "install", r.cached_download, "--no-document",
+             "--install-dir", "#{libexec}/vendor")
+    end
+
+    prefix.install_metafiles
```
dunn: If you're not changing the working directory during installation, this happens automatically.

Formula/mikutter.rb
```diff
@@ -0,0 +1,378 @@
...
+    rm_rf "vendor"
```
dunn: This shouldn't be necessary.

Formula/mikutter.rb
```diff
@@ -0,0 +1,378 @@
...
+    libexec.install Dir["*"]
+
+    if build.with? "terminal-notifier"
+      mkpath libexec/"plugin"
```
dunn: `#write` will create the path if necessary, so this and the rest of the `mkpath`s aren't needed: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/pathname.rb#L141

Formula/mikutter.rb
```diff
@@ -0,0 +1,378 @@
+class Mikutter < Formula
+  desc "Extensible Twitter client."
+  homepage "http://mikutter.hachune.net"
+  url "http://mikutter.hachune.net/bin/mikutter.3.4.6.tar.gz"
+  sha256 "143318ad39ec60c9cf83e206f1a940e9464f0fda19c61e5e13a6cf327d1db076"
+  head "git://toshia.dip.jp/mikutter.git", :branch => "develop"
+
+  depends_on "gtk+"
+  depends_on "terminal-notifier" => :recommended
+  depends_on :ruby => "2.1+"
```
dunn: The RubyRequirement doesn't use the `+`, it should just be `"2.1"` and that means 2.1 or higher.

Formula/mikutter.rb
```diff
@@ -0,0 +1,378 @@
...
+    mkpath prefix/"mikutter.app/Contents/Resources"
+    system("/usr/bin/sips", "-s", "format", "icns",
+           libexec/"core/skin/data/icon.png",
+           "--out", "#{prefix}/mikutter.app/Contents/Resources/mikutter.icns")
+
+    (prefix/"mikutter.app/Contents/Info.plist").write(info_plist)
+  end
+
+  def exec_script
+    <<-EOS.undent
```
bfontaine: You should submit them to upstream instead.

MikeMcQuaid: > You should submit them to upstream instead.

Agreed. Upstream need to be willing to support macOS.

Formula/mikutter.rb
midchildan: You mean `prefix.install_metafiles` ?

bfontaine: @midchildan Yes.

MikeMcQuaid:
As a general comment: this formula is kinda a full port to macOS. I'd strongly recommend workign with upstream to get all this stuff (including e.g. icns generation) upstream and ideally get a .app bundle that they can build and support.

Formula/mikutter.rb
```diff
@@ -0,0 +1,378 @@
...
+          if user
+            user_img = Gdk::WebImageLoader.local_path(user[:profile_image_url])
+            metadata[:title] = "@\#{user[:idname]} (\#{user[:name]})"
+            metadata[:contentImage] = user_img
+          end
+          text = text.to_s if text.is_a? Message
+          metadata[:message] = text
+
+          command = [
+            "#{HOMEBREW_PREFIX}/opt/terminal-notifier/bin/terminal-notifier",
```
bfontaine: You can’t rely on this being present because it’s only `:recommended`; not required.

Formula/mikutter.rb
bfontaine: You don’t have to depend on `HOMEBREW_PREFIX`; just use the binaries available in the `PATH`.

Formula/mikutter.rb
bfontaine: Even if you could you should use `terminal-notifier` instead of its full path.

midchildan:
@MikeMcQuaid The compatibility plugin in this formula could easily be provided by upstream or elsewhere, but it is not so easy for the rest. First, mikutter depends on formulas provided by Homebrew, and you will have to include every single one of them to package it as a standalone `.app` bundle. I'm not sure the upstream will be willing to do that. Second, mikutter depends on Hombrew-specific info such as the `HOMEBREW_PREFIX`. This is necessary because apps launched from the Finder will not include Homebrew's `PATH` by default. 
But the approach I used to make an `.app` bundle is applicable to other formulas as well.  Most of the wrapper script I wrote for launching mikutter is about dealing with differences in the way macOS and other common unix systems handle locales. Specifically, many unix apps depends on `LANG` being set to a UTF-8 locale and will not work if `LANG` is unspecified. Can't we perhaps add a helper method to make a `.app` bundle to wrap unix GUI apps?

Formula/mikutter.rb
midchildan: I could do that, but I would still need `HOMEBREW_PREFIX` for reasons given below.

midchildan:
I opened a PR in Homebrew/brew#1383

midchildan:
Fixed formula.

Formula/mikutter.rb
```diff
@@ -0,0 +1,210 @@
...
+  head "git://toshia.dip.jp/mikutter.git", :branch => "develop"
+
+  depends_on "gtk+"
+  depends_on "terminal-notifier" => :recommended
+  depends_on :ruby => "2.1"
+
+  resource "rake" do
+    url "https://rubygems.org/gems/rake-10.5.0.gem"
+    sha256 "2b55a1ad44b5c945719d8a97c302a316af770b835187d12143e83069df5a8a49"
+  end
```
MikeMcQuaid: Can you add a blank line between each of these? Thanks!

MikeMcQuaid (changes requested):

Formula/mikutter.rb
midchildan: @MikeMcQuaid done😁

MikeMcQuaid (approved):

MikeMcQuaid:

Issue #5488, opened 2016-10-01
Formulae without Sierra bottles


zmwangx:
Jenkins finally managed to chew through the daunting Sierra build queue. Below are the remaining ones that defied (not necessarily only) Sierra for one reason or another.

The list is generated with the script https://gist.github.com/zmwangx/89f07670c8e42f58434bd6d15653ecf6. You can also pass a blacklist to the script to get a more workable list (e.g., putting `qt` and `qt5` in the blacklist blocks the entire Qt tree). See `need-bottle -h`.

---

Formulae that made the top 5000 install events chart during September 2016, ordered by popularity:

- [x] qt5
- [x] xctool (https://github.com/facebook/xctool/issues/716)
- [ ] swiftlint (swift 2.2, https://github.com/realm/SwiftLint/issues/815)
- [x] pypy ([upstream issue](https://bitbucket.org/pypy/pypy/issues/2407/pypy-541-fails-to-build-on-macos-sierra), #5689)
- [ ] pyqt (qt, #1705)
- [x] pyqt5
- [x] qscintilla2 (https://github.com/Homebrew/homebrew-core/pull/5786)
- [ ] swiftgen (swift 2.2, https://github.com/AliSoftware/SwiftGen/issues/169)
- [x] qwt
- [x] allegro (#6017)
- [x] pyside
- [x] shiboken
- [x] gnuradio
- [x] upx (#6264, https://sourceforge.net/p/upx/bugs/248/, https://github.com/upx/upx/issues/4)
- [ ] infer (need to wait for 0.9.4 according to #5130)
- [x] jack (#5487)
- [x] qca
- [x] qwtpolar
- [x] binwalk
- [x] git-cola (#6210)
- [x] folly (#5548)
- [x] gpsbabel
- [x] qcachegrind
- [x] radare2
- [x] libewf (libyal/libewf#60, #5737)
- [x] dvm (#5587)
- [x] irrlicht (#5592)
- [x] tcptrace (#6223)
- [ ] imagesnap (project dead since 2011, but 10.11 bottle works on 10.12 unmodified)
- [x] uwsgi (https://github.com/Homebrew/homebrew-core/pull/5528)
- [x] log4cxx
- [ ] capnp (#5773, https://github.com/sandstorm-io/capnproto/issues/375)
- [ ] swift (need to upgrade to 3.0)
[...]

PR #6268, opened 2016-10-25
swift: migrate to boneyard

zmwangx:


MikeMcQuaid:
Why?

zmwangx:
Swift is provided by Apple through Xcode or CLT. Not many people install this formula, nothing depends on it, no one cared to bring it to 3.0, and the last time I tried to bring it up-to-date with [this patch](https://gist.github.com/zmwangx/0075ced732314f1902bdc12ef6e720e4) it [errored out](https://gist.github.com/zmwangx/85c741a6b1b167ea6d7e70d0639e3c3c) in a non-obvious way. Lacking demand and enthusiasm, let just make it history.

zmwangx:
@MikeMcQuaid I was just about to CC you since you originally added it.

MikeMcQuaid:
We're running a pretty old release that's probably worth trying to upgrade.

zmwangx:
See comment above. If you can get it to work then sure.

zmwangx:
My patch is a starting point: https://gist.github.com/zmwangx/0075ced732314f1902bdc12ef6e720e4.

MikeMcQuaid:
Given we've got some people installing/using this: 👎 on removing.

zmwangx:
Then bring it to 3.0.

zmwangx:
@MikeMcQuaid I take that to mean you'll fix it? Because I've done my fair share of work to update it, and don't plan to waste more human and CPU time on something with little return (the official channel is better). I also don't want to see swiftin #5488 forever.

MikeMcQuaid:
> I take that to mean you'll fix it?

Not necessarily but I'll take a look.

> Because I've done my fair share of work to update it, and don't plan to waste more human and CPU time on something with little return (the official channel is better).

That's up to you (as my work is up to me). We shouldn't boneyard things people are using. I won't object when no-one is using them.

> I also don't want to see swift in #5488 forever.

A maximum macOS requirement is the best short-term fix for these cases, I think.

zmwangx:
> We shouldn't boneyard things people are using.

Running `brew install swift` is not equivalent to using `/usr/bin/local/swift` (well, it's not even linked by default) regularly, especially when `/usr/bin/swift` is up-to-date and infinitely better supported (in this respect the formula swift is not comparable to any other dupe we have). A long time ago, when swift was initially open sourced, I tried `brew install swift` just out of curiosity. I surely didn't use it afterwards; in fact, I haven't even got past page fifty of *The Swift Programming Language*, despite having kept it in iBooks since Swift 1.0 beta. Given the enthusiasm in Swift in general and our relatively pathetic install stats, I wouldn't be surprised if the great majority of those install numbers come from novelty trials like mine.

Even if there are people using it, there's always `/usr/bin/swift` (or swift under whatever sdk path), so it's not like we're leaving people in the dark by boneyarding this formula. In fact, I would urge them to use Apple's swift.

MikeMcQuaid:
This isn't a conversation specific to Swift. We shouldn't boneyard things that aren't unsupported by upstream, are still installed by users and run on at least 1-2 of our supported OS versions.

zmwangx:
> This isn't a conversation specific to Swift.

Well it is, or at least I'm not arguing about anything else. Swift is special in that it is provided by Apple, and is already installed by either Xcode or CLT, which we require at least one. If we can keep it up to date then sure, but it has been having trouble for a long time and no one bothered to fix it.

What we support is ultimately decided by what is feasible for the team to support (including what outside help we can solicit), not what non-contributing users wish for. Also, in this case I doubt anyone cares about the removal of the swift formula.

MikeMcQuaid:
> Also, in this case I doubt anyone cares about the removal of the swift formula.

I care about the removal of things that aren't unsupported by upstream, are still installed by users and run on at least 1-2 of our supported OS versions.

zmwangx:
What people are running (suppose they're actually running it) is an outdated version.

PR #6307, opened 2016-10-27
swift: handle changed --build-jobs arg name in HEAD

apjanke:
  - #4300 - this picks up where that one left off

---

The `swift` HEAD build is currently broken because the `build-script` `--build-jobs` argument changed to `--jobs`. This fixes it.

Also fixes the issue where the local variable `build` was masking the `build` method in the `swift` formula, by renaming it out of the way to `build_dir`.

Picks up where #4300 left off.

### DIFF
--- a/Formula/swift.rb
+++ b/Formula/swift.rb
@@ -59,13 +59,14 @@ class Swift < Formula
 
   def install
     workspace = buildpath.parent
-    build = workspace/"build"
+    build_dir = workspace/"build"
 
     ln_sf buildpath, "#{workspace}/swift"
     resources.each { |r| r.stage("#{workspace}/#{r.name}") }
 
-    mkdir build do
-      system "#{buildpath}/utils/build-script",
+    mkdir build_dir do
+      args = [
+        "#{buildpath}/utils/build-script",
         "-R",
         "--build-subdir=",
         "--no-llvm-assertions",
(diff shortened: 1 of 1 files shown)


DomT4:
I forgot to say anything about the Swift formula before I ran off, my apologies. Just remembered when I saw your email notification _(not directed at me)_ from another thread.

Unless was some local configuration issue this version of Swift doesn't build against Xcode 8, and at the time I couldn't get the newest stable version of Swift to build on Sierra either.

apjanke:
Ah, I was building it against Xcode 7.3.1 on my box.

No worries.

I think we should still do this PR, because it will fix some build situations which are broken by things besides Xcode 8 being installed.

DomT4:
Not my decision to be part of any more 😉, just wanted to mention the build issue in case you wanted to skip CI or wait for CI and then try to debug the other issues, etc.

apjanke:
Cool. Thanks for the heads-up. Since this is HEAD, the CI won't actually test all of it. Maybe I screwed up in my local testing; I'll re-check that.

Formula/swift.rb
```diff
@@ -59,13 +59,14 @@ class Swift < Formula
...
+    build_dir = workspace/"build"
 
     ln_sf buildpath, "#{workspace}/swift"
     resources.each { |r| r.stage("#{workspace}/#{r.name}") }
 
-    mkdir build do
-      system "#{buildpath}/utils/build-script",
+    mkdir build_dir do
+      args = [
+        "#{buildpath}/utils/build-script",
```
MikeMcQuaid: Reads a bit better to keep this as part of the `system` call, I think.

MikeMcQuaid (changes requested):

MikeMcQuaid (approved):

MikeMcQuaid:
One nit but otherwise 👍 and feel free to merge without.

MikeMcQuaid:
Fixed by https://github.com/Homebrew/homebrew-core/pull/6645

Issue #3219, opened 2016-07-22
Qt 5.6.1-1: macdeployqt doesn't fixup bundle dependencies in plugins


rockihack:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

### Bug reports:

macdeployqt does not work as expected in `homebrew qt 5.6.1-1`,
though it did work in `homebrew qt 5.6.0`.

macdeployqt copies plugins (e.g. `platforms/libqcocoa.dylib`) but fails to fixup any plugin bundle dependencies with `install_name_tool -change ...`. I verified the failure with `otool -L ...` (see below). However macdeployqt still fixes the main executable ~~and bundled frameworks~~ as expected.
EDIT: Also fails to fixup bundle dependencies in frameworks. Only the Id get fixed.

### Does not work:
* homebrew macdeployqt (5.6.1-1) [app compiled with homebrew qt 5.6.1-1]

### Does work:
* homebrew macdeployqt (5.6.0) [app compiled with homebrew qt 5.6.0]
* homebrew macdeployqt (5.6.0) [app compiled with homebrew qt 5.6.1-1]
* official qt.io macdeployqt (5.6.1-1) [app compiled with official qt.io 5.6.1-1]

The problem must be with homebrew qt 5.6.1-1 since official qt.io macdeployqt 5.6.1-1 works fine.

**Notice the difference in verbose output of homebrew macdeployqt:**

5.6.1-1:
`Log:  change reference "/usr/local/opt/qt5/lib/QtGui.framework/Versions/5/QtGui"`

5.6.0:
`Log:  change reference "/usr/local/Cellar/qt5/5.6.1-1/lib/QtGui.framework/Versions/5/QtGui"`

### homebrew macdeployqt 5.6.1-1 (error):

```
Log:
Log: Deploying Qt frameworks found inside: ("src/KeePassX.app/Contents/PlugIns/platforms/libqcocoa.dylib")
Log: Using install_name_tool:
Log:  in "src/KeePassX.app/Contents/PlugIns/platforms/libqcocoa.dylib"
Log:  change reference "/usr/local/opt/qt5/lib/QtGui.framework/Versions/5/QtGui"
Log:  to "@executable_path/../Frameworks/QtGui.framework/Versions/5/QtGui"
Log: Using install_name_tool:
[...]

PR #6309, opened 2016-10-27
qt5: link brewed openssl when using postresql or mysql

### LINKED ISSUE #6032: qt5 fails to build with mysql support
### Please follow the general troubleshooting steps first:
- [ ] ~~If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?~~  N/A
### Bug reports:

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;As noted in [this comment](https://github.com/Homebrew/homebrew-core/issues/5947#issuecomment-254346583) on [Issue #5947:  'qt5 fails to upgrade to 5.6.2 on Sierra with Xcode 8.'](https://github.com/Homebrew/homebrew-core/issues/5947), I split this off from that bug report since I'm seeing issues on a different version of OS X/macOS.  [Here](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/tree/master/Qt/v5.7.0/First%20Recorded%20Attempt) are my logs.  I'm guessing the presence of a `build.sandbox.log` indicates that my problem is most likely due to some kind of sandbox violation.  Please note, however, that these are from a `brew reinstall` attempt since I thought the `brew upgrade` attempts I made beforehand might be failing due to the possibility that formula changes other than the version bump could have affected these previous build attempts (there was more than one.)  Unfortunately, this overwrote the earlier logs I had, so I can't provide them.

apjanke:

---

Fixes #6032.
Fixes #6319.

This fixes a build failure that happens when `--with-mysql` or `--with-postgresql` is specified. Does not affect the default build.

Looks like Qt5 needs to be linked against openssl when compiling in mysql or postgresql support.

This PR conditionally links `qt5` against the brewed `openssl` when mysql or postgresql support is enabled. It does the `-openssl-linked` option to ensure that the brewed openssl is linked at compile time and is always used. (The default is for Qt to detect installed openssl libraries at run time, which may pick up the system openssl instead.)

I'm wondering if we should actually make the `openssl` dependency and linkage unconditional (instead of just when using mysql or postgresql), to avoid the run-time detection possibly picking up the system openssl in other circumstances. (I don't know in what conditions that would actually happen; didn't find enough documentation on it.)

This is something of a followup to #5619, including only the openssl-related code. It differs in that it:
- does `-openssl-linked` as described above
- does not do `-no-opentransport`. I don't think it is necessary, because we don't have a reason to disable OpenTransport and always use OpenSSL instead; we just need to make sure that for the cases which do use OpenSSL, it a) builds, and b) uses the brewed `openssl` instead of the system one.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -83,6 +83,7 @@ class Qt5 < Formula
   depends_on :macos => :mountain_lion
 
   depends_on "dbus" => :optional
+  depends_on "openssl" if build.with?("mysql") || build.with?("postgresql")
   depends_on :mysql => :optional
   depends_on :postgresql => :optional
   depends_on :xcode => :build
@@ -150,6 +151,14 @@ class Qt5 < Formula
     if build.with? "qtwebkit"
       (buildpath/"qtwebkit").install resource("qt-webkit")
       inreplace ".gitmodules", /.*status = obsolete\n((\s*)project = WebKit\.pro)/, "\\1\n\\2initrepo = true"
+    end
+
+    if build.with?("mysql") || build.with?("postgresql")
+      openssl_opt = Formula["openssl"].opt_prefix
+      args << "-L#{openssl_opt}/lib"
+      args << "-I#{openssl_opt}/include"
+      args << "-openssl-linked"
+      ENV["OPENSSL_LIBS"] = "-L#{openssl_opt}/lib -lssl -lcrypto"
     end
 
     system "./configure", *args


ilovezfs:
>I'm wondering if we should actually make the openssl dependency and linkage unconditional (instead of just when using mysql or postgresql), to avoid the run-time detection possibly picking up the system openssl in other circumstances. (I don't know in what conditions that would actually happen; didn't find enough documentation on it.)

Did you notice if anything else picked up the linkage?

apjanke:
> Did you notice if anything else picked up the linkage?

No, in my tests it doesn't pick up the `ssl` linkage unless you specify `--with-mysql` or `--with-postgresql`. But my understanding from reading the Qt documentation and its `./configure --help` output is that, when needed, Qt will search for OpenSSL at run time and load it dynamically (like a plug-in, not a regular DLL), so the linkage list may not reflect the libraries it will actually use when run.

The `-openssl-linked` option causes it to use build-time linkage instead of run-time linking/loading, which is why I included it: as I understand it, doing it this way means the linkage list dumped by `otool -L` really does reflect which OpenSSL libs might be used at run time.

yezezey:
I already lost appetite for Homebrewed Qt5 because of this problem https://github.com/Homebrew/homebrew-core/issues/3219, but what's going on here?

Qt5 can not be linked with both OpenSSL and Secure Transport.
If you want to use OpenSSL, you must pass -no-securetransport to the configure script.

```
$ ./configure --help
(snip)
 *  -securetransport ... Use SecureTransport instead of OpenSSL

    -no-securetransport  Do not use SecureTransport, either use OpenSSL or do not use any SSL backend
                         at all (if combined with -no-openssl).
(snip)
```

yezezey:
And static link `-openssl-linked` is not necessary.

> Qt will search for OpenSSL at run time and load it dynamically (like a plug-in, not a regular DLL)

This is wrong. Qt will search a plugin (e.g. QtNetwork) at run time and load it dynamically,
after that the OS load OpenSSL as a regular shared library.

If I understand macOS correctly, a load module does not pick up wrong shared libraries at run time.

apjanke:
It seems that even with the default `-securetransport`, some Qt components, specifically the mysql and postgresql plugins, will still link against OpenSSL. This PR is an attempt to get them to discover the Homebrew-installed OpenSSL instead of the system OpenSSL.

When building the mysql plugin, probing for mysql in `configure` fails because Homebrew's mysql is built with openssl support, and `mysql_config --libs`, which Qt5's `configure` script uses, reports that `-lssl -lcrypto` are needed. This PR makes the probe succeed. @steve3d is working on an alternate (probably cleaner) approach in #6319.

When building the postgresql plugin, the build succeeds, but some of the binaries in the Qt5 installation end up linked against the deprecated system openssl. With this PR, they end up linked against the Homebrew-installed openssl. (I'll grab a list of the exact binaries when my computer gets around to doing a `brew install qt5 --with-postgresql` build again.)

I'm basing this on reading Qt's `./configure --help` output and these pages.
 * http://doc.qt.io/qt-5/ssl.html
 * http://doc.qt.io/qt-5/build-sources.html#openssl-v1later

I'm still not clear on how all the options interact. If there's more detailed documentation out there on Qt's build system (and specifically how it interacts with OpenSSL/SecureTransport, and how plugins work), it would be great to have a link.

yezezey:
> It  seems that even with the default -securetransport, some Qt components, specifically the mysql and postgresql plugins, will still link against OpenSSL. 

Because those plugins do not support Secure Transport.

> This PR is an attempt to get them to discover the Homebrew-installed OpenSSL instead of the system OpenSSL.

To achieve that, pass -no-securetransport to the configure script.

>  @steve3d is working on an alternate (probably cleaner) approach in #6319.

What he wants is to eliminate all OpenSSL dependency. 

> When building the postgresql plugin, the build succeeds, but some of the binaries in the Qt5 installation end up linked against the deprecated system openssl. With this PR, they end up linked against the Homebrew-installed openssl. 

Because you didn't use -no-securetransport.

steve3d:
@yezezey, what I'm trying to do is not to eliminate all OpenSSL dependency, I'm just remove OpenSSL dependency from MySQL plugin, so that MySQL plugin can be build. Please check my pr again.

yezezey:
@steve3d

MySQL plugin uses OpenSSL for connecting to a server, doesn't it?

```
$ otool -L plugins/sqldrivers/libqsqlmysql.dylib
libqsqlmysql.dylib:
	libqsqlmysql.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/opt/mysql/lib/libmysqlclient.20.dylib (compatibility version 20.0.0, current version 20.0.0)
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/qt5.7.0-mysql/lib/QtSql.framework/Versions/5/QtSql (compatibility version 5.7.0, current version 5.7.0)
	/usr/local/qt5.7.0-mysql/lib/QtCore.framework/Versions/5/QtCore (compatibility version 5.7.0, current version 5.7.0)
	/System/Library/Frameworks/DiskArbitration.framework/Versions/A/DiskArbitration (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.0.0)
```

yezezey:
Like I said before in https://github.com/Homebrew/homebrew-core/pull/5619#discussion_r82514760

`$ OPENSSL_LIBS='-L/usr/local/opt/openssl/lib -lssl -lcrypto' ./configure -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib -no-securetransport`

is the minimal requirement to properly link against Homebrewed OpenSSL.
Qt's configure script is so stupid if you omit any of above options, you will end up with wrong OpenSSL.

I don't still see why #6319 is cleaner solution.
If MySQL plugin wants to link with OpenSSL, why not properly link against brewed OpenSSL like this?
```
$ otool -L plugins/sqldrivers/libqsqlmysql.dylib | grep openssl
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
$ otool -L lib/QtNetwork.framework/QtNetwork | grep openssl
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
```

PR #6319, opened 2016-10-27
qt5: fix build with mysql/postgresql of openssl and add pch support

### LINKED ISSUE #6032: qt5 fails to build with mysql support

steve3d:

---

I know there is another PR about this. The only reason that mysql/postgresql failed to build is there is no `libssl.dynlib` to find when build mysql/postgresql. 

In my previous PR, I added `-L/usr/local/opt/openssl/lib` to global LFLAGS, this will cause all library will have this link flags, for example, if you only link QtCore, there will be this option exists and does nothing.

Qt's configure script need `mysql_config` and `pg_config` to give the link flags, but default flags of `mysql_config --libs` gives the result of `-L/usr/local/Cellar/mariadb/10.1.18/lib  -lmysqlclient -lz -lssl -lcrypto`, but the libssl and libcrypto is not placed on /usr/local/lib, so link will fail because linker have no idea where to find the libssl.

and I really don't know how can I change the result of `mysql_config`, so seems the only solution is to patch the configure script to manually add openssl's lib folder.

as to another option in my previous PR, to build plugins into QtSql is trivial, so I decided no to include those changes, and that is purely my own personal favorite ways of building qt.

also I added the pch option, I think this won't have any other effects but to speed up the compile time.

Fixes #6032.
Fixes #6319.
Fixes #6309.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -131,13 +131,10 @@ class Qt5 < Formula
 
     if build.with? "mysql"
       args << "-plugin-sql-mysql"
-      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s|-lssl|-L#{openssl_lib} -lssl|\"\`"
+      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"
     end
 
-    if build.with? "postgresql"
-      args << "-plugin-sql-psql"
-      inreplace "qtbase/configure", /QT_LFLAGS_PSQL=\`(.*)\`/, "QT_LFLAGS_PSQL=\`\\1 | sed \"s|-lssl|-L#{openssl_lib} -lssl|\"\`"
-    end
+    args << "-plugin-sql-psql" if build.with? "postgresql"
 
     if build.with? "dbus"
       dbus_opt = Formula["dbus"].opt_prefix
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -121,7 +121,6 @@ class Qt5 < Formula
       -qt-pcre
       -nomake tests
       -no-rpath
-      -system-proxies
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -121,14 +121,11 @@ class Qt5 < Formula
       -qt-pcre
       -nomake tests
       -no-rpath
-      -pch
       -system-proxies
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"
     
-    openssl_lib = Formula["openssl"].opt_lib
-
     if build.with? "mysql"
       args << "-plugin-sql-mysql"
       inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -121,12 +121,23 @@ class Qt5 < Formula
       -qt-pcre
       -nomake tests
       -no-rpath
+      -pch
+      -system-proxies
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"
+    
+    openssl_lib = Formula["openssl"].opt_lib
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
-    args << "-plugin-sql-psql" if build.with? "postgresql"
+    if build.with? "mysql"
+      args << "-plugin-sql-mysql"
+      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s|-lssl|-L#{openssl_lib} -lssl|\"\`"
+    end
+
+    if build.with? "postgresql"
+      args << "-plugin-sql-psql"
+      inreplace "qtbase/configure", /QT_LFLAGS_PSQL=\`(.*)\`/, "QT_LFLAGS_PSQL=\`\\1 | sed \"s|-lssl|-L#{openssl_lib} -lssl|\"\`"
+    end
 
     if build.with? "dbus"
       dbus_opt = Formula["dbus"].opt_prefix
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -125,7 +125,7 @@ class Qt5 < Formula
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"
-    
+
     if build.with? "mysql"
       args << "-plugin-sql-mysql"
       inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"


steve3d:
if you guys think the system-proxies is not acceptable, I can add another commit to make it a option.

ilovezfs:
>if you guys think the system-proxies is not acceptable, I can add another commit to make it a option.

It's fine, I think.

>Qt's configure script need mysql_config and pg_config to give the link flags, but default flags of mysql_config --libs gives the result of -L/usr/local/Cellar/mariadb/10.1.18/lib -lmysqlclient -lz -lssl -lcrypto, but the libssl and libcrypto is not placed on /usr/local/lib, so link will fail because linker have no idea where to find the libssl.

Do you think this is a bug in the output of mysql_config? Even if you do not know how to fix it, would it be preferable to try to fix it there instead? (Side note: it also seems bad that those flags are using the Cellar path to mariadb not the opt path)

Also, is it the same problem with postgresql's config output?

I notice the other PR does some things this PR does not:

1. Specifies location of the headers
```
"-I#{openssl_opt}/include"
```

2. Sets the OPENSSL_LIBS environment variable
```
ENV["OPENSSL_LIBS"] = "-L#{openssl_opt}/lib -lssl -lcrypto"
```

3. Specifies `-lcrypto` in addition to `-lssl`
(See above)

Why are any/all of these things unnecessary and/or undesirable?

ilovezfs:
Also, I left out
4) passes `-openssl-linked`

steve3d:
as you can see, I only added the `-L` option, because include mysql headers and postgresql headers won't force one to include openssl headers. so include those headers is really redundant.

mysql/postgresql only use openssl for connections, there is no need for their header files. 

and I just made another tests, this is definitely not qt's problem, it's because `mysql_config --libs` has `-lssl -lcrypto` caused this, I've just removed these option, also a successful build.

so mysql/postgresql plugin don't need to include any openssl headers nor link to openssl libs.

steve3d:
I've added another commit, as it turns out postgresql don't need special care at all. qt only picks postgresql's libdir instead of libs, so the only thing need to care is to remove `-lssl -lcrypto` from `mysql_config --libs` and `mysql_config --libs_r`, after removed these two baby, everything goes fine. you don't need openssl headers nor libs to build mysql plugin for qt at all.

Formula/qt5.rb
```diff
@@ -121,11 +121,19 @@ def install
       -qt-pcre
       -nomake tests
       -no-rpath
+      -pch
+      -system-proxies
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"
+    
+    openssl_lib = Formula["openssl"].opt_lib
```
apjanke: It looks like this `openssl_lib` variable is unused.

apjanke:
I'm not sure this PR does everything we need. With this PR, building with postgresql support succeeds, but the resulting binaries are linked against the system OpenSSL instead of Homebrew's OpenSSL. That's undesirable.

```
$ brew install qt5 --with-postgresql
...
$ brew linkage qt5 | grep ssl
  /usr/lib/libssl.0.9.8.dylib
$ brew audit --strict qt5
qt5:
  * 129: Trailing whitespace was found
    These object files were linked against the deprecated system OpenSSL or
    the system's private LibreSSL.
      /usr/local/Cellar/qt5/5.7.0/plugins/sqldrivers/libqsqlmysql.dylib
...
```

You might also be able to do this more cleanly by replacing calls to `mysql_config` with calls to `pkg-config --libs mysqlclient`. That one doesn't include `-lssl -lcrypto` unless you specify `--static`.

```
[~] $ mysql_config --libs
-L/usr/local/Cellar/mysql/5.7.16/lib -lmysqlclient -lssl -lcrypto
[~] $ pkg-config --libs mysqlclient
-L/usr/local/Cellar/mysql/5.7.16/lib -lmysqlclient
[~] $ pkg-config --libs mysqlclient --static
-L/usr/local/Cellar/mysql/5.7.16/lib -lmysqlclient -lssl -lcrypto
```

apjanke:
> Do you think this is a bug in the output of mysql_config?

This does seem to be an issue with `mysql_config`.

```
[~] $ mysql_config --libs
-L/usr/local/Cellar/mysql/5.7.16/lib -lmysqlclient -lssl -lcrypto
```

It's built against the brewed `openssl` instead of the system `openssl`, so it seems that their `*_config --libs` output should be pointing at the brewed openssl. For brewed programs built against this package, I don't think it matters, because those that actually link directly against openssl will have a `depends_on "openssl"` which takes care of it. But non-brewed packages might end up linking against the system openssl libs, which could be a problem. I would think that `mysql_config --libs` should include a `-L` for the brewed openssl if it's going to include `-lssl -lcrypto`.

I don't know why the `qt5` `configure` probe using `mysql_config` failed in the first place. I would think that it would have picked up the system openssl libraries, which are on the library path by default.

At first glance, `pg_config` seems to have the same problem. But its semantics are different; I don't know if `pg_config --libs` is being used the same way by clients.

yezezey:
How cad I use pch?
Is Homebrew supposed to be used like that?
When I do `brew install -s` the first time, where goes the pch?
When I do `brew install -s` in the second time, how can I use the pch?

steve3d:
@yezezey there is no way you can use pch automatically, unless you pass -pch to configure, and pch is default off on any platform except windows

yezezey:
@steve3d 
So what's the point of including -pch in this PR?

steve3d:
pch = speed up compiler

yezezey:
So how can I speed up compiler using pch with this PR?
Passing -pch to the configure script automagically speed up the compile speed?

steve3d:
I think google is your best friend for this question.

steve3d:
@apjanke, pkg-config give me this:

```
imac:~ steve$ pkg-config --libs mariadb
-L/usr/local/Cellar/mariadb/10.1.18/lib -lmysqlclient -lz -lssl -lcrypto
```

they also have `-lssl -lcrypto`, as for postgresql on my system, I have different result:

```
otool -L libqsqlpsql.dylib 
libqsqlpsql.dylib:
	libqsqlpsql.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/opt/postgresql/lib/libpq.5.dylib (compatibility version 5.0.0, current version 5.8.0)
	@rpath/QtSql.framework/Versions/5/QtSql (compatibility version 5.7.0, current version 5.7.0)
	@rpath/QtCore.framework/Versions/5/QtCore (compatibility version 5.7.0, current version 5.7.0)
	/System/Library/Frameworks/DiskArbitration.framework/Versions/A/DiskArbitration (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.0.0)
```

there is no openssl at all.

steve3d:
added a new commit, which removes unnecessary pch option, and also removed unused variable about openssl.

qt configure script will automatically enable pch. so force it to enable it is not necessary now.

RandomDSdevel:
@steve3d:  You misspelled 'trailing' as 'tailing' in the message delineating your most recent commit  for this PR.  Also, I agree that, per https://github.com/Homebrew/homebrew-core/pull/6374#issuecomment-258236345, any changes with respect to mucking around with how Qt handles generating PCH files while it's getting built that you might consider making, if any are even _needed_ (you've concluded that they no longer are,) should be split out of this PR.

Formula/qt5.rb
```diff
@@ -121,11 +121,16 @@ def install
       -qt-pcre
       -nomake tests
       -no-rpath
+      -system-proxies
```
MikeMcQuaid: Remove this for now and feel free to open another PR with it in.

Formula/qt5.rb
```diff
@@ -121,11 +121,16 @@ def install
...
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
+    if build.with? "mysql"
+      args << "-plugin-sql-mysql"
+      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"
```
MikeMcQuaid: Does this need a `depends_on "openssl"`?

MikeMcQuaid (changes requested):

Formula/qt5.rb
steve3d: no, this only removes openssl from build time(build time for mysql plugin), libmysqlclient still need openssl at runtime.

ilovezfs (approved):

MikeMcQuaid (approved):

MikeMcQuaid:

@steve3d @yezezey @apjanke Thanks for all your work on this. I just pulled this PR because it was done first.

ilovezfs:
@steve3d Thanks for doing this :white_check_mark:. Don't forget to open the `-system-proxies` is you're still interested in having us consider that change.

PR #6332, opened 2016-10-27
ffmpeg 3.2

zmwangx:

---

See https://github.com/FFmpeg/FFmpeg/blob/n3.2/Changelog.
- `openh264` 1.6 support added (FFmpeg/FFmpeg@293676c); drop vendored `openh264` 1.5.
- <s>Restore the `--with-ffplay` option and deprecate `--with-sdl`/`--with-sdl2` (recently added).  Whatever backend `ffplay` uses is just an implementation detail, users should not need to know it (and most don't), and using the dependency name in the option name not only causes user confusion (especially after years of `--with-ffplay` which is the only name that makes sense), it's also a headache when upstream switches backend, as is happening with this release.</s>
- Deprecate `--with-sdl` in favor of `--with-sdl2`, and point `--with-ffplay` to `--with-sdl2` too. FFplay now uses `sdl2` as its backend.
- Drop the AAC encoding caveats specific to the FFmpeg 3.0 release. It's been out for the better part of a year and shouldn't be news anymore. Users concerned with this caveat should have migrated already.
- Sort options and optional dependencies. They're a mess and I didn't even know where to insert options and dependencies.

More revision bumps are probably necessary. Let's see what the bot has to say.

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -1,26 +1,9 @@
 class Ffmpeg < Formula
   desc "Play, record, convert, and stream audio and video"
   homepage "https://ffmpeg.org/"
-
-  stable do
-    url "https://ffmpeg.org/releases/ffmpeg-3.1.5.tar.bz2"
-    sha256 "2400882a2c7795c74f0abebc28d267f2796510fb69ba324b0e9f16cc8dbb0d2b"
-
-    option "with-sdl", "Enable FFplay media player"
-    option "with-openh264", "Enable OpenH264 library"
-    deprecated_option "with-ffplay" => "with-sdl"
-
-    depends_on "sdl" => :optional
-    depends_on "nasm" => :build if build.with? "openh264"
-
-    # Remove when ffmpeg has support for openh264 1.6.0
-    # See https://github.com/cisco/openh264/issues/2505
-    # Master now has support, but not the 3.1.x branch
-    resource "openh264-1.5.0" do
-      url "https://github.com/cisco/openh264/archive/v1.5.0.tar.gz"
-      sha256 "98077bd5d113c183ce02b678733b0cada2cf36750370579534c4d70f0b6c27b5"
-    end
-  end
+  url "https://ffmpeg.org/releases/ffmpeg-3.2.tar.bz2"
+  sha256 "76d6cd9f5e64463a5b9940736da8a515c990bcbbe506a722e2040916cb366d74"
+  head "https://github.com/FFmpeg/FFmpeg.git"
 
   bottle do
     sha256 "201c11c27d3c4d2c65593d1f95ecbc93194da5c3d0d88193313df1d81db6d69d" => :sierra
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
> Restore the --with-ffplay option and deprecate --with-sdl/--with-sdl2 (recently added). Whatever backend ffplay uses is just an implementation detail, users should not need to know it (and most don't), and using the dependency name in the option name not only causes user confusion (especially after years of --with-ffplay which is the only name that makes sense), it's also a headache when upstream switches backend, as is happening with this release.

This creates issues with our dependency system so until we have a `variant do` or similar in Homebrew/brew we should avoid regressing back on these.

> Drop the AAC encoding caveats specific to the FFmpeg 3.0 release. It's been out for the better part of a year and shouldn't be news anymore. Users concerned with this caveat should have migrated already.

👍

zmwangx:
> This creates issues with our dependency system so until we have a variant do or similar in Homebrew/brew we should avoid regressing back on these.

The mistake here was deprecating `--with-ffplay` in the first place, which doesn't make sense (as an FFmpeg user I can tell you this for sure) and led to a funny stable/head split that `brew info` completely failed to account for. I myself lost `--with-ffplay` somewhere along the upgrade path which shouldn't have happened.

If you look at all the existing `deprecated_option`s, they are mostly adding a `with-` prefix or changing `enable-` to `with-` / `disable-` to `without-`. None of them changes an option from something that users can understand (`--with-ffplay`) to something that they can't (`--with-sdl2`).

Restoring `--with-ffplay` simply isn't a regression, and shouldn't cause problems for users if `deprecated_option` works as expected. If it doesn't, well, they were probably already screwed the last time (as I did), so recompiling `--with-ffplay` is the way to restore sanity.

zmwangx:
> None of them changes an option from something that users can understand (--with-ffplay) to something that they can't (--with-sdl2).

Oh, not only that. `--with-ffplay` was deprecated in favor of `--with-sdl`, which is outdated three weeks later.

MikeMcQuaid:
> Restoring --with-ffplay simply isn't a regression

`depends_on "sdl2" if build.with?("ffplay")` is buggy and `depends_on "sdl2" => :optional` is not so it will cause a regression.

> I myself lost --with-ffplay somewhere along the upgrade path which shouldn't have happened.

I agree this should not happen. If you can figure out how to reproduce an issue or PR would be welcome.

zmwangx:
> depends_on "sdl2" if build.with?("ffplay") is buggy

Then fix the bug?

zmwangx:
```
depends_on "sdl2" if build.with?("ffplay")
```

That structure is used all over the place and deprecating options doesn't work in most cases.

MikeMcQuaid:
> That structure is used all over the place

Yes but we want to avoid it when it's possible (as it isn't always possible).

Please put it back to how it was.

retokromer:
@zmwangx At some point it was necessary to use both `--with-sdl` and `--with-sdl2` in order to cover all situations, which is absurde.
👍 👍 👍  in favour for restoring `--with-ffplay`, which is intuitive for the user.

zmwangx:
Deprecated `--with-ffplay` and made `--with-sdl2` the primary option. (Sorry, I tried.)

nijikon:
@BrewTestBot test this please

zmwangx:
Can someone get this merged? Thanks.

dunn:
Done, thanks!

PR #6345, opened 2016-10-28
osmosis 0.45

fabrik42:

---

The audit gives a warning that no tests are defined. What's the best way to add a test for this rather simple formula?

Best,

Christian

### DIFF
--- a/Formula/osmosis.rb
+++ b/Formula/osmosis.rb
@@ -16,28 +16,27 @@ class Osmosis < Formula
 
   test do
     path = testpath/"test.osm"
-    path.write <<-EOS
-<?xml version="1.0" encoding="UTF-8"?>
-<osm version="0.6" generator="CGImap 0.5.8 (30532 thorn-05.openstreetmap.org)" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
-<bounds minlat="49.9363700" minlon="8.9159400" maxlat="49.9371300" maxlon="8.9173800"/>
-<node id="4140986569" visible="true" version="1" changeset="38789367" timestamp="2016-04-22T15:17:02Z" user="KartoGrapHiti" uid="57645" lat="49.9369693" lon="8.9163279">
-  <tag k="bench" v="yes"/>
-  <tag k="bin" v="yes"/>
-  <tag k="bus" v="yes"/>
-  <tag k="highway" v="bus_stop"/>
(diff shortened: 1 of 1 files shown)


tdsmith:
Feed it some sample data and make sure it doesn't choke, ideally! You can download a small dataset with a `resource` if there isn't one in the tarball we're downloading.

fabrik42:
Hmm sorry, but I am still unsure how to do this. I still have a couple of questions left:

Where is the dataset supposed to come from? Should I host it myself? And how can I prove that this works? You basically want to do a very simple operation to prove that it is callable?

Thanks,
Christian

dunn:
> Where is the dataset supposed to come from? 

Anywhere reliable, where it can be expected to not disappear.

> You basically want to do a very simple operation to prove that it is callable?

Yep!

fabrik42:
Ok, I just added the test. Instead of downloading something, I just provided a very simple fixture in the spec. I hope this is ok.

Thanks for the help!

Formula/osmosis.rb
```diff
@@ -13,6 +13,32 @@ def install
     bin.install "bin/osmosis"
     libexec.install %w[lib config script]
   end
+
+  test do
+    path = testpath/"test.osm"
+    path.write <<-EOS
```
MikeMcQuaid: Use EOS.undent` and intent the below lines to be two-spaces beyond the whitespace on this line.

Formula/osmosis.rb
```diff
@@ -13,6 +13,32 @@ def install
...
+  <tag k="shelter" v="yes"/>
+  <tag k="tactile_paving" v="no"/>
+  <tag k="wheelchair" v="no"/>
+  <tag k="wheelchair:description" v="Kein Kasseler Bord"/>
+</node>
+</osm>
+    EOS
+
+    `#{bin}/osmosis --read-xml file="#{path}" --write-null`
+    assert_equal 0, $?.exitstatus
```
MikeMcQuaid: Use `system` instead which will check the exit status.

MikeMcQuaid (changes requested):

Formula/osmosis.rb
fabrik42: done!

Formula/osmosis.rb
fabrik42: good point!

MikeMcQuaid (approved):

MikeMcQuaid:

fabrik42:
Thanks, you're welcome! :)

PR #6346, opened 2016-10-28
git-annex 6.20161031

ilovezfs:
monad-logger workaround for GHC 8.0.1.20161022


ilovezfs:
hurray! upstream monad-logger tagged a new release so the workaround can go away!

PR #6358, opened 2016-10-28
ghc: improve Sierra compatibility

ilovezfs:
On Sierra, use new upstream dynamic-library-dirs field to avoid

```
  ghc: panic! (the 'impossible' happened)
  malformed mach-o: load commands size (x) > 32768
```

where x is some number larger than 32768.

Note that this fix requires a corresponding update to cabal-install and that `cabal new-build` hasn't been updated to use the new field yet, so will continue to fail in the same scenarios as it already does.

Also, bump the vendored `gmp` to 6.1.1 for all macOS versions, and stop using `nm-classic` since `lvm-nm -P` is POSIX compliant now.

### DIFF
--- a/Formula/ghc.rb
+++ b/Formula/ghc.rb
@@ -1,8 +1,38 @@
+require "language/haskell"
+
 class Ghc < Formula
+  include Language::Haskell::Cabal
+
   desc "Glorious Glasgow Haskell Compilation System"
   homepage "https://haskell.org/ghc/"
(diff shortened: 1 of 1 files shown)


zmwangx:
Older GHC used to work miracles (the impossible happened). Too bad they're making it less miraculous.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

PR #6362, opened 2016-10-29
vsftpd: add support for brew services

jabenninghoff:
Add a plist to support “brew services”, add simple test (version check). Tested on OS X 10.11.


Formula/vsftpd.rb
```diff
@@ -49,6 +49,31 @@ def caveats
 
     s
   end
+
+  plist_options :startup => true, :manual => "sudo #{HOMEBREW_PREFIX}/sbin/vsftpd #{HOMEBREW_PREFIX}/etc/vsftpd.conf"
```
jabenninghoff: Thanks. I will update the pull request. In fact, the manual command can be just `sudo vsftpd`.

jabenninghoff:
Hi, I updated the pull request with a corrected manual command and added caveats for launching vsftpd as root (needed in most cases) and ownership of vsftpd.conf.

Formula/vsftpd.rb
```diff
@@ -44,11 +44,45 @@ def caveats
       s += <<-EOS.undent
         vsftpd was compiled with SSL support. To use it you must generate a SSL
         certificate and set 'enable_ssl=YES' in your config file.
+
       EOS
     end
 
+    s += <<-EOS.undent
+      vsftpd requires root privileges in most cases so you will need to run
```
MikeMcQuaid: Why does it need root privileges in most cases? I'd rather we figured out a way to configure this and the `plist` such that it can be used without root privileges and those who want to use it as `root` can do so themselves.

Formula/vsftpd.rb
```diff
@@ -44,11 +44,45 @@ def caveats
...
+      </array>
+      <key>RunAtLoad</key>
+      <true/>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    assert_match version.to_s, shell_output("#{sbin}/vsftpd -v 0>&1")
```
MikeMcQuaid: Whats `0>&1` doing?

MikeMcQuaid (changes requested):

Formula/vsftpd.rb
jabenninghoff: Technically, it is possible to run vsfptd as a user other than root, however, doing so is less secure. From `vsftpd.conf(5)`:
```
run_as_launching_user
       Set to YES if you want vsftpd to run as the user which launched  vsftpd.  This  is  useful  where  root
       access  is  not  available. MASSIVE WARNING! Do NOT enable this option unless you totally know what you
       are doing, as naive use of this option can create massive security problems. Specifically, vsftpd  does
       not / cannot use chroot technology to restrict file access when this option is set (even if launched by
       root). A poor substitute could be to use a deny_file setting such as {/*,*..*}, but the reliability  of
       this cannot compare to chroot, and should not be relied on.  If using this option, many restrictions on
       other options apply. For example, options requiring privilege such as non-anonymous logins, upload own-
       ership  changing,  connecting  from  port  20 and listen ports less than 1024 are not expected to work.
       Other options may be impacted.
```
I will update the caveat to make it clear that running as a user other than root is not recommended.

Formula/vsftpd.rb
jabenninghoff: By default, `vsftpd -v` outputs version info to stdin. To test the version properly, stdin must be redirected to stdout (0>&1).

Formula/vsftpd.rb
```diff
@@ -44,11 +44,45 @@ def caveats
       s += <<-EOS.undent
         vsftpd was compiled with SSL support. To use it you must generate a SSL
         certificate and set 'enable_ssl=YES' in your config file.
+
       EOS
     end
 
+    s += <<-EOS.undent
+      To use chroot, vsftpd requires root privileges, so you will need to run
+      `sudo vsftpd`.
```
MikeMcQuaid: Thanks for clarification.

Formula/vsftpd.rb
```diff
@@ -44,11 +44,45 @@ def caveats
...
+      You should be certain that you trust any software you grant root privileges.
```
MikeMcQuaid: Can probably remove this but 👍

Formula/vsftpd.rb
```diff
@@ -44,11 +44,45 @@ def caveats
...
+
+      The vsftpd.conf file must be owned by root in order to work correctly:
```
MikeMcQuaid: Can you elaborate on "work correctly"?  Would be good if there was some way of doing this on first startup.

Formula/vsftpd.rb
MikeMcQuaid: Bizarre, thanks!

MikeMcQuaid (changes requested):

Formula/vsftpd.rb
jabenninghoff: Yes - vsftpd will refuse to start with an error, `500 OOPS: config file not owned by correct user, or not a file`. I'll state that directly in the caveat. Is there a supported/recommended way of writing a script to handle setup? When I looked through existing formula, the only similar examples I found used a caveat to advise the user to `chown`.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #6374, opened 2016-10-29
qt5 properly enable OpenSSL

ghost:

---

I request to pull this because I see the other PRs do not figure out how to link Qt5 against brewed OpenSSL properly.

Background:
Apple stopped supporting OpenSSL.
(That's why we all have the unwanted/outdated OpenSSL 0.9.8 in /usr/lib) 
Secure Transport is Apple's version of SSL/TLS.
https://developer.apple.com/reference/security/1654508-secure_transport

Qt5 supports both OpenSSL and Secure Transport as its SSL back end.
However Qt5 can not link with both at the same time.
If you want to enable OpenSSL, you have to disable Secure Transport at compile time.
Vice versa.
i.e. -no-securetransport is mandatory if you need to enable OpenSSL.

--with-openssl
Qt classes like QSslCertificate & QSslKey do not function if Qt's SSL back end is Secure Transport.
Qt is cross-platform software, the users expect Qt will work on macOS as it does on Linux and MS-Windows.
We need a switch to enable OpenSSL without --with-mysql.

-openssl-linked is not acceptable because static linkage is not default.
If you differ from default, there must be an inevitable reason.
Unfounded fear that Qt may pick up wrong dynamic libraries at runtime is not a good reason.

-pch is useless in a Homebrew Formula because created .pch files are removed after `brew install -s qt5` is completed along with the sources and intermediate files (*.o etc).

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -73,6 +73,7 @@ class Qt5 < Formula
   option "with-oci", "Build with Oracle OCI plugin"
   option "with-qtwebkit", "Build with QtWebkit module"
   option "without-webengine", "Build without QtWebEngine module"
+  option "with-openssl", "Build with OpenSSL"
 
   deprecated_option "qtdbus" => "with-dbus"
   deprecated_option "with-d-bus" => "with-dbus"
@@ -86,6 +87,7 @@ class Qt5 < Formula
   depends_on :mysql => :optional
   depends_on :postgresql => :optional
   depends_on :xcode => :build
+  depends_on "openssl" if build.with?("openssl") || build.with?("postgresql") || build.with?("mysql")
 
   depends_on OracleHomeVarRequirement if build.with? "oci"
 
@@ -125,8 +127,18 @@ class Qt5 < Formula
 
     args << "-nomake" << "examples" if build.without? "examples"
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
     args << "-plugin-sql-psql" if build.with? "postgresql"
+    if build.with? "mysql"
+      args << "-plugin-sql-mysql"
+      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"
+    end
+
+    if build.with? "openssl"
+      args << "-openssl-linked" << "-no-securetransport"
+      openssl_opt = Formula["openssl"].opt_prefix
+      args << "-I#{openssl_opt}/include"
+      ENV["OPENSSL_LIBS"] = "-L#{openssl_opt}/lib -lssl -lcrypto"
+    end
 
     if build.with? "dbus"
       dbus_opt = Formula["dbus"].opt_prefix


Formula/qt5.rb
```diff
@@ -128,6 +130,15 @@ def install
     args << "-plugin-sql-mysql" if build.with? "mysql"
     args << "-plugin-sql-psql" if build.with? "postgresql"
 
+    # Both mysql and postgresql require openssl libs to link
+    if build.with? "openssl" or build.with? "mysql" or build.with? "postgresql"
+      openssl_opt = Formula["openssl"].opt_prefix
```
ilovezfs: You can make this
```
openssl = Formula["openssl"]
```
And then use `#{openssl.opt_lib}` and `#{openssl.opt_include}` below.

Formula/qt5.rb
```diff
@@ -86,6 +87,7 @@ class Qt5 < Formula
   depends_on :mysql => :optional
   depends_on :postgresql => :optional
   depends_on :xcode => :build
+  depends_on "openssl" => :optional
```
ilovezfs: Do you have opinions on making this default (i.e. `=> :recommended`)?

Formula/qt5.rb
```diff
@@ -73,6 +73,7 @@ class Qt5 < Formula
   option "with-oci", "Build with Oracle OCI plugin"
   option "with-qtwebkit", "Build with QtWebkit module"
   option "without-webengine", "Build without QtWebEngine module"
+  option "with-openssl", "Build with openssl"
```
ilovezfs: `=> :optional` and `=> :recommended` generate option descriptions automatically. Since this description doesn't add any additional information, you can just delete the line entirely.

Formula/qt5.rb
```diff
@@ -128,6 +130,15 @@ def install
     args << "-plugin-sql-mysql" if build.with? "mysql"
     args << "-plugin-sql-psql" if build.with? "postgresql"
 
+    # Both mysql and postgresql require openssl libs to link
+    if build.with? "openssl" or build.with? "mysql" or build.with? "postgresql"
+      openssl_opt = Formula["openssl"].opt_prefix
+      args << "-L#{openssl_opt}/lib"
+      args << "-I#{openssl_opt}/include"
+      args << "-no-securetransport"
+      ENV['OPENSSL_LIBS'] = "-L#{openssl_opt}/lib -lssl -lcrypto"
```
ilovezfs: Double quotes on OPENSSL_LIBS not single quotes

ilovezfs:
```
-openssl-linked is not acceptable because static linkage is not default.
If you differ from default, there must be an inevitable reason.
Unfounded fear that Qt may pick up wrong dynamic libraries at runtime is not a good reason.
```
Can you explain more what you mean by this? If a homebrew formula is going to use openssl, we basically never want to use the system openssl, so it's not "unfounded fear" as much as much as it's preemptively trying to enforce our will.

yezezey:
@ilovezfs 

> Can you explain more what you mean by this? If a homebrew formula is going to use openssl, we basically never want to use the system openssl, so it's not "unfounded fear" as much as it's preemptively trying to enforce our will.

My aim is ensure Qt5 is linked against brewed OpenSSL.
For that purpose static linkage is not necessary. (Other PR #6309 thinks it's necessary)

ilovezfs:
>This PR conditionally links qt5 against the brewed openssl when mysql or postgresql support is enabled. It does the -openssl-linked option to ensure that the brewed openssl is linked at compile time and is always used. (The default is for Qt to detect installed openssl libraries at run time, which may pick up the system openssl instead.)

"linked at compile time" does not mean "compiled statically" ... it means the path is decided at compile time. In brew's case the path would be to the libraries in `/usr/local/opt/openssl/lib`, which is updated whenever the openssl formula gets revision bumped or upgraded.

Is that not what `-openssl-linked` does?

yezezey:
No, -openssl-linked means static link instead of dynamic link.
If you use -openssl-linked, OpenSSL's version is fixed at compile time.

ilovezfs:
Do you have a link to the documentation for that where it says it's static?

yezezey:
Qt's official documentation does not state clearly.
I know -openssl-linked means static link because it works as expected when I build qBittorrent.
https://github.com/qbittorrent/qBittorrent/wiki/Compiling-with-MSVC-2015-%28static-linkage%29#Compiling_Qt5

Do you need more convincing evidence?

yezezey:
It seems I misunderstood the meaning of -openssl-linked.
Do you have clear definition of this option?

ilovezfs:
I think basically it means that if you were to `brew uninstall openssl`, then

- if qt 5 were built with `-openssl-linked` it would break due to the broken linkage

whereas

- if qt 5 were built with just `-openssl`, it would attempt to use another openssl (usually the system ssl) at run time if it can get its hands on one.

yezezey:
I will test that tomorrow.
Good night.

Formula/qt5.rb
DomT4: > Do you have opinions on making this default (i.e. => :recommended)?

😷

ilovezfs: @DomT4 it's not Halloween yet. You can take off the mask. What's up?

DomT4: Was just griping about OpenSSL potentially being the default for Qt5, heh.

Secure Transport is friendlier, and I think you'd cause yourself issues overriding the upstream default to use OpenSSL given people have been trying to get Homebrew to do that to `curl` for a while _(where the upstream default is actually OpenSSL)_ and we kept shooting it down because SecureTransport is shipped by the system and decently secure/maintained/etc.

ilovezfs: And I assume getting mysql or postgresql to use SecureTransport is a fool's errand?

DomT4: As far as I'm aware, yeah, there's no support for SecureTransport in either of those. Despite OpenSSL being deprecated on OS X/macOS for a fairly long time now a lot of upstreams don't seem tremendously interested in supporting yet _another_ crypto backend.

IMO, the `mysql` and `postgresql` options for `qt5` have always been more trouble than they are worth, but obviously people are using them and are determined to make them work come hell or high water. I know `postgresql` had a discussion upstream about supporting more crypto backends but I don't think SecureTransport support ever came out of that.

It's completely gross but personally I'd be tempted to make it:
```ruby
depends_on "openssl" if build.with?("postgresql") || build.with?("mysql")
```
That way you get the support without letting people deviate from the upstream SecureTransport default more than necessary.

The options here get kind of complicated by `mysql` and `postgresql` both being special dependencies and so not necessarily the Homebrew ones. The upstream pre-compiled `mysql` for example ships its own statically-linked `yassl` for the GPL version and `openssl` for the commercial version IIRC, so you could end up with Qt5 using an incompatible SSL compared to the dependency, but more or less that's a "user configuration" issue.

DomT4: `yassl` being `wolfssl` these days, of course. Forgot they'd renamed themselves.

ilovezfs:
```
depends_on "openssl" if build.with?("postgresql") || build.with?("mysql")
```
is probably right since if openssl is a non-default option as it is in the PR right now, there will be a lot of people building qt5 from source for a rather – let's say "confused" – reason.

DomT4: 👍. It's a bundle of fun. In an ideal world Qt5 would be stripped right down given the amount of hassle even the vanilla configuration causes in keeping alive and updated, let alone the options, but I've escaped having to help rescue the Qt5 formula so 😄.

Anyway, I only came to complain about Bundle doing strange things and I've ended up here, oops. Enjoy your afternoon 🙇

ilovezfs: Thanks for poking your head in. Enjoy your evening.

yezezey:
I made two versons of Qt5. /usr/local/qtnolink /usr/local/qtlinked
I used http://doc.qt.io/qt-5/qtnetwork-http-example.html as a sample program, and built two versions of http.app.
After I did `brew uninstall openssl`, both http.app failed to launch.

screen shot 2016-10-30 at 2 15 50

What I actually did
```
cd /tmp
tar xf ~/Library/Caches/Homebrew/qt5-5.7.0.tar.xz
mv qt-everywhere-opensource-src-5.7.0/ qtnolink
tar xf ~/Library/Caches/Homebrew/qt5-5.7.0.tar.xz
mv qt-everywhere-opensource-src-5.7.0/ qtlinked
cd /tmp/qtnolink
OPENSSL_LIBS='-L/usr/local/opt/openssl/lib -lssl -lcrypto' ./configure -prefix /usr/local/qtnolink -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib -no-rpath -opensource -confirm-license -release -no-securetransport -make libs -make tools -nomake examples -nomake tests -no-sql-sqlite -skip qt3d -skip qtactiveqt -skip qtandroidextras -skip qtcanvas3d -skip qtdeclarative -skip qtdoc -skip qtgraphicaleffects -skip qtimageformats -skip qtlocation -skip qtmacextras -skip qtmultimedia -skip qtquickcontrols -skip qtscript -skip qtsensors -skip qtserialport -skip qtsvg -skip qtwayland -skip qtwebchannel -skip qtwebsockets -skip qtwinextras -skip qtx11extras -skip qtxmlpatterns -skip qtwebview -skip qtwebengine -skip qtconnectivity -v
make -j4
make install
cd /tmp/qtlinked
OPENSSL_LIBS='-L/usr/local/opt/openssl/lib -lssl -lcrypto' ./configure -prefix /usr/local/qtlinked -openssl-linked -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib -no-rpath -opensource -confirm-license -release -no-securetransport -make libs -make tools -nomake examples -nomake tests -no-sql-sqlite -skip qt3d -skip qtactiveqt -skip qtandroidextras -skip qtcanvas3d -skip qtdeclarative -skip qtdoc -skip qtgraphicaleffects -skip qtimageformats -skip qtlocation -skip qtmacextras -skip qtmultimedia -skip qtquickcontrols -skip qtscript -skip qtsensors -skip qtserialport -skip qtsvg -skip qtwayland -skip qtwebchannel -skip qtwebsockets -skip qtwinextras -skip qtx11extras -skip qtxmlpatterns -skip qtwebview -skip qtwebengine -skip qtconnectivity -v
make -j4
make install
cd /tmp/qtnolink/qtbase/examples/network/http
/usr/local/qtnolink/bin/qmake
make
cd /tmp/qtlinked/qtbase/examples/network/http
/usr/local/qtlinked/bin/qmake
make
```

yezezey:
This is what I expected.
QtNetwork depends on brewed OpenSSL, but the OS can not find it.
I was right about this.
On macOS a load module does not pick up wrong dynamic libraries at runtime.

Formula/qt5.rb
yezezey: > OpenSSL being deprecated on OS X/macOS for a fairly long time

Apple deprecated OpenSSL, but OpenSSL itself is actively maintained.

yezezey:
How can I apply this patch? https://github.com/yezezey/formula-patches/blob/whisky/qt5/macdeployqt-bug-fix.patch

Made by @rockihack in https://github.com/Homebrew/homebrew-core/issues/3219
Without this patch, brewed Qt5 is completely unusable for me.

MikeMcQuaid:
> How can I apply this patch? https://github.com/yezezey/formula-patches/blob/whisky/qt5/macdeployqt-bug-fix.patch

Can you please submit this patch to the upstream Qt developers? Thanks!

yezezey:
Nope.
This problem does not occur when I do `./configure && make && make install`.
I see this is Homebrew's problem.
How do I explain to the Qt developers this patch is necessary?
Why do I have to do it for Homebrew's sake?
One of the Homebrew people should do it.

MikeMcQuaid:
> How do I explain to the Qt developers this patch is necessary?

Because it's required for Homebrew which is a widely used package manager and doesn't carry patches that aren't getting accepted upstream.

> Why do I have to do it for Homebrew's sake? One of the Homebrew people should do it.

Because you said it's unusable without that patch and because "the Homebrew people" are everyone that uses Homebrew as no-one is paid to work on it.

yezezey:
> Because  you said it's unusable without that patch and because "the Homebrew people" are everyone that uses Homebrew as no-one is paid to work on it.

I meant Homebrew maintainers, not Homebrew users.
I am not a Homebrew maintainer.

Anyway whoever submit the patch, the fix will be applied in a next release.
Right now we have to patch the current official tar ball, don't we?

yezezey:
Ok, now I have submitted the patch. https://bugreports.qt.io/browse/QTBUG-56814
I still don't see why I had to do that because any one of Homebrew developers could do far better than I did to convince them it's necessary.

Anyway let's apply the patch, shall we?

I have already reported another severe bug months ago, but no response from them so far.  https://bugreports.qt.io/browse/QTBUG-56344
Should I wait for years to apply this patch? https://github.com/Homebrew/homebrew-core/pull/5619#discussion_r82139977

ilovezfs:
I think https://bugreports.qt.io/browse/QTBUG-56124 may be the same as your https://bugreports.qt.io/browse/QTBUG-56344, and it was fixed in http://code.qt.io/cgit/qt/qtbase.git/commit/src/corelib/io/qsettings_mac.cpp?id=4cb614c7abdaa2c5e2d0a75201d51aae01e6f8c4

You may want to try that patch.

As for https://github.com/Homebrew/homebrew-core/pull/5619#discussion_r82139977, those issues have already been fixed upstream, and I included those patches in the 5.7.0 release.

Regarding the macdeployqt thing, as far as I can tell that was already considered and rejected by us in https://github.com/Homebrew/homebrew-core/issues/3219.

yezezey:
@ilovezfs 
Thanks for the info.

> Regarding the macdeployqt thing, as far as I can tell that was already considered and rejected by us in #3219.

Really?
Then why @MikeMcQuaid said to me to submit the patch to the upstream?

yezezey:
And what about the practically undocumented switch -openssl-linked?
I misunderstood it's static link, and you guys misunderstood something else because it's undocumented.
Both proven wrong by my test.
However I have no objection to this option now.
I will add it if anyone insists.

yezezey:
macdeployqt is a final process of build.
Some applications won't launch without processed by macdeployqt.
These apps highly likely crash at launch time if they are linked against brewed Qt5.
The patch works like a charm.

If Homebrew Qt5 deliberately don't support these apps,  there must be a suitable reason.
But I don't see the reason why the patch was already rejected in #3219.

Formula/qt5.rb
DomT4: Not the 0.9.8 branch, which Apple uses as the "base" for its OpenSSL. I say "base" because Apple has kept selectively plucking security fixes off of the upstream 1.0.1 and 1.0.2 branches so by now it's something of a mutant. 0.9.8 is dead upstream.

steve3d:
@yezezey , I think you need to learn some more about programming, you've got it totally wrong about pch and `-openssl-linked` option, let me explain something for you,
**First, PCH**, pch is a compile time thing, it's very specific to a project and its compile options, once you changed the compile option for a project(for example from debug to release), it's invalid. so why qt should install the build-time generated pch to user's dest installation directory, of course it should be ignored. It's NEVER meant to be used as generic compile time optimization. For example, you have this in your project:

```
#include <stdio.h>
#include <stdlib.h>
```

and if you only change the include order, this pch file would need to regenerate, and previous pch is no longer valid, so install this huge wasted mid-compiling thing is a total waste of a users disk space. test results will prove more, right, then here comes the result:

```
# ./configure -nomake examples -release -no-pch
# time make -j6 on qtbase module only
real	5m28.433s
user	28m31.658s
sys	1m54.566s

# ./configure -nomake examples -release -pch
# time make -j6 on qtbase module only
real	3m59.195s
user	20m32.265s
sys	1m33.207s
```
build time reduced from 28mins to 20mins, it's about 28% build time reduce. and during these test, I found I'm wrong about the pch option, in fact the pch option will be turned on on macOS, seems qt's crew improved their configure scripts to detect if current compiler supports pch, if yes, then turn it on.


**Second** about the `-openssl-linked` option, how can you statically link to openssl where is NO static version of openssl library exists? if you static link to openssl, there must be a `libssl.a` and `libcrypto.a` link archive exists, but can you find any of these exists? macOS never give you a static version of openssl library nor homebrew, they are all dynamic library, unless you call that link to a dynamic library is **static linking**, that's a joke right?

So what does this option really do? This option simply make QtNetwork module link to openssl dynamic library, if you don't give this option, then QtNetwork will try to use `dlopen` to open a openssl library whoever is first found in LD_LIBRARY_PATH, under this, if you run `otool -l` on QtNetwork, there will be no direct link to openssl library. @ilovezfs, your first thought about this option is correct.

As for your problems with system bounded openssl, there is a similar report long long time ago, again please be nice to your google friend.

**Third**, about the mysql/postgresql have these `-lssl -lcrypto` thing, programmers sometime made a static version of mysql/postgresql library to use, but they still want to use `mysql_config` to get those flags, and there is no flag to make mysql_config to return different flags for static linking and dynamic linking, needless to say, this two babies are mainly targeted to *nix systems, not primary on macOS, under any *nix system, the openssl or any other ssl libaray are a standard feature, so why bother different this, they can of course use the same link flags. and assume you decided to build qt as static version, there IS a static version of mysql/postgresql library exists, and they of course need to link to openssl, but for dynamic version, `otool -L libmysqlclient.dylib` shows that it also depends on `libz.1.dylib`, if you only need mysql functions, do you really need to link against openssl and zlib on a dynamic link? OF COURSE **NOT**, they are only needed when you do a static link against a object archive. so as long as we only use dynamic version of qt, it's absolutely safe to strip out the `-lssl -lcrypto` options from QtSQL's module link flags.

Please stop making this jokes and confusing more people.

steve3d:
well, seems I've made a little mistake, homebrew did provided a static version of openssl.

MikeMcQuaid:
> I think you need to learn some more about programming

@steve3d Please show more respect to @yezezey and read our Code of Conduct: https://github.com/Homebrew/brew/blob/master/CODEOFCONDUCT.md#code-of-conduct.

steve3d:
If I'm not respectful. Then I apologize for my saying, but I'm also pointing out what is wrong here, am I right?

yezezey:
@ilovezfs 
If the Homebrew developers realy don't want to fix this macdepoyqt problem, let me know.
Because in that case, brewed Qt5 is completely useless for my community,
and I don't want to spend my time to implove this Formula which we can't use.

MikeMcQuaid:
> If the Homebrew developers realy don't want to fix this macdepoyqt problem, let me know.

We do want it fixed but we need Qt patches to be submitted upstream to https://codereview.qt-project.org, accepted and ideally merged and we don't want to carry long-running patches upstream has rejected.

yezezey:
@MikeMcQuaid 
Is there someone who can help to speed up the process?
Someone who can explain the technical details about Homebrew and why macdeployqt does not work with it?

MikeMcQuaid:
> Someone who can explain the technical details about Homebrew and why macdeployqt does not work with it?

The beginning of https://github.com/Homebrew/homebrew-core/issues/3219#issuecomment-234719795 seems to sum it up pretty well.

yezezey:
Would you help me more? Because I can't compose a good report for the upstream from that comment. English is not my first language.

RandomDSdevel:
@MikeMcQuaid, @yezezey:  Shouldn't you guys stop polluting this PR with unrelated content and move discussion of the `macdeployqt` stuff over to either a reopened #3219 or a new issue?

yezezey:
@RandomDSdevel 
Good point.

@steve3d
Thanks for pointing out. I was wrong.
Since my aim is properly linking Qt5 against brewed OpenSSL, I was not thinking about Secure Transport.
If I choose to use Secure Transport,  -openssl-linked is totally necessary.
My bad.

yezezey:
If a user choose to link Qt5 against brewed OpenSSL, there is no need to do special treatment to build MySQL plugin.

The question is if a user wants to build MySQL plugin but did not choose brewed OpenSSL, what should we do?
 a) do as @steve3d suggests
 b) automatically enable brewed OpenSSL

I still think b is better solution.

yezezey:
Can I apply this QSetting bug fix, which @ilovezfs kindly show me? http://code.qt.io/cgit/qt/qtbase.git/commit/src/corelib/io/qsettings_mac.cpp?id=4cb614c7abdaa2c5e2d0a75201d51aae01e6f8c4

This bug really hurts end users.

steve3d:
I think the mysql plugin and openssl is a hard choice here, for example, I want to use qt with secure transport and with mysql/postgresql, even if I don't want openssl, but mysql/postgresql still depends on openssl, this means if one need to use homebrew's mysql/postgresql, they must have to use homebrews openssl with it.

so why not add another option, like php70 have a option of `--with-homebrew-libressl`, which will make qt use homebrew's openssl instead of secure transport, and leave mysql/postgresql plugin default off.

By the way, seems there is no one complaining about homebrew's qt don't have a mysql/postgresql plugin.

yezezey:
> I think the mysql plugin and openssl is a hard choice here, for example, I want to use qt with secure transport and with mysql/postgresql, even if I don't want openssl, but mysql/postgresql still depends on openssl, this means if one need to use homebrew's mysql/postgresql, they must have to use homebrews openssl with it.

True. And I don't see any problem in it.

> so why not add another option, like php70 have a option of --with-homebrew-libressl, which will make qt use homebrew's openssl instead of secure transport, and leave mysql/postgresql plugin default off.

That's what `brew install qt5 --with-openssl` do in this PR.

> By the way, seems there is no one complaining about homebrew's qt don't have a mysql/postgresql plugin.

True. And I don't see anyone who complains those plugins are linked against openssl as long as that is brewed one.

yezezey:
rebased

DomT4:
Can literally anyone explain to me what is wrong with the simple:
```ruby
depends_on "openssl" if build.with?("postgresql") || build.with?("mysql")
```
❓ 

As it stands this PR enforces OpenSSL on everyone automatically, whether they like it or not, which reverts the upstream default for no credible reason for anyone beside folks using the `postgresql` or `mysql` options.

yezezey:
@DomT4 
Yeah, I agree that.
It appears there is no need for that.

yezezey:
Now I confirm slight improvement in compile speed with -pch.
Should we include -pch as default?

```
 $ time brew install -v /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/qt5.rb --without-webengine --with-openssl
 real	29m52.764s
 user	76m50.673s
 sys	7m58.345s
 
 $ time brew install -v /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/qt5.rb --without-webengine --with-openssl --with-pch
 real	29m13.988s
 user	76m40.545s
 sys	7m51.648s
```

steve3d:
@yezezey, you won't see any difference here, because now qt's configure script will automatically enable pch if compiler support it. so your tests both have pch enabled.

if you want to test without pch, you have to pass `-no-pch` option to configure.

back to qt 3.x, pch won't be enabled automatically except under windows.

DomT4:
> Look the code I rebased yesterday.

_I did_. Right now, it mandates `openssl` for everyone, regardless of option usage, regardless of the upstream default. There's also an option and `if build.with?` statement that are currently redundant because you've made it mandatory.

yezezey:
@steve3d 
Thanks for the info.

yezezey:
@DomT4 
As @steve3d pointed out, Qt5 will load OpenSSL libraries even if you choose Secure Transport.
Without `OPENSSL_LIBS='-L/usr/local/opt/openssl/lib -lssl -lcrypto'  -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib`, Qt5 will load the system OpenSSL instead of Homebrew OpenSSL.

yezezey:
I forgot to add -openssl-linked

DomT4:
> Qt5 will load OpenSSL libraries even if you choose Secure Transport.

I haven't looked into this much, but if that's the case even on the vanilla, no-options build that smells very strongly like an upstream bug that should quickly be reported to them. But it is my understanding that for the _vanilla_ build there's no OpenSSL interaction.

yezezey:
> But it is my understanding that for the vanilla build there's no OpenSSL interaction.

Can you prove that?

I have not completely understand the Qt source, but I can see Qt5 actually tries to load openssl in /usr/lib

http://code.qt.io/cgit/qt/qtbase.git/tree/src/network/ssl/qsslsocket_openssl_symbols.cpp
From line 672

DomT4:
I can't comment on my certainty without commenting on the complexity that is the Qt codebase, which'll get me in trouble now I've given up my maintainer card. But it would seem completely bizarre to import OpenSSL whilst building against Secure Transport by default; I'd go so far as to call that actively misleading if it was the case. But, fairly easy to ask the Qt folk... @jakepetroules.

jakepetroules:
Please use -securetransport. OpenSSL is not needed on macOS or other Apple platforms in that configuration, and the former is both default and recommended in current versions. SecureTransport and OpenSSL are mutually exclusive.

yezezey:
So the problem is Homebrew MySQL requires OpenSSL as default and we can not build Qt5 with the vanilla build Secure Transport?
What the right thing to do?

Since my primary goal was link Qt5 against Homebrew OpenSSL, I do not want to delve into the complication with -securetransport with -plugin-sql-mysql.

steve3d:
@yezezey, I think you wrong again, say if you don't give any openssl related option to configure script, then the configure results are:

```
  Networking: 
    CoreWlan ............. yes
    getaddrinfo .......... yes
    getifaddrs ........... yes
    IPv6 ifname .......... yes
    libproxy.............. no
    OpenSSL .............. no
    SecureTransport ...... yes
```

this means QtNetwork will not link to any openssl libraries, also it won't try to use `dlopen` to load openssl libraries, so under default options, Qt will never need openssl. So without any openssl options with configure script, the source file qsslsocket_openssl_symbols.cpp won't get compiled, it's not listed in source files of QtNetwork's makefile.

the only thing needs openssl is that mysql/postgresql sql driver plugin, because they need openssl, so that makes qtsql indirectly need openssl.

yezezey:
@steve3d
Thanks for the info.

@jakepetroules
>OpenSSL is not needed on macOS or other Apple platforms in that configuration

seems  contradicted with

>the only thing needs openssl is that mysql/postgresql sql driver plugin, because they need openssl, so that makes qtsql indirectly need openssl.

What is the right thing to do if a user choose -securetransport and -plugin-sql-mysql ?

steve3d:
well, let us make sure about this: mysql/postgresql can not be built and run without openssl(homebrew openssl), right? and like I said before, the openssl dependence of mysql/postgresql plugin is only needed at run time.

so you can not build mysql/postgresql driver plugin without mysql/postgresql installed, and after you install mysql/postgresql, then openssl must have been installed, so, we don't need to explicitly depends on openssl any more, because if you need mysql plugin, mysql already depends on openssl and pulls openssl in.

So the solution is quite simple, just remove `-lssl -lcrypto` option from mysql_config's result in order to build the sql driver plugin, because this solution is very mac specific, and I doubt qt's crew will make this default behavior. so, just a simple patch to configure script like what I do in my PR.

ilovezfs:
>and after you install mysql/postgresql, then openssl must have been installed

That's actually not necessarily true since the formula's sql options uses the requirements `:mysql` and `:postgresql`, which may or may not have been satisfied by the homebrew `mysql` and `postgresql` formulae, which do happen to depend on the `openssl` formula.

yezezey:
> So the solution is quite simple, just remove -lssl -lcrypto option from mysql_config's result in order to build the sql driver plugin, because this solution is very mac specific, and I doubt qt's crew will make this default behavior. so, just a simple patch to configure script like what I do in my PR.

I have no objection to do that.

@DomT4
Are you OK with @steve3d's solution?

steve3d:
@ilovezfs , sorry, I don't understand what you said, do you mean a user can install mysql/mariadb/postgresql without openssl? I think that's not possible, I've just checked all these have requirement of openssl.

ilovezfs:
@steve3d Dominick detailed some of the possibilities here: https://github.com/Homebrew/homebrew-core/pull/6374#discussion_r85643219

Basically the user may have any random mysql or postgresql. Not necessarily Homebrew's:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements.rb#L56-L68

steve3d:
@ilovezfs , Ok, I've got the point here, if homebrew also need to take care so many things, we can just remove the mysql/postgresql things from qt5's formula, leave it to who ever needs it.

remove these or keep these are both painful.

yezezey:
So what should we do here with `./configure -securetransport -plugin-sql-mysql` ?

Qt5's formula supports only Homebrew mysql and do @steve3d's suggestion to compile successfully?
Or we drop support for the mysql plugin?

ilovezfs:
@steve3d there's precedent for depending on the formula itself instead of the special requirement, if there's a need to guarantee the plugin is going to be using openssl with its sql.

```
Josephs-MacBook-Pro:Formula joe$ ag 'depends_on "mysql"'|cat
apr-util.rb:22:  depends_on "mysql" => :optional
gdal.rb:49:  depends_on "mysql" => :optional
mediatomb.rb:26:  depends_on "mysql" => :optional
mydumper.rb:21:  depends_on "mysql"
ntopng.rb:48:  depends_on "mysql" if build.without? "mariadb"
pike.rb:37:  depends_on "mysql"         if build.with?("mysql")   || build.with?("all")
qt.rb:30:  depends_on "mysql" => :optional
Josephs-MacBook-Pro:Formula joe$ ag 'depends_on "postgresql"'|cat
apr-util.rb:21:  depends_on "postgresql" => :optional
citus.rb:16:  depends_on "postgresql"
gdal.rb:48:  depends_on "postgresql" => :optional
inspircd.rb:26:  depends_on "postgresql" => :optional
kore.rb:19:  depends_on "postgresql" => :optional
liblwgeom.rb:28:    depends_on "postgresql" => :build # don't maintain patches for HEAD
mapnik.rb:26:  depends_on "postgresql" => :optional
mapserver.rb:32:  depends_on "postgresql" => :optional unless MacOS.version >= :lion
nagios-plugins.rb:16:  depends_on "postgresql" => :optional
pgroonga.rb:16:  depends_on "postgresql"
pgrouting.rb:20:  depends_on "postgresql"
pldebugger.rb:19:  depends_on "postgresql"
postgis.rb:29:  depends_on "postgresql"
postgrest.rb:22:  depends_on "postgresql"
psql2csv.rb:9:  depends_on "postgresql"
qt.rb:31:  depends_on "postgresql" => :optional
uwsgi.rb:35:  depends_on "postgresql" => :optional
```

yezezey:
rebased.
Built successfully with `brew install -s qt5.rb --with-mysql --with-postgresql --without-webengine`

MikeMcQuaid:
> Can you prove that?

FYI: if you're proposing a change the burden of proof is on you.

steve3d:
I've checked your PR, I think you should not add `-L#{openssl_opt}/lib` as global link flags, because this will cause every qt model have this flags, 

Try create a simple qmake project and compile it you will see this, even if you only used QtCore, and this definitely don't need openssl, and there will be this link option exists.

This also apples to the CFLAGS, which will add with every qt module, if you set it globally.

DomT4:
What is likely to break if this PR does something like:
```diff
diff --git a/Formula/qt5.rb b/Formula/qt5.rb
index 304af1f..c8aa779 100644
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -82,10 +82,14 @@ class Qt5 < Formula
   # reported to fail: <https://github.com/Homebrew/homebrew/issues/45284>.
   depends_on :macos => :mountain_lion
 
-  depends_on "dbus" => :optional
-  depends_on :mysql => :optional
-  depends_on :postgresql => :optional
   depends_on :xcode => :build
+  depends_on "dbus" => :optional
+
+  # We need to know the *sql options are linked to OpenSSL to avoid
+  # potentially mixing and matching crypto libraries.
+  depends_on "mysql" => :optional
+  depends_on "postgresql" => :optional
+  depends_on "openssl" if build.with?("postgresql") || build.with?("mysql")
 
   depends_on OracleHomeVarRequirement if build.with? "oci"
 
@@ -125,8 +129,13 @@ class Qt5 < Formula
 
     args << "-nomake" << "examples" if build.without? "examples"
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
-    args << "-plugin-sql-psql" if build.with? "postgresql"
+    if build.with?("mysql") || build.with?("postgresql")
+      args << "-plugin-sql-mysql" if build.with? "mysql"
+      args << "-plugin-sql-psql" if build.with? "postgresql"
+      args << "-openssl-linked" << "-no-securetransport"
+      openssl_opt = Formula["openssl"].opt_prefix
+      ENV["OPENSSL_LIBS"] = "-L#{openssl_opt}/lib -lssl -lcrypto"
+    end
 
     if build.with? "dbus"
       dbus_opt = Formula["dbus"].opt_prefix
```
?

Honestly, my personal feeling is that the options really aren't worth this hassle, but there we go.

RandomDSdevel:
_My_ opinion on all of this is that we should just add '`depends_on "openssl" if build.with?("postgresql") || build.with?("mysql")`' (or _maybe_ make the slightly more complicated change that @DomT4 suggested) to `qt5.rb`.  Any other actionable concerns should get their fixes postponed to other issues and/or PRs.  As far as I can tell, these include:  

- Fixing the flags passed to the linker to build Qt while doing that so that said linker _only_ links MySQL and/or PostgreSQL (and/or their mutual OpenSSL dependency) to the `QMYSQL` and/or `QPSQL` plug-ins distributed as part of QtCore (or QtCore itself, if we find that we can't specify different linker flags at that fine of a grain) (at least I _think_ the `QMYSQL` and `QPSQL` plug-ins are in QtCore…,) _not_ any of Qt's _other_ internal libraries/packages.  (This may apply to compiler flags as well, but could we possibly deal with those in a separate PR…?)  
- Figure out once and for all whatever we want to do with regard to telling Homebrew to generate pre-compiled headers for Qt while building it if this is even _necessary_ any more (comments indicate it may not be since Qt may handle this on its own without Homebrew's intervention nowadays.)  
- Work with upstream to clarify what exactly the deal is when it comes to support for OpenSSL in Qt on OS X/macOS.  
  - (**NOTE:**  Given what @jakepetroules said about OpenSSL support and SecureTransport support, the latter of which now seems preferred by Qt on Apple platforms, being mutually exclusive, this may involve reporting some bugs to and/or making some contributions to Qt upstream&thinsp;—&thinsp;and perhaps MySQL and PostgreSQL as well, if needed&thinsp;—&thinsp;to un-deprecate OpenSSL support in Qt on Apple platforms&thinsp;—&thinsp;this would remain dependent on third-party package managers providing OpenSSL to users and them taking advantage of that, of course, since Apple no longer provides OpenSSL on its OSes&thinsp;—&thinsp;and, possibly, as a further step, modifying Qt and/or MySQL and/or PostgreSQL to allow users of the `QMYSQL` and `QPSQL` plug-ins to build them with support for _both_ the former _and_ latter cryptography libraries at compilation time and conditionally enable support for _either_ one at _run_ time, as they would _still_ be mutually exclusive _then_.)

ilovezfs:
>Honestly, my personal feeling is that the options really aren't worth this hassle, but there we go.

Given the amount of virtual ink that has been spilled on the conversation, and given that we now have at least two issue reports, three users, three competing PRs, three maintainers, one maintainer emeritus, and an upstream engineer involved, I'm gravitating toward this point of view.

yezezey:
As @steve3d pointed out, those SQL plugins do not need OpenSSL nor Secure Transport.
To build Qt5 successfully with `-securetransport -plugin-sql-mysql`, @steve3d's solution is the easiest way. 
We don't have to do `depends_on "openssl" if build.with?("postgresql") || build.with?("mysql")`

Anyway since my primary goal was add --with-openssl switch, those dependency complications with `-securetransport -plugin-sql-mysql` probably should be treated in a separate PR.

Sorry for this hassle.

yezezey:
@steve3d 
Thanks for the help.
I deleted -L flag.

yezezey:
We know from @steve3d's explanations, the culprit is the unwanted reference to `-lssl -lcrypto` by mysql_config.

```
$ mysql_config | grep lssl
        --libs           [-L/usr/local/Cellar/mysql/5.7.16/lib -lmysqlclient -lssl -lcrypto]
        --libs_r         [-L/usr/local/Cellar/mysql/5.7.16/lib -lmysqlclient -lssl -lcrypto]
        --libmysqld-libs [-L/usr/local/Cellar/mysql/5.7.16/lib -lmysqld -lssl -lcrypto]
```

I think you need to choose from two options.
If a user wants -plugin-sql-mysql:

**A) pure -securetransport with the ad hoc fix**
```
    if build.with? "mysql"
      args << "-plugin-sql-mysql"
      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"
    end

    if build.with? "openssl"
      args << "-openssl-linked" << "-no-securetransport"
      openssl_opt = Formula["openssl"].opt_prefix
      args << "-I#{openssl_opt}/include"
      ENV["OPENSSL_LIBS"] = "-L#{openssl_opt}/lib -lssl -lcrypto"
    end
```

**B) automatically switch to OpenSSL to eliminate the ad hoc fix.**
```
    if (build.with? "openssl") || (build.with? "mysql")
      args << "-openssl-linked" << "-no-securetransport"
      openssl_opt = Formula["openssl"].opt_prefix
      args << "-I#{openssl_opt}/include"
      ENV["OPENSSL_LIBS"] = "-L#{openssl_opt}/lib -lssl -lcrypto"
      if build.with? "mysql"
        args << "-plugin-sql-mysql"
        args << "-L#{openssl_opt}/lib"
      end
    end
```

Point out if I am wrong.

yezezey:
rebased to the plan A above.
built successfully with `brew install -s qt5 --without-webengine --with-mysql --with-postgresql`

Formula/qt5.rb
```diff
@@ -73,6 +73,7 @@ class Qt5 < Formula
   option "with-oci", "Build with Oracle OCI plugin"
   option "with-qtwebkit", "Build with QtWebkit module"
   option "without-webengine", "Build without QtWebEngine module"
+  option "with-openssl", "Build with OpenSSL"
```
MikeMcQuaid: Remove this option, thanks.

Formula/qt5.rb
```diff
@@ -86,6 +87,7 @@ class Qt5 < Formula
   depends_on :mysql => :optional
   depends_on :postgresql => :optional
   depends_on :xcode => :build
+  depends_on "openssl" if build.with?("openssl") || build.with?("postgresql") || build.with?("mysql")
```
MikeMcQuaid: Remove `build.with?("openssl")`

Formula/qt5.rb
```diff
@@ -125,8 +127,18 @@ def install
...
     args << "-nomake" << "examples" if build.without? "examples"
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
     args << "-plugin-sql-psql" if build.with? "postgresql"
+    if build.with? "mysql"
+      args << "-plugin-sql-mysql"
+      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"
+    end
+
+    if build.with? "openssl"
```
MikeMcQuaid: Remove this section.

MikeMcQuaid (changes requested):

Formula/qt5.rb
```diff
@@ -125,8 +127,18 @@ def install
 
     args << "-nomake" << "examples" if build.without? "examples"
 
-    args << "-plugin-sql-mysql" if build.with? "mysql"
     args << "-plugin-sql-psql" if build.with? "postgresql"
+    if build.with? "mysql"
+      args << "-plugin-sql-mysql"
+      inreplace "qtbase/configure", /(QT_LFLAGS_MYSQL_R|QT_LFLAGS_MYSQL)=\`(.*)\`/, "\\1=\`\\2 | sed \"s/-lssl -lcrypto//\"\`"
```
MikeMcQuaid: I suspect this will break SSL support in the MySQL client.

PR #6404, opened 2016-10-29
ircii: downgrade URL to HTTP

zmwangx:

---

Cert expired Sep 12, 2016.
#6408.

### DIFF
--- a/Formula/ircii.rb
+++ b/Formula/ircii.rb
@@ -1,7 +1,7 @@
 class Ircii < Formula
   desc "IRC and ICB client"
   homepage "http://www.eterna.com.au/ircii/"
-  url "https://ircii.warped.com/ircii-20151120.tar.bz2"
+  url "http://ircii.warped.com/ircii-20151120.tar.bz2"
   sha256 "5dfd3fd364a96960e1f57ade4d755474556858653e4ce64265599520378c5f65"
 
   bottle do


dunn:
Have you emailed the developers about it?  I can, just checking.

zmwangx:
@dunn I haven't. I checked the communications channels: according to http://www.eterna.com.au/ircii/ there's a seemingly unarchived user mailing list with "almost no traffic", then according to http://www.eterna.com.au/ircii/bugs.html the only way to report bugs is via `ircbug`. I looked at the script, and the destination is ircii-bugs@eterna.com.au, but you need to fill out a fairly confusing template, which is hard because cert expired on website isn't really a bug.

djmoch:
Another option would be to use the `--insecure` flag in `curl` to perform the download until their hosting provider can get their act together. I haven't thought through all of the implications of that, nor do I know if that's even an option you can set in the formula; I just wanted to suggest that as an option.

In any case, if I can do anything to help get this resolved soon, let me know.

zmwangx:
@djmoch Hey, are you related to the ircII project somehow, or do you know any details? (Guessing from the sentence "until their hosting provider can get their act together".)

As for `-k`, it's not supported by brew and I don't think it is any better than changing the URL to HTTP.

djmoch:
I'm not. I just did a little poking around and initially thought the expired cert was for `*.warped.com`, but now I'm not so sure. Could be a problem with the hosting provider or the ircii project. 

My feeling on HTTPS vs HTTP is that HTTPS still provides some benefit over HTTP since HTTP could theoretically be tampered with at any point in-transit, while that would cause errors in HTTPS (assuming you establish your session with the real endpoint). But that's all moot since brew doesn't support it.

zmwangx:
No one tampers with the SHA-256 check, unless your machine is hacked.

djmoch:
I was just going to mention that. You're completely right. I'm just being paranoid.

dunn:
There's also `ircii AT eterna.com.au`; I just sent an email there.

zmwangx:
Ah yeah I missed that.

ilovezfs:
Looks like there are https mirrors with the same checksum

https://mirrors.ocf.berkeley.edu/debian/pool/main/i/ircii/ircii_20151120.orig.tar.bz2
https://mirrorservice.org/sites/ftp.debian.org/debian/pool/main/i/ircii/ircii_20151120.orig.tar.bz2

zmwangx:
Hmm why the merge?

ilovezfs:
Certs expired. Checksum matches. This is good enough for me.

zmwangx:
I thought it wasn't ;)

ilovezfs:
It is when we have 117 open PRs.

djmoch:
Still no bottle?

dunn:
@djmoch I'll add one with #6896.

PR #6416, opened 2016-10-30
ghc: apply 10.12 dynlibdir patch directly to 8.0.1

ilovezfs:
avoid unfixed issue with monad-logger and 8.0 branch
add a head spec

### DIFF
--- a/Formula/ghc.rb
+++ b/Formula/ghc.rb
@@ -5,18 +5,34 @@ class Ghc < Formula
 
   desc "Glorious Glasgow Haskell Compilation System"
   homepage "https://haskell.org/ghc/"
-  revision 1
+  revision 2
 
   stable do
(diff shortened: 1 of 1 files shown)


ilovezfs:
monad-logger problem reported here: https://github.com/kazu-yamamoto/logger/issues/106

ilovezfs:
The monad-logger problem has been traced to http://git.haskell.org/ghc.git/commit/d2958bd08a049b61941f078e51809c7e63bc3354 and https://ghc.haskell.org/trac/ghc/ticket/12784 is the upstream report. The consensus so far seems to be that the real bug is in monad-logger, not GHC. So hopefully we can avoid merging this PR if a monad logger fix is in the offing.

ilovezfs:
upstream monad-logger has tagged a new release with a fix so this won't be needed

PR #6439, opened 2016-10-31
dc3dd 7.2.646

ilovezfs:


ilovezfs:
zip problem reported here https://sourceforge.net/p/dc3dd/bugs/14/

PR #6449, opened 2016-10-31
gnutls: actually disable Guile when not requested

astiob:

-----

I think this does not affect superenv (?), but this can be useful if the user requests stdenv, where GnuTLS will automatically pick up Guile if it’s installed and linked.

### DIFF
--- a/Formula/gnutls.rb
+++ b/Formula/gnutls.rb
@@ -42,7 +42,12 @@ class Gnutls < Formula
       --disable-heartbeat-support
       --without-p11-kit
     ]
-    args << "--enable-guile" << "--with-guile-site-dir" if build.with? "guile"
+
+    if build.with? "guile"
+      args << "--enable-guile" << "--with-guile-site-dir"
+    else
+      args << "--disable-guile"
+    end
 
     system "./configure", *args
     system "make", "install"


DomT4:
> this can be useful if the user requests stdenv, where GnuTLS will automatically pick up Guile if it’s installed and linked.

I'd argue if you're going to deliberately set about using `stdenv` this is precisely the behaviour you should expect and are buying into, but there we go.

MikeMcQuaid:
> if the user requests stdenv

Why are you doing this, out of interest?

astiob:
Honest answer: I use stdenv myself to build with new FSF GCC, as I have local changes related to that in stdenv (mainly optimization flags, both regular and LTO) and am too lazy to port them over to superenv. I’ve been telling myself I should do this (and possibly submit some of my changes upstream) for… years? ever since we got superenv… but I haven’t gotten around to doing this.

Still, I do believe it would be nice to have stdenv work well while it exists. At least here, the effort to make it work isn’t large, so why not? Explicitly disabling disabled features is also consistent with other formulae and won’t break if the upstream default changes in future.

> I'd argue if you're going to deliberately set about using stdenv this is precisely the behaviour you should expect and are buying into, but there we go.

I think autodetecting might be fine if you’re compiling without explicitly enabling or disabling Guile, but the real problem is that it will pick it up even when you explicitly do `brew install gnutls --without-guile` and you have no way to force it off short of unlinking Guile.

ilovezfs:
Let's not proliferate these all over the place, but since guile is already an option, I don't see the harm in honoring the `--without-` in this case if it's useful to you.

astiob:
Thanks!

PR #6450, opened 2016-10-31
harfbuzz: demote required dependencies to recommended

astiob:

-----

All of these can be disabled, so expose this ability to the user.

I also sorted the conditionals alphabetically and the dependencies alphabetically within each group. I can undo this or split it out into a separate commit if desired.

### DIFF
--- a/Formula/harfbuzz.rb
+++ b/Formula/harfbuzz.rb
@@ -23,9 +23,9 @@ class Harfbuzz < Formula
   option "with-cairo", "Build command-line utilities that depend on Cairo"
 
   depends_on "pkg-config" => :build
-  depends_on "glib"
-  depends_on "freetype"
-  depends_on "gobject-introspection"
+  depends_on "freetype" => :recommended
+  depends_on "glib" => :recommended
+  depends_on "gobject-introspection" => :recommended
   depends_on "icu4c" => :recommended
   depends_on "cairo" => :optional
   depends_on "graphite2" => :optional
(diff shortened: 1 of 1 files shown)


dunn:
We try to make sensible choices for users to simplify package installation and maintenance.  Are these options ones that a large number of users wouldn't want?

astiob:
Personally, I don’t want any GLib or GObject in my HarfBuzz. To me, they’re just unnecessary bloat, and bloat that is expensive to install and maintain. I’m not sure how large or expensive the relevant code in HarfBuzz is, but GLib itself is a pain, and this code is not important for HarfBuzz’s main functionality. I’d even go so far as to make them optional rather than recommended and I don’t know why they were required in the first place, but precisely because I don’t know, I kept the default the same as before. Over at libass (where I’m a contributor), we keep seeing people think HarfBuzz is a pain to install because it needs GLib and we have to explain that it actually really doesn’t and is quick and easy to build.

ilovezfs:
Is the motivation here something to do with linuxbrew?

astiob:
No. I’m on OS X myself, although still on 10.6. If you’re thinking that installing GLib is easy on macOS because it’s bottled, yes, but bottles aren’t always applicable; Linux and old OS X are both out of luck. And GLib comes with a bunch of dependencies that are completely irrelevant to HarfBuzz but still need to be downloaded and installed if GLib is required.

Also, while this isn’t terribly relevant for most Homebrew users, GLib is pretty large relative to HarfBuzz (without even taking into account its dependencies). If someone uses Homebrew to build a binary distribution of some software involving HarfBuzz, disabling GLib would be useful for them. This may sound far-fetched, but this [has been done](https://github.com/pigoz/mplayerosx-builds). The other environment where this might matter are embedded/tiny systems, but I guess they aren’t a real concern for us. Although I imagine it should actually be possible to set up Linuxbrew on one.

To be honest, I didn’t expect this change to cause any amount of controversy. It doesn’t harm anyone and gives more power to the user. But, of course, if you guys think this is in fact harmful, so be it.

ilovezfs:
The controversy is just that more options mean more source builds, which require support when they fail, and each option has to be maintained over time if upstream changes configure. Also, more options mean more untested combinations, since our CI only tests the default.

astiob:
If this is too many options, as a compromise, I would personally be happy with just a single option that disables both glib and gobject-introspection (while keeping freetype a required dep).

ilovezfs:
It's not necessarily a problem. Just making sure the trade-offs are clear to you. This PR may indeed be accepted ultimately.

astiob:
Right; I’m sorry if I sounded rude.

It occurred to me today that if this goes through with a nonrequired FreeType dependency, the libass formula will need to be updated to `depends_on "harfbuzz" => [:recommended, "with-freetype"]` (did I get the syntax right?). Should I do this in the same commit, add a separate commit, or just wait for now?

ilovezfs:
@astiob if freetype is :recommended, you do not need the "with-freetype" option, which actually has no impact other than triggering a source build. See https://github.com/Homebrew/brew/issues/1399.

ilovezfs:
@BrewTestBot test this please

MikeMcQuaid (approved):
I'm happy enough with this as-is when CI is happy 👍

ilovezfs:
@BrewTestBot test this please

ilovezfs:
I resolved the problems CI I was tripping on, and now we have :green_apple:

ilovezfs:
@astiob Thanks for your contribution to Homebrew! :fireworks:

PR #6451, opened 2016-10-31
theora: add devel version 1.2.0alpha1

astiob:

-----

The formula doesn’t pass audit because there is no test, but I hope this change can be accepted regardless.

**Edit:** Hm, what’s the right way to format the commit message for this change?

### DIFF
--- a/Formula/theora.rb
+++ b/Formula/theora.rb
@@ -43,23 +43,9 @@ class Theora < Formula
       --disable-examples
     ]
 
-    if build.devel?
-      # Clang's integrated assembler can't handle integer constant
-      # expressions that are used in some of the inline assembler code
-      if clang_assembler?
-        args << "--disable-asm"
-      else
-        ENV["AS_INTEGRATED_ASSEMBLER"] = nil
-      end
-    end
+    args << "--disable-asm" unless build.stable?
 
     system "./configure", *args
     system "make", "install"
-  end
-
-  def clang_assembler?
-    return true if MacOS::Xcode.installed? && MacOS::Xcode.version >= "7.0"
-    return true if ENV.compiler == :clang && DevelopmentTools.clang_build_version >= 100
-    false
   end
 end
--- a/Formula/theora.rb
+++ b/Formula/theora.rb
@@ -15,8 +15,8 @@ class Theora < Formula
 
   devel do
     url "http://downloads.xiph.org/releases/theora/libtheora-1.2.0alpha1.tar.xz"
-    sha256 "5be692c6be66c8ec06214c28628d7b6c9997464ae95c4937805e8057808d88f7"
     version "1.2.0alpha1"
+    sha256 "5be692c6be66c8ec06214c28628d7b6c9997464ae95c4937805e8057808d88f7"
   end
 
   head do
--- a/Formula/theora.rb
+++ b/Formula/theora.rb
@@ -13,6 +13,12 @@ class Theora < Formula
     sha256 "58be26743e23be63aee48186bfa9cd8a982de957efb040a6ab3030aa62753977" => :mavericks
   end
 
+  devel do
+    url "http://downloads.xiph.org/releases/theora/libtheora-1.2.0alpha1.tar.xz"
+    sha256 "5be692c6be66c8ec06214c28628d7b6c9997464ae95c4937805e8057808d88f7"
+    version "1.2.0alpha1"
+  end
+
   head do
     url "https://git.xiph.org/theora.git"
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
build failure https://bot.brew.sh/job/Homebrew%20Core/10454/version=any_sierra/console

astiob:
Sigh. Either I screwed up somewhere or it breaks on newer versions of macOS/Xcode. I’ll try to debug this in the next few days.

ilovezfs:
@astiob ping

astiob:
Yes; sorry for the delay.

The newer libtheora’s `lib/x86/mmxidct.c` emits inline assembler code with memory offsets that are arithmetic expressions rather than single numbers, e. g. `(3)*16+(%rsi)` rather than `48(%rsi)`. It appears that Apple’s build of GNU assembler that comes with their cctools (both my system’s cctools-795~45 and brewn cctools-855) supports this but the LLVM integrated assembler doesn’t. I was unable to find any information about this discrepancy or this topic in general on the Web, aside from a vague description of expressions in GNU assembler’s manual that makes no mention of where they are allowed.

I see four solutions, ranging from the most complicated option that gives the best result to the easiest option that gives the worst result:
  1. Write a patch to make libtheora evaluate these numbers during compilation rather than assembly.
  2. Fudge the compiler flags to stop Clang from using the integrated assembler for devel. What does macOS use as the system `as` these days? If it’s the LLVM integrated assembler, then also add `depends_on "cctools" => :build` for devel. Would this work? I mean, would superenv actually let it pick up the `as` from cctools rather than the system `as`?
  3. Disable all assembler optimizations in libtheora for devel by adding `--disable-asm` to the configure arguments. It can even probably be limited to `if ENV.compiler == :clang` if the system `as` is GAS (to avoid unduly penalizing builds that are capable of handling the presumably faster assembler code), or a more complicated expression involving macOS/Xcode version numbers if not.
  4. Not add the devel version at all.

Which do you prefer? Personally, I’m tempted to go with `--disable-asm` due to its simplicity. I’m not sure how much of a performance hit it might be, but having a newer version of the Theora encoder would still be good.

astiob:
Another ugly option: write a patch to exclude the particular problematic file from the build while keeping the rest of the assembler code. Similarly to other options, ideally this would come with a guarding condition to avoid punishing builds using the GNU assembler.

MikeMcQuaid:
> Disable all assembler optimizations in libtheora for devel by adding --disable-asm to the configure arguments. To avoid unduly penalizing builds that are capable of handling the presumably faster assembler code, it can probably even be limited to if ENV.compiler == :clang if the system as is GAS or a more complicated expression involving macOS/Xcode version numbers if not.

This seems fine to me. Would be good to report it upstream so that they can fix it for the stable release.

astiob:
Do you know whether (and starting from which version) the standalone `as` is the LLVM integrated assembler (by default)? `as -v < /dev/null` should tell you. I don’t have Sierra and Xcode 8 myself, and I seem to see conflicting information on the Web.

> Would be good to report it upstream so that they can fix it for the stable release.

Yeah, I want to do this. It’s hard to tell how much one can expect from further libtheora development, but their Git actually got a few new commits recently, and reporting this will do no harm anyway.

astiob:
Investigating `as/driver.c` in Apple’s cctools shows:

* Starting from Xcode 5.0 (and not in any 4.x), Clang’s integrated assembler is used when `AS_INTEGRATED_ASSEMBLER` is set (to any value) in the environment. (And always for arm64.)
* Starting from Xcode 7.0 (and not in any 6.x), Clang’s integrated assembler is used by default.

I’ve also received confirmation that Xcode 8 uses Clang’s integrated assembler by default.

Of course, starting from Xcode 4.0, it is also possible to force Clang’s integrated assembler by invoking `as -q` and to force GNU assembler by invoking `as -Q` (except for arm64), but Homebrew doesn’t do either, so this isn’t very relevant.

On the other hand, Homebrew doesn’t touch `AS_INTEGRATED_ASSEMBLER` at all, so the user’s environment can affect the choice of assembler during Homebrew builds using GCC on older versions of Xcode.

Investigating Apple’s tags in Clang’s Subversion repository shows that it uses the integrated assembler by default starting from build 100. (It doesn’t use it by default in build 78, and there are no tags in between.)

To account for all of this, my new commit disables libtheora inline assembler code on Xcode 7+ (because it will fail regardless of the compiler) and on Clang build 100+ (because it will fail in this compiler) and unsets `ENV["AS_INTEGRATED_ASSEMBLER"]` in all other cases to ensure that GCC builds with earlier Xcode succeed in building the inline assembler code by using the GNU assembler. I’m not sure about two things:

* What about CLT-only systems? Do I need to handle them separately, or does `MacOS::Xcode` do this for me?
* What about Linuxbrew etc.? (Do we care?) Does it support Clang? `clang_build_version` will be 0 if Clang isn’t Apple’s, so my condition will fail to detect the integrated assembler.

I’ve tested this as well as I could, and I guess we’ll see how it fares on an actual Sierra/Xcode 8 install (which I still don’t have, sorry) shortly.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

MikeMcQuaid:
@ilovezfs Any thoughts here?

fxcoudert:
This alpha version is 7 years old and does not compile out of the box. Development on that branch has all but stopped (8 commits over 2 years).

Either option #4. not add the devel version, or #3. `--disable-asm` everywhere, until upstream fixes it.

ilovezfs:
assuming it affects head too(?), so `unless build.stable?`

fxcoudert:
Seems ready to merge?

PR #6456, opened 2016-11-01
ffmpeg: add more dependencies and fix libbluray dependency

astiob:

-----

I wanted to also add a dependency on `celt` but discovered that it’s in the boneyard. I’m guessing we can’t have any dependencies on boneyard formulae in the core, is that right? It’s a bit unfortunate (e. g. in this case functionality is lost IIUC), but this probably isn’t the place to discuss that.

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -111,6 +111,7 @@ class Ffmpeg < Formula
     args << "--enable-ffplay" if build.with? "sdl2"
     args << "--enable-frei0r" if build.with? "frei0r"
     args << "--enable-libass" if build.with? "libass"
+    args << "--enable-libbluray" if build.with? "libbluray"
     args << "--enable-libbs2b" if build.with? "libbs2b"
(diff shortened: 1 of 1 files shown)


dunn:
> I’m guessing we can’t have any dependencies on boneyard formulae in the core, is that right? 

That's correct, but if you can get the latest version to build on macOS again we can bring it back out of the boneyard.

astiob:
In the case of CELT, it was moved to the boneyard because upstream development was halted, and the boneyard formula is in fact the latest released version as far as I can tell, so there’s not much that can be done there. (Upstream developers have moved on to Opus, but, if I understand correctly, Opus isn’t directly backwards-compatible with CELT, and this is the only available CELT implementation.)

ilovezfs:
Did you test if these options actually build?

astiob:
Yes, but then I’m on 10.6, so I have to say I’m not entirely sure that they still build on new versions of macOS.

MikeMcQuaid:

PR #6468, opened 2016-11-01
coffeescript 1.11.1 (new formula)

cglong:

### DIFF
--- a/Aliases/coffeescript
+++ b/Aliases/coffeescript
@@ -1,0 +1,1 @@
+../Formula/coffee-script.rb
--- a/Aliases/coffeescript
+++ b/Aliases/coffeescript
@@ -1,1 +1,0 @@
-../Formula/coffee-script.rb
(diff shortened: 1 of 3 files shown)


Aliases/coffeescript
```diff
@@ -0,0 +1 @@
+../Formula/coffee-script.rb
```
bfontaine: I think it should be named `coffeescript` directly since that’s the name upstream uses.

bfontaine (dismissed):

Aliases/coffeescript
cglong: I'm tempted to agree with you @bfontaine; the reason I went with this naming scheme originally is that, while the `git` repo calls it [`coffeescript`](https://github.com/jashkenas/coffeescript), the main entry point for users (`npm`) calls it [`coffee-script`](https://www.npmjs.com/package/coffee-script). I'm fine with either name, so let me know your thoughts and I can make any changes if necessary 😄

MikeMcQuaid: Let's go for `coffeescript`, thanks!

MikeMcQuaid (approved):

MikeMcQuaid:

PR #6510, opened 2016-11-02
libnice: migrate to boneyard

ilovezfs:


ilovezfs:
No reverse dependencies. Not updated in forever. Originally added for telepathy-gabble, which is now in the boneyard.

PR #6527, opened 2016-11-02
webkitgtk 2.14.1

tschoonj:
This will be fun...

### DIFF
--- a/Formula/webkitgtk.rb
+++ b/Formula/webkitgtk.rb
@@ -1,9 +1,9 @@
 class Webkitgtk < Formula
   desc "Full-featured Gtk+ port of the WebKit rendering engine"
   homepage "https://webkitgtk.org/"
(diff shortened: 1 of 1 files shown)


DomT4:
Good luck with this one. I tried a few times, it broke in a new way after every attempted fix and so I gave up and hoped you'd notice it 😄.

tschoonj:
Thanks! I managed to get it compiling apart from the gobject-introspection bindings on my El Cap machine. If this works fine on the Yosemite bot, I will be able to confirm it's an SIP/python issue...

MikeMcQuaid:
@tschoonj Feel free to ship an older version. Newest > newer > older 😁

ilovezfs:
@tschoonj Just to be clear that means dropping introspection support even if we "only" ship 2.14.1, right?

tschoonj:
Yep, but the gobject-introspection stuff never worked in 2.10.9 anyway (https://github.com/Homebrew/legacy-homebrew/issues/47000)

PR #6566, opened 2016-11-03
hadoop-native: Add build from source that includes native libraries

onlynone:

-----

Not sure exactly how this should be handled vis-a-vis the original hadoop formula. Basically, the hadoop project distributes some native libraries with their pre-compiled package, but they're only for linux. One can compile them for any platform, but that requires the source package (and some tweaks to the build process). I've done that work in this `hadoop-native` formula (based on the feedback here: http://stackoverflow.com/questions/37800605/hadoop-native-libraries-not-found-on-os-x).

I wasn't sure if it would be better to:

 * Modify the existing `hadoop` formula to be built from source
 * Create a new `hadoop-native` formula that's built from source while leaving the existing formula alone (the approach I took so far)
 * Create a new `hadoop-native` formula that only includes the native libraries from the source build and figures out a way to integrate them with the existing `hadoop` installed files

Some caveats to think about:

 * This build requires `homebrew/versions/protobuf250`
 * For the `bin/hadoop` script (and others) to be able to find the native libs, they need to be run with `JAVA_LIBRARY_PATH` set to `/usr/local/opt/openssl/lib`. I did this by creating an env wrapper script, but these are the same scripts created by `hadoop`
 * Some other xml configuration files need to be adjusted as well

I think the last two points would mean that the third option from above wouldn't work.

Would love some feedback.

### DIFF
--- a/Formula/hadoop-native.rb
+++ b/Formula/hadoop-native.rb
@@ -1,0 +1,129 @@
+class HadoopNative < Formula
+  desc "Hadoop with native libraries"
+  homepage "https://hadoop.apache.org/"
+  url "https://www.apache.org/dyn/closer.cgi?path=hadoop/common/hadoop-2.7.3/hadoop-2.7.3-src.tar.gz"
+  mirror "https://archive.apache.org/dist/hadoop/common/hadoop-2.7.3/hadoop-2.7.3-src.tar.gz"
+  sha256 "227785dc6e3e6ef8cfd64393b305d09078a209703c9c01910a1bddcf86be3054"
+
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> Modify the existing hadoop formula to be built from source

This would be best.



> This build requires homebrew/versions/protobuf250

You should use a `resource` for this instead.

Otherwise: sounds good 👍

ilovezfs:
@BrewTestBot test this please

fxcoudert:
Closing this for lack of activity at this point.

PR #6568, opened 2016-11-04
xapian 1.4.1

dunn:

-----


dunn:
@jaylett if you want to re-do your fixes to the bindings from #3643, that'd be very welcome!

jaylett:
@dunn Most of the changes were in this PR, or aren't needed for v1.4.1. I think the only thing remaining is the open question of whether to require sphinx in homebrew and use sphinx-build from there. People have been asking in other contexts if we can ship documentation in the tarballs, so I think for the time being we might be better off leaving it and seeing if it gets in people's way in future.

dunn:
OK, thanks!

PR #6586, opened 2016-11-04
rebar: install zsh/bash/fish shell completions

rjcoelho:


MikeMcQuaid:

PR #6588, opened 2016-11-04
gtkextra: use autoreconf to avoid flat_namespace

tschoonj:
The gtkextra tarball was generated with an old libtool, causing the flat_namespace to get passed to the linker...


ilovezfs:
Should this be reported?

tschoonj:
I did. He will make sure to use the latest version of autotools next time before making the tarball

ilovezfs:
Thanks!

PR #6598, opened 2016-11-04
curl: always be explicit with both ca-bundle and ca-path

lembacon:

-----

Otherwise, automatic detection of ca-bundle and/or ca-path may become problematic.

For example, `/etc/ssl` does not exist prior to macOS Sierra 10.12.2.  Starting with macOS Sierra 10.12.2 (16C32f), `/etc/ssl/cert.pem` has been shipped together with LibreSSL (see `/usr/lib/libssl.35.dylib` & `/usr/lib/libcrypto.35.dylib`), cURL will pick up `/etc/ssl/cert.pem` as the default ca bundle via automatic detection.  When building cURL with Secure Transport, the default ca bundle will provide anchor certificates exclusively, this will cause all user trust settings become ineffective.

### DIFF
--- a/Formula/curl.rb
+++ b/Formula/curl.rb
@@ -69,12 +69,16 @@ class Curl < Formula
       ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["libressl"].opt_lib}/pkgconfig"
       args << "--with-ssl=#{Formula["libressl"].opt_prefix}"
       args << "--with-ca-bundle=#{etc}/libressl/cert.pem"
+      args << "--with-ca-path=#{etc}/libressl/certs"
     elsif MacOS.version < :mountain_lion || build.with?("openssl") || build.with?("nghttp2")
       ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["openssl"].opt_lib}/pkgconfig"
       args << "--with-ssl=#{Formula["openssl"].opt_prefix}"
       args << "--with-ca-bundle=#{etc}/openssl/cert.pem"
+      args << "--with-ca-path=#{etc}/openssl/certs"
     else
       args << "--with-darwinssl"
+      args << "--without-ca-bundle"
+      args << "--without-ca-path"
     end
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Can you elaborate a bit more on this and provide some reproduction instructions for the issue you describe? Thanks!

lembacon:
@MikeMcQuaid Sure.

Basically, without this, verification of self-signed TLS certificates marked as `Always Trusted` via Keychain will fail, if `/etc/ssl/cert.pem` excluding the specific self-signed TLS certificate is presented  when building with Secure Transport.

Steps to reproduce,

1. Make sure the CA bundle at `/etc/ssl/cert.pem` or [somewhere else](https://github.com/curl/curl/blob/curl-7_51_0/acinclude.m4#L2640-L2644)  is presented.  (Note: Apple ships `/etc/ssl/cert.pem` starting macOS Sierra 10.12.2 (16C32f).)

1. Build cURL with Secure Transport enabled.

1. Add a self-signed TLS certificate to the login keychain and mark it as `Always Trusted`.

1. Use cURL to access arbitrary resources signed with the aforementioned self-signed TLS certificate.

1. cURL will exit with failure since it is [unable to verify the certificate](https://github.com/curl/curl/blob/curl-7_51_0/lib/vtls/darwinssl.c#L1828-L1832).  This is due to the fact that the CA bundle will __exclusively__ [provide anchor certificates](https://github.com/curl/curl/blob/curl-7_51_0/lib/vtls/darwinssl.c#L1804-L1813).

MikeMcQuaid:
Seems reasonable. Any objections, @ilovezfs?

ilovezfs:
Does the system curl recognize the "legitimacy" of self-signed certificates in the login keychain?

lembacon:
@ilovezfs Yes, it does.

ilovezfs:
Then this seems fine.

MikeMcQuaid:

PR #6611, opened 2016-11-04
libosmium 2.9.0 (new formula)

thomersch:

UPDATE: passes the audit now

By the way: I have been talking to the original author of (lib-)osmium, he is fine with adding it to Homebrew. I guess we can remove the blacklisting now.

### DIFF
--- a/Formula/libosmium.rb
+++ b/Formula/libosmium.rb
@@ -14,7 +14,8 @@ class Libosmium < Formula
 
   def install
     mkdir "build" do
-    system "cmake", *std_cmake_args, "-DINSTALL_GDALCPP=ON", "-DINSTALL_PROTOZERO=ON", "-DINSTALL_UTFCPP=ON", ".."
-    system "make", "install"
+      system "cmake", *std_cmake_args, "-DINSTALL_GDALCPP=ON", "-DINSTALL_PROTOZERO=ON", "-DINSTALL_UTFCPP=ON", ".."
+      system "make", "install"
+    end
   end
 end
--- a/Formula/libosmium.rb
+++ b/Formula/libosmium.rb
@@ -1,0 +1,21 @@
+class Libosmium < Formula
+  desc "Fast and flexible C++ library for working with OpenStreetMap data."
+  homepage "http://osmcode.org/libosmium/"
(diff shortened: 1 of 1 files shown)


Formula/libosmium.rb
```diff
@@ -0,0 +1,21 @@
...
+  depends_on "boost" => :build
+  depends_on "google-sparsehash" => :optional
+  depends_on "expat" => :optional
+  depends_on "gdal" => :optional
+  depends_on "proj" => :optional
+  depends_on "doxygen" => :optional
+
+  def install
+    mkdir "build"
+    chdir "build"
```
bfontaine: Use `mkdir "build" do` instead.

Formula/libosmium.rb
```diff
@@ -0,0 +1,21 @@
...
+  url "https://github.com/osmcode/libosmium/archive/v2.9.0.tar.gz"
+  sha256 "d4312c0cd9cca41fbf2927aee67cbb95d566540a76bb01fc643e8ba76db13419"
+
+  depends_on "cmake" => :build
+  depends_on "boost" => :build
+  depends_on "google-sparsehash" => :optional
+  depends_on "expat" => :optional
+  depends_on "gdal" => :optional
+  depends_on "proj" => :optional
+  depends_on "doxygen" => :optional
```
bfontaine: How common are these? We usually try to provide sane defaults and limit the number of options.

bfontaine (changes requested):

Formula/libosmium.rb
thomersch: libosmium itself is rather versatile, so if users are doing some slightly sophisticated geo transformations, they will need gdal/proj.
Expat is for additional file format parsing capabilities, so I would say it's rather useful.

Doxygen can generate documentation, not the most necessary thing, but can be useful since it's a library.

MikeMcQuaid:
Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

bfontaine (approved):

MikeMcQuaid:

PR #6638, opened 2016-11-05
ZeroMQ 4.2.0 and CZMQ 4.0.1

paddor:

-----

### DIFF
--- a/Formula/czmq.rb
+++ b/Formula/czmq.rb
@@ -1,9 +1,8 @@
 class Czmq < Formula
   desc "High-level C binding for ZeroMQ"
   homepage "http://czmq.zeromq.org/"
-  url "https://github.com/zeromq/czmq/releases/download/v3.0.2/czmq-3.0.2.tar.gz"
-  sha256 "8bca39ab69375fa4e981daf87b3feae85384d5b40cef6adbe9d5eb063357699a"
-  revision 3
+  url "https://github.com/zeromq/czmq/releases/download/v4.0.1/czmq-4.0.1.tar.gz"
+  sha256 "0fc7294d983df7c2d6dc9b28ad7cd970377d25b33103aa82932bdb7fa6207215"
 
   bottle do
     cellar :any
@@ -22,6 +21,7 @@ class Czmq < Formula
   end
 
   option :universal
+  option "with-drafts", "Build and install draft classes and methods"
 
   depends_on "pkg-config" => :build
   depends_on "libsodium" => :recommended
@@ -39,6 +39,7 @@ class Czmq < Formula
 
     args = ["--disable-dependency-tracking", "--prefix=#{prefix}"]
     args << "--with-libsodium" if build.with? "libsodium"
+    args << "--enable-drafts" if build.with?("drafts") || build.head?
 
     system "./autogen.sh" if build.head?
     system "./configure", *args
--- a/Formula/czmq.rb
+++ b/Formula/czmq.rb
@@ -23,28 +23,24 @@ class Czmq < Formula
   option :universal
   option "with-drafts", "Build and install draft classes and methods"
 
+  depends_on "asciidoc" => :build
   depends_on "pkg-config" => :build
-  depends_on "libsodium" => :recommended
+  depends_on "xmlto" => :build
+  depends_on "zeromq"
 
   conflicts_with "mono", :because => "both install `makecert` binaries"
 
-  if build.without? "libsodium"
-    depends_on "zeromq" => "without-libsodium"
-  else
-    depends_on "zeromq"
-  end
-
   def install
     ENV.universal_binary if build.universal?
+    ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
 
     args = ["--disable-dependency-tracking", "--prefix=#{prefix}"]
-    args << "--with-libsodium" if build.with? "libsodium"
-    args << "--enable-drafts" if build.with?("drafts") || build.head?
+    args << "--enable-drafts" if build.with? "drafts"
 
     system "./autogen.sh" if build.head?
     system "./configure", *args
     system "make"
-    system "(ZSYS_INTERFACE=lo0 && make check-verbose)"
+    system "make", "ZSYS_INTERFACE=lo0", "check-verbose"
     system "make", "install"
     rm Dir["#{bin}/*.gsl"]
   end
--- a/Formula/zeromq.rb
+++ b/Formula/zeromq.rb
@@ -26,31 +26,26 @@ class Zeromq < Formula
 
   deprecated_option "with-pgm" => "with-libpgm"
 
+  depends_on "asciidoc" => :build
   depends_on "pkg-config" => :build
+  depends_on "xmlto" => :build
   depends_on "libpgm" => :optional
   depends_on "libsodium" => :optional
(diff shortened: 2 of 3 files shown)


dunn:
A number of failing tests: https://bot.brew.sh/job/Homebrew%20Core/10776/#showFailuresLink

Could you also squash this two two commits, one for each formula? Thanks!

ilovezfs:
This PR is going to be problematic unless you chase https://github.com/zeromq/czmqpp/issues/43 and you may need to coordinate with https://github.com/Homebrew/homebrew-core/pull/5611.

paddor:
I'm sorry, I don't understand. What exactly do the tests verify?

Are you sure I can squash commits after already pushing to github?

ilovezfs:
>I'm sorry, I don't understand. What exactly do the tests verify?

The tests check that the other formulae that use the ones you're changing don't break as a result of the changes you've made. A formula has broken if its test block doesn't run successfully or if any of the libraries to which its programs link no longer exist.

>Are you sure I can squash commits after already pushing to github?

Yes, but you're correct that an ordinary push would fail. You have to use a force push (that's the -f option).

paddor:
Thanks for the help. I squashed the changes into two commits.

However, I have no idea what to do about fontforge and czmqpp. Leaving this here for now.

dunn:
They probably just need `revision 1` added under the `sha256` in their formula (or `revision 2` etc).  That will rebuild them against the new version of these libraries.

ilovezfs:
@dunn as I mentioned above: https://github.com/zeromq/czmqpp/issues/43.

dunn:
Oh, right.

ilovezfs:
@paddor @dunn I've added three commits here with some formula changes, and skirted the FontForge build failure by vendoring czmq 3.0.2. I boneyarded czmqpp the other day so that is no longer an issue.

ilovezfs:
FontForge reported upstream: https://github.com/fontforge/fontforge/issues/2950
CC @jtanx

Formula/czmq.rb
```diff
@@ -22,28 +21,26 @@ class Czmq < Formula
   end
 
   option :universal
+  option "with-drafts", "Build and install draft classes and methods"
```
MikeMcQuaid: Can you explain a bit more about what this does and why you'd want to enable/disable it?

Formula/zeromq.rb
```diff
@@ -11,47 +12,56 @@ class Zeromq < Formula
...
-
-    depends_on "autoconf" => :build
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
-  end
-
   option :universal
   option "with-libpgm", "Build with PGM extension"
   option "with-norm", "Build with NORM extension"
+  option "with-drafts", "Build and install draft classes and methods"
```
MikeMcQuaid: Can you explain a bit more about what this does and why you'd want to enable/disable it?

Formula/zeromq.rb
```diff
@@ -11,47 +12,56 @@ class Zeromq < Formula
...
   depends_on "libpgm" => :optional
   depends_on "libsodium" => :optional
+  depends_on "libtool" => :build
   depends_on "norm" => :optional
 
+  # Avoid known, intermittent "make check" failure
+  # upstream issue "make check error: tests/test_shutdown_stress"
+  # https://github.com/zeromq/libzmq/issues/1509
+  # Autotools dependencies can be removed from stable once this is fixed
+  patch :DATA
```
MikeMcQuaid: @ilovezfs Is there anything that should be upstreamed here?

Formula/zeromq.rb
```diff
@@ -11,47 +12,56 @@ class Zeromq < Formula
...
+
+    ENV["LIBUNWIND_LIBS"] = "-framework System"
+    ENV["LIBUNWIND_CFLAGS"] = "-I#{MacOS.sdk_path}/usr/include"
 
-    system "./autogen.sh" if build.head?
     system "./configure", *args
     system "make"
+    system "make", "check", "TESTS=''"
+    ENV.deparallelize do
+      system "sh", "-c", "ulimit -n 1200; make check"
```
MikeMcQuaid: Bit concerned that we need this and the previous call and the patch. Is it not just worth removing `make check` for now?

Formula/zeromq.rb
ilovezfs: Upstream has agreed to look into my findings:

https://github.com/zeromq/libzmq/issues/1509#issuecomment-260579648

Formula/zeromq.rb
ilovezfs:
```
/******************************************************************************/
/*  These functions are DRAFT and disabled in stable releases, and subject to */
/*  change at ANY time until declared stable.                                 */
/******************************************************************************/```

https://github.com/zeromq/libzmq/blob/06aeb47d73de5ee07de861e9f8dc3053154b5146/src/zmq_draft.h#L33-L36

The configure script enables it if you're building from a git repository or if you request the option.

Formula/czmq.rb
ilovezfs: Same situation as zeromq. It enables additional APIs. As an example,
https://github.com/zeromq/czmq/blob/58d0093d971d4563db631980ba7ff76c916d6a0a/src/zgossip_msg.h

ilovezfs:
@paddor can you explain why you added the draft options?

ilovezfs:
@MikeMcQuaid PR refreshed.

paddor:
@ilovezfs They're a new concept from Zproject, the code generation tool used in both ZeroMQ and CZMQ. With `--enable-drafts`, API classes/methods that are declared to be in _draft_ state (as opposed to _stable_) are also compiled. Otherwise, they're completely left out.

To be able to test new features that are coming up, there needs to be a way to enable these drafts. That's what the `with-drafts` option allows.

I see you removed the automatic enabling of these draft features when someone is building from HEAD. In case someone wants to actually build from HEAD **and** have draft API availalbe, they'll just have to pass `--HEAD --with-drafts`.

More information about the [C4.2 process](https://rfc.zeromq.org/spec:42/C4/).

ilovezfs:
@paddor I removed that because it's already the default for HEAD. It checks if there's a .git dir, and if there is, then it enables drafts automatically.

paddor:
@ilovezfs I didn't know about this. Is it a feature of Zproject? Can you link me to it?

ilovezfs:
@paddor 

https://github.com/zeromq/libzmq/blob/master/configure.ac#L626-L628
https://github.com/zeromq/czmq/blob/master/configure.ac#L405-L407

ilovezfs:
@paddor and yes ultimately from zproject as you said:
https://github.com/zeromq/zproject/blob/be5ae91c90de627c056e5d1a1b0dff810408c1c5/zproject_autotools.gsl#L496-L498

paddor:
@ilovezfs Thanks for pointing that out. Cool feature! I hope the bottles are built from releases (not within git repositories).

ilovezfs:
@paddor yeah it's using the upstream release tarball which doesn't have a .git directory. If we ever have to build directly from the tag, then we'll have to be careful.

ilovezfs:
:green_apple: shipped!

ilovezfs:
@paddor Thanks for your contribution to Homebrew! :tada:

PR #6645, opened 2016-11-06
swift 3.0.1

valkjsaaa:

-----

updated swift version to 3.0.1

### DIFF
--- a/Formula/swift.rb
+++ b/Formula/swift.rb
@@ -3,23 +3,23 @@ class Swift < Formula
   homepage "https://github.com/apple/swift"
 
   stable do
-    url "https://github.com/apple/swift/archive/swift-2.2.1-RELEASE.tar.gz"
-    sha256 "e971e2287055da72564356f369bad97e95821afb1ef36157e954a04a7e90753a"
+    url "https://github.com/apple/swift/archive/swift-3.0.1-RELEASE.tar.gz"
+    sha256 "5770cc45fe352d8a96d35ee35222cba0a47883b057e012a5464d411898afa074"
 
     swift_tag = "swift-#{version}-RELEASE"
     resource "cmark" do
       url "https://github.com/apple/swift-cmark/archive/#{swift_tag}.tar.gz"
-      sha256 "254d3c02bf2b03ad456fa3ad27b4da854e36318fcaf6b6f199fdb3e978a90803"
+      sha256 "07c8d491735cc0bef3a7f8796f9eff5e52011b85e025084721fbc896a469cddb"
     end
 
     resource "clang" do
       url "https://github.com/apple/swift-clang/archive/#{swift_tag}.tar.gz"
-      sha256 "40bdfa7eec0497ec69005d6a5d018b12c85aa2c0959d3408ecaaa9e34ff0415f"
+      sha256 "470e3c735a675aa37f58e4fe206a2d4141478a318dcb78c3fc8121e087e782b4"
     end
 
     resource "llvm" do
       url "https://github.com/apple/swift-llvm/archive/#{swift_tag}.tar.gz"
-      sha256 "f7977e5bb275494b5dac4490afc5d634f894ba5f209f3b2dbd5b7e520fa5fce2"
+      sha256 "5cfaa08743b29e6c8a948654b71fef608a48953b3eb928c3049c07ca279275c9"
     end
   end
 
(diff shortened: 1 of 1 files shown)


valkjsaaa:
Well, it seems that Swift requires Xcode 8 which macOS / OS X Yosemite does not support.

valkjsaaa:
This is weird. What does this supposed to mean?

Result of `brew audit swift` on macOS / OS X El Capitan

```
        Error: 1 problem in 1 formula
swift:
    Broken dylib links found:
      /System/Library/Frameworks/Intents.framework/Versions/A/Intents
```

ilovezfs:
It means the file doesn't exist, which on El Capitan is true.

ilovezfs:
@zmwangx looks like the failure has shifted from being fails-on-Sierra to being fails-on-not-Sierra lol

valkjsaaa:
@ilovezfs So are there anything wrong within this formula? Or this is a homebrew audit bug?

ilovezfs:
The formula is broken. The final binaries should not be linked against non-existent files.

valkjsaaa:
Could you give my some suggestion about how to fix that? This seems like a upstream problem to me.

ilovezfs:
This is actually probably a new `brew audit` problem.

```
otool -l /usr/local/Cellar/swift/3.0.1/lib/swift/macosx/libswiftIntents.dylib
```
shows

```
Load command 11
          cmd LC_LOAD_WEAK_DYLIB
      cmdsize 88
         name /System/Library/Frameworks/Intents.framework/Versions/A/Intents (offset 24)
   time stamp 2 Wed Dec 31 16:00:02 1969
      current version 1.0.0
```

and my guess is we shouldn't be failing something for a weak framework link against a file that doesn't exist when the build system explicitly requested `-weak_framework` due to that possibility. CC @UniqMartin

An unrelated thing that is actually a bit concerning though is the warnings
```
ld: warning: object file (/tmp/swift-20161106-41114-les5db/build/llvm-macosx-x86_64/./lib/libclangCodeGen.a(CGBuiltin.cpp.o)) was built for newer OSX version (10.11) than being linked (10.9)
```
since the configure arguments explicitly request 10.11, I'm not sure why it's only half honoring that. That seems like either it's an upstream build system bug, or we're missing a second option that should also be set to 10.11, or both. I'm tempted to just do string replacements of the offending 10.9 in the CMakeLists.txt, but it would be better to figure out why that's happening.

MikeMcQuaid:
> and my guess is we shouldn't be failing something for a weak framework link against a file that doesn't exist when the build system explicitly requested -weak_framework due to that possibility. CC @UniqMartin

CC @woodruffw in case you have thoughts, too.

woodruffw:
> and my guess is we shouldn't be failing something for a weak framework link against a file that doesn't exist when the build system explicitly requested -weak_framework due to that possibility.

This sounds correct to me. We currently enumerate linked dylibs without any consideration for what *type* they are, via `MachOFile#linked_dylibs` in ruby-macho. There's a more flexibility way to do this (`MachOFile#dylib_load_commands` + filtering), which I could expose like this (in Homebrew):

```ruby
file.dynamically_linked_libraries(except: :weak) do |dylib|
 # normal magic here
end
```

That would give us the behavior we desire in places like `LinkageChecker#check_dylibs`, which is what's currently causing the error.

Thoughts?

MikeMcQuaid:
@woodruffw Sounds like the right approach 👍

MikeMcQuaid:
Pulling this as it's an `audit` bug.

woodruffw:
@MikeMcQuaid Great! I'll have a PR for the correct behavior today.

ilovezfs:
>Pulling this as it's an audit bug.

It looks like this PR unintentionally led to dropping Yosemite support. Only a 10.11 bottle and a 10.12 bottle were pulled, and neither 3.0.1 nor 3.0.2 are building on 10.10.


3.0.1: https://bot.brew.sh/job/Homebrew%20Testing/1181/version=yosemite/console
3.0.2: https://bot.brew.sh/job/Homebrew%20Core/13231/version=yosemite/console

MikeMcQuaid:
Yes, sorry. It's likely this is an Xcode/Clang version dependency rather than an OS one so 👍 to restricting this formula to that.

ilovezfs:
Assumed Apple intends Swift 3 to have some level of Yosemite compatibility (not sure one way or another yet), I think this may very well be related to what I had mentioned above: https://github.com/Homebrew/homebrew-core/pull/6645#issuecomment-258668325

Specifically, that I don't think we're actually setting SDK to the current macOS version of the build machine in the correct way for their build system.

MikeMcQuaid:
> Specifically, I don't think we're actually setting SDK to the current macOS version of the build machine in the correct way for their build system.

Sounds likely 👍

PR #6656, opened 2016-11-06
liblastfm: use 10.11 bottle on 10.12.

MikeMcQuaid:
Needs CI rerun after https://github.com/Homebrew/homebrew-test-bot/pull/20 for proper testing.

CC @ilovezfs for thoughts on the approach.


MikeMcQuaid:
This needs Qt so isn't going to work on 10.12.

PR #6681, opened 2016-11-07
Update xhyve formula to include test resources, examples, and manpages

zbentley:

-----

These files are important resources referred to by many tutorials that use `xhyve`, but are not installed with it. This causes bad directions. For example, [this tutorial](https://gist.github.com/tanb/f8fefa22332edc7a641d), which has been forked/linked all over the place, refers to the Homebrew cache directory for those files. Not only has the cache directory moved in the past, but it also may not exist due to `brew cleanup`. As a result of those bad directions, there's confusion: for example, [this SO question](http://apple.stackexchange.com/questions/242305/install-freebsd-on-xhyve).

All of the added files are only installed when building from source. I've opened mist64/xhyve#120 to get these files added to future release tarballs.

The if/else conditional is because the manpage is only available at `HEAD` currently. It ought to be included in the next tagged release.

### DIFF
--- a/Formula/xhyve.rb
+++ b/Formula/xhyve.rb
@@ -19,11 +19,9 @@ class Xhyve < Formula
     args << "GIT_VERSION=#{version}" if build.stable?
     system "make", *args
     bin.install "build/xhyve"
-    prefix.install "test/"
-    prefix.install "xhyverun.sh"
-    if File.exist?("xhyve.1")
-      man1.install "xhyve.1"
-    end
+    pkgshare.install "test/"
+    pkgshare.install "xhyverun.sh"
+    man1.install "xhyve.1" if build.head?
   end
 
   test do
--- a/Formula/xhyve.rb
+++ b/Formula/xhyve.rb
@@ -19,6 +19,11 @@ class Xhyve < Formula
     args << "GIT_VERSION=#{version}" if build.stable?
     system "make", *args
     bin.install "build/xhyve"
+    prefix.install "test/"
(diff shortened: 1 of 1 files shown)


Formula/xhyve.rb
```diff
@@ -19,6 +19,11 @@ def install
     args << "GIT_VERSION=#{version}" if build.stable?
     system "make", *args
     bin.install "build/xhyve"
+    prefix.install "test/"
+    prefix.install "xhyverun.sh"
+    if File.exist?("xhyve.1")
```
MikeMcQuaid: `man1.install "xhyve.1" if build.head?`

MikeMcQuaid (changes requested):

Formula/xhyve.rb
```diff
@@ -19,6 +19,11 @@ def install
     args << "GIT_VERSION=#{version}" if build.stable?
     system "make", *args
     bin.install "build/xhyve"
+    prefix.install "test/"
+    prefix.install "xhyverun.sh"
```
MikeMcQuaid: Maybe `pkgshare.install` may make more sense for these?

MikeMcQuaid (approved):

MikeMcQuaid:

For future reference the preferred commit message format is `xhyve: include test, examples and manpages.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #6685, opened 2016-11-07
plplot: depend on cairo

ilovezfs:
pango already depends on cairo, and it's required for the extcairo
support that the test assumes is present

### DIFF
--- a/Formula/plplot.rb
+++ b/Formula/plplot.rb
@@ -14,6 +14,7 @@ class Plplot < Formula
 
   depends_on "cmake" => :build
   depends_on "pkg-config" => :build
+  depends_on "cairo"
   depends_on "pango"
   depends_on "freetype"
   depends_on "libtool" => :run


tschoonj:
Was this test failing without the explicit cairo dependency?

ilovezfs:
@tschoonj https://bot.brew.sh/job/Homebrew%20Core/10851/version=any_sierra/console

The error is "Requested device extcairo not available" if it's not built with the cairo dependency specified. This is probably a regression since presumably at one point the test passed.

tschoonj:
Actually there may also something else going on here. Before I patched the PLplot formula, for some time the cairo device would not get build due to the build script not being compatible with CMake 3.6.x.

I think the Sierra bottle was built before I added that patch which would explain your test failing, and why the `extcairo` doesn't show up in the list in the test output:

````
Plotting Options:
 < 1> ps         PostScript File (monochrome)
 < 2> psc        PostScript File (color)
 < 3> xfig       Fig file
 < 4> null       Null device
 < 5> mem        User-supplied memory device
 < 6> svg        Scalable Vector Graphics (SVG 1.1)
````

ilovezfs:
Yup, you're right. The final product is linked to cairo so it should be fine either way.

PR #6689, opened 2016-11-07
glib 2.50.2

tschoonj:


ilovezfs:
@tschoonj the `mimms` problem is fixed by converting the formula to use `virtualenv_install_with_resources` so you don't need to worry about it. That can be handled in a separate PR.

ilovezfs (approved):

tschoonj:
Thanks!

PR #6723, opened 2016-11-08
sdl_image: Disable ImageIO

tsparber:
-> No as there are no tests.

-----

This disables the usage of the macOS ImageIO framework to load image files instead libpng, jpeg, libtiff are used.
This fixes issues when loading for example 1-bit png files (they are  loaded as RGB instead of color indexed as they are using `libpng`) as well as other issues where pixel colors are not exactly loaded.

Currently there are no available fixes/patches for theses issues (neither for SDL2 or SDL1).

This is also needed to pass some more [pygame tests](https://bitbucket.org/pygame/pygame/src/aa82244e06e5916f77cc585c6d421354b171d6e0/.travis_osx_before_install.sh?at=default&fileviewer=file-view-default#.travis_osx_before_install.sh-47) (It also requires a `sdl_image` version with ImageIO disabled).
See also https://github.com/Homebrew/homebrew-core/pull/739.

As ImageIO is also disabled in [`sdl2_image`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/sdl2_image.rb#L31) im suggesting to always use this configuration option instead of adding another build `option`.

The [patch](https://github.com/homebrew/homebrew-core/blob/master/Formula/sdl_image.rb#L25) from https://github.com/Homebrew/homebrew-core/pull/1864 could also be removed, as it is only needed when building with ImageIO enabled.

CC @illume

### DIFF
--- a/Formula/sdl_image.rb
+++ b/Formula/sdl_image.rb
@@ -3,7 +3,7 @@ class SdlImage < Formula
   homepage "https://www.libsdl.org/projects/SDL_image"
   url "https://www.libsdl.org/projects/SDL_image/release/SDL_image-1.2.12.tar.gz"
   sha256 "0b90722984561004de84847744d566809dbb9daf732a9e503b91a1b5a84e5699"
-  revision 3
+  revision 4
 
   bottle do
     cellar :any
@@ -36,6 +36,7 @@ class SdlImage < Formula
 
     system "./configure", "--prefix=#{prefix}",
                           "--disable-dependency-tracking",
+                          "--disable-imageio",
                           "--disable-sdltest"
     system "make", "install"
   end


MikeMcQuaid:
> The patch from #1864 could also be removed, as it is only needed when building with ImageIO enabled.

Can you do this too, thanks.

Can you link to any upstream discussions about this? I'm a bit concerned about removing a default option like this, thanks.

tsparber:
> Can you link to any upstream discussions about this? I'm a bit concerned about removing a default option like this, thanks.

One upstream discussion about this [Pixel bug in Mac OS X](https://forums.libsdl.org/viewtopic.php?p=30317&sid=088b00955c814f8a6726db49ba7e001a#30317):
> We should just disable this, because it constantly causes three problems: 
> - Slightly different colors when you need pixel-perfect decoding. 
> - No 8-bit support (which is bad when you use an 8-bit grayscale .png 
> as, say, a heightmap, and actually wanted an 8-bit surface from SDL_image). 
> - The first two problems are completely non-obvious to track down when 
> you realize you have a problem, and a pain to resolve, because now you 
> have to build your own SDL_image at a minimum. 
> 
> The benefits to using ImageIO are a smaller compiled library, no need 
> for us to manage the external dependencies, and no need to worry about 
> updating the dependencies for security issues. I'm not sure this is the 
> _worst_ thing, though.
> ...

(From Ryan C. Gordon, SDL developer)

As far as my tests go, these are the exact problems I'm having/seeing which also affect others. (And ImageIO is disabled in the SDL2 formula).

Should I add an option to still use ImageIO? (I would still prefer defaulting to disabling it).

MikeMcQuaid:
You've convinced me, thanks for the explanation!

PR #6725, opened 2016-11-08
luaver: Improve description

telemachus:

-----

I suppose that many devs will already know what a "version manager" is,
but I still think it's better to give an informative description rather
than just the name of the software.

A larger question: Would a caveat be in order saying something like
"This package doesn't literally conflict with homebrew installed
versions of Lua, LuaJit, or LuaRocks, but it does not work with them
either"? That is, I worry that some (less experienced?) devs might
assume that *this* tool and homebrew's Lua items work together.

### DIFF
--- a/Formula/luaver.rb
+++ b/Formula/luaver.rb
@@ -1,5 +1,5 @@
 class Luaver < Formula
-  desc "Lua Version Manager"
+  desc "Manage and switch between versions of Lua, LuaJIT, and Luarocks"
   homepage "https://github.com/DhavalKapil/luaver"
   head "https://github.com/DhavalKapil/luaver.git"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> Would a caveat be in order saying something like
> "This package doesn't literally conflict with homebrew installed
> versions of Lua, LuaJit, or LuaRocks, but it does not work with them
> either"? That is, I worry that some (less experienced?) devs might
> assume that this tool and homebrew's Lua items work together.

Nah, I think we're fine as-is unless we get multiple reports that would have been addressed by them.

MikeMcQuaid:
Thanks again @telemachus!

PR #6736, opened 2016-11-08
[xctool] Codesign all xctool libraries

ExtremeMan:
macOS and/or Xcode at some point started requiring to sign libraries that are injected into xcodebuild and Simulator.app processes. I think we were seeing numerous reports in the past (but really vary rare) when injecting the library was failing and as a result xctool failed as a whole. Since it is very difficult to debug we have never really fixed it.

Based on latest reports from xctool users (https://github.com/facebook/xctool/issues/721) and googling I have figured that we can just codesign all xctool libraries to fix mentioned issues after install is complete.

Epic fail but I hit the same issue with my homebrew installation of xctool (yes, I need to update formula test plan). But it made it much easier to debug and fix.

### DIFF
--- a/Formula/xctool.rb
+++ b/Formula/xctool.rb
@@ -19,6 +19,14 @@ class Xctool < Formula
     bin.install_symlink "#{libexec}/bin/xctool"
   end
 
+  def post_install
+    # all libraries need to be signed to avoid codesign errors when
+    # injecting them into xcodebuild or Simulator.app.
+    Dir.glob("#{libexec}/lib/*.dylib") do |lib_file|
+      system "/usr/bin/codesign", "-f", "-s", "-", lib_file
+    end
+  end
+
   test do
     system "(#{bin}/xctool -help; true)"
   end


MikeMcQuaid:
This works but I'm surprised it did 😁. Does this need local certificates/permissions or anything on the system? Why do we have to do it on post-install?

ExtremeMan:
From codesign man:

> If identity is the single letter "-" (dash), ad-hoc signing is performed.  Ad-hoc signing does not use an identity at all, and identifies exactly one instance of code. Significant restrictions apply to the use of ad-hoc signed code; consult documentation before using this.

> Does this need local certificates/permissions or anything on the system?

So I guess no.

> Why do we have to do it on post-install?

From I have found on the internet it was recommended to first move binaries that need to be signed to a final destination and only then inside. FYI: these libraries are signed the same way when we prepare new xctool release. For some reason it doesn't work once installed (not sure how broad the problem is though).

MikeMcQuaid:
@ExtremeMan Can we try adding this in the `install` block and seeing what happens? Our tests will likely 💥 if this is going to fail for users.

ExtremeMan:
I tried that locally and it didn't work. This is the ouput:

```
** BUILD SUCCEEDED **

==> /usr/bin/codesign -f -s - /usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-ios.dylib
/usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-ios.dylib: replacing existing signature
==> /usr/bin/codesign -f -s - /usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-osx.dylib
/usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-osx.dylib: replacing existing signature
==> /usr/bin/codesign -f -s - /usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-ios.dylib
/usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-ios.dylib: replacing existing signature
==> /usr/bin/codesign -f -s - /usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-osx.dylib
/usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-osx.dylib: replacing existing signature
==> /usr/bin/codesign -f -s - /usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-fastsettings-shim.dylib
/usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-fastsettings-shim.dylib: replacing existing signature
==> /usr/bin/codesign -f -s - /usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-shim.dylib
/usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-shim.dylib: replacing existing signature
==> Cleaning
==> Finishing up
ln -s ../Cellar/xctool/0.3/bin/xctool xctool
Changing dylib ID of /usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-ios.dylib
  from /usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-ios.dylib
    to /usr/local/opt/xctool/libexec/lib/otest-query-lib-ios.dylib
Changing dylib ID of /usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-osx.dylib
  from /usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-osx.dylib
    to /usr/local/opt/xctool/libexec/lib/otest-query-lib-osx.dylib
Changing dylib ID of /usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-ios.dylib
  from /usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-ios.dylib
    to /usr/local/opt/xctool/libexec/lib/otest-shim-ios.dylib
Changing dylib ID of /usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-osx.dylib
  from /usr/local/Cellar/xctool/0.3/libexec/lib/otest-shim-osx.dylib
    to /usr/local/opt/xctool/libexec/lib/otest-shim-osx.dylib
Changing dylib ID of /usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-fastsettings-shim.dylib
  from /usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-fastsettings-shim.dylib
    to /usr/local/opt/xctool/libexec/lib/xcodebuild-fastsettings-shim.dylib
Changing dylib ID of /usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-shim.dylib
  from /usr/local/Cellar/xctool/0.3/libexec/lib/xcodebuild-shim.dylib
    to /usr/local/opt/xctool/libexec/lib/xcodebuild-shim.dylib
==> Summary
```

Most likely `Changing dylib ID of ` corrupts the signature:

```
codesign --verify -vvvv /usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-ios.dylib
/usr/local/Cellar/xctool/0.3/libexec/lib/otest-query-lib-ios.dylib: invalid signature (code or signature have been modified)
In architecture: x86_64
```

MikeMcQuaid:
Thanks for trying that out.

MikeMcQuaid (approved):

MikeMcQuaid:

For future reference the preferred commit message format is `xctool: codesign all libraries.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

Issue #6408, opened 2016-10-30
Formulae without Sierra bottles


ilovezfs:
Initially copied verbatim from https://github.com/Homebrew/homebrew-core/issues/5488. --ilovezfs (29 Oct 2016)

Jenkins finally managed to chew through the daunting Sierra build queue. Below are the remaining ones that defied (not necessarily only) Sierra for one reason or another.

The list is generated with the script https://gist.github.com/zmwangx/89f07670c8e42f58434bd6d15653ecf6. You can also pass a blacklist to the script to get a more workable list (e.g., putting `qt` and `qt5` in the blacklist blocks the entire Qt tree). See `need-bottle -h`.

---
\* means the formula now has an `:el_capitan_or_later` bottle but not a `:sierra` bottle yet.

Formulae that made the top 5000 install events chart during September 2016, ordered by popularity:
- [x] fb-adb (removed in https://github.com/Homebrew/homebrew-core/commit/9a4a8e63a667152196de72dbf4f3b97fe630b3f0)
- [x] ctorrent (redefinition of `clockid_t`)
- [x] open-zwave (https://github.com/Homebrew/brew/pull/1801)
- [x] libgroove (redefinition of `clockid_t` and `clock_gettime`; also fails on El Capitan)
- [x] vice (#8621)
- [x] libfaketime (https://github.com/Homebrew/homebrew-core/pull/18249)
- [ ] dshb (https://github.com/beltex/dshb/issues/36)
- [ ] libbladerf (redefinition of `clockid_t`)
- [ ] darkstat (redefinition of `clockid_t`)
- [x] valgrind (https://github.com/Homebrew/homebrew-core/pull/14653)

---
Obscure formulae that didn't make the top 5000 install events chart (some of these need to be thrown into boneyard — no mercy for the weak):
- [x] aiccu
- [x] blucat
- [x] buildapp
- [x] compface
- [x] dsd (#8590)
- [x] intercal (#8591)
- [x] jxrlib (#8608)
- [x] libodfgen
- [x] mimms (#8594)
- [x] nordugrid-arc
- [x] parrot (https://github.com/Homebrew/brew/pull/1801)
- [x] qpm (#8593)
- [x] qtplay (#8614)
- [x] rdate (#8611)
[...]

PR #6740, opened 2016-11-08
liblastfm: qt5 support and add a test

Moisan:

-----

Related to #1705 and #6408.

### DIFF
--- a/Formula/liblastfm.rb
+++ b/Formula/liblastfm.rb
@@ -31,7 +31,6 @@ class Liblastfm < Formula
 
   test do
     cp_r "#{share}/tests/.", testpath
-    assert_match "Totals: 3 passed, 0 failed, 0 skipped, 0 blacklisted, 0ms",
-      shell_output("./TrackTest")
+    system "./TrackTest"
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/liblastfm.rb
```diff
@@ -13,14 +14,24 @@ class Liblastfm < Formula
...
+      cd "tests" do
+        system "make"
+      end
+      share.install "tests"
     end
   end
+
+  test do
+    cp_r "#{share}/tests/.", testpath
+    assert_match "Totals: 3 passed, 0 failed, 0 skipped, 0 blacklisted, 0ms",
```
MikeMcQuaid: `0ms` needs removed here as it's failing on Sierra (and generally is likely to be flaky). Is it not enough to rely on the exit-code alone here?

MikeMcQuaid (changes requested):

Formula/liblastfm.rb
Moisan: You're right, exit-code should be enough.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #6742, opened 2016-11-09
wine: enable 64-bit WoW support

uyjulian:

-----

### DIFF
--- a/Formula/wine.rb
+++ b/Formula/wine.rb
@@ -38,6 +38,11 @@ class Wine < Formula
     MacOS.prefer_64_bit?
   end
 
+  if MacOS.version >= :el_capitan
+    option "without-win64", "Build without 64-bit support"
+    depends_on :xcode => ["8.0", :build] if build.with? "win64"
+  end
+
   # Wine will build both the Mac and the X11 driver by default, and you can switch
   # between them. But if you really want to build without X11, you can.
   depends_on :x11 => :recommended
@@ -58,12 +63,6 @@ class Wine < Formula
     url "https://bugs.winehq.org/attachment.cgi?id=52384"
     sha256 "30766403f5064a115f61de8cacba1defddffe2dd898b59557956400470adc699"
   end
-
-  # This option is currently disabled because Apple clang currently doesn't
-  # support a required feature: https://reviews.llvm.org/D1623
-  # It builds fine with GCC, however.
-  # option "with-win64",
-  #        "Build with win64 emulator (won't run 32-bit binaries.)"
 
   resource "gecko" do
     url "https://downloads.sourceforge.net/wine/wine_gecko-2.40-x86.msi", :using => :nounzip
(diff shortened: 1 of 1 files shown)


Formula/wine.rb
```diff
@@ -38,6 +38,9 @@ def require_universal_deps?
     MacOS.prefer_64_bit?
   end
 
+  option "with-win64",
+         "Build with 64-bit support (only works with Xcode 8 and newer.)"
```
MikeMcQuaid: Probably worth keeping this on the same line. Make `only works with Xcode 8 and newer.` a `depends_on :xcode => ["8.0", :build] if build.with? "win64"` instead. Add `(won't run 32-bit binaries.)`

Formula/wine.rb
```diff
@@ -88,68 +85,43 @@ def require_universal_deps?
...
   def install
-    ENV.m32 # Build 32-bit; Wine doesn't support 64-bit host builds on macOS.
-
     # Help configure find libxml2 in an XCode only (no CLT) installation.
     ENV.libxml2
 
-    args = ["--prefix=#{prefix}"]
-    args << "--disable-win16" if MacOS.version <= :leopard
-    args << "--enable-win64" if build.with? "win64"
+    args32 = ["--prefix=#{prefix}"]
```
MikeMcQuaid: Instead leave this as `args` and build it depending on whether it's 32/64-bit.

Formula/wine.rb
```diff
@@ -88,68 +85,43 @@ def require_universal_deps?
...
+
+    args64 << "--without-x" if build.without? "x11"
+
+    if build.with? "win64"
+      mkdir "wine-64-build"
+      cd "wine-64-build"
+
+      system "../configure", *args64
+
+      system "make", "install"
```
MikeMcQuaid: Will none of these files get overwritten by the 32-bit build?

MikeMcQuaid (changes requested):

Formula/wine.rb
uyjulian: No, wine and wine64 can co-exist.
wine will use 64-bit tools (for example wineserver, winedbg etc)
The order of `make install` does not matter.

Formula/wine.rb
MikeMcQuaid: Or have `args` which you then turn into `args32` and `args64` later

Formula/wine.rb
MikeMcQuaid: Ah, I see. Can omit `(won't run 32-bit binaries.)` 👍

uyjulian: Actually, it will work with 32-bit binaries. `wine` and `wine64` will run the appropriate program required for the executable (if 16-bit or 32-bit, `wine` is run, otherwise, if it is 64-bit, `wine64` is run.)

Formula/wine.rb
MikeMcQuaid: Nice. I think it'd read a bit clearer to build `args64` inside the `if build.with? "win64"` and build `args32` after this. If you use the `mkdir do` block then you don't need the manual `cd`s.

MikeMcQuaid (changes requested):

MikeMcQuaid (approved):

dunn:
@BrewTestBot test this please

MikeMcQuaid:
Crazy idea: would there be any downside of enabling 64-bit support by default if they both happily live side-by-side?

uyjulian:
@MikeMcQuaid Some applications/dlls may not work correctly in a WoW prefix  
That's okay, just use `WINEARCH=win32` when creating a prefix

MikeMcQuaid:
@uyjulian Any way to make the prefix default to 32-bit somehow?

uyjulian:
@MikeMcQuaid No, but the problem is rare. It is only needed in some cases (for example when you install native .NET 2.0 with winetricks).

In that case you can have a separate 32-bit and 64-bit wineprefix (you don't need two separate versions of wine compiled) by setting `WINEPREFIX` environment variable.

MikeMcQuaid:
Ok, let's try this with 64-bit enabled by default 👍

kaazoo:
+1

Formula/wine.rb
```diff
@@ -38,8 +38,11 @@ def require_universal_deps?
     MacOS.prefer_64_bit?
   end
 
+  option "without-win64", "Build without 64-bit support"
+
   # Wine will build both the Mac and the X11 driver by default, and you can switch
   # between them. But if you really want to build without X11, you can.
+  depends_on :xcode => ["8.0", :build] if build.with? "win64"
```
MikeMcQuaid: Last tweak: the comment above refers to `:x11` below so can you move this new line below `depends_on "pkg-config" => :build`. Post a comment when you've done that and CI is 💚 and I'll make sure to merge ASAP. Thanks!

MikeMcQuaid (changes requested):

uyjulian:
CI looks good to me

mischnic:
+1

MikeMcQuaid (approved):

Formula/wine.rb
```diff
@@ -38,10 +38,13 @@ def require_universal_deps?
     MacOS.prefer_64_bit?
   end
 
+  option "without-win64", "Build without 64-bit support"
+
   # Wine will build both the Mac and the X11 driver by default, and you can switch
   # between them. But if you really want to build without X11, you can.
   depends_on :x11 => :recommended
   depends_on "pkg-config" => :build
+  depends_on :xcode => ["8.0", :build] if build.with? "win64"
```
MikeMcQuaid: Whoops, just noticed another problem here: if we build with `win64` by default this will unfortunately mean that by default this fails on Yosemite and below.

This can probably be fixed by changing the above to:
```ruby
if MacOS.version >= :el_capitan
  option "without-win64", "Build without 64-bit support"
  depends_on :xcode => ["8.0", :build] if build.with? "win64"
end
```
so it refuses to try and build Win64 on older versions of macOS (with older versions of Xcode). Sorry I didn't spot this sooner.

MikeMcQuaid (changes requested):

uyjulian:
@MikeMcQuaid I added your fix and pushed it.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again for your contribution to Homebrew, @uyjulian, and your patience! Without people like you submitting PRs we couldn't run this project. You rock!

PR #6764, opened 2016-11-09
mercurial: --with-custom-python can be a no-op

tdsmith:
... without changing its behavior.

### DIFF
--- a/Formula/mercurial.rb
+++ b/Formula/mercurial.rb
@@ -14,11 +14,7 @@ class Mercurial < Formula
   end
 
   option "with-custom-python", "Install against the python in PATH instead of Homebrew's python"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍 but I'm wondering why we want to allow people to use a system Python for an application?

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

fxcoudert:
@BrewTestBot test this please

fxcoudert:
@MikeMcQuaid does this require bottle update? (and thus a revision bump)

ilovezfs:
No. :python resolves to "python" when bottling

PR #6770, opened 2016-11-10
Rebuild Citus 6.0

jasonmp85:

-----

The tag contents changed, so the sha changed. I believe I'm doing this correctly, just checking.

### DIFF
--- a/Formula/citus.rb
+++ b/Formula/citus.rb
@@ -2,8 +2,9 @@ class Citus < Formula
   desc "PostgreSQL-based distributed RDBMS"
   homepage "https://www.citusdata.com"
   url "https://github.com/citusdata/citus/archive/v6.0.0.tar.gz"
-  sha256 "a572d12e7f62b11512bd4b3a716cd56eda74153292adaa2340e2c1d363df13ab"
+  sha256 "90bad0e57256ca3e72c080b58877e396216cbd46c5820a8f0ec2a0ca4d062dfb"
   head "https://github.com/citusdata/citus.git"
+  revision 1
 
   bottle do
     cellar :any


MikeMcQuaid:
@jasonmp85 Can you file an issue upstream asking them why the tag changed and if they've been hacked? We need to nicely ask them to not change tags in future.

jasonmp85:
> Can you file an issue upstream asking them why the tag changed and if they've been hacked? We need to nicely ask them to not change tags in future.

In short, no hacking. In long:

So… I _am_ upstream, and _I_ changed the tag because it was less than a day old and I made a mistake in forgetting to cherry-pick something to our release branch.

I know, it's bad; I feel bad. Just didn't want to have to bump to `6.0.1` when we hadn't even built our own packages. Forgot I had already opened the homebrew pull request. Gah.

At any rate, the tag is signed with my GPG key; I'm jasonmp85 on [keybase](https://keybase.io/jasonmp85), and I've signed every single tag we've ever released, as well as every commit in all projects I am a part of. So there's that as some proof.

In addition, we sign our tarballs with a [company-wide packaging key](https://install.citusdata.com/packaging.key). The [release](https://github.com/citusdata/citus/releases/tag/v6.0.0) page has this signature for download (and indeed it is automatically downloaded and verified by the main Debian packaging system that pulls from us).

ilovezfs:
>I know, it's bad; I feel bad.

It's OK. Don't feel bad. Thanks for clarifying.

ilovezfs:
Shipped. @jasonmp85 thanks for updating the formula after the tag changed.

MikeMcQuaid:
> So… I am upstream, and I changed the tag because it was less than a day old and I made a mistake in forgetting to cherry-pick something to our release branch.

For future reference: anyone who pulled your Git repository's tags will still be on the old tags because Git doesn't update tags by default (for hopefully fairly obvious security reasons).

> I know, it's bad; I feel bad. Just didn't want to have to bump to 6.0.1 when we hadn't even built our own packages. Forgot I had already opened the homebrew pull request. Gah.

I understand this is annoying but I'd like to ask in future you consider bumping to 6.0.1 (or 6.0.000000001 or whatever) as for both Git cloners and packagers.

jasonmp85:
> For future reference: anyone who pulled your Git repository's tags will still be on the old tags because Git doesn't update tags by default (for hopefully fairly obvious security reasons).

Oh, I know!

> I understand this is annoying but I'd like to ask in future you consider bumping to 6.0.1 (or 6.0.000000001 or whatever) as for both Git cloners and packagers.

I certainly considered it. Next time I'll try to avoid tagging until absolutely sure the branch is ready. It's not quite that _I_ don't understand the severity; sometimes it's a little hard to adequately explain to non-tech people why the release needs to be bumped when we haven't even yet done a formal 6.0 announcement and I could "just delete it". Sorry about all this, and thanks for getting this merged!

PR #6790, opened 2016-11-10
openssl@1.1 1.1.0c

DomT4:

-----

### DIFF
--- a/Formula/nginx.rb
+++ b/Formula/nginx.rb
@@ -29,8 +29,15 @@ class Nginx < Formula
   deprecated_option "with-spdy" => "with-http2"
 
   depends_on "pcre"
-  depends_on "openssl@1.1"
   depends_on "passenger" => :optional
+
+  # passenger uses apr, which uses openssl, so need to keep
+  # crypto library choice consistent throughout the tree.
+  if build.with? "passenger"
+    depends_on "openssl"
+  else
+    depends_on "openssl@1.1"
+  end
 
   def install
     # Changes default port to 8080
@@ -40,7 +47,12 @@ class Nginx < Formula
     end
 
     pcre = Formula["pcre"]
-    openssl = Formula["openssl@1.1"]
+
+    if build.with? "passenger"
+      openssl = Formula["openssl"]
+    else
+      openssl = Formula["openssl@1.1"]
+    end
 
     cc_opt = "-I#{pcre.opt_include} -I#{openssl.opt_include}"
     ld_opt = "-L#{pcre.opt_lib} -L#{openssl.opt_lib}"
--- a/Formula/openssl@1.1.rb
+++ b/Formula/openssl@1.1.rb
@@ -1,9 +1,9 @@
 class OpensslAT11 < Formula
   desc "Cryptography and SSL/TLS Toolkit"
   homepage "https://openssl.org/"
-  url "https://www.openssl.org/source/openssl-1.1.0b.tar.gz"
-  mirror "https://www.mirrorservice.org/sites/ftp.openssl.org/source/openssl-1.1.0b.tar.gz"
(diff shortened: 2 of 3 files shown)


ilovezfs:
10.12
```
==> brew linkage --test passenger
==> FAILED
Missing libraries:
  /usr/local/opt/apr-util/libexec/lib/libaprutil-1.0.dylib
  /usr/local/opt/apr/libexec/lib/libapr-1.0.dylib
```
:(

DomT4:
Oh, of course. Homebrew Apr's aren't `build` only.

DomT4:
Should be fixed whenever CI picks the new push up.

ilovezfs:
Dropping 1.1 support from passenger?

ilovezfs:
@DomT4 if we back out openssl 1.1 from passenger (which I don't necessarily oppose), then I think we also need to back it out from nginx. The reason we added openssl 1.1 to passenger was that a PR adding openssl 1.1 to nginx was merged, and then I noticed that nginx has a `--with-passenger` option, and that passenger also depends on openssl. So to avoid mixing and matching openssl versions, we also added 1.1 to passenger, but it would have been equally valid to revert the nginx PR, which probably never should have been merged,

DomT4:
Will reply properly when at laptop & commit again, but Apr-util uses OpenSSL, and IIRC doesn't yet build with the 1.1 branch, so to avoid mixing crypto libraries if the Homebrew Apr is used the 1.0.2 OpenSSL needs to be used as well.

ilovezfs:
@DomT4 that's fine. We'll just need to back it out from nginx as well then, I think, or drop the passenger option from nginx.

DomT4:
Let me know how you feel about those changes.

ilovezfs:
I like it, given my current understanding of the constraints.

But Is it possible to build passenger without apr-util or to build apr-util without the openssl dep?

I'm a little concerned that the proliferation of the apr/apr-utils formula dependencies is going to interfere with the overall transition to openssl 1.1, not that I'm actually in any rush for us to transition to 1.1.

I noticed apr-util has been subsumed by apr in apr svn trunk. Do we know if apr svn trunk builds with openssl 1.1?

DomT4:
> But Is it possible to build passenger without apr-util 

Apparently not:
```
In file included from src/apache2_module/mod_passenger.c:26:
In file included from /usr/include/apache2/httpd.h:44:
In file included from /usr/include/apache2/ap_config.h:25:
/usr/include/apache2/ap_hooks.h:40:10: fatal error: 'apr_hooks.h' file not found
#include "apr_hooks.h"
         ^
1 error generated.
```

> build apr-util without the openssl dep

This is possible, yes, but I'm not sure it wouldn't cause issues with other formulae. I believe `subversion` relies on Apr's crypto elements for some aspects, for example, but don't quote me.

> I'm a little concerned that the proliferation of the apr/apr-utils formula dependencies is going to interfere with the overall transition to openssl 1.1, not that I'm actually in any rush for us to transition to 1.1.

Agreed. 1.0.2 is LTS, so there's no rush at all, but the cleaner codebase and feature set of 1.1.0 is certainly desirable. To have an important backbone of an increasing number of formulae unable to use 1.1.0 is a pain.

> I noticed apr-util has been subsumed by apr in apr svn trunk. Do we know if apr svn trunk builds with openssl 1.1?

Yup. Always thought them being separate was a bit odd, so glad to see them come together formally. Not sure about `HEAD`, I had a look and `configure` finds all the elements it specifically checks for, but `make` blows up for a seemingly unrelated reason:

```
==> make
Makefile:52: /private/tmp/apr-util-20161111-38356-qv8syg/build-outputs.mk: No such file or directory
make: *** No rule to make target `/private/tmp/apr-util-20161111-38356-qv8syg/build-outputs.mk'.  Stop.
```

ilovezfs:
Thanks, Dominyk.

DomT4:
Thanks ILZ ❤️

PR #6824, opened 2016-11-11
imagesnap: use 10.11 bottle on 10.12.

MikeMcQuaid:

-----

It's used so let's allow people to use the bottle on Sierra.

### DIFF
--- a/Formula/imagesnap.rb
+++ b/Formula/imagesnap.rb
@@ -7,7 +7,7 @@ class Imagesnap < Formula
   bottle do
     cellar :any_skip_relocation
     revision 1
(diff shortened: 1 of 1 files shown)


sjackman:
Hi, Mike. When do you use an `_or_later` tag as here, and when do you rebuild the bottle for the new OS?

MikeMcQuaid:
@sjackman We try to rebottle whenever possible and just use `_or_later` as a temporary workaround when it doesn't build on the latest OS.

sjackman:
Thanks for the explanation, Mike.

PR #6708, opened 2016-11-08
qt5: Add patch for QTBUG-51927

dagheyman:
- [X] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [X] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [X] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [X] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

Manually verified that the patched qt5 build can't reproduce the issue described [here](https://bugreports.qt.io/browse/QTBUG-56909) (duplicate of the same bug).

[Here](https://github.com/Homebrew/formula-patches/pull/84#) is the PR to add the patch in the patch repository.

Issue #6766, opened 2016-11-09
Qt v5.7.0 Still Fails to Build on OS X v10.11.6 'El Capitan' After Resolution of Issue #6032


RandomDSdevel:
### Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [ ] ~~If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?~~  N/A

### Bug reports:

The resolution of [Issue #6032:  'qt5 fails to build with mysql support'](https://github.com/Homebrew/homebrew-core/issues/6032) did not make it so that I am able to build Qt v5.7.0.  [Here](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/tree/master/Qt/v5.7.0/Third%20Recorded%20Attempt) are my logs.  Note that they again include one or more notices of a violation or more of Homebrew's build sandbox.  I also saw some warnings flash by during the build when I peeked at it, so those might be relevant as well, but I'm not sure.

RandomDSdevel:
@ilovezfs:  Could the issues I'm seeing here perhaps be related to how Homebrew Core had to add some patches originally used for Qt v5.6.1 (and _maybe_ v5.7.1-1…?) back for v5.7.0 of the software after they were removed from the edition of `qt5.rb` that propagated Qt v5.6.2 to Homebrew users?  Could you (and/or anybody else involved) have possibly forgotten to add a patch back, or am I just grasping at straws here?

MikeMcQuaid:
@RandomDSdevel I realise you may not have the patience but can you try to remove options one-by-one to figure out which one(s) are broken? I'm really tempted to remove options from the Qt5 formula (maybe all of them).

MikeMcQuaid:
For discussion: Qt5 installs in the last month with their options:

<img width="833" alt="screen shot 2016-11-10 at 10 33 25" src="https://cloud.githubusercontent.com/assets/125011/20173426/2d2c30f6-a731-11e6-9db0-9518b22b44db.png">

I'd propose from this that we make `--with docs --with-examples` (and maybe even `--with-qtwebkit`) the default and remove `--with-mysql` and `--with-postgresql` options.

RandomDSdevel:
@MikeMcQuaid:  
> @RandomDSdevel I realise you may not have the patience…

[...]

PR #6825, opened 2016-11-11
qt5: cleanup formula (many changes).

MikeMcQuaid:

-----

- Remove all options except `with-dbus`
  - Build docs/examples/qtwebkit by default
  - Remove all other options, they aren't reliable or widely used enough
- Add patch from #6708.
- Remove caveats. Unnecessary now Qt is LGPL.

Will probably hold off merging until https://github.com/Homebrew/brew/issues/1399 is fixed.

Closes #6708.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -1,19 +1,9 @@
-class OracleHomeVarRequirement < Requirement
-  fatal true
-  satisfy(:build_env => false) { ENV["ORACLE_HOME"] }
-
-  def message; <<-EOS.undent
-      To use --with-oci you have to set the ORACLE_HOME environment variable.
-      Check Oracle Instant Client documentation for more information.
-    EOS
-  end
-end
-
 # Patches for Qt5 must be at the very least submitted to Qt's Gerrit codereview
 # rather than their bug-report Jira. The latter is rarely reviewed by Qt.
 class Qt5 < Formula
   desc "Version 5 of the Qt framework"
   homepage "https://www.qt.io/"
+  revision 1
   head "https://code.qt.io/qt/qt5.git", :branch => "5.7", :shallow => false
 
   # Remove stable patches for > 5.7.0
@@ -58,6 +48,13 @@ class Qt5 < Formula
       url "https://raw.githubusercontent.com/Homebrew/formula-patches/34a4ad8/qt5/cups-sierra.patch"
       sha256 "63b5f37d694d0bd1db6d586d98f3c551239dc8818588f3b90dc75dfe6e9952be"
     end
+
+    # Upstream commit from 13 Sep 2016 "Fix crash on exit when using default property aliases with layouts"
+    # http://code.qt.io/cgit/qt/qtdeclarative.git/patch/?id=5149aa68eca6ede8836ec4f07a14d22d9da9b161
+    patch do
+      url "https://raw.githubusercontent.com/Homebrew/formula-patches/3b71525/qt5/QTBUG-51927.patch"
+      sha256 "9460c3cc5ea0f530f24cb92fc9b260a2a7b01ccbdcd0b86e3ddae719a8b53eae"
+    end
   end
 
   bottle do
@@ -68,12 +65,6 @@ class Qt5 < Formula
 
   keg_only "Qt 5 conflicts Qt 4"
 
-  option "with-docs", "Build documentation"
-  option "with-examples", "Build examples"
-  option "with-oci", "Build with Oracle OCI plugin"
-  option "with-qtwebkit", "Build with QtWebkit module"
-  option "without-webengine", "Build without QtWebEngine module"
-
   deprecated_option "qtdbus" => "with-dbus"
   deprecated_option "with-d-bus" => "with-dbus"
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
No revision bump?

MikeMcQuaid:
> No revision bump?

Done.

ilovezfs (dismissed):

ilovezfs:
@BrewTestBot test this please

apjanke:
👍 on making docs and examples build by default and minimizing options.

Do we have some numbers on how widely used the msyql and postgresql options are?

apjanke:
Oh, nevermind: I see you posted it over in #6766.

DomT4:
Aren't the docs huge? I seem to remember they took up wild disk space. Maybe that's changed. Pretty 😭 about `qtwebkit` by default given the "lol security" aspect around WebKit but have no power to dump on it anymore, so.

RandomDSdevel:
@apjanke, @DomT4, @ilovezfs, @MikeMcQuaid:  As an extension of [this comment](https://github.com/Homebrew/homebrew-core/issues/6766#issuecomment-260083889) I made on [Issue #6766, 'Qt v5.7.0 Still Fails to Build on OS X v10.11.6 'El Capitan' After Resolution of Issue #6032'](https://github.com/Homebrew/homebrew-core/issues/6766), would it possibly be a good idea to split this PR into its constituent parts and continue discussing each individual change separately?

ilovezfs:
@DomT4 looks like it balloons from 91 MB to 445 MB

ilovezfs:
@MikeMcQuaid at this point I'm tempted to just leave the formula the way it was. The examples result in a broken installation (see below), the docs are huge, the sql removal is unnecessary since it works now, and the webkit security is questionable. I think it's fine for people to add any, all, or none of those components if they want them, but the formula seems to be "correct" as-is without any of these changes, except perhaps the caveats change, which is cosmetic.

ilovezfs:
Broken installation from examples
```
Error: 1 problem in 1 formula
qt5:
    Broken dylib links found:
      libgruesensor.1.dylib
```

MikeMcQuaid:
> The examples result in a broken installation (see below)

Let's remove the examples option. Pretty sure it was me who added in the first place for my personal use.

> the docs are huge

They are generated in multiple formats. I'll try and slim them down and if not let's remove them and the option. They are available online or as a separate download.

> the sql removal is unnecessary since it works now

works (for) now. It's not widely used and will very likely break again.

> webkit security is questionable

Happy to leave that as optional if someone can point to some known security issues with the current version?

ilovezfs:
Not sure what the impetus for a change here is. The formula is actually fine as is. I'd prefer we just close this out.

MikeMcQuaid:
> Not sure what the impetus for a change here is.

Having to deal with Qt5 build issues every time there's a new macOS or upstream release. The large number of options combined with the build time means we're unable to support this formula effectively. We're having to apply increasing numbers of patches for optional functionality it's not clear anyone is using beyond "I want to build Qt5 with all the options enabled".

ilovezfs:
>we're unable to support this formula effectively

Can you point to something broken?

> build time

Your changes increase build time.

>We're having to apply increasing numbers of patches for optional functionality it's not clear anyone is using beyond "I want to build Qt5 with all the options enabled".

Every patch in the formula currently, except the examples patch, applies regardless of options.

MikeMcQuaid:
> Can you point to something broken?

I can point to the amount of work both by contributors, maintainers and the CI machines in the last few weeks that's been spent on this formula to fix options analytics shows a tiny number of people are using.

> Your changes increase build time.

As I mentioned above but perhaps wasn't clear: I'm proposing we remove examples and the option to build them.

> except the examples patch, applies regardless of options.

I'm proposing we remove the examples option and patch.

-- 

I don't feel strongly about what the defaults are for this formula but I do feel strongly that we should remove some if not all of the non-dbus options. You've generally been in favour in the past of removing options, what's changed here?

ilovezfs:
>I'm proposing we remove the examples option and patch.

My point was the patching has nothing to do with the options.

>I can point to the amount of work both by contributors, maintainers and the CI machines in the last few weeks that's been spent on this formula to fix options analytics shows a tiny number of people are using.

This is an overreaction to what happened with mysql, which was mostly my fault.

>I don't feel strongly about what the defaults are for this formula but I do feel strongly that we should remove some if not all of the non-dbus options. You've generally been in favour in the past of removing options, what's changed here?

qt-webkit was only installed by 691/31300, which is probably enough to warrant keeping it, but not enough to justify making it always-on until I have reason to dismiss @DomT4's concern, which I don't at this point.

>I don't feel strongly about what the defaults are for this formula but I do feel strongly that we should remove some if not all of the non-dbus options. You've generally been in favour in the past of removing options, what's changed here?

The ballooning from 91 MB to 445 MB and Dom's concern about the webkit option. I think this is a case of if-it-ain't-broke-don't-fix-it.

To be clear, I really don't care if we drop the options, but I fully expect the qt5 formula to continue to need patches and support galore in perpetuity regardless of the options kept or removed.

MikeMcQuaid:
> I really don't care if we drop the options

Ok, cool. Let's leave it as-is for now, then, and we'll drop them liberally next time they break.

PR #6828, opened 2016-11-11
apr-util: add openssl 1.1 support

ilovezfs:
:recommended so CI tests it, but may want to start out with it :optional

CC @DomT4

### DIFF
--- a/Formula/apr-util.rb
+++ b/Formula/apr-util.rb
@@ -15,15 +15,33 @@ class AprUtil < Formula
   keg_only :provided_by_osx, "Apple's CLT package contains apr."
 
   option :universal
+  option "without-openssl@1.1", "Build with openssl instead of openssl@1.1"
 
   depends_on "apr"
-  depends_on "openssl"
   depends_on "postgresql" => :optional
   depends_on "mysql" => :optional
   depends_on "freetds" => :optional
   depends_on "unixodbc" => :optional
   depends_on "sqlite" => :optional
   depends_on "homebrew/dupes/openldap" => :optional
+
+  depends_on "openssl@1.1" => :recommended
+
+  if build.with? "openssl@1.1"
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+
+    # backports openssl 1.1 support
+    patch do
+      url "https://mirrors.ocf.berkeley.edu/debian/pool/main/a/apr-util/apr-util_1.5.4-2.debian.tar.xz"
+      mirror "https://mirrors.kernel.org/debian/pool/main/a/apr-util/apr-util_1.5.4-2.debian.tar.xz"
+      sha256 "e8f0fdf94482c43dff69a207ecbf98cec602ab45869561800ccf46a09988caff"
+      apply "patches/openssl-1.1.patch"
+    end
+  else
+    depends_on "openssl"
+  end
 
   def install
     ENV.universal_binary if build.universal?
@@ -32,7 +50,6 @@ class AprUtil < Formula
     args = %W[
       --prefix=#{libexec}
       --with-apr=#{Formula["apr"].opt_prefix}
-      --with-openssl=#{Formula["openssl"].opt_prefix}
       --with-crypto
     ]
 
@@ -45,6 +62,13 @@ class AprUtil < Formula
       args << "--with-ldap"
       args << "--with-ldap-lib=#{Formula["openldap"].opt_lib}"
       args << "--with-ldap-include=#{Formula["openldap"].opt_include}"
+    end
+
+    if build.with? "openssl@1.1"
+      args << "--with-openssl=#{Formula["openssl@1.1"].opt_prefix}"
+      system "autoreconf", "-fiv"
+    else
+      args << "--with-openssl=#{Formula["openssl"].opt_prefix}"
     end
 
     system "./configure", *args
--- a/Formula/subversion.rb
+++ b/Formula/subversion.rb
@@ -4,6 +4,7 @@ class Subversion < Formula
   url "https://www.apache.org/dyn/closer.cgi?path=subversion/subversion-1.9.4.tar.bz2"
   mirror "https://archive.apache.org/dist/subversion/subversion-1.9.4.tar.bz2"
   sha256 "1267f9e2ab983f260623bee841e6c9cc458bf4bf776238ed5f100983f79e9299"
+  revision 1
 
   bottle do
     sha256 "a2f80a24261a9236accccb4d54c8cb611eca11d1f6df7fee0128ed6833663817" => :sierra
@@ -23,21 +24,12 @@ class Subversion < Formula
   option "with-gpg-agent", "Build with support for GPG Agent"
 
   depends_on "pkg-config" => :build
-
-  # macOS Sierra ships the APR libraries & headers, but has removed the
-  # apr-1-config & apu-1-config executables which serf demands to find
-  # those elements. We may need to adopt a broader solution if this problem
-  # expands, but currently subversion is the only breakage as a result.
-  if MacOS.version >= :sierra
-    depends_on "apr-util"
-    depends_on "apr"
-  else
-    depends_on :apr => :build
-  end
+  depends_on "apr-util"
+  depends_on "apr"
 
   resource "serf" do
-    url "https://archive.apache.org/dist/serf/serf-1.3.8.tar.bz2"
-    sha256 "e0500be065dbbce490449837bb2ab624e46d64fc0b090474d9acaa87c82b2590"
+    url "https://archive.apache.org/dist/serf/serf-1.3.9.tar.bz2"
+    sha256 "549c2d21c577a8a9c0450facb5cca809f26591f048e466552240947bdf7a87cc"
   end
 
   # Always build against Homebrew versions instead of system versions for consistency.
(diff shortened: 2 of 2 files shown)


MikeMcQuaid:
👍 on adding this option for testing but if we do go for on-by-default I think it's probably worth not having an option.

DomT4:
You'll want to check how much `brew uses apr-util` stuff also `brew uses openssl` and then see if those can be moved across to `openssl@1.1` as well, but whilst patching crypto support is hairy I think if folks want to move forwards on `openssl@1.1` transitions `apr-util` will need to be shunted across due to its increasingly wide usage.

DomT4:
Colour me pleasantly surprised `serf` actually builds with `openssl@1.1`. There's some redundant logic further down the `subversion` formula about `apr=/usr/` etc that can be shrunk since Homebrew has adopted a love of using its Apr formulae rather than clinging onto the system for 10.11 and lower.

ilovezfs:
Looks like the relevant brew uses set, assuming no undeclared dependencies on openssl, is apib, freeswitch, hydra, nginx, passenger, subversion, and subversion18.

ilovezfs:
If I also include the dependencies of the reverse dependencies, then that adds to the set curl, freetds, homebrew/dupes/openldap, ldns, libevent, libssh, libssh2, mysql, nghttp2, postgresql, python, rtmpdump, and spdylay.

ilovezfs:
reverse dependencies that use openssl:

- [ ] apib
- [ ] freeswitch
- [ ] hydra
- [ ] nginx
- [ ] passenger
- [ ] subversion
- [ ] homebrew/versions/subversion18

---

dependencies of reverse dependencies that use openssl:

- [ ] curl (optional)
- [ ] freetds
- [ ] homebrew/dupes/openldap
- [ ] ldns
- [ ] libevent
- [ ] libssh
- [ ] libssh2
- [ ] mysql
- [ ] nghttp2
- [ ] postgresql
- [ ] python
- [ ] rtmpdump
- [ ] spdylay

ilovezfs:
Working the tree back to reverse dependencies of the dependencies of the reverse dependencies, the scope reaches to

- [ ] ansible
- [ ] certbot
- [ ] efl
- [ ] eralchemy
- [ ] ffmpeg
- [ ] folly
- [ ] gearman
- [ ] getdns
- [ ] gloox
- [ ] gpac
- [ ] gst-plugins-bad
- [ ] gwenhywfar
- [ ] homebrew/dupes/openssh
- [ ] homebrew/python/pyrit
- [ ] homebrew/science/blat
- [ ] homebrew/science/gdcm
- [ ] homebrew/science/kent-tools
- [ ] homebrew/science/openimageio
- [ ] homebrew/science/root
- [ ] homebrew/science/root6
- [ ] homebrew/science/ucsc-genome-browser
- [ ] homebrew/versions/ffmpeg28
- [ ] homebrew/versions/gst-plugins-bad010
- [ ] homebrew/versions/node010
- [ ] homebrew/versions/postgresql93
- [ ] homebrew/versions/postgresql94
- [ ] homebrew/versions/thrift090
- [ ] innotop
- [ ] inspircd
- [ ] knot
- [ ] kore
- [ ] libcouchbase
- [ ] libdap
- [ ] libgda
- [ ] libimobiledevice
- [ ] libopendkim
- [ ] libtorrent-rasterbar
- [ ] libxmlsec1
- [ ] libzdb
- [ ] megatools
- [ ] midnight-commander
- [ ] mitmproxy
- [ ] mycli
- [ ] mydumper
- [ ] mysql-connector-c++
- [ ] mytop
- [ ] nagios-plugins
- [ ] net-snmp
- [ ] nrpe
- [ ] nsd
- [ ] nut
- [ ] ooniprobe
- [ ] openconnect
- [ ] opensaml
- [ ] osquery
- [ ] percona-toolkit
- [ ] percona-xtrabackup
- [ ] pgcli
- [ ] ponyc
- [ ] postgres-xc
- [ ] profanity
- [ ] psqlodbc
- [ ] pure-ftpd
- [ ] pwntools
- [ ] qca
- [ ] qt
- [ ] radare2
- [ ] rtags
- [ ] rust
- [ ] sane-backends
- [ ] sslsplit
- [ ] sstp-client
- [ ] strongswan
- [ ] sysbench
- [ ] thrift
- [ ] tlsdate
- [ ] tor
- [ ] unbound
- [ ] upscaledb
- [ ] urweb
- [ ] uwsgi
- [ ] xml-tooling-c

ilovezfs:
And taking into account the dependencies of the reverse dependencies of the dependencies of the reverse dependencies:

- [ ] afflib
- [ ] asio
- [ ] botan
- [ ] cpp-netlib
- [ ] libewf
- [ ] libstrophe
- [ ] mongodb
- [ ] neon
- [ ] pkcs11-helper
- [ ] pulseaudio
- [ ] pypy
- [ ] python3
- [ ] srtp
- [ ] xml-security-c
- [ ] xrootd
- [ ] yara

DomT4:
I think you're noticing why I was worried this could take Homebrew years to get done 😄. Debian has patched an awful lot in their unstable distro, so Homebrew could go for an en mass update using Debian patching, but uncertain if those patches went upstream, have been reviewed/accepted upstream/etc.

ilovezfs:
@DomT4 I think it also shows why simultaneously having multiple versions of the same library in core is :no_good_man:

DomT4:
Heh, well, that ship has probably sailed or is about to sail with version support, because there's pretty significant demand there to stop breaking taps on `boost` updates, etc regularly.

But OpenSSL in Homebrew is inherently complicated because almost anything that touches crypto/ssl on OS X uses or can use OpenSSL but the system ships garbage, which means relying on Homebrew's OpenSSL 100%. Even stuff like killing SSLv3 for `openssl` which is desirable is more or less impossible to do now thanks to how wide a revision season it would need.

ilovezfs:
>Heh, well, that ship has probably sailed or is about to sail with version support, because there's pretty significant demand there to stop breaking taps on boost updates, etc regularly.

We may hard code the bottle versions instead.

MikeMcQuaid:
> We may hard code the bottle versions instead.

That would be an "as well" rather than "instead"; it doesn't help anyone building from source.

DomT4:
I think any versioning solution is going to end up being overwhelmingly complicated, to be honest. I agree the version support needs to improve, but across my time in Homebrew up to last month nobody, myself included, had really fleshed out a mechanism that didn't have serious issues around either complexity, heavy maintainer burden, or potential security implications. The whole thing is a bit of an endless nightmare.

ilovezfs:
>I think any versioning solution is going to end up being overwhelmingly complicated, to be honest. I agree the version support needs to improve, but across my time in Homebrew up to last month nobody, myself included, had really fleshed out a mechanism that didn't have serious issues around either complexity, heavy maintainer burden, or potential security implications. The whole thing is a bit of an endless nightmare.

This.

ilovezfs:
The scope of this PR might as well be all of https://github.com/Homebrew/homebrew-core/issues/4543 so I'm closing this out.

MikeMcQuaid:
> I think any versioning solution is going to end up being overwhelmingly complicated, to be honest. I agree the version support needs to improve, but across my time in Homebrew up to last month nobody, myself included, had really fleshed out a mechanism that didn't have serious issues around either complexity, heavy maintainer burden, or potential security implications. The whole thing is a bit of an endless nightmare.

> more or less impossible to do now thanks to how wide a revision season it would need.

It's just going to be different. It's far from an insurmountable problem as pretty much every other package manager has better versioning support that we do and it's the number 1 recurring complaint about Homebrew. Regardless, it doesn't really matter as I'm the only person actively working on making this better so as long as I decide to keep working on it: it'll get better.

> That would be an "as well" rather than "instead"; it doesn't help anyone building from source.

Instead of the 👎 on this @ilovezfs a reason explaining what your objection is and how you propose helping people building from source would be good.

ilovezfs:
>every other package manager has better versioning support that we do 

In the past you've been extremely quick to dismiss arguments that use that as a justification. People can feel free to use one of those other package managers.

>Instead of the :-1: on this @ilovezfs a reason explaining what your objection is 

It's will be a lot of extra, ongoing – in perpetuity – maintainer work (and I'm _not_ talking about the homebrew/brew changes you're volunteering to keep working on; the point is about the work in homebrew/core) for little benefit. The cost-benefit analysis is not even close.

>and how you propose helping people building from source would be good.

We could have a source build mode available that always builds a bottle and then installs it, and keep the bottle in the cache, using the same pegging strategy as the non-source builders.

MikeMcQuaid:
> It will be a lot of extra, ongoing – in perpetuity – maintainer work (and I'm not talking about the homebrew/brew changes you're volunteering to keep working on; the point is about the work in homebrew/core) for little benefit. The cost-benefit analysis is not even close.

At the bare minimum there will be no more work than homebrew/versions while having the benefit that we don't intentionally break things. You are welcome to say "I'm not willing to maintain old versions of software" and that's fine. I have to use homebrew/versions every day to do my job and the current experience is not good enough and I'm going to make it better.

> We could have a source build mode available that always builds a bottle and then installs it, and keep the bottle in the cache, using the same pegging strategy as the non-source builders.

Which is essentially equivalent to not using `brew cleanup` or fixing it so it doesn't clean up linked directories. I'm all for that but it's not really a long-term solution.

ilovezfs:
I'm tired of wasting my breath on this conversation. Do whatever you want.

PR #6835, opened 2016-11-11
dbus: add manpages and missing build dependencies

zbentley:

-----
There are a few problems with the `dbus` formula:
- This package has many scripts, all of which have manpages. However, those manpages aren't available in many standard (e.g. linux.die) websites, and are not compiled when this package is built. We should build the manpages. This requires some new build-time dependencies.
- Building at `HEAD` (and possibly elsewhere) requires `autoconf-archive`, which is not listed as a dependency, and the build fails as a result.
- `--build-from-source` is totally broken, since all of the build toolchain dependencies are inside a `head` block, preventing the toolchain from being available unless building at `HEAD`.
- [Re]configure bootstrapping is disabled unless building at `HEAD`. This prevents documentation from building, and, aside from another ~6sec of build time (which isn't bad, considering what DBus is), is a totally unnecessary behavior fork. However, if there's good reason to leave this override in place, I am happy to undo this change.

Ordinarily I'd fix some of those in separate PRs, but they all block my ability to add the manpages (the real goal), so I'll fix 'em all in this PR.

### DIFF
--- a/Formula/dbus.rb
+++ b/Formula/dbus.rb
@@ -29,11 +29,6 @@ class Dbus < Formula
 
   depends_on "xmlto" => :build
 
-  # Docbook is a dependency of xmlto, but the XML_CATALOG_FILES env-set in
-  # install() uses files created by docbook, so best to be explicit in case
-  # the install process for docbook ever stops creating those files.
-  depends_on "docbook" => :build
-
   # Patch applies the config templating fixed in https://bugs.freedesktop.org/show_bug.cgi?id=94494
   # Homebrew pr/issue: 50219
   patch do
@@ -45,8 +40,6 @@ class Dbus < Formula
     # Fix the TMPDIR to one D-Bus doesn't reject due to odd symbols
     ENV["TMPDIR"] = "/tmp"
 
-    # Manpages won't build without a current docbook catalog. This should exist
-    # if xmlto and docbook (build dependencies of this package) are installed.
     ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
 
     system "./autogen.sh", "--no-configure" if build.head?
--- a/Formula/dbus.rb
+++ b/Formula/dbus.rb
@@ -5,6 +5,7 @@ class Dbus < Formula
   url "https://dbus.freedesktop.org/releases/dbus/dbus-1.10.12.tar.gz"
   mirror "https://mirrors.ocf.berkeley.edu/debian/pool/main/d/dbus/dbus_1.10.12.orig.tar.gz"
   sha256 "210a79430b276eafc6406c71705e9140d25b9956d18068df98a70156dc0e475d"
+  head "https://anongit.freedesktop.org/git/dbus/dbus.git"
 
   bottle do
     sha256 "8848b7e368750df3a9526f4c5d47a0649359e9e89cd9d94cb45e706402bdb66c" => :sierra
@@ -18,13 +19,16 @@ class Dbus < Formula
     sha256 "474de2afde8087adbd26b3fc5cbf6ec45559763c75b21981169a9a1fbac256c9"
   end
 
-  head do
-    url "https://anongit.freedesktop.org/git/dbus/dbus.git"
+  depends_on "autoconf" => :build
+  depends_on "autoconf-archive" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "xmlto" => :build
 
-    depends_on "autoconf" => :build
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
-  end
+  # Docbook is a dependency of xmlto, but the XML_CATALOG_FILES env-set in
+  # install() uses files created by docbook, so best to be explicit in case
+  # the install process for docbook ever stops creating those files.
+  depends_on "docbook" => :build
 
   # Patch applies the config templating fixed in https://bugs.freedesktop.org/show_bug.cgi?id=94494
   # Homebrew pr/issue: 50219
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> This requires some new build-time dependencies.

Which ones? If they are not usually built that makes me wonder if we should do so.

> Building at HEAD (and possibly elsewhere) requires autoconf-archive, which is not listed as a dependency, and the build fails as a result.

It'll just be HEAD, please scope these changes just to HEAD unless you know for sure otherwise.

> --build-from-source is totally broken, since all of the build toolchain dependencies are inside a head block, preventing the toolchain from being available unless building at HEAD.

That's not true, `--build-from-source` works fine as-is.

> However, if there's good reason to leave this override in place, I am happy to undo this change.

We only want to reconfigure bootstrapping when it's necessary for some option or HEAD. Please keep it scoped to that.

zbentley:
Sorry for the confusion. At the top level there are two changes: adding the documentation builder, and removing the "skip bootstrap" build step, which is currently conditional on `--HEAD`. I can open a separate PR for the latter if that would help.

> Which ones? If they are not usually built that makes me wonder if we should do so.

`autoconf-archive` is now required no matter what if building at `--HEAD`. With the old version of the formula, attempts to install at `HEAD` fail with the error in the attached 
[missingarchive.txt](https://github.com/Homebrew/homebrew-core/files/587276/missingarchive.txt)

> We only want to reconfigure bootstrapping when it's necessary for some option or HEAD. Please keep it scoped to that.

I'm interested in why you say that, but that's totally separate from this PR. I will ship an update in the next ~15min that reduces changes down to just manpages and the `HEAD` dependency on `autoconf-archive`.  Thanks!

100% aside from this PR: I think that it's best to have a single build process, and to avoid telling Homebrew to skip bootstrapping/use pre-built config/makefiles as a sort of "asset cache" if not building at `HEAD`. If build performance is the problem, or if the extra dependencies are considered bloat, there are bottles for that. Unless there is an important change in build dependencies between the sources of the default version and `HEAD`, I think it's best to keep the number of build paths as close to 1 as possible.

It is 100% possible/likely that I'm missing something dumb, though; I'm not very experienced at this.

Formula/dbus.rb
```diff
@@ -19,13 +18,18 @@ class Dbus < Formula
...
+    depends_on "autoconf" => :build
+    depends_on "autoconf-archive" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+  end
+
   depends_on "xmlto" => :build
 
-  # Docbook is a dependency of xmlto, but the XML_CATALOG_FILES env-set in
+  # doccbook is a dependency of xmlto, but the XML_CATALOG_FILES env-set in
```
ilovezfs: one c

Formula/dbus.rb
```diff
@@ -22,10 +22,18 @@ class Dbus < Formula
     url "https://anongit.freedesktop.org/git/dbus/dbus.git"
 
     depends_on "autoconf" => :build
+    depends_on "autoconf-archive" => :build
     depends_on "automake" => :build
     depends_on "libtool" => :build
   end
 
+  depends_on "xmlto" => :build
```
ilovezfs: I think you only need docbook.

zbentley: I'm not observing that. Without `xmlto`: 
```
==> Installing dbus
==> Downloading https://dbus.freedesktop.org/releases/dbus/dbus-1.10.12.tar.gz
Already downloaded: /Users/zbentley/Library/Caches/Homebrew/dbus-1.10.12.tar.gz
==> Downloading https://raw.githubusercontent.com/Homebrew/formula-patches/0a8a55872e/d-bus/org.freedesktop.dbus-session.plist.osx.diff
Already downloaded: /Users/zbentley/Library/Caches/Homebrew/dbus--patch-a8aa6fe3f2d8f873ad3f683013491f5362d551bf5d4c3b469f1efbc5459a20dc.diff
==> Patching
==> Applying org.freedesktop.dbus-session.plist.osx.diff
patching file bus/org.freedesktop.dbus-session.plist.in
==> ./configure --prefix=/usr/local/Cellar/dbus/1.10.12 --localstatedir=/usr/local/var --sysconfdir=/usr/local/etc --enable-xml-docs --disable-doxygen-docs --enable-launchd --with-l
Last 15 lines from /Users/zbentley/Library/Logs/Homebrew/dbus/01.configure:
checking to see if compiler understands -Wno-pointer-sign... yes
checking to see if compiler understands -Wno-error=pointer-sign... yes
checking to see if compiler understands -Wno-type-limits... yes
checking to see if compiler understands -Wno-error=type-limits... yes
checking to see if compiler understands -fno-common... yes
checking to see if compiler understands -fno-strict-aliasing... yes
checking if LD -Wl,--version-script works... no
checking for doxygen... no
checking whether to build Doxygen documentation... no
checking for xsltproc... xsltproc
checking for ducktype... no
checking for yelp-build... no
checking whether to build Ducktype documentation... no
checking for xmlto... no
checking whether to build XML documentation... configure: error: Building XML docs explicitly required, but xmlto not found

READ THIS: https://git.io/brew-troubleshooting```

Comment (ilovezfs) [inline Formula/dbus.rb]: Cool, thanks for checking.

File: Formula/dbus.rb [hunk 3]
```diff
@@ -22,10 +22,18 @@ class Dbus < Formula
...
+
+  # Docbook is a dependency of xmlto, but the XML_CATALOG_FILES env-set in
+  # install() uses files created by docbook, so best to be explicit in case
+  # the install process for docbook ever stops creating those files.
```
ilovezfs: Let's drop this entire comment and the explicit docbook dependency. ascii, deheader, pgbouncer, reposurgeon, and shadowsocks-libev all have the same situation and none declares a docbook dependency.

Formula/dbus.rb
```diff
@@ -36,13 +44,17 @@ class Dbus < Formula
   def install
     # Fix the TMPDIR to one D-Bus doesn't reject due to odd symbols
     ENV["TMPDIR"] = "/tmp"
+    
+    # Manpages won't build without a current docbook catalog. This should exist
```
ilovezfs: This comment is not needed. There are 29 cases just like this core.

Formula/dbus.rb
```diff
@@ -37,12 +45,16 @@ def install
     # Fix the TMPDIR to one D-Bus doesn't reject due to odd symbols
     ENV["TMPDIR"] = "/tmp"
 
+    # Manpages won't build without a current docbook catalog. This should exist
+    # if xmlto and docbook (build dependencies of this package) are installed.
```
ilovezfs: This comment is not needed. There are 29 cases just like this in core.

zbentley:
I removed the comment, but I have to admit I can't see why I'm getting the `version should not decrease` error.  Was there another update made? My Git-fu is not strong.

MikeMcQuaid:

> version should not decrease

This was a bug and can be ignored.

zbentley:
Sorry for all the back-and-forth @MikeMcQuaid @ilovezfs; thanks!

MikeMcQuaid:
@zbentley No apologies needed, thanks!

PR #6836, opened 2016-11-11
Zookeeper default zk env respect zoocfgdir

ddillinger:

-----

### DIFF
--- a/Formula/zookeeper.rb
+++ b/Formula/zookeeper.rb
@@ -38,7 +38,7 @@ class Zookeeper < Formula
 
   def default_zk_env
     <<-EOS.undent
-        [ -z "$ZOOCFGDIR" ] && export ZOOCFGDIR="#{etc}/zookeeper"
+      [ -z "$ZOOCFGDIR" ] && export ZOOCFGDIR="#{etc}/zookeeper"
     EOS
   end
 
--- a/Formula/zookeeper.rb
+++ b/Formula/zookeeper.rb
@@ -38,7 +38,9 @@ class Zookeeper < Formula
 
   def default_zk_env
     <<-EOS.undent
-      export ZOOCFGDIR="#{etc}/zookeeper"
+      if [ -z ${ZOOCFGDIR+x} ]; then
+          export ZOOCFGDIR="#{etc}/zookeeper"
+      fi
     EOS
   end
 
--- a/Formula/zookeeper.rb
+++ b/Formula/zookeeper.rb
@@ -38,9 +38,7 @@ class Zookeeper < Formula
 
   def default_zk_env
     <<-EOS.undent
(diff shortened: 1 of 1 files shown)


Formula/zookeeper.rb
```diff
@@ -38,7 +38,9 @@ def shim_script(target)
 
   def default_zk_env
     <<-EOS.undent
-      export ZOOCFGDIR="#{etc}/zookeeper"
+      if [ -z ${ZOOCFGDIR+x} ]; then
```
MikeMcQuaid: `[ -z "$ZOOCFGDIR" ] && export ZOOCFGDIR="#{etc}/zookeeper"`

ddillinger: This is one of those weird shell edge case best practice things where it can't tell the difference between unset and empty string unless you go the extra mile.

See [this stackoverflow](http://stackoverflow.com/a/138648290 for in depth details.

theckman:
@ddillinger that link is 404ing for me.

Formula/zookeeper.rb
theckman: @ddillinger that link is 404ing for me.

ddillinger: No idea why that link works locally but fails here, but yeah it sure does fail here. I fixed it.

theckman: @ddillinger Looking at the information in that linked StackOverflow, it seems that neither really apply strongly here. I'd also argue that trying to discern between unset and `""` would be the opposite behavior of what we want. `""` is not a valid UNIX path as far as I'm aware.

MikeMcQuaid: Please use `[ -z "$ZOOCFGDIR" ] && export ZOOCFGDIR="#{etc}/zookeeper"` as requested, thanks.

ilovezfs:
kafka test got stuck in some kind of infinite loop during `brew test --verbose kafka`

https://bot.brew.sh/job/Homebrew%20Core/11450/version=any_sierra/console

Formula/zookeeper.rb
```diff
@@ -38,7 +38,7 @@ def shim_script(target)
 
   def default_zk_env
     <<-EOS.undent
-      export ZOOCFGDIR="#{etc}/zookeeper"
+        [ -z "$ZOOCFGDIR" ] && export ZOOCFGDIR="#{etc}/zookeeper"
```
MikeMcQuaid: This doesn't need any added indentation so remove the two spaces at the beginning.

MikeMcQuaid (changes requested):

MikeMcQuaid (approved):

MikeMcQuaid:

PR #6838, opened 2016-11-11
tmux: enhance Formula to make utf8proc support optional

theckman:

-----

This change enhances the `tmux` Formula so that `utf8proc` support is an
opt-in change when building `tmux`.

There have been reports of rendering issues in versions of `tmux` when
`utf8proc` support is enabled. Specifically these issues appear to
happen when special symbols are printed, such as those commonly used in
the powerline-themed status lines and shell prompts.

The change to include `utf8proc` in `tmux` was in Homebrew PR #5665. In
that some PR some `tmux` users have reported issues with text being
rendered in builds of `tmux`. There are also some other issues on the
`tmux` tracker around similar rendering problems:

* https://github.com/tmux/tmux/issues/631
* https://github.com/tmux/tmux/issues/632
* https://github.com/tmux/tmux/issues/637
* https://github.com/tmux/tmux/issues/639

### DIFF
--- a/Formula/tmux.rb
+++ b/Formula/tmux.rb
@@ -3,7 +3,7 @@ class Tmux < Formula
   homepage "https://tmux.github.io/"
   url "https://github.com/tmux/tmux/releases/download/2.3/tmux-2.3.tar.gz"
   sha256 "55313e132f0f42de7e020bf6323a1939ee02ab79c48634aa07475db41573852b"
-  revision 1
+  revision 2
 
   bottle do
     cellar :any
@@ -22,7 +22,7 @@ class Tmux < Formula
 
   depends_on "pkg-config" => :build
   depends_on "libevent"
-  depends_on "utf8proc"
+  depends_on "utf8proc" => :optional
 
   resource "completion" do
     url "https://raw.githubusercontent.com/imomaliev/tmux-bash-completion/homebrew_1.0.0/completions/tmux"
@@ -32,11 +32,16 @@ class Tmux < Formula
   def install
     system "sh", "autogen.sh" if build.head?
 
+    args = %W[
+      --disable-Dependency-tracking
+      --prefix=#{prefix}
+      --sysconfdir=#{etc}
+    ]
+
+    args << "--enable-utf8proc" if build.with?("utf8proc")
+
     ENV.append "LDFLAGS", "-lresolv"
-    system "./configure", "--disable-dependency-tracking",
-                          "--prefix=#{prefix}",
-                          "--sysconfdir=#{etc}",
-                          "--enable-utf8proc"
+    system "./configure", *args
 
     system "make", "install"
 


Formula/tmux.rb
```diff
@@ -20,9 +20,11 @@ class Tmux < Formula
     depends_on "libtool" => :build
   end
 
+  option "with-utf8proc", "Enable utf8proc support"
```
dunn: The option text will be generated automatically by the `:optional` flag below.

Formula/tmux.rb
```diff
@@ -32,11 +34,16 @@ class Tmux < Formula
   def install
     system "sh", "autogen.sh" if build.head?
 
+    configure_args = [
+      "--disable-Dependency-tracking",
+      "--prefix=#{prefix}",
+      "--sysconfdir=#{etc}",
+    ]
```
zmwangx:
```rb
args = %W[
   ...
]
```
is the convention used throughout homebrew/core.

theckman: If that's the case why is `rubocop` not enforcing it?

Formula/tmux.rb
theckman: @dunn The branch should now have this removed. Thank you for the review.

Formula/tmux.rb
zmwangx: Well there's no point in enforcing everything. Experienced human beings are watching this repo, and should be able to point out things like this. (Also, I doubt you can ask rubocop to flag `configure_args` in favor of `args`.)

zmwangx: > For things like deciding which way to define Arrays of strings, you can absolutely alter Rubocop to enforce the style of the repo.

This does not conflict with "I doubt you can ask rubocop to flag configure_args in favor of args".

As for the way to define arrays of strings, I was talking specifically about the convention for defining the `args` array to `configure` (although I do believe `%W[]` or `%w[]` are preferred). Conventions aren't necessarily rules, and don't necessarily require enforcement, but it's nice to follow them for consistency.

> Knowing that the contribution guidelines require you to run rubocop to enforce style, and that this PR passed that check, do you feel this is absolutely a blocker on the PR?

I never said it's a blocker, and I only left a comment, not a "request changes". By the way I'm not even a maintainer at this point, so you can always ignore my opinions. But what's so hard about taking a (friendly, or at least neutral) advice? Why are you offended, if I'm reading you correctly?

theckman: I wasn't offended at all, and was trying to understand how important you felt the change was and whether I should go about changing it. I've hit a few projects where people are still not using the review system fully, so I think I'm still kinda of operating in the old model of GH reviews where a comment meant a change was requested unless otherwise specified.

I was more confused as to why I ran the style/convention checker, fixed issues before opening the PR, to then be told I've more style issues that need to be addressed. I was also trying to understand what I had missed, so that I could avoid needing more style fixes after opening a PR in the future.

zmwangx: > I wasn't offended at all.

Okay, sorry.

> I was more confused as to why I ran the style/convention checker, fixed issues before opening the PR, to then be told I've more style issues that need to be addressed.

Well, again it's safe to ignore me, but I was just trying to help. Style check is only the minimum bar to clear, it's not unreasonable that people more familiar with the code base and its trends can suggest improvements.

In this specific case, you don't need to take my word for it. Just `grep -rI 'args ='` in the repo. You'll realize the majority uses my suggested style. There are some outliers, but if you blame them you'll realize they're generally code that hasn't changed in a long time.

> I was also trying to understand what I had missed, so that I could avoid needing more style fixes after opening a PR in the future.

I don't think having to push style fixes after opening a PR is shameful or anything. Again, I wasn't trying to judge you (and I'm not in the position to judge you), I was just trying to improve the code.

theckman: I definitely don't want to ignore you when your intent was to improve me and the codebase. I'm sorry if the way I asked the questions implied an abrasive or confrontational tone, definitely not my intent.

I think I also misunderstood your comment regarding the `args` argument. The two formula I used for inspiration actually used the `[]` format, which sounds more like bad luck on my part.

zmwangx: > I'm sorry if the way I asked the questions implied an abrasive or confrontational tone, definitely not my intent.

No worries, let's get over the misunderstanding.

MikeMcQuaid:

PR #6875, opened 2016-11-13
docker-machine-driver-xhyve: fix libev issue

### LINKED ISSUE #6678: docker-machine-driver-xhyve failed to build on 10.10.5
### Bug reports:

New version of `docker-machine-driver-xhyve`, `0.3.0`, fails to build with the following message:

```
==> make build
Last 15 lines from /Users/rag/Library/Logs/Homebrew/docker-machine-driver-xhyve/04.make:
  "_ev_loop_destroy", referenced from:
      _lwt_libev_stop in liblwt-unix_stubs.a(lwt_libev_stubs.o)
  "_ev_loop_new", referenced from:
      _lwt_libev_init in liblwt-unix_stubs.a(lwt_libev_stubs.o)
  "_ev_run", referenced from:
      _lwt_libev_loop in liblwt-unix_stubs.a(lwt_libev_stubs.o)
  "_ev_set_invoke_pending_cb", referenced from:
      _lwt_libev_init in liblwt-unix_stubs.a(lwt_libev_stubs.o)
  "_ev_timer_start", referenced from:
      _lwt_libev_timer_init in liblwt-unix_stubs.a(lwt_libev_stubs.o)
  "_ev_timer_stop", referenced from:
      _lwt_libev_timer_stop in liblwt-unix_stubs.a(lwt_libev_stubs.o)
make: *** [bin/docker-machine-driver-xhyve] Error 2
```

https://gist.github.com/mtpereira/488d34625ae4639812d0c21aa41bd1cc

zchee:

-----

Change to without `make` build.
`libev` static library file was hardcoded my `Makefile`, It is not a good idea to update the version for fix it.
So, I wrote the doing with the `Makefile`.

Close #6678

### DIFF
--- a/Formula/docker-machine-driver-xhyve.rb
+++ b/Formula/docker-machine-driver-xhyve.rb
@@ -20,8 +20,11 @@ class DockerMachineDriverXhyve < Formula
   depends_on "docker-machine" => :recommended
   if build.with? "qcow2"
     depends_on "opam"
-    depends_on "libev"
+    depends_on "libev" => :build
   end
+
+  # Allow specifying version and libev location.
+  patch :DATA
 
   def install
     (buildpath/"gopath/src/github.com/zchee/docker-machine-driver-xhyve").install Dir["{*,.git,.gitignore,.gitmodules}"]
@@ -38,8 +41,12 @@ class DockerMachineDriverXhyve < Formula
 
       if build.with? "qcow2"
         build_tags << " qcow2"
+        ENV["LIBEV_FILE"] = "#{Formula["libev"].lib}/libev.a"
+
         system "opam", "init", "--no-setup"
         opam_dir = "#{buildpath}/.brew_home/.opam"
+
+        # To imitate 'eval `opam init`'
         ENV["CAML_LD_LIBRARY_PATH"] = "#{opam_dir}/system/lib/stublibs:/usr/local/lib/ocaml/stublibs"
         ENV["OPAMUTF8MSGS"] = "1"
         ENV["PERL5LIB"] = "#{opam_dir}/system/lib/perl5"
@@ -48,11 +55,11 @@ class DockerMachineDriverXhyve < Formula
         system "opam", "install", "-y", "uri", "qcow-format", "conf-libev"
       end
 
-      go_ldflags = "-w -s -X 'github.com/zchee/docker-machine-driver-xhyve/xhyve.GitCommit=Homebrew#{git_hash}'"
+      go_ldflags = "-w -s -X=github.com/zchee/docker-machine-driver-xhyve/xhyve.GitCommit=Homebrew-#{git_hash}"
       ENV["GO_LDFLAGS"] = go_ldflags
       ENV["GO_BUILD_TAGS"] = build_tags
       system "make", "lib9p"
-      system "make", "build"
+      system "make", "build", "V=1"
       bin.install "bin/docker-machine-driver-xhyve"
     end
   end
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> So, I wrote the doing without the Makefile.

Could the Makefile be patched instead? Then formula is pretty messy after this. Ideally if you are going to or have fixed this upstream already we'll just apply the upstream patch.

zchee:
@MikeMcQuaid I see. will do.

BTW, What command is can be tested efficiently? `brew audit` is only lint(IIRC), What kind of command is good for install test in the **sandbox**?
I saw `man brew`, but `install` command haven't `--dry-run` flag.
I build on my repository directory with `make`, always use the latest dev(because I'm an owner). 
So I do not want to install by `brew`

MikeMcQuaid:
@zchee You could use `brew test-bot` in CI to test installation is working correctly. I'm not quite sure what you're asking, though, could you rephase/elaborate? Thanks!

zchee:
@MikeMcQuaid Fixed with `patch :DATA`. PTAL.

zchee:
@MikeMcQuaid 
> You could use brew test-bot in CI to test installation is working correctly. I'm not quite sure what you're asking, though, could you rephase/elaborate? Thanks!

Thanks advice :)

Formula/docker-machine-driver-xhyve.rb
```diff
@@ -38,8 +40,12 @@ def install
 
       if build.with? "qcow2"
         build_tags << " qcow2"
+        ENV["LIBEV_FILE"] = "#{Formula["libev"].lib}/libev.a"
```
MikeMcQuaid: If you're using the static library you can do `depends_on "libev" => :build` above.

Formula/docker-machine-driver-xhyve.rb
```diff
@@ -23,6 +23,8 @@ class DockerMachineDriverXhyve < Formula
     depends_on "libev"
   end
 
+  patch :DATA
```
MikeMcQuaid: Can you please add a link to the upstream patch commit and explanation of why the patch is needed in a comment in the formula file. Thanks!

MikeMcQuaid (changes requested):

Formula/docker-machine-driver-xhyve.rb
zchee: That's patch is not merge to upstream yet. Because `$(LIBEV_FILE)` is just carried from
https://github.com/docker/hyperkit/blob/master/Makefile#L83
I would like to follow hyperkit as much as possible. but, Now I'm considering allowing the user to specify the path. So, there is no commit url yet.

Also, sometimes Go users set `$GO_LDFLAGS` to shell global. i.e. always strip symbol table & DWARF with `-s -w`.
In this case, if change the `+=` to `?=`, GNU Make handling shell `$GO_LDFLAGS` env(IIRC).
It means, ignores `-X=xhyve.GitCommit=(...)` trick.
There is a possibility that I can not take a commit hash on the issue, So I am not going to change upstream.

Well, If some comment is necessary, I would appreciate it if I could get your advice, because I am not good at English...

Formula/docker-machine-driver-xhyve.rb
zchee: Wait for the above reply, and fix this together.

Formula/docker-machine-driver-xhyve.rb
zchee: Ah, In other words, I will not this patch to my repository, because have side effects.
This patch is Homebrew specific.

I'll add `# Allow specifying version and libev location`.

MikeMcQuaid:
> Ah, In other words, I will not this patch to my repository, because have side effects.
> This patch is Homebrew specific.
> 
> I'll add # Allow specifying version and libev location.

Just to pull this out so it's visible: Homebrew does not accept Homebrew-specific patches that will never be merged upstream. We'll need to figure out something that can work both for your upstream project and Homebrew.

zchee:
@MikeMcQuaid Hmm...
Do you also not admit `inreplace`?

MikeMcQuaid:
> Do you also not admit inreplace?

@zchee We'd pretty strongly rather avoid it. It makes it much harder to keep the formula up to date. What's wrong with your patch being merged upstream as-is?

zchee:
@MikeMcQuaid 
> What's wrong with your patch being merged upstream as-is?

As I said on https://github.com/Homebrew/homebrew-core/pull/6875#discussion_r88623112, below are why I do not merge patches (after the `__END__` in `docker-machine-driver-xhyve.rb`) to my repository.

> $(LIBEV_FILE) is just carried from
https://github.com/docker/hyperkit/blob/master/Makefile#L83
I would like to follow hyperkit as much as possible.

> Also, sometimes Go users set `$GO_LDFLAGS` to shell global. i.e. always strip symbol table & DWARF with `-s -w`.
In this case, if change the `+=` to `?=`, GNU Make handling shell `$GO_LDFLAGS` env(IIRC).
It means, ignores `-X=xhyve.GitCommit=(...)` trick.
There is a possibility that I can not take a commit hash on the issue, So I am not going to change upstream.

In other words, without `brew`, If some users `export GO_LDFLAGS='-s -w'` and `make`, this variable will not change to commit hash. always `HEAD`.
https://github.com/zchee/docker-machine-driver-xhyve/blob/master/xhyve/version.go#L8
So, I will not be able to determine the binary's *commit level* version.

This patch have problems without `brew`.
The `brew` command is executed within the sandbox and overwrite `$GO_LDFLAGS`, so there is no problem. but it is not good for self-build users.

If following the tool manager's policy, and change my repository's `Makefile` just for `brew`, it may be difficult to debug an issue.
I can not consent to it.

MikeMcQuaid:
> In other words, without brew, If some users export GO_LDFLAGS='-s -w' and make, this variable will not change to commit hash. always HEAD.

You could instead provide a `PACKAGER` variable or similar which adds to the version rather than overriding the `GO_LDFLAGS` as a whole.

> If following the tool manager's policy, and change my repository's Makefile just for brew, it may be difficult to debug an issue.

Providing options for packagers is pretty well established as a practice.

> $(LIBEV_FILE) is just carried from
> https://github.com/docker/hyperkit/blob/master/Makefile#L83
> I would like to follow hyperkit as much as possible.

Allowing to to be overridden seems acceptable and I suspect upstream `hyperkit` would accept that too. If it isn't, you could remove the `qcow2` option entirely to remove one patch from here.

MikeMcQuaid:
@zchee What's the latest on this?

zchee:
@MikeMcQuaid I'll release v0.3.1 soon, will fix together.

MikeMcQuaid:
@zchee Great, thanks.

PR #6910, opened 2016-11-14
dmd use upstream VERSION file

John-Colvin:

-----

It was a bad idea to do this as it messes up HEAD and devel builds (version formatting for homebrew isn't necessarily the same as for dmd). Better to use a patch if a mistake is made upstream again.


ilovezfs:
Let's hope it doesn't happen again. Thanks for the PR.

PR #6928, opened 2016-11-15
Replace custom MysqlClientRequirement with the mysql requirement

Jose123456:

-----

Also added a trivial test

### DIFF
--- a/Formula/mysql-connector-c++.rb
+++ b/Formula/mysql-connector-c++.rb
@@ -1,9 +1,3 @@
-class MysqlClientRequirement < Requirement
-  fatal true
-  default_formula "mysql"
-  satisfy { which "mysql" }
-end
-
 class MysqlConnectorCxx < Formula
   desc "MySQL database connector for C++ applications"
   homepage "https://dev.mysql.com/downloads/connector/cpp/"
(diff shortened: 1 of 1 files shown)


Formula/mysql-connector-c++.rb
```diff
@@ -1,9 +1,3 @@
-class MysqlClientRequirement < Requirement
-  fatal true
-  default_formula "mysql"
-  satisfy { which "mysql" }
```
MikeMcQuaid: This is not quite equivalent as after this PR it'll now check for `mysql_config` instead. I'm fine with that but just so you're aware.

Jose123456: True, but this formula really only needs `mysql_config`. It does not need the full-blown mysql server, just a way to connect to one. I used the example [here](https://dev.mysql.com/doc/connector-cpp/en/connector-cpp-examples-complete-example-1.html) (with some minor modifications) to test this.

MikeMcQuaid:

For future reference the preferred commit message format is `mysql-connector-c++: use :mysql requirement.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #6940, opened 2016-11-15
Add dependency for Ruby 2.1

jslabovitz:
To ensure that correct version is present when --with-ruby option is given.


dunn:
Thanks! Merged it as 4940230a5d92b1f7d9791ac72edfd1dcd8e87f1c.

PR #6952, opened 2016-11-15
ruby 2.3.2

frodsan:

-----

### DIFF
--- a/Formula/ruby.rb
+++ b/Formula/ruby.rb
@@ -1,11 +1,10 @@
 class Ruby < Formula
   desc "Powerful, clean, object-oriented scripting language"
   homepage "https://www.ruby-lang.org/"
-  revision 2
 
   stable do
-    url "https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.1.tar.bz2"
-    sha256 "4a7c5f52f205203ea0328ca8e1963a7a88cf1f7f0e246f857d595b209eac0a4d"
+    url "https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.2.tar.bz2"
+    sha256 "e6ce83d46819c4120c9295ff6b36b90393dd5f6bef3bb117a06d7399c11fc7c0"
 
     # Reverts an upstream commit which incorrectly tries to install headers
     # into SDKROOT, if defined


dunn:
cc @MikeMcQuaid is this related to https://github.com/Homebrew/brew/commit/5e0b3d24e5f2a59bf715e34e8c0a62c1307d474d ?
```
        Error: 1 problem in 1 formula
ruby:
  * devel version should not decrease (from 2.4.0-preview1 to 2.4.0-beta3)
```

MikeMcQuaid:
@dunn That output is correct as `preview1` is deemed to be "newer" than `beta3` so people won't get `brew upgraded` from the prior to the latter.

ilovezfs:
So another retroactive need for version_scheme due to the new audit checks. Yuck.

MikeMcQuaid:
> So another retroactive need for version_scheme due to the new audit checks. Yuck.

Another hidden, user-affecting bug found by the new audit checks. We can either ignore the bug, fix the audit checks or fix the version detection (e.g. make `preview` detected as an alpha release if it's an established convention).

ilovezfs:
It is not user-affecting. The request to change that came from Martin literally minutes after the original push.

MikeMcQuaid:
> It is not user-affecting. The request to change that came from Martin literally minutes after the original push.

And if you upgraded in those minutes: you'd be an affected user.

ilovezfs:
I'll let that speak for itself.

MikeMcQuaid:
@ilovezfs I welcome suggestions and particularly PRs on improving these audit checks. If there's better logic for us to be taking, let's do it. It's unclear to me that it's immediately obvious what they are, though.

ilovezfs:
Reverting the false-positive-causing audit changes would be my preference. That, or running them against every existing formula, and maintaining a whitelist in perpetuity.

MikeMcQuaid:
> Reverting the false-positive-causing audit changes would be my preference. That, or running them against every existing formula, and maintaining a whitelist in perpetuity.

The issue is that the are run against the last 10 commits which means they are catching times where we've made mistakes in the past. I'm open to suggestions to what the default behaviour instead should be as e.g. just the last 1 commit will be insufficient in a PR with more than a single commit and comparing to `origin/master` is not the right behaviour when running these locally (although maybe locally they should be a no-op).

MikeMcQuaid:
> I'm open to suggestions to what the default behaviour instead should be as e.g. just the last 1 commit will be insufficient in a PR with more than a single commit and comparing to origin/master is not the right behaviour when running these locally (although maybe locally they should be a no-op).

I think this may be wrong, I think actually comparing just the last commit is sufficient.

ilovezfs:
Looking forward, yes theoretically if the audit check is always done, then just comparing the last should be sufficient. It won't help anyone stranded in a prior mistake but I don't think that was the original goal of the new checks.

ilovezfs:
Of course it assumes our version comparisons are all transitive, but we have much bigger problems if that's not the case LOL

MikeMcQuaid:
> Looking forward, yes theoretically if the audit check is always done, then just comparing the last should be sufficient. It won't help anyone stranded in a prior mistake but I don't think that was the original goal of the new checks.

Yes. PR opened in https://github.com/Homebrew/brew/pull/1522

ilovezfs:
@BrewTestBot test this please

ilovezfs:
```
Error: 1 problem in 1 formula
ruby:
  * devel revision should not decrease (from 2 to 0)
```
@MikeMcQuaid I assume we're just supposed to ignore those?

MikeMcQuaid:
@ilovezfs There was some previous code around this and it never quite made sense to me; there's no "devel revision". Yeh, just ignore it and I'll fix the audit.

ilovezfs:
@frodsan Thanks for your contribution to Homebrew! :fireworks:

frodsan:
Thanks! 👍

PR #6962, opened 2016-11-16
go: disable race detector if building without cgo

ilovezfs:
Prevents the following error when building without cgo and with race:
"go install: -race requires cgo; enable cgo by setting CGO_ENABLED=1"

CC @MikeMcQuaid

### DIFF
--- a/Formula/go.rb
+++ b/Formula/go.rb
@@ -30,7 +30,7 @@ class Go < Formula
     end
   end
 
-  option "without-cgo", "Build without cgo"
+  option "without-cgo", "Build without cgo (also disables race detector)"
   option "without-godoc", "godoc will not be installed for you"
   option "without-race", "Build without race detector"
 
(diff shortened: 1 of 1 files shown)


Formula/go.rb
```diff
@@ -45,6 +45,14 @@ class Go < Formula
   end
 
   def install
+    # Race detector only supported on amd64 platforms.
+    # https://golang.org/doc/articles/race_detector.html
+    build_race = MacOS.prefer_64_bit? && build.with?("race")
+    if build_race && build.without?("cgo")
+      opoo "race detector disabled due to --without-cgo"
```
MikeMcQuaid: I reckon we can avoid this message given the `option` text above.

ilovezfs:
@MikeMcQuaid PR refreshed.

MikeMcQuaid (approved):

PR #6832, opened 2016-11-11
goenv 20161028 (new formula)

josegonzalez:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #6980, opened 2016-11-16
goenv: move to correct directory.

woodruffw:

-----

#6832 seems to have dropped `goenv.rb` in the tap root, rather than in `Formula/`. This PR just moves it.

Edit: I also fixed the `url` and added `go` as a dep.

### DIFF
--- a/Formula/goenv.rb
+++ b/Formula/goenv.rb
@@ -1,7 +1,7 @@
 class Goenv < Formula
   desc "Go version management"
   homepage "https://github.com/syndbg/goenv"
-  url "https://github.com/syndbg/goenv/archive/20161028.tar.gz"
+  url "https://github.com/syndbg/goenv/archive/v20161028.tar.gz"
   sha256 "3f283537c2b4f64a7549a009361ac4da6115be6b2c45462b008a58f8366a8804"
   head "https://github.com/syndbg/goenv.git"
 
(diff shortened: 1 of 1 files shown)


woodruffw:
@MikeMcQuaid  The only thing causing `goenv` to fail now is the notability test. Since it was merged once before (#6832), I assume it's still fine to merge here?

(I think it might not have failed the notability test before because it was in the incorrect directory. )

Formula/goenv.rb
```diff
@@ -1,12 +1,14 @@
...
   desc "Go version management"
   homepage "https://github.com/syndbg/goenv"
-  url "https://github.com/syndbg/goenv/archive/20161028.tar.gz"
+  url "https://github.com/syndbg/goenv/archive/v20161028.tar.gz"
   sha256 "3f283537c2b4f64a7549a009361ac4da6115be6b2c45462b008a58f8366a8804"
   head "https://github.com/syndbg/goenv.git"
 
   bottle :unneeded
 
+  depends_on "go"
```
MikeMcQuaid: This isn't the case as it's used to build Go rather than using Go itself.

MikeMcQuaid:
> (I think it might not have failed the notability test before because it was in the incorrect directory. )

Yeh, it did. We should probably write tests to ensure things aren't added to the wrong directory like this again. The notability test is fine to ignore now we've included it.

woodruffw:
> We should probably write tests to ensure things aren't added to the wrong directory like this again.

I'll look into this and see if we can add something to `audit`.

woodruffw:
Test is failing because it tries to search for a list of Go versions already installed. I'll replace it with something more trivial.

MikeMcQuaid:
Thanks for the fix @woodruffw!

PR #7015, opened 2016-11-17
wu 3.10.0 (new)

chiefy:

-----

### DIFF
--- a/Formula/wu.rb
+++ b/Formula/wu.rb
@@ -1,0 +1,38 @@
+class Wu < Formula
+  desc "cli that retrieves weather data from Weather Underground."
+  homepage "https://github.com/sramsay/wu"
+  url "https://github.com/sramsay/wu/archive/3.10.0.tar.gz"
+  sha256 "0934771cdd04cb4c38b45190dbfa97fa8b3aafaf53ea75152f070c97de850223"
+
+  depends_on "go"
+
+  def install
+    ENV["GOPATH"] = buildpath
+    mkdir_p buildpath/"src/github.com/sramsay"
+    ln_sf buildpath, buildpath/"src/github.com/sramsay/wu"
+
+    cd "src/github.com/sramsay/wu" do
+      system "go", "build", "-o", bin/"wu", "-ldflags", "-X main.Version=#{version}"
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    In order for `wu` to work, you need to obtain a Weather Underground API key:
+
+      https://www.wunderground.com/weather/api/
+
+    You also need to create a `$HOME/.condrc`:
+    {
+      "key": "YOUR_API_KEY",
+      "station": "Lincoln, NE",
+      "degrees": "F"
+    }
+    EOS
+  end
+
+  test do
+    ENV["GOPATH"] = testpath.realpath
+    assert_match "You must create a .condrc file in $HOME.", shell_output("#{bin}/wu -all")
+    system "false"
+  end
+end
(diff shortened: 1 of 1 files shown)


Formula/wu.rb
```diff
@@ -0,0 +1,38 @@
+class Wu < Formula
+  desc "cli that retrieves weather data from Weather Underground."
+  homepage "https://github.com/sramsay/wu"
+  url "https://github.com/sramsay/wu/archive/3.10.0.tar.gz"
+  sha256 "0934771cdd04cb4c38b45190dbfa97fa8b3aafaf53ea75152f070c97de850223"
+
+  depends_on "go"
```
bfontaine: This depends on Go for the build only, so please use `depends_on "go" => :build`.

Formula/wu.rb
```diff
@@ -0,0 +1,38 @@
...
+      "station": "Lincoln, NE",
+      "degrees": "F"
+    }
+    EOS
+  end
+
+  test do
+    ENV["GOPATH"] = testpath.realpath
+    assert_match "You must create a .condrc file in $HOME.", shell_output("#{bin}/wu -all")
+    system "false"
```
bfontaine: You can remove this line.

Formula/wu.rb
```diff
@@ -0,0 +1,38 @@
...
+    {
+      "key": "YOUR_API_KEY",
+      "station": "Lincoln, NE",
+      "degrees": "F"
+    }
+    EOS
+  end
+
+  test do
+    ENV["GOPATH"] = testpath.realpath
```
bfontaine: You shouldn’t need this.

bfontaine (changes requested):

bfontaine (approved):

Formula/wu.rb
```diff
@@ -0,0 +1,36 @@
+class Wu < Formula
+  desc "cli that retrieves weather data from Weather Underground."
```
MikeMcQuaid: `CLI`

Formula/wu.rb
```diff
@@ -0,0 +1,36 @@
...
+    mkdir_p buildpath/"src/github.com/sramsay"
+    ln_sf buildpath, buildpath/"src/github.com/sramsay/wu"
+
+    cd "src/github.com/sramsay/wu" do
+      system "go", "build", "-o", bin/"wu", "-ldflags", "-X main.Version=#{version}"
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    In order for `wu` to work, you need to obtain a Weather Underground API key:
```
MikeMcQuaid: I don't think we need these caveats as none of them are Homebrew-specific information.

MikeMcQuaid (changes requested):

Formula/wu.rb
chiefy: @MikeMcQuaid how else should I inform the installer that there are these requirements?

MikeMcQuaid: You should not. They can read the instructions prominently at the top of the homepage: https://github.com/sramsay/wu

chiefy: @MikeMcQuaid removed. Unless I missed it, might be cool to have something in the docs about what you should / shouldn't display to the user?

MikeMcQuaid: Where would be most useful, http://www.rubydoc.info/github/Homebrew/brew/master/Formula#caveats-instance_method?

MikeMcQuaid (approved):

MikeMcQuaid:

PR #7019, opened 2016-11-17
qt: migrate to boneyard

MikeMcQuaid:
Created with `brew boneyard-formula-pr` because Qt4 is EOL.

Fixes https://github.com/Homebrew/homebrew-core/issues/1705.


ilovezfs (approved):

MikeMcQuaid:
@ilovezfs Note there's also an alias change and making qt5 not keg-only in here.

ilovezfs:
Is it valid to have an alias that is equal to the same name as something still in another homebrew tap but referring to a different formula? Seems like we'll also need to rename the qt.rb in boneyard qt4.rb, or just delete it outright.

MikeMcQuaid:
> Is it valid to have an alias that is equal to the same name as something still in another homebrew tap but referring to a different formula?

I'll test it out at some point and see what happens.

MikeMcQuaid:
Merging into https://github.com/Homebrew/homebrew-core/pull/8306.

MikeMcQuaid:
> I'll test it out at some point and see what happens.

I did, seems to work fine but I'm going to just remove the `tap_migrations` when pulling https://github.com/Homebrew/homebrew-core/pull/8306.

MikeMcQuaid:
and, actually, not commit to boneyard to avoid clashes.

scpeters:
This is just an FYI, not a complaint: this PR has broken some formulae in a tap I maintain (`osrf/simulation`) that use qt4, so I will probably fork the qt4 formula into that tap if it isn't in the boneyard. We are almost finished migrating the latest version of our software to use qt5 but want to support users on old versions as well.

Thanks for all your work maintaining homebrew!

ilovezfs:
It's been relocated to https://github.com/cartr/homebrew-qt4/blob/master/qt.rb

PR #7057, opened 2016-11-18
infer 0.9.4

jvillard:

-----

### DIFF
--- a/Formula/infer.rb
+++ b/Formula/infer.rb
@@ -1,8 +1,8 @@
 class Infer < Formula
   desc "Static analyzer for Java, C and Objective-C"
   homepage "http://fbinfer.com/"
-  url "https://github.com/facebook/infer/releases/download/v0.9.2/infer-osx-v0.9.2.tar.xz"
-  sha256 "3935f8be25982a023aba306b66804d73a7316ab833296277c1ec6c3694bfc7c7"
+  url "https://github.com/facebook/infer/releases/download/v0.9.4/infer-osx-v0.9.4.tar.xz"
+  sha256 "529d147bccf3285ddb7500c22e0c50d6e0cbdb2c7f9b11a84e8005873994b3e2"
 
   bottle do
     cellar :any
@@ -24,7 +24,7 @@ class Infer < Formula
       odie "infer: --without-clang and --without-java are mutually exclusive"
     end
 
-    opamroot = buildpath/"build"
+    opamroot = buildpath/"opamroot"
     opamroot.mkpath
     ENV["OPAMROOT"] = opamroot
     ENV["OPAMYES"] = "1"
--- a/Formula/infer.rb
+++ b/Formula/infer.rb
@@ -47,7 +47,7 @@ class Infer < Formula
 
     system "opam", "init", "--no-setup"
     ocaml_version = File.read("build-infer.sh").match(/OCAML_VERSION=\"([0-9\.]+)\"/)[1]
-    inreplace "opamroot/compilers/#{ocaml_version}/#{ocaml_version}/#{ocaml_version}.comp",
+    inreplace "#{opamroot}/compilers/#{ocaml_version}/#{ocaml_version}/#{ocaml_version}.comp",
       '["./configure"', '["./configure" "-no-graph"'
     system "./build-infer.sh", target_platform, "--yes"
     system "opam", "config", "exec", "--switch=infer-#{ocaml_version}", "--", "make", "install"
--- a/Formula/infer.rb
+++ b/Formula/infer.rb
@@ -17,6 +17,7 @@ class Infer < Formula
   depends_on "autoconf" => :build
   depends_on "automake" => :build
   depends_on "libtool" => :build
+  depends_on "ocaml" => :build
   depends_on "opam" => :build
 
   def install
@@ -44,8 +45,12 @@ class Infer < Formula
       "all"
     end
 
+    system "opam", "init", "--no-setup"
+    ocaml_version = File.read("build-infer.sh").match(/OCAML_VERSION=\"([0-9\.]+)\"/)[1]
+    inreplace "opamroot/compilers/#{ocaml_version}/#{ocaml_version}/#{ocaml_version}.comp",
+      '["./configure"', '["./configure" "-no-graph"'
     system "./build-infer.sh", target_platform, "--yes"
-    system "opam", "config", "exec", "--switch=infer-4.02.3", "--", "make", "install"
+    system "opam", "config", "exec", "--switch=infer-#{ocaml_version}", "--", "make", "install"
   end
 
   test do


jvillard:
I am confused as to why the checks failed:
- brew reports `X11: 2.7.10 => /opt/X11` in its "Configuration" section
- but `opam init` fails to compile the OCaml compiler itself because: `./libgraph.h:15:10: fatal error: 'X11/Xlib.h' file not found` ([logs](https://bot.brew.sh/job/Homebrew%20Core/11702/version=any_sierra/testReport/junit/brew-test-bot/sierra/install_infer/))

I cannot reproduce the issue locally.

Any idea how to solve this would be greatly appreciated. I've tried looking at other formulas that use `opam` but I don't see anything different in their setups.

cc @akotulski who's been looking into this with me

ilovezfs:
@jberdine does infer now have an `:x11` requirement? If so, it's likely headed for homebrew/x11 or homebrew/boneyard unfortunately.

akotulski:
@ilovezfs It doesn't as far as I know. Notice that failure happened during ocaml compilation, before we try to install any of infer dependencies.

ilovezfs:
I see. Well that should be solvable since this has no business compiling ocaml in the first place. It should use `depends_on "ocaml" => :build`.

jvillard:
@ilovezfs do you mean that Infer should depend on ocaml in addition to opam?

opam itself already depends on ocaml (with `depends_on "ocaml" => :recommended`), but actually Infer wants its own version of the ocaml compiler (4.02.3) that it installs using opam.

ilovezfs:
>depend on ocaml in addition to opam?

yes

>wants its own version of the ocaml compiler (4.02.3) that it installs using opam.

that should not be necessary

jvillard:
@ilovezfs Homebrew has OCaml 4.03.0 which is not currently supported by infer, so rely on opam to get another compiler version.

ilovezfs:
>which is not currently supported by infer

why is this the case?

jvillard:
Some of Infer's dependencies do not support 4.03.0 yet, in particular Reason.

ilovezfs:
I see https://github.com/facebook/reason/pull/604. Yuck.

jvillard:
I see what happens now: Infer installs OCaml via opam, which by default includes Graphics support, but that relies on x11. So it looks like we should tell opam to install OCaml without Graphics somehow.

ilovezfs:
Yup, ideally fixed upstream at least in a commit we can apply as a patch.

It would also be nice to have an issue filed in the infer issue tracker that it stop compiling its own OCaml as soon as https://github.com/facebook/reason/pull/604 ships.

On an unrelated note, I noticed that once it does build successfully (I threw a `depends_on :x11` at it), there's some warnings emanating from the test block right now.

jvillard:
Actually opam has no compiler switch for OCaml without x11, so it looks like we're stuck having to depend on x11 for now. Would that be ok?

The warnings in the test are harmless, they come from the library used to parse Java (javalib).

ilovezfs:
>so it looks like we're stuck having to depend on x11 for now. Would that be ok?

Nope, not without relocating this to homebrew/x11, which would be very unfortunate since that tap is likely to be deprecated entirely soon.

akotulski:
Is there a reason why in `Configuration` section of the error log there is mention for `X11 -> /opt/X11` even though there is no `X11`?

I think this is what may have tripped `opam` when building custom ocaml: it's trying to determine `-x11include` option to configure script automatically:

https://github.com/ocaml/ocaml/blob/trunk/INSTALL.adoc

akotulski:
And this is what they mean by "automatically":
https://github.com/ocaml/ocaml/blob/trunk/configure#L1614-L1754

ilovezfs:
@akotulski ah that sounds like the problem. We have XQuartz installed on the build machines so that cases where X11 gets used opportunistically like that get smoked out where `depends_on :x11` isn't used. Luckily there is a way to disable, which is already part of the ocaml formula:

https://github.com/Homebrew/homebrew-core/blob/7a4dabfc1a2acd9f01a1670fde4f0094c4fb6ffa/Formula/ocaml.rb#L46

jvillard:
I'm not sure I understand how this solution applies to the problem we have: opam builds OCaml for our infer-4.02.3 switch from its own pre-defined set of options, which don't include passing `-no-graph` to OCaml. If I understand the situation correctly, as long as OCaml thinks it found x11 at the time our build script calls opam, OCaml will try to build as if x11 was available.

akotulski:
On the bright side, it may mean that the only place where it will fail is your CI machines - elsewhere users either have working X11 or they don't have X11 at all

ilovezfs:
>I'm not sure I understand how this solution applies to the problem we have: opam builds OCaml for our infer-4.02.3 switch from its own pre-defined set of options, which don't include passing -no-graph to OCaml

You'll need to get opam to pass that flag to fix the build. The quick and dirty way would be to use opam pin.

>On the bright side, it may mean that the only place where it will fail is your CI machines

Nope. It will fail if the users have X11 just as it does during CI.

ilovezfs:
This should work:
```
diff --git a/Formula/infer.rb b/Formula/infer.rb
index c76db9c..66ec259 100644
--- a/Formula/infer.rb
+++ b/Formula/infer.rb
@@ -18,6 +18,7 @@ class Infer < Formula
   depends_on "opam" => :build
+  depends_on "ocaml" => :build
 
   def install
     if build.without?("clang") && build.without?("java")
@@ -44,6 +45,11 @@ class Infer < Formula
       "all"
     end
 
+    system "opam", "init", "--no-setup"
+    ocaml_version = File.read("build-infer.sh").match(/OCAML_VERSION=\"[0-9\.]+\"/)[0].split("=").last.delete('"')
+    inreplace "opamroot/compilers/#{ocaml_version}/#{ocaml_version}/#{ocaml_version}.comp",
+      '["./configure" "-prefix" prefix "-with-debug-runtime"]',
+      '["./configure" "-prefix" prefix "-with-debug-runtime" "-no-graph"]'
     system "./build-infer.sh", target_platform, "--yes"
     system "opam", "config", "exec", "--switch=infer-4.02.3", "--", "make", "install"
   end
```

ilovezfs:
And these are the warnings I mentioned
```
iMac-TMP:homebrew-core joe$ brew test -vd infer
/usr/local/Homebrew/Library/Homebrew/brew.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/infer.rb
Testing infer
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20161119-9843-y66mhy.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/infer.rb -vd
/usr/local/Homebrew/Library/Homebrew/test.rb (Formulary::FromPathLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/infer.rb
==> /usr/local/Cellar/infer/0.9.4/bin/infer --fail-on-bug -- clang FailingTest.c
F.==> /usr/local/Cellar/infer/0.9.4/bin/infer --fail-on-bug -- clang PassingTest.c
F.==> /usr/local/Cellar/infer/0.9.4/bin/infer --fail-on-bug -- javac FailingTest.java
Warning: unexpected attribute found: Code
Warning: unexpected attribute found: Code
Warning: unexpected attribute found: Code
Warning: unexpected attribute found: Code
F...==> /usr/local/Cellar/infer/0.9.4/bin/infer --fail-on-bug -- javac PassingTest.java
Warning: unexpected attribute found: Code
Warning: unexpected attribute found: Code
Warning: unexpected attribute found: Code
Warning: unexpected attribute found: Code
F...iMac-TMP:homebrew-core joe$
```

ilovezfs:
yup that worked. :green_apple:

jvillard:
Wow, thanks!

ilovezfs:
shipped!

PR #7065, opened 2016-11-18
quantlib 1.9: fix flag for intra-day times

coatless:

-----

Adds a flag, `--enable-intraday`, to the formula that enables 
intra-day data type during the installation of quantlib. Specifically,
the date specifications gain hours, minutes, seconds, milliseconds,
and microseconds. Related to issue in eddelbuettel/rquantlib#80
that warns on lack of intraday support.

### DIFF
--- a/Formula/quantlib.rb
+++ b/Formula/quantlib.rb
@@ -29,10 +29,9 @@ class Quantlib < Formula
   end
 
   def install
-   
     args = []
     if build.with? "intraday"
-        args << "--enable-intraday"
+      args << "--enable-intraday"
     end
 
     ENV.cxx11 if build.cxx11?
--- a/Formula/quantlib.rb
+++ b/Formula/quantlib.rb
@@ -29,15 +29,12 @@ class Quantlib < Formula
   end
 
   def install
-    args = []
-    if build.with? "intraday"
-      args << "--enable-intraday"
-    end
-
     ENV.cxx11 if build.cxx11?
     (buildpath/"QuantLib").install buildpath.children if build.stable?
     cd "QuantLib" do
       system "./autogen.sh" if build.head?
+      args = []
+      args << "--enable-intraday" if build.with? "intraday"
       system "./configure", "--disable-dependency-tracking",
                             "--prefix=#{prefix}",
                             "--with-lispdir=#{elisp}",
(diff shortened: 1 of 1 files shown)


Formula/quantlib.rb
```diff
@@ -28,13 +29,20 @@ class Quantlib < Formula
   end
 
   def install
+    args = []
```
MikeMcQuaid: Put the definition down just above `configure`

Formula/quantlib.rb
```diff
@@ -28,13 +29,20 @@ class Quantlib < Formula
...
+    if build.with? "intraday"
+      args << "--enable-intraday"
```
MikeMcQuaid: `args << "--enable-intraday" if build.with? "intraday"`

Formula/quantlib.rb
```diff
@@ -20,6 +20,7 @@ class Quantlib < Formula
   end
 
   option :cxx11
+  option "with-intraday", "Enable intraday components to dates"
```
MikeMcQuaid: Would it make sense to enable this by default? If so, would anyone want/need to remove support? If not, let's remove the option.

MikeMcQuaid (changes requested):

Formula/quantlib.rb
coatless: Enabling by default would not necessarily make sense as this was added later in QuantLib's life (v 1.7) and is still an "opt-in" feature. Primarily, some code may not have support for it in earlier quant software versions. Thus, it would be preferred to keep it as opted in and downstream can add documentation as to how to install (e.g.  https://github.com/eddelbuettel/rquantlib/wiki/RQuantLib )

MikeMcQuaid (approved):

MikeMcQuaid:

For future reference the preferred commit message format is `quantlib: add option for intra-day times`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

coatless:
@MikeMcQuaid: Thanks! I'll make sure the commit message format is better the next time around.

PR #7067, opened 2016-11-19
pike 8.0.276

hww3:
new stable version of pike

### DIFF
--- a/Formula/pike.rb
+++ b/Formula/pike.rb
@@ -1,9 +1,8 @@
 class Pike < Formula
   desc "Dynamic programming language"
   homepage "https://pike.lysator.liu.se"
-  url "https://pike.lysator.liu.se/pub/pike/all/7.8.866/Pike-v7.8.866.tar.gz"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Please fill out the steps in the PR template.

Installation is failing on all platforms: https://bot.brew.sh/job/Homebrew%20Core/11716/version=any_sierra/testReport/junit/brew-test-bot/sierra/install_pike/

Formula/pike.rb
```diff
@@ -56,10 +55,6 @@ def install
       args << "--with-abi=32"
     end
 
-    if build.without? "machine-code"
```
MikeMcQuaid: If this option was removed then `option "with-machine-code", "Enables machine code"` should be removed above, too.

MikeMcQuaid (changes requested):

hww3:
There appears to be a build problem when both X11 and the MacOS SDK are installed (as is the case on the build farm): 2 versions of GL headers. I'll get a patch committed upstream.

PR #7078, opened 2016-11-19
clipper: fix service not allowing to configure address and port from a config file

aruiz14:
- Yes, but it lacks any test as mentioned in https://github.com/Homebrew/homebrew-core/pull/3832#issuecomment-239514897
-----
Clipper allows configuring the address and port for the service in a ~/.clipper.json file, but the parameters passed from the command line (from the .plist) take precedence. As the used values are the default ones in Clipper if not specified, this change should not cause any issue for people without a config file.

### DIFF
--- a/Formula/clipper.rb
+++ b/Formula/clipper.rb
@@ -38,10 +38,6 @@ class Clipper < Formula
       <key>ProgramArguments</key>
       <array>
         <string>#{opt_bin}/clipper</string>
-        <string>--address</string>
-        <string>127.0.0.1</string>
-        <string>--port</string>
-        <string>8377</string>
       </array>
       <key>EnvironmentVariables</key>
       <dict>


woodruffw:
Since this changes the plist we include in bottles, could you add a `rebuild 1` inside the `bottle` block for this formula?

For reference, an example of this is [a2ps](https://github.com/Homebrew/homebrew-core/blob/master/Formula/a2ps.rb#L10).

MikeMcQuaid:
> Since this changes the plist we include in bottles, could you add a rebuild 1 inside the bottle block for this formula?

@woodruffw This isn't actually needed; this happens automatically when you `brew pull --bottle` and it'll pull a new bottle with a new `rebuild`.

MikeMcQuaid:

PR #7106, opened 2016-11-21
Rust: Add a Racer option

CamJN:

-----

When the `--with-racer option` is used, the rust source is retained.

Previous discussions:
https://github.com/Homebrew/legacy-homebrew/pull/48377
https://github.com/Homebrew/homebrew-core/pull/1518

### DIFF
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -11,6 +11,11 @@ class Rust < Formula
       url "https://github.com/rust-lang/cargo.git", :tag => "0.14.0", :revision => "eca9e159b6b0d484788ac757cf23052eba75af55"
     end
 
+    resource "racer" do
+      url "https://github.com/phildawes/racer/archive/2.0.3.tar.gz"
+      sha256 "0396ce9e8535ecb821d556e40758ce5dc2ba37fcfa6f96d6caa7d1a9a88acba8"
+    end
+
     # name includes date to satisfy cache
     resource "cargo-nightly-2015-09-17" do
       url "https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-09-17/cargo-nightly-x86_64-apple-darwin.tar.gz"
@@ -33,6 +38,7 @@ class Rust < Formula
   end
 
   option "with-llvm", "Build with brewed LLVM. By default, Rust's LLVM will be used."
+  option "with-racer", "Build Racer code completion tool, and retain Rust sources."
 
   depends_on "cmake" => :build
   depends_on "pkg-config" => :run
@@ -77,6 +83,21 @@ class Rust < Formula
       system "./configure", "--prefix=#{prefix}", "--local-rust-root=#{prefix}", "--enable-optimize"
       system "make"
       system "make", "install"
+    end
+
+    if build.with? "racer"
+      resource("racer").stage do
+        ENV.prepend_path "PATH", bin
+        cargo_home = buildpath/"cargo_home"
+        cargo_home.mkpath
+        ENV["CARGO_HOME"] = cargo_home
+        system bin/"cargo", "build", "--release", "--verbose"
+        (libexec/"bin").install "target/release/racer"
+        (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src")
+      end
+      # Remove any binary files; as Homebrew will run ranlib on them and barf.
+      rm_rf Dir["src/{llvm,test,librustdoc,etc/snapshot.pyc}"]
+      (pkgshare/"rust_src").install Dir["src/*"]
     end
 
     rm_rf prefix/"lib/rustlib/uninstall.sh"


aisk:
Why not just install racer via cargo?

CamJN:
Well, that's why it's opt-in.

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
       system "make", "install"
     end
 
+    if build.with? "racer"
+      resource("racer").stage do
+        (buildpath/"cargo_home").mkdir
+        ENV["PATH"] = ENV["PATH"]+":"+bin
```
ilovezfs:
```
ENV.append_path "PATH", bin
```

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
       system "make", "install"
     end
 
+    if build.with? "racer"
+      resource("racer").stage do
+        (buildpath/"cargo_home").mkdir
```
ilovezfs:
```
cargo_home = buildpath/"cargo_home"
cargo_home.mkpath
```

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
...
+        ENV["PATH"] = ENV["PATH"]+":"+bin
+        ENV["CARGO_HOME"] = buildpath/"cargo_home"
```
ilovezfs:
```
ENV["CARGO_HOME"] = cargo_home
```

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
...
+        # interactive_shell
+        system (bin/"cargo"), "build", "--release", "--verbose"
```
ilovezfs:
```
system bin/"cargo", "build", "--release", "--verbose"
```

Formula/rust.rb
```diff
@@ -33,6 +39,7 @@ class Rust < Formula
   end
 
   option "with-llvm", "Build with brewed LLVM. By default, Rust's LLVM will be used."
+  option "with-racer", "Builds racer code completion tool, and retains rust sources."
```
ilovezfs: Build not Builds

Formula/rust.rb
```diff
@@ -1,6 +1,7 @@
 class Rust < Formula
   desc "Safe, concurrent, practical language"
   homepage "https://www.rust-lang.org/"
+  revision 1
```
ilovezfs: if it's a non-default option, there's no reason to revision bump

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
...
+        (buildpath/"cargo_home").mkdir
+        ENV["PATH"] = ENV["PATH"]+":"+bin
+        ENV["CARGO_HOME"] = buildpath/"cargo_home"
+        # interactive_shell
+        system (bin/"cargo"), "build", "--release", "--verbose"
+        (libexec/"bin").install "target/release/racer"
+        (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src/current")
+      end
+      rm_rf Dir["src/{llvm,test,librustdoc,etc/snapshot.pyc}"]
+      (pkgshare/"rust_src/#{version}").install Dir["src/*"]
```
ilovezfs: this path is likely to be strange for head builds

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
...
+      resource("racer").stage do
+        (buildpath/"cargo_home").mkdir
+        ENV["PATH"] = ENV["PATH"]+":"+bin
+        ENV["CARGO_HOME"] = buildpath/"cargo_home"
+        # interactive_shell
+        system (bin/"cargo"), "build", "--release", "--verbose"
+        (libexec/"bin").install "target/release/racer"
+        (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src/current")
+      end
+      rm_rf Dir["src/{llvm,test,librustdoc,etc/snapshot.pyc}"]
```
ilovezfs: what precisely is left over after deleting these? maybe we can select that directly instead of indirectly

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
...
 
+    if build.with? "racer"
+      resource("racer").stage do
+        (buildpath/"cargo_home").mkdir
+        ENV["PATH"] = ENV["PATH"]+":"+bin
+        ENV["CARGO_HOME"] = buildpath/"cargo_home"
+        # interactive_shell
+        system (bin/"cargo"), "build", "--release", "--verbose"
+        (libexec/"bin").install "target/release/racer"
+        (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src/current")
```
ilovezfs: what is the goal of using "current" as opposed to using `opt_pkgshare`?

ilovezfs: actually, even opt_pkgshare shouldn't be needed since the sources are living together with rust here. seems like you can use the actual path, no?

ilovezfs:
I'm wondering if we should make this the default and see how many people care enough to try to opt out, as opposed to forcing a bunch of lengthy rust source builds on people who want this.

Formula/rust.rb
ilovezfs: I'd suggest making this `without-default` for now so that CI actually tests it, regardless of whether the final decision is that we make it non-default.

Formula/rust.rb
```diff
@@ -95,5 +122,8 @@ def install
     system "#{bin}/cargo", "new", "hello_world", "--bin"
     assert_equal "Hello, world!",
                  (testpath/"hello_world").cd { `#{bin}/cargo run`.split("\n").last }
+    if build.with? "racer"
+      system bin/"racer", "complete", "std::io::B"
```
ilovezfs: can we actually match something here so that we're testing something other than a 0 exit code?

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
       system "make", "install"
     end
 
+    if build.with? "racer"
+      resource("racer").stage do
+        (buildpath/"cargo_home").mkdir
+        ENV["PATH"] = ENV["PATH"]+":"+bin
+        ENV["CARGO_HOME"] = buildpath/"cargo_home"
+        # interactive_shell
```
ilovezfs: what is this referring to exactly?

ilovezfs:
Nice work!

Formula/rust.rb
```diff
@@ -79,10 +86,30 @@ def install
...
+      (pkgshare/"rust_src/#{version}").install Dir["src/*"]
+    end
+
     rm_rf prefix/"lib/rustlib/uninstall.sh"
     rm_rf prefix/"lib/rustlib/install.log"
   end
 
+  def post_install
+    if build.with? "racer"
+      (pkgshare/"rust_src").install_symlink version.to_s => "current"
```
ilovezfs: why does this need to be done in post install?

Formula/rust.rb
MikeMcQuaid: Regardless, remove this comment.

Formula/rust.rb
CamJN: A test matching something will break periodically when rust stdlib changes without actually indicating that racer doesn't work.

MikeMcQuaid: This test will never be run by CI so is probably worth removing anyway.

ilovezfs: It will be if we make this the default, as I suggested.

Formula/rust.rb
MikeMcQuaid: I'm 👍 on giving it a run through CI but 👎 on having it as default yet.

ilovezfs: Why? The rust builds take _forever_. We could even make this non-optional.

ilovezfs: I assume you'll be the one testing this then.

ilovezfs: Also worth noting in case it slipped your mind: https://github.com/Homebrew/homebrew-core/pull/6825#issuecomment-260114533

Formula/rust.rb
```diff
@@ -88,28 +87,23 @@ def install
...
+        ENV.append_path "PATH", bin
+        cargo_home = buildpath/"cargo_home"
+        cargo_home.mkpath
+        ENV["CARGO_HOME"] = cargo_home
+        system bin/"cargo", "build", "--release", "--verbose"
         (libexec/"bin").install "target/release/racer"
-        (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src/current")
+        (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src")
       end
+      # remove any binary files as they break homebrew, keep everything else
```
ilovezfs: why would they "break homebrew"?

CamJN: Something to do with ranlib freaking out when you install binary files that don't have symbols, and not being allowed to ignore it. I've run into this before too, eg first two comments here: https://github.com/Homebrew/legacy-homebrew/issues/45151

MikeMcQuaid: Is this a problem with this specific formula? Regardless, the comment could be more specific i.e. `remove libraries with no symbols`.

Formula/rust.rb
```diff
@@ -33,6 +38,7 @@ class Rust < Formula
   end
 
   option "with-llvm", "Build with brewed LLVM. By default, Rust's LLVM will be used."
+  option "without-racer", "Build Racer code completion tool, and retains Rust sources."
```
MikeMcQuaid: Now that we've had a 💚 CI run on this, let's switch it back to `with-racer`.

MikeMcQuaid (changes requested):

Formula/rust.rb
CamJN: They need not be libraries to cause the problem. It's any binary file at all that goes through the "install" procedure.

MikeMcQuaid:
@CamJN Can you note in the comment that this is only binaries/library files with no symbols that are removed? Out of interest, why are these being built with no symbols anyway?

CamJN:
They aren't all binaries or library files that's why they have no symbols. For example font files like `src/librustdoc/html/static/SourceSerifPro-Regular.woff` and stuff are binary files and so homebrew runs ranlib against them, but no-one expects symbols in a font file.

MikeMcQuaid:
@CamJN Where are/were they installed? This sounds like a Homebrew/brew bug that would be good to explore if they aren't (incorrectly) being installed into lib or bin.

CamJN:
They would get installed to the rust source dir because they are source files `(pkgshare/"rust_src").install Dir["src/*"]`.

MikeMcQuaid (approved):

MikeMcQuaid:
Ok, definitely a bug. Mind filing one to Homebrew/brew? I'm good with this now if @ilovezfs is.

Formula/rust.rb
```diff
@@ -11,6 +11,11 @@ class Rust < Formula
       url "https://github.com/rust-lang/cargo.git", :tag => "0.14.0", :revision => "eca9e159b6b0d484788ac757cf23052eba75af55"
     end
 
+    resource "racer" do
+      url "https://github.com/phildawes/racer/archive/1.2.5.tar.gz"
```
ilovezfs: I'm a bit confused by the released here: https://github.com/phildawes/racer/releases

1.1.0 is marked as latest and 1.2.5 is just a tag. But in the just-a-tag category, there's also 2.0.1, 2.0.2, and 2.0.3. Do we have to wait for https://github.com/Homebrew/homebrew-core/pull/8138 to get fixed before using 2.0.x? And if that's the case, what's the rationale for 1.2.5 over 1.1.0?

Formula/rust.rb
```diff
@@ -33,6 +38,7 @@ class Rust < Formula
   end
 
   option "with-llvm", "Build with brewed LLVM. By default, Rust's LLVM will be used."
+  option "with-racer", "Build Racer code completion tool, and retains Rust sources."
```
ilovezfs: s/retains/retain/

Formula/rust.rb
```diff
@@ -79,6 +85,21 @@ def install
...
+      resource("racer").stage do
+        ENV.append_path "PATH", bin
+        cargo_home = buildpath/"cargo_home"
+        cargo_home.mkpath
+        ENV["CARGO_HOME"] = cargo_home
+        system bin/"cargo", "build", "--release", "--verbose"
+        (libexec/"bin").install "target/release/racer"
+        (bin/"racer").write_env_script(libexec/"bin/racer", :RUST_SRC_PATH => pkgshare/"rust_src")
+      end
+      # remove any binary files as homebrew will run ranlib on them and barf, keep everything else
```
ilovezfs: please wrap the comment at the eightieth column, capitalize Homebrew, and avoid the run-on sentence

Formula/rust.rb
```diff
@@ -79,6 +85,21 @@ def install
       system "make", "install"
     end
 
+    if build.with? "racer"
+      resource("racer").stage do
+        ENV.append_path "PATH", bin
```
ilovezfs: Does it fail without this? It looks like the calls below are fully qualified paths.

Assuming it is needed, probably want prepend not append.

ilovezfs (approved):
I added some minor blah blah, but LGTM.

ilovezfs:
@CamJN thanks! This is great work. We appreciate your patience too.

CamJN:
No problem and I'll get to that bug report once I have more time to write it up in a decent manner.

MikeMcQuaid:
Thanks @CamJN from me too.

PR #7131, opened 2016-11-21
flow: remove --old-output-format from test

mroch:
Flow is removing the `--old-output-format` flag in the next release (https://github.com/facebook/flow/issues/2844).

We're also planning on changing the error messages further in the future. Simplifying this test should make it future-proof, while still being pretty confident that flow is working.


woodruffw:
Thanks for catching this before it became a failure, @mroch!

PR #7132, opened 2016-11-21
qwtpolar: use qt5

ilovezfs:
CC @MikeMcQuaid

### DIFF
--- a/Formula/qwtpolar.rb
+++ b/Formula/qwtpolar.rb
@@ -1,0 +1,82 @@
+class Qwtpolar < Formula
+  desc "Library for displaying values on a polar coordinate system"
+  homepage "http://qwtpolar.sourceforge.net/"
+  url "https://downloads.sf.net/project/qwtpolar/qwtpolar/1.1.0/qwtpolar-1.1.0.tar.bz2"
+  sha256 "e45a1019b481f52a63483c536c5ef3225f1cced04abf45d7d0ff8e06d30e2355"
+
+  bottle do
+    cellar :any
+    sha256 "e51aec713366e7406d63b0eea55f41385a24d67ff7d298f9fd479ae14dea2e3c" => :el_capitan
+    sha256 "8bd14ade82bd28887ec4bfef8098cd893806b380b2176eae6885ec5da4168a54" => :yosemite
+    sha256 "cb47115b5ca12d61ccc63a2cd681323825f6a73001a02f62a1f58c8a920ae82d" => :mavericks
+  end
+
+  option "with-examples", "Install source code for example apps"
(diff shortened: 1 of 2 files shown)


Formula/qwtpolar.rb
```diff
@@ -0,0 +1,94 @@
...
+    sha256 "cb47115b5ca12d61ccc63a2cd681323825f6a73001a02f62a1f58c8a920ae82d" => :mavericks
+  end
+
+  option "without-plugin", "Skip building the Qt Designer plugin"
+
+  depends_on "qt5"
+  depends_on "qwt"
+
+  # Update designer plugin linking back to qwtpolar framework/lib after install
+  # See: https://sourceforge.net/p/qwtpolar/patches/2/
```
MikeMcQuaid: This was submitted over 2 years ago with no response so seems to be basically unmaintained at this point. Maybe leave this open for a while to see if there's any requests to re-add?

ilovezfs: if you check the analytics, it appears to have been used quite a bit (and still is out of a third party tap). also the most recent commit is from April 2016 https://sourceforge.net/p/qwtpolar/code/commit_browser

MikeMcQuaid: Looks like from that that there was a 1.2.0 release. Wonder why we never saw a tarball? I'm fine with reinstating given that but would be good to submit some of this stuff upstream and see if we can nag them into a release.

ilovezfs: I believe the latest release is 1.1.1. Before boneyarding, we were at 1.1.0. In this PR, I bumped it to 1.1.1. It looks like 1.1.1 and 1.2.0 were created simultaneously, 1.1.1 in the 1.1 branch, and 1.2.0 in trunk. We could apply revisions 173 - 184 as patches, or build from svn and designate r184 as 1.2.0-beta1. The most significant change is probably the most recent revision which updated the colormap with
```
--- a/trunk/qwtpolar/examples/spectrogram/plotwindow.cpp
+++ b/trunk/qwtpolar/examples/spectrogram/plotwindow.cpp
@@ -13,7 +13,12 @@
     ColorMap():
         QwtLinearColorMap( Qt::darkBlue, Qt::yellow )
     {
+#if 1
+        addColorStop( 0.00, Qt::black );
+        addColorStop( 0.05, Qt::darkRed );
+#else
         addColorStop( 0.05, Qt::blue );
+#endif
         addColorStop( 0.3, Qt::cyan );
         addColorStop( 0.6, Qt::green );
         addColorStop( 0.98, Qt::red );
```

MikeMcQuaid: Sounds good 👍

PR #7134, opened 2016-11-21
qt5: fix pkg-config

stefanha:

-----

Commit 77fd56f4a96b5deb8fc0139da6519e207ec43fee ("qt5: remove spurious
pkg-config dependency") dropped the pkg-config build dependency because the
missing ./configure --pkg-config option caused Qt to be built without
pkg-config support.

pkg-config makes it easy to link Qt applications against external
libraries.  Many libraries available in Homebrew ship with pkg-config .pc
files so it makes sense to enable this feature.

This patch re-enables pkg-config so qmake CONFIG += link_pkgconfig works as
originally intended in commit b4df7fce96791c72e8fb4c84a89258eea990117a ("qt5:
add pkg-config build dependency").

Signed-off-by: Stefan Hajnoczi <stefanha@gmail.com>

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -91,6 +91,7 @@ class Qt5 < Formula
 
   depends_on "dbus" => :optional
   depends_on :mysql => :optional
+  depends_on "pkg-config" => :build
   depends_on :postgresql => :optional
   depends_on :xcode => :build
 
@@ -128,6 +129,7 @@ class Qt5 < Formula
       -qt-pcre
       -nomake tests
       -no-rpath
+      -pkg-config
     ]
 
     args << "-nomake" << "examples" if build.without? "examples"


ilovezfs:
How does this interact with https://github.com/stefanha/homebrew-core/blob/49c98c66ac7cb99511adc24a97c18c8ccfec80ec/Formula/qt5.rb#L106-L117

stefanha:
@ilovezfs I believe they are orthogonal: link_pkgconfig allows Qt applications to build against external libraries using pkg-config whereas create_pc allows Qt frameworks/libraries to generate .pc files so third-party applications can link against your framework/library.

It looks like ./configure -pkg-config is not needed by CONFIG create_pc so I don't think anything will change.

ilovezfs:
ah ok, thanks for clarifying - this looks good to me

is it really only a build-time dep built with this option?

stefanha:
@ilovezfs The pkg-config dependency is optional at run-time.  If your qmake file does not use CONFIG link_pkgconfig then there is no need to have pkg-config installed.

ilovezfs:
@stefanha :rocket: 

Thanks for the PR!

PR #7137, opened 2016-11-22
llnode fix lldb resource for Xcode < 8.0

indutny:

-----

Hello!

I'm not quite sure if I should update it with next release or not, but decided to request the feedback anyway.

While investigating a bug report, I've discovered that Yosemite (actually Xcode on Yosemite) has much older lldb which made bottles for it unusable. Is it a right way to make resource dependent on OS X version?

I'd love to make it dependent on Xcode version, but the version correspondence of Xcode to lldb is unknown to me.

Thanks!

### DIFF
--- a/Formula/llnode.rb
+++ b/Formula/llnode.rb
@@ -20,8 +20,17 @@ class Llnode < Formula
   end
 
   resource "lldb" do
-    url "https://github.com/llvm-mirror/lldb.git",
-        :revision => "839b868e2993dcffc7fea898a1167f1cec097a82"
+    if MacOS::Xcode.version >= "8.0"
+      # lldb 360.1
+      url "https://github.com/llvm-mirror/lldb.git",
+          :revision => "839b868e2993dcffc7fea898a1167f1cec097a82"
+    else
+      # It claims it to be lldb 350.0 for Xcode 7.3, but in fact it is based
+      # of 34.
+      # Xcode < 7.3 uses 340.4, so I assume we should be safe to go with this.
+      url "http://llvm.org/svn/llvm-project/lldb/tags/RELEASE_34/final/",
+          :using => :svn
+    end
   end
 
   def install


indutny:
Additionally, you seem to have quite a variety of OS X servers. May I ask you if you could help collect data on xcode vs lldb version correspondence: https://gist.github.com/indutny/b38afd4a2f9cd6277d20a6a1dda84a3e ?

Using this data, I may be able to change OS X version check to Xcode version check, which is way more reasonable.

indutny:
I've found all Xcode versions that I was curious about:

https://gist.github.com/indutny/b38afd4a2f9cd6277d20a6a1dda84a3e

Is there any way to add xcode dependency to the project?

indutny:
What I mean is that the resource that I'm using is dependent on command line tools currently installed on the machine. Depending on `lldb` version it is either newer or older lldb source.

This doesn't seem to solely depend on OS X version itself.

MikeMcQuaid:
> Is there any way to add xcode dependency to the project?

I'm still not quite sure what you're asking here, can you elaborate? Writing some non-working code for what you'd like to be able to do in a comment would be good.

indutny:
@MikeMcQuaid sorry, I totally confused everyone here with my subsequent comments.

I've pushed suggested fix to this PR. My question is: will Xcode version dependency work with bottles? As far as I can see brew uses just OS version to download an appropriate bottle. (This may be fine, assuming that Xcode 8 is on El Capitan, and Xcode 7 is on Yosemite)

MikeMcQuaid:
Yeh, just OS version is used for bottling. We use `brew doctor` to warn people on outdated Xcode versions so I think that's a same assumption for now.

MikeMcQuaid:

For future reference the preferred commit message format is `llnode: fix lldb resource for Xcode < 8.0.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

indutny:
@MikeMcQuaid thanks! Will surely do it better next time.

PR #7151, opened 2016-11-22
afflib: fix 3.7.10 to have better dependencies

umireon:

-----

* Use macOS's expat
* Enable S3 support since expat and curl are always present
* Include its python binding as optional
* Ensure to link with OSXFUSE if HOMEBREW_PREFIX is not /usr/local
* Ensure not to link with OSXFUSE

### DIFF
--- a/Formula/afflib.rb
+++ b/Formula/afflib.rb
@@ -15,25 +15,43 @@ class Afflib < Formula
   depends_on "autoconf" => :build
   depends_on "automake" => :build
   depends_on "libtool" => :build
-  depends_on "expat" => :optional
-  depends_on :osxfuse => :optional
+  depends_on "pkg-config" => :build
+
   depends_on "openssl"
 
-  # This patch fixes a bug reported upstream over there
-  # https://github.com/simsong/AFFLIBv3/issues/4
-  patch :DATA
+  depends_on "python" => :optional
+  depends_on :osxfuse => :optional
+
+  patch do
+    url "https://github.com/sshock/AFFLIBv3/pull/13.patch"
+    sha256 "dc24b0be3c17938b5b6014ba0fcd885c5f79e758c2150e3727fbda1507cdb768"
+  end
+  patch do
+    url "https://github.com/sshock/AFFLIBv3/pull/14.patch"
+    sha256 "3a078e41bd764fd45c5833335f5650f815cbfaea5fce4dca684d270742c3b34a"
+  end
+  patch do
+    url "https://github.com/sshock/AFFLIBv3/pull/15.patch"
+    sha256 "e8028dd0ca8573c7d7e51234494782d01d85b1a167b25f12140d5ea21dccea3f"
+  end
 
   def install
-    system "./bootstrap.sh"
+    args = ["--enable-s3"]
 
-    args = ["--disable-dependency-tracking", "--prefix=#{prefix}"]
+    args << "--enable-python" if build.with? "python"
 
     if build.with? "osxfuse"
-      ENV["CPPFLAGS"] = "-I/usr/local/include/osxfuse"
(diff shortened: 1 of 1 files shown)


Formula/afflib.rb
```diff
@@ -15,25 +15,33 @@ class Afflib < Formula
...
-  depends_on :osxfuse => :optional
+
   depends_on "openssl"
 
+  depends_on "python" => :optional
+  depends_on :osxfuse => :optional
+
   # This patch fixes a bug reported upstream over there
   # https://github.com/simsong/AFFLIBv3/issues/4
   patch :DATA
```
MikeMcQuaid: Would you consider sending this patch as a PR (as requested on that now-closed issue)? Thanks!

MikeMcQuaid (approved):

Formula/afflib.rb
```diff
@@ -57,21 +65,24 @@ def install
...
 +++ b/configure.ac
-@@ -241,10 +241,6 @@ AC_ARG_ENABLE(fuse,
+@@ -238,13 +238,6 @@ AC_ARG_ENABLE(fuse,
  if test "x${enable_fuse}" = "xyes" ; then
    AC_MSG_NOTICE([FUSE requested])
    CPPFLAGS="-D_FILE_OFFSET_BITS=64 -DFUSE_USE_VERSION=26 $CPPFLAGS"
 -  if test `uname -s` = Darwin ; then
 -    AC_MSG_NOTICE([FUSE IS NOT SUPPORTED ON MACOS])
 -    enable_fuse=no
 -  fi
```
umireon: This part is already merged on master: sshock/AFFLIBv3#13

MikeMcQuaid: Great. Can we use https://github.com/sshock/AFFLIBv3/pull/13.patch as a patch URL instead?

umireon:
Thank you for reviewing, @MikeMcQuaid 
sshock/AFFLIBv3#13 is included as a patch.

MikeMcQuaid:
@umireon Sorry, I mean it would be good to submit everything in `patch :DATA` on this formula as PRs. I know you didn't create them but it would be great if you could submit them.

umireon:
I am working on this but I don't think these two part won't be valid on the upstream because they remove the fallback on the error.
* https://github.com/umireon/homebrew-core/blob/843b528e3187c5b8f2ec55c8cd9bb667d78ac82f/Formula/afflib.rb#L67
* https://github.com/umireon/homebrew-core/blob/843b528e3187c5b8f2ec55c8cd9bb667d78ac82f/Formula/afflib.rb#L79-L81

MikeMcQuaid:
@umireon Cool, thanks. I think we'll need to try and find something that's valid/mergeable upstream or just remove the relevant patch.

umireon:
@MikeMcQuaid IMO the configure script of afflib is too intrusive and fuzzy, and therefore we can make a PR to fix that.
(e.g. It never fail for lack of libraries, plus it force to use the libraries under /opt/local if exists)
But making changes on configure.ac should take some time of me, because I am not familiar with autotools 😿

umireon:
The patch is reconstructed.
Yet another PR for the attached patch should be created but after sshock/AFFLIBv3#14, because this conflicts with the current master.

umireon:
@MikeMcQuaid Now I have sent all patches as PR and they are all merged: sshock/AFFLIBv3#13, sshock/AFFLIBv3#14, sshock/AFFLIBv3#15
Currently this formula have three patches but they will be able to be removed with the next release of afflibv3.

MikeMcQuaid:
@umireon Fantastic work, I'm really impressed!

umireon:
Thank you for merging, @MikeMcQuaid !

PR #7153, opened 2016-11-22
polygen: migrate to boneyard

ilovezfs:
Created with `brew boneyard-formula-pr` because it doesn't build with ocaml 4.04.0.


MikeMcQuaid:
0️⃣ installs in analytics: 👍

chenrui333 (approved):

PR #7156, opened 2016-11-22
idris: fix build with GHC 8.0.2

ilovezfs:
apply commit from Idris's trifecta dependency's upstream fixing a
trifecta bug that causes build failure with GHC 8.0.2

CC @RyanGlScott


ilovezfs:
Superseded by trifecta-1.6.1 https://github.com/ekmett/trifecta/issues/68#issuecomment-262600645

PR #7184, opened 2016-11-23
infer 0.9.4 update

akotulski:

-----

Update `sha256` of infer tarball after hotfixing it

### DIFF
--- a/Formula/infer.rb
+++ b/Formula/infer.rb
@@ -2,7 +2,7 @@ class Infer < Formula
   desc "Static analyzer for Java, C and Objective-C"
   homepage "http://fbinfer.com/"
   url "https://github.com/facebook/infer/releases/download/v0.9.4/infer-osx-v0.9.4.tar.xz"
-  sha256 "529d147bccf3285ddb7500c22e0c50d6e0cbdb2c7f9b11a84e8005873994b3e2"
+  sha256 "82a5d6e35c68149778be2dd6e82407026c507a47f6c22442ee691bc9e3f9a5c9"
 
   bottle do
     cellar :any


MikeMcQuaid:
Did the tag change or just the macOS tarball?

akotulski:
I changed tarball
I moved the tag to different hash (with hotfix)
I kept the name of the tag (via `git push --force`)

MikeMcQuaid:
Anyone who has `git fetch --tags` the existing tag will not be updated to the new one. Git doesn't update local tags which have changed because tags are intended to be static. For Homebrew, this will break the checksum for anyone who used the previous tag. As a result, would you consider either in this case or in future releasing a new version instead of repackaging an existing one? Thanks.

akotulski:
I don't fully understand how brew is using tags when deciding whether to update.
Is it only matter of tag names, or do I have to create new release on github as well?
I could create 0.9.4.1 tag (instead of reusing 0.9.4) in our repo if that helps.

MikeMcQuaid:
> I don't fully understand how brew is using tags when deciding whether to update.

We download the `infer-osx-v0.9.4.tar.xz` file and cache it and at installation time verify that the file matches the checksum.

> I could create 0.9.4.1 tag (instead of reusing 0.9.4) in our repo if that helps.

I think that would be better for the users of your GitHub repository. For Homebrew ideally the old tarball would stay with the old checksum and we could have a 0.9.4.1 tarball with the new checksum/contents and upgrade to that instead.

akotulski:
ok, does it summarize what you are suggesting?
1. create new git tag (not strictly required by brew)
2. create new github release linked to the tag (not strictly required by brew)
3. have different URL to the tarball (brew will get confused if URL stays the same)

MikeMcQuaid:
@akotulski Yep!

PR #7185, opened 2016-11-23
llvm: fix OCaml option and CLT test

ilovezfs:
OCaml option needs `ocaml-ctypes` and `ocamlfind`, so install them
temporarily with OPAM since we have no corresponding formulae.

CLT test can no longer assume the CLT's clang version is the same as the
DevelopmentTools.clang_version since if both Xcode 8 and CLT 7 are
installed (as is typically the case now for El Capitan installations),
the DevelopmentTools.clang_version will be 8.0.

CC @MikeMcQuaid


MikeMcQuaid:
Mad but fine with me 👍

PR #7187, opened 2016-11-23
ripgrep: install bash and fish completions

mernen:

-----

Sorry for not including this with the 0.3.1 patch! It didn't occur to me before.

### DIFF
--- a/Formula/ripgrep.rb
+++ b/Formula/ripgrep.rb
@@ -19,6 +19,12 @@ class Ripgrep < Formula
 
     bin.install "target/release/rg"
     man1.install "doc/rg.1"
+
+    # Completion scripts are generated in the crate's build directory, which
+    # includes a fingerprint hash. Try to locate it first
+    out_dir = Dir["target/release/build/ripgrep-*/out"].first
+    bash_completion.install "#{out_dir}/rg.bash-completion"
+    fish_completion.install "#{out_dir}/rg.fish"
   end
 
   test do


Formula/ripgrep.rb
```diff
@@ -12,13 +12,23 @@ class Ripgrep < Formula
     sha256 "70a123b6254a752f8ae87420bd07384a0cde8340d866a354e83af4c1b4dcd127" => :yosemite
   end
 
+  option "without-completions", "Disable bash/fish completions"
```
bfontaine: Why would someone use this option?

mernen: I just followed the `git` formula structure, and assumed offering an option was typical.

bfontaine: We try to limit the number of options; please remove it unless there’s a common use-case here.

mernen: No problem! Please see the new commit.

Formula/ripgrep.rb
```diff
@@ -19,6 +19,14 @@ def install
 
     bin.install "target/release/rg"
     man1.install "doc/rg.1"
+
+    # Completion scripts are generated in the crate's build directory, which
+    # includes a fingerprint hash. Try to locate it first
+    out_dir = Dir["target/release/build/ripgrep-*/out"].first
+    if out_dir
```
mernen: This test is a new addition. `out_dir` detection *should* never fail, but in case someday it does, let's not break with a cryptic error message.

mernen: What could I do? Is there a method to emit a warning? Or just fail with a clearer message?

mernen: What could I do? Is there a method to emit a warning? Or should I fail with a clearer message?

bfontaine: You can let it fail; the stacktrace will indicate the error as well as its location.

MikeMcQuaid:

PR #7190, opened 2016-11-23
add formula for red-lang

rcqls:

-----

### DIFF
--- a/Formula/red.rb
+++ b/Formula/red.rb
@@ -1,0 +1,16 @@
+class Red < Formula
+  desc "Next-gen programming language, strongly inspired by REBOL"
+  homepage "http://www.red-lang.org"
+  url "http://static.red-lang.org/dl/mac/red-061"
+  sha256 "afefaa392e5dbc1ec6d8805376ecffe86a1f6d1ce46d426800556f3c4f698693"
+  head "https://github.com/red/red.git"
+
+  def install
+    mv "red-061", "red"
+    bin.install "red"
+  end
+
+  test do
+    system "#{bin}/red", "--version"
+  end
+end


Formula/red.rb
```diff
@@ -0,0 +1,16 @@
+class Red < Formula
+  desc "Next-gen programming language, strongly inspired by REBOL"
+  homepage "http://www.red-lang.org"
+  url "http://static.red-lang.org/dl/mac/red-061"
```
MikeMcQuaid: This needs to be built from source instead.

Formula/red.rb
```diff
@@ -0,0 +1,16 @@
...
+  sha256 "afefaa392e5dbc1ec6d8805376ecffe86a1f6d1ce46d426800556f3c4f698693"
+  head "https://github.com/red/red.git"
+
+  def install
+    mv "red-061", "red"
+    bin.install "red"
```
MikeMcQuaid: `bin.install "red-061" => "red"` does the same as the previous two lines.

Formula/red.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  test do
+    system "#{bin}/red", "--version"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

MikeMcQuaid (changes requested):

Formula/red.rb
rcqls: I understand your point but the process in order to build the executable is very complex since it depends currently on rebol SDK which is not freely available and actually no more provided. 
I am not involved in the development of the red language but the creator of the language asks me to propose a PR since homebrew is in his opinion the best system to install tools in MacOSX system. 
His point is also to inform that red is an interesting language in the MacOSX world. In some future release, red will only depend on red and it will be possible (I guess) to build red from source. 
Furthermore, as a MacOSX user I do not want to use macport (or other system) to install red. From now, the installation is very basic but Homebrew is also a way to discover some interesting tools.

Formula/red.rb
rcqls: I did not know. Thanks

Formula/red.rb
rcqls: Thanks for the advice, I will do a look at cmake.rb example. But I am also waiting for your answer (see first comment) about the binary installation instead of source installation. Thanks a lot for this  excellent installation system.

Formula/red.rb
MikeMcQuaid: In that case this doesn't look like it's open-source so is not a good fit for Homebrew. Homebrew Cask may accept it.

rcqls: Ok, I understand but it is really an open-source project as you can see at [github/red](https://github.com/red/red).
From the red binary you can compile (and even cross-compile) any red source to binary.

MikeMcQuaid: That's fine, we just want to build it from source. If that means we have to use a binary cross-compiler: so be it.

rcqls: Sorry but I think that I was not clear enough. What we want to install it is exactly and only the binary cross-compiler (red-061 and very soon the red-062). With this binary you can compile any red source file. From now, red can't be bootstrapped that's why rebol2 (non open source project) is used to create the red executable and its author want to distribute it.
Do you accept the pull request?

rcqls: Will my tap be visible to Homebrew users?

rcqls: Ok, I'll ask the author. The last development maybe allows us to do it but I have some doubts. Many thanks for your help!

rcqls: I sent a PR to Homebrew cask and they think that this PR has to be a formula and not a cask. The PR is `on hold` state waiting for your answer. I ask the author and he did not propose a source-based installation.

rcqls: @MikeMcQuaid Are you doing some check? 
By the way, as a linux user I am in love with HB since with this installer system I can use my Mac mostly as a linux user. In a general purpose, if I want to share a binary  do you give me the advice to use macports or some another installer system accepting both binary and source-based installation. I thought it was not a good idea to mix different installer systems. I'll try once and I never do it again. I thought also Homebrew cask was the right tool to install a binary. So at some point, I am a bit lost ... I understand your point to prefer source-based installation but I do not understand why you do not want to accept to share it as a binary too. It is like apt-get only does one kind of installation (binary or from source) and not both at the same time. As an example, R is distributed as a cask and maybe not as a Formula (if I am not mistaken) but R can be installed from source on MacOSX. So I am lost ... sorry ...

MikeMcQuaid:
Passing on this for now until it can be self-bootstrapped in favour of Homebrew Cask PR in https://github.com/caskroom/homebrew-cask/pull/27139.

PR #7191, opened 2016-11-23
ffmpeg: add ability to disable SecureTransport

saghul:

-----

It uses private APIs which Apple doesn't allow on AppStore releases.

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -35,6 +35,7 @@ class Ffmpeg < Formula
   option "with-zimg", "Enable z.lib zimg library"
   option "without-lame", "Disable MP3 encoder"
   option "without-qtkit", "Disable deprecated QuickTime framework"
+  option "without-securetransport", "Disable use of SecureTransport"
   option "without-x264", "Disable H.264 encoder"
   option "without-xvid", "Disable Xvid MPEG-4 video encoder"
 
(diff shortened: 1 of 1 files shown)


zmwangx:
Might be worth checking whether `/usr/lib/libssl.0.9.8.dylib` is linked when `--disable-securetransport` is specified. If it is, we might want to add `--disable-openssl` unless `--with-openssl` is specified.

saghul:
@zmwangx TBH I don't know what ffmpeg needs those for. Also, AFAIK `--with-openssl` is not the default, so why would it link with the system OpenSSL? Also, its header have not been exported for the past 2 macOS releases...

zmwangx:
> TBH I don't know what ffmpeg needs those for.

FFmpeg supports a bunch of live streaming protocols — RTMP and RTSP for instance — that may require TLS, among other things.

> Also, AFAIK --with-openssl is not the default, so why would it link with the system OpenSSL?

FFmpeg's configure autodetects many optionally dependencies. However, I just checked and `--enable-openssl` defaults to `no`, so no need to worry about it.

saghul:
@zmwangx Ok.  So, is there anything else I would need to do here?

zmwangx:
I'm 👍 on this.

MikeMcQuaid:
Thanks again for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. Thanks for the help here too @zmwangx. You rock!

PR #7193, opened 2016-11-23
kops 1.4.1

WillPlatnick:
This commit adds a formula that installs kops, a tool for Kubernetes orchestration.

### DIFF
--- a/Formula/kops.rb
+++ b/Formula/kops.rb
@@ -1,0 +1,22 @@
+class Kops < Formula
+  desc "Production Grade K8s Installation, Upgrades, and Management"
+  homepage "https://github.com/kubernetes/kops"
+  url "https://github.com/kubernetes/kops/archive/v1.4.1.tar.gz"
+  sha256 "69b3c9d7e214109cfd197031091ed23963383c894e92804306629f6a32ab324b"
+  head "https://github.com/kubernetes/kops.git"
+
+  depends_on "go" => :build
+  depends_on "kubernetes-cli"
+
+  def install
+    ENV["GOPATH"] = buildpath
+    kopspath = buildpath/"src/k8s.io/kops"
+    kopspath.install Dir["*"]
+    system "make", "-C", kopspath
+    bin.install("bin/kops")
+  end
+
+  test do
+    system "#{bin}/kops", "version"
+  end
+end


chrislovecnm (changes requested):
You are going to need to depend on kubectl as well. I believe there is already a brew for it.

Formula/kops.rb
```diff
@@ -0,0 +1,28 @@
...
+  def install
+    ENV["GOPATH"] = buildpath
+    kopspath = buildpath/"src/k8s.io/kops"
+    kopspath.install Dir["*"]
+
+    cd kopspath do
+      system "make"
+    end
+
+    cd buildpath do
```
MikeMcQuaid: This is unnecessary; you'll be `cd`d back there after you leave the `cd kopspath do` block.

Formula/kops.rb
```diff
@@ -0,0 +1,28 @@
...
+  depends_on "go" => :build
+  depends_on "kubectl"
+
+  def install
+    ENV["GOPATH"] = buildpath
+    kopspath = buildpath/"src/k8s.io/kops"
+    kopspath.install Dir["*"]
+
+    cd kopspath do
+      system "make"
```
MikeMcQuaid: Alternatively, `system "make", "-C", kopspath` and skip the `cd`

Formula/kops.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    cd buildpath do
+      bin.install("bin/kops")
```
MikeMcQuaid: Is there no `make install` we can use?

MikeMcQuaid (changes requested):

Formula/kops.rb
WillPlatnick: There is not. make installs the binary into GOPATH and leaves them there.

Formula/kops.rb
WillPlatnick: Will push fix

Formula/kops.rb
WillPlatnick: Will push fix

WillPlatnick:
@MikeMcQuaid Requested changes addressed and squashed

MikeMcQuaid (approved):

MikeMcQuaid:

For future reference the preferred commit message format is `kops 1.4.1 (new formula)`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

Issue #7054, opened 2016-11-18
GMP was compiled with -march=sandybridge


smtrat:
I use a library that decided to reuse the compilation flags of gmp that are defined in __GMP_CFLAGS at the end of gmp.h. If gmp was installed as binary package via homebrew, these flags contain -march=sandybridge and -mtune=sandybridge.

Given that there are still machines out there with pre-sandybridge processors, this seems very risky. Could we change this to some sane defaults? Debian, for example, does not specify -mtune and -march at all.

For some reason, I have never encountered a problem with gmp itself. Is gmp even compiled with these options? Or are they just spurious remainders of the machine that builds the packages?

ilovezfs:
The bottles are built with -march=core2, but that's because Homebrew superenv intervenes when something other than that is passed, so it's possible the build system recorded something other than was actually used.

ilovezfs:
Yes, that's what's going on. If I build on Skylake,
```
/usr/local/opt/gmp/include/gmp.h:#define __GMP_CFLAGS "-O2 -pedantic -fomit-frame-pointer -m64 -mtune=skylake -march=broadwell"
```

And here are superenv's interventions:
https://gist.github.com/ilovezfs/ad883542d12bf1a0bb4db08a5ec3f1e8

The first shows
```
clang called with: -DHAVE_CONFIG_H -I. -I.. -I.. -O2 -pedantic -fomit-frame-pointer -m64 -mtune=skylake -march=broadwell -c memory.c -fno-common -DPIC -o .libs/memory.o
superenv removed:  -O2 -pedantic -m64 -mtune=skylake -march=broadwell
superenv added:    -pipe -w -Os -march=core2 -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers
superenv executed: clang -pipe -w -Os -march=core2 -DHAVE_CONFIG_H -I. -I.. -fomit-frame-pointer -c memory.c -fno-common -DPIC -o .libs/memory.o -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers
```

If you feel like opening a pull request that changes this so that the gmp.h installed by the bottle reflects reality more closely, we can take a look.

ilovezfs:
We'd welcome a PR addressing this but no one is actively working on it at this time.

smtrat:
[...]

PR #7215, opened 2016-11-24
Properly setting a fake CPUID for GMP

smtrat:
GMP exports its compilation options in its header file which may contain incorrect architecture optimizations.
This patch ensures that GMP builds with -mcore2 and also exports this.

See #7054

### DIFF
--- a/Formula/gmp.rb
+++ b/Formula/gmp.rb
@@ -18,7 +18,7 @@ class Gmp < Formula
 
   def install
     ENV.cxx11 if build.cxx11?
-    args = ["--prefix=#{prefix}", "--enable-cxx"]
+    args = ["--prefix=#{prefix}", "--enable-cxx", "--enable-fake-cpuid", "--enable-fat", "GMP_CPU_TYPE=core2"]
 
     if build.build_32_bit?
       ENV.m32
--- a/Formula/gmp.rb
+++ b/Formula/gmp.rb
@@ -18,15 +18,14 @@ class Gmp < Formula
 
   def install
     ENV.cxx11 if build.cxx11?
-    args = ["--prefix=#{prefix}", "--enable-cxx", "--enable-fake-cpuid", "--enable-fat", "GMP_CPU_TYPE=core2"]
+    args = ["--prefix=#{prefix}", "--enable-cxx"]
 
     if build.build_32_bit?
       ENV.m32
       args << "ABI=32"
     end
 
-    # https://github.com/Homebrew/homebrew/issues/20693
-    args << "--disable-assembly" if build.build_32_bit? || build.bottle?
+    args << "--enable-fake-cpuid" << "GMP_CPU_TYPE=#{Hardware.oldest_cpu}" << "--enable-fat" if build.build_32_bit? || build.bottle?
 
     system "./configure", *args
     system "make"
--- a/Formula/gmp.rb
+++ b/Formula/gmp.rb
@@ -19,17 +19,18 @@ class Gmp < Formula
   def install
     ENV.cxx11 if build.cxx11?
     args = ["--prefix=#{prefix}", "--enable-cxx"]
+    ENV["GMP_CPU_TYPE"] = Hardware.oldest_cpu
 
     if build.build_32_bit?
       ENV.m32
       args << "ABI=32"
     end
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
Probably want to use `#{Hardware.oldest_cpu}` not `core2` and scope this to `if build.bottle?`. Ideally, it would also respect ` --bottle-arch=`

smtrat:
I just copied `-march=core2` from https://gist.github.com/ilovezfs/ad883542d12bf1a0bb4db08a5ec3f1e8 you referenced in #7054 and assumed that everything in brew would be configured this way.
Could you do these changes? I neither use OSX myself, nor do I know how to write ruby or how these formulas are supposed to work...

Formula/gmp.rb
```diff
@@ -25,8 +25,7 @@ def install
       args << "ABI=32"
     end
 
-    # https://github.com/Homebrew/homebrew/issues/20693
-    args << "--disable-assembly" if build.build_32_bit? || build.bottle?
+    args << "--enable-fake-cpuid" << "GMP_CPU_TYPE=#{Hardware.oldest_cpu}" << "--enable-fat" if build.build_32_bit? || build.bottle?
```
MikeMcQuaid:
```ruby
if build.build_32_bit? || build.bottle?
  args << "--enable-fat" << "--enable-fake-cpuid" << "GMP_CPU_TYPE=#{Hardware.oldest_cpu}"
end
```

Also, what is `--enable-fat` doing?

MikeMcQuaid (changes requested):

MikeMcQuaid:
Installation is failing on all platforms: https://bot.brew.sh/job/Homebrew%20Core/12014/version=any_sierra/testReport/junit/brew-test-bot/sierra/install_gmp/

smtrat:
I've noticed. `--enable-fat` seems to be required when specifying `--enable-fake-cpuid`. Afaik, `--enable-fat`build a library that contains specific instructions for all platforms and selects at runtime which to choose.

There is a file that is linked into all tests that triggers these errors when `GMP_CPU_TYPE` is not set. I can fix this by setting `GMP_CPU_TYPE=#{Hardware.oldest_cpu}` for the tests, but I'm not sure whether we want to do that: It kind of suggests that end users also have to do this when regularly using libgmp (though this would not be the case, as far as I can tell).

MikeMcQuaid:
> Afaik, --enable-fatbuild a library that contains specific instructions for all platforms and selects at runtime which to choose.

This doesn't seem to be desirable for what's a fix for a cosmetic change.

smtrat:
Maybe this would be a reasonable change anyway? According to gmp, "this means more code, but gives good performance on all x86 chips." Does not seem to bad...

MikeMcQuaid:
Are they including x86_64 in that?

smtrat:
Though I'm by no means an expert here, it seems so. `mpn/fat.c` seems to deal with this and it explicitly mentions that it was originally `x86_32 code` and has been ported to `x86_64`. Furthermore, there is a whole `mpn/x86_64` directory...

MikeMcQuaid:
Ok, seems fine then if you adjust the code as suggested above and we can get it compiling.

Formula/gmp.rb
```diff
@@ -25,12 +25,11 @@ def install
...
 
     system "./configure", *args
     system "make"
-    system "make", "check"
+    system "make", "check", "GMP_CPU_TYPE=#{Hardware.oldest_cpu}"
```
MikeMcQuaid: Could this perhaps be set in both places with `ENV["GMP_CPU_TYPE"] = Hardware.oldest_cpu`?

smtrat: I guess so. I just added this change, let's see whether it works.

MikeMcQuaid (approved):

MikeMcQuaid:
Looks like this breaks a bunch of things at runtime e.g. `gcc`: https://bot.brew.sh/job/Homebrew%20Core/12495/version=any_sierra/testReport/junit/brew-test-bot/sierra/test_gcc/

ilovezfs:
superseded by https://github.com/Homebrew/homebrew-core/pull/8597

ilovezfs:
@smtrat thanks for this PR nonetheless

PR #7217, opened 2016-11-24
gpg-agent Add launchd plist file

gjasny:


gjasny:
Need to teach launchd that the agent double-forks in daemon mode.

MikeMcQuaid:
You can push changes to this PR instead of opening a new one.

gjasny:
Oh boy. After reading through [Macports #43584](https://trac.macports.org/ticket/43584) I came to the conclusion that it might be easier to use some shell magic like the [zsh gpg-plugin](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/gpg-agent).

MikeMcQuaid:
@gjasny Whatever you think is best. Feel free to close if you're no longer interested in making a PR here. Thanks either way!

chdiza:
I don't see the reason for this.  The agent starts on demand and stays active.

fxcoudert:
There seems to be general consensus that this PR is not needed. Closing accordingly.

PR #7239, opened 2016-11-25
terraform 0.7.13 (retag, fix sha)

### LINKED ISSUE #10357: Wrong version shown - v0.7.13 shows as v0.8.0-dev
Just updated to 0.7.13 but it says it is v0.8.0-dev

```shell
$ terraform --version
Terraform v0.8.0-dev
```

cblecker:

-----

Updating sha256, as the release was retagged to fix https://github.com/hashicorp/terraform/issues/10357
Note: Due to the tagging issue (wrong commit), bottles should be rebuilt for this version.

### DIFF
--- a/Formula/terraform.rb
+++ b/Formula/terraform.rb
@@ -4,7 +4,7 @@ class Terraform < Formula
   desc "Tool to build, change, and version infrastructure"
   homepage "https://www.terraform.io/"
   url "https://github.com/hashicorp/terraform/archive/v0.7.13.tar.gz"
-  sha256 "86d19f7a927bbdca3fe320e5213e25a11a4ddffd0458d449b8b1c488f4fe7295"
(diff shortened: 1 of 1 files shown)


mistydemeo:
Thanks!

MikeMcQuaid:
@cblecker In future: please don't retag Git-based projects for a few reasons:
- when you retag you break the from-source build for anyone using Homebrew (and also possibly other projects)
- when you retag we don't know until a PR is filed whether it was intentional or you've been hacked
- we (and you) now can't easily differentiate between users using the old tag or the new one as they have the same Homebrew version number (@mistydemeo: on our end we should probably make retags mandate revisions)
- when you retag anyone who has done `git fetch --tags` and runs it again will not receive an updated tag. This is because Git is designed to not update tags as they are meant to be permanent. Retagging in Git-land is equivalent to force-pushing to a public branch (i.e. bad)

As a result of all this the idea thing to do in this case is a `0.7.14` or `0.7.13.1` release. Thanks!

PR #7247, opened 2016-11-26
ffmpeg: fix fatal patch error when building head

JCount:

-----

the aac_adtstoasc_bsf regression patch was backported to 3.2 from
master, therefore do not try and apply it during head builds

_NOTE:_ This currently breaks the `HEAD` build, since brew aborts after two fatal:
```
Error: Failure while executing: /usr/bin/patch -g 0 -f -p1 -i /private/tmp/ffmpeg--patch-*
```

### DIFF
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -1,18 +1,21 @@
 class Ffmpeg < Formula
   desc "Play, record, convert, and stream audio and video"
   homepage "https://ffmpeg.org/"
-  url "https://ffmpeg.org/releases/ffmpeg-3.2.tar.bz2"
-  sha256 "76d6cd9f5e64463a5b9940736da8a515c990bcbbe506a722e2040916cb366d74"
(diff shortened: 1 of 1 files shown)


zmwangx:
My bad, wasn't in my right mind. Anyway, it doesn't matter anymore, since 3.2.1 is out. #7248.

ilovezfs:
@JCount thanks for this PR. Fortunately, this is now moot due to #7248 as @zmwangx said.

JCount:
@zmwangx @ilovezfs I thought 3.2.1 might come out before this would get merged, but I had no concrete sense of the timeline, so I figured I might as well just open this PR to fix it just in case. I am in fact glad that this is now moot, because it means the ffmpeg formula doesn't have to get split into seperate `stable` and `head` blocks.

It is rather amusing that this PR is #7247, and the very PR, 7248 is for `ffmpeg 3.2.1`. 😆

PR #7258, opened 2016-11-26
onetime 2.0.15 (devel)

DomT4:

-----

### DIFF
--- a/Formula/onetime.rb
+++ b/Formula/onetime.rb
@@ -30,11 +30,11 @@ class Onetime < Formula
   end
 
   devel do
-    url "http://red-bean.com/onetime/onetime-2.0-beta13.tar.gz"
+    url "http://red-bean.com/onetime/onetime-2.0-beta15.tar.gz"
     # FIXME: I can't rememeber why the custom version was added now, but
     # we're stuck with it now as 2.0-beta(n) is "less" than 2.0.0(n).
-    version "2.0.13"
-    sha256 "573c7f3a380a9c844dce7f8fe9dcc0086a7b0614b11a56f0b740617bc3dbfe52"
+    version "2.0.15"
+    sha256 "7f3678a70f6b83fdd961c1aa9795ee513afffb6e8684d28179cbac63add5d8f2"
   end
 
   def install


MikeMcQuaid:
Test failing on El Cap: https://bot.brew.sh/job/Homebrew%20Core/12084/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test___devel_onetime/

DomT4:
Oh joy, Python-based errors 🙈. Not sure why that's failing, to be honest, especially when Yosemite passes.

ilovezfs:
It's not reproducible on demand as far as I can tell. But this is what can happen when a test generates its input with /dev/random :)

DomT4:
Could always use Ruby's PRNG to generate a slightly more predictable/sane random output, or we could write a garbage file to use as the random source, which for testing purposes I guess doesn't really have to be cryptographically secure.

ilovezfs:
I ran it a few times and it occurred. Here's an input file that triggers it

[pad_data.txt.zip](https://github.com/Homebrew/homebrew-core/files/614293/pad_data.txt.zip)

ilovezfs:
```
$ openssl dgst -sha256 pad_data.txt
SHA256(pad_data.txt)= 6c4f1ed861faae5badf91855371cd4dad4a1fb2d9e3c963ef7b577deaec89b54
```

DomT4:
Even with that can't get it to reproduce on Sierra here. Maybe El Cap's Python has a bug or something. I don't know. Any preference over PRNG or explicitly written garbage?

ilovezfs:
Reproduces on Sierra too
```
Josephs-Mac:rand joe$  onetime -e --pad=pad_data.txt --no-trace --config=. input.txt
  File "/usr/local/bin/onetime", line 1923, in <module>
    main()
  File "/usr/local/bin/onetime", line 1841, in main
    result = encoder.finish()
  File "/usr/local/bin/onetime", line 1493, in finish
    remainder += self.pad_sess.finish()
  File "/usr/local/bin/onetime", line 1424, in finish
    remainder += self._make_fuzz(self._tail_fuzz_length)
  File "/usr/local/bin/onetime", line 1105, in _make_fuzz
    ret_data = ret_data + chr(ord(rnd_data[i]) ^ ord(pad_data[i]))
IndexError: string index out of range
Josephs-Mac:rand joe$
```

DomT4:
```
~> /usr/bin/env python --version
Python 2.7.10
~> touch input.txt
~> onetime -e --pad=pad_data.txt --no-trace --config=. input.txt
~> $?
0
```
😢

ilovezfs:
do you have the real input.txt there?

DomT4:
Ah, that broke it for me 👍.

DomT4:
I was using `touch` instead of `echo` because I'm lazy.

ilovezfs:
Interestingly it's a regression from the stable version, at least for that input file. So it should be bisectable.

DomT4:
Seems to be a regression from even 2.0.13, which I can't reproduce this problem with.

ilovezfs:
nice. that narrows it down!

DomT4:
https://github.com/kfogel/OneTime/commit/76c46de2efdd651846758e419f2ef35743b5968a

ilovezfs:
lol. does it apply cleanly?

DomT4:
That's not the fix, that's the breakage apparently. Breakage is still there in master.

ilovezfs:
Yikes. Feel like reporting it? :innocent:

DomT4:
Aye. Was gonna see if I could narrow it down to a specific line given a lot of the changes there are code comments, but maybe easier to just report the commit & explain the breakage without poking at it too much given my less-than-fond relationship with Python.

ilovezfs:
Yeah, I'd let upstream fix this nonsense.

DomT4:
https://github.com/kfogel/OneTime/issues/26

ilovezfs:
@DomT4 thanks! may want to correct `/dev/null` to `/dev/random` :)

DomT4:
Crap. Long week 😓. Thanks for spotting that.

ilovezfs:
You're welcome :)

fxcoudert:
@BrewTestBot test this please

MikeMcQuaid:
@fxcoudert 👍

DomT4:
Upstream has vanished, happens every so often, but they usually return with a wad of sudden activity at some point. I'd be tempted to just leave this formula where it is for now; I can't explain why it would suddenly be passing CI unless the initial failure is random.

ilovezfs:
It is random remember?

DomT4:
I thought it was pretty consistent on CI? I don't know. I can't remember last week, so, this PR is a stretch 🙈

ilovezfs:
It's reproducible with given broken input, but since we generate the input in the test with /dev/rand, you get success or failure depending on whether you stumble into a broken input or not.

DomT4:
Cheers 👍. I forgot this PR existed to be honest, so, remembering what happened here without re-reading the thread wasn't going to be a strong point 😓.

ilovezfs:
Heh no problem. You're not responsible for the process snafu above.

PR #7260, opened 2016-11-26
yelp-tools: convert yelp-xsl to a resource (#7243)

### LINKED ISSUE #7243: yelp-tools package is version 3.18 but related yelp-xsl package is 3.20
The Mallard / GNOME documentation generator package "yelp-tools" is version 3.18 in Homebrew core. It can use XSLT templates from the related yelp-xsl package, except it's expecting to have the same yelp-xsl version, 3.18. Unfortunately the current version in Homebrew is 3.20. This leads to:

```
/usr/local/bin/yelp-build html -o OUTPUT-DIR INPUT-DIR
warning: failed to load external entity "/usr/local/Cellar/yelp-xsl/3.18.0/share/yelp-xsl/xslt/mallard/cache/mal-cache.xsl"
cannot parse /usr/local/Cellar/yelp-xsl/3.18.0/share/yelp-xsl/xslt/mallard/cache/mal-cache.xsl
```

This temporarily hacks a solution:

```
> cd /usr/local/Cellar/yelp-xsl/
> ln -s 3.20.1 3.18.0
```

ajdavis:
*Question*: Is deleting the yelp-xsl formula the right thing to do, or should it remain in some kind of deprecated or superseded state?

yelp-tools is the only package that depends on yelp-xsl, and both
packages must have the same version. Better to make yelp-xsl a resource
of yelp-tools instead of a distinct package that can get out of sync.

Closes #7243


Formula/yelp-tools.rb
```diff
@@ -12,20 +12,45 @@ class YelpTools < Formula
...
+  end
 
   def install
+    resource("yelp-xsl").stage do
+      system "./configure", "--disable-debug",
+                            "--disable-dependency-tracking",
+                            "--disable-silent-rules",
+                            "--prefix=#{prefix}"
+      system "make", "install"
+      ENV.append_path "PKG_CONFIG_PATH", "#{prefix}/share/pkgconfig"
```
ajdavis: This syntax works, I tried this but it had no effect (ENV was the same after):
```
      ENV.append_path "PKG_CONFIG_PATH", "#{lib}/pkgconfig"
```
However, the syntax I'm using here doesn't match how other formulas do it, and "brew audit" complains:
```
yelp-tools:
  * "#{prefix}/share" should be "#{share}"
Error: 1 problem in 1 formula
```

Formula/yelp-tools.rb
```diff
@@ -12,20 +12,45 @@ class YelpTools < Formula
     sha256 "66721baa2d6c6978cceac29f35e8112e96804b6ae0197d89175867f17caca2cd" => :mavericks
   end
 
+  depends_on "gettext" => :build
+  depends_on "gtk+3" => :run
```
MikeMcQuaid: Don't need the `:run` here.

Formula/yelp-tools.rb
MikeMcQuaid: `ENV.append_path "PKG_CONFIG_PATH", "#{share}/pkgconfig"`

Formula/yelp-tools.rb
```diff
@@ -12,20 +12,45 @@ class YelpTools < Formula
...
+
   test do
     system "#{bin}/yelp-new", "task", "ducksinarow"
     system "#{bin}/yelp-build", "html", "ducksinarow.page"
     system "#{bin}/yelp-check", "validate", "ducksinarow.page"
+
+    assert (pkgshare/"xslt/mallard/html/mal2html-links.xsl").exist?
+    assert (pkgshare/"js/jquery.syntax.brush.smalltalk.js").exist?
+    assert (pkgshare/"icons/hicolor/24x24/status/yelp-note-warning.png").exist?
+    assert (share/"pkgconfig/yelp-xsl.pc").exist?
```
MikeMcQuaid: Add a second argument to these for the assertion message so we have a better output if they fail. Could consider looping through these filenames and make `"#{filename} doesn't exist"` as the assertion message

Formula/yelp-tools.rb
```diff
@@ -12,20 +12,45 @@ class YelpTools < Formula
...
+  depends_on "gtk+3" => :run
+  depends_on "intltool" => :build
   depends_on "itstool" => :build
-  depends_on "pkg-config" => :build
-  depends_on "libxslt" => :build
   depends_on "libxml2" => :build
-  depends_on "yelp-xsl"
+  depends_on "libxslt" => :build
+  depends_on "pkg-config" => :build
+
```
MikeMcQuaid: `yelp_tools_version = version`

Formula/yelp-tools.rb
```diff
@@ -12,20 +12,45 @@ class YelpTools < Formula
...
+  resource "yelp-xsl" do
+    url "https://download.gnome.org/sources/yelp-xsl/3.18/yelp-xsl-3.18.0.tar.xz"
```
MikeMcQuaid: `url "https://download.gnome.org/sources/yelp-xsl/3.18/yelp-xsl-#{yelp_tools_version}.tar.xz"` or similar to avoid these getting out of sync.

MikeMcQuaid (changes requested):

ajdavis:
Thanks for the review! New patch.

MikeMcQuaid (approved):

MikeMcQuaid:

tschoonj:
Sorry for coming a bit late to the party but I have some concerns on what was done here:

Though I do not disagree with turning yelp-xsl into a resource, I do wonder why we are forcing a version here that is three stable releases behind the latest. Is there really no way to make this work with yelp-xsl 3.20.1?

MikeMcQuaid:
@tschoonj No worries. Totally fine with updating this to the latest version now we've got the approach nailed down.

ajdavis:
Update to 3.20.1 in progress, but I could use some help: https://github.com/Homebrew/homebrew-core/pull/7386

PR #7263, opened 2016-11-26
qt5: add general plugin search path to Designer.app

dakcarto:

-----

Same as was previously done for Qt4, but with additional re-registration
of edited bundle to ensure change is noticed upon reinstalls.

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -199,6 +199,13 @@ class Qt5 < Formula
               /\n# pkgconfig\n(PKG_CONFIG_(SYSROOT_DIR|LIBDIR) = .*\n){2}\n/,
               "\n"
 
+    # Make `HOMEBREW_PREFIX/lib/qt5/plugins` an additional plug-in search path
+    # for Qt Designer to support formulae that provide Qt Designer plug-ins.
+    system "/usr/libexec/PlistBuddy",
+           "-c", "Add :LSEnvironment:QT_PLUGIN_PATH string \"#{HOMEBREW_PREFIX}/lib/qt5/plugins\"",
+           "#{bin}/Designer.app/Contents/Info.plist"
+    touch "#{bin}/Designer.app" # re-register edited bundle with Launch Services
+
     # Move `*.app` bundles into `libexec` to expose them to `brew linkapps` and
     # because we don't like having them in `bin`. Also add a `-qt5` suffix to
     # avoid conflict with the `*.app` bundles provided by the `qt` formula.


dakcarto:
Caveat: I did not fully rebuild `qt5` (seemed unnecessary), but did temporarily add similar code to a `post_install` section and it worked fine.

Formula/qt5.rb
```diff
@@ -199,6 +199,13 @@ def install
               /\n# pkgconfig\n(PKG_CONFIG_(SYSROOT_DIR|LIBDIR) = .*\n){2}\n/,
               "\n"
 
+    # Make `HOMEBREW_PREFIX/lib/qt5/plugins` an additional plug-in search path
+    # for Qt Designer to support formulae that provide Qt Designer plug-ins.
+    system "/usr/libexec/PlistBuddy",
+           "-c", "Add :LSEnvironment:QT_PLUGIN_PATH string \"#{HOMEBREW_PREFIX}/lib/qt5/plugins\"",
+           "#{bin}/Designer.app/Contents/Info.plist"
+    touch "#{bin}/Designer.app" # re-register edited bundle with Launch Services
```
MikeMcQuaid: This should be submitted as a patch to the upstream Qt buildsystem.

MikeMcQuaid (changes requested):

Formula/qt5.rb
dakcarto: How could the build know anything about the HOMEBREW_PREFIX, which (at that point, during configure) would be a different directory from the install prefix? It would have to be an extra build option, I'm guessing. Seems like something that would not be accepted.

dakcarto: I guess my point here is that that is exactly what [the runtime options](http://doc.qt.io/qt-5/deployment-plugins.html) of `qt.conf` and `QT_PLUGIN_PATH` are meant to be used for.

Usually, for custom setting for an app bundle, I would use `qt.conf`.

MikeMcQuaid: That's fine, it just feels very weird to be doing manual modification of a upstream generated plist. That should be a `configure` option. Would be good to try it with them rather than have effectively a patch we have to carry forever.

dakcarto: Found the issue with `qt.conf`. It is read in via [QSettings](http://doc.qt.io/qt-5/qsettings.html), so Plugins subgroup in `qt.conf` _should_ be able to support QStringList and QSetting's specialized arrays. However, [QLibraryInfo](http://doc.qt.io/qt-5/qlibraryinfo.html) is set up to _always_ return a string representing a single path for each subgroup of `[PATHS]` (i.e. [QLibraryInfo::LibraryLocation](http://doc.qt.io/qt-5/qlibraryinfo.html#LibraryLocation-enum) enums). Any attempt to load a list-styled .ini value, results in a single string (not a list), which causes Designer.app to not find any plugins, thereby crashing when it can not find the required `cocoa` platform plugin.

So, using `qt.conf` is no good for setting multiple paths for plugin discovery. The `QT_PLUGIN_PATH` is always appended to configured paths, so works as needed for setting the extra Homebrew-specific path. Editing Info.plist is still seems the simplest means of passing that env var into the app bundle launch.

MikeMcQuaid:
> So, using qt.conf is no good for setting multiple paths for plugin discovery. The QT_PLUGIN_PATH is always appended to configured paths, so works as needed for setting the extra Homebrew-specific path. Editing Info.plist is still seems the simplest means of passing that env var into the app bundle launch.

We're not trying to get the simplest approach, though, we're trying to simplify our long-term maintenance. This is why this should be submitted and worked with upstream.

dakcarto:
> We're not trying to get the simplest approach, though, we're trying to simplify our long-term maintenance. This is why this should be submitted and worked with upstream.

Good point. We discussed this a while back, and I think another option might be to use `configure` to help here:

1. Use `-plugindir <dir>` option set to `HOMEBREW_PREFIX/lib/qt5/plugins` during configure
2. Create the base directory structure in `HOMEBREW_PREFIX/lib/qt5/plugins`, e.g.

   ```
    $ tree /usr/local/lib/qt5/plugins
    /usr/local/lib/qt5/plugins
    ├── audio
    ├── bearer
    ├── canbus
    ├── designer
    ├── generic
    ├── geoservices
    ├── iconengines
    ├── imageformats
    ├── mediaservice
    ├── platforminputcontexts
    ├── platforms
    ├── playlistformats
    ├── position
    ├── printsupport
    ├── qmltooling
    ├── sceneparsers
    ├── sensorgestures
    ├── sensors
    └── sqldrivers
   ```
3. After `make install`, copy the `HOMEBREW_PREFIX/lib/qt5/plugins` directory back to `HOMEBREW_PREFIX/opt/qt5/plugins`
4. Remove any non-`qt5` formula plugin symlinks from the `HOMEBREW_PREFIX/opt/qt5/plugins`
5. Remove any  `qt5` formula plugins from `HOMEBREW_PREFIX/lib/qt5/plugins`

Then, the final symlinking steps of `qt5` (from formula prefix into `HOMEBREW_PREFIX/lib/qt5/plugins`) will work correctly, as well as any other formula's plugin symlink. During runtime, Qt5 references only the `HOMEBREW_PREFIX/lib/qt5/plugins` directory and everything just works.

This is a more robust solution, until there is support upstream for multiple plugin locations upon configure, as it supports all varieties of plugins, e.g. designer, sqldrivers, etc.

MikeMcQuaid:
> Use -plugindir <dir> option set to HOMEBREW_PREFIX/lib/qt5/plugins during configure

This seems like the best approach, yep. Will need to figure out how bottling/sandboxing works with this but it should be doable.

dakcarto:
@MikeMcQuaid  Ok, closed this PR, and will attempt `-plugindir` based PR. I assume that step 2-5 above will need to be in `post_install` so that they can be run after a bottle install, like `python*` formulae, yes?

MikeMcQuaid:
@dakcarto It'll likely require a bit of trial and error as the sandbox won't allow `install`ation outside of the `prefix` but: yeh, quite possibly.

PR #7273, opened 2016-11-27
ttfautohint 1.6

lemzwerg:
Also reactivate Qt support.

### DIFF
--- a/Formula/ttfautohint.rb
+++ b/Formula/ttfautohint.rb
@@ -1,8 +1,8 @@
 class Ttfautohint < Formula
   desc "Auto-hinter for TrueType fonts"
   homepage "https://www.freetype.org/ttfautohint"
-  url "https://downloads.sourceforge.net/project/freetype/ttfautohint/1.5/ttfautohint-1.5.tar.gz"
-  sha256 "644fe721e9e7fe3390ae1f66d40c74e4459fa539d436f4e0f8635c432683efd1"
+  url "https://downloads.sourceforge.net/project/freetype/ttfautohint/1.6/ttfautohint-1.6.tar.gz"
+  sha256 "e41013f710c306538ff5b2f1b4d9a5b24bda031fb73fabcaf02a21b8edd71be5"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


lemzwerg:
Can someone please help me interpret the failure messages?  What has ttfautohint to do with pdf2htmlex?  Is this *my* problem?  Were the actual build tests for ttfautohint (and ttfautohintGUI) successful?

ilovezfs:
pdf2htmlex depends on ttfautohint. So it is responsibility of a pull request that changes ttfautohint to make sure that it doesn't break pdf2htmlex.

As of https://github.com/Homebrew/legacy-homebrew/commit/cb6ca498 pdf2htmlex also uses a vendored version of fontforge.

The fontforge in the pdf2htmlex bottle links to giflib, but pdf2html doesn't declare a dependency on giflib. So `brew install pdf2htmlex` does not install giflib, and unless you happen to have giflib installed for some reason, the installation of pdf2htmlex is broken because it's linking to giflib files that aren't installed.

So pdf2htmlex is already currently broken regardless of this PR. So we have to decide whether to fix it before this PR, as part of this PR, or after this PR.

The good news is that the current version of pdf2htmlex looks like it may no longer need a vendored version of fontforge, and may be able to simply depend on the fontforge formula.

If you don't want to fix pdf2htmlex as part of this PR, I can do that for you, and then we can re-run CI on this PR. Let me know how you'd like to proceed.

lemzwerg:
Thanks for the explanation.  I'm not a real Homebrew user (running GNU/Linux all the time), only updating the formula for my ttfautohint program – and right now even away from my Mac that I use only to test whether the updated homebrew formula works...  So if you can manage to fix pdf2htmlex and re-run CI, this would be really a big help!

ilovezfs:
>Thanks for the explanation.

You're welcome.

>So if you can manage to fix pdf2htmlex and re-run CI, this would be really a big help!

Will do!

ilovezfs:
@BrewTestBot test this please

lemzwerg:
Thanks a lot for your help!  Everything seems fine now.

ilovezfs:
@lemzwerg you're welcome

ilovezfs:
@lemzwerg :rocket: thanks for your contribution to Homebrew!

PR #7278, opened 2016-11-27
pdf2htmlex: depend on cairo, freetype, and giflib

ilovezfs:


ilovezfs:
See https://github.com/Homebrew/homebrew-core/pull/7273#issuecomment-263117319.

PR #7284, opened 2016-11-27
dex2jar 2.1 alpha for multidex support

krokofant:


apjanke:
> https://github.com/pxb1988/dex2jar/releases/download/2.1-nightly-26/dex-tools-2.1-20150601.060031-26.zip

This looks like a nightly snapshot, not a stable release. Homebrew only uses stable releases for the main version of its packages. Alpha/beta/nightly and other prerelease versions should be added as a separate `devel` version.

fxcoudert:
The current stable release is still 2.0 for now. Closing this.

PR #7287, opened 2016-11-27
pangox-compat: migrate to x11

ilovezfs:
Created with `brew migrate-formula-pr`.


MikeMcQuaid:
Let's hold off any more tap migrations; I'd like to start going the other way in future.

ilovezfs:
The point is to get all of the rev-deps building without this. It can go in the boneyard as far as I'm concerned.

PR #7288, opened 2016-11-27
pangox-compat: migrate to boneyard

ilovezfs:


tschoonj:
Thanks for removing this formula! Should have done this myself last year when I migrated gtk+2 to quartz but did not realize that gtkglext had lost its pangox-compat dependency in the process...

MikeMcQuaid (approved):

PR #7303, opened 2016-11-28
libmongoc: fix note conflict with libbson

### LINKED ISSUE #7291: `brew install libmongoc`: Error: The `brew link` step did not complete successfully
```
==> Pouring libmongoc-1.4.2.el_capitan.bottle.tar.gz
The formula built, but is not symlinked into /Users/timothee/homebrew
Could not symlink include/libbson-1.0/bcon.h
Target /Users/timothee/homebrew/include/libbson-1.0/bcon.h
is a symlink belonging to libbson. You can unlink it:
  brew unlink libbson

To force the link and overwrite all conflicting files:
  brew link --overwrite libmongoc

To list all files that would be deleted:
  brew link --overwrite --dry-run libmongoc

Possible conflicting files are:
/Users/timothee/homebrew/include/libbson-1.0/bcon.h -> /Users/timothee/homebrew/Cellar/libbson/1.4.0/include/libbson-1.0/bcon.h
/Users/timothee/homebrew/include/libbson-1.0/bson-atomic.h -> /Users/timothee/homebrew/Cellar/libbson/1.4.0/include/libbson-1.0/bson-atomic.h
/Users/timothee/homebrew/include/libbson-1.0/bson-clock.h -> /Users/timothee/homebrew/Cellar/libbson/1.4.0/include/libbson-1.0/bson-clock.h

```

```
brew upgrade libbson
Error: libbson 1.4.0 already installed
```

```
brew gist-logs libmongoc
Error: No logs. 
```
(the `no logs` may be caused by https://github.com/Homebrew/legacy-homebrew/issues/44706 Unable to do `brew gist-logs` for certain formulas)

timkuijsten:

-----

fixes #7291

### DIFF
--- a/Formula/libbson.rb
+++ b/Formula/libbson.rb
@@ -12,6 +12,9 @@ class Libbson < Formula
     sha256 "4a10d2bb952da3287c4492feca37eaf481234bb904656a1fb166754f2a2c5f9f" => :mavericks
   end
 
+  conflicts_with "libmongoc",
+                 :because => "libmongoc comes with a bundled libbson"
+
   def install
     system "./configure", "--enable-silent-rules", "--prefix=#{prefix}"
     system "make", "install"
--- a/Formula/libmongoc.rb
+++ b/Formula/libmongoc.rb
@@ -11,6 +11,9 @@ class Libmongoc < Formula
     sha256 "9778915da0d362a4fe4e8c17ab1614cfe53d4c2f9cbe9a7c3ae6062fa8f3560c" => :yosemite
   end
 
+  conflicts_with "libbson",
+                 :because => "libmongoc comes with a bundled libbson"
+
   def install
     system "./configure", "--disable-debug",
                           "--disable-dependency-tracking",


apjanke:
Conflicts are symmetric: the `libbson` formula should probably also get a `conflicts_with "libmongoc"`. Would you mind adding that as part of this PR?

timkuijsten:
> Would you mind adding that as part of this PR?

Just did that :)

apjanke:
Thanks!

Looks good to me.

apjanke:
We should probably report this upstream to MongoDB, too. It's unusual behavior for a library to install its dependencies in the global location. Messes with dependency-tracking package managers. :)

timkuijsten:
Yeah, maybe @ajdavis knows of a better solution..

ajdavis:
Thanks everyone. I understand the arguments for unbundling libbson from libmongoc, but the arguments in *favor* of bundling outweigh them, so when you download libmongoc tarball you'll continue to get the libbson code along with it. *However*, when you're installing libmongoc as a Homebrew package, things should be different. I've just thought of a better approach for Homebrew.

Homebrew shouldn't consider these two packages as conflicting. Instead, libmongoc should depend on libbson. When libmongoc installs itself, it should add ``--with-libbson=system`` to its ``./configure`` arguments.

It may require a bit of tweaking to make that work, but I think it's the right approach.

Also, we need to ensure that libmongoc has the exact same version of libbson as its own version: if libmongoc is 1.4.2, libbson must also be 1.4.2.

apjanke:
> Homebrew shouldn't consider these two packages as conflicting. Instead, libmongoc should depend on libbson. When libmongoc installs itself, it should add --with-libbson=system to its ./configure arguments.

Sounds good to me. As long as there's a supported option to do this, Homebrew can work with it.

> Also, we need to ensure that libmongoc has the exact same version of libbson as its own version: if libmongoc is 1.4.2, libbson must also be 1.4.2.

We don't have a mechanism for enforcing this programmatically. We'd have to make do with comments in the formulae reminding maintainers to keep them in sync, which seems doable to me.

I did a quick pass at building libmongoc this way, and got an error in the `configure` step.

```

checking for BSON... no
configure: error:

  --------------------------------------
   The libbson-1.0 library could not be
   found on your system. Please install
   libbson-1.0  development  package or
   set --with-libbson=auto.
  --------------------------------------

```

Might be a version mismatch, since our `libmongoc` is at 1.4.2, but `libbson` is at 1.4.0. I'll try updating our `libbson` and fiddling with things and see if I can make it work.

Thanks for the input and quick response, @ajdavis!

ajdavis:
Happy to help - I'm not a Homebrew expert but I'll watch this ticket and pitch in where I can.

apjanke:
I've put in a PR #7305 to bump `libmongoc` to 1.5.0, which gets it in sync with our current `libbson`, which just got bumped to 1.5.0. That'll make it easier to do these potential changes.

apjanke:
Okay, Homebrew `libmongoc` is at 1.5.0 now, matching our `libbson`.

I tried building libmongoc with `--with-libbson=system` again, but still got that error. Full build logs here: https://gist.github.com/fc2e829805ffef20c95d9013c5da13dc

Not sure why it's not finding libbson. My pkg-config is reporting it as present.

```
$ pkg-config --libs libbson-1.0
-L/usr/local/Cellar/libbson/1.5.0/lib -lbson-1.0
```

apjanke:
At any rate, I think we should go ahead and merge this for now, adding the `conflicts_with` stanzas, since they correctly reflect the current state of our packages, and try to work on getting `libmongoc` to use our provided `libbson` after that. Any objections?

ajdavis:
Sounds good to me, for now.

apjanke:
Aha, got it working: I just needed to make `pkg-config` available to the Homebrew formula. See #7306.

With that change, `libmongoc` and `libbson` can be installed without conflict, and `libmongoc` does not install its bundled `libbson`, which is what we want for Homebrew.

timkuijsten:
Awesome progress!

Is it still possible to install 1.4.2? My package is only tested with that release.

apjanke:
> Is it still possible to install 1.4.2? My package is only tested with that release.

I'm afraid not; Homebrew only has support for one version of a package.

If you'd like to install 1.4.2, you can grab the old version of the formula and add it to a custom tap as `mongo-c-driver@1.4.2.rb` and install that; then you'll have full control over the exact version installed.

(I think you could also check out an earlier revision of the `homebrew-core` tap that still has the 1.4.2 `libmongoc` formula, install it from there, and then `brew pin libmongoc`. But that's not a very reproducible workflow.)

apjanke:
#7306 is now closed in favor of a larger refactor at #7352.

MikeMcQuaid:
https://github.com/Homebrew/homebrew-core/pull/7352 was merged but thanks for this PR.

PR #7308, opened 2016-11-28
ldc v1.1.0-beta5 (devel)

John-Colvin:

-----

### The following problem is now resolved with a patch taken from the relevant upstream pull request:

Doesn't pass audit because it installs `libLTO-ldc.dylib` in `bin`, which is currently where ldc looks for it.

2 parts to the problem: 1) where to put it and 2) how to get ldc to find it.

Things that can be done for 1):
* Allow it to install a non-executable in `bin` or
* Use a patch to modify `CmakeLists.txt` to make it install in another directory or
* Move it after `make install`

There is also the question of where it should go. `libexec` is a bit tempting, but `lib` makes sense as well.

and for 2)
* patch some code in `driver/linker.cpp` so ldc looks in a different place by default or
* patch `ldc_install.conf.in` to add `-flto-binary=#{lib}/libTLO-ldc.dylib`

Final option: wait for upstream to sort it out somehow.

### DIFF
--- a/Formula/ldc.rb
+++ b/Formula/ldc.rb
@@ -22,9 +22,14 @@ class Ldc < Formula
   end
 
   devel do
-    url "https://github.com/ldc-developers/ldc/releases/download/v1.1.0-beta3/ldc-1.1.0-beta3-src.tar.gz"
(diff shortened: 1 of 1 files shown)


dunn:
Has upstream been notified that it's a problem?

John-Colvin:
yes, upstream is aware

PR #7352, opened 2016-11-28
MongoDB drivers cleanup

### LINKED ISSUE #7291: `brew install libmongoc`: Error: The `brew link` step did not complete successfully

MikeMcQuaid:

-----

- Rename to upstream GitHub repository names
- Add aliases for various known names
- Update `mongo-cxx-driver` version
- Add tests
- Roll `libbson` into `mongo-c-driver` 

Fixes #7291.


Formula/mongo-cxx-driver.rb
```diff
@@ -0,0 +1,35 @@
...
+    system "make", "install"
+    pkgshare.install "examples"
+  end
+
+  test do
+    system ENV.cc, "-o", "test", pkgshare/"examples/mongocxx/connect.cpp",
+      "-I#{include}/mongocxx/v_noabi",
+      "-I#{include}/bsoncxx/v_noabi",
+      "-I#{Formula["mongo-c-driver"].include}/libmongoc-1.0",
+      "-I#{Formula["libbson"].include}/libbson-1.0",
```
apjanke: Should this be `Formula[mongo-c-driver]` now that `libbson` is no longer an independent formula?

Formula/mongo-c-driver.rb
```diff
@@ -0,0 +1,75 @@
...
+    (pkgshare/"mongoc").install "examples"
+  end
+
+  test do
+    system ENV.cc, "-o", "test", pkgshare/"examples/bson/json-to-bson.c",
+      "-I#{include}/libbson-1.0", "-L#{lib}", "-lbson-1.0"
+    (testpath/"test.json").write('{"name": "test"}')
+    assert_match "\u0000test\u0000", shell_output("./test test.json")
+
+    libbson = Formula["libbson"]
```
apjanke: This line looks unnecessary, since `libbson` is no longer a separate formula.

Formula/mongo-c-driver.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  bottle do
+    cellar :any
+    sha256 "1e3697dc3b2725914b720cb2aff628d1abeccfd26fc74ebe24bf3be83a5b5ba3" => :sierra
+    sha256 "f53836a2f6179cc4e65f6ef78c5a71792f9bd319676e9e8ef06d8fa5b4fafc5e" => :el_capitan
+    sha256 "c9fce4411ff385d090dc9e20529a045fa98b91801edba2732015f674fa581763" => :yosemite
+  end
+
+  mongo_c_driver_version = version
+  resource "libbson" do
```
apjanke: I'm getting a "resource defined more than once" error when trying to install. Does this `resource` need to be inside a `stable do` block to avoid it?

```
$ brew install mongo-c-driver
Error: Resource "libbson" is defined more than once
Please report this bug:
/usr/local/Homebrew/Library/Homebrew/software_spec.rb:101:in `resource'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/mongo-c-driver.rb:26:in `block in <class:MongoCDriver>'
/usr/local/Homebrew/Library/Homebrew/formula.rb:2091:in `instance_eval'
/usr/local/Homebrew/Library/Homebrew/formula.rb:2091:in `head'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/mongo-c-driver.rb:20:in `<class:MongoCDriver>'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/mongo-c-driver.rb:1:in `load_formula'
```

apjanke (dismissed):
:+1: in general on this cleanup, especially the improved names and tests.

Couple items that seem to need fixing; I couldn't get this to build locally yet.

Formula/mongo-c-driver.rb
```diff
@@ -0,0 +1,75 @@
...
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+
+    resource "libbson" do
+      url "https://github.com/mongodb/libbson.git"
+    end
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "openssl" => :recommended
```
timkuijsten: what about building with --enable-ssl=darwin instead of adding an extra dependency on openssl?

apjanke: Will that use OpenSSL or Secure Transport? Homebrew prefers using its own `openssl` instead of the system openssl, because the system openssl library is not updated, which has security implications.

timkuijsten: > The MongoDB C Driver supports the Darwin native TLS and crypto libraries. Using the native libraries there is no need to install OpenSSL. By default however, the driver will compile against OpenSSL if it detects it being available. If OpenSSL is not available, it will fallback on the native libraries. 
>
> To compile against the Darwin native TLS and crypto libraries, even when OpenSSL is available, configure the driver like so:
>
> $ ./configure --enable-ssl=darwin

From http://mongoc.org/libmongoc/current/installing.html#nativetls-macos

Formula/mongo-c-driver.rb
MikeMcQuaid: Yeh, need to fix this. Couldn't remember how we handled `resource`s for `head`.

Formula/mongo-c-driver.rb
MikeMcQuaid: Sounds good 👍

Formula/mongo-c-driver.rb
MikeMcQuaid: Thanks!

timkuijsten:
Awesome!

One note, I thought we went with libbson=system instead of bundled?

MikeMcQuaid:
@timkuijsten As nothing depends on `libbson` itself and `mongo-cxx-driver` builds both `bsoncxx` and `mongocxx` I thought it was worth just using the bundled version rather than trying to manually keep those versions in sync.

PR #7377, opened 2016-11-29
mu: fix url 404

rnkn:

-----

The old GitHub archive url resolved to 404. This seems right.

### DIFF
--- a/Formula/mu.rb
+++ b/Formula/mu.rb
@@ -4,10 +4,10 @@
 class Mu < Formula
   desc "Tool for searching e-mail messages stored in the maildir-format"
   homepage "https://www.djcbsoftware.nl/code/mu/"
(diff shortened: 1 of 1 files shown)


nijikon:
@ilovezfs should we bump revision?

ilovezfs:
@nijikon yes

rnkn:
Are revisions incremented with every formula change, or just to those for the same formula version? (e.g. would version 0.9.16 -> 0.9.17 trigger a bump?)

ilovezfs:
Only if the version stays the same but we want to force an upgrade. In this case, there's no way to be sure the old tarball and new tarball are exactly the same (unless someone has a copy of the old one lying around with the right checksum) since it's 404ing, so forcing the upgrade will make sure 0.9.16 only means one thing.

nijikon:

PR #7379, opened 2016-11-29
git-annex-map: use xdot instead of dot 

palday:

Using `xdot` instead of `dot -Tx11` gives the  correct display behavior on OS X. 

`xdot` is only marked as recommended because this is not essential behavior and you can manually generate desired output after running  `git annex map --fast` even without `xdot` if you have other software that understands the GraphViz format.

### DIFF
--- a/Formula/git-annex.rb
+++ b/Formula/git-annex.rb
@@ -26,6 +26,7 @@ class GitAnnex < Formula
   depends_on "libmagic"
   depends_on "gnutls"
   depends_on "quvi"
+  depends_on "xdot" => :recommended
 
   resource "esqueleto-2.4.3" do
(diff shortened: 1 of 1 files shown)


nijikon:
@ilovezfs need second opinion, imho it's ok

ilovezfs:
@nijikon it'll need to be reported [here](http://git-annex.branchable.com/bugs/) before it goes in, and we'll need some indication from @joeyh that he intends to incorporate it in some way.

palday:
I'm not sure it's really a bug, but rather a packaging issue. Is there a release of `dot` for Mac that supports `-Tx11`?
________________________________

joeyh:
I was not aware of xdot. It's a better interactive viewer so I will make
git-annex use it by default.

-- 
see shy jo

ilovezfs (approved):

nijikon:

PR #7381, opened 2016-11-29
clipper 0.4

00dani:
Note that the formula doesn't pass `brew audit --strict`, because it doesn't have a `test do` block. It didn't have one before this commit either, though.


woodruffw:
Not a blocking issue, but a `test` block would be nice. Since it looks like `clipper` has a somewhat involved setup, perhaps just a test for socket creation or `--help`/`--version` if the former doesn't work out? Thanks!

00dani:
Sure thing! The `test` I just pushed actually tries to copy some text through `clipper`, confirming it works using `pbpaste`. 👌

nijikon:

PR #7386, opened 2016-11-29
yelp-tools 3.20.1

ajdavis:
This is a work in progress. I've added two dependencies that appear to be required now, but linking fails with "ld: unknown option: --no-as-needed".

```
==> ./configure --prefix=/usr/local/Cellar/yelp-tools/3.20.1
... snip ...
==> make install
... snip ...
  CCLD     libyelp/libyelp.la
  CCLD     libyelp/web-extension/libyelpwebextension.la
  CCLD     yelp
ld: unknown option: --no-as-needed
make[2]: *** [yelp] Error 1
make[1]: *** [install-recursive] Error 1
make: *** [install] Error 2
```

Perhaps we should disable some of the new yelp features in 3.20 by passing some disable options to configure? I'm taking a break from this for now.

### DIFF
--- a/Formula/yelp-tools.rb
+++ b/Formula/yelp-tools.rb
@@ -20,11 +20,9 @@ class YelpTools < Formula
   depends_on "libxslt" => :build
   depends_on "pkg-config" => :build
 
-  yelp_tools_version = version
-
   resource "yelp-xsl" do
-    url "https://download.gnome.org/sources/yelp-xsl/3.18/yelp-xsl-#{yelp_tools_version}.tar.xz"
-    sha256 "893620857b72b3b43ee3b462281240b7ca4d80292f469552827f0597bf60d2b2"
+    url "https://download.gnome.org/sources/yelp-xsl/3.20/yelp-xsl-3.20.1.tar.xz"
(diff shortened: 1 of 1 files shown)


tschoonj:
Wow, you used `yelp` instead of `yelp-tools` in the url, which is a very different package. 

As far as I can tell, there has been no `yelp-tools` 3.20.x release, so you will have to continue using 3.18.0. Only the `yelp-xsl` resource needs to get an update here.

ajdavis:
Jeez, sorry about that! Updated. This installs yelp-tools 3.18.0 and yelp-xsl 3.20.1. It does not reintroduce #7243, however, because the yelp-xsl contents are now a resource of yelp-tools instead of a separate formula.

tschoonj:
Thanks for changing this.

Would you mind changing the urls back to download.gnome.org?

ajdavis:
Done, thanks.

tschoonj:
urls are still wrong I am afraid 😢

ajdavis:
Thanks for your patience, I'm new here. Updated.

tschoonj:
Don't worry, you're doing fine :+1:

woodruffw:
Thanks @ajdavis!

PR #7410, opened 2016-11-30
qjson: restore from boneyard

ilovezfs:
0.9.0 added qt5 support


MikeMcQuaid:
I wasn't going to do it but as you have: 👏 👍

ilovezfs:
@MikeMcQuaid well I feel obligated given that I filed https://github.com/flavio/qjson/issues/95 hehe

PR #2549, opened 2016-07-01
berkeley-db 6.2.23

zmwangx:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

I happened to notice that Berkeley DB is outdated (when I was reading [this](http://arstechnica.com/information-technology/2016/07/how-oracles-business-as-usual-is-threatening-to-kill-java/) 😜), so I'm submitting this PR. Since it is only a drive-by update, I might not commit to fixing potential failures of dependents.

I also added a test, but am having some trouble with the sandbox:

```
$ brew test --no-sandbox berkeley-db
Testing berkeley-db
==> /usr/bin/clang++ test.cpp -o test -Os -w -pipe -march=native -mmacosx-version-min=10.11 -F/usr/local/Frameworks -L/usr/local/lib -F/usr/local/Frameworks -Wl,-headerpad_max_install_names -I/usr/local/Cellar/berkeley-db/6.2.23/include -L/usr/local/Cellar/berkeley-db/6.2.23/lib -ldb_cxx
==> ./test
```

```
$ brew test --sandbox berkeley-db
Testing berkeley-db
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160701-46650-1dnexye.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-core/Formula/berkeley-db.rb --sandbox
==> /usr/bin/clang++ test.cpp -o test -Os -w -pipe -march=native -mmacosx-version-min=10.11 -F/usr/local/Frameworks -L/usr/local/lib -F/usr/local/Frameworks -Wl,-headerpad_max_install_names -I/usr/local/Cellar/berkeley-db/6.2.23/include -L/usr/local/Cellar/berkeley-db/6.2.23/lib -ldb_cxx
[...]

PR #7421, opened 2016-11-30
berkeley-db: up formula version to 6.2.23, enable dbm, depend on java

maxim-belkin:

-----

This PR does several things:
1. updates `berkeley-db` to 6.2.23.
1. updates default options to build with java and sql support
1. adds `:java` dependency
1. adds `dbm` support (needed for some formulas)

Related (and closed) PRs: #2549 and #3263

```txt
$ brew audit --strict berkeley-db
berkeley-db:
  * JARs were installed to "/usr/local/Cellar/berkeley-db/6.2.23/lib"
    Installing JARs to "lib" can cause conflicts between packages.
    For Java software, it is typically better for the formula to
    install to "libexec" and then symlink or wrap binaries into "bin".
    See "activemq", "jruby", etc. for examples.
    The offending files are:
      /usr/local/Cellar/berkeley-db/6.2.23/lib/db.jar
Error: 1 problem in 1 formula
```

### DIFF
--- a/Formula/berkeley-db.rb
+++ b/Formula/berkeley-db.rb
@@ -1,8 +1,8 @@
 class BerkeleyDb < Formula
   desc "High performance key/value database"
   homepage "https://www.oracle.com/technology/products/berkeley-db/index.html"
-  url "http://download.oracle.com/berkeley-db/db-6.1.26.tar.gz"
-  sha256 "dd1417af5443f326ee3998e40986c3c60e2a7cfb5bfa25177ef7cadb2afb13a6"
+  url "http://download.oracle.com/berkeley-db/db-6.2.23.tar.gz"
+  sha256 "47612c8991aa9ac2f6be721267c8d3cdccf5ac83105df8e50809daea24e95dc7"
 
   bottle do
     cellar :any
@@ -13,10 +13,10 @@ class BerkeleyDb < Formula
     sha256 "8f5a87bc3336e01f8528ae8aaae24c83e8fa10e8f01ee65e8cac4af7d0786bf1" => :mountain_lion
   end
 
-  option "with-java", "Compile with Java support."
-  option "with-sql", "Compile with SQL support."
+  option "without-sql", "Build without SQL support."
+  option "without-dbm", "Build without dbm interface."
 
-  deprecated_option "enable-sql" => "with-sql"
+  depends_on :java => [:recommended, :build]
 
   def install
     # BerkeleyDB dislikes parallel builds
@@ -32,6 +32,7 @@ class BerkeleyDb < Formula
     ]
     args << "--enable-java" if build.with? "java"
     args << "--enable-sql" if build.with? "sql"
+    args << "--enable-dbm" if build.with? "dbm"
 
     # BerkeleyDB requires you to build everything from the build_unix subdirectory
     cd "build_unix" do
--- a/Formula/berkeley-db.rb
+++ b/Formula/berkeley-db.rb
@@ -13,9 +13,6 @@ class BerkeleyDb < Formula
     sha256 "8f5a87bc3336e01f8528ae8aaae24c83e8fa10e8f01ee65e8cac4af7d0786bf1" => :mountain_lion
   end
 
-  option "without-sql", "Build without SQL support."
-  option "without-dbm", "Build without dbm interface."
-
   depends_on :java => [:optional, :build]
 
   def install
@@ -29,10 +26,11 @@ class BerkeleyDb < Formula
       --mandir=#{man}
       --enable-cxx
       --enable-compat185
+      --enable-sql
+      --enable-sql_codegen
+      --enable-dbm
     ]
     args << "--enable-java" if build.with? "java"
-    args << "--enable-sql" if build.with? "sql"
-    args << "--enable-dbm" if build.with? "dbm"
 
     # BerkeleyDB requires you to build everything from the build_unix subdirectory
     cd "build_unix" do
--- a/Formula/berkeley-db.rb
+++ b/Formula/berkeley-db.rb
@@ -16,7 +16,7 @@ class BerkeleyDb < Formula
   option "without-sql", "Build without SQL support."
   option "without-dbm", "Build without dbm interface."
 
-  depends_on :java => [:recommended, :build]
+  depends_on :java => [:optional, :build]
 
   def install
     # BerkeleyDB dislikes parallel builds
--- a/Formula/berkeley-db.rb
+++ b/Formula/berkeley-db.rb
@@ -29,6 +29,7 @@ class BerkeleyDb < Formula
       --enable-sql
       --enable-sql_codegen
       --enable-dbm
+      --enable-stl
     ]
     args << "--enable-java" if build.with? "java"
 
--- a/Formula/darkice.rb
+++ b/Formula/darkice.rb
@@ -22,6 +22,9 @@ class Darkice < Formula
   depends_on "jack"
 
   def install
+    # Fixes  "invalid conversion from ‘const float*’ to ‘float*’ [-fpermissive]"
+    # Upstream issue Oct 25, 2016 https://github.com/rafael2k/darkice/issues/119
+    # Suggested fix  Oct 25, 2016 https://github.com/rafael2k/darkice/pull/120
     ["aacPlusEncoder.cpp", "FaacEncoder.cpp", "OpusLibEncoder.cpp", "VorbisLibEncoder.cpp"].each do |f|
       inreplace "src/#{f}", ", converterData.data_in", ", const_cast<float*>( converterData.data_in )"
     end
--- a/Formula/darkice.rb
+++ b/Formula/darkice.rb
@@ -3,6 +3,7 @@ class Darkice < Formula
   homepage "http://www.darkice.org/"
   url "https://downloads.sourceforge.net/project/darkice/darkice/1.3/darkice-1.3.tar.gz"
   sha256 "2c0d0faaa627c0273b2ce8b38775a73ef97e34ef866862a398f660ad8f6e9de6"
+  revision 1
 
   bottle do
     cellar :any
--- a/Formula/darkice.rb
+++ b/Formula/darkice.rb
@@ -22,7 +22,7 @@ class Darkice < Formula
   depends_on "jack"
 
   def install
-    # Fixes  "invalid conversion from ‘const float*’ to ‘float*’ [-fpermissive]"
+    # Fixes  "invalid conversion from 'const float*' to 'float*' [-fpermissive]"
     # Upstream issue Oct 25, 2016 https://github.com/rafael2k/darkice/issues/119
     # Suggested fix  Oct 25, 2016 https://github.com/rafael2k/darkice/pull/120
     ["aacPlusEncoder.cpp", "FaacEncoder.cpp", "OpusLibEncoder.cpp", "VorbisLibEncoder.cpp"].each do |f|
--- a/Formula/darkice.rb
+++ b/Formula/darkice.rb
@@ -22,6 +22,9 @@ class Darkice < Formula
   depends_on "jack"
 
   def install
+    ["aacPlusEncoder.cpp", "FaacEncoder.cpp", "OpusLibEncoder.cpp", "VorbisLibEncoder.cpp"].each do |f|
+      inreplace "src/#{f}", ", converterData.data_in", ", const_cast<float*>( converterData.data_in )"
+    end
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
                           "--sysconfdir=#{etc}",
--- a/Formula/bogofilter.rb
+++ b/Formula/bogofilter.rb
@@ -3,6 +3,7 @@ class Bogofilter < Formula
   homepage "http://bogofilter.sourceforge.net"
   url "https://downloads.sourceforge.net/project/bogofilter/bogofilter-1.2.4/bogofilter-1.2.4.tar.bz2"
   sha256 "e10287a58d135feaea26880ce7d4b9fa2841fb114a2154bf7da8da98aab0a6b4"
+  revision 1
 
   bottle do
     sha256 "92301b235bfd5088a5b20fa263de2c012f31f90860f3b1b1fd93a918d93fff7c" => :sierra
--- a/Formula/dbxml.rb
+++ b/Formula/dbxml.rb
@@ -3,6 +3,7 @@ class Dbxml < Formula
   homepage "https://www.oracle.com/us/products/database/berkeley-db/xml/overview/index.html"
   url "http://download.oracle.com/berkeley-db/dbxml-6.0.18.tar.gz"
   sha256 "5851f60a47920718b701752528a449f30b16ddbf5402a2a5e8cde8b4aecfabc8"
+  revision 1
 
   bottle do
     cellar :any
--- a/Formula/dirt.rb
+++ b/Formula/dirt.rb
@@ -3,6 +3,7 @@ class Dirt < Formula
   homepage "https://github.com/tidalcycles/Dirt"
   url "https://github.com/tidalcycles/Dirt/archive/1.1.tar.gz"
   sha256 "bb1ae52311813d0ea3089bf3837592b885562518b4b44967ce88a24bc10802b6"
+  revision 1
   head "https://github.com/tidalcycles/Dirt.git"
 
   bottle do
--- a/Formula/gnu-cobol.rb
+++ b/Formula/gnu-cobol.rb
@@ -1,7 +1,7 @@
 class GnuCobol < Formula
   desc "Implements much of the COBOL 85 and COBOL 2002 standards"
   homepage "http://www.opencobol.org/"
-  revision 3
+  revision 4
 
   stable do
     url "https://downloads.sourceforge.net/project/open-cobol/gnu-cobol/1.1/gnu-cobol-1.1.tar.gz"
--- a/Formula/gnuradio.rb
+++ b/Formula/gnuradio.rb
@@ -3,7 +3,7 @@ class Gnuradio < Formula
   homepage "https://gnuradio.squarespace.com/"
   url "https://gnuradio.org/releases/gnuradio/gnuradio-3.7.9.1.tar.gz"
   sha256 "9c06f0f1ec14113203e0486fd526dd46ecef216dfe42f12d78d9b781b1ef967e"
-  revision 1
+  revision 2
 
   bottle do
     rebuild 1
--- a/Formula/jack.rb
+++ b/Formula/jack.rb
@@ -10,6 +10,7 @@ class Jack < Formula
   homepage "http://jackaudio.org"
   url "http://jackaudio.org/downloads/jack-audio-connection-kit-0.125.0.tar.gz"
   sha256 "3517b5bff82139a76b2b66fe2fd9a3b34b6e594c184f95a988524c575b11d444"
+  revision 1
 
   bottle do
     sha256 "09037cc758f7fbb1906e9f57722143ac6fad688dafc8a3737b23ee6cf60f3277" => :sierra
--- a/Formula/jigdo.rb
+++ b/Formula/jigdo.rb
@@ -3,7 +3,7 @@ class Jigdo < Formula
   homepage "http://atterer.org/jigdo/"
   url "http://atterer.org/sites/atterer/files/2009-08/jigdo/jigdo-0.7.3.tar.bz2"
   sha256 "875c069abad67ce67d032a9479228acdb37c8162236c0e768369505f264827f0"
-  revision 2
+  revision 3
 
   bottle do
     rebuild 2
--- a/Formula/ltc-tools.rb
+++ b/Formula/ltc-tools.rb
@@ -4,6 +4,7 @@ class LtcTools < Formula
   url "https://github.com/x42/ltc-tools/archive/v0.6.4.tar.gz"
   sha256 "8fc9621df6f43ab24c65752a9fee67bee6625027c19c088e5498d2ea038a22ec"
   head "https://github.com/x42/ltc-tools.git"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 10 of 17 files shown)


Formula/berkeley-db.rb
```diff
@@ -13,10 +13,10 @@ class BerkeleyDb < Formula
     sha256 "8f5a87bc3336e01f8528ae8aaae24c83e8fa10e8f01ee65e8cac4af7d0786bf1" => :mountain_lion
   end
 
-  option "with-java", "Compile with Java support."
-  option "with-sql", "Compile with SQL support."
+  option "without-sql", "Build without SQL support."
+  option "without-dbm", "Build without dbm interface."
 
-  deprecated_option "enable-sql" => "with-sql"
+  depends_on :java => [:recommended, :build]
```
MikeMcQuaid: Shouldn't build with Java by default because macOS doesn't ship with Java by default.

MikeMcQuaid (changes requested):

maxim-belkin:
Looks like there is a number of failures due to non-existent 
```
/usr/local/opt/berkeley-db/lib/libdb-6.1.dylib
```

zmwangx:
I attempted this in #2549. rpm was the only blocker back then.

maxim-belkin:
@zmwangx yes, I saw that one and referenced it in the original post

zmwangx:
Ah right. Just wanted to point it out to potentially save some research.

MikeMcQuaid:
@BrewTestBot test this please

maxim-belkin:
What's going on with downloaded files? They have the same SHA256 :)
```
=> brew fetch --retry bogofilter --build-bottle --force
==> FAILED
==> Downloading https://downloads.sourceforge.net/project/bogofilter/bogofilter-1.2.4/bogofilter-1.2.4.tar.bz2
==> Downloading from http://downloads.sourceforge.net/#!/project/bogofilter/bogofilter-1.2.4/bogofilter-1.2.4.tar.bz2
Downloaded to: /Users/brew/Library/Caches/Homebrew/bogofilter-1.2.4.tar.bz2
SHA256: 818a4b8bbcb50878a8b1b9f71b4274d242ab46bf860c74676e98dec1d0248821
```
and
```
==> brew fetch --retry darkice --build-bottle --force
==> FAILED
==> Downloading https://downloads.sourceforge.net/project/darkice/darkice/1.3/darkice-1.3.tar.gz
==> Downloading from http://downloads.sourceforge.net/#!/project/darkice/darkice/1.3/darkice-1.3.tar.gz
Downloaded to: /Users/brew/Library/Caches/Homebrew/darkice-1.3.tar.gz
SHA256: 818a4b8bbcb50878a8b1b9f71b4274d242ab46bf860c74676e98dec1d0248821
```

maxim-belkin:
And thanks for starting this test because it is blocking some of my work in Linuxbrew-land.

maxim-belkin:
@MikeMcQuaid could you please restart the build?

maxim-belkin:
my gist-logs for [rpm](https://gist.github.com/1d688342a9349ee589f44cf98f34c3fd) and [bogofilter](https://gist.github.com/753a60127e5210295f31afb9175b85fd)

MikeMcQuaid (approved):

maxim-belkin:
fixed `rpm` test locally.
I have a solution for `darkice` but it requires modifications to several source code files in `darkice` OR one header file in `libsamplerate`

maxim-belkin:
Fixed `darkice` and `rpm`'s testing section. Ready to be re-tested ( @MikeMcQuaid ) .

maxim-belkin:
Weird failure on El Capitan: the only part that failed is the development build of `moc`. Note that El Capitan uses SDK for `10.12`

```log
==> brew install --devel moc --verbose
==> FAILED
```
then:
```log
==> Installing moc
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20161209-86963-1nsoq7g.sb nice /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/build.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/moc.rb --verbose --devel
==> Downloading http://ftp.daper.net/pub/soft/moc/unstable/moc-2.6-alpha2.tar.xz
Already downloaded: /Users/brew/Library/Caches/Homebrew/moc-2.6-alpha2.tar.xz
==> Verifying moc-2.6-alpha2.tar.xz checksum
tar xJf /Users/brew/Library/Caches/Homebrew/moc-2.6-alpha2.tar.xz
==> ./configure --disable-debug --prefix=/usr/local/Cellar/moc/2.6-alpha2_1
...
... some output ...
...
Making install in vorbis
/bin/sh ../../libtool  --tag=CC   --mode=compile clang -DHAVE_CONFIG_H -I. -I../..    -I/usr/local/Cellar/libogg/1.3.2/include -I/usr/local/Cellar/libvorbis/1.3.5/include -I../.. -D_THREAD_SAFE -pthread -g -O2 -Wall -W -I/usr/local/Cellar/jack/0.125.0_1/include -I/usr/local/Cellar/libsamplerate/0.1.9/include  -MT libvorbis_decoder_la-vorbis.lo -MD -MP -MF .deps/libvorbis_decoder_la-vorbis.Tpo -c -o libvorbis_decoder_la-vorbis.lo `test -f 'vorbis.c' || echo './'`vorbis.c
libtool: compile:  clang -DHAVE_CONFIG_H -I. -I../.. -I/usr/local/Cellar/libogg/1.3.2/include -I/usr/local/Cellar/libvorbis/1.3.5/include -I../.. -D_THREAD_SAFE -pthread -g -O2 -Wall -W -I/usr/local/Cellar/jack/0.125.0_1/include -I/usr/local/Cellar/libsamplerate/0.1.9/include -MT libvorbis_decoder_la-vorbis.lo -MD -MP -MF .deps/libvorbis_decoder_la-vorbis.Tpo -c vorbis.c  -fno-common -DPIC -o .libs/libvorbis_decoder_la-vorbis.o
In file included from vorbis.c:35:
In file included from ../../decoder.h:6:
In file included from ../../io.h:5:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/pthread.h:60:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/time.h:177:5: error: conflicting types for 'clock_gettime'
int clock_gettime(clockid_t __clock_id, struct timespec *__tp);
    ^
../../compat.h:77:5: note: previous declaration is here
int clock_gettime (int clk_id, struct timespec *ts);
    ^
1 error generated.
make[2]: *** [libvorbis_decoder_la-vorbis.lo] Error 1
make[1]: *** [install-recursive] Error 1

==> Formula
Path: /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/moc.rb
```

ilovezfs:
This fixes it
```
diff --git a/Formula/moc.rb b/Formula/moc.rb
index 12c9aa7..9d39a8c 100644
--- a/Formula/moc.rb
+++ b/Formula/moc.rb
@@ -45,6 +45,11 @@ class Moc < Formula
   depends_on "homebrew/dupes/ncurses" => :optional
 
   def install
+    # Fix "error: conflicting types for 'clock_gettime'"
+    if build.devel? && MacOS.version == :el_capitan && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+      inreplace Dir["{*.c,compat.h}"], "clock_gettime", "moc_clock_gettime", false
+    end
+
     system "autoreconf", "-fvi" if build.head?
     system "./configure", "--disable-debug", "--prefix=#{prefix}"
     system "make", "install"
```
@maxim-belkin mind reporting it upstream to moc? :innocent:

maxim-belkin:
A perfect stranger ( wink-wink @zmwangx ) already reported it [here](https://moc.daper.net/node/1576)
I can implement your fix and bump up the revision of development version of moc to 2.5.3. Should I do that?

ilovezfs:
@maxim-belkin the name rings a bell.

>I can implement your fix and bump up the revision of development version of moc to 2.5.3. Should I do that?

Pretty please, with sugar on top. Also add a link to rings-a-bell's report.

zmwangx:
I've been having some back-and-force with upstream maintainer John Fitzgerald (over private email), with me explaining the clock_gettime clusterfuck and both of us trying to come up ways to fix this in upstream code base satisfactorily. For instance, I've suggested `AC_CHECK_FUNCS` or `AC_SEARCH_LIBS` (reminding him of `SDKROOT` and our `ac_cv_func_clock_gettime` and `ac_cv_search_clock_gettime`); or renaming the function and define it unconditionally for Darwin:

```diff
diff --git a/compat.c b/compat.c
index df02a1d..5b0bcc5 100644
--- a/compat.c
+++ b/compat.c
@@ -49,13 +49,13 @@ char *strcasestr (const char *haystack, const char *needle)
#endif

/* OSX doesn't provide clock_gettime(3) so fall back to gettimeofday(2). */
-#ifndef HAVE_CLOCK_GETTIME
+#if !defined HAVE_CLOCK_GETTIME || defined __APPLE__
#include <sys/time.h>
#include <assert.h>

#include "common.h"

-int clock_gettime (int clk_id ASSERT_ONLY, struct timespec *ts)
+int my_clock_gettime (int clk_id ASSERT_ONLY, struct timespec *ts)
{
	int result;
	struct timeval tv;
diff --git a/compat.h b/compat.h
index 71a011e..27c0283 100644
--- a/compat.h
+++ b/compat.h
@@ -50,10 +50,12 @@ extern "C" {
char *strcasestr (const char *haystack, const char *needle);
#endif

-#ifndef HAVE_CLOCK_GETTIME
+#if !defined HAVE_CLOCK_GETTIME || defined __APPLE__
#define CLOCK_REALTIME 0
struct timespec;
-int clock_gettime (int clk_id, struct timespec *ts);
+int my_clock_gettime (int clk_id, struct timespec *ts);
+#else
+#define my_clock_gettime clock_gettime
#endif

#ifdef __cplusplus
diff --git a/interface.c b/interface.c
index e8c3e08..976f023 100644
--- a/interface.c
+++ b/interface.c
@@ -2581,7 +2581,7 @@ static time_t rounded_time ()
	struct timespec exact_time;
	time_t curr_time;

-	if (clock_gettime (CLOCK_REALTIME, &exact_time) == -1)
+	if (my_clock_gettime (CLOCK_REALTIME, &exact_time) == -1)
		interface_fatal ("clock_gettime() failed: %s", xstrerror (errno));

	curr_time = exact_time.tv_sec;
```

He was unsatisfied with both approaches and decided that he would use a `host_os` based check (no no to me) in his latest email to me (from yesterday).

Whatever the outcome is, rest assured that upstream really wants to fix this.

ilovezfs:
@zmwangx you may want to share this with him: https://github.com/jsquyres/libfabric/commit/469779935f9995b88a6f486fe3c4c4496989af36

zmwangx:
@ilovezfs We've already discussed `AC_LINK_IFELSE` and such, I'm sure he won't be thrilled to carry that `ofi_clock_gettime.m4` in moc's codebase. However, feel like contributing to Autoconf Archive, or asking Jeff Squyres to do that?

ilovezfs:
@zmwangx considering Jeff actually decided to remove clock_gettime entirely from their codebase that somehow feels wrong lol

ilovezfs:
@zmwangx https://github.com/ofiwg/libfabric/pull/2508.

zmwangx:
@ilovezfs John considered that (there's only one clock_gettime call after all, and it's not even essential, it's only called when seeking forward/backward), but he had concerns over precision of alternate implementations.

I feel bad for all these poor Linux souls who are now forced to clean after Apple's mess.

zmwangx:
By the way, re rpm @maxim-belkin: are you sure `/var/lib/rpm/sqldb` is the correct replacement for `/var/lib/rpm/Packages`? Last time I looked at the this I thought it might be due to a subtly broken build, but I don't know enough about rpm (well, I know almost nothing) to tell either way.

ilovezfs:
@zmwangx there are two separate issues really: how to detect the evil case, and what to do about it. Assuming the former is not a problem, it seems like changing the name of the function, using a toll free bridge when the platform has the real clock_gettime, and otherwise using one of the preexisting compat implementations is pretty uncontroversial, no?

zmwangx:
@ilovezfs 

> ... pretty uncontroversial, no?

To me it's pretty uncontroversial. However, I wrote

> 3. Ignore Darwin's clock_gettime. Something like the patch attached at
  the end.

and John wrote

> Yeah, I did come up with a similar patch, but decided it got too cryptic.

A little bit of context: a while ago John discarded a lot of compat stuff and tried to make the code base as POSIX as possible. It's understandable he doesn't want something like `moc_clock_gettime` in the C code. But then you into all the problems with the heavily contended symbol `clock_gettime`.

ilovezfs:
Given the limitations of the language, a wrapper seems preferable to littering the code with #if this #else that macros.

zmwangx:
@ilovezfs You will have cpp conditionals anyway. A point can be made to limit tweaks to `compat.{c,h}` and pretend elsewhere (i.e. `interface.c`) that you have unfettered access to realtime extensions.

ilovezfs:
@zmwangx yup, which is why "It's understandable he doesn't want something like moc_clock_gettime in the C code," is not actually understandable to me.

zmwangx:
@ilovezfs Well, if you could "pretend elsewhere (i.e. interface.c) that you have unfettered access to realtime extensions", then you would just call `clock_gettime`, not `moc_clock_gettime`...

ilovezfs:
@zmwangx I see, but tough to avoid including time.h indirectly at some point, no?

zmwangx:
@ilovezfs Never mind, got the patch: https://raw.githubusercontent.com/Homebrew/patches/b7814c0/moc/moc-r2935%2Bclock_gettime-3.patch

Basically:

```
dnl OSX / MacOS doesn't provide clock_gettime(3) prior to darwin-16.0.0
dnl so fall back to gettimeofday(2).
case "$host_os" in
[*darwin[1-9].*] | [*darwin1[0-5].*])
	HAVE_CLOCK_GETTIME="no"
	;;
*darwin*)
	HAVE_CLOCK_GETTIME="yes"
	;;
*)
	AC_CHECK_LIB([rt], [clock_gettime],
		[HAVE_CLOCK_GETTIME="yes"],
		[HAVE_CLOCK_GETTIME="no"])
	;;
esac
```

```diff
diff --git a/Formula/moc.rb b/Formula/moc.rb
index 12c9aa70cd..0756bae400 100644
--- a/Formula/moc.rb
+++ b/Formula/moc.rb
@@ -11,9 +11,18 @@ class Moc < Formula
   end
 
   devel do
-    url "http://ftp.daper.net/pub/soft/moc/unstable/moc-2.6-alpha2.tar.xz"
-    sha256 "0a3a4fb11227ec58025f7177a3212aca9c9955226a2983939e8db662af13434b"
+    url "http://ftp.daper.net/pub/soft/moc/unstable/moc-2.6-alpha3.tar.xz"
+    sha256 "a27b8888984cf8dbcd758584961529ddf48c237caa9b40b67423fbfbb88323b1"
 
+    patch do
+      url "https://raw.githubusercontent.com/Homebrew/patches/b7814c0/moc/moc-r2935%2Bclock_gettime-3.patch"
+      sha256 "601b5cdf59db67f180f1aaa6cc90804c1cb69c44cdecb2e8149338782e4f21a8"
+    end
+
+    # Remove build deps when next devel release comes out
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "gettext" => :build
     depends_on "popt"
   end
 
@@ -45,7 +54,8 @@ class Moc < Formula
   depends_on "homebrew/dupes/ncurses" => :optional
 
   def install
-    system "autoreconf", "-fvi" if build.head?
+    # Remove build.devel? when next devel release comes out
+    system "autoreconf", "-fvi" if build.head? || build.devel?
     system "./configure", "--disable-debug", "--prefix=#{prefix}"
     system "make", "install"
   end
```

MikeMcQuaid:
> +    # Remove build deps when next devel release comes out

Probably worth noting a specific version or even adding a `test` or this will get left indefinitely.

zmwangx:
> Probably worth noting a specific version or even adding a test or this will get left indefinitely.

My patch is only meant for testing... I think

```rb
inreplace Dir["{*.c,compat.h}"], "clock_gettime", "moc_clock_gettime", false
```

is good enough for now.

zmwangx:
By the way, the `clock_gettime` issue is subtle in moc. Even if `clock_gettime` is linked on 10.11, running `mocp` won't fail immediately; only if you actually play a song in it and try to seek forward or backward will it error out with the dreaded

```
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
```

Of course we can't test that.

The only reliable test that I know of is

```rb
system "/usr/bin/nm #{bin}/mocp | grep clock_gettime"
```

maxim-belkin:
build on sierra failed...

zmwangx:
> build on sierra failed...

Because the replacement needs to happen on 10.12 too (`HAVE_CLOCK_GETTIME` isn't set there thanks to looking for `clock_gettime` in the nonexistent librt). Now that there's an official patch though, let's just use that one (at the expense of having to regenerate configure).

I've opened #7734 to bump moc devel so that we don't need to keep rebuilding everything else here. Might as well rebase after #7734 is merged; berkeley-db 6.2.23 has been delayed for the better part of a year.

maxim-belkin:
OK, I'll rebase this PR and remove all the fixes for `moc` once your PR is merged.

maxim-belkin:
rebased and force-pushed

maxim-belkin:
@ilovezfs have a look at error messages on Jenkins -- I think those errors are Jenkins' problem. Can you restart the build?

maxim-belkin:
@zmwangx regarding `rpm`: I don't know but I see `sqldb`, not `Packages` when I run tests. I checked that this file is an sql database with `Packages` table in it. It could be a combination of factors that leads to `sqldb` file and not a bunch of Berkeley-DB files one of which if `Packages`

ilovezfs:
@BrewTestBot test this please

maxim-belkin:
Strange failures again. Is there something wrong with Jenkins or does it indicate a problem with formulae?

MikeMcQuaid:
Will try to rebuild.

MikeMcQuaid:
@BrewTestBot test this please

maxim-belkin:
Can someone take a look at this PR? Looks like all the checks are passed now.

MikeMcQuaid (approved):

MikeMcQuaid:
Looks good to me. @ilovezfs 👍 ?

Formula/darkice.rb
```diff
@@ -21,6 +22,9 @@ class Darkice < Formula
   depends_on "jack"
 
   def install
+    ["aacPlusEncoder.cpp", "FaacEncoder.cpp", "OpusLibEncoder.cpp", "VorbisLibEncoder.cpp"].each do |f|
```
ilovezfs: Has this been reported?

ilovezfs:
Does it really still need ENV.deparallelize for berkeley-db?

Formula/berkeley-db.rb
```diff
@@ -13,10 +13,10 @@ class BerkeleyDb < Formula
     sha256 "8f5a87bc3336e01f8528ae8aaae24c83e8fa10e8f01ee65e8cac4af7d0786bf1" => :mountain_lion
   end
 
-  option "with-java", "Compile with Java support."
-  option "with-sql", "Compile with SQL support."
+  option "without-sql", "Build without SQL support."
```
ilovezfs: if sql and dbm don't add dependencies, why are they optional as opposed to always passed?

Formula/berkeley-db.rb
```diff
@@ -13,10 +13,10 @@ class BerkeleyDb < Formula
...
+  option "without-dbm", "Build without dbm interface."
 
-  deprecated_option "enable-sql" => "with-sql"
+  depends_on :java => [:optional, :build]
```
ilovezfs: is it really only a build time dep?

Formula/darkice.rb
maxim-belkin: Thanks for asking. I was going to say "no", but then looked around and found [this PR](https://github.com/rafael2k/darkice/pull/120). So yes, this has been reported

ilovezfs: Cool. Please add the link, date, and brief description like these: https://gist.github.com/ilovezfs/62a9eb715dc044f7202cbf86caaf713b

maxim-belkin: within the `echo do` loop or right before it?

Formula/berkeley-db.rb
maxim-belkin: Do you mean we should convert "without-sql" to "with-sql" OR `depends_on "sql" => :build` ? I guess the latter but want to make sure

maxim-belkin:
I've taken all suggestions into account and now formula builds with `sql` and `dbm` by default:
```configure
--enable-sql
--enable-sql_codegen
--enable-dbm
```
I don't see any additional dependencies because of this change.

`java` is `build` only dependency
parallel build (still) fails.

[Options](http://docs.oracle.com/cd/E17076_04/html/installation/build_unix_conf.html) to consider:

`--enable-sql_compat`: builds sqlite3.

I see that `sqlite` formula is keg-only because of `sqlite3` and that's why I didn't enable this option in this PR. Let me know if I should add it here.

maxim-belkin:
added `--enable-stl`

maxim-belkin:
@ilovezfs I've addressed your comments/suggestions

ilovezfs (approved):

ilovezfs:
@maxim-belkin thanks for your contribution to Homebrew! I did some minor surgery on the PR so that it's one commit per formula and so that it uses the standard commit subjects, but preserved your commit credits for the changes to berkeleydb, darkice, and rpm.

maxim-belkin:
Cool! Thank you!

PR #7427, opened 2016-11-30
grib-api 1.19.0

jfsimoneau:

-----

grib-api 1.19.0

### DIFF
--- a/Formula/grib-api.rb
+++ b/Formula/grib-api.rb
@@ -1,10 +1,9 @@
 class GribApi < Formula
   desc "Encode and decode grib messages (editions 1 and 2)"
   homepage "https://software.ecmwf.int/wiki/display/GRIB/Home"
-  url "https://software.ecmwf.int/wiki/download/attachments/3473437/grib_api-1.18.0-Source.tar.gz"
-  mirror "https://distfiles.macports.org/grib_api/grib_api-1.18.0-Source.tar.gz"
-  sha256 "dfffeeb4df715b234907cb12d6729617bed0df0ff023337c2dd3cd20ab58199e"
-  revision 2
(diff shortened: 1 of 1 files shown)


woodruffw:
Holding off on merging this until the mirror reappears. Might be worth prodding upstream/the mirror owners in a few days if it doesn't happen on its own.

ilovezfs:
@woodruffw the mirror is macports so we probably should open a PR to upgrade their port file if we want the mirror updated: https://github.com/macports/macports-ports/blob/master/science/grib_api/Portfile

If we're concerned about mirrors and don't want to do the MacPorts PR, we could switch over to the Debian mirrors' tarballs. They're .xz instead of .tar.gz so the checksum won't match (we'd just use them instead of the upstream link). Alternatively, we could just stick with the upstream tarball and hope it doesn't 404. The last option would be to use --mirror when pulling this and host our own tarball.

woodruffw:
@ilovezfs I'll open a MacPorts PR, let's just go with Debian mirrors if they don't get around to it within a day or so.

woodruffw:
macports/macports-ports#71 has the changes.

ilovezfs:
:+1:

woodruffw:
There hasn't been much movement from the MacPorts team on this (I was told that my bump was merged, but hasn't appeared in the tree...), so let's go ahead and switch to Debian here @jfsimoneau

jfsimoneau:
I added the mirror link to Debian's package. When Takeshi Enomoto, the maintainer of grib-api on MacPorts, approves your pull request, I'll change it back to MacPorts so the checksum matches.

ilovezfs:
Actually @woodruffw meant as a full replacement since the checksums must match or the tarball will be rejected by brew. I've updated and squashed the PR accordingly, as well as fixed the audit error.

ilovezfs:
@jfsimoneau thanks for your contribution to Homebrew!

PR #7435, opened 2016-12-01
readline 7.0.1

JCount:

-----

This may also require revision bumps for other formulae.

### DIFF
--- a/Formula/readline.rb
+++ b/Formula/readline.rb
@@ -3,6 +3,7 @@ class Readline < Formula
   homepage "https://tiswww.case.edu/php/chet/readline/rltop.html"
   url "https://ftpmirror.gnu.org/readline/readline-7.0.tar.gz"
   mirror "https://ftp.gnu.org/gnu/readline/readline-7.0.tar.gz"
+  version "7.0.1"
   sha256 "750d437185286f40a369e1e4f4764eda932b9459b5ec9a731628393dd3d32334"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
The CI issue with postgresql turns out to be a `brew test` bug and a `brew test-bot` bug. While `brew test foo` requires `bar` to be linked if `foo` depends on `bar`, `brew test bar` will not require `bar` to be linked, which would only be appropriate if `bar` is `keg_only`. The postgresql error is what you get if you run `brew test postgresql` with postgresql unlinked. The reason it is unlinked is that it `conflicts_with` with postgres-xc, and so test-bot will unlink postgresql so that it can install and test postgres-xc. But it doesn't relink postgresql after the postgres-xc test is completed, and since `brew test` doesn't error out requiring postgresql be linked before running, the weird error output in the CI log occurs because postgresql is tested while unlinked.

CC @MikeMcQuaid

ilovezfs:
@JCount thanks for the :gift: Your contribution to Homebrew is much appreciated.

MikeMcQuaid:
@ilovezfs `brew test`s should be written such that they work unlinked.

ilovezfs:
@MikeMcQuaid then shouldn't we unlink before running the test?

MikeMcQuaid:
@ilovezfs Could do, yep.

ilovezfs:
Would be quite interesting to see how many fail!

JCount:
I see the Jenkins standing in a sea of 🔥

ilovezfs:
@JCount :see_no_evil:

PR #7462, opened 2016-12-02
subversion: simplify apr logic

DomT4:

-----

Follows what the rest of Homebrew has started doing since ditching the requirement.

### DIFF
--- a/Formula/subversion.rb
+++ b/Formula/subversion.rb
@@ -4,6 +4,7 @@ class Subversion < Formula
   url "https://www.apache.org/dyn/closer.cgi?path=subversion/subversion-1.9.5.tar.bz2"
   mirror "https://archive.apache.org/dist/subversion/subversion-1.9.5.tar.bz2"
   sha256 "8a4fc68aff1d18dcb4dd9e460648d24d9e98657fbed496c582929c6b3ce555e5"
+  revision 1
 
   bottle do
     sha256 "c94de8b8fd132839a839a7a42770a94b3fe99a59ac40c4073b3e2cbc2e8af000" => :sierra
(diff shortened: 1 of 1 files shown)


ilovezfs:
Do you think it should be revision bumped?

DomT4:
Eh. Mostly depends how much you want to prioritise local installation consistency really.

My point of view is that getting rid of the Apr requirement was premature, but I think the project might as well be consistent about not keeping around logic that was designed to facilitate using the system as widely as possible in formulae that used `:apr` previously. The confusion around this resulted in:

```ruby
depends_on :apr => :build
```
Being changed to:
```ruby
depends_on "apr" => :build
```
The other day, which is flat out incorrect whichever way you look at it. Homebrew's `apr` is never `:build` only, well, unless you static build I guess.

ilovezfs:
Let's go ahead and revision bump since the bottle contents and run-time dependencies will have changed.

Formula/subversion.rb
```diff
@@ -126,6 +116,9 @@ def install
       --enable-optimize
       --with-zlib=/usr
       --with-sqlite=#{Formula["sqlite"].opt_prefix}
+      --with-apr=#{Formula["apr"].opt_prefix}
+      --with-apr-util=#{Formula["apr-util"].opt_prefix}
+      --with-apxs=no
```
ilovezfs: what's this?

DomT4:
Done & re-pushed. If Jenkins had already started the build it may well crash the new build thanks to that delightful ancient bug.

ilovezfs:
@BrewTestBot test this please

Formula/subversion.rb
DomT4: Mmm. Don't quote me on this because I'm working from memory, but I believe `apxs` would use the system's `apxs` which links/references to the system's Apr, and you'd have overlapping Apr linkage.

DomT4: Yeah, we put it in a while back. Pretty sure someone gave me a detailed explanation as to why it would be needed at the time, but by now it has well & truly escaped my brain.

To be honest though I've never really had cause to question or look into it much because it has always seemed to work & nobody's gone _"Hey why doesn't xyz configuration handle xyz use case"_ etc.

ilovezfs:
Thanks, Mr. T

DomT4:
Thanks ILZ 😁

PR #7470, opened 2016-12-02
dub v1.1.1

John-Colvin:

-----

~~Doesn't pass `audit` because it doesn't like recommended dependencies for some reason~~. Can someone explain? I read about it https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#specifying-other-formulae-as-dependencies so I assumed it would be ok to use.

### DIFF
--- a/Formula/dub.rb
+++ b/Formula/dub.rb
@@ -1,11 +1,10 @@
 class Dub < Formula
   desc "Build tool for D projects"
   homepage "https://code.dlang.org/about"
-  url "https://github.com/dlang/dub.git",
-      :tag => "v1.0.0",
-      :revision => "b59af2b8befb4fad4157d8c9cc86dba707b2fc87"
+  url "https://github.com/dlang/dub/archive/v1.1.1.tar.gz"
+  sha256 "9931c03fa908e83632553572bdcfb3570b8175a57b00b9ff7c9b69b991662b89"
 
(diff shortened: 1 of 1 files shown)


Formula/dub.rb
```diff
@@ -14,13 +13,7 @@ class Dub < Formula
...
   end
 
-  devel do
-    url "https://github.com/dlang/dub.git",
-        :tag => "v1.1.0-beta.1",
-        :revision => "1824d16056a680fcca52d3ce87fb43d3679ff3ee"
-  end
-
-  depends_on "pkg-config" => :build
+  depends_on "pkg-config" => :recommended
```
MikeMcQuaid: `depends_on "pkg-config" => [:recommended, :build]` but I do wonder the benefits of making this optional?

MikeMcQuaid (changes requested):

Formula/dub.rb
John-Colvin: `dub` uses it at runtime if it can find it, but is also perfectly happy without. It's no longer used for building `dub` itself unless you're building it with an old compiler.

MikeMcQuaid (approved):

MikeMcQuaid:

For future reference the preferred commit message format is `dub 1.1.1` (no `v`). Thanks!

PR #7476, opened 2016-12-02
kops: 1.4.1 (fix) 

WillPlatnick:
```kops version``` uses an ENV variable that needs to be present at build to denote the version. Without this, people don't know the version they are on.

### DIFF
--- a/Formula/kops.rb
+++ b/Formula/kops.rb
@@ -16,6 +16,7 @@ class Kops < Formula
   depends_on "kubernetes-cli"
 
   def install
+    ENV["VERSION"] = version unless build.head?
     ENV["GOPATH"] = buildpath
     kopspath = buildpath/"src/k8s.io/kops"
     kopspath.install Dir["*"]
--- a/Formula/kops.rb
+++ b/Formula/kops.rb
@@ -4,6 +4,7 @@ class Kops < Formula
   url "https://github.com/kubernetes/kops/archive/v1.4.1.tar.gz"
   sha256 "69b3c9d7e214109cfd197031091ed23963383c894e92804306629f6a32ab324b"
   head "https://github.com/kubernetes/kops.git"
+  revision 1
 
   bottle do
     cellar :any_skip_relocation

### CODE REVIEW

File: Formula/kops.rb [hunk 1]
```diff
@@ -5,17 +5,11 @@ class Kops < Formula
   sha256 "69b3c9d7e214109cfd197031091ed23963383c894e92804306629f6a32ab324b"
   head "https://github.com/kubernetes/kops.git"
 
-  bottle do
-    cellar :any_skip_relocation
-    sha256 "99fc900bb11b242b4d3eca456fc7956233a8efa92f3dee7b321a005de0e94a28" => :sierra
-    sha256 "ed71d71b5031e0918478dec06b9064cf7e3f5b907128e98ec85a187801a27f8e" => :el_capitan
-    sha256 "0eee45caca5eb2a67ab88a90f9da226e99538d5198a196067b2a224a816bd6e0" => :yosemite
-  end
```
MikeMcQuaid: Please leave this as-is, it'll be updated when we pull.

Formula/kops.rb
```diff
@@ -5,17 +5,11 @@ class Kops < Formula
...
-
   depends_on "go" => :build
   depends_on "kubernetes-cli"
 
   def install
+    ENV["VERSION"] = "1.4.1" unless build.head?
```
MikeMcQuaid: `ENV["VERSION"] = version`.

What should the behaviour be if it's building HEAD?

MikeMcQuaid (changes requested):

Formula/kops.rb
WillPlatnick: Fixed.

Formula/kops.rb
WillPlatnick: The makefile will grab the git commit for HEAD and set the environment variable appropriately.

WillPlatnick:
@MikeMcQuaid Squashed and committed, I think that should fix everything raised.

WillPlatnick:
@MikeMcQuaid Alrighty, using the version variable now.

nijikon:

PR #7350, opened 2016-11-28
ttylog 0.29

ilovezfs:
Created with `brew bump-formula-pr`.

PR #7479, opened 2016-12-02
ttylog: migrate to boneyard

ilovezfs:
Created with `brew boneyard-formula-pr` because the latest version is incompatible with macOS and upstream does not intend to fix it: https://github.com/rocasa/ttylog/issues/6.


ilovezfs:
See Homebrew/homebrew-core#7350.

MikeMcQuaid:
👍

PR #7497, opened 2016-12-02
sync_gateway 1.3.1

ilovezfs:
use depot_tools vendoring

### DIFF
--- a/Formula/sync_gateway.rb
+++ b/Formula/sync_gateway.rb
@@ -2,9 +2,8 @@ class SyncGateway < Formula
   desc "Make Couchbase Server a replication endpoint for Couchbase Lite"
   homepage "http://docs.couchbase.com/sync-gateway"
   url "https://github.com/couchbase/sync_gateway.git",
-      :tag => "1.2.1",
-      :revision => "26c202a800226ce599cbaf9b2fcc4576a924d45e"
-
+      :tag => "1.3.1",
+      :revision => "660b1c92fadce1a9c7e692dfe7c5b741772d1dd2"
   head "https://github.com/couchbase/sync_gateway.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


woodruffw:
```
==> make buildit
./build.sh
./build.sh: line 20: cd: /src/github.com/couchbase/sync_gateway: No such file or directory
make: *** [buildit] Error 1
```

Formula/sync_gateway.rb
```diff
@@ -16,10 +15,41 @@ class SyncGateway < Formula
...
+  end
 
   def install
-    system "make", "buildit"
-    bin.install "bin/sync_gateway"
+    ENV["GOPATH"] = buildpath/"godeps"
+    repo_cache = HOMEBREW_CACHE/"repo_cache/#{name}/.repo"
+    repo_cache.mkpath
+
+    inreplace "manifest/default.xml", "git://", "https://" unless build.head?
```
MikeMcQuaid: Can you explain this a bit? Probably warrants a comment regardless.

Formula/sync_gateway.rb
```diff
@@ -16,10 +15,41 @@ class SyncGateway < Formula
...
+    url "https://chromium.googlesource.com/chromium/tools/depot_tools.git",
+      :revision => "935b93fb9bf367510eece7db8ee3e383b101c36d"
+  end
 
   def install
-    system "make", "buildit"
-    bin.install "bin/sync_gateway"
+    ENV["GOPATH"] = buildpath/"godeps"
+    repo_cache = HOMEBREW_CACHE/"repo_cache/#{name}/.repo"
+    repo_cache.mkpath
```
MikeMcQuaid: Is this a new pattern?

ilovezfs: Yeah, sync_gateway, like v8, uses https://chromium.googlesource.com/chromium/tools/depot_tools.git. So this is to support caching with that vendor system without having to resort to `go_resource` blocks.

Formula/sync_gateway.rb
ilovezfs: Upstream has already "upgraded" the URLs for their go dependencies to use https in HEAD, but as of the 1.3.1 stable tag, they were all still `git://`, so this makes sure we use https with stable, too. It can be removed for the next release.

PR #7503, opened 2016-12-03
gradle: Add option to install gradle-all

octylFractal:
- [ ] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)? ❌ **I cannot run `RuboCop` as it currently segfaults on my system.**

-----

When writing Gradle plugins it's nice to have the source on hand. This PR adds the ability to install the `-all` distribution of Gradle, which, while larger, contains docs, source, and other useful files for a developer. I'm not sure if there is a recommended way to do such a thing, so I patched something together from what I could find on the documentation. If there's a better way, just tell me and I'd be glad to do it.

### DIFF
--- a/Formula/gradle.rb
+++ b/Formula/gradle.rb
@@ -1,8 +1,10 @@
 class Gradle < Formula
   desc "Build system based on the Groovy language"
   homepage "https://www.gradle.org/"
-  url "https://downloads.gradle.org/distributions/gradle-3.2.1-bin.zip"
-  sha256 "9843a3654d3e57dce54db06d05f18b664b95c22bf90c6becccb61fc63ce60689"
+  url "https://downloads.gradle.org/distributions/gradle-3.2.1-all.zip"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> while larger

What's the size difference?

If this has a different URL then it should be a different formula, probably. Alternatively, if it's not a huge difference we can just always use `-all` and selectively install the bits we need.

kenzierocks:
The `-bin` zip is 66M and the `-all` zip is 85M. My reasoning for keeping the formula together is that they should both have the same version, and I'm worried that one of them might get behind a few versions if no one uses it.

MikeMcQuaid:
@kenzierocks Let's change this one to use the `-all` zip and only install the same bits as currently by default unless the option is passed.

kenzierocks:
Alright, made that change.

MikeMcQuaid:

PR #7535, opened 2016-12-04
Node@*: import from homebrew/versions.

MikeMcQuaid:

-----

Homebrew/versions PR: https://github.com/Homebrew/homebrew-versions/pull/1471

If any of these don't build they will likely be 💥. If we are going to not import any then https://nodejs.org/en/download/releases/ would imply dropping `node@5` (but analytics show it's widely used).

### DIFF
--- a/Formula/node.rb
+++ b/Formula/node.rb
@@ -1,5 +1,5 @@
 class Node < Formula
-  desc "Platform built on the V8 JavaScript runtime to build network applications"
+  desc "Platform built on V8 to build network applications"
   homepage "https://nodejs.org/"
   url "https://nodejs.org/dist/v7.2.0/node-v7.2.0.tar.xz"
   sha256 "486d4db7ef659521ad2fafefca877638da07bef61e2aee090207ff52149294fb"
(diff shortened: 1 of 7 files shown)


dunn:
Weird
```
==> Downloading https://nodejs.org/dist/v0.12.17/node-v0.12.17.tar.xz
/usr/bin/curl --fail --remote-time --location --user-agent Homebrew/1.1.2-17-ga6a09d6 (Macintosh; Intel macOS 10.12.1) curl/7.49.1 https://nodejs.org/dist/v0.12.17/node-v0.12.17.tar.xz -C 0 -o /Users/brew/Library/Caches/Homebrew/node@0.12-0.12.17.tar.xz.incomplete --silent
Warning: Formula reports different SHA256: 4ce3a862eb28be752fbd65fe032c1d55cbbc1145af39292766eea701f67ba5f6
Downloaded to: /Users/brew/Library/Caches/Homebrew/node@0.12-0.12.17.tar.xz
SHA256: e2e82229200ef973e033d6dae910d789ab440e913f8c4c6225bf5e6e4a446420

DomT4:
Personally I'd just keep the LTS & last couple stable version before the current one and point people to `nvm` or such if they want anything older.

Alternatively all the versioned Nodes could use `Formula["node"].post_install` in their `post_install` block which reduces the sheer amount of logic carried around and make this less of a pain for people to look after going forwards.

MikeMcQuaid:
> Weird

@dunn Yeh, forgot to update a checksum when doing a bump.

> Personally I'd just keep the LTS & last couple stable version before the current one and point people to nvm or such if they want anything older.

So long as they build, are used and we're not getting support issues: I'd rather we packaged them. In general I'm going to migrate everything over from homebrew/versions that's still actively used and builds.

> Alternatively all the versioned Nodes could use Formula["node"].post_install in their post_install block which reduces the sheer amount of logic carried around and make this less of a pain for people to look after going forwards.

I'd rather not start down the hole of having formulae depend on others in that way. If it becomes a maintenance burden we can fix it then rather than preemptively.

DomT4:
> I'd rather not start down the hole of having formulae depend on others in that way.

It's less starting and more going back to it. We used to lean on that mechanism for `gnutls`'s `postinstall` step for example, until there was a brief change in the `openssl` `postinstall` step, at which point I removed it and added the whole logic block.

I only suggested it because we tweaked `node` quite a bit and forgot to swiftly apply those changes to the other `node` formulae in Versions at the time, but you run the show around here and I left, so is all your's to play with.

PR #7540, opened 2016-12-04
ucspi-tcp: vendor IPv6 patch

zmwangx:

-----

fefe.de disappeared from the Internet. The current patch was downloaded from the Wayback Machine [1], then unzipped.

Please merge https://github.com/Homebrew/formula-patches/pull/94.

P.S. I caught this when I was validating the checksums of all patches in core.

[1] https://web.archive.org/web/20160403064123/http://www.fefe.de/ucspi/ucspi-tcp-0.88-ipv6.diff19.bz2

### DIFF
--- a/Formula/ucspi-tcp.rb
+++ b/Formula/ucspi-tcp.rb
@@ -3,6 +3,7 @@ class UcspiTcp < Formula
   homepage "https://cr.yp.to/ucspi-tcp.html"
   url "https://cr.yp.to/ucspi-tcp/ucspi-tcp-0.88.tar.gz"
   sha256 "4a0615cab74886f5b4f7e8fd32933a07b955536a3476d74ea087a3ea66a23e9c"
+  revision 1
 
   bottle do
     cellar :any_skip_relocation
--- a/Formula/ucspi-tcp.rb
+++ b/Formula/ucspi-tcp.rb
@@ -12,9 +12,11 @@ class UcspiTcp < Formula
     sha256 "67eb31db588a2299c5e69a4a60f3c56d07624a58e52e77cff2e58be554085d9f" => :mavericks
   end
 
+  # IPv6 patch
+  # Used to be https://www.fefe.de/ucspi/ucspi-tcp-0.88-ipv6.diff19.bz2
   patch do
-    url "https://www.fefe.de/ucspi/ucspi-tcp-0.88-ipv6.diff19.bz2"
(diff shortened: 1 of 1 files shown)


zmwangx:
For the record, the patch has been in the formula since it was added in 2013. See af367447d4f.

nijikon:
This needs a revision bump.

zmwangx:
The patch is exactly the same. A revision bump is pointless.

nijikon:
You are right, I missed that you only changed the url.

nijikon:

nijikon:
Closed in cd699a75af4e6725779af0bcae72524d0169cd61

zmwangx:
@nijikon Actually, https://github.com/Homebrew/formula-patches/pull/94 hasn't been merged, and after merging it, the commit 2b3e4da needs to be changed to the merged commit in formula-patches' master branch...

ilovezfs:
@zmwangx Mike taught me a scary trick recently ... the URL actually works with `formula-patches` in it before it's ever committed.

zmwangx:
@ilovezfs Yeah I know it works, but it feels wrong... Homebrew/homebrew-core master should carry patches from Homebrew/formula-patches master, IMO.

ilovezfs:
Yeah, I wish it didn't work, but in terms of the PRs we might as well reap the rewards of the exploit and use the final url as soon as it resolves.

ilovezfs:
@zmwangx I was lazy and used rebase merge, so here we go https://raw.githubusercontent.com/Homebrew/formula-patches/7ea2f2e/ucspi-tcp/patch-0.88-ipv6.diff

zmwangx:
> Yeah, I wish it didn't work, but in terms of the PRs we might as well reap the rewards of the exploit and use the final url as soon as it resolves.

I think the commit could be swapped out when pulling to avoid another CI build. If `brew pull` is smart enough you might even be able do that automagically...

> I was lazy and used rebase merge, so here we go

Might directly pushing to master? Not really worth another PR...

By the way, putting vendored patches right in this repo would stop all this madness. (Yeah, I do remember the out-of-tree problem and debate on DSL.) macports/macports-ports is only 231M in size (repo + checked out tree), and their number of packages (10304) almost triples ours, so it shouldn't be scary.

MikeMcQuaid:
> By the way, putting vendored patches right in this repo would stop all this madness.

There is no madness. We have a URL that works and is cached and checksummed for those who need it.

PR #7542, opened 2016-12-05
argyll-cms 1.9.2

jslabovitz:
While building argyll-cms from source, I noticed that brew attempted to download a 1.9.0 tarball. However, that doesn't exist -- it's 1.9.2 now. This patch gets the right file.


nijikon:

PR #7556, opened 2016-12-05
go: Remove broken CGO_ENABLED options

msiebuhr:

-----

It produces builds that behave in non-standard ways (i.e. produces
non-obvious errors when cross-compiling).

See https://github.com/golang/go/issues/18189#issuecomment-issuecomment-264745541

### DIFF
--- a/Formula/go.rb
+++ b/Formula/go.rb
@@ -1,6 +1,7 @@
 class Go < Formula
   desc "The Go programming language"
   homepage "https://golang.org"
+  revision 1
 
   stable do
     url "https://storage.googleapis.com/golang/go1.7.4.src.tar.gz"
--- a/Formula/go.rb
+++ b/Formula/go.rb
@@ -63,7 +63,7 @@ class Go < Formula
     cd "src" do
       ENV["GOROOT_FINAL"] = libexec
       ENV["GOOS"]         = "darwin"
-      ENV["CGO_ENABLED"]  = build.with?("cgo") ? "1" : "0"
+      ENV["CGO_ENABLED"]  = "0" if build.without?("cgo")
       system "./make.bash", "--no-clean"
     end
 
@@ -74,7 +74,7 @@ class Go < Formula
 
     # Race detector only supported on amd64 platforms.
     # https://golang.org/doc/articles/race_detector.html
-    if build.with?("cgo") && build.with?("race") && MacOS.prefer_64_bit?
+    if buid.with?("cgo") && build.with?("race") && MacOS.prefer_64_bit?
       system bin/"go", "install", "-race", "std"
     end
 
@@ -120,6 +120,11 @@ class Go < Formula
     if build.with? "godoc"
       assert File.exist?(libexec/"bin/godoc")
       assert File.executable?(libexec/"bin/godoc")
+    end
+
+    if build.with? "cgo"
+      ENV["GOOS"] = "freebsd"
+      system bin/"go", "build", "hello.go"
     end
   end
 end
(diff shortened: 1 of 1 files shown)


ilovezfs:
This is an option for a reason, and I don't think we're building it any non-standard fashion, so I'm closing this unless upstream itself removes the option or says we should.

MikeMcQuaid:
> This is an option for a reason

It was an option originally added because it didn't work with Clang and defaulted to off. It was changed to default-on in September 2013 so I think it's worth considering removing it now.

ilovezfs:
I seem to recall cases where it was required for other things to build. I don't think it makes sense to remove this unless upstream is advising as such.

MikeMcQuaid:
@ilovezfs Worth investigating a bit more either way, I think. Death to options 😉

MikeMcQuaid:
@msiebuhr Could you add something to the `test do` block that uses CGO functionality so we can verify that that's still working after this change? Thanks!

msiebuhr:
Quoting the linked bug:

> >CGO_ENABLED=0
>
> You don't need to set this value, it always defaults to the correct value. I think what is going on is brew are setting it, and that is messed you up. I think this bug should be assigned to the maintainer of the homebrew formular for Go.

I've added a test that cross-builds to freebsd, though I haven't tried running it with a broken install of Go.

Formula/go.rb
```diff
@@ -121,5 +119,10 @@ def caveats; <<-EOS.undent
       assert File.exist?(libexec/"bin/godoc")
       assert File.executable?(libexec/"bin/godoc")
     end
+    def freebsd_build
+      ENV["GOOS"]="freebsd"
+      system bin/"go", "build", "-o", "/dev/null", "hello.go"
+    end
+    freebsd_build
```
MikeMcQuaid: Instead of creating a function do:

```ruby
if build.with? "cgo"
  ENV["GOOS"] = "freebsd"
  system bin/"go", "build", "hello.go"
end
```

Formula/go.rb
```diff
@@ -40,7 +40,6 @@ class Go < Formula
     end
   end
 
-  option "without-cgo", "Build without cgo (also disables race detector)"
```
MikeMcQuaid: This this option here.

Formula/go.rb
```diff
@@ -63,7 +62,6 @@ def install
     cd "src" do
       ENV["GOROOT_FINAL"] = libexec
       ENV["GOOS"]         = "darwin"
-      ENV["CGO_ENABLED"]  = build.with?("cgo") ? "1" : "0"
```
MikeMcQuaid: Let's leave the option but not set it when not needed:

```ruby
ENV["CGO_ENABLED"] = "0" if build.without?("cgo")
```

Formula/go.rb
```diff
@@ -74,7 +72,7 @@ def install
 
     # Race detector only supported on amd64 platforms.
     # https://golang.org/doc/articles/race_detector.html
-    if build.with?("cgo") && build.with?("race") && MacOS.prefer_64_bit?
```
MikeMcQuaid: Leave `if build.with?("cgo")` here.

MikeMcQuaid (dismissed):

msiebuhr:
Thanks. Fixed per the given suggestions.

tsuna:
Thanks for doing this, I was confused when I started testing go1.8beta2 and initially reported this upstream at golang/go#18360, where we quickly concluded that the formula was working by chance and that `CGO_ENABLED` shouldn't be set while building Go unless explicitly disabled with `--without-cgo`.

ilovezfs (approved):

ilovezfs:
@msiebuhr thanks for the PR! :rocket:

PR #7557, opened 2016-12-05
LLVM@3*: import from homebrew/versions

MikeMcQuaid:

-----

Homebrew/versions PR: https://github.com/Homebrew/homebrew-versions/pull/1475

Also, remove various unused options from analytics.

### DIFF
--- a/Formula/llvm@3.4.rb
+++ b/Formula/llvm@3.4.rb
@@ -1,0 +1,248 @@
+class LlvmAT34 < Formula
+  desc "Next-gen compiler infrastructure"
+  homepage "http://llvm.org/"
+
+  stable do
+    url "http://llvm.org/releases/3.4.2/llvm-3.4.2.src.tar.gz"
+    sha256 "17038d47069ad0700c063caed76f0c7259628b0e79651ce2b540d506f2f1efd7"
+
+    resource "clang" do
+      url "http://llvm.org/releases/3.4.2/cfe-3.4.2.src.tar.gz"
+      sha256 "5ba6f5772f8d00f445209356a7daf83c5bca2da5acd10de517ad2359ae95bc10"
+    end
+
+    resource "clang-tools-extra" do
+      url "http://llvm.org/releases/3.4/clang-tools-extra-3.4.src.tar.gz"
+      sha256 "ba85187551ae97fe1c8ab569903beae5ff0900e21233e5eb5389f6ceab1028b4"
+    end
+
+    resource "polly" do
+      url "http://llvm.org/releases/3.4/polly-3.4.src.tar.gz"
+      sha256 "9f1a5fb73dddc0afe47a0f4108dea818e0d1d16485899141957f87f75fa50ee7"
+    end
+
+    resource "libcxx" do
+      url "http://llvm.org/releases/3.4.2/libcxx-3.4.2.src.tar.gz"
+      sha256 "826543ee2feb5d3313b0705145255ebb2ed8d52eace878279c2525ccde6e727c"
+    end
+
+    if MacOS.version <= :snow_leopard
+      resource "libcxxabi" do
+        url "http://llvm.org/git/libcxxabi.git", :branch => "release_32"
+      end
+    end
+  end
+
+  head do
+    url "http://llvm.org/git/llvm.git", :branch => "release_34"
+
+    resource "clang" do
+      url "http://llvm.org/git/clang.git", :branch => "release_34"
+    end
+
+    resource "clang-tools-extra" do
+      url "http://llvm.org/git/clang-tools-extra.git", :branch => "release_34"
+    end
+
+    resource "polly" do
+      url "http://llvm.org/git/polly.git", :branch => "release_34"
+    end
+
+    resource "libcxx" do
+      url "http://llvm.org/git/libcxx.git", :branch => "release_34"
+    end
+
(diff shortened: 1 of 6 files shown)


MikeMcQuaid:
@BrewTestBot test this please

equal-l2:
llvm@3.4 and llvm@3.5 should depend on isl newer than 0.12 because they need `isl/val.h`, which is introduced at isl 0.12

MikeMcQuaid:
Thanks @equal-l2. The issue is cloog@0.18 is depending on isl@0.11 so that'll need handled too.

equal-l2:
@MikeMcQuaid 
The version difference between `cloog` and `cloog@0.18` is on the patch version.
Maybe are they interchangeable?

MikeMcQuaid:
> The version difference between cloog(0.18.4) and cloog@0.18(0.18.0) is on the patch version.
> Maybe are they interchangeable?

@equal-l2 Good idea, will try that out. Thanks a lot for the help!

MikeMcQuaid:
@equal-l2 We have a tricky problem here due to the new version handling. cloog(0.18.4) and cloog@0.18(0.18.0) use different ISLs and GMPs. We'll need them to use the same as the GCC that uses them. Thoughts on how to resolve this? CC @ilovezfs for an example of something we could easily shoot ourselves in the foot on without https://github.com/Homebrew/brew/pull/1681.

ilovezfs:
>We'll need them to use the same as the GCC that uses them.

Is that just an aesthetic preference or something that we know actually compromises the functionality of the compiler?

MikeMcQuaid:
> Is that just an aesthetic preference or something that we know actually compromises the functionality of the compiler?

Partly it's to avoid descending into madness of having a single piece of software depend on two versions of the same thing. It may or may not compromise the functionality of the compiler but actually detecting that may prove difficult.

ilovezfs:
>Partly it's to avoid descending into madness of having a single piece of software depend on two versions of the same thing.

That's circular reasoning if it's supposed to be a demonstration of "something we could easily shoot ourselves in the foot" with.

MikeMcQuaid:
@ilovezfs It would be if I didn't make the second argument: it could break things in ways that would be very subtle (i.e. at runtime). We saw similarly with mixed C++ standard libraries: things fail infrequently and extremely weirdly.

ilovezfs:
>It would be if I didn't make the second argument: it could break things in ways that would be very subtle (i.e. at runtime).

That sounds plausible though if the given releases are ABI compatible it's not clear to me what form that breakage would take.

>We saw similarly with mixed C++ standard libraries: things fail infrequently and extremely weirdly.

The C++ situation is actually still very broken despite the C++ standards warnings because we have no way to require the use of the same compiler for dependencies and reverse dependencies, and the mixed and matched compilers are rarely actually fully ABI compatible. I recently had an upstream quite unhappy about how we don't enforce the use of the same compiler because it's causing our users to file bogus support issues with them.

equal-l2:
Is it impossible to replace all isl@0.11 dependency with isl 0.12?

At least, cloog@0.18 builds on isl 0.12.

MikeMcQuaid:
I'll try, thanks @equal-l2.

equal-l2:
Well, it found that LLVM 3.5 needs isl 0.13 or newer, unfortunately.

```
Analysis/Dependences.cpp:237:20: error: use of undeclared identifier 'isl_ctx_get_max_operations'
  long MaxOpsOld = isl_ctx_get_max_operations(S.getIslCtx());
                   ^
Analysis/Dependences.cpp:238:3: error: use of undeclared identifier 'isl_ctx_set_max_operations'
  isl_ctx_set_max_operations(S.getIslCtx(), OptComputeOut);
  ^
Analysis/Dependences.cpp:292:44: error: use of undeclared identifier 'isl_error_quota'
  if (isl_ctx_last_error(S.getIslCtx()) == isl_error_quota) {
                                           ^
Analysis/Dependences.cpp:300:3: error: use of undeclared identifier 'isl_ctx_reset_operations'; did you mean 'isl_ctx_reset_error'?
  isl_ctx_reset_operations(S.getIslCtx());
  ^~~~~~~~~~~~~~~~~~~~~~~~
  isl_ctx_reset_error
/usr/local/opt/isl@0.12/include/isl/ctx.h:236:6: note: 'isl_ctx_reset_error' declared here
void isl_ctx_reset_error(isl_ctx *ctx);
     ^
Analysis/Dependences.cpp:301:3: error: use of undeclared identifier 'isl_ctx_set_max_operations'
  isl_ctx_set_max_operations(S.getIslCtx(), MaxOpsOld);
  ^
5 errors generated.
```

`isl_ctx_get_max_operations` and etc. are introduced on isl 0.13.

Maybe isl 0.14 will work.

equal-l2:
Now I feel like it is reasonable to drop LLVM 3.4 & 3.5 support...

MikeMcQuaid:
Pulled just 3.7 and 3.8 for now.

tmcdonell:
llvm@3.8 and llvm@3.7 are now hard-coded at `--enable-targets=host`, and the option `--with-all-targets` is gone. Was there a reason for this, or is that a regression?

MikeMcQuaid:
@tmcdonell We are cutting down on options because they aren't tested by CI. We may be open to changing the defaults instead.

PR #7599, opened 2016-12-06
link-grammar 5.3.13

nijikon:

-----

### DIFF
--- a/Formula/link-grammar.rb
+++ b/Formula/link-grammar.rb
@@ -1,8 +1,8 @@
 class LinkGrammar < Formula
   desc "Carnegie Mellon University's link grammar parser"
   homepage "http://www.abisource.com/projects/link-grammar/"
-  url "http://www.abisource.com/downloads/link-grammar/5.3.3/link-grammar-5.3.3.tar.gz"
-  sha256 "832e1aa61d5eafbb7d3f4429847125067dcb6a3ece8c55ee8ffb1cd8c8db57d4"
+  url "http://www.abisource.com/downloads/link-grammar/5.3.13/link-grammar-5.3.13.tar.gz"
+  sha256 "a0f1665e046d6e40dfab305d3eab728ede2c509dd798c12154c52590bd81e3ca"
 
   bottle do
     sha256 "f6e69a200b473a60a6bafe8abe02862b9e114f10a4277badf63b03f2a818fbab" => :sierra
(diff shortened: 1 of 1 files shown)


dunn:
Urgh:
```
link-grammar:
      /usr/local/Cellar/link-grammar/5.3.13/lib/python2.7/site-packages/linkgrammar/_clinkgrammar.5.so
      /usr/local/Cellar/link-grammar/5.3.13/lib/python2.7/site-packages/linkgrammar/_clinkgrammar.so

**Comment by nijikon** [comment]
What we can do here? /cc @tdsmith

**Comment by tdsmith** [comment]
Fix the makefile!

Possible we could handle these problems automatically by stripping the linkages using an approach like https://github.com/alexzielenski/optool; maybe you could convince @woodruffw to implement it ;)

**Comment by woodruffw** [comment]
Ruby-macho can do this!

The current check is just this (in `check_python_framework_links`):

```ruby
python_modules = Pathname.glob lib/"python*/site-packages/**/*.so"
framework_links = python_modules.select do |obj|
  dlls = obj.dynamically_linked_libraries
  dlls.any? { |dll| /Python\.framework/.match dll }
end
```

Replacing it with this should work:

```ruby
python_modules = Pathname.glob lib/"python*/site-packages/**/*.so"
python_modules.each do |obj|
  macho = MachO::MachOFile.new(obj.to_s)
  macho.dylib_load_commands.each do |dylib|
    if  /Python\.framework/.match dylib.name.to_s 
      macho.delete_command(dylib)
    end
  end
end
```

(Not sure if that should remain in checks, since it modifies the binary.)

That's a bit longer than it needs to be, I can add abstract all that away in a `MachO::Tools` method if preferred.

tdsmith:
Hey, neat! This could save a lot of trouble if it works (at the cost of letting upstreams get away with bad habits...).

Agree this wouldn't be a check; the cleaner should do if we want it to happen magically. (Implementation note: we should only do this to things with realpaths in site-packages. Some packages like lldb build a single all-singing, all-dancing dylib that is both an importable Python module and, separately, a library that includes functionality that embeds a Python interpreter; the latter functionality will break without this linkage. If a dylib in lib/ is just symlinked into site-packages, we should probably assume something like that is going on.)

woodruffw:
> at the cost of letting upstreams get away with bad habits...

Yeah, this is the dilemma. Perhaps we should perform the linkage removals *and* spit out a warning that we had to?

> If a dylib in lib/ is just symlinked into site-packages, we should probably assume something like that is going on.)

So something like `next if File.symlink?(obj)`? Just want to make sure I'm understanding correctly :smile:

tdsmith:
I think so and yes!

woodruffw:
Got it, I'll open a brew PR in a bit!

woodruffw:
Opened Homebrew/brew#1769 for this.

tdsmith:
@woodruffw and I looked further and it's not that easy; we'd have to do additional rewriting of the bundles to indicate that the symbols should be found at runtime instead of from a linked library.

So we're back to "fix the makefile!"

nijikon:
@tdsmith is it really so crucial to check this linking? What consequences does it bring if we ship it like this?

ilovezfs:
```
    ENV["PYTHON_LIBS"] = "-undefined dynamic_lookup"
    inreplace "bindings/python/Makefile.am",
      "$(PYTHON2_LDFLAGS) -module -no-undefined",
      "$(PYTHON2_LDFLAGS) -module"
```

PR #7642, opened 2016-12-07
gcc: Hack version_suffix to fix --HEAD builds

ts826848:
This shouldn't be merged as-is; I just want something up so I'm not making changes blindly.

This addresses the issue mentioned by @ilovezfs [here](https://github.com/Homebrew/homebrew-core/pull/4783#issuecomment-265492644). I thought `version_suffix` would be called once, during the install of gcc, so a simple test for a `--HEAD` build would be enough to determine if the version number needed to be changed.

It appears that this broke other --HEAD builds that used gcc, since brew ended up looking for `gcc-7`/`g++-7`, failed, and fell back to gcc6, which may not be installed.

The solution here is to separate the logic for determining the version suffix into two parts -- one to be used for a gcc install, and the other to be used by other formulae.

@tdsmith suggested in IRC that build object being passed around could be altered. I could look into that if the hack here isn't acceptable.

If this is alright, don't merge this right away since I want to get a proper commit message written.

### DIFF
--- a/Formula/gcc.rb
+++ b/Formula/gcc.rb
@@ -73,6 +73,10 @@ class Gcc < Formula
   end
 
   def version_suffix
+    version.to_s.slice(/\d/)
+  end
+
+  def build_version_suffix
     if build.head?
       (stable.version.to_s.slice(/\d/).to_i + 1).to_s
     else
@@ -114,10 +118,10 @@ class Gcc < Formula
     args = [
       "--build=#{arch}-apple-darwin#{osmajor}",
       "--prefix=#{prefix}",
-      "--libdir=#{lib}/gcc/#{version_suffix}",
+      "--libdir=#{lib}/gcc/#{build_version_suffix}",
       "--enable-languages=#{languages.join(",")}",
       # Make most executables versioned to avoid conflicts.
-      "--program-suffix=-#{version_suffix}",
+      "--program-suffix=-#{build_version_suffix}",
       "--with-gmp=#{Formula["gmp"].opt_prefix}",
       "--with-mpfr=#{Formula["mpfr"].opt_prefix}",
       "--with-mpc=#{Formula["libmpc"].opt_prefix}",
@@ -160,7 +164,7 @@ class Gcc < Formula
 
     # Ensure correct install names when linking against libgcc_s;
     # see discussion in https://github.com/Homebrew/homebrew/pull/34303
-    inreplace "libgcc/config/t-slibgcc-darwin", "@shlib_slibdir@", "#{HOMEBREW_PREFIX}/lib/gcc/#{version_suffix}"
+    inreplace "libgcc/config/t-slibgcc-darwin", "@shlib_slibdir@", "#{HOMEBREW_PREFIX}/lib/gcc/#{build_version_suffix}"
 
     mkdir "build" do
       unless MacOS::CLT.installed?
@@ -175,7 +179,7 @@ class Gcc < Formula
       system "make", "install"
 
       if build.with?("fortran") || build.with?("all-languages")
-        bin.install_symlink bin/"gfortran-#{version_suffix}" => "gfortran"
+        bin.install_symlink bin/"gfortran-#{build_version_suffix}" => "gfortran"
       end
     end
 
@@ -183,7 +187,7 @@ class Gcc < Formula
     # with system compilers.
     # Since GCC 4.8 libffi stuff are no longer shipped.
     # Rename man7.
-    Dir.glob(man7/"*.7") { |file| add_suffix file, version_suffix }
+    Dir.glob(man7/"*.7") { |file| add_suffix file, build_version_suffix }
     # Even when suffixes are appended, the info pages conflict when
     # install-info is run. TODO fix this.
     info.rmtree
@@ -216,7 +220,7 @@ class Gcc < Formula
         return 0;
       }
     EOS
-    system "#{bin}/gcc-#{version_suffix}", "-o", "hello-c", "hello-c.c"
+    system "#{bin}/gcc-#{build_version_suffix}", "-o", "hello-c", "hello-c.c"
     assert_equal "Hello, world!\n", `./hello-c`
 
     (testpath/"hello-cc.cc").write <<-EOS.undent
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> The solution here is to separate the logic for determining the version suffix into two parts -- one to be used for a gcc install, and the other to be used by other formulae.

@ilovezfs Which other formulae use this logic? Is it just me or does it not feel weird using other formulae like this?

ilovezfs:
@MikeMcQuaid it's not just you. I don't think there should be any special version logic in the gcc formula. HEAD, devel, and stable specs of gcc are all supposed to satisfy the requirement `depends_on "gcc"` regardless of whether the dependent is itself stable, devel, or head. If there are known failure cases we want to handle with a fails_with, that logic belongs in brew not the formula. If upstream reports HEAD is gcc 6, that's up to them.

MikeMcQuaid:
@ilovezfs Does this change work for you? I'm not sure I understand the original bug.

ilovezfs:
The original bug is

```
brew install gcc
brew install --HEAD lightstringgraph
```

fails with the error

```
Error: No available formula with the name "gcc6" 
```

I think the underlying brew bug(s) are unfixed but this PR successfully dances around the breakage above if that's your question.

MikeMcQuaid:
> think the underlying brew bug(s) are unfixed

Yeh, I think it'd be preferable to fix this in both places if you're up to that @aw1621107?

ilovezfs:
Also I've now tested what happens when your gcc install is itself a HEAD installation with this PR in place before building, and :boom:
```
iMac-TMP:homebrew-core joe$ brew uninstall lightstringgraph
Error: No such keg: /usr/local/Cellar/lightstringgraph
iMac-TMP:homebrew-core joe$ brew install --HEAD lightstringgraph 
==> Installing lightstringgraph from homebrew/science
==> Using the sandbox
Error: No available formula with the name "gcc7" 
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

aw1621107:
>I don't think there should be any special version logic in the gcc formula.

Agreed. This was just the first thing I came up with.

>Yeh, I think it'd be preferable to fix this in both places if you're up to that @aw1621107?

Sure, I can take a look. Probably will take me a bit.

>Also I've now tested what happens when your gcc install is itself a HEAD installation with this PR in place before building, and 💥

Uh... I guess this might be "fixable" by having version_suffix return `installed_version.to_s.slice(/\d/)`, but that's not really any better.

Should I open a new PR for a better fix, or just force push and replace this one?

MikeMcQuaid:
> Should I open a new PR for a better fix, or just force push and replace this one?

Force push here, thanks!

nijikon:
@aw1621107 ping

aw1621107:
@nijikon Sorry, been pretty busy of late. I'll probably have some time this coming week to take a look.

MikeMcQuaid:
Interested in @fxcoudert's thoughts here.

fxcoudert:
I never really understood why Homebrew's GCC head is installed as `gcc-7` (currently) instead of `gcc-head` (which seems clearer that it's not a released version).

Regarding the rest of the PR, I do not understand well enough how Brew itself goes looking for GCC versions.

MikeMcQuaid:
> instead of gcc-head (which seems clearer that it's not a released version)

This makes sense to me. @ilovezfs, @aw1621107?

ilovezfs:
`fails_with :gcc => "7"` seems like it should be possible before stable is 7.

aw1621107:
> instead of gcc-head (which seems clearer that it's not a released version).

It looks like that was what I actually had at first when I opened #4783. I could have sworn that I was told to not do that which is why it was changed to `gcc-7`, but I can't seem to find any record of that either on GitHub or in my irssi logs.

Should that change be reverted while this gets sorted out?

aw1621107:
Never mind, reason for not using `gcc-HEAD` is [here](https://github.com/Homebrew/brew/pull/955#issuecomment-246927445)

MikeMcQuaid:
I agree with past me with my blacklist reasoning. Sorry to contradict myself. Mainly asking @fxcoudert for thoughts due to being a GCC maintainer. Still interested and please contradict me here; I defer to your GCC experience here.

fxcoudert:
GCC head consider itself GCC 7:

```
$ cat gcc-trunk/gcc/BASE-VER 
7.0.1
```

and the reasoning to avoid `gcc-HEAD` seems rational. Now, I do not fully understand what is the problem on the homebrew deps side, so I will leave that to you :)

Formula/gcc.rb
```diff
@@ -73,6 +73,10 @@ def osmajor
   end
 
   def version_suffix
```
MikeMcQuaid: This method seems to no longer be used?

MikeMcQuaid:
@aw1621107 Let me know if you have other questions that need addressed before updating this, thanks.

bfontaine:
Anything new here?

aw1621107:
School just finished up, so I should be able to work on this soon™

ilovezfs:
Chill out bot

fxcoudert:
My opinion: the meaning of GCC “head” varies a lot over time. To me, it makes no sense to refer to it as `stable+1`. Right now, head's features and bugs might be so different from what will become GCC 8 that it makes little sense for them to be grouped under the same label.

The HEAD of a compiler is very unstable, and thus depending on it really is at the user's own peril. If using it breaks other stuff, then switch back to something stable.

So, I think what makes most sense is `gcc-HEAD`.

---

The original reason for moving _away_ from `gcc-HEAD` (as I understand from @ilovezfs & @MikeMcQuaid's comments) is ([source](https://github.com/Homebrew/brew/pull/955#issuecomment-247020620)):

> You don't actually want to blacklist HEAD, you want to blacklist the version it's based on (as if that version gets e.g. released as a stable one you probably want to blacklist that too).

But I think that's simply not true, because for most of its life cycle HEAD is pretty different from the future released version.

ilovezfs:
Thanks for the PR @aw1621107! I don't think the original bug is still reproducible now (though I'm not sure why not) so I'm going to close this out. (It's also currently moot because no one updated the compiler list to recognize `gcc-8` and because HEAD doesn't even build.)

Given what @fxcoudert said, installing gcc --HEAD as `gcc-HEAD` might be a better idea than installing it as `gcc-8`. The ability to use `fails_with :gcc => "8"` as a way to blacklist HEAD, which was the reason for going with `gcc-8` instead of `gcc-HEAD`, is of dubious value if HEAD and the eventual GCC 8 are not similar to each other.

PR #7649, opened 2016-12-08
caveats: eliminate `brew --prefix` calls in shell rcs

zmwangx:

-----

`brew --prefix` is slow, while shell rcs have to be fast. On my mid-2015 15'' rMBP with 2.5 GHz i7 and 512 GB SSD, `brew --prefix` takes 40–50 ms:

```zsh
$ time (repeat 1000 brew --prefix >/dev/null)
( repeat 1000; do; brew --prefix > /dev/null; done; )  23.20s user 18.35s system 82% cpu 50.538 total
```

It should be slower on HDD and worse still on NFS and likes. Moreover, `brew --prefix` is often called in a construct like

```sh
[ -f "$(brew --prefix)/path/to/script" ] && . "$(brew --prefix)/path/to/script"
```

i.e., it is called twice, consuming ~100 ms. As someone who has been obsessed with the art of Zsh customizations for years, I can confidently say that a 100 ms delay during shell init is in the noticeable range. We don't want to inflate shell init time of unsuspecting users; meanwhile, those advanced users who want to have a set of shell rcs that work accross multiple hosts with brew installed at different locations, or those who have multiple brew installations on a single host that are dynamically switched should know how to use `brew --prefix` (or better, use different hardcoded values for different values for different hosts) without us telling them.

This PR replaces all `$(brew --prefix)/...` in caveats with `#{HOMEBREW_PREFIX}/...` or `#{etc}/...`.

Note that `dvm` is also revisioned to use standard `bash_completion.install`, so if merged bottle needs to be pulled for `dvm`.

### DIFF
--- a/Formula/dvm.rb
+++ b/Formula/dvm.rb
@@ -28,7 +28,7 @@ class Dvm < Formula
 
       system "make", "VERSION=#{version}", "UPGRADE_DISABLED=true"
       prefix.install "dvm.sh"
-      prefix.install "bash_completion"
+      bash_completion.install "bash_completion" => "dvm"
       (prefix/"dvm-helper").install "dvm-helper/dvm-helper"
       prefix.install_metafiles
     end
@@ -37,12 +37,7 @@ class Dvm < Formula
   def caveats; <<-EOS.undent
     dvm is a shell function, and must be sourced before it can be used.
     Add the following command to your bash profile:
-
-        [[ -s "$(brew --prefix dvm)/dvm.sh" ]] && source "$(brew --prefix dvm)/dvm.sh"
-
-    To enable tab completion of commands, add the following command to your bash profile:
-        [[ -s "$(brew --prefix dvm)/bash_completion" ]] && source "$(brew --prefix dvm)/bash_completion"
-
+        [ -f #{opt_prefix}/dvm.sh ] && . #{opt_prefix}/dvm.sh
     EOS
   end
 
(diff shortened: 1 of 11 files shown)


zmwangx:
Forgot to mention: `#{HOMEBREW_PREFIX}` and `#{etc}` are already (more) widely used in such caveats. A very crude search:

```sh
for f in Formula/*; do sed -n '/def caveats/,/plist/p' $f | grep -e HOMEBREW_PREFIX -e '#{etc}' && echo $f; done | grep -v plist_options
```

MikeMcQuaid (approved):

zmwangx:
@MikeMcQuaid There's no performance difference between three lines and one, or between `source` and `.`, and there are more caveats with a variety of styles anyway, but since you asked I normalized every formula touched here to POSIX `test` (`[`) and `.`, converted single pipeline conditionals to lists, and removed quotes around paths (fewer existing paths were quoted than not, and Homebrew shouldn't be installed to a path that needs quoting anyway, or many build or runtime scripts would break).

Formula/dvm.rb
```diff
@@ -3,6 +3,7 @@ class Dvm < Formula
   homepage "https://github.com/getcarina/dvm"
   url "https://github.com/getcarina/dvm/archive/0.6.4.tar.gz"
   sha256 "7ff3bc5271432b7f5da535cd04228bf212dd2965ed15998837d660452094ff0e"
+  revision 1
```
zmwangx: Sure, I dropped the revision. Need to push a rebottle then, since the content has changed.

MikeMcQuaid:
Thanks again @zmwangx!

PR #7650, opened 2016-12-08
eigen 3.3.1

xu-cheng:


ilovezfs:
Please make sure any dependents this breaks in science are at least identified. This formula may actually belong in science because its dependents are almost all over there and it seems to break things most of the time it gets updated in core.

nijikon:
@ilovezfs Maybe we should move it to science then?

ilovezfs:
@nijikon Maybe, but the overall goal of late is to migrate things into core, not out of core. CC @MikeMcQuaid

MikeMcQuaid:
Yeh, things that are heavily used in homebrew/science will start gradually moving to core (after I've done homebrew/versions entirely).

ilovezfs:
In terms of eigen specifically,
```
Josephs-MacBook-Pro:~ joe$ brew uses eigen --include-optional
cgal                              homebrew/science/dgtal            homebrew/science/orocos-kdl
ompl                              homebrew/science/g2o              homebrew/science/pcl
open-babel                        homebrew/science/gnudatalanguage  homebrew/science/shogun
homebrew/science/ceres-solver     homebrew/science/iqtree           homebrew/science/statismo
homebrew/science/cmdstan          homebrew/science/openbr           homebrew/science/trilinos
homebrew/science/cppad            homebrew/science/opencv
homebrew/science/cufflinks        homebrew/science/opencv3
```

ilovezfs:
Actually, this upgrade should be fine. The last upgrade (3.2.10 -> 3.3.0) silently broke at least five formulae in science: ceres-solver (required), cmdstan (optional), cppad (required), dgtal (optional), and orocos-kdl (required), all but one of which are still broken (ceres-solver was fixed). There may be additional cases since I've not tested all of them.

Of the outstanding broken cases, only dgtal is fixed in head.

But this upgrade (3.3.0 -> 3.3.1) is unlikely to cause any additional breakage.

ilovezfs:
:rocket:

xu-cheng:
Thanks all. 

BTW, sorry being less of active recently. But I hope that I can resume my involvement to Homebrew early next year.

PR #7654, opened 2016-12-08
python3: version detection --HEAD packages

yan12125:

-----

As title, it's 3.7 now or the install fails

### DIFF
--- a/Formula/python3.rb
+++ b/Formula/python3.rb
@@ -57,19 +57,6 @@ class Python3 < Formula
   # X11.
   patch :DATA if build.with? "tcl-tk"
 
-  def lib_cellar
-    prefix/"Frameworks/Python.framework/Versions/#{xy}/lib/python#{xy}"
-  end
-
-  def site_packages_cellar
-    lib_cellar/"site-packages"
-  end
-
-  # The HOMEBREW_PREFIX location of site-packages.
-  def site_packages
-    HOMEBREW_PREFIX/"lib/python#{xy}/site-packages"
-  end
-
   # setuptools remembers the build flags python is built with and uses them to
   # build packages later. Xcode-only systems need different flags.
   pour_bottle? do
@@ -88,6 +75,9 @@ class Python3 < Formula
     # and not into some other Python the user has installed.
     ENV["PYTHONHOME"] = nil
     ENV["PYTHONPATH"] = nil
+
+    xy = (buildpath/"configure.ac").read.slice(/PYTHON_VERSION, (3\.\d)/, 1)
+    lib_cellar = prefix/"Frameworks/Python.framework/Versions/#{xy}/lib/python#{xy}"
 
     args = %W[
       --prefix=#{prefix}
@@ -198,7 +188,7 @@ class Python3 < Formula
     rm bin/"2to3"
 
     # Remove the site-packages that Python created in its Cellar.
-    site_packages_cellar.rmtree
+    (prefix/"Frameworks/Python.framework/Versions/#{xy}/lib/python#{xy}/site-packages").rmtree
 
     %w[setuptools pip wheel].each do |r|
       (libexec/r).install resource(r)
@@ -213,6 +203,10 @@ class Python3 < Formula
   end
 
   def post_install
+    xy = (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+    site_packages = HOMEBREW_PREFIX/"lib/python#{xy}/site-packages"
+    site_packages_cellar = prefix/"Frameworks/Python.framework/Versions/#{xy}/lib/python#{xy}/site-packages"
+
     # Fix up the site-packages so that user-installed Python software survives
     # minor updates, such as going from 3.3.2 to 3.3.3:
 
@@ -267,7 +261,8 @@ class Python3 < Formula
       library_dirs << Formula["homebrew/dupes/tcl-tk"].opt_lib
     end
 
-    cfg = lib_cellar/"distutils/distutils.cfg"
+    cfg = prefix/"Frameworks/Python.framework/Versions/#{xy}/lib/python#{xy}/distutils/distutils.cfg"
+
     cfg.atomic_write <<-EOF.undent
       [install]
       prefix=#{HOMEBREW_PREFIX}
@@ -278,11 +273,9 @@ class Python3 < Formula
     EOF
   end
 
-  def xy
-    version.to_s.slice(/(3\.\d)/) || "3.6"
-  end
-
   def sitecustomize
+    xy = (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+
     <<-EOF.undent
       # This file is created by Homebrew and is executed on each python startup.
       # Don't print from here, or else python command line scripts may fail!
@@ -313,7 +306,7 @@ class Python3 < Formula
           # the Cellar site-packages is a symlink to the HOMEBREW_PREFIX
           # site_packages; prefer the shorter paths
           long_prefix = re.compile(r'#{rack}/[0-9\._abrc]+/Frameworks/Python\.framework/Versions/#{xy}/lib/python#{xy}/site-packages')
-          sys.path = [long_prefix.sub('#{site_packages}', p) for p in sys.path]
+          sys.path = [long_prefix.sub('#{HOMEBREW_PREFIX/"lib/python#{xy}/site-packages"}', p) for p in sys.path]
 
           # Set the sys.executable to use the opt_prefix
           sys.executable = '#{opt_bin}/python#{xy}'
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> As title, it's 3.7 now or the install fails

Can we be smarter about this and make it e.g. the stable version + 1? Otherwise I suspect the same thing will happen for 3.8.

yan12125:
> Can we be smarter about this

I guess it's possible. Let me try.

yan12125:
How about change the version scheme for --HEAD packages? Currently it's HEAD-8688a664995c, and 3.7-8688a664995c should make life easier.

yan12125:
I got a version detection mechanism. Hope I'm not writing Ruby like Python :)

Formula/python3.rb
```diff
@@ -310,7 +310,22 @@ def post_install
   end
 
   def xy
-    version.to_s.slice(/(3\.\d)/) || "3.6"
```
MikeMcQuaid: This should do it:
```ruby
     if build.head?
        /3\.(\d+)/ =~ stable.version.to_s
        "3.#{$1.to_i + 1}"
      else
       version.to_s.slice(/(3\.\d)/)
     end
   end
```

ilovezfs: Can't we use the version reported by upstream when you invoke `python --version`?

MikeMcQuaid: @ilovezfs We need this pre-build.
@yan12125 Yes. This could use `(devel || stable).version.to_s` instead.

ilovezfs: pre-build meaning when?

yan12125: OK I found alpha versions are included in some previous commits, for example 75f72923eb6fe4fc6d4d1ad9a2669301274267da, so ```devel``` version is not reliable. Also, updating the devel version may take hours or even days in Homebrew, during which things may break.

@ilovezfs's suggestion is possible. Currently in python3.rb, ```#{xy}``` is used only after ```make``` is invoked, while I guess it's dangerous to assume it will never change in the future. My proposal is indeed lengthy and verbose yet robust, and only upstream changes can break it.

ilovezfs: What file is `python --version` getting the number from originally? I'd use that, since it'll be available whenever as long as the source has been unpacked, which is why I asked what "pre-build" means.

yan12125: > What file is ```python --version``` getting the number from originally

It's from PY_VERSION macro of patchlevel.h. In ```install``` step, I can just read ```Include/patchlevel.h```. However I don't know how to access the source tree in ```post_install```. Is it possible? In installed files, patchlevel.h can be found in following paths:
```
$ find -L /usr/local -name patchlevel.h                                                                                    
/usr/local/Cellar/python3/HEAD-da510d1aa683_3/Frameworks/Python.framework/Versions/3.7/Headers/patchlevel.h
/usr/local/Cellar/python3/HEAD-da510d1aa683_3/Frameworks/Python.framework/Versions/3.7/include/python3.7m/patchlevel.h
/usr/local/Frameworks/Python.framework/Versions/3.7/Headers/patchlevel.h
/usr/local/Frameworks/Python.framework/Versions/3.7/include/python3.7m/patchlevel.h
/usr/local/opt/python3/Frameworks/Python.framework/Versions/3.7/Headers/patchlevel.h
/usr/local/opt/python3/Frameworks/Python.framework/Versions/3.7/include/python3.7m/patchlevel.h
/usr/local/var/homebrew/linked/python3/Frameworks/Python.framework/Versions/3.7/Headers/patchlevel.h
/usr/local/var/homebrew/linked/python3/Frameworks/Python.framework/Versions/3.7/include/python3.7m/patchlevel.h
```
All requires knowing "3.7" first.

ilovezfs: In post install you can do
```
Utils.popen_read(bin/"python3", "--version").match(/ (\d+\.\d+).*/)[1]
```
or
```
Utils.popen_read(bin/"python3", "--version").split(" ").last.to_f.to_s
```
or
```
prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
```
or, or, or ...

ilovezfs: If it's the installed files, you can read it out of any of them in post install.

ilovezfs: For example, `File.read(Dir["#{prefix}/Frameworks/**/patchlevel.h"].pop)`

yan12125: Thanks for all those suggestions. I use the third approach. IMO either parsing or wildcard matching is fragile than direct detection.

Formula/python3.rb
```diff
@@ -310,7 +310,22 @@ def post_install
...
+        if File.file?("configure.ac")
+            # In Python's source tree
+            ver = File.read("configure.ac").slice(/PYTHON_VERSION, (3\.\d)/, 1)
+        else
+            # Installed packages
+            ver = (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+        end
+    end
+
+    raise "Cannot determine Python version" unless ver
```
MikeMcQuaid: I think this probably warrants a wider refactoring where this function is turning into a `xy` variable in `install` and `post_install`. If this function is called elsewhere it will 💥.

yan12125: Before ```make install``` configure.ac is available in the source tree, and after that command Frameworks/Versions/3.x folder is there. I guess if there are future breakages only tweaking paths is necessary.

yan12125: How is my previous idea here: https://github.com/Homebrew/homebrew-core/pull/7654#issuecomment-265752532?

MikeMcQuaid: Nah, the version scheme for HEAD is consistent across packages so would rather not change it, thanks.

yan12125: Thanks for the answer. Should I try refactoring in this pull request?

MikeMcQuaid: Yes, please! Focus on migrating `xy` (and the functions that call it) into variables in `install` and `post_install` instead. Thanks!

yan12125: Got it. Thanks for all the patience for a Ruby newbie. Refactoring this might take some time :)

yan12125:
Hi, I've changed lots of version handling codes in python3.rb. Could anyone have a look?

Formula/python3.rb
```diff
@@ -75,16 +75,16 @@ class Python3 < Formula
...
-    lib_cellar/"site-packages"
+  def site_packages_cellar(xy)
+    lib_cellar(xy)/"site-packages"
   end
 
   # The HOMEBREW_PREFIX location of site-packages.
-  def site_packages
+  def site_packages(xy)
     HOMEBREW_PREFIX/"lib/python#{xy}/site-packages"
   end
```
MikeMcQuaid: Let's just inline all these functions into variables in the needed methods e.g. `install`, `post_install`, etc.

Formula/python3.rb
```diff
@@ -107,6 +107,9 @@ def install
     ENV["PYTHONHOME"] = nil
     ENV["PYTHONPATH"] = nil
 
+    ver = version.to_s.slice(/(3\.\d)/) || File.read("configure.ac").slice(/PYTHON_VERSION, (3\.\d)/, 1)
```
MikeMcQuaid: I think you could just do:
```
(buildpath/"configure.ac").read.slice(/PYTHON_VERSION, (3\.\d)/, 1)
```
unless there's times when that file isn't present? If so, it'd be best to use `build.head?`, `build.stable?` etc. rather than the `||`.

Formula/python3.rb
```diff
@@ -230,34 +233,42 @@ def install
     end
   end
 
+  def ver_from_cellar
+    version.to_s.slice(/(3\.\d)/) || (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+  end
```
MikeMcQuaid: Inline and duplicate this function into e.g. `post_install`

MikeMcQuaid (changes requested):

yan12125:
I've expanded most utility functions into their call sites except ```sitecustomize```, which I think too big to be inlined. But is it good to have so many duplicated codes?

Formula/python3.rb
yan12125: configure.ac is always available during the build time (both stable release tarballs and mercurial repository has it), so I just drop ```version.to_s.slice(/(3\.\d)/)```

yan12125:
Rebased as 3.6.0 is landed. Any updates?

The latest tests on Jenkins failed in fdroidserver. Seems the error is unrelated to Python?
```
==> brew fetch --retry fdroidserver
==> brew install fdroidserver
==> brew linkage --test fdroidserver
==> FAILED
Missing libraries:
  /usr/local/opt/libtiff/lib/libtiff.5.dylib
  /usr/local/opt/webp/lib/libwebp.6.dylib
  /usr/local/opt/webp/lib/libwebpdemux.2.dylib
  /usr/local/opt/webp/lib/libwebpmux.2.dylib
```

MikeMcQuaid (approved):

MikeMcQuaid:
Looks good to me now. @ilovezfs, any thoughts/objections?

ilovezfs:
LGTM.

MikeMcQuaid:

PR #7655, opened 2016-12-08
ola 0.10.3

ilovezfs:


ilovezfs:
A `make check` failure, not new to this version (also affects at 0.10.2), reported upstream: https://github.com/OpenLightingProject/ola/issues/1165.

Formula/ola.rb
```diff
@@ -1,48 +1,35 @@
 class Ola < Formula
   desc "Open Lighting Architecture for lighting control information"
   homepage "https://www.openlighting.org/ola/"
-  url "https://github.com/OpenLightingProject/ola/releases/download/0.10.2/ola-0.10.2.tar.gz"
```
peternewman: The release is here now:
https://github.com/OpenLightingProject/ola/releases/download/0.10.3/ola-0.10.3.tar.gz

peternewman:
Has the "protobuf-c --with-python" issue been fixed @ilovezfs ? If so, and given you've enabled the python libs by default, you may as well add the RDM tests, which are more widely used. They also need python's numpy now.

ilovezfs:
@peternewman I'm not sure what that issue is/was since there's no such option at this point: https://github.com/Homebrew/homebrew-core/blob/master/Formula/protobuf-c.rb

peternewman:
It was referring to https://github.com/Homebrew/legacy-homebrew/pull/48575#issuecomment-177348713

Issue #7213, opened 2016-11-24
unfs3 is broken


royger:
Hello,

unfs3 doesn't seem to work on OS X at all, I've tried several options (-d, -s) and although the destination can mount the share, the data seems to be completely unavailable, and I have to end up force-unmounting it (umount -f) because proceeses trying to access it just hung. I've tried dtracing the unfsd process, but I don't seem to spot anything obviously wrong. I'm using 10.12.1, maybe this was working on previous OS X versions? If not it should be removed IMHO.

Thanks, Roger.

MikeMcQuaid:
To help us debug this issue can you explain:
- What you were trying to do (and why)
- What happened
- What you expected to happen
- Step-by-step reproduction instructions (with as minimal input data as possible)

Also, you have not followed the requested steps on [the Troubleshooting page](https://github.com/Homebrew/brew/blob/master/docs/Troubleshooting.md#troubleshooting)

Please follow (all of) these steps and post the information here so we can help you with your problem.

Thanks!

royger:
Hello,

Sorry, I completely forgot about this.

 - What you were trying to do (and why):

I'm trying to run a NFS server from user-space using unfs3. This should allow me to share folders from OS X to any other OS with NFS support (mostly UNIX ones). I'm doing this because the current nfsd shipped with OS X (10.12.1) crashes the kernel (this is already reported and acknowledged by Apple).

- What happened:

I've created a sample exports file, used it with unfs and mounted it from a FreeBSD box, the result is that when trying to access a folder or subfolder of that share the client gets stuck.

- What you expected to happen:

The client should be able to access the share.

- Step-by-step reproduction instructions (with as minimal input data as possible):

1. Create a sample exports file (in my case I'm using /Users/royger/exports), and it contains the following line:

/Users/royger/src    nuc(rw,anonuid=501,anongid=20,all_squash)

I'm exporting folder ~/src, and all accesses should be mapped to the id of my user (501) and the group id of staff (20).

2. Start unfsd:

$ sudo /usr/local/sbin/unfsd -e /Users/royger/exports -d

3. Mount the share from another box:

$ mount mac:/Users/royger/src src

4. Try to access the share:

$ ls src/foo

5. ls gets stuck.

I've done what the Troubleshooting guide describes, but AFAICT this is not a homewbrew issue, it's just that unfs3 simply doesn't work on OS X[0] so it should be removed from homebrew.

Roger.

[...]

PR #7667, opened 2016-12-08
unfs3: migrate to boneyard

MikeMcQuaid:
Created with `brew boneyard-formula-pr` because not functional on macOS (see https://github.com/Homebrew/homebrew-core/issues/7213).

Fixes https://github.com/Homebrew/homebrew-core/issues/7667.


MikeMcQuaid:
@BrewTestBot test this please

ryansch:
@MikeMcQuaid This formula works just fine on macOS Sierra.  It's a dependency of dinghy (https://github.com/codekitchen/dinghy) which is now uninstallable without jumping through hoops.

codekitchen:
@MikeMcQuaid to echo what @ryansch said... my project [Dinghy](https://github.com/codekitchen/dinghy) uses unfs3 installed via homebrew. I can say for certain that there are at least 100 people inside and outside my company using this setup, and it's been working fine for over 2 years, starting with OS X 10.9 I believe, and still working fine in MacOS 10.12.

So while I haven't dug into why the initial reporter couldn't get it working, I'm very confident in saying that the homebrew install of unfs3 works on OS X, and should not have been moved to boneyard. What can I do to help get it re-instated? In the meantime I'm going to add it to [Dinghy's homebrew tap](https://github.com/codekitchen/homebrew-dinghy) so that people can install Dinghy again. Thanks.

ilovezfs:
@codekitchen if you can open a PR adding it to core and CI is :green_apple: that would be a good first step.

codekitchen:
Thanks @ilovezfs. I pushed up https://github.com/Homebrew/homebrew-core/pull/7825 as a straight revert of this commit, but it looks like that's not going to pass CI, since it appears that CI now fails if there's no `test do` block in the formula and unfs3 didn't have one.

It'd be nice if we could grandfather this back in, but if necessary I can have a go at adding a test block. It may take me a bit.

codekitchen:
OK I got a chance to look over the docs on adding a test, and I added what I think is a reasonable test block. Moving this conversation over to #7825

PR #7712, opened 2016-12-09
lua 5.3.3

hatzka-nezumi:


ilovezfs:
This is in homebrew/versions: https://github.com/Homebrew/homebrew-versions/blob/master/lua53.rb.

PR #7731, opened 2016-12-09
Idris 0.99

Termina1:

-----

### DIFF
--- a/Formula/idris.rb
+++ b/Formula/idris.rb
@@ -5,8 +5,8 @@ class Idris < Formula
 
   desc "Pure functional programming language with dependent types"
   homepage "http://www.idris-lang.org"
-  url "https://github.com/idris-lang/Idris-dev/archive/v0.12.3.tar.gz"
-  sha256 "c6f410cddddbc53c4779d3612be40ef4e4f1f11ce8083a811825763daf30ee4d"
+  url "https://github.com/idris-lang/Idris-dev/archive/v0.99.tar.gz"
+  sha256 "bc5b497807dacca741b1ab971c145f705230bb4430961166f54dd46c4249e6af"
   head "https://github.com/idris-lang/Idris-dev.git"
 
   bottle do


woodruffw:
Thanks @Termina1!

ilovezfs:
Pretty sure 0.99 is still considered pre-release which is why I did not do this bump.

ilovezfs:
Yeah: http://www.idris-lang.org/towards-version-1-0/

"We’ll shortly be releasing Idris version 0.99, which is essentially an alpha release of 1.0, and we’ll keep releasing updates until 1.0 is ready, probably around February 2017"

But I suppose it's _possible_ upstream considers 0.99 to supersede 0.12.3, even if it's calling 0.99 **alpha**? Of course in that case, I'm not sure why idris is in core at all.

Termina1:
@ilovezfs I asked on irc channel if this version is stable, and received an answer "stable as the language is".

ilovezfs:
@Termina1 OK. Looks like we're the only ones that have moved to 0.99 but so be it.
```
Distributions 	Arch:0.12.3, LTSHaskell:0.12.3, NixOS:0.12.3, Tumbleweed:0.12.3
```

Termina1:
May be we could ask @jfdm to tell us more?

woodruffw:
Hmm, I didn't realize. If we're the only ones on 0.99, it might make sense to roll back.

ilovezfs:
@woodruffw given that @edwinb uploaded it to Hackage, I think it'll be fine.

jfdm:
Dear all;

When @Termina1 asked his question about Idris' stability it came across as being a general questions and not in reference to home brew formula. In response I referred to the blog posting: 

  http://www.idris-lang.org/towards-version-1-0/

in which our notion of stability and what '1.0' is defined.

Yes, Version 0.99 of Idris is stable, but we make explicit reference to language stability. Some parts of Idris' internals might be revised, and may need some refinement. Specifically:

> Mostly, what we mean by calling a release “1.0” is that there are large parts of the language and libraries that we now consider stable, and we promise not to change them without also changing the version number appropriately. 

To some degree, and this is a personal opinion, notions of '1.0' and alpha and beta and production readiness are not necessarily conducive to research software. This is partly why alpha was not used in the release notes, and why we define our own notion of stability.

@ilovezfs Idris developers are not responsible for downstream release of Idris outside precompiled binaries. There will naturally be a lag: 0.99 is less than a week old.

@woodruffw If the version number in home-brew is the same as seen in Hackage then it is fine.

If future it would be nice to if people contacted: the developers of Idris directly; the mailing list; or ask in IRC with direct reference to home-brew.

ilovezfs:
@jfdm the bottom line here is whether upstream wants us distributing 0.99 or 0.12.3 right now.

Termina1:
@jfdm I guess, this is why I mentioned you in this thread, I'm not sure how to be more specific than that. Also, there wasn't a question about Idris overall stability, but specifically about 0.99 version. But I apologise if I interpreted you words in a wrong way.

jfdm:
> the bottom line here is whether upstream wants us distributing 0.99 or 0.12.3 right now.

Version 0.99. It is a *normal* release that just so happens to be a precursor to our v1.0.

ilovezfs:
@jfdm thanks for making that explicit. I apologize for reading too much into "essentially an alpha release of 1.0." Around here, we usually won't even package software that designates itself "beta," so "alpha" is like putting a red flag in front of a bull :)

PR #7752, opened 2016-12-10
libbpg 0.9.7

bilhackmac:


Formula/libbpg.rb
```diff
@@ -29,6 +29,8 @@ def install
     args << "USE_X265=" if build.without? "x265"
 
     system "make", "install", "prefix=#{prefix}", "CONFIG_APPLE=y", *args
+
+    share.install Dir["html/bpgdec*.js"]
```
MikeMcQuaid: How are these useful in `share`? I'm trying to figure out how they'd be used?

bilhackmac: I don't really know where to put these files. They must be copied to a web site to enable viewing bpg images in your browsers.

MikeMcQuaid: Cool. Just use `pkgshare` instead of `share`.

MikeMcQuaid (changes requested):

bilhackmac:
Update done

MikeMcQuaid (approved):

nijikon:

PR #7760, opened 2016-12-11
ndenv 0.4.0 (new formula)

matadon:

-----

`ndenv` is a Node version manager based off `rbenv`, and in a similar vein, this formula is based on the `rbenv` formula.

### DIFF
--- a/Formula/ndenv.rb
+++ b/Formula/ndenv.rb
@@ -1,0 +1,31 @@
+class Ndenv < Formula
+  desc "Node version manager"
+  homepage "https://github.com/riywo/ndenv"
+  url "https://github.com/riywo/ndenv/archive/v0.4.0.tar.gz"
+  sha256 "1a85e4c0c0eee24d709cbc7b5c9d50709bf51cf7fe996a1548797a4079e0b6e4"
+  head "https://github.com/riywo/ndenv.git"
+
+  depends_on "node-build" => :recommended
+
+  def install
+    inreplace "libexec/ndenv" do |s|
+      if HOMEBREW_PREFIX.to_s != "/usr/local"
+        s.gsub! ":/usr/local/etc/ndenv.d", \
+            ":#{HOMEBREW_PREFIX}/etc/ndenv.d\\0"
+      end
+    end
+
+    if build.head?
+      git_revision = `git rev-parse --short HEAD`.chomp
+      inreplace "libexec/rbenv---version", /^(version=)"([^"]+)"/, \
+          %Q(\\1"\\2-g#{git_revision}")
+    end
+
+    prefix.install "bin", "completions", "libexec"
+    system "#{bin}/ndenv", "rehash"
+  end
+
+  test do
+    shell_output "eval \"$(#{bin}/ndenv init -)\" && ndenv versions"
+  end
+end


Formula/ndenv.rb
```diff
@@ -0,0 +1,41 @@
...
+  desc "Node version manager"
+  homepage "https://github.com/riywo/ndenv"
+  url "https://github.com/riywo/ndenv/archive/v0.4.0.tar.gz"
+  sha256 "1a85e4c0c0eee24d709cbc7b5c9d50709bf51cf7fe996a1548797a4079e0b6e4"
+  head "https://github.com/riywo/ndenv.git"
+
+  depends_on "node-build" => :recommended
+
+  def install
+    inreplace "libexec/ndenv" do |s|
```
MikeMcQuaid:
```
if HOMEBREW_PREFIX.to_s != "/usr/local"
  inreplace "libexec/ndenv", ":/usr/local/etc/ndenv.d", ":#{HOMEBREW_PREFIX}/etc/ndenv.d\\0"
end
```

Also, it's worth requesting to upstream to provide a dedicated `configure` or similar to handle this rather than our patching.

Formula/ndenv.rb
```diff
@@ -0,0 +1,41 @@
...
+      if HOMEBREW_PREFIX.to_s != "/usr/local"
+        s.gsub!(":/usr/local/etc/ndenv.d",
+            ":#{HOMEBREW_PREFIX}/etc/ndenv.d\\0")
+      end
+    end
+
+    if build.head?
+      git_revision = `git rev-parse --short HEAD`.chomp
+      inreplace("libexec/rbenv---version", /^(version=)"([^"]+)"/,
+          %Q(\\1"\\2-g#{git_revision}"))
```
MikeMcQuaid: Similarly, would be good to do this somehow without patching.

Formula/ndenv.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    prefix.install("bin", "completions", "libexec")
```
MikeMcQuaid: `completions` should instead be installed with e.g. `bash_completion.install`

Formula/ndenv.rb
```diff
@@ -0,0 +1,41 @@
...
+    system("#{bin}/ndenv", "rehash")
```
MikeMcQuaid: Skip the parentheses here and everywhere above and move this line into a `post_install` block.

Formula/ndenv.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def caveats; <<-EOS.undent
+    ndenv stores data under ~/.ndenv by default. If you absolutely need to
+    store everything under Homebrew's prefix, include this in your profile:
+      export NDENV_ROOT=#{var}/ndenv
+
+    To enable shims, run this and follow the instructions:
+      ndenv init
+    EOS
+  end
```
MikeMcQuaid: Remove all these caveats, they aren't Homebrew specific.

MikeMcQuaid (changes requested):

Formula/ndenv.rb
matadon: Can do.

Formula/ndenv.rb
matadon: This is set up the same way that the `rbenv` formula handles its post-install instructions. Took a quick look, and `postgresql` does the same. Is there a better way that I could handle this?

ilovezfs:
@BrewTestBot test this please

Formula/ndenv.rb
MikeMcQuaid: They aren't necessary at all so we can just pass on them. We've got other formulae (e.g. `rbenv`) where they should probably be removed but we're stricter on new formulae than existing ones.

MikeMcQuaid: @matadon Remove entirely I think, thanks!

matadon:
Pulled `caveats` as requested.

Formula/ndenv.rb
matadon: So, upstream doesn't have any actual `configure` or build script -- it's just a collection of shell scripts.

MikeMcQuaid: Yeh, it'd be worth requesting something long-term from them.

MikeMcQuaid (approved):

MikeMcQuaid:

Issue #7763, opened 2016-12-11
[Question] Make a formula depend on Xcode 8 only if it's built from sources


AliSoftware:
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [X] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Context of the Problem

I'm the author of SwiftGen, which is a tool written in Swift and published via Homebrew ([Formula here](https://github.com/Homebrew/homebrew-core/blob/master/Formula/swiftgen.rb))

Because SwiftGen is written in Swift 3, it requires Xcode 8+ to build from source. To address this, I've [added the `depends_on :xcode => "8.0"` line in my Formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/swiftgen.rb).

The problem is that once Homebrew's Jenkins have built the bottles for SwiftGen and people start installing SwiftGen from them, thus getting the precompiled version of SwiftGen instead of having to build it from Source, they still have this requirement, while it's not needed.

### TL;DR

Is it possible to depend on a version of Xcode only for building the formula from source, but discard it when installed from bottle?

#### Only use the Xcode 8 dependency if built from source, not downloaded from bottle?

The consequence of all that being that Homebrew refuses to install the formula if the user has `xcode-select -p` set to Xcode 7… even if, when installed from the bottles, it's not needed to have Xcode 8 installed and `xcode-select`'d

Is there a way to indicate in my formula that th Xcode 8 dependency is only for building SwiftGen from source, but not necessary when downloaded via the bottle?

#### Depend on installed Xcode, not selected Xcode?

Wouldn't it be better if `depends_on :xcode => xcv` would check if Xcode version xcv was *present* on the machine, even if it wasn't the one selected? For example if the user has both Xcode 7 and Xcode 8 on their machine but `xcode-select`'d Xcode 7, it's still possible to build with Xcode 8 by setting the `DEVELOPER_DIR=…` env variable in the shell executing the installation commands.

That's what I actually do in SwiftGen's Rakefile — using `mdfind` to search for all Xcodes present on the user's system with a given version then get its path and use it as a `DEVELOPER_DIR` to execute my commands. This way, it works even if the Xcode I want to use isn't the default selected one.

MikeMcQuaid:
[...]

PR #7764, opened 2016-12-11
swiftgen: fix Xcode dependency

AliSoftware:

-----

Xcode 8 dependency was always only needed for building SwiftGen from source, not for running it.
See #7763

---

_Note: I've wondered about adding `revision 1`, but since this change doesn't require SwiftGen's bottles to be rebuilt and the generated and installed version would be exactly the same, so this formula doesn't change anything for people who already installed SwiftGen 4.0.1 I think that doesn't really require to generate a `4.0.1_1` version that would be the exact same as 4.0.1 binary-wise (and lib-dependency-wise), as all it does is loosen the requirements._


MikeMcQuaid:
Thanks again for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #7792, opened 2016-12-12
mmark 1.3.4 (new formula)

dericed:


dericed:
In ` brew audit --new-formula mmark.rb` I only get this error:

>  * GitHub fork (not canonical repository)

I hope the patch can be considered despite this as mmark contains significant development not present in the canonical repo. mmark has 1,134 commits while the canonical blackfriday has 391.

woodruffw:
Pinging @ilovezfs since I'm not sure what our policy is here -- both seem to be under semi-active development, `mmark` has more commits but `blackfriday` seems to be more publicized.

ilovezfs:
@woodruffw there's no hard rule, but typically would be guided by who has the lion's share of users.

dericed:
Cc'ing @miekg (maintainer of mmark), but want to note that mmark and blackfriday have different user groups. As an IETF contributor I regularly use mmark, but haven't had any need to use blackfriday which is very different. mmark is 877 commits ahead and 134 commits behind blackfriday. I know with other forks, such as libav and ffmpeg, homebrew has supported installers for both rather than only the canonical one.

miekg:
[ Quoting <notifications@github.com> in "Re: [Homebrew/homebrew-core] mmark ..." ]
>Cc'ing @miekg (maintainer of mmark), but want to note that mmark and blackfriday have different user groups. As an IETF contributor I regularly use mmark, but haven't had any need to use blackfriday which is very different. mmark is 877 commits ahead and 134 commits behind blackfriday. I know with other forks, such as libav and ffmpeg, homebrew has supported installers for both rather than only the canonical one.

mmark was forked from blackfriday but has diverged significantly and it *not*
blackfriday anymore.

I have vague plans to use blackfriday's AST parser and redo mmark using that -
it would then still be a complete fork.

Mmark is now a new markdown dialect.

This no-fork policy is silly

ilovezfs:
>This no-fork policy is silly

You can save the editorializing. As I said above, there is no hard rule here. The audit check is only run on new formulae so if we decide it's fine, then it won't subsequently cause an audit failure the next time the formula goes through CI as a non-new formula.

MikeMcQuaid:
@miekg Also, if you wish you can get someone who works at GitHub (i.e. me) to make it not a fork any more if you'd like that. That said, there's no reason to do that purely for Homebrew's benefit.

MikeMcQuaid (approved):

MikeMcQuaid:

For future reference the preferred commit message format is `mmark 1.3.4 (new formula)`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #7793, opened 2016-12-12
zsh-autosuggestions: install zsh-autosuggestions with zsh_function

zachwhaley:

-----

Instead of installing the zsh-autosuggestions zsh script to `prefix` and asking the user to

### DIFF
--- a/Formula/zsh-autosuggestions.rb
+++ b/Formula/zsh-autosuggestions.rb
@@ -7,14 +7,14 @@ class ZshAutosuggestions < Formula
   bottle :unneeded
 
   def install
-    pkgshare.install "zsh-autosuggestions.zsh"
+    zsh_function.install "zsh-autosuggestions.zsh" => "zsh-autosuggestions"
   end
 
   def caveats
     <<-EOS.undent
     To activate the autosuggestions, add the following at the end of your .zshrc:
 
-      source #{HOMEBREW_PREFIX}/share/zsh-autosuggestions/zsh-autosuggestions.zsh
+      autoload -Uz zsh-autosuggestions && zsh-autosuggestions
 
     You will also need to force reload of your .zshrc:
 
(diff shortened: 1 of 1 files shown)


zachwhaley:
Hello @zsh-users, since you are the maintainers of this project, I thought it might be prudent to get your opinion on this review.

Thanks!

zachwhaley:
I'm not sure why, but the Yosemite test seems to be the only one failing o.O
https://bot.brew.sh/job/Homebrew%20Core/13108/version=yosemite/

zmwangx:
`zsh-autosuggestions` isn't meant to be a function, so you're polluting the global namespace, and less experienced users might be tricked into thing it's a function that should be called repeatedly. Not sure it's the best idea.

zmwangx:
Oh, and not to mention you're going to break people's existing runcoms with this questionable change. Same goes for the other two PRs.

zachwhaley:
@zmwangx Thanks for the info!

I'm curious now, why is this kind of "function" loading used for things like promptinit, compinit, and colors?
They seem to be similar to zsh-git-prompt in that they are not functions, and they require an initial run of the "function" after an autoload, e.g. `autoload -Uz prompinit && prompinit`

Thanks,
Zach

zmwangx:
> why is this kind of "function" loading used for things like promptinit, compinit, and colors?

Yeah that's a good point. One reason I can think of is that people expect use to stock functionalities without knowing where the files are installed to; those are implementation details to some extent. External packages, on the other hand, are fully controlled by you. An alternative approach (that could be built in) is something like `zmodload`, where a bunch of parameters and functions (not just the initializer) are available as soon as you load a module. (I'm only talking about the intended behavior here; actual zsh modules are a different breed.) IMO that's much more elegant. Prezto, for instance, has `pmodload` that works like this (sourcing a bunch of init scripts and loading a directory of functions), and its kinship to `zmodload` is obvious. `antigen bundle` (`-antigen-load`) basically works this way, too.

Also, when a function is named like `*init` it's much more obvious that it's an initializer which isn't intended to be called afterwards. Which is true for both `promptinit` and `compinit`.

Still, the great majority of functions in `$prefix/share/zsh/functions` are meant to be functions.

PR #7794, opened 2016-12-12
nuxeo 8.10

akervern:


akervern:
Hi, I got those error during the first build. But after doing some `brew tap XXX` as expected the build passed; I thought it was related to my environment. 
So... I just don't understand where those errors are tied from. Any hints?

dunn:
I think there's an issue with Jenkins.  Hold tight.

ilovezfs:
upgrading xcode.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

woodruffw:
Thanks @akervern! For future reference, `<name> <version>` is the preferred format for commits that update a packages' version.

PR #7795, opened 2016-12-12
WIP: zsh-git-prompt: install zsh-git-prompt with zsh_function

zachwhaley:

-----

Instead of installing the zsh-git-prompt zsh script to `prefix` and asking the user to


zachwhaley:
Hello @olivierverdier, since you are the maintainers of this project, I thought it might be prudent to get your opinion on this review.

Thanks!

zmwangx:
For the same reason #7798 and #7793, this is an abuse of the function concept, and it pollutes the namespace. The inreplace makes it even less ideal.

PR #7798, opened 2016-12-12
WIP: antigen: install with zsh_function

zachwhaley:

-----

Instead of installing the antigen zsh script to `prefix` and asking the user to

### DIFF
--- a/Formula/antigen.rb
+++ b/Formula/antigen.rb
@@ -8,17 +8,17 @@ class Antigen < Formula
   bottle :unneeded
 
   def install
-    pkgshare.install "bin/antigen.zsh"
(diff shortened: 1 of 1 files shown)


zmwangx:
Similar to https://github.com/Homebrew/homebrew-core/pull/7793#issuecomment-266573598, `antigen.zsh` isn't a function definition; it's an init script with configuration statements and function definitions that should be sourced once. The difference from the `zsh-autosuggestions` case is that here your autoloaded `antigen` "function" will replace itself with an actual `antigen` upon first execution, which arguably makes it even more confusing. (`autoload -Uz antigen && antigen help` doesn't work, `autoload -Uz antigen && antigen && antigen help` works.) Again, 👎

PR #7801, opened 2016-12-12
termsql 0.3 (new formula)

kaeff:
Hi,

I'd like to propose this new formula. I'm not the package author, nor experienced with writing formulas for python projects, but I've made sure to follow the docs as best as I can. Happy for your feedback, thanks in advance!

cf. https://github.com/tobimensch/termsql/issues/3

### DIFF
--- a/Formula/termsql.rb
+++ b/Formula/termsql.rb
@@ -10,7 +10,7 @@ class Termsql < Formula
 
   def install
     # Replace path to man pages in hard-coded setup.py
-    inreplace "setup.py", %r{/usr/share/man/man1/}, man1
+    inreplace "setup.py", %r{/usr/share/man/man1/}, "#{man1}/"
 
     system "python", *Language::Python.setup_install_args(prefix)
   end
--- a/Formula/termsql.rb
+++ b/Formula/termsql.rb
@@ -17,6 +17,6 @@ class Termsql < Formula
 
   test do
     # Run an example query (count processes)
-    system "ls -lha /usr/bin/* | termsql -w -r 8 \"select count(*) from tbl\""
+    assert_equal "bar", pipe_output("#{bin}/termsql 'select COL1 from tbl'", "foo bar baz").strip
   end
 end
--- a/Formula/termsql.rb
+++ b/Formula/termsql.rb
@@ -1,0 +1,20 @@
+class Termsql < Formula
+  desc "Convert text into SQL table and query it"
+  homepage "https://tobimensch.github.io/termsql/"
+  url "https://github.com/tobimensch/termsql.git"
+  version "0.3"
+
+  depends_on :python => ["sqlite3", "sqlparse"]
+
+  def install
+    # Replace path to man pages in hard-coded setup.py
+    inreplace "setup.py", %r{/usr/share/man/man1/}, man1
+
+    system "python", *Language::Python.setup_install_args(prefix)
+  end
+
+  test do
+    # Run an example query (count processes)
+    system "ls -lha /usr/bin/* | termsql -w -r 8 \"select count(*) from tbl\""
+  end
+end
--- a/Formula/termsql.rb
+++ b/Formula/termsql.rb
@@ -1,8 +1,10 @@
 class Termsql < Formula
   desc "Convert text into SQL table and query it"
   homepage "https://tobimensch.github.io/termsql/"
-  url "https://github.com/tobimensch/termsql.git"
-  version "0.3"
+
+  url "https://github.com/tobimensch/termsql/archive/22501e5c982ec04229643802e36756feb4687e89.tar.gz"
+  version "22501e5c982ec04229643802e36756feb4687e89"
+  sha256 "d8fabe96f57153a2e2852995d66f4506cf593bae4a6b9d7d76d3aef6150fb07e"
 
   depends_on :python => ["sqlite3", "sqlparse"]
 


Formula/termsql.rb
```diff
@@ -0,0 +1,21 @@
...
+  desc "Convert text into SQL table and query it"
+  homepage "https://tobimensch.github.io/termsql/"
+  url "https://github.com/tobimensch/termsql.git", :using => :git
+  version "0.3"
+
+  depends_on :python => ["sqlite3", "sqlparse"]
+
+  def install
+    # Replace path to man pages in hard-coded setup.py
+    new_contents = File.read("setup.py").gsub(%r{\/usr\/share\/man\/man1\/}, man1)
```
woodruffw: This can be done with `inreplace` instead, to avoid separate read/writes:

```ruby
inreplace "setup.py" do |s|
  s.gsub! %r{/usr/share/man/man1/}, man1
end
```

You should also report it upstream (build parameters never hurt :wink:)

woodruffw (changes requested):

Formula/termsql.rb
```diff
@@ -0,0 +1,21 @@
+class Termsql < Formula
+  desc "Convert text into SQL table and query it"
+  homepage "https://tobimensch.github.io/termsql/"
+  url "https://github.com/tobimensch/termsql.git", :using => :git
```
woodruffw: Should be fine to leave `:using => :git` off due to the `.git` suffix.

Formula/termsql.rb
```diff
@@ -0,0 +1,21 @@
...
+  version "0.3"
```
woodruffw: Where is this version coming from? I can't find any tags/releases on upstream.

woodruffw:
Thanks for taking the time to contribute @kaeff! I've left some comments for you to look over, let me know if you have any questions.

kaeff:
Thanks @woodruffw ! 

I asked @tobimensch about adding a version tag and making the man1 path configurable here: tobimensch/termsql#3 . I've guessed the version from running `termsql -v` after installing. Is that an acceptable workaround?

woodruffw:
Thanks for raising the issues with upstream.

Unfortunately, getting the version from `-v` isn't an acceptable workaround -- Homebrew needs some kind of immutable tag or archive (and corresponding `sha256`). If this wasn't the case, any upstream could break their OS X build at any point and we would be caught fielding support for them.

kaeff:
How about using the commitish as the version?

woodruffw:
Unfortunately, no. The version should be a stable release put out by upstream. 

@tobimensch Could you put out either a tag or a full release for `termsql`? That's the main blocker on this formula.

Formula/termsql.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  url "https://github.com/tobimensch/termsql/archive/22501e5c982ec04229643802e36756feb4687e89.tar.gz"
+  version "22501e5c982ec04229643802e36756feb4687e89"
+  sha256 "d8fabe96f57153a2e2852995d66f4506cf593bae4a6b9d7d76d3aef6150fb07e"
+
+  depends_on :python => ["sqlite3", "sqlparse"]
+
+  def install
+    # Replace path to man pages in hard-coded setup.py
+    inreplace "setup.py", %r{/usr/share/man/man1/}, man1
```
ilovezfs: `man1` doesn't have a trailing slash, so this looks sketchy to me

woodruffw: Shouldn't be necessary according to the spec (https://docs.python.org/2/distutils/setupscript.html#installing-additional-files), but we can append a slash anyway for consistency.

woodruffw: @bfontaine I think I understood that, I guess I'm not clear on what the problem is. Looking at the [file itself](https://github.com/tobimensch/termsql/blob/master/setup.py):

```
data_files=[('/usr/share/man/man1/', ['termsql.1.gz'])]
```

should be changed to something like this:

```
data_files=[('/usr/local/Cellar/termsql/0.3/share/man/man1', ['termsql.1.gz'])]
```

There's no trailing slash there, but it shouldn't matter to `setup.py` according to the spec above. Please correct me if I'm misunderstanding something!

MikeMcQuaid: Let's use "#{man1}/" here, thanks.

bfontaine: @woodruffw Thanks, that’s clear. I didn’t look at the file itself and thought we might replace e.g. `"/usr/share/man/man1/foo.1"` with `"#{man1}foo.1"`.

Formula/termsql.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    system "python", *Language::Python.setup_install_args(prefix)
+  end
+
+  test do
+    # Run an example query (count processes)
+    system "ls -lha /usr/bin/* | termsql -w -r 8 \"select count(*) from tbl\""
```
MikeMcQuaid: Use `pipe_input` here instead of a manual pipe and `assert_match` on the output.

MikeMcQuaid (changes requested):

kaeff:
@woodruffw @MikeMcQuaid I've addressed your requested changes. Could you have another look? The upstream tag is still missing unfortunately.

Formula/termsql.rb
```diff
@@ -0,0 +1,22 @@
+class Termsql < Formula
+  desc "Convert text into SQL table and query it"
+  homepage "https://tobimensch.github.io/termsql/"
+
+  url "https://github.com/tobimensch/termsql/archive/22501e5c982ec04229643802e36756feb4687e89.tar.gz"
+  version "22501e5c982ec04229643802e36756feb4687e89"
+  sha256 "d8fabe96f57153a2e2852995d66f4506cf593bae4a6b9d7d76d3aef6150fb07e"
+
+  depends_on :python => ["sqlite3", "sqlparse"]
```
MikeMcQuaid: These should be vendored instead e.g. like `ansible.rb`

Formula/termsql.rb
```diff
@@ -0,0 +1,22 @@
+class Termsql < Formula
+  desc "Convert text into SQL table and query it"
+  homepage "https://tobimensch.github.io/termsql/"
+
+  url "https://github.com/tobimensch/termsql/archive/22501e5c982ec04229643802e36756feb4687e89.tar.gz"
```
MikeMcQuaid: Still need an upstream tag before we can include this, I'm afraid.

MikeMcQuaid (changes requested):

fxcoudert:
Closing for now, please reopen if circumstances change.

PR #7825, opened 2016-12-13
unfs3: fix removal to boneyard

codekitchen:
This formula was migrated to boneyard because of a report of it not working at all
on OS X, but it does work. See the discussion in
https://github.com/Homebrew/homebrew-core/pull/7667

This reverts commit 0b48a2e0efa738545e2cf278da6ea7e6d4b718de.

### DIFF
--- a/Formula/unfs3.rb
+++ b/Formula/unfs3.rb
@@ -1,0 +1,30 @@
+class Unfs3 < Formula
+  desc "User-space NFSv3 server"
+  homepage "http://unfs3.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/unfs3/unfs3/0.9.22/unfs3-0.9.22.tar.gz"
+  sha256 "482222cae541172c155cd5dc9c2199763a6454b0c5c0619102d8143bb19fdf1c"
+
+  bottle do
+    cellar :any_skip_relocation
+    sha256 "39efb568edcd8eb1898deb664c1f016baed60b60cb9d6031743da32b6f615cd3" => :sierra
+    sha256 "b1387de21ce9672d5caea47ff223fdbca37e0ed08137a252ae14c7c80dea36e1" => :el_capitan
+    sha256 "8daaa9ce7c48d3a0efbf72dac2e7cdb429aa7aa8475b64837eaa4f5159b2a4d8" => :yosemite
+    sha256 "43e89011e0472b5a7aea6583f930d05d18a449474617f5d11b10b10b52badc66" => :mavericks
+  end
+
+  head do
+    url "http://svn.code.sf.net/p/unfs3/code/trunk/"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+  end
+
+  def install
+    ENV.j1 # Build is not parallel-safe
+    system "./bootstrap" if build.head?
+    system "./configure", "--disable-debug", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+end
--- a/Formula/unfs3.rb
+++ b/Formula/unfs3.rb
@@ -27,4 +27,11 @@ class Unfs3 < Formula
     system "make"
     system "make", "install"
   end
+
+  test do
+    (testpath/"exports").write <<-EOS.undent
+    "#{Dir.pwd}" 192.168.0.1(ro)
+    EOS
+    system("#{sbin}/unfsd", "-T", "-e", (testpath/"exports").to_s)
+  end
 end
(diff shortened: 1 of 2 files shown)


codekitchen:
The first commit is just a straight revert of the removal, the second commit adds a test block so that this will past CI. The test block verifies that unfsd can run and parse a sample config file.

twardnw:
The tests are passing, can we get this resolved? This broke the installation of our development environment installer.

MikeMcQuaid:
Sorry but I'm 👎 on adding this back. This software isn't supported by upstream, hasn't been updated since 2009. CC @ilovezfs @woodruffw @nijikon for thoughts here.



> The tests are passing, can we get this resolved? This broke the installation of our development environment installer.

@twardnw Please be patient. Homebrew is run by volunteers in their free time and provided for no financial cost to you. I've been away on a work trip so have been unable to get to this.

woodruffw:
I think boneyard is best here, since unmaintained projects inevitably break and leave us with the mess.

Downstream developers can still install from boneyard, and this better reflects the status/reliability of the package.

codekitchen:
Hm well, that's disappointing but at least I can now understand your reasoning. We'll work around it, I may just pull it into Dinghy's Homebrew tap and deal with any issues as they arise. We are trying to deprecate Dinghy itself, but there are still issues with Docker for Mac we're waiting on fixes for.

Can I suggest more transparency around the process of removing a formula? Understand that from our perspective, we saw a program that we'd been installing via Homebrew for years, and that we use daily, removed apparently just because one person reported an issue getting it to work. Knowing the full reasoning would have been helpful.

Of course that said, we do appreciate all you folks do. Thanks.

MikeMcQuaid:
> Can I suggest more transparency around the process of removing a formula? Understand that from our perspective, we saw a program that we'd been installing via Homebrew for years, and that we use daily, removed apparently just because one person reported an issue getting it to work. Knowing the full reasoning would have been helpful.

That was most of the reasoning. Adding to that was the very small number of installs in analytics and upstream appearing dead and not supporting officially macOS.

Things you could do to help:
- keep an eye on the issue tracker. I appreciate the UI doesn't make it that easy to filter based on formualae
- encourage anyone you know using `unfs3` with analytics disabled to no longer disable analytics

I'll let some others wade in on this, though.

woodruffw:
Visibility seems to be a recurring issue with boneyardings/removals. It should be pretty easy to set up a  commit hook that filters down to `/migrate to boneyard/` commits and tweets using MacHomebrew (or a dedicated account).

Thoughts?

ilovezfs:
@woodruffw I'm not excited about the prospect of publishing obituaries on Twitter.

>removed apparently just because one person reported an issue getting it to work.

For what it's worth, that's just not how boneyarding decisions are ever made. In almost all cases it is some combination of any or all of the following:

1. zero or very few installs according to the analytics
2. absentee upstream
3. issues we've filed are not responded to for an extended period
4. persistent build, run-time, or test issues
5. upstream that indicates an intention not to fix known macOS specific issue(s)
6. violation of general brew guidelines (e.g., does not work with clang on latest macOS and latest Xcode), usually long-standing, occasionally new
7. depends on formula(e) with a combination of any of the above

I think generally when users are asking us to get more specific about why we boneyarded something, they're really just saying "I wish you had not boneyarded that formula."

So "increasing transparency" by getting more specific about which subset of the above issues is active in a particular boneyarding decision is unlikely to make anyone happier. The answer is almost always the same: you can create your own tap. The response from the user is also almost always the same: I still wish you hadn't boneyarded it.

woodruffw:
@ilovezfs Obituaries on the main twitter probably wouldn't be very good, but I think it would be nice if we had a more visible artifact of boneyarding than just `tap_migrations`. Perhaps a separate twitter, or a informational message after checking for `homebrew/boneyard` specifically in `tap_migrations` when an install search fails?

codekitchen:
That's a great list @ilovezfs. I'm not sure why you think it wouldn't be helpful to communicate those reasons when removing a formula. Having that reasoning explained on the commit removing this formula would've been enough for us to understand why it was removed, and I wouldn't have tried to get this PR in.

I'm sure it wouldn't satisfy everybody in every case, but that doesn't make it worthless. I think users would appreciate it, but you folks will of course have to decide if it's worth the added work.

twardnw:
Sorry @MikeMcQuaid, wasn't wanting to seem impatient, the indication in #7667 was that all this needed was passing tests to get back in, I was just adding feedback that we had also been relying on it.

Since it's not going to be moved back in, what's the best way of installing it as a dependency? Should we add it to our tap, or just list `boneyard/unfs3` as the dependency in the Formula?

woodruffw:
@twardnw Adding it to your tap is probably best even if the boneyarded version currently works, since you may need to patch it in the future.

twardnw:
thanks @woodruffw

woodruffw:
Closing this, thanks to all for input.

PR #7831, opened 2016-12-13
libhttpseverywhere 0.2.9 (new formula)

tschoonj:


Formula/libhttpseverywhere.rb
```diff
@@ -0,0 +1,76 @@
...
+    (testpath/"test.c").write <<-EOS.undent
+      #include <httpseverywhere.h>
+
+      int main(int argc, char *argv[]) {
+        GType type = https_everywhere_context_get_type();
+        return 0;
+      }
+    EOS
+    ENV.libxml2
+    gettext = Formula["gettext"]
```
tschoonj: I don't think that would be useful as I cannot loop over them to get the include and link flags.

woodruffw: I was going to suggest that you could with a `map`, but I didn't notice that a few of the include flags had extra directories on the end. My bad!

PR #7849, opened 2016-12-14
artifactory 4.15.0

yogeva:


woodruffw:
Thanks for contributing, @yogeva! For future reference, we prefer that the package name be lowercased in the commit (e.g., `artifactory 4.15.10`).

PR #7860, opened 2016-12-14
android-sdk 25.2.4

tahnok:

-----

Android has an updated set of build tools <https://developer.android.com/studio/releases/sdk-tools.html>, so time to bump the formula

The layout of the archive has changed somewhat. Previously the `install` method would execute so that there was a `tools` folder to install, in addition to a README and an ignored `add-ons` and `platforms`.

The new structure just contains a `tools` folder which brew is inside of during `install`. Moving up a directory and installing the whole folder seems like the simplest approach, but I could also write some ruby to build up a hash of the files in the current directory to `tools/` and pass that to `prefix`

There is also a new `sdkmanager` tool that it's a new location that needs to be added to `bin`

### DIFF
--- a/Formula/android-sdk.rb
+++ b/Formula/android-sdk.rb
@@ -3,10 +3,9 @@ require "base64"
 class AndroidSdk < Formula
   desc "Android API libraries and developer tools"
   homepage "https://developer.android.com/index.html"
-  url "https://dl.google.com/android/android-sdk_r24.4.1-macosx.zip"
-  version "24.4.1"
-  sha256 "ce1638cb48526a0e55857fc46b57eda4349e6512006244ad13dd6c1361c74104"
-  revision 1
+  url "https://dl.google.com/android/repository/tools_r25.2.4-macosx.zip"
+  version "25.2.4"
+  sha256 "399f107a905c367bd23b64bf34d4c0313d63ba6292376992021a93cbce6e9d89"
 
   bottle do
     cellar :any
@@ -31,11 +30,43 @@ class AndroidSdk < Formula
 
   # Version of the android-build-tools the wrapper scripts reference.
   def build_tools_version
-    "23.0.1"
+    "25.0.1"
   end
 
   def install
-    prefix.install "tools", "SDK Readme.txt" => "README"
+    prefix.install_metafiles
+
+    (prefix/"tools").install [
+      buildpath/"android",
+      buildpath/"ant",
+      buildpath/"apps",
+      buildpath/"bin",
+      buildpath/"bin64",
+      buildpath/"ddms",
+      buildpath/"draw9patch",
+      buildpath/"emulator",
+      buildpath/"emulator-check",
+      buildpath/"emulator64-arm",
+      buildpath/"emulator64-crash-service",
+      buildpath/"emulator64-mips",
+      buildpath/"emulator64-x86",
+      buildpath/"hierarchyviewer",
+      buildpath/"jobb",
+      buildpath/"lib",
+      buildpath/"lib64",
+      buildpath/"lint",
+      buildpath/"mksdcard",
+      buildpath/"monitor",
+      buildpath/"monkeyrunner",
+      buildpath/"proguard",
+      buildpath/"qemu",
+      buildpath/"screenshot2",
+      buildpath/"source.properties",
+      buildpath/"support",
+      buildpath/"templates",
+      buildpath/"traceview",
+      buildpath/"uiautomatorviewer",
+    ]
 
     %w[android ddms draw9patch emulator
        emulator-arm emulator-x86 hierarchyviewer lint mksdcard
@@ -47,13 +78,17 @@ class AndroidSdk < Formula
       EOS
     end
 
-    %w[zipalign].each do |tool|
-      (bin/tool).write <<-EOS.undent
+    (bin/"zipalign").write <<-EOS.undent
         #!/bin/bash
-        TOOL="#{prefix}/build-tools/#{build_tools_version}/#{tool}"
+        TOOL="#{prefix}/build-tools/#{build_tools_version}/zipalign"
         exec "$TOOL" "$@"
-      EOS
-    end
+    EOS
+
+    (bin/"sdkmanager").write <<-EOS.undent
+        #!/bin/bash
+        TOOL="#{prefix}/tools/bin/sdkmanager"
+        exec "$TOOL" "$@"
+    EOS
 
     %w[dmtracedump etc1tool hprof-conv].each do |tool|
       (bin/tool).write <<-EOS.undent
@@ -103,7 +138,7 @@ class AndroidSdk < Formula
     end
 
     # automatically install platform and build tools
-    system "echo y | bash #{bin}/android --verbose update sdk --no-ui --all --filter platform-tools,build-tools-#{build_tools_version}"
+    system "echo y | ANDROID_SDK_HOME=$HOME bash #{bin}/android --verbose update sdk --no-ui --all --filter platform-tools,build-tools-#{build_tools_version}"
   end
 
   def caveats; <<-EOS.undent
@@ -117,6 +152,7 @@ class AndroidSdk < Formula
   end
 
   test do
-    assert_match version.to_s, shell_output("#{prefix}/tools/emulator -version")
+    assert_match "Android Debug Bridge", shell_output("#{bin}/adb version")
+    system "#{bin}/emulator", "-list-avds"
   end
 end


woodruffw:
Audit failure:

```
android-sdk:
    Broken dylib links found:
      /tmp/darwin-x86_64-clang-3.5/lib/libc++.1.dylib
        libQt5Widgets.5.dylib
        libQt5Gui.5.dylib
        libQt5Core.5.dylib
        libQt5Network.5.dylib
        libQt5Svg.5.dylib
        libQt5DBus.5.dylib
        libQt5PrintSupport.5.dylib
        libQt5Sql.5.dylib
```

Test failure:

```
==> /usr/local/Cellar/android-sdk/25.2.3/tools/emulator -version
Error: android-sdk: failed
<0> expected but was
<1>.
```

tahnok:
Hrrm, I don't think that test works at all. If i re-install 24.4.1 and run `brew test android-sdk` I get the same failure... I will remove it and replace it with something that works

woodruffw:
Thanks for fixing the test.

I'm not sure what's causing those broken dylib linkages in `audit`. Best guess is missing runtime dependencies (`qt5`)...

tahnok:
It seems like it's broken in the archive from google. They seem to have just released a 25.2.4 so I will see if that fixes this

woodruffw:
Sounds good. Feel free to amend this PR for the new version.

tahnok:
No dice in updating to 25.2.4, although I did notice a lot of similar message to this in the install logs:

```
Warning: Could not fix /tmp/darwin-x86_64-clang-3.5/lib/libc++.1.dylib in /usr/local/Cellar/android-sdk/25.2.4/tools/emulator
```

(full log in this gist: https://gist.github.com/tahnok/abbc604a2165ee521f9288a09172bc1f)

I'm wayyyy out of my depth here, but is brew failing to patch these properly or something? Would it be better to leave them as is so at least they are consistent?

woodruffw:
Yeah, it looks like the relocation code is failing under an edge case (probably something to do with `libc++` being in `/tmp`, for whatever reason).

I'll look into this.

woodruffw:
I'm not sure where `/tmp/darwin-x86_64-clang-3.5/lib/libc++.1.dylib` is coming from, but my best guess is that the build is dropping it in `/tmp` as a hack. This screws with our relocation process, as we expect shared libraries and other relocatables to only be in a few reasonable places (under the `prefix`, in the cellar, under `lib`, etc.) where we know they'll be permanently.

I don't want to say it's for certain that it's an upstream problem *quite* yet, since there could be something I'm missing, but I *think* that it is.

woodruffw:
According to my Mach-O dumper, these are the only dylibs that show up in `libQt5Core.5.dylib`:

```
DYNAMIC LIBRARIES:
  /System/Library/Frameworks/DiskArbitration.framework/Versions/A/DiskArbitration
  /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit
  /usr/lib/libz.1.dylib
  /usr/lib/libSystem.B.dylib
  /System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices
  /usr/lib/libstdc++.6.dylib
  /usr/lib/libobjc.A.dylib
```

So the mystery deepens.

woodruffw:
Nope, there it is, in the `emulator` executable:

```
DYNAMIC LIBRARIES:
  /tmp/darwin-x86_64-clang-3.5/lib/libc++.1.dylib
  /usr/lib/libSystem.B.dylib
  /System/Library/Frameworks/AudioUnit.framework/Versions/A/AudioUnit
  /System/Library/Frameworks/AVFoundation.framework/Versions/A/AVFoundation
  /System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa
  /System/Library/Frameworks/CoreAudio.framework/Versions/A/CoreAudio
  /System/Library/Frameworks/CoreMedia.framework/Versions/A/CoreMedia
  /System/Library/Frameworks/CoreVideo.framework/Versions/A/CoreVideo
  /System/Library/Frameworks/ForceFeedback.framework/Versions/A/ForceFeedback
  /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit
  /System/Library/Frameworks/QTKit.framework/Versions/A/QTKit
  /System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices
```

So upstream has done something hacky in the build system, probably to avoid a change in clang's libc++.

ilovezfs:
I still wouldn't rule out it's us.

woodruffw:
Well, here's the old (24.4.1) `emulator` dylibs:

```
DYNAMIC LIBRARIES:
  /usr/lib/libstdc++.6.dylib
  /usr/lib/libSystem.B.dylib
  /System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa
  /System/Library/Frameworks/QTKit.framework/Versions/A/QTKit
  /System/Library/Frameworks/CoreVideo.framework/Versions/A/CoreVideo
  /System/Library/Frameworks/AVFoundation.framework/Versions/A/AVFoundation
  /System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices
```

It looks like libc++ changed out from under them, and so they just copied the old one to `/tmp` and continued the build from there. Unless I'm misunderstanding how we do relocation, that's not a situation that we can cleanly handle. It's also not a good practice by any measure (they should have just included their version of libc++ in `lib64`).

ilovezfs:
stupid question: why are we installing precompiled binaries?

woodruffw:
I was wondering the same thing...

ilovezfs:
@woodruffw let's pretend that's acceptable _arguendo_. You could use your fancy macho library to delete those right? :)

woodruffw:
Yeah, definitely. I'm just not sure what it would do when executed :thinking:

ilovezfs:
@woodruffw I suspect they're not needed. For example, this ugliness didn't break anything: https://github.com/Homebrew/homebrew-core/blob/master/Formula/haskell-stack.rb#L53-L64

woodruffw:
Cool. I'll come up with a snippet that @tahnok can add to this that should "fix" things.

ilovezfs:
Interesting: 
```
bash-3.2$ find . |grep dylib|while read f; do otool -L $f|grep -q clang-3.5 && echo $f; done
./lib64/lib64EGL_translator.dylib
./lib64/lib64GLES_CM_translator.dylib
./lib64/lib64GLES_V2_translator.dylib
./lib64/lib64OpenglRender.dylib
./lib64/lib64emugl_test_shared_library.dylib
```

There are more broken ones though right?

ilovezfs:
Oh I see some are actually programs not dylibs. yikes

woodruffw:
I *think* those ones are being linked in correctly - `keg_relocate` should be smart enough to find anything referenced in `lib` or `lib64`.

tahnok:
I modified the audit command to print files that have broken dylibs

```
/usr/local/Cellar/android-sdk/25.2.4/tools/emulator
/usr/local/Cellar/android-sdk/25.2.4/tools/emulator-check
/usr/local/Cellar/android-sdk/25.2.4/tools/emulator64-arm
/usr/local/Cellar/android-sdk/25.2.4/tools/emulator64-crash-service
/usr/local/Cellar/android-sdk/25.2.4/tools/emulator64-mips
/usr/local/Cellar/android-sdk/25.2.4/tools/emulator64-x86
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/lib64EGL_translator.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/lib64GLES_CM_translator.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/lib64GLES_V2_translator.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/lib64OpenglRender.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/lib64emugl_test_shared_library.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/bearer/libqcorewlanbearer.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/bearer/libqgenericbearer.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/generic/libqtuiotouchplugin.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/iconengines/libqsvgicon.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/imageformats/libqgif.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/imageformats/libqico.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/imageformats/libqjpeg.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/imageformats/libqsvg.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/platforms/libqcocoa.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/platforms/libqminimal.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/platforms/libqoffscreen.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/printsupport/libcocoaprintersupport.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/lib64/qt/plugins/sqldrivers/libqsqlite.dylib
/usr/local/Cellar/android-sdk/25.2.4/tools/qemu/darwin-x86_64/qemu-system-aarch64
/usr/local/Cellar/android-sdk/25.2.4/tools/qemu/darwin-x86_64/qemu-system-armel
/usr/local/Cellar/android-sdk/25.2.4/tools/qemu/darwin-x86_64/qemu-system-i386
/usr/local/Cellar/android-sdk/25.2.4/tools/qemu/darwin-x86_64/qemu-system-mips64el
/usr/local/Cellar/android-sdk/25.2.4/tools/qemu/darwin-x86_64/qemu-system-mipsel
/usr/local/Cellar/android-sdk/25.2.4/tools/qemu/darwin-x86_64/qemu-system-x86_64
```

woodruffw:
They really butchered this release...

We could delete those broken install names entirely (might explode on dynamic load) or replace them with `/usr/lib/libstdc++.6.dylib` (might explode on first ABI change). I'm not sure which one is really better here...

woodruffw:
@tahnok We're considering just whitelisting this from `audit` because the bad linkage doesn't seem to actually cause crashes. Could you confirm that the executables actually run?

tahnok:
Yup, I can run every tool that is linked and successfully launch the emulator

Formula/android-sdk.rb
```diff
@@ -31,11 +30,12 @@ class AndroidSdk < Formula
...
   # Version of the android-build-tools the wrapper scripts reference.
   def build_tools_version
-    "23.0.1"
+    "25.0.1"
   end
 
   def install
-    prefix.install "tools", "SDK Readme.txt" => "README"
+    Dir.chdir "../"
+    prefix.install "tools"
```
ilovezfs: instead of changing directories (by the way `cd "foo"` works, no need for `Dir.chdir`), probably should do this:
```
(prefix/"tools").install buildpath.children - [buildpath/".brew_home"]
```

Formula/android-sdk.rb
```diff
@@ -55,6 +55,14 @@ def install
       EOS
     end
 
+    %w[sdkmanager].each do |tool|
```
ilovezfs: there's only one, so you can kill the outer loop and do `(bin/"sdkmanager").write`

woodruffw:
Some of the binaries seem to also have been renamed/deleted since the last release, so it may be worth going over them quickly to make sure we're not missing any when writing the shell wrappers.

Formula/android-sdk.rb
```diff
@@ -117,6 +125,7 @@ def caveats; <<-EOS.undent
   end
 
   test do
-    assert_match version.to_s, shell_output("#{prefix}/tools/emulator -version")
+    system bin/"adb version"
```
ilovezfs: this is actually spawning another shell. should be
```
system bin/"adb", "version"
```
or
```
system "#{bin}/adb", "version"
```
Ditto for the one below.

Formula/android-sdk.rb
```diff
@@ -117,6 +125,7 @@ def caveats; <<-EOS.undent
   end
 
   test do
-    assert_match version.to_s, shell_output("#{prefix}/tools/emulator -version")
```
ilovezfs: why can't we still match?

tahnok: `emulator -version` isn't a recognized command. I don't think it was working in 24 either though

Formula/android-sdk.rb
tahnok: I thought there was a way to do this that was less hacky, thanks

Formula/android-sdk.rb
ilovezfs: OK, but presumably `adb version` is spitting out something match-worthy, no?

tahnok: It's downloaded separately as part of the `build-tools` so it's quite possible that it will change at some point. Google has a bad habit of doing that...

It currently outputs
```
Android Debug Bridge version 1.0.36
Revision 9f299f282ec9-android
```
Should I check that it says "Android Debug Bridge" in the output?

tahnok:
According to the changelog `sdkmanager` is the only new tool. All the other tools that get linked work for me.

Should I squash these commits down to a single one?

Formula/android-sdk.rb
```diff
@@ -31,11 +30,11 @@ class AndroidSdk < Formula
 
   # Version of the android-build-tools the wrapper scripts reference.
   def build_tools_version
-    "23.0.1"
+    "25.0.1"
   end
 
   def install
-    prefix.install "tools", "SDK Readme.txt" => "README"
+    (prefix/"tools").install buildpath.children - [buildpath/".brew_home"]
```
tahnok: Nope, it's gone

ilovezfs: looks like there's a NOTICE.txt instead, so can do
```
prefix.install_metafiles
(prefix/"tools").install buildpath.children - [buildpath/".brew_home"]
```

Formula/android-sdk.rb
```diff
@@ -31,11 +30,12 @@ class AndroidSdk < Formula
...
   # Version of the android-build-tools the wrapper scripts reference.
   def build_tools_version
-    "23.0.1"
+    "25.0.1"
   end
 
   def install
-    prefix.install "tools", "SDK Readme.txt" => "README"
+    prefix.install_metafiles
+    (prefix/"tools").install buildpath.children - [buildpath/".brew_home"]
```
MikeMcQuaid: Let's specify these specifically rather than dumping everything in there.

tahnok:
Is there a way to see the output of `install`? I think I know what the problem is, but I want to double check.

It seems like sometimes the android tool can't fetch one of the XML files and so the `android sdk install` command fails. I'm not entirely sure how that should be fixed. Is there a precedent somewhere in brew?

woodruffw:
You can use the `--debug` flag with `install`, which will give you an option to drop either into a shell in the build tree or an `irb` session for debugging.

tahnok:
Yeah, I've been using that but I was wondering if I could see the logs from jenkins for the process. I suspect the issue is a flaky remote server. One of the solutions that people use is `--no-https` flag but that feels super 🤢  (see my latest commit)

tahnok:
I'm super stumped.

```
 bash /usr/local/Cellar/android-sdk/25.2.4/bin/android --verbose update sdk --no-ui --no-https --all --filter platform-tools,build-tools-25.0.1
```

reliably fails to do what it is supposed to when run during the brew install process. But since it's the last command the install process will complete successfully. Once done, I can successfully run the command.

Is there come kind of sandbox the install process is run inside of? I dropped into a shell during the install process (by introducing a fake error and choosing the shell option) and inside of that shell the command will reliably fail.

Not really sure how to fix at this point...

woodruffw:
> Is there come kind of sandbox the install process is run inside of? I dropped into a shell during the install process (by introducing a fake error and choosing the shell option) and inside of that shell the command will reliably fail.

Yeah, the install is sandboxed. If this is indeed a sandbox-induced failure, then I'm not sure what to do :worried:. 

Any output from that failing command?

ilovezfs:
well there's always `--no-sandbox` ...

tahnok:
What does the sandbox prevent?

The command uses the network to fetch a few xml files, which it caches in `~/.android/cache` and then it downloads some files to `$ANDROID_HOME` which I assume is part of the bottle?

How does `--no-sandbox` work?

tahnok:
Yeah, I checked the syslog while installing and I noticed

```
kernel: (Sandbox) SandboxViolation: java(38039) deny(1) file-write-create /Users/wes/.android/cache
```

Not sure what to do at this point to get this merged. Can I whitelist writing to that file for this formula?

MikeMcQuaid:
> Can I whitelist writing to that file for this formula?

No, sorry. We'll need to figure out how to coax it into writing into `$HOME` which is set to under the `buildpath`.

tahnok:
I've filed an issue upstream: https://code.google.com/p/android/issues/detail?id=231993&thanks=231993&ts=1484168383

tahnok:
Aha! I think I figured it out. The tool will respect `ANDROID_SDK_HOME`, so I just need to set that to $HOME and it _should_ work in the sandbox

tahnok:
More dylib failures... Were new checks added?

MikeMcQuaid:
@tahnok Nope.

tahnok:
Ok, so the failure is with `/usr/local/Cellar/android-sdk/25.2.4/build-tools/25.0.1/lib64/libbcinfo.dylib` which needs to be relinked? I don't think that's possible since it's precompiled in the tarball

Again, I find myself out of my depth. Can this issue be worked around? Or is this formula not suitable for brew?

MikeMcQuaid:
We can ignore dylib checks for this formula.

MikeMcQuaid:
Was that the last problem?

MikeMcQuaid:
@BrewTestBot test this please

tahnok:
Yes, dylib problems where the only remaining problems. Obviously the commit history is pretty messy, but otherwise this works and the tests in the formula itself pass.

tahnok:
@MikeMcQuaid  Should I rebase and squash these commits?

MikeMcQuaid:
@tahnok Yes, please!

tahnok:
Updated

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
Non-blocking issue: https://bot.brew.sh/job/Homebrew%20Core/16198/version=sierra/testReport/junit/brew-test-bot/sierra/audit_android_sdk___online/

Blocking issue: https://bot.brew.sh/job/Homebrew%20Core/16198/version=sierra/testReport/junit/brew-test-bot/sierra/bottle_android_sdk/

tahnok:
I don't know if that blocking issue can be fixed. The binaries are coming from google and I have no idea how to go about patching that

for the context of others, here is the error:

```
Error: Updated load commands do not fit in the header of /usr/local/Cellar/android-sdk/25.2.4/build-tools/25.0.1/lib64/libbcinfo.dylib. /usr/local/Cellar/android-sdk/25.2.4/build-tools/25.0.1/lib64/libbcinfo.dylib needs to be relinked, possibly with -headerpad or -headerpad_max_install_names
```

woodruffw:
Oh, boy. That error means that they need to build their binaries with `-headerpad`. I'm not even sure how they managed to break them that badly - it's extremely rare for load commands to actually run up against the header padding boundary.

MikeMcQuaid:
I think we need to figure out how or if we can build this from source and also ask if this is open source at all. At this point Homebrew Cask is looking like a much better delivery mechanism.

tahnok:
I've done some research and I don't think compiling all these tools from source is going to be feasible. The code seems to be scattered all over the place, and they seem to be built with non-standard tools. Eg: here's the source repo for just the platform tools: https://github.com/android/platform_system_core which seems to be missing any kind of makefile.

I'm not super up to date on the brew ecosystems, but if casks are designed for pre-compiled binaries than yes, I would agree that's a better place for this to live.

MikeMcQuaid:
> I'm not super up to date on the brew ecosystems, but if casks are designed for pre-compiled binaries than yes, I would agree that's a better place for this to live.

Yeh, that seems a better bet given what you've been saying. Is that something you can work on? We'd be happy to help you.

tahnok:
Sure, I can look into that. Going to close this for now since it seems unmergable

PR #7898, opened 2016-12-15
Added caveats to support usage with zsh.

lylemoffitt:
Zsh provides a similar utility with the builtin `times`, which aliased as the reserved word `time`. If `gnu-time` is compiled `--with-default-names`, then it will not be visible in PATH, because it will be overshadowed by Zsh's reserved word. This PR amends the formula with a `caveats` section to notify the user of this and recommends that they modify their `.zshc` with a (provided) command that will disable the reserved word. 

-----

### DIFF
--- a/Formula/gnu-time.rb
+++ b/Formula/gnu-time.rb
@@ -35,6 +35,18 @@ class GnuTime < Formula
     system "make", "install"
   end
 
+  def caveats
+    <<-EOS.undent
+    If you compiled "with-default-names" and you use zsh, then
+    this command will be overshadowed by a built-in. To fix this,
+    add the following to your .zshrc:
+
+        ## Disable builtin `time` command so that gnu-time may be used instead
+        if ! [ -x "$(which time)" ] ;then disable -r time ; fi
+
+    EOS
+  end
+
   test do
     system bin/"gtime", "ruby", "--version"
   end


zmwangx:
1. `time` is not a builtin, it's a reserved word used to form a complex command. Meanwhile, `times` is a very different builtin defined by POSIX (called a special built-in utility by POSIX, see [doc](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_27)), which doesn't take any argument and prints the accumulated times in the shell.

2. There's a POSIX utility `time` (see [doc](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/time.html)), fulfilled by `/usr/bin/time` on macOS. `gnu-time` duplicates this role, just like `coreutils` duplicates most other POSIX utilities.

3. `time` as a shell keyword/reserved word (different jargon for different shells) is very widespread, because timing implemented in the shell is more accurate and versatile. For instance, `time` is a reserved word in `zsh` and `mksh`, and a keyword in `bash` and `ksh93`. It's certainly not limited to `zsh`, and `time` as a shell reserved word masking the external utility `time(1)` is extremely common. If you really want to use the `time` executable (not sure why), probably the "correct" way is `command time` (you can alias `time` to `command time` if you want to), as opposed to disabling a reserved word.

In summary, reserved word `time` masking external `time(1)` is a common in most modern interactive shell usage; certainly not Homebrew specific. It doesn't deserve a mention in caveats.

woodruffw:
I agree with @zmwangx here. This kind of masking behavior is to be expected when installing programs that share names with builtins and core utilities. It's not unique to OS X, Homebrew, or zsh.

We appreciate the contribution, but we can't merge this. Sorry!

lylemoffitt:
As a user I would have preferred a heads up from the package that this behavior was to be expected. I know this might be obvious to someone with experience, but it's abrasive to people who haven't encountered this kind of thing. Would a change in the language of the `caveats` be appropriate? Is there a `notes` section that could be added to the formula instead?

woodruffw:
@lylemoffitt My apologies if I came off as abrasive :bowing_man: 

We mostly use caveats for knowledge that's central to the package's behavior with respect to Homebrew and/or OS X. Since (GNU) `date` masks the builtin `date` on any *nix it's installed on, we would be loosening the standard to include more general knowledge. This, in turn, imposes a greater maintenance burden and tasks us with determining which knowledge is obscure enough to warrant inclusion (something nobody agrees on!).

> Is there a notes section that could be added to the formula instead?

As far as I know, there isn't anything like that currently. I encourage you to raise the idea on [brew](https://github.com/Homebrew/brew/), although changes to the formula DSL are pretty rare (since they're more or less permanent once introduced).

lylemoffitt:
Oh. No worries! I was saying that software behaviors/features/entailments that are undocumented because "most people" would already know.... is abrasive. I understand that it's difficult to find a balance between having to explain too much (which burdens the dev) and explaining too little (which can burden the user). 

I can understand wanting to have a pretty strict lower (more stringent) bound for `caveats`, based on "need to know" as determined by whomever... but why limit the upper (more lax) bound? The only way this stuff gets expanded is if someone volunteers their time anyway. As maintainers I can see your concerns, but why can't the only criteria be: (1) if it meets the lower bound and (2) if it's accurate? Maybe I'm ignorant (I've never maintained a FOSS project), but that doesn't seem unreasonable.

----

The problem (aside from dev/maintenance stuff) with creating an extra `notes` section is the overlap. If you task it as just extra info, then it overlaps with the expected contents of the homepage/readme (which is assumed to cover the basics). If you task it as more important than that, then you end up bumping into the `caveats`.

ilovezfs:
Caveats are for information _specific to Homebrew_.

zmwangx:
Run

```zsh
for formula in "$(brew --repo homebrew/core)"/Formula/*.rb(:r:t); do caveats="$(brew info $formula | sed '1,/==> Caveats/d')"; [[ -n $caveats ]] && { print $'\e[32m'$formula$'\e[0m'; print -R $caveats; print }; done
```

in Zsh to get a feel for what actually belongs to caveats.

lylemoffitt:
Thank you @zmwangx, I was just about to use something like that.

I've been a homebrew user for a while (even though, this is my first PR). The most common thing I see caveats used for is extra install/usage instructions. The relation to homebrew usually crops up as: 
1) Something homebrew doesn't want to do itself (like automatically edit your `.*rc` files for you or modifying stuff outside homebrew's purview). 
2) Providing the user with path/config information that is derived as a result of where/how homebrew installed it.

This is limited scope is good for maintainers, but bad for users. It leaves them high-and-dry if there is anything else they might need to know to finish the installation. It also leaves them in the dark when upgrading as they don't have release notes or other warning info when upgrading/installing. As of right now, if they want that they need to go to the homepage. But the problem there is that the home page is often out of date or missing the necessary info altogether. 

---
tl;dr - If the scope of caveats is so narrowly defined than you really should have a notes section (or something).

zmwangx:
Caveats were never meant to replace or extend official docs (which is unfeasible to begin with; it's hard enough to maintain up-to-date documentation, not to mention *other people's* documentation), or be a getting started guide, or worse still, be a command line crash course.

> release notes

This has been discussed before, but the conclusion was it's pretty much impossible for us to do, because even when they're available, they come in all shapes and forms and styles and levels of verbosity.

> if they want that they need to go to the homepage.

They should.

Based on experience I don't think many people have complained about the lack of caveats.

ilovezfs:
For what it's worth, our thrust of late has been to attempt to remove as many caveats as possible, so a notes field is most likely DOA as a proposal.

MikeMcQuaid:
I agree with @zmwangx @ilovezfs and @woodruffw here. We have to limit Homebrew's scope (due to limited time and resources) to installing software. If you aren't sure what to do once that software has been installed or don't e.g. know enough about Unix yet then I definitely agree it'd be great if we could teach you that but I think that's really the responsibility of other projects. Sorry!

lylemoffitt:
@MikeMcQuaid 
> We have to limit Homebrew's scope

I *totally* get this. I'm just wishing there was someway to be more helpful to the user. Not necessarily educate, just give them a heads-up.

@zmwangx 
> This has been discussed before [...] it's pretty much impossible for us to do

I'm not suggesting **full** release notes. Just notes pertinent to the formula install. For example, change in formula dependency/conflicts or installation options; these might not necessarily be on the homepage, and they don't really belong in the caveats either.

MikeMcQuaid:
@lylemoffitt The probably with adding more notes is: the more we ask people to read, the less likely they are to read any of it (as time has repeatedly shown), unfortunately.

lylemoffitt:
I'm inclined to agree with you. It's a generally accepted truth that people don't read the release notes (or even the documentation) for any project until they need to. The problem comes when they need to and it's not there. Being able to pull up the revision notes for a formula you installed recently could be really helpful in figuring out why you suddenly have a problem (seemingly out of nowhere). 

I think part of the issue is that we're already showing the users more than they need. Detailing the step-by-step install process for every package isn't really necessary every time you upgrade, and neither is showing the caveats. The former should show up when there's a problem, and the latter is needed the first time you install. Really the only things that should be printed when you upgrade is what's changed and what failed (and why). But, maybe that's just me; maybe it should be configurable? I don't know, I'm just trying to advocate for a better UX with more/better tools.

woodruffw:
Just an idea: If revision/release notes are something you *really* want Homebrew to provide, you can always write your own `brew notes` subcommand. Reading from CHANGELOGs and RELEASE files would be a good start.

lylemoffitt:
@woodruffw For right now I'm more trying to see how much traction the idea has. I also don't have a clear idea yet of what would be most helpful. So I'm trying to cultivate some helpful feedback on what people feel is/isn't missing. 

> Reading from CHANGELOGs and RELEASE files would be a good start.

I like that idea in principle. I actually have no idea how reliable those files are in terms of formatting or helpfulness. From a practical standpoint, simply improving the visibility of things like that might be the best short term approach. I had been considering ways to try and make `brew-log <formula>` more contextually helpful (when shown during `brew upgrade`) and less arcare for the average user.

woodruffw:
Yeah, you're not going to get any sort of uniform formatting on changelogs or release notes. I think the best approach, if you were to tackle those, is just to dump them (shortened?).

I wouldn't be surprised if this is a feature that some users want, but it's unlikely to gain traction officially, for the reasons mentioned by the others (maintenance, relatively low use, etc).

lylemoffitt:
> dump them (shortened?)

Hmm.. I'd have to do a survey and see if any general trends are visible. They might fall into common patterns that could be parsed/formatted. The real problem is incorporating them into the upgrade process in a coherent (and helpful) way. This would involve more than just making a new command, I'd have to alter the upgrade code itself.

> it's unlikely to gain traction officially

That brings me to a side-question... What's the process like for progressive feature addition in homebrew? For example, `brew-desc` was a separate tap for a long time. Then one day is was a built-in. How did that happen? There doesn't seem to be any cultivation or curation of beta-features in homebrew. As a user I have to go out of my way to find them. They only become part of homebrew main if enough people have also managed to find that tap, and then also advocate for its addition.

MikeMcQuaid:
> That brings me to a side-question... What's the process like for progressive feature addition in homebrew? For example, brew-desc was a separate tap for a long time. Then one day is was a built-in. How did that happen? There doesn't seem to be any cultivation or curation of beta-features in homebrew. As a user I have to go out of my way to find them. They only become part of homebrew main if enough people have also managed to find that tap, and then also advocate for its addition.

We added it when it had complete coverage (which took a few years). At that point it wasn't `brew desc` folks advocating for inclusion but Homebrew maintainers (i.e. me) who helped to get it included. This is better because it means when the feature is added to Homebrew itself it's useful and works consistently.

lylemoffitt:
That's great and all, but what about the other extensions? How do they make progress towards that level of inclusion? Just about the only thing they can do to gain exposure is be put on the [list](https://github.com/Homebrew/brew/blob/master/docs/External-Commands.md) (which never seems to get updated). Beyond that you're relying on people actively searching github for such a command, which proves more difficult than it should be.

---
Bottom line: Why isn't there a `homebrew/homebrew-extern` repo for external commands that people can contribute to?

ilovezfs:
@lylemoffitt that sounds like an interesting idea.

MikeMcQuaid:
> Bottom line: Why isn't there a homebrew/homebrew-extern repo for external commands that people can contribute to?

We did this in the past and what ends up happening is they become unmaintained by their original authors but widely used enough to get us bug reports but not enough such that they get fixed.

There's nothing stopping you creating your own tap and promoting it yourself. What you seem to want, though, is for Homebrew to maintain and promote your idea. As said, we may consider that in future but you'd need to implement it first for the majority of formulae for it to be useful. What's not reasonable is the expectation (explicitly or implicitly) that the onus is on us to implement, support and/or promote this.

lylemoffitt:
> What you seem to want, though, is for Homebrew to maintain and promote your idea.

[TANSTAAFL](https://en.wikipedia.org/wiki/There_ain't_no_such_thing_as_a_free_lunch). I am merely trying to understand why things are the way they are, so that I can try and help improve things. I don't want to repeat a mistake that has already been learned from. I also don't want to burden Homebrew with anything that isn't worth its upkeep.

> We did this in the past [...]

This is the kind of thing I really want to know. It sounds like what's needed is a layer of indirection that keeps bug reports with the original authors, but retains ease of access that comes with a centralized repo.

MikeMcQuaid:
> This is the kind of thing I really want to know. It sounds like what's needed is a layer of indirection that keeps bug reports with the original authors, but retains ease of access that comes with a centralized repo.

That's pretty fundamentally not possible, unfortunately.

ilovezfs:
For what it's worth, attempting to google for third-party Homebrew external commands is not exactly straightforward and returns a lot less than is actually out there without going through search engine gymnastics to exclude homebrew organization results and 3rd party formulae results and just ordinary forks people were using for PRs.

lylemoffitt:
@ilovezfs Glad I'm not the only one who feels this. 

---
@MikeMcQuaid 
> That's pretty fundamentally not possible, unfortunately.

Why? Please explain. I know that because a lot of ext cmds are done in tap form, that makes it impossible to do in a regular formula repo, but I thought a repo of mini `Brewfile`s (like those made and used by `brew bundle` could get solve that. But maybe this isn't about the impl., maybe it's about the culture? What did you have in mind?

MikeMcQuaid:
@lylemoffitt It's impossible to keep bug reports with original authors while maintaining a centralised repo. Some of the benefits/drawbacks of open-source:
- You can't force anyone to work on anything
- You can't have maintainers for something they don't have commit access to
- Maintainers end up maintaining anything committed to their repository (for good or ill)

In the case of `brew desc` (and this command): it's not useful until it gets adequate coverage. If we put it in the main repository the onus for getting that coverage is on the maintainers and there's an implicit signalling that the command is "ready" even when it is not. When it's in a tap you (or whoever) maintains then getting it to "ready" is on you and whoever else you can convince to work on it and we can consider including it when it's popular, complete and useful enough.

Other projects are run in different ways and I'm sure there's strengths and weaknesses to various approaches but this is how we're doing it in Homebrew. Note that even commands like e.g. `brew bundle` and `brew services` that could have lived in this repository are kept in separate, official taps which ends up easing development work and contribution. It's worth looking at how "ready" each of these were (and Homebrew Cask) before they were made official. Homebrew has been around for a long time at this point so the bar is pretty high for such things, unfortunately.

MikeMcQuaid:
> For what it's worth, attempting to google for third-party Homebrew external commands is not exactly straightforward and returns a lot less than is actually out there without going through search engine gymnastics to exclude homebrew organization results and 3rd party formulae results and just ordinary forks people were using for PRs.

I agree here. I don't think including 3rd-party contributions in Homebrew/brew is the way forward though for the various reasons listed above. Discoverability of 3rd-party taps is an interesting problem that someone could make better but it's not something that I personally see as as high priority as many other things (i.e. I won't work on it but someone else could/should).

lylemoffitt:
> It's impossible [...]

I totally agree with and understand all three points here, but I still don't see why it's impossible. Maybe i'm just too caught up in the impl I have in my head? I'm thinking that you could get that kind of re-direction by having  a repo of simplified (and constrained) shell scripts, following the `brew bundle` model (i.e. all commands are prepended with `brew`). For example the package for an extcmd like `brew-growl` from the [list](https://github.com/Homebrew/brew/blob/master/docs/External-Commands.md) could be represented with a simple YAML like this:

```yaml
name:   brew-growl
desc:   Get Growl notifications for Homebrew
home:   https://github.com/secondplanet/homebrew-growl
install:    # Prepend each line with 'brew' before executing
    - tap secondplanet/growl
    - install brew-growl
uninstall:  # Same as above
    - uninstall brew-growl
    - untap secondplanet/growl
test:       # Each must: [$?==0] after 'install' & [$?!=0] after 'uninstall'
    - command growl
    - growl on
```
Granted, even this example has problems, b/c `brew-growl.rb` doesn't list its dependency on `growlnotify`. However, I think it also get's a lot of stuff right:
1) The problem caused by missing requirements would be caught by at least one of the tests.
2) Our package only stores what commands to execute, which leaves the burden of maintenance on the original dev.
3) Because the commands are universal and generic, we have very loose coupling with the formula/tap, which minimizes maintenance burden when the tap/formula gets updated, modified, or even changes repos.
4) Using `brew` commands gives us a basic guarantee against arbitrary shell commands and dependencies.
5) Having tests allows us to verify installs, detect breaking changes, and collect (and report) metrics about whether or not an extcmd "works" for a majority of users. This could potentially be reported to the maintainer.
6) Storing a GitHub homepage means we can relate metrics to the user and indicate the health/volatility of the extension.
7) It's a simple and popular format with a very low barrier to entry, which should make it as easy as possible for developers to add and maintain their associated extcmds package.

---
All in all, I think this goes a long way towards meetings the goals of such a system. To clarify they are:
1) Maintaining a relationship (with mutual feedback) between extcmd devs and the users who want to experiment with and use them.
2) Lower barriers to entry for users looking for more features in Homebrew.
3) Lower barriers to entry for devs who wish to add more features to Homebrew.
4) Minimize maintenance upkeep for the maintainers of both the extcmd repo and the extcmd tap/formula.
5) Ease integration of extcmds into brew by providing clear measures for reliability and desirability.

---
Please let me know what you think. Am I off the mark? Where/Why will this system not work? What could be improved? Does this system address the problems of the previous impl?

MikeMcQuaid:
@lylemoffitt The only advantage of that approach would be discoverability. If you think that would be a good system: please feel free to implement it and push it forward. It's been a pleasant discussion but we're going pretty far off topic now and we're far into the realm of "stuff you'd like that you should consider implementing". Let's call it at this. Thanks again.

PR #7899, opened 2016-12-15
rgbds: update homepage/desc, add head

frarees:

-----

* On macOS Sierra building fails as it doesn't properly load `PNGFLAGS`. Now includes a valid `PNGFLAGS` override in the formula
* Adds libpng and pkg-config as dependencies, needed at build time
* Matches name with original capitalization
* Updates homepage as it was down

### DIFF
--- a/Formula/rgbds.rb
+++ b/Formula/rgbds.rb
@@ -1,6 +1,6 @@
 class Rgbds < Formula
-  desc "Rednex GameBoy development system"
-  homepage "https://www.anjbe.name/rgbds/"
+  desc "Rednex GameBoy Development System"
+  homepage "https://github.com/bentley/rgbds"
   url "https://github.com/bentley/rgbds/releases/download/v0.2.4/rgbds-0.2.4.tar.gz"
   sha256 "a7d32f369c6acf65fc0875c72873ef21f4d3a5813d3a2ab74ea604429f7f0435"
 
@@ -14,8 +14,11 @@ class Rgbds < Formula
     sha256 "c08b856ecf4cb57390ec99241d0bc87ba623536ab7e3e11d5cc23334230ff1cd" => :mavericks
   end
 
+  depends_on "pkg-config" => :build
+  depends_on "libpng" => :build
+
   def install
-    system "make", "install", "PREFIX=#{prefix}", "MANPREFIX=#{man}"
+    system "make", "install", "PREFIX=#{prefix}", "MANPREFIX=#{man}", "PNGFLAGS=-I#{Formula["libpng"].opt_include}"
   end
 
   test do


Formula/rgbds.rb
```diff
@@ -14,8 +14,11 @@ class Rgbds < Formula
     sha256 "c08b856ecf4cb57390ec99241d0bc87ba623536ab7e3e11d5cc23334230ff1cd" => :mavericks
   end
 
+  depends_on "pkg-config" => :build
```
woodruffw: Although it looks like upstream has hardcoded `pkg-config` in their Makefile, so we'll need it anyways. Ignore the removal bit, but the rest of the previous comment still applies.

frarees: Thanks! I've updated the install task. I'll see if I can help upstream to modify the Makefile to make it possible to not depend on `pkg-config` when brewing.

woodruffw: I was about to ask, thanks for thinking in advance!

woodruffw:
Did you mean to add a `devel` block with this PR? Asking because the title is "rgbds 0.2.4, 0.2.5 (devel)".

Formula/rgbds.rb
frarees: @woodruffw I get `/usr/local/opt/libpng/include/libpng16` (system) with the Formula["libpng"] method (which I am not familiarized at), and `/usr/local/Cellar/libpng/1.6.26/include/libpng16` (brew) with `pkg-config`. I feel like the brew'd version is desired over the system one, but I am unsure. What do you think?

frarees:
@woodruffw I was mostly interested in being able to install via brew the latest on upstream (HEAD). I've added devel following the contrib guidelines, but I might misunderstood. Should I just refer to 0.2.4, then?

woodruffw:
> I've added devel following the contrib guidelines, but I might misunderstood

Yeah, I think you might have confused `head` with `devel`. `head` is for builds directly from SCM or another dynamic source, while `devel` is for release candidates, betas, etc.

It's not a problem, I just wanted to make sure you were on the right page. I'll update the PR's title.

Formula/rgbds.rb
frarees: @woodruffw oh, I see. If that's the case I don't get why HEAD breaks using the Formula["libpng"] method. In ~1 hour I will have some spare time to get to the issue.

frarees:
Thanks!

Formula/rgbds.rb
frarees: Think I figured out. It should work on every scenario now. Another time I will take a look at upstream to see if changes on the Makefile are possible concerning dependencies. Thanks again!

woodruffw: No problem. Just out of curiosity, what was the issue?

frarees: I wasn't including `-I` in `PNGFLAGS` as `pkg-config` was. Rewrote the sentence a bit as well.

woodruffw: Thanks!

woodruffw:
Thanks @frarees!

PR #7901, opened 2016-12-15
Add plist_options and placeholder test for automysqlbackup

GauthamGoli:

-----

Could you please help me define what should be the plist_options? 
I set `startup => true` because, the backups have to be taken regardless, but please correct me if I am wrong. 
What should be the `startup_user` and `manual` options?

### DIFF
--- a/Formula/automysqlbackup.rb
+++ b/Formula/automysqlbackup.rb
@@ -28,7 +28,7 @@ class Automysqlbackup < Formula
     EOS
   end
 
-  plist_options :startup => true
+  plist_options :manual => "automysqlbackup"
 
   def plist; <<-EOS.undent
     <?xml version="1.0" encoding="UTF-8"?>
@@ -60,5 +60,6 @@ class Automysqlbackup < Formula
   end
 
   test do
+    system "#{sbin}/automysqlbackup", "--help"
   end
 end
--- a/Formula/automysqlbackup.rb
+++ b/Formula/automysqlbackup.rb
@@ -28,6 +28,8 @@ class Automysqlbackup < Formula
     EOS
   end
 
+  plist_options :startup => true
+
   def plist; <<-EOS.undent
     <?xml version="1.0" encoding="UTF-8"?>
     <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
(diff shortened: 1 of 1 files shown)


Formula/automysqlbackup.rb
```diff
@@ -28,6 +28,8 @@ def caveats; <<-EOS.undent
     EOS
   end
 
+  plist_options :startup => true
```
MikeMcQuaid: You can omit `:startup => true` and instead specify `:manual` e.g. `:manual => "automysqlbackup"`

Formula/automysqlbackup.rb
```diff
@@ -56,4 +58,7 @@ def plist; <<-EOS.undent
     </plist>
     EOS
   end
+
+  test do
```
MikeMcQuaid: Can this test be modified to do something e.g. run `--version` or `--help` at a minimum? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

MikeMcQuaid (changes requested):

Formula/automysqlbackup.rb
GauthamGoli: Thanks for the quick response. A good test according to my understanding would run the command, check if databases are backed up, but would require the database credentials in the configuration file and hence cannot be implemented. 
So, I updated the test block to run `--help` instead.

MikeMcQuaid: That'd be ideal, yeh. You could manually start a database in the test. Alternatively, having it do an `assert_match` on a  `--version` would be great at a minimum. Thanks!

GauthamGoli: There is no `--version` switch available for `automysqlbackup` 

- The default configuration file has a version number in its first line but that is previous version number, so can't be used to write a test for current version. 
- Running `automysqlbackup` without any options returns a long output which contains the version number, provided `mysql` is already installed. Can I use this for writing an `assert_match` to match version using regex ? But I guess, this may not be a good test as the output might change and this test might fail.

MikeMcQuaid: How about `--help`?

MikeMcQuaid: @GauthamGoli Great, thanks. You can ignore that failure, we'll merge it regardless.

MikeMcQuaid (approved):

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
Thanks again @GauthamGoli!

GauthamGoli:
Thank you too for helping me get this PR merged! :D

PR #7910, opened 2016-12-15
Remove 32-bit options

MikeMcQuaid:
No supported versions of macOS default to 32-bit and these can cause massive issues with dependencies.


ilovezfs (approved):

ilovezfs:
10 failed steps per node.

PR #7943, opened 2016-12-16
openshift-cli: fix socat dependency not global

elyscape:

-----

The dependency on socat was added to the devel block in #4105 when the prerelease version was an alpha for 1.3.0. This has since been released, so socat is needed as a regular dependency.


nijikon:

PR #7947, opened 2016-12-16
fantom: install fixes and tweak options to install everything by default

afrankvt:

-----

Generally we expect Fantom to be installed "as-is" - so stripping stuff by default can cause some problems in several cases.  So this switches the behavior to only strip if specified.  Also fixes a problem where one of the Fantom property files was not properly configured on macOS.


MikeMcQuaid:
How big are the examples and/or source? If they aren't dramatically bigger than the existing installation it's worth just installing them all and not offering options.

afrankvt:
I'm not really sure how much these options will get used - but the src is ~9MB so seems like someone may want to use that if they are a power user (and/or pulling the src directly somewhere else).  The examples are more negligible (~130k) - but I do agree with @mistydemeo - makes sense to keep those options separate - updated PR to break out src/example as separate options.

MikeMcQuaid:
@afrankvt Given they are both fairly small let's just remove the options altogether. Options don't get CI test coverage so we want to remove them unless they are going to be really widely used. Thanks!

afrankvt:
@MikeMcQuaid that sounds good - though my branch got hosed up - going to reopen a new clean PR

PR #7955, opened 2016-12-16
moreutils: fix compatibility with task-spooler, pwntools, & parallel

lylemoffitt:
The formula `moreutils` is a meta-package that provides a number of tools. Presently, it has listed conflicts with `task-spooler`, `pwntools`, and `parallel`. This PR adds installation options to obviate those conflicts by simply not installing the infringing binary. 

The formula previously has the option to install `without-parallel`, but this wasn't reflected in the `conflicts_with` section. This PR also fixes that.

----

### DIFF
--- a/Formula/moreutils.rb
+++ b/Formula/moreutils.rb
@@ -15,12 +15,14 @@ class Moreutils < Formula
   end
 
   option "without-parallel", "Build without the 'parallel' tool."
+  option "without-errno",    "Build without the 'errno' tool, for compatibility with 'pwntools'."
+  option "without-ts",       "Build without the 'ts' tool, for compatibility with 'task-spooler'."
 
   depends_on "docbook-xsl" => :build
 
-  conflicts_with "parallel", :because => "Both install a `parallel` executable."
-  conflicts_with "pwntools", :because => "Both install an `errno` executable."
-  conflicts_with "task-spooler", :because => "Both install a `ts` executable."
+  conflicts_with "parallel", :because => "Both install a `parallel` executable."  if build.with? "parallel"
+  conflicts_with "pwntools", :because => "Both install an `errno` executable."    if build.with? "errno"
+  conflicts_with "task-spooler", :because => "Both install a `ts` executable."    if build.with? "ts"
 
   resource "Time::Duration" do
     url "https://cpan.metacpan.org/authors/id/N/NE/NEILB/Time-Duration-1.20.tar.gz"
@@ -53,6 +55,14 @@ class Moreutils < Formula
     if build.without? "parallel"
       inreplace "Makefile", /^BINS=.*\Kparallel/, ""
       inreplace "Makefile", /^MANS=.*\Kparallel\.1/, ""
+    end
+    if build.without? "errno"
+      inreplace "Makefile", /^BINS=.*\Kerrno/, ""
+      inreplace "Makefile", /^MANS=.*\Kerrno\.1/, ""
+    end
+    if build.without? "ts"
+      inreplace "Makefile", /^PERLSCRIPTS=.*\Kts/, ""
+      inreplace "Makefile", /^MANS=.*\Kts\.1/, ""
     end
     system "make", "all"
     system "make", "check"


ilovezfs:
I cannot say I'm thrilled with the direction this formula is going. It seems like an invitation to slice and dice anything that's keg_only into a non-keg_only version with a pile of options.

lylemoffitt:
@ilovezfs I'm not really thrilled about it either. I'm just trying to make it compatible with the rest of the formulas. The ultimate solution should be to put each of these tools in a separate formula. Normally when there's a conflict between formula, you're choosing between whole replacements of one or the other. In this case `ts` and `errno` are completely different programs from their nominal counterparts.

ilovezfs:
Wouldn't the usual way to handle this be to `brew unlink moreutils` and then add symlink(s) to /usr/local/opt/moreutils/bin/$tool somewhere in your PATH for the desired subset?

lylemoffitt:
Catch22: How do you `brew link` if you can't install it in the first place? Moreover, how do you direct the user to do all that? Even if you put all that install logic in the formula, doesn't that just leave you with the same or more install options?

ilovezfs:
>How do you brew link if you can't install it in the first place

by brew unlinking the conficting one first, or by making the formula keg_only

>Moreover, how do you direct the user to do all that?

out of scope for the formula itself

>Even if you put all that install logic in the formula

I mean outside of brew itself, not something that's part of the formula.

ilovezfs:
basically
```
brew unlink $conflict
brew install moreutils
brew unlink moreutils
brew link $conflict
ln -s /usr/local/opt/moreutils/bin/foo $somewhereinyourpath
```

If it were keg_only
```
brew install moreutils
ln -s /usr/local/opt/moreutils/bin/foo $somewhereinyourpath
```

ilovezfs:
Just for some perspective, in the last 30 days:

1,139 `brew install moreutils`

13 `brew install moreutils --without-parallel`

lylemoffitt:
> by making the formula keg_only

I thought this was for stuff that was only needed as a build dependency of something else? These are user-land tools conflicting with other user-land tools.

> `ln -s /usr/local/opt/moreutils/bin/foo $somewhereinyourpath`

I mean, I get what you're saying. But you're basically just directing the user to link something themselves that brew could have linked for them. This is made worse because `moreutils` isn't one command, its 15. So really, you're asking the user to do *this* as the last step:

```shell
ln -s /usr/local/opt/moreutils/bin/isutf8  $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/ifdata  $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/ifne    $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/pee     $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/sponge  $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/mispipe $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/lckdo   $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/paralle $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/errno   $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/vidir   $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/vipe    $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/ts      $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/combine $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/zrun    $somewhereinyourpath
ln -s /usr/local/opt/moreutils/bin/chronic $somewhereinyourpath
```

Which is super annoying. The only real thing that forcing the user to link it for themselves is good for is that it involves them in the renaming process.

lylemoffitt:
> Just for some perspective, in the last 30 days:

That's super cool! How did you get those stats?

> 13 brew install moreutils --without-parallel

I suspect that this number might be so low in part because it doesn't really do anything for the user. Installing `without-parallel` doesn't allow you to install the `parallel` package from another formula. The `conflicts_with` is static, not conditional.

ilovezfs:
>That's super cool! How did you get those stats?

Google analytics! https://github.com/Homebrew/brew/blob/master/docs/Analytics.md

User can do this:
```
ln -s /usr/local/opt/moreutils/bin/* $somewhereinyourpath
rm $somewhereinyourpath/one_that_must_die
```

ilovezfs:
> Installing without-parallel doesn't allow you to install the parallel package from another formula. The conflicts_with is static, not conditional.

Right. So user will still be blocked when attempting to install the other formula. You'd have to inspect the INSTALL_RECEIPT.json (i.e., the `tab`) for moreutils in the conflicting formulae to make the conflicts conditional.

lylemoffitt:
>You'd have to inspect the INSTALL_RECEIPT.json (i.e., the tab) for moreutils in the conflicting formulae to make the conflicts conditional.

For handling `without-parallel` it's a one-liner in the formula.
```ruby
  conflicts_with "parallel", :because => "Both install a `parallel` executable."  if build.with? "parallel"
```

ilovezfs:
That only works if `moreutils` is the second one you're trying to install.

lylemoffitt:
The bottom line here is I don't get why this is better. If you require the user to do the linking for you, you have to put that kind of instruction in the `caveats` of something. This requires the same complexity of formula. Furthermore, if you uninstall the package later, the link won't be tracked by homebrew and will end up pointing to nothing.

----
> That only works if moreutils is the second one you're trying to install.

Well that's depressing. Here I was thinking this was going to be easy. 

---
The way I see it, there are really three options here. I'd like to know why you think the linking is the best.
1) Provide `without-foo` for every conflicting `foo`.
2) Provide a separate formula for each installed tool.
3) Manually link (and rename) the offending tool/s.

Each has it's drawbacks.

ilovezfs:
>if you uninstall the package later, the link won't be tracked by homebrew and will end up pointing to nothing.

that doesn't break anything
```
Josephs-MacBook-Pro:~ joe$ ln /bin/ls myls
Josephs-MacBook-Pro:~ joe$ ln -s ../myls mu/ls
Josephs-MacBook-Pro:~ joe$ which ls
/Users/joe/mu/ls
Josephs-MacBook-Pro:~ joe$ rm myls 
override rwxr-xr-x  root/wheel compressed for myls? y
Josephs-MacBook-Pro:~ joe$ which ls
/bin/ls
```

>Provide without-foo for every conflicting foo.

would require inspecting tabs

>Provide a separate formula for each installed tool.

15 new formulae to maintain

ilovezfs:
So looks like the way to handle the case where moreutils is installed first, is to use install -f.
```
brew install --without-parallel moreutils
brew install -f parallel
```
is valid so let's see if anyone ends up using these.

ilovezfs:
@lylemoffitt thanks for the PR!

lylemoffitt:
Not to put too fine a point on it, but the example you gave about links is erroneous for a couple of reasons:
1) You used a hard-link for myls. Homebrew doesn't use hard-links. It's copies the executable to the Cellar and creates a sym-link to it in `/usr/local/bin`. 
2) The result of `which` contradicts the point you're making. In the first call it recognizes that `mu/ls` is a valid executable in PATH. However, in the second call it proves that it's not. If you check `mu/ls` you will find that it still points to `myls` (which isn't there).
```
bash-4.4$ ls -l mu/ls
lrwxr-xr-x  1 Lyle  wheel  7 Dec 16 22:47 mu/ls -> ../myls
```

---
Now, I don't bring this up to be pedantic, only to prove my point that managing system link (hard or sym) is tricky. It's part of the reason we use package managers to handle this stuff for us.

---
>> Provide a separate formula for each installed tool.

>15 new formulae to maintain

This is less of a problem when they all use the same tar ball and make file, but the point *is* valid. Unfortunately, it seems that Homebrew is just built that way. It's mechanics very much assume that formula have a 1-1 relationship with the binaries they install. I think multiple formula is the right approach given that, but I'd like it if there was a way to not make it burdensome. Is there a way to write formula that wouldn't be some cumbersome and would adhere to DRY when possible?

What do you think? What's your ideal solution for this kind of problem (in general)?

ilovezfs:
>    You used a hard-link for myls. Homebrew doesn't use hard-links. It's copies the executable to the Cellar and creates a sym-link to it in /usr/local/bin.
>    The result of which contradicts the point you're making. In the first call it recognizes that mu/ls is a valid executable in PATH. However, in the second call it proves that it's not. If you check mu/ls you will find that it still points to myls (which isn't there).

That is intentional. The point is that it falls back to /bin/ls as expected despite the symlink that points nowhere being in the path.

>What do you think? What's your ideal solution for this kind of problem (in general)?

I don't think there is one. moreutils is an incoherent jumble as its name suggests, so here we are.

lylemoffitt:
Alright. Thanks.

derimagia:
Just to note, I've been using this for a while not and the main issue I have is that `brew upgrade` doesn't seem to keep track of the `-f` on parallel so it halts up the entire upgrade process to unlink moreutils first (or run `brew upgrade -f parallel`.

ilovezfs:
Yup, force linking isn't persistent. `conflicts_with` means what it says.

PR #7956, opened 2016-12-16
WIP upx 3.92

mistydemeo:

-----

UPX 3.92 fixes support for building/running on Sierra. The tests are currently not running: compressing executables is broken (every result returns `NotCompressibleException`), but decompressing executables works.

### DIFF
--- a/Formula/upx.rb
+++ b/Formula/upx.rb
@@ -23,21 +23,11 @@ class Upx < Formula
   end
 
   test do
-    (testpath/"hello-c.c").write <<-EOS.undent
-      #include <stdio.h>
-      int main()
-      {
-        puts("Hello, world!");
-        return 0;
-      }
-    EOS
-    system "cc", "-o", "hello-c", "hello-c.c"
-    assert_equal "Hello, world!\n", `./hello-c`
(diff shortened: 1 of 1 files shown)


ilovezfs:
Odd. I thought this release was supposed to have full Sierra support.

mistydemeo:
Yeah, I thought so too.

mistydemeo:
After some debugging, it looks like the issue is that 3.92 is producing moderately larger executables compared to 3.91. When creating an executable, upx compares the size of the compressed executable to the original, and raises `NotCompressibleException` if it isn't at least 512 bytes smaller: https://github.com/upx/upx/blob/master/src/packer.cpp#L328-L335

Our test program produces a binary that's 8432 bytes, and that seems to be around the threshold that no longer gets an appreciable size gain. With both tests running on El Cap, 3.91 produces a 4096-byte output while 3.92 produces a 8816 byte binary - e.g., a 384-byte *increase*.

3.92 *does* still work fine on non-trivial binaries, so I think an adjustment to the test is in order.

mistydemeo:
Updated to test against UPX's own binary, which gets a compression ratio of about 33.78% and exits 0.

PR #7962, opened 2016-12-17
go: Fix cross-compilation.

tsuna:
The build was forcing CGO_ENABLED=1, which is not necessary.  But as of
Go 1.8, not only it's not necessary, it's also harmful, as the value of
CGO_ENABLED used during Go's build gets recorded and used by default [1]
when building any Go code, which prevents the compiler from disabling cgo
automatically when cross-compiling. [2]

[1] See golang/go#12808
[2] See golang/go#18360

-----


tsuna:
Hmm, sorry, this is a dup of #7556.  Should've checked existing PRs first.

PR #6817, opened 2016-11-11
pyqt: migrate to boneyard

MikeMcQuaid:
Created with `brew boneyard-formula-pr` because Qt4 is EOL. CC @cartr because Qt4 boneyarding.

PR #7976, opened 2016-12-18
gnuradio 3.7.10.1; restored optional components

kolen:

-----

- Bumped version to 3.7.10.1
- Restored optional components removed in 2a6d326a61507caf279352b86f3ef1116c3a8a1f (#6817): python, gnuradio-companion and GUIs
  
  Everything python-related was removed in 2a6d326a61507caf279352b86f3ef1116c3a8a1f, seemingly by mistake, when trying to remove pyQt (gnuradio formula of that version installed everything if enabling python). I restored these components (except pyQt which requires Qt 4.x which no longer builds on latest mac os)

  I set most components as `:recommended` because I think primary use case of gnuradio is to run interactive DSP workflows with gnuradio-companion and WX UI components. If not building gnuradio-companion, it's usable only as c++/python lib (which is sometimes useful too).

  I didn't made it possible to build it entirely without python — why if python is always available in MacOS?
- Removed building custom version cmake, gnuradio builds ok with latest cmake now.
- Added some basic tests 
- Changed homepage to current homepage
- Added `head`

Formula contains patch from gnuradio/gnuradio#1146 to make it properly link against python (``-undefined dynamic_lookup``).

### DIFF
--- a/Formula/gnuradio.rb
+++ b/Formula/gnuradio.rb
@@ -1,9 +1,9 @@
 class Gnuradio < Formula
   desc "SDK providing the signal processing runtime and processing blocks"
-  homepage "https://gnuradio.squarespace.com/"
-  url "https://gnuradio.org/releases/gnuradio/gnuradio-3.7.9.1.tar.gz"
-  sha256 "9c06f0f1ec14113203e0486fd526dd46ecef216dfe42f12d78d9b781b1ef967e"
-  revision 2
+  homepage "https://gnuradio.org/"
+  url "https://gnuradio.org/releases/gnuradio/gnuradio-3.7.10.1.tar.gz"
+  sha256 "63d7b65cc4abe22f47b8f41caaf7370a0a502b91e36e29901ba03e8838ab4937"
+  head "https://github.com/gnuradio/gnuradio.git"
 
   bottle do
     sha256 "7f2b54d889d4d568736e9ae8221b0a652015d6abb18b1466a5637f92a007884f" => :sierra
@@ -11,34 +11,37 @@ class Gnuradio < Formula
     sha256 "464555aaac52b55b9e09ec2b0b2e0e4cabc024b655a1bf815d7ddde59aec92fe" => :yosemite
   end
 
-  depends_on "pkg-config" => :build
+  # Fixes linkage of python (swig) bindings directly to python
+  # https://github.com/gnuradio/gnuradio/pull/1146
+  # Merged into master, probably will be in next release
+  patch do
+    url "https://github.com/gnuradio/gnuradio/pull/1146.patch"
+    sha256 "fbf9842292cc1c2cfcf708d648d01c9a42fb98c5bba542a272dcdac504454d26"
+  end
 
-  depends_on :python if MacOS.version <= :snow_leopard
+  option "without-python", "Build without python support"
+
+  depends_on "pkg-config" => :build
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
   depends_on "boost"
   depends_on "cppunit"
   depends_on "fftw"
   depends_on "gsl"
   depends_on "zeromq"
+  depends_on "numpy" if build.with? :python
+  depends_on "swig" => :build if build.with? :python
+  depends_on "cmake" => :build
 
   # For documentation
-  depends_on "doxygen" => :build
-  depends_on "sphinx-doc" => :build
+  depends_on "doxygen" => [:build, :optional]
+  depends_on "sphinx-doc" => [:build, :optional]
 
   depends_on "uhd" => :recommended
-  depends_on "sdl" => :recommended
-  depends_on "jack" => :recommended
+  depends_on "sdl" => :optional
+  depends_on "jack" => :optional
   depends_on "portaudio" => :recommended
-
-  # gnuradio is known not to compile against CMake >3.3.2 currently.
-  resource "cmake" do
-    url "https://cmake.org/files/v3.3/cmake-3.3.2.tar.gz"
-    sha256 "e75a178d6ebf182b048ebfe6e0657c49f0dc109779170bad7ffcb17463f2fc22"
-  end
-
-  resource "numpy" do
-    url "https://pypi.python.org/packages/source/n/numpy/numpy-1.10.1.tar.gz"
-    sha256 "8b9f453f29ce96a14e625100d3dcf8926301d36c5f622623bf8820e748510858"
-  end
+  depends_on "pygtk" => :optional
+  depends_on "wxpython" => :optional
 
   # cheetah starts here
   resource "Markdown" do
@@ -66,33 +69,7 @@ class Gnuradio < Formula
     ENV["CHEETAH_INSTALL_WITHOUT_SETUPTOOLS"] = "1"
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
 
-    resource("cmake").stage do
-      args = %W[
-        --prefix=#{buildpath}/cmake
-        --no-system-libs
-        --parallel=#{ENV.make_jobs}
-        --datadir=/share/cmake
-        --docdir=/share/doc/cmake
-        --mandir=/share/man
-        --system-zlib
-        --system-bzip2
-      ]
-
-      # https://github.com/Homebrew/homebrew/issues/45989
-      if MacOS.version <= :lion
-        args << "--no-system-curl"
-      else
-        args << "--system-curl"
-      end
-
-      system "./bootstrap", *args
-      system "make"
-      system "make", "install"
-    end
-
-    ENV.prepend_path "PATH", buildpath/"cmake/bin"
-
-    res = %w[Markdown Cheetah lxml numpy]
+    res = %w[Markdown Cheetah lxml]
     res.each do |r|
       resource(r).stage do
         system "python", *Language::Python.setup_install_args(libexec/"vendor")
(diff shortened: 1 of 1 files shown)


mistydemeo:
Looks like the test is failing; it can't import numpy.

kolen:
Seems that it was using numpy manually installed on my system with pip instead of numpy installed into ``vendor/.../site-packages`` (despite using sandbox in tests, not sure how sandbox works). Retrying with numpy from homebrew.

tdsmith:
There is no good way to handle numpy for the gnuradio bindings. :( As a matter of policy we can't depend on Homebrew's numpy package in core.

The problem is that, as you noticed, installing the numpy distribution to a hidden path isn't good enough because it needs to be present in sys.path when gnuradio is imported. A way to make sure numpy is present while avoiding linking conflicts with a system-installed numpy is to install numpy to a private prefix (e.g. `libexec/numpy`) and then do something like:

`(lib/"python2.7/site-packages/homebrew-gnuradio-numpy.pth").write libexec/"numpy/python2.7/site-packages"`

to put the vendored numpy at the end of the global sys.path.

Alas, this approach doesn't work with system Python, because Apple ships a random old bad version of numpy that appears in sys.path before any of our packages will. If we try and override it, we'll end up overriding the installed numpy version in all cases, which will frustrate numpy users.

It seems like some options are:

* accept that this doesn't work with system Python and add a dependency on "python" so that CI works
* make the Python bindings optional and have them depend on a NumpyRequirement to force users to install their own numpy
* ?

Formula/gnuradio.rb
```diff
@@ -31,17 +42,8 @@ class Gnuradio < Formula
...
-    url "https://cmake.org/files/v3.3/cmake-3.3.2.tar.gz"
-    sha256 "e75a178d6ebf182b048ebfe6e0657c49f0dc109779170bad7ffcb17463f2fc22"
-  end
-
-  resource "numpy" do
-    url "https://pypi.python.org/packages/source/n/numpy/numpy-1.10.1.tar.gz"
-    sha256 "8b9f453f29ce96a14e625100d3dcf8926301d36c5f622623bf8820e748510858"
-  end
+  depends_on "pygtk" => :recommended
+  depends_on "wxpython" => :recommended
```
tdsmith: This uses two different windowing toolkits?

Formula/gnuradio.rb
```diff
@@ -1,9 +1,9 @@
 class Gnuradio < Formula
   desc "SDK providing the signal processing runtime and processing blocks"
-  homepage "https://gnuradio.squarespace.com/"
-  url "https://gnuradio.org/releases/gnuradio/gnuradio-3.7.9.1.tar.gz"
-  sha256 "9c06f0f1ec14113203e0486fd526dd46ecef216dfe42f12d78d9b781b1ef967e"
-  revision 1
+  homepage "http://gnuradio.org/"
+  url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.10.1.tar.gz"
```
kolen: `https://gnuradio.org/releases/gnuradio/gnuradio-3.7.10.1.tar.gz` redirects to `http://gnuradio.org/releases/gnuradio/gnuradio-3.7.10.1.tar.gz`.

Formula/gnuradio.rb
```diff
@@ -12,7 +12,15 @@ class Gnuradio < Formula
...
 
+  # Fixes linkage of python (swig) bindings directly to python
+  # https://github.com/gnuradio/gnuradio/pull/1146
+  patch do
+    url "https://github.com/gnuradio/gnuradio/pull/1146.patch"
+    sha256 "385b62cc46ff07696f7ff4ff89bbb4891909947027f9712be040937d477f6d81"
+  end
+
   option :universal
+  option "with-documentation", "Build with documentation"
```
MikeMcQuaid: Instead call this `with-doxygen` and make it `[:build, :optional]` and, if necessary, add a `deprecated_option`.

Formula/gnuradio.rb
```diff
@@ -31,16 +41,12 @@ class Gnuradio < Formula
...
   depends_on "jack" => :recommended
   depends_on "portaudio" => :recommended
-
-  # gnuradio is known not to compile against CMake >3.3.2 currently.
-  resource "cmake" do
-    url "https://cmake.org/files/v3.3/cmake-3.3.2.tar.gz"
-    sha256 "e75a178d6ebf182b048ebfe6e0657c49f0dc109779170bad7ffcb17463f2fc22"
-  end
+  depends_on "pygtk" => :recommended
+  depends_on "wxpython" => :recommended
```
MikeMcQuaid: Please make these `:optional`. We default to non-GUI by default mostly now.

MikeMcQuaid (changes requested):

ilovezfs:
I'd prefer we not do this. https://github.com/gnuradio/gnuradio/issues/1090 is already and the works, and I've already re-written this formula for Qt 5 compatibility locally pending resolution of that issue.

MikeMcQuaid:
@ilovezfs Seems reasonable to me.

kolen:
Looks like Mac OS 10.12.2 has numpy `1.8.0rc1`, gnuradio [lists `1.1.0` as minimum](http://gnuradio.org/doc/doxygen/build_guide.html), so maybe it's suitable. It works for me for basic tasks in gnuradio-companion. Added `NumpyRequirement` which checks its version. Made GUI and python optional.

I'm not sure if dependencies between dependencies should be indicated some way (if `--with-pygtk` is specified `--with-python` must be also specified, currently if building with pygtk but without python pygtk components will not be built silently).

Formula/gnuradio.rb
kolen: Changed to TLS anyway, it works and maybe some day their server will serve file directly on SSL.

MikeMcQuaid:
> There is no good way to handle numpy for the gnuradio bindings. :( As a matter of policy we can't depend on Homebrew's numpy package in core.

@tdsmith Let's import `numpy` to homebrew/core instead.

Formula/gnuradio.rb
kolen: I disabled both GUI libraries by default.

But it doesn't make sense to pick only one. Gnuradio-companion, the main gnuradio GUI, visual signal flow editor, uses GTK, however GUI blocks (for example that display spectrogram or allow to change parameters as GUI slider) use either WXWidgets or Qt (seems that most blocks use wx). I think most users who need gnuradio-companion need all three toolkits or at least gtk + wx.

Formula/gnuradio.rb
kolen: Changed.

Formula/gnuradio.rb
kolen: Changed.

MikeMcQuaid:
> @tdsmith Let's import numpy to homebrew/core instead.

I did this.

kolen:
- Updated to use `numpy` from homebrew-core
- Set config file locations to ``/usr/local/etc/gnuradio/conf.d/`` (I'm not sure if either previous version of gnuradio used `/usr/local/etc` by default or there was no linting for it yet)

There's some problem with Jenkins, tried to push 2 times, my build appeared in the queue then disappeared. Trying again (Expected build number: 1​55​58).

MikeMcQuaid:
@BrewTestBot test this please

nijikon:
Is there anything else that prevents this from being merged?

Formula/gnuradio.rb
```diff
@@ -1,46 +1,48 @@
...
+    url "https://github.com/gnuradio/gnuradio/pull/1146.patch"
+    sha256 "fbf9842292cc1c2cfcf708d648d01c9a42fb98c5bba542a272dcdac504454d26"
+  end
+
   option :universal
 
   depends_on "pkg-config" => :build
 
-  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on :python => :recommended
```
ilovezfs: This should be
```
depends_on :python => :recommended if MacOS.version <= :snow_leopard
```
and then have a
```
option "without-python", "Build without python support"
```
above.

ilovezfs (dismissed):

Formula/gnuradio.rb
```diff
@@ -1,46 +1,48 @@
...
-  depends_on "doxygen" => :build
-  depends_on "sphinx-doc" => :build
+  depends_on "doxygen" => [:build, :optional]
+  depends_on "sphinx-doc" => [:build, :optional]
 
   depends_on "uhd" => :recommended
-  depends_on "sdl" => :recommended
-  depends_on "jack" => :recommended
+  depends_on "sdl" => :optional
+  depends_on "jack" => :optional
```
ilovezfs: what is the point of removing these features? Anyone that needs them would now need to build from source.

Formula/gnuradio.rb
```diff
@@ -115,10 +93,17 @@ def install
                             gr-blocks testing gr-pager gr-noaa gr-channels
                             gr-audio gr-fcd gr-vocoder gr-fec gr-digital
                             gr-dtv gr-atsc gr-trellis gr-zeromq]
+    if build.with? "python"
+      enabled_components << "python"
+      enabled_components << "gr-utils"
+      enabled_components << "grc" if build.with? "pygtk"
```
ilovezfs: we shouldn't be nesting if's like this

Formula/gnuradio.rb
```diff
@@ -135,6 +120,79 @@ def install
...
+        def main(top_block_cls=top_block, options=None):
+            tb = top_block_cls()
+            tb.start()
+            tb.wait()
+
+        main()
+      EOS
+      system "python", (testpath/"test.py")
+
+      Dir.chdir(testpath) do
```
ilovezfs: we just use `cd` for these

Formula/gnuradio.rb
kolen: Based on [this discussion](https://github.com/Homebrew/homebrew-core/pull/7976#discussion_r92953220)

Formula/gnuradio.rb
kolen: But then if I build with ``--without-python --with-pygtk`` it will still build with python. In my case it skips python. What is right way?

EvanKrall:
Thanks for working on this, @kolen.

ilovezfs:
Please ignore any of my above remarks.

kolen:
Is it ready for merge?

fxcoudert:
@BrewTestBot test this please

fxcoudert:
Looks decent to merge for me. A newer PR will be needed for the latest version (3.7.11) but that can come later.

@MikeMcQuaid you were the last reviewer, any objection?

MikeMcQuaid (approved):

fxcoudert:
Thanks @kolen and all who participated!

PR #7978, opened 2016-12-18
grpc 1.0.1 (new formula)

maz:
This commit adds grpc (Google's RPC framework) as a formula. The formula
for HEAD is straightforward. For grpc 1.0.1, some extra steps are
required to regenerate source files that have been checked into the grpc
tree. As of grpc 1.0.1, these source files are generated with protoc 3,
while Homebrew ships protoc 3.1. To rectify this, a few files are
in-replaced, and a separate build script is run. As the comments in the
formula indicate, these hacks should be able to be removed with the next
release of grpc.

-----

### DIFF
--- a/Formula/grpc.rb
+++ b/Formula/grpc.rb
@@ -1,0 +1,44 @@
+class Grpc < Formula
+  desc "Google's open-source RPC framework, based on protocol buffers."
+  homepage "http://www.grpc.io/"
+  head "https://github.com/grpc/grpc.git"
+
+  stable do
+    # The script run below, which is no longer needed in versions newer than
+    # 1.0.1, requires fun sed things, which don't seem to work with mac's sed.
+    url "https://github.com/grpc/grpc/archive/v1.0.1.tar.gz"
+    sha256 "efad782944da13d362aab9b81f001b7b8b1458794751de818e9848c47acd4b32"
+    depends_on "gnu-sed" => :build
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "protobuf"
+  depends_on "openssl"
+
+  def install
+    ENV["prefix"] = prefix
+    ENV["HAS_SYSTEM_PROTOBUF"] = "true"
+    ENV["HAS_PKG_CONFIG"] = "true"
+    ENV["HAS_EMBEDDED_PROTOBUF"] = "0"
+
+    if build.stable?
+      # This isn't an issue after version 1.0.1. In version 1.0.1, some
+      # compiled protobufs are checked in. Those protobufs are compiled with
+      # an older version of protoc than the one in homebrew. Moreover, the
+      # script to build these files is broken, which this inreplace fixes.
+      inreplace "tools/codegen/extensions/gen_reflection_proto.sh" do |s|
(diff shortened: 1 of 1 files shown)


Formula/grpc.rb
```diff
@@ -0,0 +1,44 @@
+class Grpc < Formula
+  desc "Google's open-source RPC framework, based on protocol buffers."
+  homepage "http://www.grpc.io/"
+  head "https://github.com/grpc/grpc.git"
+
+  stable do
```
woodruffw: I don't think a `stable` block is necessary here, this can just be collapsed into the formula body.

maz: The dependency on `sed` is only required on stable, not on HEAD (and presumably future releases). So, what's the preferred syntax for that? I tried moving the checksum and url to the formula body, but then the audit system complained that, as long as there was a stable block, the url and checksum should be there.

woodruffw: Hmm, I don't think we currently have a way to specify stable-only build dependencies, unfortunately. 

I guess in that case a `stable` block is fine, since there's no point in pulling in `gnu-sed` for head builds.

ilovezfs: >Hmm, I don't think we currently have a way to specify stable-only build dependencies, unfortunately.

It's intentional.

Formula/grpc.rb
```diff
@@ -0,0 +1,44 @@
...
+        s.gsub! 'INCLUDE_DIR="grpc++/ext"', 'INCLUDE_DIR="grpc++\\/ext"'
+      end
+      system "make", "grpc_cpp_plugin"
+      system "tools/codegen/extensions/gen_reflection_proto.sh"
+    end
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/grpc_cpp_plugin < /dev/null"
```
MikeMcQuaid: Use `pipe_input` here and ideally `assert_match` to test the output value.

MikeMcQuaid (changes requested):

Formula/grpc.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  depends_on "pkg-config" => :build
+  depends_on "protobuf"
+  depends_on "openssl"
+
+  def install
+    ENV["prefix"] = prefix
+    ENV["HAS_SYSTEM_PROTOBUF"] = "true"
+    ENV["HAS_PKG_CONFIG"] = "true"
+    ENV["HAS_EMBEDDED_PROTOBUF"] = "0"
```
MikeMcQuaid: If any of these are needed only for a single `make` it'd be preferable to specify them only for that. e.g. `system "make", "install", "PREFIX=#{prefix}"`

MikeMcQuaid (changes requested):

Formula/grpc.rb
maz: They're needed for both `make`s. However, once the next version of grpc is released, the additional `make` won't be required. So, should this be left as-is?

MikeMcQuaid: Yep, thanks.

fxcoudert:
`grpc` was added by #9942. Closing this PR.

PR #7983, opened 2016-12-18
dnscrypt-proxy 1.8.0

ilovezfs:


DomT4:
FWIW you can find the release tarball [here](https://download.dnscrypt.org/dnscrypt-proxy/) if it's not on GitHub yet.

ilovezfs:
might as well as always run autoconf if it's not going to be on GitHub reliably

PR #8016, opened 2016-12-19
git: Add option to build with LibreSSL

oherrala:

-----

Git can already be build with Homebrew'd OpenSSL. With this change it's possible to build with Homebrew'd LibreSSL.

Building with OpenSSL _and_ LibreSSL is disallowed.

### DIFF
--- a/Formula/git.rb
+++ b/Formula/git.rb
@@ -14,13 +14,13 @@ class Git < Formula
   option "with-blk-sha1", "Compile with the block-optimized SHA1 implementation"
   option "without-completions", "Disable bash/zsh completions from 'contrib' directory"
   option "with-brewed-openssl", "Build with Homebrew OpenSSL instead of the system version"
+  option "with-brewed-libressl", "Build with Homebrew LibreSSL"
   option "with-brewed-curl", "Use Homebrew's version of cURL library"
   option "with-brewed-svn", "Use Homebrew's version of SVN"
   option "with-persistent-https", "Build git-remote-persistent-https from 'contrib' directory"
 
   depends_on "pcre" => :optional
   depends_on "gettext" => :optional
-  depends_on "openssl" if build.with? "brewed-openssl"
   depends_on "curl" if build.with? "brewed-curl"
   depends_on "go" => :build if build.with? "persistent-https"
   # Trigger an install of swig before subversion, as the "swig" doesn't get pulled in otherwise
@@ -29,6 +29,13 @@ class Git < Formula
     depends_on "swig"
     depends_on "subversion" => "with-perl"
   end
+
+  if build.with?("brewed-openssl") && build.with?("brewed-libressl")
+    odie "Options --with-brewed-openssl and --with-brewed-libressl are mutually exclusive."
+  end
+
+  depends_on "openssl" if build.with? "brewed-openssl"
+  depends_on "libressl" if build.with? "brewed-libressl"
 
   resource "html" do
     url "https://www.kernel.org/pub/software/scm/git/git-htmldocs-2.11.0.tar.xz"
@@ -91,7 +98,9 @@ class Git < Formula
       CFLAGS=#{ENV.cflags}
       LDFLAGS=#{ENV.ldflags}
     ]
-    args << "NO_OPENSSL=1" << "APPLE_COMMON_CRYPTO=1" if build.without? "brewed-openssl"
+    if build.without?("brewed-openssl") && build.without?("brewed-libressl")
+      args << "NO_OPENSSL=1" << "APPLE_COMMON_CRYPTO=1"
+    end
 
     system "make", "install", *args
 


DomT4:
Is this supported upstream? I'd completely missed `git` had added `libressl` support.

IMO having either option be called `brewed-cryptolibrary` is silly these days because Homebrew hasn't permitted using the system OpenSSL or LibreSSL for a while, which is why `git-imap-send` gets deleted unless you use Homebrew's `openssl`.

Might just be best to deprecate those options onto `with-openssl` and `with-libressl` options and then just `depends_on "openssl" => :optional` etc.

oherrala:
@DomT4 Git is compiled against LibreSSL for example in OpenBSD and apparently it works. OpenSSL and LibreSSL are pretty much API compatible, with some exceptions.

I agree with `with-openssl` and `with-libressl` options. I can look into it. It makes the formula cleaner.

MikeMcQuaid:
I'm a hard 👎 on this and am adding code to ensure we stop doing this on other formulae too (https://github.com/Homebrew/homebrew-core/pull/8040, https://github.com/Homebrew/brew/pull/1700). I don't think it's sustainable or sensible to add LibreSSL options everywhere OpenSSL is used. Homebrew has standardised on OpenSSL.

DomT4:
I don't think there's evidence of a serious maintainer impact on the current status of LibreSSL/OpenSSL handling, which I think I'm still in a decent position from the past to state. There should be more care taken to ensure crypto/ssl libraries don't cross over each other in the same linkage tree, but I think a sharp outlawing is overreach & I don't really understand what it _gains_ you.

I think you're inviting trouble to be honest. There are upstreams that for example prefer to link to LibreSSL or GnuTLS but we kick them over to OpenSSL because it is supported, and mostly Homebrew gets away with it, but there have been times where that has backfired in the past and upstreams have gone huh, yeah, we don't really support that backend, it works but we're not actively testing for it.

In the PR you linked to for example `h2o` uses LibreSSL upstream, but I kept it on `openssl`, but we're explicitly going against the upstream default there. That may well come as a surprise to upstream if we ever cause a bug to be punted up there. I believe `cURL` links to `openssl` by default rather than Secure Transport as well because it is feature incomplete against Secure Transport. There's a really weird mix of rules on ssl/crypto usage across Homebrew and I don't think your proposed change is actually making the situation any better.

I have no power here any more and ultimately you'll do what you wish to, but since this used to be my backyard maybe my opinion counts for the smallest something, I'd consider it a misstep to make a _broad_ decision.

ilovezfs:
Just to look at it empirically for a moment,

https://github.com/Homebrew/homebrew-core/pull/8040
+6 −46

so 40 lines for 5 formulae, 8 lines per formula.

There are 332 formulae with openssl dependency, so putting aside the original 5 as sunk cost, that's an additional 327 potentially. Let's say only 1/3 become eligible at some point in the near future for libressl support. That would be 109 formulae.

At 8 lines per formula, that's 872 lines.

If a line takes approximately 1 minute of maintainer time, that's 14.5 hours over whatever period you think 1/3 eligibility entails.

DomT4:
To be fair though, some of that code is pure fat I suspect. `superenv` should and generally does more than handle sufficiently which is furthest up the `$PATH` and can be found without the need for a flag. And we'd been talking for ages that Homebrew should _know_ how to handle overlapping SSL/Crypto deps outside of having to do it individually.

I'm less fussed about the optional removal than the confusion around OpenSSL being declared the standard, even when that standard doesn't mesh with the upstream standard. I think you're opening a bit of a can of worms on that. If your goal is simplicity there's not much simple about:

"We use OpenSSL everywhere regardless of the upstream default where there's support, but only ever Homebrew's OpenSSL, and it can use either `openssl` or `openssl@1.1`, unless it can use the system SecureTransport or CommonCrypto then it should use the system, but if it has linkage to both the system crypto elements and OpenSSL elements the parts that link the OpenSSL elements should be removed"

ilovezfs:
>I'm less fussed about the optional removal than the confusion around OpenSSL being declared the standard, even when that standard doesn't mesh with the upstream standard.

I agree with you. Even if we're not going to maintain dual options, I don't think we should prefer openssl over libressl if upstream recommends the latter. Trying to eliminate ambiguity by making the standard "hard libressl requirement" would be a mistake.

MikeMcQuaid:
> I'm less fussed about the optional removal than the confusion around OpenSSL being declared the standard, even when that standard doesn't mesh with the upstream standard.

It is our standard because a) it's provided by macOS and b) it's dramatically more widely used. What that means on a package-by-package basis is up to us to decide. I don't think we'll find (m)any other package managers that will just defer entirely to upstream on this issue given that one is designed to be a drop-in replacement of the other. Of course if upstream don't support OpenSSL we shouldn't try to shoehorn it in.

ilovezfs:
>It is our standard because a) it's provided by macOS and b) it's dramatically more widely used. What that means on a package-by-package basis is up to us to decide. I don't think we'll find (m)any other package managers that will just defer entirely to upstream on this issue given that one is designed to be a drop-in replacement of the other. Of course if upstream don't support OpenSSL we shouldn't try to shoehorn it in.

"Prefer openssl over libressl but no shoehorning" sounds right to me.

DomT4:
>  it's provided by macOS

This argument doesn't really work these days. So is `libressl`, which is more widely used by the system than `openssl` today. We also don't use the system `openssl`, ever. Anyway, is shipped, so I'll let the discussion die. Not like this isn't FOSS where I can't just go undo things that bother me.

PR #8039, opened 2016-12-19
Remove ENV.libxml2 no-op.

MikeMcQuaid:

-----

With superenv (used on all these formulae) this now does nothing.


MikeMcQuaid:
Some failures but none related to this change (which is removing a no-op anyway) so pulling.

PR #8040, opened 2016-12-19
Remove LibreSSL options where OpenSSL is already used.

MikeMcQuaid:

-----

Formulae should not depend on both OpenSSL and LibreSSL (even
optionally). This is to avoid descending into madness where every
formulae that could use LibreSSL has to have option and switching logic.

Homebrew has standardised on OpenSSL and will do so everywhere that
LibreSSL is not a hard requirement.

CC @ilovezfs for thoughts.


ilovezfs:
Yeah I don't see any reason to try to maintain dual options for their own sake.

nijikon (approved):

MikeMcQuaid:
Fixed `h2o` style failure and pushed.

tayler1:
Doh. You don't need to maintain both libraries.
Libressl is already bundled into h2o source.

PR #8053, opened 2016-12-20
zstd 1.1.2

nijikon:

-----

### DIFF
--- a/Formula/zstd.rb
+++ b/Formula/zstd.rb
@@ -1,8 +1,8 @@
 class Zstd < Formula
   desc "Zstandard is a real-time compression algorithm"
   homepage "http://zstd.net/"
-  url "https://github.com/facebook/zstd/archive/v1.1.1.tar.gz"
-  sha256 "a92965a2480fecc6b7bf491106e93cbf69a298b865b6d623755b3a482d77e614"
+  url "https://github.com/facebook/zstd/archive/v1.1.2.tar.gz"
+  sha256 "980b8febb0118e22f6ed70d23b5b3e600995dbf7489c1f6d6122c1411cdda8d8"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

apjanke:
Audit failed due to a non-executable file in `bin`.

```
failed: brew audit zstd --online

Stacktrace

        Error: 1 problem in 1 formula
zstd:
  * Non-executables were installed to "/usr/local/Cellar/zstd/1.1.2/bin"
    The offending files are:
      /usr/local/Cellar/zstd/1.1.2/bin/zstdless
```

Looks like a broken script in the installation? It doesn't have an executable permissions bit or a shebang line.

```
[/usr/local/Cellar/zstd/1.1.2/bin]
$ ls -l zstdless
-r--r--r-- 1 janke admin 18 Dec 22 17:05 zstdless
[/usr/local/Cellar/zstd/1.1.2/bin]
$ cat zstdless
zstdcat $@ | less
```

What's weird is that `zstdless` is `+x` in the source distribution and is installed with `@install -m 755 zstdless $(DESTDIR)$(BINDIR)/zstdless`, which seems like the right permissions.

Seems like a bug that should be reported upstream.

apjanke:
Interestingly, I installed it from outside `brew`, and `zstdless` kept its `+x` permissions. Is there something in `brew install` that modifies permissions of files based on whether they have a shebang or something similar?

apjanke:
Reported the lack of shebang upstream: https://github.com/facebook/zstd/issues/496

But I don't think this is actually an upstream issue, since `zstdless` seems to be losing its `+x` bit specifically under `brew`. I don't know why. Will look in to it when I have some time.

tdsmith:
>  Is there something in brew install that modifies permissions of files based on whether they have a shebang or something similar?

Yes, the cleaner does this.

apjanke:
Okay, I see [where the cleaner does this](https://github.com/Homebrew/brew/blob/8f85eb64c4875fed514150dec755cb267a96df74/Library/Homebrew/cleaner.rb#L103-L115). Thanks.

I submitted [an upstream patch](https://github.com/facebook/zstd/pull/497) to add a shebang. We could pull that in and get it working. Or wait to see if it's merged in a near-future version.

ilovezfs:
:+1: to merging it with your fix and not waiting on upstream. We broke it, we bought it.

nijikon:
@apjanke Nice work, added the patch, waiting for CI.

PR #8055, opened 2016-12-20
folly: fix reference to gflags .pc file

### LINKED ISSUE #7977: Folly is not usable with PKG_CHECK_MODULES due to missing libgflags dependency
I am trying to develop my own software which depends on Facebook's Folly library. The Homebrew-installed version of Folly (as of today: `folly: stable 2016.12.12.00 (bottled), HEAD`) is not findable with pkg-config (specifically, the PKG_CHECK_MODULES autoconf macro). When using PKG_CHECK_MODULES with libfolly, pkg-config fails because it claims to be missing libgflags:

```
# The following command is exactly what PKG_CHECK_MODULES will do:
$ pkg-config --exists --print-errors "libfolly >= 0.57"
Package libgflags was not found in the pkg-config search path.
Perhaps you should add the directory containing `libgflags.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libgflags', required by 'libfolly', not found
```

(To reproduce this issue, first `brew install folly`, and then run the `pkg-config` command shown above, which should succeed with no output.)

This is because the libfolly.pc installed in /usr/local/lib depends on "libgflags":

```
$ cat /usr/local/lib/pkgconfig/libfolly.pc
prefix=/usr/local/Cellar/folly/2016.12.12.00
...
Requires.private:  libgflags libglog openssl libevent
...
```

Yet the gflags pkg-config file is named "gflags.pc", not "libgflags.pc":

```
$ brew list gflags
...
/usr/local/Cellar/gflags/2.2.0/lib/pkgconfig/gflags.pc
...
```

I'm not sure whether the folly or the gflags recipe is at issue here, but they don't agree. In Homebrew's folly.rb file, I see that we are explicitly jamming dependencies into folly's configuration:

```sh
      # Build system relies on pkg-config but gflags removed
      # the .pc files so now folly cannot find without flags.
      ENV["GFLAGS_CFLAGS"] = Formula["gflags"].opt_include
      ENV["GFLAGS_LIBS"] = Formula["gflags"].opt_lib
```

And there are no similar alterations that I can see in the gflags recipe. But while the above helps Folly to build, it's still broken with respect to anything else that *depends* on Folly.

I'm not sure what the right course of action here is -- is this a Homebrew issue or an issue with the upstream Folly source?

mistydemeo:
Fixes #7977.

-----

cc @josh-berry

### DIFF
--- a/Formula/folly.rb
+++ b/Formula/folly.rb
@@ -55,10 +55,9 @@ class Folly < Formula
         inreplace "portability/Time.h", "typedef uint8_t clockid_t;", ""
       end
 
-      # Build system relies on pkg-config but gflags removed
-      # the .pc files so now folly cannot find without flags.
-      ENV["GFLAGS_CFLAGS"] = Formula["gflags"].opt_include
-      ENV["GFLAGS_LIBS"] = Formula["gflags"].opt_lib
+      # Fixes the .pc file, which references the gflags .pc under the wrong name.
+      # Applied upstream: https://github.com/facebook/folly/pull/531
+      inreplace "configure.ac", "[libgflags]", "[gflags]"
 
       system "autoreconf", "-fvi"
       system "./configure", "--prefix=#{prefix}", "--disable-silent-rules",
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

PR #8064, opened 2016-12-20
handbrake 1.0.0 (new formula)

JMoVS:

-----
~~Release version will be tagged in the coming days.~~

~~UPDATE: Release version has been tagged. Now awaiting buildable source tar ball from devs~~

**UPDATE 2: Ready to be merged now when the testbot doesn't complain. Please also consider merging https://github.com/caskroom/homebrew-cask/pull/28232 in one go then**

~~Please wait before merging to give the handbrake devs a chance to look over it.~~

### DIFF
--- a/Formula/handbrake.rb
+++ b/Formula/handbrake.rb
@@ -1,8 +1,7 @@
 class Handbrake < Formula
   desc "Open-source video transcoder available for Linux, Mac, and Windows."
   homepage "https://handbrake.fr/"
-  url "https://github.com/HandBrake/HandBrake.git", :revision => "d6e2e09ee7c00d9ed371d0a9b16ca87b6ee417e1"
-  version "1.0.0-alpha1"
+  url "https://github.com/HandBrake/HandBrake.git", :revision => "3f4a076243d2821ae78bc97aaed0ca47301d0d65", :tag => "1.0.0"
   head "https://github.com/HandBrake/HandBrake.git"
 
   depends_on "autoconf" => :build
--- a/Formula/handbrake.rb
+++ b/Formula/handbrake.rb
@@ -1,7 +1,9 @@
 class Handbrake < Formula
   desc "Open-source video transcoder available for Linux, Mac, and Windows."
   homepage "https://handbrake.fr/"
-  url "https://github.com/HandBrake/HandBrake.git", :revision => "3f4a076243d2821ae78bc97aaed0ca47301d0d65", :tag => "1.0.0"
+  url "https://github.com/HandBrake/HandBrake.git",
+      :tag => "1.0.0",
+      :revision => "3f4a076243d2821ae78bc97aaed0ca47301d0d65"
   head "https://github.com/HandBrake/HandBrake.git"
 
   depends_on "autoconf" => :build
--- a/Formula/handbrake.rb
+++ b/Formula/handbrake.rb
@@ -1,9 +1,8 @@
 class Handbrake < Formula
   desc "Open-source video transcoder available for Linux, Mac, and Windows."
   homepage "https://handbrake.fr/"
-  url "https://github.com/HandBrake/HandBrake.git",
-      :tag => "1.0.0",
-      :revision => "3f4a076243d2821ae78bc97aaed0ca47301d0d65"
+  url "https://handbrake.fr/rotation.php?file=HandBrake-1.0.0.tar.bz2"
+  sha256 "15fb4593c70d75621212e4499f018c3c93d7ce39f1083bf527d4616ded0044c5"
   head "https://github.com/HandBrake/HandBrake.git"
 
   depends_on "autoconf" => :build
--- a/Formula/handbrake.rb
+++ b/Formula/handbrake.rb
@@ -1,0 +1,26 @@
+class Handbrake < Formula
+  desc "Open-source video transcoder available for Linux, Mac, and Windows."
+  homepage "https://handbrake.fr/"
+  url "https://github.com/HandBrake/HandBrake.git", :revision => "d6e2e09ee7c00d9ed371d0a9b16ca87b6ee417e1"
+  version "1.0.0-alpha1"
+  head "https://github.com/HandBrake/HandBrake.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "cmake" => :build
+  depends_on "libtool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "yasm" => :build
+
+  def install
+    system "./configure", "--prefix=#{prefix}",
+                          "--disable-xcode",
+                          "--disable-gtk"
+    system "make", "-C", "build"
+    system "make", "-C", "build", "install"
+  end
+
+  test do
+    system bin/"HandBrakeCLI", "--help"
+  end
+end


Formula/handbrake.rb
```diff
@@ -0,0 +1,28 @@
...
+  url "https://github.com/HandBrake/HandBrake.git", :revision => "72e77efa1c309be6a5e525a9484ec131707af848"
+  version "1.0.0-alpha1"
+  head "https://github.com/HandBrake/HandBrake.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "cmake" => :build
+  depends_on "libtool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "yasm" => :build
```
MikeMcQuaid: Definitely no linkage here?

Formula/handbrake.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def install
+    system "./configure", "--debug=none",
+                          "--disable-xcode",
+                          "--disable-gtk"
+    cd "build" do
+      system "make"
+      bin.install "HandBrakeCLI"
```
MikeMcQuaid: No `make install` we can use?

Formula/handbrake.rb
JMoVS: yes. I chatted with @ilovezfs about it. They have crazy upstream projects (where they do report bugs to that regularly break stuff in minor point updates that are not detectable in tests (at east automatically). I made sure and provided a pull to them which got pulled in by now to at least verify the downloads of the libs via sha256

ilovezfs:
```
iMac-TMP:~ joe$ brew linkage handbrake
System libraries:
  /System/Library/Frameworks/AudioToolbox.framework/Versions/A/AudioToolbox
  /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit
  /usr/lib/libSystem.B.dylib
  /usr/lib/libbz2.1.0.dylib
  /usr/lib/libc++.1.dylib
  /usr/lib/libiconv.2.dylib
  /usr/lib/libobjc.A.dylib
  /usr/lib/libz.1.dylib
```

Formula/handbrake.rb
JMoVS: I don't think so. cc: @ilovezfs  Last time I checked, didn't seem like it. Referring to: https://handbrake.fr/docs/en/latest/developer/build-mac.html

ilovezfs: Nope. Building without `--disable-xcode` gives you a `make install` that sticks Handbrake.app in /tmp/HandBrake.dst/Applications and HandBrakeCLI in /tmp/HandBrake.dst/usr/local/bin/HandBrakeCLI, which is useless.

And building with `--disable-xcode` gives you a `make install` that does nothing.

bradleysepos: HandBrake developer. Will see about fixing `make install` on Mac, thanks.

JMoVS: Don't know, whether I'm the first one to make this joke, but seriously:

I see something good is brewing up (the joke works in German at least...)

ilovezfs:
@BrewTestBot test this please

Formula/handbrake.rb
bradleysepos: https://github.com/bradleysepos/HandBrake/commit/6ccbfa43e09d1c8c22c4a1c2749a5bf4b6e017e2

ilovezfs:
updated formula to use `--prefix=#{prefix}` and `make install`

Formula/handbrake.rb
```diff
@@ -0,0 +1,26 @@
+class Handbrake < Formula
+  desc "Open-source video transcoder available for Linux, Mac, and Windows."
+  homepage "https://handbrake.fr/"
+  url "https://github.com/HandBrake/HandBrake.git", :revision => "d6e2e09ee7c00d9ed371d0a9b16ca87b6ee417e1"
+  version "1.0.0-alpha1"
```
rakkesh: current stable version is `0.10.5`.

bradleysepos: HandBrake 1.0.0 will be released in a matter of days. I believe the intent is to update the formula to the new release before merging.

vitorgalvao:
Any reason **not** to call it `Handbrakelci`/`handbrakecli.rb`? It is the correct name of the tool, after all.

ilovezfs:
Perhaps the cask should be handbrake-app or handbrake-gui.

adidalal:
It will be `handbrake-app`, if functionally different from the CLI-only Formula.

ilovezfs:
@adidalal @vitorgalvao basically the choices seem to be

both named handbrake
cask named handbrake, formula named handbrakecli (or handbrake-cli)
cask named handbrake-app, formula named handbrake
cask named handbrake-app, formula named handbrakecli (or handbrake-cli)

JMoVS:
I am in for:
Formula: handbrake
Cask: handbrake-app

vitorgalvao:
I see no reason why we should rename the cask. **The GUI is the main thing**. The CLI tool is *specifically* called `CLI`. Our documented case for renaming is reasonable: when the GUI is *just* a dumb wrapper *on the more important* CLI. Not at all the case, here.

In this specific case, making the cask have `app` and the formula not have `cli` is downright contradicting the names given by the developer and having them both have wrong names for no reason.

bradleysepos:
The two most difficult things in software development: cache invalidation and naming things and off-by-one errors.

ilovezfs:
I'd much rather do `brew install handbrake`.

ilovezfs:
If for example we add an option to build the GUI (i.e., everything exactly the same but without the `--disable-xcode`) it will look ridiculous to do this:
```
brew install --with-gui handbrake-cli
```

bradleysepos:
I like the idea of `brew install [--with-gui] handbrake`.

JMoVS:
I am not familiar enough with homebrew in general.

But when I install anything from main homebrew, I **never ever** expect it to have a GUI. I'd only expect it to install a GUI if I explicitly somehow mention the GUI in the install process - which leads to my proposal of "handbrake" for the formula and "handbrake-app" for the cask.

But I'm not familar enough with homebrew. Maybe my expectations are wrong ;-)

ilovezfs:
your expectation is correct. if there is a gui that is optional, it is almost always either not available at all or available as a non-default option.

vitorgalvao:
> I'd much rather do `brew install handbrake`.

I can say the same about every cask where we added the `app` suffix.

**I want to be absolutely clear this is not a jab**. But personal preference is not an argument for what is best. If it were we would haver never discussed this rule.

I agree with your point about the name if adding the option to build the GUI, so lets leave both with the same name (it’d still be nonsensical to rename the cask in this case).

JMoVS:
@vitorgalvao I am not super familiar with casks, but the impression I got was it was more ore less clearly for GUI apps. So at least to me (as a user), it wouldn't seem confusing to install handbrake (the app) via cask and install handbrake (the cli) via homebrew-core.

ilovezfs:
>But personal preference is not an argument for what is best.

Of course it is. https://plato.stanford.edu/entries/hume-aesthetics/#HumeEssaTast

vitorgalvao:
> I am not super familiar with casks, but the impression I got was it was more ore less clearly for GUI apps.

It’s also for closed-source CLI apps, fonts, prefpanes, audio plugins, and more. A more accurate description would be “almost anything that is not an open-source CLI tool”.

> So at least to me (as a user), it wouldn't seem confusing to install handbrake (the app) via cask and install handbrake (the cli) via homebrew-core.

And I wish every other user understood that as clearly. Unfortunately, they don’t (this is not speculation), which is why we need these differentiating rules. If that weren’t the case, *zero* casks would have the `-app` suffix.

vitorgalvao:
> > But personal preference is not an argument for what is best.

> Of course it is. https://plato.stanford.edu/entries/hume-aesthetics/#HumeEssaTast

Let me expand and refine the argument to what I meant: “Regarding matters of naming a user-facing (the user will see/type/interact with/need to understand) something on a largely-user-centric software project, what is “best” can be defined as what is more logical and clearer to the user. From there it follows a single maintainer’s personal preference is in no way representative of what is best (as defined previously) to said users (non-maintainers)”.

To put it in ridiculous terms, if my personal preference were to call this formula `travao-de-mao` (Portuguese for “handbrake”) because I am Portuguese, my personal preference would not be an argument ([technically](https://youtu.be/hou0lU8WMgo) it would still be an argument, just a bad one) for what is best to this hugely-non-Portuguese user base.

ilovezfs:
I'd definitely be in favor of naming it `travao-de-mao`

JMoVS:
@vitorgalvao I think, if it's unclear for users what a cask is, then the current cask should be renamed to handbrake-gui or handbrake-app.

Though I don't see a need for clarification for homebrew (classic?) users who are already on the command line and are using a package manager to install command line tools.

Should I open a PR on casks to rename handbrake to handbrake-app, bump the version as well?

vitorgalvao:
> I think, if it's unclear for users what a cask is, then the current cask should be renamed to handbrake-gui or handbrake-app.

> Though I don't see a need for clarification for homebrew (classic?) users who are already on the command line and are using a package manager to install command line tools.

For some users the difference between a formula and a cask is sometimes unclear. But you are deeply wrong in assuming the default is knowing what the formula is. A huge number of our users use HBC *exclusively*. Many of those never type `brew` without `cask`, and they even preferred if we didn’t remove some of the casks in favour of the formulas. For them, it’s just easier and makes more sense to use HBC.

It is frequent to have people submit CLI open-source tools to HBC, to which I tell them (so much so it’s an item in the FAQ and I have an expanding snippet) to try HB first. A number of those don’t even want to do it, because casks are easier to make to them, and they don’t even know where to start with a formula..

> Should I open a PR on casks to rename handbrake to handbrake-app, bump the version as well?

No. Like I said, the GUI is the main thing. We’re not going to rename it. Same argument for `transmission`, for example. In that case it’s even stronger, actually, as I’m willing to bet the vast majority know about the GUI but not the CLI.

vitorgalvao:
As a bit of a side note, we can stop having the `-gui`. We’re never going to make it `-gui`. `-app` has been decided on and is [clearly documented](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md#with-a-formula).

JMoVS:
@vitorgalvao quoting your linked documentation:

> GUI app is simply a wrapper of the CLI tool but has been decided to provide enough value to warrant the inclusion of both the formula and the cask

sounds to me like handbrake-gui (which one _could_ interprete as a wrapper for the cli version) should be named -app all along

vitorgalvao:
> sounds to me like handbrake-gui (which one could interprete as a wrapper for the cli version) should be named -app all along

And yet, in the official tools, the CLI version is the one where it was considered necessary to add the suffix, not the GUI version. That rule is for cases where the GUI is an afterthought, not **the main advertised version on the website**.

There’s no point in discussing the meaning of that rule with me, as I’m the one who wrote it (in conversation with @adidalal and @MikeMcQuaid) as I did most of that document. Regardless of how it is written, I know the spirit of it. You may argue the rule as written is unclear, but arguing I’m misinterpreting does not hold.

JMoVS:
Well, the more I'm thinking about it -- why even at all have the app version in the cask? The standalone app has a sparkle updater and is available through handbrake, as @bradleysepos can confirm. For users that want to use brew, I think the way forward is rather to make the option "--with-GUI" work in main homebrew rather than have the cask...

Thoughts/Comments?

vitorgalvao:
> The standalone app has a sparkle updater

And HB [does not want self-updating apps](https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md#we-dont-like-tools-that-upgrade-themselves), so that would need to be removed when building it.

You touched on two points why many users prefer some casks as casks and don’t want them as formulae: a formula will take longer to install (needs to build the GUI) and cannot auto-update.

JMoVS:
ah ok. Good notes.

Then, considering all this, I still believe the best way is to then make both handbrake without any postfix in the name.

MikeMcQuaid:
> If for example we add an option to build the GUI

Let's never do this. I'd like Homebrew to be the place where the only GUIs we build are those that upstream are never going to turn into a .app (e.g. GTK apps).

> That rule is for cases where the GUI is an afterthought, not the main advertised version on the website.

👍 

> why even at all have the app version in the cask? The standalone app has a sparkle updater and is available through handbrake

This applies to many/most casks. Homebrew Cask has value because it allows one to script installation of any given package.



> Then, considering all this, I still believe the best way is to then make both cask and formula called handbrake without any postfix in the name.

I agree.

ilovezfs:
>Let's never do this.

Why? It's trivial to add and some users will want it.

MikeMcQuaid:
> Why?

Homebrew does not work well for GUI applications and even worse for .apps. Homebrew was designed for command-line tools. Homebrew Cask (which is part of Homebrew itself, now) handles these better and uses upstream supported binaries which both works better for end-users and reduces our support burden. Users who really really still want to build it from source can do so in a third-party tap.

ilovezfs:
I'm not happy with the idea that the "policy" is shifting from GUI options are cool if they're non-default to GUI options are evil if there's a cask. That really reduces the usefulness of Homebrew.

MikeMcQuaid:
> I'm not happy with the idea that the "policy" is shifting from GUI options are cool if they're non-default to GUI options are evil if there's a cask. That really reduces the usefulness of Homebrew.

That's been the written policy since (almost or literally) Day 1. We had very minimal or no complaints when moving a bunch of things to Homebrew Cask and a further reduced support burden. Good design involves setting scope appropriately. We do not have infinite time or people so we need to constrain scope. Again, nothing is to stop anyone from supporting such things in non-Homebrew organisation taps if they choose to do so.

ilovezfs:
Why do we still have the emacs cocoa option then?

MikeMcQuaid:
> ### Sometimes there are exceptions
> 
> ... Documentation tends to lag behind current decision-making. Although some rejections may seem arbitrary or strange they are based from years of experience making Homebrew work acceptably for our users.

https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md#sometimes-there-are-exceptions

In that specific case: because there's no officially supported upstream binary provided for the Cocoa option.

PR #8129, opened 2016-12-22
meson: don't use virtualenv

ilovezfs:


aleb:
Currently meson fails when started from a python script. This commit fixes the problem.

Steps to reproduce:
```
➜  git clone https://github.com/pitivi/gst-transcoder.git
➜  cd gst-transcoder
➜  ./configure
```

Until now no problem can be noticed, but you can check that the `mesonbuild/build.ninja` file generated by meson in this step contains `/usr/local/bin/python3`: 
```
➜  grep python3 mesonbuild/build.ninja
 command = "/usr/local/bin/python3" "/usr/local/Cellar/meson/0.37.1/bin/meson" --internal symbolextractor $in $out $CROSS
```

Acording to folks in #mesonbuild, this means that `/usr/local/bin/meson`, which is a python script, has been executed with `/usr/local/bin/python3` instead of `/usr/local/Cellar/meson/0.37.1/libexec/bin/python3` which is specified in its shebang line!

Because of this, when the generated build system is used to actually build the thing, it fails:
```
$ make
cd /Users/aleb/tmp/gst-transcoder/mesonbuild && ninja
[8/9] Generating symbol file 'gsttrans...a/libgsttranscoder-1.0.dylib.symbols'.
FAILED: gsttranscoder-1.0@sha/libgsttranscoder-1.0.dylib.symbols 
"/usr/local/bin/python3" "/usr/local/Cellar/meson/0.37.1/bin/meson" --internal symbolextractor libgsttranscoder-1.0.dylib 'gsttranscoder-1.0@sha/libgsttranscoder-1.0.dylib.symbols' 
/usr/local/opt/python3/bin/python3.5
  File "/usr/local/Cellar/meson/0.37.1/bin/meson", line 18, in <module>
    from mesonbuild import mlog, mesonmain, mesonlib
ImportError: cannot import name 'mlog'
make: *** [all] Error 1
```

Notice "/usr/local/Cellar/meson/0.37.1/bin/meson" then fails because it's run with the homebrew python3 instead of the python3 in its virtual env.

Also notice it prints "/usr/local/opt/python3/bin/python3.5", because I added the following two lines after the shebang line in `/usr/local/bin/meson`:
```
#!/usr/local/Cellar/meson/0.37.1/libexec/bin/python3
import sys
print(sys.executable)
```



Trying to minimize the steps for reproducing the problem, I got to this:
```
➜  cd gst-transcoder/mesonbuild
➜  python3 -c 'import subprocess; subprocess.check_call(["/usr/local/bin/meson", "../gst-transcoder"])'     
/usr/local/bin/python3
[...]
➜  python3 -c 'import subprocess; subprocess.check_call("/usr/local/bin/meson ../gst-transcoder", shell=True)'
/usr/local/bin/python3
[...]
```

Normally these should print "/usr/local/Cellar/meson/0.37.1/libexec/bin/python3", because that's what's in the shebang line.

In #python I was told (by Yhg1s) that "Python really just does execve() with the strings you give it. Its own executable isn't somehow involved." Do you think it's a bug in execve? @tdsmith

tdsmith:
The build.ninja output means sys.executable is bad during configuration: https://github.com/mesonbuild/meson/blob/a2528a881640913dfc71fab7f15225d7e7d9a567/mesonbuild/backend/ninjabackend.py#L1288

Annnnnnnnd: sys.executable is not set correctly when invoking a virtualenv stub as a subprocess of another Python process.

cf:
```
$ virtualenv -p python3 test
$ test/bin/python3 -c 'import sys; print(sys.executable)'
/Users/tim/test/bin/python3
$ /usr/local/bin/python3 -c 'import subprocess; subprocess.call(["/Users/tim/test/bin/python3", "-c", "import sys; print(sys.executable)"])'
/usr/local/bin/python3
$ /usr/local/bin/python3 -c 'import subprocess, os; del os.environ["__PYVENV_LAUNCHER__"]; subprocess.call(["/Users/tim/test/bin/python3", "-c", "import sys; print(sys.executable)"])'
/Users/tim/test/bin/python3
```

This is https://github.com/pypa/virtualenv/issues/845, which I am sad to report relates to https://bugs.python.org/issue22490, which I had hoped I had washed my hands of.

tdsmith:
Anyway, yeah, sure, let's not use a virtualenv here.

ilovezfs:
Shipped.

PR #8130, opened 2016-12-22
lasso 2.5.1 (new formula)

haegrr:

-----

[Lasso](http://lasso.entrouvert.org) implements SAML 2.0 and related protocols.

**Note**: the core C library comes with a bunch of language bindings. I gave up trying to get them to install properly, and thus they are all disabled.

### DIFF
--- a/Formula/lasso.rb
+++ b/Formula/lasso.rb
@@ -1,0 +1,40 @@
+class Lasso < Formula
+  desc "Library for Liberty Alliance and SAML protocols"
+  homepage "http://lasso.entrouvert.org"
+  url "https://dev.entrouvert.org/releases/lasso/lasso-2.5.1.tar.gz"
+  sha256 "be105c8d400ddeb798419eafa9522101d0f63dc42b79b7131b6010c4a5fc2058"
+
+  depends_on "pkg-config" => :build
+  depends_on "libxmlsec1"
+  depends_on "glib"
+  depends_on "openssl"
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--disable-java",
+                          "--disable-perl",
+                          "--disable-php5",
+                          "--disable-python",
+                          "--prefix=#{prefix}",
+                          "--with-pkg-config=#{ENV["PKG_CONFIG_PATH"]}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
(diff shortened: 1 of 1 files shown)


Formula/lasso.rb
```diff
@@ -0,0 +1,42 @@
...
+      #include <lasso/lasso.h>
+
+      int main() {
+        return lasso_init();
+      }
+    EOS
+    system ENV.cc, "test.c",
+                   "-I#{Formula["glib"].include}/glib-2.0",
+                   "-I#{Formula["glib"].lib}/glib-2.0/include",
+                   "-I/usr/include/libxml2",
```
bfontaine: Shouldn’t that use `Formula["libxml2"].include`?

Formula/lasso.rb
```diff
@@ -0,0 +1,42 @@
+class Lasso < Formula
+  desc "Library for Liberty Alliance and SAML protocols"
+  homepage "http://lasso.entrouvert.org"
+  url "https://dev.entrouvert.org/releases/lasso/lasso-2.5.1.tar.gz"
+  sha256 "be105c8d400ddeb798419eafa9522101d0f63dc42b79b7131b6010c4a5fc2058"
+
+  depends_on "pkg-config" => :build
+  depends_on "libxml2"
```
MikeMcQuaid: Can this use the system `libxml2`?

Formula/lasso.rb
MikeMcQuaid: Yes but ideally it'll just use the system version.

MikeMcQuaid (changes requested):

Formula/lasso.rb
protomouse: Yep. Also libxslt.

Formula/lasso.rb
ilovezfs: Without the CLT,
```
robotunicorn # ls /usr/include                                              [~]
ls: /usr/include: No such file or directory
```
so I think we'd need to vendor the header if it needs it to use the system version, no?

ilovezfs: Ah it's in the SDK too, so I think we want

```
"-I#{MacOS.sdk_path}/usr/include/libxml2"
```

```
"-I#{MacOS.sdk_path}/usr/include/libxslt"
```
unless it's already working without adding anything.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #8138, opened 2016-12-22
rust 1.14.0

alex:

-----

### DIFF
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -3,12 +3,12 @@ class Rust < Formula
   homepage "https://www.rust-lang.org/"
 
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.13.0-src.tar.gz"
-    sha256 "ecb84775ca977a5efec14d0cad19621a155bfcbbf46e8050d18721bb1e3e5084"
+    url "https://static.rust-lang.org/dist/rustc-1.14.0-src.tar.gz"
+    sha256 "c790edd2e915bd01bea46122af2942108479a2fda9a6f76d1094add520ac3b6b"
 
     resource "cargo" do
       # git required because of submodules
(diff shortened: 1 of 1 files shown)


ilovezfs:
also llvm appears to be broken, though breakage apparently is preexisting and wasn't caught on the prior release: https://gist.github.com/ilovezfs/1eb14b90948d2079bb6e5abdaeac6a5f

ilovezfs:
CC @alexcrichton any ideas on the CI breakage https://bot.brew.sh/job/Homebrew%20Core/13572/version=sierra/console and/or the `--with-llvm` breakage https://gist.github.com/ilovezfs/1eb14b90948d2079bb6e5abdaeac6a5f ?

yurikoles:
It doesn't build. I tried same changes myself, it fails with message CARG_CFG not found.

alexcrichton:
@ilovezfs I think the CI breakage is relevant here:

```
```

To build Cargo you'll need a previous version of Cargo installed somewhere to build with. For example the previously packaged Cargo should work ok to build the next Cargo.

For the latter it looks like LLVM was linked against libffi and Rust didn't detect that. I think that may be a case where we don't detect that library should be linked (or a bug in llvm-config, I forget). I'd recommend not compiling against the system LLVM, however, as using the upstream version is guaranteed to have the most bugfixes.

ilovezfs:
@alexcrichton cool. Thanks for the help. Note that the LLVM is the upstream LLVM, specifically the homebrew llvm formula (https://github.com/Homebrew/homebrew-core/blob/master/Formula/llvm.rb):
https://github.com/alex/homebrew-core/blob/c4217951cac69bdd25a1a3bb40206ce804b1cde9/Formula/rust.rb#L56

nijikon:
@BrewTestBot test this please

ilovezfs:
Closing in favor of #8708 but thanks for the PR nonetheless.

PR #8161, opened 2016-12-23
python 3.6.0

skycaptain:

-----

### DIFF
--- a/Formula/python3.rb
+++ b/Formula/python3.rb
@@ -1,32 +1,14 @@
 class Python3 < Formula
   desc "Interpreted, interactive, object-oriented programming language"
   homepage "https://www.python.org/"
-  revision 3
-
+  url "https://www.python.org/ftp/python/3.6.0/Python-3.6.0.tar.xz"
+  sha256 "b0c5f904f685e32d9232f7bdcbece9819a892929063b6e385414ad2dd6a23622"
   head "https://hg.python.org/cpython", :using => :hg
-
-  stable do
-    url "https://www.python.org/ftp/python/3.5.2/Python-3.5.2.tar.xz"
-    sha256 "0010f56100b9b74259ebcd5d4b295a32324b58b517403a10d1a2aa7cb22bca40"
-
-    # Patch for pyport.h macro issue
-    # https://bugs.python.org/issue10910
-    # https://trac.macports.org/ticket/44288
-    patch do
-      url "https://bugs.python.org/file30805/issue10910-workaround.txt"
-      sha256 "c075353337f9ff3ccf8091693d278782fcdff62c113245d8de43c5c7acc57daf"
-    end
-  end
 
   bottle do
     sha256 "4a43600ceb2875c13200ace82fea1a9a119973b831f5503fa57cf8db44fbb155" => :sierra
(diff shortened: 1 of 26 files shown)


tdsmith:
Thanks for this! I'm inclined to bump the devel spec to 3.6.0 but leave stable on the 3.5 branch until there are numpy, scipy, and cryptography wheels for 3.6, since that seemed to be the pain point when we transitioned from 3.4 to 3.5, but I'm open to feedback.

ilovezfs:
Nope, we should ship 3.6.0 as stable. Upstream has designated as such, and that's what people will want for the bottles.

tdsmith:
Yes, we should and we will; the question in my mind, which Mike raised after the 3.5 transition, is whether it's helpful to do that immediately, before the packaging ecosystem has caught up.

ilovezfs:
I don't see this as remotely controversial. If 3.6 isn't break anything in Homebrew: :ship:

tdsmith:
With respect to the instant PR, all of the failing packages will need to have a revision bump.

ilovezfs:
squashed, and revision bumps added.

pirate:
Pardon me for asking if this isn't the right place, but why the revision bumps on other packages?  Is it so they will upgrade and reinstall the python3 dependency in the process in order to get python3.6?

ilovezfs:
@pirate it forces those formulae to be rebuilt from source even though their own versions aren't changing, and then triggers upgrades for users to those newly built bottles.

alex:
As someone who makes wheels; I'm +1 on homebrew upgrading and breaking us; without users complaining to us there's a real chance we'll forget.

MikeMcQuaid:
I'm personally fine shipping this as-is but if @tdsmith would rather wait until other Python stuff is sorted I'm fine with that too. I definitely think if we're going to hold off on those packages being ready that they should be moved into Homebrew/homebrew-core.

skycaptain:
Remaining errors in checks are imho innocuous and are not related to this pr.

Formula/python3.rb
```diff
@@ -47,17 +29,17 @@ class Python3 < Formula
   depends_on :x11 if build.with?("tcl-tk") && Tab.for_name("homebrew/dupes/tcl-tk").with?("x11")
   depends_on "sphinx-doc" => [:build, :optional]
 
-  skip_clean "bin/pip3", "bin/pip-3.4", "bin/pip-3.5"
-  skip_clean "bin/easy_install3", "bin/easy_install-3.4", "bin/easy_install-3.5"
+  skip_clean "bin/pip3", "bin/pip-3.4", "bin/pip-3.5", "bin/pip-3.6"
+  skip_clean "bin/easy_install3", "bin/easy_install-3.4", "bin/easy_install-3.5", "bin/easy_install-3.6"
 
   resource "setuptools" do
```
zmwangx: I realize this won't be a popular opinion for people waiting for this PR to land, but I think setuptools should be upgraded to 32.2.0, or at least 32.1.0 to match the version in [`python.rb`](https://github.com/Homebrew/homebrew-core/blob/7330a126bf338b5216dc3b31bb65485cbedbbbf1/Formula/python.rb#L44-L47).

ilovezfs: done.

ilovezfs:
@skycaptain thanks for your speedy pull request and your contribution to Homebrew! :snake:

PR #8220, opened 2016-12-25
ucg 0.3.1 (new formula)

equal-l2:

-----

About `env :std`, see gvansickle/ucg#110.

### DIFF
--- a/Formula/ucg.rb
+++ b/Formula/ucg.rb
@@ -1,8 +1,8 @@
 class Ucg < Formula
   desc "grep-like tool for searching large bodies of source code"
   homepage "https://github.com/gvansickle/ucg"
-  url "https://github.com/gvansickle/ucg/releases/download/0.3.2/universalcodegrep-0.3.2.tar.gz"
-  sha256 "df57c877164454a5001caf791a24cd119afe4196070cb8d3cc741a7a43805c3e"
+  url "https://github.com/gvansickle/ucg/releases/download/0.3.3/universalcodegrep-0.3.3.tar.gz"
+  sha256 "116d832bbc743c7dd469e5e7f1b20addb3b7a08df4b4441d59da3acf221caf2d"
   head "https://github.com/gvansickle/ucg.git"
 
   depends_on "pkg-config" => :build
@@ -24,7 +24,6 @@ class Ucg < Formula
 
   test do
     (testpath/"test.txt").write("Hello World!")
-    lines = shell_output("#{bin}/ucg 'Hello World' #{testpath}").strip.split(":")
-    assert_equal "Hello World!", lines[2]
+    assert_match "Hello World!", shell_output("#{bin}/ucg 'Hello World' #{testpath}")
   end
 end
--- a/Formula/ucg.rb
+++ b/Formula/ucg.rb
@@ -27,7 +27,7 @@ class Ucg < Formula
 
   test do
     (testpath/"test.txt").write("Hello World!")
-    lines = `#{bin}/ucg 'Hello World' #{testpath}`.strip.split(":")
+    lines = shell_output("#{bin}/ucg 'Hello World' #{testpath}").strip.split(":")
     assert_equal "Hello World!", lines[2]
   end
 end
--- a/Formula/ucg.rb
+++ b/Formula/ucg.rb
@@ -1,0 +1,33 @@
+class Ucg < Formula
+  desc "grep-like tool for searching large bodies of source code"
+  homepage "https://github.com/gvansickle/ucg"
+  url "https://github.com/gvansickle/ucg/releases/download/0.3.1/universalcodegrep-0.3.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/ucg.rb
```diff
@@ -0,0 +1,33 @@
...
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.txt").write("Hello World!")
+    lines = `#{bin}/ucg 'Hello World' #{testpath}`.strip.split(":")
```
bfontaine: Please use `shell_output("code")` instead of `` `code` ``.

bfontaine (changes requested):

Formula/ucg.rb
equal-l2: Fixed. Thanks.

bfontaine (approved):

Formula/ucg.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  depends_on "pkg-config" => :build
+  depends_on "autoconf" => :build
+  depends_on "libtool" => :build
+  depends_on "automake" => :build
+  depends_on "argp-standalone" => :build
+  depends_on "pcre2"
+
+  # superenv's '-march' flag conflicts with build script
+  env :std
```
MikeMcQuaid: We can't accept new formulae with `env :std`. This will need to be fixed so this works. Please remove it so I can help debug the failure. Thanks!

MikeMcQuaid (changes requested):

Formula/ucg.rb
equal-l2: This software won't build with `-march=` options, since it determines environment-specific optimize option by itself.

Is it able to get rid of `-march` by another way?

gvansickle: Hi @equal-l2 ,
I believe I've resolved the `-march` issue on upstream master as of 6467465abfbcdddef20a7bde4263e1479feb5264.  It should now build correctly with a `-march` in the CXX or CC env vars.  Can you try master/HEAD and confirm?  I'm also rolling a tarball for release as 0.3.2 which will pick up these changes.

gvansickle: Hi @equal-l2 ,

The 0.3.2 release is out in tarball form here: https://github.com/gvansickle/ucg/releases/tag/0.3.2
This release should not require the `env :std` to build under Homebrew.

Let me know how that tarball works; `travis-ci` is showing no problems on any of its OS X configurations, so I'm pretty confident this will go smoothly.

Thanks again @equal-l2 and everyone else,
GRVS

Formula/ucg.rb
```diff
@@ -0,0 +1,30 @@
...
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.txt").write("Hello World!")
+    lines = shell_output("#{bin}/ucg 'Hello World' #{testpath}").strip.split(":")
+    assert_equal "Hello World!", lines[2]
```
MikeMcQuaid: Test is failing: https://bot.brew.sh/job/Homebrew%20Core/13835/version=sierra/testReport/junit/brew-test-bot/sierra/test_ucg/

Use `assert_match`, don't `strip` or `split` or `[2]` (we don't need that level of precision and it may fail in future) and add a `1` second argument to `shell_output`.

MikeMcQuaid (changes requested):

Formula/ucg.rb
gvansickle: Ok @equal-l2 , third time's the charm:
https://github.com/gvansickle/ucg/releases/tag/0.3.3

This tarball fixes the SSE4.2 issue which was causing the test to fail.  I've added a regression test to the testsuite to catch anything like that in the future.  Travis-CI is showing all green for xcode6.4 through xcode8.2: https://travis-ci.org/gvansickle/ucg/builds/187776582

Thanks again, and thanks all for your efforts and patience,
GRVS

equal-l2: @MikeMcQuaid 
Thanks for the advice.
Using `assert_match` is right. I'll do so.
But return value should not be `1` (should be `0`), so we won't need the second argument.

@gvansickle 
Thanks for the fix again!

MikeMcQuaid: https://bot.brew.sh/job/Homebrew%20Core/13835/version=sierra/testReport/junit/brew-test-bot/sierra/test_ucg/ demonstrates that it is `1`, unfortunately.

equal-l2: @MikeMcQuaid 
New release 0.3.3 should fix it.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #8184, opened 2016-12-24
libgit2 0.25.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #8221, opened 2016-12-26
libgit2 0.25.0

JCount:

-----
**libgit 0.25.0**

-----

- **gitless 0.8.5** is a new version and should probably be put in a separate PR before this is merged

- **libgit2-glib** currently only has support through the libgit2-0.24.x version series

- **gnome-builder** and **gitg** depend on **libgit2-glib** so they are held up at least until libgit2-glib gets support for libgit2 0.25.0

-----
cc @ilovezfs since this originated from his #8184

### DIFF
--- a/Formula/gitless.rb
+++ b/Formula/gitless.rb
@@ -5,6 +5,7 @@ class Gitless < Formula
   homepage "http://gitless.com/"
   url "https://github.com/sdg-mit/gitless/archive/v0.8.5.tar.gz"
   sha256 "c93f8f558d05f41777ae36fab7434cfcdb13035ae2220893d5ee222ced1e7b9f"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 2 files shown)


tschoonj:
Doesn't look like libgit2 and its GNOME dependents will see an update soon. Could easily be March before this happens.

From the libgit2 website:

> libgit2 is a portable, pure C implementation of the Git core methods provided as a re-entrant linkable library with a solid API, allowing you to write native speed custom Git applications in any language which supports C bindings.

Not sure why they call it solid when they break backwards compatibility about once a year 😢

JCount:
Thanks for the info @tschoonj, I thought that something like that might be the case.

In that case, it might be worth doing a PR for a simple version bump for **gitless 0.8.5** since it is still compatible with the current **libgit2 0.24.5** formula. Any thoughts anyone?

ilovezfs:
Is there a GNOME issue about this yet?

JCount:
I couldn't find one, but honestly the GNOME bugzilla was making my head spin.

tschoonj:
I opened a bug: https://bugzilla.gnome.org/show_bug.cgi?id=776506

JCount:
Haha, @tschoonj I was about to hit enter on essentially the same bug. 😀

tschoonj:
You're welcome 😄 

Also, no objections here if you want to update `gitless` in a separate PR

JCount:
Ok, will do, especially since there is not roadmap for libgit2-glib.

ilovezfs:
@JCount OK, I've shipped https://github.com/Homebrew/homebrew-core/pull/8275 so you should be back in business here.

JCount:
@ilovezfs @tschoonj Looks like we are in business here.

ilovezfs:
sweet!

ilovezfs:
Shipped! Thanks @JCount

PR #8222, opened 2016-12-26
p11-kit 0.23.3

novel:
Update to a newer p11-kit version. The patch for p11-kit-proxy.dylib symlinking was merged upstream, so drop it from the formula.


nijikon:

PR #8229, opened 2016-12-26
p11-kit: add HEAD

novel:


ilovezfs (dismissed):
Instead of `if build.head?` the head-specific dependencies should be in a `head do` block, and the head url should be moved into that block as well.

novel:
Updated to add the "head do" block.

MikeMcQuaid:

PR #8230, opened 2016-12-26
apache-brooklyn-cli 0.10.0

ilovezfs:
use glide vendoring


neykov:
Thanks for the update @ilovezfs.

PR #8238, opened 2016-12-26
mono 4.6.2.16

mfilippov:
mono 4.6.2.16
I switch build from tar archive to github tag. Because mono team has delay between publish release and build tar archive.

### DIFF
--- a/Formula/mono.rb
+++ b/Formula/mono.rb
@@ -1,8 +1,7 @@
 class Mono < Formula
   desc "Cross platform, open source .NET development framework"
   homepage "http://www.mono-project.com/"
-  url "https://download.mono-project.com/sources/mono/mono-4.6.2.7.tar.bz2"
-  sha256 "b69ead7db7590277630685f3fcad84a5ac2f83e40563f0dc8cf139b10cbe6b31"
+  url "https://github.com/mono/mono.git", :tag => "mono-4.6.2.16", :revision => "ac9e222cd99a3ba55a3232598bd53bd3c397f03f"
 
   bottle do
     sha256 "dd214b33b7b66785327aecaac80d2260cca4e5e97dbd0efcb0d2c04d9eba18a9" => :sierra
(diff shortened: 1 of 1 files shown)


Formula/mono.rb
```diff
@@ -46,7 +46,8 @@ def install
 
     args << "--build=" + (MacOS.prefer_64_bit? ? "x86_64": "i686") + "-apple-darwin"
 
-    system "./configure", *args
+    system "./autogen.sh", *args
+    system "make", "get-monolite-latest"
```
MikeMcQuaid: What's this doing? Is it downloading from the internet or using Git submodules?

MikeMcQuaid:
> I switch build from tar archive to github tag. Because mono team has delay between publish release and build tar archive.

What about building from the GitHub tarball instead of tag? Does this use Git submodules at all? Thanks!

mfilippov:
@MikeMcQuaid  yes mono use submodules in repo.

MikeMcQuaid:
@mfilippov Ok, thanks. Still need question on `get-monolite-latest` figured out.

mfilippov:
@MikeMcQuaid get-monolite-latest download small subset of mono required for build on machine without mono.

MikeMcQuaid:
@mfilippov Could we do so in a `resource` block instead, by any chance? Does it use a checksum?

Relatedly, what is the delay between GitHub release and tarball archive? Thanks!

mfilippov:
1. I think I can make get monolite manualy, but mono has builtin check monolite version and self maintainable copy it is not good idea I think soo. 
2. Mono 4.6.2.16 released 22 dec 2016 but tarbal archive not available yet.

MikeMcQuaid:
> Mono 4.6.2.16 released 22 dec 2016 but tarbal archive not available yet.

Ok. Has this happened before? It could be someone just forgot before they went on vacation. If we can use the upstream tarball that'd be preferable so let's hold off on this for a couple more weeks.

mfilippov:
@MikeMcQuaid Ok, but It happened before I watch for this.

mfilippov:
@MikeMcQuaid I asked Xamarin team about the delay. This delay connected with holiday.

PR #5427, opened 2016-09-30
Add optional dependency for on p11-kit for gnutls

toonetown:
- [X] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [X] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [X] Does your submission pass `brew audit --new-formula <formula>` (after doing `brew install <formula>`)?

---

After the discussion and comment at https://github.com/Homebrew/homebrew-core/issues/5390#issuecomment-250530292, this PR adds an optional dependency on `p11-kit`.  Similar changes are included within #5416, however, this change is limited to _JUST_ `gnutls`.

Issue #5390, opened 2016-09-29
Proposoal to resurrect p11-kit


toonetown:
In January, p11-kit was sent to the boneyard, apparently because of [upstream's lack of maintenance for OS X](https://github.com/Homebrew/legacy-homebrew/issues/47124).

However, the two unit test failures that happen in that [upstream bug](https://bugs.freedesktop.org/show_bug.cgi?id=91602) appear to be due to the fact that we are running the build as non-root, or in some sort of sandboxed environment.  The code for those unit tests p11-kit actually contains an option (set by environment variable) to skip those particular tests:

https://cgit.freedesktop.org/p11-glue/p11-kit/tree/common/test-compat.c#n138
https://cgit.freedesktop.org/p11-glue/p11-kit/tree/p11-kit/test-conf.c#n451

Information on why these tests fail in certain environments is pretty sparse - but from what I can figure (*see below), it appears to be something to do with setguid and sandboxing or something like that.

In the original (homebrew-legacy) issue from January where it was decided to boneyard, @apjanke noted that:

```bash
$ brew uses p11-kit
gnutls
```

And this is where *I* ended up getting involved.  I'm interested in adding PKCS#11 support in to openconnect - but according to @dwmw2 (from the openconnect project), it already exists there...*IF* gnutls is installed using p11-kit.  That got me thinking that perhaps it could be used by more applications if gnutls were compiled with it:

```bash
$ brew uses gnutls
anjuta                      libvirt                     qemu                      
bitlbee                     libxmlsec1                  task                      
freediameter                loudmouth                   taskd                     
git-annex                   mailutils                   ucommon                   
glib-networking             minbif                      upscaledb                 
gnu-smalltalk               ncdc                        vte3                      
gtk-vnc                     openconnect                 weechat                   
gwenhywfar                  pianobar                    wine                      
knot                        pianod                      wireshark                 
libinfinity                 pidgin                      homebrew/fuse/s3fs        
libsoup                     profanity                   homebrew/versions/gnupg21
```

One of the main benefits of including p11-kit (and support for it in gnutls) would be the ability for openconnect to use certificates from the OS X Keychain if an appropriate PKCS#11 driver is installed (likely [KeychainToken](https://github.com/slushpupie/KeychainToken) - but perhaps others).

[...]

PR #8274, opened 2016-12-28
Update openconnect to depend on gnutls with p11-kit support.

toonetown:

-----
Once PR #5427 is merged in and p11-kit support is added to gnutls, we should make that dependency happen (since openconnect can leverage the p11 support in gnutls).

If we decide to make p11-kit support `:recommended` in gnutls (rather than `:optional`), then this PR not be necessary.

### DIFF
--- a/Formula/gnutls.rb
+++ b/Formula/gnutls.rb
@@ -5,6 +5,7 @@ class Gnutls < Formula
   mirror "https://gnupg.org/ftp/gcrypt/gnutls/v3.4/gnutls-3.4.17.tar.xz"
   mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/gnutls/v3.4/gnutls-3.4.17.tar.xz"
   sha256 "9b50e8a670d5e950425d96935c7ddd415eb6f8079615a36df425f09a3143172e"
+  revision 1
 
   bottle do
     sha256 "613d98861f14e5a880d8a9567b42d680a118f5ee38946a0cf70c6952293a9dd2" => :sierra
@@ -16,6 +17,7 @@ class Gnutls < Formula
   depends_on "libtasn1"
   depends_on "gmp"
   depends_on "nettle"
+  depends_on "p11-kit" => :recommended
   depends_on "guile" => :optional
   depends_on "unbound" => :optional
 
@@ -34,8 +36,13 @@ class Gnutls < Formula
       --sysconfdir=#{etc}
       --with-default-trust-store-file=#{etc}/openssl/cert.pem
       --disable-heartbeat-support
-      --without-p11-kit
     ]
+
+    if build.with? "p11-kit"
+      args << "--with-p11-kit"
+    else
+      args << "--without-p11-kit"
+    end
 
     if build.with? "guile"
       args << "--enable-guile" << "--with-guile-site-dir"
--- a/Formula/openconnect.rb
+++ b/Formula/openconnect.rb
@@ -3,6 +3,7 @@ class Openconnect < Formula
   homepage "http://www.infradead.org/openconnect.html"
   url "ftp://ftp.infradead.org/pub/openconnect/openconnect-7.08.tar.gz"
   sha256 "1c44ec1f37a6a025d1ca726b9555649417f1d31a46f747922b84099ace628a03"
+  revision 1
 
   bottle do
     sha256 "6ca71944ac9c1f0f8324cf3eb3b494dc3838036265b786f479003e41dd053e32" => :sierra


toonetown:
This build was aborted by @ilovezfs - is there a reason why?

ilovezfs:
The option is non-default so CI isn't actually testing anything. You'll need to temporarily make it required or recommended for the CI to be meaningful.

toonetown:
The option here actually is non-default - so in this one, we would want to run the checks, wouldn't we?  The change for this is to enable p11-kit support when building openconnect.  p11-kit support is `:optional` on gnutls - but the openconnect team (@dwmw2) wants to have it included by default on OS X.

This can be done in two ways:
 1. Set p11-kit as a `:recommended` option in the gnutls formula.  This is what has been suggested by the gnutls maintainer at https://github.com/Homebrew/homebrew-core/issues/5390#issuecomment-250766012:
    >@nmav
    >As the maintainer of gnutls, I'd strongly suggest making the p11-kit variant the default. The PKCS#11 functionality is essential in modern gnutls versions, especially for primary platforms like OSX which you would like your users to use smart cards, HSMs and the keychaintoken.

 2.  Set the `with-p11-kit` option to be included on the gnutls dependency within the openconnect formula.  That is what this pull request does.  This PR has a dependency upon #5427 - which is why there are two commits in this PR (the first commit is the commit from #5427, the second commit is actually the openconnect->p11-kit dependency.

If we go with option 1, then this PR is completely obsolete.  I would actually suggest going that route, given that the gnutls guys and the openconnect guys *both* are pushing for p11-kit support added in by default.  From what I can tell, the main difference is that there is a `p11tool` command line utility that is added, as well as a few additional library entry points.

This PR is just for option 2 above.

ilovezfs:
@DomT4 any thoughts here?

DomT4:
I'm pretty wary of contributing anything to Homebrew that may be in disagreement with Mike lest I end up simply wasting my breath after last week to be honest, but since you asked, if the upstream maintainers of `gnutls` are pushing for it I think it's worth not pushing back against that and adopting it by default for `gnutls` and forcing a revision bump.

toonetown:
If consensus is to go with option 1 (my preference as well), then I would probably suggest updating #5427 to include that dependency and closing this PR.

toonetown:
...or would it be preferable for me to just create a new PR for `gnutls+p11-kit :recommended`, and close both this and #5427 to keep things clean?

MikeMcQuaid:
If `gnutls` is far more fully-featured on `openconnect` and upstream favours it over `openssl` then I'd be in favour of just making it the default and dropping the `openssl` option there.

Re: `gnutls` itself: given the `gnutls` maintainer has strongly recommended we make it on-by-default I'm fine with doing that, too, now. If we do so it'd be good to also figure out how to include and enable the `p11-kit` macOS keychain integration you mentioned, too.

MikeMcQuaid:
> I'm pretty wary of contributing anything to Homebrew that may be in disagreement with Mike lest I end up simply wasting my breath after last week to be honest, but since you asked

@DomT4 I don't think this is fair, really. I do always listen to and consider what you say and you know better than anyone how often I took your side on disagreements when you were maintaining Homebrew. Now that you are not maintaining Homebrew your opinion is still valued but I'm now relying on your argument rather than the amount of work you're putting in to justify it. This is how you advocated people be treated when you were a maintainer and, as I've stated before, we'd always welcome you back.

DomT4:
> I don't think this is fair, really. I do always listen to and consider what you say and you know better than anyone how often I took your side on disagreements when you were maintaining Homebrew. Now that you are not maintaining Homebrew your opinion is still valued but I'm now relying on your argument rather than the amount of work you're putting in to justify it. This is how you advocated people be treated when you were a maintainer and, as I've stated before, we'd always welcome you back.

My comment was a little sharp, but I very much got the impression the discussion was over before I even stuck my oar in last week or whenever it was now, and that left me a little salty. The _"war against `libressl`"_ was the priority and there wasn't much room for common ground even after I pointed out my objections weren't based around the `git` PR being closed, or not accepting new `optional` deps for `libressl` as had been the policy when I was still here.

The argument I ended up making was that in some places you were dumping on the upstream default in favour of `openssl` and given there's not absolute compatibility between `openssl` and `libressl` long-term you may end up causing yourself or an upstream trouble, especially the upstreams that build by default using `libressl` like at least `h2o` and `openssh`. I also pointed out these days `libressl` is far more a system-default on macOS than `openssl` is, and indeed, running `otool` through `/usr/bin` on 10.12+ will reveal as much. By this point it's not like `libressl` is a new addition to macOS, it's been around for over a year now with Apple showing various levels of dedication in keeping it up-to-date.

I very much appreciate now I'm no longer regularly contributing and did step down from maintaining the project my voice is simply one amongst the crowd, which is why I almost always keep my trap shut about anything that makes me 🙄 a bit, but even amongst this crowd I've been the one who pushed for or pushed through some of the major changes to the crypto library handling in Homebrew for the last couple years, from pushing a lot harder for us to mandate using our `openssl` rather than the system's and adding a hoard of revisions to make that transition, to deprecating SSLv2 from our `openssl` _way_ before other popular macOS package managers, to introducing `openssl@1.1` and starting work on integrating that before I ran away. It's not like these are waters many other people have swam in as much as I have for the last ~2 years or however long it has been now.

My point wasn't _"Listen to me, I'm a former maintainer, insert grumpy face here"_ it was _"Hey, I've been the one manning the lighthouse on this for the last couple years, I'm not sure how great an idea this is, and this is why"_ but it genuinely felt like unless I could push back with the power of being a maintainer I wasn't going to get anywhere against your argument. And maybe that's not completely fair, but I'm nothing if not human and certainly not immune to feeling things based on perception rather than absolute raw fact.

I'm aware of the position I advocated when I was here, as failing and defective as my memory often is, but I wasn't ever suggesting for example that if Tim stepped away we stop listening to his thoughts or advice on `python` stuff, even if we didn't end up following that advice. The feeling was that whatever I said was just bouncing off a wall rather than actually being returned to me with consideration, and again, I reference my point above about being human here.

You know I don't pick fights with you simply because I _disagree with the principle_ of doing something, otherwise I wouldn't have shut up and quietly let analytics or other things sail through with tacit approval from myself, and indeed offered a robust defence time and time again for it & other things being implemented. I am wildly careful to not just lamp you for stuff arbitrarily.

MikeMcQuaid:
> The argument I ended up making was that in some places you were dumping on the upstream default in favour of openssl and given there's not absolute compatibility between openssl and libressl long-term you may end up causing yourself or an upstream trouble, especially the upstreams that build by default using libressl like at least h2o and openssh.

I appreciate that. When it becomes a problem we'll consider making changes there.

> It's not like these are waters many other people have swam in as much as I have for the last ~2 years or however long it has been now.

I'm aware of that and we're still grateful for that work. Some of it, though (i.e. adding LibreSSL options to a fair few places) hasn't been matched by widespread upstream or user adoption so it's not worth the maintenance burden.

Where I've definitely failed here is I tended to just defer to you on some of these things as long as they didn't cause user issues when you were around. I should have discussed things and made my views clearer at the time and I appreciate and apologise for the grief that's resulted from my not doing so.

toonetown:
Once #5427 is merged, this PR can be closed, as it will no longer be necessary.

> @MikeMcQuaid 
> If `gnutls` is far more fully-featured on `openconnect` and upstream favours it over `openssl` then I'd be in favour of just making it the default and dropping the `openssl` option there.

I think that upstream favours `gnutls` over `openssl`...but I do believe that `openssl` is still supported (and may be a possible valid option if someone doesn't want to pull in all the dependencies of `gnutls`).  Perhaps the upstream maintainer - @dwmw2 - could comment on what the upstream opinion of `gnutls`-vs-`openssl` is?

In either case, if we remove `openssl` as an option in the formula, I can submit that as a separate PR - and this one can still be closed once #5427 is merged.

toonetown:
Actually, given the comment at https://github.com/Homebrew/homebrew-core/issues/5390#issuecomment-250762636
> @dwmw2
> But if the configure script finds that the `gnutls_pkcs11_add_provider()` function isn't present in GnuTLS, it'll infer that GnuTLS was built without PKCS#11 support and it'll build OpenConnect without PKCS#11 support accordingly.

I think that once #5427 is merged, we need to bump the `revision` of this so that the configure script will pull in the `gnutls_pkcs11_add_provider()` function.  I will update this PR accordingly.

MikeMcQuaid:
https://github.com/Homebrew/homebrew-core/pull/5427 was merged 🎉

dunn:
Could you go ahead and bump the revision now? Thanks!

PR #8275, opened 2016-12-28
libgit2-glib: vendor libgit2 0.24.x

ilovezfs:
Re-evaluate whether to continue vendoring libgit2 when libgit2-glib gets
libgit2 0.25.x compatibility.

See https://github.com/Homebrew/homebrew-core/pull/8221#issuecomment-269237260.

CC @JCount @tschoonj


tschoonj:
Not pretty but I can certainly live with this.

JCount:
Me too

MikeMcQuaid:
Fine with me 👍 . Rather than vendoring, though, we may want to consider shipping a `libgit@0.24` formula instead. We may want until the versioning support is more fleshed out (i.e. homebrew/versions is torn down) for that and/or that there's more than one formulae depending on this older version.

ilovezfs:
This is specifically intended to avoid such an @ formula ;)

PR #8278, opened 2016-12-28
servus 1.5.0 (new formula)

staticfloat:

-----

### DIFF
--- a/Formula/servus.rb
+++ b/Formula/servus.rb
@@ -1,0 +1,25 @@
+class Servus < Formula
+  desc "Library and Utilities for zeroconf networking"
+  homepage "https://github.com/HBPVIS/Servus"
+  url "https://github.com/HBPVIS/Servus/archive/1.5.0.tar.gz"
+  sha256 "42fb9c060f17f040ad3c7563f5e87c89f5a221a5aa7da21384fc26b9c725ecc8"
+
+  depends_on "cmake" => :build
+  depends_on "boost"
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+
+    # Install tests to share/Servus/tests so that we can test later
+    mkdir share/"Servus/tests"
+    (share/"Servus/tests").install Dir["tests/{serializable,servus,uint128_t,uri}"]
+  end
+
+  test do
+    system share/"Servus/tests/serializable"
+    system share/"Servus/tests/servus"
+    system share/"Servus/tests/uint128_t"
+    system share/"Servus/tests/uri"
+  end
+end
--- a/Formula/servus.rb
+++ b/Formula/servus.rb
@@ -15,94 +15,94 @@ class Servus < Formula
   test do
     # Embed "serializeable" test from the servus 1.5.0 source
     (testpath/"test.cpp").write <<-EOS.undent
-#define BOOST_TEST_MODULE servus_serializable
-#include <boost/test/unit_test.hpp>
+      #define BOOST_TEST_MODULE servus_serializable
+      #include <boost/test/unit_test.hpp>
 
-#include <servus/uint128_t.h>
-#include <servus/serializable.h>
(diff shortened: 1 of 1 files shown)


Formula/servus.rb
```diff
@@ -0,0 +1,30 @@
+class Servus < Formula
+  desc "Library and Utilities for zeroconf networking"
+  homepage "https://github.com/HBPVIS/Servus"
+  url "https://github.com/HBPVIS/Servus/archive/1.5.0.tar.gz"
+  sha256 "42fb9c060f17f040ad3c7563f5e87c89f5a221a5aa7da21384fc26b9c725ecc8"
+
+  bottle do
+    cellar :any
+    sha256 "fb13c90a02d860b5442a046a66a98f190788b2d5f576c574b891751a14353893" => :el_capitan
+  end
```
MikeMcQuaid: Remove `bottle do` blocks on new formulae.

Formula/servus.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  depends_on "cmake" => :build
+  depends_on "boost"
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+
+    # Install tests to share/Servus/tests so that we can test later
```
MikeMcQuaid: Would rather use an integration test instead i.e. actually using the library and utilities.

MikeMcQuaid (changes requested):

Formula/servus.rb
staticfloat: Whoops, that snuck in during my testing, fixed now.

Formula/servus.rb
staticfloat: I'm not sure what you mean by this; these tests are the builtin tests that the library provides, that run through the various functionality of the library.  The test executables are dynamically linked against the library and use the library's functions, ensuring that the library works as intended.  What would you prefer?

MikeMcQuaid: This generally isn't how we run tests; it'd be ideal to just run something in `bin` or if it's primarily a library link against something in `lib`.

staticfloat: This formula has nothing in `bin` to test with (and you wouldn't want to put these executables into `bin` because they have very generic names like `uri` that might conflict with other tools).  Would you prefer that I take the source code of these tests, embed it into the formula, compile it at test time and test with that, kind of like how [libxml2 does it](https://github.com/Homebrew/homebrew-core/blob/master/Formula/libxml2.rb#L54-L71)?

ilovezfs: Another example: https://github.com/Homebrew/homebrew-core/blob/master/Formula/ibex.rb#L54-L71

MikeMcQuaid:
> Would you prefer that I take the source code of these tests, embed it into the formula, compile it at test time and test with that, kind of like how libxml2 does it?

Yes, doing that for a single test sounds good, thanks.

staticfloat:
Done

Formula/servus.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    # Embed "serializeable" test from the servus 1.5.0 source
+    (testpath/"test.cpp").write <<-EOS.undent
+#define BOOST_TEST_MODULE servus_serializable
```
MikeMcQuaid: Indent all the lines by six spaces. Ideally strip this test down so it's a bit more minimal i.e. just checks a link works but feel free to leave it as-is.

staticfloat:
Test source indented.  I left it as-is because I think there's value in having the source code here the same as the tests from the 1.5.0 source file; when `servus` changes its API, we will have a reference that we can pull future code from, no need to maintain a separate test case.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #8286, opened 2016-12-29
lnav: use GitHub for homepage

ilovezfs:
lnav.org has an SSL problem currently, causing brew audit failure


ilovezfs:
Reported here:  https://github.com/tstack/lnav/issues/401

PR #8290, opened 2016-12-29
osquery: vendor boost 1.62.0 and depend on openssl

ilovezfs:
- vendor Boost 1.62.0 since osquery is not yet compatible with 1.63.0
- depend on openssl
- drop libressl dependency (as it wasn't actually being used anyway)

Without the Boost vendoring, this was failing to build due to the Boost 1.63.0 upgrade: https://bot.brew.sh/job/Homebrew%20Testing/1209/version=el_capitan/console

Note that osquery 1.7.3 is several versions behind the current stable
version (2.2.1), which is also be broken with boost 1.63.0.

CC @theopolis

### DIFF
--- a/Formula/osquery.rb
+++ b/Formula/osquery.rb
@@ -5,7 +5,7 @@ class Osquery < Formula
   url "https://github.com/facebook/osquery.git",
     :tag => "1.7.3",
     :revision => "6901aa644a9bcc0667207008db71471abf756b82"
-  revision 6
+  revision 7
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


ilovezfs:
Boost 1.63.0 breakage reported to osquery upstream here: https://github.com/facebook/osquery/issues/2888.

MikeMcQuaid (commented):
Does this build all of Boost or just some of it? Given the vendoring and difficulty building a newer version I'm tempted to say we just boneyard this.

ilovezfs:
```
 +        --with-libraries=filesystem,regex,system 
```
https://github.com/Homebrew/homebrew-core/pull/8290/files#diff-ef2935cce4dec080563243e3608d91b3R79

It only builds those three libraries, which is fast.

As for boneyarding, analytics shows 1,340 installs in the last 30 days.

MikeMcQuaid:
Ok, cool, that's fine with me. Kinda regardless of the number of installs it feels a bit 💩 to be sitting on an old version for so long but I'm happy to leave it if you are.

ilovezfs:
@MikeMcQuaid I recall that I was able to work around https://github.com/facebook/osquery/issues/2362 by similarly vendoring a static aws, so that should still be a way to get it current again.

MikeMcQuaid (approved):

PR #8299, opened 2016-12-29
compcert: build with Coq 8.5pl3 from OPAM

ilovezfs:
CompCert does not yet support Coq 8.6.

See https://github.com/AbsInt/CompCert/pull/150.


ilovezfs:
@BrewTestBot test this please

PR #8303, opened 2016-12-29
recipes 0.4.2 (new formula)

matthiasclasen:
This commit adds a formula for GNOME recipes, which is an
application for cooking and managing recipes.

### DIFF
--- a/Formula/recipes.rb
+++ b/Formula/recipes.rb
@@ -1,7 +1,7 @@
 class Recipes < Formula
   desc "Formula for GNOME recipes"
   homepage "https://wiki.gnome.org/Apps/Recipes"
-  url "https://mclasen.fedorapeople.org/recipes-0.4.2.tar.xz"
+  url "https://download.gnome.org/sources/recipes/0.4/recipes-0.4.2.tar.xz"
   sha256 "9554d4f5d97eb9cd4032de0e4f9cc27a218c32d022dd1917a7e9efbd379c5bc1"
 
   depends_on "gtk+3"
--- a/Formula/recipes.rb
+++ b/Formula/recipes.rb
@@ -1,0 +1,24 @@
+class Recipes < Formula
+  desc "Formula for GNOME recipes"
+  homepage "https://wiki.gnome.org/Apps/Recipes"
+  url "https://mclasen.fedorapeople.org/recipes-0.4.2.tar.xz"
+  sha256 "9554d4f5d97eb9cd4032de0e4f9cc27a218c32d022dd1917a7e9efbd379c5bc1"
+
+  depends_on "gtk+3"
+  depends_on "gnome-icon-theme"
+
+  def install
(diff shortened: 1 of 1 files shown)


Formula/recipes.rb
```diff
@@ -0,0 +1,24 @@
+class Recipes < Formula
+  desc "Formula for GNOME recipes"
+  homepage "https://wiki.gnome.org/Apps/Recipes"
+  url "https://mclasen.fedorapeople.org/recipes-0.4.2.tar.xz"
```
MikeMcQuaid: This URL is failing: https://bot.brew.sh/job/Homebrew%20Core/13831/version=sierra/testReport/junit/brew-test-bot/sierra/fetch_recipes/

Formula/recipes.rb
```diff
@@ -0,0 +1,24 @@
...
+  sha256 "9554d4f5d97eb9cd4032de0e4f9cc27a218c32d022dd1917a7e9efbd379c5bc1"
+
+  depends_on "gtk+3"
+  depends_on "gnome-icon-theme"
+
+  def install
+    inreplace "data/Makefile.in", "gtk-update-icon-cache", "gtk3-update-icon-cache"
```
MikeMcQuaid: Why is this needed? It's worth filing an upstream issue to detect this correctly.

Formula/recipes.rb
```diff
@@ -0,0 +1,24 @@
...
+                          "--disable-silent-rules",
+                          "--disable-debug",
+                          "--prefix=#{prefix}",
+                          "--disable-autoar",
+                          "--disable-gspell"
+    system "make", "install"
+  end
+
+  test do
+    system "recipes", "--help"
```
MikeMcQuaid: Audit failing: https://bot.brew.sh/job/Homebrew%20Core/13831/version=sierra/testReport/junit/brew-test-bot/sierra/audit_recipes___online___new_formula/

MikeMcQuaid (changes requested):

Formula/recipes.rb
matthiasclasen: This was copied from the file-roller formula. I should have copied the comment as well, I guess:

    # forces use of gtk3-update-icon-cache instead of gtk-update-icon-cache. No bugreport should
    # be filed for this since it only occurs because Homebrew renames gtk+3's gtk-update-icon-cache
    # to gtk3-update-icon-cache in order to avoid a collision between gtk+ and gtk+3.

MikeMcQuaid: Yeh, can you add that comment, thanks.

tschoonj:
Looks like the URL you provided doesn't work: use https://download.gnome.org/sources/recipes/0.4/recipes-0.4.2.tar.xz instead?

You may want to add support for `gspell`, since that formula is in Homebrew already. Also, you will be able to add support for `gnome-autoar` (#8330) when we merge that formula in.

MikeMcQuaid (approved):

PR #8309, opened 2016-12-30
global: --with-universal-ctags

ivanbrennan:
GNU Global supports using Universal Ctags and Exuberant Ctags as plug-in parsers, but the current formula only exposes the Exuberant Ctags option.

This is problematic, as Universal Ctags is the actively maintained/developed implementation. The Global documentation describes how to configure Universal Ctags and makes no mention of Exuberant Ctags.

Expose `--with-universal-ctags` option to remedy this.

To install:

    $ brew install global --with-universal-ctags

To use:

    $ export GTAGSCONF=/usr/local/share/gtags/gtags.conf
    $ export GTAGSLABEL=new-ctags
    $ gtags                         # gtags invokes Universal Ctags internally

References:
  https://www.gnu.org/software/global/manual/global.html#Plug_002din
  https://github.com/universal-ctags/homebrew-universal-ctags

### DIFF
--- a/Formula/global.rb
+++ b/Formula/global.rb
@@ -23,12 +23,14 @@ class Global < Formula
   end
 
   option "with-ctags", "Enable Exuberant Ctags as a plug-in parser"
+  option "with-universal-ctags", "Enable Universal Ctags as a plug-in parser"
   option "with-pygments", "Enable Pygments as a plug-in parser (should enable exuberent-ctags too)"
   option "with-sqlite3", "Use SQLite3 API instead of BSD/DB API for making tag files"
 
   deprecated_option "with-exuberant-ctags" => "with-ctags"
 
   depends_on "ctags" => :optional
+  depends_on "universal-ctags/universal-ctags/universal-ctags" => :optional
 
   skip_clean "lib/gtags"
 
@@ -50,6 +52,10 @@ class Global < Formula
 
     if build.with? "ctags"
       args << "--with-exuberant-ctags=#{Formula["ctags"].opt_bin}/ctags"
+    end
+
+    if build.with? "universal-ctags"
+      args << "--with-universal-ctags=#{Formula["universal-ctags"].opt_bin}/ctags"
     end
 
     if build.with? "pygments"
@@ -78,7 +84,7 @@ class Global < Formula
        int c2func (void) { return 0; }
        void cfunc (void) {int cvar = c2func(); }")
     EOS
-    if build.with?("pygments") || build.with?("ctags")
+    if build.with?("pygments") || build.with?("ctags") || build.with?("universal-ctags")
       (testpath/"test.py").write <<-EOS
         def py2func ():
              return 0
@@ -88,7 +94,7 @@ class Global < Formula
     end
     if build.with? "pygments"
       assert shell_output("#{bin}/gtags --gtagsconf=#{share}/gtags/gtags.conf --gtagslabel=pygments .")
-      if build.with? "ctags"
+      if build.with?("ctags") || build.with?("universal-ctags")
         assert_match "test.c", shell_output("#{bin}/global -d cfunc")
         assert_match "test.c", shell_output("#{bin}/global -d c2func")
         assert_match "test.c", shell_output("#{bin}/global -r c2func")


ivanbrennan:
Tests are currently failing because the formula can't find the `universal-ctags` formula (I ran tests locally, but only after running `brew tap universal-ctags/universal-ctags`.

I'm looking into either auto-tapping or providing an informative error in such cases -- not sure which is the preferred behavior.

ivanbrennan:
I updated the `depends_on` call to correctly reference the tapped `universal-ctags` formula:
```diff
 -  depends_on "universal-ctags" => :optional
 +  depends_on "universal-ctags/universal-ctags/universal-ctags" => :optional
```

Formula/global.rb
```diff
@@ -23,12 +24,14 @@ class Global < Formula
...
 
   option "with-ctags", "Enable Exuberant Ctags as a plug-in parser"
+  option "with-universal-ctags", "Enable Universal Ctags as a plug-in parser"
   option "with-pygments", "Enable Pygments as a plug-in parser (should enable exuberent-ctags too)"
   option "with-sqlite3", "Use SQLite3 API instead of BSD/DB API for making tag files"
 
   deprecated_option "with-exuberant-ctags" => "with-ctags"
 
   depends_on "ctags" => :optional
+  depends_on "universal-ctags/universal-ctags/universal-ctags" => :optional
```
MikeMcQuaid: We don't want homebrew/core formulae depending on non-Homebrew organisation taps as options. Could this be submitted to homebrew/core?

Formula/global.rb
```diff
@@ -4,6 +4,7 @@ class Global < Formula
   url "https://ftpmirror.gnu.org/global/global-6.5.6.tar.gz"
   mirror "https://ftp.gnu.org/gnu/global/global-6.5.6.tar.gz"
   sha256 "122f9afa69a8daa0f64c12db7f02981fe573f51a163fa3829ed4f832cd281505"
+  revision 1
```
MikeMcQuaid: Don't need this unless you modify the non-option installation.

Formula/global.rb
```diff
@@ -117,6 +124,18 @@ def pyfunc ():
       assert_no_match(/test\.py/, shell_output("#{bin}/global -r py2func # correctly fails"))
       assert_no_match(/test\.py/, shell_output("#{bin}/global -s pyvar   # correctly fails"))
     end
+    if build.with? "universal-ctags"
```
MikeMcQuaid: Probably not worth adding this; these tests are only for CI and thus will never be run with options so this code will never be run.

MikeMcQuaid (changes requested):

Formula/global.rb
ivanbrennan: I wasn't sure whether exposing a new option warranted/required adding a `revision`. I'm starting to suspect not, but I'll leave this for now to get some outside feedback.

MikeMcQuaid: It definitely doesn't so you can remove it, thanks. A `revision` will force a reinstall and, in some cases, rebuild for everyone using this.

Formula/global.rb
ivanbrennan: @MikeMcQuaid The [universal-ctags tap](https://github.com/universal-ctags/homebrew-universal-ctags) mentions,

> When Universal Ctags has a tagged release, this formula should be moved to the official Homebrew repository. This repo should then be deleted.

Indeed, [universal ctags](https://github.com/universal-ctags/ctags) still has no tagged release. I'll check with the maintainers on the status of that. Is a tagged release a prerequisite for submitting to homebrew/core?

MikeMcQuaid: It is, yep.

dunn:
Do you have any idea when a stable release is expected?

ivanbrennan:
@dunn I found multiple universal ctags issues discussing their plans for a release, but it all seems up in the air at the moment. For reference:
https://github.com/universal-ctags/ctags/issues/1038
https://github.com/universal-ctags/ctags/issues/664
https://github.com/universal-ctags/ctags/issues/658
https://github.com/universal-ctags/ctags/issues/909
https://github.com/universal-ctags/ctags/issues/662
https://github.com/universal-ctags/ctags/issues/652

Apologies for the barrage of links, maybe too much information, but that's what I found.

dunn:
Thanks!  If there's no definite ETA, you can still leave this PR open, but we're gonna want that stable release before merging.

ivanbrennan:
Sounds good. Thanks @dunn

MikeMcQuaid:
Closing until we've got a stable release.

PR #8330, opened 2016-12-30
gnome-autoar 0.1.1 (new formula)

matthiasclasen:
This commit adds a formula for GNOME autoar, which is
a convenience library providing GIO-style async wrappers
around libarchive.

### DIFF
--- a/Formula/gnome-autoar.rb
+++ b/Formula/gnome-autoar.rb
@@ -1,5 +1,5 @@
 class GnomeAutoar < Formula
-  desc "A convenient library for archive handling"
+  desc "GNOME library for archive handling"
   homepage "https://github.com/GNOME/gnome-autoar"
   url "https://download.gnome.org/sources/gnome-autoar/0.1/gnome-autoar-0.1.1.tar.xz"
   sha256 "f65cb810b562dc038ced739fbf59739fd5df1a8e848636e21f363ded9f349ac9"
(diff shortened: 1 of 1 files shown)


Formula/gnome-autoar.rb
```diff
@@ -0,0 +1,29 @@
...
+                          "--disable-schemas-compile"
+    system "make", "install"
+  end
+
+  def post_install
+    system "#{Formula["glib"].opt_bin}/glib-compile-schemas", "#{HOMEBREW_PREFIX}/share/glib-2.0/schemas"
+  end
+
+  test do
+    system "false"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

MikeMcQuaid (changes requested):

MikeMcQuaid:
Please fill out the pull request template as requested. There's a bunch of audit failures: https://bot.brew.sh/job/Homebrew%20Core/13867/version=sierra/testReport/junit/brew-test-bot/sierra/audit_gnome_autoar___online___new_formula/

Maintainers note: we can waive the notability requirement here because it's from the GNOME project. CC @tschoonj who may be interested.

tschoonj:
Hi @matthiasclasen,

Last week I created a Formula for this package, as a dependency for GNOME recipes, but since I couldn't get that one to compile, I forgot about the whole thing.

Am I correct in assuming that todays recipes 0.4.2 release now supports macOS as well? And that you plan on creating a formula for that one too when this one gets merged in?

Anyway, let me help you a bit here: I will commit a proper test in a bit...

matthiasclasen:
Thanks for your help, tschoonj !

And yes, recipes 0.4.2 should be buildable on OS X with --disable-autoar. Once this formula is in, getting it to build with the dependencies should just be a matter of fixing some more compiler warnings.

MikeMcQuaid (approved):

PR #8333, opened 2016-12-30
gcc@4.*, cloog@0.18: use isl@0.12.

MikeMcQuaid:
Suggested by @equal-l2 in #7557.


equal-l2:
You should add `isl@0.12` formula in advance.
Copying from `isl@0.11` and change version will work.

MikeMcQuaid:
@equal-l2 lol, whoops. Forgot to commit that. Thanks!

PR #8306, opened 2016-12-30
qt5 5.7.1 and delete qt4

DomT4:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

```
==> Summary
🍺  /usr/local/Cellar/qt5/5.7.1: 8,273 files, 242.3M, built in 83 minutes 35 seconds
```

It's that time again.

Fixes https://github.com/Homebrew/homebrew-core/issues/1705.

PR #8334, opened 2016-12-30
Rename qt5 to qt

MikeMcQuaid:
And adjust formulae accordingly.

CC @cartr as I'll hold off on this until you're ready.

### DIFF
--- a/Aliases/pyqt5
+++ b/Aliases/pyqt5
@@ -1,0 +1,1 @@
+../Formula/pyqt.rb
--- a/Aliases/qt
+++ b/Aliases/qt
@@ -1,1 +1,0 @@
-../Formula/qt5.rb
--- a/Aliases/qt5
+++ b/Aliases/qt5
@@ -1,0 +1,1 @@
+../Formula/qt.rb
--- a/Formula/cgal.rb
+++ b/Formula/cgal.rb
@@ -12,18 +12,19 @@ class Cgal < Formula
   end
 
   option :cxx11
-  option "with-qt5", "Build ImageIO and QT5 compoments of CGAL"
+  option "with-qt", "Build ImageIO and Qt components of CGAL"
   option "with-eigen", "Build with Eigen3 support"
   option "with-lapack", "Build with LAPACK support"
 
-  deprecated_option "imaging" => "with-qt5"
-  deprecated_option "with-imaging" => "with-qt5"
+  deprecated_option "imaging" => "with-qt"
+  deprecated_option "with-imaging" => "with-qt"
   deprecated_option "with-eigen3" => "with-eigen"
+  deprecated_option "with-qt5" => "with-qt"
 
   depends_on "cmake" => :build
   depends_on "mpfr"
 
-  depends_on "qt5" => :optional
+  depends_on "qt" => :optional
   depends_on "eigen" => :optional
 
   if build.cxx11?
@@ -42,7 +43,7 @@ class Cgal < Formula
       -DCMAKE_INSTALL_NAME_DIR=#{HOMEBREW_PREFIX}/lib
     ]
 
-    if build.without? "qt5"
+    if build.without? "qt"
       args << "-DWITH_CGAL_Qt5=OFF" << "-DWITH_CGAL_ImageIO=OFF"
     else
       args << "-DWITH_CGAL_Qt5=ON" << "-DWITH_CGAL_ImageIO=ON"
--- a/Formula/cockatrice.rb
+++ b/Formula/cockatrice.rb
@@ -5,6 +5,7 @@ class Cockatrice < Formula
       :tag => "2017-03-14-Release",
       :revision => "6e723b2a992022ba343d45d881b3c92b9d1c6ba2"
   version "2017-03-14"
+  revision 1
   head "https://github.com/Cockatrice/Cockatrice.git"
 
   bottle do
@@ -20,9 +21,9 @@ class Cockatrice < Formula
   depends_on "protobuf"
 
   if build.with? "server"
-    depends_on "qt5" => "with-mysql"
+    depends_on "qt" => "with-mysql"
   else
-    depends_on "qt5"
+    depends_on "qt"
   end
 
   fails_with :clang do
--- a/Formula/codequery.rb
+++ b/Formula/codequery.rb
@@ -3,6 +3,7 @@ class Codequery < Formula
   homepage "https://github.com/ruben2020/codequery"
   url "https://github.com/ruben2020/codequery/archive/v0.19.0.tar.gz"
   sha256 "c8fbb133ec281cdc58d81824a1d3b1761e5496fcfd37229e68060beb0e0c7fc9"
+  revision 1
 
   bottle do
     cellar :any
@@ -12,7 +13,7 @@ class Codequery < Formula
   end
 
   depends_on "cmake" => :build
-  depends_on "qt5"
+  depends_on "qt"
 
   def install
     args = std_cmake_args
--- a/Formula/color-code.rb
+++ b/Formula/color-code.rb
@@ -3,6 +3,7 @@ class ColorCode < Formula
   homepage "http://colorcode.laebisch.com/"
   url "http://colorcode.laebisch.com/download/ColorCode-0.8.5.tar.gz"
   sha256 "7c128db12af6ab11439eb710091b4a448100553a4d11d3a7c8dafdfbc57c1a85"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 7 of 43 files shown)


RandomDSdevel:
@MikeMcQuaid:  As noted in [this comment](https://github.com/Homebrew/homebrew-core/pull/8306#issuecomment-269824169) on #8306, even just having `qt` be an _alias_ to `qt5` has already started causing issues for the small subset of Homebrew users who, prior to said PR's resolution, had the latest Homebrew-delivered instances of both Qt versions 4.x and 5.x installed on their systems.  As I stated in the afore-mentioned comment, I, as a member of that small user subset, will submit an issue about this (it would have happened by now, but I _kind_ of got distracted reading the release notes for the stuff I have on my plate to `brew upgrade` as of today…😆.)

RandomDSdevel:
@MikeMcQuaid:  As reported by GitHub's notification of #8342's existence and mention of this PR in said issue, said issue is now ready and we can continue discussion of my problem there.

MikeMcQuaid:
Most of the failures here are due to https://github.com/Homebrew/homebrew-core/issues/8392

MikeMcQuaid:
Note this is blocked on https://github.com/Homebrew/brew/issues/1770.

MikeMcQuaid:
This looks to be working. Any objections, folks? I can fix up the minor audit fails when I pull.

MikeMcQuaid:
And @cartr for final thoughts (and to check you have a `pyqt@4` if you want one).

ilovezfs:
This seems to be incomplete.

```
iMac-TMP:homebrew-core joe$ ag qt5|grep depend
Formula/cockatrice.rb:23:    depends_on "qt5" => "with-mysql"
Formula/cockatrice.rb:25:    depends_on "qt5"
Formula/color-code.rb:14:  depends_on "qt5"
Formula/gammaray.rb:23:  depends_on "homebrew/science/vtk" => [:optional, "with-qt5"]
Formula/gearboy.rb:15:  depends_on "qt5"
Formula/gearsystem.rb:15:  depends_on "qt5"
Formula/mgba.rb:32:  depends_on "qt5" => :recommended
Formula/qtads.rb:33:  depends_on "qt5"
Formula/quazip.rb:14:  depends_on "qt5"
Formula/qwt.rb:17:  depends_on "qt5"
Formula/qwtpolar.rb:16:  depends_on "qt5"
Formula/qxmpp.rb:15:  depends_on "qt5"
Formula/tarsnap-gui.rb:15:  depends_on "qt5"
Formula/treefrog.rb:23:  depends_on "qt5" => qt5_build_options
Formula/ttfautohint.rb:29:  depends_on "qt5" => :optional
Formula/urh.rb:20:  depends_on "pyqt5"
Formula/wireshark.rb:33:  depends_on "qt5" => :optional
Formula/zurl.rb:16:  depends_on "qt5"
```

cartr:
Not sure the extent to which this affects core's formulae, but one issue I've been running into while working on my side of the change is that lots of formulae hard-code Qt's formula name as part of their build process, and that changing the formula name breaks them because they can't find Qt any more. (You may need to bump the `revision` and re-bottle the dependent formulae.) 

It would be worth checking that by installing a bunch of formulae that depend on `qt5`, switching over to the branch that renames it to `qt`, then seeing if they still work.

ilovezfs:
@cartr that is odd. Formulae names are not really exposed to build systems, so the problems you're describing probably have some other cause.

cartr:
@ilovezfs In this case, I'm pretty sure the formula name is being exposed to the build system of other formulae. Here's my understanding of how it works, at least with Qt 4:

1. When Qt builds, we pass it a path that includes its formula name (e.g. `/usr/local/Cellar/qt/4.8.2`) as part of the `--prefix` configure argument.
2. Qt's configure script saves that path into a C++ file that is then built as part of `qmake`.
3. Other formulae run `qmake` as part of their build processes to determine where Qt is installed.
4. `qmake` uses `/usr/local/Cellar/qt/4.8.2` as the path to Qt in the Makefiles and `-query` output it generates, and other formulae end up compiling that path into themselves and saving it in config files.
5. Formulae attempt to [clean this up](https://github.com/cartr/homebrew-qt4/blob/4587160ec0495a1ce274502938e53dba83f19f34/pyside.rb#L51) by using `inreplace` to replace `/usr/local/Cellar/qt/4.8.2` with `/usr/local/opt/qt`.
6. The fact that Qt 4's opt prefix is `/usr/local/opt/qt` (and, by extension, that its formula name is `qt`) gets compiled into other formulae's binaries.

I haven't done super in-depth testing for Qt 5, but I suspect the same is true there, as `qmake` still has its path compiled in to the binary:

```
carter@Carter-MBP ~> /usr/local/Cellar/qt5/5.8.0_1/bin/qmake -query
QT_SYSROOT:
QT_INSTALL_PREFIX:/usr/local/Cellar/qt5/5.8.0_1
QT_INSTALL_ARCHDATA:/usr/local/Cellar/qt5/5.8.0_1
QT_INSTALL_DATA:/usr/local/Cellar/qt5/5.8.0_1
QT_INSTALL_DOCS:/usr/local/Cellar/qt5/5.8.0_1/doc
[...]
carter@Carter-MBP ~> strings /usr/local/Cellar/qt5/5.8.0_1/bin/qmake | grep "/usr/local/Cellar/"
qt_epfxpath=/usr/local/Cellar/qt5/5.8.0_1
qt_prfxpath=/usr/local/Cellar/qt5/5.8.0_1
qt_hpfxpath=/usr/local/Cellar/qt5/5.8.0_1
carter@Carter-MBP ~>
```

ilovezfs:
@cartr that all sounds normal though, no? Going forward those paths will be `/usr/local/opt/qt` for qt 5.x and `/usr/local/opt/qt@4` for qt 4.x, or have I misunderstood something?

cartr:
@ilovezfs The issue is that those paths are changing, which will cause any already-built formula that hard-coded them to be wrong.

ilovezfs:
@cartr right, which is why new bottles for the formulae in the PR will be pulled as well. Are there things missing from the PR that you think need new bottles too?

cartr:
@ilovezfs Maybe I'm misunderstanding how Homebrew works, but I didn't think `brew` re-installed formulae/re-downloaded bottles unless the formula's version number changed. If someone installs a formula that depends on `qt5`, `qt5` gets renamed to `qt`, and then they run `brew upgrade`, will it download the new bottles? (What about if the user compiled the formulae from source, will Homebrew re-compile it?)

ilovezfs:
>If someone installs a formula that depends on qt5, qt5 gets renamed to qt, and then they run brew upgrade, will it download the new bottles?

Only if we also add or bump the `revision`.

In the case of a formula rename there will be an optlink create for the old name, so `/usr/local/opt/qt5` and `/usr/local/opt/qt` will both point to ` ../Cellar/qt/5.8.0_1`.

>What about if the user compiled the formulae from source, will Homebrew re-compile it?

Also only if there is a revision bump. So you will definitely want to revision bump everything in the qt4 tap after you change the formula name to qt@4 since you won't be able to have a formula_rename entry for the formula without directly conflicting with the /usr/local/opt/qt link that points to qt5.

cartr:
> In the case of a formula rename that also has an alias for the old name, there will be an optlink created for the old name, so `/usr/local/opt/qt5` and `/usr/local/opt/qt` will both point to `../Cellar/qt/5.8.0_1`.

Ah, I didn't know about that feature.

ilovezfs:
>Ah, I didn't know about that feature.

It was recently added. So ordinarily, this would be a less breaking change since you'd have both /usr/local/opt/qt and /usr/local/opt/qt@4 both pointing to the qt4 cellar. But since we're "reclaiming" the `qt` name and the `/usr/local/opt/qt` opt link, it'll be particularly important for you to revision bump everything due to the hard coding of paths you described.

ilovezfs (approved):
I think it's safe to ship this as-is and we can do the rest in follow-up PR(s)

cartr:
Would it be possible to wait a couple of days before merging this? Re-building the binary bottles for my tap will take considerable time on my end.

ilovezfs:
@cartr you could do it asynchronously by initially pushing revision bumps with the bottle blocks removed, and then re-adding the bottle blocks as the bottles get built.

cartr:
@ilovezfs I'm aware, but several of the Qt4-related formulae I package have very long build times and I'd like to avoid breaking people's Travis CI builds if possible. After the most-used packages (PyQt, PySide, Qt, WebKit, and their dependencies) are built I'll merge the name change in and we'll be good to go.

MikeMcQuaid:
@cartr Pushing a new build now and will likely pull when it's working. Apologies that I won't wait; we can't really set a precedent of holding PRs for things in taps we intentionally don't support, sorry 😭.

ilovezfs (approved):

MikeMcQuaid:
Thanks for review and help.

PR #8345, opened 2016-12-31
Fix audit errors related to :apr usage

GauthamGoli:

-----

This PR fixes `:apr is deprecated. Usage should be "apr-util"` audit error.

### DIFF
--- a/Formula/apib.rb
+++ b/Formula/apib.rb
@@ -16,7 +16,7 @@ class Apib < Formula
     sha256 "c538b9389b1849704720383f03fb2fe4eaf145ecd01d7dca5fd08d95e0e961ba" => :mountain_lion
   end
 
-  depends_on :apr => :build
+  depends_on "apr"
   depends_on "openssl"
 
   def install


Formula/ganglia.rb
```diff
@@ -19,7 +19,7 @@ class Ganglia < Formula
   end
 
   depends_on "pkg-config" => :build
-  depends_on "apr"
+  depends_on "apr-util"
```
ilovezfs: There's no need to change this since it was already fixed. `"apr"` is valid. The audit check refers to `:apr`.

Formula/apib.rb
```diff
@@ -16,7 +16,7 @@ class Apib < Formula
     sha256 "c538b9389b1849704720383f03fb2fe4eaf145ecd01d7dca5fd08d95e0e961ba" => :mountain_lion
   end
 
-  depends_on :apr => :build
+  depends_on "apr-util" => :build
```
ilovezfs: This should not be `=> :build` and you probably just need `"apr"`

DomT4: There's quite a lot of that in the core that's wrong. I guess whenever it was changed nobody picked up on/remembered the `=> :build` was always a hack that needed discarding whenever Homebrew's `apr` was  used.

DomT4:
If you feel like it, `tomcat-native` needs the option removed since it's pointless these days.

Formula/apib.rb
ilovezfs: also finding the audit text weird since it's recommending `depends_on "apr-util"` even though often `depends_on "apr"` is sufficient

MikeMcQuaid:
Thanks again @GauthamGoli!

PR #8310, opened 2016-12-30
mps-youtube 0.2.7.1 (new formula)

ed1d1a8d:
Adds a new formula for mps-youtube, a terminal based YouTube player
and downloader.

- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #8369, opened 2017-01-01
mps-youtube: fix (Add youtube_dl as resource)

aleksandrs-ledovskis:

-----

Fixes following warning on `mpsyt` launch:

  > WARNING:root:pafy: youtube-dl not found; falling back to internal backend.
  > This is not as well maintained as the youtube-dl backend
  > To hide this message, set the environmental variable PAFY_BACKEND to"internal".

From what I've inferred, having `youtube-dl` formula installed doesn't
satisfy `youtube_dl` python library dependency and placing it as package
is necessary for pafy's [`import`](https://github.com/mps-youtube/pafy/blob/66d0f262dd230a947015b3f1fa114551d3bc2bc1/pafy/pafy.py#L48) to find.

---

/cc @tonynater 
Follows: #8310

### DIFF
--- a/Formula/mps-youtube.rb
+++ b/Formula/mps-youtube.rb
@@ -14,7 +14,6 @@ class MpsYoutube < Formula
 
   depends_on "python3"
   depends_on "mpv" => :recommended
-  depends_on "youtube-dl" => :recommended
   depends_on "mplayer" => :optional
 
   resource "pafy" do
@@ -22,11 +21,18 @@ class MpsYoutube < Formula
     sha256 "11e0cb83bd9e636bc4d0d6f7d7ce964f4975c6f0e037fe285ef2acedafcf7bb2"
   end
 
+  resource "youtube_dl" do
+    url "https://pypi.python.org/packages/33/53/be5fd3d2e8b4af17cce81cdece45426856868692e2b7821a957108ff302e/youtube_dl-2016.12.31.tar.gz"
+    sha256 "cdba662fc6ff00b9b972da3bf4ac76d3db95841767871155a9518bfc6afb9a82"
+  end
+
   def install
     virtualenv_install_with_resources
   end
 
   test do
-    system "#{bin}/mpsyt", "/september,", "da 1,", "q"
+    Open3.popen3("#{bin}/mpsyt", "/september,", "da 1,", "q") do |_, _, stderr|
+      assert_empty stderr.read, "Some warnings were raised"
+    end
   end
 end


MikeMcQuaid:
> From what I've inferred, having youtube-dl formula installed doesn't
> satisfy youtube_dl python library dependency and placing it as package
> is necessary for pafy's import to find.

Can we try to figure out why this is not the case? I think it'd be better to not have this vendor.

aleksandrs-ledovskis:
@MikeMcQuaid 

Seems that `youtube-dl` formula includes only binary application & shell completions. 
Compare [formula's file listing](https://gist.github.com/aleksandrs-ledovskis/531f9d2b1b99c66bb85dab3d28030820) with [package's listing](https://gist.github.com/aleksandrs-ledovskis/7fd601c6aa50d52a434180bc6f768969) (building `youtube-dl` with `-s` flag didn't change anything.)

I'm not a _pythonista_, but this library lookup issue seems logical given binary-only nature of installed `youtube-dl`. Maybe you have a note of some good example Python based formula I could scavenge some tricks from?

aleksandrs-ledovskis:
I've found [this guide](https://github.com/Homebrew/brew/blob/master/docs/Python-for-Formula-Authors.md), however replacing `depends_on "youtube-dl" => :recommended` with either `depends_on "youtube_dl" => :python` or `depends_on "youtube_dl" => :python3` didn't help - an  aforementioned warning reappeard

MikeMcQuaid:
@aleksandrs-ledovskis We can't accept it as-is but @tdsmith may have a suggestion of how to resolve this (or if it's even worth the effort).

tdsmith:
The way we install Python applications into private virtualenvs isolates them from other environments, as a design goal; this isn't supposed to work. If something wants to use youtube-dl as a library, I'd imagined the client would vendor it instead of depending on the formula.

That said: we could drop a pth file into this formula's virtualenv to add the youtube-dl virtualenv to the module search path anyway. I'm not sure that's good policy in general since it sort of depends on an implementation detail but I don't really think it will cause trouble here. If we do, we should add a CI test that youtube-dl is being discovered correctly by this app.

tdsmith:
After actually looking at how youtube-dl gets installed, it turns out none of the previous comment applies! Sorry.

We install youtube-dl using its Makefile, which does sort of a funny thing that produces a single zipped binary that includes its entire Python environment; there are no importable artifacts on disk. Vendoring youtube-dl is the only reasonable thing to do here unless we want to change how youtube-dl is installed, which I don't feel a strong need to do, but would probably be fine.

tdsmith (approved):

JCount:
A pressing issue with vendoring youtube-dl like this in my view is the mere frequency at which youtube-dl itself has to be updated, meaning this would also have to follow the same breakneck update schedule. That said I have no solution to the actual issue at the heart of this PR.

tdsmith:
If youtube-dl actually _breaks_ regularly, this formula seems like an okay
reason to switch our youtube-dl to a flat install and depend on it here.

JCount:
I cannot comment on whether it truly "breaks" or not, but there have been 9 updates to the formula since the beginning of December, 2016. Which seems rather rapid when considering vendoring something. If `youtube-dl` could be turned into a required dependency that might be preferable, but I do not know my way around the homebrew's python virtual env very well, to know whether that would be feasible. 

@tdsmith you have much more knowledge there.

MikeMcQuaid:
I agree that that implies depending on it would be worthwhile. I also think fixing it up so it can be used here correctly is the right outcome 👍

zmwangx:
As a youtube-dl user: no, it does NOT break regularly. youtube-dl supports ~700 sites; just look at the [extractor](https://github.com/rg3/youtube-dl/tree/master/youtube_dl/extractor) directory. YouTube functionality alone is very stable, it doesn't change very much. Look at the [history of `youtube.py`](https://github.com/rg3/youtube-dl/commits/master/youtube_dl/extractor/youtube.py), only a couple of commits in the past few months, mostly non-consequential. This mps-youtube only deals with youtube as far as I can tell, so youtube-dl doesn't need to be updated often.

If youtube-dl upstream chooses to zipapp their stuff I don't see why we should do our own thing.

MikeMcQuaid:
> If youtube-dl upstream chooses to zipapp their stuff I don't see why we should do our own thing.

@zmwangx Because it doesn't install it in a form we can reuse in other formulae.

zmwangx:
> Because it doesn't install it in a form we can reuse in other formulae.

I don't think the  Python interface of youtube-dl is supposed to be reused in the first place. There's no public API documentation. The fact that it is on PyPI is probably due to demand (just my guess, take it with a grain of salt); I learnt from first hand experience that whenever you build an open source Python application, people will ask for it on PyPI, whether it's suitable or not.

Also, is there any other formula that needs to reuse it?

tdsmith:
The youtube-dl README gives examples of using it as a Python library so I guess it does have an API.

We actually don't even need to flatten youtube-dl; the zipapp is directly importable. Sort of a weird legacy feature. Try:
`PYTHONPATH=$(brew --prefix youtube-dl)/bin/youtube-dl python -c "import youtube_dl"`

To make it work, we can add something like this to mps-youtube:
`(libexec/"lib/python2.7/site-packages/youtube-dl.pth").write "#{Formula["youtube-dl"].opt_bin/"youtube-dl"}\n"`

zmwangx:
> The youtube-dl README gives examples of using it as a Python library so I guess it does have an API.

Yeah, never noticed that (very short) section, sorry.

tdsmith:
-- though I'd like to reiterate that in the general case depending on library functionality of Python apps delivered as formulas is not expected to work and vendoring is the expected solution.

There are two ways to interpret rapid updates as signal: every release of any library has the opportunity to break dependents, so rapid updates may be a signal that a library _should_ be vendored. (Looking at you, poppler.)

ilovezfs:
>We actually don't even need to flatten youtube-dl; the zipapp is directly importable. Sort of a weird legacy feature. Try:
PYTHONPATH=$(brew --prefix youtube-dl)/bin/youtube-dl python -c "import youtube_dl"

>To make it work, we can add something like this to mps-youtube:
(libexec/"lib/python2.7/site-packages/youtube-dl.pth").write "#{Formula["youtube-dl"].opt_bin/"youtube-dl"}\n"

Why is this making my stomach do somersaults?

ilovezfs:
I checked analytics. So in the last 30 days, there were 93 installs of mps-youtube, and there were over 71,000 install of youtube-dl. So really I'm happy with whatever as long as it does *not* involve changing the youtube-dl formula for the sake of mps-youtube.

aleksandrs-ledovskis:
> there were 93 installs of mps-youtube

Probably, because it was only added (in #8310) "5 days ago"? 😅 

---

Back on topic, what path do we take here?

**a)** Accept MR its current/simplest form.

_Pros:_  it will fix a warning and make almost a hundred users happy. Even though @MikeMcQuaid have vetoed it previously, this change seems sound enough to me and even [self-proclaimed] Pythonist @tdsmith. 

_Cons:_ In a long(er) term, YouTube might start to change its API's and that will require frequent(er?) update to `mps-youtube` formula definition.

**b)**  Rewrite MR using `(libexec/"lib/python2.7/site-packages/youtube-dl.pth").write "#{Formula["youtube-dl"].opt_bin/"youtube-dl"}\n"` quirk. 

_Pros:_  Updates & bugfixes to fetching backend will be auto-applied to `mps-youtube`

_Cons:_ Will introduce silent dependency on a way how `youtube-dl` installs itself. Any potential problems in newly released `youtube-dl` versions (which might have incompatibility with API backed) might also be introduced without knowledge.

**c)** Investigate some non-disruptive changes to `youtube-dl` formula

_Pros:_ Any other software/developer that wants to use `youtube_dl` Python library locally can do so freely.

_Cons:_ Will require some explaining to let that change MR pass.

ilovezfs:
(A) makes the most sense to me.

zmwangx:
> Cons: In a long(er) term, YouTube might start to change its API's and that will require frequent(er?) update to mps-youtube formula definition.

This ("frequent") doesn't make much sense. YouTube might change their API some day, which would require a youtube-dl update, and an update to the vendored resource in mps-youtube, if mps-youtube doesn't bump itself. But I don't see YouTube changing their API from day to day, or month to month.

ilovezfs:
Shipping as-is.

ilovezfs:
@aleksandrs-ledovskis thanks for the thoughtful work.

PR #8376, opened 2017-01-01
a*: import from homebrew/versions.

MikeMcQuaid:
Import all `a*` formulae from homebrew/versions. Any that do not build may be removed if they take too much effort to fix.


ilovezfs:
Here we go again.

MikeMcQuaid:
@ilovezfs https://www.youtube.com/watch?v=DSlSaGcc0QM

PR #8400, opened 2017-01-02
redis: fix launchd compatibility

antstorm:
-----

`launchd` will fail to start `redis` when config is set to `daemonize yes`. This small change will force `daemonize no` when running `redis` via `lauchd` (as well as `brew services`).

### DIFF
--- a/Formula/redis.rb
+++ b/Formula/redis.rb
@@ -62,6 +62,7 @@ class Redis < Formula
         <array>
           <string>#{opt_bin}/redis-server</string>
           <string>#{etc}/redis.conf</string>
+          <string>--daemonize no</string>
         </array>
         <key>RunAtLoad</key>
         <true/>


Formula/redis.rb
```diff
@@ -62,6 +62,7 @@ def plist; <<-EOS.undent
         <array>
           <string>#{opt_bin}/redis-server</string>
           <string>#{etc}/redis.conf</string>
+          <string>--daemonize no</string>
```
MikeMcQuaid: Should this be two parameters perhaps? `--daemonize` and `no`?

antstorm: Not sure, made sense to keep them together as these should never be separated, but I don't know what's the preferred style… quick search didn't reveal any established standards.

MikeMcQuaid:
Thanks @antstorm!

Issue #8392, opened 2017-01-02
qt5: cmake find_package(Qt5Core) doesn't work when linked to /usr/local


scpeters:
# Please follow the general troubleshooting steps first:

- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [X] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Bug reports:

I have a 3rd-party package that uses cmake to find qt5 with a `find_package(Qt5Core)` call. The `find_package` call works if the `CMAKE_PREFIX_PATH` is set to `/usr/local/opt/qt5` but fails if  `CMAKE_PREFIX_PATH` is set to `/usr/local` with the following console output:

~~~
cmake ..
-- The C compiler identification is AppleClang 7.0.2.7000181
-- The CXX compiler identification is AppleClang 7.0.2.7000181
-- Check for working C compiler: /usr/local/opt/ccache/libexec/cc
-- Check for working C compiler: /usr/local/opt/ccache/libexec/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/local/opt/ccache/libexec/c++
-- Check for working CXX compiler: /usr/local/opt/ccache/libexec/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /usr/local/lib/cmake/Qt5Core/Qt5CoreConfig.cmake:15 (message):
  The imported target "Qt5::Core" references the file

     "/usr/local/.//mkspecs/macx-clang"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/usr/local/lib/cmake/Qt5Core/Qt5CoreConfigExtras.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  /usr/local/lib/cmake/Qt5Core/Qt5CoreConfigExtras.cmake:50 (_qt5_Core_check_file_exists)
[...]

PR #8408, opened 2017-01-02
qt5: make keg-only again.

MikeMcQuaid:
Temporarily workaround for #8392 until we can get an upstream-suitable patch.


ilovezfs:
No brew link for you, qt5!

ilovezfs (approved):

PR #8468, opened 2017-01-04
osquery 2.2.1

ilovezfs:

-----

CC @theopolis

### DIFF
--- a/Formula/osquery.rb
+++ b/Formula/osquery.rb
@@ -3,9 +3,8 @@ class Osquery < Formula
   homepage "https://osquery.io"
   # pull from git tag to get submodules
   url "https://github.com/facebook/osquery.git",
-    :tag => "1.7.3",
-    :revision => "6901aa644a9bcc0667207008db71471abf756b82"
-  revision 7
+      :tag => "2.2.1",
+      :revision => "8fcb3659eeeac56847ee6d26b138b6bddc941a6a"
 
   bottle do
     sha256 "e5e53203da14c962be57dcab14948dbb8c177f14e4f810ce75561c5bf29937c6" => :sierra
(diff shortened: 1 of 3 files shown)


Formula/osquery.rb
```diff
@@ -122,9 +215,18 @@ def install
...
+      -Daws-cpp-sdk-firehose_library:FILEPATH=#{libexec}/aws-sdk-cpp/lib/libaws-cpp-sdk-firehose.a
+      -Daws-cpp-sdk-kinesis_library:FILEPATH=#{libexec}/aws-sdk-cpp/lib/libaws-cpp-sdk-kinesis.a
+      -Daws-cpp-sdk-sts_library:FILEPATH=#{libexec}/aws-sdk-cpp/lib/libaws-cpp-sdk-sts.a
+      -Dboost_filesystem-mt_library:FILEPATH=#{libexec}/boost/lib/libboost_filesystem-mt.a
+      -Dboost_regex-mt_library:FILEPATH=#{libexec}/boost/lib/libboost_regex-mt.a
+      -Dboost_system-mt_library:FILEPATH=#{libexec}/boost/lib/libboost_system-mt.a
+      -Dcppnetlib-client-connections_library:FILEPATH=#{libexec}/cpp-netlib/lib/libcppnetlib-client-connections.a
+      -Dcppnetlib-uri_library:FILEPATH=#{libexec}/cpp-netlib/lib/libcppnetlib-uri.a
+      -Dlinenoise_library:FILEPATH=#{libexec}/linenoise/lib/liblinenoise.a
+      -Dthrift_library:FILEPATH=#{libexec}/thrift/lib/libthrift.a
```
MikeMcQuaid: Could probably avoid installing all the `.a`s into `libexec` as they will be statically linked happily from the `buildpath`

MikeMcQuaid (approved):

Formula/osquery.rb
ilovezfs: The test needs to be able to reference them as well.

MikeMcQuaid: Cool. Might be worth considering a test that runs the application rather than building the library if possible.

ilovezfs:
@MikeMcQuaid PR refreshed.

theopolis:
If this is vendoring `cpp-netlib` perhaps its formula should be removed? Some of the headers it installs will not compile in downstream applications with boost 1.62+ and from what I can see only `osquery` was using it as a dependency.

ilovezfs:
Analytics show `cpp-netlib` has only 162 installs in the last thirty days, so it's indeed pretty marginal.

ilovezfs:
:green_apple: hurray

ilovezfs:
Shipped.

ilovezfs:
>If this is vendoring cpp-netlib perhaps its formula should be removed? Some of the headers it installs will not compile in downstream applications with boost 1.62+ and from what I can see only osquery was using it as a dependency.

@theopolis I've opened a PR to boneyard cpp-netlib here: https://github.com/Homebrew/homebrew-core/pull/8498.

PR #8498, opened 2017-01-04
cpp-netlib: migrate to boneyard

ilovezfs:
Created with `brew boneyard-formula-pr` because it's incompatible with Boost 1.63.0 and has no dependents. CC @theopolis @MikeMcQuaid.


MikeMcQuaid:
👍

PR #8509, opened 2017-01-05
wireguard-tools 0.0.20170105

zx2c4:
Fairly simple version bump. This release adds a new tool that's Linux only and a systemd unit, so we explicitly opt-out of those. We also set up bash-completion.


nijikon:

PR #8522, opened 2017-01-05
gnupg2: install gpg-zip in #{bin}

### LINKED ISSUE #8327: gpg-zip missing from gnupg2
Could gpg-zip be included in gnupg2? Or is there another way to get it? Thanks!

ilovezfs:
Also remove the lines deleting the gpg-zip man page, since upstream
stopped installing gpg-zip.1 in October 2015 according to the ChangeLog.

Closes #8327.


DomT4:
I'd probably just put it in `bin` personally. That's where it is for `gnupg` 1.x.

ilovezfs:
@DomT4 upstream specifically removed it from bin and yanked the man page so that feels a little weird ...

DomT4:
Fair comment. Just think `pkgshare` is a weird place for it. `libexec` is already filled with various utilities so maybe worth dumping in there?

ilovezfs:
@DomT4 let's just go with `bin`

Issue #8502, opened 2017-01-05
`qt5` Fails to Build on macOS v10.11.6 With Reported Options


RandomDSdevel:
### Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Bug reports:

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Well, _now_ I feel even _more_ stupid than I did _before_.  After reading the advice that @MikeMcQuaid left via [this comment](https://github.com/Homebrew/homebrew-core/issues/8342#issuecomment-269873212) in the mess that was [this issue](https://github.com/Homebrew/homebrew-core/issues/8342) just prior to resolving it, I apparently _forgot_ about it and went ahead and manually mucked about with Homebrew state by running `brew uninstall -vd --force --ignore-dependencies qt qt5` or some such insanity on my part…  That being said, however, I'm not entirely sure that's really the point here.  After all, I'm _pretty_ sure I would have seen the same problem crop up had I just waited for 1.1.6 to be released and bring [the relevant fix](https://github.com/Homebrew/brew/issues/1758) for my previous craziness to my local `stable` branch.  _Anyway_, what you _probably_ want to hear is the actual _bug report_, so I'd better write that out in the next paragraph once I'm done miming bashing my head into a wall in this one…
[...]

PR #8523, opened 2017-01-05
qt5: enable all options

ilovezfs:
intended to test #8502, and not necessarily intended to be merged

CC @MikeMcQuaid @DomT4


MikeMcQuaid:
Good idea to test. If this fails I'd probably like us to remove this option and not re-add it.

DomT4:
Seems to have built okay on CI at least.

ilovezfs:
So it looks like #8502 may actually be about the `--with-examples` option: https://github.com/Homebrew/homebrew-core/issues/8502#issuecomment-270740472

ilovezfs:
Looks like `--with-examples` is also fine except for minor audit whining:
```
==> brew audit qt5 --online
==> FAILED
Error: 1 problem in 1 formula
qt5:
    Broken dylib links found:
      libgruesensor.1.dylib

```

ilovezfs:
OK, let's try all options enabled. :spaghetti:

ilovezfs:
Build with all options enabled was also fine. Poor qt5 and it's falsely accused, much maligned and scapegoated options. :goat:

DomT4:
Possibly the first time a breakage wasn't explicitly the `qt5` formula's fault 😉.

PR #8577, opened 2017-01-07
python3: fix caveats when python3 not installed

mistydemeo:
eec96a68d82 updated the caveats to look at the actual contents of the
prefix, but this doesn't work if either

a) python3 is not installed, or
b) an outdated version is installed

The result was that some users were getting exceptions raised on any
operations that evaluated `#caveats`.

-----

### DIFF
--- a/Formula/python3.rb
+++ b/Formula/python3.rb
@@ -314,7 +314,11 @@ class Python3 < Formula
   end
 
   def caveats
-    xy = (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+    if prefix.exist?
+      xy = (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+    else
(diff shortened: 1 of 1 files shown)


mistydemeo:
The test failures are from dependents, and weren't caused by this branch.

Formula/python3.rb
```diff
@@ -314,7 +314,11 @@ def sitecustomize
   end
 
   def caveats
-    xy = (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+    if prefix.exist?
+      xy = (prefix/"Frameworks/Python.framework/Versions").children.first.basename.to_s
+    else
+      xy = version.to_s.slice(/(3\.\d)/) || "3.6"
```
MikeMcQuaid: This isn't going to be correct and may be missed when doing a version bump (which is why the original refactoring was done). Could we consider just omitting the relevant lines of the caveats instead?

PR #8593, opened 2017-01-07
qpm: fix build

ilovezfs:
git submodules were competing with the go resouces


ilovezfs:
@BrewTestBot test this please

PR #8597, opened 2017-01-07
gmp: use --build configure option

### LINKED ISSUE #7054: GMP was compiled with -march=sandybridge

ilovezfs:
and remove --disable-assembly.

CC @smtrat @MikeMcQuaid @fxcoudert

Closes #7054.

### DIFF
--- a/Formula/gmp.rb
+++ b/Formula/gmp.rb
@@ -16,11 +16,8 @@ class Gmp < Formula
 
   def install
     ENV.cxx11 if build.cxx11?
-    args = ["--prefix=#{prefix}", "--enable-cxx"]
-
-    # https://github.com/Homebrew/homebrew/issues/20693
-    args << "--disable-assembly" if build.bottle?
-
+    args = %W[--prefix=#{prefix} --enable-cxx]
+    args << "--build=core2-apple-darwin#{`uname -r`.to_i}" if build.bottle?
     system "./configure", *args
     system "make"
     system "make", "check"


Formula/gmp.rb
```diff
@@ -16,12 +16,13 @@ class Gmp < Formula
 
   def install
     ENV.cxx11 if build.cxx11?
-    args = ["--prefix=#{prefix}", "--enable-cxx"]
 
-    # https://github.com/Homebrew/homebrew/issues/20693
-    args << "--disable-assembly" if build.bottle?
+    # XNU release major version
+    osmajor = MacOS.version.to_s[3..-1].to_i + 4
```
MikeMcQuaid: What does the output end up being? Should rather shell out for this if we can; fairly opaque as-is (not your fault)

Formula/gmp.rb
```diff
@@ -16,12 +16,13 @@ class Gmp < Formula
...
 
-    system "./configure", *args
+    system "./configure", "--prefix=#{prefix}",
+                          "--enable-cxx",
+                          "--build=core2-apple-darwin#{osmajor}"
```
MikeMcQuaid: Let's just do this when `build.bottle?`

MikeMcQuaid (changes requested):

Formula/gmp.rb
ilovezfs: This is how XNU is versioned
10.6 -> 10
10.7 -> 11
10.8 -> 12
10.9 -> 13
...
10.12 -> 16

There's no reason to shell out if we know the answer.

MikeMcQuaid: Other formula generally use for this
```
ruby
`uname -r`.chomp
```
I'd be 👍 on adding a Homebrew/brew method that exposed this too.

ilovezfs: That's the full version. `uname -r`.chomp
=> "12.6.0"

So you still end up having to do string manipulations then to extract the major version. Running processes for answers we already know is pretty gross lol

ilovezfs:
pr refreshed with useless shell invocation

MikeMcQuaid (approved):

PR #8617, opened 2017-01-08
[obsolete] ldc 1.1.0-beta6

timotheecour:
- [ x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [ x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?

-----
Updates ldc to 1.1.0 branch, and also fixes https://github.com/Linuxbrew/homebrew-core/issues/852

```
brew audit --strict ldc
ldc:
  * devel version 1.1.0-beta3 is older than stable version 1.1.0
Error: 1 problem in 1 formula
```

But not sure what else to put here since latest tar.gz release in https://github.com/ldc-developers/ldc/releases/ would be still older than 1.1.0 branch name

NOTE: it supersedes https://github.com/Linuxbrew/homebrew-core/pull/1525

### DIFF
--- a/Formula/ldc.rb
+++ b/Formula/ldc.rb
@@ -5,8 +5,10 @@ class Ldc < Formula
   stable do
     # for the sake of LLVM 3.9 compatibility
     url "https://github.com/ldc-developers/ldc.git",
-        :branch => "release-1.1.0"
-    version "1.1.0"
+        :branch => "release-1.1.0",
+        :revision => "a86f87637d4880709ee0bde3ae741e6617b835b2"
+    # NOTE sure what suffix should be after 1.1.0
+    version "1.1.0-alpha1"
 
     resource "ldc-lts" do
       url "https://github.com/ldc-developers/ldc/releases/download/v0.17.2/ldc-0.17.2-src.tar.gz"
--- a/Formula/ldc.rb
+++ b/Formula/ldc.rb
@@ -5,9 +5,8 @@ class Ldc < Formula
   stable do
     # for the sake of LLVM 3.9 compatibility
     url "https://github.com/ldc-developers/ldc.git",
-        :branch => "release-1.0.1",
-        :revision => "3461e00f3531f855f9fc6e92515d7affb8201827"
-    version "1.0.1-alpha1"
+        :branch => "release-1.1.0"
+    version "1.1.0"
 
     resource "ldc-lts" do
       url "https://github.com/ldc-developers/ldc/releases/download/v0.17.2/ldc-0.17.2-src.tar.gz"
--- a/Formula/ldc.rb
+++ b/Formula/ldc.rb
@@ -5,10 +5,8 @@ class Ldc < Formula
   stable do
     # for the sake of LLVM 3.9 compatibility
     url "https://github.com/ldc-developers/ldc.git",
-        :branch => "release-1.1.0",
-        :revision => "a86f87637d4880709ee0bde3ae741e6617b835b2"
-    # NOTE sure what suffix should be after 1.1.0
-    version "1.1.0-alpha1"
+        :tag => "v1.1.0-beta6"
+    version "1.1.0-beta6"
 
     resource "ldc-lts" do
       url "https://github.com/ldc-developers/ldc/releases/download/v0.17.2/ldc-0.17.2-src.tar.gz"
--- a/Formula/ldc.rb
+++ b/Formula/ldc.rb
@@ -22,17 +22,6 @@ class Ldc < Formula
     sha256 "8d827625dea278303befcbb997d3e1713b7d85e257705a30da72ce1519eab147" => :yosemite
   end
 
-  devel do
-    url "https://github.com/ldc-developers/ldc/releases/download/v1.1.0-beta3/ldc-1.1.0-beta3-src.tar.gz"
-    sha256 "cf4aeb393eada610aa3bad18c3ae6a5de94250eaa968fe2d1b0a6afdf8ea54f6"
-    version "1.1.0-beta3"
-
-    resource "ldc-lts" do
-      url "https://github.com/ldc-developers/ldc/releases/download/v0.17.2/ldc-0.17.2-src.tar.gz"
-      sha256 "8498f0de1376d7830f3cf96472b874609363a00d6098d588aac5f6eae6365758"
-    end
-  end
-
   head do
     url "https://github.com/ldc-developers/ldc.git", :shallow => false
 


dunn:
> But not sure what else to put here since latest tar.gz release in https://github.com/ldc-developers/ldc/releases/ would be still older than 1.1.0 branch name

If there's no new development release, you can remove the `devel do` block entirely.

timotheecour:
> If there's no new development release, you can remove the devel do block entirely.
View changes

Done.

NOTE: 1.1.0 is not released yet, so I named it (rather arbitrarily) version "1.1.0-alpha1"
Pleaese let me know if you have a better suggestion or if it's ok

Formula/ldc.rb
```diff
@@ -5,8 +5,10 @@ class Ldc < Formula
   stable do
     # for the sake of LLVM 3.9 compatibility
     url "https://github.com/ldc-developers/ldc.git",
-        :branch => "release-1.1.0"
-    version "1.1.0"
+        :branch => "release-1.1.0",
+        :revision => "a86f87637d4880709ee0bde3ae741e6617b835b2"
+    # NOTE sure what suffix should be after 1.1.0
+    version "1.1.0-alpha1"
```
dunn: Sorry; if this isn't a stable release, it should remain as the devel version.

timotheecour: there's a catch 22 situation:
* initially I had made this as a linuxbrew only change https://github.com/Linuxbrew/homebrew-core/pull/1525 because it fixes a long standing issue (brew install ldc has been broken for months https://github.com/Linuxbrew/homebrew-core/issues/852)
* but @maxim-belkin  told me to make this a homebrew PR which I did
* if this remains as devel version, linuxbrew ldc remains broken
* furthermore, before this PR, the code was:

:branch => "release-1.0.1",
```
:branch => "release-1.0.1",
    :revision => "3461e00f3531f855f9fc6e92515d7affb8201827"
    version "1.0.1-alpha1"
```

so the `stable` branch was already not a stable release. 
This PR doesn't make anything worse, and it fixes linuxbrew ldc which was broken for months.

Please advise

timotheecour (commented):
see catch-22 comment

maxim-belkin:
> This PR doesn't make anything worse, and it fixes linuxbrew ldc which was broken for months.

Linuxbrew/homebrew-core#1525 was not Linux-specific and, therefore, was redirected here.

timotheecour:
ok please let me know if you want me to change anything else.

MikeMcQuaid:
May as well make the stable release LDC 1.1.0 beta 6 rather than using an alpha?

timotheecour:
> May as well make the stable release LDC 1.1.0 beta 6 rather than using an alpha?

done

ilovezfs:
I think we have some confusion here. I am the one who called it "alpha" since upstream was saying they would tag an actual 1.0.1 release, as opposed to only committing the LLVM 3.9 fixes to the 1.0.1 branch. The fact that hasn't happened does not imply that the 1.1.0 branch is suddenly non-beta.

The stable spec should be left alone. The devel spec should be upgraded to 1.1.0-beta6.

timotheecour:
The stable spec is broken, so on linux `brew install ldc` has been broken for *months*

ilovezfs:
@timotheecour you should let upstream know and see if they're willing to backport the Linux fixes to the 1.0.1 branch.

timotheecour:
The whole point of this was to fix https://github.com/Linuxbrew/homebrew-core/issues/852 ...

timotheecour:
> you should let upstream know and see if they're willing to backport the Linux fixes to the 1.0.1 branch.

ping @klickverbot what do you suggest here...

ilovezfs:
Linuxbrew is welcome to jump the gun and designate a beta as stable rather than pursuing a backport to the stable branch. But that's not a reason for us to do that in Homebrew core if the problem has no impact on macOS and 1.1.0 is still considered beta by upstream.

timotheecour:
I give up, I'll just use my own tap.

timotheecour:
> Linuxbrew is welcome to jump the gun and designate a beta as stable
before this PR, stable was already using an alpha...

> that's not a reason for us to do that in Homebrew core
@maxim-belkin @rwhogg looks like this PR is problematic in homebrew-core. Initially I had suggested doing it in linuxbrew [https://github.com/Linuxbrew/homebrew-core/pull/1525] but you suggested moving it here. Shall I move it back to linuxbrew after all? ldc was broken for months and this would finally fix it... I don't understand the pushback

ilovezfs:
>before this PR, stable was already using an alpha...

As I said above, " I am the one who called it "alpha" since upstream was saying they would tag an actual 1.0.1 release."

rwhogg:
> @maxim-belkin @rwhogg looks like this PR is problematic in homebrew-core. Initially I had suggested doing it in linuxbrew Linuxbrew#1525 but you suggested moving it here. Shall I move it back to linuxbrew after all? ldc was broken for months and this would finally fix it... I don't understand the pushback

My apologies both to @timotheecour and to the Homebrew maintainers here: I'm not familiar with `ldc` or its development process and I didn't know that 1.1.0 is currently not considered "stable".

That said:

Linuxbrew has made certain tradeoffs in order to remain in sync with Homebrew. Some of those mean that some formulae are going to suffer more than others. The alternative is that we start diverging in versions and suddenly on merges there's even less certainty than there already is about how to proceed.

> But that's not a reason for us to do that in Homebrew core if the problem has no impact on macOS and 1.1.0 is still considered beta by upstream.

I'm not a Homebrew maintainer, but I nevertheless agree with ilovezfs' argument.

I was often confused and frustrated by this early on too, but I eventually came to understand Homebrew's position. Pretend, for a moment, that Homebrew was a fork of Linuxbrew instead of the other way around, and that Linuxbrew only supported Linux. Would Linuxbrew want to accept all sorts of PR's that fix issues only relevant on macOS, especially if they risked making the Linux experience worse? I'd say "no" and I'd probably make the same sorts of arguments ilovezfs did.

Once again, I'm sorry you've had a bad experience with this.

@sjackman, is there anything else that should be said?

ilovezfs:
For what it's worth, Debian stable is on 0.14.0: https://packages.debian.org/jessie/ldc

maxim-belkin:
> I give up, I'll just use my own tap.

Never give up: look at my PR to update llvm to 3.8.0 :)

From what I understand, "stable" ldc builds fine on macOS but not on Linux? This should be made clear when submitting a PR to Linuxbrew. If that's correct, reopen Linuxbrew/homebrew-core#1525 and add a comment to the line `:branch` about that.

sjackman:
> @sjackman, is there anything else that should be said?

Only that `ldc` stable is currently broken on Linux, and we need to resolve that situation. The simples possible fix is to backport in a `patch` the commit that makes it possible to build `stable` on Linux if that's a simple commit that applies easily to the current stable release. If not, then perhaps promoting the `devel` release to `stable` on Linuxbrew is the way to go.

timotheecour:
=> moved to https://github.com/Linuxbrew/homebrew-core/pull/1534/files

sjackman:
The current stable release of `ldc` in Homebrew is `1.0.1-alpha1`, which was never released by upstream.  Would Homebrew consider updating stable to `1.1.0-beta6`, which has been released by upstream, or do you prefer to wait for a stable release of 1.1.0?

ilovezfs:
1.0.1-alpha1 is 1.0.0 with the LLVM 3.9 patch, so I'd lean toward waiting for the actual 1.1.0 before promoting 1.1.* to stable. It might be worth pestering upstream about when 1.1.0 is actually going to be released.

PR #8618, opened 2017-01-08
nim 0.16.0 + add linenoise option

subsetpark:

-----

Add option to build the nim compiler with linenoise for readline-like
behavior in the REPL (nim secret).

### DIFF
--- a/Formula/nim.rb
+++ b/Formula/nim.rb
@@ -12,7 +12,7 @@ class Nim < Formula
     sha256 "602c726784ef9fb42538ac9945c8fad86c4aacdd8a9689f71e04a4d5f9c951e2" => :yosemite
   end
 
-  option "with-linenoise", "Build the compiler with linenoise to support readline keybindings"
+  option "without-linenoise", "Leave out readline keybindings via linenoise"
 
   resource "nimble" do
     url "https://github.com/nim-lang/nimble/archive/v0.7.10.tar.gz"
--- a/Formula/nim.rb
+++ b/Formula/nim.rb
@@ -1,9 +1,8 @@
 class Nim < Formula
   desc "Statically typed, imperative programming language"
   homepage "http://nim-lang.org/"
-  url "http://nim-lang.org/download/nim-0.15.2.tar.xz"
-  sha256 "905df2316262aa2cbacae067acf45fc05c2a71c8c6fde1f2a70c927ebafcfe8a"
-  head "https://github.com/nim-lang/Nim.git", :branch => "devel"
+  url "http://nim-lang.org/download/nim-0.16.0.tar.xz"
+  sha256 "9e199823be47cba55e62dd6982f02cf0aad732f369799fec42a4d8c2265c5167"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


woodruffw:
@subsetpark Do you know if other distributions ship nim with this enabled? If so, we might want to enable it by default and do `build.without?` instead.

subsetpark:
@woodruffw I don't believe so:
- It's not required on Windows;
- `koch boot`, the standard compiler bootstrap command, requires the `useLinenoise` flag to include it.

ilovezfs:
https://github.com/gentoo/gentoo/blob/master/dev-lang/nim/nim-0.15.2.ebuild#L33

Would it be better to `depends_on "readline"`?

woodruffw:
@subsetpark Got it, thanks for checking!

subsetpark:
@woodruffw the standard tarball available on nim-lang.org comes with C source files, and a shell script to compile them (which the recipe runs); the functionality in this option is *not* available in those C files. Hence the recipe effectively has to generate new sources.

subsetpark:
@ilovezfs `useGnuReadline` is actually an older flag that is now a synonym for `useLinenoise`; the project moved over to linenoise in order to drop that dependency.

ilovezfs:
So the gentoo distro enables this by default.
https://github.com/gentoo/gentoo/blob/master/dev-lang/nim/nim-0.15.2.ebuild#L17

subsetpark:
We definitely could too; it doesn't introduce any other dependencies. Basically on the pro side:
- You get to use readline commands in the repl

On the against side:
- It introduces two new compilation steps into the build;
- It's a pretty niche feature (The 'repl' is actually an undocumented command -- literally, `nim secret`; the actual nim compiler functions identically)

As a non-member I couldn't speak nearly well enough to the Homebrew 'philosophy' as to which side to favor.

ilovezfs:
Since it doesn't add dependencies and it doesn't sound like anyone would have a reason to _want_ a build without it, it seems worth the extra build steps to avoid adding an option to the formula.

subsetpark:
OK, I'll adjust.

 Today also has a new  release: 0.16.0. Shall I update the  versionin the same PR?

woodruffw:
Yeah, adding 0.16.0 to this PR sounds good to me. No point in having two rebottles when we can make do with one :smile:

ilovezfs:
>Make linenoise opt out

Let's just turn it on unconditionally without having an opt-out.

subsetpark:
@woodruffw I added 0.16.0. That update deprecates a few things that the previous version was relying on, so there's some relatively substantial changes to the install process for both `head` and stable.

Formula/nim.rb
```diff
@@ -63,6 +61,6 @@ def install
       license = "MIT"
       requires "nim >= 0.15.0"
     EOS
-    assert_equal "name: \"hello\"", shell_output("#{bin}/nimble dump").split("\n")[1].chomp
+    assert_equal "name: \"hello\"", shell_output("#{bin}/nimble dump").split("\n")[0].chomp
```
MikeMcQuaid: `shell_output("#{bin}/nimble dump").lines.first` would be preferable.

MikeMcQuaid (approved):

Formula/nim.rb
nijikon: @subsetpark can you apply this please? I want to merge and then do #8647

subsetpark: @nijikon All done.

subsetpark: @nijikon This PR also takes care of the upgrade to 0.16.0 however.

nijikon:
@subsetpark Oh, you updated to `0.16.0`, sweet 👍

nijikon (approved):

nijikon:

Issue #8447, opened 2017-01-03
diffoscope: Depends on findutils with default names


rwhogg:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Bug reports:

diffoscope fails at runtime (it builds fine) with the following error if `findutils` is not built with default names:

```bash
% diffoscope docs docs.current
Traceback (most recent call last):
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/__main__.py", line 146, in main
    sys.exit(run_diffoscope(parsed_args))
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/__main__.py", line 122, in run_diffoscope
    parsed_args.file1, parsed_args.file2)
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/comparators/__init__.py", line 86, in compare_root_paths
    return compare_directories(path1, path2)
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/comparators/directory.py", line 95, in compare_directories
    return FilesystemDirectory(path1).compare(FilesystemDirectory(path2))
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/comparators/directory.py", line 130, in compare
    find_diff = Difference.from_command(FindAll, self.path, other.path)
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/difference.py", line 350, in from_command
    difference = Difference.from_feeder(feeder1, feeder2, path1, path2, *args, **kwargs)
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/difference.py", line 300, in from_feeder
    unified_diff = diff(feeder1, feeder2)
  File "/usr/local/Cellar/diffoscope/42_1/libexec/lib/python3.6/site-packages/diffoscope/difference.py", line 269, in diff
[...]

PR #8622, opened 2017-01-08
diffoscope 67

ilovezfs:


ilovezfs:
Fixes https://github.com/Homebrew/homebrew-core/issues/8447. CC @rwhogg

rwhogg:
Thanks a bunch @ilovezfs! Works for me now too.

PR #8630, opened 2017-01-09
Scala 2.10 and 2.11 (new formulas)

qayshp:

-----

My organization (and likely at least a few others) are still using `Scala 2.11` for legacy projects, and noticed when it was removed in [homebrew-versions bb202ed](https://github.com/Homebrew/homebrew-versions/commit/bb202eda53e7a9174caee91ad6c5c7b2d01b61e9). Somewhat surprizingly to me, `scala210` is still in `homebrew-versions`. We have set up our own tap for now, as suggested by @MikeMcQuaid in [this comment](https://github.com/Homebrew/homebrew-versions/commit/ba8b293656ef8fb51eca2d3be022e69ad20277f4#commitcomment-20383905). I'm not sure what analytics are used to determine that a formula is "used by a very small number of people", and if you don't think keeping these formula around is worth it, I would completely understand.

With that said:

This PR adds `Scala 2.11` almost exactly as it was from [`homebrew-versions` `scala211.rb`](https://github.com/Homebrew/homebrew-versions/blob/2bbcc3f2009e55230d41d9c07dd538163350d8fb/scala211.rb).

This PR also updates the `Scala 2.10` formula from [`homebrew-versions` `scala210.rb`](https://github.com/Homebrew/homebrew-versions/blob/838311f4f825af653315c3f8511bbbad1bf6e5c2/scala210.rb) from homebrew-versions from `2.10.5` to `2.10.6` since `2.10` is still in homebrew-versions, and homebrew-versions is now deprecated. This would allow the `scala210` formula to be removed from there.

Both `2.10` and `2.11` formulas are added in this one PR because otherwise `brew audit --strict <formula>` will fail as a result of the `conflicts_with` in each formula pointing at a formula that wouldn't yet exist.

### DIFF
--- a/Formula/scala@2.10.rb
+++ b/Formula/scala@2.10.rb
@@ -1,0 +1,65 @@
+class ScalaAT210 < Formula
+  desc "JVM-based programming language"
+  homepage "https://www.scala-lang.org/"
+  url "http://downloads.lightbend.com/scala/2.10.6/scala-2.10.6.tgz"
+  mirror "https://downloads.typesafe.com/scala/2.10.6/scala-2.10.6.tgz"
+  mirror "http://www.scala-lang.org/files/archive/scala-2.10.6.tgz"
+  sha256 "54adf583dae6734d66328cafa26d9fa03b8c4cf607e27b9f3915f96e9bcd2d67"
+
+  option "with-docs", "Also install library documentation"
+  option "with-src", "Also install sources for IDE support"
+
+  depends_on :java => "1.6+"
+
+  conflicts_with "scala", :because => "Differing version of same formula"
+  conflicts_with "scala@2.11", :because => "Differing version of same formula"
+
+  resource "docs" do
+    url "https://downloads.lightbend.com/scala/2.10.6/scala-docs-2.10.6.txz"
+    mirror "http://www.scala-lang.org/files/archive/scala-docs-2.10.6.txz"
+    sha256 "e9b5694255607ba069dcc0faa3ab1490164115ae000129c03100b196fce2025a"
+  end
+
+  resource "src" do
+    url "https://github.com/scala/scala/archive/v2.10.6.tar.gz"
+    sha256 "06d7467ff628ebac615c5e60af155e0b4cbbf4c31d10c03a45e9615d5b1e0420"
+  end
+
+  resource "completion" do
+    url "https://raw.githubusercontent.com/scala/scala-tool-support/0a217bc/bash-completion/src/main/resources/completion.d/2.9.1/scala"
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
> I'm not sure what analytics are used to determine that a formula is "used by a very small number of people"

http://docs.brew.sh/Analytics.html. If you or anyone else in your organisation are opting out, I'd advise not doing so.

> This would allow the scala210 formula to be removed from there.

Could you create a homebrew/versions PR to do that, too? That'd be super helpful.

Additionally, add to `tap_migrations.json` in the homebrew/versions PR and the `formula_renames.json` file in this PR.

> if you don't think keeping these formula around is worth it, I would completely understand.

I'm fine with adding them, thanks for the explanation 👍 

Thanks!

Formula/scala@2.10.rb
```diff
@@ -0,0 +1,68 @@
+class ScalaAT210 < Formula
+  desc "JVM-based programming language"
+  homepage "https://www.scala-lang.org/"
+  url "http://downloads.lightbend.com/scala/2.10.6/scala-2.10.6.tgz"
+  mirror "https://downloads.typesafe.com/scala/2.10.6/scala-2.10.6.tgz"
+  mirror "http://www.scala-lang.org/files/archive/scala-2.10.6.tgz"
+  sha256 "54adf583dae6734d66328cafa26d9fa03b8c4cf607e27b9f3915f96e9bcd2d67"
+
+  bottle do
+  end
```
MikeMcQuaid: Can remove this empty bottle block.

MikeMcQuaid (changes requested):

qayshp:
I have done what was suggested above.

- created [PR 1507 on `homebrew/versions`](https://github.com/Homebrew/homebrew-versions/pull/1507) to remove `scala210` and migrate `scala211`.
-- Removes `scala210` formula
-- Marks `scala211` and `scala210` as migrated to `homebrew/core`
- Updated this PR with rebased commits of the `*.rb` files after removing the empty 'bottle' block.
- Updated this PR with `formula_renames.json` for `scala210` and `scala211` migrations.

I tried my best to follow commit message conventions I saw in git blame on the files I was modifying, but please let me know if anything needs to be changed.

Thanks!

qayshp:
I hope I addressed everything, but let me know if there are any other changes you would suggest.

MikeMcQuaid:
Sorry for the delay, have been away on a work trip, will take a look at this next week.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #8659, opened 2017-01-09
mysql.rb: Look for my.cnf in both /etc /etc/mysql

sjackman:
Both directories are searched by default.
See http://dev.mysql.com/doc/refman/5.7/en/option-files.html


nijikon:

sjackman:
Thanks for merging, Tomasz!

MikeMcQuaid:
What if they both exist 😉

sjackman:
Fire and brimstone. 🔥

Issue #7852, opened 2016-12-14
Installing universal boost fails with "Failed changing install name"


omegahm:
### Bug reports:

When reinstalling boost with universal, I get an error with "Failed changing install name"

```bash
$ brew reinstall boost --universal
==> Reinstalling boost --universal
==> Using the sandbox
==> Downloading https://downloads.sourceforge.net/project/boost/boost/1.62.0/boost_1_62_0.tar.bz2
Already downloaded: /Users/ohm/Library/Caches/Homebrew/boost-1.62.0.tar.bz2
==> ./bootstrap.sh --prefix=/usr/local/Cellar/boost/1.62.0 --libdir=/usr/local/Cellar/boost/1.62.0/lib --without-icu --without-libraries=python,context,coroutine,mpi
==> ./b2 headers
==> ./b2 --prefix=/usr/local/Cellar/boost/1.62.0 --libdir=/usr/local/Cellar/boost/1.62.0/lib -d2 -j4 --layout=tagged --user-config=user-config.jam install threading=multi,single link=shared,static address-mod
Error: Failed changing install name in /usr/local/Cellar/boost/1.62.0/lib/libboost_chrono-mt.dylib
  from libboost_system-mt.dylib
    to @loader_path/libboost_system-mt.dylib
Error: Failed to fix install linkage
The formula built, but you may encounter issues using it or linking other
formula against it.
==> Caveats

Building of Boost.Context and Boost.Coroutine is disabled as they are
only supported on x86_64.
==> Summary
🍺  /usr/local/Cellar/boost/1.62.0: 12,401 files, 661.7M, built in 25 minutes
```

Installing boost without the universal binary works "fine", but then I cannot use both x64 and i386 boost, when compiling my own stuff.

I have also tried with various combinations of `--with-icu4c`, `--with-mpi` and `--universal`, which all fails in different ways.

Log-files are present here: https://gist.github.com/anonymous/e1f7a08dc18cc1102608e847a1d2d6c7

MikeMcQuaid:
CC @woodruffw for thoughts. This may just result in us dropping `--universal` for `boost`.

woodruffw:
I'm building boost now to test this locally, but my bet is that they've just removed i386 support for Boost.Chrono. If that's the case, our options are disabling `--universal` entirely or just selectively disabling Chrono when `--universal` is toggled (just as we do with Boost.Context and Boost.Coroutine currently).

woodruffw:
[...]

PR #8660, opened 2017-01-09
boost: Disable universal builds.

woodruffw:

-----

Previously, universal builds had removed several modules from i386/ppc
targets to get the build to succeed. This is becoming increasingly
unsustainable as i386 and PPC age and are no longer supported.

See #7852.

### DIFF
--- a/Formula/boost.rb
+++ b/Formula/boost.rb
@@ -14,7 +14,6 @@ class Boost < Formula
 
   env :userpaths
 
-  option :universal
   option "with-icu4c", "Build regexp engine with icu support"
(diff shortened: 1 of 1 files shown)


woodruffw:
`ncmpcpp`'s test is failing on El Cap and Sierra. Not sure if that's a side effect of `boost` changing, since it's just a `--version` test and `ncmpcpp` isn't built as a universal binary...

MikeMcQuaid:
If it doesn't fail for you locally before or after the PR I'd say just :shipit:.

PR #8661, opened 2017-01-09
ios-webkit-debug-proxy HEAD

artygus:
Add HEAD to ios-webkit-debug-proxy formula


Formula/ios-webkit-debug-proxy.rb
```diff
@@ -26,4 +28,8 @@ def install
     system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}"
     system "make", "install"
   end
+
+  test do
+    system "#{bin}/ios_webkit_debug_proxy", "--help"
```
woodruffw: Any chance of a more in-depth test here?

artygus: @woodruffw I don't think so. The main purpose of the utility is to proxy requests from usbmuxd daemon, which could be either started or not and a running device, or at least iOS Simulator

woodruffw: Got it, thanks for the explanation.

nijikon:

PR #8664, opened 2017-01-09
lrdf 0.6.1

ilovezfs:
remove unused openssl dependency
use autogen.sh
add a test


dunn:
Everything up to and including `configure` can be replaced with autogen.sh now https://github.com/swh/LRDF/blob/master/autogen.sh

PR #8677, opened 2017-01-10
Apache Ant 1.9 (new formula)

stringbean:
Version of ant formula for Apache Ant 1.9.

-----
Apache Ant 1.10.0 (provided by the `ant` formula) is the first version which requires Java 8. This adds a 1.9 version formula for developers that still need Java 7 (or below) support. The Ant team are still maintaining the 1.9.x branch and 1.9.8 was released at the same time as 1.10.0.

### DIFF
--- a/Formula/ant@1.9.rb
+++ b/Formula/ant@1.9.rb
@@ -5,12 +5,10 @@ class AntAT19 < Formula
   version "1.9"
   sha256 "6edeaab09fc0bb1eeeb6b6be9c0e463813b073c40189f760e70c85fa288d5125"
 
-  keg_only :provided_by_osx if MacOS.version < :mavericks
+  keg_only "Older version of formula"
 
   option "with-ivy", "Install ivy dependency manager"
   option "with-bcel", "Install Byte Code Engineering Library"
-
-  conflicts_with "ant", :because => "Different versions of same formula"
 
   resource "ivy" do
     url "https://www.apache.org/dyn/closer.cgi?path=ant/ivy/2.4.0/apache-ivy-2.4.0-bin.tar.gz"
--- a/Formula/ant@1.9.rb
+++ b/Formula/ant@1.9.rb
@@ -5,7 +5,7 @@ class AntAT19 < Formula
   version "1.9"
   sha256 "6edeaab09fc0bb1eeeb6b6be9c0e463813b073c40189f760e70c85fa288d5125"
 
-  keg_only "Older version of formula"
+  keg_only :versioned_formula
 
   option "with-ivy", "Install ivy dependency manager"
   option "with-bcel", "Install Byte Code Engineering Library"
--- a/Formula/ant@1.9.rb
+++ b/Formula/ant@1.9.rb
@@ -7,19 +7,6 @@ class AntAT19 < Formula
 
   keg_only :versioned_formula
 
-  option "with-ivy", "Install ivy dependency manager"
-  option "with-bcel", "Install Byte Code Engineering Library"
-
-  resource "ivy" do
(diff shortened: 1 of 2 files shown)


Formula/ant@1.9.rb
```diff
@@ -0,0 +1,69 @@
+class AntAT19 < Formula
+  desc "Java build tool"
+  homepage "https://ant.apache.org/"
+  url "https://www.apache.org/dyn/closer.cgi?path=ant/binaries/apache-ant-1.9.8-bin.tar.bz2"
+  version "1.9"
+  sha256 "6edeaab09fc0bb1eeeb6b6be9c0e463813b073c40189f760e70c85fa288d5125"
+
+  keg_only :provided_by_osx if MacOS.version < :mavericks
```
MikeMcQuaid: Just make this `keg_only "Older version of formula"` and then remove the `conflicts_with`s in both this and the `ant` formulae.

MikeMcQuaid (changes requested):

Formula/ant@1.9.rb
stringbean: Can I ask why this should be `keg_only`? 

Ant is a cli build tool and as far as I can tell other versioned cli tools (such as ansible or scala) are set to `conflicts_with`.

MikeMcQuaid: They are being updated en-masse in https://github.com/Homebrew/homebrew-core/pull/9001.

andreild: Thanks for creating this PR stringbean, we also need exactly this formula.

How's it going with the `keg_only` thing?

stringbean: @MikeMcQuaid should be good now.

Formula/ant@1.9.rb
```diff
@@ -0,0 +1,67 @@
+class AntAT19 < Formula
+  desc "Java build tool"
+  homepage "https://ant.apache.org/"
+  url "https://www.apache.org/dyn/closer.cgi?path=ant/binaries/apache-ant-1.9.8-bin.tar.bz2"
+  version "1.9"
+  sha256 "6edeaab09fc0bb1eeeb6b6be9c0e463813b073c40189f760e70c85fa288d5125"
+
+  keg_only "Older version of formula"
```
MikeMcQuaid: Can use `keg_only :versioned_formula` now

Formula/ant@1.9.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  option "with-ivy", "Install ivy dependency manager"
+  option "with-bcel", "Install Byte Code Engineering Library"
```
MikeMcQuaid: As this is a versioned formula: let's pick sensible defaults and remove these options. Which are you planning on using personally?

MikeMcQuaid (changes requested):

Formula/ant@1.9.rb
stringbean: Personally we don't use either of these. I would lean towards users managing these instead of Homebrew - Ant has 25 [optional dependencies](http://ant.apache.org/manual-1.9.x/install.html#librarydependencies) so where do you draw the line?

MikeMcQuaid: We use our discretion. We've tried to remove options on older versions as they are even less likely to be tested. Let's just remove these, thanks.

growtofill:
This pull request seems to be stuck, although the scope of the changes seems clear. @MikeMcQuaid should I create another one and get it merged?

MikeMcQuaid (approved):

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:

PR #8679, opened 2017-01-10
sqlparse 0.2.2 (new formula)

ilovezfs:

-----

Closes https://github.com/Homebrew/homebrew-python/issues/383.

CC @tdsmith

### DIFF
--- a/Formula/sqlparse.rb
+++ b/Formula/sqlparse.rb
@@ -1,0 +1,23 @@
+class Sqlparse < Formula
+  include Language::Python::Virtualenv
+
+  desc "Non-validating SQL parser"
+  homepage "https://github.com/andialbrecht/sqlparse"
+  url "https://files.pythonhosted.org/packages/55/ce/3944e990b03f80f36f0050b407ad46cde192a210d473f0d705b554bddd1d/sqlparse-0.2.2.tar.gz"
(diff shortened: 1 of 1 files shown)


tdsmith:
-1 on making this `import`able; if someone wants to use this in a Python project they should use pip.

tdsmith:
Specifically: I think this shouldn't be installed to `prefix`.

ilovezfs:
The docs make it look like importing it is the primary use case: https://sqlparse.readthedocs.io/en/latest/

tdsmith:
My preference is that we expose only the CLI component or decline to ship it at all; I think the former is reasonable if you want to do it.

ilovezfs:
That's fine. It's what I had originally before reading the documentation.

tdsmith:
For completeness, some of the reasons I feel that way are that:

* pip doesn't have a sense of ownership and will happily try and upgrade or downgrade any modules in sys.path while trying to satisfy the requirements of something it's been asked to install, which can lead to confusion
* pip and Homebrew both dump files in the HOMEBREW_PREFIX site-packages, which can lead to linking conflicts if pip has already installed something
* having pip manage importables avoids our having to worry about installing for python3.x

Now that the Python world has wheels I think homebrew-python has mostly outlived its utility.

ilovezfs:
@tdsmith PR refreshed

tdsmith (approved):

PR #8688, opened 2017-01-10
new formula for bit

amitgilad3:

-----


Formula/bit.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  depends_on "node"
+
+  def install
+    system "npm", "install", "-g", *Language::Node.std_npm_install_args(libexec)
+    bin.install_symlink Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+    system bin/"bit", "--version"
```
nijikon: This is not a proper test, sorry.

nijikon (changes requested):

Formula/bit.rb
amitgilad3: we currently do do not have a way to test the brew-formula for Bit, can we submit the formula without a test?

ilovezfs: elaborate please

nijikon (approved):

Formula/bit.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "node"
+
+  def install
+    system "npm", "install", "-g", *Language::Node.std_npm_install_args(libexec)
+    bin.install_symlink Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+    assert_equal "successfully initialized an empty bit scope.\n",
+      shell_output("#{bin}/bit init")
```
ilovezfs:
```
    assert_match "successfully initialized", shell_output("#{bin}/bit init")
    assert_match "dependencies", File.read("bit.json")
```
otherwise it'll fail because of the update message

ilovezfs:
And speaking of update messages ... if this has an autoupdate mechanism, which "There is a new version of Bit, would you like to update? [Y/n]" seems to imply, it'll need to be disabled.

nijikon:
If there is a new version we should probably add that :)

nijikon:
Is there a page with information about current release? Page from the formula has no information at all.

amitgilad3:
bit is new and so is the site .
the site of bit is under construction but  will hold all the necessary info in the near future

ilovezfs:
```
bit:
  * File should end with a newline
  * C: 19: col 4: Final newline missing.
```
Something ate the new line character at the end of the file.

```
ex -cwq Formula/bit.rb
```
will fix it

nijikon:

PR #8692, opened 2017-01-10
pldiff 1.0.4 (new formula)

scottrigby:

-----

CLI utility tool to help manage changes between plist files.
    
Includes:
- pldiff bash binary
- roff formatted manpage
- Asciidoc manpage source
- Install script

This is a very new tool, and is my first submitted Homebrew formula. Feedback welcome ⚙️

### DIFF
--- a/Formula/pldiff.rb
+++ b/Formula/pldiff.rb
@@ -1,0 +1,16 @@
+class Pldiff < Formula
+  desc "pldiff - Compares plist files line by line"
+  homepage "https://github.com/scottrigby/pldiff"
+  url "https://github.com/scottrigby/pldiff/archive/1.0.4.tar.gz"
+  sha256 "0093027aee4ceb049902a9fcea28923477c87c2f88813e2370fb4c12195b9ccd"
+
+  depends_on "colordiff" => :recommended
+
+  def install
+    system "/bin/sh", "install.sh", prefix
+  end
+
+  test do
+    system "false"
+  end
+end
--- a/Formula/pldiff.rb
+++ b/Formula/pldiff.rb
@@ -1,5 +1,5 @@
 class Pldiff < Formula
-  desc "pldiff - Compares plist files line by line"
+  desc "Compares plist files line by line"
   homepage "https://github.com/scottrigby/pldiff"
   url "https://github.com/scottrigby/pldiff/archive/1.0.4.tar.gz"
   sha256 "0093027aee4ceb049902a9fcea28923477c87c2f88813e2370fb4c12195b9ccd"


scottrigby:
Hi @nijikon - could the build have failed because of the warning I received `Description shouldn't include formula name`, and my commits were spaced too closely together? Looking at the [Jenkins console output](https://bot.brew.sh/job/Homebrew%20Core/14400/console) didn't give me anything to go on, but since it's building OK on my machine (and someone else I just asked to test this formula), perhaps that's it? Does your CI have an easy way to trigger a new build (without pushing some insignificant text change)?

woodruffw:
The build failed because of this audit check:

```
pldiff:
```

We generally like to see packages gain a bit of popularity and maturity before we include them in the official taps. We also generally like to see a non-owner/maintainer submit the package to us as a sign that the package has an active userbase.

You can always put this formula in a [personal tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md), but we can't merge it as-is. Sorry!

Formula/pldiff.rb
```diff
@@ -0,0 +1,16 @@
+class Pldiff < Formula
+  desc "Compares plist files line by line"
+  homepage "https://github.com/scottrigby/pldiff"
+  url "https://github.com/scottrigby/pldiff/archive/1.0.4.tar.gz"
+  sha256 "0093027aee4ceb049902a9fcea28923477c87c2f88813e2370fb4c12195b9ccd"
+
+  depends_on "colordiff" => :recommended
+
+  def install
+    system "/bin/sh", "install.sh", prefix
```
woodruffw: Just for future reference, you don't need the explicit `/bin/sh` here :smile:

Formula/pldiff.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  test do
+    system "false"
```
woodruffw: Formulae should also have a test block that invokes the program itself, usually in a way more complex than just a `--version` or `--help` test.

scottrigby:
@woodruffw That's totally fine (re popularity, maturity, and non-maintainer submission). I mainly want to learn how to do this properly, and this is a help 😄  Can we still communicate here about the comments you've left, so that I can get better at this? It would also help me get this in a better place if I want to add this as a custom tap.

Formula/pldiff.rb
scottrigby: @woodruffw ah ok, i was looking at other existing formulas for examples and this is how some others have done it. So just to be clear, are you saying `sh` is already discoverable so I can can just do this?:
```diff
+    system "sh", "install.sh", prefix
```

Also I saw the `prefix` var worked properly – I got this from some existing Formulas I looked through – but from the [cookbook]() it looked as if it's recommended instead to use the `#{prefix}` syntax. Is there a standard, or best practice for Homebrew Forumlas about this?

woodruffw:
@scottrigby Sure! I'm happy to provide further feedback.

Formula/pldiff.rb
woodruffw: Yeah, just `sh` works here!

The choice between `prefix` and `#{prefix}` really just depends on whether you're using it for interpolation or not. In your case, bare `prefix` is fine since it's a separate argument. If it were being inserted into an argument (e.g. `Pathname#/`: `prefix/"foo"`), we prefer the interpolation style (`"#{prefix}/foo"`).

scottrigby: @woodruffw Gotcha, this makes sense.

I created a custom tap repo, and moved this formula there. I also made this change you suggested there as a separate commit https://github.com/scottrigby/homebrew-macos-configurator/commit/31d43aa55771b1d16903b264d26098b633f59eaa and gave props 😄 Thanks!

Formula/pldiff.rb
scottrigby: @woodruffw Understood, I just didn't add this yet. I would think for this command, the test would make two temp files (using `EOF`) with one minor change, then call the `pldiff` command and then measure expected output from the CLI. Does that sound reasonable?

woodruffw: Absolutely, that's a good test :+1:

Formula/pldiff.rb
woodruffw: No problem, happy to help.

Formula/pldiff.rb
scottrigby: @woodruffw just following up… I added the test https://github.com/scottrigby/homebrew-macos-configurator/commit/ebda1ab2f82213d63b8cc30c24f82985806bec18, and it works perfectly:
```diff
diff --git a/Formula/pldiff.rb b/Formula/pldiff.rb
index ca9c348..80057a5 100644
--- a/Formula/pldiff.rb
+++ b/Formula/pldiff.rb
@@ -11,6 +11,10 @@ class Pldiff < Formula
   end
 
   test do
-    system "false"
+    system "defaults write #{testpath}/a foo -string bar"
+    system "defaults write #{testpath}/b foo -string baz"
+    o = shell_output("echo $(pldiff #{testpath}/a.plist #{testpath}/b.plist)")
+    assert_match "- <string>bar</string>", o
+    assert_match "+ <string>baz</string>", o
   end
 end
```
I think I'm all sorted for this formula. Thanks again!

PR #8697, opened 2017-01-10
mingw-w64-binutils 2.27 (new formula)

fxcoudert:
Cross-compiling binutils for mingw-w64 (Windows) target, 32 and 64-bit

-----

### DIFF
--- a/Formula/mingw-w64-binutils.rb
+++ b/Formula/mingw-w64-binutils.rb
@@ -32,7 +32,7 @@ class MingwW64Binutils < Formula
     system "#{bin}/x86_64-w64-mingw32-as", "-o", "test.o", "test.s"
     assert_match "file format pe-x86-64", shell_output("#{bin}/x86_64-w64-mingw32-objdump -a test.o")
     system "#{bin}/x86_64-w64-mingw32-ld", "-o", "test.exe", "test.o"
-    assert_match "PE32+ executable (console) x86-64", shell_output("file test.exe")
+    assert_match /PE32\+ executable .* x86-64/, shell_output("file test.exe")
 
     # Assemble a simple 32-bit routine
     (testpath/"test32.s").write <<-EOS.undent
@@ -46,6 +46,6 @@ class MingwW64Binutils < Formula
     system "#{bin}/x86_64-w64-mingw32-as", "--32", "-o", "test32.o", "test32.s"
     assert_match "file format pe-i386", shell_output("#{bin}/x86_64-w64-mingw32-objdump -a test32.o")
     system "#{bin}/x86_64-w64-mingw32-ld", "-m", "i386pe", "-o", "test32.exe", "test32.o"
-    assert_match "PE32 executable (console) Intel 80386", shell_output("file test32.exe")
+    assert_match /PE32 executable .* Intel 80386/, shell_output("file test32.exe")
   end
 end
--- a/Formula/mingw-w64-binutils.rb
+++ b/Formula/mingw-w64-binutils.rb
@@ -20,10 +20,6 @@ class MingwW64Binutils < Formula
   end
 
   test do
-    # Check target and version number
-    assert_match "x86_64-w64-mingw32", shell_output("#{bin}/x86_64-w64-mingw32-as --version")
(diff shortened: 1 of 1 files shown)


Formula/mingw-w64-binutils.rb
```diff
@@ -0,0 +1,26 @@
...
+                          "--target=x86_64-w64-mingw32",
+                          "--enable-targets=x86_64-w64-mingw32,i686-w64-mingw32",
+                          "--prefix=#{prefix}",
+                          "--with-sysroot=#{prefix}"
+    system "make"
+    system "make", "install"
+
+    # Info pages and localization files conflict with native tools
+    info.rmtree
+    (share/"locale").rmtree
```
woodruffw: This might be a problem for non-English/default locale users. Any way around this?

fxcoudert: In the spirit of "minor conflicts" (as described in the cookbook), this is what is also done by other developer tool packages in homebrew-core, like gdb (https://github.com/Homebrew/homebrew-core/blob/master/Formula/gdb.rb). I followed the practice.

Another option would be to leave the locale files in the keg, and not symlinking them into `/usr/local/share/locale`. I have not found how to specify that in the recipe.

lnxg33k:
👍 We need this man.

fxcoudert:
Can I get an approval? Or a response to the earlier comment? (and if deemed necessary a link to doc explaining how to keep the locale files in the keg and prevent homebrew from symlinking them into `/usr/local/share/locale`)

Formula/mingw-w64-binutils.rb
woodruffw: Thanks for the clarification. This seems like the right practice here.

woodruffw:
@fxcoudert Sorry, I've been a bit busy. I'll merge this today.

Formula/mingw-w64-binutils.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  test do
+    assert_match "x86_64-w64-mingw32", shell_output("#{bin}/x86_64-w64-mingw32-as --version")
```
woodruffw: Can we build and link a simple assembly file here, instead of just testing `--version`?

fxcoudert:
@woodruffw tests added. Cannot add proper tests for `ld` because we don't have any runtime libraries to link at that stage.

(Once this is merged, I will submit a PR for mingw-w64 headers, runtime and compiler.)

woodruffw:
Shouldn't it be possible to run `ld` against two object files? We don't need to actually link against a library, just execute `ld` in a way that provides a reasonably good guarantee of functionality. Skipping `i386` is fine too, I think a combined `as`/`ld` test for `x86_64` provides a good guarantee.

Thanks for adding the `as` test. You can go ahead and remove the old `--version` tests.

fxcoudert:
I added some bare bones linker tests. They link without CRT so they won't produce actual working executables, but at least we can check the format produced.

Given that binutils are multitarget (and it has been known to break in the past), I would rather have the tests enabled for both 32 and 64-bit targets, so that we know the multitarget configuration was correct.

woodruffw:
> Given that binutils are multitarget (and it has been known to break in the past), I would rather have the tests enabled for both 32 and 64-bit targets, so that we know the multitarget configuration was correct.

That makes sense. Thanks for the explanation!

woodruffw:
Thanks for contributing @fxcoudert! This is a great addition to Homebrew :tada:

MikeMcQuaid:
Let's hold off merging any more mingw stuff until the compiler is ready. It's unclear to me how this is useful without the compiler.

fxcoudert:
This one was a dependency of the compiler. Next the CRT and compiler should go in together (their build systems are intertwined): https://github.com/Homebrew/homebrew-core/pull/9037

MikeMcQuaid:
Personally I think they should all be in a single formula as they aren't particularly useful without each other.

fxcoudert:
With separate binutils, we have the potential for cross-compilers other than GCC (or other versions of GCC)

MikeMcQuaid:
https://en.wikipedia.org/wiki/You_aren't_gonna_need_it 😉

PR #8699, opened 2017-01-10
rust: fix issue #8656; allow CARGO_PATH

cema-sp:
Fixes #8656

This patch allows to specify `CARGO_PATH` with previous **cargo** version to be added to `PATH` on build.

Without this fix I couldn't brew *HEAD* version of  **rust** package.
Now I use following command to brew with previous **cargo**:

~~~sh
CARGO_PATH="/usr/local/Cellar/rust/1.13.0/bin" brew install --HEAD --with-cargo rust
~~~

- [x] Guidelines?
- [x] Open pull requests?
- [x] Local build with `brew install --build-from-source rust`?
- [ ] Pass `brew audit --strict rust` (after doing `brew install rust`)?

Brew audit doesn't pass with following error:

~~~
rust:
  * The installation was broken.
    Broken dylib links found:
      /usr/local/Cellar/rust/HEAD-78c892d/lib//private/tmp/rust-20170110-47474-1gbw9ld/build/x86_64-apple-darwin/stage1-rustc/x86_64-apple-darwin/release/deps/librustc_driver-3b169b4b1f044fbc.dylib
     ...
~~~

Any suggestions on this error?

PR #8708, opened 2017-01-10
rust 1.14.0

### LINKED ISSUE #8656: Rust failed to build on Mac OS 10.12.2
Trying to install **rust** `HEAD` with following command: `brew install --HEAD rust`, I have an error:

~~~
~~~

Link to gist-logs: https://gist.github.com/cema-sp/4085820d0073d76cee591bdeeb2c6057

As I understood from [this cargo issue](https://github.com/rust-lang/cargo/issues/3437), one should have previous version of binary OR compile it from source. But compilation doesn't work on my laptop.

I have tried it a number of times and resolved the issue providing _./configure_ script `cargo` binary from stable **rust** bottle:

~~~sh
./configure --prefix=/usr/local/Cellar/rust/HEAD-7ef1a69 --local-rust-root=/usr/local/Cellar/rust/HEAD-7ef1a69 --enable-optimize --cargo=/usr/local/Cellar/rust/1.13.0/bin/cargo
make && make install
~~~

Any suggestions?
____

- [x] Confirmed this is problem with specific formulae and not Homebrew/brew?

ilovezfs:

-----

Supersedes #5490.
Supersedes #8138.
Supersedes #8699.
Fixes #8656.

CC @nomoon @alex @cema-sp

### DIFF
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -3,23 +3,18 @@ class Rust < Formula
   homepage "https://www.rust-lang.org/"
 
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.13.0-src.tar.gz"
-    sha256 "ecb84775ca977a5efec14d0cad19621a155bfcbbf46e8050d18721bb1e3e5084"
+    url "https://static.rust-lang.org/dist/rustc-1.14.0-src.tar.gz"
+    sha256 "c790edd2e915bd01bea46122af2942108479a2fda9a6f76d1094add520ac3b6b"
 
     resource "cargo" do
-      # git required because of submodules
-      url "https://github.com/rust-lang/cargo.git", :tag => "0.14.0", :revision => "eca9e159b6b0d484788ac757cf23052eba75af55"
+      url "https://github.com/rust-lang/cargo.git",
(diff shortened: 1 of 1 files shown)


nijikon:
❤️

MikeMcQuaid (approved):

aktau:
This broke installing rust on Linuxbrew (not that this should be tested in the main homebrew repo I think, just leaving a note).

ilovezfs:
interesting. can you please file an issue with Linuxbrew so we can determine the problem and where/who needs to fix it?

aktau:
I already had before I posted here: https://github.com/Linuxbrew/homebrew-core/issues/1588.

ilovezfs:
@aktau yes I think you're on to something. Linuxbrew, downstream, will need to alter that block to use the prebuilt Linux one.

aktau:
> @aktau yes I think you're on to something. Linuxbrew, downstream, will need to alter that block to use the prebuilt Linux one.

Is it still possible to self-build one (or force that with some --without-prebuilt-cargo flag)? Because that would also be a solution. I tried manually generating the path for the prebuilt linux one but failed. I wonder where those are found.

PR #8729, opened 2017-01-11
source-to-image 1.1.4 (new formula)

neckhair:
This adds the formula for Openshift's source-to-image tool.

-----

### DIFF
--- a/Formula/source-to-image.rb
+++ b/Formula/source-to-image.rb
@@ -1,0 +1,14 @@
+class SourceToImage < Formula
+  desc "Tool for building source and injecting into docker images"
+  homepage "https://github.com/openshift/source-to-image"
+  url "https://github.com/openshift/source-to-image/releases/download/v1.1.3/source-to-image-v1.1.3-ddb10f1-darwin-amd64.tar.gz"
+  sha256 "b5d179e1da3e5e3d4b206a767c69bc6d5860d82227a475ecc17c4279bfb0f862"
+
+  def install
+    bin.install "s2i"
+  end
+
+  test do
+    system "#{bin}/s2i", "-h"
+  end
+end
--- a/Formula/source-to-image.rb
+++ b/Formula/source-to-image.rb
@@ -9,6 +9,7 @@ class SourceToImage < Formula
   end
 
   test do
-    system "#{bin}/s2i", "-h"
+    system "#{bin}/s2i", "create", "testimage", testpath
+    assert_match(/Dockerfile/, shell_output("ls").chomp)
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/source-to-image.rb
```diff
@@ -0,0 +1,14 @@
+class SourceToImage < Formula
+  desc "Tool for building source and injecting into docker images"
+  homepage "https://github.com/openshift/source-to-image"
+  url "https://github.com/openshift/source-to-image/releases/download/v1.1.3/source-to-image-v1.1.3-ddb10f1-darwin-amd64.tar.gz"
```
bfontaine: We need to build from source here.

Formula/source-to-image.rb
```diff
@@ -0,0 +1,14 @@
...
+  sha256 "b5d179e1da3e5e3d4b206a767c69bc6d5860d82227a475ecc17c4279bfb0f862"
+
+  def install
+    bin.install "s2i"
+  end
+
+  test do
+    system "#{bin}/s2i", "-h"
```
bfontaine: Can’t we write a test that does something more substantial than e.g. `-v` or `-h`?

bfontaine (changes requested):

Formula/source-to-image.rb
neckhair: As source-to-image is in Go the user will need that on his machine to build it. Could you explain, please, why we can't just download the already built binary?

Formula/source-to-image.rb
```diff
@@ -9,6 +9,7 @@ def install
   end
 
   test do
-    system "#{bin}/s2i", "-h"
+    system "#{bin}/s2i", "create", "testimage", testpath
+    assert_match(/Dockerfile/, shell_output("ls").chomp)
```
bfontaine: You may want to use `assert (testpath/"Dockerfile").exist?` instead of parsing `ls`’s output.

Formula/source-to-image.rb
bfontaine: We want to have full control over what we build; we provide pre-built binaries ("bottles") so users who don’t want to build from source won’t have to install Go.

See https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md#we-dont-like-binary-formulae

neckhair:
@bfontaine Read some more documentation for building formulas. I'm building it now from source and implemented your suggestions for the test.

bfontaine (approved):

Formula/source-to-image.rb
```diff
@@ -0,0 +1,20 @@
...
+  depends_on "go" => :build
+
+  def install
+    system "hack/build-go.sh"
+    bin.install "_output/local/bin/darwin/amd64/s2i"
+  end
+
+  test do
+    system "#{bin}/s2i", "create", "testimage", testpath
+    assert (testpath/"Dockerfile").exist?
```
MikeMcQuaid: Can you add a second argument here `, "Dockerfile does not exist!"` as the default message is pretty terrible. Thanks!

neckhair:
@MikeMcQuaid Done.

dunn:
Thank you! Merged as 091598078999e215e0999a933170c021e6e53958.

Issue #1828, opened 2016-06-08
Upgrade: ImageMagick 7.0.x-y


rakkesh:
Formulae dependent on imagemagick (in official Homebrew taps)

❌ - Build failed
⚠️ - Formula needs a revision update
✅ - Formula works as is, without a revision update

---

   Formula     => Dependency ( Formula revision )

---
- [x] asciinema2gif            => required (⚠️)
- [x] autotrace                => recommended (⚠️)
- [x] [blockhash](https://github.com/commonsmachinery/blockhash)                => required (❌ [build failed](https://gist.github.com/02477d5bcbbbf2b4697991d18cd48add), failure related to change in path of `wand/MagickWand.h`)
- [x] [dmtx-utils](https://sourceforge.net/p/libdmtx/bugs/)               => required (❌ [build failed](https://gist.github.com/eeccc370b67b5f10c7bd3585cd434067))
- [x] gifify                   => required (✅, shell-script)
- [x] negfix8                  => required (✅, shell-script)
- [x] neofetch                 => required (✅, shell-script)
- [x] nuxeo                    => required (✅, shell-script)
- [x] pdf-redact-tools         => required (✅)
- [x] plowshare                => required (✅, shell-script)
- [x] pstoedit                 => required (⚠️)
- [x] spaceman-diff            => required (✅, shell-script)
- [x] synfig                   => required (❌ [build failed](https://gist.github.com/b69b505e494bd876d6e0ec3fedfaabf1))
- [x] vcs                      => required (✅, shell-script)
- [x]  xplanetfx                => required (✅)
- [x] [zbar](https://sourceforge.net/p/zbar/support-requests/)                     => required (❌[build failed](https://gist.github.com/f51997dc48699f868b17bd77404c873a), failure related to change in path of `wand/MagickWand.h`)
- [x] homebrew/games/mgba      => recommended (❌[build failed](https://gist.github.com/337dfe1700c929c4ba77679334d670a8), failure related to change in path of `wand/MagickWand.h`)
- [x] homebrew/gui/geeqie      => recommended (✅)
- [ ] homebrew/science/ogdraw  => required (too much of a hassle to install Perl dependencies)
- [x] homebrew/science/therion => required (⚠️)
- [x] homebrew/science/weblogo => required (✅, perl-script)
[...]

PR #8740, opened 2017-01-11
dmtx-utils: migrate to boneyard

ilovezfs:
Created with `brew boneyard-formula-pr` because it's marked as abandoned on SourceForge and because upstream has not responded to the report of Imagemagick 7 problems for over 6 months.


ilovezfs:
See #1828 and https://sourceforge.net/p/libdmtx/bugs/58/.

ilovezfs:
also apparently no installs according to analytics

PR #8747, opened 2017-01-11
luaver 1.0.0-p1

jdonaldson:
This commit cleans up the luaver installation process.

luaver is not a proper executable, rather it is a set of shell functions
that are typically sourced from a ~/.bashrc

The first change is to not install the sourced file to /usr/local/bin,
but rather to /usr/local/share, and then provide instructions for
sourcing in the caveats.

The next change is to use a git-style installation, rather than a
release download.  This will make it easier to manage versions via tags,
and osx-specific patches via revisions.

I've bumped the version here with a patch (-p1), which will include the
latest osx-specific fix, which was added directly after the 1.0 version.

-----

### DIFF
--- a/Formula/luaver.rb
+++ b/Formula/luaver.rb
@@ -1,42 +1,37 @@
 class Luaver < Formula
   desc "Manage and switch between versions of Lua, LuaJIT, and Luarocks"
   homepage "https://github.com/DhavalKapil/luaver"
+
+  url "https://github.com/DhavalKapil/luaver.git",
+      :revision => "61861f2a966645a663971cd4e26d2a6104d075f6"
+  version "1.0.0-p1"
   head "https://github.com/DhavalKapil/luaver.git"
 
-  stable do
-    url "https://github.com/DhavalKapil/luaver/archive/v1.0.0.tar.gz"
-    sha256 "5c7eb65ea9b3cb4f10987c1d564e6e86bd6ad4f9d829a799121c83f0d85bd390"
-
-    patch do
-      url "https://github.com/DhavalKapil/luaver/pull/9.patch"
-      sha256 "390929ef53b6e8e372451049eeffbfa51b9e31d7071e386478b871a835aba2db"
-    end
-  end
-
   bottle :unneeded
-
   depends_on "wget" => :run
 
   def install
-    bin.install "luaver"
+    share.install "luaver"
   end
 
   def caveats; <<-EOS.undent
-    Add the following at the end of the correct file yourself:
-      if which luaver > /dev/null; then . `which luaver`; fi
+    You should add the following to your .bashrc (or equivalent):
+      if [ -f #{HOMEBREW_PREFIX}/share/luaver ]; then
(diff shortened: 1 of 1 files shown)


jdonaldson:
CC @umireon

jdonaldson:
second push fixed some audit nits.

nijikon:
Why do we install to `share`?

jdonaldson:
I had seen other brew packages install to `share` if they were intended to be sourced rather than executed directly. [git-bash-prompt](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bash-git-prompt.rb) comes to mind.

I'm open to putting this somewhere else, I just figured /usr/local/bin was not appropriate.  Let me know if there's a better place.

dunn:
`pkgshare` would be better, thanks!

Formula/luaver.rb
```diff
@@ -1,42 +1,37 @@
...
-    url "https://github.com/DhavalKapil/luaver/archive/v1.0.0.tar.gz"
-    sha256 "5c7eb65ea9b3cb4f10987c1d564e6e86bd6ad4f9d829a799121c83f0d85bd390"
 
-    patch do
-      url "https://github.com/DhavalKapil/luaver/pull/9.patch"
-      sha256 "390929ef53b6e8e372451049eeffbfa51b9e31d7071e386478b871a835aba2db"
-    end
-  end
+  url "https://github.com/DhavalKapil/luaver.git",
+      :revision => "61861f2a966645a663971cd4e26d2a6104d075f6"
```
MikeMcQuaid: We need to use a tag.

Formula/luaver.rb
```diff
@@ -1,42 +1,37 @@
...
+  version "1.0.0-p1"
+  head "https://github.com/DhavalKapil/luaver.git"
 
   bottle :unneeded
-
   depends_on "wget" => :run
 
   def install
-    bin.install "luaver"
+    share.install "luaver"
```
MikeMcQuaid: `pkgshare`, as mentioned.

Formula/luaver.rb
```diff
@@ -1,42 +1,37 @@
...
-      sha256 "390929ef53b6e8e372451049eeffbfa51b9e31d7071e386478b871a835aba2db"
-    end
-  end
+  url "https://github.com/DhavalKapil/luaver.git",
+      :revision => "61861f2a966645a663971cd4e26d2a6104d075f6"
+  version "1.0.0-p1"
+  head "https://github.com/DhavalKapil/luaver.git"
 
   bottle :unneeded
-
```
MikeMcQuaid: Leave this space.

MikeMcQuaid (changes requested):

umireon:
We should wait for a new release that will happen soon, or simply remove the `git-style installation'.
I think the other changes are reasonable.

MikeMcQuaid:
> We should wait for a new release that will happen soon

Yep 👍

jdonaldson:
In that case, I'll close, and interested parties can use this PR for future reference.

PR #8749, opened 2017-01-11
libui 3.1 (new formula)

ricardobeat:
Adds [libui](https://github.com/andlabs/libui), a cross-platform GUI library for C.

-----

### DIFF
--- a/Formula/libui.rb
+++ b/Formula/libui.rb
@@ -1,0 +1,50 @@
+class Libui < Formula
+  desc "portable GUI library for C"
+  homepage "https://github.com/andlabs/libui"
+  url "https://github.com/andlabs/libui/archive/alpha3.1.tar.gz"
+  sha256 "945ee9abacffafa93b5cf941bddfbe0bd0d4b3a8c7f8b0256779e24249f57679"
+
+  head "https://github.com/andlabs/libui/libui.git"
+
+  option "without-shared-library", "Build static library only (defaults to building dylib only)"
+  option "with-examples", "Build examples"
+  option "with-tester", "Build tester program"
+
+  depends_on "cmake" => :build
+
+  def install
+    args = []
+    args << "-DBUILD_SHARED_LIBS=OFF" if build.without? "shared-library"
+    args << ".."
+
+    mkdir "build" do
+      system "cmake", *args
+      system "make"
+      system "make", "examples" if build.with? "examples"
+      system "make", "tester" if build.with? "tester"
+
+      Dir.chdir("out")
+      libexec.install "controlgallery", "cpp-multithread", "histogram" if build.with? "examples"
+      libexec.install "test" if build.with? "test"
+    end
+
+    include.install "ui.h", "ui_darwin.h"
+    lib.install Dir["build/out/libui.*"]
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <ui.h>
+      uiWindow *mainwin;
+      int main(void) {
+        mainwin = uiNewWindow("libui test", 640, 480, 1);
+        uiQuit();
+        return 0;
+      }
+    EOS
+    system ENV.cc, "test.c", "-o", "test",
+           "-I#{include}", "-L#{lib}", "-lui",
+           "-framework", "Cocoa"
+    system "./test"
+  end
+end


woodruffw:
We don't normally include alpha-only packages in `homebrew-core`, so we may pass on this until it's stable (other maintainers can chime in and overrule me here).

Formula/libui.rb
```diff
@@ -0,0 +1,49 @@
...
+      system "make"
+      system "make", "examples" if build.with? "examples"
+      system "make", "tester" if build.with? "tester"
+
+      Dir.chdir("out")
+      libexec.install "controlgallery", "cpp-multithread", "histogram" if build.with? "examples"
+      libexec.install "test" if build.with? "test"
+    end
+
+    lib.install "common", "darwin", "ui.h", "ui_darwin.h", Dir["build/out/libui.*"]
```
woodruffw: You probably don't want to install header files in `lib` (they go in `include` instead). That's what's causing this CI error:

```
libui:
  * Non-libraries were installed to "/usr/local/opt/libui/lib"
    Installing non-libraries to "lib" is discouraged.
    The offending files are:
      /usr/local/opt/libui/lib/ui.h
      /usr/local/opt/libui/lib/ui_darwin.h
```

woodruffw: The `Dir["build/out/libui.*"]` may also not do anything, since you call `Dir.chdir("out")` before it (and you're already in `out`).

ricardobeat: @woodruffw thanks for reviewing! I'll move those over to `include`. The last `.install` is outside the `mkdir` block, `Dir["build/out/libui.*"]` links the `.dylib / .a` files.

ricardobeat:
I was also not sure about installing sample apps to `libexec`, is that common pratice? I just mimicked what I say in the `glfw` Formula.

woodruffw:
> I was also not sure about installing sample apps to libexec, is that common pratice?

Yeah, I think that should be fine.

Formula/libui.rb
```diff
@@ -0,0 +1,50 @@
...
+    args = []
+    args << "-DBUILD_SHARED_LIBS=OFF" if build.without? "shared-library"
+    args << ".."
+
+    mkdir "build" do
+      system "cmake", *args
+      system "make"
+      system "make", "examples" if build.with? "examples"
+      system "make", "tester" if build.with? "tester"
+
```
woodruffw: Do you know if this project's `cmake` generates a `make install` target we can use instead of manually calling `libexec`/`lib`/`include`? The most common `cmake` flow looks something like this, usually:

```ruby
args = std_cmake_args
args << "-DBUILD_SHARED_LIBS=OFF" if build.without? "shared-library"

mkdir "build" do
  system "cmake", "..", *args
  system "make"
  # additional targets
  system "make", "install"
end
```

Formula/libui.rb
```diff
@@ -0,0 +1,50 @@
...
+      Dir.chdir("out")
```
bfontaine: Use `cd "out" do` instead of `Dir.chdir`.

Formula/libui.rb
```diff
@@ -0,0 +1,50 @@
+class Libui < Formula
+  desc "portable GUI library for C"
+  homepage "https://github.com/andlabs/libui"
+  url "https://github.com/andlabs/libui/archive/alpha3.1.tar.gz"
+  sha256 "945ee9abacffafa93b5cf941bddfbe0bd0d4b3a8c7f8b0256779e24249f57679"
+
+  head "https://github.com/andlabs/libui/libui.git"
+
+  option "without-shared-library", "Build static library only (defaults to building dylib only)"
+  option "with-examples", "Build examples"
```
bfontaine: Is there a reason not to build examples?

Formula/libui.rb
```diff
@@ -0,0 +1,50 @@
...
+  option "with-tester", "Build tester program"
```
bfontaine: Is there a reason not to build with this program?

Formula/libui.rb
```diff
@@ -0,0 +1,50 @@
+class Libui < Formula
+  desc "portable GUI library for C"
+  homepage "https://github.com/andlabs/libui"
+  url "https://github.com/andlabs/libui/archive/alpha3.1.tar.gz"
```
bfontaine: We need a stable release, not an alpha or beta release.

bfontaine (changes requested):

Formula/libui.rb
ricardobeat: It doesn't, it just drops the built libraries in the `out` folder :(

Formula/libui.rb
ricardobeat: The test or examples are not required for the functioning of the library, I didn't want to add [possible] unnecessary build dependencies. I honestly don't even know how users would reach these executables in `libexec` normally.

Formula/libui.rb
ricardobeat: Despite being v3, ~2.8k commits, 4k stars, this library doesn't have a version tagged `stable` yet. Is there an alternate tap this could be submitted to?

woodruffw: There *was* `homebrew-head-only`, but we not longer provide an official tap for non-stable formulae. Until this gets a stable release, the best action is probably to keep it in a personal tap.

Formula/libui.rb
bfontaine: We prefer to limit the number of build options; since they don’t have dependencies let’s build those and remove the options.

MikeMcQuaid:
Passing on this until they release a stable version. Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

PR #8751, opened 2017-01-11
iwyu 0.7 (new formula)

jasonmp85:

-----

So there's a bit of a history here. I initially opened Homebrew/legacy-homebrew#32610 over two years ago because I found IWYU fairly obnoxious to build and install correctly and wanted a simple Homebrew formula for coworkers, etc.

That request was closed for two primary reasons:

  * Building `include-what-you-use` requires LLVM development headers, not immediately available on OS X, creating a dependency on LLVM (making builds _very slow_)
  * Though `include-what-you-use` provides binaries which might sidestep the above issue, they need access to `clang`'s built-in headers and expect them at a particular relative path, so I had to write somewhat brittle code to ensure this structure. Upgrading Xcode would have broken my Formula

To get around the above, I created my [own tap](https://github.com/jasonmp85/homebrew-iwyu) and added a `caveat` for the Xcode upgrade case. In addition, I found that I [hadn't covered the C++ case very well](https://github.com/jasonmp85/homebrew-iwyu/pull/5), so I fixed that, too.

Eventually, upstream [did two things](https://github.com/include-what-you-use/include-what-you-use/issues/247):

  * They added code to auto-search Xcode's header paths, if available (fixing the C++ issue)
  * Binaries now come bundled with compiler built-in headers, which can be installed alongside the binary directly

I've tweaked my tap's formula to use these new developments, meaning I no longer need to symlink to version-specific Xcode directories (!). The C++ headers will be found automatically, and the built-ins are included with the tarball.

Unless I'm mistaken, I think that means this is eligible for inclusion in `homebrew-core`?

### DIFF
--- a/Formula/iwyu.rb
+++ b/Formula/iwyu.rb
@@ -1,0 +1,59 @@
+require "English"
+
+class Iwyu < Formula
+  desc "analyze includes in C and C++ source files"
+  homepage "https://include-what-you-use.org"
+  url "https://include-what-you-use.org/downloads/include-what-you-use-0.7-x86_64-apple-darwin.tar.gz"
+  version "0.7"
+  sha256 "ba343d452b5d7b999c85387dff1750ca2d01af0c4e5a88a2144b395fa22d1575"
+
+  def install
+    # include-what-you-use looks for a lib directory one level up from its bin-
+    # dir, but putting it directly in prefix results in it getting linked in
+    # /usr/local/lib
+    #
+    # To solve this, we put iwyu in a custom dir and make bin links.
+    iwyu_subdir_path = (prefix / "iwyu")
+    iwyu_subdir_path.mkpath
+
+    # just copy everything from the tarball's lib directory
+    iwyu_subdir_path.install("lib")
+
+    # selectively include certain binaries; give them better names
+    iwyu_bindir = (iwyu_subdir_path / "bin")
+    iwyu_bindir.mkpath
+
+    iwyu_bindir.install("bin/fix_includes.py" => "fix_include")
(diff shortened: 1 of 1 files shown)


woodruffw:
We generally avoid binary formulae in homebrew core, so I think it would be good to get a working build going (burdening the CI with a large build shouldn't be a problem, and the majority of users will be installing from bottles).

jasonmp85:
Uuuuggghhh… OK.

FWIW, I feel as though the cleanest approach will be to add the entire LLVM and Clang source trees as resources to the `include-what-you-use` recipe, build everything, then build `include-what-you-use` in-tree before copying out resources needed. That… could take a while. Just declaring `llvm` as a dependency isn't enough, IIRC.

In addition, I'm a scrub a CMake, so this could take a while. But at least now it's theoretically possible.

jasonmp85:
(Ummmm… does Cask accept binary packages, then, maybe? As an interim to get rid of my tap?)

woodruffw:
Yeah, Cask does binary packages only. I'm not sure what their policy on non-`.app` packages is, though :confused:

MikeMcQuaid:
Cask won't accept this is it can be built from source but it's just too hard. It's worth digging in why our installed LLVM isn't sufficient. Are we not installing everything or is it using private headers?

jasonmp85:
> Are we not installing everything or is it using private headers?

I'm not sure if I actually know the up-to-date answer to that question. It's possible enough is installed now, but I'll have to check. I'll close this issue and re-open with a source build if I can get it figured out/when I have the time.

PR #8756, opened 2017-01-12
Add back the old imagemagick under the name imagemagick@6

crazymykl:
This is necessary since `rmagick` does not work with ImageMagick 7.0.x.

Apologies if I'm not following protocol 100%, it was not super clear how new versions of existing formulae should be formatted.

I bought back the formula for the last 6.9.x ImageMagick nearly verbatim, only changing the class name.

### DIFF
--- a/Formula/imagemagick@6.rb
+++ b/Formula/imagemagick@6.rb
@@ -1,0 +1,133 @@
+class ImagemagickAT6 < Formula
+  desc "Tools and libraries to manipulate images in many formats"
+  homepage "https://www.imagemagick.org/"
+  # Please always keep the Homebrew mirror as the primary URL as the
+  # ImageMagick site removes tarballs regularly which means we get issues
+  # unnecessarily and older versions of the formula are broken.
+  url "https://dl.bintray.com/homebrew/mirror/imagemagick-6.9.7-4.tar.xz"
+  mirror "https://www.imagemagick.org/download/ImageMagick-6.9.7-4.tar.xz"
+  sha256 "68842c55ed9c958b84aae17974961cefff4212bf7146f09fd15c09dbdc2d9629"
+
+  keg_only "Older version of imagemagick"
+
+  option "with-fftw", "Compile with FFTW support"
+  option "with-hdri", "Compile with HDRI support"
+  option "with-opencl", "Compile with OpenCL support"
+  option "with-openmp", "Compile with OpenMP support"
+  option "with-perl", "Compile with PerlMagick"
+  option "with-quantum-depth-8", "Compile with a quantum depth of 8 bit"
+  option "with-quantum-depth-16", "Compile with a quantum depth of 16 bit"
+  option "with-quantum-depth-32", "Compile with a quantum depth of 32 bit"
+  option "without-magick-plus-plus", "disable build/install of Magick++"
+  option "without-modules", "Disable support for dynamically loadable modules"
+  option "without-threads", "Disable threads support"
+  option "with-zero-configuration", "Disables depending on XML configuration files"
+
+  deprecated_option "enable-hdri" => "with-hdri"
+  deprecated_option "with-jp2" => "with-openjpeg"
+
+  depends_on "pkg-config" => :build
+  depends_on "libtool" => :run
+  depends_on "xz"
+
+  depends_on "jpeg" => :recommended
+  depends_on "libpng" => :recommended
+  depends_on "libtiff" => :recommended
+  depends_on "freetype" => :recommended
+
+  depends_on "fontconfig" => :optional
+  depends_on "little-cms" => :optional
+  depends_on "little-cms2" => :optional
+  depends_on "libwmf" => :optional
+  depends_on "librsvg" => :optional
+  depends_on "liblqr" => :optional
+  depends_on "openexr" => :optional
+  depends_on "ghostscript" => :optional
+  depends_on "webp" => :optional
+  depends_on "openjpeg" => :optional
+  depends_on "fftw" => :optional
+  depends_on "pango" => :optional
+  depends_on :perl => ["5.5", :optional]
+
+  needs :openmp if build.with? "openmp"
+
+  skip_clean :la
+
+  def install
+    args = %W[
+      --disable-osx-universal-binary
+      --prefix=#{prefix}
+      --disable-dependency-tracking
+      --disable-silent-rules
+      --enable-shared
+      --enable-static
+    ]
+
+    if build.without? "modules"
+      args << "--without-modules"
+    else
+      args << "--with-modules"
+    end
+
+    if build.with? "opencl"
+      args << "--enable-opencl"
+    else
+      args << "--disable-opencl"
+    end
+
+    if build.with? "openmp"
+      args << "--enable-openmp"
+    else
+      args << "--disable-openmp"
+    end
+
+    if build.with? "webp"
+      args << "--with-webp=yes"
+    else
+      args << "--without-webp"
+    end
+
+    if build.with? "openjpeg"
+      args << "--with-openjp2"
+    else
+      args << "--without-openjp2"
+    end
+
+    args << "--without-gslib" if build.without? "ghostscript"
+    args << "--with-perl" << "--with-perl-options='PREFIX=#{prefix}'" if build.with? "perl"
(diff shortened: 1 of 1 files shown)


ilovezfs:
I do not think sufficient need for this formula has been demonstrated, especially given that there are no dependents in Homebrew itself that need it. The formula will be broken in terms of the ability of build systems to detect it via pkg-config unless it's made keg only or is marked as conflicting with the main imagemagick formula, which will then interfere with the installation of formulae that depend on imagemagick. My recommendation is that you host this in your own tap for the time being.

ledermann:
This gist allows installing 6.9.7-3:
https://gist.github.com/ledermann/0825a492abd2d465fc589be2f4e3bff1

ilovezfs:
@ledermann you should host that in your own tap. This is not an appropriate venue for posting personal formulae files.

zmwangx:
> My recommendation is that you host this in your own tap for the time being.

The problem with that, in the case of IM, is that it is updated frequently, and whenever an update is issued the old tarball is removed from the server. That's why we brew mirror the tarballs.

But indeed, a formula like this wouldn't work very well. The best compromise is to mark it as keg only, and deal with the hassle of having to specify the opt paths (`PATH`, `LDFLAGS`, `CPPFLAGS`, `PKG_CONFIG_PATH`) where necessary (which isn't always possible in obscure build systems), or force link it if you don't need any revdeps.

At the very least it comes down to

> I do not think sufficient need for this formula has been demonstrated.

ilovezfs:
>The problem with that, in the case of IM, is that it is updated frequently, and whenever an update is issued the old tarball is removed from the server. 

Perhaps "someone" should convince upstream to knock that behavior off.

zmwangx:
> Perhaps "someone" should convince upstream to knock that behavior off.

That "someone" might need to foot the bill (which might be negligible; I've no idea).

ilovezfs:
Is there something deficient about the _orig tarballs on all Debian mirrors? https://mirrors.ocf.berkeley.edu/debian/pool/main/i/imagemagick/

zmwangx:
Not sure whether dfsg stripping causes problems, I assume not. Two issues:

1. Only a handful of versions are available, most are skipped;
1. Debian removes old tarballs on a regular basis, too.

ilovezfs:
>Only a handful of versions are available, most are skipped;

Is there some benefit the plethora of updates is providing?

>Debian removes old tarballs on a regular basis, too.

I guess this is one of the reasons `brew update` and the Internet exist.

zmwangx:
> Is there some benefit the plethora of updates is providing?

zmwangx:
> Is there some benefit the plethora of updates is providing?

Probably not until you have a CVE.

> I guess this is one of the reasons brew update and the Internet exist.

Well, we're talking about user-hosted taps here... Not really our problem, but still.

ilovezfs:
User-hosted taps get updated by brew update.

zmwangx:
It depends on said user updating the formula diligently :)

ilovezfs:
>It depends on said user updating the formula diligently 

Indeed.

crazymykl:
Again, the `rmagick` gem, by far the most common reason I have seen ImageMagick installed on macOS, does not build with major version 7. It is unlikely that a version that builds against both 6 and 7 will be forthcoming.,

`rmagick` has [almost 8 million downloads](https://rubygems.org/gems/rmagick/versions/2.15.4).

ilovezfs:
perhaps we've just given them a reason to add support for 7

crazymykl:
There's also the question of deployment. A lot of apps are ultimately hosted on OSes which won't have a package for ImageMagick@7 for year, or possibly ever (e. g. CentOS).

Flipping the question: what exists that only builds against ImageMagick@7? Maybe 7 should be made keg-only until it is better supported in the wild than version 6.

ilovezfs:
It handles all of the Homebrew dependents. There's no reason to make it keg_only.

zmwangx:
> by far the most common reason I have seen ImageMagick installed on macOS

Experience varies, but I'm pretty sure IM on macOS is more often used as standalone tools. Most of those 8 million downloads could well be on Linux web servers, and I'm pretty sure PHP bindings for IM are more prevalent than the Ruby binding on servers.

> what exists that only builds against ImageMagick@7?

Homebrew ships cutting edge software. That's the nature of the project.

crazymykl:
I have made a tap, for whoever finds this after googling rmagick homebrew imagemagick 7.

`brew install crazymykl/homebrew-custom/imagemagick@6`

jacquerie:
Thanks, @crazymykl! I'm using your tap because [Wand 0.4.4](http://docs.wand-py.org/en/0.4.4/) appears to support only ImageMagick@6. With ImageMagick@7 I was getting:
```
ImportError: MagickWand shared library not found.
You probably had not installed ImageMagick library.
Try to install:
  brew install freetype imagemagick
```
(typing this for the benefit of people searching for the error above).

MikeMcQuaid:
I talked with @ilovezfs and my understanding of the conversation was that we (or at least I) am happy to accept this as long as it's keg-only.

Formula/imagemagick.rb
```diff
@@ -130,6 +130,9 @@ def install
 
   def caveats
     s = <<-EOS.undent
+      rmagick 2.x does not work with ImageMagick 7, and some invokations of
```
crazymykl: Not 100% thrilled with the verbiage here.

crazymykl:
Updated. Let me know if there's more to do.

Formula/imagemagick.rb
```diff
@@ -129,11 +129,17 @@ def install
   end
 
   def caveats
-    s = <<-EOS.undent
+    base = <<-EOS.undent
+      rmagick 2.x does not work with ImageMagick 7, and some invocations of
+      `convert` may have changed syntax.
```
MikeMcQuaid: Let's omit this; there's more than just `rmagick` and this isn't in keeping with other formulae updates.

Formula/imagemagick@6.rb
```diff
@@ -0,0 +1,148 @@
...
+  # unnecessarily and older versions of the formula are broken.
+  url "https://dl.bintray.com/homebrew/mirror/imagemagick-6.9.7-4.tar.xz"
+  mirror "https://www.imagemagick.org/download/ImageMagick-6.9.7-4.tar.xz"
+  sha256 "68842c55ed9c958b84aae17974961cefff4212bf7146f09fd15c09dbdc2d9629"
+
+  bottle do
+    sha256 "9258c4fc88187040a23e8c62e451ef725eb504b721181eba4e6dda382884bf8e" => :sierra
+    sha256 "6bb941944b07a17da426a1fae986ef8b13ba091dd52728e2ed59adb19d5fcce0" => :el_capitan
+    sha256 "7246f8e60983e429ee088e8a3d2f2b3fb780c868cc5f5d3effff4146ddafe12e" => :yosemite
+  end
```
MikeMcQuaid: Remove the bottle block, it'll get re-added when we pull.

Formula/imagemagick@6.rb
```diff
@@ -0,0 +1,148 @@
...
+    system "make", "install"
+  end
+
+  def caveats
+    s = <<-EOS.undent
+      For full Perl support you may need to adjust your PERL5LIB variable:
+        export PERL5LIB="#{HOMEBREW_PREFIX}/lib/perl5/site_perl":$PERL5LIB
+    EOS
+    s if build.with? "perl"
+  end
```
MikeMcQuaid: Probably just worth removing these caveats (or adjusting them to point to `opt_lib` )

Formula/imagemagick@6.rb
```diff
@@ -0,0 +1,148 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "libtool" => :run
+  depends_on "xz"
+
+  depends_on "jpeg" => :recommended
+  depends_on "libpng" => :recommended
+  depends_on "libtiff" => :recommended
+  depends_on "freetype" => :recommended
+
+  depends_on :x11 => :optional
```
MikeMcQuaid: Let's remove this option.

MikeMcQuaid (changes requested):

crazymykl:
I did the thing.

MikeMcQuaid (approved):

MikeMcQuaid:

MikeMcQuaid:
@crazymykl You may want to now remove your tap and/or add the according `tap_migrations.json` line.

PR #8767, opened 2017-01-12
ruby*: import from homebrew/versions

hashar:
Import ruby versions from homebrew/versions. They are all coinstallable
since https://github.com/Homebrew/homebrew-versions/pull/1509

The naming scheme is adjusted, add ruby-versions names translations to
formula_renames.json

Test/audit:
 - [X] ruby1.8.7: test OK, audit --strict OK
 - [X] ruby1.9.3: test OK, audit --strict Ok
 - [X] ruby2.0: test OK, audit --strict OK
 - [X] ruby2.1: test OK, audit --strict OK
 - [X] ruby2.2: test OK, audit --strict OK 
 - [X] ruby2.3: test OK, audit --strict OK
-----


hashar:
This is not the current versions of the formulae on homebrew-versions but a pull request I have sent there to make the rubies versions to install with a suffix by default [homebrew-versions #1509](https://github.com/Homebrew/homebrew-versions/pull/1509).

@MikeMcQuaid [proposed to migrate the ruby formulae](https://github.com/Homebrew/homebrew-versions/pull/1509#issuecomment-272016412) from homebrew-versions to homebrew-core hence this pull request (  ).

We would want to hold merging this #8767 until homebred-version PR is merged in.

hashar:
That is a single commit copy pasting the formulae.  Not sure how to import the commit history, might not be so easy to handle :(

nijikon:
I would really appreciate if each version would be a separate commit also this should be ruby@18, ruby@19 etc, check node for example.

hashar:
(I have fixed up a few brew audit issues, class names were missing `AT`).

> I would really appreciate if each version would be a separate commit

Not sure there is much point but I will split the commit in several bits once I get the test results.

> This should be ruby@18, ruby@19 etc, check node for example.

If I remember correctly, before 2.0, ruby would break compatibility between "patch" versions. Thus 1.8.0 and 1.8.7 are not quite the same rubies. Ditto for  ruby 1.9.3 which is back compat with 1.9.1 but not with 1.9.0.  So we want to keep the three bits version numbering scheme.  That also aligns with what is in homebrew-versions.

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+    sha256 "f40e11d81cdae7051600b72ea5511b9995420618a13b65be439f6978e634d8c5" => :yosemite
+  end
+
+  option :universal
+  option "without-suffix", "Do not suffix commands with '187' (will become keg-only)"
+  option "with-doc", "Install documentation"
+  option "with-tcltk", "Install with Tcl/Tk support"
+
+  provided_by_osx = [:mountain_lion, :lion, :snow_leopard].include? MacOS.version.to_sym
+  if provided_by_osx && (build.without? "suffix")
```
ilovezfs: I do not think we should do conditional keg_only like this. There's already a DSL such as ` keg_only :provided_pre_mountain_lion`. Why do they need to be keg_only when provided by macOS? The main ruby formula isn't.

ilovezfs (changes requested):

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
+class RubyAT187 < Formula
```
nijikon: ruby@1.8

Formula/ruby@1.9.3.rb
```diff
@@ -0,0 +1,180 @@
+class RubyAT193 < Formula
```
nijikon: ruby@1.9

nijikon (changes requested):

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+  revision 2
+
+  bottle do
+    sha256 "bc89a74b22eec03454e42a37ce12df2fd31765e3ed9f87673fb3d073f0d3db05" => :sierra
+    sha256 "85ffdf739e1f4f74979e59616933e57a3a384bfc2214277b34120295ab0a695d" => :el_capitan
+    sha256 "f40e11d81cdae7051600b72ea5511b9995420618a13b65be439f6978e634d8c5" => :yosemite
+  end
+
+  option :universal
+  option "without-suffix", "Do not suffix commands with '187' (will become keg-only)"
```
ilovezfs: This is contrary to the behavior of other such options.

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+  provided_by_osx = [:mountain_lion, :lion, :snow_leopard].include? MacOS.version.to_sym
+  if provided_by_osx && (build.without? "suffix")
+    keg_only :provided_by_osx
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "readline" => :recommended
+  depends_on "gdbm" => :optional
+  depends_on "openssl"
+  depends_on :x11 if build.with? "tcltk"
```
ilovezfs: Since this has been deprecated, do we really still need it in the older rubies?

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+    sha256 "a39e95acb86f52fdca38b1484b2eed61a7f76551119079d7a9f8c67c5be78831"
+  end
+
+  def install
+    # Compilation with `superenv` breaks because the Ruby build system sets
+    # `RUBYLIB` and `RUBYOPT` and this disturbs the wrappers for `cc` and
+    # `clang` that are provided by Homebrew and are implemented in Ruby. We
+    # prefix those invocations with `env` to clean the environment for us.
+    scrub_env = "/usr/bin/env RUBYLIB= RUBYOPT="
+    inreplace "mkconfig.rb", "    if /^prefix$/ =~ name\n",
```
ilovezfs: this seems like it should be a patch in formula-patches

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+              EOS
+
+    args = %W[
+      --prefix=#{prefix}
+      --enable-shared
+      --with-sitedir=#{HOMEBREW_PREFIX}/lib/ruby/site_ruby
+      --with-vendordir=#{HOMEBREW_PREFIX}/lib/ruby/vendor_ruby
+    ]
+
+    if build.universal?
```
ilovezfs: do we really want universal options on what are basically new formulae?

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+
+    system "./configure", *args
+
+    # Ruby has been configured to look in the HOMEBREW_PREFIX for the
+    # sitedir and vendordir directories; however we don't actually want to create
+    # them during the install.
+    #
+    # These directories are empty on install; sitedir is used for non-rubygems
+    # third party libraries, and vendordir is used for packager-provided libraries.
+    inreplace "instruby.rb" do |s|
```
ilovezfs: this also seems like it could go in a patch

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+      mv (prefix/"gem/bin").children, bin
+      mv (prefix/"gem/lib").children, lib/"ruby/#{abi_version}"
+      rm_r prefix/"gem"
+    end
+  end
+
+  def post_install
+    # Customize rubygems to look/install in the global gem directory
+    # instead of in the Cellar, making gems last across reinstalls
+    config_file = lib/"ruby/#{abi_version}/rubygems/defaults/operating_system.rb"
```
ilovezfs: Can abi_version not be hard coded on versioned formulae?

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+    config_file.unlink if config_file.exist?
```
ilovezfs: I think I'd prefer `rm`

ilovezfs: Also rm_f would eliminate the need for the exist check

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+    config_file.write rubygems_config
+
+    # Create the sitedir and vendordir that were skipped during install
+    ruby="#{bin}/ruby#{program_suffix}"
+    %w[sitearchdir vendorarchdir].each do |dir|
+      mkdir_p `#{ruby} -rrbconfig -e 'print RbConfig::CONFIG["#{dir}"]'`
+    end
+
+    # Create the version-specific bindir used by rubygems
+    mkdir_p rubygems_bindir
```
ilovezfs: tend to use `mkpath` for these

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+  end
+
+  def abi_version
```
ilovezfs: probably doesn't need to be a function for the versioned formulae

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+    By default, binaries installed by gem will be placed into:
+      #{rubygems_bindir}
+
+    You may want to add this to your PATH.
+    EOS
+  end
+
+  test do
+    hello_text = shell_output("#{bin}/ruby#{program_suffix} -e 'puts :hello'")
+    assert_equal "hello\n", hello_text
```
ilovezfs: would be better to use `chomp` here

Formula/ruby@1.8.7.rb
nijikon: Purge that

Formula/ruby@1.8.7.rb
nijikon: Purge that

Formula/ruby@1.8.7.rb
nijikon: rm_f

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+
+  provided_by_osx = [:mountain_lion, :lion, :snow_leopard].include? MacOS.version.to_sym
+  if provided_by_osx && (build.without? "suffix")
+    keg_only :provided_by_osx
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "readline" => :recommended
+  depends_on "gdbm" => :optional
+  depends_on "openssl"
```
ilovezfs: is there an option to use secure transport?

Formula/ruby@1.9.3.rb
```diff
@@ -0,0 +1,180 @@
...
+  desc "Powerful, clean, object-oriented scripting language"
+  homepage "https://www.ruby-lang.org/"
+  url "https://cache.ruby-lang.org/pub/ruby/1.9/ruby-1.9.3-p551.tar.bz2"
+  sha256 "b0c5e37e3431d58613a160504b39542ec687d473de1d4da983dabcf3c5de771e"
+  revision 4
```
ilovezfs: they're new formulae so they do not need revisions

Formula/ruby@2.0.rb
```diff
@@ -0,0 +1,187 @@
+class RubyAT20 < Formula
+  desc "Powerful, clean, object-oriented scripting language"
+  homepage "https://www.ruby-lang.org/"
+  url "https://cache.ruby-lang.org/pub/ruby/2.0/ruby-2.0.0-p648.tar.bz2"
+  sha256 "087ad4dec748cfe665c856dbfbabdee5520268e94bb81a1d8565d76c3cc62166"
+  revision 1
```
ilovezfs: no need for revisions on new formulae

Formula/ruby@2.0.rb
```diff
@@ -0,0 +1,187 @@
...
+
+  bottle do
+    sha256 "5e1529fbbff248b8064c00433a1cd7ee5acd0f98c0e840bf0a127c5b7384a870" => :sierra
```
ilovezfs: the old bottle blocks are invalid so should be removed

Formula/ruby@2.2.rb
```diff
@@ -0,0 +1,191 @@
+class RubyAT22 < Formula
+  desc "Powerful, clean, object-oriented scripting language"
+  homepage "https://www.ruby-lang.org/"
+  url "https://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.5.tar.bz2"
+  sha256 "22f0c6f34c0024e0bcaaa8e6831b7c0041e1ef6120c781618b833bde29626700"
+  revision 2
```
ilovezfs: ditto

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
+class RubyAT187 < Formula
+  desc "Powerful, clean, object-oriented scripting language"
+  homepage "https://www.ruby-lang.org/"
+  url "https://cache.ruby-lang.org/pub/ruby/1.8/ruby-1.8.7-p374.tar.bz2"
+  sha256 "b4e34703137f7bfb8761c4ea474f7438d6ccf440b3d35f39cc5e4d4e239c07e3"
+  revision 2
```
ilovezfs: ditto

nijikon:
Actually, why we want to add those? We provide the latest version which we use in our formulas. If somebody needs an older version he/she should install that via rvm or some other software.

Formula/ruby@1.8.7.rb
hashar: I have marked a given ruby version as keg-only only if the same version is provided by Apple. Leopard (10.5) has ruby 1.8.6, Mavericks (10.9) has ruby 2.0.0.

I wasn't aware of the provided_pre shortcut. Note the comparison is strict so should be `:provided_pre_mavericks`.    That could be done if we drop Leopard entirely which comes with ruby 1.8.6.

For reference from opensource.apple.com:

| Ver | Name | Ruby version
|--|--|--
| 10.12 | sierra | 2.0.0-p648
| 10.11 | el_capitan | 2.0.0-p648
| 10.10 | yosemite | 2.0.0-p481
| 10.9 | mavericks | 2.0.0-p481
| 10.8 | mountain_lion | 1.8.7-p357
| 10.7 | lion | 1.8.7-p357
| 10.6 | snow_leopard | 1.8.7-p174
| 10.5 | leopard | 1.8.6-p287

Formula/ruby@1.8.7.rb
hashar: From a previous comment https://github.com/Homebrew/homebrew-core/pull/8767#issuecomment-272139269

That should be the equivalent of an ABI version.  Turns out on previous comment I was wrong for ruby 1.8 and we could use `ruby@1.8` and `class RubyAT18` indeed.

Formula/ruby@1.9.3.rb
hashar: From a previous comment https://github.com/Homebrew/homebrew-core/pull/8767#issuecomment-272139269

Given you are the second person to find it misleading, and that ruby 1.9.0 is long gone. I guess we can just go with `ruby@1.8` and `class RubyAT19`.  It is less pure, but that should be good enough for nowadays usage.

Formula/ruby@1.8.7.rb
hashar: The point of my patch is supporting installation of multiple rubies on the same machine, each having a different suffix.  So you can have `/usr/local/bin/ruby187` and `/usr/local/bin/ruby21` alongside.

I thought one might want to install `ruby2.3` and have it installed as `/usr/local/bin/ruby`.  A use case is one dealing with legacy softwares that do not work on ruby2.4 but yet still have to be maintained and the user willing to have that 2.3 as `ruby`.

For Wikimedia, they have that specific use case. They use Debian Jessie with ruby2.1  and a server configuration management system Puppet at 3.7 which does not work with ruby 2.4.  Hence for developers they want to have their `ruby` to be ruby 2.1 (and not 2.0 from Apple or 2.4 from Homebrew). 

Hope that clarify the intent behind `option "without-suffix"`.

Formula/ruby@1.8.7.rb
hashar: That patches comes from https://github.com/Homebrew/homebrew-versions/commit/656096eac302fd0ae8d49d56b73f0346811f3b72 by @UniqMartin 

The hotfix is only on ruby 1.8.7, is it really worth porting it to a formula-patches?  If so, I would rather do that in a follow up patch.

Formula/ruby@1.8.7.rb
hashar: I have dig in the git log, the original intent in homebrew-version is to have each specific formula to be mostly similar.  So you have bits of codes that are essentially the same, and one can just change a few methods to vary per version.

One can try a `diff ruby187.rb ruby193.rb` or `diff ruby20.rb ruby22.rb`.  With the use of a variable the diff is rather small.

homebrew-php has a nice shared abstraction that is reused by formulae. Maybe we can head to ward that later on.

Formula/ruby@1.8.7.rb
hashar: If that matters `rm_f` ignores any error, but that probably does not matter.

Formula/ruby@1.8.7.rb
hashar: Given one is an alias of the other, does it really matter?   Maybe the preference for `mkpath` should be enforced via `brew style`.

Formula/ruby@1.9.3.rb
hashar: Not so new since they have been introduced in homebrew-versions. We could even bump those revisions for each of them?

hashar:
> Actually, why we want to add those? We provide the latest version which we use in our formulas. If somebody needs an older version he/she should install that via rvm or some other software.

For the same reason there are several node versions when one could uses nvm?   In my experience with end users, it is much easier to just `brew install ruby@2.3`.  See my rationale above on https://github.com/Homebrew/homebrew-core/pull/8767#discussion_r95886521

hashar:
Thanks for the reviews everyone. I have read and acknowledge all of them, I haven't updated the code though because it is late here and that gives you an opportunity to follow up/clarify where needed :-}

Formula/ruby@1.8.7.rb
MikeMcQuaid: Let's just use RubyAT18, yeh.

Formula/ruby@1.8.7.rb
MikeMcQuaid: To be explicit in this comment: can remove this `revision`.

Formula/ruby@1.8.7.rb
```diff
@@ -0,0 +1,205 @@
...
+
+  bottle do
+    sha256 "bc89a74b22eec03454e42a37ce12df2fd31765e3ed9f87673fb3d073f0d3db05" => :sierra
+    sha256 "85ffdf739e1f4f74979e59616933e57a3a384bfc2214277b34120295ab0a695d" => :el_capitan
+    sha256 "f40e11d81cdae7051600b72ea5511b9995420618a13b65be439f6978e634d8c5" => :yosemite
+  end
```
MikeMcQuaid: Remove this `bottle`

Formula/ruby@1.8.7.rb
MikeMcQuaid: Let's pick one: either install these as `ruby` and make them keg-only or install them with a suffix. Let's remove the option regardless. For Wikimedia it sounds like they may want `rbenv` instead.

Formula/ruby@1.8.7.rb
MikeMcQuaid: I agree, this should not be conditional.

Formula/ruby@1.8.7.rb
MikeMcQuaid: Yep, lets remove this option.

Formula/ruby@1.8.7.rb
MikeMcQuaid: I think the hot fix is fine in this case.

Formula/ruby@1.8.7.rb
MikeMcQuaid: Yep, let's remove it.

Formula/ruby@1.8.7.rb
MikeMcQuaid: For the sake of just porting this from homebrew/versions I think it's fine as-is.

Formula/ruby@1.8.7.rb
MikeMcQuaid: Yeh, agreed on `rm_f`

Formula/ruby@1.8.7.rb
MikeMcQuaid: Doesn't really matter but may as well cleanup while you're here, thanks.

Formula/ruby@1.8.7.rb
MikeMcQuaid: I don't have strong feelings here either way.

Formula/ruby@1.9.3.rb
MikeMcQuaid: Yeh, agreed on nowadays usage 👍

Formula/ruby@1.9.3.rb
MikeMcQuaid: Can just remove them as they will be considered new on migration.

MikeMcQuaid (changes requested):

MikeMcQuaid:
> Thanks for the reviews everyone. I have read and acknowledge all of them, I haven't updated the code though because it is late here and that gives you an opportunity to follow up/clarify where needed :-}

@hashar No worries, should be pretty much complete now, thanks!

dunn:
This needs rebasing, please!

hashar:
@dunn I will follow up eventually :}

swrobel:
@hashar any plans to get back to this or is it worth considering a new PR?

ilovezfs:
@MikeMcQuaid so are we closing this given that these are covered by https://github.com/Homebrew/homebrew-core/pull/10862 ?

MikeMcQuaid:
Yes, thanks anyway @hashar!

PR #8785, opened 2017-01-12
autorest 0.17.3 (new formula)

mandrean:


Formula/autorest.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    libexec.install Dir["*"]
+    (bin/"autorest").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/autorest.exe "$@"
+    EOS
+  end
+
+  test do
+    assert_match "Microsoft (R) AutoRest #{version}", shell_output("#{bin}/autorest | head -1")
```
bfontaine: Can’t we write a test that does something more substantial than printing the version?

bfontaine (dismissed):

Formula/autorest.rb
```diff
@@ -0,0 +1,21 @@
...
+    libexec.install Dir["*"]
+    (bin/"autorest").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/autorest.exe "$@"
+    EOS
+  end
+
+  test do
+    system "curl", "-#SLO", "https://raw.githubusercontent.com/Azure/autorest/master/Samples/petstore/petstore.json"
+    assert_match "Finished generating CSharp code for petstore.json.", shell_output("#{bin}/autorest -n test -i petstore.json | tail -1")
```
mandrean: @bfontaine Better? The tests now actually run through the core functionality of the app.

mandrean:
@bfontaine @ilovezfs @nijikon

Formula/autorest.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    libexec.install Dir["*"]
+    (bin/"autorest").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/autorest.exe "$@"
+    EOS
+  end
+
+  test do
+    system "curl", "-#SLO", "https://raw.githubusercontent.com/Azure/autorest/master/Samples/petstore/petstore.json"
```
ilovezfs: This should be fetched using a resource block.

Formula/autorest.rb
```diff
@@ -0,0 +1,21 @@
...
+  url "https://github.com/Azure/autorest/releases/download/AutoRest-0.16.0/autorest.0.16.0.zip"
+  sha256 "bc909aaf3db2105a5d709c940c4d120c5b71047539c2068d484ced1be47650c3"
+
+  depends_on "mono"
+
+  def install
+    libexec.install Dir["*"]
+    (bin/"autorest").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/autorest.exe "$@"
```
ilovezfs: The file name is actually AutoRest.exe, which is an important distinction on a case-sensitive file system (I got test failure as a result)

Formula/autorest.rb
ilovezfs: shell_output("#{bin}/autorest -n test -i petstore.json").lines.last would be preferable to `| tail -1`

mandrean:
@ilovezfs Now? Added changes according to your requests.

Formula/autorest.rb
```diff
@@ -0,0 +1,27 @@
...
+  sha256 "bc909aaf3db2105a5d709c940c4d120c5b71047539c2068d484ced1be47650c3"
+
+  depends_on "mono"
+
+  resource "swagger" do
+    url "https://raw.githubusercontent.com/Azure/autorest/master/Samples/petstore/petstore.json"
+    sha256 "8de4043eff83c71d49f80726154ca3935548bd974d915a6a9b6aa86da8b1c87c"
+  end
+
+  def install
```
woodruffw: Any chance of a source build instead of installing a binary?

We've made exceptions in the past or certain large/difficult-to-build packages, but we strongly prefer a source build where possible.

mandrean: @woodruffw According to this link, it takes approximately 100 minutes to build AutoRest?... https://github.com/Azure/autorest/blob/master/docs/developer/guide/building-code.md

woodruffw: 100 minutes isn't long by our standards - `gcc` and `qt` both take substantially longer on average. That's part of the reason we provide bottles -- they give us all of the control of source builds without any of the time issues.

I'm personally in favor of keeping this in a [custom tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md) until upstream makes their build procedure more clear/practical on OS X.

MikeMcQuaid: @woodruffw FYI we have two exceptions for building from source: Mono and Java as the outputs are not strictly binaries. I'm fine with allowing this as-is.

Formula/autorest.rb
```diff
@@ -0,0 +1,27 @@
...
+    libexec.install Dir["*"]
+    (bin/"autorest").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/AutoRest.exe "$@"
+    EOS
+  end
+
+  test do
+    resource("swagger").stage do
+      assert_match "Finished generating CSharp code for petstore.json.", shell_output("#{bin}/autorest -n test -i petstore.json | tail -1")
```
MikeMcQuaid: Newline after `,` to make this a little more readable.

Formula/autorest.rb
MikeMcQuaid: (as it'll work as-is on all macOS versions and Linux)

MikeMcQuaid: @woodruffw Let me know if you're 👍 and where you think this is worth documenting

woodruffw: 100% fine, thanks for clarifying!

I realized that https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md#we-dont-like-binary-formulae covers this, but I wasn't reading it closely enough :smile:

Formula/autorest.rb
mandrean: @MikeMcQuaid Better?

mandrean:
Updated this to use Nuget (https://www.nuget.org/packages/AutoRest/0.17.3) as source instead of GitHub, because Nuget has the latest releases (0.17.3) while GitHub only has a really old one (0.16.0)

(nupkg's are just zip files with another extension)

All the usual stuff works fine for me after this change (test/audit/install etc.)

Formula/autorest.rb
```diff
@@ -0,0 +1,28 @@
...
+    (bin/"autorest").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/AutoRest.exe "$@"
+    EOS
+  end
+
+  test do
+    resource("swagger").stage do
+      assert_match "Finished generating CSharp code for petstore.json.",
+        shell_output("#{bin}/autorest -n test -i petstore.json | tail -1")
```
MikeMcQuaid: Can drop the `| tail -1` as it'll match regardless.

Formula/autorest.rb
```diff
@@ -0,0 +1,28 @@
+class Autorest < Formula
+  desc "Swagger (OpenAPI) Specification code generator"
+  homepage "https://github.com/Azure/autorest"
+  url "https://api.nuget.org/packages/autorest.0.17.3.nupkg"
+  sha256 "b3f5b67ae1a8aa4f0fd6cf1e51df27ea1867f0c845dbb13c1c608b148bd86296"
+
+  depends_on "mono"
+
+  resource "swagger" do
+    url "https://raw.githubusercontent.com/Azure/autorest/master/Samples/petstore/petstore.json"
```
MikeMcQuaid: Use a versioned link like e.g. https://raw.githubusercontent.com/Azure/autorest/7ae1623f03b3abe646a3ec5ba34a9c60ba5d0781/Samples/petstore/petstore.json

MikeMcQuaid (changes requested):

Formula/autorest.rb
mandrean: @MikeMcQuaid Done!

Formula/autorest.rb
mandrean: @MikeMcQuaid Done as well. Used a different commit hash than yours (used same as v0.17.3 release tag)

MikeMcQuaid (approved):

MikeMcQuaid:

PR #8789, opened 2017-01-12
cloudflare-ddns 1.0.1 (new formula)

wyattjoh:

-----

### DIFF
--- a/Formula/cloudflare_ddns.rb
+++ b/Formula/cloudflare_ddns.rb
@@ -1,0 +1,23 @@
+class CloudflareDdns < Formula
+  desc "Cloudflare Dynamic DNS Updater"
+  homepage "https://github.com/wyattjoh/cloudflare-ddns"
+  url "https://github.com/wyattjoh/cloudflare-ddns/archive/v1.0.1.tar.gz"
+  sha256 "e13ab6c203eb646ebb23384c0136f469b8e72b05d31f43f288ff9671e601809c"
+  head "https://github.com/wyattjoh/cloudflare-ddns.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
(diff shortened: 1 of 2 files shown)


woodruffw:
Thanks for your submission. Unfortunately, we can't merge this into `homebrew-core` in its current state - we generally like to see packages gain a bit more attention before adding them to core. It's also generally a good idea to get a user (instead of a maintainer) to submit the package, so that we can gauge demand.

In the mean time, you can always host it in your [own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md). I'll also leave some notes so that you can get an idea of what our review process is like for future submissions.

Formula/cloudflare_ddns.rb
```diff
@@ -0,0 +1,23 @@
...
+    dir = buildpath/"src/github.com/wyattjoh/cloudflare-ddns"
+    dir.install buildpath.children
+    cd dir do
+      system "go", "build", "-o", bin/"cloudflare-ddns"
+      prefix.install_metafiles
+    end
+  end
+
+  test do
+    system bin/"cloudflare-ddns", "--help"
```
woodruffw: We try to use `--help` and `--version` tests as a last resort, preferring tests of more realistic functionality (even just an error message from a command that needs credentials).

If `--help` or `--version` is really all that's practical, testing their output with `shell_output` and `assert_match` is usually a good middle ground.

wyattjoh: Thanks for the feedback! I had originally copied what `heroku` had done, but I've amended my fork of the formula to match your recommended flow.

If it ever get's popular enough (which, seems unlikely) I'll use this as a feedback.

wyattjoh: https://github.com/wyattjoh/homebrew-core/commit/5d609f7f017bd51cf4f2d68c0489444bf85ac947

PR #8790, opened 2017-01-12
OpenCoarrays 1.8.2 (new formula)

zbeekman:

-----

### DIFF
--- a/Formula/opencoarrays.rb
+++ b/Formula/opencoarrays.rb
@@ -1,0 +1,50 @@
+class Opencoarrays < Formula
+  desc "open-source coarray Fortran ABI, API, and compiler wrapper"
+  homepage "http://opencoarrays.org"
+  url "https://github.com/sourceryinstitute/opencoarrays/releases/download/1.8.2/OpenCoarrays-1.8.2.tar.gz"
+  sha256 "dc5b98a62de85e38500f431b05290415d4a4a87001c0ba03fbe57403ad62010b"
+
+  head "https://github.com/sourceryinstitute/opencoarrays.git"
+
+  option "without-test", "Skip build time tests (not recommended)"
+
+  depends_on "gcc"
+  depends_on :fortran
+  depends_on :mpi => :cc
+  depends_on "cmake" => :build
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make"
+      system "ctest", "--output-on-failure", "--schedule-random" if build.with? "test"
+      system "make", "install"
+    end
+  end
+
+  test do
+    ENV.fortran
+    (testpath/"tally.f90").write <<-EOS.undent
+      program main
+        use iso_c_binding, only : c_int
+        use iso_fortran_env, only : error_unit
+        implicit none
+        integer(c_int) :: tally
+        tally = this_image() ! this image's contribution
+        call co_sum(tally)
+        verify: block
+          integer(c_int) :: image
+          if (tally/=sum([(image,image=1,num_images())])) then
+             write(error_unit,'(a,i5)') "Incorrect tally on image ",this_image()
+             error stop 2
+          end if
+        end block verify
+        ! Wait for all images to pass the test
+        sync all
+        if (this_image()==1) write(*,*) "Test passed"
+      end program
+    EOS
+    system "#{bin}/caf", "tally.f90", "-o", "tally"
+    system "#{bin}/cafrun", "-np", "3", "./tally"
+  end
+end


zbeekman:
This is a resurrection of https://github.com/Homebrew/legacy-homebrew/pull/46547

This comment may still need to be resolved, once I figure out how: https://github.com/Homebrew/legacy-homebrew/pull/46547

A bunch of issues have been fixed with the upstream build system... I'm looking into the build failures here, but one might be because I just force pushed to delete some erroneous comments that I accidentally left in the Formula...

zbeekman:
> I'm looking into the build failures here, but one might be because I just force pushed to delete some erroneous comments that I accidentally left in the Formula

I'm a derp... I had the formula in the wrong place... this should be good to go, assuming tests pass... I'll look into creating a Requirements class for GCC, as discussed [here](https://github.com/Homebrew/legacy-homebrew/pull/46547#r55076413), but as I said in the old PR, my ruby-fu is a bit lacking...

zbeekman:
> I'll look into creating a Requirements class for GCC, as discussed here, but as I said in the old PR, my ruby-fu is a bit lacking...

I looked at https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/mpi_requirement.rb and https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/java_requirement.rb and I think coming up with something like this for GCC may be beyond my ruby capabilities...

zbeekman:
This PR should be good to go pending review by a Homebrew team member

zbeekman:
Sorry for the noise... one final comment: If you merge this PR, could you please enable the bottle download analytics for public access on bintray? Thanks! :bow: :pray:

Formula/open-coarrays.rb
```diff
@@ -0,0 +1,57 @@
+class OpenCoarrays < Formula
+  desc "open-source coarray Fortran ABI, API, and compiler wrapper"
+  homepage "http://opencoarrays.org"
+  url "https://github.com/sourceryinstitute/opencoarrays/releases/download/1.8.2/OpenCoarrays-1.8.2.tar.gz"
+  sha256 "dc5b98a62de85e38500f431b05290415d4a4a87001c0ba03fbe57403ad62010b"
+
+  head "https://github.com/sourceryinstitute/opencoarrays.git"
+
+  option "without-test", "Skip build time tests (not recommended)"
```
woodruffw: If it's not recommended, I think we can just nix it entirely :wink:

woodruffw:
Thanks for reviving this! I made one (minor) comment, but this LGTM. The original reviewers will probably have other comments.

Formula/open-coarrays.rb
zbeekman: Well.... I don't know... Sometimes users would rather have the software because it's hard to get one's hands on Fortran compilers supporting recent 2008 and 2015 (or 16? or 17? not sure what the standard ended up getting called...) coarray fortran features. On some systems, e.g., when linux brew pulls this in down stream, you might encounter test failures... but after seeing that you may still want to install OpenCoarrays and proceed cautiously with the library anyway... If you don't provide this build time option, than a test failure will completely prevent users from installing via Homebrew (or Linuxbrew). For this reason I would much rather keep this option in.

woodruffw: Good point, ignore this then.

zbeekman: (btw this is [the official GFortran back end](https://gcc.gnu.org/wiki/CoarrayLib) implementation of Coarray Fortran, added in the 2008 standard.)

zbeekman:
OK, sorry to force push again, but I was just double checking the release notes and for simplicity it looks like we should restrict to GFortran >= 5.4, but, at the moment the restriction is >= 5.3. I'm going to adjust the fails_with each.do again to reflect this...

Formula/open-coarrays.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  depends_on :fortran
+  depends_on :mpi => [:cc, :f90]
+  depends_on "cmake" => :build
+
+  # GFortran >= 5.4 required; no other Homebrew supported compilers are compatible
+  # but the mpicc compiler may be clang (which is ABI compatible with GCC)
+  fails_with :gcc
+  fails_with :gcc_4_0
+  [("4.3".."4.9"), ("5.0".."5.3")].each do |series|
```
MikeMcQuaid: Not sure `5.0` will do anything; we have a `gcc@5` rather than `gcc@5.1` formula

MikeMcQuaid:
> @MikeMcQuaid 's comment regarding implementing a requirements class for GFortran to replace L15-L21, a la https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/java_requirement.rb This is, however, far beyond my abilities with Ruby.

I'm fine with this PR (almost) as-is without a requirement 👍

Formula/open-coarrays.rb
zbeekman: so, would this do the trick then:

```ruby
 [("4.3".."4.9"), "5", ("5.1".."5.3")].each do |series|
```

?

MikeMcQuaid: I'd have thought, `[("4.3".."4.9"), "5"]` would be sufficient but would be good to test it out.

Formula/open-coarrays.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  option "without-test", "Skip build time tests (not recommended)"
+
+  depends_on :fortran
+  depends_on :mpi => [:cc, :f90]
+  depends_on "cmake" => :build
+
+  # GFortran >= 5.4 required; no other Homebrew supported compilers are compatible
+  # but the mpicc compiler may be clang (which is ABI compatible with GCC)
+  fails_with :gcc
```
MikeMcQuaid: `fails_with :gcc_4_2`

zbeekman: @MikeMcQuaid Add this too? OK.

MikeMcQuaid: Aye, cheers.

Formula/open-coarrays.rb
zbeekman: Yes, but I think 5.4 works without issue... However, I would be just as happy to completely disallow 5.x branch entirely since there are added features/better support/fewer bugs in GFortran 6.x.... I'll defer to your preference, @MikeMcQuaid 

I can test this, but it might necessitate rebuilding GCC 5 from source....

zbeekman: > I can test this, but it might necessitate rebuilding GCC 5 from source....

Actually I can try getting it from the brew-versions tap, if you think that will work and if they provide bottles... I'll try it out and see.

zbeekman:
@MikeMcQuaid somewhat related question: When GFortran 7 comes out (should be later this month) libgfortran.so is getting bumped from 3 to 4 I think... currently libgfortran.3.so soon libgfortran.4.so with GCC 7.... Is there anyway that I should be getting ready for this? Currently if I install GCC using --head OpenCoarrays works as expected, except for the fact that I need to soft link lib/gcc/6 from gcc 6's cellar into /usr/local/opt/lib/gcc/6 so that MPICH/OpenMPI will work as expected... otherwise it complains that it can't find /usr/local/opt/lib/gcc/6/.../libgfortran.3.dylib

MikeMcQuaid:
@zbeekman Sorry, I'm not sure I really understand the question.

zbeekman:
> @zbeekman Sorry, I'm not sure I really understand the question.

I guess my question is this: Some packages, like OpenMPI/MPICH depend on libgfortran.3.dylib which gets linked into /usr/local/opt/lib/gcc/6/... If I unlink GCC 6 so that I can link GCC 7/HEAD, then OpenMPI/MPICH stop working and start complaining about missing /usr/local/opt/lib/gcc/6/.../libgfortran.3.dylib. However, GCC 7/trunk provides /usr/local/opt/lib/7/.../libgfortran.4.dylib.  Is there a recommended way that formulas like MPICH that are linked against /usr/local/opt/lib/gcc/6/.../libgfortran.3.dylib can still be used, even when GCC 6 is unlinked because I'm using GCC 7/trunk?

Anyway this may be a moot point, because when GCC 7 is released (a few days or weeks from now) homebrew will have to deal with updated formulae and the migration from libgfortran.3.dylib to libgfortran.4.dylib with the ABI incompatibilities anyway, at which point someone, through necessity, will have found an acceptable resolution to my issue....

zbeekman:
> I'm fine with this PR (almost) as-is without a requirement

I've tried my hand at throwing together a hacked up requirement on GFortran, but I don't think it's working... I would love any comments/feedback from anyone willing to tolerate my unsound and incomplete understanding of Ruby 🙃 💎 It is here: https://github.com/Homebrew/brew/pull/1864

zbeekman:
IMO, this PR is ready to be merged, pending approval that these latest changes address @MikeMcQuaid's previous comments.

If anyone thinks that https://github.com/Homebrew/brew/issues/1864 is on the right track and close to being functional, and merged, then we can adjust the `fails_with` to become `depends_on :gfortran => "5.4+"` but I'm doubtful that it's worth the hassle.

I also had a misunderstanding about naming the file and class names, and how it relates to preferred formula names, so I have updated formula filename to be opencoarrays.rb, the class name to be Opencoarrays since the project name is OpenCoarrays (no hyphen or underscore)

MikeMcQuaid:
@zbeekman Do you need a `fails_with :clang`? I'm wondering from https://github.com/Homebrew/brew/pull/1864 given your requirements (and some sleep) if just `depends_on "gcc" => :build` is sufficient for your use-case.

zbeekman:
OK, I think I may need to back up a little bit here. This is what OpenCoarrays needs:

 - CMake (build dependency)
 - An MPI 3 Implementation (build and runtime dependency)
 - A modern C compiler (build dependency)
 - GFortran >= 5.4 (runtime dependency, AND needed at build time for system introspection and ensuring API/ABI match)

@MikeMcQuaid said:
> @zbeekman Do you need a fails_with :clang? I'm wondering from Homebrew/brew#1864 given your requirements (and some sleep) if just depends_on "gcc" => :build is sufficient for your use-case.

The easiest thing to do would be to specify a dependency on GCC >= 6 and ensure that the `--without-fortran` flag was*NOT* passed. I just tried to catch up on the new formula versioning schemes by reading through https://github.com/Homebrew/brew/pull/1767/files and https://github.com/Homebrew/brew/issues/620 ... I did some experiments and it seems I can't specify `=> "with-fortran"` for the GCC dependency (because it is a default...) and I also can't specify gcc@6. Despite this, I think the combinations of `depends_on "gcc"` (build and run, do I need to add `=> [:build, :run]`?) and `depends_on :fortran` should be strong enough to nearly guarantee in almost all cases GFortran being used as the Fortran compiler...

Perhaps rather than a new requirements I just examine the output of `$FC --version` from within opencoarrays.rb and bail/complain if it's not GFortran and not >= 6? Something like this to get the version: https://github.com/zbeekman/brew/blob/31a26307c97d431f79555bbec8daa97d3f907277/Library/Homebrew/requirements/gfortran_requirement.rb#L8

Here is more of the full story about OpenCoarrays, "starting at the beginning":

OpenCoarrays is a library providing multi-image (multi core, shared memory or distributed shared memory) Coarray Fortran (CAF) for GFortran. The library itself can be compiled with any modern enough C compiler. However GFortran is needed *at runtime* because, well, there's no point of installing OpenCoarrays without it.[1] In addition, GFortran is needed at compile time, because OpenCoarrays gets called by the code generated by GFortran, and OpenCoarrays must be built to ensure that it's API/ABI matches the GFortran API/ABI. There are some diagnostics, system introspection and conditional compilation that happen when building OpenCoarrays. Lastly, OpenCoarrays provides wrapper scripts, *a la* MPI, to compile and execute Coarray Fortran programs.

[1]: This isn't strictly true, it can work to varying degrees with other Fortran compilers.... and ongoing work to provide source to source transformations so that it can be used with ANY Fortran compiler is under way...

MikeMcQuaid:
> Despite this, I think the combinations of depends_on "gcc" (build and run, do I need to add => [:build, :run]?) and depends_on :fortran should be strong enough to nearly guarantee in almost all cases GFortran being used as the Fortran compiler...

Yes, just do `depends_on "gcc"` and `depends_on :fortran` unconditionally for `with-fortran`. Is it worth this option being the default?

zbeekman:
> Is it worth this option being the default?

I don't understand what you're asking.



> Yes, just do depends_on "gcc" and depends_on :fortran unconditionally for with-fortran

Yes I have this in the latest push

Formula/opencoarrays.rb
```diff
@@ -0,0 +1,54 @@
...
+  option "without-test", "Skip build time tests (not recommended)"
+
+  depends_on "gcc"
+  depends_on :fortran
+  depends_on :mpi => :cc
+  depends_on "cmake" => :build
+
+  fails_with :gcc
+  fails_with :gcc_4_0
+  fails_with :gcc_4_2
```
zbeekman: Do I still need these? We do want a relatively modern C compiler... but I don't know that this is still necessary... won't `depends_on "gcc"` mean that GCC is linked and we're using clang or brewed gcc?

MikeMcQuaid:
> I don't understand what you're asking.

Sorry, ignore me.

Formula/opencoarrays.rb
MikeMcQuaid: You can remove all of these and assume Clang or GCC will be used, yes.

Formula/opencoarrays.rb
```diff
@@ -0,0 +1,55 @@
...
+  depends_on :fortran
+  depends_on :mpi => :cc
+  depends_on "cmake" => :build
+
+  def install
+    fc_string = Utils.popen_read(ENV.fc, "--version").split("\n").first.chomp
+    ohai "Building OpenCoarrays with the following fortran compiler: \n" + fc_string
+    odie "OpenCoarrays requires GFortran" unless fc_string =~ /gnu/i
+    fc_version = fc_string.split(" ").last
+    odie "OpenCoarrays requires GFortran 5.4 or newer" unless Gem::Version.new(fc_version) >= Gem::Version.new("5.4.0")
```
zbeekman: @MikeMcQuaid I'm not sure if this is the best/correct way to do this, but I have added checks on the Fortran compiler by parsing some of the output of `ENV.fc --version` I print diagnostic info (the first line of that output) with `ohai` and I throw errors if it's not `GNU Fortran` and also `>= 5.4.0` using `odie`.

Unless you want me to adjust any of this, I'm quite happy with it and think that we're finally good to go.

MikeMcQuaid: Let's just leave it with `depends_on "gcc"` and `depends_on :fortran`. That'll be sufficient.

zbeekman: so to clarify, you think I should remove the `odie` and version parsing logic?

zbeekman:
@MikeMcQuaid I've made the adjustments you requested and removed the `odie` hackage...

If/when you merge, can you please enable the public bintray download statistics for the bottle? Cheers!

MikeMcQuaid:
> If/when you merge, can you please enable the public bintray download statistics for the bottle? Cheers!

This is done by default now, FYI: 
https://github.com/Homebrew/homebrew-test-bot/blob/46178ff5bb968e01c2153fda7e79a5129f0740a9/cmd/brew-test-bot.rb#L1050-L1051

MikeMcQuaid:

PR #8792, opened 2017-01-13
consul-template 0.18.0-rc2 (devel)

cblecker:

-----

A number of changes to incorporate the new make/install process required for devel/head, remove the need for a patch, but still allow the current stable to build. Marked the sections in this formula that can be fixed up once devel becomes stable.

### DIFF
--- a/Formula/consul-template.rb
+++ b/Formula/consul-template.rb
@@ -1,11 +1,26 @@
-require "language/go"
-
 class ConsulTemplate < Formula
   desc "Generic template rendering and notifications with Consul"
   homepage "https://github.com/hashicorp/consul-template"
-  url "https://github.com/hashicorp/consul-template.git",
-      :tag => "v0.16.0",
-      :revision => "efa462daa2b961bff683677146713f4008555fba"
+  head "https://github.com/hashicorp/consul-template.git"
+
+  stable do
+    url "https://github.com/hashicorp/consul-template.git",
+        :tag => "v0.16.0",
+        :revision => "efa462daa2b961bff683677146713f4008555fba"
+
+    # The stable block can be removed on next stable version as the below
+    # go_resource dependencies are no longer required.
+    # Ref: https://github.com/hashicorp/consul-template/issues/793
+    require "language/go"
+    go_resource "github.com/mitchellh/gox" do
+      url "https://github.com/mitchellh/gox.git",
+          :revision => "c9740af9c6574448fd48eb30a71f964014c7a837"
+    end
+    go_resource "github.com/mitchellh/iochan" do
+      url "https://github.com/mitchellh/iochan.git",
+          :revision => "87b45ffd0e9581375c491fef3d32130bb15c5bd7"
+    end
+  end
 
   bottle do
     cellar :any_skip_relocation
@@ -16,50 +31,35 @@ class ConsulTemplate < Formula
 
   devel do
     url "https://github.com/hashicorp/consul-template.git",
-        :tag => "v0.18.0-rc1",
-        :revision => "933192dc0b6bb642af2549848e0681670550e095"
-    version "0.18.0-rc1"
-
-    # Upstream issue "Request: Makefile target that doesn't use docker"
-    # Reported 18 Nov 2016 https://github.com/hashicorp/consul-template/issues/793
-    patch :DATA
-  end
-
-  head do
-    url "https://github.com/hashicorp/consul-template.git"
-
-    patch :DATA
+        :tag => "v0.18.0-rc2",
+        :revision => "e446ab16fb567e1ddeb88ef29406b53bdb4ee387"
+    version "0.18.0-rc2"
   end
 
   depends_on "go" => :build
 
-  go_resource "github.com/mitchellh/gox" do
-    url "https://github.com/mitchellh/gox.git",
-        :revision => "c9740af9c6574448fd48eb30a71f964014c7a837"
-  end
-
-  # gox dependency
-  go_resource "github.com/mitchellh/iochan" do
-    url "https://github.com/mitchellh/iochan.git",
-        :revision => "87b45ffd0e9581375c491fef3d32130bb15c5bd7"
-  end
-
   def install
     ENV["GOPATH"] = buildpath
-
+    arch = MacOS.prefer_64_bit? ? "amd64" : "386"
+    ENV["XC_OS"] = "darwin"
+    ENV["XC_ARCH"] = arch
     dir = buildpath/"src/github.com/hashicorp/consul-template"
-    dir.install buildpath.children
-
-    Language::Go.stage_deps resources, buildpath/"src"
-    ENV.prepend_create_path "PATH", buildpath/"bin"
-    cd("src/github.com/mitchellh/gox") { system "go", "install" }
+    dir.install buildpath.children - [buildpath/".brew_home"]
 
     cd dir do
-      system "make", "dev"
-      system "make", "test"
+      if build.stable?
+        # On next stable release, this if/else block can be replaced below with
+        # just the steps in the else block.
+        Language::Go.stage_deps resources, buildpath/"src"
+        ENV.prepend_create_path "PATH", buildpath/"bin"
+        cd(buildpath/"src/github.com/mitchellh/gox") { system "go", "install" }
+        system "make", "dev"
+        bin.install "bin/consul-template"
+      else
+        system "make", "bin-local"
+        bin.install "pkg/darwin_#{arch}/consul-template"
+      end
     end
-
-    bin.install "bin/consul-template"
   end
 
   test do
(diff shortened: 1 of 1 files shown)


Formula/consul-template.rb
```diff
@@ -16,50 +18,56 @@ class ConsulTemplate < Formula
...
+
+      if build.stable?
+        # This can be removed on next stable version
+        Language::Go.stage_deps resources, gopath/"src"
+        cd(gopath/"src/github.com/mitchellh/gox") { system "go", "install" }
+        system "make", "dev"
+        bin.install "bin/consul-template"
+      end
+
+      if build.devel? | build.head?
```
ilovezfs: This should be `||` but you can drop the entire line and make it an `else` at line 62.

Formula/consul-template.rb
```diff
@@ -16,50 +18,56 @@ class ConsulTemplate < Formula
...
-
-    patch :DATA
+        :tag => "v0.18.0-rc2",
+        :revision => "e446ab16fb567e1ddeb88ef29406b53bdb4ee387"
+    version "0.18.0-rc2"
   end
 
   depends_on "go" => :build
 
+  # This can be removed on next stable version
```
ilovezfs: Does it still use gox?

Formula/consul-template.rb
```diff
@@ -16,50 +18,56 @@ class ConsulTemplate < Formula
...
-    dir.install buildpath.children
-
-    Language::Go.stage_deps resources, buildpath/"src"
-    ENV.prepend_create_path "PATH", buildpath/"bin"
-    cd("src/github.com/mitchellh/gox") { system "go", "install" }
-
-    cd dir do
-      system "make", "dev"
-      system "make", "test"
+    contents = Dir["{*,.git,.gitignore}"]
```
ilovezfs: there's no need for this if you use `buildpath.chidlren` as was previously here.

Formula/consul-template.rb
```diff
@@ -16,50 +18,56 @@ class ConsulTemplate < Formula
...
+
+    ENV["GOPATH"] = gopath
+    ENV.prepend_create_path "PATH", gopath/"bin"
+    arch = MacOS.prefer_64_bit? ? "amd64" : "386"
+    ENV["XC_OS"] = "darwin"
+    ENV["XC_ARCH"] = arch
+
+    cd gopath/"src/github.com/hashicorp/consul-template" do
+      # Ensure a clean git tree for version output
+      system "git", "reset", "--hard"
```
ilovezfs: The reset and clean should not be necessary. If they are, this needs to be reported as a bug to upstream.

Formula/consul-template.rb
```diff
@@ -16,50 +18,56 @@ class ConsulTemplate < Formula
...
+        # This can be removed on next stable version
+        Language::Go.stage_deps resources, gopath/"src"
+        cd(gopath/"src/github.com/mitchellh/gox") { system "go", "install" }
+        system "make", "dev"
+        bin.install "bin/consul-template"
+      end
+
+      if build.devel? | build.head?
+        # This can replace the above stable steps
+        system "make", "bin-local"
```
ilovezfs: does this build other architectures as well?

Formula/consul-template.rb
```diff
@@ -16,50 +18,56 @@ class ConsulTemplate < Formula
...
-    cd("src/github.com/mitchellh/gox") { system "go", "install" }
-
-    cd dir do
-      system "make", "dev"
-      system "make", "test"
+    contents = Dir["{*,.git,.gitignore}"]
+    gopath = buildpath/"gopath"
+    (gopath/"src/github.com/hashicorp/consul-template").install contents
+
+    ENV["GOPATH"] = gopath
```
ilovezfs: GOPATH should be buildpath and this should be at the top of install as it was.

Formula/consul-template.rb
cblecker: In stable, yes. In devel+head, no. This is mentioned in https://github.com/hashicorp/consul-template/issues/793

Formula/consul-template.rb
cblecker: Fixed.

Formula/consul-template.rb
cblecker: Fixed.

Formula/consul-template.rb
cblecker: I used this from the kubernetes-cli formula that also has a build time check for a clean branch. The problem detected here was `.brew_home` is in `buildpath.children`, so this is dirtying the branch. I've made a change to prevent `.brew_home` from being copied, and now the branch is reporting clean.

Formula/consul-template.rb
cblecker: Yes, [it does](https://github.com/hashicorp/consul-template/blob/master/Makefile). I set `XC_OS` and `XC_ARCH` earlier up to restrict it to only the active architecture.

Formula/consul-template.rb
cblecker: Changed to `else`.

cblecker (commented):
Thanks for the review @ilovezfs! I'm pushing a new commit revision with changes.

Formula/consul-template.rb
```diff
@@ -16,50 +18,49 @@ class ConsulTemplate < Formula
...
-    dir.install buildpath.children
-
-    Language::Go.stage_deps resources, buildpath/"src"
-    ENV.prepend_create_path "PATH", buildpath/"bin"
-    cd("src/github.com/mitchellh/gox") { system "go", "install" }
-
-    cd dir do
-      system "make", "dev"
-      system "make", "test"
+    gopath = buildpath
```
ilovezfs: You can just use `buildpath` everywhere you're using `gopath` right now. There's no need for a separate variable.

Formula/consul-template.rb
```diff
@@ -16,50 +18,49 @@ class ConsulTemplate < Formula
...
+    (gopath/"src/github.com/hashicorp/consul-template").install buildpath.children - [buildpath/".brew_home"]
+
+    ENV["GOPATH"] = gopath
```
ilovezfs: this should be
```
  def install
    ENV["GOPATH"] = buildpath
```

Formula/consul-template.rb
```diff
@@ -16,50 +18,49 @@ class ConsulTemplate < Formula
...
-
-    Language::Go.stage_deps resources, buildpath/"src"
-    ENV.prepend_create_path "PATH", buildpath/"bin"
-    cd("src/github.com/mitchellh/gox") { system "go", "install" }
-
-    cd dir do
-      system "make", "dev"
-      system "make", "test"
+    gopath = buildpath
+    (gopath/"src/github.com/hashicorp/consul-template").install buildpath.children - [buildpath/".brew_home"]
```
ilovezfs: Does this install actually break without ` - [buildpath/".brew_home"]`

If not, this should just be
```
(buildpath/"src/github.com/hashicorp/consul-template").install buildpath.children
```

Formula/consul-template.rb
ilovezfs: These go_resource blocks should be moved into a `stable do` since they're only used by the stable spec. The block goes above the bottle block, and also should contain the stable url, :tag, :revision.

Formula/consul-template.rb
```diff
@@ -16,50 +18,49 @@ class ConsulTemplate < Formula
...
+
+    ENV["GOPATH"] = gopath
+    ENV.prepend_create_path "PATH", gopath/"bin"
```
ilovezfs: The `buildpath/"bin"` directory only needs to be in the `PATH` if it's the stable spec and we're using `gox`, right? This can be moved into the `if build.stable?` below.

Formula/consul-template.rb
cblecker: The install doesn't completely break, but when you run `consul-template --version` the output notes the git tree is dirty (adding a `-CHANGES` to the version). This is what was being solved by the `git reset` and `git clean` you commented on here:  https://github.com/Homebrew/homebrew-core/pull/8792#discussion_r95954955.

ilovezfs: That's funny. So the clean was an indirect way to blow away the home directory?

cblecker: Yup. Used `--interactive` to debug why the git tree was dirty. There are another couple formulas I've contributed to that have a similar issue, so I'm going to look at cleaning those up the same way.

Formula/consul-template.rb
cblecker: Fixed.

Formula/consul-template.rb
cblecker: Fixed.

Formula/consul-template.rb
cblecker: Fixed.

Formula/consul-template.rb
cblecker: Fixed.

Formula/consul-template.rb
```diff
@@ -16,50 +31,34 @@ class ConsulTemplate < Formula
...
-    dir.install buildpath.children
-
-    Language::Go.stage_deps resources, buildpath/"src"
-    ENV.prepend_create_path "PATH", buildpath/"bin"
-    cd("src/github.com/mitchellh/gox") { system "go", "install" }
-
-    cd dir do
-      system "make", "dev"
-      system "make", "test"
+    (buildpath/"src/github.com/hashicorp/consul-template").install buildpath.children - [buildpath/".brew_home"]
```
ilovezfs: Let's move this below `ENV["XC_ARCH"] = arch`

Also, when it exceeds the eightieth column as here, you'll want to do

```
    dir = buildpath/"src/github.com/hashicorp/consul-template"
    dir.install buildpath.children - [buildpath/".brew_home"] 

    cd dir do
```

Formula/consul-template.rb
```diff
@@ -16,50 +31,34 @@ class ConsulTemplate < Formula
...
+    arch = MacOS.prefer_64_bit? ? "amd64" : "386"
+    ENV["XC_OS"] = "darwin"
+    ENV["XC_ARCH"] = arch
+
+    cd buildpath/"src/github.com/hashicorp/consul-template" do
+      if build.stable?
+        # This can be removed on next stable version
```
ilovezfs: It's unclear what "This" refers to here.

ilovezfs:
>Thanks for the review @ilovezfs! I'm pushing a new commit revision with changes.

You're welcome!

Formula/consul-template.rb
cblecker: Fixed.

Formula/consul-template.rb
cblecker: Fixed.

ilovezfs (approved):
Thanks for making all those changes!

cblecker:
Build passed. I think we might be ready for a merge :)

ilovezfs:
Merged! Thanks @cblecker

PR #8148, opened 2016-12-23
gitlab-ci-multi-runner 1.9.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #8830, opened 2017-01-14
gitlab-ci-multi-runner 1.9.2

JCount:

-----
also add ARM prebuilt image,
static build time

### DIFF
--- a/Formula/gitlab-ci-multi-runner.rb
+++ b/Formula/gitlab-ci-multi-runner.rb
@@ -4,8 +4,8 @@ class GitlabCiMultiRunner < Formula
   desc "The official GitLab CI runner written in Go"
   homepage "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner"
   url "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner.git",
-      :tag => "v1.9.0",
-      :revision => "82714ae882225d77636f8780197309e59cd20ecf"
+      :tag => "v1.9.2",
+      :revision => "ade6572ddb2d1e75243c31333fb392ac2a745685"
   head "https://gitlab.com/gitlab-org/gitlab-ci-multi-runner.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@tmaczukin mind verifying this is [as you requested](https://github.com/Homebrew/homebrew-core/pull/8148#issuecomment-270194212)?

Formula/gitlab-ci-multi-runner.rb
```diff
@@ -41,21 +48,24 @@ def install
     end
 
     cd dir do
-      resource("prebuilt-x86_64.tar.xz").stage do
-        system "go-bindata", "-pkg", "docker", "-nocompress", "-nomemcopy",
-                             "-nometadata", "-o",
-                             "#{dir}/executors/docker/bindata.go",
-                             "prebuilt-x86_64.tar.xz"
+      ["prebuilt-x86_64.tar.xz", "prebuilt-arm.tar.xz"].each do |r|
```
tmaczukin: I don't know how Homebrew works internally, so maybe my comment is invalid, but from what I see here it looks that the `go-bindata -pkg ...` command will be executed twice: first time with `prebuilt-x86_64.tar.xz` as the argument and second time with `prebuilt-arm.tar.xz` as the argument. If yes, then the second execution will overwrite the `bindata.go` file. And that's not the goal which we want to achieve.

It should be a one execution of this command with both files as the final arguments, so something looking like:
```
go-bindata -pkg docker \
           -nocompress \
           -nomemcopy \
           -nometadata \
           -o executors/docker/bindata.go
           prebuilt-x86_64.tar.xz
           prebuilt-arm.tar.xz
```

With this both files will be available for Docker executor as assets stored internally in `bindata.go` file.

tmaczukin:
@ilovezfs Thanks for mentioning me. Generally it looks good. However I left one comment in a place that worries me. If this will be fixed or confirmed that the current implementation works as it should, then I think it can be merged.

ilovezfs:
@tmaczukin should be fixed now

tmaczukin:
Looks good for me :)

PR #8833, opened 2017-01-14
watchexec 1.6.0 (new formula)

sbdchd:

-----

See https://github.com/mattgreen/watchexec

### DIFF
--- a/Formula/watchexec.rb
+++ b/Formula/watchexec.rb
@@ -1,0 +1,21 @@
+class Watchexec < Formula
+  desc "Execute commands when watched files change"
+  homepage "https://github.com/mattgreen/watchexec"
+  url "https://github.com/mattgreen/watchexec/archive/1.6.0.tar.gz"
+  sha256 "0ce7b26bbc05400a835b5a06e13ef69aac90a6d93ef4b989241258494f3c967f"
+
+  depends_on "rust" => :build
+
+  def install
+    system "cargo", "build", "--release"
+    bin.install "target/release/watchexec"
+    man1.install "doc/watchexec.1"
+  end
+
+  test do
+    o = IO.popen("#{bin}/watchexec -- echo 'saw file change'")
+    sleep 0.1
+    Process.kill("INT", o.pid)
+    assert_match "saw file change", o.read
+  end
+end


Formula/watchexec.rb
```diff
@@ -0,0 +1,18 @@
...
+  depends_on "rust" => :build
+
+  def install
+    system "cargo", "build", "--release"
+    bin.install "target/release/watchexec"
+    man1.install "doc/watchexec.1"
+  end
+
+  test do
+    system "#{bin}/watchexec", "--version"
```
woodruffw: Can we get a more nuanced test than just `--version`?

sbdchd: It is difficult to make a more complete test since `watchexec` works by monitoring for filechange events and then executing a command. It does not exit after being called.

I tried the following.

```diff
diff --git a/Formula/watchexec.rb b/Formula/watchexec.rb
index d6a59e5..ee3c0a0 100644
--- a/Formula/watchexec.rb
+++ b/Formula/watchexec.rb
@@ -13,6 +13,9 @@ class Watchexec < Formula
   end
 
   test do
-    system "#{bin}/watchexec", "--version"
+    o = IO.popen("#{bin}/watchexec -- echo 'saw file change'")
+    sleep 0.1
+    Process.kill("INT", o.pid)
+    assert_equal "saw file change\n", o.read
   end
 end
```

Which works but is not very pleasant. 

*Note:* I don't know ruby.

`watchexec` is pretty similar to [`watchman`][1], which also has a pretty rudimentary test.

[1]: https://github.com/Homebrew/homebrew-core/blob/7dfe465b7cb48be58bedc4349a5357953a53133e/Formula/watchman.rb#L47

woodruffw: Yeah, testing programs that don't exit is always ungraceful. Your new test looks fine to me under the circumstances, although you should use `assert_match` instead of `assert_equal` (it produces nicer output on failure) and remove the `\n`.

sbdchd:
updated

woodruffw:
Thanks! We'll leave this open for a bit to give other maintainers a chance to review, but this looks good to me.

Formula/watchexec.rb
mattgreen: I can add an option (such as `-1`) to support stopping after a single change is detected. This would clean up the test well and mitigate potential issues with the `sleep` call not being long enough if you're unlucky.

It'd be good for integration testing for myself, as well!

Tracking issue: https://github.com/mattgreen/watchexec/issues/35

MikeMcQuaid:

PR #8875, opened 2017-01-14
libming 0.4.7

ilovezfs:
add a test
always depend on giflib
always build the python and perl extensions
remove the php option since it violates the sandbox


woodruffw:
Outdated patch:

```
patching file py_ext/setup.py.in
Hunk #1 FAILED at 19.
1 out of 1 hunk FAILED -- saving rejects to file py_ext/setup.py.in.rej
Error: Failure while executing: /usr/bin/patch -g 0 -f -p1
```

ilovezfs:
PR refreshed.

PR #8909, opened 2017-01-14
libiscsi: fix options and dependencies, add tests

sitsofe:

-----

- Make with-noinst work again.
- Add explicit with-cunit if the user wants iscsi-test-cu . If the
  option is not specified then iscsi-test-cu won't be installed even if
  the cunit package is found and iscsi-test-cu is built.
- Remove popt dependency (libiscsi hasn't required it for years).
- Add basic tests.

### DIFF
--- a/Formula/libiscsi.rb
+++ b/Formula/libiscsi.rb
@@ -12,23 +12,36 @@ class Libiscsi < Formula
     sha256 "c89e40197b9aebd712c67d43fbe7a4e085ddb7a1b58861c6e98444271fd9e383" => :yosemite
   end
 
-  option "with-noinst", "Install the noinst binaries (e.g. iscsi-test-cu)"
+  option "with-noinst", "Install the noinst binaries (examples, tests)"
+  option "with-cunit", "Install iscsi-test-cu, the iSCSI target test suite"
 
   depends_on "autoconf" => :build
   depends_on "automake" => :build
   depends_on "libtool" => :build
-  depends_on "cunit" if build.with? "noinst"
-  depends_on "popt"
+  depends_on "cunit" => :optional
 
   def install
+    if build.without? "cunit"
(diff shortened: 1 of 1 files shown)


Formula/libiscsi.rb
```diff
@@ -12,23 +12,36 @@ class Libiscsi < Formula
...
 
   def install
+    if build.without? "cunit"
+      inreplace "test-tool/Makefile.am", "bin_PROGRAMS =", "noinst_PROGRAMS ="
+    end
     if build.with? "noinst"
       # Install the noinst binaries
-      inreplace "Makefile.am", "noinst_PROGRAMS +=", "bin_PROGRAMS +="
+      inreplace ["tests/Makefile.am", "examples/Makefile.am"], "noinst_PROGRAMS =", "bin_PROGRAMS ="
     end
```
MikeMcQuaid: I think we should just default to the options that don't require `inreplace` patching and remove the others. If upstream wants to provide these as options they should do so as `configure` options.

sitsofe: The problem with just using the defaults is that if you don't happen to have cunit installed you won't get `iscsi-test-cu` which is a large part of using libiscsi for me and thus a use case I don't want to lose. If you look at what others have done (e.g. http://pkgs.fedoraproject.org/cgit/rpms/libiscsi.git/tree/libiscsi.spec ) you will see they always depend on cunit (but if you just need libiscsi for qemu you probably didn't need the cunit dependency..).

Not having the other utilities is less of an issue. If people feel strongly they should never be installed that's fine by me (I only added the noinst hack way-back-when so that I could have iscsi-test-cu installed).

Thoughts?

MikeMcQuaid: Ok, that's fine with me, thanks.

MikeMcQuaid (approved):

MikeMcQuaid:

sitsofe:
@MikeMcQuaid - no YOU rock for taking the time to review and give feedback. Cheers!

PR #8945, opened 2017-01-15
gcc: add gcc@6 alias.

MikeMcQuaid:
This is useful now https://github.com/Homebrew/brew/pull/1847 is merged in order to allow people to not upgrade to gcc 7 when it comes along if they `brew install gcc@6`.

CC @shanematley and @ilovezfs as we talked about this.

### DIFF
--- a/Aliases/gcc@6
+++ b/Aliases/gcc@6
@@ -1,0 +1,1 @@
+../Formula/gcc.rb
--- a/formula_renames.json
+++ b/formula_renames.json
@@ -27,6 +27,7 @@
   "gcc49": "gcc@4.9",
   "gcc5": "gcc@5",
   "gcc6": "gcc",
+  "gcc@6": "gcc",
   "geode": "apache-geode",
   "glfw3": "glfw",
   "gmp4": "gmp@4",


shanematley:
That's great. Thanks @MikeMcQuaid!

ilovezfs:
Is the description actually accurate? If you
```
brew install gcc@6
```
It will be converted to `gcc` and installed as such, and brew will "forget" that it was installed by an aliased name and happily upgrade the user to version 7.

That's different from a formula having
```
depends_on "gcc@6"
```
which will be dynamically converted to whatever gcc@6 is pointing to each time it's installed.

ilovezfs:
Ah I see this is what 1847 is addressing. Hm. So if I understand the consequence of that commit, it means that `ag` will never be able to mean something other that `the_silver_searcher` now. Quite a land grab for aliases.

ilovezfs:
I've opened a PR to revert https://github.com/Homebrew/brew/pull/1847 as I don't think that was such a hot idea.

ilovezfs (approved):

MikeMcQuaid:
@ilovezfs It's on my todo list to try and make this 💥 and fix that case in Homebrew/brew.

PR #8924, opened 2017-01-15
apt-dater 1.0.3

geoffmacintosh:
- [X] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [X] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [X] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [X] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----
There's already a pull request  at #5763, but it links the old website.

PR #8989, opened 2017-01-17
Rollback apt-dater to v0.9.0

MikeLund:

-----

Does not pass audit only because of "stable version should not decrease (from 1.0.3 to 0.9.0)".

Please see https://github.com/Homebrew/homebrew-core/pull/8924#issuecomment-273152828 for reason to rollback. Thanks.

### DIFF
--- a/Formula/apt-dater.rb
+++ b/Formula/apt-dater.rb
@@ -1,8 +1,8 @@
 class AptDater < Formula
   desc "Manage package updates on remote hosts using SSH"
   homepage "https://github.com/DE-IBH/apt-dater"
-  url "https://github.com/DE-IBH/apt-dater/archive/v1.0.3.tar.gz"
-  sha256 "891b15e4dd37c7b35540811bbe444e5f2a8d79b1c04644730b99069eabf1e10f"
+  url "https://github.com/DE-IBH/apt-dater/archive/v0.9.0.tar.gz"
+  sha256 "1c361dd686d66473b27db4af8d241d520535c5d5a33f42a35943bf4e16c13f47"
 
   bottle do
     sha256 "e4062b5a3202e29cd57cf95328348c20022857ec4486056dcc8bf9c10da5e97a" => :sierra
--- a/Formula/apt-dater.rb
+++ b/Formula/apt-dater.rb
@@ -3,6 +3,7 @@ class AptDater < Formula
   homepage "https://github.com/DE-IBH/apt-dater"
   url "https://github.com/DE-IBH/apt-dater/archive/v0.9.0.tar.gz"
   sha256 "1c361dd686d66473b27db4af8d241d520535c5d5a33f42a35943bf4e16c13f47"
+  version_scheme 1
 
   bottle do
     sha256 "e4062b5a3202e29cd57cf95328348c20022857ec4486056dcc8bf9c10da5e97a" => :sierra


mbi:
Agreed, FWIW: 1.0.3 seems toxic at first sight.

MikeMcQuaid:
0.9.0 is also marked as unstable so an earlier version seems like the best bet.

> Does not pass audit only because of "stable version should not decrease (from 1.0.3 to 0.9.0)".

That means it needs a `version_scheme` bump added.

If there's no stable version in over 5 years I think that somewhat implies this software isn't maintained and I'm unsure if we should be packaging it at all.

nijikon:
Done.

MikeMcQuaid:
Reverted in 2bdb07adaf as this needs a `version_scheme`, to be before 0.9.0 and discussion on whether it should be maintained by us at all.

MikeLund:
I use apt-dater v0.9.0 every week with no issues. It serves its purpose and doesn't really need any changes. Apt doesn't really change much either, so I don't see this version breaking in the near-future. Thus I don't support removing it.

How would one use `version_scheme` by the way? Didn't see much description in the docs.

MikeMcQuaid:
Add `version_scheme 1` to the formula.

MikeLund:
@MikeMcQuaid thanks, please see new commit, passes audit now.

MikeMcQuaid:
> 0.9.0 is also marked as unstable so an earlier version seems like the best bet.

To repeat myself more explicitly: given 0.9.0 is also listed as unstable I think it's either worth going on 0.8.6 or leaving the version as-is.

MikeLund:
1. Changes from 0.9.0 from 0.8.6 didn't make any regressions, only added an optional feature (clustering) and fixed some bugs. The addition of clustering is why it gets tagged "unstable". Users can just not use that feature if doesn't work well for them, and us who like it can continue to use it.

2. Under this situation you propose (using 0.8.6), you will break the configs for people who use clustering from v9, and reintroducing bugs fixed in v9. There is no real benefit other than /saying/ the version is more stable.

3. 0.9.0 has been out and "unstable" since 2012. There is no major issue with it.


Stability-flags from an inactive hobby-project shouldn't be taken so strictly, especially when it "works" and has been around for so long. If I am missing the point, please explain it a bit better, because I simply do not understand how it helps any brewer to use a less-stable version of a software simply because it claims to be stable. Thanks much.

MikeMcQuaid:
> Under this situation you propose (using 0.8.6), you will break the configs for people who use clustering from v9, and reintroducing bugs fixed in v9. There is no real benefit other than /saying/ the version is more stable.

> Stability-flags from an inactive hobby-project shouldn't be taken so strictly, especially when it "works" and has been around for so long.

I think both the above arguments could be made for this PR as a whole.

>  I simply do not understand how it helps any brewer to use a less-stable version of a software simply because it claims to be stable.

We either go by what upstreams claim to be stable or the latest released version. It feels strange to me to deliberately be on neither. When do we upgrade? What if there's a security vulnerability?

MikeLund:
Yeah I mean normally I'd agree with that stability logic; no one wants to run unstable stuff without opting in to it. In a case where the project is dormant, I might be inclined for graveyard. But this version is fine. It's a simple userland utility. There is not much risk of security issues, it's mostly just a frontend to automate updating Debian packages.

But I'll stop trying to go around in circles. Merging 0.9.0 LGTM, but if that's unacceptable then using 0.8.6 would still be better than the 1.x mess. So let me know the final decision and I can test it out, thank you.

MikeMcQuaid:
Went for 0.9.0. Thanks for the PR @MikeLund!

PR #8494, opened 2017-01-04
gdk-pixbuf 2.36.3

tschoonj:

PR #8991, opened 2017-01-17
gdk-pixbuf 2.36.4

tschoonj:
Do not merge this one in: it is broken for the same reason as #8494


tschoonj:
@matthiasclasen would you mind having a look at this? I also file a bug on the GNOME bugzilla over this at https://bugzilla.gnome.org/show_bug.cgi?id=776892

Thanks!

ilovezfs:
@tschoonj from https://trac.macports.org/ticket/53334#comment:4 it sounds like the problem is the reverse, namely that it's fixed now, and the accident was the increase, as opposed to being broken now, and the accident being the decrease.

nijikon:
#9940

PR #9001, opened 2017-01-17
Use `keg_only :versioned_formula`.

MikeMcQuaid:
Use `keg_only :versioned_formula` for versioned formulae and do some general version cleanup.

Note, this should not be merged until 1.1.8 has been tagged with this new DSL.


MikeMcQuaid:
Before merging this I think I should add a message to keg-only formulae to tell people how to set their PATH rather than needing to `brew link --force` all the time.

alexdnaz:
what is keg only formulae??????

alyssais:
> Before merging this I think I should add a message to keg-only formulae to tell people how to set their PATH rather than needing to `brew link --force` all the time.

For those following along at home that's Homebrew/brew#1902.

alexdnaz:
brew link --force

alexdnaz:
Please call me

RandomDSdevel:
@alexdnaz:  From [here](http://www.rubydoc.info/github/Homebrew/brew/master/Formula#keg_only-class_method) in [Homebrew's Ruby-Doc documentation](http://www.rubydoc.info/github/Homebrew/brew/master/):  

> `.keg_only(reason, explanation = "") ⇒ Object`
>
> Software that will not be sym-linked into the `brew --prefix` will only live in its Cellar. Other formulae can depend on it and then brew will add the necessary includes and libs (etc.) during the brewing of that other formula. But generally, keg_only formulae are not in your PATH and not seen by compilers if you build your own software outside of Homebrew. This way, we don't shadow software provided by macOS.
>
> `keg_only :provided_by_macos`
>
> `keg_only "because I want it so"`

MikeMcQuaid:
@BrewTestBot test this please

DomT4:
Gentle reminder if people come back to this in future not to apply this to the `lua` formulae which have been designed to coexist.

MikeMcQuaid:
@DomT4 For now the plan is to allow formulae that can coexist to do so. If this causes too much conceptual confusion longer-term we're likely to make more keg-only and provide a better command/interface to handle the linking.

MikeMcQuaid:
Fixed up the conflicts and broken alias here locally and verified the test failures didn't reproduce with local bottles.

ilovezfs:
:tada:

PR #9011, opened 2017-01-17
riak 2.2.0

schrepfler:
The build process audit did throw this warning but the riak binary does return a version correctly so I don't have enough info what's the best way to proceed.

```
==> Installing or updating 'rubocop' gem
Fetching: rubocop-0.45.0.gem (100%)
Successfully installed rubocop-0.45.0
1 gem installed
riak:
    These object files were linked against the deprecated system OpenSSL or
    the system's private LibreSSL.
      /usr/local/Cellar/riak/2.2.0/libexec/lib/crypto-3.1/priv/lib/crypto.so
    Broken dylib links found:
      /Users/buildbot/masters/riak/riak-build-osx-108-64/build/distdir/riak-2.2.0/deps/eleveldb/c_src/system/lib/libsnappy.1.dylib
```

-----

### DIFF
--- a/Formula/riak.rb
+++ b/Formula/riak.rb
@@ -9,7 +9,6 @@ class Riak < Formula
 
   depends_on :macos => :mountain_lion
   depends_on :arch => :x86_64
-  depends_on "openssl"
 
   def install
     logdir = var + "log/riak"
--- a/Formula/riak.rb
+++ b/Formula/riak.rb
@@ -9,6 +9,7 @@ class Riak < Formula
 
   depends_on :macos => :mountain_lion
   depends_on :arch => :x86_64
+  depends_on "openssl"
 
   def install
     logdir = var + "log/riak"
--- a/Formula/riak.rb
+++ b/Formula/riak.rb
@@ -1,9 +1,9 @@
 class Riak < Formula
   desc "Distributed database"
   homepage "https://basho.com/products/#riak"
-  url "https://s3.amazonaws.com/downloads.basho.com/riak/2.1/2.1.4/osx/10.8/riak-2.1.4-OSX-x86_64.tar.gz"
-  version "2.1.4"
-  sha256 "ece75fa1d1ac89525162537ac3cb258c37da5adf0f03b6a04cc49e9f29cbfb8a"
+  url "http://s3.amazonaws.com/downloads.basho.com/riak/2.2/2.2.0/osx/10.8/riak-2.2.0-OSX-x86_64.tar.gz"
+  version "2.2.0"
+  sha256 "51ea63d6efaa3bba4efb0ca13de81da2e2662b6691b4132cf552ca7635c8a857"
 
   bottle :unneeded
 
--- a/Formula/riak.rb
+++ b/Formula/riak.rb
@@ -1,6 +1,6 @@
 class Riak < Formula
   desc "Distributed database"
-  homepage "https://basho.com/products/#riak"
+  homepage "http://basho.com/products/riak-kv/"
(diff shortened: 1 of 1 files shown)


schrepfler:
Declaring explicitly openssl did not help (but I guess it didn't hurt either.) Not sure we can fix the last error though.

MikeMcQuaid:
@schrepfler Can you report this to the upstream project? Thanks!

schrepfler:
As I don't think this recipe is in any sense worse than the previous version, I'd suggest we merge it anyways. The binary distribution is what basho ships so I don't think we can retroactively fix this version. I've sent a DM to a person in the basho team, hopefully it will be picked up in the next release.

schrepfler:
Any thoughts?

MikeMcQuaid:
We've been verbally told this would be fixed for a while. We can't ship this version but we can block this version from shipping until we have at least an upstream bug report. Is that something you could do? Thanks!

schrepfler:
Bug created upstream.

schrepfler:
Can this now be merged as I don't fully understand why would this be a problem. This is using the officially distributed binary which if I understand correctly has a broken symlink to an internal library. However if this is not really breaking functionally the package or pollutes system libraries why would it be blocked from joining the recipes? I've also reported the issue upstream so we can also hope to get it fixed in some future release.

MikeMcQuaid:
> * object files were linked against system openssl
>     These object files were linked against the deprecated system OpenSSL or
>     the system's private LibreSSL.
>     Adding `depends_on "openssl"` to the formula may help.
>       /usr/local/Cellar/riak/2.2.0/libexec/lib/crypto-3.1/priv/lib/crypto.so

@schrepfler Thanks for filing the issue. Can you file an additional issue for the OpenSSL problem? Thanks!

Formula/riak.rb
```diff
@@ -1,14 +1,15 @@
...
+  homepage "http://basho.com/products/riak-kv/"
+  url "http://s3.amazonaws.com/downloads.basho.com/riak/2.2/2.2.0/osx/10.8/riak-2.2.0-OSX-x86_64.tar.gz"
+  version "2.2.0"
+  sha256 "51ea63d6efaa3bba4efb0ca13de81da2e2662b6691b4132cf552ca7635c8a857"
 
   bottle :unneeded
 
   depends_on :macos => :mountain_lion
   depends_on :arch => :x86_64
+  depends_on "openssl"
```
MikeMcQuaid: Please remove this until it's used.

MikeMcQuaid (changes requested):

Formula/riak.rb
schrepfler: @angrycub can we get a bug raised against the basho bug tracker wrt using the system openssl and post it here?

schrepfler: @MikeMcQuaid removing openssl dependency.

angrycub: @mikemcquaid.  https://github.com/basho/riak/issues/888 Is the upstream issue that is tracking the OpenSSL linkage issue.   The broken dylib issue is https://github.com/basho/eleveldb/issues/236.

MikeMcQuaid: @angrycub Thanks!

MikeMcQuaid:
@schrepfler As a user of this formula: could you consider trying to add a test? Even one using `--version` or `--help` or similar would be better than nothing. Thanks!

schrepfler:
Test added.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks for your contribution to Homebrew (and patience)! Without people like you submitting PRs we couldn't run this project. You rock!

schrepfler:
No problem 😀🙏🏻

PR #5948, opened 2016-10-15
Add Redex

kageiit:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

---

PR #9012, opened 2017-01-18
apk-patch-size-estimator 0.2 (new formula)

MichaelEvans:
This adds apk-patch-size-estimator to Homebrew. This tool Estimates the size of Google Play patches and the new gzipped APK.

-----

### DIFF
--- a/Formula/apk-patch-size-estimator.rb
+++ b/Formula/apk-patch-size-estimator.rb
@@ -1,0 +1,19 @@
+class ApkPatchSizeEstimator < Formula
+  desc "Estimates the size of Play patches / gzipped APKs."
+  homepage "https://github.com/googlesamples/apk-patch-size-estimator"
+  url "https://github.com/googlesamples/apk-patch-size-estimator/archive/0.2.tar.gz"
+  sha256 "cbab54effbbbb2abf3ddb58cf004e8ee0e38b520fbc5266f08a5fd6a8cf9c9ff"
+  depends_on "bsdiff"
+
+  def install
+    libexec.install Dir["*"]
+
+    Dir.glob("#{libexec}/*.py") do |script|
+      bin.install_symlink script => File.basename(script, ".py").tr("_", "-")
+    end
+  end
+
+  test do
+    assert_match /^usage: apk-patch-size-estimator/, shell_output("#{bin}/apk-patch-size-estimator --help").strip
+  end
+end


Formula/apk-patch-size-estimator.rb
```diff
@@ -0,0 +1,19 @@
...
+  def install
+    libexec.install Dir["*"]
+
+    Dir.glob("#{libexec}/*.py") do |script|
+      bin.install_symlink script => File.basename(script, ".py").tr("_", "-")
+    end
+  end
+
+  test do
+    assert_match /^usage: apk-patch-size-estimator/, shell_output("#{bin}/apk-patch-size-estimator --help").strip
```
MikeMcQuaid: Any chance of downloading and running this on an actual `apk`?

Formula/apk-patch-size-estimator.rb
```diff
@@ -0,0 +1,19 @@
...
+  desc "Estimates the size of Play patches / gzipped APKs."
+  homepage "https://github.com/googlesamples/apk-patch-size-estimator"
+  url "https://github.com/googlesamples/apk-patch-size-estimator/archive/0.2.tar.gz"
+  sha256 "cbab54effbbbb2abf3ddb58cf004e8ee0e38b520fbc5266f08a5fd6a8cf9c9ff"
+  depends_on "bsdiff"
+
+  def install
+    libexec.install Dir["*"]
+
+    Dir.glob("#{libexec}/*.py") do |script|
```
MikeMcQuaid: `Pathname.glob` will allow you to do `script.basename(".py").to_s.tr("_", "-")` below. Fine either way but may be a little nicer.

Formula/apk-patch-size-estimator.rb
MichaelEvans: I guess that's doable - but I can't really think of a stable place to host an APK / which APK to use. I'd be glad to do it if you can think of one though :)

Formula/apk-patch-size-estimator.rb
MichaelEvans: Just curious - what would be the pros/cons? I would be glad to change it to whatever is preferred though.

MichaelEvans:
@MikeMcQuaid Any update for this?

Formula/apk-patch-size-estimator.rb
MikeMcQuaid: A Google-hosted one feels like a good fit. We can download them in the formula with a `resource`.

MichaelEvans: I can't seem to find any APKs that are hosted by Google anywhere (besides the Play Store). If you know of one, I'd be glad to update though.

MikeMcQuaid: Check out https://github.com/Homebrew/homebrew-core/pull/5948 where they've used one hosted by Facebook

MichaelEvans: Will do! Thanks

Formula/apk-patch-size-estimator.rb
MichaelEvans: So I can't seem to figure out how to do it - but the `lib` folder inside `libexec` needs to be copied into `bin` (and remain as a subdirectory). Is there a way to symlink or something that directory from one to the other?

MikeMcQuaid: I'm not sure I understand. Can you elaborate with specific paths?

MichaelEvans: Sorry for the delay. There's a project setup like this

```
.
│-- lib
│    └-- file-by-file-tools.jar
└-- apk_patch_size_estimator.py
```

Perhaps there's a better way to copy these files in that structure properly.

MikeMcQuaid: How many scripts are there here?

MichaelEvans: Just those two - the .jar and the .py file

JCount:
Closing per inactivity. Thank you for your work nonetheless @MichaelEvans. If you end up deciding to work on this again, feel free to open a new PR.

PR #9023, opened 2017-01-18
carrot2 3.15.0 (new formula)

ankane:

-----

Hey, thanks again for all the great work on Homebrew 💯  This adds Carrot2 to the collection.

- Andrew

### DIFF
--- a/Formula/carrot2.rb
+++ b/Formula/carrot2.rb
@@ -44,7 +44,7 @@ class Carrot2 < Formula
     inreplace testpath/"bin/carrot2", "cd #{libexec}", "cd #{testpath}/libexec"
     begin
       pid = fork { exec testpath/"bin/carrot2" }
-      sleep 10
+      sleep 5
       output = shell_output("curl -s -F dcs.c2stream=@#{libexec}/examples/shared/data-mining.xml http://localhost:8080/dcs/rest")
       assert_match /data mining/m, output
     ensure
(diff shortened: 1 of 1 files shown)


Formula/carrot2.rb
```diff
@@ -0,0 +1,45 @@
...
+        <array>
+          <string>#{opt_bin}/carrot2</string>
+        </array>
+      </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    system bin/"carrot2", "--help"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

ankane:
Thanks for the quick response @MikeMcQuaid. I added a better test. Needed to copy everything into `testpath` due to sandbox since Carrot2 really wants to write to `libexec/logs`.

woodruffw:
El Cap failure; looks like a race condition with the `sleep`, so perhaps make it a bit longer?

ankane:
Thanks @woodruffw. I reread the docs and it looks like the latest version depends on Java 1.8+. When I bumped that, it fixed it for El Capitan.

woodruffw:
Thanks! We can bump that sleep back down, then - sorry for leading you in the wrong direction.

ankane:
Restored. No need to apologize - I didn't realize the tests were failing. I really appreciate all the work you and the other Homebrew members put it.

woodruffw:
Thanks for contributing, @ankane!

PR #9024, opened 2017-01-18
mesos: Disable optimized build to not trigger libcxx UB.

### LINKED ISSUE #8707: Segfault with Mesos 1.1.0 formula when using the Docker image provider
- [x] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

Currently when you run `mesos-slave --master=127.0.0.1:5050 --work_dir=/tmp/slave --containerizers=mesos --image_providers=docker` you will immediately get a segfault:
```
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0110 13:32:58.590873 2100015104 main.cpp:243] Build: 2016-11-16 18:56:50 by brew
I0110 13:32:58.591126 2100015104 main.cpp:244] Version: 1.1.0
I0110 13:32:58.593508 2100015104 containerizer.cpp:200] Using isolation: posix/cpu,posix/mem,filesystem/posix
*** Aborted at 1484073178 (unix time) try "date -d @1484073178" if you are using GNU date ***
PC: @        0x10a29a9ca std::__1::__tree<>::__assign_multi<>()
*** SIGSEGV (@0x0) received by PID 47913 (TID 0x7fff7d2bb000) stack trace: ***
    @     0x7fff899b252a _sigtramp
    @     0x7fdb13029530 (unknown)
    @        0x10a29a96c flags::FlagsBase::operator=()
    @        0x10a299924 mesos::uri::fetcher::Flags::operator=()
    @        0x10a298902 mesos::uri::fetcher::create()
    @        0x10a284fd3 mesos::internal::slave::docker::Store::create()
    @        0x10a24841f mesos::internal::slave::Store::create()
    @        0x10a22f25a mesos::internal::slave::Provisioner::create()
    @        0x10a1a2c6f mesos::internal::slave::MesosContainerizer::create()
    @        0x10a141426 mesos::internal::slave::Containerizer::create()
    @        0x109cb8d80 main
    @     0x7fff9a3995ad start
Segmentation fault: 11
```

This seems to be caused from an issue with libc++ with clang++ 3.8. There's an open issue about this here https://issues.apache.org/jira/browse/MESOS-5745
If you look at a few of the comments further down you'll see that it should be fixed with clang 3.9.

Can Mesos be built with the newest version of clang? If not, can gcc be used instead? I wouldn't mind contributing any changes that are needed.

What is everyone's thought on this?

bbannier:
This fixes #8707.

-----

### DIFF
--- a/Formula/mesos.rb
+++ b/Formula/mesos.rb
@@ -53,6 +53,16 @@ class Mesos < Formula
   def install
     ENV.java_cache
 
+    # Disable optimizing as libc++ does not play well with optimized clang
+    # builds (see https://llvm.org/bugs/show_bug.cgi?id=28469 and
+    # https://issues.apache.org/jira/browse/MESOS-5745).
+    #
+    # NOTE: We cannot use `--disable-optimize` since we also pass e.g.,
+    # CXXFLAGS via environment variables. Since compiler flags are passed via
+    # environment variables the Mesos build will silently ignore flags like
+    # `--[disable|enable]-optimize`.
+    ENV.O0 unless DevelopmentTools.clang_build_version >= 900
+
     # work around to avoid `_clock_gettime` symbol not found error.
     if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
       ENV["ac_have_clock_syscall"] = "no"
@@ -138,6 +148,23 @@ class Mesos < Formula
   test do
     require "timeout"
 
+    # Make sure we are not affected by MESOS-6910 and related issues.
+    agent = fork do
+      exec "/Users/bbannier/src/homebrew/sbin/mesos-agent",
+          "--master=127.0.0.1:5050",
+          "--work_dir=/tmp/mesos.slave.brew",
+          "--image_providers=docker"
+    end
+    begin
+      Timeout.timeout(2) do
+        Process.wait agent
+      end
+    rescue Timeout::Error
+      Process.kill "TERM", agent
+    end
+    assert $?.exitstatus, "agent process died, check MESOS-6606-related behavior"
+
+    # Make tests for minimal functionality.
     master = fork do
       exec "#{sbin}/mesos-master", "--ip=127.0.0.1",
                                    "--registry=in_memory"


Formula/mesos.rb
```diff
@@ -53,6 +53,16 @@ class Mesos < Formula
...
 
+    # Disable optimizing as libc++ does not play well with optimized clang
+    # builds (see https://llvm.org/bugs/show_bug.cgi?id=28469 and
+    # https://issues.apache.org/jira/browse/MESOS-5745).
+    #
+    # NOTE: We cannot use `--disable-optimize` since we also pass e.g.,
+    # CXXFLAGS via environment variables. Since compiler flags are passed via
+    # environment variables the Mesos build will silently ignore flags like
+    # `--[disable|enable]-optimize`.
+    ENV.O0
```
MikeMcQuaid: May as well pass `--disable-optimize` too. Also, you can guard this with something like `unless DevelopmentTools.clang_build_version >= 900`

bbannier: I tested this with the LLVM upstream reproducer. The test case linked there fails for any of `-Os`, `-O1`, or `-O2`, so I suspect the optimization pass exploiting the UB we are hitting is active in each of these optimization levels. The reproducer passes with `-O0`.

ilovezfs: Apparently it was fixed in 3.9: https://llvm.org/bugs/show_bug.cgi?id=19302

Maybe we should use our llvm's clang here since it's 3.9.1?

MikeMcQuaid: @ilovezfs I'd rather we used default clang without optimisations over relying on LLVM clang.

ilovezfs: O0 is pretty grotesque to have in core. It's confined to clisp, ohcount, scheme48, stk, and twemcache right now. I guess the illustrious club could get a new member, but I'd almost rather just give it the boot than build an awful version.

MikeMcQuaid: If it's temporary until (and guarded on) a new Clang I'm fine with it and seems preferable to another dep.

ilovezfs: Does it build properly with gcc?

scheme48 does this
```
ENV.O0 if ENV.compiler == :clang
```

bbannier: @ilovezfs: I am not able to build this with either gcc-5 or gcc-6 (fails in bundled 3rdparty tools) so I believe making this conditional on clang is not that useful in the long run. Also, the real issue is with the used stdlib, not with the compiler, and theoretically users might build using clang with libstdc++, or with gcc and libcxx (not sure if that's a practical concern for Homebrew).

@MikeMcQuaid: Re:guarding on the clang version: AFAIK Apple assigns its clang version numbers kind of arbitrary, and there is e.g., no guarantee that clang-3.9 would be the basis of Apple clang-9.0.0 (I think it is not yet released).

I guess we should try to fix https://issues.apache.org/jira/browse/MESOS-6606 upstream (e.g., to gracefully degrade on setups with this broken libcxx); after that we should be able to remove the non-standard configuration here.

ilovezfs: >I guess we should try to fix https://issues.apache.org/jira/browse/MESOS-6606 upstream (e.g., to gracefully degrade on setups with this broken libcxx); after that we should be able to remove the non-standard configuration here.

That would be lovely.

bbannier: @MikeMcQuaid: I have updated the patch to include a `TODO` to make this conditional once we can be certain the clang-9.0.0 and later fix this libcxx issue.

@ilovezfs: re:degrade gracefully in upstream, there seems to be little support for this since it would e.g., lead to automatic builds passing, but not using explicitly passed flags (e.g., `-O2` would be override with `-O0` in some cases). This is an issue for e.g., automated builds. I would like to suggest keeping an unconditional `-O0` here, and making it unconditional once clang-9.0.0 and the associated libcxx is released. Alternatively we could check here in Homebrew if a build with implicit `-O2` configures, and only drop to `-O0` if it does not -- this looks too complicated for my taste.

bbannier:
FYI, https://issues.apache.org/jira/browse/MESOS-6606 disabling building with optimizations with a broken libcxx has been merged upstream in all current releases. It performs a configure check which explicitly checks if with the picked tools and settings the known libcxx would be triggered.

@MikeMcQuaid @ilovezfs, what is holding this PR back from getting merged?

MikeMcQuaid:
@ilovezfs @nijikon Do either of you object to this being merged?

ilovezfs:
If this change has been merged upstream, can we not apply the commit now as a patch?

bbannier:
@ilovezfs: The upstream patch breaks the build during the configure phase when a broken libcxx is identified. On the Homebrew side, one would still need to disable optimizations on such systems for the build to succeed (if one wants to build with optimizations by default); the upstream build does not implicitly fix flags, but just rejects broken setups.

The last time we talked, we were arguing whether optimizations should be disabled completely or conditionally on the used clang version in the hope that the unreleased clang-900 fixes that issue (@MikeMcQuaid seemed to have preferred making this conditional). I was arguing that one should disable optimizations unconditionally since it is unknown whether clang-900 would really fix this issue.

ilovezfs:
@bbannier thanks for explaining. Is there any way to test in the `test do` for a broken build? That would make the `>= 900` OK I think.

bbannier:
@ilovezfs: Well, we could make disabling optimizations conditional and let Mesos' `configure` do the checking. If then clang-900 does not fix the issue, any attempt to build Mesos with such a clang+libcxx would blow up (fail during Mesos' `configure`). That would prevent the issue from slipping in unseen.

ilovezfs:
@bbannier right but that assumes that the same optimizations flags will be used during configure as during the build, which is relying on implementation details of superenv.

bbannier:
@ilovezfs: Is that something to expect? From the point of view of the software being built, if the build setup requests one set of flags at `./configure` time, but overrides that with incompatible other flags at `make` time, I am not sure there's much the software build setup could do. I personally wouldn't try to handle that.

Any idea how to guard against that reliably? There would be an option to scan Mesos' `./configure` output in order to detect if it would be safe to enable optimizations at `make` time, but then again, why wouldn't Homebrew decide on one set of flags?

ilovezfs:
As I said, you can use an appropriate test in `test do`.

ilovezfs (changes requested):
Please guard with `unless DevelopmentTools.clang_build_version >= 900` and add a test for a broken build in `test do`

bbannier:
I pushed an update, @ilovezfs.

ilovezfs (approved):

ilovezfs (approved):
LGTM when :green_apple:

MikeMcQuaid:

PR #9037, opened 2017-01-18
mingw-w64 Cross Compiler (binutils, gcc, and headers)

stephengroat:
A complete cross compiler solution for mingw-w64. Defaultly compiles
both i686 and x86_64 compilers (x86_64 currently has multilib disabled).
Can be split out into binutils and gcc, but I think that it makes more
sense to gather the complete cross compiler solution into one
Formula. Also has a test for compiler but currently cannot test in wine.

Related to #8697

-----

### DIFF
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -45,10 +45,10 @@ class MingwW64 < Formula
           system "make"
           system "make", "install"
         end
-        rm_rf( %W[
-          #{prefix}/#{name}/share/info
-          #{prefix}/#{name}/share/locale
-        ])
+        rm_rf(%W[
+                #{prefix}/#{name}/share/info
+                #{prefix}/#{name}/share/locale
+              ])
       end
       ENV.prepend_path "PATH", "#{prefix}/#{name}/bin"
 
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -178,14 +178,14 @@ class MingwW64 < Formula
       EOS
     compiler = { ".c" => "gcc", ".cc" => "g++", ".f90" => "gfortran" }.freeze
     target_archs.keys.each do |target_arch|
-      Dir["hello*{#{compiler.keys.join(',')}}"].each do |src|
+      Dir["hello*{#{compiler.keys.join(",")}}"].each do |src|
         exe = "#{target_arch}-#{src.tr(".", "-")}.exe"
         system "#{bin}/#{target_arch}-#{compiler[File.extname(src)]}", "-o", exe, src
         assert_match "file format pei-#{target_archs[target_arch]}", shell_output("#{bin}/#{target_arch}-objdump -a #{exe}")
         if target_arch.start_with?("x86_64")
           assert_match "PE32+ executable (console) x86-64, for MS Windows", shell_output("file #{exe}")
         elsif target_arch.start_with?("i686")
-          assert_match "PE32 executable (console) Intel 80386, for MS Windows", shell_output("file #{exe}")
+          assert_match "PE32 executable for MS Windows (console) Intel 80386 32-bit", shell_output("file #{exe}")
         end
       end
     end
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -182,10 +182,6 @@ class MingwW64 < Formula
         exe = "#{target_arch}-#{src.tr(".", "-")}.exe"
         system "#{bin}/#{target_arch}-#{compiler[File.extname(src)]}", "-o", exe, src
         assert_match "file format pei-#{target_archs[target_arch]}", shell_output("#{bin}/#{target_arch}-objdump -a #{exe}")
-        if target_arch.start_with?("x86_64")
-          assert_match "PE32+ executable (console) x86-64, for MS Windows", shell_output("file #{exe}")
-        elsif target_arch.start_with?("i686")
-          assert_match "PE32 executable for MS Windows (console) Intel 80386 32-bit", shell_output("file #{exe}")
         end
       end
     end
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -182,7 +182,6 @@ class MingwW64 < Formula
         exe = "#{target_arch}-#{src.tr(".", "-")}.exe"
         system "#{bin}/#{target_arch}-#{compiler[File.extname(src)]}", "-o", exe, src
         assert_match "file format pei-#{target_archs[target_arch]}", shell_output("#{bin}/#{target_arch}-objdump -a #{exe}")
-        end
       end
     end
   end
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -53,6 +53,7 @@ class MingwW64 < Formula
       args = %W[
         --host=#{target_arch}
         --prefix=#{prefix}/#{target_arch}
+        --libdir=#{lib}/#{name}
       ]
       args << "--disable-multilib" if target_arch.start_with?("i686") || (target_arch.start_with?("x86_64") && build.without?("multilib"))
 
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -32,8 +32,8 @@ class MingwW64 < Formula
       resource("binutils").stage do
         args = %W[
           --target=#{target_arch}
-          --prefix=#{prefix}/mingw-w64
-          --with-sysroot=#{prefix}/mingw-w64
+          --prefix=#{prefix}/#{name}
+          --with-sysroot=#{prefix}/#{name}
         ]
         if target_arch.start_with?("i686") || (target_arch.start_with?("x86_64") && build.without?("multilib"))
           args << "--enable-targets=#{target_arch}" << "--disable-multilib"
@@ -45,14 +45,16 @@ class MingwW64 < Formula
           system "make"
           system "make", "install"
         end
-        info.rmtree
-        (share/"locale").rmtree
+        rm_rf( %W[
+          #{prefix}/#{name}/share/info
+          #{prefix}/#{name}/share/locale
+        ])
       end
-      ENV.prepend_path "PATH", "#{prefix}/mingw-w64/bin"
+      ENV.prepend_path "PATH", "#{prefix}/#{name}/bin"
 
       args = %W[
         --host=#{target_arch}
-        --prefix=#{prefix}/mingw-w64/#{target_arch}
+        --prefix=#{prefix}/#{name}/#{target_arch}
       ]
       args << "--disable-multilib" if target_arch.start_with?("i686") || (target_arch.start_with?("x86_64") && build.without?("multilib"))
 
@@ -62,19 +64,19 @@ class MingwW64 < Formula
         system "make", "install"
       end
 
-      ln_s "#{prefix}/mingw-w64/#{target_arch}", "#{prefix}/mingw-w64/mingw"
+      ln_s "#{prefix}/#{name}/#{target_arch}", "#{prefix}/#{name}/mingw"
 
       resource("gcc").stage buildpath/"gcc"
 
       gcc_args = %W[
         --target=#{target_arch}
-        --prefix=#{prefix}/mingw-w64
-        --with-sysroot=#{prefix}/mingw-w64
+        --prefix=#{prefix}/#{name}
+        --with-sysroot=#{prefix}/#{name}
         --enable-version-specific-runtime-libs
         --with-bugurl=https://github.com/Homebrew/homebrew-core/issues
         --enable-languages=c,c++,fortran
-        --with-ld=#{prefix}/mingw-w64/bin/#{target_arch}-ld
-        --with-as=#{prefix}/mingw-w64/bin/#{target_arch}-as
+        --with-ld=#{prefix}/#{name}/bin/#{target_arch}-ld
+        --with-as=#{prefix}/#{name}/bin/#{target_arch}-as
         --with-gmp=#{Formula["gmp"].opt_prefix}
         --with-mpfr=#{Formula["mpfr"].opt_prefix}
         --with-mpc=#{Formula["libmpc"].opt_prefix}
@@ -100,8 +102,8 @@ class MingwW64 < Formula
         CPP=#{target_arch}-cpp
         LD=#{target_arch}-gcc
         --host=#{target_arch}
-        --prefix=#{prefix}/mingw-w64/#{target_arch}
-        --with-sysroot=#{prefix}/mingw-w64/#{target_arch}
+        --prefix=#{prefix}/#{name}/#{target_arch}
+        --with-sysroot=#{prefix}/#{name}/#{target_arch}
       ]
 
       if target_arch.start_with?("i686")
@@ -123,7 +125,7 @@ class MingwW64 < Formula
         system "make", "install"
       end
 
-      ln_s "../../lib/gcc/#{target_arch}/lib/libgcc_s.a", "#{prefix}/mingw-w64/#{target_arch}/lib"
+      ln_s "../../lib/gcc/#{target_arch}/lib/libgcc_s.a", "#{prefix}/#{name}/#{target_arch}/lib"
 
       ENV["LDPATH"] = "#{target_arch}/#{lib}"
       args = %W[
@@ -131,7 +133,7 @@ class MingwW64 < Formula
         CXX=#{target_arch}-g++
         CPP=#{target_arch}-cpp
         --host=#{target_arch}
-        --prefix=#{prefix}/mingw-w64/#{target_arch}
+        --prefix=#{prefix}/#{name}/#{target_arch}
       ]
 
       mkdir "mingw-w64-libraries/winpthreads/build-#{target_arch}" do
@@ -178,8 +180,8 @@ class MingwW64 < Formula
     target_archs.keys.each do |target_arch|
       Dir["hello*{#{compiler.keys.join(",")}}"].each do |src|
         exe = "#{target_arch}-#{src.tr(".", "-")}.exe"
-        system "#{prefix}/mingw-w64/bin/#{target_arch}-#{compiler[File.extname(src)]}", "-o", exe, src
-        assert_match "file format pei-#{target_archs[target_arch]}", shell_output("#{prefix}//mingw-w64/bin/#{target_arch}-objdump -a #{exe}")
+        system "#{prefix}/#{name}/bin/#{target_arch}-#{compiler[File.extname(src)]}", "-o", exe, src
+        assert_match "file format pei-#{target_archs[target_arch]}", shell_output("#{prefix}/#{name}/bin/#{target_arch}-objdump -a #{exe}")
       end
     end
   end
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -46,6 +46,7 @@ class MingwW64 < Formula
           system "make"
           system "make", "install"
         end
+        # remove share/locale from conflicting with other binutils installs
         rm_rf("#{share}/locale")
       end
       ENV.prepend_path "PATH", bin.to_s
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -104,6 +104,7 @@ class MingwW64 < Formula
         --host=#{target_arch}
         --prefix=#{prefix}/#{target_arch}
         --with-sysroot=#{prefix}/#{target_arch}
+        --libdir=#{lib}/#{name}
       ]
 
       if target_arch.start_with?("i686")
@@ -134,6 +135,7 @@ class MingwW64 < Formula
         CPP=#{target_arch}-cpp
         --host=#{target_arch}
         --prefix=#{prefix}/#{target_arch}
+        --libdir=#{lib}/#{name}
       ]
 
       mkdir "mingw-w64-libraries/winpthreads/build-#{target_arch}" do
--- a/Formula/mingw-w64.rb
+++ b/Formula/mingw-w64.rb
@@ -5,7 +5,7 @@ class MingwW64 < Formula
   url "https://downloads.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v5.0.1.tar.bz2"
   sha256 "9bb5cd7df78817377841a63555e73596dc0af4acbb71b09bd48de7cf24aeadd2"
 
-  # option "with-multilib", "Compile x86_64 compiler with multilib support"
+  # option "without-multilib", "Compile x86_64 compiler with multilib support"
 
   depends_on "texinfo" => :build
   depends_on "mpfr"
@@ -32,8 +32,9 @@ class MingwW64 < Formula
       resource("binutils").stage do
         args = %W[
           --target=#{target_arch}
-          --prefix=#{prefix}/#{name}
-          --with-sysroot=#{prefix}/#{name}
+          --prefix=#{prefix}
+          --with-sysroot=#{prefix}
+          --datarootdir=#{share}/#{name}
         ]
         if target_arch.start_with?("i686") || (target_arch.start_with?("x86_64") && build.without?("multilib"))
           args << "--enable-targets=#{target_arch}" << "--disable-multilib"
@@ -45,16 +46,13 @@ class MingwW64 < Formula
           system "make"
           system "make", "install"
         end
-        rm_rf(%W[
-                #{prefix}/#{name}/share/info
-                #{prefix}/#{name}/share/locale
-              ])
+        rm_rf("#{share}/locale")
       end
-      ENV.prepend_path "PATH", "#{prefix}/#{name}/bin"
+      ENV.prepend_path "PATH", bin.to_s
 
       args = %W[
         --host=#{target_arch}
-        --prefix=#{prefix}/#{name}/#{target_arch}
+        --prefix=#{prefix}/#{target_arch}
       ]
       args << "--disable-multilib" if target_arch.start_with?("i686") || (target_arch.start_with?("x86_64") && build.without?("multilib"))
 
@@ -64,19 +62,21 @@ class MingwW64 < Formula
         system "make", "install"
       end
 
-      ln_s "#{prefix}/#{name}/#{target_arch}", "#{prefix}/#{name}/mingw"
+      ln_s "#{prefix}/#{target_arch}", "#{prefix}/mingw"
 
       resource("gcc").stage buildpath/"gcc"
 
       gcc_args = %W[
         --target=#{target_arch}
-        --prefix=#{prefix}/#{name}
-        --with-sysroot=#{prefix}/#{name}
+        --prefix=#{prefix}
+        --with-sysroot=#{prefix}
+        --libdir=#{lib}/#{name}
+        --datarootdir=#{share}/#{name}
         --enable-version-specific-runtime-libs
         --with-bugurl=https://github.com/Homebrew/homebrew-core/issues
         --enable-languages=c,c++,fortran
-        --with-ld=#{prefix}/#{name}/bin/#{target_arch}-ld
-        --with-as=#{prefix}/#{name}/bin/#{target_arch}-as
+        --with-ld=#{bin}/#{target_arch}-ld
+        --with-as=#{bin}/#{target_arch}-as
         --with-gmp=#{Formula["gmp"].opt_prefix}
         --with-mpfr=#{Formula["mpfr"].opt_prefix}
         --with-mpc=#{Formula["libmpc"].opt_prefix}
@@ -102,8 +102,8 @@ class MingwW64 < Formula
         CPP=#{target_arch}-cpp
         LD=#{target_arch}-gcc
         --host=#{target_arch}
-        --prefix=#{prefix}/#{name}/#{target_arch}
-        --with-sysroot=#{prefix}/#{name}/#{target_arch}
+        --prefix=#{prefix}/#{target_arch}
+        --with-sysroot=#{prefix}/#{target_arch}
       ]
 
       if target_arch.start_with?("i686")
@@ -125,7 +125,7 @@ class MingwW64 < Formula
         system "make", "install"
       end
 
-      ln_s "../../lib/gcc/#{target_arch}/lib/libgcc_s.a", "#{prefix}/#{name}/#{target_arch}/lib"
+      ln_s "../../lib/gcc/#{target_arch}/lib/libgcc_s.a", "#{prefix}/#{target_arch}/lib"
 
       ENV["LDPATH"] = "#{target_arch}/#{lib}"
       args = %W[
@@ -133,7 +133,7 @@ class MingwW64 < Formula
         CXX=#{target_arch}-g++
         CPP=#{target_arch}-cpp
         --host=#{target_arch}
-        --prefix=#{prefix}/#{name}/#{target_arch}
+        --prefix=#{prefix}/#{target_arch}
       ]
 
       mkdir "mingw-w64-libraries/winpthreads/build-#{target_arch}" do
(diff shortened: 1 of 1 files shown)


Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+  homepage "https://mingw-w64.org/"
+  # https://sourceforge.net/p/mingw-w64/wiki2/Cross%20Win32%20and%20Win64%20compiler/
+  url "https://downloads.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v4.0.6.tar.bz2"
+  sha256 "0c407394b0d8635553f4fbca674cdfe446aac223e90b4010603d863e4bdd015c"
+
+  option "without-i686", "Compile without i686 compiler"
+  option "without-x86_64", "Compile without x86_64 compiler"
+  # option "without-multilib", "Compile x86_64 copmiler without multilib support"
+
+  depends_on "gcc" => :build
```
fxcoudert: As far as I can tell, all these build well with the system compiler…

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+  depends_on "texinfo" => :build
+  depends_on "mpfr"
+  depends_on "libmpc"
+  depends_on "cloog"
+  depends_on "wine" => :optional
```
fxcoudert: Is the wine dependency for testing only? If so, maybe we can check the produced files in ways other than running them (objdump, etc.)

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+
+  option "without-i686", "Compile without i686 compiler"
+  option "without-x86_64", "Compile without x86_64 compiler"
+  # option "without-multilib", "Compile x86_64 copmiler without multilib support"
+
+  depends_on "gcc" => :build
+  depends_on "texinfo" => :build
+  depends_on "mpfr"
+  depends_on "libmpc"
+  depends_on "cloog"
```
fxcoudert: GCC is not using cloog anymore, since version 5

stephengroat: thanks, i'll remove it and try

Formula/mingw-w64.rb
stephengroat: i thought about that, but a true "test" would be running, not just objdump

fxcoudert:
For what it's worth, I am working on a similar formula, the working version is here: https://gist.github.com/fxcoudert/9717a66e34ecde14f145e688053bff97

For some reason, with my setup the multilib compiler does not accept to build (issue reported there: https://sourceforge.net/p/mingw-w64/bugs/580/). Other than that, the main differences I see are:
 - I have split binutils into their own recipe
 - you do not need to configure and build gcc twice, you can simply build `all-gcc` before the mingw-w64 CRT, and then go back to build `all` in the same GCC build directory as before

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+      elsif target_arch.start_with?("i686")
+        args.push("--enable-lib32", "--disable-lib64")
+      end
+      # elsif target_arch.start_with?("x86_64") and build.with?("multilib")
+      #  args.push("--enable-lib32", "--enable-lib64")
+
+      chdir "mingw-w64-crt" do
+        mkdir "build-#{target_arch}" do
+          system "../configure", *args
+          system "make", "-j1"
```
fxcoudert: mingw-w64-crt does build in parallel without problem

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+          --enable-checking=release
+          --enable-lto
+          --enable-libssp
+          --enable-libcc1
+          --enable-shared
+          --with-dwarf
+          --enable-fully-dynamic-string
+          --with-bugurl=https://github.com/Homebrew/homebrew-core/issues
+          --enable-threads=win32
+          --enable-languages=c,c++,objc,obj-c++
```
fxcoudert: Not sure if objc and obj-c++ are heavily used on Windows, but please add Fortran to the list :)

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+          --with-gnu-as
+          --with-gnu-ld
```
fxcoudert: Configure will automatically detect that it has GNU as and GNU ld

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+          args.insert(7, "--disable-multilib")
+        end
+        # elsif target_arch.start_with?("x86_64") and build.with?("multilib")
+        #  args.insert(7, "--enable-64bit")
+        #  args.insert(7, "--enable-targets=x86_64-w64-mingw32,i686-w64-mingw32")
+        #  args.insert(7, "--enable-multilib")
+        mkdir "build-#{target_arch}" do
+          system "../configure", *args
+          system "make", "all-target-libgcc"
+          system "make", "install-target-libgcc"
```
fxcoudert: I do not think all-target-libgcc needs to be done first. A simple "make" then "make install" should work here.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+      end
+
+      resource("gcc").stage do
+        args = %W[
+          CC=gcc-6
+          CXX=g++-6
+          CPP=cpp-6
+          LD=gcc-6
+          --build=#{arch}-apple-darwin#{osmajor}
+          --host=#{arch}-apple-darwin#{osmajor}
```
fxcoudert: host defaults to build, and build is automatically detected. I think only `--target` needs to be specified.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,303 @@
...
+          --target=#{target_arch}
+          --prefix=#{prefix}
+          --with-sysroot=#{prefix}
+          --disable-nls
+          --with-system-zlib
+          --enable-version-specific-runtime-libs
+          --enable-libstdcxx-time=yes
+          --enable-stage1-checking
```
fxcoudert: Because the cross-compiler is not bootstrapped, this probably has no effect

stephengroat:
also want to include @cosmo0920, who also has done work in https://github.com/cosmo0920/homebrew-mingw_w64

stephengroat:
I think separating out `binutils` and `gcc` is probably a good idea. i've added #9047 to update `binutils` to have options for i686, x86_64, and multilib

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,225 @@
...
+    (testpath/"hello.f90").write <<-EOS.undent
+    program hello
+    print *, "Hello, world!"
+    end program hello
+    EOS
+    target_archs.each do |target_arch|
+      system "#{bin}/#{target_arch}-gcc", "-o", "hello-c.exe", "hello.c"
+      # TODO: add wine for test and figure how to run, currently
+      # runs on i686
+      # assert_equal "Hello, world!\r\n", `wine hello-c.exe`
```
cosmo0920: How about test this with the following code?

```ruby
if build.with?("wine")
  assert_equal "Hello, world!\r\n", shell_output(wine hello-c.exe)
end
```

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,225 @@
...
+  # option "without-multilib", "Compile x86_64 copmiler without multilib support"
+
+  depends_on "texinfo" => :build
+  depends_on "mpfr"
+  depends_on "libmpc"
+  depends_on "isl"
+  depends_on "gmp"
+  depends_on "wine" => :optional
+
+  resource "binutils" do
```
cosmo0920: We should depends on `mingw-w64-binutils` instead of monolithic mingw-w64 Formula.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,225 @@
...
+      odie "Options --without-i686 and --without-x86_64 are mutually exclusive"
+    end
+    archs = []
+    archs.push("i686-w64-mingw32") if build.with?("i686")
+    archs.push("x86_64-w64-mingw32") if build.with?("x86_64")
+  end
+
+  def install
+    target_archs.each do |target_arch|
+      resource("binutils").stage do
```
cosmo0920: If we decide to depends on `mingw-w64-bintulis` in this formula, we can remove `resource("binutils")` block from here.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,219 @@
+class MingwW64 < Formula
+  desc "Minimalist GNU for Windows with GCC."
+  homepage "https://mingw-w64.org/"
+  # https://sourceforge.net/p/mingw-w64/wiki2/Cross%20Win32%20and%20Win64%20compiler/
+  url "https://downloads.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v5.0.1.tar.bz2"
+  sha256 "9bb5cd7df78817377841a63555e73596dc0af4acbb71b09bd48de7cf24aeadd2"
+
+  option "without-i686", "Compile without i686 compiler"
+  option "without-x86_64", "Compile without x86_64 compiler"
+  # option "without-multilib", "Compile x86_64 copmiler without multilib support"
```
MikeMcQuaid: Remove these options.

MikeMcQuaid (changes requested):

Formula/mingw-w64.rb
stephengroat: will do, i just want to get `i686` and `x86_64` binutils in, currently being discussed in #9047

Formula/mingw-w64.rb
stephengroat: which ones? the `i686` and `x86_64`? the multilib `x86_64` build currently does not work (broke sometime in version 4), so having a separate `i686` build is currently the only way to get 32 bit binaries compiled

MikeMcQuaid: What do you do if you need to build 32-bit and 64-bit binaries?

stephengroat: Currently, compile both `x86_64` toolchain and `i686` toolchain. I made the option to disable either one in case someone is building from source, since each toolchain build is time consuming

Formula/mingw-w64.rb
```diff
@@ -123,6 +123,7 @@ def install
       mkdir "#{buildpath}/gcc/build-#{target_arch}" do
         system "../configure", *gcc_args
         system "make", "all-gcc"
+        system "make", "check-gcc"
```
fxcoudert: Please don't `make check-gcc` as part of the build, it runs the full testsuite and can take several hours

stephengroat:
it only took an extra 10 minutes on my box. i'll post how long the travis
run took, if it's too long i'll take it out.
it's a much better test than anything someone can write in `test` and is
recommend for binary distributions

Formula/mingw-w64.rb
stephengroat: only took 37 minutes for the full build

fxcoudert:
The GCC testsuite regularly has several failures (especially so for mingw!), even for released versions. And because make check fails silently (i.e. testsuite failures are reported on stdout but ignored), I do not see how running "make check" is useful, unless we do something with the output (parse the number of failures and compare it to a threshold?).

MikeMcQuaid:
> Another issue is this: the GCC testsuite regularly has several failures (especially so for mingw!), even for released versions. And because make check fails silently (i.e. testsuite failures are reported on stdout but ignored), I do not see how running "make check" is useful, unless we do something with the output (parse the number of failures and compare it to a threshold?).

I agree.

Formula/mingw-w64.rb
MikeMcQuaid: Let's remove all the options, thanks.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,220 @@
...
+    }
+    EOS
+    (testpath/"hello.f90").write <<-EOS.undent
+    program hello
+    print *, "Hello, world!"
+    end program hello
+    EOS
+    target_archs.each do |target_arch|
+      system "#{bin}/#{target_arch}-gcc", "-o", "hello-c.exe", "hello.c"
+      # TODO: add wine for test and figure how to run, currently
```
MikeMcQuaid: I don't think we need a wine test but good idea.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,220 @@
...
+          system "make"
+          system "make", "install-strip"
+        end
+      end
+    end
+  end
+
+  test do
+    (testpath/"hello.c").write <<-EOS.undent
+    #include <stdio.h>
```
MikeMcQuaid: Indent these by two characters more.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,220 @@
...
+      args = %W[
+        CC=#{target_arch}-gcc
+        CXX=#{target_arch}-g++
+        CPP=#{target_arch}-cpp
+        --host=#{target_arch}
+        --prefix=#{prefix}/#{target_arch}
+      ]
+
+      chdir "mingw-w64-libraries/winpthreads" do
+        mkdir "build-#{target_arch}" do
```
MikeMcQuaid: You could do the `mkdir` in a single step to avoid the `chdir`

Formula/mingw-w64.rb
MikeMcQuaid: Actually I'd rather see this resource it and remove the other formula as I don't think `mingw-w64-binutils` is useful alone.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,220 @@
...
+  option "without-i686", "Compile without i686 compiler"
+  option "without-x86_64", "Compile without x86_64 compiler"
+  # option "without-multilib", "Compile x86_64 copmiler without multilib support"
+
+  depends_on "texinfo" => :build
+  depends_on "mpfr"
+  depends_on "libmpc"
+  depends_on "isl"
+  depends_on "gmp"
+  depends_on "wine" => :optional
```
MikeMcQuaid: Remove this as I'm assuming it's only useful for testing.

MikeMcQuaid (changes requested):

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,220 @@
...
+      gcc_args = %W[
+        --target=#{target_arch}
+        --prefix=#{prefix}
+        --with-sysroot=#{prefix}
+        --disable-nls
+        --with-system-zlib
+        --enable-version-specific-runtime-libs
+        --enable-libstdcxx-time=yes
+        --enable-stage1-checking
+        --enable-checking=release
```
fxcoudert: `--enable-stage1-checking` is useless on a cross-compiler (i.e. non bootstrapped) build.
`--enable-checking=release`, `--enable-lto`, `--enable-libssp`, and `--enable-shared` are on by default, so we can remove them.

Regarding `--enable-libcc1`, I am not sure how useful it is for a cross-compiler, but it should also be the default if supported.

fxcoudert (changes requested):

Formula/mingw-w64.rb
cosmo0920: I see. 
I wouldn't like mega formula, but it is good point.
`mingw-w64-binutils` is not useful alone, I guess.

fxcoudert:
I have managed to get the multilib build to work. The build sequence I use is the following (with integrated `binutils`, as decided by @MikeMcQuaid): https://gist.github.com/fxcoudert/538c3d1ff2e55a348018a00e03d2f156

This will allow to remove the options, building the cross-compiler as 64-bit macOS binary that can target both 32-bit and 64-bit Windows.

vszakats:
I'm not sure where the limitation comes from, but so far all multilib mingw-w64 builds I've seen will support [SJLJ](https://stackoverflow.com/a/15685229/1732433) exception handling for both 32-bit and 64-bit targets, while the optimal (and therefore more popular) method is [DWARF](https://en.wikipedia.org/wiki/DWARF) for 32-bit and [SEH](https://en.wikipedia.org/wiki/Microsoft-specific_exception_handling_mechanisms#SEH) for 64-bit. I've only seen the latter two being supported by non-multilib builds, specifically in the mingw-64 builds hosted on MSYS2 and the standalone mingw-w64 builds distributed via SF. The MacPorts mingw-w64 packages are non-multilib as well (using SEH for 64-bit and what appears to be SJLJ for 32-bit, though I may be wrong here).

If such limitation exists indeed, I think it'd be better to support the non-multilib builds, one to target 32-bit with DWARF and one for 64-bit with SEH.

Of one popular mingw-w64 user, QT, requires the DWARF build, so using the multilib build doesn't seem to be an option. Relevant QT docs here:
   - <https://wiki.qt.io/Mingw-64-bit>
   - <https://wiki.qt.io/MinGW>

vszakats:
Maybe worth a mention that in my experience, packages are consistently more difficult to get built for Windows using the multilib mingw-w64 builds. Often it breaks internal assumptions about toolchain prefix and expected output, default output depends on the host build (to 32-bit for 32-hosted and 64-bit for 64-bit hosted), and most builds require to manually pass either `-m32` or `-m64` (and some other options for `windres`) to get the intended result. Sometimes these are not possible to do without altering/patching the build system (I've submitted numerous patches for curl and OpenSSL to address these).

vszakats:
Thank you @stephengroat for keeping the option.

I'd argue it'd be better to also make it the default though. It'd keep parity with the rest of the package managers (MSYS2, Linux distros (including Arch, Ubuntu), MacPorts).

Here are the download statistics from sourceforge.net mingw-w64 builds, where both DWARF/SEH (non-multilib) and SJLJ (multilib) is offered:

32-bit hosted: https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/6.3.0/threads-posix/
64-bit hosted: https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/6.3.0/threads-posix/

SJLJ (multilib): 397 + 369
DWARF / SEH (non-multilib): 6493 + 4186

stephengroat:
@fxcoudert @vszakats could you look at the [last jenkins log](https://bot.brew.sh/job/Homebrew%20Core/15047/version=yosemite/console)? this is the error i've always been getting when trying to compile multilib builds

Formula/mingw-w64.rb
MikeMcQuaid: I think we should do the opposite and remove the `mingw-w64-binutils` formula in this PR.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
...
+      elsif target_arch.start_with?("x86_64") && build.without?("multilib")
+        gcc_args << "--enable-64bit" << "--disable-multilib"
+      elsif target_arch.start_with?("x86_64") && build.with?("multilib")
+        gcc_args << "--enable-targets=all"
+      end
+
+      mkdir "#{buildpath}/gcc/build-#{target_arch}" do
+        system "../configure", *gcc_args
+        system "make", "all-gcc"
+        system "make", "check-gcc"
```
MikeMcQuaid: Definitely don't need a `check` here.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
...
+
+      mkdir "mingw-w64-headers/build-#{target_arch}" do
+        system "../configure", *args
+        system "make"
+        system "make", "install"
+      end
+
+      ln_s "#{prefix}/#{target_arch}", "#{prefix}/mingw"
+
+      resource("gcc").stage buildpath/"gcc"
```
MikeMcQuaid: Not disagreeing just curious: why do we need a full GCC as well as mingw?

fxcoudert: Building the mingw runtime from source needs a cross-compiler. Moreover, I think the most common use case of a macOS-hosted mingw toolchain is to cross-compile stuff to Windows from your Mac, for which you need all three components: binutils, mingw headers/runtime, and cross-compiler.

MikeMcQuaid: Thanks. Again stupid question: we can't reuse the existing GCC formula for any of this?

fxcoudert: (No stupid questions! sometimes I'm not sure about that stuff, and I'm listed as GCC maintainer, so…)

I don't think so… cross-compilation is really quite different from bootstrapped compiler.

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
+class MingwW64 < Formula
+  desc "Minimalist GNU for Windows with GCC."
+  homepage "https://mingw-w64.org/"
+  # https://sourceforge.net/p/mingw-w64/wiki2/Cross%20Win32%20and%20Win64%20compiler/
+  url "https://downloads.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v5.0.1.tar.bz2"
+  sha256 "9bb5cd7df78817377841a63555e73596dc0af4acbb71b09bd48de7cf24aeadd2"
+
+  option "with-multilib", "Compile x86_64 compiler without multilib support"
```
fxcoudert: The message should be "**with** multilib support"

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
...
+      end
+
+      chdir "#{buildpath}/gcc/build-#{target_arch}" do
+        system "make"
+        system "make", "install"
+      end
+
+      ln_s "../../lib/gcc/#{target_arch}/lib/libgcc_s.a", "#{prefix}/#{target_arch}/lib"
+
+      ENV["LDPATH"] = "#{target_arch}/#{lib}"
```
fxcoudert: Why use `ENV["LDPATH"]` here instead of adding `LDPATH=` to the `args` below?

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
...
+        CXX=#{target_arch}-g++
+        CPP=#{target_arch}-cpp
+        --host=#{target_arch}
+        --prefix=#{prefix}/#{target_arch}
+      ]
+
+      mkdir "mingw-w64-libraries/winpthreads/build-#{target_arch}" do
+        system "../configure", *args
+        system "make"
+        system "make", "install-strip"
```
fxcoudert: Why strip? (especially as the rest of the installed CRT is, rightfully, not stripped)

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
...
+  end
+
+  test do
+    (testpath/"hello.c").write <<-EOS.undent
+      #include <stdio.h>
+      int main()
+      {
+      puts("Hello, world!");
+      return 0;
+      }
```
fxcoudert: We could test the presence of the mingw-w64 headers by expanding the testcase slightly:

```
      #include <stdio.h>
      #include <windows.h>
      int main (void)
      {
        fprintf (stdout, "Hello, console!");
        MessageBox (NULL, TEXT("Hello, GUI!"), TEXT("HelloMsg"), 0);
        exit (EXIT_SUCCESS);
      }
```

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
...
+    (testpath/"hello.f90").write <<-EOS.undent
+      program hello
+      print *, "Hello, world!"
+      end program hello
+      EOS
+    target_archs.each do |target_arch|
+      system "#{bin}/#{target_arch}-gcc", "-o", "hello-c.exe", "hello.c"
+      system "#{bin}/#{target_arch}-g++", "-o", "hello-cc.exe", "hello.cc"
+      system "#{bin}/#{target_arch}-gfortran", "-o", "hello-f90.exe", "hello.f90"
+    end
```
fxcoudert: Can we test, at least slightly, that the executables have the right format?

```
assert_match "PE32+ executable (console) x86-64, for MS Windows", shell_output("file hello-f90.exe")
assert_match "file format pei-x86-64", shell_output("#{bin}/x86_64-w64-mingw32-objdump -a hello-f90.exe")
```

(that's for 64-bit, but we can adapt the patterns to match each target)

fxcoudert (changes requested):
Some more comments, but I think we're getting there…

Formula/mingw-w64.rb
stephengroat: I've been switching it back and forth since it currently still doesn't work. I'm going to comment the option out right now but leave all of the messages in the formula since I think someone will eventually get it working

Formula/mingw-w64.rb
stephengroat: saw examples of other homebrew formulas, thought this was preferable

Formula/mingw-w64.rb
stephengroat: unstripped

Formula/mingw-w64.rb
```diff
@@ -0,0 +1,177 @@
...
+        --enable-version-specific-runtime-libs
+        --with-bugurl=https://github.com/Homebrew/homebrew-core/issues
+        --enable-languages=c,c++,fortran
+        --with-ld=#{bin}/#{target_arch}-ld
+        --with-as=#{bin}/#{target_arch}-as
+        --with-gmp=#{Formula["gmp"].opt_prefix}
+        --with-mpfr=#{Formula["mpfr"].opt_prefix}
+        --with-mpc=#{Formula["libmpc"].opt_prefix}
+        --with-isl=#{Formula["isl"].opt_prefix}
+        MAKEINFO=#{Formula["texinfo"].opt_bin}/makeinfo
```
fxcoudert: MAKEINFO here should not be necessary, because the `depends_on "texinfo" => :build` puts it in the PATH.

stephengroat: trying a removal

Formula/mingw-w64.rb
fxcoudert: It may be working, but I think LDPATH should be for host linker, not target linker. An alternative is `--with-ld=` and `--with-as` (see: https://gist.github.com/fxcoudert/538c3d1ff2e55a348018a00e03d2f156) which I think helps it work in multilib (I got it working with the formula in that link).

Formula/mingw-w64.rb
stephengroat: added

Formula/mingw-w64.rb
stephengroat: added something in, monitoring for 32 bit case

Formula/mingw-w64.rb
stephengroat: this is for the mingw libraries, shouldn't that be for the target linker?

for the gcc build, i use the `--with-ld` and `as`

fxcoudert: Sorry, missed that… For the CRT, the solution I found simplest is to remove CC and CXX from the environment so that the configure script can guess the correct compiler, assembler and linker: 

```
      ENV.delete "CC"
      ENV.delete "CXX"
```

and then you don't need to specify LD or LDPATH.

stephengroat:
since we can't assume what other people have, i think we should explicitly
list them. when I do that (setting `CC` and `CXX`) i still get multilib
compile failures.

any ideas?

fxcoudert:
With the formula exactly as it is now, by uncommenting the multilib option, I can build with multilib enabled on my x86_64 macOS Sierra:

```
$ /usr/local/Cellar/mingw-w64b/5.0.1/bin/x86_64-w64-mingw32-gcc a.c
$ file a.exe                                                        
a.exe: PE32+ executable (console) x86-64, for MS Windows
$ /usr/local/Cellar/mingw-w64b/5.0.1/bin/x86_64-w64-mingw32-gcc a.c -m32
$ file a.exe                                                            
a.exe: PE32 executable (console) Intel 80386, for MS Windows
```

The exact formula I tested is this one: https://gist.github.com/fxcoudert/9ae2749935be0b05004e6d9f0abb27a4
Can you tell me what is your multilib failure so I can look into it?

stephengroat:
https://github.com/Homebrew/homebrew-core/pull/9037#issuecomment-274182110

fxcoudert:
Testing the current formula, it builds fine but tries to install libraries that conflict with native GCC compiler (GCC 6.3.0 from `gcc.rb` formula). We should symlink binaries into /usr/local/bin, but the libs should stay within the kegs and not be symlinked into /usr/local/lib

MikeMcQuaid:
> We should symlink binaries into /usr/local/bin, but the libs should stay within the kegs and not be symlinked into /usr/local/lib

Agreed. This should not have any conflicts with other GCC formulae.

fxcoudert:
@MikeMcQuaid I have been trying to test that, but I don't know homebrew well enough: how can you tell it that certain files or directories should not be symlinked/installed into /usr/local/lib?

MikeMcQuaid:
@fxcoudert You cannot. Generally in that case we either don't install those files or install them under `libexec` instead (sometimes by setting the `--prefix=` to `libexec`)

stephengroat:
what specific paths are conflicting?

fxcoudert:
Full list is there: https://gist.github.com/fxcoudert/05aafd8ce2575d0f41d23419cfd5d6a7
Pretty much everything in `/usr/local/lib/gcc/6/gcc`. I'm not even sure why the native compiler wants to symlink stuff in that directory, anyway.

fxcoudert:
One solution is thus to configure and install everything into `#{prefix}/toolchain`, and to create symlinks in `bin` (i.e. `#{prefix}/bin` to the binaries in `#{prefix}/toolchain/bin`. Sounds right?

MikeMcQuaid:
@fxcoudert Yes but let's use `libexec` which is more typical for Homebrew's usage.

fxcoudert:
@MikeMcQuaid Given that gcc already installs stuff in libexec, we'd then have `libexec/bin`, `libexec/lib`, and… `libexec/libexec`. That sounds confusing to me.

MikeMcQuaid:
@fxcoudert You've convinced me. How about `prefix/mingw` or similar? `toolchain` feels a bit weird but if it's the best then so be it.

fxcoudert:
Whatever we choose, it will have subdirectories `bin`, `include`, `lib`, `libexec`, `mingw`, `share` `i686-w64-mingw32` and `x86_64-w64-mingw32`. I'm not particularly fond of "toolchain" as directory name, but I guess `mingw/mingw` is a bit weird too :)

MikeMcQuaid:
@fxcoudert Again good point. `toolchain` feels like the best so far, agreed.

stephengroat:
`toolchain` seems a little generic, people might have other toolchains installed. does brew host any other cross compilers to use as a template? what about use the `target_arch` as the folder to install in?

fxcoudert:
`target_arch` is a bad idea for the same reason as `libexec` and `mingw` (because it will lead to `target_arch/target_arch` subdirs), and also because I think it's better to have the two toolchains coexist (to keep us as close as possible to a multilib dir, which we can enable at some point).

MikeMcQuaid:
> target_arch is a bad idea for the same reason as libexec and mingw (because it will lead to target_arch/target_arch subdirs)

Agreed.

stephengroat:
what about the formula's name? `mingw-w64`

MikeMcQuaid:
> what about the formula's name? mingw-w64

Seems fine, it matches the upstream project.

stephengroat:
@fxcoudert any idea when changing the `lib` directory the c compiler still works but the c++ and fortran compilers can't find `libgcc_s.a`?

fxcoudert:
@stephengroat after a long investigation, it appears to be a problem with the `--enable-version-specific-runtime-libs` option and the symbolic links you are creating. I have continued working on this formula, starting from your PR, removing those and simplifying a bit. I think we're close to getting that merged!

stephengroat:
@fxcoudert i added you to the `stephengroat/homebrew-core` repo. just push the changes right there, hopefully it'll be faster then having me try to copy them in

fxcoudert:
@stephengroat thanks… because I didn't have push access to your branch, I created a new, clean pull request at https://github.com/Homebrew/homebrew-core/issues/9726

stephengroat:
Instead of opening another PR with #9726, if people are fine without multilib builds and conflicting with `gcc`, I can configure this PR to do both. #9726 is currently configured without multilib and conflicts with `gcc`

fxcoudert:
@stephengroat #9726 does not conflict with `gcc` (I have both installed at the moment), and can build with multilib (although it is currently commented out, as in this PR).

I used this PR as starting point for #9726. I don't care whether it's in this one (with fixes) or the other, but I would like to see this formula okayed and merged. I thought using a clean PR with comments might make it easier for a maintainer to judge the current state of things and approve.

PR #9038, opened 2017-01-18
Fix cloog formula dependencies

stephengroat:
Fixing cloog dependencies from hard versions to following head(s).

-----

### DIFF
--- a/Formula/cloog.rb
+++ b/Formula/cloog.rb
@@ -20,8 +20,8 @@ class Cloog < Formula
   end
 
   depends_on "pkg-config" => :build
-  depends_on "gmp@4"
-  depends_on "isl@0.12"
+  depends_on "gmp"
+  depends_on "isl"
 
   def install
     system "./autogen.sh" if build.head?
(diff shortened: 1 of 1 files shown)


ilovezfs:
This requires changing all of the GCC reverse dependencies as well.

stephengroat:
What gcc reverse dependencies? I don't see anything when i run `brew deps --tree gcc`?

The website current states that `cloog` should be targeting `gmp-6.0.0` or higher (http://www.bastoul.net/cloog/download.php)

ilovezfs:
https://bot.brew.sh/job/Homebrew%20Core/14924/

stephengroat:
I understand, just there is also an issue that `0.18.4` states a requirement of 

> GMP 6.0.0 (or any later version).

ilovezfs:
@stephengroat it's possible those gcc's and their other dependencies could be updated to use gmp 6.

MikeMcQuaid:
> Fixing cloog dependencies from hard versions to following head(s).

Can you elaborate what this was planned to fix?

stephengroat:
Sure, I was trying to address #9037 which has cyclic dependencies due to the library requirement(s) of this package

ilovezfs:
Can you elaborate on what the cyclic dependencies are there?

stephengroat:
update to show failing tests in current configuration

ilovezfs:
Did I miss something, or did it fail both times regardless of the versions?

stephengroat:
worked on my box

MikeMcQuaid:
> Can you elaborate on what the cyclic dependencies are there?

@stephengroat Still need this information.

MikeMcQuaid:
Also, regardless, this is going to be a non-starter. The only things that currently depend on Cloog also depend on GMP@4 so we need the dependency tree as-is or it's going to break them. Regardless of whether this is preferable or not: you need to fix the formulae currently depending on `cloog` at the same time as making any adjustments like this and, if they can't live side-by-side with the mingw stuff, it can't be merged.

MikeMcQuaid:
> Regardless of whether this is preferable or not: you need to fix the formulae currently depending on cloog at the same time as making any adjustments like this and, if they can't live side-by-side with the mingw stuff, it can't be merged.

PR #9045, opened 2017-01-18
stormpath-cli: fix description

rdegges:
Updating the stormpath-cli formula description so it matches the official Github
description.


woodruffw:
Thanks @rdegges!

PR #9047, opened 2017-01-18
mingw-w64-binutils i686 and x86_64 builds

stephengroat:
To be compatible with i686 only mingw builds, this change makes
seperate i686 and x86_64 binutils builds for mingw-w64. It also
gives the option to disable multilib

-----

### DIFF
--- a/Formula/mingw-w64-binutils.rb
+++ b/Formula/mingw-w64-binutils.rb
@@ -74,12 +74,11 @@ class MingwW64Binutils < Formula
         system "#{bin}/#{target_arch}-ld", "-m", "i386pe", "-o", "test32.exe", "test32.o"
         assert_match "PE32 executable", shell_output("file test32.exe")
       end
-      if target_arch.start_with?("x86_64")
-        system "#{bin}/#{target_arch}-as", "-o", "test.o", "test.s"
-        assert_match "file format pe-x86-64", shell_output("#{bin}/#{target_arch}-objdump -a test.o")
-        system "#{bin}/#{target_arch}-ld", "-o", "test.exe", "test.o"
-        assert_match "PE32+ executable", shell_output("file test.exe")
-      end
+      next unless target_arch.start_with?("x86_64")
+      system "#{bin}/#{target_arch}-as", "-o", "test.o", "test.s"
+      assert_match "file format pe-x86-64", shell_output("#{bin}/#{target_arch}-objdump -a test.o")
+      system "#{bin}/#{target_arch}-ld", "-o", "test.exe", "test.o"
+      assert_match "PE32+ executable", shell_output("file test.exe")
     end
   end
 end
--- a/Formula/mingw-w64-binutils.rb
+++ b/Formula/mingw-w64-binutils.rb
@@ -37,7 +37,7 @@ class MingwW64Binutils < Formula
 
       mkdir "build-#{target_arch}" do
         system "../configure", *args
-        system "make"
+        system "make", "check"
         system "make", "install"
       end
 
--- a/Formula/mingw-w64-binutils.rb
+++ b/Formula/mingw-w64-binutils.rb
@@ -11,18 +11,40 @@ class MingwW64Binutils < Formula
     sha256 "39e04cd72f16ce87abac183cfa9d1bfe3b979378a938f4a8b987b79ea696261d" => :yosemite
   end
 
-  def install
-    system "./configure", "--disable-werror",
-                          "--target=x86_64-w64-mingw32",
-                          "--enable-targets=x86_64-w64-mingw32,i686-w64-mingw32",
-                          "--prefix=#{prefix}",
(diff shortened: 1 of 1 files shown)


Formula/mingw-w64-binutils.rb
```diff
@@ -11,18 +11,40 @@ class MingwW64Binutils < Formula
...
+  option "without-i686", "Compile without i686 mingw target"
+  option "without-x86_64", "Compile without x86_64 mingw target"
+  option "without-multilib", "Compile x86_64 mingw target without multilib"
+
+  def targets
+    if build.without?("i686") && build.without?("x86_64")
+      odie "Options --without-i686 and --without-x86_64 are mutually exclusive"
+    end
+    archs = []
+    archs.push("i686-w64-mingw32") if build.with?("i686")
```
dunn: We usually use `archs << "i686-w64-mingw32'` etc.; it'd be nice to maintain the convention.

stephengroat: no problem, i'll fix this on all prs i have open!

fxcoudert:
Given that binutils build perfectly fine as multilib (i.e. with both i386 and x86_64 targets), why do we want to provide the option of building without one or the other?

Also, why do 2 separate builds of the two archs?

ilovezfs:
Universal and 32-bit-only builds are both deprecated going forward, so this should only really have a 64 bit build.

fxcoudert:
@ilovezfs I think there is a misunderstanding: this is a cross-compiler, so when discussing i386 and x86_64 we're not discussing the host (i.e. the macOS machine, for which the build is 64-bit only), but the target (i.e. the Windows machine for which code is generated by the cross-compiler).

MikeMcQuaid:
> this is a cross-compiler, so when discussing i386 and x86_64 we're not discussing the host (i.e. the macOS machine, for which the build is 64-bit only), but the target (i.e. the Windows machine for which code is generated by the cross-compiler).

That seems fine then, thanks 👍 



> Given that binutils build perfectly fine as multilib (i.e. with both i386 and x86_64 targets), why do we want to provide the option of building without one or the other?
> 
> Also, why do 2 separate builds of the two archs?

I agree. Again, please explain the justification for this PR more @stephengroat.

stephengroat:
Since multilib builds currently don't work with `mingw-w64` gcc builds, currently to build `i686` binaries you need the `i686` tools (the `mingw-w64` gcc compilation process fails if you use `x86_64` multilib binutils when trying to compile `i686` gcc)

MikeMcQuaid:
> Since multilib builds currently don't work with mingw-w64 gcc builds, currently to build i686 binaries you need the i686 tools (the mingw-w64 gcc compilation process fails if you use x86_64 multilib binutils when trying to compile i686 gcc)

What do you do if you need to build 32-bit and 64-bit binaries?

stephengroat:
Currently, have both `i686` tools and `x86_64` tools. I just wanted to create the option to disable either build in case you want to build from source, since each build is about 20 minutes.

Default builds both toolchains

MikeMcQuaid:
Most users will consume this as a binary package. Let's just remove the non-default options, thanks.

stephengroat:
can i leave the `without--multilib` option? otherwise, it's impossible to have a non-multilib `x86_64` compiler. `mingw-w64` can be finicky, so have the control to get a non-multilib package sometimes helps

MikeMcQuaid:
Then we can make non-multilib the default.

stephengroat:
done

Formula/mingw-w64-binutils.rb
```diff
@@ -11,18 +11,32 @@ class MingwW64Binutils < Formula
     sha256 "39e04cd72f16ce87abac183cfa9d1bfe3b979378a938f4a8b987b79ea696261d" => :yosemite
   end
 
+  option "with-multilib", "Compile x86_64 mingw target with multilib support"
```
MikeMcQuaid: Don't want this option, thanks.

MikeMcQuaid (changes requested):

MikeMcQuaid:
I think this formula should probably just be removed and installed as part of https://github.com/Homebrew/homebrew-core/pull/9037 as I can't see a use for it alone. CC @woodruffw for thoughts.

fxcoudert:
The case I can see for providing two separate packages for binutils and compiler is that this could allow several compilers, either other than GCC or other versions of GCC, without recompiling/reinstalling binutils every time.

MikeMcQuaid:
> The case I can see for providing two separate packages for binutils and compiler is that this could allow several compilers, either other than GCC or other versions of GCC, without recompiling/reinstalling binutils every time.

When we have separate compilers we should consider it. When we have one compiler we should not. 

I don't see us providing multiple versions of MinGW.

stephengroat:
since focus is on #9037, closing this

PR #9053, opened 2017-01-19
webkitgtk 2.14.5

nijikon:

-----

### DIFF
--- a/Formula/webkitgtk.rb
+++ b/Formula/webkitgtk.rb
@@ -2,9 +2,8 @@ class Webkitgtk < Formula
   desc "Full-featured Gtk+ port of the WebKit rendering engine"
   homepage "https://webkitgtk.org/"
   homepage "https://webkitgtk.org"
-  url "https://webkitgtk.org/releases/webkitgtk-2.14.1.tar.xz"
-  sha256 "2e2d76c328de65bed6e0e4f096b2720a366654b27fc1af0830ece90bc4b7ceb5"
-  revision 2
+  url "https://webkitgtk.org/releases/webkitgtk-2.14.5.tar.xz"
+  sha256 "3ca8f1c33a9b43d6c753dcac1c0788656930e06382b10fdf5c2805ea8f96369f"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
@tschoonj weren't we avoiding upgrading this further for the time being?

tschoonj (commented):
Let's first see if this compiles properly...

@nijikon Have you tried compiling this locally? And tried running the MiniBrowser or any other software package that uses webkitgtk?

DomT4:
Good news, more CVEs have been assigned to the [existing version](https://webkitgtk.org/security/WSA-2017-0002.html) 😓.

MikeMcQuaid:
I think we need to think seriously (but sadly) about killing this formula and everything that depends on it.

DomT4:
Nudged for that a while back without much luck, but I think Homebrew has now shown a consistent problem updating this formula on macOS without having to constantly go through the report bug => wait for fix => apply fix cycle, whilst in the mean time known CVEs sit there and get shipped out to users.

IIRC it's not actually used by other formulae much _by default_. And maybe this version and everything that can use it now work fine, but that hasn't been the long-term situation so far 😢.

ilovezfs:
:-1: to killing this over some new CVs and irrelevant prior breakage.

DomT4:
Yes, Homebrew definitely has no prior history of killing formulae that have persistently proven prone to breakage on macOS. None at all. And who needs security 👀.

ilovezfs:
Not taking the bait. Nope, not gonna do it.

tschoonj:
Feel free to boneyard it. The quartz backend is not really supported upstream anyway.

ilovezfs:
I defer to @tschoonj since the MiniBrowser is unhappy https://gist.github.com/ilovezfs/c5a735bb9cf803048de011640c37c4ec

nijikon:
What is our decision? Are we killing it?

ilovezfs:
>I defer to @tschoonj since the MiniBrowser is unhappy https://gist.github.com/ilovezfs/c5a735bb9cf803048de011640c37c4ec (which is 2.14.4 built locally)

Yes. :fire:

PR #9082, opened 2017-01-20
cmocka: add static lib and cmockery support to cmocka

stephengroat:

-----
add static library and cmockery support to cmocka

### DIFF
--- a/Formula/cmocka.rb
+++ b/Formula/cmocka.rb
@@ -15,12 +15,13 @@ class Cmocka < Formula
 
   def install
     args = std_cmake_args
+    args << "-DWITH_STATIC_LIB=ON" << "-DWITH_CMOCKERY_SUPPORT=ON" << "-DUNIT_TESTING=ON"
     if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
       args << "-DHAVE_CLOCK_GETTIME:INTERNAL=0"
     end
 
     mkdir "build" do
-      system "cmake", "..", "-DUNIT_TESTING=On", *args
+      system "cmake", "..", *args
       system "make"
       system "make", "install"
     end


Formula/cmocka.rb
```diff
@@ -15,12 +15,13 @@ class Cmocka < Formula
 
   def install
     args = std_cmake_args
+    args << "-DWITH_STATIC_LIB=ON" << "-DWITH_CMOCKERY_SUPPORT=ON" << "-DUNIT_TESTING=ON"
```
dunn: Would anyone ever not want these options?

stephengroat: `STATIC_LIB` should always be on (to compile it, it's pretty small)
`CMOCKERY_SUPPORT` is backwards compatibility with the old cmockery test suite
`UNIT_TESTING` should work, but currently fails a test in the `make test` if built in ruby. should be run, currently working on a work around

MikeMcQuaid: We don't want to run `make test` so can probably remove `-DUNIT_TESTING`

stephengroat: why not run `make test`? A binary is being released, seems very important to test it

the only test that fails is a catch for seg faults, it's because of how ruby's `system` works. i'm trying to figure out how to avoid running it right now

MikeMcQuaid: Paging @ilovezfs. In short: this doesn't catch Homebrew bugs but good `test do` tests do.

stephengroat: I'm really not trying to be a problem and really appreciate all of the great work that homebrew does.

I think that `test do` catches homebrew installation issues. package unit tests, when they exist, should be run to confirm that the package is compiled correctly before a `bottle` is released.

MikeMcQuaid: Package unit tests should be run by the upstream project. If and when they catch Homebrew bugs, we should run them. They don't seem to, though. `test do` catches Homebrew or upstream issues if the software doesn't work as expected.

stephengroat: i think it's important for binary releases, though, since it's always very difficult to perfectly replicate compile-time environments. running `make check`, `make test`, etc. assures that the tests the packages maintainers wrote still work in your environment, which they should.

if not, is it really a good idea to release a binary that seems to not run properly? a lot of times, i don't think anyone other than original maintainers can confirm whether failing tests are an issue or not, but releasing a binary releases prevent an end user from knowing that a binary is potentially misbehaving

MikeMcQuaid: We disagree that it's important because it does not catch bugs in our experiences. Our `test do` blocks are better tests.

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:

For future reference the preferred commit message format is `cmocka: add static lib and cmockery support.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #9103, opened 2017-01-20
Rename boot-clj formula to boot

lukateras:
The tool (and the provided executable) is called "boot", and thus, this name for the formula is more straightforward.

### DIFF
--- a/Formula/boot.rb
+++ b/Formula/boot.rb
@@ -1,4 +1,4 @@
-class BootClj < Formula
+class Boot < Formula
   desc "Build tooling for Clojure"
   homepage "http://boot-clj.com"
   url "https://github.com/boot-clj/boot-bin/releases/download/2.5.2/boot.sh",


MikeMcQuaid:
My concern is this name is super generic and not googleable.

ilovezfs:
Agreed. I'd prefer we leave this name open.

yegortimoshenko:
Quick `brew search` showed me that there are many more generic formula names in Homebrew: [`go`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/go.rb) (could have been `golang` instead), [`joe`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/joe.rb), [`unicorn`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/unicorn.rb), [`neko`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/neko.rb) (could have been `nekovm`), etc...

I find it more straightforward when name of the formula matches the name of the tool, and executable. 
As far as I can see, `boot-clj` is the only prefixed formula in the repository.

Boot gains popularity, especially in CLJS community. Mind that Ruby once was a generic name, too.

yegortimoshenko:
On googleability: for me it's not that bad, actually.

screen shot 2017-01-20 at 16 58 25

The name definitely feels generic, but I believe it's not something that should (or can) be fixed from within a package manager.

MikeMcQuaid:
Passing on this for now, sorry.

PR #9154, opened 2017-01-22
ldns 1.7.0

aabdnn:
This commit updates ldns to 1.7.0

### DIFF
--- a/Formula/dnscrypt-proxy.rb
+++ b/Formula/dnscrypt-proxy.rb
@@ -3,6 +3,7 @@ class DnscryptProxy < Formula
   homepage "https://dnscrypt.org"
   url "https://github.com/jedisct1/dnscrypt-proxy/archive/1.9.4.tar.gz"
   sha256 "a79d5da0133344d38f8b3d3355c16269f11c15fbeedd0521e1a657b00ac503bb"
+  revision 1
   head "https://github.com/jedisct1/dnscrypt-proxy.git"
 
   bottle do
--- a/Formula/ldns.rb
+++ b/Formula/ldns.rb
@@ -1,9 +1,8 @@
 class Ldns < Formula
   desc "DNS library written in C"
   homepage "https://nlnetlabs.nl/projects/ldns/"
-  url "https://nlnetlabs.nl/downloads/ldns/ldns-1.6.17.tar.gz"
-  sha256 "8b88e059452118e8949a2752a55ce59bc71fa5bc414103e17f5b6b06f9bcc8cd"
-  revision 2
+  url "https://nlnetlabs.nl/downloads/ldns/ldns-1.7.0.tar.gz"
+  sha256 "c19f5b1b4fb374cfe34f4845ea11b1e0551ddc67803bd6ddd5d2a20f0997a6cc"
 
   bottle do
     sha256 "24224df22aaf86b534c0162afe2de2d977c7c49959abd2f43aaadb43b19533a9" => :sierra
@@ -12,14 +11,14 @@ class Ldns < Formula
   end
 
   depends_on "swig" => :build
-  depends_on "openssl"
+  depends_on "openssl@1.1"
 
   def install
     args = %W[
       --prefix=#{prefix}
       --with-drill
       --with-examples
-      --with-ssl=#{Formula["openssl"].opt_prefix}
+      --with-ssl=#{Formula["openssl@1.1"].opt_prefix}
       --with-pyldns
       PYTHON_SITE_PKG=#{lib}/python2.7/site-packages
     ]


woodruffw:
```
Error: freeswitch contains conflicting version recursive dependencies:
  openssl, openssl@1.1
View these with `brew deps --tree freeswitch`.
```

woodruffw:
Does this actually fail without `openssl@1.1`? We might want to hold off on that dependency bump in order to get this through.

ilovezfs:
There's a moratorium on new openssl@1.1 dependencies: https://github.com/Homebrew/homebrew-core/issues/4543#issuecomment-260371674

aabdnn:
@woodruffw ldns 1.7.0 can build against openssl 1.0.x, but then we have to pass some options to "configure" to turn off some functions of ldns that require openssl 1.1. Upstream recommends building against openssl 1.1.

MikeMcQuaid:
If this moves to OpenSSL 1.1 then Freeswitch and everything it depends on must be too.

ilovezfs:
which means

curl, xml-security-c, xml-tooling-c, opensaml, apr-util, and once you hit apr-utils, you can just keep going

MikeMcQuaid:
@ilovezfs Yeh so in this case: almost certainly not worth it.

aabdnn:
Guys, I don't understand. How is freeswitch depending on openssl 1.1 when I've only added it as a dependency for ldns?

ilovezfs:
Due to https://github.com/Homebrew/homebrew-core/blob/master/Formula/freeswitch.rb#L32

aabdnn:
Okay, perhaps I am too thick or missing something obvious. ldns doesn't need freeswitch. Why is the ldns build being held up by a different package?

ilovezfs:
It's the reverse. `freeswitch` currently depends on ldns: https://github.com/Homebrew/homebrew-core/blob/master/Formula/freeswitch.rb#L31

MikeMcQuaid:
> Why is the ldns build being held up by a different package?

Because we need to make sure we don't break our other packages through upgrading. This is standard practise for package managers.

aabdnn:
Aaaaah, now the penny has dropped. Thank you for patiently explaining it to me @ilovezfs.
I see two options. One is personal, where I build ldns 1.7.0 on a personal branch and use it privately without updating Homebrew. The other is to create an ldns@1.7 formula.
What do you guys think?

ilovezfs:
```
checking for libldns >= 1.6.6... checking for ldns_str2rdf_a in -lldns... no
configure: error: You need to either install libldns-dev or disable mod_enum in modules.conf
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: ./configure --disable-dependency-tracking --enable-shared --enable-static --prefix=/usr/local/Cellar/freeswitch/1.6.13_1 --exec_prefix=/usr/local/Cellar/freeswitch/1.6.13_1
```
so it may be possible to drop the ldns dependency from freeswitch.

The other one that's in the same boat is dnscrypt-proxy, which would need to be switched over to openssl@1.1.

MikeMcQuaid:
@aabdnn I'd consider just using OpenSSL 1.0 for now.

aabdnn:
@MikeMcQuaid I can't use openssl 1.0, because then some functions have to be disabled in ldns 1.7, and I want to use those specific functions (relating to DANE verification).

ilovezfs:
Actually dnscrypt-proxy only uses openssl via ldns, but isn't itself linked to it.

With mod-enum disabled, freeswitch does build successfully without needing ldns (diff below).

Here's a description of what the mod-enum feature is that would get sacrificed:

>ENUM (Electronic Number Mapping System) is a system that uses standard phone numbers as a key to lookup addressing and identification information that can potentially let Freeswitch servers communicate directly, thus bypassing the PSTN, reducing cost, and improving audio quality. It is based on the E.164 standard and uses standard DNS systems to translate telephone numbers into a special DNS record type called a NAPTR record (NamingAuthority PoinTer Resource Record) that might contain information for 1-(800)-555-1234 such as:

```
$ORIGIN 4.3.2.1.5.5.5.0.0.8.1.e164.arpa.
IN NAPTR 100 10 "U" "E2U+sip" "!^.*$!sip:customer-service@example.com!" .
IN NAPTR 102 10 "U" "E2U+email" "!^.*$!mailto:information@example.com!" .

```

(https://freeswitch.org/confluence/display/FREESWITCH/mod_enum)

```
diff --git a/Formula/freeswitch.rb b/Formula/freeswitch.rb
index dc7a5e6..b4ca3a7 100644
--- a/Formula/freeswitch.rb
+++ b/Formula/freeswitch.rb
@@ -28,7 +28,6 @@ class Freeswitch < Formula
 
   depends_on "curl"
   depends_on "jpeg"
-  depends_on "ldns"
   depends_on "openssl"
   depends_on "pcre"
   depends_on "sqlite"
@@ -147,6 +146,9 @@ class Freeswitch < Formula
   #------------------------ End sound file resources --------------------------
 
   def install
+    inreplace "build/modules.conf.in", "applications/mod_enum",
+                                       "#applications/mod_enum"
+
     system "./bootstrap.sh", "-j"
 
     # tiff will fail to find OpenGL unless told not to use X
bash-3.2$ 
```

ilovezfs:
https://github.com/Homebrew/homebrew-core/pull/9189 should allow you to use openssl@1.1 here.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

MikeMcQuaid:
Think we just need a `revision 1` added to `dnscrypt-proxy` and we're good to go.

ilovezfs:
I rebased on master, and pushed a revision for dnscrypt-proxy

ilovezfs:
@aabdnn shipped with the openssl@1.1 dependency! Thanks for your contribution to Homebrew.

PR #9164, opened 2017-01-22
Import formulae from homebrew/gui

MikeMcQuaid:
These all showed usage with analytics, aren't in Cask, don't have upstream binaries available and will help us kill another tap.


MikeMcQuaid:
Pulled without `xournal` which failed to apply a patch on all platforms.

PR #9189, opened 2017-01-23
freeswitch 1.6.14

ilovezfs:
Remove dependency on ldns, which means the mod_enum feature must be
disabled, to avoid OpenSSL version conflicts.


MikeMcQuaid:
I'm getting uncomfortable with the amount of vendoring we're doing to work around package management issues. In this case I personally think we should just continue to depend on OpenSSL 1.0.

ilovezfs:
The vendor is only needed to keep the mod_enum feature, which is trivial to disable:
```
    inreplace "build/modules.conf.in", "applications/mod_enum",
                                       "#applications/mod_enum"
```

PR #9193, opened 2017-01-23
Import formulae from homebrew/python.

MikeMcQuaid:
The other two are going to homebrew/science because they have hard homebrew/science dependencies.

Also see https://github.com/Homebrew/homebrew-python/pull/388 and https://github.com/Homebrew/homebrew-science/pull/4894.

### DIFF
--- a/Formula/numpy.rb
+++ b/Formula/numpy.rb
@@ -1,0 +1,74 @@
+class Numpy < Formula
+  desc "Package for scientific computing with Python"
+  homepage "http://www.numpy.org"
+  url "https://files.pythonhosted.org/packages/b7/9d/8209e555ea5eb8209855b6c9e60ea80119dab5eff5564330b35aa5dc4b2c/numpy-1.12.0.zip"
+  sha256 "ff320ecfe41c6581c8981dce892fe6d7e69806459a899e294e4bf8229737b154"
+
+  head do
+    url "https://github.com/numpy/numpy.git"
+
+    resource "Cython" do
+      url "https://files.pythonhosted.org/packages/b7/67/7e2a817f9e9c773ee3995c1e15204f5d01c8da71882016cac10342ef031b/Cython-0.25.2.tar.gz"
+      sha256 "f141d1f9c27a07b5a93f7dc5339472067e2d7140d1c5a9e20112a5665ca60306"
+    end
+  end
+
+  option "without-python", "Build without python2 support"
+
+  depends_on :fortran => :build
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
+  depends_on :python3 => :optional
+
+  resource "nose" do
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
CC @tdsmith for thoughts on whether all these should be imported and if any of them should be keg-only.

tdsmith:
Packaging Python libraries is swimming against the current for us; Homebrew can't do it well, Homebrew can't even interoperate well with pip, and the python ecosystem has made huge strides in delivering pre-built packages. A few libraries haven't caught up and tap support for those is fine but I don't like the precedent of moving these into core because we might attract more of them and that would be bad. I don't even think that moving numpy and scipy into core really provides value for most people most of the time; pip will work better. It makes some of our dependency logic for bindings a little simpler but accepting that a complex package like gnuradio is going to be a little inelegant might be better than carrying python libraries in earnest in core. Delivering these as keg only wouldn't be helpful. 

Anyway, -1 on anything but numpy and scipy; -0 on those. On my phone right now but willing to elaborate on anything here.

MikeMcQuaid:
@tdsmith Great, thanks. Will wait around for a longer response. Just so it's clear: the homebrew/python tap is going to get burned down anyway. Is it worth importing any of these into e.g. homebrew/science if needed instead or just killing them entirely? With `numpy` and `scipy`: I'm basically figuring that it's useful if we would ever need this for our formulae, if it saves mass vendoring or compilation time with bottles. If none of these are the case: let's not add.

tdsmith:
I don't really understand why killing the tap is helpful but that doesn't
change the calculus. Scipy is legitimately slow to build and needs numpy.

MikeMcQuaid:
> I don't really understand why killing the tap is helpful

We're killing all taps that are maintained purely by a subset of maintainers. This is because they offer a worse user experience, don't get e.g. updated/CI tested for reverse dependency failures and generally the same level of support and care as homebrew/core.

> but that doesn't change the calculus.

Cool, good to know.

> Scipy is legitimately slow to build and needs numpy.

Ok, cool. Let's add both and nothing else.

ilovezfs:
I'm trying to reconcile the viewpoint expressed above with the insanity of the practice of copy-pasting the exact same patch to `cryptography` into every formula that needs it. Ditto `pillow`.

tdsmith:
You could try asking either project to do a release.

ilovezfs:
That's completely beside the point.

MikeMcQuaid:
@BrewTestBot test this please

tdsmith:
Insofar as that's two (responsive and friendly!) projects out of hundreds of resourced packages, I'm not sure I am :)

> Will wait around for a longer response.

FWIW, I didn't have anything to add unless you wanted me to elaborate on something.

MikeMcQuaid:
> FWIW, I didn't have anything to add unless you wanted me to elaborate on something.

@tdsmith No, all good, thanks. Just didn't want you to feel rushed or anything.

PR #9198, opened 2017-01-23
Import from homebrew/completions

MikeMcQuaid:
Move all formulae into homebrew/core.


MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
@BrewTestBot test this please

alyssais:
Should completion formulae maybe skip the notability check? It's unlikely somebody's going to star the completions for a tool, so we might be better judging based on the popularity of the tool itself.

MikeMcQuaid:
@alyssais In general no, for this import yes.

MikeMcQuaid:
I didn't import `ctest-completion` due to a test failure and `composer-completion` due to it not having a repo but just being a Gist.

Issue #9202, opened 2017-01-23
Installing Git HEAD throws lime replacement error.


rye:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Bug reports:

Installing Git with the `--HEAD` flag fails (as of about 2 days ago) with the following error message:

```
Error: ["gitk-git/gitk", ["expected replacement of \"lime\" with \"\\\"#99FF00\\\"\""]]
```

Important to note that the build _does not even start_ as this seems to be a validation on the clone.

Here are logs from the HEAD install.

[`brew gist-logs git`](https://gist.github.com/anonymous/99d6c61f01d1f46779c05a98e52c5dca)

### Commentary:

I believe that this probably has to do with [a fix that was applied](https://github.com/Homebrew/homebrew-core/blob/feb75dce332e7e8bfa0aaef53190e83e01dab875/Formula/git.rb#L52-L55) to the Git formula to replace the color in order to solve #115.

Is this fix still necessary?  Removing [the linked lines](https://github.com/Homebrew/homebrew-core/blob/feb75dce332e7e8bfa0aaef53190e83e01dab875/Formula/git.rb#L52-L55) from the Git formula results in a successful build, and gitk works in the build.

I will open a PR shortly.

Issue #115, opened 2016-04-07
git: cannot run gitk


cybertk:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

### Bug reports:

After install **git** with `brew install git`, **gitk** is failed to run with following error

```
$ gitk
Error in startup script: unknown color name "lime"
    (processing "-fore" option)
    invoked from within
"$ctext tag conf m2 -fore [lindex $mergecolors 2]"
    (procedure "makewindow" line 347)
    invoked from within
"makewindow"
    (file "/usr/local/bin/gitk" line 12434)
```

mikemcquaid:
I can reproduce this. Can you see if you get the same result building `git` outside Homebrew? If so, please report this upstream and let us know. Thanks!

tdsmith:
What's `which wish`? I don't see this with Homebrew gitk + the system Tcl.

mikemcquaid:
`/usr/bin/wish` for me. Do you have X11 installed? (I don't)

DomT4:
I can reproduce the failure locally as well.

tdsmith:
I have Xquartz.

mikemcquaid:
@DomT4 Do you? If not my guess is that's the failure and we should probably just remove `gitk` (Apple doesn't install it in their `git`).

DomT4:
Yup, and if I punt `/opt/X11/bin` out of the `$PATH` it magically works again.

apjanke:
I can also reproduce on OS X 10.9.5 with latest XQuartz (2.7.8) installed, system Tcl/Tk.

[StackOverflow is on it](http://stackoverflow.com/questions/34637896/gitk-will-not-start-on-mac-unknown-color-name-lime). Seems like `lime` is new in something between Tcl 8.5.9 and 8.6.4.

The patch is trivial: https://github.com/Homebrew/homebrew-core/pull/236. With this, `gitk` works again for me on 10.9.5. Want to do this?

apjanke:
[...]

PR #9203, opened 2017-01-23
Git: Don't require replacement of "lime" with "99ff00" in gitk

rye:
It appears that the issue that the covered lines were introduced to solve, #115, was resolved upstream, and that we don't need to do this anymore.  As it stands, I get build errors when installing Git with the --HEAD flag, and this PR resolves them. (See #9202.)

### DIFF
--- a/Formula/git.rb
+++ b/Formula/git.rb
@@ -49,11 +49,6 @@ class Git < Formula
     ENV["PYTHON_PATH"] = which "python"
     ENV["PERL_PATH"] = which "perl"
 
-    # Support Tcl versions before "lime" color name was introduced
-    # https://github.com/Homebrew/homebrew-core/issues/115
-    # https://www.mail-archive.com/git%40vger.kernel.org/msg92017.html
-    inreplace "gitk-git/gitk", "lime", '"#99FF00"'
-
     perl_version = /\d\.\d+/.match(`perl --version`)
 
     if build.with? "brewed-svn"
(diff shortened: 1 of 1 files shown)


DomT4:
Are you sure this isn't an issue for stable still, even if fixed in `HEAD`? Homebrew's testing wouldn't catch that particular failure, so needs to be checked.

rye:
I haven't confirmed this.  Bisecting, it appears that the commit that fixed this was git/git@6e8fda5fd2d30d9e9e83af146b64c5b13ee7f2ef, which isn't present in `v2.11.0` (stable).  So, this seems like it might break things.  I'm unable to test this extensively, but I can tell that the commit that resolves this is only in HEAD.

Good catch.

I will fix this to only do this for HEAD builds for now.  We can remove the entire chunk of code once this commit in HEAD gets released.

Formula/git.rb
```diff
@@ -52,7 +52,10 @@ def install
     # Support Tcl versions before "lime" color name was introduced
     # https://github.com/Homebrew/homebrew-core/issues/115
     # https://www.mail-archive.com/git%40vger.kernel.org/msg92017.html
-    inreplace "gitk-git/gitk", "lime", '"#99FF00"'
+    #
+    # This has been resolved in Git (6e8fda5fd), which is currently present
+    # in HEAD but not in the stable.  This should be removed later.
+    inreplace "gitk-git/gitk", "lime", '"#99FF00"' if build.head?
```
DomT4: `if build.stable?`.

rye: Fix incoming.

dunn:
Thanks! Merged this as e600190ff23c179f6ed19070c64b41626a0d65dc.

PR #1593, opened 2016-05-31
llvm 3.8.0

maxim-belkin:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

---

Other changes: bring back libcxxabi, add libunwind, libffi, and add shared-libs
option. Linux and OS X compatible. Extra dependencies for Linux:
libedit, libxml2, ncurses, valgrind (recommended), zlib, and gcc. New
llvm build flags: LLVM_TARGETS_TO_BUILD=X86 (speeds up build process),
LLVM_ENABLE_DOXYGEN (if build with doxygen), BUILD_SHARED_LIBS (build
all libraries as shared instead of static, DLLVM_ENABLE_FFI and related
DFFI_INCLUDE_DIR and FFI_LIBRARY_DIR.

PR #9212, opened 2017-01-24
llvm@3.8: fix missing build targets

tmcdonell:
This seems to have been dropped in the transition to homebrew-core. Reverts to the old behaviour of building x86, ARM, NVPTX and AMDGPU targets by default, with this flag building all targets. This matches the behaviour of the llvm formula.

-----

### DIFF
--- a/Formula/llvm.rb
+++ b/Formula/llvm.rb
@@ -130,7 +130,6 @@ class Llvm < Formula
   option "with-python", "Build bindings against custom Python"
   option "with-shared-libs", "Build shared instead of static libraries"
   option "without-libffi", "Do not use libffi to call external functions"
-  option "with-all-targets", "Build all targets. Default targets: AMDGPU, ARM, NVPTX, and X86"
 
   depends_on "libffi" => :recommended # http://llvm.org/docs/GettingStarted.html#requirement
   depends_on "graphviz" => :optional # for the 'dot' tool (lldb)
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
@tmcdonell We are cutting down on options because they aren't tested by CI. We may be open to changing the defaults instead. Which are the targets you personally need?

tmcdonell:
Right, that sounds fair.

I currently require x86 and NVPTX. Probably also AMDGPU soon, but not right now.

I could update the patch, or close and submit a new one, if you like?

tmcdonell:
I have changed this to build all targets, which is the default LLVM build configuration by the way. This makes the formula maximally useful to all users.

MikeMcQuaid:
@tmcdonell That makes sense to me. Could you do the same on the other `llvm@` and `llvm.rb` formulae? Thanks!

tmcdonell:
Done! For the `llvm.rb` formula this means removing the `--with-all-targets` option, since it is now the default.

ilovezfs:
Just for the record, removing them was a rather deliberate decision made in https://github.com/Homebrew/homebrew-core/pull/1593#r65803232 to reduce build times.

MikeMcQuaid:
Thanks @ilovezfs. Given it's updated relatively infrequently I'm 🆒 with longer build times here. Can always add an option if from-source people want it different.

MikeMcQuaid:

PR #9230, opened 2017-01-24
Remove PowerPC support from GCC formula

fxcoudert:
GCC 6.3.0 and trunk do not actively support powerpc OS X targets. The latest successful build and test-results reported were in 2014 for darwin8 (https://gcc.gnu.org/ml/gcc-testresults/2014-07/msg02093.html). Making the GCC formula shorter and simpler will makes maintenance easier.

- Drop support for powerpc
- Drop dependency on cctools that was ppc-only
- Forcing `AS` and `AS_FOR_TARGET` was pre-Leopard only and not required anymore
- `--enable-plugin` is the default (see gcc's `config/gcc-plugin.m4`)
- `--enable-libstdcxx-time=yes` is not necessary anymore as configure checks now work for darwin (see `libstdc++-v3/acinclude.m4`)

-----

### DIFF
--- a/Formula/gcc.rb
+++ b/Formula/gcc.rb
@@ -1,17 +1,9 @@
 class Gcc < Formula
   def arch
-    if Hardware::CPU.type == :intel
-      if MacOS.prefer_64_bit?
-        "x86_64"
-      else
-        "i686"
-      end
-    elsif Hardware::CPU.type == :ppc
-      if MacOS.prefer_64_bit?
-        "powerpc64"
-      else
-        "powerpc"
-      end
+    if MacOS.prefer_64_bit?
+      "x86_64"
+    else
+      "i686"
     end
   end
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
:+1:. Would this apply to any of the versioned GCC formulae, too?

MikeMcQuaid (approved):

fxcoudert:
@MikeMcQuaid it depends what your criteria for maintaining the formulas are. If they are only expected to build on the recent macOS versions, yes. If you want to have formulas that can build on all macOS versions supported by GCC itself, no. For example, gcc 4.9 should compile fine on 10.5 (Leopard).

MikeMcQuaid:
@fxcoudert Yeh, we avoid deliberately breaking backwards compatibility (particularly around compilers) for this stuff when we can. Does that mean `gcc@5` should also have the same change made? Thanks!

fxcoudert:
The "build status" pages on the GCC website list 4.9 as the last release branch for which a successful  powerpc-apple-darwin build was reported: https://gcc.gnu.org/gcc-4.9/buildstat.html

However, GCC 5 did build for at least some time during its development, with the latest test results posted in April 2015: https://gcc.gnu.org/ml/gcc-testresults/2015-04/msg01438.html

So I think GCC 5 is borderline: maybe better not to touch it (if it ain't broken…).

MikeMcQuaid:
Ok, cool, thanks. Happy to ship as-is then 👍

MikeMcQuaid:

For future reference the preferred commit message format is `gcc: remove powerpc support `. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #9232, opened 2017-01-24
qt5 5.8.0

ilovezfs:
use -dbus-runtime and remove the dbus option
use a shim for mysql_config instead of inreplace
remove old PKG_CONFIG_LIBDIR inreplace

### DIFF
--- a/Formula/qt5.rb
+++ b/Formula/qt5.rb
@@ -3,11 +3,10 @@
 class Qt5 < Formula
   desc "Cross-platform application and UI framework"
   homepage "https://www.qt.io/"
-  url "https://download.qt.io/official_releases/qt/5.7/5.7.1/single/qt-everywhere-opensource-src-5.7.1.tar.xz"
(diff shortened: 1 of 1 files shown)


Formula/qt5.rb
```diff
@@ -68,28 +63,27 @@ def install
...
+      (buildpath/"brew_shim/mysql_config").write <<-EOS.undent
+        #!/bin/sh
+        if [ x"$1" = x"--libs" ]; then
+          mysql_config --libs | sed "s/-lssl -lcrypto//"
+        else
+          exec mysql_config "$@"
+        fi
+      EOS
+      chmod 0755, "brew_shim/mysql_config"
+      args << "-mysql_config" << buildpath/"brew_shim/mysql_config"
```
MikeMcQuaid: Nice approach 👍

MikeMcQuaid (approved):

PR #9234, opened 2017-01-24
Force using building GTK from source

manuseiffel:
This allows to use the patched version of GTK that allows EiffelStudio to run.

### DIFF
--- a/Formula/eiffelstudio.rb
+++ b/Formula/eiffelstudio.rb
@@ -12,7 +12,7 @@ class Eiffelstudio < Formula
   end
 
   depends_on "pkg-config" => :build
-  depends_on "gtk+"
+  depends_on "gtk+" => "build-from-source"
 
   def ise_platform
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
This won't do anything. What's the issue you're having?

manuseiffel:
Last time I tried without the change, it would take the non-patched version of GTK so EiffelStudio was not working properly.

MikeMcQuaid:
Can you point to the issue that explains this?

manuseiffel:
On Homebrew, there was this recent PR #9120 which corresponds to a patch I submitted for the following bug: https://bugzilla.gnome.org/show_bug.cgi?id=757187

MikeMcQuaid:
Ok but that should be included by the normal build and not require building from source. Additionally, it's probably worth asking again on that bug if it can be merged.

PR #9238, opened 2017-01-24
wine 2.0

uyjulian:


Formula/wine.rb
```diff
@@ -8,16 +8,9 @@ class Wine < Formula
...
   stable do
-    url "https://dl.winehq.org/wine/source/1.8/wine-1.8.6.tar.bz2"
-    mirror "https://downloads.sourceforge.net/project/wine/Source/wine-1.8.6.tar.bz2"
-    sha256 "b1797896eb3b63aab8a4753cc756d6211a0e85460146a1b52063ec79c13906d3"
-
-    # Patch to fix screen-flickering issues. Still relevant on 1.8. Broken on 1.9.10.
-    # https://bugs.winehq.org/show_bug.cgi?id=34166
-    patch do
-      url "https://bugs.winehq.org/attachment.cgi?id=52485"
-      sha256 "59f1831a1b49c1b7a4c6e6af7e3f89f0bc60bec0bead645a615b251d37d232ac"
```
apetresc: Should this patch actually be removed, or just replaced with https://bugs.winehq.org/attachment.cgi?id=55968 ?

apetresc:
The `-devel` target needs to be updated so it's not behind the new `-stable`. Though there isn't a devel release yet, so maybe just remove it for now.

Formula/wine.rb
```diff
@@ -57,13 +44,20 @@ def require_universal_deps?
...
     url "https://bugs.winehq.org/attachment.cgi?id=52384"
     sha256 "30766403f5064a115f61de8cacba1defddffe2dd898b59557956400470adc699"
   end
 
+  # Patch to fix screen-flickering issues. Still relevant on 2.0.
+  # https://bugs.winehq.org/show_bug.cgi?id=34166
+  patch do
+    url "https://bugs.winehq.org/attachment.cgi?id=55968"
+    sha256 "1b5086798ce6dc959b3cbb8f343ee236ae06c7910e4bbae7d9fde3f162f03a79"
+  end
```
MikeMcQuaid: If these patches aren't getting merged upstream we should strongly consider removing them. Any information on why upstream won't merge them?

MikeMcQuaid (changes requested):

apetresc:
@MikeMcQuaid [This comment](https://bugs.winehq.org/show_bug.cgi?id=34166#c4) explains it in some detail. Basically, they consider it a bug in Apple's OpenGL implementation, so they view hacking around it in Wine itself as some sort of moral compromise they're not willing to make, even though it does resolve the issue.

MikeMcQuaid:
@apetresc We need to get some resolution here, we can't just carry these unmerged patches forever. I've commented in https://bugs.winehq.org/show_bug.cgi?id=14939 and https://bugs.winehq.org/show_bug.cgi?id=34166 and let's hold off on this merge until there's more news.

xrisk:
Hi, any updates on this? Wine 2.0 patch is kinda important...

apetresc:
@xrisk: It looks like Mike is still going back-and-forth with the Wine developers in those two tickets, with responses as recently as this morning. So, it's still being figured out. As expected, they're resisting merging the patches upstream (for some seemingly good reasons, too).

MikeMcQuaid (approved):

MikeMcQuaid:
@uyjulian Can you rebase this on `master`? Thanks!

MikeMcQuaid:
Thanks for your contribution to Homebrew (and patience)! Without people like you submitting PRs we couldn't run this project. You rock!

alexchandel:
Just a heads up, it looks like there are newer versions of both patches than the ones we currently apply.

Also, the patch for https://bugs.winehq.org/show_bug.cgi?id=14939 was purportedly committed, so we may need to stop applying it to `HEAD` and `devel` builds soon.

PR #9249, opened 2017-01-25
Import from homebrew/x11.

MikeMcQuaid:
Deprecate homebrew/x11 and move all formulae here.

https://github.com/Homebrew/homebrew-x11/pull/292


MikeMcQuaid:
Pulled without `bochs` (doesn't build on Sierra/El Capitan), `curlftpfs` (will migrate to homebrew/fuse for now) and `mesalib-glw` (`with-static` is broken so going to create a PR just for that formula without that option)

PR #9278, opened 2017-01-26
Add libstemmer support to Groonga

nomoon:

-----

This is necessary in order to be able to use the `TokenFilterStem` token filter.

### DIFF
--- a/Formula/groonga.rb
+++ b/Formula/groonga.rb
@@ -3,6 +3,7 @@ class Groonga < Formula
   homepage "http://groonga.org/"
   url "http://packages.groonga.org/source/groonga/groonga-6.1.5.tar.gz"
   sha256 "bd404dca8860b4bb7af72d77020c95b32926f8976fecfe3ae2b9f8792e26105e"
+  revision 1
 
   bottle do
     sha256 "3d0724639c516f89b1351e7ff8024728bbe03a15f858800be88f937dd9527bff" => :sierra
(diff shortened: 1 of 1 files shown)


ilovezfs:
@BrewTestBot test this please

Formula/groonga.rb
```diff
@@ -41,6 +43,11 @@ class Groonga < Formula
     sha256 "bc83d1e5e0f32d4b95e219cb940a7e3f61f0f743abd3bd47c2d436a34e503870"
   end
 
+  resource "stemmer" do
+    url "https://github.com/snowballstem/snowball.git",
+      :revision => "71936098048f915a797de990776564c924d40b6a"
```
MikeMcQuaid: Any chance of addressing https://github.com/snowballstem/snowball/issues/7?

nomoon: I wish. The official notes from the new Snowball page suggest that if a "stable" source distribution is required, one should use the old download available at http://snowball.tartarus.org/download.html, which is from 2014 and nonetheless still contains no numeric version number. There has been no update on @snowballstem/snowball#7 for over a year. I could ping them, however.

It appears that the [dovecot](https://github.com/Homebrew/homebrew-core/blob/master/Formula/dovecot.rb#L27) and [sphinx](https://github.com/Homebrew/homebrew-core/blob/master/Formula/sphinx.rb#L28) formulae have run into this same problem and also chose to pick an arbitrary commit for the resource.

nomoon: Ubuntu, CentOS, etc. all have libstemmer packages which I believe are based on the 2014 source tar, which is certainly better than nothing, so I could alter the PR to use that if necessary.  groonga/groonga#430 suggests that the upstream project will eventually integrate libstemmer similarly to my implementation here.

MikeMcQuaid:

For future reference the preferred commit message format is `groonga: add libstemmer option.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

PR #9289, opened 2017-01-26
MonetDB 11.25.5

monetdb-team:


MikeMcQuaid:
Squash commits by running `brew update; git checkout $YOUR_BRANCH; git rebase --interactive origin/master` and then mark each commit other than the first as `squash` or `fixup` then `git push -f`. Thanks!

PR #9290, opened 2017-01-26
openssl@1.1 1.1.0d

alex:

-----


alex:
I don't understand the test failure here, looks like maybe a lib that depends on openssl@1.1 has a broken build?

DomT4:
Apparently `fdroidserver` is missing some dependencies. I would gently hazard a guess that at some point in went through part of a larger PR and scooped up some deps it usually wouldn't have access to.

You can either bump the revision without changing the deps and see if it solves the problem or you can add:

```ruby
depends_on "libtiff"
depends_on "webp"
```
to the formula, whilst bumping the revision number.

alex:
@DomT4 attempting with the modified deps, let's see if that works

alex:
Good call, that fixed it.

nijikon:

PR #9297, opened 2017-01-26
qt@5.7 5.7.1 (new formula)

ilovezfs:
This is a stop-gap for the following problems:

- [x] gammaray: The imported target "Qt5::UiPlugin" references the file "/usr/local/opt/qt5/lib/QtUiPlugin.framework" but this file does not exist. (https://github.com/Homebrew/homebrew-core/pull/9319)
- [x] gpsbabel: call to member function 'compare' is ambiguous (https://github.com/Homebrew/homebrew-core/pull/9322)
- [x] pyqt5: Error: This version of PyQt5 and the commercial version of Qt have incompatible
licenses. (https://github.com/Homebrew/homebrew-core/pull/9308)
- [x] qwt: qwt_designer_plugin.h:23: Error: Undefined interface (https://github.com/Homebrew/homebrew-core/pull/9321)
- [x] qwtpolar: qwt_designer_plugin.h:23: Error: Undefined interface (https://github.com/Homebrew/homebrew-core/pull/9321)
- [x] homebrew/science/octave: build hangs (https://github.com/Homebrew/homebrew-science/pull/4918)
- [x] homebrew/games/pc6001vx: qtmultimediadefs.h:1:10: fatal error: '../../src/multimedia/qtmultimediadefs.h' file not found (https://github.com/Homebrew/homebrew-games/pull/781)

### DIFF
--- a/Formula/qt@5.7.rb
+++ b/Formula/qt@5.7.rb
@@ -1,0 +1,163 @@
+# Patches for Qt5 must be at the very least submitted to Qt's Gerrit codereview
+# rather than their bug-report Jira. The latter is rarely reviewed by Qt.
+class QtAT57 < Formula
+  desc "Cross-platform application and UI framework"
+  homepage "https://www.qt.io/"
+  url "https://download.qt.io/official_releases/qt/5.7/5.7.1/single/qt-everywhere-opensource-src-5.7.1.tar.xz"
+  mirror "https://www.mirrorservice.org/sites/download.qt-project.org/official_releases/qt/5.7/5.7.1/single/qt-everywhere-opensource-src-5.7.1.tar.xz"
+  sha256 "46ebca977deb629c5e69c2545bc5fe13f7e40012e5e2e451695c583bd33502fa"
+  head "https://code.qt.io/qt/qt5.git", :branch => "5.7", :shallow => false
+
+  keg_only "Older version of qt5"
+
+  option "with-docs", "Build documentation"
+  option "with-examples", "Build examples"
+  option "with-qtwebkit", "Build with QtWebkit module"
+
+  deprecated_option "qtdbus" => "with-dbus"
+  deprecated_option "with-d-bus" => "with-dbus"
+
+  # OS X 10.7 Lion is still supported in Qt 5.5, but is no longer a reference
+  # configuration and thus untested in practice. Builds on OS X 10.7 have been
+  # reported to fail: <https://github.com/Homebrew/homebrew/issues/45284>.
+  depends_on :macos => :mountain_lion
+
+  depends_on "dbus" => :optional
+  depends_on :mysql => :optional
+  depends_on "pkg-config" => :build
+  depends_on :postgresql => :optional
+  depends_on :xcode => :build
+
+  # http://lists.qt-project.org/pipermail/development/2016-March/025358.html
+  resource "qt-webkit" do
+    url "https://download.qt.io/community_releases/5.7/5.7.1/qtwebkit-opensource-src-5.7.1.tar.xz"
+    sha256 "a46cf7c89339645f94a5777e8ae5baccf75c5fc87ab52c9dafc25da3327b5f03"
+  end
+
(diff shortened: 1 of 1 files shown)


dakcarto:
+1 for this. Not sure why 5.8.0 was pushed out to Homebrew in #9232 before anything that depends upon it was tested and updated to work with it. Kind of like living with only a Linux distro channel of `unstable` or `testing`. Just curious, has anyone ever investigated having `master` (stable) and `unstable` Homebrew branches before? 

You can add to the list:

- [ ] qgis3 formulae: The imported target "Qt5::UiPlugin" references the file "/usr/local/opt/qt5/lib/QtUiPlugin.framework" but this file does not exist. <-- does this seems like a `qt5` install issue?

ilovezfs:
>Not sure why 5.8.0 was pushed out to Homebrew in #9232 before anything that depends upon it was tested and updated to work with it.

Because all test blocks passed, which is our standard. If you have a problem with that, I'd suggest opening some PRs improving the test blocks. Note that regardless of the quality of the test blocks, that standard does not, and will not, include rebuilding every reverse dependency to check for build failure.

dakcarto:
Missing QtUiPlugin.framework reported here:
https://bugreports.qt.io/browse/QTBUG-57656

ilovezfs:
@dakcarto yeah that was just bizarre. Thank you. It seems to build but not get installed by make install, but what's built only seems to have the header files.

gborowik:
When I do
brew install --build-from-source pyqt5 --with-python --without-python3
for qt@5.7 still have this
Error: This version of PyQt5 and the commercial version of Qt have incompatible
licenses.

ilovezfs:
@gborowik right. And that's only the tip of the iceberg. Once you hack passed that there are more serious issues. This is addressed by #9308 until Riverbank catches up.

ilovezfs:
@dakcarto I'm confused. I don't think there currently is a formula for qgis3, is there?

MikeMcQuaid:
> Kind of like living with only a Linux distro channel of unstable or testing.

Homebrew is a rolling release distribution so: yes, it's like Debian unstable. People could create their own stable distributions with taps but personally I don't see the need; Debian stable is useful because you want stability on e.g. your libc libraries.

dakcarto:
@ilovezfs wrote:
> @dakcarto I'm confused. I don't think there currently is a formula for qgis3, is there?

Sorry for the confusion there (and delay on my response). There is a `qgis3-dev`, but just in a [very specific QGIS project tap](https://github.com/qgis/homebrew-qgisdev).

With regards to `qt@5.7` and the general `qt5` formulae, there is a _very significant_ blocking problem with using `qt@5.7` for low-level dependencies, like `pyqt5`, with other software the _is_ capable of building against `qt5` (and states that as its `depends_on`). This is the dreaded **"hey you are loading two versions of Qt, I'm going to segfault now"** (or something like that 😉).

Essentially, for any application like QGIS, which ends up with mixed Qt-dependent dependencies, this is a total blocking issue: instant crash. So, while this formula is a good idea, it is not a wise solution for low-level dependencies. Those really should be updated prior to rolling out an incompatible new Qt version. Certainly, IMO common deps like `pyqt5` should be compatible before such a roll out.

dakcarto:
I am currently trying to utilize all the `qt@5.7` formulae now, to build QGIS 3. Will post results, detailing my noted problem above, if it presents.

ilovezfs:
> There is a qgis3-dev, but just in a very specific QGIS project tap.

Outside of scope for us, then.

>This is the dreaded "hey you are loading two versions of Qt, I'm going to segfault now" (or something like that :wink:).

If you're talking about a problem with a formula that uses qt@5.7, please file an issue. If you're talking about a problem with something outside of the Homebrew with which you're attempting to use qt@5.7, it is your responsibility to make sure you're not linking against multiple versions.

>Certainly, IMO common deps like pyqt5 should be compatible before such a roll out.

We've already been through this above. I'm locking this issue now.

PR #9323, opened 2017-01-27
chakra 1.4.0 (new formula)

mathiasbynens:

-----

Ref. https://github.com/Microsoft/ChakraCore/issues/2278

### DIFF
--- a/Formula/chakra.rb
+++ b/Formula/chakra.rb
@@ -1,0 +1,39 @@
+class Chakra < Formula
+  desc "The core part of the Chakra JavaScript engine that powers Microsoft Edge"
+  homepage "https://github.com/Microsoft/ChakraCore"
+  url "https://github.com/Microsoft/ChakraCore/archive/v1.4.0.tar.gz"
+  sha256 "f6b4a3bce9e69bb3ea619a82805460906ef53d4f9da98b51e554dd48d6cdfdb3"
+
+  depends_on "cmake" => :build
+  depends_on "icu4c" => :optional
+
+  # Remove for > 1.4.0
+  # Fix build failure on case-sensitive file systems
+  # https://github.com/Microsoft/ChakraCore/pull/2421
+  patch do
+    url "https://github.com/Microsoft/ChakraCore/commit/bb5c393.patch"
+    sha256 "5694f260c910d11c59211766592002c27b3c1ecbb0c712f84e32c115d44666ed"
+  end
+
+  def install
+    # Build fails with -Os default
+    # Upstream issue from 26 Jan 2016 https://github.com/Microsoft/ChakraCore/issues/2417
+    # Fixed in master https://github.com/obastemur/ChakraCore/commit/cda81f4
+    ENV.O3
+
+    args = ["--static"]
+    if build.with? "icu4c"
+      args << "--icu=#{Formula["icu4c"].opt_include}"
+    else
+      args << "--no-icu"
+    end
+    system "./build.sh", *args
+
+    bin.install "BuildLinux/Release/ch" => "chakra"
+  end
+
+  test do
+    (testpath/"test.js").write("print('Hello world!');\n")
+    assert_equal "Hello world!", shell_output("#{bin}/chakra test.js").chomp
+  end
+end


mathiasbynens:
@obastemur [mentioned](https://github.com/Microsoft/ChakraCore/issues/2417#issuecomment-275652850) the issue could be this:

> IIRC Homebrew pushes for recent libc++ for osx 10.9+..
> 
> While installing with homebrew, you should use `--env=std` argument with brew install ... command. The rest should be just fine.

In that case, how can I force Homebrew to default to old libc++ for this formula, i.e. act as if `--env=std` was always supplied?

Formula/chakra.rb
```diff
@@ -0,0 +1,28 @@
...
+  def install
+    if build.with? "icu4c"
+      icu4c = Formula["icu4c"]
+      icu = "--icu=#{icu4c.opt_include}"
+    else
+      icu = "--no-icu"
+    end
+
+    system "./build.sh", "--static", "--test-build", icu
+    prefix.install "BuildLinux/Release/ch" => "chakra"
```
MikeMcQuaid: Why `BuildLinux`?

ilovezfs: It's hard coded: https://ci.dot.net/job/Microsoft_ChakraCore/job/master/job/osx_osx_release_static/lastBuild/console

ilovezfs:
The problem is not the c++ library, it's the optimization. Adding `ENV.O3` to the install method should fix it for you.

Formula/chakra.rb
MikeMcQuaid: Gentle nudge that upstream might want to consider renaming at some point 😉

Formula/chakra.rb
```diff
@@ -0,0 +1,28 @@
...
+  url "https://github.com/Microsoft/ChakraCore/archive/v1.4.0.tar.gz"
+  sha256 "f6b4a3bce9e69bb3ea619a82805460906ef53d4f9da98b51e554dd48d6cdfdb3"
+
+  depends_on "cmake" => :build
+  depends_on "icu4c" => :optional
+
+  def install
+    if build.with? "icu4c"
+      icu4c = Formula["icu4c"]
+      icu = "--icu=#{icu4c.opt_include}"
```
MikeMcQuaid: I'd populate an `args` array here instead as that's more typical formula form and inline `icu4c` as you're only using that variable once

ilovezfs:
This has a case insensitive file system bug, which can be fixed with
```
inreplace "pal/src/locale/unicode.cpp", "corefoundation", "CoreFoundation"
```

Formula/chakra.rb
```diff
@@ -0,0 +1,28 @@
...
+    else
+      icu = "--no-icu"
+    end
+
+    system "./build.sh", "--static", "--test-build", icu
```
ilovezfs: should be a release build, not a test build

Formula/chakra.rb
ilovezfs: This should be `bin.install` not `prefix.install`

ilovezfs:
I suggest the following:

```
class Chakra < Formula
  desc "The core part of the Chakra Javascript engine that powers Microsoft Edge"
  homepage "https://github.com/Microsoft/ChakraCore"
  url "https://github.com/Microsoft/ChakraCore/archive/v1.4.0.tar.gz"
  sha256 "f6b4a3bce9e69bb3ea619a82805460906ef53d4f9da98b51e554dd48d6cdfdb3"

  depends_on "cmake" => :build
  depends_on "icu4c" => :optional

  def install
    ENV.O3

    # Fix build failure on case-sensitive file systems
    inreplace "pal/src/locale/unicode.cpp", "corefoundation", "CoreFoundation"

    args = ["--static"]
    if build.with? "icu4c"
      args << "--icu=#{Formula["icu4c"].opt_include}"
    else
      args << "--no-icu"
    end
    system "./build.sh", *args

    bin.install "BuildLinux/Release/ch" => "chakra"
  end

  test do
    (testpath/"test.js").write("print('Hello world!');\n")
    assert_equal "Hello world!", shell_output("#{bin}/chakra test.js").chomp
  end
end
```

Formula/chakra.rb
mathiasbynens: What would inlining look like exactly? Would `"--icu=#{Formula['icu4c'].opt_include}"` work? (I’m not a Ruby programmer 😊)

Formula/chakra.rb
mathiasbynens: The test build enables command-line flags so most of the time people will want a test build. I’ll make this an option then

Formula/chakra.rb
mathiasbynens: Ah never mind, @ilovezfs already shows this below:

```
"--icu=#{Formula["icu4c"].opt_include}"
```

Formula/chakra.rb
obastemur: ``--target`` argument is added to set custom path and will be part of release 2.0. We also don't like `BuildLinux` naming. I have to admit, we've been lazy on changing it..

BTW; `BuildLinux/Release/ch` is a wrong path for `test-build`. It should be `BuildLinux/Test/ch`

mathiasbynens:
Thanks for the helpful reviews, @ilovezfs and @MikeMcQuaid! ♥ I’ve updated the PR.

ilovezfs:
@MikeMcQuaid are we permitting the use of libstdc++ in new formulae these days?

MikeMcQuaid:
@ilovezfs not intentionally but in this case we probably wouldn't have noticed and there's no obvious hacks happening. 
@mathiasbynens Why is `libc++` not supported, any ideas?

mathiasbynens:
> @mathiasbynens Why is `libc++` not supported, any ideas?

I have no idea — maybe @obastemur does? (Note: I am not affiliated with Microsoft or the ChakraCore team.)

obastemur:
> @mathiasbynens Why is libc++ not supported, any ideas?

https://github.com/Microsoft/ChakraCore/issues/2417#issuecomment-275849276

MikeMcQuaid:
Will wait until @ilovezfs's conversation there has concluded and leave it up to him whether we merge this given the stdlib situation.

ilovezfs:
-Os issue is fixed. You can drop the ENV.O3, by adding this patch from upstream:

```
  patch do
    url "https://github.com/obastemur/ChakraCore/commit/54874cfbe139daad5ae821c5e2710b6764632bd4.patch"
    sha256 "8f08b9708e43b9f3324187573455f3a1c01fe6861e39f7f909a60aced314c0ae"
  end
```

However, the deployment target/mmacosx-version-min and C++ issues remain unresolved: https://github.com/Microsoft/ChakraCore/issues/2417#issuecomment-276072593

mathiasbynens:
Patch updated!

mathiasbynens:
The build now fails on Yosemite:

```
[100%] Linking CXX executable ch
  "Js::JavascriptString::ConcatDestructive(Js::JavascriptString*)", referenced from:
      Js::JavascriptMath::AddLeftDead(void*, void*, Js::ScriptContext*, Js::JavascriptNumber*) in libChakra.Jsrt.a(JavascriptMathOperators.cpp.o)
      Js::JavascriptMath::Add_FullHelper(void*, void*, Js::ScriptContext*, Js::JavascriptNumber*, bool) in libChakra.Jsrt.a(JavascriptMathOperators.cpp.o)
make[2]: *** [bin/ch/ch] Error 1
make[1]: *** [bin/ch/CMakeFiles/ch.dir/all] Error 2
make: *** [all] Error 2
See error details above. Exit code was 2
```

cc @obastemur

obastemur:
@mathiasbynens it works on master.

we will have our first official cross-platform release soon, we may target that instead of 1.4.0? In the meantime, I will make sure we have this scenario covered in our CI. So, we wouldn't break it going forward.

mathiasbynens:
Can’t wait for that release!

Formula/chakra.rb
```diff
@@ -0,0 +1,35 @@
...
+  url "https://github.com/Microsoft/ChakraCore/archive/v1.4.0.tar.gz"
+  sha256 "f6b4a3bce9e69bb3ea619a82805460906ef53d4f9da98b51e554dd48d6cdfdb3"
+
+  depends_on "cmake" => :build
+  depends_on "icu4c" => :optional
+
+  def install
+    # Build fails with -Os default
+    # Upstream issue from 26 Jan 2016 https://github.com/Microsoft/ChakraCore/issues/2417
+    # Fixed in master https://github.com/obastemur/ChakraCore/commit/cda81f4
```
MikeMcQuaid: Can we backport this commit as a patch so it gets removed on the next release?

Formula/chakra.rb
```diff
@@ -0,0 +1,35 @@
...
+    ENV.O2
+
+    # Fix build failure on case-sensitive file systems
+    # https://github.com/Microsoft/ChakraCore/pull/2421
+    inreplace "pal/src/locale/unicode.cpp", "corefoundation", "CoreFoundation"
```
MikeMcQuaid: Similarly here.

MikeMcQuaid (changes requested):

Formula/chakra.rb
ilovezfs: Nope. It causes the build to fail on Yosemite.

Formula/chakra.rb
ilovezfs: Yes, this one could be applied as a patch.

MikeMcQuaid (approved):

ilovezfs:
@mathiasbynens Thanks for the new formula! :new:

mathiasbynens:
Thanks for the help! 👍

PR #9324, opened 2017-01-27
mesalib-glw: import from homebrew/x11.

MikeMcQuaid:
Remove `--enable-static` option which doesn't seem to work.

Skip adding a test for now for speed of tearing down homebrew/x11.


MikeMcQuaid:
Fixed the audits I could do trivially and don't affect the build in 7f46992192.

PR #9343, opened 2017-01-27
v8 5.8.48

oaleynik:

-----

This defaults v8 to 5.8.48, targets `d8` only and allows build latest HEAD with --HEAD option

Based on: https://github.com/pinepain/homebrew-devtools/blob/master/Formula/v8%405.7.rb

### DIFF
--- a/Formula/v8.rb
+++ b/Formula/v8.rb
@@ -3,8 +3,9 @@
 class V8 < Formula
   desc "Google's JavaScript engine"
   homepage "https://github.com/v8/v8/wiki"
-  url "https://github.com/v8/v8-git-mirror/archive/5.1.281.47.tar.gz"
-  sha256 "63c9933227d6912689ea6bc012eea6a1fabaf526ac04bc245d9381e3ea238bf6"
+  url "https://chromium.googlesource.com/chromium/tools/depot_tools.git"
+  version "5.6.326"
+  head "https://chromium.googlesource.com/chromium/tools/depot_tools.git"
 
   bottle do
     cellar :any
@@ -14,100 +15,71 @@ class V8 < Formula
     sha256 "7bcd1bbd66c11305eeea0c36ca472de8a639f511abe0909c8815b1208dbce7b6" => :mavericks
   end
 
-  option "with-readline", "Use readline instead of libedit"
-
   # not building on Snow Leopard:
   # https://github.com/Homebrew/homebrew/issues/21426
   depends_on :macos => :lion
 
-  depends_on :python => :build # gyp doesn't run under 2.6 or lower
-  depends_on "readline" => :optional
-  depends_on "icu4c" => :optional
+  # depot_tools/GN require Python 2.7+
+  depends_on :python => :build
 
   needs :cxx11
 
-  # Update from "DEPS" file in tarball.
-  # Note that we don't require the "test" DEPS because we don't run the tests.
-  resource "gyp" do
-    url "https://chromium.googlesource.com/external/gyp.git",
-        :revision => "4ec6c4e3a94bd04a6da2858163d40b2429b8aad1"
-  end
-
-  resource "icu" do
-    url "https://chromium.googlesource.com/chromium/deps/icu.git",
-        :revision => "c291cde264469b20ca969ce8832088acb21e0c48"
-  end
-
-  resource "buildtools" do
-    url "https://chromium.googlesource.com/chromium/buildtools.git",
-        :revision => "80b5126f91be4eb359248d28696746ef09d5be67"
-  end
-
-  resource "common" do
-    url "https://chromium.googlesource.com/chromium/src/base/trace_event/common.git",
-        :revision => "c8c8665c2deaf1cc749d9f8e153256d4f67bf1b8"
-  end
-
-  resource "swarming_client" do
-    url "https://chromium.googlesource.com/external/swarming.client.git",
-        :revision => "df6e95e7669883c8fe9ef956c69a544154701a49"
-  end
-
-  resource "gtest" do
-    url "https://chromium.googlesource.com/external/github.com/google/googletest.git",
-        :revision => "6f8a66431cb592dad629028a50b3dd418a408c87"
-  end
-
-  resource "gmock" do
-    url "https://chromium.googlesource.com/external/googlemock.git",
-        :revision => "0421b6f358139f02e102c9c332ce19a33faf75be"
-  end
-
-  resource "clang" do
-    url "https://chromium.googlesource.com/chromium/src/tools/clang.git",
-        :revision => "faee82e064e04e5cbf60cc7327e7a81d2a4557ad"
-  end
-
   def install
-    # Bully GYP into correctly linking with c++11
-    ENV.cxx11
-    ENV["GYP_DEFINES"] = "clang=1 mac_deployment_target=#{MacOS.version}"
-    # https://code.google.com/p/v8/issues/detail?id=4511#c3
-    ENV.append "GYP_DEFINES", "v8_use_external_startup_data=0"
+    (buildpath/"depot_tools").install buildpath.children - [buildpath/".brew_home"]
+    ENV["DEPOT_TOOLS_UPDATE"] = "0"
+    ENV.prepend_path "PATH", buildpath/"depot_tools"
 
-    if build.with? "icu4c"
-      ENV.append "GYP_DEFINES", "use_system_icu=1"
-      i18nsupport = "i18nsupport=on"
-    else
-      i18nsupport = "i18nsupport=off"
+    system "gclient", "root"
+    system "gclient", "config", "--spec", <<-EOS.undent
+      solutions = [
+        {
+          "url": "https://chromium.googlesource.com/v8/v8.git",
+          "managed": False,
+          "name": "v8",
+          "deps_file": "DEPS",
+          "custom_deps": {},
+        },
+      ]
+      target_os = [ "mac" ]
+      target_os_only = True
+    EOS
+
+    cache_location = ENV["CACHE_DEPS"]
+    no_sync = ENV["NO_SYNC"]
+
+    if cache_location && File.directory?(cache_location)
+      cp_r cache_location, "v8"
     end
 
-    # fix up libv8.dylib install_name
-    # https://github.com/Homebrew/homebrew/issues/36571
-    # https://code.google.com/p/v8/issues/detail?id=3871
-    inreplace "tools/gyp/v8.gyp",
-              "'OTHER_LDFLAGS': ['-dynamiclib', '-all_load']",
-              "\\0, 'DYLIB_INSTALL_NAME_BASE': '#{opt_lib}'"
+    system "gclient", "sync", "-vvv", "-j #{Hardware::CPU.cores}", "-r", version unless no_sync
 
-    (buildpath/"build/gyp").install resource("gyp")
-    (buildpath/"third_party/icu").install resource("icu")
-    (buildpath/"buildtools").install resource("buildtools")
-    (buildpath/"base/trace_event/common").install resource("common")
-    (buildpath/"tools/swarming_client").install resource("swarming_client")
-    (buildpath/"testing/gtest").install resource("gtest")
-    (buildpath/"testing/gmock").install resource("gmock")
-    (buildpath/"tools/clang").install resource("clang")
+    if !no_sync
+      rm_rf cache_location
+      cp_r "v8", cache_location
+    end
 
-    system "make", "native", "library=shared", "snapshot=on",
-                   "console=readline", i18nsupport,
-                   "strictaliasing=off"
+    cd "v8" do
+      arch = MacOS.prefer_64_bit? ? "x64" : "x86"
+      output_name = "#{arch}.release"
+      output_path = "out.gn/#{output_name}"
 
-    include.install Dir["include/*"]
+      # Configure build
+      gn_args = {
+        :is_debug => false,
+        :is_component_build => true,
+        :v8_use_external_startup_data => false,
+      }
 
-    cd "out/native" do
-      rm ["libgmock.a", "libgtest.a"]
-      lib.install Dir["lib*"]
-      bin.install "d8", "mksnapshot", "process", "shell" => "v8"
+      gn = "gn gen #{output_path} --args=\"#{gn_args.map { |k, v| "#{k}=#{v}" }.join(" ")}\""
+      system gn
+      system "ninja", "-j #{Hardware::CPU.cores}", "-v", "-C", output_path, "d8"
+
+      include.install Dir["include/*"]
+
+      cd output_path do
+        lib.install Dir["lib*.dylib"]
+        bin.install "d8" => "v8"
+      end
     end
   end
 


nijikon:
@BrewTestBot test this please

Formula/v8.rb
```diff
@@ -3,115 +3,77 @@
...
+      include.install Dir["include/*"]
 
-    cd "out/native" do
-      rm ["libgmock.a", "libgtest.a"]
-      lib.install Dir["lib*"]
-      bin.install "d8", "mksnapshot", "process", "shell" => "v8"
+      cd output_path do
+        lib.install Dir["lib*"]
+        bin.install "d8"
+      end
```
mathiasbynens: Could even do `bin.install "d8" => "v8"` IMHO. The `v8` shell in the code base is just an example with little real-world use — `d8` should be used instead. But it makes sense to give it a more appropriate name (just like `"ch" => "chakra"` when installing `brew install chakra`).

oaleynik: @mathiasbynens can I amend this commit and force push to the branch? Or it is ok to have two commits?

oaleynik: Indeed. I've been going to make that (d8 => v8), but was not sure is it ok :)

mathiasbynens: Since GitHub supports squash-and-merge nowadays, I’d assume Homebrew core devs are fine with either. Not sure though

oaleynik: @mathiasbynens pushed as separated commit. I can change it if will be needed :) Thanks for suggestion!

pinepain:
5.8 breaks BC with 5.1 which is default now. Maybe let's use versioned formula -`v8@5.8`?

pinepain:
**To clarify:** v8 releases doesn't follow semantic versioning at all: it even breaks BC in path releases, which happens on everyday basis, with [php-v8](https://github.com/pinepain/php-v8) I track v8 changes to public API almost on everyday basis. If there are packages who depends on default `v8`, they may stop working. Also, prior to 5.6 (or 5.7?) you had to build your binaries with v8 static  lib, which is not necessary nowadays as v8 build complete dynamic library.

I'd suggest just bring my [`v8@5.7`](https://github.com/pinepain/homebrew-devtools/blob/master/Formula/v8%405.7.rb) in, update it patch version to lkgr (last known good release) and creage `v8@5.8` from it with lkgr patch version. Later default `v8` could be phased out when in will be out of use (it could be tracked with homebrew build-in analytic and bintray downloads statistic).

P.S.: Apart of that - 👍, thanks for bringing it up!

mathiasbynens:
Versioning is being discussed here: https://github.com/Homebrew/brew/issues/620

pinepain:
@mathiasbynens I'm aware of, thanks. Do you think it would be better to move further discussion about this PR BC & versioning there?

mathiasbynens:
@pinepain Thanks for your work on creating and maintaining V8 formulae!

[My comment](https://github.com/Homebrew/homebrew-core/pull/9343#issuecomment-275856360) wasn’t aimed at you specifically — it was meant as a useful remark for anyone reading this thread now that you’ve brought up versioning.

> Do you think it would be better to move further discussion about this PR BC & versioning there?

@ilovezfs already [said as much](https://github.com/Homebrew/homebrew-core/pull/7168#issuecomment-265132022), so yeah :) FWIW your proposal sounds 👍 to me.

pinepain:
@mathiasbynens got it, good remark, last time @ilovezfs opened my eyes with poiting to that discussion which i somehow missed. Will bring it up there and create few PRs if we come up to some decision. 

As to exposing d8 as v8, I'd rather symlink or create some proxy script.

pinepain:
> This defaults v8 to 5.8.48, targets d8 only

@oaleynik Does this formula builds all necessary headers and libs so that other devs can reuse it, or it **d8 only**, as you originally wrote?

If it d8-only, it limit it usage significantly and definitely breaks BC, at this time at least `php{5x,7x}-v8js` formulae depends on it.

oaleynik:
@pinepain it is `d8` only. Intent is to have easy way for developers to check newest features in different JavaScript engines. It can make sense to create `v8-shell` formula which exposes `d8` as `v8` only.

Ref: https://github.com/mathiasbynens/todo/issues/17

oaleynik:
IMHO for me it feels that `v8` formula should reference latest stable `V8`. At least I would expect to get latest `V8` if I run `brew install v8`

pinepain:
@oaleynik 👍, one day it would be this way, but nowadays it may or even will unnecessary break things we may not even aware of. I'd suggest to go with some `v8-shell@x.y` which would depend on `v8@x.y` and expose `d8` binary as `v8` (see my note about ICU data file) or with version-independent  `v8-version` which would expose/switch current `d8/v8`.

FTR, latest stable is 5.6, not 5.8.

oaleynik:
The idea is to have the way to check *latest* features. Not use the latest stable version. But I agree that `v8-shell` would be better as it clearly indicates what it exposes.

pinepain:
@oaleynik I would also to have latest feature =), but v8 releases pop up almost every hour, sometimes even more often. You may want to build them with `--head`, or ask google v8 folks to package every it build for homebrew (if you will, ask please to build for ubuntu/debs and for centos/rhel families, internet won't forget that 😃). I'd prefer not to add --head option to version at this time, mostly because it's a bit cumbersome to get latest patch release version, but I will try to get it with `x.y-lkgr`, which should be far more enough for vast majority cases, normally lkgr is just a few patch versions behind (read, few hours behind). 

So will you be OK if I just PR my formula to homebrew-core (sure, adjusting to 5.8)? Then you can go with simple `v8-shell` (though, I'd suggest to make it versioned, `v8-shell@x.y` to prevent any misunderstandings or go with some d8 version switcher when multiple v8 versions installed).

oaleynik:
@pinepain I agree with either. IMHO, before installing anything I would know what I need (dev deps or shell or whatever...) so I don't see an issue with having `up-to-date` version of V8 under `v8` formula. On the other hand, I think it will be correct to have formula named after its intent - `v8-devel`, `v8-shell`, etc.. too.

pinepain:
@oaleynik have you looked what formulae depends on `v8`? Note, that homebrew-core is not the only tap, it is just a default one. Without being sure that nothing will be broken, I would not recommend to upgrade this core formula. Also notice, that because of internal changes to v8, js code run withing different v8 versions may provide different result, so if any depend on `v8` formula, they may be very surprised, but I would count just direct dependencies as the latter one is is incidental dependency and undetectable from out position.

P.S. having `--head` for 5.8 is really cumbersome as there is not lkgr yet, maybe it will be referenced later.

oaleynik:
@pinepain in `homebrew-core` is see just one package which depends on `v8` https://github.com/Homebrew/homebrew-core/blob/0e6b8fcf56ca7976b1ae8489323771fed4e6813a/Formula/gjstest.rb

and one which has `v8` as optional:
https://github.com/Homebrew/homebrew-core/blob/496e1baa572036290610551bb7f5db19a87e27f0/Formula/uwsgi.rb

I'm not sure about all the taps around

MikeMcQuaid:
> oaleynik commented 22 hours ago • edited
> @pinepain in homebrew-core I see just one package which depends on v8 https://github.com/Homebrew/homebrew-core/blob/0e6b8fcf56ca7976b1ae8489323771fed4e6813a/Formula/gjstest.rb
> 
> and one which has v8 as optional:
> https://github.com/Homebrew/homebrew-core/blob/496e1baa572036290610551bb7f5db19a87e27f0/Formula/uwsgi.rb
> 
> I'm not sure about all the taps around

As long as we don't break these I'm fine with updating the V8 version here.

Formula/v8.rb
```diff
@@ -3,111 +3,73 @@
...
-  url "https://github.com/v8/v8-git-mirror/archive/5.1.281.47.tar.gz"
-  sha256 "63c9933227d6912689ea6bc012eea6a1fabaf526ac04bc245d9381e3ea238bf6"
-
-  bottle do
-    cellar :any
-    sha256 "8106efc14371982af11a66d8db533dc0589bc240950e0e445467cf6ce8871393" => :sierra
-    sha256 "487f2ca72096ee27d13533a6dad2d472a92ba40ef518a45226f19e94d4a79242" => :el_capitan
-    sha256 "dc9af3e08eda8a4acd1ff3c6b47a4c5170a92dbab7d2d79958a14d8aa42eefac" => :yosemite
-    sha256 "7bcd1bbd66c11305eeea0c36ca472de8a639f511abe0909c8815b1208dbce7b6" => :mavericks
-  end
```
MikeMcQuaid: Leave this as-is, it'll be updated when we merge.

MikeMcQuaid (changes requested):

MikeMcQuaid:
Various CI failures: 
- https://bot.brew.sh/job/Homebrew%20Core/15494/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_v8___online/
- https://bot.brew.sh/job/Homebrew%20Core/15494/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/style_v8/
- https://bot.brew.sh/job/Homebrew%20Core/15494/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_gjstest/

pinepain:
@MikeMcQuaid this PR is fundamentally broken, not even mention that it breaks dependencies (not only listed one).

I would vote to use my PR #9392 which also provides working d8  (d8 from this PR won't work properly because of how ICU data file loaded by default, in referenced PR I patched that).

MikeMcQuaid:
> @MikeMcQuaid this PR is fundamentally broken, not even mention that it breaks dependencies (not only listed one).

@pinepain It may be but it can be improved. I'll comment on your PR.

pinepain:
@MikeMcQuaid sure it could be, but even if we copy the formula from my PR we can't do such update in `v8` formula itself as it will break all deps. V8 changed significantly between 5.1 and 5.8 in many aspects, from breaking API/ABI to building process.

MikeMcQuaid:
> as it will break all deps

We should try to fix these deps and report issues to upstream projects. In this PRs case it only seems to break a single formula.

pinepain:
@MikeMcQuaid this PR drop all .dylib building, so any deps won't work with it. As well as relocating d8 will lead to semi-working shell (see my comments above). Also, as you mentioned, 5.8 is not a stable (5.6 is). But @oaleynik interested in having latest v8 **shell** with latest ES features, if I get it right. While I'm absolutely for this changes, I'd rather keep them in a separate version, like I submitted.

> We should try to fix these deps and report issues to upstream projects
it might take from weeks to months, as a lot of changed in V8 API between 5.1 to 5.8 (I work with V8 actively since 4.8).

MikeMcQuaid:
> it might take from weeks to months, as a lot of changed in V8 API between 5.1 to 5.8 (I work with V8 actively since 4.8).

That's fine. That's our typical process for major upgrades like this.

MikeMcQuaid:
> this PR drop all .dylib building, so any deps won't work with it.

We should continue to build dylibs or have dependencies build statically.

pinepain:
OK,  let's sum up what we want to achieve:
 - do we want to upgrade v8 to 5.8 in this formula? 
 - do we want to upgrade v8 to just a stable version?

MikeMcQuaid:
If we upgrade v8 it should be to a stable version. We could consider adding 5.8 as a `devel` block regardless.

pinepain:
@oaleynik ftr, I added [v8@5.8](https://github.com/pinepain/homebrew-devtools/blob/master/Formula/v8%405.8.rb) to [pinepain/devtools](https://github.com/pinepain/homebrew-devtools), so if you up to give it a try, just `brew tap pinepain/devtools` && `brew install v8@5.8`. The `d8` exposed as `v8`, the only thing you will need to do, is to add opt_bin path to your `$PATH`.  I tested it and I guess it's what you need. Any later improvements are highly welcomed and thank you a lot for bringing the issue up! You did enormous work! I know how it time consuming to play around v8 and how many hidden snags does it have, so thank you one more time.

MikeMcQuaid:
> What is the main issue having v8@5.8 versioned formula?

It'll have to be maintained indefinitely and it's not a stable release which is against our packaging guidelines.

oaleynik:
@MikeMcQuaid I've cleared all the audit errors. This is my first PR to Homebrew - I've missed initially that I can't run audit on my fork (or probably I've done it wrong). Branch has been rebased over latest upstream master and all intermediary commits have been squashed. Also, I've set version to 5.6.326 which is the version of V8 used in Chrome stable. It is still possible to build from HEAD. @pinepain there is no `5.8-lgkr` tag or branch for V8 FYI.

@pinepain Although I found your https://github.com/Homebrew/homebrew-core/pull/9343#issuecomment-275927404 a bit insulting and disrespectful, which can turndown any desire to continue contributing to Homebrew - I highly appriciate all your help and suggestions. I believe that you have not had in mind any means to attack somebody or be aggressive. As I said a few comments above your concerns regarding such an optimistic update of v8 are reasonable. But I also think that having latest (well, at least latest stable with an ability to build from HEAD) V8 under `v8` formula will be correct as well. 

I'm not at the position of making the final decision here, however I believe that all together we can find the optimal solution!

pinepain:
@oaleynik I'm sorry that you found it in this way, though it was targets not a author personality, but the profound effect of what this changes will make on all dependencies.

It you want to contribute, just take formula from my pr, rename formula class to `V8` and replace version with 5.6.326.42. If I get you right, you need working d8, if it is, that's probably what you want. Thought, it would be just a 5.6.

pinepain:
> It'll have to be maintained indefinitely and it's not a stable release which is against our packaging guidelines.

@MikeMcQuaid yeah, makes sense, then I'll keep it in a separate tap as before.

MikeMcQuaid:
> @oaleynik I'm sorry that you found it in this way, though it was targets not your personality, but the profound effect of what this changes will make on all dependencies.

Just FYI: "I'm sorry that you X" is not an apology (it doesn't imply your actions are at fault but the other person was) and I agree that there's definitely more diplomatic ways of phrasing things than "fundamentally broken". Let's all try and assume the best in each other: everyone here is trying to make Homebrew better.

> @MikeMcQuaid yeah, makes sense, then I'll keep it in a separate tap as before.

If you're not going to be making your own bottles then a `devel` build here is probably just as good but ultimately: up to you.

pinepain:
To prevent further misunderstanding, I'll drop few lines in Russian.

@oaleynik Олег, я действительно не хотел ничего такого обидного написать, извини. Основноая идея в первоначальном виде данные измменения реально ломает очень много чего, т.е. не просто зависимости он ломает, а даже концепцию данной формулы. d8 просто консольная утилита, к слову, довольно примитивная если сравнивать с нодой ([src](https://github.com/v8/v8/blob/master/src/d8.cc), в то время как осоновная логика здесь в динамических библиотеках, сборка которых и выхлоп собственно и очень сильно помемнялась. Раньше собирались статические и динамические библиотеки, но примерно в 5.4 в v8 наконец-то довернули систему сборки и статические стали ненужны. И если для зависимостей решить проблему сборки путем исключения линкования статических библиотек не есть большая проблема, то вот тот ворох изменений в API v8 вряд ли кто-то будет разгребать. Вот собствено что я и имел в виду, но изложил из рук вон плохо. Впредь я таки лучше распишу личний раз чем просто писать "fundamentally broken".

@MikeMcQuaid:
 - Can we depends on --devel from other formula?
 - Is it ok to bump v8 version in devel?
 - Are there any chance that homebrew will bottle devel version in forseable future? Are there any discussions or some other resources where I can read.

MikeMcQuaid:
> Can we depends on --devel from other formula?

We cannot. That's another good case for a separate formula in your tap (until we need it in homebrew-core).

> Is it ok to bump v8 version in devel?

Yep.

> Are there any chance that homebrew will bottle devel version in forseable future? Are there any discussions or some other resources where I can read.

If someone submits a PR for it: yep. There's no active plans for feature issue tracking this currently.

PR #9389, opened 2017-01-28
crystal-lang: vendor libevent 2.0.22 during the build

ilovezfs:
The crystal-lang bootstrap binary is linked against our libevent 2.0.22,
so this will unblock the libevent 2.1.8 PR.

### DIFF
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -33,8 +33,28 @@ class CrystalLang < Formula
     sha256 "31de819c66518479682ec781a39ef42c157a1a8e6e865544194534e2567cb110"
   end
 
+  resource "libevent-2.0.22" do
+    url "https://github.com/libevent/libevent/releases/download/release-2.0.22-stable/libevent-2.0.22-stable.tar.gz"
+    sha256 "71c2c49f0adadacfdbe6332a372c38cf9c8b7895bb73dabeaa53cdcc1d4e1fa3"
+  end
+
   def install
+    resource("libevent-2.0.22").stage do
+      system "./configure", "--disable-dependency-tracking",
+                            "--disable-debug-mode",
+                            "--prefix=#{buildpath}/vendor/libevent"
+      ENV.deparallelize do
+        system "make"
+        system "make", "install"
+      end
+    end
+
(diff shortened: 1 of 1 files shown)


ilovezfs:
@asterite the crystal-lang bootstrap binary is causing :boom: in https://github.com/Homebrew/homebrew-core/pull/9344 because it is dynamically linked against Homebrew's libevent 2.0.22 (libevent-2.0.5.dylib). This PR is one way to work around the situation. Alternatively, we could wait for a new bootstrap binary from you that is linked against libevent statically.

Either way, it would be great if the next iteration of your bootstrap binary were statically linked against libevent so that this doesn't happen again. I presume we have the same problem for libgc.1.dylib (bdw-gc formula) the next time that gets an ABI breaking upgrade.

asterite:
@ilovezfs Hm, that's going to be tough, because there's no easy way to say "I want to statically link this library" in Crystal. I guess it's time we start thinking how to do that...

Formula/crystal-lang.rb
```diff
@@ -33,9 +33,29 @@ class CrystalLang < Formula
...
+        system "make"
+        system "make", "install"
+      end
+    end
+
     (buildpath/"boot").install resource("boot")
 
+    macho = MachO.open("#{buildpath}/boot/embedded/bin/crystal")
+    macho.change_dylib("/usr/local/opt/libevent/lib/libevent-2.0.5.dylib",
+                       "#{buildpath}/vendor/libevent/lib/libevent-2.0.5.dylib")
```
nijikon: Why there is `2.0.5` here if we build agains `2.0.22`?

nijikon: Nevermind, this is weird.

```
bash-3.2$ ls libevent*.dylib
libevent-2.0.5.dylib		libevent_core-2.0.5.dylib	libevent_extra-2.0.5.dylib	libevent_openssl-2.0.5.dylib	libevent_pthreads-2.0.5.dylib
libevent.dylib			libevent_core.dylib		libevent_extra.dylib		libevent_openssl.dylib		libevent_pthreads.dylib
```

PR #9392, opened 2017-01-28
[WIP] v8@5.8 5.8.60 (new formula)

pinepain:
-----

As discussed and proposed in #9343 and https://github.com/Homebrew/brew/issues/620#issuecomment-275857021

Rationale: I actively use latest v8 in [php-v8](https://github.com/pinepain/php-v8) PHP extension and I plan to make it available on homebrew (within `homebrew/php` tap). I already have formulae in [pinepain/devtools](https://github.com/pinepain/homebrew-devtools) tap, but want to make it available to larger audience. As building v8 is time-consuming, it would be nice to have it built once and hosted for all of us. Also, other formulae may utilize this version and thus move to versioned approach to reduce PITA level with preserving BC.

I would take about this formulae and will update it as mentioned in https://github.com/Homebrew/brew/issues/620#issuecomment-275857021, on monthly basis (or weekly, but it may abuse build system as building v8 is very time and resource consuming) to any v8  5.8.x changes. After v8 5.8 become stable, I will update it only if BC-compatible changes appears in public interface (`v8.h` only at this time as other things are highly volatile at this time).

FTR, I already package v8 for Ubuntu for more then a year, (see libv8-* PPAs at https://launchpad.net/~pinepain, packaging scripts are under https://github.com/pinepain/ppa-packaging), so I more or less into it and have some perspective of how much fun is to build v8. 

There are some PRs about updating v8 version at this time:
 - #9343 - which focuses only on having d8 and based on my formula which also includes d8, but as that PR targets updating `v8` version from 5.1 to 5.8 it will break all dependent formulae, v8 doesn't follow semver at all and can introduce BC-incompatible changes in patch versions.
- #7168 - which did an awesome work to adapt formula to new v8 build approach (they changed it somewhere at 5.4 or 5.5). This PR targets to update v8 from 5.1 to 5.4 which doesn't make much practical sense as it will also break BC and 5.4 is a bit outdated nowadays (5.6 is stable one). Also, this PR is not finished for some reasons, though, I highly appreciate it author, @Alhadis efforts. I used this PR as a reference to update own v8 version under [pinepain/devtools](https://github.com/pinepain/homebrew-devtools) tap.

Credits:
 - all `v8` formula contributors and maintainer;
 - @Alhadis for his efforts to deal with v8 build system changes in #7168;
 - @oaleynik for bringing up the question to update v8 in #9343 and all;
 - @ilovezfs, @mathiasbynens with commenters and reviewers on that PRs and formulae.

### DIFF
--- a/Formula/v8@5.8.rb
+++ b/Formula/v8@5.8.rb
@@ -3,8 +3,8 @@
 class V8AT58 < Formula
   desc "Google's JavaScript engine"
   homepage "https://github.com/v8/v8/wiki"
-  url "https://chromium.googlesource.com/chromium/tools/depot_tools.git"
-  version "5.8.60"
+  url "https://github.com/v8/v8/archive/5.8.69.tar.gz"
+  sha256 "15271aa814a8e00f8a1b493084513eb5e0b88d6a4ca662f26f0a446bfe568d99"
 
   keg_only "Provided V8 formula is co-installable and it is not installed in the library path."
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> After v8 5.8 become stable

I'd rather we didn't merge 5.8 in any form until it's designated stable. I'd also rather we added the newer version as `v8` and made the older version a `v8@`.

pinepain:
@MikeMcQuaid while v8 doesn't follow semver, "stable" is a bit volatile state for v8. What considered "stable" is 5.6, which is not bad, but it misses a lot of what 5.8 get. I'd rather keep rolling release alongside v8.

MikeMcQuaid:
@pinepain We could add that as a `devel` block but then `v8.rb` should install 5.6 by default.

pinepain:
@MikeMcQuaid will --devel use binaries or it will build it from sources? If it will be built from sources, it's not quite useful as building v8 is enormous task even for top up to date macs. If homebrew doesn't want to have this version, I'm absolutely fine to keep using my [pinepain/devtools](https://github.com/pinepain/homebrew-devtools) tap.

pinepain:
Offtopic a bit, @MikeMcQuaid do you know by chance how to build and serve bottles for non-homebrew taps? I would really appreciate some reference, links or examples.

MikeMcQuaid:
> Offtopic a bit, @MikeMcQuaid do you know by chance how to build and serve bottles for non-homebrew taps? I would really appreciate some reference, links or examples.

@pinepain Building is easy; `brew tap-new` will create a `.travis.yml` file for you. Uploading is the hard bit that we have no general solution for.

> @MikeMcQuaid will --devel use binaries or it will build it from sources? If it will be built from sources, it's not quite useful as building v8 is enormous task even for top up to date macs.

`--devel` will not use binaries yet. We'd consider a PR for this, though, to Homebrew/brew.

Alhadis:
> Also, this PR is not finished for some reasons, though, I highly appreciate it author, @Alhadis efforts.

Just to clarify: I paused because I was waiting on an update from @ilovezfs after he started refactoring the build logic. He was concerned Mike would "lose his cookies" over the fact I had to tamper with a build configuration file to get the necessary libraries compiled. He never reported progress in #7168, so the PR drifted into awkward silence.

Also, you're welcome regarding the GN-work. It was fun.

pinepain:
@Alhadis you did a great work there! The v8 build system is a real rocket science. Thanks a lot! 

I don't see a simple way to build v8 except using "hack" I used, when url points to `depot_tools` repo instead of v8 and then `gclient sync` v8 with deps. The proper way might be to write separate download strategy for v8 (which goes far beyond my personal needs).

ilovezfs:
@pinepain Note that sync_gateway uses depo_tools with caching in the Homebrew cache and not messing with the standard URL. I suggest you copy that.

Formula/v8@5.8.rb
```diff
@@ -0,0 +1,118 @@
+# Track Chrome stable.
+# https://omahaproxy.appspot.com/
+class V8AT58 < Formula
+  desc "Google's JavaScript engine"
+  homepage "https://github.com/v8/v8/wiki"
+  url "https://chromium.googlesource.com/chromium/tools/depot_tools.git"
```
ilovezfs: This should be a resource not a replacement for the stock URL. Please see the sync_gateway formula for how to use depot_tools with the Homebrew cache and without messing with the standard URL.

Alhadis:
Honestly, it's best to let `depot_tools` handle as much of the setup as possible. I haven't actually gone through this PR's changes, but I'm assuming it's basically picking up where I left off...

> Note that sync_gateway uses depo_tools with caching in the Homebrew cache and not messing with the standard URL. I suggest you copy that.

Yes, please do. Without caching, rebuilding with different options means redownloading EVERYTHING all over again. Total pain-in-the-butt,  and the entire reason I asked for @ilovezfs's intervention in the first place.

Formula/v8@5.8.rb
pinepain: Ideally, it should be, what's here is dirty hack.

The issue that cloning v8 repo with git doesn't make sense as the only supported way to do that is to use `gclient`, at least that's what v8 docs says (https://github.com/v8/v8/wiki/Using%20Git).

So here we may cache gclient synced v8 folder to reuse it later.

pinepain:
> rebuilding with different options means redownloading EVERYTHING all over again

we don't have that options (thankfully), but yeah, it what I used `DEV_CACHE_DEPS` env var. Terrible hack which we need to get rid off.

Formula/v8@5.8.rb
ilovezfs: The issue is getting the relevant manifest file and then syncing things up as in the sync_gateway formula, which then clones everything in the cache, having nothing to do with the original url download: https://github.com/Homebrew/homebrew-core/blob/master/Formula/sync_gateway.rb#L37-L51

pinepain: Thanks for the `sync_gateway` suggestion, I see how do they handle caching, that's what I should also do. Though, I'm not sure about how should I handle `url`, as it would be misleading to clone v8 (and a waste of time as it huge) and not using it later.

pinepain: I don't get the part about manifest, sorry. In a which way it could be used by depot_tools to fetch v8?

ilovezfs: How large the source archive tarball on GitHub? I assume it has the relevant manifest file?

pinepain: I still can't get about manifest. V8 doesn't use manifest like sync_gateway or did I missed something?

ilovezfs: I see. So gclient is serving that function I guess?

ilovezfs: You'll need to find a way to reuse the cache, and sync it up with the new v8 version as the version changes. It won't really be useful if it requires the user to blow it away manually. I assume gclient sync is able to do that without re-downloading everything if you pass it the current version.

pinepain: `gclient sync` + generated solutions. At least this is (unfortunately) the official way to build v8. 

To give a better perspective how nice v8 is, on ubuntu this sync also fetch full toolchin to build v8 and don't even use system one (there is an option to use system, but it doesn't work. Officially.)

pinepain: As I see it for now, we can specify url which to point v8 repo, and move depot_tools to resources section, which will be a valid approach. Then fetch v8 with depot tools (or take it from cache), sync and update cache. After that we can resume standard build process.

ilovezfs: The source archive tarball is only 22M. And it should let us update the formula with `brew bump-formula-pr` since the URL has the version number in it. Also, it has the license file, changelog, etc. which get installed automatically brew in the prefix, so it's not a total waste as long as you leave it unpacked in the buildpath, and do the other operations in a subdirectory and/or in the cache.

pinepain: Should I use the form `url "https://chromium.googlesource.com/v8/v8.git", :tag => "5.8.60"`?

Alhadis:
Guys, just letting you know I'm unsubscribing. If you need my input, just ping me again or e-mail me directly (address is on my profile). =) Good luck!

Formula/v8@5.8.rb
ilovezfs: I'd try to get away with just https://github.com/v8/v8/archive/5.8.68.tar.gz

pinepain:
@ilovezfs I made some changes regards sources, PTAL.

ahundt:
@pinepain I know OSRF figured out how to host their own bottles, maybe ask them? https://github.com/osrf/homebrew-simulation

MikeMcQuaid:
It's relatively simple once they are built: upload them somewhere. The "somewhere" is why we don't have a non-Bintray generic solution for this.

pinepain:
I see bintray has free plans for OSS projects, I'll look into brew sources to see how it works as I can't find any manual on example how people upload binaries and how to update bottle block. I guess it's not very popular among developers.

MikeMcQuaid:
There's no specific document but the `brew test-bot --help` command will give you more information and the source is here: https://github.com/homebrew/homebrew-test-bot

ahundt:
I did ask them what they do as well for my own interest, bc they may have some useful advise of gotchas (no response yet):
https://github.com/osrf/homebrew-simulation/issues/214

pinepain:
Thank you! I'll give a detailed look to both test-bot sources and to OSRF build scripts. Meanwhile @mathiasbynens asked v8 folks for to provide some prebuilt binaries as a part of their building process - https://bugs.chromium.org/p/v8/issues/detail?id=5918.

pinepain:
Closing as it doesn't match homebrew-core stability requirement.

PR #9403, opened 2017-01-29
gnuplot: move gd to recommended

equal-l2:

-----
Also remove some dependencies which are not used by gnuplot directly, but used by gd.

### DIFF
--- a/Formula/gnuplot.rb
+++ b/Formula/gnuplot.rb
@@ -127,10 +127,10 @@ class Gnuplot < Formula
 
   test do
     system "#{bin}/gnuplot", "-e", <<-EOS.undent
-      set terminal png;
-      set output "#{testpath}/image.png";
+      set terminal dumb;
+      set output "#{testpath}/graph.txt";
       plot sin(x);
     EOS
-    File.exist? testpath/"image.png"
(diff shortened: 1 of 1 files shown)


ilovezfs:
Why is this a useful change?

equal-l2:
This was useful at least just for me because I never used terminals provided by gd.

I think cairo provides better support for png than gd, and I don't use gif and jpeg, which are provided by gd.

equal-l2:
I think the removal of dependencies is a bit more useful, since gnuplot itself doesn't need them, and `webp`, `tiff` are not used at all (no corresponding terminals).

Formula/gnuplot.rb
```diff
@@ -82,6 +77,10 @@ def install
 
     args << "--with-pdf=#{pdflib}" if build.with? "pdflib-lite"
 
+    if build.without? "gd"
+      args << "--without-gd"
```
ilovezfs: this can go on one line
```
args << "--without-gd" if build.without? "gd"
```

ilovezfs (approved):

MikeMcQuaid:

PR #9414, opened 2017-01-29
llvm@3.9 (new formula)

tmcdonell:

-----

Versioned install of LLVM-3.9. Useful for people who need to install multiple versions of LLVM simultaneously (like me), or who need to install a specific version (also useful in documentation).

LLVM-4.0 is currently in RC and scheduled to be released in a few days. Once that is out we can bump the `llvm.rb` formula and LLVM-3.9 will still be available via this formula.

I included the fix to `llvm@3.8` in this PR, since that was causing problems installing multiple versions simultaneously, but I could probably split that out. I currently have `llvm@3.7`, `llvm@3.8`, and `llvm@3.9` formula installed and they seem to be working together nicely.

### DIFF
--- a/Formula/llvm@3.8.rb
+++ b/Formula/llvm@3.8.rb
@@ -180,10 +180,11 @@ class LlvmAT38 < Formula
 
     system "make", "-C", libcxx_buildpath, "install", *libcxx_make_args
 
-    (share/"clang/tools").install Dir["tools/clang/tools/scan-{build,view}"]
-    inreplace share/"clang/tools/scan-build/bin/scan-build", "$RealBin/bin/clang", bin/"clang-#{ver}"
-    bin.install_symlink share/"clang/tools/scan-build/bin/scan-build", share/"clang/tools/scan-view/bin/scan-view"
-    man1.install_symlink share/"clang/tools/scan-build/man/scan-build.1"
+    (share/"clang-#{ver}/tools").install Dir["tools/clang/tools/scan-{build,view}"]
+    inreplace share/"clang-#{ver}/tools/scan-build/bin/scan-build", "$RealBin/bin/clang", install_prefix/"bin/clang"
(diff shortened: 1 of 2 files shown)


ilovezfs:
The current formula is already 3.9.x, so at most we'd consider adding an alias.

tmcdonell:
@ilovezfs As I mentioned in the comment, LLVM is about to land 4.0, so this is still necessary. At worst I am a few days early with the PR.

Additionally, for my own use can, LLVM is one of the dependencies of my software and it requires a specific version to be installed. It is much easier to write in the documentation/installation instructions `brew install llvm@3.9`; rather than `brew install llvm; brew link --force llvm` and hope for the best. I think this is a valid use case.

ilovezfs:
which is why I said we'd accept a PR for an alias

ilovezfs:
See https://github.com/Homebrew/homebrew-core/pull/8945.

MikeMcQuaid:
> I included the fix to llvm@3.8 in this PR, since that was causing problems installing multiple versions simultaneously, but I could probably split that out.

Yeh, a separate PR would be best for this, thanks.

> As I mentioned in the comment, LLVM is about to land 4.0, so this is still necessary. At worst I am a few days early with the PR.

I'd recommend creating a PR for both LLVM 4.0 and LLVM@3.9 in the same one.

tmcdonell:
@MikeMcQuaid sure, I'll create a separate PR for that.

I appreciate your courtesy to actually read the commit message. Thanks.

ilovezfs:
Cute.

MikeMcQuaid:
> I appreciate your courtesy to actually read the commit message. Thanks.

GitHub's UI is PR-focused rather than commit-focused and Homebrew's maintainers are busy people so it's best to put information you need to be read in the PR message as well as the commit message.

PR #9400, opened 2017-01-28
New formula: telegram-cli

alexreg:
Test not included because any worthwhile test is non-trivial.

PR #9422, opened 2017-01-29
New formula: telegram-cli (UPDATED)

alexreg:
Sorry, a new pull-request since I forgot to create a separate branch on my homebrew-core fork. https://github.com/Homebrew/homebrew-core/pull/9400 can be ignored now.


Formula/telegram-cli.rb
```diff
@@ -0,0 +1,60 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "readline"
+  depends_on "libevent"
+  depends_on "openssl"
+  depends_on "libconfig" => :recommended
+  depends_on "jansson" => :recommended
+  depends_on "lua" => :recommended
+  depends_on "python" => :recommended
+
+  patch :DATA
```
MikeMcQuaid: > We’ll keep the patch for now, then remove it when/if it’s accepted upstream, if that’s okay?

Still need a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Would probably also like to see it merged first.

Formula/telegram-cli.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def install
+    args = %W[
+      --prefix=#{prefix}
+      CFLAGS=-I#{Formula["readline"].include}
+      CPPFLAGS=-I#{Formula["readline"].include}
+      LDFLAGS=-L#{Formula["readline"].lib}
+    ]
```
MikeMcQuaid: > The `configure` script is doing something weird, as @tdsmith and I diagnosed yesterday on IRC. If they’re not included, then it picks up the system *readline* library, which is a problem, because it lacks some required feature.

Please file an upstream issue about this, too, thanks.

Formula/telegram-cli.rb
```diff
@@ -0,0 +1,60 @@
...
+  head "https://github.com/vysheng/tg.git"
+
+  depends_on "pkg-config" => :build
+  depends_on "readline"
+  depends_on "libevent"
+  depends_on "openssl"
+  depends_on "libconfig" => :recommended
+  depends_on "jansson" => :recommended
+  depends_on "lua" => :recommended
+  depends_on "python" => :recommended
```
MikeMcQuaid: > It’s nice to have the option though, isn’t it? I mean, they’re already on by default (they should be —without-* options). Maybe libconfig I can make essential, if you prefer.

We prefer to remove options unless they provide really useful functionality. Removing dependencies isn't one, really.

MikeMcQuaid (changes requested):

alexreg:
> On 29 Jan 2017, at 17:49, Mike McQuaid <notifications@github.com> wrote:
> 
> @MikeMcQuaid requested changes on this pull request.
> 
> In Formula/telegram-cli.rb <https://github.com/Homebrew/homebrew-core/pull/9422#pullrequestreview-18998086>:
> 
> > +  # Use the tag instead of the tarball to get submodules
> +  url "https://github.com/vysheng/tg.git",
> +      :tag => "1.3.1",
> +      :revision => "5935c97ed05b90015418b5208b7beeca15a6043c"
> +  head "https://github.com/vysheng/tg.git"
> +
> +  depends_on "pkg-config" => :build
> +  depends_on "readline"
> +  depends_on "libevent"
> +  depends_on "openssl"
> +  depends_on "libconfig" => :recommended
> +  depends_on "jansson" => :recommended
> +  depends_on "lua" => :recommended
> +  depends_on "python" => :recommended
> +
> +  patch :DATA
> We’ll keep the patch for now, then remove it when/if it’s accepted upstream, if that’s okay?
> 
> Still need a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Would probably also like to see it merged first.
> 
Patch explanation added.

Here’s the submission upstream: https://github.com/vysheng/tg/issues/1296 <https://github.com/vysheng/tg/issues/1296>

The maintainer doesn’t seem particularly active, so it may not get merged for a while. Your call though.

> 
> In Formula/telegram-cli.rb <https://github.com/Homebrew/homebrew-core/pull/9422#pullrequestreview-18998086>:
> 
> > +  depends_on "libevent"
> +  depends_on "openssl"
> +  depends_on "libconfig" => :recommended
> +  depends_on "jansson" => :recommended
> +  depends_on "lua" => :recommended
> +  depends_on "python" => :recommended
> +
> +  patch :DATA
> +
> +  def install
> +    args = %W[
> +      --prefix=#{prefix}
> +      CFLAGS=-I#{Formula["readline"].include}
> +      CPPFLAGS=-I#{Formula["readline"].include}
> +      LDFLAGS=-L#{Formula["readline"].lib}
> +    ]
> The configure script is doing something weird, as @tdsmith <https://github.com/tdsmith> and I diagnosed yesterday on IRC. If they’re not included, then it picks up the system readline library, which is a problem, because it lacks some required feature.
> 
> Please file an upstream issue about this, too, thanks.
> 
https://github.com/vysheng/tg/issues/1298

> 
> In Formula/telegram-cli.rb <https://github.com/Homebrew/homebrew-core/pull/9422#pullrequestreview-18998086>:
> 
> > +  desc "Command-line interface for Telegram."
> +  homepage "https://github.com/vysheng/tg"
> +  # Use the tag instead of the tarball to get submodules
> +  url "https://github.com/vysheng/tg.git",
> +      :tag => "1.3.1",
> +      :revision => "5935c97ed05b90015418b5208b7beeca15a6043c"
> +  head "https://github.com/vysheng/tg.git"
> +
> +  depends_on "pkg-config" => :build
> +  depends_on "readline"
> +  depends_on "libevent"
> +  depends_on "openssl"
> +  depends_on "libconfig" => :recommended
> +  depends_on "jansson" => :recommended
> +  depends_on "lua" => :recommended
> +  depends_on "python" => :recommended
> It’s nice to have the option though, isn’t it? I mean, they’re already on by default (they should be —without-* options). Maybe libconfig I can make essential, if you prefer.
> 
> We prefer to remove options unless they provide really useful functionality. Removing dependencies isn't one, really.
> 

The software can still be completely functional without Python and Lua, and many may not want to pull in those big formulas, so I’ve left these as recommended, which I think is fair. I’ve made *libconfig* and *jansson* normal dependencies however. Fair enough?

Formula/telegram-cli.rb
tdsmith: What do you expect should happen? I think this is masking a superenv flag filtering bug, actually, but I didn't dig into it.

Formula/telegram-cli.rb
```diff
@@ -0,0 +1,61 @@
...
+  head "https://github.com/vysheng/tg.git"
+
+  depends_on "pkg-config" => :build
+  depends_on "readline"
+  depends_on "libevent"
+  depends_on "openssl"
+  depends_on "libconfig"
+  depends_on "jansson"
+  depends_on "lua" => :recommended
+  depends_on "python" => :recommended
```
tdsmith: what features are enabled/disabled without lua and python?

alexreg:
Libraries for scripting a Telegram client, I believe.

Formula/telegram-cli.rb
tdsmith: okay, the problem is:
- the zlib check happens before the readline check
- the zlib check is dumb and adds `-I/usr/include` to CPPFLAGS
- SDKROOT is set, so /usr/include is not in the compiler's default search list, so -I/usr/include actually adds /usr/include to the top of the search list
- -I/usr/include appears on the command line before our -I flag for Homebrew's readline appears, so configure gets the wrong readline and barfs

We don't filter any flags for configure invocations so this is actually kind of annoying to fix.

MikeMcQuaid:
@alexreg Please use the GitHub UI rather than replying to emails inline. It's really time consuming to read your replies, thanks.

Formula/telegram-cli.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  # Look for the configuration file under /usr/local/etc rather than /etc on OS X.
```
MikeMcQuaid: Again: **add a link to the upstream patch submission** in a comment in the formula file. Thanks!

Formula/telegram-cli.rb
```diff
@@ -0,0 +1,61 @@
...
+  def install
+    args = %W[
+      --prefix=#{prefix}
+      CFLAGS=-I#{Formula["readline"].include}
+      CPPFLAGS=-I#{Formula["readline"].include}
+      LDFLAGS=-L#{Formula["readline"].lib}
+    ]
+
+    args << "--disable-libconfig" if build.without? "libconfig"
+    args << "--disable-json" if build.without? "jansson"
```
MikeMcQuaid: These are no longer optional

MikeMcQuaid:
> The software can still be completely functional without Python and Lua, and many may not want to pull in those big formulas, so I’ve left these as recommended, which I think is fair. I’ve made *libconfig* and *jansson* normal dependencies however. Fair enough?

If you don't personally have use for these dependencies then `:optional` feels better. Otherwise, that's fine.

Formula/telegram-cli.rb
MikeMcQuaid: @alexreg This isn't a Homebrew bug. Upstream should allow specifying readline with `--with-readline=` or similar in configure.

Formula/telegram-cli.rb
MikeMcQuaid: Also, please submit a pull request upstream rather than a patch as an issue.

MikeMcQuaid:
> The maintainer doesn’t seem particularly active, so it may not get merged for a while. Your call though.

That doesn't bode terribly well for this being merged so we'll wait for now.

Formula/telegram-cli.rb
tdsmith: It looks like pretty stock autoconf; it would have worked before we started setting SDKROOT more aggressively.

One way around is to set `--with-zlib=#{MacOS.sdk_path}/usr/include`, probably?

MikeMcQuaid: 👍 to try that. `--with-readline` is used in quite a few formulae, though. If it's stock autoconf then maybe that would work?

Formula/telegram-cli.rb
alexreg: Ah, you didn't say you wanted it in the file. Understood.

Formula/telegram-cli.rb
MikeMcQuaid: @alexreg ping on this

Formula/telegram-cli.rb
MikeMcQuaid: @alexreg any news on the pull request?

Formula/telegram-cli.rb
MikeMcQuaid: @alexreg any thoughts on this?

Formula/telegram-cli.rb
alexreg: Afraid not... maintainer a bit inactive, as I suspected.

Formula/telegram-cli.rb
alexreg: Oh, I could swear I replied, sorry. Just the client libraries, for people plugging into the interface with their own code.

Formula/telegram-cli.rb
alexreg: @MikeMcQuaid No luck I'm afraid. I've tried putting in both the above `--with-zlib` and `--with-readline=#{Formula["readline"].opt_prefix}`, but the same error as previously discussed.

Formula/telegram-cli.rb
MikeMcQuaid: @alexreg Sorry, I mean: did you submit one? I'm just seeing a patch inline in an issue.

alexreg: Okay, a PR is submitted, and the comment in the formula links to it now. Phew, a lot of work for a little patch! :) Hope all is in order to merge now.

MikeMcQuaid:
Test is failing: https://bot.brew.sh/job/Homebrew%20Core/16220/version=sierra/testReport/junit/brew-test-bot/sierra/test_telegram_cli/

Formula/telegram-cli.rb
```diff
@@ -0,0 +1,60 @@
+class TelegramCli < Formula
+  desc "Command-line interface for Telegram."
+  homepage "https://github.com/vysheng/tg"
+  # Use the tag instead of the tarball to get submodules
```
MikeMcQuaid: Can omit this comment

Formula/telegram-cli.rb
MikeMcQuaid: Let's default them to off.

Formula/telegram-cli.rb
MikeMcQuaid: May be worth using https://github.com/vysheng/tg/pull/1306.patch as a patch URL. Also worth noting CI seems to be failing on that PR; may be worth seeing if that's the case on master (not blocking this though)

Formula/telegram-cli.rb
MikeMcQuaid: Thanks for trying.

MikeMcQuaid (changes requested):

Formula/telegram-cli.rb
alexreg: I presume you mean the comment? I saw it in a lot of formulas, but gladly done.

alexreg:
@MikeMcQuaid Okay, those amendments should be pushed now. :)

alexreg:
Oh, and the PR on *tg* is failing even on the master branch – I guess it would be better to be against the tagged release, for you guys, although not for the project maintainer...

MikeMcQuaid:
> Oh, and the PR on tg is failing even on the master branch – I guess it would be better to be against the tagged release, for you guys, although not for the project maintainer...

Good to know, thanks.

Test is still failing and audit is too.

alexreg:
The audit is fixed now. As for the test, I'm not quite sure what's going on. Maybe you have a (short) timeout on how long it should run for? Here's the console output: https://gist.github.com/603d1d22bdb0eea678b0619557b976fb

Alternatively, perhaps there could be a problem with the command returning an exit code of 1? (I know, it's rather odd behaviour.)

alexreg:
Okay, ignore the above about the test case and the exit code of 1, that's fixed now. Hopefully ready to merge!

Formula/telegram-cli.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    system "./configure", *args
+    system "make"
+
+    bin.install "bin/telegram-cli" => "telegram"
+    (etc/"telegram-cli").install "server.pub"
+  end
+
+  test do
+    shell_output "#{bin}/telegram -h", 1
```
MikeMcQuaid: Use `assert_match` here too match something in the output. This should be the last thing.

MikeMcQuaid (changes requested):

Formula/telegram-cli.rb
alexreg: Done. Uff, hope so! This is exhausting. ;)

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again for your contribution to Homebrew and patience! Without people like you submitting PRs we couldn't run this project. You rock!

PR #9428, opened 2017-01-30
Actually use readline when enabled, add --with-editline as alternative

decklin:
- [ ] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)? No, there is a warning about the universal option, but that was there already.

-----

I noticed that the `sqlite3` terminal ui was using editline, not Readline. You can verify this with:

- `otool -L /usr/local/Cellar/sqlite/3.16.2/bin/sqlite3`
- editline does not have reverse-incremental-search, so pressing ctrl-R does nothing

I poked around and found this comment in the configure script:

    #-------------------------------------------------------------------------
    # Two options to enable readline compatible libraries:
    #
    #   --enable-editline
    #   --enable-readline
    #
    # Both are enabled by default. If, after command line processing both are
    # still enabled, the script searches for editline first and automatically
    # disables readline if it is found. So, to use readline explicitly, the
    # user must pass "--disable-editline". To disable command line editing
    # support altogether, "--disable-editline --disable-readline".

After peeling my hand off my forehead, I implemented this and was able to rebuild the formula with Readline linked in (`otool` reports `/usr/local/opt/readline/lib/libreadline.7.dylib`).

Building with Readline is still the default, but now it works. I added the option `--with-editline`, which will cause the with Readline setting to be ignored in favor of editline. If you for some reason build with `--without-readline` but don't explicitly enable editline, configure will still find editline, which I think is fine.

Important question before merging this: users probably have existing `~/.sqlite_history` files. editline and Readline use different formats for this file, so if you installed this upgrade without a backup of your history file, you could lose that data. Would adding a caveat be appropriate/required/enough for this? It could say something along the lines of:

    To convert your existing history file so that it can be used by Readline, run:
      sed -i~ 's/\\040/ /g' ~/.sqlite_history

Which seems to have been sufficient for me.

### DIFF
--- a/Formula/sqlite.rb
+++ b/Formula/sqlite.rb
@@ -67,9 +67,14 @@ class Sqlite < Formula
 
     ENV.universal_binary if build.universal?
 
-    system "./configure", "--prefix=#{prefix}",
-                          "--disable-dependency-tracking",
-                          "--enable-dynamic-extensions"
+    args = [
+      "--prefix=#{prefix}",
+      "--disable-dependency-tracking",
+      "--enable-dynamic-extensions",
+    ]
+    args << "--enable-readline" << "--disable-editline" if build.with? "readline"
+
+    system "./configure", *args
     system "make", "install"
 
     if build.with? "functions"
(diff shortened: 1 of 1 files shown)


Formula/sqlite.rb
```diff
@@ -16,6 +16,7 @@ class Sqlite < Formula
 
   option :universal
   option "with-docs", "Install HTML documentation"
+  option "with-editline", "Use editline (note: will override readline)"
```
MikeMcQuaid: Let's punt on this option and just always use readline, thanks.

MikeMcQuaid (changes requested):

decklin:
Sure, sounds good. I will also take a shot at adding the caveat.

decklin:
OK, ready. `?w=1` on the diff will suppress the noise from indenting the existing caveat.

To test the runtime checks in the new caveat, I rebuilt and reinstalled, followed the instructions to run sed, reinstalled again, and verified that the instructions were not shown again.

Building `--without-readline` works as expected.

The one thing to perhaps worry about with history is if you accidentally run Apple's `/usr/bin/sqlite3` or another non-Readline version, it'll fail to read your Readline history. But I think this is a pretty rare situation and will probably only be encountered by expert users.

Formula/sqlite.rb
```diff
@@ -67,9 +67,16 @@ def install
...
-    system "./configure", "--prefix=#{prefix}",
-                          "--disable-dependency-tracking",
-                          "--enable-dynamic-extensions"
+    args = [
+      "--prefix=#{prefix}",
+      "--disable-dependency-tracking",
+      "--enable-dynamic-extensions",
+    ]
+    if build.with? "readline"
+      args += ["--enable-readline", "--disable-editline"]
```
MikeMcQuaid: `args << "--enable-readline" << "--disable-editline" if build.with? "readline"`

Formula/sqlite.rb
```diff
@@ -85,25 +92,45 @@ def install
...
+          If the program is built so that loading extensions is permitted,
+          the following will work:
+           sqlite> SELECT load_extension('#{lib}/libsqlitefunctions.dylib');
+           sqlite> select cos(radians(45));
+           0.707106781186548
       EOS
     end
+    if build.with? "readline"
+      user_history = "~/.sqlite_history"
+      hist_path = File.expand_path(user_history)
```
MikeMcQuaid: `user_history_path`

MikeMcQuaid (changes requested):

MikeMcQuaid:
> To test the runtime checks in the new caveat, I rebuilt and reinstalled, followed the instructions to run sed, reinstalled again, and verified that the instructions were not shown again.

@decklin Just to be super explicit: you saw those instructions displayed and then not displayed when rerun? We override `HOME` at various times so wanted to make sure this wasn't one of them. Thanks!

decklin:
Sorry for the delay. Correct, I saw them displayed, actually edited the history file to remove `\040`, rerun, then saw them not displayed. Which was as intended.

Formula/sqlite.rb
decklin: ok

Formula/sqlite.rb
decklin: sure

MikeMcQuaid (approved):

MikeMcQuaid:

For future reference the preferred commit message format is `sqlite: fix readline/editline handling.`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

decklin:
Thanks, will do in the future.

PR #9444, opened 2017-01-30
ice: add python support

AndiDog:

-----

This makes PHP support optional and adds Python support. Please check if I'm doing the Python executable/library/site-packages selection right - I basically followed `audit` output instead of hardcoded linkage against a certain libpython.

@externl FYI ;)

### DIFF
--- a/Formula/ice.rb
+++ b/Formula/ice.rb
@@ -18,12 +18,14 @@ class Ice < Formula
 
   option "with-additional-compilers", "Build additional Slice compilers (slice2py, slice2js, slice2rb)"
   option "with-java", "Build Ice for Java and the IceGrid GUI app"
+  option "without-python", "Build without Ice for Python"
   option "without-xcode-sdk", "Build without the Xcode SDK for iOS development (includes static libs)"
 
   depends_on "mcpp"
   depends_on "lmdb"
   depends_on :java => ["1.8+", :optional]
   depends_on :macos => :mavericks
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
 
   def install
     # Ensure Gradle uses a writable directory even in sandbox mode
@@ -37,8 +39,17 @@ class Ice < Formula
       "CONFIGS=shared cpp11-shared #{build.with?("xcode-sdk") ? "xcodesdk cpp11-xcodesdk" : ""}",
       "PLATFORMS=all",
       "SKIP=slice2confluence #{build.without?("additional-compilers") ? "slice2py slice2rb slice2js" : ""}",
-      "LANGUAGES=cpp objective-c #{build.with?("java") ? "java java-compat" : ""}",
+      "LANGUAGES=cpp objective-c #{build.with?("java") ? "java java-compat" : ""} #{build.with?("python") ? "python" : ""}",
     ]
+
+    if build.with? "python"
+      args << "PYTHON_LIB_NAME=-Wl,-undefined,dynamic_lookup"
+      cd "python" do
+        inreplace "config/install_dir", "print(e.install_dir)", "print('#{lib}/python2.7/site-packages')"
+        inreplace "config/Make.rules", /^python_ldflags\s*:=\s*-L\$\(PYTHON_LIB_DIR\) -l\$\(PYTHON_LIB_NAME\)$/, "python_ldflags := -Wl,-undefined,dynamic_lookup"
+      end
+    end
+
     system "make", "install", *args
   end
 
@@ -76,5 +87,8 @@ class Ice < Formula
     system ENV.cxx, "-DICE_CPP11_MAPPING", "-std=c++11", "-c", "-I#{include}", "-I.", "Test.cpp"
     system ENV.cxx, "-L#{lib}", "-o", "test", "Test.o", "Hello.o", "-lIce++11"
     system "./test"
+    if build.with? "python"
+      system "python", "-c", "import Ice; Ice.initialize().destroy()"
+    end
   end
 end


externl:
I'm not sure I see the benefit of adding optional flags for python on php. Both interpreters are installed by default with the command line tools / Xcode and their build times are miniscule.

What is homebrew's default stance on this?

AndiDog:
You're actually making a good point here.

Here's the battle:
```
/path/to/homebrew-core/Formula$ ag -l without-python | wc -l ; ag -l with-python | wc -l
      25
      41
```

which... umm... doesn't tell me anything 👅  What about I just change the formula to use `--without-xyz` options and by default include all the scripting languages? Makes sense for me 👍

externl:
TBO I'm still not sure I understand the need for any `--with[out]` flags for python or php.

Formula/ice.rb
```diff
@@ -12,6 +12,8 @@ class Ice < Formula
   end
 
   option "with-java", "Build Ice for Java and the IceGrid Admin app"
+  option "with-php", "Build Ice for PHP"
+  option "with-python", "Build Ice for Python"
```
MikeMcQuaid: Use `depends_on :python => :optional`

Formula/ice.rb
```diff
@@ -87,10 +89,25 @@ def install
...
+    if build.with? "php"
+      cd "php" do
+        args << "install_phpdir=#{share}/php"
+        args << "install_libdir=#{lib}/php/extensions"
+        system "make", "install", *args
+      end
+    end
+
+    if build.with? "python"
+      python = PythonRequirement.new
```
MikeMcQuaid: don't need to use this once the `depends_on` is done above as it'll be first in the PATH.

Formula/ice.rb
```diff
@@ -127,8 +144,13 @@ class HelloI : public Test::Hello {
...
-    system "/usr/bin/php", "-d", "extension_dir=#{lib}/php/extensions",
-                           "-d", "extension=IcePHP.dy",
-                           "-r", "extension_loaded('ice') ? exit(0) : exit(1);"
+    if build.with? "php"
+      system "/usr/bin/php", "-d", "extension_dir=#{lib}/php/extensions",
+                             "-d", "extension=IcePHP.dy",
+                             "-r", "extension_loaded('ice') ? exit(0) : exit(1);"
+    end
+    if build.with? "python"
+      system PythonRequirement.new.which_python.to_s, "-c", "import Ice; Ice.initialize().destroy()"
```
MikeMcQuaid: Just use `"python"` rather than `which_python`

MikeMcQuaid (dismissed):

externl:
I think `without` is preferable to `with` for php and python.

Formula/ice.rb
AndiDog: With

```
option "without-python", "Build without Ice for Python"
depends_on :python => :optional
```

`brew info ice` shows both `--with-python` and `--without-python`... With the workaround `depends_on :python => :optional if build.without?("python")`, only `--without-python` is shown. Is that a good way to do it?

JCount: `depends_on :python => :optional if build.without?("python")` is definitely not a good way to do it. If you think it should be built with python by default, you should use `depends_on :python => :recommended` which will automatically produce a `--without-python` option.

AndiDog:
Changed to `without` semantics and `depends_on :python => :recommended`. Please check if the `inreplace` use is okay. It doesn't work without since else the Ice script `python/config/install_dir` finds the global site-packages dir instead of the one of the sandbox.

Formula/ice.rb
```diff
@@ -12,10 +12,13 @@ class Ice < Formula
   end
 
   option "with-java", "Build Ice for Java and the IceGrid Admin app"
+  option "without-php", "Build without Ice for PHP"
+  option "without-python", "Build without Ice for Python"
 
   depends_on "mcpp"
   depends_on :java => ["1.7+", :optional]
   depends_on :macos => :mavericks
+  depends_on :python => :recommended
```
ilovezfs: you should be able to change this to `depends_on :python => :recommended if MacOS.version <= :snow_leopard`

Formula/ice.rb
```diff
@@ -87,10 +90,22 @@ def install
...
-      system "make", "install", *args
+    if !build.without? "php"
+      cd "php" do
+        args << "install_phpdir=#{share}/php"
+        args << "install_libdir=#{lib}/php/extensions"
+        system "make", "install", *args
+      end
+    end
+
+    if !build.without? "python"
```
ilovezfs: if build.with? "python"

Formula/ice.rb
```diff
@@ -127,8 +142,13 @@ class HelloI : public Test::Hello {
...
     system "./test", "--Ice.InitPlugins=0"
-    system "/usr/bin/php", "-d", "extension_dir=#{lib}/php/extensions",
-                           "-d", "extension=IcePHP.dy",
-                           "-r", "extension_loaded('ice') ? exit(0) : exit(1);"
+    if !build.without? "php"
+      system "/usr/bin/php", "-d", "extension_dir=#{lib}/php/extensions",
+                             "-d", "extension=IcePHP.dy",
+                             "-r", "extension_loaded('ice') ? exit(0) : exit(1);"
+    end
+    if !build.without? "python"
```
ilovezfs: if build.with? "python"

AndiDog:
Remaining lints fixed

AndiDog:
Didn't see latest review comment - trying `depends_on :python => :recommended if MacOS.version <= :snow_leopard` now and then update again.

AndiDog:
All review comments resolved, no more audit lints now. I could only test build on Sierra, and works fine.

AndiDog:
Can you help with the build failure output? `brew test ice` is working fine for me.

MikeMcQuaid:
@AndiDog Perhaps you have PYTHONPATH or similar set locally or are using the non-system Python? Setting `PYTHONPATH` in the test feels right but probably indicates this should only be build optionally with Python support to avoid system/Homebrew Python decision issues.

AndiDog:
The missing piece for me is to understand what `depends_on :python` is actually doing? Does it pull in homebrew-python as dependency? Or will it choose homebrew-python (if installed) or system python (fallback) and set up PATH/PYTHONPATH accordingly?

I can reproduce the test failure, but to fix it, I need that understanding.

Since Ice builds native modules, my preference would be to choose homebrew-python if installed, and otherwise build the modules for system python.

AndiDog:
How should I continue on this matter? Any help on the above questions? I'd _like_ to support both system and brew python, but don't know how.

AndiDog:
I managed to get the tests working (just a mistake on my side related to make arguments). Should be okay to merge now if it passes code review and automated build.

JCount:
@BrewTestBot test this please

ilovezfs (approved):

ilovezfs (dismissed):
please resolve the conflicts with master and force push

externl:
FYI @AndiDog, I plan on making PRs to the homebrew-php with separate php fromulae for Ice 3.7.

AndiDog:
I saw PHP was removed, anyway, so I'm only going to add Python support. Will make it optional because also slice2py and friends are optional by now. Currently I'm building my change and will then push the modified commit.

externl:
@AndiDog Cool. We made them optional because installing the gem or pypi package destroys `slice2py` and `slice2rb`. All the rest are included. 

Will your Python support also install `slice2py`?

AndiDog:
... actually we said that `without` makes more sense for Python, so I left it so. Let's see if this builds fine. My Python support should not install `slice2py` by default - that is decided by `--with-additional-compilers` from what I can see.

externl:
Ok, that makes sense.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
```
mkdir -p ../python/python
clang++ -dynamiclib  -current_version 3.7.0 -compatibility_version 3.7.0 -o ../python/python/IcePy.so modules/IcePy/build/macosx/shared/pic/Checksum.o modules/IcePy/build/macosx/shared/pic/FileTracker.o modules/IcePy/build/macosx/shared/pic/Grammar.o modules/IcePy/build/macosx/shared/pic/MD5.o modules/IcePy/build/macosx/shared/pic/MD5I.o modules/IcePy/build/macosx/shared/pic/Parser.o modules/IcePy/build/macosx/shared/pic/Preprocessor.o modules/IcePy/build/macosx/shared/pic/Python.o modules/IcePy/build/macosx/shared/pic/Scanner.o modules/IcePy/build/macosx/shared/pic/SliceUtil.o modules/IcePy/build/macosx/shared/pic/PythonUtil.o modules/IcePy/build/macosx/shared/pic/StringLiteralUtil.o modules/IcePy/build/macosx/shared/pic/BatchRequestInterceptor.o modules/IcePy/build/macosx/shared/pic/Communicator.o modules/IcePy/build/macosx/shared/pic/Connection.o modules/IcePy/build/macosx/shared/pic/ConnectionInfo.o modules/IcePy/build/macosx/shared/pic/Current.o modules/IcePy/build/macosx/shared/pic/Dispatcher.o modules/IcePy/build/macosx/shared/pic/Endpoint.o modules/IcePy/build/macosx/shared/pic/EndpointInfo.o modules/IcePy/build/macosx/shared/pic/ImplicitContext.o modules/IcePy/build/macosx/shared/pic/Init.o modules/IcePy/build/macosx/shared/pic/Logger.o modules/IcePy/build/macosx/shared/pic/ObjectAdapter.o modules/IcePy/build/macosx/shared/pic/Operation.o modules/IcePy/build/macosx/shared/pic/Properties.o modules/IcePy/build/macosx/shared/pic/PropertiesAdmin.o modules/IcePy/build/macosx/shared/pic/Proxy.o modules/IcePy/build/macosx/shared/pic/Slice.o modules/IcePy/build/macosx/shared/pic/Thread.o modules/IcePy/build/macosx/shared/pic/Types.o modules/IcePy/build/macosx/shared/pic/Util.o modules/IcePy/build/macosx/shared/pic/ValueFactoryManager.o  -mmacosx-version-min=10.9 -arch x86_64 -Wl,-rpath,@loader_path/../../cpp/lib -L../cpp/lib -lIceDiscovery -lIceLocatorDiscovery -lIceSSL -lIce -L/usr/local/opt/mcpp/lib -lmcpp -Wl,-undefined,dynamic_lookup
install -p ../python/python/IcePy.so /usr/local/Cellar/ice/3.7.0/lib/python2.7/site-packages
chmod a+rx /usr/local/Cellar/ice/3.7.0/lib/python2.7/site-packages/IcePy.so
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C python install
mkdir -p ../../python/python/Ice/.depend
make[2]: *** No rule to make target `/slice2py', needed by `../../python/python/Ice/BuiltinSequences_ice.py'.  Stop.
make[1]: *** [install] Error 2
make: *** [install] Error 1
```

ilovezfs:
@AndiDog the build failed with the above error.

JCount:
Pruning this due to the lack of a response and activity in general. Please comment if you would like to reopen and address the needed changes.

AndiDog:
I have to test the different combinations, and it seems the default options don't build. Due to missing ccache support, and since I'm doing this on the side, this will take a while... will reopen when I found a solution, thx.

AndiDog:
I tested both the standard build and the one with Python support. Changes pushed to my branch. Ready to reopen and have the bot test it as well.

JCount:
@AndiDog You'll need to open a new PR because it seems that I can't reopen this one. (The UI has a message about your branch being force-pushed or recreated.)

PR #9456, opened 2017-01-30
watchman: launchctl unload post-install

wez:
This is taking a run at: https://github.com/facebook/watchman/issues/358

Watchman installs and updates its own launchd.plist file on behalf of the user; it is not integrated with `brew services`.
In the linked issue, watchman would stop working for some homebrew users.  The problem was that the launchd.plist used socket activation to set up the service, and because launchd is maintaining the socket, and the socket connection attempt just hangs, watchman never had a chance to fix up the plist and things stayed broken.

To help solve this, we're forcing an unload in the post-install portion of the recipe.

-----

### DIFF
--- a/Formula/watchman.rb
+++ b/Formula/watchman.rb
@@ -42,6 +42,17 @@ class Watchman < Formula
   def post_install
     (var/"run/watchman").mkpath
     chmod 042777, var/"run/watchman"
+    # Older versions would use socket activation in the launchd.plist, and since
+    # the homebrew paths are versioned, this meant that launchd would continue
+    # to reference the old version of the binary after upgrading.
+    # https://github.com/facebook/watchman/issues/358
+    # To help swing folks from an older version to newer versions, force unloading
+    # the plist here.  This is needed even if someone wanted to add brew services
+    # support; while there are still folks with watchman <4.8 this is still an
+    # upgrade concern.
+    home = ENV["HOME"]
+    system "launchctl", "unload",
+           "-F", "#{home}/Library/LaunchAgents/com.github.facebook.watchman.plist"
   end
 
   test do


MikeMcQuaid:
So, to confirm I've understood,  `watchman` did it's own `launchctl` setup and this was broken so you're now removing it on post-install? How will this be fixed for existing installs inside or outside of Homebrew? Is it worth perhaps integrating `watchman` with Homebrew Services at the same time?

wez:
Watchman does its own launctl setup, but because of a bug in launchd with socket activation (if you remove the binary referenced by the plist, launch will just leave the socket unconnected with no error response), combined with a bug in launchtl with not returning an error code on failure (the man page hilariously describes this as a backwards compatibility concern), combined with the strongly versioned paths in homebrew, homebrew users are sometimes left with a hung watchman instance after upgrading from an older watchman version.

Watchman 4.8 (which we're going to tag next week) removes the socket activation component for new installs, but will also have to deal with this upgrade concern before that takes effect.  In the meantime (until all homebrew users are on 4.8+) we need homebrew to give us a signal when the installed version changes; the easiest way to do this is a post install script to unregister the plist deployed by watchman.

It is technically possible to configure watchman to not do its own launchd stuff, but it means deploying a global configuration file and a helper script that will cause the service to start at the appropriate time.  That's something that I'd like to see happen, but not something I have bandwidth to tackle at the moment, and we'd probably still want to do something like the post install step to swing users over to the brew services approach.

Does that make sense?

wez:
Friendly ping on this.  I've got the 4.8 version update queued up behind this one :)

MikeMcQuaid:
@wez Pulling this temporarily but just FYI that it feels a little weird to fix non-Homebrew-specific issues like this. Thanks anyway.

PR #9472, opened 2017-01-31
Lua: attempt to remove patching.

MikeMcQuaid:
Let's see how badly this explodes and try to remove the patches that relate to different versions of Lua. As-is Lua is in [violation of our docs here](https://github.com/Homebrew/brew/blob/master/docs/Versions.md) because it requires buildsystem patching to maintain side-by-side versions.

### DIFF
--- a/Formula/lua.rb
+++ b/Formula/lua.rb
@@ -3,7 +3,7 @@ class Lua < Formula
   homepage "https://www.lua.org/"
   url "https://www.lua.org/ftp/lua-5.2.4.tar.gz"
   sha256 "b9e2e4aad6789b3b63a056d442f7b39f0ecfca3ae0f1fc0ae4e9614401b69f4b"
-  revision 4
+  revision 5
 
   bottle do
     cellar :any
@@ -23,10 +23,6 @@ class Lua < Formula
   option "with-completion", "Enables advanced readline support"
   option "without-sigaction", "Revert to ANSI signal instead of improved POSIX sigaction"
   option "without-luarocks", "Don't build with Luarocks support embedded"
-
-  # Be sure to build a dylib, or else runtime modules will pull in another static copy of liblua = crashy
-  # See: https://github.com/Homebrew/homebrew/pull/5043
-  patch :DATA
 
   # completion provided by advanced readline power patch
   # See http://lua-users.org/wiki/LuaPowerPatches
@@ -55,7 +51,6 @@ class Lua < Formula
     # Subtitute formula prefix in `src/Makefile` for install name (dylib ID).
     # Use our CC/CFLAGS to compile.
     inreplace "src/Makefile" do |s|
-      s.gsub! "@LUA_PREFIX@", prefix
       s.remove_make_var! "CC"
       s.change_make_var! "CFLAGS", "#{ENV.cflags} -DLUA_COMPAT_ALL $(SYSCFLAGS) $(MYCFLAGS)"
       s.change_make_var! "MYLDFLAGS", ENV.ldflags
@@ -68,15 +63,6 @@ class Lua < Formula
     system "make", "macosx", "INSTALL_TOP=#{prefix}", "INSTALL_MAN=#{man1}"
     system "make", "install", "INSTALL_TOP=#{prefix}", "INSTALL_MAN=#{man1}"
     (lib/"pkgconfig/lua.pc").write pc_file
-
-    # Fix some software potentially hunting for different pc names.
-    bin.install_symlink "lua" => "lua5.2"
-    bin.install_symlink "lua" => "lua-5.2"
-    bin.install_symlink "luac" => "luac5.2"
-    bin.install_symlink "luac" => "luac-5.2"
-    (include/"lua5.2").install_symlink include.children
-    (lib/"pkgconfig").install_symlink "lua.pc" => "lua5.2.pc"
-    (lib/"pkgconfig").install_symlink "lua.pc" => "lua-5.2.pc"
 
     # This resource must be handled after the main install, since there's a lua dep.
     # Keeping it in install rather than postinstall means we can bottle.
@@ -91,8 +77,6 @@ class Lua < Formula
         system "make", "install"
 
         (pkgshare/"5.2/luarocks").install_symlink Dir["#{libexec}/share/lua/5.2/luarocks/*"]
-        bin.install_symlink libexec/"bin/luarocks-5.2"
-        bin.install_symlink libexec/"bin/luarocks-admin-5.2"
         bin.install_symlink libexec/"bin/luarocks"
         bin.install_symlink libexec/"bin/luarocks-admin"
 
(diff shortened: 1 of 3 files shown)


DomT4:
I would gently request Homebrew doesn't go down this path. I use all three Luas fairly often and I maintain enough in my own tap as it is 😓.

IMO this is more of a `python`/`python3` situation than anything else. If people absolutely need it maybe stick unversioned symlinks in `libexec` but I think the current system specifically here is useful to the widest variety of people. Note this was already a compromise after people whinged loudly about how `luarocks` was handled in the past.

I don't think, generally speaking, making other package managers `keg_only` is a good path to take, which is why I nudged Homebrew to stop providing _most_ `node` versions whilst I was here in favour of `nvm` and such. Obviously you chose to go down a different route, but such is your right.

ilovezfs:
>IMO this is more of a python/python3 situation than anything else.

A `keg_only` `python@3` (and/or python@2) may be in the offing.

DomT4:
Crumbs, I'm glad you have to tell Tim that and not me.

MikeMcQuaid:
> I would gently request Homebrew doesn't go down this path. I use all three Luas fairly often and I maintain enough in my own tap as it is 😓.

The main problem I have with the current approach is it requires hefty patching that will never be upstreamed. This is something we frown heavily on elsewhere (for good reason) and I'm not sure this should be an exception.

> but I think the current system specifically here is useful to the widest variety of people

See https://github.com/Homebrew/brew/issues/1937 for an example of it not being useful to someone, currently.

> I don't think, generally speaking, making other package managers keg_only is a good path to take

I agree. I'd like to figure out how to get LuaRocks to install somewhere persistent as part of this PR and would welcome the help 🙇.

DomT4:
> The main problem I have with the current approach is it requires hefty patching that will never be upstreamed.

I'm pretty sure this is `lua` on macOS more than the current system, to be honest. The patch isn't related specifically to the way this is currently handled but more that upstream doesn't bother building a shared library despite a lot of things expecting a shared library and shared libraries being the default on macOS.

Even if you `keg_only` all the lua versions the issues the patch fixes don't go away, and you will cause users issues removing it. I agree it should be upstreamed, and I don't know if they'll touch it, but it is an upstream issue and not an issue unique to Homebrew. Besides sometimes forgetting to update the patch on update it has been an issue-free patch for years now.

> See Homebrew/brew#1937 for an example of it not being useful to someone, currently.

I did. Fundamentally though I don't see us agreeing here because I think you've veered off the path of usability with the way the version system has been implemented, and that's part of why I left in the end; me and you and ILZ were going to constantly butt heads to the end of time and I decided happiness sounded a better alternative, heh. Obviously you disagree with me here and that's completely fine. People disagree, such is life 😄.

I believe I noted to @telemachus some time ago that `lua53` only got special handling in Versions because it was _in_ Versions and that would change once it lived in core. It was a mistake to give it that special treatment at the time, to be honest, and I should have appreciated doing so would create expectations later.

MikeMcQuaid:
> I did. Fundamentally though I don't see us agreeing here because I think you've veered off the path of usability with the way the version system has been implemented, and that's part of why I left in the end; me and you and ILZ were going to constantly butt heads till the end of time and I decided happiness sounded a better alternative, heh. Obviously you disagree with me here and that's completely fine. People disagree, such is life 😄.

It's worth noting it's not so much about agreeing or disagreeing here as much as accepting that something is going to happen and in the end it's going to be decided by the person doing the work.

> I'm pretty sure this is lua on macOS more than the current system, to be honest. The patch isn't related specifically to the way this is currently handled but more that upstream doesn't bother building a shared library despite a lot of things expecting a shared library and shared libraries being the default on macOS.

If everything is expecting us to apply a long-running patch: that too feels like a smell. I'd like to try this approach and see how far we get. If it doesn't work: so be it but I think we'll at least move these to being `keg_only` with the original names.

DomT4:
> It's worth noting it's not so much about agreeing or disagreeing here as much as accepting that something is going to happen and in the end it's going to be decided by the person doing the work.

Well, with respect, I put in plenty of work but fairly frequently felt like I was going to end up losing the argument regardless of merit once the lead maintainer system set in, but I accepted my fault in not raising that as a concern when you asked me initially and consequently walked away rather than end up killing a fairly positive relationship with you.

> If everything is expecting us to apply a long-running patch: that too feels like a smell. I'd like to try this approach and see how far we get. If it doesn't work: so be it but I think we'll at least move these to being keg_only with the original names.

I think it's very likely you'll end up introducing awkward, mysterious crashes as was the initial problem, but I can't stop you. Again, I respect that it should be upstreamed and I think it strange upstream doesn't natively (at last check) try to build a sensibly versioned `dylib` on macOS, but given the history of this patch I've no idea if anyone ever raised it with them as a discussion.

telemachus:
@DomT4 I do now recall you warning me about upcoming changes to how brew handled Lua. I didn't really absorb how bad (for me, at least) those changes would be.

From my point of view, I want something simple: the version of Lua I choose (any one version) linked as `/usr/local/bin/lua`. It is frustrating that this is not now possible. That said, I grant that behind the scenes it may take a lot of unsimple work to make such a thing possible. I'll deal with whatever the outcome is, but I'm definitely not happy with the current situation, and I wanted to register that.

MikeMcQuaid:
@DomT4 You did put in the work, I agree, but the bulk of the Homebrew/brew core changes required were going to end up meaning it'd be on me. My being lead maintainer wasn't so much a reality change as a labelling change: I have been the person making most of the high-level technical decisions on Homebrew/brew since `@jacknagel` left.

I agree that this may produce some issues but we can deal with them if and when we see them. There's a bunch of patches like this that can carry on forever through inertia alone.

> From my point of view, I want something simple: the version of Lua I choose (any one version) linked as /usr/local/bin/lua. It is frustrating that this is not now possible.

@telemachus I suspect we'll get you half-way there and have something you can use by adding one directory to your PATH to pick your version.

DomT4:
@telemachus Homebrew has always struggled with versioned stuff, to be honest, or at least for as long as I've used it. The balance between usability and maintainability leaps from one side to the other without yet finding a happy middle ground. Versioned stuff is hard; users expect things to just work which is perfectly understandable but the kind of systems that "just work" for users are incredibly maintainer-heavy.

I think the idea there's a happy middle ground for everyone is pretty much fantasy, but I remain willing to be convinced, even whilst being deeply not thrilled with Homebrew's latest implementation. But every effort I've seen so far either dumps on users or dumps on maintainers actually being able to live life and not spend 8 hours a day here.

IMO the solution for `lua` if people need it is to stick unversioned symlinks in `libexec/bin`, especially since the unversioned `luarocks` symlinks _already_ live there. I think that's not a bad solution to `keg_only` being really user/third-party-tools hostile whilst still allowing users to essentially choose which `lua` becomes the "primary" one in the `$PATH`. I don't think `keg_only` has ever been or will ever be that user friendly.

@MikeMcQuaid 

> My being lead maintainer wasn't so much a reality change as a labelling change: I have been the person making most of the high-level technical decisions on Homebrew/brew since jacknagel left.

I don't disagree you've been the person making the most high-level decisions and indeed being the person shaping direction the most often, and I don't want to get into internal politics here because that always got me into trouble, but I'm unsure if you realise how much of a power grab it felt like and how much it made disagreeing with you feel like a futile exercise. And again, in all fairness you gave me a chance in advance to disagree with the idea and I didn't, so a lot of my feelings specifically around how it ended up are on me and not you.

I don't want to get into it with you given that I left precisely to retain the positive memories & feelings I have for this project and the amount of freedom and chances you gave me here, so I'll leave it at that. It is what it is.

MikeMcQuaid:
> I don't think keg_only has ever been or will ever be that user friendly.

That's something we aim to change.

> I'm unsure if you realise how much of a power grab it felt like and how much it made disagreeing with you feel like a futile exercise

Just because this has become quite a public discussion it's worth noting https://github.com/Homebrew/brew/graphs/contributors?type=a and https://github.com/Homebrew/homebrew-core/graphs/contributors?type=a (and perhaps the list of conferences I've represented Homebrew at) are the reasons I've ended up in the position I have. Frankly, I've put more time into Homebrew over a longer period of time than anyone else. I know you agreed with that reasoning in the past and perhaps still do but I just didn't want to leave the "power grab" comment unchallenged; I don't think it helps Homebrew or open-source in general to imply that influence is achieved by more than hard work and hours invested.

MikeMcQuaid:
Closing this out but will be on my list for later thinking about how to better handle version consistency.

PR #9516, opened 2017-02-01
conductr-cli 1.0.1 (new formula)

huntc:
Introduces the Lightbend ConductR CLI installation

-----

### DIFF
--- a/Formula/conductr-cli.rb
+++ b/Formula/conductr-cli.rb
@@ -1,0 +1,14 @@
+class ConductrCli < Formula
+  desc "The Lightbend ConductR CLI"
+  homepage "https://conductr.lightbend.com"
+  url "https://bintray.com/lightbend/generic/download_file?file_path=conductr-cli-1.0.1-Mac_OS_X-x86_64.zip"
+  version "1.0.1"
+  sha256 "fff2d3caa7cbbefc3da4a158b34a46bf51338ae921985cd4fb35d46e3c8eccf2"
+
+  def install
+    bin.install Dir["*"]
+  end
+
+  test do
+  end
+end


Formula/conductr-cli.rb
```diff
@@ -0,0 +1,14 @@
+class ConductrCli < Formula
+  desc "The Lightbend ConductR CLI"
+  homepage "https://conductr.lightbend.com"
+  url "https://bintray.com/lightbend/generic/download_file?file_path=conductr-cli-1.0.1-Mac_OS_X-x86_64.zip"
```
woodruffw: We require source installations for non-portable binaries (i.e., anything not Mono or Java).

Is this program open-source?

woodruffw (changes requested):

Formula/conductr-cli.rb
huntc: Yes, this is open source: https://github.com/typesafehub/conductr-cli

woodruffw: Awesome. A tarball from the [releases](https://github.com/typesafehub/conductr-cli/releases) should probably be used, then.

huntc: It isn't possible to link to a releases based tar ball as we must manually produce a native image via pyinstaller. Happy to receive other thoughts. Thanks.

huntc: Ah actually I see that I can drop binaries on to a releases page also. I didn't appreciate that. However I'd prefer to leave the binary at bintray if that's ok given that we have other things referring to it there. Would that be ok with you?

woodruffw: The problem isn't necessarily it being hosted on bintray; we just don't include binary formulae in homebrew-core (more info [here](https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md#we-dont-like-binary-formulae)).

If building is completely out of the question, this might be better suited for homebrew-cask or a custom tap.

huntc: Thanks for the pointers. I'm not sure why brew cask is required vs regular brew. Is the intent to signal to the user that they are indeed installing a binary? If so then I'm unsure they'd care that much - most users just copy/paste commands. Feel free to reply or simply close. I appreciate the dialogue.

woodruffw: > Is the intent to signal to the user that they are indeed installing a binary?

Not exactly, since we provide binary packages ourselves (bottles) via our build infrastructure. The problem with shipping binaries *not* built by us is twofold:

1. OS X's "public" APIs/ABIs change subtly with each release, making it difficult to build a single binary that works across multiple versions.
2. As a result of 1, we (the Homebrew) maintainers would have to spend more time debugging binaries that we know nothing about (since they didn't come from us) and fielding support that upstreams should general deal with.

Inclusion in Cask better reflects the user-upstream relationship that binaries have and partially shields us from OS X-induced breakage (for a good example of this, check out our [clock_gettime tag](https://github.com/Homebrew/homebrew-core/issues?utf8=%E2%9C%93&q=label%3Aclock_gettime%20)).

Thanks for submitting and discussing, but we can't include this in core. I still recommend seeing if Cask is interested, or placing it in a custom tap - `brew` will happily install binaries from custom formulae.

PR #9519, opened 2017-02-01
bochs 2.6.8 (new formula)

Jnosh:
The bochs formula was formerly part of homebrew-x11 but was not migrated since it failed to build on Sierra.

Old formula: https://github.com/Homebrew/homebrew-x11/blob/2ba3b8da7ed5dd7a8087465f580f6d032b6e5a2a/bochs.rb

GIthub issue: https://github.com/Homebrew/homebrew-x11/issues/263

This formula is based on that one with two changes:

1) Workaround to make it build
Removed the `--enable-sb16` configure argument since the the soundblaster module is responsible for the build failures on Sierra.

2) Switch from using x11 to sdl2. Dependencies have been adjusted accordingly.

-----

### DIFF
--- a/Formula/bochs.rb
+++ b/Formula/bochs.rb
@@ -1,0 +1,79 @@
+class Bochs < Formula
+  desc "Open source IA-32 (x86) PC emulator written in C++"
+  homepage "http://bochs.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/bochs/bochs/2.6.8/bochs-2.6.8.tar.gz"
+  sha256 "79700ef0914a0973f62d9908ff700ef7def62d4a28ed5de418ef61f3576585ce"
+
+  option "with-gdb-stub", "Enable GDB Stub"
+
+  depends_on "pkg-config" => :build
+  depends_on "sdl2"
+
+  def install
+    args = %W[
+      --prefix=#{prefix}
+      --with-sdl2
+      --with-nogui
+      --enable-disasm
+      --disable-docbook
+      --enable-x86-64
+      --enable-pci
+      --enable-all-optimizations
+      --enable-plugins
+      --enable-cdrom
+      --enable-a20-pin
+      --enable-fpu
+      --enable-alignment-check
+      --enable-large-ramfile
+      --enable-debugger-gui
+      --enable-readline
+      --enable-iodebug
+      --enable-xpm
+      --enable-show-ips
+      --enable-logging
+      --enable-usb
+      --enable-ne2000
+      --enable-cpu-level=6
+      --enable-clgd54xx
+      --with-term
+    ]
+
+    if build.with? "gdb-stub"
+      args << "--enable-gdb-stub"
+    else
+      args << "--enable-debugger"
+    end
+
+    system "./configure", *args
+
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    require "open3"
+
+    (testpath/"bochsrc.txt").write <<-EOS.undent
+        panic: action=fatal
+        error: action=report
+        info: action=ignore
+        debug: action=ignore
+      EOS
+
+    expected = <<-ERR.undent
+        Bochs is exiting with the following message:
+        \[BIOS  \] No bootable device\.
+      ERR
+
+    command = "#{bin}/bochs -qf bochsrc.txt"
+    if build.without? "gdb-stub"
+      # When the debugger is enabled, bochs will stop on a breakpoint early
+      # during boot. We can pass in a command file to continue when it is hit.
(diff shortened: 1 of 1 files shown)


Formula/bochs.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    system "./configure", *args
+
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system bin/"bochs", "--help"
```
woodruffw: It'd be nice to get a more in-depth test here. Then again, since this is an emulator, that may not be possible...

Jnosh:
Hi @woodruffw,

Thanks for the feedback. I tried to improve the test; feedback is very welcome. I have zero experience with ruby so I hope I'm not doing anything too stupid :) The test completes successfully on my machine.

What I'm doing now is to start emulation with bochs and capture it's ouput. Eventually it will fail once the BIOS doesn't detect any valid boot options, which I check for to see if we successfully reach that point.

That should at least cover the basics since the basic parts of the CPU, some devices, BIOS need to be working to get that far.

Currently I am also testing with the GUI enabled and bochs will create an SDL window. Please let me know if that is a problem for the buildbots or discouraged.


To test more thoroughly would require some OS image file. There are some example bochs images at https://sourceforge.net/projects/bochs/files/Disk%20Images/.
I imagine a more thorough test could be implemented by downloading one of these images and checking that we can run it for a few seconds without bochs panicking.
Please let me know if that would still be within the scope of a homebrew test and I'll try to implement it.

- Jnosh

Jnosh:
Adding a more extensive test is definitively a good idea.

The buildbot version crashes right after loading the `libbx_usb_uhci.so` plugin during startup.
The same version built by brew on my local system works fine so it's probably related to my local configuration. I'll try to track this down.

woodruffw:
> I have zero experience with ruby so I hope I'm not doing anything too stupid :) The test completes successfully on my machine.

Looks good to me! 

> Currently I am also testing with the GUI enabled and bochs will create an SDL window. Please let me know if that is a problem for the buildbots or discouraged.

I don't *think* it is (I think other formulae create windows during test), but I'm not positive. If it goes :green_apple: after we fix that `.so` problem, it should be fine.

> The buildbot version crashes right after loading the libbx_usb_uhci.so plugin during startup.
> The same version built by brew on my local system works fine so it's probably related to my local configuration. I'll try to track this down.

:+1:. A cursory google search on `libbx_usb_uhci.so` didn't reveal much, but one frequent source of problems with building on OS X is the `.so -> .dylib` change. If you can find a `libbx_usb_uhci.dylib` in your libraries, then that's probably the issue.

Jnosh:
> 👍. A cursory google search on libbx_usb_uhci.so didn't reveal much, but one frequent source of problems with building on OS X is the .so -> .dylib change. If you can find a libbx_usb_uhci.dylib in your libraries, then that's probably the issue.

Sorry, turns out that is an issue on my end. If I run the bottle form the buildbot in a Sierra VM it runs fine. (No idea what the issue is but it's not related to the buildbot failure).

Anyway, I could replicate the failure in the VM. Bochs insists on stopping on a debugger breakpoint during boot and waits for user input.
I failed to notice this originally because I had built my local copy with the gdb-stub that disables the debugger and subsequently forgot about that.

I haven't found a way to disable that breakpoint but it's possible to pass a file with debugger input to bochs so I'm using that to just have it continue at that breakpoint.

From what I've seen the formulas don't generally contain comments so I didn't describe this in the formula itself. Let me know if I should add one.

Formula/bochs.rb
```diff
@@ -0,0 +1,78 @@
...
+    require "open3"
+
+    (testpath/"bochsrc.txt").write <<-EOS.undent
+        panic: action=fatal
+        error: action=report
+        info: action=ignore
+        debug: action=ignore
+      EOS
+
+    (testpath/"debugger.txt").write("c\n")
```
woodruffw: Just put this in the conditional block below.

Formula/bochs.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    expected = <<-ERR.undent
+        Bochs is exiting with the following message:
+        \[BIOS  \] No bootable device\.
+      ERR
+
+    command = "#{bin}/bochs -qf bochsrc.txt"
+    if build.without? "gdb-stub"
+      command << " -rc debugger.txt"
```
woodruffw: I'd document the reasons for this condition, but otherwise this looks good to be.

Jnosh: Done & thanks for the help and review.

woodruffw:
Thanks @Jnosh, and congratulations on your first contribution/formula! :tada:

PR #9063, opened 2017-01-19
knot 2.4.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #9526, opened 2017-02-01
knot 2.4.0

fcelda:
This PR updates the knot formula to 2.4.0. There is an existing PR #9063 for Knot DNS, but it doesn't address following issues:

- The changes also include upstream patch to address compilation issues on macOS older than Sierra.
- The default configuration is tweaked to log onto stderr only. launchd will collect the logs so it's no reason to log with syslog as well.

(I'm Knot DNS developer.)

### DIFF
--- a/Formula/knot.rb
+++ b/Formula/knot.rb
@@ -4,8 +4,13 @@ class Knot < Formula
   revision 1
 
   stable do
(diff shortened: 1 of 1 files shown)


nijikon:

PR #9530, opened 2017-02-01
ivykis 0.41 (new formula)

MrAnno:

-----

### DIFF
--- a/Formula/ivykis.rb
+++ b/Formula/ivykis.rb
@@ -1,8 +1,8 @@
 class Ivykis < Formula
   desc "Async I/O-assisting library"
-  homepage "http://libivykis.sourceforge.net"
-  url "https://github.com/buytenh/ivykis/archive/v0.41.tar.gz"
-  sha256 "b074441cc33aa0281f393c0a172136f0e0648266b89405b31ac819d64ea89e48"
+  homepage "https://sourceforge.net/projects/libivykis"
+  url "https://downloads.sourceforge.net/project/libivykis/0.41/ivykis-0.41.tar.gz"
+  sha256 "2c934539a59029851b1332c8143ba9f21b79fb0185dd68d6eb259ae1b61ef3c5"
 
   depends_on "automake" => :build
   depends_on "autoconf" => :build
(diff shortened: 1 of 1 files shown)


Formula/ivykis.rb
```diff
@@ -0,0 +1,31 @@
+class Ivykis < Formula
+  desc "Async I/O-assisting library"
+  homepage "http://libivykis.sourceforge.net"
+  url "https://github.com/buytenh/ivykis/archive/v0.41.tar.gz"
```
woodruffw: Let's fetch it from here instead: https://sourceforge.net/projects/libivykis/files/0.41/

Formula/ivykis.rb
```diff
@@ -0,0 +1,31 @@
+class Ivykis < Formula
+  desc "Async I/O-assisting library"
+  homepage "http://libivykis.sourceforge.net"
```
woodruffw: That page leads to this, which looks like the real homepage: https://sourceforge.net/projects/libivykis/

woodruffw (changes requested):

woodruffw:
Debian et al. provide this, so I think we can ignore the notability warning here. It'll also disappear once the `url` link is moved to sourceforge.

MrAnno:
Thanks for the review.

I've updated the PR based on your notes. I had some issues with the sha256 sum because of the SourceForge mirrors, that was the reason I switched to the GitHub URL.

woodruffw:
No problem!

> I had some issues with the sha256 sum because of the SourceForge mirrors, that was the reason I switched to the GitHub URL.

Could you elaborate? Was the `sha256` varying across mirrors?

MrAnno:
> Could you elaborate? Was the sha256 varying across mirrors?

Sorry, I was wrong. All the available SourceForge mirrors have the same checksum, I just mixed up the GitHub and the SourceForge tarball, so I entered a wrong sha256 value.

woodruffw (approved):

woodruffw:
Thanks @MrAnno, and congratulations on your first contribution/formula! :tada:

PR #9539, opened 2017-02-01
mercurial 4.1

ngoldbaum:
https://www.mercurial-scm.org/wiki/WhatsNew#Mercurial_4.1_.282017-2-1.29

One possible issue that whoever merges this might want to consider is that this version of mercurial bundles zstd for improved compression performance. Since it's using private APIs to do this it does not make sense to use a version of zstd provided by homebrew. For more details about this, see https://www.mercurial-scm.org/pipermail/mercurial-packaging/2016-November/000178.html as well as replies to that thread.


tdsmith (approved):

nijikon:

PR #9572, opened 2017-02-02
knot-resolver 1.2.1 (new formula)

fcelda:
New formula for [Knot Resolver](https://www.knot-resolver.cz/) — Minimalistic, caching, DNSSEC-validating DNS resolver.

- The daemon and library are installed and fully working.
- DNSSEC trust anchor is intentionally provided in the formula. No bootstrapping is performed by default.
- Advanced cache backends (Redis and memcached) are optional.
- Documentation is not installed due to some missing Python dependencies (I might fix that in the future).
- Some advanced modules require various Lua modules which are not included yet (I might fix in the future).
- Go bindings are disabled as there are basically no existing modules yet.

-----

### DIFF
--- a/Formula/knot-resolver.rb
+++ b/Formula/knot-resolver.rb
@@ -1,0 +1,99 @@
+class KnotResolver < Formula
+  desc "Minimalistic, caching, DNSSEC-validating DNS resolver"
+  homepage "https://www.knot-resolver.cz"
+  url "https://secure.nic.cz/files/knot-resolver/knot-resolver-1.2.1.tar.xz"
+  sha256 "1b6f55ea1dfec90f45c437f23e1ab440e478570498161d0f8a8f94a439305f8c"
+
+  head do
+    url "https://gitlab.labs.nic.cz/knot/resolver.git"
+  end
+
+  depends_on "knot"
+  depends_on "luajit"
+  depends_on "libuv"
+  depends_on "gnutls"
+  depends_on "lmdb"
+
+  depends_on "cmocka" => :build
+  depends_on "pkg-config" => :run
+
+  option "without-nettle", "Compile without DNS cookies support"
+  option "with-hiredis", "Compile with Redis cache storage support"
+  option "with-libmemcached", "Compile with memcached cache storage support"
+  depends_on "nettle" => :recommended
+  depends_on "hiredis" => :optional
+  depends_on "libmemcached" => :optional
+
+  def install
+    %w[all check lib-install daemon-install modules-install].each do |target|
+      system "make", target, "PREFIX=#{prefix}"
+    end
+
+    (buildpath/"config").write(config)
+    (etc/"kresd").install "config"
+
+    (buildpath/"root.keys").write(root_keys)
+    (var/"kresd").install "root.keys"
+  end
+
+  def config; <<-EOS.undent
+    -- vim:syntax=lua:
+    -- Refer to manual: http://knot-resolver.readthedocs.org/en/latest/daemon.html#configuration
+
+    net.listen(net.lo0)
+
+    -- Unmanaged DNSSEC root TA
+    trust_anchors.file = 'root.keys'
+
+    -- Load useful modules
+    modules = {
+        'stats',    -- Track internal statistics
+        'policy',   -- Block queries to local zones/bad sites
+        'predict',  -- Prefetch expiring/frequent records
+    }
+
+    -- Cache size
+    cache.size = 100 * MB
+    EOS
+  end
+
+  # DNSSEC root anchor published by IANA
+  def root_keys; <<-EOS.undent
+    . 172800 IN DS 19036 8 2 49aac11d7b6f6446702e54a1607371607a1a41855200fd2ce1cdde32f24e8fb5
+    EOS
+  end
+
+  plist_options :startup => true
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+    <dict>
+      <key>Label</key>
+      <string>#{plist_name}</string>
+      <key>WorkingDirectory</key>
+      <string>#{var}/kresd</string>
+      <key>RunAtLoad</key>
+      <true/>
+      <key>ProgramArguments</key>
+      <array>
+        <string>#{sbin}/kresd</string>
+        <string>-c</string>
+        <string>#{etc}/kresd/config</string>
+      </array>
+      <key>StandardInPath</key>
+      <string>/dev/null</string>
+      <key>StandardOutPath</key>
+      <string>/dev/null</string>
+      <key>StandardErrorPath</key>
+      <string>#{var}/log/kresd.log</string>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    system sbin/"kresd", "--version"
+  end
+end
--- a/Formula/knot-resolver.rb
+++ b/Formula/knot-resolver.rb
@@ -26,32 +26,12 @@ class KnotResolver < Formula
       system "make", target, "PREFIX=#{prefix}"
     end
 
-    (buildpath/"config").write(config)
+    cp "etc/config.personal", "config"
+    inreplace "config", /^\s*user\(/, "-- user("
     (etc/"kresd").install "config"
 
     (buildpath/"root.keys").write(root_keys)
     (var/"kresd").install "root.keys"
-  end
-
-  def config; <<-EOS.undent
-    -- vim:syntax=lua:
-    -- Refer to manual: http://knot-resolver.readthedocs.org/en/latest/daemon.html#configuration
-
-    net.listen(net.lo0)
-
-    -- Unmanaged DNSSEC root TA
-    trust_anchors.file = 'root.keys'
-
-    -- Load useful modules
-    modules = {
-        'stats',    -- Track internal statistics
-        'policy',   -- Block queries to local zones/bad sites
-        'predict',  -- Prefetch expiring/frequent records
-    }
-
-    -- Cache size
-    cache.size = 100 * MB
-    EOS
   end
 
   # DNSSEC root anchor published by IANA
--- a/Formula/knot-resolver.rb
+++ b/Formula/knot-resolver.rb
@@ -3,11 +3,7 @@ class KnotResolver < Formula
   homepage "https://www.knot-resolver.cz"
   url "https://secure.nic.cz/files/knot-resolver/knot-resolver-1.2.1.tar.xz"
   sha256 "1b6f55ea1dfec90f45c437f23e1ab440e478570498161d0f8a8f94a439305f8c"
-  revision 1
-
-  head do
-    url "https://gitlab.labs.nic.cz/knot/resolver.git"
-  end
+  head "https://gitlab.labs.nic.cz/knot/resolver.git"
 
   depends_on "knot"
   depends_on "luajit"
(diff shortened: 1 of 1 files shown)


vcunat:
Also, `xxd` or `hexdump` is required during build.  BTW, I see we currently miss that requirement in documentation.

fcelda:
`hexdump` is part of the macOS base system, so that should be fine.

Formula/knot-resolver.rb
```diff
@@ -0,0 +1,101 @@
+class KnotResolver < Formula
+  desc "Minimalistic, caching, DNSSEC-validating DNS resolver"
+  homepage "https://www.knot-resolver.cz"
+  url "https://secure.nic.cz/files/knot-resolver/knot-resolver-1.2.1.tar.xz"
+  sha256 "1b6f55ea1dfec90f45c437f23e1ab440e478570498161d0f8a8f94a439305f8c"
+  revision 1
```
MikeMcQuaid: Can remove this as it's a new formula.

Formula/knot-resolver.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  head do
+    url "https://gitlab.labs.nic.cz/knot/resolver.git"
+  end
```
MikeMcQuaid: `head "https://gitlab.labs.nic.cz/knot/resolver.git"`

Formula/knot-resolver.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  depends_on "knot"
+  depends_on "luajit"
+  depends_on "libuv"
+  depends_on "gnutls"
+  depends_on "lmdb"
+
+  depends_on "cmocka" => :build
+  depends_on "pkg-config" => :run
```
MikeMcQuaid: Are you sure this is needed at runtime and not build time?

Formula/knot-resolver.rb
```diff
@@ -0,0 +1,101 @@
...
+    end
+
+    (buildpath/"config").write(config)
+    (etc/"kresd").install "config"
+
+    (buildpath/"root.keys").write(root_keys)
+    (var/"kresd").install "root.keys"
+  end
+
+  def config; <<-EOS.undent
```
MikeMcQuaid: Is there no default or example configuration file we can install instead?

Formula/knot-resolver.rb
```diff
@@ -0,0 +1,101 @@
...
+        'predict',  -- Prefetch expiring/frequent records
+    }
+
+    -- Cache size
+    cache.size = 100 * MB
+    EOS
+  end
+
+  # DNSSEC root anchor published by IANA
+  def root_keys; <<-EOS.undent
```
MikeMcQuaid: Is there no default or example configuration root keys we can install instead?

MikeMcQuaid (changes requested):

Formula/knot-resolver.rb
vcunat: Right, `pkg-config` should be build-time-only dependency.

Formula/knot-resolver.rb
vcunat: Standard "make install" would install a few examples, so they could be copied instead.  It's also well usable without any config file at all.

MikeMcQuaid: Using the `make install` examples or no config file would be ideal.

Formula/knot-resolver.rb
vcunat: There's only one set of root keys, as [published by IANA](https://data.iana.org/root-anchors/root-anchors.xml).  `kresd` is able to fetch it over https when needed, but for that it would require some additional lua libs to be present on the system.

Formula/knot-resolver.rb
fcelda: I think it should be build time dependency as well. But the cookbook exclusively mentions that pkg-config is run dependency: https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#specifying-other-formulae-as-dependencies

The formula also contains libkresd.pc.

Formula/knot-resolver.rb
fcelda: OK. I will fix this.

Formula/knot-resolver.rb
fcelda: The standard examples are not usable on macOS. At least, we have to disable dropping of privileges. Also running kresd without configuration doesn't allow use of modules. There is also no easy way for the user to tweak the command line parameters if we provide only plist.

I think providing configuration file this way is a good balance.

Formula/knot-resolver.rb
fcelda: These files are not strictly configuration files. It's a secure trust anchor for DNSSEC (similarly to CA certificates in PKI). And it's kind of a chicken-and-egg problem. We need to provide the anchor securely.

Most Linux distributions decided to do so in package form. The trust is already established towards the distribution and the integrity of the package can be verified using GPG signatures for instance. There is either a single package that contains the anchor and other DNSSEC-aware validators can use it (Debian). Or the anchor is shipped with each component that needs it (Fedora).

Bootstrapping via HTTPS relies on certificate authorities. Some consider this concept broken because CAs are difficult to audit. In contrast a distribution represent a single point of trust. Knot Resolver can use HTTPS for bootstrap and will also authenticate the downloaded file using out-of-band X.509, however that requires us to include ICANN public certificate instead.

I don't mind removing the trust anchor from the formula. But it makes sense to me to include it this way, because it means less moving parts and allows the users to use DNSSEC validation immediately.

fcelda:
@MikeMcQuaid Thank you for the review. I fixed the evident problems. I'm not sure about the pkg-config as the formula cookbook says something different. And I've provided explanation for shipping configuration file and the DNSSEC trust anchor. Please, can you take a look again?

fcelda:
@vcunat Thanks for the comments.

Formula/knot-resolver.rb
vcunat: I'm just curious (and unknowledgeable of Mac*) &ndash; what's wrong about dropping the privileges?

Formula/knot-resolver.rb
vcunat: > Bootstrapping via HTTPS relies on certificate authorities.

Not that much, IMO, as kresd verifies the identity explicitly, only against the certificate that it installs. (Or would install by `make install`.)

vcunat:
@fcelda: when we do this, do a no-brainer bump to 1.2.2, please :-)

Formula/knot-resolver.rb
vcunat: Still, providing the file via package manager is probably the "best solution" (TM) here.

Formula/knot-resolver.rb
MikeMcQuaid: That's just an example. Make it `:build`, thanks.

Formula/knot-resolver.rb
MikeMcQuaid: So to be explicit: there's no `make install` or sample file we can install? I'm looking for ways to avoid hard-coding so much data in the formula.

Formula/knot-resolver.rb
MikeMcQuaid: @vcunat Homebrew runs everything as the same user.
@fcelda It would be preferable to `inreplace` an existing example configuration rather than reproducing so much of it in the formula.

MikeMcQuaid (changes requested):

Formula/knot-resolver.rb
vcunat: The (upstream) package does provide `make install` and e.g. in nixpkgs we use it even for ["darwin" builds](http://hydra.nixos.org/job/nixpkgs/trunk/knot-resolver.x86_64-darwin).

MikeMcQuaid: Sorry, I mean: there's no `make install` that provides these sample files?

vcunat: It does provide them.

MikeMcQuaid: @vcunat Cool, thanks: then we should use them and modify them rather than maintain our own versions here.

Formula/knot-resolver.rb
fcelda: OK, updated.

Formula/knot-resolver.rb
fcelda: @MikeMcQuaid I checked again all sample configuration files shipped in the tarball (there is five of them). And none is usable out of the box on macOS.

The most suitable one (also the simplest one) misses configuration of networking, includes dropping of privileges, and cache size is rather low. That's at least three inreplace calls. And also renaming of some files.

So for the sake of simplicity, I would like to keep this as it is. The provided config is cleaned up version of the config I'm talking about above.

Formula/knot-resolver.rb
fcelda: @MikeMcQuaid @vcunat There is still some confusion in this discussion.

The file `root_keys` is no sample or configuration file. It's secure trust anchor for DNSSEC. The anchor rarely changes. First key was published in Jun 2010 and the new one in Feb 2017. Don't expect that we would have to change the file frequently.

Also, the trust anchor is *not* installed by `make install`. The `iccan-ca.pem` certificate is installed instead which is something different. In case `root_keys` is not provided, kresd will try to bootstrap this file by downloading the anchor from IANA website and by using the ICANN certificate to verify it's content. This may or may not work.

I understand the concern of _harcoding_ something into the formula. But I still think we should follow what for instance Linux distributions do and provide `root_keys` in the _package_. This is most comfortable for users. And also a robust solution because we can be sure that they will always have the trust anchor.

MikeMcQuaid: @fcelda Can the `root_keys` not be downloaded from anywhere instead? A `resource` block would be preferable.

Formula/knot-resolver.rb
MikeMcQuaid: It's not more simple as-is, though, it's significantly more code for the maintainers of Homebrew and contributors to this formula to keep up to date. Three `inreplace` calls and a rename is fine and fairly standard for software like this.

Formula/knot-resolver.rb
fcelda: I don't think so. The authoritative source is [crazy XML](http://data.iana.org/root-anchors/root-anchors.xml).

Formula/knot-resolver.rb
fcelda: OK. I will change that once I have some time.

fcelda:
@MikeMcQuaid OK. Let's give it another try. I've replaced the config file in formula by an upstream config file. It's the one based on `config.personal`. To minimize a number of inreplace changes, I just commented out the privilege dropping and kept everything else intact. I admit that it's better. :)

Formula/knot-resolver.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    cp "etc/config.personal", "config"
+    inreplace "config", /^\s*user\(/, "-- user("
+    (etc/"kresd").install "config"
+
+    (buildpath/"root.keys").write(root_keys)
+    (var/"kresd").install "root.keys"
+  end
+
+  # DNSSEC root anchor published by IANA
```
MikeMcQuaid: Last thing (I think): can you provide a source URL for this so if it ever needs update people know where to go looking? Thanks!

MikeMcQuaid:
> To minimize a number of inreplace changes, I just commented out the privilege dropping and kept everything else intact. I admit that it's better. :)

Thanks for both doing that but having the humility to admit it. You're great ❤️

Formula/knot-resolver.rb
vcunat: It's this XML: https://data.iana.org/root-anchors/root-anchors.xml

MikeMcQuaid: Cool. A link for that in this formula and I think we're good to go 👍

fcelda: It's there. :)

fcelda:
@MikeMcQuaid I've added the link to the IANA website as a source for DNSSEC TA. So we should be good to go. :-)

dunn:
Thanks for your patience! Merged as 1ac53486cf008da3099e200e28f22cd986c546e3.

PR #9598, opened 2017-02-03
boost: Split boost-mpi into separate formula

mkuron:

-----

This pull request moves the building of Boost::MPI into a separate formula, _boost-mpi_. Previously, it was available as an option of the _boost_ formula, _boost --with-mpi_. Having it as a separate formula has two advantages: Boost will not need to be rebuilt in its entirety when Boost::MPI is installed, and Boost::MPI will be available as a bottle, thus speeding up installation. Furthermore, _boost-python_ has always been a separate formula.

### DIFF
--- a/Formula/boost-mpi.rb
+++ b/Formula/boost-mpi.rb
@@ -1,0 +1,98 @@
+class BoostMpi < Formula
+  desc "C++ library for C++/MPI interoperability"
+  homepage "https://www.boost.org/"
+  url "https://downloads.sourceforge.net/project/boost/boost/1.63.0/boost_1_63_0.tar.bz2"
+  sha256 "beae2529f759f6b3bf3f4969a19c2e9d6f0c503edcb2de4a61d1428519fcb3b0"
+  head "https://github.com/boostorg/boost.git"
+
+  bottle do
+    cellar :any
+  end
+
+  option :cxx11
+
+  if build.cxx11?
+    depends_on "boost" => "c++11"
+    depends_on "open-mpi" => "c++11"
+  else
+    depends_on "boost"
+    depends_on :mpi => [:cc, :cxx]
+  end
+
+  def install
+    # "layout" should be synchronized with boost
+    args = ["--prefix=#{prefix}",
+            "--libdir=#{lib}",
+            "-d2",
+            "-j#{ENV.make_jobs}",
+            "--layout=tagged",
+            "--user-config=user-config.jam",
+            "threading=multi,single",
+            "link=shared,static"]
+
+    # Build in C++11 mode if boost was built in C++11 mode.
+    # Trunk starts using "clang++ -x c" to select C compiler which breaks C++11
+    # handling using ENV.cxx11. Using "cxxflags" and "linkflags" still works.
+    if build.cxx11?
+      args << "cxxflags=-std=c++11"
+      if ENV.compiler == :clang
+        args << "cxxflags=-stdlib=libc++" << "linkflags=-stdlib=libc++"
+      end
+    elsif Tab.for_name("boost").cxx11?
+      odie "boost was built in C++11 mode so boost-mpi must be built with --c++11."
+    end
+
+    open("user-config.jam", "a") do |file|
+      file.write "using darwin : : #{ENV.cxx} ;\n"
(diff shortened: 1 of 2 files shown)


mkuron:
The only test that failed on the test bot are related to the ncmpcpp, which I didn't touch. The test succeeds for me locally, and #8660 also mentions it failing on the bot, so it's probably fine.

Formula/boost-mpi.rb
```diff
@@ -0,0 +1,98 @@
+class BoostMpi < Formula
+  desc "C++ library for C++/MPI interoperability"
+  homepage "https://www.boost.org/"
+  url "https://downloads.sourceforge.net/project/boost/boost/1.63.0/boost_1_63_0.tar.bz2"
+  sha256 "beae2529f759f6b3bf3f4969a19c2e9d6f0c503edcb2de4a61d1428519fcb3b0"
+  head "https://github.com/boostorg/boost.git"
+
+  bottle do
+    cellar :any
+  end
```
MikeMcQuaid: Remove this; bottles will be added when we pull.

MikeMcQuaid (changes requested):

MikeMcQuaid:
@ilovezfs Thoughts on this?

ilovezfs:
Is there a downside to just making it the default instead?

mkuron:
Making it the default would mean that boost depends on openmpi. Many people and a lot of C++ software use boost, but most of them don't need openmpi. Furthermore, openmpi depends on a Fortran compiler, which most people definitely don't need.

ilovezfs:
And won't this PR then mean that the boost bottle will have been built with clang but the boost-mpi bottle will have been built with gcc?

mkuron:
No, why should it? It is built in the same way as the boost-python bottle.

ilovezfs:
It needs gfortran6 from gcc no?

mkuron:
openmpi needs gfortran, but only to build its Fortran bindings. These are obviously not used by boost, which is C++.

ilovezfs:
Any reason for not providing the same options?
```
   option "without-single", "Disable building single-threading variant"
   option "without-static", "Disable building static library variant"
```

Also on any thoughts on whether the icu4c module could/should be similarly split out?

mkuron:
No reason for not providing these options. boost-python doesn't provide them either, but I can add them to both if you want that for consistency.

Regarding icu4c: that's not a separate boost module. It's only a different backend for Boost::Regex.

ilovezfs:
>No reason for not providing these options. boost-python doesn't provide them either, but I can add them to both if you want that for consistency.

Only asking because the analytics show some people combining the mpi option with --without-single. But it probably is fine to have all 4 since it shouldn't materially increase build-time for just one module.

>Regarding icu4c: that's not a separate boost module. It's only a different backend for Boost::Regex.

Ah OK. So then I revise my question:
Also any thoughts on whether the Boost::Regex module could/should be similarly split out?

mkuron:
> Only asking because the analytics show some people combining the mpi option with --without-single. But it probably is fine to have all 4 since it shouldn't materially increase build-time for just one module.

Exactly. While single vs. single+double doubles the build time, static vs static+dynamic only takes a few seconds extra because only linking happens twice. boost-mpi builds really quickly anyway, so doubling the compile time is no big deal. The complete boost, on the other hand, takes significant time to compile.

> Also any thoughts on whether the Boost::Regex module could/should be similarly split out?

Boost::Regex is built by default in the boost formula. We could create two formulae, boost-regex and boost-regex-icu4c, that declare conflicts on each other. However, many formulae have optional dependencies on icu4c and `--with-icu4c` options and none have separate formulae with and without icu4c. So until somebody comes up with a compelling reason to create separate boost-regex and boost-regex-icu4c formulae, I'd leave Boost::Regex in the main boost formula. The more sensible thing would probably be to make `--with-icu4c` the default -- but having never used Boost::Regex, that is not for me to decide. If somebody needs that change and can argue their case, they can submit another pull request.

ilovezfs (approved):

MikeMcQuaid (approved):

MikeMcQuaid:

PR #9603, opened 2017-02-03
tor: add post_install step that creates /usr/local/var/lib/tor

reedriley:

-----

By default, the version of `tor` installed won't start because of a missing folder.  We should create it in the post-install step.

### DIFF
--- a/Formula/tor.rb
+++ b/Formula/tor.rb
@@ -39,6 +39,10 @@ class Tor < Formula
     system "make", "install"
   end
 
+  def post_install
+    mkdir_p "#{HOMEBREW_PREFIX}/var/lib/tor"
+  end
+
   plist_options :manual => "tor"
 
   def plist; <<-EOS.undent


Formula/tor.rb
```diff
@@ -39,6 +39,10 @@ def install
     system "make", "install"
   end
 
+  def post_install
+    mkdir_p "#{HOMEBREW_PREFIX}/var/lib/tor"
```
DomT4: `(var/"lib/tor").mkpath`

DomT4:
Sorry to circle back to this. Are you sure this is correct, and not as a result of a personal configuration locally? The default configuration file [writes to](https://gitweb.torproject.org/tor.git/tree/src/config/torrc.sample.in#n52) `$HOME/.tor`, but can be uncommented to write to the `var/lib` directory.

reedriley:
@DomT4, technically you're correct - this can be overridden via a `$HOME/.torrc`.  But because homebrew builds `tor` with a prefix all the documentation (i.e., `man tor`) and the default values (i.e., for `DataDirectory`) point to `/usr/local/var/lib/tor`.  (It also tries to read the configuration from `/usr/local/etc/tor/torrc` unless you have one in `$HOME/.torrc` as well.)

If we want to consider a sample `.torrc` from the tor website "the default configuration", then I'd argue we probably want to configure our package that way during install as well.

MikeMcQuaid:
> If we want to consider a sample .torrc from the tor website "the default configuration", then I'd argue we probably want to configure our package that way during install as well.

I'm happy with this approach.

DomT4:
Presuming you simply mean removing `--localstatedir=#{var}` I think that would be fine. I do find it somewhat bemusing that a configuration flag causes the manpage to point to the directory in `var/lib` when the UNIX default is `~/.tor` unless you choose to set it to something else in a configuration file.

JCount:
@reedriley ping on this

JCount:
Closing per inactivity. Thank you for your work nonetheless @reedriley. If you end up deciding to work on this again, feel free to open a new PR.

PR #9607, opened 2017-02-03
spidermonkey @ 1486087028

mathiasbynens:

-----

The existing SpiderMonkey formula builds a heavily outdated version from source. This new formula installs a more recent version by downloading a precompiled binary off the official servers.

### DIFF
--- a/Formula/spidermonkey.rb
+++ b/Formula/spidermonkey.rb
@@ -1,54 +1,17 @@
 class Spidermonkey < Formula
-  desc "JavaScript-C Engine"
+  desc "Mozilla's JavaScript engine, as used in Firefox"
   homepage "https://developer.mozilla.org/en/SpiderMonkey"
-  url "https://archive.mozilla.org/pub/mozilla.org/js/js185-1.0.0.tar.gz"
-  version "1.8.5"
-  sha256 "5d12f7e1f5b4a99436685d97b9b7b75f094d33580227aa998c406bbae6f2a687"
-  revision 2
-
-  head "https://hg.mozilla.org/tracemonkey/archive/tip.tar.gz"
-
-  bottle do
-    cellar :any
-    sha256 "62193341691f6f35a1d844409c587b431aa7540b70c02d90451e2cb3623788de" => :sierra
-    sha256 "5e7789a8ba4e3259364bd3ae827037ba83bf3a076633799bf8f5869b885db399" => :el_capitan
-    sha256 "38d1b7f54b5dbdd4a0e28e3a1077aed2ada42a9266cfaddeda6a08d761a2d8b2" => :yosemite
-  end
-
-  conflicts_with "narwhal", :because => "both install a js binary"
-
-  depends_on "readline"
-  depends_on "nspr"
+  url "https://archive.mozilla.org/pub/firefox/tinderbox-builds/mozilla-release-macosx64/1486087028/jsshell-mac.zip"
+  version "1486087028"
+  sha256 "68ade82a1f5840ef575af51cd25bfc09a623049e35cb51bf6f914269701b5d94"
 
   def install
-    cd "js/src" do
-      # Remove the broken *(for anyone but FF) install_name
-      inreplace "config/rules.mk",
-        "-install_name @executable_path/$(SHARED_LIBRARY) ",
-        "-install_name #{lib}/$(SHARED_LIBRARY) "
-    end
-
-    mkdir "brew-build" do
-      system "../js/src/configure", "--prefix=#{prefix}",
-                                    "--enable-readline",
-                                    "--enable-threadsafe",
-                                    "--with-system-nspr",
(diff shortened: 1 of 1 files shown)


mathiasbynens:
***Do not merge this yet!***

Let’s see where the discussion on https://bugzilla.mozilla.org/show_bug.cgi?id=1336514 goes first.

Formula/spidermonkey.rb
```diff
@@ -1,54 +1,17 @@
...
-      inreplace "js-config", /JS_CONFIG_LIBS=.*?$/, "JS_CONFIG_LIBS=''"
-      # These need to be in separate steps.
-      system "make"
-      system "make", "install"
-
-      # Also install js REPL.
-      bin.install "shell/js"
-    end
+    lib.install "libmozglue.dylib", "libnss3.dylib"
+    bin.install "js" => "spidermonkey"
```
mathiasbynens: Ideally, this would be aliased as `sm` in addition to `spidermonkey`. What’s the best way to do that within a formula? `bin.install "js" => "spidermonkey", "js" => "sm"` doesn’t seem to work.

Formula/spidermonkey.rb
```diff
@@ -1,54 +1,17 @@
 class Spidermonkey < Formula
-  desc "JavaScript-C Engine"
+  desc "SpiderMonkey is Mozilla's JavaScript engine, as used in Firefox"
```
bfontaine: We can use `desc "Mozilla's JavaScript engine, as used in Firefox"` not to repeat the formula’s name.

Formula/spidermonkey.rb
```diff
@@ -1,54 +1,17 @@
...
-    sha256 "62193341691f6f35a1d844409c587b431aa7540b70c02d90451e2cb3623788de" => :sierra
-    sha256 "5e7789a8ba4e3259364bd3ae827037ba83bf3a076633799bf8f5869b885db399" => :el_capitan
-    sha256 "38d1b7f54b5dbdd4a0e28e3a1077aed2ada42a9266cfaddeda6a08d761a2d8b2" => :yosemite
-  end
-
-  conflicts_with "narwhal", :because => "both install a js binary"
-
-  depends_on "readline"
-  depends_on "nspr"
+  url "https://archive.mozilla.org/pub/firefox/tinderbox-builds/mozilla-release-macosx64/1486087028/jsshell-mac.zip"
```
bfontaine: We need to build from source.

mathiasbynens: How strong is that requirement? Building SpiderMonkey takes a long time, and building other JavaScript engines like V8 takes even longer. (For V8 specifically, the build process is quite cumbersome and it changes all the time — see #9392 for more background. CC @pinepain)

Formula/spidermonkey.rb
mathiasbynens: Thanks! Updated.

Formula/spidermonkey.rb
MikeMcQuaid: It's a hard blocker.

mathiasbynens: Ah, I found https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md#we-dont-like-binary-formulae which explains this requirement a bit further. Feel free to close this PR — I’ll create a tap instead.

Still, if anyone knows [how to `bin.install` a file to two distinct locations (two names)](https://github.com/Homebrew/homebrew-core/pull/9607#discussion_r99463543), I’d appreciate it if you could tell me :)

mathiasbynens: Tap with formulae for ECMAScript/JavaScript engines (downloaded from official sources instead of built from source, for performance reasons): https://github.com/mathiasbynens/homebrew-ecmascript

Formula/spidermonkey.rb
mathiasbynens:
```
bin.install "js" => "spidermonkey"
bin.install_symlink "spidermonkey" => "sm"
```

Formula/spidermonkey.rb
MikeMcQuaid: Why is performance better when not build from source?

mathiasbynens: @MikeMcQuaid I meant `brew install spidermonkey` is much faster if all you need to do is download a tarball and extract it, compared to building the exact same binary from source.

MikeMcQuaid: We build binary packages for people for this same reason.

MikeMcQuaid: Which in this case is `cellar :any` so can be used by anyone using the latest three versions of macOS

mathiasbynens: /me reads https://github.com/Homebrew/brew/blob/master/docs/Bottles.md

PR #9616, opened 2017-02-03
wine 2.1 (devel)

neutric:
Created with `brew bump-formula-pr`.

PR #9618, opened 2017-02-03
wine 2.1 (devel)

uyjulian:


neutric:
Thanks for your contribution! This change was merged in an earlier PR (see #9616) and can thus be closed.

PR #9623, opened 2017-02-04
open-sp: Fix default catalog path

petere:
The previous value `#{HOMEBREW_PREFIX}/share/sgml/catalog` didn't make much sense, since that is not a configuration file location.  Change it to `#{etc}/sgml/catalog`, which is similar to the default XML catalog path used by most formulae.

Previous change was Homebrew/legacy-homebrew#15955. The claims there about OpenJade are not correct AFAICT. I have an OpenJade package at https://github.com/petere/homebrew-sgml/blob/master/openjade.rb, which does the right thing (I think) about catalog installation and works with my change.

### DIFF
--- a/Formula/open-sp.rb
+++ b/Formula/open-sp.rb
@@ -18,7 +18,7 @@ class OpenSp < Formula
                           "--mandir=#{man}",
                           "--disable-doc-build",
                           "--enable-http",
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):

nijikon:

PR #9635, opened 2017-02-04
yaz: add universal build option

gmcharlt:
No; patch intentionally (re)adds the deprecated universal build option for the sake of a Mono app called MarcEdit, which [expects](http://marcedit.reeset.net/marcedit-mac-installation-instructions) yaz to be built with `--universal` and `--without-icu4c`
-----

### DIFF
--- a/Formula/yaz.rb
+++ b/Formula/yaz.rb
@@ -18,10 +18,16 @@ class Yaz < Formula
     depends_on "libtool" => :build
   end
 
+  # to support an app called MarcEdit that uses yaz for Z39.50 client
+  # functionality and expects it to be installed with  --universal and
+  # --without-icu4c
+  option :universal
+
   depends_on "pkg-config" => :build
   depends_on "icu4c" => :recommended
 
   def install
+    ENV.universal_binary if build.universal?
     system "./buildconf.sh" if build.head?
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",


ilovezfs:
We're auditing all universal options out of existence. Perhaps MarcEdit itself just needs to be built without universal.

gmcharlt:
My understanding in discussion with the maintainer of MarcEdit is that it's not as simple as that on MarcEdit's end. I can't speak to the details, so I've alerted him to this discussion in case he cares to participate. In any event, the current state of affairs is a showstopper for users who want to use the Z39.50 client functionality of MarcEdit on OS X.

reeset:
I will put a note here -- I'm the MarcEdit maintainer, and as much as I would like to build MarcEdit only for 64-bit systems, I'm unable to.  The tool uses Xamarin's toolset which allows me to fuse object C and C# on the MacOS.  I'd like to think, long-term, these issues will be sorted, but until they are, my preference has been to point users to homebrew for installation as it helps to handle the dependency issues that come up when installing yaz, while not breaking other programs that may rely on the 64-bit library being present.  You are certainly welcome to remove this option, and I'll have to work out a different method for users wanting to use Z39.50 with MarcEdit -- but it would be a breaking change for the + or - 800 or so active Mac users of the tool (do they all use z39.50 -- probably not, but I don't know that).

ilovezfs:
@reeset I suggest you copy a universal build of yaz into your app and distribute it directly.

reeset:
So, originally, this is what I did.  I build my own version of yaz an included it with the application, but I found  that I ran into dependency issues across systems.  If this bit drops, that's the path that I'll likely take -- but it's not just that simple.  I'll need to strip the component of all external dependencies, and then maintain my own build repository of yaz going forward.

ilovezfs:
What dependencies does yaz have? The formula doesn't seem to have any ...

reeset:
Yaz makes use of a variety of xml/xslt and character encoding libraries.

ilovezfs:
Those are all provided by the system
```
  /usr/lib/libSystem.B.dylib
  /usr/lib/libedit.3.dylib
  /usr/lib/libexslt.0.dylib
  /usr/lib/libiconv.2.dylib
  /usr/lib/libicucore.A.dylib
  /usr/lib/libncurses.5.4.dylib
  /usr/lib/libxml2.2.dylib
  /usr/lib/libxslt.1.dylib
  /usr/lib/libz.1.dylib
```

reeset:
Regardless, I'm find if you want to eventually take this build option away.  I'd just appreciate a little bit of time to make accommodations and provide a work around.  I don't have the time right now to fix this kind of breaking change on my end, and Galen letting me know about this today was the first I'd heard about it (I'm not a Mac user if I can help it).  If you want to pick a time to drop it from the profile, that's fine, but I'd appreciate some time (maybe a month) so that I can accommodate the changes on my end, update the documentation, and (this is more important), ensure that this doesn't break the workflows of the various libraries that rely on the tool.

reeset:
I realize the system has them -- but I'm just telling what I found when testing in the wild.  The tool is used from OSX 10.8 - present, and just sending as didn't always work.  This was the only way I found to reliably make this work in all cases, and it was really Misty that helped make that happen.

ilovezfs:
@reeset any reason you can't copy the formula into your own tap and point users to that? https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap

That should take a matter of minutes to set up.

reeset:
Maybe -- I'm not really familiar with homebrew.  I can certainly take a look.  If its straightforward, sure -- I could do that.

ilovezfs:
@reeset please try that. It should be straightforward. I suggest you also rename the formula to like yaz-universal.rb and the class to YazUniversal to avoid name collisions with our formula. If you need any help, feel free to come bug me on IRC.

reeset:
Great -- I will take a look at that this afternoon once I get back from coaching our local robotics team.  Best.

PR #9641, opened 2017-02-04
Remove universal support

nijikon:

-----

- [ ] fix all the comments
- [ ] move patches to formula-patches
- [ ] vendor deps into the wine formula

### DIFF
--- a/Formula/freeimage.rb
+++ b/Formula/freeimage.rb
@@ -23,8 +23,6 @@ class Freeimage < Formula
     sha256 "9d1e914ae20deb7066caf5f1cf52c3d48c0c04ccd36b791170c7e1fcb3528a36" => :mountain_lion
   end
 
-  option :universal
-
   patch :DATA
 
   # fix GCC 5.0 compile.
@@ -34,7 +32,6 @@ class Freeimage < Formula
   end
 
   def install
-    ENV.universal_binary if build.universal?
     system "make", "-f", "Makefile.gnu"
     system "make", "-f", "Makefile.gnu", "install", "PREFIX=#{prefix}"
     system "make", "-f", "Makefile.fip"
@@ -50,7 +47,7 @@ index b59c419..6e177fc
 --- a/Makefile.fip
 +++ b/Makefile.fip
 @@ -5,8 +5,9 @@ include fipMakefile.srcs
- 
+
  # General configuration variables:
  DESTDIR ?= /
 -INCDIR ?= $(DESTDIR)/usr/include
@@ -58,11 +55,11 @@ index b59c419..6e177fc
 +PREFIX ?= /usr/local
 +INCDIR ?= $(DESTDIR)$(PREFIX)/include
 +INSTALLDIR ?= $(DESTDIR)$(PREFIX)/lib
- 
+
  # Converts cr/lf to just lf
  DOS2UNIX = dos2unix
 @@ -35,9 +36,9 @@ endif
- 
+
  TARGET  = freeimageplus
  STATICLIB = lib$(TARGET).a
 -SHAREDLIB = lib$(TARGET)-$(VER_MAJOR).$(VER_MINOR).so
@@ -73,7 +70,7 @@ index b59c419..6e177fc
 +VERLIBNAME = lib$(TARGET).$(VER_MAJOR).dylib
  HEADER = Source/FreeImage.h
  HEADERFIP = Wrapper/FreeImagePlus/FreeImagePlus.h
- 
+
 @@ -49,7 +50,7 @@ all: dist
  dist: FreeImage
 	mkdir -p Dist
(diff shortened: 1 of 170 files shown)


MikeMcQuaid (approved):

Formula/gegl.rb
```diff
@@ -44,14 +42,6 @@ def install
       --disable-docs
```
ilovezfs: The array should be removed.

Formula/jsoncpp.rb
```diff
@@ -28,10 +26,6 @@ def install
       -DJSONCPP_WITH_TESTS=OFF
       -DJSONCPP_WITH_POST_BUILD_UNITTEST=OFF
```
ilovezfs: don't really need this array either

Formula/libdc1394.rb
```diff
@@ -39,7 +36,7 @@ def install
 --- a/dc1394/macosx/capture.c
 +++ b/dc1394/macosx/capture.c
 @@ -150,7 +150,7 @@ callback (buffer_info * buffer, NuDCLRef dcl)
-
```
ilovezfs: looks like white space was changed in the patch

Formula/libfreenect.rb
```diff
@@ -13,20 +13,13 @@ class Libfreenect < Formula
     sha256 "f714532e1b21365063746846544a340dac70cf0c5cc877a207dd17284ee100b7" => :mavericks
   end
 
-  option :universal
-
   depends_on "cmake" => :build
   depends_on "libusb"
 
   def install
     args = std_cmake_args
```
ilovezfs: can just move std_cmake_args down to the system "cmake" line, with the -DBUILD_OPENNI2_DRIVER=ON inline

Formula/openssl.rb
```diff
@@ -67,58 +63,18 @@ def install
     dirs = []
 
     archs.each do |arch|
```
ilovezfs: don't need iteration if there's just one

Formula/openssl@1.1.rb
```diff
@@ -70,57 +66,17 @@ def install
     dirs = []
 
     archs.each do |arch|
```
ilovezfs: ditto

Formula/subversion.rb
```diff
@@ -252,21 +230,21 @@ def caveats
 +++ b/configure
 @@ -26153,6 +26153,8 @@ fi
  SWIG_CPPFLAGS="$CPPFLAGS"
-
```
ilovezfs: looks like whitespace was changed in the patch

Formula/freeimage.rb
```diff
@@ -50,19 +47,19 @@ def install
 --- a/Makefile.fip
 +++ b/Makefile.fip
 @@ -5,8 +5,9 @@ include fipMakefile.srcs
-
```
ilovezfs: looks like whitespace was changed the patch

Formula/openssl.rb
DomT4: Technically there is still two here, because he's left in the possibility the CPU being built against is 32-bit, just above this blob.

DomT4: It remains redundant, but the above block also needs tweaking, since `each` implies more than one which will no longer be possible. That's what I meant to say, more clearly 😉. Not sure if the retention of building on 32-bit CPUs was intentional or not.

fxcoudert:
`wine.rb` says:

```
  # note that all wine dependencies should declare a --universal option in their formula,
  # otherwise homebrew will not notice that they are not built universal
  def require_universal_deps?
    MacOS.prefer_64_bit?
  end
```

Among the `wine` dependencies, there are several (`jpeg`, `libtiff`, `little-cms2` to name just a few) that are included in this PR. Won't that break `wine`?

nijikon:
@fxcoudert this is not finished yet, those will be brought back so wine will work.

DomT4:
The list of recursive `wine` deps is kind of... extensive:
```
fontconfig
freetype
gd
gmp
gnutls
jasper
jpeg
libffi
libgphoto2
libicns
libpng
libtasn1
libtiff
libtool
libunistring
libusb
libusb-compat
little-cms2
net-snmp
nettle
openssl
p11-kit
sane-backends
webp
```

MikeMcQuaid:
At some point those wine deps should also just be vendored into the wine formula.

nijikon:
@MikeMcQuaid I think that we should do this now.

MikeMcQuaid:
@nijikon I'm 👍 on that, I just didn't want to commit to doing the work personally.

jtsylve:
Why is support for universal being dropped?  I can't find the discussion.

MikeMcQuaid:
@jtsylve There isn't a wider discussion really anywhere public but the justification is detailed here where the `brew audit` rule was created: macOS has been 64-bit only since 10.6 so universal options are deprecated.

jtsylve:
I'd like to offer some decent to this decision.  While OS X has been 64-bit only for quite some time, many off us still have to support legacy 32-bit only applications, which means we need to be able to link against 32-bit version of libraries.  Removing the universal support only hinders those in similar situations.  I'm not sure what you gain by removing it.

MikeMcQuaid:
> I'd like to offer some decent to this decision. While OS X has been 64-bit only for quite some time, many of us still have to support legacy 32-bit only applications, which means we need to be able to link against 32-bit version of libraries. Removing the universal support only hinders those in similar situations. I'm not sure what you gain by removing it.

We do not test formulae options (including universal) in CI and it increases our support surface and burden as well as decreasing the overall quality of Homebrew when these options fail.

While it's unfortunate you still have to maintain those applications it's worth noting that Homebrew is a tool run by volunteers and provided for free. It's still possible for you to maintain your own taps or just build universal libraries outside of Homebrew.

jtsylve:
Will support for `ENV.universal_binary` be removed completely?  I don't mind editing my local formulas to add the one or two lines it would take to re-enable the functionality

MikeMcQuaid:
> Will support for ENV.universal_binary be removed completely? I don't mind editing my local formulas to add the one or two lines it would take to re-enable the functionality

@jtsylve We have no plans to do that. Currently it's just removing support from Homebrew/homebrew-core rather than Homebrew/brew.

nijikon:
Close in favor of the new ones, thanks @fxcoudert!

Issue #9621, opened 2017-02-04
rocksdb now installs librocksdb_lite


tmm1:
As of https://github.com/Homebrew/homebrew-core/commit/6445d90c4723fd9f27e1e0e4ac4ab8f913f3f31a, the rocksdb package now builds rocksdb_lite by default. RocksDBLite is a separate version of rocksdb, and is not compatible with the original. Any application which currently has a dependency on the rocksdb homebrew package is now broken, because the package no longer providers librocksdb.dylib

I think it makes sense to restore the original behavior. osquery can still depend on librocksdb.dylib and work fine with it. Or, we can add a new `rocksdb_lite` formula which installs librocksdb_lite.dylib, since both versions can be installed side-by-side.

cc @ilovezfs What do you think?

ilovezfs:
I think given that this is the first time this is being raised here that the majority of our users are better served by having the bottles be rocksdb_lite.

tmm1:
What about having different formulae?

ilovezfs:
That would be highly unusual.

tmm1:
Okay then what if the bottle contains both versions? The libs were explicitly designed by the rocksdb team to live side by side.

rocksdb is more than four years old and there are bindings available for every single programming language. Many of these are not compatible with rocksdb_lite, which is only a few months old.

Further rocksdb_lite was designed primarily for mobile applications. It is missing feature one expects would be included when installing a package named rocksdb: https://github.com/facebook/rocksdb/blob/master/ROCKSDB_LITE.md

ilovezfs:
Would be happy to look at a PR for that.

PR #9642, opened 2017-02-04
rocksdb: build both regular and lite versions

tmm1:
cc https://github.com/Homebrew/homebrew-core/issues/9621 @ilovezfs

### DIFF
--- a/Formula/rocksdb.rb
+++ b/Formula/rocksdb.rb
@@ -3,6 +3,7 @@ class Rocksdb < Formula
   homepage "http://rocksdb.org"
   url "https://github.com/facebook/rocksdb/archive/v5.0.2.tar.gz"
   sha256 "5e39d2131ebdb92c30eda3d32861f489f4843fad50cc2fbd5d234bc4415948ca"
+  revision 1
 
   bottle do
     cellar :any
@@ -10,9 +11,6 @@ class Rocksdb < Formula
     sha256 "024bd14ab972cca0421b6adb9d9fac80570337077b1eaa6f1a0db0b0aafa161c" => :el_capitan
     sha256 "b058bea89af8825ab24bcebe61418ed20a8d504805292dda9d5addca181d13e1" => :yosemite
   end
-
-  option "without-lite", "Don't build mobile/non-flash optimized lite version"
-  option "with-tools", "Build tools"
 
   needs :cxx11
   depends_on "snappy"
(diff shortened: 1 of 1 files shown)


Formula/rocksdb.rb
```diff
@@ -22,15 +22,21 @@ class Rocksdb < Formula
   def install
     ENV.cxx11
     ENV["PORTABLE"] = "1" if build.bottle?
-    if build.with? "lite"
-      ENV.append_to_cflags "-DROCKSDB_LITE=1"
-      ENV["LIBNAME"] = "librocksdb_lite"
-    end
-    system "make", "clean"
```
ilovezfs: do we still need to run make clean here too?

Formula/rocksdb.rb
```diff
@@ -11,7 +12,6 @@ class Rocksdb < Formula
     sha256 "b058bea89af8825ab24bcebe61418ed20a8d504805292dda9d5addca181d13e1" => :yosemite
   end
 
-  option "without-lite", "Don't build mobile/non-flash optimized lite version"
   option "with-tools", "Build tools"
```
ilovezfs: any thoughts on always installing the tools too?

Formula/rocksdb.rb
```diff
@@ -3,6 +3,7 @@ class Rocksdb < Formula
   homepage "http://rocksdb.org"
   url "https://github.com/facebook/rocksdb/archive/v5.0.2.tar.gz"
   sha256 "5e39d2131ebdb92c30eda3d32861f489f4843fad50cc2fbd5d234bc4415948ca"
+  revision 2
```
ilovezfs: revision 1 (we count from 0 implicitly)

Formula/rocksdb.rb
tmm1: Yea that seems like a good idea. Should I keep the option (`without-tools`) or remove it?

Formula/rocksdb.rb
tmm1: Are the build directories ever re-used? On a fresh build `make clean` shouldn't be necessary.

Formula/rocksdb.rb
```diff
@@ -22,15 +22,21 @@ class Rocksdb < Formula
...
-    if build.with? "lite"
-      ENV.append_to_cflags "-DROCKSDB_LITE=1"
-      ENV["LIBNAME"] = "librocksdb_lite"
-    end
-    system "make", "clean"
+
+    # build regular rocksdb
     system "make", "static_lib"
     system "make", "shared_lib"
     system "make", "tools" if build.with? "tools"
```
ilovezfs: are the tools identical for lite and non-lite?

Formula/rocksdb.rb
ilovezfs: nope, but I have no idea why that make clean was originally there

Formula/rocksdb.rb
ilovezfs: You'd just remove it. We only use "deprecated_option" when it's being mapped to something other than nothing.

tmm1:
> are the tools identical for lite and non-lite?

I believe the regular tools would work with both, since they contain a superset of lite features.

@igorcanadi @theopolis can you confirm?

Formula/rocksdb.rb
tmm1: Ok I put it back.

Formula/rocksdb.rb
tmm1: Alright, removed.

Formula/rocksdb.rb
tmm1: thanks

Formula/rocksdb.rb
MikeMcQuaid: > Are the build directories ever re-used?

They are not.

tmm1:
Looks like the tests for the tools are failing- doesn't seem like they ever worked, but that was masked by the fact that the tools weren't built by default.

tmm1:
> I believe the regular tools would work with both, since they contain a superset of lite features.

Actually it looks like most of the tools do not even support LITE mode:

https://github.com/facebook/rocksdb/blob/master/tools/dump/rocksdb_dump.cc#L14
https://github.com/facebook/rocksdb/blob/master/tools/dump/rocksdb_undump.cc#L14
https://github.com/facebook/rocksdb/blob/master/tools/ldb.cc#L18
https://github.com/facebook/rocksdb/blob/master/tools/sst_dump.cc#L18

tmm1:
@ilovezfs Can this be merged? Happy to address any remaining concerns.

ilovezfs:
@tmm1 do you know why the tools tests fail? Is it just that the exit code isn't zero for `--help`? If so, you can use `assert_match "something expected", shell_output("#{bin}/foo --help", 1)` assuming the exit code were 1 not 0.

tmm1:
Yes they all exit 1. I can switch to assertions, thanks for the tip.

ilovezfs:
Great, thanks!

tmm1:
@ilovezfs All set.

ilovezfs (approved):

ilovezfs:
Shipped! Thanks @tmm1. This formula is much better now.

theopolis:
Thanks for the work here folks! Excited that Rocks is now more flexible

PR #9649, opened 2017-02-04
tenyr 0.9.3 (significant version update)

kulp:
Stop patching (which was incorrect -- see http://bit.ly/2jPhJYX)

Depend on sdl2_image softly

Disable JIT build until we have AsmJit somehow

Add basic test using check_sim

Update install script for new version

Specify that HEAD is on develop branch

Depend on pkg-config during build

-----

### DIFF
--- a/Formula/tenyr.rb
+++ b/Formula/tenyr.rb
@@ -57,6 +57,10 @@ class Tenyr < Formula
   end
 
   test do
-    system "make", "check_sim"
+    # sanity test assembler, linker and simulator
+    `echo 'B <- 9'     | tas --output=a.to -`
+    `echo 'C <- B * 3' | tas --output=b.to -`
+    `tld --output=test.texe a.to b.to`
+    `tsim -vvvv test.texe | grep -o 'C [0-9a-f]\\{8\\}' | tail -n1` == "C 0000001b\n"
   end
 end
--- a/Formula/tenyr.rb
+++ b/Formula/tenyr.rb
@@ -1,10 +1,10 @@
 class Tenyr < Formula
-  desc "32-bit computing environment"
+  desc "32-bit computing environment (including simulated CPU)"
   homepage "http://tenyr.info/"
-  url "https://github.com/kulp/tenyr/archive/v0.6.0.tar.gz"
-  sha256 "fc9ed9f6f46c79470163ed5e9f2f820aab74c7bd972315862d3cca547988684c"
+  url "https://github.com/kulp/tenyr/archive/v0.9.3.tar.gz"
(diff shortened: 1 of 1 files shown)


woodruffw:
`test` failed, looks like this make target no longer exists:

```
==> make check_sim
make: *** No rule to make target `check_sim'.  Stop.
Error: tenyr: failed
Failed executing: make check_sim
```

kulp:
I think the problem is that I used File.chdir inside the `install` rule. Could that cause this ?

I am new to writing Homebrew recipes. How can I test the `test` part of a recipe locally through `brew` ?

woodruffw:
> I think the problem is that I used File.chdir inside the install rule. Could that cause this ?

Possibly, yeah. You can do this instead:

```ruby
cd "path/to/whatever" do
  # whatever
end
```

> I am new to writing Homebrew recipes. How can I test the test part of a recipe locally through brew ?

`brew test tenyr` should work.

kulp:
I was under the mistaken impression that the `test` method was called within the context of a source build, but I now see why this is not the case. Instead of using a rule from the build makefile, I will create a small inline test case for now.

Formula/tenyr.rb
```diff
@@ -14,43 +14,53 @@ class Tenyr < Formula
...
+    cd builddir do
+      bin.install "tsim", "tas", "tld"
+      lib.install Dir["*.dylib"]
+    end
+  end
 
- struct symbol *symbol_find(struct symbol_list *list, const char *name);
+  test do
+    # sanity test assembler, linker and simulator
+    `echo 'B <- 9'     | tas --output=a.to -`
```
woodruffw: These should probably be wrapped with `system` instead, and file creation should be isolated:

```ruby
(testpath/"part1").write "B <- 9"
(testpath/"part2").write "C <- B * 3"

system "tas", "--output=a.to", "part1"
system "tas", "--output=b.to", "part2"
 # etc...

assert_match "C 0000001b", `tsim -vvvv test.texe`
```

kulp: Thank you, I will make the suggested changes.

MikeMcQuaid:

For future reference the preferred commit message format is `tenyr 0.9.3`. Please complete the issue template in future PRs where this format is detailed in the linked [Contributing guidelines](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md). Thanks!

kulp:
When I had read the contributing guidelines, I wasn't sure where these significant changes fell between the two examples in "The preferred commit message format for simple version updates is `foobar 7.3` and for fixes is `foobar: fix flibble matrix`." From now on I think things are clear. Thanks !

PR #9660, opened 2017-02-05
[moved] dmd: install shared libraries if they're generated

timotheecour:
eg, on linux, it'll allow the .so to be generated; likewise on OSX when these become generated

-----


MikeMcQuaid:
@timotheecour I've told you many times at this point that we don't want to accept changes that apply just to Linux at not macOS. We'll accept this change when they are actually generated on macOS.

PR #9666, opened 2017-02-05
blueutil: Drop dependency on Xcode, use makefile

chdiza:

-----

Version 1.1.0 introduced, and 1.1.2 repaired, a makefile that allows the user to avoid Xcode altogether.    Use it and drop the ":build" dep on Xcode.

### DIFF
--- a/Formula/blueutil.rb
+++ b/Formula/blueutil.rb
@@ -13,12 +13,9 @@ class Blueutil < Formula
     sha256 "ad5d8d0d865f9d369090698432c822c7842a4b59d8c5bc1627cd54959cb3329c" => :yosemite
   end
 
-  depends_on :xcode => :build
-
   def install
-    # Set to build with SDK=macosx10.6, but it doesn't actually need 10.6
-    xcodebuild "SDKROOT=", "SYMROOT=build"
-    bin.install "build/Release/blueutil"
+    system "make"
+    bin.install "blueutil"
   end
 
   test do
--- a/Formula/blueutil.rb
+++ b/Formula/blueutil.rb
@@ -14,8 +14,11 @@ class Blueutil < Formula
   end
 
   def install
-    system "make"
-    bin.install "blueutil"
+    bin.mkpath
+    inreplace "Makefile",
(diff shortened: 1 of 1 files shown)


Formula/blueutil.rb
```diff
@@ -13,12 +13,9 @@ class Blueutil < Formula
...
   end
 
-  depends_on :xcode => :build
-
   def install
-    # Set to build with SDK=macosx10.6, but it doesn't actually need 10.6
-    xcodebuild "SDKROOT=", "SYMROOT=build"
-    bin.install "build/Release/blueutil"
+    system "make"
+    bin.install "blueutil"
```
ilovezfs: can we convince them to add an install target too ...

chdiza: There is one; I'll need to make an inreplace, though.  I'll add this later today, I can't do it right this second.

chdiza: All that gets installed is the binary.  Theres no manpage.  Seems simpler to just `bin.install`, but whatever.

ilovezfs: Preference for `make install` over `bin.install` is one of the tenets of Brewism.

ilovezfs:
I think you can
```
system "make", "DESTDIR=#{prefix}", install
```

chdiza:
> I think you can
> system "make", "DESTDIR=#{prefix}", "install"

No, I think that will hardcode `/usr/local`, and then there'd have to be an inreplace to fix that.

chdiza:
(The Makefile hardcodes `/usr/local`, and setting `DESTDIR` to our `#{prefix}` doesn't undo the hardcoding.)

ilovezfs:
Ah, you can just do `prefix=#{prefix}` but why are they only echoing the command not executing it?

ilovezfs:
```
    inreplace "Makefile", "@echo $(INSTALL_PROGRAM)", "$(INSTALL_PROGRAM)"
    bin.mkpath
    system "make", "prefix=#{prefix}", "install"
```
The first two lines should not be necessary and are reportable bugs.

chdiza:
It's not worth the effort to me when `bin.install` seems to exist for exactly this kind of case.

Never mind.

PR #9685, opened 2017-02-06
Import formulae from homebrew/games.

MikeMcQuaid:
Companion to https://github.com/Homebrew/homebrew-games/pull/788.


ilovezfs:
Looks like something went wrong. Only 16/174.

MikeMcQuaid:
@ilovezfs Have rerun with `--debug`, let's see what happens.

JCount:
@MikeMcQuaid The issue seems to be with the `blastem` formula itself. It still depends on `homebrew/python/pillow` instead of `homebrew/science/pillow`. This seems cause the failure of entire job when the CI cannot find the `pillow` formula after tapping `homebrew/python`.

mistydemeo:
I'll look at fixing anything here that has issues before we make a decision on removing anything.

MikeMcQuaid:
@mistydemeo That's fine but I'd like to time box it. How does a week sound (i.e. ship this on Wednesday 15th). I would say that the things that don't install at all as-is without fixes are unlikely to be widely used so personally don't think they are good candidates for inclusion.

mistydemeo:
> That's fine but I'd like to time box it. How does a week sound (i.e. ship this on Wednesday 15th).

I've been at a conference, so I haven't had a chance to look before today. Let's timebox it for a week today (18th AEST).

mistydemeo:
Oops, missed this was already shipped.

PR #9699, opened 2017-02-06
gpgme-1.8.0

jstedfast:

-----

### DIFF
--- a/Formula/gpgme.rb
+++ b/Formula/gpgme.rb
@@ -1,9 +1,9 @@
 class Gpgme < Formula
   desc "Library access to GnuPG"
   homepage "https://www.gnupg.org/related_software/gpgme/"
-  url "https://www.gnupg.org/ftp/gcrypt/gpgme/gpgme-1.7.0.tar.bz2"
-  mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/gpgme/gpgme-1.7.0.tar.bz2"
-  sha256 "71f55fed0f2b3eaf7a606e59772aa645ce3ffff322d361ce359951b3f755cc48"
+  url "https://www.gnupg.org/ftp/gcrypt/gpgme/gpgme-1.8.0.tar.bz2"
+  mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/gpgme/gpgme-1.8.0.tar.bz2"
+  sha256 "596097257c2ce22e747741f8ff3d7e24f6e26231fa198a41b2a072e62d1e5d33"
 
   bottle do
     cellar :any
@@ -11,6 +11,8 @@ class Gpgme < Formula
     sha256 "9764f539b4b51161127d95687c5611f9571ae2abde1dce66ba014a0f33d2bcc1" => :el_capitan
     sha256 "4a4c3e90175b02ded974a8a385b6d9799bcc2873c5a1087c558f350eb103934d" => :yosemite
   end
+
+  patch :DATA
 
   depends_on "gnupg2"
   depends_on "libgpg-error"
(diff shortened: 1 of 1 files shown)


Formula/gpgme.rb
```diff
@@ -12,6 +12,8 @@ class Gpgme < Formula
     sha256 "4a4c3e90175b02ded974a8a385b6d9799bcc2873c5a1087c558f350eb103934d" => :yosemite
   end
 
+  patch :DATA
```
nijikon: We need a comment why this was added.

nijikon (changes requested):

Formula/gpgme.rb
MikeMcQuaid: To be more explicit: can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Given this is security software we're unlikely to apply this until upstream accepts it. Thanks!

jstedfast: The patch is needed in order to compile the c++ sources. Without the patch, it fails due to `strdup()` not being defined. The patch is based on the fact that gpgme-1.7.0 does not #include "config.h" in those sources and does compile w/o error.

That said, I've just emailed the gnupg-devel list with a proper solution which involves changes to configure.ac to define `_XOPEN_SOURCE` to `700` instead of `500`.

jstedfast: FWIW, this is my email to gnupg-devel (seems to be sitting in moderation queue since it hasn't shown up in the archives yet):

```
Hey all,
 
It turns out that gpgme-1.8.0 does not build on Mac OS X (tested only on Sierra 10.12.2 with Xcode 8.2), but 1.7.0 does.
 
The compile failures are all from lang/cpp/src/*result.cpp due to strdup() being undeclared.
 
Source files such as src/get-env.c, data.c, etc also warn about strdup() being undeclared but do not error out. I suspect the issue is due to c++ being more strict.
 
I did a quick comparison of gpgme-1.7.0 and 1.8.0 and discovered the problem was that 1.8.0 added #include “config.h”, so it appears that something in config.h is making the c++ compiler fail on Mac OS X.
 
I took a look at /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/string.h and it only defined strdup() if __DARWIN_C_LEVEL >= 200112L (stpcpy(), another source of compiler *warnings*, is likewise only defined if __DARWIN_C_LEVEL >= 200809L).
 
After a bit of investigation, I eventually came to sys/cdefs.h which has the following:
 
/* Deal with various X/Open Portability Guides and Single UNIX Spec. */
#ifdef _XOPEN_SOURCE
#if _XOPEN_SOURCE - 0L >= 700L && (!defined(_POSIX_C_SOURCE) || _POSIX_C_SOURCE - 0L < 200809L)
#undef _POSIX_C_SOURCE
#define _POSIX_C_SOURCE         200809L
#elif _XOPEN_SOURCE - 0L >= 600L && (!defined(_POSIX_C_SOURCE) || _POSIX_C_SOURCE - 0L < 200112L)
#undef _POSIX_C_SOURCE
#define  _POSIX_C_SOURCE                            200112L
#elif _XOPEN_SOURCE - 0L >= 500L && (!defined(_POSIX_C_SOURCE) || _POSIX_C_SOURCE - 0L < 199506L)
#undef _POSIX_C_SOURCE
#define  _POSIX_C_SOURCE                            199506L
#endif
#endif
 
Sure enough, gpgme’s config.h defines _XOPEN_SOURCE to be 500.
 
After modifying the definition of _XOPEN_SOURCE to 700 in my config.h, gpgme-1.8.0 built fine although there are still a few warnings in the c sources:
 
conversion.c:519:14: warning: implicit declaration of function 'timegm' is invalid in C99 [-Wimplicit-function-declaration]
      return timegm (&buf);
             ^
 
data.c:196:11: warning: comparison of unsigned enum expression < 0 is always false [-Wtautological-compare]
  if (enc < 0 || enc > GPGME_DATA_ENCODING_MIME)
      ~~~ ^ ~
 
key.c:293:16: warning: expression which evaluates to zero treated as a null pointer constant of type 'char *' [-Wnon-literal-null-conversion]
    sig->uid = '\0';
               ^~~~
 
engine-gpg.c:1160:11: warning: comparison of unsigned enum expression >= 0 is always true [-Wtautological-compare]
                  if (r >= 0)
                      ~ ^  ~
engine-gpg.c:2512:15: warning: implicit declaration of function 'asprintf' is invalid in C99 [-Wimplicit-function-declaration]
          if (asprintf (r_line,
              ^
 
engine-gpgsm.c:399:11: warning: implicit declaration of function 'asprintf' is invalid in C99 [-Wimplicit-function-declaration]
      if (asprintf (&optstr, "OPTION display=%s", dft_display) < 0)
          ^
engine-gpgsm.c:639:21: warning: comparison of unsigned enum expression >= 0 is always true [-Wtautological-compare]
              if (r >= 0 && status_fnc && !cb_err)
                  ~ ^  ~
engine-gpgsm.c:990:10: warning: comparison of unsigned enum expression >= 0 is always true [-Wtautological-compare]
          if (r >= 0)
              ~ ^  ~
 
 
Looking into asprintf() being undefined, it turns out stdio.h only defines asprintf() on Mac OS X if __DARWIN_C_LEVEL >= __DARWIN_C_FULL and __DARWIN_C_FULL is defined to be 900000L, so configure.ac may also need to define __DARWIN_C_LEVEL as 900000L for Mac OS X rather than trying to define _XOPEN_SOURCE alone.
```

I'll add a link to the mailing-list archived email once it shows up there.

jstedfast: https://lists.gnupg.org/pipermail/gnupg-devel/2017-February/032550.html

MikeMcQuaid: Thanks! Please add the link and short description as an inline Ruby comment and let's wait for a response.

jstedfast: Hey guys, sorry for my delay on updating this. I've been busy but will try to get back to this when I get a chance. Just wanted to let you guys know.

jstedfast: Sorry it's been taking me so long to get back to this... I'm half thinking that it might be worth waiting for a gpgme 1.8.1?

dunn: Up to you!

fxcoudert:
Looks like it was decided to wait on the release of gpgme 1.8.1. Closing for now, please reopen (or open a new PR with a link to this one) when it is the case.

PR #9705, opened 2017-02-07
Update libpointing 1.0.4

casiez:
This new version of libpointing fixes the bug of seize parameter not working as well as the bug of "system:" URI wrongly modifying the system's transfer function is some conditions.


nijikon:

PR #9733, opened 2017-02-07
gnuplot: add libcerf support

equal-l2:

-----

`libcerf` enables gnuplot to use some special functions, `cerf` for example.

### DIFF
--- a/Formula/gnuplot.rb
+++ b/Formula/gnuplot.rb
@@ -26,6 +26,7 @@ class Gnuplot < Formula
   option "with-tex", "Build with LaTeX support"
   option "with-aquaterm", "Build with AquaTerm support"
   option "without-gd", "Build without gd based terminals"
+  option "with-libcerf", "Build with libcerf support"
 
   deprecated_option "with-x" => "with-x11"
   deprecated_option "pdf" => "with-pdflib-lite"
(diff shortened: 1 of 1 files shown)


woodruffw:
I think our current policy in core is to avoid cross-tap dependencies, wherever possible.

ping @ilovezfs for an opinion

ilovezfs:
Optional cross tap dependencies are permissible. However, in the last month libcerf has 17 installs, so I'm thinking this option isn't really going to be used.

equal-l2:
Actually I don't know any package using `libcerf` other than gnuplot.
Providing `libcerf` in `resource` block might be better.

ilovezfs:
>Providing libcerf in resource block might be better.

Can you build it in statically? How long does it take to build?

equal-l2:
> Can you build it in statically?

Yes.

> How long does it take to build?

The build takes only a minute including `configure` step.

ilovezfs:
Sounds fine to me.

Out of interest: what does it do? :)

equal-l2:
> what does it do?

You mean what `libcerf` itself does?
`libcerf` provides implementation of `cerf` and `cerfc` math functions, which are "reserved" in C99 but not implemented.
`libcerf` also provides some special math functions.

With libcerf, gnuplot can use these math functions.

MikeMcQuaid (approved):
Will leave the final call up to @ilovezfs

ilovezfs (approved):

JCount:
CI git checkout went :boom:

JCount:
Thank you for your contribution to Homebrew, @equal-l2!

PR #9753, opened 2017-02-08
Import formulae from homebrew/games (take two).

MikeMcQuaid:
Try to import the formulae that failed to build on Sierra in https://github.com/Homebrew/homebrew-core/pull/9685.

@mistydemeo @tomyun I'll leave this open for about a week after which things that still can't build with Clang on Sierra will be deleted from homebrew/games rather than imported.

@tomyun I'll give you write access to my fork and @mistydemeo should already have it.

### DIFF
--- a/Formula/raine.rb
+++ b/Formula/raine.rb
@@ -1,0 +1,38 @@
+class Raine < Formula
+  desc "680x0 arcade emulator"
+  homepage "http://raine.1emulation.com/"
+  url "https://github.com/zelurker/raine/archive/0.64.13.tar.gz"
+  sha256 "0af13e67744ac81f987687a3f83703bc844897a6a1b828a19d82f96dfe8ab719"
+  head "https://github.com/zelurker/raine.git"
+
+  depends_on "gettext" => "universal"
+  depends_on "libpng" => "universal"
+  depends_on "sdl" => "universal"
+  depends_on "sdl_image" => "universal"
+  depends_on "sdl_sound" => ["universal", "with-flac", "with-libogg", "with-libvorbis"]
+  depends_on "sdl_ttf" => "universal"
+  depends_on "muparser" => "universal"
+  depends_on "nasm" => :build
+
+  depends_on "flac" => "universal" # for sdl_sound
+  depends_on "freetype" => "universal" # for sdl_ttf
+  depends_on "xz" => "universal" # for sdl_sound, etc.
+
+  def install
+    ENV.m32
+    inreplace "makefile" do |s|
+      s.gsub! /-framework (SDL|SDL_image|SDL_ttf)/, "-l\\1"
+      s.gsub! %r{/usr/local/lib/libSDL_sound\.a}, "-lSDL_sound"
+      s.gsub! %r{/usr/local/lib/libintl\.a}, "-lintl"
+      s.gsub! %r{/usr/local/lib/libmuparser\.a}, "-lmuparser"
+    end
+    system "make"
+    system "make", "install"
+    prefix.install "Raine.app"
+    bin.write_exec_script "#{prefix}/Raine.app/Contents/MacOS/raine"
+  end
+
+  test do
+    assert_match /RAINE \(680x0 Arcade Emulation\) #{version} /, shell_output("#{bin}/raine -n")
+  end
+end
--- a/Formula/raine.rb
+++ b/Formula/raine.rb
@@ -144,8 +144,9 @@ class Raine < Formula
       s.gsub! %r{/usr/local/lib/libSDL_sound\.a}, "-lSDL_sound"
       s.gsub! %r{/usr/local/lib/libintl\.a}, "-lintl"
       s.gsub! %r{/usr/local/lib/libmuparser\.a}, "-lmuparser"
+      s.gsub! %r{-I/usr/local/include}, ENV.cflags
     end
-    system "make"
+    system "make", "LD=#{ENV.cxx} #{ENV.ldflags}"
     system "make", "install"
     prefix.install "Raine.app"
     bin.write_exec_script "#{prefix}/Raine.app/Contents/MacOS/raine"
--- a/Formula/raine.rb
+++ b/Formula/raine.rb
@@ -35,6 +35,7 @@ class Raine < Formula
       "sdl_sound" => %w[--disable-dependency-tracking
                         --disable-sdltest
                         --disable-mikmod
+                        --disable-modplug
                         --disable-physfs
                         --disable-speex],
       "sdl_ttf"   => %W[--disable-debug
--- a/Formula/raine.rb
+++ b/Formula/raine.rb
@@ -143,7 +143,7 @@ class Raine < Formula
     end
 
     inreplace "makefile" do |s|
-      s.gsub! /-framework SDL_ttf/, "#{buildpath}/lib/libSDL_ttf.a #{buildpath}/lib/libfreetype.a -lbzip2"
+      s.gsub! /-framework SDL_ttf/, "#{buildpath}/lib/libSDL_ttf.a #{buildpath}/lib/libfreetype.a -lbz2"
       s.gsub! /-framework (SDL_image|SDL)/, "#{buildpath}/lib/lib\\1.a"
       s.gsub! %r{/usr/local/lib/libSDL_sound\.a}, "#{buildpath}/lib/libSDL_sound.a"
       s.gsub! %r{/usr/local/lib/libintl\.a}, "#{buildpath}/lib/libintl.a"
--- a/Formula/raine.rb
+++ b/Formula/raine.rb
@@ -48,6 +48,9 @@ class Raine < Formula
                         --disable-shared]).dup
   end
 
+  depends_on "nasm" => :build
+  depends_on "pkg-config" => :build
+
   resource "gettext" do
     url "https://ftpmirror.gnu.org/gettext/gettext-0.19.8.1.tar.xz"
     mirror "https://ftp.gnu.org/gnu/gettext/gettext-0.19.8.1.tar.xz"
@@ -108,9 +111,6 @@ class Raine < Formula
     sha256 "0666ef55da72c3e356ca85b6a0084d56b05dd740c3c21d26d372085aa2c6e708"
   end
 
-  depends_on "nasm" => :build
-  depends_on "pkg-config" => :build
-
   def install
     ENV.m32
 
@@ -157,7 +157,7 @@ class Raine < Formula
     frameworks = %w[ApplicationServices AppKit AudioToolbox AudioUnit Carbon
                     CoreFoundation CoreGraphics CoreServices Foundation IOKit]
 
-    system "make", "LD=#{ENV.cxx} #{ENV.ldflags} #{frameworks.map {|f| "-framework #{f}"}.join(" ")}"
+    system "make", "LD=#{ENV.cxx} #{ENV.ldflags} #{frameworks.map { |f| "-framework #{f}" }.join(" ")}"
     system "make", "install"
     prefix.install "Raine.app"
     bin.write_exec_script "#{prefix}/Raine.app/Contents/MacOS/raine"
--- a/Formula/raine.rb
+++ b/Formula/raine.rb
@@ -25,6 +25,9 @@ class Raine < Formula
                         --without-git
                         --without-cvs
                         --without-xz],
+      "muparser"  => %w[--disable-debug
+                        --disable-dependency-tracking
+                        --disable-samples],
       "sdl"       => %w[--without-x],
       "sdl_image" => %w[--disable-dependency-tracking
                         --disable-imageio
--- a/Formula/raine.rb
+++ b/Formula/raine.rb
@@ -27,6 +27,7 @@ class Raine < Formula
                         --without-xz],
       "muparser"  => %w[--disable-debug
                         --disable-dependency-tracking
+                        --disable-shared
                         --disable-samples],
       "sdl"       => %w[--without-x],
       "sdl_image" => %w[--disable-dependency-tracking
@@ -41,9 +42,10 @@ class Raine < Formula
       "sdl_ttf"   => %W[--disable-debug
                         --disable-dependency-tracking
                         --disable-sdltest
-                        --with-freetype-prefix=#{libexec}],
+                        --with-freetype-prefix=#{buildpath}],
     }.fetch(package, %w[--disable-debug
-                        --disable-dependency-tracking]).dup
+                        --disable-dependency-tracking
+                        --disable-shared]).dup
   end
 
   resource "gettext" do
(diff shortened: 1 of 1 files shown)


ilovezfs:
- [ ] arx-libertatis
- [ ] blastem (https://github.com/Homebrew/homebrew-core/pull/9756)
- [ ] brogue (https://github.com/Homebrew/homebrew-core/pull/9757)
- [ ] chocolate-doom
- [ ] inform6 (https://github.com/Homebrew/homebrew-core/pull/9758)
- [ ] magnetix
- [ ] nethack4 (https://github.com/Homebrew/homebrew-core/pull/9754)
- [ ] openmsx
- [ ] openrct2 (https://github.com/Homebrew/homebrew-core/pull/9759)
- [ ] pytouhou (https://github.com/Homebrew/homebrew-core/pull/9760)
- [ ] raine
- [ ] simutrans
- [ ] slashem (https://github.com/Homebrew/homebrew-core/pull/9761)
- [ ] taisei (https://github.com/Homebrew/homebrew-core/pull/9762)

mistydemeo:
> @mistydemeo @tomyun I'll leave this open for about a week after which things that still can't build with Clang on Sierra will be deleted from homebrew/games rather than imported.

As mentioned in the previous PR, I've been at a conference and haven't had a chance to look before now - mind making it a week today (18th AEST)?

mistydemeo:
Raine itself isn't failing; what's failing is `sdl_image --universal`, because we removed webp's universal option in 2e69f2c7e9ecb992e0afc1092f15e3f6c9b50d95.

mistydemeo:
openmsx fails due to clang crashing, which is an Apple/clang bug rather than an openmsx bug:

```
Compiling video/scalers/ScalerFactory.cc...
clang: error: unable to execute command: Floating point exception: 8
clang: error: clang frontend command failed due to signal (use -v to see invocation)
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin16.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
clang: note: diagnostic msg: PLEASE submit a bug report to http://developer.apple.com/bugreporter/ and include the crash backtrace, preprocessed source, and associated run script.
clang: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /tmp/HQ3xScaler-9b2370.cpp
clang: note: diagnostic msg: /tmp/HQ3xScaler-9b2370.sh
clang: note: diagnostic msg: 

********************
```

mistydemeo:
The arx-libertatis bug appears to have already been fixed upstream: https://github.com/arx/ArxLibertatis/commit/39fb9a0e3a6888a6a5f040e39896e88750c89065

ilovezfs:
@mistydemeo right ... but still would need an old CMake to build (or more patch(es) from HEAD), and even then once it built, it failed to launch in my 10.12 vm (could be because it's vm, but I didn't investigate further given the CMake trouble).

mistydemeo:
`HEAD` launches fine for me, after some testing; I think it likely only fails to launch in a VM.

ilovezfs:
Neat! I wonder if they'd be willing to make a new tag.

ilovezfs:
>Raine itself isn't failing; what's failing is sdl_image --universal, because we removed webp's universal option in 2e69f2c.

That's a show stopper since we're removing _all_ 32-bit and universal options, except the ones used by wine, and those will also be removed as soon as they've all been vendored into the wine formula.

mistydemeo:
> That's a show stopper since we're removing all 32-bit and universal options, except the ones used by wine, and those will also be removed as soon as they've all been vendored into the wine formula.

My understanding is that key portions of Raine are written in 32-bit ASM, and it can't be built as 64-bit at all.

MikeMcQuaid:
@mistydemeo We can always remove and re-add these so would rather if we stuck to a week here so the tap migration process can continue. Sounds like there's at least a few (if not all remaining) that we're not going to be able to realistically get fixed by next week so delete and re-add is probably their best bet.

ilovezfs:
>openmsx fails due to clang crashing, which is an Apple/clang bug rather than an openmsx bug:

it compiles with ENV.O0

mistydemeo:
> it compiles with ENV.O0

Sounds like a good workaround.

@MikeMcQuaid Do you want me to add commits to this branch, or submit a PR against the branch?

> We can always remove and re-add these so would rather if we stuck to a week here so the tap migration process can continue.

A week is fine, I'd just rather it's a week from when I started looking at it!

mistydemeo:
Pushed a few changes:

* arx-libertatis: fix stable build by backporting a few changes
* arx-libertatis: add HEAD
* openmsx: use `-O0` to work around clang bug

Formula/arx-libertatis.rb
```diff
@@ -25,6 +25,23 @@ class ArxLibertatis < Formula
...
 
+    # The patches for these aren't straightforward to backport because of
+    # other changes; these minimal inreplaces get it building.
+    # HEAD is fine, and the next stable release will contain these changes.
+    if build.stable?
+      inreplace "src/platform/Time.cpp", "clock_t", "clockid_t"
+
+      # Version parsing is broken in the current stable; fixed upstream.
+      # This hardcodes the current version based on data from VERSION.
+      inreplace "src/core/Version.cpp.in" do |s|
```
ilovezfs: These version issues are what using the old CMake addressed, but hard-coding them temporarily instead of vendoring the "correct" CMake seems fine to me.

mistydemeo: Yeah, given that this will be fixed in the next stable version, I think some minimal hardcoding is probably best.

Formula/arx-libertatis.rb
```diff
@@ -25,6 +25,23 @@ class ArxLibertatis < Formula
   def install
     args = std_cmake_args
 
+    # The patches for these aren't straightforward to backport because of
+    # other changes; these minimal inreplaces get it building.
+    # HEAD is fine, and the next stable release will contain these changes.
```
ilovezfs: The last stable release was in 2013. It would be good to get some kind of commitment out of upstream to at least create a new tag at some point soon.

mistydemeo: Agreed. Development is clearly very active, but with four years between releases I would really like to see another.

ilovezfs: Mind creating "Can haz tag plz?" issue so we can at least track where they stand on this?

Formula/openmsx.rb
```diff
@@ -21,6 +21,10 @@ class Openmsx < Formula
   end
 
   def install
+    # Fixes a clang crash; this is an LLVM/Apple bug, not an openmsx bug
+    # https://github.com/Homebrew/homebrew-core/pull/9753
```
ilovezfs: maybe also add the rdar://30475877 reference?

mistydemeo:
simutrans is failing because of QTKit:

```
sound/core-audio_sound.mm:12:9: fatal error: 'QTKit/QTMovie.h' file not found
#import <QTKit/QTMovie.h>
        ^
```

ilovezfs:
>simutrans is failing because of QTKit:

Indeed. So the best we could is an or_later bottle, but I think it's probably best to nix the formula.

mistydemeo:
The two files using QTKit are very small - I'll see if I can work up a patch to port to AVFoundation.

ilovezfs:
>The two files using QTKit are very small - I'll see if I can work up a patch to port to AVFoundation.

:+1:

ilovezfs:
@mistydemeo note it's also the same situation for magnetix:
```
In file included from /tmp/magnetix-20170211-88069-107zbbw/magnetiX_src/MXFSTableView.m:36:
In file included from /tmp/magnetix-20170211-88069-107zbbw/magnetiX_src/MXFSController.h:36:
/tmp/magnetix-20170211-88069-107zbbw/magnetiX_src/MagneticController.h:36:9: fatal error: 
      'QTKit/QTMovie.h' file not found
#import <QTKit/QTMovie.h>
        ^
1 error generated.
```

mistydemeo:
I guess this means I have to install Xcode. :(

mistydemeo:
Simutrans patch finished, and posted here: http://forum.simutrans.com/index.php?topic=16675.new#new

ilovezfs:
Nice work!

mistydemeo:
Ported magnetiX - branch here: https://github.com/mistydemeo/magnetiX/tree/avfoundation I've emailed the developer with the patch.

ilovezfs:
Nice!

So looks like `raine` will be the only casualty?

mistydemeo:
With Raine, if we're not willing to accept a version with vendored dependencies, I'll migrate it into my own tap with those changes.

ilovezfs:
The analytics show only 13 installs of `raine` since July, so I'm not sure it's worth the effort, but that's up to you.

mistydemeo:
It's a nice bit of emulation history. Raine was a big deal in the late 90s! I'm willing to put in the effort for a little nostalgia.

ilovezfs:
Sounds fine to me.

ilovezfs:
@mistydemeo I tested simutrans patch. Builds and launches fine. Only snag was a checksum mismatch on the language_pack:
```
==> Verifying simutrans--text-Base+texts.zip checksum
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
ChecksumMismatchError: SHA256 mismatch
Expected: 4c711c343db25e4055bf62b54c3bd8d96da5d43148db1c7767a72e586336790b
Actual: 202816f67750cfb9decdfca3bacfebbc5bfc3474c8703239418edc093ce3774d
Archive: /Users/joe/Library/Caches/Homebrew/simutrans--text-Base+texts.zip
```

mistydemeo:
That seems to have been stealth-patched, yeah. I'll push a fix for it.

Formula/magnetix.rb
```diff
@@ -7,6 +7,14 @@ class Magnetix < Formula
 
   depends_on :xcode => :build
 
+  # Port audio code from QTKit to AVFoundation
+  # Required since 10.12 SDK no longer includes QTKit.
+  # Submitted by email to the developer.
+  patch do
```
ilovezfs: Probably should scope this to `if DevelopmentTools.clang_build_version >= 800`, no?

Formula/simutrans.rb
```diff
@@ -1,11 +1,22 @@
...
 
+  stable do
+    url "https://downloads.sourceforge.net/project/simutrans/simutrans/120-1-3/simutrans-src-120-1-3.zip"
+    version "120.1.3"
+    sha256 "2d29b849fc39d25a0580091e1377270bddb2cae36c0fc32bd7c2d0f1d7ccfb84"
+
+    # Port Mac audio code from QTKit to AVFoundation
+    # Required since 10.12 SDK no longer includes QTKit.
+    # Submitted upstream: http://forum.simutrans.com/index.php?topic=16675.0
+    patch do
```
ilovezfs: ditto

Formula/magnetix.rb
mistydemeo: Given it's a major path, I'd rather not end up shipping two different versions of the audio code - would make support easier if we're only supporting one version.

ilovezfs: @mistydemeo as you may have guessed, I'm just anticipating what @MikeMcQuaid will/would/may say here hehe

MikeMcQuaid:
Wow, great work @mistydemeo and @ilovezfs 💖 

> With Raine, if we're not willing to accept a version with vendored dependencies

@mistydemeo I'd be happy to accept the version with vendored dependencies 👍 . I don't have a problem with formulae using universal internally if they need it, just the proliferation of universal options/builds.

mistydemeo:
Pushed an update. The vendored dependencies all work at this point, but raine itself is having a little trouble finding them.

Formula/raine.rb
```diff
@@ -0,0 +1,157 @@
...
+                    /(CGDirectPaletteRef.+)$/,
+                    "#if (MAC_OS_X_VERSION_MIN_REQUIRED < 1070)\n\\1\n#endif"
+        elsif name == "sdl_ttf"
+          inreplace "SDL_ttf.c",
+                    "for ( row = 0; row < glyph->bitmap.rows; ++row ) {",
+                    "for ( row = 0; row < glyph->pixmap.rows; ++row ) {"
+        end
+
+        args = configure_args(r.name)
+        args << "--prefix=#{libexec}"
```
ilovezfs: can they not be built statically instead and installed only in the buildpath?

mistydemeo: I gave it a shot, but SDL_image isn't building: https://gist.github.com/fc4f62f12ad00ae68333b4e8c5db6fb0

Formula/raine.rb
```diff
@@ -0,0 +1,158 @@
...
+  depends_on "nasm" => :build
+  depends_on "pkg-config" => :build
+
+  def install
+    ENV.m32
+
+    ENV.prepend "PATH", "#{libexec}/bin", File::PATH_SEPARATOR
+    ENV.append_to_cflags "-I#{libexec}/include"
+    ENV.append "LDFLAGS", "-L#{libexec}/lib"
+    ENV.prepend "PKG_CONFIG_PATH", "#{libexec}/lib/pkgconfig", File::PATH_SEPARATOR
```
ilovezfs: This seems highly unusual. Do we have a brew bug?

MikeMcQuaid:
> mind making it a week today (18th AEST)?

This has rolled by and given it's still not building and wouldn't pass our notability requirements for a new formula I think it's worth punting on this. Still pretty good that all but one of the homebrew/games formulae has made it in, though.

Formula/raine.rb
mistydemeo: `ENV.prepend_path` and `ENV.append_path` silently do nothing if the passed directory doesn't exist, which it won't yet in this circumstance. I could also have used `prepend_create_path` instead of forcing the `prepend`.

mistydemeo:
Sorry for not commenting. I've pushed a working build for a dynamically-built Raine, and I'm 95% to a statically-built one. I'm testing a build now.

mistydemeo:
Pushed a static formula.

mistydemeo:
The only thing broken is linking against the system libbIp2.

mistydemeo:
Build works. I'll update the audit warnings.

Wrt notability, Raine has been around since the 90s; the move to GitHub is a recent one.

Formula/raine.rb
ilovezfs: Yes, please either use `ENV.prepend_create_path` or defer calling `ENV.prepend_path` until the directory actually exists.

mistydemeo:
Merged!

ilovezfs:
:confetti_ball:

MikeMcQuaid:
Nice work both of you 🎉!

PR #9759, opened 2017-02-08
openrct2: import from homebrew/games

ilovezfs:

### DIFF
--- a/Formula/openrct2.rb
+++ b/Formula/openrct2.rb
@@ -1,0 +1,36 @@
+class Openrct2 < Formula
+  desc "Open source re-implementation of RollerCoaster Tycoon 2"
+  homepage "https://openrct2.website"
+  url "https://github.com/OpenRCT2/OpenRCT2/archive/v0.0.6.tar.gz"
+  sha256 "c4e04359ef91bedf58fbd9da57c39ba59066840a467aa0346c4a6353b0ae33f5"
+  head "https://github.com/OpenRCT2/OpenRCT2.git", :branch => "develop"
+
+  depends_on "cmake" => :build
+  depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):

LRFLEW:
I just noticed this was merged when I was updating. I am a dev on the OpenRCT2 project, and I largely work with the macOS port. I wanted to put a few comments about the formula here.

1. I never knew there was a homebrew/games, nor that OpenRCT2 was on it. There was some discussion with a few devs about getting the project into homebrew some time ago, but I hadn't realized it had happened. (The lack of visibility is probably why you moved it to the main repo)
2. The CMake build on macOS is technically supported, but we tend to do a poor job of maintaining it. The CMake project is tested in CI for the Linux builds, but we don't have a CI for the macOS build, so we only learn that something's wrong with it when a dev manually tries it. We usually recommend using the Xcode project to create an app bundle (though that would then go in the cask group from what I understand). I can see about adding a macOS CMake build to the CI (though we already have a massive number of builds as-is).
3. I noticed the launch patch that was added for the data folder. The issue was largely missed because so few of us use the CMake build on macOS, and fewer still use the install target (I don't know of anybody who does). I can get a fix in for that for the next release.
4. The formula should get a way to install the develop version of the project. Our develop builds are built off of the current head of the Git repository, and is triggered on every push (we use CI to automate the process). Many players use that instead of the releases, and with the multiplayer requiring players to all be on (roughly) the same version, adding a way to build the develop version will make this formula more complete.

ilovezfs:
>We usually recommend using the Xcode project to create an app bundle (though that would then go in the cask group from what I understand

This is the case when upstream itself provides a .app. Is that something OpenRCT2 would be able to do?

PR #9765, opened 2017-02-08
streem 201611 (new formula)

yamamuteki:

-----

### DIFF
--- a/Formula/streem.rb
+++ b/Formula/streem.rb
@@ -1,0 +1,17 @@
+class Streem < Formula
+  desc "prototype of stream based programming language"
+  homepage "https://github.com/matz/streem"
+  url "https://github.com/matz/streem/archive/201611.tar.gz"
+  sha256 "807a9ccf3162efb061a15a333943050b329f0232cceb55492e5cd582c09a7fb4"
+  head "https://github.com/matz/streem.git"
+
+  def install
+    system "make"
+    cp_r "bin", prefix
+  end
+
+  test do
+    hello_text = shell_output("#{bin}/streem -e '[\"hello\"] | stdout'")
+    assert_equal "hello\n", hello_text
+  end
+end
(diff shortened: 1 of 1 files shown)


Formula/streem.rb
```diff
@@ -0,0 +1,17 @@
+class Streem < Formula
+  desc "prototype of stream based programming language"
+  homepage "https://github.com/matz/streem"
+  url "https://github.com/matz/streem/archive/201611.tar.gz"
+  sha256 "807a9ccf3162efb061a15a333943050b329f0232cceb55492e5cd582c09a7fb4"
+  head "https://github.com/matz/streem.git"
+
+  def install
+    system "make"
+    cp_r "bin", prefix
```
dunn: Is there no `make install`? Otherwise, change this to `prefix.install "bin"`.

Formula/streem.rb
```diff
@@ -0,0 +1,17 @@
+class Streem < Formula
+  desc "prototype of stream based programming language"
```
dunn: Capitalize "Prototype".

Formula/streem.rb
yamamuteki: Thanks. I changed to `prefix.install "bin"` because there was not `make install`.

Formula/streem.rb
yamamuteki: Thanks. I fixed capitalize.

Formula/streem.rb
dunn: Would you mind asking the developers if they could add an install target?

yamamuteki: I will ask the developers, but can not you just merge with this? Probably I will make it take time, including pull requests.

dunn:
From https://github.com/matz/streem/issues/157:
> Although streem is not stable enough for daily use.

We don't accept formulae until they have stable releases.  Feel free to resubmit when that's available!

yamamuteki:
OK. Thanks.

PR #9777, opened 2017-02-09
Kibana: fix plugin script

### LINKED ISSUE #9770: kibana 5.2.0 formula doesn't work properly
Kibana is now at its 5.2.0 version.
Even if `homebrew` after a `brew update/brew upgrade` command says that the kibana package is upgraded to 5.2.0 it is actually still at its 5.0.0 version as shown in [kibana formula link](http://brewformulas.org/Kibana).

The error that made me realise is shown in this [issue](https://github.com/elastic/kibana/issues/10240)

jasontedor:
This commit fixes the plugin script to use the Kibana-installed node.js
instead of node.js from the path which could be the wrong version of
node.js.

Closes #9770

### DIFF
--- a/Formula/kibana.rb
+++ b/Formula/kibana.rb
@@ -49,6 +49,7 @@ class Kibana < Formula
     prefix.install Dir["build/kibana-#{version}-#{platform.sub("x64", "x86_64")}/{bin,config,node_modules,optimize,package.json,src,webpackShims}"]
 
     inreplace "#{bin}/kibana", %r{/node/bin/node}, "/libexec/node/bin/node"
+    inreplace "#{bin}/kibana-plugin", %r{/node/bin/node}, "/libexec/node/bin/node"
 
     cd prefix do
       inreplace "config/kibana.yml", "/var/run/kibana.pid", var/"run/kibana.pid"


ilovezfs (approved):

Formula/kibana.rb
```diff
@@ -49,6 +49,7 @@ def install
     prefix.install Dir["build/kibana-#{version}-#{platform.sub("x64", "x86_64")}/{bin,config,node_modules,optimize,package.json,src,webpackShims}"]
 
     inreplace "#{bin}/kibana", %r{/node/bin/node}, "/libexec/node/bin/node"
+    inreplace "#{bin}/kibana-plugin", %r{/node/bin/node}, "/libexec/node/bin/node"
```
MikeMcQuaid: Could maybe do a `Dir["#{bin}/*"].each` to cover any future libraries that were added in here? I have no strong feelings either way.

jasontedor: Thank you for the suggestion @MikeMcQuaid. For now I would prefer to keep it explicit like this. I do know that right now there is nothing in the works that would lead to there being a third script here. If this changes in the future, I think your suggestion would be a better approach.

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
Thanks again @jasontedor!

jasontedor:
Thanks @MikeMcQuaid!

salvob41:
@jasontedor hi, is this really fixed? because the same error keeps showing if I want to install ¡x-pack` with Kibana plugin (installed by brew)

jasontedor:
Relates #11360.

PR #9582, opened 2017-02-02
rust 1.15.0

mrageh:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ ] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [ ] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #9778, opened 2017-02-09
rust: fix checksum

JCount:
Reported to homebrew: https://github.com/Homebrew/homebrew-core/pull/9582#issuecomment-278526514
Upstream's response to query about checksum change: https://github.com/Homebrew/homebrew-core/pull/9582#issuecomment-278535967
Reason:
> we've been having some infrastructure hiccups unfortunately and we think we may have accidentally overwritten some artifacts. Sorry if that's causing any problems! I can confirm though that f655e4fac9c2abb93eb579e29c408e46052c0e74b7655cd222c63c6743457673 is now the current correct sha256 of the source tarball. - alexcrichton

### DIFF
--- a/Formula/rust.rb
+++ b/Formula/rust.rb
@@ -4,7 +4,7 @@ class Rust < Formula
 
   stable do
     url "https://static.rust-lang.org/dist/rustc-1.15.0-src.tar.gz"
-    sha256 "33f3747d231ede34b56c6cc0ae6be8cbaa29d2fdb39d86f25693dceb9fc5f164"
+    sha256 "f655e4fac9c2abb93eb579e29c408e46052c0e74b7655cd222c63c6743457673"
 
     resource "cargo" do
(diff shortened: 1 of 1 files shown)


alexcrichton:
Again, sorry for the troubles here! 

FWIW the old archive is still available at https://static.rust-lang.org/dist/2017-01-31/rustc-1.15.0-src.tar.gz (e.g. the old checksum matches https://static.rust-lang.org/dist/2017-01-31/rustc-1.15.0-src.tar.gz.sha256). The new archive is also at https://static.rust-lang.org/dist/2017-02-08/rustc-1.15.0-src.tar.gz which matches the checksum in this PR (available at https://static.rust-lang.org/dist/2017-02-08/rustc-1.15.0-src.tar.gz.sha256)

ilovezfs:
@alexcrichton Thanks! No worries.

ilovezfs (approved):

PR #9803, opened 2017-02-09
postgresql 9.6.2

neutric:

-----

### DIFF
--- a/Formula/check_postgres.rb
+++ b/Formula/check_postgres.rb
@@ -3,7 +3,7 @@ class CheckPostgres < Formula
   homepage "https://bucardo.org/wiki/Check_postgres"
   url "https://bucardo.org/downloads/check_postgres-2.22.0.tar.gz"
   sha256 "29cd8ea0a0c0fcd79a1e6afb3f5a1d662c1658eef207ea89276ddb30121b85a8"
-  revision 1
+  revision 2
 
   head "https://github.com/bucardo/check_postgres.git"
 
--- a/Formula/citus.rb
+++ b/Formula/citus.rb
@@ -3,6 +3,7 @@ class Citus < Formula
   homepage "https://www.citusdata.com"
   url "https://github.com/citusdata/citus/archive/v6.0.1.tar.gz"
   sha256 "642c194ad229a30f9e960a78d4400566fc08dec3fce1694990592d964d26bf3d"
+  revision 1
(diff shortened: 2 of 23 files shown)


neutric:
I'll bump the revision for packages in `brew uses postgresql` as soon as the above is green.

neutric:
The source file for `pg_top` is not accessible at the moment, as pgFoundry, the host, is undergoing a rebuild from scratch. That'll make the revision built fail. Maybe get it from [here](https://github.com/markwkm/pg_top)? Furthermore, `pgcli`, `pgpool-ii`, `postgis`, and `temporal_tables`, have new versions, and [`libpqxx`](http://pqxx.org/development/libpqxx/) [recommends building from trunk](http://pqxx.org/development/libpqxx/wiki/DownloadPage). Should those formulae be updated here or can that be done in a separate PR once this is merged?

neutric:
Zabbix DL failed for Yosemite, the rest looks fine.

neutric:
Related PRs: #9810, #9817

neutric:
@ilovezfs Can you selectively restart Jenkins for Yosemite only?

ilovezfs:
No, but we can just do zabbix in a separate PR

neutric:
So remove it here by resetting git to the commit before zabbix and then force push? Would a testthisplease instead be too burdensome on the infrastructure? And what about those formulae using postgresql that have newer versions? If I remember correctly from yesterday, at least one of them didn't look straight-forward, which is why I wanted to keep them out of here.

ilovezfs:
>So remove it here by resetting git to the commit before zabbix and then force push?

Nope, you don't need to do anything.

>Would a testthisplease instead be too burdensome on the infrastructure?

Not really, but it's not necessary.

>And what about those formulae using postgresql that have newer versions?

Which are you thinking of? You can open PRs for them after this ships.

ilovezfs:
@neutric thanks for doing this big PR. Mind opening a `zabbix: revision for postgresql` PR now? Note the revision subjects usually don't include the version number and have a lower case r in revision (I standardized them before pushing this PR).

neutric:
@ilovezfs Thanks for fixing the subjects! Working on the zabbix PR...

ilovezfs:
@neutric note you can use https://fossies.org/linux/misc/zabbix-3.2.3.tar.gz as a mirror for it too

neutric:
Will do!

PR #9840, opened 2017-02-10
micro 1.1.4

zyedidia:

-----

### DIFF
--- a/Formula/micro.rb
+++ b/Formula/micro.rb
@@ -4,8 +4,8 @@ class Micro < Formula
   desc "Modern and intuitive terminal-based text editor"
   homepage "https://github.com/zyedidia/micro"
   url "https://github.com/zyedidia/micro.git",
-    :tag => "v1.1.3",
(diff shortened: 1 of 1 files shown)


ilovezfs:
@zyedidia are you considering using a go vendor directory in the future? That would be great from a downstream perspective.

Also, in the absence of vendoring, currently "make build" is running
```
==> make build
go get -d ./cmd/micro
```
which could potentially invoke a naked `go get` if someone fails to fully update the resource blocks in the future.

Do you mind changing it to use `make build-quick`, which appears to skip the deps target?

zyedidia:
I have considered using govendor at various points but I never really put in the time to actually use it in the project. Honestly I think it's unlikely that I switch to govendor, especially considering that an official dependency management system for Go might be coming soon.

I have now switched the install to use `build-quick` instead of `build`.

ilovezfs:
Thanks!

>considering that an official dependency management system for Go might be coming soon

cool

ilovezfs:
Shipped! Thanks @zyedidia

PR #9852, opened 2017-02-10
Deprecate keybase Formula as Cask installs CLI

joshka:
Keybase is now installed as an App, which installs the CLI during installation and hence should only be available as a Cask, not a Formula.

The cask up to date with the latest version of the app, as well as installing the CLI during installation rather than first run of the App so the CLI can be used immediately. See https://github.com/caskroom/homebrew-cask/pull/29956

CC: @malgorithms

---


MikeMcQuaid:
Is this primarily used as a GUI or CLI?

joshka:
Both. The CLI is older though. This is relevant now as Keybase [recently added a chat feature](https://keybase.io/blog/keybase-chat) to the GUI / command line. I'm not sure how long the app has been around as I've only ever seen the CLI tool before this came out.

https://github.com/keybase/client/issues/215#issuecomment-75612070 seems to indicate a requirement by the Keybase authors to install both CLI and GUI at the same time.
>In the long run, what we should aim for is one install of Keybase that gets you everything.
>
>the GUI
>the CLI, properly in your path
>kbfs + keybased, both set to start up automatically
>So, ideally, someone has keybase or not. Not parts of it, or a mishmash of versions they might piece together. Obviously our requirements as devs may be different.

The [official download](https://keybase.io/download) is a .dmg, with an app to drag to the Applications folder. Running the app for the first time installs the CLI tool (I've added that install logic to the cask instead).

ilovezfs:
This sounds like it's primarily CLI if the GUI is just a recent feature addition.

joshka:
@ilovezfs the GUI has been around for some time (the comment above mentions it in February 2015, and other commits seem to suggest that it's at least a year old). I installed it via homebrew and didn't realize it had a GUI.

The primary non-homebrew installation instructions install an app, which secondarily installs the CLI. To me this says the GUI is the primary, with a CLI as a supporting artifact. @cjb or @gabriel might have more info on this.

ilovezfs:
According to the analytics, we have 490 installs in the last 30 days, which indicates to me there are more than a few users just using the CLI.

joshka:
@ilovezfs is that number new installs or does it include existing installs that are upgrades etc.? I cannot speak for the keybase devs on this, but it seems like the website docs, previous messaging from devs, etc. is clear, the cli is a part of the app installation, not the other way around. When keybase first came out it was a cli app and the installation instructions on the website mentioned installing via homebrew. The current installation instructions no longer mention homebrew as the installation technique and refer to keybase as a gui app.

From https://keybase.io/docs/the_app/install_macos:
> Steps
>
> 1. Buy an Apple computer.
> 2. Download and open Keybase.dmg.
> 3. Drag Keybase into your Applications folder & run it.

ilovezfs:
419 are install_on_request events.

joshka:
My understanding of the Cask vs Formula split is programs that have apps generally belong in cask, while command line apps belong in brew. I don't understand how you determine that this belongs in here rather than only in cask. It makes sense to me that this be should be deprecated.

Another point to note is that installing the app overwrites the version installed by homebrew regardless of whether the app is installed manually or via hbc.

ilovezfs:
That have a .app that is writing binaries into /usr/local/bin? :see_no_evil:

joshka:
yes

ilovezfs:
What a mess

ilovezfs:
```
vm1011:~ joe$ which -a keybase
/usr/local/bin/keybase
vm1011:~ joe$ ls -l /usr/local/bin/keybase 
lrwxr-xr-x  1 joe  admin  60 Feb 11 19:47 /usr/local/bin/keybase -> /Applications/Keybase.app/Contents/SharedSupport/bin/keybase
vm1011:~ joe$
```
yikes.

joshka:
If this is merged, effectively that mess goes away... things just work.

MikeMcQuaid:
Let's leave this open for a bit and see if there's strong opinions in favour or issues. I suspect as-is this will just annoy people.

ilovezfs:
Would be nice if upstream used `/etc/paths.d` to avoid the needless conflict.

joshka:
I contacted a few people on the keybase team that said they'd get back with us on this issue. The `/et/paths.d` approach does seem like a better option than symlinking however.

ilovezfs:
Thanks @joshka

gabriel:
Hey, @gabriel from Keybase. 

We'd prefer the cask install since we need to sign the app bundle and go binaries (this is important for privileged helper and keychain access). Also there are things, like the filesystem and chat that won't work without the app bundle. Also some of our newer users might be really confused if there is a keybase formula that only supports some really basic command line options and seems broken.

gabriel:
Regarding the symlink... the app (bundle) install does symlink `keybase` into /usr/local/bin (essentially unlinking if there is an existing keybase formula), which is important for things to work, if people upgrade from the brew formula. Also it turns out a lot of people override the behavior of the path_helper (for /etc/paths.d) in their terminal programs, so we only use that method if /usr/local/bin isn't available.

MikeMcQuaid:
Ok, will defer to upstream here. Thanks @joshka!

krono:
Note to @gabriel: The old non-cask version works for Mavericks, the cask one does not, because the dmg-version does not. I know Mavericks is "not supported" but making it even harder to install is discouraging :(

cjb:
Hm, I'm not @gabriel, but Mavericks doesn't get security updates any more, is that right? Keybase is security software dealing with crypto keys -- I think it's probably a good thing that it's difficult to install on OS versions without vulnerability patches available.

krono:
@cjb Mavericks security updates ended in late 2016, right. For reasons not relevant here, an upgrade is no way for me. However, just doing a `brew update` today picked up this change and just removed keybase from my Mac. Is this expected behavior?

cjb:
Ah, bummer -- no, I hadn't expected that to happen. But I don't expect we'll want to change it given the reasoning above. Sorry for the nasty surprise.

krono:
Yes, I figured that…  
I just copied the old formular to my own tap, wich works for now…

specious:
I found this discussion because I was confused why I have both a brew formula and a cask named `keybase` listed as installed simultaneously.

`brew edit keybase` pulled up the install script referencing `https://github.com/keybase/client/archive/v1.0.18.tar.gz` as the source, meanwhile my active keybase version reports back as:

```
$ keybase --version
keybase version 1.0.38-20171220205307+f5d54bc77
```

After reading this thread, I ran `brew uninstall keybase`, which fortunately left my newer cask install perfectly intact:

```
$ ls -l `which keybase`
lrwxr-xr-x  1 specious  admin  60 Feb 22  2017 /usr/local/bin/keybase -> /Applications/Keybase.app/Contents/SharedSupport/bin/keybase
```

Now the legacy `keybase` formula is unambiguously gone from my system:

```
$ brew info keybase
Error: No available formula with the name "keybase"
It was migrated from homebrew/core to caskroom/cask.
You can access it again by running:
  brew tap caskroom/cask

$ brew edit keybase
Error: No available formula with the name "keybase"
```

I figured I'd contribute this empirical observation to assist in bringing clarity and to alleviate confusion.

PR #9613, opened 2017-02-03
wget 1.19

ilovezfs:
Created with `brew bump-formula-pr`.

PR #9873, opened 2017-02-11
wget 1.19.1

zmwangx:

-----

- Make `openssl` mandatory and `libidn` recommended per discussions in #9613;
- Test: write retrieved page to `/dev/null` (the hundreds of kilobytes of output on stdout is useless).

### DIFF
--- a/Formula/wget.rb
+++ b/Formula/wget.rb
@@ -3,9 +3,9 @@
 class Wget < Formula
   desc "Internet file retriever"
   homepage "https://www.gnu.org/software/wget/"
-  url "https://ftpmirror.gnu.org/wget/wget-1.18.tar.xz"
-  mirror "https://ftp.gnu.org/gnu/wget/wget-1.18.tar.xz"
-  sha256 "b5b55b75726c04c06fe253daec9329a6f1a3c0c1878e3ea76ebfebc139ea9cc1"
+  url "https://ftpmirror.gnu.org/wget/wget-1.19.1.tar.gz"
+  mirror "https://ftp.gnu.org/gnu/wget/wget-1.19.1.tar.gz"
+  sha256 "9e4f12da38cc6167d0752d934abe27c7b1599a9af294e73829be7ac7b5b4da40"
 
   bottle do
     sha256 "26c26f28a4fad3f57602722219e9c70574c4d8191e59639052ce02b3e9d84508" => :sierra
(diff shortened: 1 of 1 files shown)


Formula/wget.rb
```diff
@@ -23,16 +23,15 @@ class Wget < Formula
...
-  option "with-iri", "Enable iri support"
   option "with-debug", "Build with debug support"
+  option "without-libidn", "Build without IRI support"
 
   depends_on "pkg-config" => :build
   depends_on "pod2man" => :build if MacOS.version <= :snow_leopard
-  depends_on "openssl" => :recommended
-  depends_on "libidn" if build.with? "iri"
+  depends_on "openssl"
+  depends_on "libidn" => :recommended
```
MikeMcQuaid: What does `libidn` do? Not sure what IRI support is but I'd personally rather avoid the extra dependency given how infrequently this option is specified: 

screen shot 2017-02-11 at 21 32 01

zmwangx:
@MikeMcQuaid [Internationalized domain name](https://en.wikipedia.org/wiki/Internationalized_domain_name).

"IRI support" is somewhat confusing because Wget can handle Unicode in path, query and fragment (which is simply a matter of percent-encoding) just fine without libidn; however, it requires libidn to support Unicode characters in the domain name.

As I said in https://github.com/Homebrew/homebrew-core/pull/9613#issuecomment-277555936, I'm recommending libidn on principle: 1. Wget defaults to supporting IDNs (when libidn is available, that is); 2. The web is increasingly internationalized; I've certainly encountered IDNs in the wild, and I expect to see more. (I do realize you can translate IDNs elsewhere and feed the ASCII version to a Wget without IDN support, e.g., copying from browser's address bar.)

However, I just realized that Wget 1.19 dropped support for libidn in favor of libidn2 (commit https://github.com/bettermirror/wget/commit/00ae9b4ee2caa6640776e9777e7ef5a374d082c0). Therefore, I removed libidn-related code from the formula.

Someone could contribute a libidn2 formula to bring the option back.

ilovezfs:
Also probably important to note the security concerns with idn: https://en.wikipedia.org/wiki/Internationalized_domain_name#ASCII_spoofing_concerns

>The use of Unicode in domain names makes it potentially easier to spoof web sites as the visual representation of an IDN string in a web browser may make a spoof site appear indistinguishable to the legitimate site being spoofed, depending on the font used. For example, Unicode character U+0430, Cyrillic small letter a, can look identical to Unicode character U+0061, Latin small letter a, used in English. As a concrete example, using Cyrillic letters а, е (“Ie”/“Ye”, U+0435, looking essentially identical to Latin letter e), Belarusian-Ukrainian і (U+0456, essentially identical to Latin letter i), р (“Er”, U+0440, essentially identical to Latin letter p), we form the URL wіkіреdіа.org (xn--wkd-8cdx9d7hbd.org in encoded form), which is virtually indistinguishable from the visual representation of the legitimate wikipedia.org (possibly depending on fonts).

MikeMcQuaid (approved):

MikeMcQuaid:

PR #9875, opened 2017-02-11
compcert 3.0

beurdouche:

-----

### DIFF
--- a/Formula/compcert.rb
+++ b/Formula/compcert.rb
@@ -1,9 +1,8 @@
 class Compcert < Formula
   desc "Formally verified C compiler"
   homepage "http://compcert.inria.fr"
-  url "http://compcert.inria.fr/release/compcert-2.7.1.tgz"
-  sha256 "446199fb66c1e6e47eb464f2549d847298f3d7dcce9be6718da2a75c5dd00bee"
-  revision 1
+  url "http://compcert.inria.fr/release/compcert-3.0.tgz"
+  sha256 "f5e42c528849fffa840a4c820ad4cc6257a7e518f3b3b0fe54c06fa60270dd0d"
 
   bottle do
     cellar :any_skip_relocation
@@ -12,15 +11,11 @@ class Compcert < Formula
     sha256 "d1b1c7c3ea3edf4e9f40b9c30d3d3bc4786e927dd8659928c697c0eb673c33c7" => :yosemite
   end
 
+  option "with-config-x86_64", "Build Compcert with ./configure 'x86_64'"
+
   depends_on "menhir" => :build
   depends_on "ocaml" => :build
   depends_on "opam" => :build
-
-  # remove for > 2.7.1; allow Coq version 8.5pl3
-  patch do
-    url "https://github.com/AbsInt/CompCert/commit/a8f87aa.patch"
-    sha256 "fb1b35503ae106a28b276521579fcf862772615414dca3ae3fabc4ed736ab5de"
-  end
 
   def install
     ENV.permit_arch_flags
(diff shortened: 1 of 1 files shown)


beurdouche:
? @ilovezfs

ilovezfs:
@beurdouche sorry for not commenting right away. So ... a couple things. First, we never put `revision 0`. The counting of revisions does start from 0, but we only add an explicit revision when it's >= 1.

Second, are you sure we can flat out replace the ia32-macosx with x86_64-macosx, and don't need to build both or make one optional and the other default? I saw in the Changelog that it said:

>The x86 port of CompCert was extended to produce x86-64 bit code in
  addition to the original x86-32 bit (IA32) code.  (This is the first
  instantiation of the new support for 64-bit targets described
  above.) Support for x86-64 is currently available for Linux and MacOS X.
  (Run the configure script with 'x86_64-linux' or 'x86_64-macosx'.)
  This is an early port: several ABI incompatibilities remain.

(http://compcert.inria.fr/release/Changelog)

beurdouche:
@ilovezfs no worries. 
After thinking a bit more about it, I think you are right. It might be more reasonable to keep `ia32-macosx` as the default for this release and offer the option for building `x86_64`.
I'll discuss with Xavier Leroy (the author) to make sure that we can switch default and option in a future release.
Also I just noticed a bug in the version number (hasn't been incremented to 3.0.0) and reported it upstream. Maybe we can do a patch just to fix the VERSION file ?
WDYT ?

ilovezfs:
Yup. All sounds correct to me. If the version number does get patched upstream right away, but a new tarball isn't published, then it would be best to apply the commit as a patch, but if not an `inreplace` will work.

beurdouche:
Ok, the version problem has been fixed and the release has been re-tagged upstream, so no patch necessary.

Remains: 
[] updating the script with the new tarball/hash
[] creating the option to build with ia32 or x86_64

I'll try to do that tomorrow except if someone manages to do it before..

ilovezfs:
@beurdouche did you get a chance to talk with Xavier?

beurdouche:
@ilovezfs Ok I discussed with Xavier by email. He suggested to separate Compcert32 (more stable) and Compcert64 (more experimental) for now, so I proceeded by adding an option to compile `x86_64` if the user chooses it. Please note that while it seems to work properly I am not super familiar with options for Homebrew and, in particular, naming conventions. The PR is currently ready to be merged from my perspective, please don't hesitate to make adjustments where you see fit ;)

beurdouche:
Ps: just squashed the commits

Formula/compcert.rb
```diff
@@ -36,8 +31,15 @@ def install
     (Pathname.pwd/"opamroot").mkpath
     system "opam", "init", "--no-setup"
     system "opam", "install", "coq=8.5.3"
-    system "opam", "config", "exec", "--",
-           "./configure", "-prefix", prefix, "ia32-macosx"
+
+    if build.without? "config-x86_64"
```
MikeMcQuaid: Use `build.with?` instead; it's a little easier to follow.

MikeMcQuaid (changes requested):

MikeMcQuaid:
> Compcert64 (more experimental) for now, so I proceeded by adding an option to compile x86_64 if the user chooses it.

Maybe it's worth removing the option unless you personally use it?

ilovezfs:
The option is the cool new toy.

beurdouche:
@MikeMcQuaid @ilovezfs, I actually suspect that the option will be very used, yes :)
JustChanged the code to use `build.with ?` instead, thank you for the suggestion.

MikeMcQuaid:
> @MikeMcQuaid @ilovezfs, I actually suspect that the option will be very used, yes :)

@beurdouche Will you personally be using it or not?

beurdouche:
@MikeMcQuaid Yes, I am using it, as well as half of the lab.

ilovezfs:
@BrewTestBot test this please

MikeMcQuaid:
@beurdouche In that case I might be in favour for saying it should just be the default?

ilovezfs:
@MikeMcQuaid upstream has basically explicitly said that it's not considered stable yet and there are API incompatibilities.

beurdouche:
@MikeMcQuaid we discussed it but we think it is too early. We will probably change the default in 3.1 or 3.2 after another discussion with the authors.

MikeMcQuaid (approved):

ilovezfs:
@BrewTestBot test this please

beurdouche:
@BrewTestBot test this please

beurdouche:
@ilovezfs Good to go ;)

MikeMcQuaid:

beurdouche:
@ilovezfs @MikeMcQuaid Thanks a lot to both of you ! ;)

PR #9877, opened 2017-02-12
yosys 0.7 (new formula)

joshhead:
Yosys is a framework for Verilog RTL synthesis. It is a part of an
open source tool chain for FPGA development with Lattice iCE40
FPGA's. See Project IceStorm at http://www.clifford.at/icestorm/
It is also supports synthesis for several other targets.

It has an (optional) dependency on ABC, which is used for some
advanced optimizations. Not all versions of Yosys work with all
versions of ABC, so Yosys includes its own ABC binary. It normally
checks out a mercurial repository and builds a specific ABC commit
from source. This recipe uses a tar.gz resource for the ABC source
instead. I didn't try to make ABC a separate recipe since it has no
tagged releases.

-----

### DIFF
--- a/Formula/yosys.rb
+++ b/Formula/yosys.rb
@@ -27,13 +27,9 @@ class Yosys < Formula
 
   def install
     args = []
-
     resource("abc").stage buildpath/"abc"
-
     args << "ENABLE_PLUGINS=0" if build.without? "libffi"
-
     args << "ENABLE_READLINE=0" if build.without? "readline"
-
     system "make", "install", "PREFIX=#{prefix}", "PRETTY=0", "ABCREV=default", *args
   end
 
--- a/Formula/yosys.rb
+++ b/Formula/yosys.rb
@@ -1,0 +1,77 @@
+class Yosys < Formula
+  desc "Framework for Verilog RTL synthesis"
+  homepage "http://www.clifford.at/yosys/"
+  url "https://github.com/cliffordwolf/yosys/archive/yosys-0.7.tar.gz"
+  sha256 "3df986d0c6bf20b78193456e11c660f2ad935cc126537c2dc5726e78896d6e6e"
+
+  option "without-abc", "Build without support for ABC"
+  option "without-tcl", "Build without support for Tcl language bindings"
+
+  depends_on "python3"
+  depends_on "libffi" => :recommended
+  depends_on "readline" => :recommended
+  depends_on "pkg-config" => :build
+  depends_on "bison" => :build
+
+  # Prevent the makefile from trying to automatically locate homebrew and macports libraries
+  patch :DATA
+
+  # This ABC revision is specified in the makefile.
+  # The makefile by default checks it out using mercurial,
+  # but this recipe instead downloads a tar.gz archive.
+  resource "abc" do
+    url "https://bitbucket.org/alanmi/abc/get/eb6eca6807cc.tar.gz"
+    sha256 "ae9acddad38a950d48466e2f66de8116f2d21d03c78f5a270fa3bf77c3fd7b5b"
+  end
+
+  def install
+    extra_args = []
+
+    if build.with? "abc"
+      resource("abc").stage buildpath/"abc"
+      extra_args.push "ABCREV=default"
+    else
+      extra_args.push "ENABLE_ABC=0"
+    end
+
+    if build.without? "tcl"
+      extra_args.push "ENABLE_TCL=0"
+    end
+
+    if build.without? "readline"
+      extra_args.push "ENABLE_READLINE=0"
+    end
+
+    if build.without? "libffi"
+      extra_args.push "ENABLE_PLUGINS=0"
+    end
+
+    system "make", "install", "PREFIX=#{prefix}", "PRETTY=0", *extra_args
+  end
+
+  test do
+    system "#{bin}/yosys", "-p", "hierarchy; proc; opt; techmap; opt;", "-o", "synth.v", "#{pkgshare}/adff2dff.v"
+  end
+end
+
+__END__
+diff --git a/Makefile b/Makefile
+index 0a61fe65..061102b3 100644
+--- a/Makefile
++++ b/Makefile
+@@ -58,15 +58,6 @@ SED = sed
+ BISON = bison
+ 
+ ifeq (Darwin,$(findstring Darwin,$(shell uname)))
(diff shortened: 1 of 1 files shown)


Formula/yosys.rb
```diff
@@ -0,0 +1,77 @@
...
+  option "without-tcl", "Build without support for Tcl language bindings"
+
+  depends_on "python3"
+  depends_on "libffi" => :recommended
+  depends_on "readline" => :recommended
+  depends_on "pkg-config" => :build
+  depends_on "bison" => :build
+
+  # Prevent the makefile from trying to automatically locate homebrew and macports libraries
+  patch :DATA
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Formula/yosys.rb
```diff
@@ -0,0 +1,77 @@
+class Yosys < Formula
+  desc "Framework for Verilog RTL synthesis"
+  homepage "http://www.clifford.at/yosys/"
+  url "https://github.com/cliffordwolf/yosys/archive/yosys-0.7.tar.gz"
+  sha256 "3df986d0c6bf20b78193456e11c660f2ad935cc126537c2dc5726e78896d6e6e"
+
+  option "without-abc", "Build without support for ABC"
+  option "without-tcl", "Build without support for Tcl language bindings"
```
MikeMcQuaid: When would you want to disable these? Unless there's a great reason and/or it takes ages to build them: let's kill the options.

Formula/yosys.rb
```diff
@@ -0,0 +1,77 @@
...
+  # This ABC revision is specified in the makefile.
+  # The makefile by default checks it out using mercurial,
+  # but this recipe instead downloads a tar.gz archive.
+  resource "abc" do
+    url "https://bitbucket.org/alanmi/abc/get/eb6eca6807cc.tar.gz"
+    sha256 "ae9acddad38a950d48466e2f66de8116f2d21d03c78f5a270fa3bf77c3fd7b5b"
+  end
+
+  def install
+    extra_args = []
```
MikeMcQuaid: `args = []`

Formula/yosys.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    if build.with? "abc"
+      resource("abc").stage buildpath/"abc"
+      extra_args.push "ABCREV=default"
```
MikeMcQuaid: `args << "ABCREV=default"`

Formula/yosys.rb
```diff
@@ -0,0 +1,77 @@
...
+    else
+      extra_args.push "ENABLE_ABC=0"
+    end
+
+    if build.without? "tcl"
+      extra_args.push "ENABLE_TCL=0"
```
MikeMcQuaid: `args << "ENABLE_TCL=0" if build.without? "tcl"`

Formula/yosys.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    if build.without? "readline"
+      extra_args.push "ENABLE_READLINE=0"
```
MikeMcQuaid: Similarly one-line this.

Formula/yosys.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    if build.without? "libffi"
+      extra_args.push "ENABLE_PLUGINS=0"
```
MikeMcQuaid: Similarly one-line this.

MikeMcQuaid (changes requested):

Formula/yosys.rb
joshhead: I included the options because the Makefile supports them and because the :recommended dependencies can be disabled. There's no great need that I can think of. I'll take them out.

Formula/yosys.rb
joshhead: There is a recent upstream change that addresses this issue. I'll make the comment clearer and add a link.

https://github.com/cliffordwolf/yosys/pull/303

With the upstream changes, brew or macports libraries search paths are automatically added only if `brew` or `port` are found on $PATH. So I think it won't be a problem when building in the sandbox.

In general, is there a recommended way for build processes to automatically locate libraries from homebrew or other third-party package managers? Is it discouraged?

Formula/yosys.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def install
+    args = []
+
+    resource("abc").stage buildpath/"abc"
+
+    args << "ENABLE_PLUGINS=0" if build.without? "libffi"
+
+    args << "ENABLE_READLINE=0" if build.without? "readline"
+
```
MikeMcQuaid: Can remove all the spaces above.

Formula/yosys.rb
MikeMcQuaid: > In general, is there a recommended way for build processes to automatically locate libraries from homebrew or other third-party package managers? Is it discouraged?

Using `brew --prefix` etc. is the best bet.

Formula/yosys.rb
```diff
@@ -0,0 +1,116 @@
...
+  depends_on "readline" => :recommended
+  depends_on "pkg-config" => :build
+  depends_on "bison" => :build
+
+  # The makefile in Yosys 0.7 adds library search paths from macports, which a homebrew build
+  # should not be using. It also prints warnings about a missing brew command.
+  # See https://github.com/cliffordwolf/yosys/pull/303 for discussion.
+  # The patch is based on the Makefile changes in this upstream commit:
+  # https://github.com/cliffordwolf/yosys/commit/a431f4ee311b9563f546201d255e429e9ce58cfa
+  patch :DATA
```
MikeMcQuaid: Does `https://github.com/cliffordwolf/yosys/commit/a431f4ee.patch` or `https://github.com/cliffordwolf/yosys/pull/303.patch` work at all? CC @ilovezfs for thoughts on whether this would be good to have in the formula-patches repo.

MikeMcQuaid (changes requested):

Formula/yosys.rb
joshhead: Neither patch applies, due to some other differences in `Makefile` and changes in `README.md` which does not exist (there is a `README`).

MikeMcQuaid (approved):

MikeMcQuaid:

joshhead:
Thanks!

PR #9879, opened 2017-02-12
Add wsk (OpenWhisk CLI binary) formula

shaftoe:
- [ ] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)? **Unfortunately seems almost impossible to follow them strictly, I first tried as a `Cask` but I've been asked to try here instead because even if it's distributed as a binary, it's maintained as an Open Source project**
- [ ] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting? **Same caveat as point 1**

-----

### DIFF
--- a/Formula/wsk.rb
+++ b/Formula/wsk.rb
@@ -1,0 +1,21 @@
+class Wsk < Formula
+  desc "OpenWhisk CLI"
+  homepage "https://console.ng.bluemix.net/openwhisk/cli"
+  url "https://openwhisk.ng.bluemix.net/cli/go/download/mac/amd64/OpenWhisk_CLI-mac.zip"
+  version "0.1.0"
+
+  def install
+    bin.install "wsk"
+  end
+
+  def caveats
+    <<-EOS.undent
+      Until OpenWhisk project does not provide an official version for the CLI tools,
+      to upgrade we must uninstall/install the formula.
+    EOS
+  end
+
+  test do
+    system "true"
+  end
+end


Formula/wsk.rb
```diff
@@ -0,0 +1,21 @@
+class Wsk < Formula
+  desc "OpenWhisk CLI"
+  homepage "https://console.ng.bluemix.net/openwhisk/cli"
+  url "https://openwhisk.ng.bluemix.net/cli/go/download/mac/amd64/OpenWhisk_CLI-mac.zip"
```
MikeMcQuaid: This needs to be built from source from a versioned source.

Formula/wsk.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    bin.install "wsk"
+  end
+
+  def caveats
+    <<-EOS.undent
+      Until OpenWhisk project does not provide an official version for the CLI tools,
+      to upgrade we must uninstall/install the formula.
+    EOS
+  end
```
MikeMcQuaid: Remove these caveats.

Formula/wsk.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  test do
+    system "true"
```
MikeMcQuaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

MikeMcQuaid (changes requested):

MikeMcQuaid:
> if it's distributed as a binary, it's maintained as an Open Source project

We need to build it from source. If it's open source this should be possible.

Formula/wsk.rb
shaftoe: I thought about that, but there's no `--version` or similar available from the CLI tool. Actually, there seems to be no official version at all for the OpenWhisk project: no Git tags, no (CHANGELOG|VERSION) file. Invoking the binary alone prints help to stdout and returns 0, I guess it's better then nothing

Formula/wsk.rb
shaftoe: The only version available is the commit id for the master branch, there are no tags nor other versions available in the project, or at least I can't find any. I guess the only option is https://github.com/openwhisk/openwhisk/archive/master.zip

shaftoe:
@MikeMcQuaid ok, I'll try to make it, but this will force the user to install Java + Gradle + Docker... feels very overkill for having only a small binary installed as suggested by the official documentation. Wouldn't this be a better fit as a Cask?

shaftoe:
I've spent some time trying to make it install via Brew, but failed (I managed to compile it from CLI though). It's failing with this message:
```
==> gradle :tools:cli:distDocker

FAILURE: Build failed with an exception.

* What went wrong:
Failed to load native library 'libnative-platform.dylib' for Mac OS X x86_64.
```

I'm sorry but I have to give up, I've got not enough time nor will to invest in this, I'll try again to submit the request to Cask. Thanks for your time!

markusthoemmes:
It's not dependend on Java, Gradle or Docker I think. It should be compilable using plain Go just fine.

PR #9933, opened 2017-02-13
boost*: import from homebrew/versions

JCount:
Goes along with Homebrew/homebrew-versions#1523


MikeMcQuaid:
@JCount I'd pull this as-is, fix conflicts/`universal` audit failures locally.

PR #9935, opened 2017-02-13
nss: fix heinous os specific install logic

carlorosati:
- [x ] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x ] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x ] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

Addresses downstream issue https://github.com/Linuxbrew/homebrew-core/issues/1620

### DIFF
--- a/Formula/nss.rb
+++ b/Formula/nss.rb
@@ -43,7 +43,7 @@ class Nss < Formula
     # rather than copying the referenced file.
     cd "../dist"
     bin.mkpath
-    Dir.glob("Darwin*/bin/*") do |file|
+    Dir.glob("*.OBJ/bin/*") do |file|
       cp file, bin unless file.include? ".dylib"
     end
 
@@ -53,7 +53,7 @@ class Nss < Formula
 
     lib.mkpath
     libexec.mkpath
-    Dir.glob("Darwin*/lib/*") do |file|
+    Dir.glob("*.OBJ/lib/*") do |file|
       if file.include? ".chk"
         cp file, libexec
       else


MikeMcQuaid:
> Addresses downstream issue Linuxbrew#1620

This is Homebrew, we don't make changes for Linux here.

carlorosati:
@MikeMcQuaid 

Is that so?

https://github.com/Homebrew/brew/pull/2274

MikeMcQuaid:
@carlorosati Yes, that's a separate repository for the package manager rather than the packages. Regardless, whatever point you're trying to make here you're being rude.

carlorosati:
@MikeMcQuaid Why do you allow Linux changes in the package manager but not the packages?

MikeMcQuaid:
@carlorosati It's much easier to make the package manager cross-platform without harming readability. Ideally we'll end up at a point where Linuxbrew can run an unpatched Homebrew/brew. This will never happen on homebrew-core without a cross-platform formula abstraction layer which will need to break backwards compatibility which we will never do. Linuxbrew should create their own homebrew-core (not a fork) and maintain formulae known to work on Linux configured how best makes sense on Linux.

carlorosati:
Ahh. Now that makes sense! I wish you had said so before! Your original comment came off as rude too.

carlorosati:
Just out of curiosity has this been discussed before? I'm curious if we could solve this without having two repositories. What about versioning newer formulas?

MikeMcQuaid:
> Ahh. Now that makes sense! I wish you had said so before! Your original comment came off as rude too.

I hope you appreciate that we don't always have time to provide full justifications. In the time I've typed out these replies I could have fixed a simple bug. I'm happy to do so on this occasion but that's not universal.

> Just out of curiosity has this been discussed before? I'm curious if we could solve this without having two repositories. What about versioning newer formulas?

It's been discussed in various PRs and issues over the years. It's technically possible to solve this without two repositories but we're not willing to do so on homebrew-core.

carlorosati:
> I'm happy to do so on this occasion but that's not universal.

Let's add it to the readme then? https://github.com/Homebrew/homebrew-core/pull/11381

PR #9943, opened 2017-02-14
Fix audit --strict warning about sourceforge link

miguelarauj1o:

-----

``` shell
FORMULAE:
  * http://FORMULAE.sourceforge.net/ should be `https://FORMULAE.sourceforge.io/`
```

### DIFF
--- a/Formula/ant-contrib.rb
+++ b/Formula/ant-contrib.rb
@@ -1,6 +1,6 @@
 class AntContrib < Formula
   desc "Collection of tasks for Apache Ant"
-  homepage "http://ant-contrib.sourceforge.net/"
+  homepage "https://ant-contrib.sourceforge.io/"
   url "https://downloads.sourceforge.net/project/ant-contrib/ant-contrib/1.0b3/ant-contrib-1.0b3-bin.tar.gz"
   version "1.0b3"
   sha256 "6e58c2ee65e1f4df031796d512427ea213a92ae40c5fc0b38d8ac82701f42a3c"
--- a/Formula/archivemail.rb
+++ b/Formula/archivemail.rb
@@ -1,6 +1,6 @@
 class Archivemail < Formula
   desc "Tool for archiving and compressing old email in mailboxes"
-  homepage "http://archivemail.sourceforge.net/"
+  homepage "https://archivemail.sourceforge.io/"
   url "https://downloads.sourceforge.net/project/archivemail/archivemail-0.9.0.tar.gz"
   sha256 "4b430e2fba6f24970a67bd61eef39d7eae8209c7bef001196b997be1916fc663"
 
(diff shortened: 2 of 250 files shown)


ilovezfs:
what does `--warning` mean here?

miguelarauj1o:
@ilovezfs, it's should be --strict. :/

ilovezfs:
@miguelarauj1o I actually think this are hard audit failures even without --strict, no?

miguelarauj1o:
I got theses warnings after run 'brew audit --strict'

ilovezfs:
```
Josephs-MacBook-Pro:~ joe$ brew audit vpcs
vpcs:
  * http://vpcs.sourceforge.net/ should be `https://vpcs.sourceforge.io/`
Error: 1 problem in 1 formula
```
so looks like they're failures even without `--strict`.

Also I think there are a few more: https://gist.github.com/ilovezfs/b4c5f091e38ebb6c4bd35428ceeca623

ilovezfs:
@miguelarauj1o don't worry about updating these just yet. We can probably do this in a single commit.

miguelarauj1o:
Yes, you're right, these are hard audit failures. I don't need pass --strict flag to get them.

Using ```sed```, I made an shell script to handle all cases that respect this structure: ```"http://FORMULAE.sourceforge.net/"```. 
So,
``` shell
chadwick:
  * http://chadwick.sourceforge.net/doc/index.html should be `https://chadwick.sourceforge.io/`
[...]
dumb:
  * http://dumb.sourceforge.net/index.php?page=about should be `https://dumb.sourceforge.io/`
```
```sed``` will ignore theses examples from your gist.

I can modify it to get the other cases.
Also, @ilovezfs, should I fix all commit messages (--warning to --strict)?

MikeMcQuaid:
We can update the commit messages in future but these are fine for now (or at least we don't want another CI run). Thanks @miguelarauj1o!

miguelarauj1o:
Great!

ilovezfs:
@MikeMcQuaid this explicitly had do not merge on it. Not sure why this was merged as is.

MikeMcQuaid:
@ilovezfs Apologies; because the email notifications don't include label changes and when I click through to a long thread it jumps to the comment I missed it. I think we should consider replacing "do not merge" with PR reviews as they are more prominent at the bottom of a PR next to the merge. Thoughts?

ilovezfs:
>Apologies

That's all right. Unfortunately the only actual remedy here is a force push, which is against the repository's religion.

>I think we should consider replacing "do not merge" with PR reviews

This won't work because you cannot perform a GitHub PR review on your own PRs, and the vast majority of my use of `do not merge` is on my own PRs so we need some other solution that covers that case as well.

MikeMcQuaid:
> This won't work because you cannot perform a GitHub PR review on your own PRs, and the vast majority of my use of do not merge is on my own PRs so we need some other solution that covers that case as well.

A failing CI job feels like it would be the best solution.

PR #9964, opened 2017-02-14
dfmt 0.5.0-beta.5 (devel)

John-Colvin:

-----

### DIFF
--- a/Formula/dfmt.rb
+++ b/Formula/dfmt.rb
@@ -4,6 +4,7 @@ class Dfmt < Formula
   url "https://github.com/Hackerpilot/dfmt.git",
       :tag => "v0.4.5",
       :revision => "4fe021df9771d83c325c879012842402a28ca5c7"
+  version_scheme 1
 
   head "https://github.com/Hackerpilot/dfmt.git", :shallow => false
 
@@ -16,9 +17,9 @@ class Dfmt < Formula
 
   devel do
     url "https://github.com/Hackerpilot/dfmt.git",
-        :tag => "v0.5.0-beta4",
-        :revision => "4a4704896ba960c3ffc22c25b38557d89ce13b79"
-    version "0.5.0-beta4"
+        :tag => "v0.5.0-beta.5",
+        :revision => "9fb13d0cafb3a9f0252e7e45277c37c28889731c"
+    version "0.5.0-beta.5"
   end
 
   depends_on "dmd" => :build
--- a/Formula/dfmt.rb
+++ b/Formula/dfmt.rb
@@ -4,7 +4,6 @@ class Dfmt < Formula
   url "https://github.com/Hackerpilot/dfmt.git",
       :tag => "v0.4.5",
       :revision => "4fe021df9771d83c325c879012842402a28ca5c7"
-  version_scheme 1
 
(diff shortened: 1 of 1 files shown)


Formula/dfmt.rb
```diff
@@ -4,6 +4,7 @@ class Dfmt < Formula
   url "https://github.com/Hackerpilot/dfmt.git",
       :tag => "v0.4.5",
       :revision => "4fe021df9771d83c325c879012842402a28ca5c7"
+  version_scheme 1
```
John-Colvin: Had to add this because apparently `beta4 > beta.5`, but I don't like it

ilovezfs: if we add this it can never be removed from the formula

Formula/dfmt.rb
```diff
@@ -16,9 +17,9 @@ class Dfmt < Formula
 
   devel do
     url "https://github.com/Hackerpilot/dfmt.git",
-        :tag => "v0.5.0-beta4",
-        :revision => "4a4704896ba960c3ffc22c25b38557d89ce13b79"
-    version "0.5.0-beta4"
+        :tag => "v0.5.0-beta.5",
+        :revision => "9fb13d0cafb3a9f0252e7e45277c37c28889731c"
+    version "0.5.0-beta.5"
```
ilovezfs: why not make this
```
version "0.5.0-beta5"
```

John-Colvin: Can do, but what's the general solution here? What is the preferred format for devel version strings? I'd prefer to just move to that across all the formulae I maintain, to minimise future work, even if it means adding/incrementing `version_scheme`.

ilovezfs: alpha < beta < rc is understood. Words like "preview" are not.

0.5.0-beta5 would be more typical than 0.5.0-beta.5, just as as 0.5.0-rc1 would be more typical than 0.5.0-rc.1

The way to get it to pick up the version without specifying it manually is to use a tarball or tag with the name leading.

https://cmake.org/files/v3.8/cmake-3.8.0-rc1.tar.gz -> parses as "3.8.0-rc1"
https://cmake.org/files/v3.8/3.8.0-rc1.tar.gz -> parses as "1"

This presents a challenge for devel versions that use designations like "alpha", "beta", and "rc" in the tag name but don't also prepend the software name, since you're forced to manually specify it every time.

So the solution is to use weird tag names like "dfmt-0.5.0-beta5" instead of "v0.5.0-beta5" or to provide a devel tarball with the name "dfmt-0.5.0-beta5.tar.gz"

You can run
```
bash-4.4$ brew irb
==> Interactive Homebrew Shell
Example commands available with: brew irb --examples
irb(main):001:0> Version.new("0.5.0-beta.5") > Version.new("0.5.0-beta4")
=> false
irb(main):002:0> Version.new("0.5.0-beta.5") < Version.new("0.5.0-beta4")
=> true
irb(main):003:0> 
```
to check how any two versions compare.

The general solution would be to teach our parser to understand tag names that have alpha, beta, and rc in them and that don't start with a word.

nijikon:

Formula/dfmt.rb
John-Colvin: http://semver.org/ suggests that `rc.1` is the preferred format.

PR #9968, opened 2017-02-14
deis 2.11.0

jayjun:

-----

### DIFF
--- a/Formula/deis.rb
+++ b/Formula/deis.rb
@@ -1,8 +1,10 @@
+require "language/go"
+
 class Deis < Formula
-  desc "Deploy and manage applications on your own servers"
+  desc "The CLI for Deis Workflow"
   homepage "https://deis.io/"
-  url "https://github.com/deis/workflow-cli/archive/v2.10.0.tar.gz"
-  sha256 "de5a9b1dd8746059c77b63472ae8d9a788e3f3cccdb190f8f307e3fdbb309700"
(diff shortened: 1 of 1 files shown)


ilovezfs:
This formula needs to be changed not to use `go get`

jayjun:
I don't believe there's a workaround.

[The original source](https://github.com/deis/workflow-cli/blob/master/version/version.go) uses the Go linker to pass the current version number into the build.

Without it, the `deis version` command won't work.

jayjun:
@ilovezfs Found official Darwin binaries. Removed Go dependency in 29049be.

MikeMcQuaid:
@jayjun We need to build from source, sorry.

jayjun:
@MikeMcQuaid The [official build](https://github.com/deis/workflow-cli/blob/master/Makefile#L42) uses `gox`, but I can't use `go get`.

Can I clarify that we need to find a way to use `go build` alone?

MikeMcQuaid:
@jayjun We prefer upstream vendored resources or submodules but if not you can use `go_resource` (`grep` other formulae to see usage).

jayjun:
Thanks for the `go_resource` hint @MikeMcQuaid.

305b7d7 removes `go get` and builds from source.

dunn:
Looks great, thanks! Merged as f3648f06d9d188117599e537a504b8fd8445cafd.

PR #9978, opened 2017-02-14
postgres: fix caveats for fresh install

jmgarnier:
Edit caveats to add a script to create postgres role for a fresh install.

**=> I did not do last 2 steps as I only edited the "caveats" text.**

-----

### DIFF
--- a/Formula/postgresql.rb
+++ b/Formula/postgresql.rb
@@ -106,6 +106,14 @@ class Postgresql < Formula
     you may need to remove the previous version first. See:
       https://github.com/Homebrew/homebrew/issues/2510
 
+    For a fresh install, you need to setup the postgres role:
+    ```
+    psql template1
+    CREATE ROLE postgres LOGIN
+    SUPERUSER INHERIT CREATEDB CREATEROLE REPLICATION;
+    \q
+    ```
+    
     To migrate existing data from a previous major version (pre-9.0) of PostgreSQL, see:
       https://www.postgresql.org/docs/9.6/static/upgrading.html
 
--- a/Formula/postgresql.rb
+++ b/Formula/postgresql.rb
@@ -113,7 +113,7 @@ class Postgresql < Formula
     SUPERUSER INHERIT CREATEDB CREATEROLE REPLICATION;
     \q
     ```
-    
+
     To migrate existing data from a previous major version (pre-9.0) of PostgreSQL, see:
       https://www.postgresql.org/docs/9.6/static/upgrading.html
 


jmgarnier:
I guess  both commits should be squashed if this will be merged some day

MikeMcQuaid (changes requested):
You only need to do this if you've previously run `createuser -s -r postgres`, right? I've never personally had to do this locally so it doesn't seem to be universally applicable so I'm currently 👎 on including this.

jmgarnier:
Thanks @MikeMcQuaid  I am not an expert Postgres user but it seems I am not the first to be hit by this missing step - no Superuser is created by homebrew - when `brew install postgresql` the 1st time. See http://stackoverflow.com/a/35308200/575950

Until I created this role, I could not run `psql` and run migrations on a fresh Phoenix project:
image

Is there a point doing more research to come up with a clearer message?

MikeMcQuaid:
Did you run `createuser -s -r postgres`?

jmgarnier:
I ran the equivalent, that I copy pasted from the blog post
```sql
CREATE ROLE postgres LOGIN
SUPERUSER INHERIT CREATEDB CREATEROLE REPLICATION;
```

Looking at https://www.postgresql.org/docs/9.6/static/sql-createrole.html user = role. 
Your version is shorter.  But I prefer explicit long parameters. What do you think if I amend my PR with the following SQL:
```sql
createuser --superuser --createrole postgres
```

MikeMcQuaid:
Passing on this for now but we'll consider it in future if this seems to be a more widespread problem rather than specific to that blog post. Thanks anyway!

jmgarnier:
It was not specific to that blog post. I was setting a fresh phoenix app and I had to setup `postgres`. 

Anyway, If I ever need to make a fresh install of posgres with homebrew, I'll remember now and know where to look :)

Thanks for taking the time to review

jmgarnier:
Anyway, I can see why you got confused, I was not very clear.

I will open another PR with a clearer goal - help first time install - and hopefully it will help someone some day 😃

MikeMcQuaid:
@jmgarnier I think we can pass on a PR for now, thanks. I and many other users seem to have no need for this on a first time install. It may be specific to Phoenix but that's not relevant to Homebrew.

jmgarnier:
Fair enough, no PR then 😸  . 

For my own sake, I will ask some friends with more experience, I really need to understand how things work. I will come back with another PR if it's relevant to a broader scope than just "Phoenix".

Thanks again. I will make sure my next contributions are more straight forward.

PR #10031, opened 2017-02-16
kontena-cli 1.1.1 (new formula)

kke:

-----

**Note:** executable name is "kontena", formula is "kontena-cli" as this formula only installs the command line client which is what most people want on their local machine. Formulas for the other components in the stack can be created in the future.

**Note2:** trickery with picking up the ruby and gem commands and installing dependencies are related to https://github.com/Homebrew/brew/issues/2010 and https://github.com/Homebrew/brew/issues/2022

### DIFF
--- a/Formula/kontena-cli.rb
+++ b/Formula/kontena-cli.rb
@@ -35,23 +35,31 @@ class KontenaCli < Formula
     # Add a build tag to "kontena --version" string
     inreplace buildpath/"cli/lib/kontena/main_command.rb", "RUBY_PLATFORM +", "RUBY_PLATFORM + '+homebrew' +"
 
-    # Install completions
-    inreplace buildpath/"cli/lib/kontena/scripts/init", "#!/usr/bin/env bash", "# Completion for kontena-cli"
-    inreplace buildpath/"cli/lib/kontena/scripts/init", "DIR=$( cd \"$( dirname \"$src\" )\" && pwd )", "DIR=#{libexec}/kontena/scripts"
+    # Patch completion init script
+    inreplace buildpath/"cli/lib/kontena/scripts/init" do |s|
+      s.gsub! "#!/usr/bin/env bash", "# Completion for kontena-cli"
+      s.gsub! "DIR=$( cd \"$( dirname \"$src\" )\" && pwd )", "DIR=#{libexec}/kontena/scripts"
+    end
+
+    # The same script works for both shells. The original 'init' needs to exist in the directory or
+    # kontena whois --bash-completion-path crashes.
     cp buildpath/"cli/lib/kontena/scripts/init", buildpath/"cli/lib/kontena/scripts/kontena.zsh"
     cp buildpath/"cli/lib/kontena/scripts/init", buildpath/"cli/lib/kontena/scripts/kontena.bash"
+
+    # Create a "standard" loader for the completion function
     (buildpath/"cli/lib/kontena/scripts/_kontena").write(zsh_completer)
-    bash_completion.install buildpath/"cli/lib/kontena/scripts/kontena.bash" => "kontena.bash"
-    zsh_completion.install buildpath/"cli/lib/kontena/scripts/kontena.zsh" => "kontena.zsh"
     zsh_completion.install buildpath/"cli/lib/kontena/scripts/_kontena" => "_kontena"
 
-    bin.install Dir["cli/bin/*"]
+    bash_completion.install buildpath/"cli/lib/kontena/scripts/kontena.bash" => "kontena.bash"
+    zsh_completion.install buildpath/"cli/lib/kontena/scripts/kontena.zsh" => "kontena.zsh"
+
+    bin.install buildpath/"cli/bin/kontena"
     libexec.install Dir["cli/lib/*"]
     prefix.install buildpath/"cli/README.md"
     prefix.install buildpath/"CHANGELOG.md"
+    # Required during operation:
     prefix.install buildpath/"cli/LOGO"
     prefix.install buildpath/"cli/VERSION"
-    doc.install Dir["docs/*"]
   end
 
   test do
--- a/Formula/kontena-cli.rb
+++ b/Formula/kontena-cli.rb
@@ -10,8 +10,9 @@ class KontenaCli < Formula
   depends_on :ruby => "2.1"
 
   def install
-    ruby_command = which("ruby")
-    gem_command = which("gem")
+    # avoid the outdated system ruby for now
+    ruby_command = which_all("ruby").detect { |path| path.to_s != "/usr/bin/ruby" }
+    gem_command = File.join(ruby_command.dirname, 'gem')
 
     # Make --version indicate it is a HEAD build
     if build.head?
@@ -89,7 +90,7 @@ class KontenaCli < Formula
 
   def libexec_to_gem_path
     <<-EOS.undent
-      "# add libexec to gem path
+      # add libexec to gem path
       ENV['GEM_PATH'] = '#{libexec}'
       Gem.paths = ENV
 
--- a/Formula/kontena-cli.rb
+++ b/Formula/kontena-cli.rb
@@ -81,6 +81,23 @@ class KontenaCli < Formula
     zsh_completion.install buildpath/"cli/lib/kontena/scripts/kontena.zsh" => "kontena.zsh"
   end
 
+  def caveats
+    unless Dir['/usr/local/{opt,var}/rbenv/shims/kontena'].empty?
+      <<-EOS.undent
+        You seem to use rbenv and have a previously installed kontena-cli that
+        may get loaded instead of the homebrew installed version.
+
+        To uninstall the previous installation copy and paste this into the
+        terminal:
+
+          for ruby in $(rbenv whence kontena); do \\
+            rbenv shell $ruby; gem uninstall --force -a -x kontena-cli; \\
+          done; \\
+          rbenv rehash
+      EOS
+    end
+  end
+
   test do
     assert_match "+homebrew", shell_output("#{bin}/kontena --version")
     assert_match "login", shell_output("#{bin}/_kontena_completer kontena master")
--- a/Formula/kontena-cli.rb
+++ b/Formula/kontena-cli.rb
@@ -1,0 +1,120 @@
+class KontenaCli < Formula
+  desc "Command-line client for Kontena container & microservices platform"
+  homepage "https://kontena.io/"
+  url "https://github.com/kontena/kontena/archive/v1.1.1.tar.gz"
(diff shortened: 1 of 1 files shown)


kke:
Hold, I just realized something.

MikeMcQuaid:
> Note2: trickery with picking up the ruby and gem commands and installing dependencies are related to Homebrew/brew#2010 and Homebrew/brew#2022

@kke Can you submit PRs to Homebrew/brew for this? I'm hesitant to include this logic in every relevant Ruby formula.

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+  option "with-ruby-autodiscovery", "Select a Ruby interpreter on every run instead of linking to the Ruby used during installation"
+  option "with-development", "Install development libraries"
+
+  depends_on :ruby => "2.1"
+
+  def install
+    # The depends_on :ruby is a bit flawed. It will validate that a ruby with version >= X exists
+    # but does not patch PATH to return that one
+    ruby_command = pick_a_ruby("2.1")
+    gem_command = gem_for_ruby(ruby_command)
```
MikeMcQuaid: This logic should go to Homebrew/brew

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+
+    # Make --version indicate it is a HEAD build
+    inreplace (buildpath/"cli/VERSION"), /(\d+\.\d+\.\d+).*/, "\\1-head" if build.head?
+
+    # Also add a "homebrew" build tag to "kontena --version" string
+    inreplace buildpath/"cli/lib/kontena/main_command.rb", "RUBY_PLATFORM +", "RUBY_PLATFORM + '+homebrew' +"
```
MikeMcQuaid: Let's make lines <= 80 characters. Also, ideally this wouldn't require an `inreplace` but upstream would provide some sort of script to configure this.

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+
+    # If building from tar.gz sources, git ls-files obviously doesn't work
+    unless build.head?
+      inreplace buildpath/"cli/kontena-cli.gemspec", "`git ls-files -z`.split(\"\\x0\")", "Dir['LOGO', 'README.md', 'LICENSE.txt', 'VERSION','bin/*', 'lib/**/*']"
```
MikeMcQuaid: Let's make lines <= 80 characters. You can use a Git checkout (url with `:revision`, `:tag`) instead of patching if upstream assumes this is being built from Git.

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+    end
+
+    # Build the gem from sources and install it and its dependencies to buildpath/out
+    Dir.chdir buildpath/"cli" do
```
MikeMcQuaid: `(buildpath/"cli").cd`

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+      system gem_command, "build", "--norc", "kontena-cli.gemspec"
+      install_args = ["install", Dir["kontena-cli-*.gem"].first, "--install-dir", buildpath/"out", "--no-env-shebang", "--no-wrappers", "--norc", "--no-document"]
+      install_args << "--development" if build.with?("development")
```
MikeMcQuaid: Any reason to not do this by default?

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+
+    # If building from tar.gz sources, git ls-files obviously doesn't work
+    unless build.head?
+      inreplace buildpath/"cli/kontena-cli.gemspec", "`git ls-files -z`.split(\"\\x0\")", "Dir['LOGO', 'README.md', 'LICENSE.txt', 'VERSION','bin/*', 'lib/**/*']"
+    end
+
+    # Build the gem from sources and install it and its dependencies to buildpath/out
+    Dir.chdir buildpath/"cli" do
+      system gem_command, "build", "--norc", "kontena-cli.gemspec"
+      install_args = ["install", Dir["kontena-cli-*.gem"].first, "--install-dir", buildpath/"out", "--no-env-shebang", "--no-wrappers", "--norc", "--no-document"]
```
MikeMcQuaid: Let's make this <= 80 characters

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+      install_args << "--development" if build.with?("development")
+      system gem_command, *install_args
+    end
+
+    # Patch the exec script
+    inreplace buildpath/"cli/bin/kontena" do |s|
+      s.gsub! "#!/usr/bin/env ruby", build.with?("ruby-autodiscovery") ? adaptive_shebang : "#!#{ruby_command}"
```
MikeMcQuaid: 👎 to this option if it requires patching. Instead, request upstream provide this as an option.

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+      s.gsub! "# add self to libpath", "# add libexec to gem path\nENV['GEM_PATH'] = '#{libexec}'\nGem.paths = ENV\n\n# add self to libpath"
```
MikeMcQuaid: This should use `EOS.undent` for the second argument. Also, can you clarify why this is needed? It just seems to be a comment?

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+    end
+    bin.install buildpath/"cli/bin/kontena"
+
+    # Same for autocompleter
+    inreplace buildpath/"cli/lib/kontena/scripts/completer" do |s|
+      s.gsub! "#!/usr/bin/env ruby", build.with?("ruby-autodiscovery") ? adaptive_shebang : "#!#{ruby_command}"
+      s.gsub! "# add self to libpath", "# add libexec to gem path\nENV['GEM_PATH'] = '#{libexec}'\nGem.paths = ENV\n\n# add self to libpath"
+      s.gsub! "require_relative '..", "require 'kontena"
```
MikeMcQuaid: Same comments here as above.

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+    end
+    bin.install buildpath/"cli/lib/kontena/scripts/completer" => "_kontena_completer"
+
+    rm_rf buildpath/"out/cache"
+    rm_rf buildpath/"out/bin"
```
MikeMcQuaid: These aren't installed anyway so do they need to be deleted?

Formula/kontena-cli.rb
```diff
@@ -0,0 +1,154 @@
...
+    end
+    bin.install buildpath/"cli/bin/kontena"
+
+    # Same for autocompleter
+    inreplace buildpath/"cli/lib/kontena/scripts/completer" do |s|
+      s.gsub! "#!/usr/bin/env ruby", build.with?("ruby-autodiscovery") ? adaptive_shebang : "#!#{ruby_command}"
+      s.gsub! "# add self to libpath", "# add libexec to gem path\nENV['GEM_PATH'] = '#{libexec}'\nGem.paths = ENV\n\n# add self to libpath"
+      s.gsub! "require_relative '..", "require 'kontena"
+    end
+    bin.install buildpath/"cli/lib/kontena/scripts/completer" => "_kontena_completer"
```
MikeMcQuaid: If this is completion it should probably go into `bash_completion.install`

MikeMcQuaid: Regardless, it doesn't look like it should go into `bin`.

MikeMcQuaid (changes requested):

Formula/kontena-cli.rb
kke: They are installed on the next line? Perhaps I just don't understand how it works.

Formula/kontena-cli.rb
kke: The comment is just used to find a suitable spot in the file to insert the code.

It's needed because the app is installed as a gem and the exec script needs to modify GEM_PATH to find it.

Formula/kontena-cli.rb
kke: (I am from the upstream)

Removed the trickery.

Formula/kontena-cli.rb
kke: There's very little reason to do it at all.

MikeMcQuaid:
Test is failing: https://bot.brew.sh/job/Homebrew%20Core/16801/version=sierra/testReport/junit/brew-test-bot/sierra/test_kontena_cli/

MikeMcQuaid:
As you're upstream a more general comment: there's a lot of logic in here that I think should belong upstream. Ideally there's be a single provided `Makefile` or `Rakefile` or `script/build` or similar upstream that does all the necessary `inreplace` operations and installation (like `make install`). My concern is that this formula as-is will be very difficult for us to keep up to date without intimate knowledge of upstream. Thoughts?

kke:
The upstream automatically builds and publishes the app to rubygems. If OSX just shipped with an up-to-date Ruby, then `gem install kontena-cli` would be enough for the users.

With brew we can have a newer Ruby as dependency and another bonus is an easy way to install `--HEAD`.

Some of the logic in this formula is just "nice to have", such as inserting stuff to the `--version` string for easy identification of the installation method and to verify that you are in fact running the brewed version.

Some of the patching is to avoid problems with the `#!/usr/bin/env ruby` shebang possibly giving a wrong ruby if you're using rbenv, rvm, chruby or who knows what.

The shell autocompletion calls a ruby script, that's why the `_kontena_completer` is in `bin/`. It could be left inside the libexec as in the normal `gem install` installation, but figuring out the path seems difficult.

You can expect the formula to be kept up to date by the upstream when new versions are released.

MikeMcQuaid:
> You can expect the formula to be kept up to date by the upstream when new versions are released 

Sorry, this isn't how Homebrew works. We need things to be such that we're happy maintaining them without upstream. That said, given this doesn't need a bottle it may be preferable for you to consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead which would allow upstream complete control (and less pedantry from us).

kke:
In casks you have this appcast thing that could be used to check for new versions automatically, how do you keep up to date with updates without upstream or someone to make PR's?

MikeMcQuaid:
@kke We use GitHub's releases API and https://github.com/youtux/homebrew-livecheck

kke:
Formula simplified due to improved support in upstream.
Waiting for https://github.com/kontena/kontena/pull/1889 - closing until relevant.

PR #10036, opened 2017-02-16
cn 1.0.0 (new formula)

JayBrown:
cn is short for "Common Numerics"
It is originally part of Apple's Open Source Project, taken from the CommonCrypto library, published open source under APSL v2.0.
However, it is not installed in macOS as a binary in /usr/bin.
It adds some important functionality to macOS:
not only Base16 and Base32 encoding/decoding plus additional options for Base64, but also many CRC hash functions, which can work alongside the native crc32 and cksum/sum commands:
several CRC-8 hashes
several CRC-16 hashes
several CRC-32 hashes, incl. CRC-32C (Castagnoli), which is used by Google Storage (compare the gsutil CLI)
plus one CRC-64 hash (ECMA 182)
Please note that auditing produced some errors regarding the system make commands; however, they've been entered like this by design, and installation worked fine.
cn will probably always remain at v1.0.0, because we should not expect an update from Apple

-----

### DIFF
--- a/Formula/cn.rb
+++ b/Formula/cn.rb
@@ -1,0 +1,19 @@
+class Cn < Formula
+  desc "CLI for CommonNumerics routines (CRC-8/16/32/64, Base16/32/64)"
+  homepage "https://github.com/JayBrown/cn"
+  url "https://github.com/JayBrown/cn/archive/1.00.tar.gz"
+  version "1.0.0"
+  sha256 "4a11ccf8ba82f344f4d8faffc9a3cc28c5971d00f8e0f85d2682bb70e351e660"
+
+  def install
+    system "make prefix=#{prefix}"
+    system "make clean"
+    system "make cn"
+    bin.install "cn"
+    man1.install "cn.1"
+  end
+
+  test do
+    system "#{bin}/cn", "version"
+  end
+end


Formula/cn.rb
```diff
@@ -0,0 +1,19 @@
+class Cn < Formula
+  desc "CLI for CommonNumerics routines (CRC-8/16/32/64, Base16/32/64)"
+  homepage "https://github.com/JayBrown/cn"
+  url "https://github.com/JayBrown/cn/archive/1.00.tar.gz"
+  version "1.0.0"
+  sha256 "4a11ccf8ba82f344f4d8faffc9a3cc28c5971d00f8e0f85d2682bb70e351e660"
```
bfontaine: This seems wrong:

```
curl -sL 'https://github.com/JayBrown/cn/archive/1.00.tar.gz' | shasum -a 256
ac0472649f09d20f396767e1f9fa185a59639bd17fc32153f5dd4a2135dbebc9  -
```

Formula/cn.rb
```diff
@@ -0,0 +1,19 @@
+class Cn < Formula
+  desc "CLI for CommonNumerics routines (CRC-8/16/32/64, Base16/32/64)"
+  homepage "https://github.com/JayBrown/cn"
+  url "https://github.com/JayBrown/cn/archive/1.00.tar.gz"
+  version "1.0.0"
```
bfontaine: Why not use `1.0` from the tarball URL?

Formula/cn.rb
```diff
@@ -0,0 +1,19 @@
...
+  sha256 "4a11ccf8ba82f344f4d8faffc9a3cc28c5971d00f8e0f85d2682bb70e351e660"
+
+  def install
+    system "make prefix=#{prefix}"
```
bfontaine: Please use `system "make", "prefix=#{prefix}"`.

Formula/cn.rb
```diff
@@ -0,0 +1,19 @@
...
+    system "make clean"
+    system "make cn"
+    bin.install "cn"
+    man1.install "cn.1"
```
bfontaine: Is there no `make install` we can use?

Formula/cn.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  test do
+    system "#{bin}/cn", "version"
```
bfontaine: Please write a test that does something more substantial than just printing the version.

bfontaine (changes requested):

bfontaine:
This is a self-submission of a relatively new project with no popularity (0 stars on GitHub), so we’ll pass on this for now. Sorry!

You can maintain the formula [in your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md) and submit it here again when it’s more popular. Thanks!

JayBrown:
^ (1) Already created my own 3rd-party tap; installing fine with the .rb I originally submitted, and it's working great. That's sufficient for me, although an actual brew formula would be nice, since this one is a "secret CLI" from Apple (open source). (2) The hash seems to have been a copy/paste mistake from an earlier version, which should've been pasted the other way around; at any rate, with the right hash it works fine. (3) As I wrote in the commit, `system "make", "prefix=#{prefix}"` doesn't work, only `system "make prefix=#{prefix}"` does! (I copied this from another formula's .rb, and I don't see why it should be prohibited here, when others are using it.)

bfontaine:
> As I wrote in the commit, `system "make", "prefix=#{prefix}"` doesn't work, only `system "make prefix=#{prefix}"` does!

This seems strange. Did you had a specific error? I just checked the [`Makefile`](https://github.com/JayBrown/cn/blob/master/Makefile) and it doesn’t contain any `prefix` variable so you can just run `system "make"`.

> I copied this from another formula's .rb, and I don't see why it should be prohibited here, when others are using it.

This formula was probably written before we started enforcing this rule.

JayBrown:
It's possible that it was an older formula; the corresponding webpage was gone.

I didn't get a specific error, as I recall, just a hint to visit the troubleshooting page. ;) 

As for the Makefile, it's just the setup for a super simple way to install a `*.c` document, with `make clean`, and then `make <formula>`. Afterwards you only need to put or symlink the binary into a /bin, therefore the `bin.install "cn"` in the `.rb`.

PR #9304, opened 2017-01-26
go 1.7.5

tsuna:
Created with `brew bump-formula-pr`.

PR #7452, opened 2016-12-01
go 1.7.4

cblecker:
Created with `brew bump-formula-pr`.

PR #10048, opened 2017-02-16
go 1.8

ttaylorr:

-----

### DIFF
--- a/Formula/go.rb
+++ b/Formula/go.rb
@@ -3,16 +3,15 @@ class Go < Formula
   homepage "https://golang.org"
 
   stable do
-    url "https://storage.googleapis.com/golang/go1.7.5.src.tar.gz"
-    mirror "https://fossies.org/linux/misc/go1.7.5.src.tar.gz"
-    version "1.7.5"
-    sha256 "4e834513a2079f8cbbd357502cccaac9507fd00a1efe672375798858ff291815"
+    url "https://storage.googleapis.com/golang/go1.8.src.tar.gz"
+    mirror "https://fossies.org/linux/misc/go1.8.src.tar.gz"
+    version "1.8"
+    sha256 "406865f587b44be7092f206d73fc1de252600b79b3cacc587b74b5ef5c623596"
 
-    go_version = version.to_s.split(".")[0..1].join(".")
+    # revision should be tip of tree commit in release-branch.goX.Y
     resource "gotools" do
       url "https://go.googlesource.com/tools.git",
-          :branch => "release-branch.go#{go_version}",
-          :revision => "6220cba6419b2bf78aad19d85c347ecc0fda2b53"
+          :revision => "5682db0e919ed9cfc6f52ac32e170511a106eb3b"
     end
   end
 
@@ -20,16 +19,6 @@ class Go < Formula
     sha256 "7d96560afe4d231cb7c63911fc1e8324a5092306503e8cbe8b7b89d5dcbffe9d" => :sierra
     sha256 "8d4d7111294a186a032c7fd9534ae343f01e84e5b45b56118d7e20a896fb0926" => :el_capitan
     sha256 "311a87565de855a024df90eed45d95c84e3c71dfcc44941a5755ce6b412755cb" => :yosemite
-  end
-
-  devel do
-    url "https://storage.googleapis.com/golang/go1.8rc3.src.tar.gz"
-    version "1.8rc3"
-    sha256 "38b1c1738f111f7bccdd372efca2aa98a7bad1ca2cb21767ba69f34ae007499c"
-
-    resource "gotools" do
-      url "https://go.googlesource.com/tools.git"
-    end
   end
 
   head do
@@ -46,8 +35,7 @@ class Go < Formula
 
   depends_on :macos => :mountain_lion
 
-  # Should use the last stable binary release to bootstrap.
-  # More explicitly, leave this at 1.7 when 1.7.1 is released.
+  # Don't update this unless this version cannot bootstrap the new version.
   resource "gobootstrap" do
     url "https://storage.googleapis.com/golang/go1.7.darwin-amd64.tar.gz"
     version "1.7"
@@ -55,8 +43,6 @@ class Go < Formula
   end
 
   def install
-    ENV.permit_weak_imports
-
     (buildpath/"gobootstrap").install resource("gobootstrap")
     ENV["GOROOT_BOOTSTRAP"] = buildpath/"gobootstrap"
 
@@ -94,7 +80,8 @@ class Go < Formula
   end
 
   def caveats; <<-EOS.undent
-    As of go 1.2, a valid GOPATH is required to use the `go get` command:
+    As of go 1.2, a valid GOPATH is required to use the `go get` command.
+    If $GOPATH is not specified, $HOME/go will be used by default:
       https://golang.org/doc/code.html#GOPATH
 
     You may wish to add the GOROOT-based install location to your PATH:
--- a/Formula/godep.rb
+++ b/Formula/godep.rb
@@ -3,6 +3,7 @@ class Godep < Formula
   homepage "https://godoc.org/github.com/tools/godep"
   url "https://github.com/tools/godep/archive/v79.tar.gz"
   sha256 "3dd2e6c4863077762498af98fa0c8dc5fedffbca6a5c0c4bb42b452c8268383d"
+  revision 1
   head "https://github.com/tools/godep.git"
 
   bottle do
@@ -25,7 +26,7 @@ class Godep < Formula
     (testpath/"Godeps/Godeps.json").write <<-EOS.undent
       {
         "ImportPath": "github.com/tools/godep",
-        "GoVersion": "go1.7",
+        "GoVersion": "go1.8",
         "Deps": [
           {
             "ImportPath": "golang.org/x/tools/cover",
--- a/Formula/govendor.rb
+++ b/Formula/govendor.rb
@@ -3,6 +3,7 @@ class Govendor < Formula
   homepage "https://github.com/kardianos/govendor"
   url "https://github.com/kardianos/govendor/archive/v1.0.8.tar.gz"
   sha256 "7e887b84c7a9278473f39ae8a74440ffc17b329aa193e9304d170d458f8785c7"
+  revision 1
   head "https://github.com/kardianos/govendor.git"
 
   bottle do


wendigo:
That was fast :)
```
brew bump-formula-pr --strict go
Error: These open pull requests may be duplicates:
go 1.8 https://github.com/Homebrew/homebrew-core/pull/10048
Duplicate PRs should not be opened. Use --force to override this error.
```

Formula/go.rb
```diff
@@ -49,9 +39,9 @@ class Go < Formula
   # Should use the last stable binary release to bootstrap.
   # More explicitly, leave this at 1.7 when 1.7.1 is released.
```
bcomnes: probably need to update the comment here

ttaylorr: Thanks, fixed 👍

tsuna:
The `godep` check failed with a weird error:
```
==> /usr/local/Cellar/godep/79/bin/godep restore
godep: [WARNING]: godep should only be used inside a valid go package directory and
godep: [WARNING]: may not function correctly. You are probably outside of your $GOPATH.
godep: [WARNING]:	Current Directory: /private/tmp/godep-test-20170216-53415-iarcjp
godep: [WARNING]:	$GOPATH: /private/tmp/godep-test-20170216-53415-iarcjp
godep: Dep (golang.org/x/tools/cover) restored, but was unable to load it with error:
	Package (bufio) not found
godep: Error checking some deps.
Error: godep: failed
Failed executing: /usr/local/Cellar/godep/79/bin/godep restore
```

ttaylorr:
> The `godep` check failed with a weird error:

Strange. Anyone else getting this?

twexler:
@ttaylorr I was able to replicate this failure. I am not entirely sure why it's happening yet, it looks like there is some issue with `runtime.GOROOT()`/`$GOROOT`. `godep` doesn't seem to be searching those paths for imports on 1.8.

DomT4:
`godep` at the very least will need a `revision` bump, as it always does on major bumps.

cblecker:
The caveat should also be edit, as there is a functionality change in GOPATH:
```
    As of go 1.8, a default GOPATH will be created for you if it is not manually
    set:
      https://golang.org/doc/code.html#hdr-GOPATH_environment_variable
```

Formula/go.rb
```diff
@@ -47,11 +37,11 @@ class Go < Formula
   depends_on :macos => :mountain_lion
 
   # Should use the last stable binary release to bootstrap.
-  # More explicitly, leave this at 1.7 when 1.7.1 is released.
+  # More explicitly, leave this at 1.8 when 1.8.1 is released.
```
DomT4: Unless 1.7 has known issues bootstrapping 1.8 I'd probably leave this how it was. The original intention here was that no version of `go` should be bootstrapped by the same version of `go`, so that if there's a problem bootstrapping you know it's with the new from-source release rather than having to work out whether the problem is the binary or the from-source part.

DomT4: (I appreciate when I tweaked this blob of code I made the comment confusing as all hell, apologies)

ttaylorr: I think this argument makes more sense to me than https://github.com/Homebrew/homebrew-core/pull/10048#discussion_r101629955. Fixed!

bcomnes: Makes sense when you put it this way ;)

prsolucoes:
+1
Waiting for it!!!!

Formula/go.rb
DomT4: Apologies 😸. I've noticed with some pain that since I left the project code comments I added as a maintainer made sense sometimes exclusively to me, heh.

jingweno:
:+1:

twexler:
> godep at the very least will need a revision bump, as it always does on major bumps.

This is the issue, looks like `godep` just needs to be built from source again with 1.8.

twexler:
Alternatively the formula could be modified to always set `ENV['GOROOT']` from the output of `go env GOROOT`.

ttaylorr:
> > godep at the very least will need a revision bump, as it always does on major bumps.
> 
> This is the issue, looks like `godep` just needs to be built from source again with 1.8.

I was unable to confirm that this is a correct fix. After bumping the version in `godep`'s formula:

```diff
diff --git a/Formula/godep.rb b/Formula/godep.rb
index 6ee335ae..a9c003cc 100644
--- a/Formula/godep.rb
+++ b/Formula/godep.rb
@@ -25,7 +25,7 @@ class Godep < Formula
     (testpath/"Godeps/Godeps.json").write <<-EOS.undent
       {
         "ImportPath": "github.com/tools/godep",
-        "GoVersion": "go1.7",
+        "GoVersion": "go1.8",
         "Deps": [
           {
             "ImportPath": "golang.org/x/tools/cover",
``` 

and rebuilding it from source:

```sh
/u/l/H/L/T/h/homebrew-core (master!) $ go version
go version go1.8 darwin/amd64
/u/l/H/L/T/h/homebrew-core (master!) $ brew install --build-from-source godep
Updating Homebrew...
==> Using the sandbox
==> Downloading https://github.com/tools/godep/archive/v79.tar.gz
Already downloaded: /Users/ttaylorr/Library/Caches/Homebrew/godep-79.tar.gz
==> go build -o /usr/local/Cellar/godep/79/bin/godep
🍺  /usr/local/Cellar/godep/79: 3 files, 8.9M, built in 4 seconds
/u/l/H/L/T/h/homebrew-core (master!) $ godep version
godep v79 (darwin/amd64/go1.8)
```

the tests still failed:

```sh
/u/l/H/L/T/h/homebrew-core (master!) $ brew test --verbose godep
Testing godep
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20170216-14330-1dawnu0.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/godep.rb --verbose
==> /usr/local/Cellar/godep/79/bin/godep restore
godep: [WARNING]: godep should only be used inside a valid go package directory and
godep: [WARNING]: may not function correctly. You are probably outside of your $GOPATH.
godep: [WARNING]:       Current Directory: /private/tmp/godep-test-20170216-14331-1fam7n9
godep: [WARNING]:       $GOPATH: /private/tmp/godep-test-20170216-14331-1fam7n9
```

@twexler -- were you able to spot anything I missed?

twexler:
> @twexler -- were you able to spot anything I missed?
Do you have go 1.7.5 available?

polds:
I just looked and fossies.org doesn't have the 1.8 package yet either.

twexler:
> @twexler -- were you able to spot anything I missed?
This is a bit backwards(the first test is *not* the bottled version) because I screwed up, but I believe this demonstrates that `--build-from-source` fixes it. 

```
[twexler:~/dev/tmp] 2 $ godep version
godep v79 (darwin/amd64/go1.8)
[twexler:~/dev/tmp] $ brew test godep
Testing godep
==> Using the sandbox
==> /usr/local/Cellar/godep/79/bin/godep restore
⌁45% [twexler:~/dev/tmp] $ brew reinstall godep
==> Reinstalling godep
==> Downloading https://homebrew.bintray.com/bottles/godep-79.sierra.bottle.tar.gz
Already downloaded: /Users/twexler/Library/Caches/Homebrew/godep-79.sierra.bottle.tar.gz
==> Pouring godep-79.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/godep/79: 3 files, 8.6M
⌁45% [twexler:~/dev/tmp] $ brew test godep
Testing godep
==> Using the sandbox
==> /usr/local/Cellar/godep/79/bin/godep restore
Last 15 lines from /Users/twexler/Library/Logs/Homebrew/godep/test.01.godep:
2017-02-16 15:04:58 -0800

/usr/local/Cellar/godep/79/bin/godep
restore

godep: [WARNING]: godep should only be used inside a valid go package directory and
godep: [WARNING]: may not function correctly. You are probably outside of your $GOPATH.
godep: [WARNING]:	Current Directory: /private/tmp/godep-test-20170216-59376-z4jxm2
godep: [WARNING]:	$GOPATH: /private/tmp/godep-test-20170216-59376-z4jxm2
godep: Dep (golang.org/x/tools/cover) restored, but was unable to load it with error:
	Package (bufio) not found
godep: Error checking some deps.
Error: godep: failed
Failed executing: /usr/local/Cellar/godep/79/bin/godep restore
/usr/local/Homebrew/Library/Homebrew/formula.rb:1834:in `block in system'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1772:in `open'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1772:in `system'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/godep.rb:37:in `block in <class:Godep>'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1671:in `block (2 levels) in run_test'
/usr/local/Homebrew/Library/Homebrew/formula.rb:884:in `with_logging'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1670:in `block in run_test'
/usr/local/Homebrew/Library/Homebrew/extend/fileutils.rb:14:in `block in mktemp'
/usr/local/Homebrew/Library/Homebrew/extend/fileutils.rb:74:in `block in run'
/usr/local/Homebrew/Library/Homebrew/extend/fileutils.rb:74:in `chdir'
/usr/local/Homebrew/Library/Homebrew/extend/fileutils.rb:74:in `run'
/usr/local/Homebrew/Library/Homebrew/extend/fileutils.rb:13:in `mktemp'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1664:in `run_test'
/usr/local/Homebrew/Library/Homebrew/test.rb:28:in `block in <main>'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/timeout.rb:66:in `timeout'
/usr/local/Homebrew/Library/Homebrew/test.rb:27:in `<main>'
[twexler:~/dev/tmp] 1 $ godep version
godep v79 (darwin/amd64/go1.7.5)
[twexler:~/dev/tmp] $
```

DomT4:
> @twexler -- were you able to spot anything I missed?

Heh, the test actually passed there on `godep` 😉. The warning about `GOPATH` is expected. A failure more explicitly blows up:
```
Error: godep: failed
Failed executing: /usr/local/Cellar/godep/79_1/bin/godep restore
```

Running this process:
```
brew pull https://github.com/Homebrew/homebrew-core/pull/10048
brew upgrade -s -v go
brew edit godep
```

```diff
diff --git a/Formula/godep.rb b/Formula/godep.rb
index 6ee335aec5..bdf5d8459c 100644
--- a/Formula/godep.rb
+++ b/Formula/godep.rb
@@ -3,6 +3,8 @@ class Godep < Formula
   homepage "https://godoc.org/github.com/tools/godep"
   url "https://github.com/tools/godep/archive/v79.tar.gz"
   sha256 "3dd2e6c4863077762498af98fa0c8dc5fedffbca6a5c0c4bb42b452c8268383d"
+  revision 1
+
   head "https://github.com/tools/godep.git"
 
   bottle do
@@ -25,7 +27,7 @@ class Godep < Formula
     (testpath/"Godeps/Godeps.json").write <<-EOS.undent
       {
         "ImportPath": "github.com/tools/godep",
-        "GoVersion": "go1.7",
+        "GoVersion": "go1.8",
         "Deps": [
           {
             "ImportPath": "golang.org/x/tools/cover",
```
```
brew install godep -s -v
brew test godep -v
```
Produces no problems locally.

ttaylorr:
@DomT4, @twexler -- thanks for your help, both of your comments clarified things a bunch. Judging from past PRs (in particular: https://github.com/Homebrew/homebrew-core/pull/3924), it looks like the thing to do is increment the `revision` on all formulae that depend on Go.

I'll do that now.

twexler:
> I'll do that now.

Sweet! Excited for this.

DomT4:
> it looks like the thing to do is increment the revision on all formulae that depend on Go.

Note: Only the ones that use `go` at runtime. Any that are `:build` don't need bumping, or we'll be here for days.

ttaylorr:
> Note: Only the ones that use `go` at runtime.

Thanks. I gathered from running `ack` 👍

twexler:
@ttaylorr You should also add a comment to `go.rb` mentioning this.

ttaylorr:
Thanks again, all. For posterity, Go 1.8 still supports Mountain Lion, but:

> Go 1.8 now only supports OS X 10.8 or later. This is likely the last Go release to support 10.8. Compiling Go or running binaries on older OS X versions is untested.

(ref: https://beta.golang.org/doc/go1.8#ports)

So Go 1.9 will need to update this in the formula.

Formula/go.rb
```diff
@@ -3,6 +3,10 @@ class Go < Formula
   homepage "https://golang.org"
 
   stable do
+    # Version bumps of this formula require formulae that depend on Go at
```
twexler: Is it all version bumps, or just major/minor?

twexler: #6096, #7452, and #9304 would seem to indicate it's only major/minor and not patch versions

ttaylorr: I assume it would vary between formulae, so I left it unspecific.

twexler: Ah, okay.

Formula/go.rb
```diff
@@ -3,16 +3,20 @@ class Go < Formula
   homepage "https://golang.org"
 
   stable do
-    url "https://storage.googleapis.com/golang/go1.7.5.src.tar.gz"
-    mirror "https://fossies.org/linux/misc/go1.7.5.src.tar.gz"
-    version "1.7.5"
-    sha256 "4e834513a2079f8cbbd357502cccaac9507fd00a1efe672375798858ff291815"
+    # Version bumps of this formula require formulae that depend on Go at
+    # runtime to be rebuilt in order to pass their tests. This can be done by
+    # bumping those formula's revision.
```
wkschwartz: s/those formula's revision./those formulas' revisions./

DomT4: If you want to be particularly consistent about it, Homebrew's preferred plural for formula is formulae 😉, so something like:
```
This can be done by bumping the revision of the `brew uses go` formulae.
```

DomT4: Also this comment should ideally be at the top of the formula rather than nested, and I'd perhaps limit it to "major version bumps" given those are the most problematic. But at this point I'd be tempted to leave it and let @ilovezfs tidy up when he merges 😈

Formula/go.rb
cblecker: Having patch version bumped go myself multiple times, these extra revision bumps on dependant formula were not needed. Based on what we know now, I'd recommend clarifying this to just major/minor.

Formula/go.rb
ttaylorr: Thanks for the suggestions. I don't want to bikeshed this longer than we need to, so I'm alright with deferring to @ilovezfs's judgement, if he's OK with contributing here.

I've ticked the 'Allow edits from maintainers.' box, so anyone from Homebrew core is welcome to touch this.

silentred:
Looking forward to this upgrade!

Formula/go.rb
```diff
@@ -3,16 +3,20 @@ class Go < Formula
...
+    mirror "https://fossies.org/linux/misc/go1.8.src.tar.gz"
+    version "1.8"
+    sha256 "406865f587b44be7092f206d73fc1de252600b79b3cacc587b74b5ef5c623596"
 
     go_version = version.to_s.split(".")[0..1].join(".")
     resource "gotools" do
       url "https://go.googlesource.com/tools.git",
           :branch => "release-branch.go#{go_version}",
-          :revision => "6220cba6419b2bf78aad19d85c347ecc0fda2b53"
+          :revision => "12a9eeeaa0fb4dc5c00e510411b39ab6e4b8796b"
```
ilovezfs: This appears to be the hash of the tree not the top commit, and I think specifying branch at all is actually incorrect, since that's really only for head specs that use a branch other than master.

Formula/go.rb
```diff
@@ -94,7 +88,8 @@ def install
   end
 
   def caveats; <<-EOS.undent
-    As of go 1.2, a valid GOPATH is required to use the `go get` command:
+    As of go 1.2, a valid GOPATH is required to use the `go get` command.
+    If $GOPATH is not specified, $HOME/go will be used default:
```
ilovezfs: by default

Formula/go.rb
ilovezfs: Probably no comment is needed since nothing has changed.

ilovezfs (changes requested):
Looks like bootstrap wasn't updated to

https://storage.googleapis.com/golang/go1.8.darwin-amd64.tar.gz

Formula/go.rb
ilovezfs: In the case of godep, it's currently required every time even on revision bumps. The other revision bumps in this PR are probably unnecessary altogether.

Formula/go.rb
ilovezfs: I don't think I really follow this argument. Does upstream actually say x.y should not be built with x.y itself or is this brew superstition?

DomT4: It's just a tool we've used in the past for easier build failure debugging, but upstream themselves the last time I checked used Go 1.4 to bootstrap. We moved off of that system because for a while the 1.4 branch died on Sierra, which may or may not still be the case.

DomT4: _(More broadly I think you might as well scrap the code comment and replace it with something along the lines of don't update this unless this version cannot bootstrap the new version, and effectively pin it in place)_

ilovezfs: I believe they've fixed that (definitely MacPorts has). At some point (not this PR, of course), I think we should consider building our own 1.4 bootstrapper from source and avoid the upstream binaries altogether.

In any case, assuming for the sake of argument that I accept that this use-the-prior-version-to-bootstrap approach is useful, why 1.7 not 1.7.5?

DomT4: I believe they quietly cut a new release, and we just never moved back to 1.4 because 1.7 wasn't causing any issues and changes to the `go` formula are rarely simple 😄, but I could be misremembering. I know there was talk of cutting a new 1.4 release simply for the sake of bootstrapping at least.

But yes, you're correct that having a `gobootstrap` or such from source would've allowed us to patch the 1.4 build failure at the time and keep using that branch for the process. The downside is that building even Go 1.4 is not ultra-quick in comparison to the existing process, and there hasn't been many clear upsides in the past.

> In any case, assuming for the sake of argument that I accept that this use-the-prior-version-to-bootstrap approach is useful, why 1.7 not 1.7.5?

No reason at all. I simply wanted to avoid the confusion & hassle of people updating the bootstrap for every minor release. Less moving parts, less potential for breakage, etc etc.

ilovezfs:
@DomT4 PR refreshed.

ilovezfs (approved):

DomT4:
👍 LGTM, for what that's worth these days heh.

ilovezfs:
>LGTM, for what that's worth these days heh.

I still have your badge in the drawer whenever you'd like it back :)

DomT4:
> I still have your badge in the drawer whenever you'd like it back :)

But retirement has been such a quiet place where I get to randomly work on my own mildly obscure [side projects](https://github.com/DomT4/homebrew-autoupdate) 🙈. But the desire to put up with me again, which I don't quite understand why you would want to heh, is appreciated as always.

ilovezfs:
@ttaylorr thanks for the PR! :rocket:

ilovezfs:
>But the desire to put up with me again, which I don't quite understand why you would want to heh, is appreciated as always.

:bowing_man:

freeformz:
FWIW ....

Given this error message:

```
godep: [WARNING]: godep should only be used inside a valid go package directory and
godep: [WARNING]: may not function correctly. You are probably outside of your $GOPATH.
godep: [WARNING]:	Current Directory: /private/tmp/godep-test-20170216-59376-z4jxm2
godep: [WARNING]:	$GOPATH: /private/tmp/godep-test-20170216-59376-z4jxm2
```

The current directory (`/private/tmp/godep-test-20170216-59376-z4jxm2`) is not a valid location for go packages to exist at. The minimal filepath would be: `/private/tmp/godep-test-20170216-59376-z4jxm2/src/<anything>`. So godep is telling you that it should "only be used inside of a valid go package directory". This may not be the best wording and I'm open to suggestions to fix the wording. I should also add a test to see if `! strings.HasPrefix(os.Getwd(), $GOPATH)` and only then  make the $GOPATH suggestion.

ilovezfs:
That's usually a warning (not an error) that occurs when we use the symlink black magic.

PR #10065, opened 2017-02-17
mariadb@10.0 10.0.28 (new formula)

kuahyeow:

-----

### DIFF
--- a/Formula/mariadb@10.0.rb
+++ b/Formula/mariadb@10.0.rb
@@ -1,0 +1,161 @@
+class MariadbAT100 < Formula
+  desc "Drop-in replacement for MySQL"
+  homepage "https://mariadb.org/"
+  url "https://fossies.org/linux/misc/mariadb-10.0.28.tar.gz"
+  sha256 "0a5033d56f1c5403df7fabd519ccbdc4da308b30a1ecd52e06a19be182a9bab2"
+
+  keg_only :versioned_formula
+
+  option "with-test", "Keep test when installing"
+  option "with-bench", "Keep benchmark app when installing"
+  option "with-embedded", "Build the embedded server"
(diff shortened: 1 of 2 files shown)


kuahyeow:
This formula has to have `conflicts_with` as it provides the same binaries as `mysql`

kuahyeow:
Have opened https://github.com/Homebrew/homebrew-versions/pull/1526 as well

makigumo:
Maybe it's a good idea to weed out the `build.devel?`, `build.stable?` stuff for a simpler formula.

Also it probably needs to `conflicts_with` the unversioned `mariadb` package.

kuahyeow:
> Maybe it's a good idea to weed out the build.devel?, build.stable? stuff for a simpler formula.

Good idea, have done so.

> Also it probably needs to conflicts_with the unversioned mariadb package.

Hmm, I'm not sure. Seems like `brew audit` is explicitly telling me to use `keg_only :versioned_formula` to deal with rather than `conflicts_with`. Seems to be consistent if I look at the formula for mysql@5.5 for example (https://github.com/Homebrew/homebrew-core/blob/master/Formula/mysql%405.5.rb).

So currently there's seems to be audit check against any `conflicts_with` in a versioned formula. I'm guessing this new formula is a new exception ?

makigumo:
Yes you are right, I believe this is the corresponding issue: https://github.com/Homebrew/brew/pull/1851

MikeMcQuaid:
> This formula has to have conflicts_with as it provides the same binaries as mysql

Once a formula is `keg_only` the `conflicts_with` are no longer needed so can be removed. Thanks!

kuahyeow:
Thanks for the guidance @MikeMcQuaid. I have removed the `conflicts_with`.

fxcoudert:
Thanks @kuahyeow for your contribution!

PR #10085, opened 2017-02-17
wine 2.2 (devel)

uyjulian:


ilovezfs:
@MikeMcQuaid @mistydemeo well as @JCount pointed out to me, it seems that upstream has managed to break the current flickering patch such that it no longer applies cleanly. @mistydemeo any update on the status of the rdar you opened?

uyjulian:
I'm thinking about just removing all the patches, and adding wine-staging support later on.

ilovezfs:
Users will not be entirely thrilled with the flickering if we remove the patches.

JCount:
I believe that I have been able to successfully adapt the patch for the screen flickering issue to apply cleanly and correctly to the 2.2 development release of wine. I haven't done any rigorous testing, however. I do wish there were a proper fix.

LinusU:
I'm currently working on updating the patch for Wine 2.2, if the build will just finish on my computer before I have to run I should be able to submit it shortly...

Please do not merge this without the flickering patch updated since it breaks a ton of games on macOS. The developers of Wine thinks that this is bug in the macOS OpenGL APIs which is why it isn't merged upstream. I can understand their situation but I think it's highly relevant for brew to continue to carry this patch. Please see the bug for more information: https://bugs.winehq.org/show_bug.cgi?id=34166

ilovezfs (changes requested):
let's wait until the patch that upstream should have merged long ago but hasn't is rebased

LinusU:
Soooo, for some reason I cannot get any of the programs that I wanted to test this on running on a pure 2.0 build. This should be unrelated to this specific patch since I haven't applied it yet, but without getting a program that has the problem to run, I cannot test to see if the patch works...

Does anyone have any recommendations on games/applications that I could try?

ilovezfs:
>Soooo, for some reason I cannot get any of the programs that I wanted to test this on running on a pure 2.0 build.

Do they work with 2.1?

LinusU:
I haven't had time to test yet, but it seems like to logical next step 👍

jrtc27:
The versions of wine-mono and wine-gecko have been outdated for a while; I don't think sourceforge has the updated versions, but http://dl.winehq.org/wine/ has several new versions.

fxcoudert:
Patch was rebased in #10386, which tests fine, so closing this PR.

PR #10090, opened 2017-02-17
Remove some non-Homebrew-specific caveats

alyssais:

-----

Following [this discussion](https://github.com/Homebrew/homebrew-core/commit/b30889b47eca461706ca57b6908f221129b7ffa4#commitcomment-20939142) with @MikeMcQuaid, I went through the caveats of all the formulae I had installed and removed any non-Homebrew-specific caveats.

There were some I wasn't sure of, so I didn't include them in this PR:

- gpg-agent
- python
- python3

### DIFF
--- a/Formula/chruby.rb
+++ b/Formula/chruby.rb
@@ -20,28 +20,7 @@ class Chruby < Formula
 
   def caveats; <<-EOS.undent
     Add the following to the ~/.bashrc or ~/.zshrc file:
-
       source #{opt_share}/chruby/chruby.sh
-
-    By default chruby will search for Rubies installed into /opt/rubies/ or
-    ~/.rubies/. For non-standard installation locations, simply set the RUBIES
-    variable after loading chruby.sh:
-
-      RUBIES+=(
-        /opt/jruby-1.7.0
-        $HOME/src/rubinius
-      )
-
-    If you are migrating from another Ruby manager, set `RUBIES` accordingly:
-
-      RVM:   RUBIES+=(~/.rvm/rubies/*)
-      rbenv: RUBIES+=(~/.rbenv/versions/*)
-      rbfu:  RUBIES+=(~/.rbfu/rubies/*)
-
-    To enable auto-switching of Rubies specified by .ruby-version files,
-    add the following to ~/.bashrc or ~/.zshrc:
-
-      source #{opt_share}/chruby/auto.sh
     EOS
   end
 end
--- a/Formula/go.rb
+++ b/Formula/go.rb
@@ -79,16 +79,6 @@ class Go < Formula
     end
   end
 
-  def caveats; <<-EOS.undent
-    As of go 1.2, a valid GOPATH is required to use the `go get` command.
-    If $GOPATH is not specified, $HOME/go will be used by default:
-      https://golang.org/doc/code.html#GOPATH
-
-    You may wish to add the GOROOT-based install location to your PATH:
-      export PATH=$PATH:#{opt_libexec}/bin
-    EOS
-  end
-
   test do
     (testpath/"hello.go").write <<-EOS.undent
     package main
--- a/Formula/liquidprompt.rb
+++ b/Formula/liquidprompt.rb
@@ -17,13 +17,6 @@ class Liquidprompt < Formula
       if [ -f #{HOMEBREW_PREFIX}/share/liquidprompt ]; then
         . #{HOMEBREW_PREFIX}/share/liquidprompt
       fi
-
-    If you'd like to reconfigure options, you may do so in ~/.liquidpromptrc.
-    A sample file you may copy and modify has been installed to
-      #{HOMEBREW_PREFIX}/share/liquidpromptrc-dist
-
-    Don't modify the PROMPT_COMMAND variable elsewhere in your shell config;
-    that will break things.
     EOS
   end
 
--- a/Formula/tor.rb
+++ b/Formula/tor.rb
@@ -39,15 +39,6 @@ class Tor < Formula
     system "make", "install"
   end
 
-  def caveats; <<-EOS.undent
-    You will find a sample `torrc` file in #{etc}/tor.
-    It is advisable to edit the sample `torrc` to suit
-    your own security needs:
-      https://www.torproject.org/docs/faq#torrc
-    After editing the `torrc` you need to restart tor.
-    EOS
-  end
-
   plist_options :manual => "tor"
 
   def plist; <<-EOS.undent
--- a/Formula/zsh-git-prompt.rb
+++ b/Formula/zsh-git-prompt.rb
@@ -11,11 +11,8 @@ class ZshGitPrompt < Formula
   end
 
   def caveats; <<-EOS.undent
-    First, make sure zsh-git-prompt is loaded from your .zshrc:
+    Make sure zsh-git-prompt is loaded from your .zshrc:
       source "#{opt_prefix}/zshrc.sh"
-
-    Then include $(git_super_status) in your PROMPT or RPROMPT, e.g.:
-      PROMPT='%B%m%~%b$(git_super_status) %# '
     EOS
   end
 
--- a/Formula/zsh-syntax-highlighting.rb
+++ b/Formula/zsh-syntax-highlighting.rb
@@ -20,17 +20,7 @@ class ZshSyntaxHighlighting < Formula
   def caveats
     <<-EOS.undent
     To activate the syntax highlighting, add the following at the end of your .zshrc:
-
       source #{HOMEBREW_PREFIX}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
-
-    You will also need to force reload of your .zshrc:
-
-      source ~/.zshrc
-
-    Additionally, if you receive "highlighters directory not found" error message,
-    you may need to add the following to your .zshenv:
-
-      export ZSH_HIGHLIGHT_HIGHLIGHTERS_DIR=#{HOMEBREW_PREFIX}/share/zsh-syntax-highlighting/highlighters
     EOS
   end
 
--- a/Formula/awscli.rb
+++ b/Formula/awscli.rb
@@ -33,14 +33,6 @@ class Awscli < Formula
   def caveats; <<-EOS.undent
     The "examples" directory has been installed to:
       #{HOMEBREW_PREFIX}/share/awscli/examples
-
-    Before using aws-cli, you need to tell it about your AWS credentials.
-    The quickest way to do this is to run:
-      aws configure
-
-    More information:
-      https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html
-      https://pypi.python.org/pypi/awscli#getting-started
     EOS
   end
 
--- a/Formula/zsh-history-substring-search.rb
+++ b/Formula/zsh-history-substring-search.rb
@@ -10,10 +10,4 @@ class ZshHistorySubstringSearch < Formula
     inreplace "README.md", "source zsh-history", "source #{opt_prefix}/zsh-history"
     prefix.install Dir["*.zsh"]
   end
-
-  def caveats; <<-EOS.undent
-    For setup instructions:
-      more "#{opt_prefix}/README.md"
-    EOS
-  end
 end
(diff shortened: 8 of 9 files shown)


Formula/go.rb
```diff
@@ -79,16 +79,6 @@ def install
     end
   end
 
-  def caveats; <<-EOS.undent
```
ilovezfs: let's leave go alone

Formula/liquidprompt.rb
```diff
@@ -12,21 +12,6 @@ def install
     share.install "liquidprompt"
   end
 
-  def caveats; <<-EOS.undent
```
ilovezfs: this tells me relevant locations, which are homebrew specific

Formula/chruby.rb
```diff
@@ -17,31 +17,4 @@ class Chruby < Formula
   def install
     system "make", "install", "PREFIX=#{prefix}"
   end
-
-  def caveats; <<-EOS.undent
-    Add the following to the ~/.bashrc or ~/.zshrc file:
-
-      source #{opt_share}/chruby/chruby.sh
```
ilovezfs: this location is homebrew specific information

Formula/zsh-syntax-highlighting.rb
```diff
@@ -17,23 +17,6 @@ def install
     system "make", "install", "PREFIX=#{prefix}"
   end
 
-  def caveats
-    <<-EOS.undent
-    To activate the syntax highlighting, add the following at the end of your .zshrc:
-
-      source #{HOMEBREW_PREFIX}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
```
ilovezfs: again: the location is homebrew specific information

ilovezfs (changes requested):
Removing caveats that inform about homebrew-specific locations, which will not be identified specifically in the upstream docs, is counterproductive.

Formula/go.rb
alyssais: because…?

alyssais:
> Removing caveats that inform about homebrew-specific-locations, which will not be identified specifically in the upstream docs, is counterproductive.

I'm inclined to agree, but I'll wait to see what @MikeMcQuaid thinks since I think that contradicts https://github.com/Homebrew/homebrew-core/commit/b30889b47eca461706ca57b6908f221129b7ffa4#commitcomment-20939307. :)

Formula/go.rb
ilovezfs: because every line in that formula has been bike-shedded to death already

Formula/zsh-git-prompt.rb
```diff
@@ -10,15 +10,6 @@ def install
     prefix.install Dir["*.{sh,py}"]
   end
 
-  def caveats; <<-EOS.undent
-    First, make sure zsh-git-prompt is loaded from your .zshrc:
-      source "#{opt_prefix}/zshrc.sh"
```
ilovezfs: homebrew specific.

Formula/go.rb
DomT4: Agree with @ilovezfs here, FWIW.

Formula/tor.rb
```diff
@@ -39,15 +39,6 @@ def install
     system "make", "install"
   end
 
-  def caveats; <<-EOS.undent
```
DomT4: Beside logging every website I visit and making it accessible to random chunks of the government on the flimsiest of reasons I'm fortunate to live in a fairly free country, so I can't be certain of this, but there's a potential these caveats remain a useful nudge towards documentation in places where the "reasonable defaults" applied if you don't edit the `torrc` and remove the `.sample` from the filename are not entirely reasonable.

Formula/chruby.rb
MikeMcQuaid: Let's trim the caveats down to just this line and the lines above and skip the empty line.

Formula/go.rb
MikeMcQuaid: @ilovezfs I think we can remove the "As of go 1.2" at this point. Personally I think we should just remove all the caveats now as they aren't Homebrew specific.

Formula/liquidprompt.rb
MikeMcQuaid: Let's trim this just to lines 16-19. The rest are overkill.

Formula/tor.rb
MikeMcQuaid: This is not Homebrew specific information. Homebrew specific information is what the caveats are for so 👍 to removing them.

Formula/zsh-git-prompt.rb
MikeMcQuaid: Let's remove the "First" and then just reference the `source` and preceding line and skip the rest.

Formula/zsh-syntax-highlighting.rb
MikeMcQuaid: Let's show just the first line of text and the first `source` and skip the rest.

MikeMcQuaid:
> I'm inclined to agree, but I'll wait to see what @MikeMcQuaid thinks since I think that contradicts b30889b#commitcomment-20939307. :)

I'll update the docs in Homebrew/brew to clarify.

MikeMcQuaid:
> I'll update the docs in Homebrew/brew to clarify.

https://github.com/Homebrew/brew/pull/2063

alyssais:
Made all those changes apart from `go`.

Formula/go.rb
ilovezfs: The location of the GOROOT is Homebrew specific information.

Formula/zsh-syntax-highlighting.rb
```diff
@@ -20,17 +20,7 @@ def install
...
       source #{HOMEBREW_PREFIX}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
-
-    You will also need to force reload of your .zshrc:
-
-      source ~/.zshrc
-
-    Additionally, if you receive "highlighters directory not found" error message,
-    you may need to add the following to your .zshenv:
-
-      export ZSH_HIGHLIGHT_HIGHLIGHTERS_DIR=#{HOMEBREW_PREFIX}/share/zsh-syntax-highlighting/highlighters
```
ilovezfs: Homebrew specific information.

Formula/go.rb
ilovezfs: I'm fine with deleting the five words "As of go 1.2, " though :laughing:

Formula/chruby.rb
```diff
@@ -20,28 +20,7 @@ def install
...
-    If you are migrating from another Ruby manager, set `RUBIES` accordingly:
-
-      RVM:   RUBIES+=(~/.rvm/rubies/*)
-      rbenv: RUBIES+=(~/.rbenv/versions/*)
-      rbfu:  RUBIES+=(~/.rbfu/rubies/*)
-
-    To enable auto-switching of Rubies specified by .ruby-version files,
-    add the following to ~/.bashrc or ~/.zshrc:
-
-      source #{opt_share}/chruby/auto.sh
```
ilovezfs: Homebrew specific location

ilovezfs (changes requested):
>Made all those changes apart from go.

The `go` caveat is fine as-is. Let's leave it alone.

Formula/tor.rb
DomT4: I think we've seen pretty clear evidence in the past removing all basic usage nudges results in more issues being opened against Homebrew along the lines of x software doesn't work please help. As much as I agree with a lot of caveats being trimmed back I do think long-term you're increasing your own burden here. Not a discussion I expect to win, but FWIW.

MikeMcQuaid: I don't agree, sorry.

> Not a discussion I expect to win, but FWIW.

Please either quit with these caveats or quit with the comments that precede them, thanks. If you believe your opinion to be genuinely ignored: self-censor. If you think you'll be listened to: don't self-sabotage your own argument.

Formula/go.rb
MikeMcQuaid: I think the default GOPATH situation is fine; why would it not be? Similarly, why would I want to add the GOROOT location to my path? I've written Go in anger and this provides no insight as to why I'd do that so seems useless.

MikeMcQuaid: But I don't care enough: feel free to 🚢 either way

Formula/tor.rb
DomT4: > Please either quit with these caveats or quit with the comments that precede them, thanks. If you believe your opinion to be genuinely ignored: self-censor. If you think you'll be listened to: don't self-sabotage your own argument.

For flip sake Mike, I wasn't taking a dig at you, and I wish you'd stop reacting like I'm out to get you every time I breathe something you consider negative.

My comment was purely a notation that whilst I thought my comment would be taken on board I didn't expect it to shift the winds significantly enough for the work here to be discarded. I don't expect my dislike of President Trump's anti-immigration rhetoric to make a single drop of difference, and I'd have to be an idiot to think otherwise, but it doesn't mean I'm going to sit there and go _"Yes I love this, it's great"_.

MikeMcQuaid: > For flip sake Mike, I wasn't taking a dig at you, and I wish you'd stop reacting like I'm out to get you every time I breathe something you consider negative.

I don't think everything is a dig at me but I do find the "here's what I think but no-one listens to me now" thinking pretty tiresome when it's repeated.

DomT4: That's how you're choosing to read it, which is frustrating. My point is more "I don't expect anyone to _have_ to adhere to my concerns", whereas when I was the one handling the extreme majority of the work around formulae pre-ILZ unless someone else wanted to take that level of work on people kind of did.

Formula/liquidprompt.rb
```diff
@@ -17,13 +17,6 @@ def caveats; <<-EOS.undent
       if [ -f #{HOMEBREW_PREFIX}/share/liquidprompt ]; then
         . #{HOMEBREW_PREFIX}/share/liquidprompt
       fi
-
-    If you'd like to reconfigure options, you may do so in ~/.liquidpromptrc.
-    A sample file you may copy and modify has been installed to
-      #{HOMEBREW_PREFIX}/share/liquidpromptrc-dist
```
JCount: You might potentially want to keep this part as well, since it points the where Homebrew installed the sample `.liquidpromptrc`, but it isn't nearly as important as the first section of the caveat, which is required for basic functionality.

alyssais: I don't think it's worth keeping — that sample file will be available online for people that want it, and I can't see why it would ever be important to anybody where Homebrew kept its copy of it. Honestly, there's probably not a lot of point in Homebrew even installing it.

ilovezfs: We routinely install such sample files.

alyssais:
I feel this change is pretty trivial in terms of importance, and it's **way** more controversial than I expected it to be when I proposed it. To be honest, I can already feel some pretty nasty burnout coming on as a result of my frustration with not knowing what I'm supposed to be doing here, and I'm not willing to let that happen.

So, I'm going to close this PR. I don't plan on spending any more time on this — it's just not worth the amount of effort it's turned out to require. I could be doing much more useful work instead. If somebody else wants to pursue this, though, they are of course welcome to pick it up. :)

***

To be clear: I'm not making a dig at anybody here. There's a difference of opinions on what's best here, and that's okay! It's how we build good software. This just isn't enough of a priority for me for it to be worth it.

Formula/tor.rb
MikeMcQuaid: @DomT4 If someone is consistently misreading your communication there's at least a chance your communication could be improved. I've made a fairly simple request which you're free to ignore if you choose.

MikeMcQuaid:
I pushed everything here that didn't involve Homebrew specific locations and left Go alone (last commit was bcc3675711).

ilovezfs:
:confetti_ball:

JCount:
🍁

Formula/tor.rb
DomT4: Honestly, I can't be bothered to continue an argument a day later Mike. I've just taken up running my own fork as a primary, it's easier than this.

MikeMcQuaid: @DomT4 Ok, up to you. I'm not wanting to restart and argument either; I'm trying to indicate what would be a more effective communication style here given both of us having better things to do than bicker with one another.

Issue #9845, opened 2017-02-10
Symlink instructions for cups-pdf don't work on El Capitan 


greylurk:
The instructions for cups-pdf that are echoed upon install are no longer working in OS-X El Capitan.

https://github.com/Homebrew/homebrew-core/blob/master/Formula/cups-pdf.rb#L36

The /usr/share/cups/model/ so when attempting to install a symlink into that directory, you'll get an error message: 

```
$ sudo ln -sf /usr/local/opt/cups-pdf/share/cups/model/CUPS-PDF.ppd /usr/share/cups/model/CUPS-PDF.ppd
ln: /usr/share/cups/model/CUPS-PDF.ppd: Operation not permitted
```

It appears that this is blocked by Apple's "System Integrity Protection" (SIP) layer that was introduced in El-Capitan. See https://derflounder.wordpress.com/2015/10/01/system-integrity-protection-adding-another-layer-to-apples-security-model/ for details

ilovezfs:
Thanks for pointing this out. I suspect we'll remove the formula.

ilovezfs:
Formula boneyarded. https://github.com/Homebrew/homebrew-core/pull/10121

bryan3000000:
It is not necessary to link or copy the PPD file to /usr/share/cups/model/.  Instead, the printer can be installed by 1) manually choosing the PPD file from the Add Printer dialog (choose Use: other), or 2) using the "-P" option to specify the location of the PPD file when installing the printer using lpadmin.  I see no reason to remove this formula.

breandan:
@bryan3000000 Thanks, worked for me. If you encounter this issue, try running `lpadmin -p CUPS-PDF -P /opt/local/share/cups/model/CUPS-PDF.ppd`.

PR #10121, opened 2017-02-19
cups-pdf: migrate to boneyard

ilovezfs:
Created with `brew boneyard-formula-pr` because it needs write-access to SIP protected locations.


ilovezfs:
See https://github.com/Homebrew/homebrew-core/issues/9845.

PR #10130, opened 2017-02-19
vim 8.0.0341

### LINKED ISSUE #8716: --with-mzscheme does not compile vim 8 with mzscheme
Hey guys, just noticed this today. I don't need mzscheme but someone might, someday. Brew seems fine; this seems to be the Vim 8 formula.

I'm working with a test system on a fresh install. The requested steps:
```bash
$ brew update
Already up-to-date.

$ brew doctor 
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: A newer Command Line Tools release is available.
Update them from Software Update in the App Store.

Warning: Putting non-prefixed coreutils in your path can cause gmp builds to fail.

Warning: Putting non-prefixed findutils in your path can cause python builds to fail.
```

### Bug reports:
The config
```bash
$ brew config 
HOMEBREW_VERSION: >1.1.0 (no git repository)
ORIGIN: https://github.com/Homebrew/brew
HEAD: 9cce3414fbd39d146d97911dbb3bf07fffaa1080
Last commit: 27 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: c0e2a9033c42e0911f87a24c8960dca9d80890f4
Core tap last commit: 48 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: single-core 64-bit ivybridge
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.11.0 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /usr/bin/ruby => /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby
Java: N/A
macOS: 10.12-x86_64
Xcode: N/A
CLT: 8.0.0.0.1.1472435881
X11: N/A
```

The Steps and output:
```bash
brew install vim --override-system-vi --disable-nls --with-lua --with-mzscheme --with-tcl
Warning: vim: --disable-nls was deprecated; using --without-nls instead!
Warning: vim: --override-system-vi was deprecated; using --with-override-system-vi instead!
==> Using the sandbox
==> Downloading https://github.com/vim/vim/archive/v8.0.0165.tar.gz
Already downloaded: /Users/vagrant/Library/Caches/Homebrew/vim-8.0.0165.tar.gz
==> ./configure --prefix=/usr/local --mandir=/usr/local/Cellar/vim/8.0.0165/share/man --enable-multibyte --with-tlib=ncurses --ena
==> make
==> make install prefix=/usr/local/Cellar/vim/8.0.0165 STRIP=/usr/bin/true
🍺  /usr/local/Cellar/vim/8.0.0165: 1,400 files, 22.2M, built in 1 minute 54 seconds

$ vim --version | egrep 'mzschem'
+byte_offset     -gettext         -mzscheme        +textobjects
```

ilovezfs:

-----

remove mzscheme option

Fixes https://github.com/Homebrew/homebrew-core/issues/8716.

~~CC @todd-dsm @yuemori @minhyeong-kim mind testing this fixes mzscheme for you?~~

### DIFF
--- a/Formula/vim.rb
+++ b/Formula/vim.rb
@@ -1,8 +1,8 @@
 class Vim < Formula
   desc "Vi \"workalike\" with many additional features"
   homepage "http://www.vim.org/"
-  url "https://github.com/vim/vim/archive/v8.0.0337.tar.gz"
-  sha256 "a0eab15ac801a1fda1b65c93e6603f20f447762f5be65893fcb464a84b8c5a72"
+  url "https://github.com/vim/vim/archive/v8.0.0341.tar.gz"
+  sha256 "c91548d7bc01c00866acfdacab6b5cd1b8532b0ab1e90b4dbaa32885d33e0dea"
   head "https://github.com/vim/vim.git"
 
   bottle do
@@ -18,7 +18,7 @@ class Vim < Formula
   option "without-nls", "Build vim without National Language Support (translated messages, keymaps)"
   option "with-client-server", "Enable client/server mode"
 
-  LANGUAGES_OPTIONAL = %w[lua mzscheme python3 tcl].freeze
+  LANGUAGES_OPTIONAL = %w[lua python3 tcl].freeze
   LANGUAGES_DEFAULT  = %w[perl python ruby].freeze
 
   if MacOS.version >= :mavericks


ilovezfs:
@DomT4 or shall I just take the mzscheme option out back and dispense with it?

DomT4:
I didn't even know there was any desire for us to support the option, so 😅. What's with the hardcoding of `/usr/local/` for this?

If we're making changes, something that I've been meaning to do for a while is enforce a single `lua` option here. If you build with `--with-lua --with-luajit` it links to only `luajit`, so the options should be mutually exclusive.

ilovezfs:
>What's with the hardcoding of /usr/local/ for this?

It's the location of `mzscheme` if you've installed `minimal-racket` and then run `raco pkg install -i main-distribution`.

The other way to get mzcheme is with the `racket` cask, but that installs a putrid `/Applications/Racket v6.8/` which instead installs a symlink `/usr/local/bin/mzscheme` pointing to `/Applications/Racket v6.8/bin/mzscheme`, and the space in the path breaks vim's build system. I can fix that by moving `/Applications/Racket v6.8/` to `/Applications/Racket` and then correcting the symlinks, etc.

So the only mzscheme that's actually feasible to support sanely is the `minimal-racket` formula followed by user installation of the non-minimal bits with the built-in package manager thingy.

ilovezfs:
Regarding `/usr/local` specifically, I have no idea where `raco` will end up installing `mzscheme` if you have have a non-standard Homebrew prefix.

Also, arguably `vim` should fix configure so it doesn't break with the spaces but I'm not terribly sympathetic to the plight of that wretched cask.

DomT4:
I hate to suggest it, but this very much sounds like one of those nasty _"Could use a custom requirement"_ problems to check someone actually has `mzscheme` installed, and potentially locate it if the `HOMEBREW_PREFIX` is not `/usr/local`.

ilovezfs:
Indeed, which is also the point at which my finger involuntarily moves to the `delete` key.

DomT4:
Aye, unless there's wild demand I wouldn't personally consider it reasonable to expect Homebrew to tack it into the formula given it needs ideally both a custom requirement & an `inreplace` to work as intended consistently.

ilovezfs:
@DomT4 option terminated.

Issue #10133, opened 2017-02-19
PERL5LIB conflicts with system perl produce a bad story, what's the solution?


johnhawkinson:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [ ] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Bug report summary
Having installed homebrew's perl package, and followed the guidelines from the formula's Caveats section, I have PERL5LIB set to ~/perl5/lib/perl5, and have some related environment variables similarly set as well.

The result of this is that sometimes the system perl (/usr/bin/perl) breaks mysteriously. Ironically this is often via Homebrew. For instance, upgrading `openssl` failed like this:
```
==> make test
Last 15 lines from /Users/jhawk/Library/Logs/Homebrew/openssl/05.make:
Server's key = 8A7DEB34EAF1EDC085FD2CB54B8F2E8DC9A23D2FCF7B0AD27A1235FE61E6EAFC2
C25ED98B191A98665BE630539F561B69D0BE0462C6A140047B1C962469F21E77BCDC180194885355
61EF6D0AC391CDF70BE4BB787C3E4F883740D094A2A57620673FCE676FEA50AD734BB8186F8DF27B
651E62161F7CE5D71ED77A7B5BA8CFA
CMS consistency test
/usr/bin/perl cms-test.pl
Usage: DynaLoader::dl_find_symbol(libhandle, symbolname) at /Users/jhawk/perl5/l
ib/perl5/darwin-thread-multi-2level/XSLoader.pm line 89.
Compilation failed in require at /System/Library/Perl/5.18/darwin-thread-multi-2
level/IO/Handle.pm line 269.
```

A workaround is to unset PERL5LIB, but that means I don't have access my locally installed perl modules, which is a problem.

How is this supposed to work? Having a $PERL5LIB that breaks /usr/bin/perl doesn't seem a tenable situation.

The system perl is v5.18.2, the homebrew /usr/local/bin/perl is v5.24.0.
Installed perl modules are all in the $PERL5LIB top-level, and $PERL5LIB/5.24.0/ contains only an empty `darwin-thread-multi-2level/` subdir.

### Caveats and .bashrc

Here's what the Caveats section says:
```
pb3:test jhawk$ brew info perl
...
==> Caveats
[...]

PR #9291, opened 2017-01-26
openssl 1.0.2k

alex:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ ] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [ ] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #10141, opened 2017-02-19
openssl: Ignore PERL5LIB while we ignore PERL

johnhawkinson:
Per #10133. If we use /usr/bin/perl with a PERL5LIB intended for /usr/local/bin/perl, perl can break in bizarre ways, like:

```
CMS consistency test
/usr/bin/perl cms-test.pl
Usage: DynaLoader::dl_find_symbol(libhandle, symbolname) at /Users/jhawk/perl5/l
ib/perl5/darwin-thread-multi-2level/XSLoader.pm line 89.
```

-----

### DIFF
--- a/Formula/openssl.rb
+++ b/Formula/openssl.rb
@@ -44,8 +44,10 @@ class Openssl < Formula
 
   def install
     # OpenSSL will prefer the PERL environment variable if set over $PATH
-    # which can cause some odd edge cases & isn't intended. Unset for safety.
+    # which can cause some odd edge cases & isn't intended. Unset for safety,
+    # along with perl modules in PERL5LIB.
     ENV.delete("PERL")
+    ENV.delete("PERL5LIB")
 
     # Load zlib from an explicit path instead of relying on dyld's fallback
     # path, which is empty in a SIP context. This patch will be unnecessary


ilovezfs:
@johnhawkinson did you verify this PR fixes the issue(s) you were experiencing?

ilovezfs (dismissed):
We should probably do this in Homebrew/brew so it fixes the same problem elsewhere.

ilovezfs:
I could see a case for just doing it here since we don't `ENV.delete("PERL")` everywhere at this point, so it's arguably an openssl specific issue.

johnhawkinson:
@ilovezfs:
> did you verify this PR fixes the issue(s) you were experiencing?

I did so _in concept_ by having `brew update openssl` fail, and having `PERL5LIB=  brew update openssl` succeed, although I did not affirmatively test this individual line here by uninstalling my successful openssl build and rebuilding it.

> I could see a case for just doing it here since we don't ENV.delete("PERL") everywhere at this point, so it's arguably an openssl specific issue.

I think you are correct -- deleting `PERL` without deleting `PERL5LIB` is what triggers the problem. Perhaps openssl shouldn't delete `PERL` all by its lonesome, but that's a bigger issue and will likely have other Consequences.

ilovezfs:
@DomT4 any thoughts here? I see openssl@1.1 went all in on brew perl but that wasn't propagated back to this formula.

DomT4:
Only for Mountain Lion and below I think? `openssl@1.1` needs a quite a bit newer version of `perl` to test against, and the tests are run by default for obvious reasons. The older `openssl` is fine testing against older system `perl` versions, at least, the last time we were actually able to verify that.

DomT4:
I do still think unsetting it using `superenv` is correct though. Any formulae that actually _need_ to use a `PERL5LIB` should be setting one in the formula for sandboxing reasons, amongst other things.

ilovezfs:
Got it. This seems fine to me then and like it's fixing a genuine bug.

ilovezfs:
>I do still think unsetting it using superenv is correct though. Any formulae that actually need to use a PERL5LIB should be setting one in the formula for sandboxing reasons, amongst other things.

Right and can't we say the same about PYTHONPATH?

DomT4:
Tim's the expert there, but I'd argue that comes with the same set of potential problems, aye.

johnhawkinson:
I'm not sure I follow what the `PYTHONPATH` issues are, but I do think users are much more likely to set a persistent `PERL5LIB` (to allow cpan to work!) than they are to set a persistent `PYTHONPATH,` because pip/easy_install/etc. do not require setting `PYTHONPATH`. 

That's not to say `PYTHONPATH` can't generate the same problems (but it may say they are less common; of course, these days perl is less common than python, so...). And I'm still missing some context :)

ilovezfs:
The CI failures are pre-existing: https://bot.brew.sh/job/Homebrew%20Testing/1310/#showFailuresLink

I'm not sure if the last openssl upgrade got merged with this happening or if they've developed since then.

ilovezfs:
It seems the last openssl upgrade was merged with them happening: https://github.com/Homebrew/homebrew-core/pull/9291#issuecomment-275562608

MikeMcQuaid (approved):

PR #10146, opened 2017-02-20
mariadb-connector-odbc 2.0.13 (new formula)

unilynx:

-----

This adds a formula for the mariadb odbc connector, which should be useful now that the mysql odbc connector ended up on the boneyard.

I'm not sure about adding /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/ as an include path in the changes to FindIconv.cmake, there should be a cleaner way to do that..

### DIFF
--- a/Formula/mariadb-connector-odbc.rb
+++ b/Formula/mariadb-connector-odbc.rb
@@ -1,0 +1,57 @@
+class MariadbConnectorOdbc < Formula
+  desc "MariaDB Standardized database driver"
+  homepage "https://downloads.mariadb.org/connector-odbc/"
+  url "https://github.com/MariaDB/mariadb-connector-odbc/archive/2.0.13.tar.gz"
+  sha256 "6b77e879335ea32dad40d42de95930df8ad557faa887b414a09ab7ead3ce963c"
+
+  depends_on "cmake" => :build
+  depends_on "mariadb-connector-c"
+  depends_on "unixodbc"
+  depends_on "openssl"
+
+  patch :DATA
+
+  def install
+    system "cmake", ".", "-DWITH_UNIXODBC=1", "-DMARIADB_FOUND=1", "-DWITH_OPENSSL=1", "-DOPENSSL_INCLUDE_DIR=/usr/local/opt/openssl/include/", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    output = shell_output("#{Formula["unixodbc"].bin}/dltest #{lib}/libmaodbc.dylib")
+    assert_equal "SUCCESS: Loaded #{lib}/libmaodbc.dylib\n", output
+  end
+end
+
+__END__
+diff --git a/cmake/FindIconv.cmake b/cmake/FindIconv.cmake
+index f0768f7..a3f3eaa 100644
+--- a/cmake/FindIconv.cmake
++++ b/cmake/FindIconv.cmake
+@@ -9,6 +9,7 @@ IF(APPLE)
+   find_path(ICONV_INCLUDE_DIR iconv.h PATHS
+             /opt/local/include/
+             /usr/include/
++            /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/
+             NO_CMAKE_SYSTEM_PATH)
+ ELSE()
+   find_path(ICONV_INCLUDE_DIR iconv.h)
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 4104ad0..a9bbf32 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -127,6 +127,15 @@ IF(NOT WIN32)
+   ELSE()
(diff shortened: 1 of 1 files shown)


Formula/mariadb-connector-odbc.rb
```diff
@@ -0,0 +1,66 @@
...
+  # So the next best choice seems to be github:
+  url "https://github.com/MariaDB/mariadb-connector-odbc/archive/2.0.13.tar.gz"
+  sha256 "6b77e879335ea32dad40d42de95930df8ad557faa887b414a09ab7ead3ce963c"
+
+  depends_on "cmake" => :build
+  depends_on "mariadb-connector-c"
+  depends_on "unixodbc"
+  depends_on "openssl"
+
+  patch :DATA
```
bfontaine: Why is this patch necessary? Was it submitted to upstream?

Formula/mariadb-connector-odbc.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def install
+    args = std_cmake_args
+    args << "-DWITH_UNIXODBC=1"
+    args << "-DMARIADB_FOUND=1"
+    args << "-DWITH_OPENSSL=1"
+    args << "-DOPENSSL_INCLUDE_DIR=/usr/local/opt/openssl/include/"
+
+    system "cmake", ".", *args
```
bfontaine: You can probably avoid making an array here, e.g.:

```rb
system "cmake", ".", "-DWITH_UNIXODBC=1", …, *std_cmake_args
```

Formula/mariadb-connector-odbc.rb
unilynx: It doesn't build without. upstream requires even more fixes to build outside of a Homebrew formula, these are now submitted as https://github.com/MariaDB/mariadb-connector-odbc/pull/13

Formula/mariadb-connector-odbc.rb
unilynx: Never coded in Ruby :-) Will change.

Formula/mariadb-connector-odbc.rb
fxcoudert: Great. Can you remove the `patch :DATA` and use the [upstream commit URL](https://github.com/MariaDB/mariadb-connector-odbc/commit/677f3c9d2ddb2100d113289f0314520c46ef5680.diff) in a `patch do` block?

Formula/mariadb-connector-odbc.rb
```diff
@@ -0,0 +1,60 @@
+class MariadbConnectorOdbc < Formula
+  desc "MariaDB Standardized database driver"
+  homepage "https://downloads.mariadb.org/connector-odbc/"
+  # MariaDB's download page doesn't let you download without going through the site, probably a cookie-based check
+  # The connector is packaged at http://archive.mariadb.org/connector-odbc-2.0.13/source/mariadb-connector-odbc-2.0.13-ga-src.tar.gz, but there's no https here
+  # So the next best choice seems to be github:
+  url "https://github.com/MariaDB/mariadb-connector-odbc/archive/2.0.13.tar.gz"
```
fxcoudert: It's fine to download from github if it's the official project, so you can remove the comment

fxcoudert (changes requested):

Formula/mariadb-connector-odbc.rb
unilynx: @fxcoudert unfortunately the patch I've sent them doesn't apply to the 2.0.13.tar.gz I'm downloading as base package. 

The patch as submitted includes an update to .gitingore but that file isn't in the 2.0.13.tar.gz due to https://github.com/MariaDB/mariadb-connector-odbc/blob/master/.gitattributes - so the patch fails with a 'No file to patch'

Not sure how to work around that.

fxcoudert: Please add a link to https://github.com/MariaDB/mariadb-connector-odbc/pull/13 above the patch, saying the patch can be removed for the next version.

Formula/mariadb-connector-odbc.rb
```diff
@@ -0,0 +1,57 @@
+class MariadbConnectorOdbc < Formula
+  desc "MariaDB Standardized database driver"
+  homepage "https://downloads.mariadb.org/connector-odbc/"
+  url "https://github.com/MariaDB/mariadb-connector-odbc/archive/2.0.13.tar.gz"
```
fxcoudert: There is a 3.0.0 release, can we use that?

unilynx: That's still marked as Alpha, it might be better to wait until upstream considers it stable ?

It would also require updating the mariadb-connector-c formula to 3 or higher, which is currently at 2.2.2. The current upstream stable version for that is at 2.3.2, which is still not sufficient to compile the 3.0 odbc connector.

So you'd be upgrading at least two packages to an alpha version.

fxcoudert:
Closing, as software is marked alpha and no recent movement.

Issue #10144, opened 2017-02-20
chezscheme fails to build on 10.11


ararslan:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [ ] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Bug reports:

Full Gist: https://gist.github.com/ararslan/21597dcac66032d8283f5944a6c1bbe6

The formula for Chez Scheme version 9.4 fails to build since `_clock_gettime` isn't found. My understanding is that this stems from my use of Xcode 8 on OS X 10.11, which defines certain symbols to be able to cross-compile for Sierra, but doesn't actually make them available on the local system. Cloning the Chez repository and building, manually setting `-Wl,-no_weak_imports`, works fine. Given that https://github.com/Homebrew/homebrew-core/issues/3727 has been implemented, I would have expected this to no longer be an issue. Isn't weak importing now disabled by default, at least on El Capitan, or is there something I'm missing?

Quick view of the failure:
```
[23:27:07] alex:~ $ brew install chezscheme
==> Using the sandbox
==> Downloading https://github.com/cisco/ChezScheme/archive/v9.4.tar.gz
Already downloaded: /Users/alex/Library/Caches/Homebrew/chezscheme-9.4.tar.gz
==> ./configure --installprefix=/usr/local/Cellar/chezscheme/9.4 --threads --installschemename=chez
==> make install
Last 15 lines from /Users/alex/Library/Logs/Homebrew/chezscheme/02.make:
dyld: Symbol not found: _clock_gettime
  Referenced from: /private/tmp/chezscheme-20170219-22209-4vcmnj/ChezScheme-9.4/ta6osx/s/../bin/ta6osx/scheme
  Expected in: /usr/lib/libSystem.B.dylib

/bin/sh: line 1: 23254 Done                    echo '(reset-handler abort)' '(base-exception-handler (lambda (c) (fresh-line) (display-condition c) (newline) (reset)))' '(keyboard-interrupt-handler (lambda () (display "interrupted---aborting\n") (reset)))' '(optimize-level 3)' '(debug-level 0)' '(generate-inspector-information #f)' '(subset-mode (quote system))' '(compile-file "cmacros.ss" "cmacros.so")'
[...]

PR #10159, opened 2017-02-20
chezscheme: fix the build on 10.11 with Xcode 8

ilovezfs:

-----

Fixes https://github.com/Homebrew/homebrew-core/issues/10144.

### DIFF
--- a/Formula/chezscheme.rb
+++ b/Formula/chezscheme.rb
@@ -12,6 +12,16 @@ class Chezscheme < Formula
   depends_on :x11 => :build
 
   def install
+    # dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
+    if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+      inreplace "c/stats.c" do |s|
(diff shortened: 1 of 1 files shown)


ararslan:
Awesome, thanks @ilovezfs! Would it be simpler to pass `LDFLAGS=-Wl,-no_weak_imports` in the call to `./configure`, or is something like this the brew-preferred solution?

ilovezfs:
@ararslan that works if a) Xcode 7 and CLT 7 are not in the picture at all (they are) and if b) the use of clock_gettime is guarded by a variable like HAVE_CLOCK_GETTIME, and some kind of configure check for its availability (it's not).

ilovezfs (commented):
needs to be reported upstream

ararslan:
Good to know. Thanks for the explanation and thanks again for the fix! You rock! 🎸

ararslan:
Reported upstream: https://github.com/cisco/ChezScheme/issues/146

ilovezfs:
Shipped! @ararslan thanks for taking care of the upstream report. I've added a link to it in the formula.

PR #10180, opened 2017-02-21
crystal-lang 0.21.0

asterite:

-----

Thank you as usual! 🙇

### DIFF
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -3,7 +3,6 @@ class CrystalLang < Formula
   homepage "https://crystal-lang.org/"
   url "https://github.com/crystal-lang/crystal/archive/0.21.0.tar.gz"
   sha256 "4dd01703f5304a0eda7f02fc362fba27ba069666097c0f921f8a3ee58808779c"
-  revision 2
   head "https://github.com/crystal-lang/crystal.git"
 
   bottle do
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -1,8 +1,8 @@
 class CrystalLang < Formula
   desc "Fast and statically typed, compiled language with Ruby-like syntax"
   homepage "https://crystal-lang.org/"
-  url "https://github.com/crystal-lang/crystal/archive/0.20.5.tar.gz"
-  sha256 "ee1e5948c6e662ccb1e62671cf2c91458775b559b23d74ab226dc2a2d23f7707"
+  url "https://github.com/crystal-lang/crystal/archive/0.21.0.tar.gz"
+  sha256 "4dd01703f5304a0eda7f02fc362fba27ba069666097c0f921f8a3ee58808779c"
   revision 2
   head "https://github.com/crystal-lang/crystal.git"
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
@asterite don't the libevent and bdw-gc blocks need to be updated, too?

asterite:
@ilovezfs How?

ilovezfs:
@asterite on inspection, it looks like the new bootstrap binary has still been built against the older libevent. Is there a reason you haven't updated it to be built against 2.1.8 like the release binary here in the bottle will be?

asterite:
The "new" bootstrap binary is not new, it's the old one from the previous version. It has already been built with the older libevent. I can't change it.

ilovezfs:
@asterite ah OK, that explains it. Are you saying the current version of crystal-cannot bootstrap itself?

asterite:
Yes, it can, with the older libevent. Well, I think you need the old libevent and link against the new libevent. But honestly I don't know how to do that :-(

ilovezfs:
@asterite presumably the bottle from https://github.com/Homebrew/homebrew-core/blob/master/Formula/crystal-lang.rb#L13 could be used, no?

asterite:
@ilovezfs Maybe. How can I change the formula to use that for boot?

ilovezfs:
@asterite you don't have to. This should be fine. I was just surprised to see that the bootstrap binary hadn't been updated since your CI already seems to test the builds against the latest libraries in Homebrew, so I assumed you were just producing a new boostrap binary each time, and that the newest one would be built against the current libevent.

asterite:
Ah, I see... No, we use the bootstrap binary from the previous build. In fact for other builds we have (centos, debian) we are still using the old libevent and libgc, because they work well. Updating them will probably involve lots of conflicts which right now I don't want to deal with. But I guess after this release we'll take a look at them so we are up to date :-)

ilovezfs:
@asterite cool. I assume ever since we updated libevent in Homebrew, and we rebuilt crystal against it, that you haven't noticed any problems resulting from that upgrade?

asterite:
Hmm... actually, I never updated crystal to the last formula, I was using the one before libevent was upgraded. But I just tried it and it seems to work :-)

I guess next releases will work if I build the current binary against libevent 2.1.8. I'll try to do that now, but if I run into conflicts I'll probably just go with the old versions for now.

ilovezfs:
@asterite OK great. I look forward to hearing the results!

asterite:
Yay, the build passed! 🎉 

I tried to build against the newer libevent but got an error when trying to compile it:

```
Error:

    bufferevent_openssl.c:66:10: fatal error: 'openssl/bio.h' file not found
#include <openssl/bio.h>
         ^
1 error generated.
make[1]: *** [libevent_openssl_la-bufferevent_openssl.lo] Error 1
make: *** [all] Error 2
```

Unfrotunately I don't know have time to look at it right now, so this release will be linked against the old libevent, but for the next release we'll make sure to upgrade to the latest versions.

ilovezfs:
@asterite you probably need to
```
 export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/local/opt/openssl/lib/pkgconfig
```

asterite:
@ilovezfs I got a different error, probably the openssl version is incorrect (as far as I know we are not linking against openssl, so maybe that's the issue):

```
    sample/le-proxy.c:33:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
         ^
1 error generated.
make[1]: *** [sample/le-proxy.o] Error 1
sample/openssl_hostname_validation.c:42:10: fatal error: 'openssl/x509v3.h' file not found
#include <openssl/x509v3.h>
         ^
1 error generated.
make[1]: *** [sample/openssl_hostname_validation.o] Error 1
sample/https-client.c:42:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
         ^
1 error generated.
make[1]: *** [sample/https-client.o] Error 1
make: *** [all] Error 2
```

But don't worry, I'll make sure that the versions are up to date for the next release.

Is anything else needed to merge this PR?

ilovezfs:
@asterite not worried at all! Also, I hope you have a chance to look into making the bootstrap binary statically built so we can avoid the vendoring and macho rewrites.

>Is anything else needed to merge this PR?

Nope. Also, just as an aside, this PR uncovered a `brew` bug! https://github.com/Homebrew/brew/pull/1754#issuecomment-281379947

ilovezfs:
Shipped! Thanks again @asterite :star2:

PR #10203, opened 2017-02-21
snake 1.0.0 (new formula)

hasahmed:
Just a simple snake game written in pure java.

- [✓] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [✓] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [✓] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [✓] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

### DIFF
--- a/Formula/snake.rb
+++ b/Formula/snake.rb
@@ -1,0 +1,16 @@
+class Snake < Formula
+  desc "simple snake game. Written in pure java"
+  homepage "http://silo.cs.indiana.edu:32903/snake/snakehome.html"
+  url "http://silo.cs.indiana.edu:32903/snake/snake.tar.gz"
+  version "1.0.0"
+  sha256 "9a078047dd9b553f458703afad98ebdcf27907382d05abf99900662de05178dd"
+  depends_on "ant" => :build
+  depends_on :java
+  def install
+    system "ant", "compile", "jar"
+    bin.install "snake", "snake.jar"
+  end
+  test do
+    system "java", "Snake", "--version"
+  end
+end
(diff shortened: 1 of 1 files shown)


Formula/snake.rb
```diff
@@ -0,0 +1,17 @@
...
+  desc "simple snake game. Written in pure java"
+  homepage "http://silo.cs.indiana.edu:32903/snake/snakehome.html"
+  url "http://silo.cs.indiana.edu:32903/snake/snake.tar.gz"
+  version "1.0.0"
+  sha256 "0ce0cf8cd1cc77c67be06ea66f11909f43d5557e7c74595a37b6af85c9788abe"
+  depends_on "ant" => :build
+  depends_on :java
+  def install
+    system "ant", "compile", "jar"
+    system "mv " "snake.jar " "#{prefix}/"
```
rakkesh: Use the `mv` Ruby method instead of `system "mv `

Formula/snake.rb
```diff
@@ -0,0 +1,17 @@
...
+    bin.install "snake"
+  end
+  test do
+    system "java", "Snake", "--version"
```
rakkesh: Error from CI - 
```
java Snake --version
Error: Could not find or load main class Snake
Error: snake: failed
```

fxcoudert:
Thanks for your submission. Unfortunately, we can't accept this as-is: it appears to be personal project, with not enough of a user base or community to ensure that it will be useful to a large number of Homebrew users, or that it will be maintained in the long term (both upstream and here on Homebrew). We also generally prefer that packages be submitted by someone other than their primary maintainer — as it gives us an indication that a community will be keeping the package alive. In the mean time, you can always host this formula in a [personal tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

Sorry, and thanks again!

PR #10204, opened 2017-02-21
spandsp: switch to http for homepage and url

ilovezfs:
upstream SSL certificate expired 8 Jan 2017


ilovezfs:
reported to upstream email contact

PR #10207, opened 2017-02-21
chromaprint: remove examples.

MikeMcQuaid:
Remove examples because they introduce a recursive (optional) dependency from ffmpeg to chromaprint.


MikeMcQuaid:
Passing on this for now because @ilovezfs thought another approach would be best.

PR #10217, opened 2017-02-22
Bazel 0.3.2.

benjyw:
This is a copy of the main Bazel formula as it was at 0.3.2. The only changes are reordering the `url`, `sha256` and `head`, in order to pass `brew audit --strict`.

This formula is needed because the main Bazel formula is currently up to 0.4.4, but various repos (or past commits in those repos) will not build with 0.4.4 and require this earlier version.

### DIFF
--- a/Formula/bazel@0.3.2.rb
+++ b/Formula/bazel@0.3.2.rb
@@ -5,16 +5,9 @@ class BazelAT032 < Formula
   sha256 "9692ac3318a40e8a0530f68bbfc473ae5f6a4a5c0fe08d2f88612ca4d40ba54a"
   head "https://github.com/bazelbuild/bazel.git"
 
-  bottle do
-    cellar :any_skip_relocation
-    rebuild 1
-    sha256 "8c28efdb500103fc2fe69f5d32b2b433e00345c55c5003cdc53f76e9aec3a22e" => :sierra
-    sha256 "9baa84e3f84d5060d39166ff682f7280e93dd173713b83dd5c95c23dda78c942" => :el_capitan
-    sha256 "526bcf3e24b646c1f1648d47868654a16e8f566dcf1575034b039f7e17aa91d4" => :yosemite
-  end
-
   depends_on :java => "1.8+"
   depends_on :macos => :yosemite
+  keg_only :versioned_formula
 
   def install
     ENV["EMBED_LABEL"] = "#{version}-homebrew"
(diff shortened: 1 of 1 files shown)


Formula/bazel@0.3.2.rb
```diff
@@ -0,0 +1,56 @@
...
+  sha256 "9692ac3318a40e8a0530f68bbfc473ae5f6a4a5c0fe08d2f88612ca4d40ba54a"
+  head "https://github.com/bazelbuild/bazel.git"
+
+  bottle do
+    cellar :any_skip_relocation
+    rebuild 1
+    sha256 "8c28efdb500103fc2fe69f5d32b2b433e00345c55c5003cdc53f76e9aec3a22e" => :sierra
+    sha256 "9baa84e3f84d5060d39166ff682f7280e93dd173713b83dd5c95c23dda78c942" => :el_capitan
+    sha256 "526bcf3e24b646c1f1648d47868654a16e8f566dcf1575034b039f7e17aa91d4" => :yosemite
+  end
```
fxcoudert: The `bottle do` section will be introduced when the PR is merged

Formula/bazel@0.3.2.rb
```diff
@@ -0,0 +1,56 @@
...
+
```
fxcoudert: The formula would conflict with unversioned `bazel`, and thus need `keg_only :versioned_formula`

fxcoudert (changes requested):

fxcoudert:
Do we want to maintain a versioned formula? Nothing in homebrew depends on `bazel` currently, so it is hard to estimate the need for this version specifically…

MikeMcQuaid:
> various repos (or past commits in those repos) will not build with 0.4.4 and require this earlier versio

Can you list these, please? Thanks!

benjyw (commented):
The motivating example for me is Kythe (https://github.com/google/kythe).  The most recent release is v0.0.26, which requires Bazel 0.3.2 and doesn't build under 0.4 (current HEAD in that repo does build with 0.4, but there is no release that does).

Likewise, it appears that versions of TensorFlow (https://github.com/tensorflow/tensorflow/) prior to the recent v1.0.0 release (e.g., v0.12.0) require Bazel 0.3.2.

There aren't many repos out there using Bazel, the only examples I know of are Google open-source projects.

If you think this isn't useful enough to justify being in core, I can create my own tap for bazel-related formulae.

fxcoudert:
A personal tap is probably better. The formula otherwise looks good, with just one minor style issue flagged up by audit:

  * `keg_only` (line 10) should be put before `depends_on` (line 9)

benjyw:
Sounds good, I've put a tap up here:

https://github.com/benjyw/homebrew-bazel

Will close this pull request. Thanks!

PR #10239, opened 2017-02-22
python@3.5 3.5.2 (new formula)

davidparsson:

-----

Heavily based on an old revision (510018d3) of `Formula/python3.rb`, when 3.5 was the version being installed. What's changed is:

- File name changed to `python@3.5.rb`
- Class name changed to `PythonAT35`
-  `revision` and `head` statements as well as `devel` block removed
- `:universal` option and `if build.universal?` block removed

When it comes to other packages being dependent on this, is the current name `python@3.5` or `python3@3.5` preferable?

Is there anything else to consider when adding a new version of an existing package?

This is my first contribution.

### DIFF
--- a/Formula/python@3.5.rb
+++ b/Formula/python@3.5.rb
@@ -1,0 +1,410 @@
+class PythonAT35 < Formula
+  desc "Interpreted, interactive, object-oriented programming language"
+  homepage "https://www.python.org/"
+
+  stable do
+    url "https://www.python.org/ftp/python/3.5.2/Python-3.5.2.tar.xz"
+    sha256 "0010f56100b9b74259ebcd5d4b295a32324b58b517403a10d1a2aa7cb22bca40"
+
+    # Patch for pyport.h macro issue
+    # https://bugs.python.org/issue10910
+    # https://trac.macports.org/ticket/44288
+    patch do
+      url "https://bugs.python.org/file30805/issue10910-workaround.txt"
+      sha256 "c075353337f9ff3ccf8091693d278782fcdff62c113245d8de43c5c7acc57daf"
+    end
+  end
+
+  bottle do
+    sha256 "4a43600ceb2875c13200ace82fea1a9a119973b831f5503fa57cf8db44fbb155" => :sierra
+    sha256 "301854e1a52fd577f84015eb5ef07e0e369c794b8894fb6850ae84d9391f740e" => :el_capitan
+    sha256 "075f5e63188dfcd5a13ef1f9658b26aa2d8d943fbc5c5c50e05fc973bc38f7c1" => :yosemite
+  end
+
+  option "with-tcl-tk", "Use Homebrew's Tk instead of macOS Tk (has optional Cocoa and threads support)"
+  option "with-quicktest", "Run `make quicktest` after the build"
+  option "with-sphinx-doc", "Build HTML documentation"
+
+  deprecated_option "quicktest" => "with-quicktest"
+  deprecated_option "with-brewed-tk" => "with-tcl-tk"
+
+  depends_on "pkg-config" => :build
+  depends_on "readline" => :recommended
+  depends_on "sqlite" => :recommended
+  depends_on "gdbm" => :recommended
+  depends_on "openssl"
+  depends_on "xz" => :recommended # for the lzma module added in 3.3
+  depends_on "homebrew/dupes/tcl-tk" => :optional
+  depends_on :x11 if build.with?("tcl-tk") && Tab.for_name("homebrew/dupes/tcl-tk").with?("x11")
+  depends_on "sphinx-doc" => [:build, :optional]
+
+  skip_clean "bin/pip3", "bin/pip-3.4", "bin/pip-3.5"
+  skip_clean "bin/easy_install3", "bin/easy_install-3.4", "bin/easy_install-3.5"
+
+  resource "setuptools" do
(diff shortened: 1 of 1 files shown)


ilovezfs:
>When it comes to other packages being dependent on this, is the current name python@3.5 or python3@3.5 preferable?

python@3.5

davidparsson:
If it helps, the output of `git diff 510018d:Formula/python3.rb add/python@3.5:Formula/python@3.5.rb` is:

```diff
diff --git a/510018d:Formula/python3.rb b/add/python@3.5:Formula/python@3.5.rb
index 5ae99a5a1..3941e0e66 100644
--- a/510018d:Formula/python3.rb
+++ b/add/python@3.5:Formula/python@3.5.rb
@@ -1,9 +1,6 @@
-class Python3 < Formula
+class PythonAT35 < Formula
   desc "Interpreted, interactive, object-oriented programming language"
   homepage "https://www.python.org/"
-  revision 3
-
-  head "https://hg.python.org/cpython", :using => :hg
 
   stable do
     url "https://www.python.org/ftp/python/3.5.2/Python-3.5.2.tar.xz"
@@ -24,12 +21,6 @@ class Python3 < Formula
     sha256 "075f5e63188dfcd5a13ef1f9658b26aa2d8d943fbc5c5c50e05fc973bc38f7c1" => :yosemite
   end
 
-  devel do
-    url "https://www.python.org/ftp/python/3.6.0/Python-3.6.0rc2.tar.xz"
-    sha256 "ed07453330af6677d0a670e187344922a67208fc6bdaea9fade66dc08bf763a4"
-  end
-
-  option :universal
   option "with-tcl-tk", "Use Homebrew's Tk instead of macOS Tk (has optional Cocoa and threads support)"
   option "with-quicktest", "Run `make quicktest` after the build"
   option "with-sphinx-doc", "Build HTML documentation"
@@ -145,11 +136,6 @@ class Python3 < Formula
       s.gsub! "/usr/local/ssl", Formula["openssl"].opt_prefix
     end
 
-    if build.universal?
-      ENV.universal_binary
-      args << "--enable-universalsdk" << "--with-universal-archs=intel"
-    end
-
     if build.with? "sqlite"
       inreplace "setup.py" do |s|
         s.gsub! "sqlite_setup_debug = False", "sqlite_setup_debug = True"
```

MikeMcQuaid:
Can you explain why you need this formula? We're trying to think about which versioned formulae we add or not. Thanks!

davidparsson:
@MikeMcQuaid, sure! I'm a developer and our target platform is Python 3.5, so for me having this would allow me to use the same version (with the same system libraries) as the target platform.

The same reasoning could of course be applied to every other released Python version.

MikeMcQuaid:
@davidparsson Have you investigated `pyenv` at all?

tdsmith:
Thanks for the PR. I've been uncomfortable supporting old versions of Python in Homebrew before, in part because without special handling, installed scripts will step on each other in confusing ways. pyenv handles this very well and delivers a much better experience than Homebrew alone can, so I'd recommend that. 

3.5.3 is the newest 3.5 release.

davidparsson:
The problem I've had with`pyenv` is that I've had a hard time installing other Python libraries using Homebrew, specifically `pyqt5`, and get that up and running properly. But if you think this would introduce more problems than it would solve, then this is of course not the way to go.

MikeMcQuaid:
> The problem I've had withpyenv is that I've had a hard time installing other Python libraries using Homebrew, specifically pyqt5, and get that up and running properly.

Note this same problem with persist with this formula, FWIW.

davidparsson:
> Note this same problem with persist with this formula, FWIW.
I was afraid that was the case. Then this would be of no benefit to me, so I'll close this.

Thanks a lot for your helpful, friendly, and timely responses!

tdsmith:
One thing that you can do is to create a tap with this formula and a
version of PyQt5 that depends on it.

lopopolo:
Why no python@3.5 when there is ruby@2.3 and node@6?

MikeMcQuaid:
@lopopolo because no-one has given a good reason yet.

tdsmith:
That is a good reason, but nothing in Homebrew uses Django 1.10. Pyenv
provides a better experience than Homebrew can if you need a specific or
multiple Python version/s.

lopopolo:
What in homebrew depends on ruby 1.8.7?

Could the same not be said of nodenv, nvm, rbenv, and rvm?

tdsmith:
Possibly! Consistency could be nice but my own personal reluctance is that
I'd be de facto responsible for supporting and integrating the different
Python formulas and handling tickets for pip doing surprising things, and I
don't have much motivation for that when pyenv already does this, better.

tdsmith:
Along those lines, though, I'd be happy to point people towards a tap with
old Python versions, maybe in the mold of the deadsnakes ppa, and if it
proves to be useful maybe we'll adopt it.

lopopolo:
Thank you for taking the time to respond to me. I have been using pyenv and will continue to do so as it seems to be the best option.

MikeMcQuaid:
> I have been using pyenv and will continue to do so as it seems to be the best option.

We agree. Not sure I really understand your comments given that statement...

lopopolo:
> Not sure I really understand your comments given that statement...

I used pyenv out of necessity and all things being equal would prefer to install python@3.5 via Homebrew. I thought it was clear that I learned something from this conversation about the tradeoffs involved that NOW makes it clear to me that pyenv is the best option.

PR #10273, opened 2017-02-23
tensorflow_c 1.0.0 (new formula)

kali:

-----

### DIFF
--- a/Formula/libtensorflow.rb
+++ b/Formula/libtensorflow.rb
@@ -1,22 +1,21 @@
 class Libtensorflow < Formula
   desc "C interface for Google's OS library for Machine Intelligence"
   homepage "https://www.tensorflow.org/"
-  url "https://github.com/tensorflow/tensorflow.git", :tag => "v1.0.0",
-    :revision => "07bb8ea2379bd459832b23951fb20ec47f3fdbd4"
+  url "https://github.com/tensorflow/tensorflow/archive/v1.0.0.tar.gz"
+  sha256 "db8b3b8f4134b7c9c1b4165492ad5d5bb78889fcd99ffdffc325e97da3e8c677"
 
   depends_on "bazel" => :build
-  depends_on "pkg-config" => :run
 
   def install
-    ENV["PYTHON_BIN_PATH"]=`which python`.strip
-    ENV["CC_OPT_FLAGS"]="-march=native"
-    ENV["TF_NEED_JEMALLOC"]="1"
-    ENV["TF_NEED_GCP"]="0"
-    ENV["TF_NEED_HDFS"]="0"
-    ENV["TF_ENABLE_XLA"]="0"
-    ENV["USE_DEFAULT_PYTHON_LIB_PATH"]="1"
-    ENV["TF_NEED_OPENCL"]="0"
-    ENV["TF_NEED_CUDA"]="0"
+    ENV["PYTHON_BIN_PATH"] = which("python").to_s
+    ENV["CC_OPT_FLAGS"] = "-march=native"
+    ENV["TF_NEED_JEMALLOC"] = "1"
+    ENV["TF_NEED_GCP"] = "0"
+    ENV["TF_NEED_HDFS"] = "0"
+    ENV["TF_ENABLE_XLA"] = "0"
+    ENV["USE_DEFAULT_PYTHON_LIB_PATH"] = "1"
+    ENV["TF_NEED_OPENCL"] = "0"
+    ENV["TF_NEED_CUDA"] = "0"
     system "./configure"
 
     system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow.so"
@@ -32,7 +31,6 @@ class Libtensorflow < Formula
   end
 
   test do
-    # test a call on TF_Version()
     (testpath/"test.c").write <<-EOS.undent
       #include <stdio.h>
       #include <tensorflow/c/c_api.h>
--- a/Formula/libtensorflow.rb
+++ b/Formula/libtensorflow.rb
@@ -1,4 +1,4 @@
-class TensorflowC < Formula
+class Libtensorflow < Formula
   desc "C interface for Google's OS library for Machine Intelligence"
   homepage "https://www.tensorflow.org/"
   url "https://github.com/tensorflow/tensorflow.git", :tag => "v1.0.0",
@@ -19,14 +19,14 @@ class TensorflowC < Formula
     ENV["TF_NEED_CUDA"]="0"
     system "./configure"
 
-    system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow_c.so"
-    lib.install "bazel-bin/tensorflow/libtensorflow_c.so"
+    system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow.so"
+    lib.install "bazel-bin/tensorflow/libtensorflow.so"
     (include/"tensorflow/c").install "tensorflow/c/c_api.h"
-    (lib/"pkgconfig/tensorflow_c.pc").write <<-EOS.undent
-      Name: tensorflow_c
-      Description: Tensorflow c lib
+    (lib/"pkgconfig/tensorflow.pc").write <<-EOS.undent
+      Name: tensorflow
+      Description: Tensorflow library
       Version: #{version}
-      Libs: -L#{lib} -ltensorflow_c
+      Libs: -L#{lib} -ltensorflow
       Cflags: -I#{include}
     EOS
   end
@@ -40,7 +40,7 @@ class TensorflowC < Formula
         printf("%s", TF_Version());
       }
     EOS
-    system ENV.cc, "-L#{lib}", "-ltensorflow_c", "-o", "test_tf", "test.c"
+    system ENV.cc, "-L#{lib}", "-ltensorflow", "-o", "test_tf", "test.c"
     assert_equal version, shell_output("./test_tf")
   end
 end
--- a/Formula/tensorflow_c.rb
+++ b/Formula/tensorflow_c.rb
@@ -1,8 +1,8 @@
 class TensorflowC < Formula
   desc "C interface for Google's OS library for Machine Intelligence"
   homepage "https://www.tensorflow.org/"
-  url "https://github.com/tensorflow/tensorflow", :using => :git,
-    :tag => "v1.0.0", :revision => "07bb8ea2379bd459832b23951fb20ec47f3fdbd4"
+  url "https://github.com/tensorflow/tensorflow.git", :tag => "v1.0.0",
+    :revision => "07bb8ea2379bd459832b23951fb20ec47f3fdbd4"
 
   depends_on "bazel" => :build
   depends_on "pkg-config" => :run
@@ -11,32 +11,26 @@ class TensorflowC < Formula
     system 'echo "\n\n\n\n\n\n\n\n\n" | ./configure'
     system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow_c.so"
     lib.install "bazel-bin/tensorflow/libtensorflow_c.so"
-    cp "tensorflow/c/c_api.h", "tensorflow_c_api.h"
-    include.install "tensorflow_c_api.h"
-    pc = <<-EOF.gsub(/^\s+/, "")
+    include.install "tensorflow/c/c_api.h" => "tensorflow_c_api.h"
+    (lib/"pkgconfig/tensorflow_c.pc").write <<-EOS.undent
       Name: tensorflow_c
       Description: Tensorflow c lib
       Version: #{version}
       Libs: -L#{lib} -ltensorflow_c
       Cflags: -I#{include}
-    EOF
-    mkdir_p(lib/"pkgconfig")
-    File.open(lib/"pkgconfig/tensorflow_c.pc", "w") { |f| f.write(pc) }
+    EOS
   end
 
   test do
     # test a call on TF_Version(), checking .h, .so, pkg-config setup.
-    File.open("test.c", "w") do |f|
-      f.write(<<-EOF.gsub(/^\s+/, ""))
+    (testpath/"test.c").write <<-EOS.undent
       #include <stdio.h>
       #include <tensorflow_c_api.h>
       int main() {
-        printf("%s\\n", TF_Version());
+        printf("%s", TF_Version());
       }
-      EOF
-    end
+    EOS
     system "sh", "-c", "gcc `pkg-config --libs --cflags tensorflow_c` -o test_tf test.c"
-    found_version = `./test_tf`.strip
-    assert found_version == version
+    assert_equal version, shell_output("test_tf")
   end
 end
--- a/Formula/tensorflow_c.rb
+++ b/Formula/tensorflow_c.rb
@@ -31,6 +31,6 @@ class TensorflowC < Formula
       }
     EOS
     system "sh", "-c", "gcc `pkg-config --libs --cflags tensorflow_c` -o test_tf test.c"
-    assert_equal version, shell_output("test_tf")
+    assert_equal version, shell_output("./test_tf")
   end
 end
--- a/Formula/tensorflow_c.rb
+++ b/Formula/tensorflow_c.rb
@@ -21,7 +21,7 @@ class TensorflowC < Formula
 
     system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow_c.so"
     lib.install "bazel-bin/tensorflow/libtensorflow_c.so"
-    include.install "tensorflow/c/c_api.h" => "tensorflow_c_api.h"
+    (include/"tensorflow/c").install "tensorflow/c/c_api.h"
     (lib/"pkgconfig/tensorflow_c.pc").write <<-EOS.undent
       Name: tensorflow_c
       Description: Tensorflow c lib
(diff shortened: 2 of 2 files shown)


Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
+class TensorflowC < Formula
+  desc "C interface for Google's OS library for Machine Intelligence"
+  homepage "https://www.tensorflow.org/"
+  url "https://github.com/tensorflow/tensorflow", :using => :git,
```
bfontaine: Please add `.git` at the end of the URL and remove `:using => :git` (it’s already the default).

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+    :tag => "v1.0.0", :revision => "07bb8ea2379bd459832b23951fb20ec47f3fdbd4"
+
+  depends_on "bazel" => :build
+  depends_on "pkg-config" => :run
+
+  def install
+    system 'echo "\n\n\n\n\n\n\n\n\n" | ./configure'
```
bfontaine: Why is this needed?

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+    system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow_c.so"
+    lib.install "bazel-bin/tensorflow/libtensorflow_c.so"
+    cp "tensorflow/c/c_api.h", "tensorflow_c_api.h"
+    include.install "tensorflow_c_api.h"
```
bfontaine: Use `include.install "tensorflow/c/c_api.h" => "tensorflow_c_api.h"` instead.

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+    pc = <<-EOF.gsub(/^\s+/, "")
+      Name: tensorflow_c
+      Description: Tensorflow c lib
+      Version: #{version}
+      Libs: -L#{lib} -ltensorflow_c
+      Cflags: -I#{include}
+    EOF
+    mkdir_p(lib/"pkgconfig")
+    File.open(lib/"pkgconfig/tensorflow_c.pc", "w") { |f| f.write(pc) }
```
bfontaine: Use `(lib/"pkgconfig/tensorflow_c.pc").write pc` instead. You don’t even need to make a variable for it.

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+  depends_on "bazel" => :build
+  depends_on "pkg-config" => :run
+
+  def install
+    system 'echo "\n\n\n\n\n\n\n\n\n" | ./configure'
+    system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow_c.so"
+    lib.install "bazel-bin/tensorflow/libtensorflow_c.so"
+    cp "tensorflow/c/c_api.h", "tensorflow_c_api.h"
+    include.install "tensorflow_c_api.h"
+    pc = <<-EOF.gsub(/^\s+/, "")
```
bfontaine: Use `EOS.undent` instead (and `EOS` instead of `EOF` at the end).

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+      Name: tensorflow_c
+      Description: Tensorflow c lib
+      Version: #{version}
+      Libs: -L#{lib} -ltensorflow_c
+      Cflags: -I#{include}
+    EOF
+    mkdir_p(lib/"pkgconfig")
```
bfontaine: You don’t need this.

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+    File.open(lib/"pkgconfig/tensorflow_c.pc", "w") { |f| f.write(pc) }
+  end
+
+  test do
+    # test a call on TF_Version(), checking .h, .so, pkg-config setup.
+    File.open("test.c", "w") do |f|
```
bfontaine: Use `(testpath/"test.c").write <<-EOS.undent`.

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+      #include <stdio.h>
+      #include <tensorflow_c_api.h>
+      int main() {
+        printf("%s\\n", TF_Version());
+      }
+      EOF
+    end
+    system "sh", "-c", "gcc `pkg-config --libs --cflags tensorflow_c` -o test_tf test.c"
+    found_version = `./test_tf`.strip
+    assert found_version == version
```
bfontaine: Use `assert_equal version, shell_output("test_tf")`. You don’t need the `.strip` if you remove `\\n` in the C code.

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,42 @@
...
+    File.open("test.c", "w") do |f|
+      f.write(<<-EOF.gsub(/^\s+/, ""))
+      #include <stdio.h>
+      #include <tensorflow_c_api.h>
+      int main() {
+        printf("%s\\n", TF_Version());
+      }
+      EOF
+    end
+    system "sh", "-c", "gcc `pkg-config --libs --cflags tensorflow_c` -o test_tf test.c"
```
bfontaine: Use the formula paths instead; formulae should be testable without being linked.

bfontaine (changes requested):

Formula/tensorflow_c.rb
kali: I wanted to check the generated .pc was fine. If I understand correctly, you'd prefer me to replace the pkg-config call by something like -L#{lib} and so on ?

Formula/tensorflow_c.rb
kali: This one is very sad. Tensorflow `./configure` is not an autotools like thing. It will expect you to hit return 9 times to accept to default to 9 questions. There is not provision for non interactive run.

Alternatively, I could do a `yes ''| ./configure` but `yes` would fail systematically with a borken pipe when configure ends and it would get tricky to detect a ./configure fail.

If you have a better idea, I'm all hears.

MikeMcQuaid: @kali I'd advise submitting a patch upstream (or at least an issue) that fixes this problem. This is particularly hostile to any package manager packaging this software as-is.

Formula/tensorflow_c.rb
bfontaine: Yes, thanks.

Formula/tensorflow_c.rb
kali: ok, it is possible to pass environment variables matching the default to get the configure script running with no input. Still very fragile, though.

fxcoudert:
Test is failing with:

```
==> brew test tensorflow_c --verbose
==> FAILED
Testing tensorflow_c
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20170224-99173-14yz7vt.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/tensorflow_c.rb --verbose
Error: tensorflow_c: failed
Calling ENV.gcc is deprecated!
Use ENV.gcc_4_2 instead.
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/tensorflow_c.rb:48:in `block in <class:TensorflowC>'
```

It should probably be `ENV.cc` instead of `ENV.gcc`.

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,46 @@
...
+    ENV["TF_NEED_JEMALLOC"]="1"
+    ENV["TF_NEED_GCP"]="0"
+    ENV["TF_NEED_HDFS"]="0"
+    ENV["TF_ENABLE_XLA"]="0"
+    ENV["USE_DEFAULT_PYTHON_LIB_PATH"]="1"
+    ENV["TF_NEED_OPENCL"]="0"
+    ENV["TF_NEED_CUDA"]="0"
+    system "./configure"
+
+    system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow_c.so"
```
asimshankar: Could you use `tensorflow:libtensorflow.so` instead?
That is the target that is used as part of the binary releases of the TensorFlow C library.

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,46 @@
+class TensorflowC < Formula
+  desc "C interface for Google's OS library for Machine Intelligence"
+  homepage "https://www.tensorflow.org/"
+  url "https://github.com/tensorflow/tensorflow.git", :tag => "v1.0.0",
+    :revision => "07bb8ea2379bd459832b23951fb20ec47f3fdbd4"
+
+  depends_on "bazel" => :build
+  depends_on "pkg-config" => :run
+
+  def install
```
asimshankar: Can the release tarballs of the TensorFlow C library be used to avoid building from source?

That is recommended for building language bindings for TensorFlow (for example, the [Go bindings](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/go) and hopefully the Rust bindings will switch to using this binary target when possible too)

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,46 @@
...
+    ENV["PYTHON_BIN_PATH"]=`which python`.strip
+    ENV["CC_OPT_FLAGS"]="-march=native"
+    ENV["TF_NEED_JEMALLOC"]="1"
+    ENV["TF_NEED_GCP"]="0"
+    ENV["TF_NEED_HDFS"]="0"
+    ENV["TF_ENABLE_XLA"]="0"
+    ENV["USE_DEFAULT_PYTHON_LIB_PATH"]="1"
+    ENV["TF_NEED_OPENCL"]="0"
+    ENV["TF_NEED_CUDA"]="0"
```
asimshankar: Would it make sense for these to be options, so that GPU support etc. can be enabled with flags like `--with-cuda` or something to the `brew install` command?

asimshankar (commented):
A TensorFlow team member chiming in here.

Firstly, this is great! Thanks for doing this, adding a brew formula for the TensorFlow C library has been on our TODO list for a while. I was hoping to get to it over the next couple of weeks.

A few comments below. Also, would `libtensorflow` be a more appropriate name for this formula?

Thanks much,

-- Asim

Formula/tensorflow_c.rb
MikeMcQuaid: No, sorry. Homebrew requires building from source.

Formula/tensorflow_c.rb
MikeMcQuaid: We'd rather minimise our number of options until requested by users.

Formula/tensorflow_c.rb
MikeMcQuaid: It seems like `libtensorflow.dylib` should be used as this is macOS where `.so` indicates a plugin rather than library.

asimshankar: Oh, my comment was about the bazel targets and not the filename.
I was suggesting that the build command use the bazel target [`//tensorflow:libtensorflow.so`](https://github.com/tensorflow/tensorflow/blob/e75c0fc/tensorflow/BUILD#L300) instead of [`//tensorflow:libtensorflow_c.so`](https://github.com/tensorflow/tensorflow/blob/e75c0fc/tensorflow/BUILD#L309) (which is what is being done right now)

Formula/tensorflow_c.rb
```diff
@@ -0,0 +1,46 @@
...
+    system "./configure"
+
+    system "bazel", "build", "--compilation_mode=opt", "--copt=-march=native", "tensorflow:libtensorflow_c.so"
+    lib.install "bazel-bin/tensorflow/libtensorflow_c.so"
+    include.install "tensorflow/c/c_api.h" => "tensorflow_c_api.h"
```
asimshankar: Is the renaming necessary?
Can the include be installed in `<include_dir>/tensorflow/c/c_api.h` to avoid the rename?

My concern with renaming the file is that unless the same rename is followed in all other forms of distribution (such as on other platforms/OSes), this rename makes client code less portable.

MikeMcQuaid: Agreed.

Formula/tensorflow_c.rb
MikeMcQuaid: > Oh, my comment was about the bazel targets and not the filename.

Gotcha 👍

kali:
@asimshankar Thanks for your interest, just wanted to make sure you have also seen the rest of our efforts/struggle over there: https://github.com/snipsco/tensorflow-build and there https://medium.com/snips-ai/how-we-made-tensorflow-run-on-a-raspberry-pi-using-rust-7478f7a31329#.kgvgxpech . Feedback welcome :)

Formula/tensorflow_c.rb
kali: @asimshankar I'm not sure what you recommend here, actually. Are these two libraries the same thing ? I know that the Rust bindings, for instance, looks for `tensorflow_c` in pkg-config. Unless I'm confused, building `//tensorflow:libtensorflow.so` with bazel does not produce `libtensorflow_c.so` .

asimshankar: @kali: Yes, these two libraries are the same thing and the fact that there are two blaze targets that do exactly the same thing was an accident. So the idea is to suggest usage of `llibtensorflow.so` and remove `libtensorflow_c.so`.

I recently updated the Rust bindings to use `libtensorflow.so` instead of `libtensorflow_c.so` in https://github.com/tensorflow/rust/pull/56 and https://github.com/tensorflow/rust/pull/60,  and then plan to remove the `libtensorflow_c.so` target completely.

Formula/libtensorflow.rb
```diff
@@ -0,0 +1,46 @@
...
+  def install
+    ENV["PYTHON_BIN_PATH"]=`which python`.strip
+    ENV["CC_OPT_FLAGS"]="-march=native"
+    ENV["TF_NEED_JEMALLOC"]="1"
+    ENV["TF_NEED_GCP"]="0"
+    ENV["TF_NEED_HDFS"]="0"
+    ENV["TF_ENABLE_XLA"]="0"
+    ENV["USE_DEFAULT_PYTHON_LIB_PATH"]="1"
+    ENV["TF_NEED_OPENCL"]="0"
+    ENV["TF_NEED_CUDA"]="0"
```
bfontaine: Please leave a space before and after `=`. You can also probably use `which("python").to_s` instead of shelling out with `` `which python`.strip ``.

Formula/libtensorflow.rb
```diff
@@ -0,0 +1,46 @@
...
+      Name: tensorflow
+      Description: Tensorflow library
+      Version: #{version}
+      Libs: -L#{lib} -ltensorflow
+      Cflags: -I#{include}
+    EOS
+  end
+
+  test do
+    # test a call on TF_Version()
```
bfontaine: You can remove this comment.

Formula/libtensorflow.rb
```diff
@@ -0,0 +1,46 @@
+class Libtensorflow < Formula
+  desc "C interface for Google's OS library for Machine Intelligence"
+  homepage "https://www.tensorflow.org/"
+  url "https://github.com/tensorflow/tensorflow.git", :tag => "v1.0.0",
+    :revision => "07bb8ea2379bd459832b23951fb20ec47f3fdbd4"
+
+  depends_on "bazel" => :build
+  depends_on "pkg-config" => :run
```
bfontaine: Is it really needed?

Formula/libtensorflow.rb
```diff
@@ -0,0 +1,46 @@
+class Libtensorflow < Formula
+  desc "C interface for Google's OS library for Machine Intelligence"
+  homepage "https://www.tensorflow.org/"
+  url "https://github.com/tensorflow/tensorflow.git", :tag => "v1.0.0",
+    :revision => "07bb8ea2379bd459832b23951fb20ec47f3fdbd4"
```
bfontaine: Can’t we build from the tarball?

bfontaine (changes requested):

Formula/libtensorflow.rb
kali: I could not find a tgz. The "install from source" page pulls from github. @asimshankar, is there something I'm missing ?

MikeMcQuaid: https://github.com/tensorflow/tensorflow/archive/v1.0.0.tar.gz from https://github.com/tensorflow/tensorflow/releases/tag/v1.0.0

kali: thx.

bfontaine (approved):

kali:
Hey there... could we move this forward and merge it ? Is there something more I can / should do ?

fxcoudert:
Thanks @kali for your contribution!

carlthome:
Newbie brew user here!

1. I thought `brew install tensorflow` would get the TensorFlow Python API, not the C API. IMO it would be good to require `brew install libtensorflow` as that's typically what the non-Python stuff is called in TensorFlow's docs.
2. `brew install tensorflow` "successfully" installed the formula despite my system not having Bazel installed.
```sh
macbook:~ carl$ brew install tensorflow
Warning: libtensorflow 1.6.0 is already installed
macbook:~ carl$ bazel
-bash: bazel: command not found
macbook:~ carl$ brew install bazel
bazel: Java 1.8 is required to install this formula.
JavaRequirement unsatisfied!
You can install with Homebrew-Cask:
 brew cask install caskroom/versions/java8
You can download from:
 https://www.oracle.com/technetwork/java/javase/downloads/index.html
```
Shouldn't the install halt on https://github.com/Homebrew/homebrew-core/blob/master/Formula/libtensorflow.rb#L37 or am I missing something obvious?

bfontaine:
@carlthome you should open a separate issue.

PR #10303, opened 2017-02-24
pygobject3: add python3 support

actionless:
without this option `Language::Python.each_python` will never return `python3` for this build

-----

### DIFF
--- a/Formula/pygobject3.rb
+++ b/Formula/pygobject3.rb
@@ -13,6 +13,7 @@ class Pygobject3 < Formula
 
   option :universal
   option "without-python", "Build without python2 support"
+  option "without-python3", "Build without python3 support"
 
   depends_on "pkg-config" => :build
   depends_on "libffi" => :optional


actionless:
`audit --strict` is not passing neither before nor after my change (because of `:universal`)

DomT4:
> without this option Language::Python.each_python will never return python3 for this build

Are you sure? It should if you do `brew install pygobject3 --with-python3` no?

actionless:
@DomT4 it'll skip then iteration here anyway: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/language/python.rb#L19

but i can imagine this being a bug in brew itself, not in the formula -- i am not familiar with homebrew so don't know how should it handle the default value if option is not presented in a formula

DomT4:
The option is present in the formula though. Optional dependencies without an explicit `option` are treated as options and assigned a default option text:
```
==> Options
--universal
	Build a universal binary
--with-libffi
	Build with libffi support
--with-python3
	Build with python3 support
--without-python
	Build without python2 support
```

ilovezfs:
```
iMac-TMP:~ joe$ brew instal --with-python3 pygobject3
==> Installing dependencies for pygobject3: py3cairo
==> Installing pygobject3 dependency: py3cairo
==> Downloading https://homebrew.bintray.com/bottles/py3cairo-1.10.0_3.el_capitan.bottle.tar.gz
==> Pouring py3cairo-1.10.0_3.el_capitan.bottle.tar.gz
Warning: python modules have explicit framework links
  /usr/local/opt/py3cairo/lib/python3.6/site-packages/cairo/_cairo.cpython-36m-darwin.so
==> Summary
🍺  /usr/local/Cellar/py3cairo/1.10.0_3: 10 files, 153.9K
==> Installing pygobject3 --with-python3
==> Using the sandbox
==> Downloading https://download.gnome.org/sources/pygobject/3.22/pygobject-3.22.0.tar.xz
==> Downloading from https://mirror.umd.edu/gnome/sources/pygobject/3.22/pygobject-3.22.0.tar.xz
==> hello python: python
==> ./configure --prefix=/usr/local/Cellar/pygobject3/3.22.0 PYTHON=python
==> make install
==> make clean
==> hello python: python3
==> ./configure --prefix=/usr/local/Cellar/pygobject3/3.22.0 PYTHON=python3
==> make install
==> make clean
🍺  /usr/local/Cellar/pygobject3/3.22.0: 165 files, 3.4M, built in 24 seconds
iMac-TMP:~ joe$ git -C `brew --repo homebrew/core` diff -- Formula/pygobject3.rb
diff --git a/Formula/pygobject3.rb b/Formula/pygobject3.rb
index 33b0f2f..8a910f4 100644
--- a/Formula/pygobject3.rb
+++ b/Formula/pygobject3.rb
@@ -26,6 +26,7 @@ class Pygobject3 < Formula
     ENV.universal_binary if build.universal?
 
     Language::Python.each_python(build) do |python, _version|
+ohai "hello python: #{python}"
       system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}", "PYTHON=#{python}"
       system "make", "install"
       system "make", "clean"
```

actionless:
but isn't that logical to always assume `--with-python3` if `python3` package is installed by brew?

DomT4:
Well, that's not what you're trying to do in this specific PR with the current code, but no, that's explicitly not how Homebrew's `superenv` system is designed to work.

There are certainly formulae still around which go _"Oh you have `libidn` installed, cool, I'll link to that"_ but the `superenv` system is based around slapping build systems on the wrist when they try to do that and forcing formulae to specify valid linkage and/or potential linkage, albeit with the caveat that if something for example requires `cairo` you don't also need to explicitly write `depends_on "freetype"`, `depends_on "fontconfig"`, etc into the same formula.

ilovezfs:
In this case, you'll need to do
```
brew install --with-python3 pygobject3
```
if you want the python3 support, but thank you for the PR nonetheless.

PR #10305, opened 2017-02-24
borg: use GitHub for homepage

ilovezfs:


ilovezfs:
Reported https://github.com/ok-borg/borg/issues/52

crufter:
The problem is not actually with the frontend, that could be on github. But the server is also down. Will fix after work.

ilovezfs:
@crufter thanks

ilovezfs:
@crufter hmm it still seems to be down.

PR #10340, opened 2017-02-25
m4ri 20140914 (new formula)

imhotepisinvisible:


fxcoudert:
Thanks for your submission. I am not sure we can accept this package as-is: from the Bitbucket repository statistics, it seems to have very low activity: 1 fork, 4 watchers, 1 pull request, no release since 2014… It does not seem to meet the broad audience of Homebrew, and would probably be better if hosted in a [personal tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

Sorry, and thanks again!

MikeMcQuaid:
Passing on this for now as @fxcoudert says, sorry!

imhotepisinvisible:
OK, thanks for your consideration.  To note, it is a component of Sage Math ( http://doc.sagemath.org/html/en/reference/matrices/sage/matrix/matrix_mod2_dense.html ) which of course has a large audience, so it seemed to me like it could be a candidate for inclusion.  It is in the Ubuntu/Debian repos ( http://packages.ubuntu.com/source/xenial/libm4ri ), but I don't know what your policy is on matching that content.

ilovezfs:
This is probably a better candidate for homebrew/science.

imhotepisinvisible:
OK, thanks I will submit it to there.  BTW, their website doesn't work: https://brew.sh/homebrew-science/

ilovezfs:
https://github.com/Homebrew/homebrew-science/

imhotepisinvisible:
Yes I found that, but they link to the brew.sh website in their description, which doesn't work

ilovezfs:
@imhotepisinvisible yup, I see what you mean. I let the science channel know on Slack :)

PR #10347, opened 2017-02-25
Add Python 3 support to protobuf

Noctem:

-----
Optionally build the protobuf package for Python 3, also update setuptools and appdirs.

### DIFF
--- a/Formula/protobuf.rb
+++ b/Formula/protobuf.rb
@@ -3,6 +3,7 @@ class Protobuf < Formula
   homepage "https://github.com/google/protobuf/"
   url "https://github.com/google/protobuf/archive/v3.2.0.tar.gz"
   sha256 "2a25c2b71c707c5552ec9afdfb22532a93a339e1ca5d38f163fe4107af08c54c"
+  revision 1
   head "https://github.com/google/protobuf.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Formula/protobuf.rb
```diff
@@ -81,30 +108,8 @@ def install
...
-        args = Language::Python.setup_install_args libexec
-        args << "--cpp_implementation"
-        system "python", *args
-      end
-      site_packages = "lib/python2.7/site-packages"
-      pth_contents = "import site; site.addsitedir('#{libexec/site_packages}')\n"
-      (prefix/site_packages/"homebrew-protobuf.pth").write pth_contents
-    end
+    pyinstall("python") if build.with? "python"
+    pyinstall("python3") if build.with? "python3"
```
MikeMcQuaid: Use `Language::Python.each_python(build) do |python, python_version|` or `python_versions.each` over some array instead of creating a new function.

MikeMcQuaid (changes requested):

fxcoudert:
This might need a `revision` bump due to the updated resources.

bfontaine:
Ping on the comments above?

Noctem:
So is a revision bump actually needed? I can add it if it is.

fxcoudert:
Yes, please add the revision bump and squash the commits.

dunn:
Thanks! Merged as 5cb2a3ff831eceeb694930180468f66035302125.

Noctem:
FYI: the *.pth* file seems to trigger this bug when `import`ing pip: pypa/pip#4216

Changing it to just the path (instead of importing and using `site`) seems to fix it, but I don't know off the top of my head if that has other effects. Apparently an upcoming pip release will fix the bug, so I guess intervention probably isn't necessary here.

PR #10337, opened 2017-02-25
Remove universal support, except for wine dependencies

fxcoudert:
- [X] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [X] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [X] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [X] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

This is based on #9641, where I have fixed the comments by @ilovezfs. It removes the universal option from 147 formulas, but does not touch the 17 formulas that are `wine` dependencies (as they need to be vendored into the wine formula before universal is removed from them).

PR #10348, opened 2017-02-26
Remove universal support 1/n

fxcoudert:
Removing universal, PR 1/n: 29 changes out of a total of 147. This is split from #10337, so the testbot does not go crazy.


ilovezfs (approved):
gcore needs its own PR, a revision bump, and its new bottle pulled, but the rest look good to me.

ilovezfs:
Builds 1315 - 1345 on the Homebrew Testing job

PR #10349, opened 2017-02-26
Remove universal support 2/n

fxcoudert:
Removing universal, PR 2/n: 25 changes out of a total of 147. This is split from #10337, so the testbot does not go crazy.

### DIFF
--- a/Formula/jsoncpp.rb
+++ b/Formula/jsoncpp.rb
@@ -12,8 +12,6 @@ class Jsoncpp < Formula
     sha256 "9f33d38dc6d459622705c4982204e0eea49a5d9a14440a032695dd1b72214cad" => :yosemite
   end
 
-  option :universal
-
   needs :cxx11
 
   depends_on "cmake" => :build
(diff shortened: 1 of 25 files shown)


Formula/jsoncpp.rb
```diff
@@ -12,27 +12,19 @@ class Jsoncpp < Formula
...
-      -DJSONCPP_WITH_CMAKE_PACKAGE=ON
-      -DJSONCPP_WITH_TESTS=OFF
-      -DJSONCPP_WITH_POST_BUILD_UNITTEST=OFF
-    ]
-    if build.universal?
-      ENV.universal_binary
-      cmake_args << "-DCMAKE_OSX_ARCHITECTURES=#{Hardware::CPU.universal_archs.as_cmake_arch_flags}"
-    end
-    system "cmake", ".", *cmake_args
+    system "cmake", ".", *cmake_args,
```
ilovezfs: I think you mean `std_cmake_args`

Formula/libepoxy.rb
```diff
@@ -11,8 +11,6 @@ class Libepoxy < Formula
     sha256 "f95aff4f5d3aed6991335ea6f67e3377a8a99200a554ca2cb0026c5e20630523" => :yosemite
   end
 
-  option :universal
```
ilovezfs: ` ENV.universal_binary if build.universal?` still needs to be removed from this one.

ilovezfs (changes requested):

ilovezfs:
Builds 1346 - 1370 on the Homebrew Testing job

PR #10350, opened 2017-02-26
Remove universal support 3/n

fxcoudert:
Removing universal, PR 3/n: 26 changes out of a total of 147. This is split from #10337, so the testbot does not go crazy.


ilovezfs (approved):

ilovezfs:
Builds 1372 - 1397 on the Homebrew Testing job

PR #10351, opened 2017-02-26
Remove universal support 4/n

fxcoudert:
Removing universal, PR 3/n: 27 changes out of a total of 147. This is split from #10337, so the testbot does not go crazy.

### DIFF
--- a/Formula/py2cairo.rb
+++ b/Formula/py2cairo.rb
@@ -15,8 +15,6 @@ class Py2cairo < Formula
     sha256 "784e49b2f15f30af7f4e255eb2263c6e99ae4e0d0ec961412ff033d0954fd298" => :mountain_lion
   end
 
-  option :universal
-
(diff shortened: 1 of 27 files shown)


Formula/py2cairo.rb
```diff
@@ -31,7 +29,7 @@ def install
     ENV.append_to_cflags `python-config --includes`
 
     # Python extensions default to universal but cairo may not be universal
```
ilovezfs: This comment should be revised since universal is no longer a possibility here.

ilovezfs (approved):
assuming the py2cairo comment is updated

fxcoudert:
@ilovezfs thanks for the review!

ilovezfs:
You're welcome.

ilovezfs:
Builds 1398 - 1424 on the Homebrew Testing job

PR #10352, opened 2017-02-26
Remove universal support 5/n (n=5)

fxcoudert:
Removing universal, PR 5/n (n=5): 35 changes out of a total of 147. This is split from #10337, so the testbot does not go crazy.

### DIFF
--- a/Formula/subversion.rb
+++ b/Formula/subversion.rb
@@ -16,7 +16,6 @@ class Subversion < Formula
   deprecated_option "perl" => "with-perl"
   deprecated_option "ruby" => "with-ruby"
 
-  option :universal
   option "with-java", "Build Java bindings"
   option "with-perl", "Build Perl bindings"
   option "with-ruby", "Build Ruby bindings"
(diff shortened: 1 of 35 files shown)


Formula/subversion.rb
```diff
@@ -16,7 +16,6 @@ class Subversion < Formula
   deprecated_option "perl" => "with-perl"
   deprecated_option "ruby" => "with-ruby"
 
-  option :universal
```
ilovezfs:
```
    resource("serf").stage do
      # SConstruct merges in gssapi linkflags using scons's MergeFlags,
      # but that discards duplicate values - including the duplicate
      # values we want, like multiple -arch values for a universal build.
      # Passing 0 as the `unique` kwarg turns this behaviour off.
      inreplace "SConstruct", "unique=1", "unique=0"
```
This is probably no longer needed unless there are other duplicates. If it is still needed, the comment should be updated.

Formula/wxmac.rb
```diff
@@ -64,7 +64,6 @@ class Wxmac < Formula
     end
   end
 
-  option :universal
```
ilovezfs:
```
      # Otherwise, even in superenv, the internal libtiff can pick
      # up on a nonuniversal xz and fail
      # https://github.com/Homebrew/legacy-homebrew/issues/22732
      "--without-liblzma",
```
We probably don't need this argument anymore.

ilovezfs (approved):
assuming the `subversion` and `wxmac` changes are made

ilovezfs:
Builds 1426 - 1460 on the Homebrew Testing job

PR #10367, opened 2017-02-26
bochs: fix configure flags

Nax:
Makes the configure flags a bit more sane:

 * Removed `--enable-xpm`, as it has no effect unless linking with X11.
 * Added `--enable-avx`, as AVX is quite standard now, and the bochs implementation is solid.
 * Added `--enable-vmx`, same as AVX, standard and solid.
 * Added `--enable-smp`, no cost when there is only one CPU.
 * Added `--enable-long-phy-addres`, so bochs can use more than 4GiB of RAM.

-----

### DIFF
--- a/Formula/bochs.rb
+++ b/Formula/bochs.rb
@@ -34,13 +34,16 @@ class Bochs < Formula
       --enable-debugger-gui
       --enable-readline
       --enable-iodebug
-      --enable-xpm
       --enable-show-ips
       --enable-logging
       --enable-usb
       --enable-ne2000
       --enable-cpu-level=6
       --enable-clgd54xx
+      --enable-avx
+      --enable-vmx
+      --enable-smp
+      --enable-long-phy-addres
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Will AVX/VMX be enabled even when the CPU doesn't support them?

Nax:
I don't think you understood what these flags are for.

Bochs is an x86 (and x86_64) emulator. These flags control what instruction sets are available to the emulated CPUs.
It's perfectly possible to emulate AVX on a non-AVX CPU.

However, even is these flags are enabled, the emulated CPU needs to be compatible with the instruction sets. For example, emulating an old i3 will not give you access to AVX instructions.

MikeMcQuaid:
Gotcha, thanks. I did not understand.

MikeMcQuaid (approved):

nijikon:

PR #10413, opened 2017-02-27
mame & rom-tools 0.183

mbcoguno:

-----

* ~~Add option to build MAME's tools inside the MAME formula.~~
* Update lua@5.3's comments.
* Change rom-tools build's main target, where ldplayer target is not longer exist.

### DIFF
--- a/Formula/mame.rb
+++ b/Formula/mame.rb
@@ -1,9 +1,9 @@
 class Mame < Formula
   desc "Multiple Arcade Machine Emulator"
   homepage "http://mamedev.org/"
-  url "https://github.com/mamedev/mame/archive/mame0179.tar.gz"
-  version "0.179"
-  sha256 "d1616ef32b884c3e7913378ebf5282b6f846f895f419eb92a9068770581e081b"
+  url "https://github.com/mamedev/mame/archive/mame0183.tar.gz"
+  version "0.183"
+  sha256 "c12b3051f2f11331a38f557eac7f3074166e48155133b2f3e7cc323df56ce8b0"
   head "https://github.com/mamedev/mame.git"
 
   bottle do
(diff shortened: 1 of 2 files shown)


mbcoguno:
I have a question:
I want to merge rom-tools inside MAME formula, what will I do for the rom-tools formula? Delete it or boneyard it?

Formula/mame.rb
```diff
@@ -47,10 +55,39 @@ def install
...
+      man1.install Dir["docs/man/*.1"]
+    end
   end
 
   test do
     assert shell_output("#{bin}/mame -help").start_with? "MAME v#{version}"
     system "#{bin}/mame", "-validate"
+
+    if build.with? "tools"
+      # system "#{bin}/aueffectutil" # segmentation fault
```
fxcoudert: Why are we installing a tool that segfaults?

mbcoguno: Because it's part of MAME's tools.
But since Homebrew won't accept unstable formulae, and some MAME tools never get working, I'll remove the SEGFAULT binary in next few days.

fxcoudert:
Regarding whether to integrate `rom-tools` into `mame`, the questions are:
- is `mame` usable without the `rom-tools`?
- what about other mame packages, such as `advancemame` and `advancemenu`?

mbcoguno:
> is mame usable without the rom-tools?

Generally, MAME is usable without tools. Mostly are old computer's disk image creation tools. Arcade machine emulator isn't use it.
See <http://docs.mamedev.org/tools/index.html>

> what about other mame packages, such as advancemame and advancemenu?

They're MAME's forks. They're not affected.

mbcoguno:
I'll simply remove rom-tools. If someone think boneyard is better please let me know.

fxcoudert:
> They're MAME's forks. They're not affected.

Do they build the rom-tools? If not, wouldn't it be better to have rom-tools as a separate formula, so that all users (of mame and its forks) can benefit?

mbcoguno:
I think it's not bad, although these tools aren't working well in macOS, and rarely user use rom-tools bottle.
Preserve rom-tools right now unless they need boneyard.

fxcoudert:
Thanks @mbcoguno for your contribution!

PR #10422, opened 2017-02-27
Add back in universal option for some (optional) wine dependencies

JCount:
Add back in universal options to wine dependencies, mainly ones part of the optional dependency tree.


JCount:
CC @fxcoudert @ilovezfs @nijikon 

I just want to let you guys know why there are going to be some commits re-adding universal options.

JCount:
Failures due to `brew audit --online` getting 404s because of some instability in fossies.org's network.

PR #10431, opened 2017-02-28
sha1dc git (new formula) Detect SHAttered collisions

johnhawkinson:

-----
The reference implementation to detect SHA-1 collisions from https://github.com/cr-marcstevens/sha1collisiondetection as linked from https://shattered.io/

### DIFF
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -1,24 +1,18 @@
 class Sha1dc < Formula
   desc "Tool to detect SHA-1 collisions in files, including SHAttered"
   homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
+  url "https://github.com/cr-marcstevens/sha1collisiondetection/archive/stable-v1.0.2.tar.gz"
+  sha256 "ab1a5624d12417b5af9ffe6662546b6932c39c0e4c4bef822851a2d0e53df790"
 
   # The "master" branch is unusably broken and behind the
   # "simplified_c90" branch that's the basis for release.
   head "https://github.com/cr-marcstevens/sha1collisiondetection.git",
-       :using => :git,
-       :branch => "simplified_c90"
-
-  devel do
-    url "https://github.com/cr-marcstevens/sha1collisiondetection/archive/development-v1.0.1.tar.gz"
-    sha256 "a414aad7ab1da30193d053ad31651d39a449c2f9f74eac777e467929d0de3c93"
-    version "dev-v1.0.1"
-  end
+       :using => :git
 
   depends_on "libtool" => :build
   depends_on "coreutils" => :build # GNU install
 
   def install
-
     system "make", "INSTALL=ginstall", "PREFIX=#{prefix}", "install"
     (pkgshare/"test").install Dir["test/*"]
   end
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -2,22 +2,22 @@ class Sha1dc < Formula
   desc "Tool to detect SHA-1 collisions in files, including SHAttered"
   homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
   url "https://github.com/cr-marcstevens/sha1collisiondetection.git",
-      :revision => "cb10f50d20a1c6c9efbf526c5da510c492d595b9"
-  version "HEAD-cb10f0"
-  head "https://github.com/cr-marcstevens/sha1collisiondetection.git"
+      :revision => "6b1c12f3606d0959f8658d44a324fcf5f4497f88"
+  version "HEAD-6b1c12"
 
   depends_on "libtool" => :build
 
   def install
     inreplace "Makefile", "libtool", "glibtool"
 
-    # By default tries to build with HAVEAVX=1, fails.
-    inreplace "Makefile", "HAVEAVX=1", "HAVEAVX=0"
-    system "make"
+    ENV.deparallelize
 
-    bin.install "bin/sha1dcsum"
-    bin.install "bin/sha1dcsum_partialcoll"
-    lib.install "bin/.libs/libdetectcoll.a"
+    # By default tries to build with HAVEAVX=1, fails.
+    system "make", "HAVEAVX=0"
+
+    system "make", "PREFIX=#{prefix}", "install"
+    (pkgshare/"test").install Dir["test/*"]
+
   end
 
   test do
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -26,8 +26,8 @@ class Sha1dc < Formula
   end
 
   test do
-    system "#{bin}/sha1dcsum #{pkgshare}/test/shattered-1.pdf | fgrep -q '*coll*'"
-    system "#{bin}/sha1dcsum #{pkgshare}/test/shattered-2.pdf | fgrep -q '*coll*'"
-    system "#{bin}/sha1dcsum_partialcoll #{pkgshare}/test/sha1_reducedsha_coll.bin | fgrep -q '*coll*'"
+    assert_match "*coll*", shell_output("#{bin}/sha1dcsum #{pkgshare}/test/shattered-1.pdf")
+    assert_match "*coll*", shell_output("#{bin}/sha1dcsum #{pkgshare}/test/shattered-2.pdf")
+    assert_match "*coll*", shell_output("#{bin}/sha1dcsum_partialcoll #{pkgshare}/test/sha1_reducedsha_coll.bin")
   end
 end
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -2,8 +2,8 @@ class Sha1dc < Formula
   desc "Tool to detect SHA-1 collisions in files, including SHAttered"
   homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
 
-  # The "master" branch is unusably broken and behind the "simplified_c90" branch
-  # that's the basis for release.
+  # The "master" branch is unusably broken and behind the
+  # "simplified_c90" branch that's the basis for release.
   head "https://github.com/cr-marcstevens/sha1collisiondetection.git",
        :using => :git,
        :branch => "simplified_c90"
@@ -15,15 +15,12 @@ class Sha1dc < Formula
   end
 
   depends_on "libtool" => :build
-  depends_on "coreutils" => :build
+  depends_on "coreutils" => :build # GNU install
 
   def install
     # Workaround a poor Makefile choice that prevents cmdline override of INSTALL
     inreplace "Makefile", "INSTALL ?= install", "INSTALL ?= ginstall"
 
-    ENV.deparallelize
-
-    system "make", "LIBTOOL=glibtool"
     system "make", "LIBTOOL=glibtool", "PREFIX=#{prefix}", "install"
     (pkgshare/"test").install Dir["test/*"]
   end
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -3,7 +3,7 @@ class Sha1dc < Formula
   homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
   url "https://github.com/cr-marcstevens/sha1collisiondetection.git",
       :revision => "6b1c12f3606d0959f8658d44a324fcf5f4497f88"
-  version "HEAD-6b1c12"
+  version "git-6b1c12"
 
   depends_on "libtool" => :build
 
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -14,14 +14,15 @@ class Sha1dc < Formula
 
     # By default tries to build with HAVEAVX=1, fails.
     system "make", "HAVEAVX=0"
-
+    # Some bug in the Makefile...
+    system "mkdir", "-p", "#{prefix}/bin", "#{prefix}/lib" 
     system "make", "HAVEAVX=0", "PREFIX=#{prefix}", "install"
     (pkgshare/"test").install Dir["test/*"]
-
   end
 
   test do
-    # Build system's 'make' ran 'make test'
-    system "#{bin}/sha1dcsum"
+    system "#{bin}/sha1dcsum #{pkgshare}/test/shattered-1.pdf | fgrep -q '*coll*'"
+    system "#{bin}/sha1dcsum #{pkgshare}/test/shattered-2.pdf | fgrep -q '*coll*'"
+    system "#{bin}/sha1dcsum_partialcoll #{pkgshare}/test/sha1_reducedsha_coll.bin | fgrep -q '*coll*'"
   end
 end
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -15,7 +15,7 @@ class Sha1dc < Formula
     # By default tries to build with HAVEAVX=1, fails.
     system "make", "HAVEAVX=0"
 
-    system "make", "PREFIX=#{prefix}", "install"
+    system "make", "HAVEAVX=0", "PREFIX=#{prefix}", "install"
     (pkgshare/"test").install Dir["test/*"]
 
   end
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -1,0 +1,27 @@
+class Sha1dc < Formula
+  desc "Tool to detect SHA-1 collisions in files, including SHAttered"
+  homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
+  url "https://github.com/cr-marcstevens/sha1collisiondetection.git",
+      :revision => "cb10f50d20a1c6c9efbf526c5da510c492d595b9"
+  version "HEAD-cb10f0"
+  head "https://github.com/cr-marcstevens/sha1collisiondetection.git"
+
+  depends_on "libtool" => :build
+
+  def install
+    inreplace "Makefile", "libtool", "glibtool"
+
+    # By default tries to build with HAVEAVX=1, fails.
+    inreplace "Makefile", "HAVEAVX=1", "HAVEAVX=0"
+    system "make"
+
+    bin.install "bin/sha1dcsum"
+    bin.install "bin/sha1dcsum_partialcoll"
+    lib.install "bin/.libs/libdetectcoll.a"
+  end
+
+  test do
+    # Build system's 'make' ran 'make test'
+    system "#{bin}/sha1dcsum"
+  end
+end
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -18,10 +18,8 @@ class Sha1dc < Formula
   depends_on "coreutils" => :build # GNU install
 
   def install
-    # Workaround a poor Makefile choice that prevents cmdline override of INSTALL
-    inreplace "Makefile", "INSTALL ?= install", "INSTALL ?= ginstall"
 
-    system "make", "LIBTOOL=glibtool", "PREFIX=#{prefix}", "install"
+    system "make", "INSTALL=ginstall", "PREFIX=#{prefix}", "install"
     (pkgshare/"test").install Dir["test/*"]
   end
 
--- a/Formula/sha1dc.rb
+++ b/Formula/sha1dc.rb
@@ -1,22 +1,30 @@
 class Sha1dc < Formula
   desc "Tool to detect SHA-1 collisions in files, including SHAttered"
   homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
-  url "https://github.com/cr-marcstevens/sha1collisiondetection.git",
-      :revision => "6b1c12f3606d0959f8658d44a324fcf5f4497f88"
-  version "git-6b1c12"
+
+  # The "master" branch is unusably broken and behind the "simplified_c90" branch
+  # that's the basis for release.
+  head "https://github.com/cr-marcstevens/sha1collisiondetection.git",
+       :using => :git,
+       :branch => "simplified_c90"
+
+  devel do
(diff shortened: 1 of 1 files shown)


Formula/sha1dc.rb
```diff
@@ -0,0 +1,27 @@
+class Sha1dc < Formula
+  desc "Tool to detect SHA-1 collisions in files, including SHAttered"
+  homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
+  url "https://github.com/cr-marcstevens/sha1collisiondetection.git",
+      :revision => "cb10f50d20a1c6c9efbf526c5da510c492d595b9"
```
MikeMcQuaid: Need a stable, released tag (for which we'd used the tarball) for this to be accepted.

Formula/sha1dc.rb
```diff
@@ -0,0 +1,27 @@
...
+  version "HEAD-cb10f0"
+  head "https://github.com/cr-marcstevens/sha1collisiondetection.git"
+
+  depends_on "libtool" => :build
+
+  def install
+    inreplace "Makefile", "libtool", "glibtool"
+
+    # By default tries to build with HAVEAVX=1, fails.
+    inreplace "Makefile", "HAVEAVX=1", "HAVEAVX=0"
```
MikeMcQuaid: Submit the two above `inreplace`s to the upstream developers of this project and add a link to the upstream PR submission as a comment in the formula file. Thanks!

Formula/sha1dc.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "make"
+
+    bin.install "bin/sha1dcsum"
+    bin.install "bin/sha1dcsum_partialcoll"
+    lib.install "bin/.libs/libdetectcoll.a"
+  end
+
+  test do
+    # Build system's 'make' ran 'make test'
```
MikeMcQuaid: Can remove this comment.

Formula/sha1dc.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "#{bin}/sha1dcsum"
```
MikeMcQuaid: Would be good to have a slightly more involved test that does something.

Formula/sha1dc.rb
```diff
@@ -0,0 +1,27 @@
...
+  depends_on "libtool" => :build
+
+  def install
+    inreplace "Makefile", "libtool", "glibtool"
+
+    # By default tries to build with HAVEAVX=1, fails.
+    inreplace "Makefile", "HAVEAVX=1", "HAVEAVX=0"
+    system "make"
+
+    bin.install "bin/sha1dcsum"
```
MikeMcQuaid: Is there no `make install` that can be used?

MikeMcQuaid (changes requested):

Formula/sha1dc.rb
johnhawkinson: There isn't one — and of course we're in a world of relatively bleeding-edge software, what with SHAttered only being released less than a week ago.

Formula/sha1dc.rb
johnhawkinson: The upstream developer has no platform-specific build system.
Putting either of these into their Makefile for all platforms wouldn't be correct.
And trying to add in platform support in their Makefile for OS X seems like something they wouldn't be wild about accepting.

Formula/sha1dc.rb
johnhawkinson: Correct, there is no `make install` target :(

Formula/sha1dc.rb
johnhawkinson: OK. The intent was to justify the lack of tests.

Formula/sha1dc.rb
johnhawkinson: So, doing so would essentially duplicate the build system's tests. Is that OK?

Also, then, the source tree has reference test files in the test/ subdirectory. How/where should I install them in order to facilitate the tests? Should I use `share.install` and install them in the `share/` toplevel? Or in `share/test`? Or somewhere else?

Or is there some way to reference files in the source tree from the test section of the recipe? That would be ideal but I did not find one.

johnhawkinson (commented):
replies. Sorry I'm new at this.

Formula/sha1dc.rb
MikeMcQuaid: Worth creating an upstream issue to request one in cases like this.

Formula/sha1dc.rb
MikeMcQuaid: Having these values be variables so e.g. `system "make", "HAVEAVX=0", "LIBTOOL=glibtool"` would be a happy middle ground useful to other people without causing maintenance overhead.

Formula/sha1dc.rb
MikeMcQuaid: In a case like this it's worth trying to submit a trivial one upstream (that accepts `PREFIX` as an argument)`

Formula/sha1dc.rb
MikeMcQuaid: It's fine, the test below is sufficient.

Formula/sha1dc.rb
MikeMcQuaid: > So, doing so would essentially duplicate the build system's tests. Is that OK?

Yup; this is a packaging test rather than just functionality test.

> Or somewhere else?

`(pkgshare/"test")` is probably best.

Something simple that just e.g. verifies a collision would be 👍. Feel free to punt on this until upstream resolves the other requirements stated above, though; don't want you to waste the effort.

MikeMcQuaid:
> replies. Sorry I'm new at this.

No apologies needed. You submitted a PR and that alone is appreciated regardless.

Formula/sha1dc.rb
johnhawkinson: https://github.com/cr-marcstevens/sha1collisiondetection/issues/11

Formula/sha1dc.rb
johnhawkinson: Oh, duh. That works great for HAVEAVX (which itself is a workaround setting. Looks like https://github.com/cr-marcstevens/sha1collisiondetection/pull/4 tries to fix a bunch of this stuff more properly). LIBTOOL probably has to stay as a hack pending those kinds of fixes given that

```
LT_CC:=glibtool --tag=CC --mode=compile $(CC)
```
is probably too much to try to express on the command line, and there are pending changes to this area of the Makefile anyhow. (I didn't review open pull requests on the project before writing the recipe -- my bad)

Formula/sha1dc.rb
johnhawkinson: Well, I guess there's an actual test, sure. But if there isn't, it seems like an explanation is owed...whatever :)

Formula/sha1dc.rb
johnhawkinson: It seems orthogonal to the upstream issues.

johnhawkinson (commented):
Replies. Will push changes in a bit.

johnhawkinson:
Umm, help. Without any changes from the previously-submitted formula I am now getting an "Error: empty installation" and I don't know why. The installation seems anything but empty. Any tips?:

```
pb3:homebrew-core jhawk$ brew uninstall sha1dc
Unsetting PERL5LIB to protect you from yourself
Uninstalling /usr/local/Cellar/sha1dc/HEAD-cb10f50... (6 files, 3.9MB)
pb3:homebrew-core jhawk$ brew install  sha1dc
Unsetting PERL5LIB to protect you from yourself
==> Using the sandbox
==> Cloning https://github.com/cr-marcstevens/sha1collisiondetection.git
Updating /Users/jhawk/Library/Caches/Homebrew/sha1dc--git
==> Checking out revision cb10f50d20a1c6c9efbf526c5da510c492d595b9
==> make
Error: Empty installation
pb3:homebrew-core jhawk$ ls -lR /usr/local/Cellar/sha1dc/
total 0
drwxr-xr-x  7 jhawk  admin  238 Feb 28 04:47 HEAD-cb10f50

/usr/local/Cellar/sha1dc//HEAD-cb10f50:
total 24
-rw-r--r--  1 jhawk  admin   591 Feb 28 04:47 INSTALL_RECEIPT.json
-rw-r--r--  1 jhawk  staff  1270 Feb 28 02:32 LICENSE.txt
-rw-r--r--  1 jhawk  staff  3281 Feb 28 02:32 README.md
drwxr-xr-x  4 jhawk  admin   136 Feb 28 04:47 bin
drwxr-xr-x  3 jhawk  admin   102 Feb 28 04:47 lib

/usr/local/Cellar/sha1dc//HEAD-cb10f50/bin:
total 5264
-rwxr-xr-x  1 jhawk  staff  1344332 Feb 28 04:47 sha1dcsum
-rwxr-xr-x  1 jhawk  staff  1344332 Feb 28 04:47 sha1dcsum_partialcoll

/usr/local/Cellar/sha1dc//HEAD-cb10f50/lib:
total 2736
-rw-r--r--  1 jhawk  staff  1397728 Feb 28 04:47 libdetectcoll.a
pb3:homebrew-core jhawk$ 
```

johnhawkinson:
In re https://github.com/Homebrew/brew/issues/2229, @ MikeMcQuaid said:

> Let's talk about it there. I'm not really sure what you're being asked and why it would be necessary to install .la files. Ideally make install handles all such things instead; if there's no make install then submitting one upstream is the best bet for other packagers to have an easier life too.

So, when program's build system builds a library foo with `libtool`, it produces `libfoo.la` and `.libs/libfoo.a` or maybe `.libs/libfoo.dylib`. From the `libtool` perspective, you're not supposed to have to know or think about the `.libs` directory, it's a private internal secret part of libtool.

So I would expect that homebrew would have a better answer than "reach inside the `.libs` directory and pull out the real files." If that's the answer, well, OK, it's nonintuitive enough to me that it would be good to have it in the Cookbook.

On the other hand, if it's not the right thing, well, then I don't know at all what the answer is, and it would be nice to know :)

Also, the only example I could find was `rtmidi.rb` which makes me think I'm missing something? It does this:
```

  def install
    system "./autogen.sh", "--no-configure"
    system "make"
    lib.install Dir[".libs/*.a", ".libs/*.dylib"]
    include.install Dir["*.h"]
  end
```

Thanks.

zbeekman:
@johnhawkinson dang you beat me to this, I was going to work up a Formula PR... On the road today, but if you're still stuck tomorrow I'll have a look.

Also, upstream author/maintainer seems pretty unresponsive. I'm wondering if we might just want to create & maintain a dedicated Mac OS fork of the project that includes all the fixes, and, possibly, a more robust build system (autotools or CMake). Maybe I'll ping upstream with an email in a few days if we don't see any action on his end. Would also solve the lack of release issue.

MikeMcQuaid:
> If that's the answer, well, OK, it's nonintuitive enough to me that it would be good to have it in the Cookbook.
> 
> On the other hand, if it's not the right thing, well, then I don't know at all what the answer is, and it would be nice to know :)

The answer is basically "upstream should handle this in their `make install` and we should submit one to them if they don't have one already.

Formula/sha1dc.rb
MikeMcQuaid: I was more thinking: we're not going to be able to accept this unless upstream makes a release tag so in case that didn't happen I didn't want you to have spent too much effort here.

Formula/sha1dc.rb
MikeMcQuaid: @johnhawkinson Looks like adding a `LIBTOOL` variable is doable. I've made comments on https://github.com/cr-marcstevens/sha1collisiondetection/pull/4#pullrequestreview-24293217

johnhawkinson:
> The answer is basically "upstream should handle this in their make install and we should submit one to them if they don't have one already.

OK. So to put it differently, "For relatively simple packages, it's ok to use `bin.install`; but for anything at the level of complexity where libtool applies, it is a requirement to fix the upstream to include a `make install` rule."

I guess the corollary would be in the cases where it was truly necessary, to do something like:
`system "glibtool", "--mode=install", "install," "-c", "libwhatever.la", "#{lib}"`
or something. And I can certainly see why you'd want to avoid that.

Formula/sha1dc.rb
johnhawkinson: Gotcha. While it seems clear they're add a release tag (they have said so), what's the actual homebrew rule here? I suppose I wouldn't have put any of this packaging effort in if I'd realized it would block on the upstream designating a tag (although I did learn more about homebrew in the process...).

It can't possibly be the rule that homebrew won't package software whose upstream doesn't bother to maintain adequate versioning, since that's a not-infrequent issue in reality. Or is the requirement that I would have to fork it myself and publish a tag? (Not a business I want to be in, at least in this particular case, which has active maintenance).

Or is this a maintainer judgement call, i.e. that this package isn't important enough to merit such treatment, but some other package might be? I don't disagree with that evaluation, but I'm just trying to understand the rules of the road.

ilovezfs: >It can't possibly be the rule that homebrew won't package software whose upstream doesn't bother to maintain adequate versioning, since that's a not-infrequent issue in reality.

The rule it cannot possibly be is the rule.

johnhawkinson:
> Umm, help. Without any changes from the previously-submitted formula I am now getting an "Error: empty installation" and I don't know why. The installation seems anything but empty. Any tips?:

Did I mention I wasn't a Ruby programmer? Its probably obvious. Applying:

```
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -677,6 +677,15 @@ class FormulaInstaller
 
     formula.update_head_version
 
+    if !formula.prefix.directory?
+      ohai "prefix.directory?: "
+      ohai formula.prefix
+      ohai formula
+      ohai formula.prefix.directory?
+    end
+    if Keg.new(formula.prefix).empty_installation?
+      ohai "keg empty?"
+    end
     if !formula.prefix.directory? || Keg.new(formula.prefix).empty_installation?
       raise "Empty installation"
     end
```

gives me:

```
==> prefix.directory?: 
==> /usr/local/Cellar/sha1dc/HEAD-cb10f0
==> sha1dc
==> false
Error: /usr/local/Cellar/sha1dc/HEAD-cb10f0 is not a directory
/usr/local/Homebrew/Library/Homebrew/keg.rb:186:in `initialize'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:686:in `new'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:686:in `build'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:315:in `install'
/usr/local/Homebrew/Library/Homebrew/cmd/reinstall.rb:47:in `reinstall_formula'
/usr/local/Homebrew/Library/Homebrew/cmd/reinstall.rb:18:in `block in reinstall'
/usr/local/Homebrew/Library/Homebrew/cmd/reinstall.rb:13:in `each'
/usr/local/Homebrew/Library/Homebrew/cmd/reinstall.rb:13:in `reinstall'
/usr/local/Homebrew/Library/Homebrew/brew.rb:94:in `<main>'
```

To which my response is kinda "huh? How is it not a directory? It has files in it..."
(Should I be moving these dumb questions to https://discourse.brew.sh/t/help-debugging-formuas/645?)

MikeMcQuaid:
> It can't possibly be the rule that homebrew won't package software whose upstream doesn't bother to maintain adequate versioning, since that's a not-infrequent issue in reality.

 This is the case and the reason being that if upstream won't provide versions they don't really provide "stable" software.

@johnhawkinson The "empty installation" looks like a bug. Could you try to create as minimal a test formula as possible and submit it to Homebrew/brew's issue tracker? Thanks!

johnhawkinson:
There's now a development release tag, updated the recipe to deal with it and it's foibles, as well to use a HEAD tag. Not sure if a development release is enough to be ready to merge or not.

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+    sha256 "a414aad7ab1da30193d053ad31651d39a449c2f9f74eac777e467929d0de3c93"
+    version "dev-v1.0.1"
+  end
+
+  depends_on "libtool" => :build
+  depends_on "coreutils" => :build
+
+  def install
+    # Workaround a poor Makefile choice that prevents cmdline override of INSTALL
+    inreplace "Makefile", "INSTALL ?= install", "INSTALL ?= ginstall"
```
MikeMcQuaid: Why do we need `ginstall`? Regardless, this should be reported upstream.

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    ENV.deparallelize
```
MikeMcQuaid: This should be reported upstream

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    system "make", "LIBTOOL=glibtool"
```
MikeMcQuaid: May be worth doing `ENV["LIBTOOL"] = "glibtool"` to avoid repetition

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "make", "LIBTOOL=glibtool", "PREFIX=#{prefix}", "install"
```
MikeMcQuaid: Does this `make install` work without running `make` before it?

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+    (pkgshare/"test").install Dir["test/*"]
+  end
+
+  test do
+    system "#{bin}/sha1dcsum #{pkgshare}/test/shattered-1.pdf | fgrep -q '*coll*'"
```
MikeMcQuaid: Use `assert_match` and `shell_output` instead of `fgrep`

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "#{bin}/sha1dcsum #{pkgshare}/test/shattered-2.pdf | fgrep -q '*coll*'"
```
MikeMcQuaid: Use `assert_match` and `shell_output` instead of `fgrep`

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "#{bin}/sha1dcsum_partialcoll #{pkgshare}/test/sha1_reducedsha_coll.bin | fgrep -q '*coll*'"
```
MikeMcQuaid: Use `assert_match` and `shell_output` instead of `fgrep`

MikeMcQuaid (changes requested):

Formula/sha1dc.rb
```diff
@@ -0,0 +1,36 @@
...
+  # The "master" branch is unusably broken and behind the "simplified_c90" branch
+  # that's the basis for release.
+  head "https://github.com/cr-marcstevens/sha1collisiondetection.git",
+       :using => :git,
+       :branch => "simplified_c90"
+
+  devel do
+    url "https://github.com/cr-marcstevens/sha1collisiondetection/archive/development-v1.0.1.tar.gz"
+    sha256 "a414aad7ab1da30193d053ad31651d39a449c2f9f74eac777e467929d0de3c93"
+    version "dev-v1.0.1"
```
MikeMcQuaid: What does `dev-v1.0.1` mean for a version? Is this a beta of 1.0.1?

MikeMcQuaid (changes requested):

Formula/sha1dc.rb
johnhawkinson: Done.

You know, https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#add-a-test-to-the-formula
recommends `cmake` and `tinyxml2` as good examples, and I reviewed both of them and none of them use those. Can we get a better recommendation in the cookbook, please? [Happy to file a PR if I knew which one to recommend...]

johnhawkinson (commented):
Done. Some changes held back pending upstream.

Formula/sha1dc.rb
johnhawkinson: Having, queried, in https://github.com/cr-marcstevens/sha1collisiondetection/pull/19#issuecomment-284256955 an answer is offered:

> Version numbers (major.minor.patch) will be increasing and marked as stable or development. E.g. development-v1.0.1, and now stable-v1.0.2.

johnhawkinson:
I think this is now good-to-go, @MikeMcQuaid ?

Sorry for all the false starts along the way, my first time packaging a formula, and also first time on something bleeding edge with packaging issues.

johnhawkinson:
I think this is now good-to-go, @MikeMcQuaid ?

Sorry for all the false starts along the way, my first time packaging a formula, and also first time on something bleeding edge with packaging issues.

Formula/sha1dc.rb
```diff
@@ -0,0 +1,25 @@
+class Sha1dc < Formula
+  desc "Tool to detect SHA-1 collisions in files, including SHAttered"
+  homepage "https://github.com/cr-marcstevens/sha1collisiondetection"
+  url "https://github.com/cr-marcstevens/sha1collisiondetection/archive/stable-v1.0.2.tar.gz"
+  sha256 "ab1a5624d12417b5af9ffe6662546b6932c39c0e4c4bef822851a2d0e53df790"
+
+  # The "master" branch is unusably broken and behind the
+  # "simplified_c90" branch that's the basis for release.
+  head "https://github.com/cr-marcstevens/sha1collisiondetection.git",
+       :using => :git
```
DomT4: `:using => :git` is assumed here, no need to write it explicitly, unless Homebrew has changed something that I missed.

johnhawkinson: Yeah, no, leftover from earlier experimentation...fixed.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #10433, opened 2017-02-28
gtk+3 3.22.9

tschoonj:


tschoonj:
Thanks for the info, pushed a new commit for this

PR #10451, opened 2017-02-28
kibana 5.2.2

jasontedor:
This pull request bumps the version on the Kibana formula from 5.2.1 to 5.2.2.


fxcoudert:
The build fails. Maybe it needs a newer node, like `v6.9.5`?

ilovezfs:
Error is
```
Running "_build:notice" task
Fatal error: ENOENT: no such file or directory, open '/private/tmp/kibana-20170305-1870-55wef5/.node_binaries/6.9.5/linux-x64/LICENSE'
```
It "fixes" it to do
```
bash-3.2$ mkdir -p .node_binaries/6.9.5/linux-x64
bash-3.2$ cp .node_binaries/6.9.5/darwin-x64/LICENSE .node_binaries/6.9.5/linux-x64/
```
after it fails.

meteormatt:
Maybe because elastic/kibana#10344

fxcoudert:
If it's that issue, could we use the patch committed?
https://github.com/elastic/kibana/commit/3ebb3b1644f60b60811787d9d509c7c71604506b.diff

jasontedor:
I've pushed a commit that will address the build failure here.

neutric:
Thank you for your help in keeping Homebrew awesome, @jasontedor!

jasontedor:
Thank you @neutric!

PR #10456, opened 2017-03-01
llvm: Symlink OpenMP header/libs to separate directory

ts826848:
This allows other formulae to use omp.h and libomp.dylib without pulling
in other libraries installed with LLVM that conflict with system ones
(e.g. libc++.dylib)

- [] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

For [PR #2234 at Homebrew/brew](https://github.com/Homebrew/brew/pull/2234).


Formula/llvm.rb
```diff
@@ -272,6 +272,13 @@ def install
     # install llvm python bindings
     (lib/"python2.7/site-packages").install buildpath/"bindings/python/llvm"
     (lib/"python2.7/site-packages").install buildpath/"tools/clang/bindings/python/clang"
+
+    # Install symlinks for omp.h/libomp so they can be used in other formulae
+    mkdir lib/"libomp"
+    ["libgomp.dylib", "libiomp5.dylib", "libomp.dylib"].each { |libomp_dylib|
+      (lib/"libomp").install_symlink lib/"#{libomp_dylib}"
```
DomT4: The use of `install_symlink` here should make the `mkdir` above redundant.

Formula/llvm.rb
```diff
@@ -272,6 +272,13 @@ def install
     # install llvm python bindings
     (lib/"python2.7/site-packages").install buildpath/"bindings/python/llvm"
     (lib/"python2.7/site-packages").install buildpath/"tools/clang/bindings/python/clang"
+
+    # Install symlinks for omp.h/libomp so they can be used in other formulae
+    mkdir lib/"libomp"
+    ["libgomp.dylib", "libiomp5.dylib", "libomp.dylib"].each { |libomp_dylib|
```
MikeMcQuaid: Check `brew style`: https://bot.brew.sh/job/Homebrew%20Core/17250/version=sierra/testReport/junit/brew-test-bot/sierra/style_llvm/

I'd also use just `|dylib|` here.

Formula/llvm.rb
```diff
@@ -272,6 +272,13 @@ def install
     # install llvm python bindings
     (lib/"python2.7/site-packages").install buildpath/"bindings/python/llvm"
     (lib/"python2.7/site-packages").install buildpath/"tools/clang/bindings/python/clang"
+
+    # Install symlinks for omp.h/libomp so they can be used in other formulae
```
MikeMcQuaid: `# Install symlinks for OpenMP header and libraries so they can be used by superenv`

ilovezfs:
I'm not sure I understand why this isn't an upstream bug if we think this is necessary.

Formula/llvm.rb
aw1621107: Should `lib/"#{dylib}"` be replaced with `lib/dylib.to_s`, or does `dylib.to_s` need parentheses?

aw1621107:
@ilovezfs Sorry, I'm not quite sure what you're getting at. Could you elaborate?

Formula/llvm.rb
MikeMcQuaid: `lib/dylib` works without the `.to_s` as these are strings as-is.

MikeMcQuaid:
@ilovezfs I think so long as we build and install libc++ but want to keep using our own one: it's not their bug.

Formula/llvm.rb
aw1621107: Oh, ok. Fixed!

ilovezfs:
Sorry but doing this looks like a mistake to me.

ilovezfs (changes requested):
I don't think we should be making changes to the llvm formula to support superenv. This should be handled in superenv itself if it's necessary.

MikeMcQuaid:
> I don't think we should be making changes to the llvm formula to support superenv. This should be handled in superenv itself if it's necessary.

@ilovezfs How? You can't add a library or include directory path without opening up everything else in that directory (which we don't want to do as we don't want to use the libc++ there).

ilovezfs:
If you have "-L#{MacOS.sdk_path}/usr/lib -L/usr/local/opt/llvm/lib" it should prefer the brew libc++.

MikeMcQuaid:
> If you have "-L#{MacOS.sdk_path}/usr/lib -L/usr/local/opt/llvm/lib" it should prefer the Xcode libc++.

:+1: to that approach. @aw1621107 it may be worth checking the ordering in superenv and seeing if that behaviour is as expected.

aw1621107:
Is that behavior one we can count on to not change in the future?

MikeMcQuaid:
Yes.

aw1621107:
Alright, I'll see what I can do.

JCount:
Pruning this due to the lack of a response and activity in general. Please comment if you would like to reopen and address the needed changes.

PR #10477, opened 2017-03-01
emacs: fix --with-imagemagick using imagemagick@6

### LINKED ISSUE #10453: Emacs formula actually requires imagemagick@6 to support --with-imagemagick
- [x] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

Emacs depends on imagemagick 6 rather than 7 to integrate imagemagick support. Installing with `--with-imagemagick` appears to work, except that `./configure` reports imagemagick was not found and the final build doesn't have imagemagick support (this can be checked by looking at the output of `M-x report-emacs-bug`: the list of configured features should include "IMAGEMAGICK").

I can easily get the formula to build with `brew switch imagemagick 6.9.7-3`, but it should be possible to depend on `imagemagick@6` and use it instead. Unfortunately I couldn't figure out how to do this - setting `PKG_CONFIG_PATH` works for me when I build Emacs manually (at least the `./configure` output is right), but not when I tried to modify the formula.

With some pointers on how to depend on an old version of a package including `pkg-config` (or even just the right way to debug this) I'd be happy to submit a pull request to fix this issue.

`brew config`:

```
HOMEBREW_VERSION: 1.1.10-425-gfa34aa2a2
ORIGIN: https://github.com/Homebrew/brew
HEAD: fa34aa2a21473a4e4d360e9561f8ca89f2415ded
Last commit: 2 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: f58aa629a583e8e4c15cff2661c0d2a9f80c3078
Core tap last commit: 4 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.12.0 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /usr/local/bin/ruby => /usr/local/Cellar/ruby/2.4.0/bin/ruby
Java: N/A
macOS: 10.11.6-x86_64
Xcode: 8.2.1
CLT: 8.2.0.0.1.1480973914
X11: 2.7.11 => /opt/X11
```

`brew doctor`:
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: Unbrewed header files were found in /usr/local/include.
If you didn't put them there on purpose they could cause problems when
building Homebrew formulae, and may need to be deleted.

Unexpected header files:
  /usr/local/include/python2.7/greenlet/greenlet.h
  /usr/local/include/python3.5m/greenlet/greenlet.h
  /usr/local/include/python3.6m/greenlet/greenlet.h

Warning: Some directories in your path end in a slash.
Directories in your path should not end in a slash. This can break other
doctor checks. The following directories should be edited:
  /Users/tchajed/code/sw/coq/bin/
```

### PR DESCRIPTION
**Pull request by tchajed**

-----

Now `brew install --with-cocoa --with-imagemagick@6 emacs` installs Emacs with support for imagemagick.

Closes #10453

### DIFF
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -39,7 +39,7 @@ class Emacs < Formula
   depends_on "dbus" => :optional
   depends_on "gnutls" => :optional
   depends_on "librsvg" => :optional
-  depends_on "imagemagick" => :optional
+  depends_on "imagemagick@6" => :optional
   depends_on "mailutils" => :optional
 
   def install
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -71,11 +71,15 @@ class Emacs < Formula
       args << "--without-gnutls"
     end
 
+    # Note that if ./configure is passed --with-imagemagick but can't find the
+    # library it does not fail but imagemagick support will not be available.
+    # See: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24455
     if build.with? "imagemagick@6"
       args << "--with-imagemagick"
     else
       args << "--without-imagemagick"
     end
+
     args << "--with-modules" if build.with? "modules"
     args << "--with-rsvg" if build.with? "librsvg"
     args << "--without-pop" if build.with? "mailutils"
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -34,6 +34,7 @@ class Emacs < Formula
   deprecated_option "cocoa" => "with-cocoa"
   deprecated_option "keep-ctags" => "with-ctags"
   deprecated_option "with-d-bus" => "with-dbus"
+  deprecated_option "imagemagick" => "imagemagick@6"
 
   depends_on "pkg-config" => :build
   depends_on "dbus" => :optional
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -40,6 +40,8 @@ class Emacs < Formula
   depends_on "dbus" => :optional
   depends_on "gnutls" => :optional
   depends_on "librsvg" => :optional
+  # Emacs does not support ImageMagick 7:
+  # Reported on 2017-03-04: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25967
   depends_on "imagemagick@6" => :optional
   depends_on "mailutils" => :optional
 
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -71,7 +71,11 @@ class Emacs < Formula
       args << "--without-gnutls"
     end
 
-    args << "--with-imagemagick" if build.with? "imagemagick"
+    if build.with? "imagemagick@6"
+      args << "--with-imagemagick"
+    else
+      args << "--without-imagemagick"
+    end
     args << "--with-modules" if build.with? "modules"
     args << "--with-rsvg" if build.with? "librsvg"
     args << "--without-pop" if build.with? "mailutils"

### CODE REVIEW

**Review by ilovezfs — DISMISSED** [review]
Please add
```
deprecated_option "imagemagick" => "imagemagick@6"
```

Also, this problem will need to be reported upstream with a link to the report in the formula.

**Comment by ilovezfs** [comment]
Also,
```
 args << "--with-imagemagick" if build.with? "imagemagick"
```
needs to be
```
 args << "--with-imagemagick" if build.with? "imagemagick@6"
```

tchajed:
I'm not sure if Emacs will consider this a bug: the `./configure` script searches for `pkg-config Wand`, which isn't provided by imagemagick 7 (it only has MagickWand). As far as Emacs is concerned the appropriate version of imagemagick isn't installed.

Maybe we can link to [Emacs bug 24455](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24455): `./configure --with-imagemagick` does not error out if imagemagick isn't found? That would fix the issue in the sense that the failure wouldn't have been silent.

ilovezfs:
The issue in 24455 is indeed a concern, so good find there, but I'd like to hear what the imagemagick 7 plan is before we start adding versioned dependencies in perpetuity.

tchajed:
Passing `--with-imagemagick` is actually an unrelated bug: `--with-imagemagick` is the default.

I've fixed this as well, but could make it a separate pull request if desired.

ilovezfs:
I don't think it's unrelated in that even if it's the default, an explicit `--with-foo` should cause configure to fail if the build plan is going to actually end up being to build without foo.

tchajed:
It might be not super concerning that Emacs depends on imagemagick 6, since imagemagick has agreed to support that version. The [last release](https://legacy.imagemagick.org/script/changelog.php) of imagemagick 6 was on Feb 18, 2017 (!), even though they started version 7 in 2011.

In any case it doesn't look like the bug tracker has anything about switching (https://debbugs.gnu.org/cgi/pkgreport.cgi?package=emacs;include=subject%3Aimagemagick). I found this post on `emacs-devel` indicating that some emacs devs think ImageMagick 7 is too buggy to switch to and would prefer version 6.

I agree it would be nice if Emacs upgraded, but for now it's probably fine.

ilovezfs:
>In any case it doesn't look like the bug tracker has anything about switching 

Please create one.

tchajed:
Fair enough that --with-imagemagick being passed all along is a bug: the real issue seems to be that `./configure` doesn't fail.

I'll report an Emacs bug about switching to imagemagick 7 soon.

zmwangx:
> The last release of imagemagick 6 was on Feb 18, 2017 (!), even though they started version 7 in 2011.

FYI that's not true. IMv7 was released in January 2016. When you read IM's changelog, among all the entries with a common version number, only the latest one indicates the actual release.

tchajed:
Oh I see - the changelog starts from when it forked off, but the first release of IMv7 took a few years.

ilovezfs:
@tchajed thank you for filing the bug report.

ilovezfs:
@zmwangx any thoughts on how to handle the case where someone passes --with-imagemagick@6 but doesn't pass --with-cocoa, since that seems to result in no imagemagick support?

zmwangx:
> any thoughts on how to handle the case where someone passes --with-imagemagick@6 but doesn't pass --with-cocoa, since that seems to result in no imagemagick support?

I personally don't use an IM-capable Emacs (actually I use Emacs in a tty 99.9% of the time), but AFIAK IM support in Emacs is only useful for displaying images. Without `--with-cocoa` (or the historical `--with-x11`), Emacs is only going to work in ttys; what's the point of IM support in that case?

By the way, in response to

> I'd like to hear what the imagemagick 7 plan is before we start adding versioned dependencies in perpetuity.

There's a patch or two floating around on the mailing list without much interest, justifiably so because the whole world isn't moving to IMv7 (just look at any popular server distro).

Even if a patch is accepted tomorrow, it won't make it into Emacs 25.2 (which is already on rc2), and given Emacs's release cycle, it will be many months before IMv7 is supported in a stable.

ilovezfs:
Merged. Thanks @tchajed

tchajed:
Sweet, thanks for your help!

ilovezfs:
You're welcome

PR #10488, opened 2017-03-01
uniutils 2.27 (new formula)

onlynone:

-----

### DIFF
--- a/Formula/uniutils.rb
+++ b/Formula/uniutils.rb
@@ -1,0 +1,43 @@
+# coding: utf-8
+
+class Uniutils < Formula
+  desc "Manipulate and analyze Unicode text"
+  homepage "https://billposer.org/Software/unidesc.html"
+  url "https://billposer.org/Software/Downloads/uniutils-2.27.tar.gz"
+  sha256 "c662a9215a3a67aae60510f679135d479dbddaf90f5c85a3c5bab1c89da61596"
+
+  # Allow build with clang. This patch was reported to debian here:
+  # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=740968
+  # And emailed to the upstream source at billposer@alum.mit.edu
+  patch :DATA
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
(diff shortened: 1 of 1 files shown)


woodruffw:
`audit` failures:

```
uniutils:
  * Stable: The URL http://billposer.org/Software/Downloads/uniutils-2.27.tar.gz could use HTTPS rather than HTTP
  * The URL http://billposer.org/Software/unidesc.html could use HTTPS rather than HTTP
```

Formula/uniutils.rb
```diff
@@ -0,0 +1,41 @@
+# coding: utf-8
+
+class Uniutils < Formula
+  desc "Manipulate and analyze Unicode text"
+  homepage "http://billposer.org/Software/unidesc.html"
+  url "http://billposer.org/Software/Downloads/uniutils-2.27.tar.gz"
+  sha256 "c662a9215a3a67aae60510f679135d479dbddaf90f5c85a3c5bab1c89da61596"
+
+  # Allow build with clang
+  patch :DATA
```
fxcoudert: Has the patch been submitted upstream? Please do so, and add a comment to a commit so it can be clear when we need to remove it later.

onlynone: The patch was sent to debian. It's unclear whether the original author ever saw it. I just emailed him and have updated the formula with a comment reflecting that.

onlynone:
I addressed those audit failures. But I'm curious why my local `brew audit --strict uniutils` didn't report anything.

woodruffw:
@onlynone The `--online` flag should enable them. They're not enabled by default (maybe they should be) because they're more time consuming.

Formula/uniutils.rb
```diff
@@ -0,0 +1,43 @@
+# coding: utf-8
```
fxcoudert: the line specifying UTF-8 encoding is unnecessary

onlynone: I think that was added by emacs automatically when I inserted the ü into the file. Editor encoding hints are something that should definitely be taken out?

fxcoudert:
Thanks @onlynone for your contribution!

PR #10498, opened 2017-03-02
gnupg2: add gnupg 2.1 versioned formula

cao:
This is a port from homebrew/versions with the following changes:

* Updated version from 2.1.18 to 2.1.19
* Removed necessary patches for 2.1.18
* Use `test` option (instead of ignoring it)
* Use `keg_only :versioned_formula` and remove conflicts_with
* Removed SHA256 for bottles (cannot build them locally)

Tested and found to be working on MacOS Sierra 10.12.3 with `--without-test`, see comment below https://github.com/Homebrew/homebrew-core/pull/10498#issuecomment-283540523.

### DIFF
--- a/Formula/gnupg2@2.1.rb
+++ b/Formula/gnupg2@2.1.rb
@@ -1,0 +1,95 @@
+class Gnupg2AT21 < Formula
+  desc "GNU Privacy Guard: a free PGP replacement"
+  homepage "https://www.gnupg.org/"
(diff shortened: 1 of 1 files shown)


cao:
This appears to be a regression upstream in computing the path to tools properly.

Specifically, it tries to access the tools in the set prefix, which does not yet exist (as we have not installed the tools there yet). Installing first, then checking works without an issue.

ilovezfs:
The plan is to upgrade gnupg itself to 2.1.x, so passing on this but thanks for the PR nonetheless.

PR #10500, opened 2017-03-02
bit: migrate to boneyard

JCount:
Currently bit tries to automatically update daily after install
and there is apparently no way to change this potentially insecure
behavior that is contrary to Homebrew's version-based release
system and general security principles.

This problematic autoupdating mechanism was identified in the past,
but was mistakenly believed to have been resolved in the
corresponding PR. However, it was not, and until measures are taken
to address this issue it would be inappropriate to continue
distributing this software in its current state.


eyalse:
@JCount There is one HTTPS call, that is triggered when the user uses the bit cli within a day interval.
This call's purpose is to check if there is a new version available via brew, prompting the user to ```brew update bit```.

I would be happy to fix any issue that is not aligned with Homebrew's version-based release system and general security principles.

JCount:
@eyalse First of all I must apologize, my commit message was overly harsh. I misinterpreted `bit`'s actions and misread some of your update code. However, the daily check, as it is currently written, seems as if it ends up serving no real purpose since `bit` won't get updated. The reason for this is that `brew upgrade bit` will only serve to upgrade `bit` if the formula has been updated and `brew update` run as well. Therefore, this mechanism would only make sense if the formula were being continually kept up-to-date. In terms of versioning, I am not certain what to say since versioning is used but in a fairly opaque manner.

eyalse:
@JCount No worries and thanks for being crystal clear. As you've mentioned there is a path where calling isLatestVersion ends up serving no real purpose but my opinion is that giving the user the ability to choose wether he'd like to be up to date or not does serve a real purpose as we'll be updating the formula on each release, thus  the call serves a meaningful purpose of prompting the user that there is a new version (if and only if there is one). What would make sense applying here in your opinion?

JCount:
@eyalse If you integrate updating the formula into your workflow whenever you cut a new release, then it does seem reasonable to leave an update check in `bit` since not all Homebrew users regularly run `brew update`, which I can see would potentially leave your users with outdated installations. It also might make sense to add a configuration option that would disable this behavior, but this may be useless or even detrimental to your users, I would not know. What may be handy, however, is if there were a CLI flag that would disable the automatic update behavior for that execution of `bit`. The reason, and desire, for this is predominantly for CI and testing. Previously, `bit`'s auto-update mechanism's somewhat unpredictable behavior was producing erroneous failures on the CI. Therefore, having some way of temporarily disabling the update mechanism be helpful to Homebrew, and perhaps others as well.

Also, to make the auto-update more effective, or rather fix a bit of a corner case, you might want to consider changing the update command to `brew update && brew upgrade bit`. This will ensure that the `bit` formula is up-to-date, even in the unlikely event that the user has set `HOMEBREW_NO_AUTO_UPDATE`.

eyalse:
@JCount added a `--skip-update` option to `bit` commands and created a [new PR](https://github.com/Homebrew/homebrew-core/pull/10688) .

fxcoudert:
For Homebrew to consider it, we need "skip update" to be the default.

ilovezfs:
I think it means `--skip-checking-for-update` as opposed to `--skip-upgrading-in-place`

JCount:
@eyalse Thank you for working with me to what I hope is a mutually satisfactory solution. I will now work on moving `bit` out of the `boneyard` and back into `homebrew/core` (using the PR that upstream created).

JCount:
Added back into `homebrew/core` in https://github.com/Homebrew/homebrew-core/commit/a6f92f98955c127ac9d2b5ed340200695a8f9ccb.

eyalse:
@JCount much obliged  🙌

PR #10532, opened 2017-03-03
wine: include dependencies

fxcoudert:
Including inside the `wine` formula all its universal dependencies (recursively), building them in the right order. `gnutls` (and its dependencies) are not included because, as far as I can tell, they do not support being built as universal libraries currently (and have not for some time).

This is work in progress: right now it builds, but it fails to locate several libraries (including `libpng` and `freetype`) at runtime. Comments or help are welcome.

### DIFF
--- a/Formula/wine.rb
+++ b/Formula/wine.rb
@@ -492,9 +492,9 @@ class Wine < Formula
   end
 
   test do
-    assert_equal shell_output("#{bin}/wine hostname.exe 2>/dev/null"), shell_output("hostname")
+    assert_equal shell_output("hostname"), shell_output("#{bin}/wine hostname.exe 2>/dev/null")
     if build.with? "win64"
-      assert_equal shell_output("#{bin}/wine64 hostname.exe 2>/dev/null"), shell_output("hostname")
+      assert_equal shell_output("hostname"), shell_output("#{bin}/wine64 hostname.exe 2>/dev/null")
     end
   end
 end
--- a/Formula/wine.rb
+++ b/Formula/wine.rb
@@ -25,12 +25,6 @@ class Wine < Formula
     sha256 "bfb9abf63691c93df28d9599aaa866dc2b4e27209b3b7b546df8a37d7d9d1e6e"
   end
 
-  # note that all wine dependencies should declare a --universal option in their formula,
-  # otherwise homebrew will not notice that they are not built universal
-  def require_universal_deps?
-    MacOS.prefer_64_bit?
-  end
-
   if MacOS.version >= :el_capitan
     option "without-win64", "Build without 64-bit support"
     depends_on :xcode => ["8.0", :build] if build.with? "win64"
@@ -40,15 +34,8 @@ class Wine < Formula
   # between them. But if you really want to build without X11, you can.
   depends_on :x11 => :recommended
   depends_on "pkg-config" => :build
-  depends_on "freetype"
-  depends_on "jpeg"
-  depends_on "libgphoto2"
-  depends_on "little-cms2"
-  depends_on "libicns"
-  depends_on "libtiff"
-  depends_on "sane-backends"
-  depends_on "gnutls"
-  depends_on "libgsm" => :optional
+  depends_on "cmake" => :build
+  depends_on "makedepend" => :build
 
   # Patch to fix texture compression issues. Still relevant on 2.0.
   # https://bugs.winehq.org/show_bug.cgi?id=14939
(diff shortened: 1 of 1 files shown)


ilovezfs:
Can we build it statically like we're doing with `raine`? That should automatically make the finding-them-at-runtime issue be moot, and will make it faster.

ilovezfs:
Looks like there's a `brew audit --online` bug that caused this to get stuck on CI for 6 hours trying to run that command.

Formula/wine.rb
```diff
@@ -65,49 +52,406 @@ def require_universal_deps?
...
+      ENV["MAKEFLAGS"] = saved_makeflags
+      ENV["HOMEBREW_OPTFLAGS"] = saved_homebrew_optflags
+    end
+  end
 
   def install
-    if build.with? "win64"
-      args64 = ["--prefix=#{prefix}"]
-      args64 << "--enable-win64"
+    mkdir_p "#{prefix}/deps/bin"
```
ilovezfs: You can use `ENV.prepend_create_path` instead.

Formula/wine.rb
```diff
@@ -65,49 +52,406 @@ def require_universal_deps?
...
+    ENV.prepend_path "PATH", "#{prefix}/deps/bin"
```
ilovezfs: Let's use `libexec` not `deps`

Formula/wine.rb
```diff
@@ -65,49 +52,406 @@ def require_universal_deps?
...
+    mkdir_p "#{prefix}/deps/lib/pkgconfig"
+    ENV.prepend_path "PKG_CONFIG_PATH", "#{prefix}/deps/lib/pkgconfig"
```
ilovezfs: `ENV.prepend_create_path` again

Formula/wine.rb
```diff
@@ -65,49 +52,406 @@ def require_universal_deps?
...
+        end
+      end
+    end
+
+    # Help wine find our libraries at runtime
+    args_soname = ["ac_cv_lib_soname_png=#{prefix}/deps/lib/libpng16.16.dylib",
+                   "ac_cv_lib_soname_tiff=#{prefix}/deps/lib/libtiff.5.dylib",
+                   "ac_cv_lib_soname_jpeg=#{prefix}/deps/lib/libjpeg.8.dylib",
+                   "ac_cv_lib_soname_sane=#{prefix}/deps/lib/libsane.1.dylib",
+                   "ac_cv_lib_soname_freetype=#{prefix}/deps/lib/libfreetype.6.dylib"]
```
ilovezfs: I think you can do
```
    %w[freetype jpeg png sane tiff].each do |dep|
      ENV["ac_cv_lib_soname_#{dep}"] = (libexec/"lib/lib#{dep}.dylib").realpath
    end
```

Formula/wine.rb
```diff
@@ -134,6 +478,16 @@ def caveats
     s
   end
 
+  def post_install
+    # For fontconfig
+    ohai "Regenerating font cache, this may take a while"
+    system "#{prefix}/deps/bin/fc-cache", "-frv"
```
ilovezfs: Are we sure this doesn't need to be done before the wine build?

ilovezfs (dismissed):

Formula/wine.rb
```diff
@@ -65,49 +52,406 @@ def require_universal_deps?
   end
 
   resource "gecko" do
-    url "https://downloads.sourceforge.net/wine/wine_gecko-2.40-x86.msi", :using => :nounzip
-    sha256 "1a29d17435a52b7663cea6f30a0771f74097962b07031947719bb7b46057d302"
+    url "https://dl.winehq.org/wine/wine-gecko/2.47/wine_gecko-2.47-x86.msi", :using => :nounzip
+    sha256 "3b8a361f5d63952d21caafd74e849a774994822fb96c5922b01d554f1677643a"
   end
```
JCount: you also need (something like) this:
```
resource "gecko-64" do
    url "https://dl.winehq.org/wine/wine-gecko/2.47/wine_gecko-2.47-x86_64.msi"
    sha256 "c565ea25e50ea953937d4ab01299e4306da4a556946327d253ea9b28357e4a7d"
end
```
For the 64 bit build of wine, both the x86 and x86_64 versions are needed and should be placed in the same "gecko" folder. See https://wiki.winehq.org/Gecko#Installing for details. 😄

Formula/wine.rb
fxcoudert: Yes, it needs to be done before programs depending on `fontconfig` (i.e. wine) run for the first time.

Formula/wine.rb
fxcoudert: Fixed. But how come this wasn't present before? 🤔

JCount: Perhaps because no one bothered to figure out why the 64 bit version of the web engine was missing? 🤷‍♂️

fxcoudert:
Bottle building fails due to brew issue: https://github.com/Homebrew/brew/pull/1469#pullrequestreview-25123081

ilovezfs:
@DomT4 mind looking this beast over?

DomT4:
Aye, happy to look it over later today, no problem. This seems a bit light on resources compared to the current `brew deps wine` list; did some of those deps turn out to be no longer necessary and we just hadn't noticed?

ilovezfs:
@DomT4 gnutls and its deps, and libicns

fxcoudert:
- `gnutls` (and its dependencies `gmp libffi libtasn1 libunistring nettle p11-kit`) are not needed [since wine 1.3.16](https://github.com/Homebrew/homebrew-core/commit/291d53f4070fbc80e03e34d3061c6f8337a775b1) (wine now includes a separate implementation of that code using [macOS frameworks](https://github.com/wine-mirror/wine/blob/master/dlls/secur32/schannel_macosx.c)).
- `libicns` (and its dep `jasper`) are no longer needed, this dep was dropped from wine in 2011
- `libgsm` was optional and is used in 0.14% of wine installs according to analytics. Its own formula says "package is no longer developed upstream"

Formula/wine.rb
```diff
@@ -64,53 +51,421 @@ def require_universal_deps?
...
 
   resource "mono" do
-    url "https://downloads.sourceforge.net/wine/wine-mono-4.5.6.msi", :using => :nounzip
-    sha256 "ac681f737f83742d786706529eb85f4bc8d6bdddd8dcdfa9e2e336b71973bc25"
+    url "https://dl.winehq.org/wine/wine-mono/4.6.4/wine-mono-4.6.4.msi", :using => :nounzip
+    sha256 "91b7d58177b9a9355edf007dab94535471aebdddae12734ceb4a219d2ecc4152"
+  end
+
+  resource "openssl" do
```
DomT4: There's no alternative here given the universal situation but my gut feeling is that people will forget to update this `openssl` ASAP when there's a new release. That was the case where we'd vendored crypto stuff before, which isn't always an issue because not everything has `openssl`'s significant & mainstream profile.

Formula/wine.rb
```diff
@@ -64,53 +51,421 @@ def require_universal_deps?
...
+        ENV.deparallelize
+        ENV.permit_arch_flags
+
+        # OpenSSL will prefer the PERL environment variable if set over $PATH
+        # which can cause some odd edge cases & isn't intended. Unset for safety,
+        # along with perl modules in PERL5LIB.
+        ENV.delete("PERL")
+        ENV.delete("PERL5LIB")
+
+        # Load zlib from an explicit path instead of relying on dyld's fallback
```
DomT4: Unless `wine` is using TLS compression for some reason, you could probably just configure this vendored `openssl` with `no-zlib` instead of carrying a duplicate `inreplace` until the end of days.

Formula/wine.rb
```diff
@@ -64,53 +51,421 @@ def require_universal_deps?
...
+                              *depflags
+        system "make", "install"
+      end
+
+      resource("fontconfig").stage do
+        system "./configure", "--disable-dependency-tracking",
+                              "--prefix=#{libexec}",
+                              "--disable-static",
+                              "--with-add-fonts=/System/Library/Fonts,/Library/Fonts,~/Library/Fonts",
+                              "--localstatedir=#{var}",
```
DomT4: I'd probably keep `localstatedir` stuff inside `libexec` with everything else, or do something like `var/vendored_wine_fontconfig`.

It's not completely impossible that `fontconfig` changes something in future or fixes that annoying `fc-cache` bug & you `brew upgrade fontconfig` and everything is rosy, then you `brew install wine` and you've broken that stuff in `var` again.

Formula/wine.rb
```diff
@@ -64,53 +51,421 @@ def require_universal_deps?
...
+        ln_s "darwin13.h", "include/net-snmp/system/darwin14.h"
+        ln_s "darwin13.h", "include/net-snmp/system/darwin15.h"
+        ln_s "darwin13.h", "include/net-snmp/system/darwin16.h"
+
+        system "./configure", "--disable-debugging",
+                              "--prefix=#{libexec}",
+                              "--disable-static",
+                              "--enable-ipv6",
+                              "--with-defaults",
+                              "--with-persistent-directory=#{var}/db/net-snmp",
```
DomT4: Same sort of comment here as above. Not sure it's sensible to have things share these permanent directories when they're unlikely to be kept perfectly in-sync over time.

Formula/wine.rb
```diff
@@ -64,53 +51,421 @@ def require_universal_deps?
...
+      resource("sane-backends").stage do
+        save_env do
+          # Cannot have "patch do" here
+          Pathname.pwd.install resource("sane-backends-patch")
+          system "patch", "-p1", "-i", "1.0.25-missing-types.patch"
+
+          ENV.deparallelize
+          system "./configure", "--disable-dependency-tracking",
+                                "--prefix=#{libexec}",
+                                "--localstatedir=#{var}",
```
DomT4: Same note here as above again.

Formula/wine.rb
```diff
@@ -134,6 +489,16 @@ def caveats
...
+    ohai "Regenerating font cache, this may take a while"
+    system "#{libexec}/bin/fc-cache", "-frv"
+
+    # For net-snmp
+    (var/"db/net-snmp").mkpath
+    (var/"log").mkpath
+  end
+
   test do
     system "#{bin}/wine", "--version"
```
DomT4: Maybe tempting here to add a basic smoke test to ensure that `otool -L` shows `wine` linked correctly to the vendored tools in `libexec` given how large a change this is.

DomT4:
> libicns (and its dep jasper) are no longer needed, this dep was dropped from wine in 2011

This might be some kind of record for how long a dependency has stuck around in a formula after upstream stopped supporting it 😅.

Formula/wine.rb
ilovezfs: Perhaps that's something audit could address.

Formula/wine.rb
ilovezfs: Stuffing the cache into the bottle feels sketchy but using a separate cache sounds correct.

Formula/wine.rb
MikeMcQuaid: > Perhaps that's something audit could address.

Yes or ideally we have some sort of subformula situation where the URL/mirror/SHA for a resource like this just uses the OpenSSL one.

I'm less concerned about Wine specifically in this case because I'd be surprised if people are using it for as serious crypto as most other dependents.

Formula/wine.rb
MikeMcQuaid: > using a separate cache sounds correct.

👍

Formula/wine.rb
MikeMcQuaid: Ideally I think we'd just use a `resource` for some downloaded command-line `.exe` that we actually run.

Formula/wine.rb
ilovezfs: That doesn't actually itself resolve it as an audit actually would since it would only help if someone rebuilt wine as opposed to prompting them to do so.

Formula/wine.rb
ilovezfs: I think it's perfectly possible for wine dependents to need TLS so I don't think I agree with this.

Formula/wine.rb
ilovezfs: Yes. Note that except for a couple of the deps, it's not actually linked to specific paths since it uses dlopen.

Formula/wine.rb
DomT4: That's unrelated. TLS compression isn't mandatory for TLS, and indeed, if you're using it you might as well stick your hand in a fire (See `CRIME`, etc) 😅.

Like `ssl3` it's something that should have been disabled in the `openssl` formula a fair while ago, but nobody has any desire to revision 350+ formulae to cover any potential assumptions built into other formulae about `zlib` being available, so it has stuck around like a nasty cold.

Formula/wine.rb
DomT4: > Ideally I think we'd just use a resource for some downloaded command-line .exe that we actually run.

Aye, was just thinking locating a "legit" `.exe` that's likely to be around for a long time might be a pain, but probably not that much of a pain I guess.

Formula/wine.rb
ilovezfs: OK, but if it breaks someone's favorite game, I'm sending them to you.

Formula/wine.rb
fxcoudert: Wine actually comes with some tools that we can test. The problem is, the first time they are run `wine` will want to create its config and display a GUI telling us to wait (and no flag to turn it off). So I don't know if the CI will like it :)

fxcoudert:
Thanks a lot @ilovezfs @DomT4 @MikeMcQuaid for your comments! Waiting now for the CI one last time…

ilovezfs (approved):
LGTM. Feel free to ship when :green_heart:

fxcoudert:
Fixed by ca2b723507b79e91d9fcede838859b61058dd991

vszakats:
First of all thanks for this, I'm looking forward migrating from the wine-devel cask.

One remaining concern here, and sorry for hijacking a closed PR, but I had to install the formula to notice. The infamous fontconfig cache regeneration is taking place as a post-install task and this may be a problem when installing the formula on CI machines. These are headless and certainly not GUI, so fonts should not matter. The regeneration process however takes several minutes, which reduces the useful time to run the actual build and adds an uncomfortable wait to each CI session.

I had committed a patch to address this problem to the MSYS2 project here: https://github.com/Alexpux/MINGW-packages/commit/fdea2f97bef6a728c2df6ec287fff2b032f6d25f
It solved the problem properly.

Do you think it would be feasible to implement a similar setting to the `wine` formula? I'm unsure about the best way, but maybe an `--without-fc-cache` or similar option would do. Or an envvar, like in MSYS2.

Any opinions?

PR #10556, opened 2017-03-03
sslscan: add option to statically link OpenSSL

hanazuki:

-----

This patches adds `--with-static-openssl` build option to the sslscan formula, which makes sslscan be statically linked with a custom-build OpenSSL with weak cipher suites enabled, usually disabled in OpenSSL provided by system or installed by Homebrew.
It is useful when scanning servers to detect if a weak cipher is enabled.


ilovezfs (changes requested):
I made some tweaks and am sending it through CI with this as the default. Then I'll switch it back to being optional before pulling.

DomT4:
You might as well make the insecure static OpenSSL the default here IMO. Even a recent 1.0.2 release from outside Homebrew has `SSLv2` disabled, which you may still want/need to check for.

ilovezfs:
>You might as well make the insecure static OpenSSL the default here IMO. Even a recent 1.0.2 release from outside Homebrew has SSLv2 disabled, which you may still want/need to check for.

On Slack to @JCount "8:51]  now the question is whether that should actually be the default"

:)

DomT4:
👍 Beyond that, you might want to consider the horror of installing 1.0.2f and pinning it there. That was [the last OpenSSL](https://www.openssl.org/news/secadv/20160301.txt) to by-default support `SSLv2` and as such wouldn't need further configuration to get you the weakest `openssl` possible, whilst still also being able to check for TLSv1.1 and above (which the system OpenSSL cannot do).

>  OpenSSL 1.0.2g and 1.0.1s deploy the following mitigation against DROWN: SSLv2 is now by default disabled at build-time. Builds that are not configured with "enable-ssl2" will not support SSLv2. Even if "enable-ssl2" is used, users who want to negotiate SSLv2 via the version-flexible SSLv23_method() will need to explicitly call either of: SSL_CTX_clear_options(ctx, SSL_OP_NO_SSLv2); or SSL_clear_options(ssl, SSL_OP_NO_SSLv2); as appropriate. Even if either of those is used, or the application explicitly uses the version-specific SSLv2_method() or its client or server variants, SSLv2 ciphers vulnerable to exhaustive search key recovery have been removed. Specifically, the SSLv2 40-bit EXPORT ciphers, and SSLv2 56-bit DES are no longer available. In addition, weak ciphers in SSLv3 and up are now disabled in default builds of OpenSSL. Builds that are not configured with "enable-weak-ssl-ciphers" will not provide any "EXPORT" or "LOW" strength ciphers. Reported by Nimrod Aviram and Sebastian Schinzel on 29th December 2015.

ilovezfs:
@DomT4 I wonder if we should use peter's fork here too

DomT4:
@ilovezfs If you did I would probably declare that in caveats. I don't think Peter's fork has been audited by anyone externally and Homebrew using it is essentially vouching that nothing has been snuck into that codebase. I've used Peter's fork before for testing, but I'd feel more reluctant to make that decision of trust for everyone.

ilovezfs:
@DomT4 sure, though we're already using it in sslyze currently.

DomT4:
@ilovezfs Aye, I know. That's just where I am on these sorts of things; using unofficial crypto forks always makes me nervous 😄. But ultimately if anything negative happens the responsibility will fall on `homebrew/core` maintainers so if you're comfortable the chances of that are significantly small enough that's a choice I'll respect. They who take the risk make the decisions and all that.

ilovezfs:
scape:goat:

DomT4:
With great power comes being shouted at by a hundred angry people if you break their use case/etc 😭

JCount:
@hanazuki I am closing this in favor of https://github.com/Homebrew/homebrew-core/pull/13439, but thank you very much for your contribution nonetheless. 😄

PR #10559, opened 2017-03-03
Import Redis formulae for versions 2.8, 2.6, 1.3

DaxServer:


SrihariThalla:
Homebrew versions PR - https://github.com/Homebrew/homebrew-versions/pull/1537

Formula/redis@2.6.rb
```diff
@@ -0,0 +1,73 @@
...
+  bottle do
+    cellar :any
+    sha256 "1c0c9565f242a1e34d171950a341e2f38372e49cc5420162c73cf2c39ef2c682" => :yosemite
+    sha256 "f4abb52ec0ef232b85aae9c0e3bbc6cfdfca3e43daeef38e46f3eaeb35644b45" => :mavericks
+    sha256 "4d936a5010e9b50f1c9c347192f11e0b8e365a1948455f6b27415d1a28e35b50" => :mountain_lion
+  end
+
+  keg_only :versioned_formula
+
+  conflicts_with "redis",
```
SrihariThalla: I don't think this is necessary?

Formula/redis@2.6.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def install
+    # Architecture isn't detected correctly on 32bit Snow Leopard without help
+    ENV["OBJARCH"] = MacOS.prefer_64_bit? ? "-arch x86_64" : "-arch i386"
+
+    # Head and stable have different code layouts
+    src = (buildpath/"src/Makefile").exist? ? buildpath/"src" : buildpath
+    system "make", "-C", src, "CC=#{ENV.cc}"
+
+    %w[benchmark cli server check-dump check-aof].each { |p| bin.install src/"redis-#{p}" }
```
SrihariThalla: I guess `redis@2.6`?

Formula/redis@2.8.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def install
+    # Architecture isn't detected correctly on 32bit Snow Leopard without help
+    ENV["OBJARCH"] = MacOS.prefer_64_bit? ? "-arch x86_64" : "-arch i386"
+
+    # Head and stable have different code layouts
+    src = (buildpath/"src/Makefile").exist? ? buildpath/"src" : buildpath
+    system "make", "-C", src, "CC=#{ENV.cc}"
+
+    %w[benchmark cli server check-dump check-aof sentinel].each { |p| bin.install src/"redis-#{p}" => "redis28-#{p}" }
```
SrihariThalla: I guess `redis@2.8 instead of `redis28`? And will it affect existing installations?

Formula/redis@2.6.rb
rakkesh: This is fine in it's current form. It would not affect existing installations.

`%w[benchmark cli server check-dump check-aof].each { |p| bin.install src/"redis-#{p}" }` would create executables `redis-benchmark redis-cli redis-server redis-check-dump redis-check-aof` in `/usr/local/Cellar/redis@2.6/2.6.17/bin` and then symlink them to `/usr/local/opt/redis@2.6/2.6.17/bin`.

Formula/redis@2.6.rb
```diff
@@ -0,0 +1,73 @@
+class Redis26 < Formula
```
rakkesh: Versioned formula name should be `RedisAT26`.

Formula/redis@2.6.rb
rakkesh: `conflicts_with` is not necessary as this formula is `keg_only :versioned_formula`

Formula/redis@2.8.rb
rakkesh: Check my other comment made on `redis@2.6`.

`%w[benchmark cli server check-dump check-aof sentinel].each { |p| bin.install src/"redis-#{p}"`

formula_renames.json
```diff
@@ -78,6 +78,9 @@
   "polarssl": "mbedtls",
   "protobuf250": "protobuf@2.5",
   "protobuf260": "protobuf@2.6",
+  "redis26": "redis@2.6",
+  "redis28": "redis@2.8",
+  "redis1310": "redis@1.3",
   "qt55": "qt@5.5",
   "racket": "minimal-racket",
```
rakkesh: Ideally, `redis` versions should go after `racket`.

Formula/redis@2.8.rb
```diff
@@ -0,0 +1,74 @@
...
+      s.gsub! "/var/run/redis.pid", "#{var}/run/redis-2.8.pid"
+      s.gsub! "dir ./", "dir #{var}/db/redis28/"
+      s.gsub! "\# bind 127.0.0.1", "bind 127.0.0.1"
+    end
+
+    etc.install "redis.conf" => "redis28.conf" unless (etc/"redis28.conf").exist?
+    etc.install "sentinel.conf" => "redis28-sentinel.conf" unless (etc/"redis28-sentinel.conf").exist?
+  end
+
+  plist_options :manual => "redis28-server #{HOMEBREW_PREFIX}/etc/redis28.conf"
```
rakkesh: You could use `"#{HOMEBREW_PREFIX}/opt/redis@28/bin/redis-server` and eliminate the need to rename the executables.

formula_renames.json
SrihariThalla: Yeah! Not sure what I was thinking while writing them..!

Formula/redis@2.6.rb
```diff
@@ -0,0 +1,73 @@
...
+  url "http://download.redis.io/releases/redis-2.6.17.tar.gz"
+  sha256 "5a65b54bb6deef2a8a4fabd7bd6962654a5d35787e68f732f471bbf117f4768e"
+  head "https://github.com/antirez/redis.git", :branch => "2.6"
+
+  bottle do
+    cellar :any
+    sha256 "1c0c9565f242a1e34d171950a341e2f38372e49cc5420162c73cf2c39ef2c682" => :yosemite
+    sha256 "f4abb52ec0ef232b85aae9c0e3bbc6cfdfca3e43daeef38e46f3eaeb35644b45" => :mavericks
+    sha256 "4d936a5010e9b50f1c9c347192f11e0b8e365a1948455f6b27415d1a28e35b50" => :mountain_lion
+  end
```
MikeMcQuaid: Remove all bottle blocks.

MikeMcQuaid (changes requested):

Formula/redis@2.6.rb
SrihariThalla: Sure. I thought HomebrewBot would update them?

Formula/redis@1.3.rb
```diff
@@ -0,0 +1,65 @@
+class Redis1310 < Formula
+  desc "Persistent key-value database, with built-in net interface"
+  homepage "http://redis.io/"
+  url "https://github.com/antirez/redis/archive/v1.3.12.tar.gz"
+  sha256 "326d8866317f7bcaafba80409b847700620025dbd34e638f85f05bc1c13beabd"
+  head "https://github.com/antirez/redis.git", :branch => "unstable"
```
SrihariThalla: I couldn't find the branch for `1.3` and so I have kept the default branch for the repo. Please advice!

Formula/redis@1.3.rb
```diff
@@ -0,0 +1,65 @@
...
+        <key>WorkingDirectory</key>
+        <string>#{var}</string>
+        <key>StandardErrorPath</key>
+        <string>#{var}/log/redis.log</string>
+        <key>StandardOutPath</key>
+        <string>#{var}/log/redis.log</string>
+      </dict>
+    </plist>
+        EOS
+  end
```
MikeMcQuaid: Mind adding a simple `system "#{bin}/redis28-server", "--version"`-like test? Thanks ❤️

MikeMcQuaid (changes requested):

Formula/redis@1.3.rb
SrihariThalla: Thanks, Mike! This formula adds a `redis-server` instead of `redis28` like prefix. So, I have added `redis-server`.

I didn't want to change how these were added previously in the https://github.com/Homebrew/homebrew-versions/pull/1537. If `redisXX-server` is preferred, let me know. I'll update all the files.

MikeMcQuaid (approved):

ilovezfs:
Should we do just `redis@1`?

MikeMcQuaid:
@ilovezfs I reckon given we're using double digits for the two series it's worth doing the same for the first too.

ilovezfs:
That seems peculiar. gcc@5 ...

ilovezfs:
Either way: it should be standardized. gcc@5 should be changed to gcc@5.4 (yuck) or this one changed to redis@1 (less yuck).

MikeMcQuaid:
I don't agree as it depends on the versioning scheme; GCC changed theirs with GCC5 onwards whereas Redis didn't.

ilovezfs:
So we will have no consistency other than consulting the oracle.

ilovezfs:
Also: I think we actually only should take 2.8.24 (last release in 2.8 series was 2.8.24 not 2.8.21), and we should not take 2.6 or 1.3.

Regarding 1.3: it's not even a stable release. According to upstream, "An even minor marks a stable release, like 1.2, 2.0, 2.2, 2.4, 2.6, 2.8. Odd minors are used for unstable releases, for example 2.9.x releases are the unstable versions of what will be Redis 3.0 once stable."

Also, according to analytics, homebrew/versions/redis1310 had 1 install in the last 30 days.

Regarding 2.6: homebrew/versions/redis26 had 12 installs in the last 30 days, whereas homebrew/versions/redis28 had 121 in the last 30 days.

For context: redis, the main formula in core, had 43,549 in the last 30 days.

MikeMcQuaid:
> consulting the oracle.

`s/oracle/upstream versioning scheme`

> Regarding 1.3: it's not even a stable release. According to upstream, "An even minor marks a stable release, like 1.2, 2.0, 2.2, 2.4, 2.6, 2.8. Odd minors are used for unstable releases, for example 2.9.x releases are the unstable versions of what will be Redis 3.0 once stable."
> 
> Also, according to analytics, homebrew/versions/redis1310 had 1 install in the last 30 days.

Let's punt on 1.3 👍

ilovezfs:
>s/oracle/upstream versioning scheme

Just make sure to keep your divining rod handy.

ilovezfs:
>Let's punt on 1.3

Why are we keeping 2.6? 12 installs vs. 121 (2.8) and 43,549 (3.0).

Does it need to drop to 8?

MikeMcQuaid:
> Does it need to drop to 8?

Needs to drop to 0️⃣ 😉 (or, easier, just break).

ilovezfs:
:roll_eyes:

MikeMcQuaid:
@SrihariThalla Just for future reference we always want only one formula changed per commit so I've amended that and dropped Redis 1.3.

Formula/redis.rb
```diff
@@ -1,6 +1,6 @@
 class Redis < Formula
   desc "Persistent key-value database, with built-in net interface"
-  homepage "http://redis.io/"
+  homepage "https://redis.io/"
   url "http://download.redis.io/releases/redis-3.2.8.tar.gz"
   sha256 "61b373c23d18e6cc752a69d5ab7f676c6216dc2853e46750a8c4ed791d68482c"
   head "https://github.com/antirez/redis.git", :branch => "unstable"
```
SrihariThalla: I believe the branch should be [3.2](https://github.com/antirez/redis/tree/3.2) ?

formula_renames.json
```diff
@@ -93,6 +93,10 @@
   "protobuf260": "protobuf@2.6",
   "qt55": "qt@5.5",
   "racket": "minimal-racket",
+  "redis26": "redis@2.6",
+  "redis28": "redis@2.8",
+  "qt55": "qt@5.5",
```
SrihariThalla: Duplicated!

formula_renames.json
```diff
@@ -93,6 +93,10 @@
...
+  "racket": "minimal-racket",
```
SrihariThalla: Duplicated!

SrihariThalla:
@MikeMcQuaid Sure, I'll keep that in mind for further commits! And where can I check analytics (installs counter)?

Formula/redis.rb
MikeMcQuaid: We tend to leave this at `unstable` so it stays the same on version bumps.

formula_renames.json
MikeMcQuaid: Fixed when I `brew pull`d, thanks!

formula_renames.json
MikeMcQuaid: Fixed when I `brew pull`d, thanks!

MikeMcQuaid:
> And where can I check analytics (installs counter)?

Unfortunately this isn't public yet but I hope to implement that in future.

MikeMcQuaid:

SrihariThalla:
Worked like a charm! (Although I had to put in `PATH` and start the service manually!)

screenshot 2017-03-13 16 50 30

I have a question:

1. Since I had to put the Redis into `PATH` manually; I understand that not everyone wants everything in `PATH`, but I do. Can I expect that to be done automatically by Homebrew, in a foreseeable future?

MikeMcQuaid:
> Since I had to put the Redis into PATH manually; I understand that not everyone wants everything in PATH, but I do. Can I expect that to be done automatically by Homebrew, in a foreseeable future?

It'll likely never be done automatically but I'm up for figuring out how to improve the messaging and/or tooling to help there. Any thoughts?

PR #10563, opened 2017-03-03
abcde: add missing dependency md5sha1sum

lsolanka:
abcde uses md5sum to calculate hash for the cddb data. On Mavericks md5sum does
not seem to be available.

### DIFF
--- a/Formula/abcde.rb
+++ b/Formula/abcde.rb
@@ -17,6 +17,7 @@ class Abcde < Formula
   depends_on "cdrtools"
   depends_on "id3v2"
   depends_on "mkcue"
+  depends_on "md5sha1sum"
   depends_on "flac" => :optional
   depends_on "lame" => :optional
   depends_on "vorbis-tools" => :optional


fxcoudert:
Thanks @lsolanka for your contribution!

ilovezfs:
I've reverted this in https://github.com/Homebrew/homebrew-core/commit/2d223464a8eb79ecf2b1268fd9b4523be64e8525. This is actually an upstream bug.

```
if [ X$(uname) = "XFreeBSD" ] ; then
        HTTPGET=fetch
        MD5SUM=md5
        NEEDCDROMID=y
        OSFLAVOUR=FBSD
elif [ X$(uname) = "XDarwin" ] ; then
        HTTPGET=curl
        OSFLAVOUR=OSX
        # We should have diskutil in OSX, but let's be sure...
        NEEDDISKUTIL=y
        CDROMREADERSYNTAX=cddafs
        # We won't find the eject program in OSX, and doing checkexec will fail further below...
        unset EJECT
elif [ X$(uname) = "XOpenBSD" ] ; then
```

The correct command is `md5` on macOS just as it is on FreeBSD.

fxcoudert:
Good catch, thanks!

JCount:
`md5` probably needs to be specified as `/sbin/md5` because `/sbin` is not generally in the **PATH**.

ilovezfs:
It's not?

lsolanka:
Indeed, sorry, originally thought that this is missing there on purpose.

ilovezfs:
@lsolanka no problem at all. Mind reporting this upstream so we can get it fixed?

lsolanka:
Yeah, tried earlier in their bugzilla but it seems to have problem registering me so I posted to the mailing list: https://lists.einval.com/pipermail/abcde-users/2017-March/000233.html

ilovezfs:
Thanks!

lsolanka:
Still think it might be worth merging this as abcde is quite unresponsive...it wouldn't do any harm would it? Up to you though, I'm not fussed.

JCount:
see https://github.com/Homebrew/homebrew-core/pull/11250

PR #10572, opened 2017-03-03
par2 0.6.14

garybernhardt:

-----

This upgrades par2. I had to update the formula to point to its new home on GitHub, which is confirmed as the official repository by the author: "Yes, this repository on github is now the primary repository for par2cmdline." from PeterBClements in https://github.com/Parchive/par2cmdline/issues/72.

### DIFF
--- a/Formula/par2.rb
+++ b/Formula/par2.rb
@@ -1,8 +1,8 @@
 class Par2 < Formula
   desc "Parchive: Parity Archive Volume Set for data recovery"
-  homepage "https://parchive.sourceforge.io/"
-  url "https://downloads.sourceforge.net/project/parchive/par2cmdline/0.4/par2cmdline-0.4.tar.gz"
-  sha256 "9e32b7dbcf7bca8249f98824757d4868714156fe2276516504cd26f736e9f677"
+  homepage "https://github.com/Parchive/par2cmdline"
+  url "https://github.com/Parchive/par2cmdline/archive/v0.6.14.tar.gz"
+  sha256 "2fd831ba924d9f0ecd9242ca45551b6995ede1ed281af79aa30e7490d5596e7a"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


Formula/par2.rb
```diff
@@ -13,39 +13,55 @@ class Par2 < Formula
...
-  end
-
-  # http://parchive.cvs.sourceforge.net/viewvc/parchive/par2-cmdline/par2creatorsourcefile.cpp?r1=1.4&r2=1.5
-  patch :p0 do
-    url "https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/app-arch/par2cmdline/files/par2cmdline-0.4-offset.patch?revision=1.2"
-    sha256 "c4820b11376c9932ece944752ddd388fb50fcbcd47aaadda073997142952d969"
-  end
+  # The tests for this version of par2cmdline use the `iflag` option to dd,
+  # which doesn't exist in OS X. `iflag` has been removed as of commit
+  # 855f3096c9f52cd0fbdb9e237c6c9624f7b90ea6, but no release has been cut since
```
fxcoudert: Can you instead use a `patch do` block with the URL (and SHA-256) of the patch?

Formula/par2.rb
```diff
@@ -13,39 +13,55 @@ class Par2 < Formula
...
+  # then.
+  patch :p1, :DATA
 
   def install
-    system "./configure", "--disable-debug", "--disable-dependency-tracking",
-                          "--prefix=#{prefix}"
+    system "aclocal"
+    system "automake", "--add-missing"
+    system "autoconf"
```
fxcoudert: Maybe worth opening an issue upstream to suggest their releases include the generated files? (avoiding to run `aclocal`, `automake` and `autoconf` in the future)

fxcoudert (changes requested):

Formula/par2.rb
garybernhardt: This patch isn't available on the web; I generated it locally to make the tests pass. It will also disappear as of the next release of par2cmdline, since the problem has already been fixed upstream; they just haven't a cut a release yet. Do you mind leaving it in as it is, since it's temporary and there's no obvious place for me to host it, being an unofficial patch?

Formula/par2.rb
garybernhardt: Their README explicitly lists these install steps, so they seem intentional. Single source of truth and all that, I think?

Formula/par2.rb
fxcoudert: You give the upstream commit URL, can we use that?

garybernhardt: Ha... I feel silly for not thinking of that. Unfortunately, that commit doesn't apply cleanly on its own. Applying 855f3096c9f52cd0fbdb9e237c6c9624f7b90ea6 first fixes the problem, but then the tests fail. I assume that continuing this and applying a third (or maybe more) commit(s) wouldn't be desirable. I don't see anything in the formula cookbook about what to do in this situation. What do you recommend?

fxcoudert (approved):

fxcoudert:
@BrewTestBot test this please

fxcoudert:
Thanks @garybernhardt for your contribution to Homebrew!

garybernhardt:
Thank you! All of the automation is really nice. `brew audit`, the automatic bottle builds, etc.

PR #10577, opened 2017-03-04
node@5: delete unsupported version

ilovezfs:


MikeMcQuaid:
👎 given these are still being used by people and aren't causing us problems.

ilovezfs:
They violate our guidelines for versioned formulae since they're not supported by upstream, and they are actively corrupting the main node installations. Other than that, they're perfectly fine.

ilovezfs:
Specifically: https://github.com/nodejs/LTS/#lts-schedule

- support for 0.12.x ended 31 Dec 2016.
- support for 0.10.x ended 31 Oct 2016.
- support for 5.x ended 23 Jun 2016.

"Please be aware that this will likely be the final official release of the 0.12 release line. As per our LTS schedule, support for 0.12 ends on December 31st 2016. Please plan your migration to a newer, supported release line." (https://nodejs.org/en/blog/release/v0.12.18/)

MikeMcQuaid:
Let's remove `node@4` given (as @ilovezfs pointed out on Slack) it's unsupported and there are unpatched CVEs. We can move all other versioned nodes to use `conflicts_with` again to get past the `npm`-overwriting situation.

PR #10585, opened 2017-03-04
[hub] Add fish completion

alloy:

-----

This links the completions for the fish shell that are added in hub 2.3.0. However, that version is still a pre-release state, so this will only work when using `--HEAD`. I guess we should wait with merging this till 2.3.0 proper is released? /cc @mislav

### DIFF
--- a/Formula/hub.rb
+++ b/Formula/hub.rb
@@ -48,6 +48,8 @@ class Hub < Formula
     if build.with? "completions"
       bash_completion.install "etc/hub.bash_completion.sh"
       zsh_completion.install "etc/hub.zsh_completion" => "_hub"
+      # TODO: Remove the conditional when hub 2.3.0 is released.
+      fish_completion.install "etc/hub.fish_completion" => "hub.fish" unless build.stable?
     end
   end
 


bfontaine:
> However, that version is still a pre-release state, so this will only work when using `--HEAD`

You can snap an `if build.head?` at the end of the line then.

alloy:
Ah sure, can do that. So then it’s up to the next person that updates the formula for the proper release to remove it again, I assume?

bfontaine:
Yes. You can add a comment above it like ``# Remove `if build.head?` when hub 2.3.0 is released``.

ilovezfs:
Won't this also work for the devel spec?

alloy:
@ilovezfs I don't know what you're referring to, can you elaborate?

ilovezfs:
https://github.com/alloy/homebrew-core/blob/1d3d5d7d13ad8eb087d6fae76fefb330d67aa1e4/Formula/hub.rb#L16-L20 which is what you get if you pass `--devel`

alloy:
Oh I see. Yeah I think it's in that release too, Will double-check and update 👍

ilovezfs:
You can use `unless build.stable?` to cover both the devel & head cases

alloy:
Done, thanks for the help 👍

bfontaine:
Thank *you* @alloy 😉

PR #10594, opened 2017-03-04
bash-completion 2.5

mernen:
Based on [homebrew-versions/bash-completion2](https://github.com/Homebrew/homebrew-versions/blob/master/bash-completion2.rb), with the following changes:

* Updated to version 2.5 (homebrew-versions was still on 2.4)
* Fixed description: it claimed to be for Bash 4.0+, but AFAICT 4.1+ has always been required
* Added `depends_on "bash"`, to ensure it’s only installed with Bash 4
* `conflicts_with` replaced with `keg_only :versioned_formula`, as per Homebrew v1.1.8

Additionally, I’ve updated the notice at the top of `bash-completion`.

-----

### DIFF
--- a/Formula/bash-completion@2.rb
+++ b/Formula/bash-completion@2.rb
@@ -1,0 +1,31 @@
+class BashCompletionAT2 < Formula
+  desc "Programmable completion for Bash 4.1+"
+  homepage "https://github.com/scop/bash-completion"
+  url "https://github.com/scop/bash-completion/releases/download/2.5/bash-completion-2.5.tar.xz"
+  sha256 "b0b9540c65532825eca030f1241731383f89b2b65e80f3492c5dd2f0438c95cf"
+  head "https://github.com/scop/bash-completion.git"
+
+  keg_only :versioned_formula
+
+  depends_on "bash"
+
+  def install
+    inreplace "bash_completion", "readlink -f", "readlink"
+
+    system "./configure", "--prefix=#{prefix}", "--sysconfdir=#{etc}"
+    ENV.deparallelize
+    system "make", "install"
+  end
+
+  def caveats; <<-EOS.undent
+    Add the following to your ~/.bash_profile:
+      if [ -f #{opt_share}/bash-completion/bash_completion ]; then
+        . #{opt_share}/bash-completion/bash_completion
+      fi
+    EOS
+  end
+
+  test do
+    system "test", "-f", "#{share}/bash-completion/bash_completion"
+  end
+end
--- a/Formula/bash-completion@2.rb
+++ b/Formula/bash-completion@2.rb
@@ -5,9 +5,9 @@ class BashCompletionAT2 < Formula
   sha256 "b0b9540c65532825eca030f1241731383f89b2b65e80f3492c5dd2f0438c95cf"
   head "https://github.com/scop/bash-completion.git"
 
-  keg_only :versioned_formula
-
   depends_on "bash"
+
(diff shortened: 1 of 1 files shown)


Formula/bash-completion@2.rb
```diff
@@ -0,0 +1,31 @@
...
+    ENV.deparallelize
+    system "make", "install"
+  end
+
+  def caveats; <<-EOS.undent
+    Add the following to your ~/.bash_profile:
+      if [ -f $(brew --prefix)/share/bash-completion/bash_completion ]; then
+        . $(brew --prefix)/share/bash-completion/bash_completion
+      fi
+    EOS
```
JCount: This is currently incorrect because `keg_only :versioned_formula` are not linked to `$(brew --prefix)/share`. It should be:
```
      if [ -f #{opt_share}/bash-completion/bash_completion ]; then
        . #{opt_share}/bash-completion/bash_completion
      fi
```

JCount (changes requested):
Also, please break the change to the `bash-completion` formula into a separate commit. Thanks.

mernen:
Done.

Formula/bash-completion.rb
```diff
@@ -1,5 +1,5 @@
 # NOTE: version 2.0 is out, but it requires Bash 4, and macOS ships
-# with 3.2.48. See homebrew-versions for a 2.0 formula.
+# with 3.2.48. If you've upgraded bash, use bash-completion@2 instead.
```
Numbermaniac: Correct me if I'm wrong, but I believe it's now 3.2.57.

```
> /bin/bash -c "echo \$BASH_VERSION"
3.2.57(1)-release
```

JCount (approved):

Formula/bash-completion.rb
JCount: It doesn't really matter since the key point is that the system bash is <4. However, I'll fix it when I merge.

mernen: Indeed! I’ve also updated the line above to just say “version 2”, since 2.0 feels outdated, and there’s little point in keeping it fully up-to-date whenever newer versions are out.

JCount:
The contents of the PR itself are ready to be merged, good job @mernen. 💯 

However, there is a slight issue of needing to probably coordinate merging this with migrating homebrew/versions' `bash-completion2` to homebrew/core; and using this as a drop in replacement.

MikeMcQuaid (approved):

mernen:
Thanks, this is the only reason I still need homebrew/versions.

JCount:
The issue here currently is that this actually conflicts with the main `bash-completion` formula. Both try to install a script to set up auto-sourcing in `$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh` and there does not appear to be an easy way to change this behavior without breaking other things. 

@MikeMcQuaid Any guidance here would be greatly appreciated, since this is bypassing `keg_only :versioned_formula`.

MikeMcQuaid:
> My current plan is to whitelist this to allow a conflicts_with "bash-completion@2" since there is no real reason to have both installed simultaneously.

:+1:

JCount:
I am going to have to split `bash-completion` into a separate PR, because here it fails to build on the CI due to artifacts from the build of `bash-completion@2` not being completely cleaned up, creating conflicts.

JCount:
@mernen Thank you for your contribution, but even more thank you for your patience while I sorted some issues out!! 🏆

PR #10601, opened 2017-03-05
node@6, node@4: use same npm as node

ilovezfs:

-----

Avoids (or reverses) npm downgrades for the main node formula caused by
installing node@6 or node@4.


ilovezfs:
@MikeMcQuaid should we also bump node's revision to reverse any downgrades caused by 0.10 or 0.12 installs?

MikeMcQuaid (approved):

MikeMcQuaid:
> @MikeMcQuaid should we also bump node's revision to reverse any downgrades caused by 0.10 or 0.12 installs?

@ilovezfs Seems reasonable (for the keg only change if nothing else)

PR #10602, opened 2017-03-05
ee: migrate to boneyard

ilovezfs:
Created with `brew boneyard-formula-pr` because the homepage is gone and there were 10 installs in the last 30 days.


ClearPathDigital:
For those that wish to continue using ee, https://github.com/ClearPathDigital/homebrew-ee

PR #10617, opened 2017-03-05
ctorrent: migrate to boneyard

fxcoudert:
Created with `brew boneyard-formula-pr` because not maintained since 2008, does not build on Sierra.


docdawning:
I'm sad ctorrent's out. I rather like this app.

PR #10621, opened 2017-03-06
Drop universal from non wine-related formulas

fxcoudert:
3 formulas that are not wine dependencies were forgotten in the previous patches for `ENV.universal_binary` removal.

Well, `libtasn1` actually is a dependency of wine optional dependency `libgsm`, but `libgsm` itself is not built as universal, so…


JCount:
👏

PR #10648, opened 2017-03-06
mandoc 1.14.1

mkhl:


neutric:
Thanks for your contribution, @mkhl!

PR #10649, opened 2017-03-06
lowdown 0.1.9 (new formula)

mkhl:


Formula/lowdown.rb
```diff
@@ -0,0 +1,25 @@
...
+  def install
+    localconfig = [
+      # Sane prefixes.
+      "PREFIX=#{prefix}",
+      "INCLUDEDIR=#{include}",
+      "BINDIR=#{bin}",
+      "LIBDIR=#{lib}",
+      "MANDIR=#{man}",
+    ]
+    (buildpath/"configure.local").write localconfig.join("\n")
```
bfontaine: You don’t need to create an array for that. Write a string instead:
```ruby
(buildpath/"configure.local").write <<-EOS.undent
  PREFIX=#{prefix}
  INCLUDEDIR=#{include}
  ...
EOS
```

mkhl: @bfontaine Thank you for your feedback. I imitated the [mandoc](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mandoc.rb) formula because it uses the same configure system, and to be honest I prefer the array personally, but if you want me to I’ll switch to a heredoc.

bfontaine: The `mandoc.rb` formula uses an array for the initial values; then append append some depending on the options it’s build with; then join everything together. This is a common pattern in formulae with options; but here it’s not necessary.

mkhl:
> The mandoc.rb formula uses an array for the initial values; then append append some depending on the options it’s build with; then join everything together. This is a common pattern in formulae with options; but here it’s not necessary.

Exactly, and it’s just as unnecessary there as it is here.

Again, do you _want_ me to use a heredoc?

MikeMcQuaid:
> Again, do you want me to use a heredoc?

Yes, please.

mkhl:
* [X] heredoc

fxcoudert:
[Github repo](https://github.com/kristapsdz/lowdown) has 5 forks, 10 stars. Perhaps it would be appropriate to host it in a personal tap?

To1ne:
I've created a tap at https://github.com/To1ne/homebrew-lowdown

PR #10654, opened 2017-03-06
Fix typo

berkus:


neutric:
Thanks for catching that, @berkus!

PR #10658, opened 2017-03-06
crystal-lang 0.21.1

asterite:

-----

Now that the previous boot binary is linked against the latest libevent and bdw-gc it's safe to remove those specific resources.

Thank you as always! 🙇

### DIFF
--- a/Formula/crystal-lang.rb
+++ b/Formula/crystal-lang.rb
@@ -1,8 +1,8 @@
 class CrystalLang < Formula
   desc "Fast and statically typed, compiled language with Ruby-like syntax"
   homepage "https://crystal-lang.org/"
-  url "https://github.com/crystal-lang/crystal/archive/0.21.0.tar.gz"
-  sha256 "4dd01703f5304a0eda7f02fc362fba27ba069666097c0f921f8a3ee58808779c"
+  url "https://github.com/crystal-lang/crystal/archive/0.21.1.tar.gz"
+  sha256 "eb93de3b388641fd2a09a2d10ab9fe40a58148d888fc48e4583a86ee1ddb1085"
   head "https://github.com/crystal-lang/crystal.git"
 
   bottle do
@@ -24,9 +24,9 @@ class CrystalLang < Formula
   depends_on "libyaml" if build.with? "shards"
 
   resource "boot" do
-    url "https://github.com/crystal-lang/crystal/releases/download/0.20.5/crystal-0.20.5-1-darwin-x86_64.tar.gz"
-    version "0.20.5"
-    sha256 "79462c8ff994b36cff219c356967844a17e8cb2817bb24a196a960a08b8c9e47"
+    url "https://github.com/crystal-lang/crystal/releases/download/0.21.0/crystal-0.21.0-1-darwin-x86_64.tar.gz"
+    version "0.21.0"
+    sha256 "e92abb33a9a592febb4e629ad68375b2577acd791a71220b8dc407904be469ee"
   end
 
   resource "shards" do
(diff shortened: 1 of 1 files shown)


ilovezfs (changes requested):
It is not safe to remove them unless the bootstrap binary is built statically because the paths will still not match if the user has a non-standard Homebrew prefix.

1. Please put the resource blocks back and update the version numbers in them and in the install method
2. Have you had a chance to look into a static build so that we can actually remove these permanently?

asterite:
@ilovezfs I just force-pushed the changes you requested.

Sorry, I didn't have time to look at the issue of statically linking libevent. I tried to, but it's more complex than what I thought.

ilovezfs:
@asterite hmm I'm not seeing them reflected in the PR.

asterite:
@ilovezfs Hm, strange. In the diff I only see changes to urls and shas.

asterite:
@ilovezfs If I download the boot binary for 0.21.0 (which is what is going to be used for 0.21.1) and run `otool -L` on the binary I get:

```
$ otool -L crystal 
crystal:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
	/usr/lib/libedit.3.dylib (compatibility version 2.0.0, current version 3.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/usr/lib/libpcre.0.dylib (compatibility version 1.0.0, current version 1.1.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
```

If I run it on the previous boot binary (0.20.5) I get:

```
$ otool -L embedded/bin/crystal 
embedded/bin/crystal:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
	/usr/lib/libedit.3.dylib (compatibility version 2.0.0, current version 3.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/usr/lib/libpcre.0.dylib (compatibility version 1.0.0, current version 1.1.0)
	/usr/local/opt/bdw-gc/lib/libgc.1.dylib (compatibility version 2.0.0, current version 2.3.0)
	/usr/local/opt/libevent/lib/libevent-2.0.5.dylib (compatibility version 7.0.0, current version 7.9.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
```

So it seems the last boot binary (0.21.0) has statically linked libevent and bdw-gc for some reason I can't understand now. I'll have to investigate that. But that means that the specific resource blocks and changes in the mach-o files are not needed.

I updated the PR once again to reflect those changes (I basically went back to the original PR)

asterite:
I actually think that libevent and bdw-gc are supposed to be statically linked in the binary. Maybe some mistake prevented that from happening in 0.20.5.

ilovezfs:
If so, it would be good to know what the problem was.

asterite:
@ilovezfs I run the build process on my machine. Maybe I forgot to unlink libevent and bdw-gc when I ran it and it linked the libraries I had in that moment instead of the static ones. In fact libpcre is also supposed to be statically linked [as far as I can see](https://github.com/crystal-lang/omnibus-crystal/blob/master/config/software/pcre.rb#L31) but it's not because I never unlink it.

The build process must be definitely improved. The main problem is that I didn't write it in the first place so I don't know much about how it works, but more people are joining the project to help me.

ilovezfs:
@asterite not that the resulting brew build is dynamically linked:
```
🍺  /usr/local/Cellar/crystal-lang/0.21.1: 786 files, 56.8MB, built in 21 minutes
iMac-TMP:~ joe$ brew linkage crystal-lang
System libraries:
  /usr/lib/libSystem.B.dylib
  /usr/lib/libiconv.2.dylib
  /usr/lib/libncurses.5.4.dylib
  /usr/lib/libpcre.0.dylib
  /usr/lib/libz.1.dylib
Homebrew libraries:
  /usr/local/opt/bdw-gc/lib/libgc.1.dylib (bdw-gc)
  /usr/local/opt/libevent/lib/libevent-2.1.6.dylib (libevent)
  /usr/local/opt/libyaml/lib/libyaml-0.2.dylib (libyaml)
  /usr/local/opt/llvm/lib/libc++.1.dylib (llvm)
  /usr/local/opt/pcre/lib/libpcre.1.dylib (pcre)
```

I'm not complaining since a statically linked bootstrap binary, and a dynamically linked release build are exactly what we want, I think ... unless you're saying you think the release builds here should also be statically linked.

asterite:
@ilovezfs Yes, that's OK, because the resulting binary is not going to be used as the next boot binary.

ilovezfs (approved):

ilovezfs:
Shipped! Thanks @asterite

PR #10665, opened 2017-03-07
psql: fix allow intuitive connection by adding createdb

### LINKED ISSUE #7682: Postgredb formula database <user> does not exist problem
After PostgreSQL installation on _os x 10.12.1 (16B2555)_, when you call `psql` it gives errors. Details are below. Should I fix this by contributing to the homebrew formula?

### Reproducing

1. `rm -rf /usr/local/var/postgres*`
2. `brew install postgresql`
3. `brew services start postgresql`
4. `psql` -- boom: error:

    ```bash
    psql: FATAL:  database "<user>" does not exist
    ```

### Fix

Do the reproducing steps then run just `createdb`. `psql` will work.

I think that the formula should have been calling `createdb`.

### Details

[```brew gist-logs postgresql```](https://gist.github.com/inancgumus/bbfcf2912adedcebccc05fa8350c003d)

harshnarang8:
Fixes #7682.
Using `initdb` is necessary for creation of a filesystem like folder for `postgresql`. However, running `psql` after installing postgresql results in a fatal error, i.e. `database "user" does not exist`. Running `createdb` fixes the error, as it creates a database with the name of the user. Also, simply running `system "createdb"` may lead to displaying an error if the user already had the database in his system. So, calling creadb through open3 executes the command, and does not display the mentioned error which is not required to be displayed to the user.

-----

### DIFF
--- a/Formula/postgresql.rb
+++ b/Formula/postgresql.rb
@@ -94,11 +94,13 @@ class Postgresql < Formula
   end
 
   def post_install
+    require "open3"
     (var/"log").mkpath
     (var/"postgres").mkpath
     unless File.exist? "#{var}/postgres/PG_VERSION"
       system "#{bin}/initdb", "#{var}/postgres"
     end
+    _stdout, _stderr, _status = Open3.capture3("createdb")
   end
 
   def caveats; <<-EOS.undent
--- a/Formula/postgresql.rb
+++ b/Formula/postgresql.rb
@@ -94,13 +94,14 @@ class Postgresql < Formula
   end
 
   def post_install
-    require "open3"
     (var/"log").mkpath
     (var/"postgres").mkpath
     unless File.exist? "#{var}/postgres/PG_VERSION"
       system "#{bin}/initdb", "#{var}/postgres"
+      system "brew", "services", "start", "postgresql"
+      system "createdb"
+      system "brew", "services", "stop", "postgresql"
     end
-    _stdout, _stderr, _status = Open3.capture3("createdb")
   end
 
   def caveats; <<-EOS.undent


MikeMcQuaid:
Can you provide a much more detailed explanation of why this is needed or useful? Thanks!

Formula/postgresql.rb
```diff
@@ -94,11 +94,13 @@ def install
   end
 
   def post_install
+    require "open3"
     (var/"log").mkpath
     (var/"postgres").mkpath
     unless File.exist? "#{var}/postgres/PG_VERSION"
       system "#{bin}/initdb", "#{var}/postgres"
     end
+    _stdout, _stderr, _status = Open3.capture3("createdb")
```
MikeMcQuaid: `system "#{bin}/createdb"`

MikeMcQuaid (changes requested):

MikeMcQuaid:
Ok, I see this is from https://github.com/Homebrew/homebrew-core/issues/7682#issuecomment-284549247. Can you add `Fixes #7682` to your initial PR comment and a summary of the solution?

I'm still not sure I quite understand why a database with the same name as the default user is useful.

MikeMcQuaid:
> I'm still not sure I quite understand why a database with the same name as the default user is useful.

That said, I'm happy to merge this regardless.

Formula/postgresql.rb
harshnarang8: @MikeMcQuaid the command may lead to displaying the error: 
`createdb: database creation failed: ERROR:  database "harshn" already exists`
I think displaying the error is not required, and the command doesn't change anything if the db exists already. (in this case harshn) That's why I thought opting for Open3 instead of system would be better.

MikeMcQuaid: It would be good to verify if it's necessary. I'd suggest perhaps just putting this in the `unless` above?

harshnarang8: While I think that should be done, there may be a chance that the `{var}/postgres/PG_VERSION` could exist due to a previous installation and uninstallation (where the database in question is non existant), rendering `unless` ineffective.
I also thought of using another block to test the existence of the db, but couldn't think of one, especially since postgres has its own ways of storing and reading database data. I'd be glad if you could provide a workaround, as even I find using Open3 clumsy, and more of a hack.

MikeMcQuaid: There's a chance of that but it seems that it handles the most typical new installation case so I think it's better. The only other option is a test for the user database existing if you have any ideas.

Formula/postgresql.rb
```diff
@@ -98,6 +98,9 @@ def post_install
     (var/"postgres").mkpath
     unless File.exist? "#{var}/postgres/PG_VERSION"
       system "#{bin}/initdb", "#{var}/postgres"
+      system "brew", "services", "start", "postgresql"
+      system "createdb"
```
MikeMcQuaid: `system "#{bin}/createdb"`

Formula/postgresql.rb
```diff
@@ -98,6 +98,9 @@ def post_install
     (var/"postgres").mkpath
     unless File.exist? "#{var}/postgres/PG_VERSION"
       system "#{bin}/initdb", "#{var}/postgres"
+      system "brew", "services", "start", "postgresql"
```
MikeMcQuaid: We can't do this; is there any way of achieving the same result without it?

MikeMcQuaid (changes requested):

Formula/postgresql.rb
harshnarang8: I figured something out, perhaps we could do something like :
`system "#{bin}/initdb", "#{var}/postgres"`
`system "pg_ctl -D #{var}/postgres -l #{var}/postgres/server.log start"`
`sleep(0.5)`
`system "createdb"`
`system "pg_ctl -D #{var}/postgres stop -s -m fast"`
I think so because
- `initdb` works fine
- `pg_ctl` command is for starting the postgresql server (does the job of brew services start)
- `sleep` is required because the server actually takes time to start, maybe due to a process which continues even after the system call ( I tested a variant without sleep, and the script couldn't detect the started server. Though running each command individually did the trick. )
- Next `pg_ctl` is used to stop the server

MikeMcQuaid: Is there any way of doing this without starting the server? If not, I fear this may not be a change we can do.

harshnarang8: [http://stackoverflow.com/questions/40040824/how-to-initialize-a-postgresql-database-without-running-the-postgresql-server](url)
This is a possible way to go without starting and stopping the server, but essentially the same stuff happens in the background.

MikeMcQuaid: Something like that seems superior if it still works if a server is already running (I suspect it won't, sadly). I think this may just be worth adding to the `caveats` if anything.

harshnarang8: @MikeMcQuaid I'm sorry, but I'm unfamiliar with adding to the `caveats`. Does it mean adding a warning text after installation has been completed?

MikeMcQuaid: @harshnarang8 Yes but I think we should probably just pass on this, sorry.

MikeMcQuaid:
Passing on this as there's no clean way to do it automatically, sorry!

PR #10670, opened 2017-03-07
Update geoserver.rb

yoghi:
fix SHA256 hash


ilovezfs (changes requested):
We'll need to know why the file has changed before taking any action here. Has the fact that the checksum changed been reported upstream yet? If not, please do so, and post a link to the issue here.

fxcoudert:
Updated to later version by #11579

PR #10778, opened 2017-03-08
llvm: build libcxx experimental library

outcoldman:

-----

In most cases when you want to get latest clang with libcxx you also want to try experimental features, currently formula installs all experimental headers, but does not give you access to the experimental library, which is required for using some of the experimental features, like filesystem. 

By default I include experimental library when llvm is built with libcxx. It can be disabled with `--without-experimental`.

Also added a test to verify that experimental features can be used.

### DIFF
--- a/Formula/llvm.rb
+++ b/Formula/llvm.rb
@@ -130,6 +130,7 @@ class Llvm < Formula
   option "with-python", "Build bindings against custom Python"
   option "with-shared-libs", "Build shared instead of static libraries"
   option "without-libffi", "Do not use libffi to call external functions"
+  option "without-experimental", "Do not build libc++experimental library"
 
   depends_on "libffi" => :recommended # http://llvm.org/docs/GettingStarted.html#requirement
   depends_on "graphviz" => :optional # for the 'dot' tool (lldb)
@@ -227,7 +228,13 @@ class Llvm < Formula
       args << "-DLLVM_BUILD_LLVM_DYLIB=ON"
     end
 
-    args << "-DLLVM_ENABLE_LIBCXX=ON" if build_libcxx?
+    if build_libcxx?
+      args << "-DLLVM_ENABLE_LIBCXX=ON"
+      if build.with? "experimental"
+        args << "-DLIBCXX_ENABLE_EXPERIMENTAL_LIBRARY=ON"
+        args << "-DLIBCXX_INSTALL_EXPERIMENTAL_LIBRARY=ON"
+      end
+    end
 
     if build.with?("lldb") && build.with?("python")
       args << "-DLLDB_RELOCATABLE_PYTHON=ON"
@@ -335,6 +342,19 @@ class Llvm < Formula
       }
     EOS
 
+    if build.with? "experimental"
+      (testpath/"test_experimental.cpp").write <<-EOS.undent
+        #include <iostream>
+        #include <experimental/filesystem>
+
+        int main()
+        {
+            std::cout << std::experimental::filesystem::current_path() << std::endl;
+            return 0;
+        }
+      EOS
+    end
+
     # Testing Command Line Tools
     if MacOS::CLT.installed?
       libclangclt = Dir["/Library/Developer/CommandLineTools/usr/lib/clang/#{MacOS::CLT.version.to_i}*"].last { |f| File.directory? f }
@@ -383,6 +403,19 @@ class Llvm < Formula
               "-Wl,-rpath,#{lib}", "test.cpp", "-o", "test"
       assert_includes MachO::Tools.dylibs("test"), "@rpath/libc++.1.dylib"
       assert_equal "Hello World!", shell_output("./test").chomp
+
+      # link against installed libc++experimental
+      if build.with? "experimental"
+        system "#{bin}/clang++", "-v", "-nostdinc",
+                "-std=c++11", "-stdlib=libc++",
+                "-lc++experimental",
+                "-I#{opt_include}/c++/v1",
+                "-I#{libclangxc}/include",
+                "-I#{MacOS.sdk_path}/usr/include",
+                "-L#{lib}",
+                "-Wl,-rpath,#{lib}", "test_experimental.cpp", "-o", "test_experimental"
+        assert_match testpath.to_s, shell_output("./test_experimental").chomp
+      end
     end
   end
 end


outcoldman:
Don't believe that this is actually my change which introduces this issue with tests.

outcoldman:
Ok, seems like old tests did not take in account `rpath`, I would assume that maintainers of these tests had `LDFLAGS` in their environment, so this is why they have not seen it, but curious how this test passed before.

outcoldman:
ping @DomT4 @maxim-belkin to review this change (based on previous changes)

Formula/llvm.rb
```diff
@@ -130,6 +130,7 @@ class Llvm < Formula
   option "with-python", "Build bindings against custom Python"
   option "with-shared-libs", "Build shared instead of static libraries"
   option "without-libffi", "Do not use libffi to call external functions"
+  option "without-experimental", "Build libc++experimental library (only if libc++ build is enabled)"
```
maxim-belkin: my comment in pseudo-code:
```
if option.starts_with? "without"
  assert_match "Do not", option.text
end
```

outcoldman: Good point, fixed. I actually had `with-experimental` at beginning, but decided that it does not hurt to include it by default, but forgot to change the description.
Fixed now.

outcoldman:
@maxim-belkin good for merge?

maxim-belkin:
I'm a Linuxbrew maintainer, not Homebrew :)

AraHaan:
Why? I require the experimental library to build most of my projects on other systems that I target for all platforms even though I write them on Windows 7.

outcoldman:
@maxim-belkin ah... formalities 

@AraHaan I am a little bit confused, are you saying that you vote for this change or not?

p.s. seems like 4.0.0 is getting released in following days, so maybe we will not even need that

Formula/llvm.rb
```diff
@@ -130,6 +130,7 @@ class Llvm < Formula
   option "with-python", "Build bindings against custom Python"
   option "with-shared-libs", "Build shared instead of static libraries"
   option "without-libffi", "Do not use libffi to call external functions"
+  option "without-experimental", "Do not build libc++experimental library"
```
DomT4: Is there any reason to not make this a complete default if you `build_libcxx?`?

Formula/llvm.rb
```diff
@@ -381,8 +402,21 @@ def caveats
               "-I#{MacOS.sdk_path}/usr/include",
               "-L#{lib}",
               "-Wl,-rpath,#{lib}", "test.cpp", "-o", "test"
-      assert_includes MachO::Tools.dylibs("test"), "#{opt_lib}/libc++.1.dylib"
+      assert_includes MachO::Tools.dylibs("test"), "@rpath/libc++.1.dylib"
```
DomT4: Why the change here? Seems suspicious this would need changing.

Formula/llvm.rb
```diff
@@ -381,8 +402,21 @@ def caveats
...
+      if build.with? "experimental"
+        system "#{bin}/clang++", "-v", "-nostdinc",
+                "-std=c++11", "-stdlib=libc++",
+                "-lc++experimental",
+                "-I#{opt_include}/c++/v1",
+                "-I#{libclangxc}/include",
+                "-I#{MacOS.sdk_path}/usr/include",
+                "-L#{lib}",
+                "-Wl,-rpath,#{lib}", "test_experimental.cpp", "-o", "test_experimental"
+        assert_equal "\"#{Pathname.pwd}\"", shell_output("./test_experimental").chomp
```
DomT4: I suspect a simpler:
```ruby
assert_equal testpath, shell_output("./test_experimental").chomp
```
May work here?

Formula/llvm.rb
DomT4: Ah, I did wonder if related to a commit in `brew` the other day, and ILZ confirmed it with a PR before I'd even asked. https://github.com/Homebrew/homebrew-core/pull/10808

ilovezfs: @DomT4 see https://github.com/Homebrew/homebrew-core/pull/10808. This belongs in a totally separate PR.

outcoldman: Well, the tests are failing (see https://bot.brew.sh/job/Homebrew%20Core/17783/), the question is for original author of this test, because when you build llvm from this formula it sets installation_path for the libraries as @rpath. You can use `otool -l [path to libc++ library] to see that LC_ID_DYLIB is set to `@rpath`. 

It is possible that LLVM have changed it from being absolute by default to relative, but I am surprised how it got updated in homebrew-core without seeing breakage in the tests. And I could not find any flags in libcxx project which can change this behavior.

Formula/llvm.rb
ilovezfs: if it's experimental, I'd think we'd want it opt-in or not in the formula at all

outcoldman: For some reason this formula did not have it at the beginning, so I just keep a possibility to allowing people to keep the old behavior if they want to. But to be honest I do not see any reasons, other than if you already have libc++experimental built somewhere and now this one will give you a conflict

Formula/llvm.rb
DomT4: Aye, see the brief discussion above. I thought the change that broke `llvm`'s test was going to get reverted, but I presume it is sticking around at least for now.

ilovezfs: No, this was a breaking change made in brew itself in order to fix `rust`. There are any number of formulae, among which is `llvm`, that are impacted.

Formula/llvm.rb
outcoldman: Well, in `test_experimental` I actually call cwd, so in test I compare current working directories, if `testpath` happens to be the same as current working directory is good, but my guess it might change in future? For me current implementation is more bullet proof, but I am ok to change it to suggested if you insist.

Formula/llvm.rb
DomT4: > if it's experimental, I'd think we'd want it opt-in or not in the formula at all

Probably the latter given the higher bar for option justification these days. I was just curious as to whether this breaks any existing use case, and if it doesn't, whether it was worth chucking it in by default.

Formula/llvm.rb
DomT4: If Homebrew breaks the `testpath` helper there's going to be a whole world of pain beyond this formula, heh. But yes, unless you're moving to a new directory deliberately inside `test do` you should be in `testpath`.

maxim-belkin: ```assert_equal testpath, "\"#{Pathname.pwd}\""```

Formula/llvm.rb
maxim-belkin: @ilovezfs do you have a ref to that change? just curious...

ilovezfs: @maxim-belkin yup:
https://github.com/Homebrew/brew/pull/2036
https://github.com/rust-lang/rust/issues/39870

outcoldman:
@ilovezfs just to summarize,
- do you want to take #10808 and rebase my PR on top of it?
- are we ok with having `without-experimental` or do we want to switch it to `with-experimental` (exclude by default)?
- and I will change test to use `testpath` (possible it will be `assert_equal "\"#{testpath}\"", shell_output("./test_experimental").chomp`, as filesystem adds quotes by default.

Formula/llvm.rb
maxim-belkin: Muchas gracias, señor

ilovezfs:
>do you want to take #10808 and rebase my PR on top of it?

Yes, but only after I've shipped #10808.

>are we ok with having without-experimental or do we want to switch it to with-experimental (exclude by default)?

Not sure yet. I need to learn more about what this is.

>and I will change test to use testpath (possible it will be assert_equal "\"#{testpath}\"", shell_output("./test_experimental").chomp, as filesystem adds quotes by default.

I'd probably just use `assert_match` like
```
assert_match testpath, shell_output("./test_experimental")
```

AraHaan:
The problem with that is that switching it will break doozer as many rely on experimental and they cannot specify formulae options in.doozer.json as of yet making all OSX builds using this broken.

outcoldman:
@AraHaan but wait, currently brew does not ship experimental, so I am actually adding it to the lib folder. Are you voting for this change, because it will help people who uses doozer.json, so they can start using experimental lib right away?

AraHaan:
Ok thanks. And then I could reenable OSX with this so I can try to figure out build cache on it so it can package the build binaries for OSX instead of me having to use an 32 and 64 bit OSX VM on VirtualBox to try to ship the binaries.

AraHaan (approved):

ilovezfs:
@outcoldman it should be safe to rebase now.

Formula/llvm.rb
MikeMcQuaid: Given we don't actually use this libc++ ourselves at all I'm tempted to say this should be enabled with `with-libcxx` (so by default). Thoughts?

outcoldman:
@ilovezfs rebased, squashed, changed also test to suggested

Formula/llvm.rb
ilovezfs: It seems sketchy to replace stock libcxx, even if it's only optional for whatever Very Good Reasons we have, with one that has experimental features enabled.

AraHaan (approved):

Formula/llvm.rb
ilovezfs: I see we have ` option "without-libcxx"`: even more sketchy then.

AraHaan:
Thanks for adding a test for <experimental/filesystem>.

Formula/llvm.rb
```diff
@@ -383,6 +403,19 @@ def caveats
...
+      if build.with? "experimental"
+        system "#{bin}/clang++", "-v", "-nostdinc",
+                "-std=c++11", "-stdlib=libc++",
+                "-lc++experimental",
+                "-I#{opt_include}/c++/v1",
+                "-I#{libclangxc}/include",
+                "-I#{MacOS.sdk_path}/usr/include",
+                "-L#{lib}",
+                "-Wl,-rpath,#{lib}", "test_experimental.cpp", "-o", "test_experimental"
+        assert_equal testpath, shell_output("./test_experimental").chomp
```
DomT4: I misspoke/mistyped earlier; ILZ correctly corrected me. This works:
```ruby
assert_match testpath.to_s, shell_output("./test_experimental").chomp
```

outcoldman: fixed, trying again

AraHaan (approved):

AraHaan:
Sorry for approving this 3 times. I do not know ruby however I know a little C++, Python, Visual Basic, C#, and JS.

outcoldman:
All green! @ilovezfs @DomT4 thank you for feedback and helping with improvements!

AraHaan:
Can this be merged now plz?

MikeMcQuaid:
@AraHaan Please be patient.

AraHaan:
Alright. So When could this be merged?

outcoldman:
@AraHaan btw, I believe you can install in brew right from the PR, by specifying link to the right formula file. You can google to find a way to find it. 

Not sure what exactly is holding merge of this change.

MikeMcQuaid:
> Not sure what exactly is holding merge of this change.

This:

> It seems sketchy to replace stock libcxx, even if it's only optional for whatever Very Good Reasons we have, with one that has experimental features enabled.

and this: 

> I need to learn more about what this is.

On: 

> So When could this be merged?

when it's ready and we're happy with it. That could take a while. Again, please be patient.

outcoldman:
@MikeMcQuaid but this change does not change the behavior of `libcxx`, it is only about adding `libexperimental`.

outcoldman:
Ok, I don't believe this change is required anymore, considering that 4.0.0 is out and after homebrew will bump to it - there are no need to specify this option, because it will be ON by default starting from 4.0.0.

ilovezfs:
@outcoldman is `libc++experimental.a` of any use if you've built `--without-libcxx`?

maxim-belkin:
I haven't looked at llvm source code yet, but I would assume that it would then use system's libcxx.

It wouldn't make much sense on Linux, though

outcoldman:
@ilovezfs I am not 100% sure if this is compatible, but I second what @maxim-belkin said

ilovezfs:
So the question is really whether we think it's worth supporting an independent option or if it's acceptable to ship it if and only if the user has not passed `--without-libcxx`.

ilovezfs:
And yet another option would be to ship it unconditionally regardless of the status of `--without-libcxx`.

AraHaan:
> I haven't looked at llvm source code yet, but I would assume that it would then use system's libcxx.
>
> It wouldn't make much sense on Linux, though

Well the problem with the system libcxx is that it only supports up to C++ 11, NOT C++ 14. An another issue is you could try to install boost filesystem all you WANT but if the optimization level of boost filesystem is -02 and I am using -03 it WILL cause my code to segfault. This is why I rely on the experimental filesystem instead so I can use -03 with it and not worry about segmentation faults.

fxcoudert:
> 4.0.0 is out and after homebrew will bump to it - there are no need to specify this option, because it will be ON by default starting from 4.0.0

Closing accordingly.

PR #10783, opened 2017-03-08
postgres-xc: append -D_XOPEN_SOURCE to CFLAGS, rm patching

JCount:
append -D_XOPEN_SOURCE to CFLAGS, rm patching


MikeMcQuaid:
👏

PR #10803, opened 2017-03-09
Update DCMTK to the newest snapshot

waldyrious:

-----

The audit does not pass:
```
$ brew audit --strict dcmtk
dcmtk:
  * Stable: version 3.6.1_20170228 should not end with an underline and a number
```

But this is expected, since the snapshots are currently being used as stable version, which is indicated in the comment immediately above the diff.

### DIFF
--- a/Formula/dcmtk.rb
+++ b/Formula/dcmtk.rb
@@ -2,9 +2,9 @@ class Dcmtk < Formula
   desc "OFFIS DICOM toolkit command-line utilities"
   homepage "http://dicom.offis.de/dcmtk.php.en"
 
-  # Current snapshot used for stable now.
+  # Current snapshot used for stable now (using - instead of _ in the version field).
   url "http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20170228.tar.gz"
-  version "3.6.1_20170228"
+  version "3.6.1-20170228"
   sha256 "8de2f2ae70f71455288ec85c96a2579391300c7462f69a4a6398e9ec51779c11"
 
   head "git://git.dcmtk.org/dcmtk.git"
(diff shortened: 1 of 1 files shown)


Formula/dcmtk.rb
```diff
@@ -3,9 +3,9 @@ class Dcmtk < Formula
   homepage "http://dicom.offis.de/dcmtk.php.en"
 
   # Current snapshot used for stable now.
-  url "http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20161102.tar.gz"
-  version "3.6.1-20161102"
-  sha256 "657adb3811e0c5c08d8f143a6d878afcd92fac7dcf0d3c89860eecffd5a1a873"
+  url "http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20170228.tar.gz"
+  version "3.6.1_20170228"
```
fxcoudert: Per Homebrew policy, please use `3.6.1-20170228` in `version` line (with the hyphen, even if it does not match the URL above).

fxcoudert (changes requested):

Formula/dcmtk.rb
waldyrious: Oh, so that's how you make it pass the audit :) can I add a comment so others don't overlook this?

fxcoudert: It was already that way (see before your change), so I think it's self-explanatory.

waldyrious: That's my point exactly -- just like I missed the subtle difference, others could in the future as well. I'll just add to the existing comment above this block.

dunn:
Thanks! Merged as 486ab1458277f68dded9dcd691895fea828bf761.

PR #10804, opened 2017-03-09
mention --version and --help in bump-formula-pr

waldyrious:

-----

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ First time contributing to Homebrew? Read our [Code of Conduct](https://github.c
 ### Submit a version upgrade for the `foo` formula
 
 * check if the same upgrade has been already submitted by [searching the open pull requests for `foo`](https://github.com/Homebrew/homebrew-core/pulls?utf8=✓&q=is%3Apr+is%3Aopen+foo).
-* `brew bump-formula-pr --strict foo` with `--url=...` and `--sha256=...` or `--tag=...` and `--revision=...` arguments.
+* `brew bump-formula-pr --strict foo` with `--url=...` and `--sha256=...`, or `--tag=...` and `--revision=...` arguments (and optionally `--version=...` as well). Run `brew bump-formula-pr --help` for additional instructions.
 
 ### Add a new formula for `foo` version `2.3.4` from `$URL`
 


MikeMcQuaid:
It'll throw an error if `--version` is required, no? As a result I think it's worth keeping this text compact with what will work for the majority of cases.

waldyrious:
It does, but running the same command with `--version` added now fails because the new version is cached somewhere, and brew claims there's no version change -- so the error message ends up not helping much here. Maybe `bump-formula-pr` could clear the cache of the version  if it fails?

MikeMcQuaid:
CC @ilovezfs for thoughts.

ilovezfs:
I've never seen that behavior as far as I know, so I'd need an example. Regardless, `--version=` is for overrides, so saying it can optionally be provided isn't accurate anyway. It's either mandatory or proscribed depending on the circumstance.

waldyrious:
Unfortunately I don't have the terminal output I could paste here, but here are the relevant commands from `.bash_history` reflecting my attempt at doing this:

```
$ brew bump-formula-pr --strict dcmtk --url=http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20170228.tar.gz --sha256=8de2f2ae70f71455288ec85c96a2579391300c7462f69a4a6398e9ec51779c11
# Got an error related to the version being the same

$ brew bump-formula-pr --help

$ brew bump-formula-pr --strict dcmtk --url=http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20170228.tar.gz --sha256=8de2f2ae70f71455288ec85c96a2579391300c7462f69a4a6398e9ec51779c11 --version=3.6.1_20170228
# Got an error due to the new version ending in an underscore followed by a number, which isn't supposed to be used for stable versions

$ brew bump-formula-pr dcmtk --url=http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20170228.tar.gz --sha256=8de2f2ae70f71455288ec85c96a2579391300c7462f69a4a6398e9ec51779c11 --version=3.6.1_20170228
# This invocation (without strict) then failed because I already had a fork of homebrew-core (cloned from Linuxbrew for a PR I sent a while ago) under my github account

$ brew bump-formula-pr dcmtk --url=http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20170228.tar.gz --sha256=8de2f2ae70f71455288ec85c96a2579391300c7462f69a4a6398e9ec51779c11 --version=3.6.1_20170228
# Second attempt, after removing the exising homebrew-core fork
# This time it failed because the "current" version, as far as Homebrew was concerned, was the same as the one I was passing as a parameter
```

I tried doing `brew uninstall dcmtk && brew install dcmtk` to restore the correct "current" version (which should be 3.6.1-20161102), but it still kept the version as 3.6.1_20170228.

Later on I eventually found a lock file and removed it, but I can't find the log of that command to give you the exact file. It ended in ".brewing" IIRC. Maybe I should try this doing `brew bump-formula-pr dcmtk --url=http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20170228.tar.gz --sha256=8de2f2ae70f71455288ec85c96a2579391300c7462f69a4a6398e9ec51779c11 --version=3.6.1_20170228` again to see if it works now that I've removed the lock?

I wonder if `bump-formula-pr` automatically cleaned up if the push to github fails would prevent this issue from tripping an unexperienced Homebrew contributor (like me), since the last command in the list above would have (hopefully) succeeded.

If that would be the case, then I'd agree with shortening this PR to only add a mention of the `--help` option.

ilovezfs:
Thanks for the extended information. Yes, if bump-formula-pr fails during the git operations, the original file is not restored. I've seen that happen when the branch already exists.

waldyrious:
@ilovezfs is there something I can help with in that regard? I doubt that I'd be able to submit a PR to do that cleanup, but at the very least I could create an issue to track that :)

ilovezfs:
We could put this entire block https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb#L279-L325 inside a begin ... ensure and then unconditionally restore the original file at the end, but I think that could lead to some weird outcomes depending on exactly where it fails along the way.

waldyrious:
Ok, what do you suggest then? I'm happy to help with whatever is within my reach.

JCount:
Closing per inactivity and ambiguous end goals at this point. Thank you for your work nonetheless @waldyrious. If you end up deciding to work on addressing this general issue you have found again, feel free to open a new PR.

waldyrious:
>If you end up deciding to work on addressing this general issue

@JCount I believe the "ball" was in @ilovezfs' court to offer his thoughts on what he believes would be a sensible path to fix the underlying issue, or am I reading this wrong? I'm happy to wait as long as necessary until he has the time and disposition to do so, but having this issue closed, I'm afraid, would just lead to this being forgotten...

PR #10809, opened 2017-03-09
deis 2.12.0

jayjun:

-----

### DIFF
--- a/Formula/deis.rb
+++ b/Formula/deis.rb
@@ -1,10 +1,8 @@
-require "language/go"
-
 class Deis < Formula
   desc "The CLI for Deis Workflow"
   homepage "https://deis.io/"
-  url "https://github.com/deis/workflow-cli/archive/v2.11.0.tar.gz"
-  sha256 "e35744bd5223d9099f719c69fc1eda648c44075cd29014dd82da5a2a141f3382"
+  url "https://github.com/deis/workflow-cli/archive/v2.12.0.tar.gz"
+  sha256 "5d4dbd7f21774b139013cf02c19b27b5a110b472744364ba26ad187f3200bf77"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


JCount:
@jayjun It seems that some of the `go_resource`dependencies can be updated. In particular:
```
diff --git a/Formula/deis.rb b/Formula/deis.rb
index f8cf334d2d..3955755322 100644
--- a/Formula/deis.rb
+++ b/Formula/deis.rb
@@ -42,12 +42,12 @@ class Deis < Formula
 
   go_resource "golang.org/x/text" do
     url "https://github.com/golang/text.git",
-        :revision => "85c29909967d7f171f821e7a42e7b7af76fb9598"
+        :revision => "f28f36722d5ef2f9655ad3de1f248e3e52ad5ebd"
   end
 
   go_resource "golang.org/x/net" do
     url "https://github.com/golang/net.git",
-        :revision => "b4690f45fa1cafc47b1c280c2e75116efe40cc13"
+        :revision => "a6577fac2d73be281a500b310739095313165611"
   end
 
   go_resource "github.com/docopt/docopt-go" do
@@ -67,7 +67,7 @@ class Deis < Formula
 
   go_resource "golang.org/x/crypto" do
     url "https://github.com/golang/crypto.git",
-        :revision => "453249f01cfeb54c3d549ddb75ff152ca243f9d8"
+        :revision => "728b753d0135da6801d45a38e6f43ff55779c5c2"
   end
 
   go_resource "gopkg.in/yaml.v2" do
```
(I tested these changes, and the formula still builds, but I did not test the actual functionality.)

jayjun:
Thanks @JCount. Sorry for the oversight, still new to Homebrew.

I've updated the formula to match Deis 2.12.0's [`glide.lock`](https://github.com/deis/workflow-cli/blob/v2.12.0/glide.lock) file exactly. Should be a 100% reproducible build.

JCount:
@ilovezfs The move made to synchronize the go resources with the versions in upstream's `glide.lock` makes sense in a variety of ways. However, for a number of the go resources, this actually causes a much older revision to be used, and the effective "version" to decrease. Should we abandon the idea of tracking the `glide.lock` file because the specified revisions are simply too old?

ilovezfs (changes requested):
We should depend on glide and use it directly, and get rid of the resource blocks altogether. See
algernon, apache-brooklyn-cli, cayley, collector-sidecar, consul-backinator, dvm, ironcli, kubernetes-helm, mupen64plus, peco, piknik, snap-telemetry, and terragrunt.

jayjun:
@ilovezfs @JCount No worries, latest commit depends on Glide as suggested.

Formula/deis.rb
```diff
@@ -13,78 +11,20 @@ class Deis < Formula
...
     deispath.install Dir["{*,.git}"]
-    Language::Go.stage_deps resources, buildpath/"src"
 
-    cd "src/github.com/deis/workflow-cli" do
+    cd deispath do
+      system "glide", "install"
       system "go", "build", "-o", "build/deis",
         "-ldflags",
-        "'-X=github.com/deis/workflow-cli/version.Version=v2.11.0'"
+        "'-X=github.com/deis/workflow-cli/version.Version=v2.12.0'"
```
ilovezfs: should be able to make it =v#{version}

jayjun: Yes, it works. Updated with `#{version}`.

ilovezfs (approved):

JCount:
@jayjun Thank you for your contribution to Homebrew, and also for being so responsive to feedback! 💯

PR #10815, opened 2017-03-09
rebar@3: import from homebrew/versions

rjcoelho:

Related PR: https://github.com/Homebrew/homebrew-versions/pull/1542

### DIFF
--- a/Formula/rebar@3.rb
+++ b/Formula/rebar@3.rb
@@ -1,0 +1,23 @@
+class RebarAT3 < Formula
+  desc "Erlang build tool"
+  homepage "https://github.com/erlang/rebar3"
+  url "https://github.com/erlang/rebar3/archive/3.3.5.tar.gz"
+  sha256 "b17661bedaf2060179e0b7b2c4a64350b7a303d809397325ee4be38cc1dc9058"
+
+  head "https://github.com/rebar/rebar3.git"
+
+  depends_on "erlang"
+
+  def install
+    system "./bootstrap"
+    bin.install "rebar3"
+
+    bash_completion.install "priv/shell-completion/bash/rebar3"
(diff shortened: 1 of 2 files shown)


Formula/rebar@3.rb
```diff
@@ -0,0 +1,23 @@
...
+  head "https://github.com/rebar/rebar3.git"
+
+  depends_on "erlang"
+
+  def install
+    system "./bootstrap"
+    bin.install "rebar3"
+
+    bash_completion.install "priv/shell-completion/bash/rebar3"
+    zsh_completion.install "priv/shell-completion/zsh/_rebar3" => "_rebar3"
```
ilovezfs:
```
zsh_completion.install "priv/shell-completion/zsh/_rebar3"
```
should be sufficient since the file name isn't changing

Formula/rebar@3.rb
```diff
@@ -0,0 +1,23 @@
+class RebarAT3 < Formula
+  desc "Erlang build tool"
+  homepage "https://github.com/erlang/rebar3"
+  url "https://github.com/erlang/rebar3/archive/3.3.5.tar.gz"
+  sha256 "b17661bedaf2060179e0b7b2c4a64350b7a303d809397325ee4be38cc1dc9058"
+
```
ilovezfs: this blank line isn't needed

ilovezfs:
I wonder if we should just upgrade the `rebar` formula to version 3 and delete the versions formula. @nijikon I think you were the one who suggested that, right?

rjcoelho:
@ilovezfs no because rebar3 isn't backward compatible. Once you brew update your rebar2 projects would no longer compile.

ilovezfs:
Sure. But we don't have anything using rebar2

MikeMcQuaid:

PR #10823, opened 2017-03-09
use secure Git access with savannah.gnu.org repos

vszakats:

-----

### DIFF
--- a/Formula/bash.rb
+++ b/Formula/bash.rb
@@ -2,7 +2,7 @@ class Bash < Formula
   desc "Bourne-Again SHell, a UNIX command interpreter"
   homepage "https://www.gnu.org/software/bash/"
 
-  head "http://git.savannah.gnu.org/r/bash.git"
+  head "https://git.savannah.gnu.org/git/bash.git"
 
   stable do
     url "https://ftpmirror.gnu.org/bash/bash-4.4.tar.gz"
--- a/Formula/coreutils.rb
+++ b/Formula/coreutils.rb
@@ -12,7 +12,7 @@ class Coreutils < Formula
   end
 
   head do
-    url "git://git.sv.gnu.org/coreutils"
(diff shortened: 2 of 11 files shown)


woodruffw:
Let's go ahead and merge #10825, #10826, and #10827 into this PR (unless there's a reason not to that I'm not aware of).

vszakats:
@woodruffw My last known instruction from maintainers was to always groups such changes into smaller PRs to avoid exhausting the CI sessions. That's what I'm doing since then. Are you saying that this is not undesired and one can pack in any number of changes inside a single PR? [I'd be happy about it, but not necessarily in context of the already submitted PRs :) ]

woodruffw:
If another maintainer said to split it up, that's fine. Thanks for clarifying :smile:

ilovezfs:
It would be good to have an audit check for these.

ilovezfs:
Thanks @vszakats

vszakats:
@ilovezfs Thank you. (A `brew audit --strict --online` was done for all of them, in every PR.)

ilovezfs:
@vszakats right. I mean that it doesn't catch these.

vszakats:
Ah okay, I definitely agree. Though these (or at least `svn.code.sf.net`) were special cases, so there didn't seem to be an obvious way to do it. It also seems non-trivial to recommend the correct replacement URL.

ilovezfs:
It seems safe to say http://git.savannah.gnu.org is never OK now though right?

vszakats:
These are never OK:
```
(http|git)://git.(savannah|sv).gnu.org/
(http|git)://git.(savannah|sv).gnu.org/
http://hg.(savannah|sv).gnu.org/
(http|svn)://svn.code.sf.net/
(http|git)://git.code.sf.net/
(http|git)://*.git.sourceforge.net/gitroot/
http://(download|download-mirror).savannah.gnu.org/
```

PR #10834, opened 2017-03-10
git-completion 2.12.0 (new formula)

lucidaeon:

-----

### DIFF
--- a/Formula/git-completion.rb
+++ b/Formula/git-completion.rb
@@ -1,0 +1,23 @@
+class GitCompletion < Formula
+  desc "Bash and Zsh completion for Git"
+  homepage "https://github.com/git/git"
+  url "https://github.com/git/git/archive/v2.12.0.tar.gz"
+  sha256 "1f94b3bbe2928b63ce7d743438052ad07f21b20e91ca5b9161e03fd220269016"
+
+  head "https://github.com/git/git.git"
+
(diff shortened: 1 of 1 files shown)


Formula/git-completion.rb
```diff
@@ -0,0 +1,20 @@
+class GitCompletion < Formula
+  desc "Bash completion for Git"
```
fxcoudert: Bash and Zsh

laurentr: Fixed!

AraHaan:
Mind squashing the commits?

laurentr:
@AraHaan Done.

fxcoudert:
The `git`formula already has an option (on by default) to install bash and zsh completions (and also the `git-prompt.sh` script). So why provide these as a separate formula?

fxcoudert:
The `conflicts_with` is a minimum, but it doesn't address the question: why would we want this? (Also, for zsh, system git comes with completion already!)

laurentr:
For folks who don't have git installed from brew.

fxcoudert:
macOS already has zsh completion installed in `/usr/share/zsh/5.2/functions/_git`. And for `bash`, if it is not already installed, then it needs to correspond to the system's `git` version, right?

laurentr:
Yeah, makes sense.

PR #10836, opened 2017-03-10
commandbox 3.6.0

bdw429s:
commandbox 3.6.0


neutric:
Thanks for taking the time to update `commandbox`, @bdw429s!

bdw429s:
@neutric Wow, thanks for the fast merge!  Ironically, I just found a small issues with the 3.6.0 binary so I'll probably be sending an updated sha over soon.  I was hoping to update the pull but you were too fast :)

PR #10838, opened 2017-03-10
gcc 7-20170305 (devel)

petere:
I needed a gcc-7 to test some software, so I found this handy. Maybe others could use it too.

They removed all the java stuff from gcc-7, so some of the build options could be reworked, but I figured that could wait until this becomes the stable version and gcc-6 gets moved off to its own formula.

They don't appear to have a generic download site for the snapshots, so I just grabbed a mirror from their list (https://gcc.gnu.org/mirrors.html).

### DIFF
--- a/Formula/gcc.rb
+++ b/Formula/gcc.rb
@@ -29,6 +29,11 @@ class Gcc < Formula
     sha256 "807d107ada2e8774a5bf1461d3f8dc636d1135bdfde5a64f03ccabc063a11328" => :yosemite
   end
 
+  devel do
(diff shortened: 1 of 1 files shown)


ilovezfs:
Other than doubling CI time, does this get us anything that `brew install --HEAD gcc` doesn't already provide?

fxcoudert:
We have turned down at least one previous request to do that. Please use `brew install --HEAD gcc` if you want to build the latest development branch of GCC. Although GCC project does emit snapshots, they are automatically released and not particularly stable compared to head.

petere:
It gets me the same versions that, say, Debian and Fedora are using.

Issue #10486, opened 2017-03-01
highlight 3.35: SHA256 seems to have changed


rwhogg:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is a problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

### Bug reports:

`highlight` source builds now fail due to a SHA256 mismatch:

```
==> Verifying highlight-3.35.tar.bz2 checksum
Error: SHA256 mismatch
Expected: 29b2111531230317fc6228b5f15ad0839448f20d65256279ac68d08319fa7a75
Actual: 8a14b49f5e0c07daa9f40b4ce674baa00bb20061079473a5d386656f6d236d05
Archive: /Users/rhogg/Library/Caches/Homebrew/highlight-3.35.tar.bz2
To retry an incomplete download, remove the file above.
```

The actual checksum is listed on the `highlight` homepage; I didn't see any reference to this change but didn't look too thoroughly.

```
brew doctor
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: You have unlinked kegs in your Cellar
Leaving kegs unlinked can lead to build-trouble and cause brews that depend on
those kegs to fail to run properly once built. Run `brew link` on these:
  gcc@4.8
  httpd24
  php70

Warning: Some installed formula are missing dependencies.
You should `brew install` the missing dependencies:
  brew install let-alist php55

Run `brew missing` for more details.

Warning: Some keg-only formula are linked into the Cellar.
Linking a keg-only formula, such as gettext, into the cellar with
`brew link <formula>` will cause other formulae to detect them during
the `./configure` step. This may cause problems when compiling those
other formulae.

Binaries provided by keg-only formulae may override system binaries
with other strange results.

You may wish to `brew unlink` these brews:
  mysql@5.6
bob@work:Homebrew rhogg (master $)$ brew config
HOMEBREW_VERSION: 1.1.10-432-ge02223960
ORIGIN: https://github.com/Homebrew/brew
HEAD: e022239608ca535fa3b56305eff8dd091b3fc008
Last commit: 3 hours ago
[...]

PR #10841, opened 2017-03-10
switch to HTTPS for SVN access on sf.net #1

vszakats:

### DIFF
--- a/Formula/astyle.rb
+++ b/Formula/astyle.rb
@@ -3,7 +3,7 @@ class Astyle < Formula
   homepage "https://astyle.sourceforge.io/"
   url "https://downloads.sourceforge.net/project/astyle/astyle/astyle%202.06/astyle_2.06_macos.tar.gz"
   sha256 "ad26b437365add1ec718b0f5f7c03ef0297616528619c2d1de19e940cd18d88a"
-  head "svn://svn.code.sf.net/p/astyle/code/trunk/AStyle"
+  head "https://svn.code.sf.net/p/astyle/code/trunk/AStyle"
 
   bottle do
     cellar :any_skip_relocation
--- a/Formula/ctags.rb
+++ b/Formula/ctags.rb
@@ -25,7 +25,7 @@ class Ctags < Formula
   end
 
   head do
-    url "http://svn.code.sf.net/p/ctags/code/trunk"
+    url "https://svn.code.sf.net/p/ctags/code/trunk"
     depends_on "autoconf" => :build
   end
 
--- a/Formula/dasm.rb
+++ b/Formula/dasm.rb
@@ -1,9 +1,9 @@
 class Dasm < Formula
   desc "Macro assembler with support for several 8-bit microprocessors"
-  homepage "https://dasm-dillon.sourceforge.io"
+  homepage "https://dasm-dillon.sourceforge.io/"
   url "https://downloads.sourceforge.net/project/dasm-dillon/dasm-dillon/2.20.11/dasm-2.20.11-2014.03.04-source.tar.gz"
   sha256 "a9330adae534aeffbfdb8b3ba838322b92e1e0bb24f24f05b0ffb0a656312f36"
-  head "svn://svn.code.sf.net/p/dasm-dillon/code/trunk"
+  head "https://svn.code.sf.net/p/dasm-dillon/code/trunk"
 
   bottle do
     cellar :any_skip_relocation
--- a/Formula/dosbox.rb
+++ b/Formula/dosbox.rb
@@ -12,7 +12,7 @@ class Dosbox < Formula
   end
 
   head do
-    url "http://svn.code.sf.net/p/dosbox/code-0/dosbox/trunk"
+    url "https://svn.code.sf.net/p/dosbox/code-0/dosbox/trunk"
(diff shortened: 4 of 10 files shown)


DomT4:
Isn't this going to have the same issues as always? That at least everyone not running Sierra (and possibly even Sierra users) can't access these `svn` checkouts securely? Or did they finally fix the incredibly annoying certificate errors?

vszakats:
Fair question. I'll bring alive my El Cap machine for a test.

ilovezfs:
@DomT4 we're considering that a non-bug bug. It deliberately skips the svn tests everywhere except Sierra now.

ilovezfs:
@DomT4 https://github.com/Homebrew/brew/pull/2262

vszakats:
(Still needs manual confirmation of the cert on El Capitan when using `/usr/bin/svn`.)

DomT4:
> we're considering that a non-bug bug. It deliberately skips the svn tests everywhere except Sierra now.

Yes, I know, I saw the commit at the time. I use a forked Homebrew so I check pretty much everything 😅. But I still think you'll have the situation at least on early versions of Sierra or everything including and before El Cap where users cannot execute `brew fetch` on these secure SVN URLs for at least SF, which has always been problematic in the past.

vszakats:
To address pre-Sierra cases, the protocol could automatically be downgraded on these systems along with a screen warning, if the current user prompt may be a blocker.

ilovezfs:
@DomT4 that sounds like a `brew fetch bug` if it fails instead of prompting :)

DomT4:
> @DomT4 that sounds like a brew fetch bug if it fails instead of prompting :)

It prompts you to accept the certificate IIRC, but Mike has always been (understandably, IMO) firm on Homebrew not requiring interactivity once a command has been issued. Maybe that's changed, but simply seems like a way to break automatic `brew fetch` executions on things like Travis, scripting, etc.

DomT4:
El Cap CI on the other PR confirms issues FWIW:
```
==> brew install --build-bottle --verbose spim
==> FAILED
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20170310-89304-j02f3n.sb nice /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/build.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/spim.rb --build-bottle --verbose
==> Cloning https://svn.code.sf.net/p/spimsimulator/code
==> Checking out 681
svn checkout https://svn.code.sf.net/p/spimsimulator/code /Users/brew/Library/Caches/Homebrew/spim--svn -r 681
svn: E170013: Unable to connect to a repository at URL 'https://svn.code.sf.net/p/spimsimulator/code'
svn: E230001: Server SSL certificate verification failed: issuer is not trusted
Error: Failed to download resource "spim"
Failure while executing: svn checkout https://svn.code.sf.net/p/spimsimulator/code /Users/brew/Library/Caches/Homebrew/spim--svn -r 681
```

Several of those.

vszakats:
Worst case, I can downgrade everything to HTTP://, so that it looks organised and they can be bumped to TLS in the future with a simple s&r. Or, I can add `MacOS.version` branching to fall back to HTTP:// on pre-Sierra (though I may need advise on the preferred style for that).

ilovezfs:
I think the cert only has to be accepted once. So we could just accept it on the CI boxes.

ilovezfs:
@DomT4 

>To address pre-Sierra cases, the protocol could automatically be downgraded on these systems along with a screen warning, if the current user prompt may be a blocker.

perhaps you'd be interested in working on adding that to the subversion download strategy? :innocent:

DomT4:
> perhaps you'd be interested in working on adding that to the subversion download strategy? 😇

Heh, I'm going to have to politely decline. I tried the opposite thing last month and didn't get anywhere 😸 https://github.com/Homebrew/brew/pull/1962

ilovezfs:
Another option would be to pass `--trust-server-cert --non-interactive` on < Sierra.

vszakats:
Sierra clean.

ilovezfs:
>Sierra clean.

Sounds like a cologne.

ilovezfs:
The head URLs are fine because they only go through audit not fetch.

I think until someone addresses this in Homebrew/brew, we should branch the stable URLs so that they don't fail CI.

An exception is is exult since it's now on GitHub: https://github.com/exult/exult/releases switching it over.

vszakats:
Okay, moving stable URL updates to a separate PR.

ilovezfs:
>Okay, moving stable URL updates to a separate PR.

yup that makes sense

ilovezfs:
Let's also remove `highlight` from the PR since steps mentioned here have not been followed yet to my knowledge: https://github.com/Homebrew/homebrew-core/issues/10486#issuecomment-284225257

ilovezfs:
Fine to update highlight's head url though if you want

vszakats:
Okay, removed anything but the head URL update from `hightlight`.

ilovezfs:
Thanks @vszakats. Shipped!

PR #10842, opened 2017-03-10
commandbox 3.6.0

bdw429s:
updated hashes

### DIFF
--- a/Formula/commandbox.rb
+++ b/Formula/commandbox.rb
@@ -2,7 +2,7 @@ class Commandbox < Formula
   desc "CFML embedded server, package manager, and app scaffolding tools"
   homepage "https://www.ortussolutions.com/products/commandbox"
   url "https://downloads.ortussolutions.com/ortussolutions/commandbox/3.6.0/commandbox-bin-3.6.0.zip"
-  sha256 "85c4336b50385eaf0194f1ff55bd79659f36a516ab0f2434c054a474b48ee7bb"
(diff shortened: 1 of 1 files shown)


bdw429s:
@neutric This is the same release I sent earlier today, but I needed to update the binaries on S3, so sending over an update with the new sha hashes.  Thanks!

fxcoudert:
Thanks @bdw429s for your contribution!

neutric:
Thanks also for providing context for the hash change. :D

ilovezfs:
In the future: please bump the version rather than just changing the checksums. As it is, people who have already downloaded the file are going to be getting checksum mismatch errors now.

>but I needed to update the binaries on S3

Why? This is still an unexplained change.

bdw429s:
@ilovezfs @nijikon I realize updating a binary is not desired.  Our build process had placed the wrong binary on S3.  I realized it right away but the original pull was merged so fast I didn't get a chance to update it ( https://github.com/Homebrew/homebrew-core/pull/10836 ).  All of the release material and documentation for the 3.6.0 release had been written so it was not a trivial change to bump the version. 

Please note, the first homebrew version had literally only been merged a couples hours prior and in the middle of the night so I would be surprised if anyone at all had downloaded it.  Had the old binary been there for weeks, I wouldn't have done this.

ilovezfs:
OK, thanks. Typically, uploading a duplicate named https://downloads.ortussolutions.com/ortussolutions/commandbox/3.6.0/commandbox-bin-3.6.0.1.zip would be the way to go in that case.

>in the middle of the night 

:night_with_stars: 

We have users all over the world ;)

PR #10858, opened 2017-03-10
Bulk import from homebrew/versions (2/7)

MikeMcQuaid:


Formula/ffmpeg@2.8.rb
```diff
@@ -0,0 +1,190 @@
+class FfmpegAT28 < Formula
+  desc "Play, record, convert, and stream audio and video"
+  homepage "https://ffmpeg.org/"
+  url "https://ffmpeg.org/releases/ffmpeg-2.8.6.tar.bz2"
```
retokromer: The last release of the 2.8 branch is 2.8.11 not the 2.8.6 (the current last release is 3.2.4).

ilovezfs:
CC @woodruffw RE: dylib ID of libglfw.dylib

ilovezfs:
@MikeMcQuaid 🍏 modulo the unholy download.sourceforge.net false positives.

MikeMcQuaid:
Thanks @ilovezfs ❤️

PR #10859, opened 2017-03-10
Bulk import from homebrew/versions (3/7)

MikeMcQuaid:


fxcoudert:
Some audits complaining about lack of test. More interesting is:

```
kibana:
      ln -s ../Formula/kibana.rb kibana@5.2
```

and 

```
kibana@4.1:
  * Use Language::Node for npm install args
kibana@4.4:
  * Use Language::Node for npm install args
```

Formula/jboss-as@5.rb
```diff
@@ -0,0 +1,27 @@
+class JbossAsAT5 < Formula
+  desc "JBoss Application Server"
+  homepage "https://www.jboss.org/jbossas"
+  url "https://downloads.sf.net/project/jboss/JBoss/JBoss-5.1.0.GA/jboss-5.1.0.GA.zip"
```
vszakats: It'd be nice to use the `downloads.sourceforge.net` domain to stay uniform.

fxcoudert:
Merged in #11260

MikeMcQuaid:
Thanks @fxcoudert!

PR #10861, opened 2017-03-10
Bulk import from homebrew/versions (5/7)

MikeMcQuaid:


fxcoudert:
```
open-mpi@1.6:
  * Stable: The URL http://www.open-mpi.org/software/ompi/v1.6/downloads/openmpi-1.6.5.tar.bz2 could use HTTPS rather than HTTP
  * Formula should have a desc (Description).
  * The URL http://www.open-mpi.org/ could use HTTPS rather than HTTP

percona-server@5.5:
  * `revision` (line 10) should be put before `keg_only` (line 8)
  * Stable: The URL http://www.percona.com/downloads/Percona-Server-5.5/Percona-Server-5.5.41-37.0/source/tarball/percona-server-5.5.41-37.0.tar.gz could use HTTPS rather than HTTP
  * The URL http://www.percona.com could use HTTPS rather than HTTP
  * Use '--with-test' instead of '--with-tests'. Migrate '--with-tests' with `deprecated_option`.

perl@5.14:
  * Use '--with-test' instead of '--with-tests'. Migrate '--with-tests' with `deprecated_option`.
  * W: 10: col 5: Literal `"OS X ships Perl and overriding that can cause unintended issues"` used in void context.
  * C: 10: col 5: Inconsistent indentation detected.

perl@5.18:
  * Use '--with-test' instead of '--with-tests'. Migrate '--with-tests' with `deprecated_option`.
  * W: 9: col 5: Literal `"OS X ships Perl and overriding that can cause unintended issues"` used in void context.
  * C: 9: col 5: Inconsistent indentation detected.

perl@5.18:
  * Use '--with-test' instead of '--with-tests'. Migrate '--with-tests' with `deprecated_option`.
  * W: 9: col 5: Literal `"OS X ships Perl and overriding that can cause unintended issues"` used in void context.
  * C: 9: col 5: Inconsistent indentation detected.

phantomjs@1.98:
  * Formula should have a desc (Description).
  * The URL http://www.phantomjs.org/ is not reachable
```

fxcoudert:
@BrewTestBot test this please

fxcoudert:
Merged with two more HTTP->HTTPS changes in the perl formulas. Only remaining audit issue was:

`percona-server@5.5:   * A `test do` test block should be added`

Given that the main `percona-server` formula itself does not have a test block, I think we can worry about that later.

chdiza:
How come in `homebrew-core` there is now `play@1.4` but not a plain `play`?

fxcoudert:
```
$ brew info play 
Play 2.3 replaces the play command with activator:
  brew install typesafe-activator

You can read more about this change at:
  https://www.playframework.com/documentation/2.3.x/Migration23
  https://www.playframework.com/documentation/2.3.x/Highlights23
```

ilovezfs:
Probably should have been deleted rather than imported.

chdiza:
@fxcoudert 

```
$ brew update
Already up-to-date.
$ brew edit play
Error: No available formula with the name "play"
$ brew desc play
Error: No available formula with the name "play"
```

chdiza:
Also, `cd` to the `Formula` dir:

```
$ ls | grep play
easyrpg-player.rb
libreplaygain.rb
madplay.rb
mplayer.rb
mplayershell.rb
play@1.4.rb
svtplay-dl.rb
tcpreplay.rb
wordplay.rb
```

ilovezfs:
The imported ones were added to homebrew/versions in https://github.com/Homebrew/homebrew-versions/pull/1232.

The original play formula was deleted in https://github.com/Homebrew/legacy-homebrew/pull/30262.

ilovezfs:
In the last 30 days homebrew/versions/play14 had 1 install, so there's that.

chdiza:
I don't particularly care whether play14 lives or dies. I don't even know what it is (which is why I tried to `brew desc` and `brew edit` it, and their failure led to my question).  But it seems to me that *if* it lives there should be an alias or a dummy formula or *something* that would make `brew [desc|edit] play` work.

chdiza:
There are others besides `play` like this.  I'll open an issue.

MikeMcQuaid:
`play` can just be deleted 👍

PR #10862, opened 2017-03-10
Bulk import from homebrew/versions (6/7)

MikeMcQuaid:


ilovezfs:
@MikeMcQuaid shouldn't it be ruby@1.8 and ruby@1.9?

ilovezfs:
and I guess redis@1.3 not redis@1.310?

ilovezfs:
or even redis@1?

MikeMcQuaid:
Fixed Ruby. Rebar is happening in https://github.com/Homebrew/homebrew-core/pull/10815 and Redis in https://github.com/Homebrew/homebrew-core/pull/10559

PR #10863, opened 2017-03-10
Bulk import from homebrew/versions (7/7)

MikeMcQuaid:


fxcoudert:
@BrewTestBot test this please

fxcoudert:
Fail to build: `subversion@1.8` and `unison@2.40`.
Test fails for `srtp@1.5` and `zeromq@4.0` (missing header in both cases).
The rest look good.

ilovezfs:
subversion problem is that the inline patch was not copied verbatim so the white space is borked.

ilovezfs:
I pushed a fix for unison@2.40

ilovezfs:
subversion
```
Possible unintended interpolation of @1 in string at (eval 26) line 1.
```

ilovezfs:
Build succeeds if the formula is named subversion18 instead of subversion@1.8.

ilovezfs:
I've pushed a fix for subversion@1.8.

ilovezfs:
@BrewTestBot test this please

fxcoudert:
Some style issues, some deprecated 

```
unison@2.40:
  * Stable: The URL http://www.seas.upenn.edu/~bcpierce/unison/download/releases/unison-2.40.128/unison-2.40.128.tar.gz could use HTTPS rather than HTTP
  * The URL http://www.cis.upenn.edu/~bcpierce/unison/ could use HTTPS rather than HTTP
subversion@1.8:
  * `depends_on` (line 38) should be put before `resource` (line 21)
  * :apr is deprecated. Usage should be "apr-util"
  * `env :userpaths` in formulae is deprecated
selenium-server-standalone@2.45:
```

MikeMcQuaid:
Let's hold off merging this until we have a new tag with https://github.com/Homebrew/brew/pull/1732.

@fxcoudert Thanks, will look at cleaning them up.

MikeMcQuaid:
Thanks for the help @fxcoudert, @ilovezfs!

ilovezfs:
yw

:confetti_ball:

PR #10903, opened 2017-03-11
gdbm 1.13

chdiza:

-----

The new `configure` flag is a workaround for what I believe is a newly-introduced bug.  The new version of `gdbm` "enhances" the enclosed `gdbmtool` to use `readline` functionality.  The problem (I think) is that upstream's detection of readline on macOS is broken.  Without the new flag, building fails with:
```
libtool: link: clang -g -O2 -o .libs/gdbm_load gdbm_load.o  ./libgdbmapp.a
./.libs/libgdbm.dylib -ldbm
  "_history_list", referenced from:
      _input_history_handler in input-rl.o
clang: error: linker command failed with exit code 1 (use -v to see
invocation)
make[3]: *** [gdbmtool] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [all] Error 2
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```
This happens outside of HB also.  (My hunch: their `configure` script thinks readline is present even if it isn't.)

I've reported this upstream, via email (since they appear to have no obviously functioning public bug tracker).

The version bump contains more than just the enhancement of `gdbmtool`, so I thought it was worth making a PR here.  If they fix the bug, we can remove the flag later.

### DIFF
--- a/Formula/gdbm.rb
+++ b/Formula/gdbm.rb
@@ -1,9 +1,9 @@
 class Gdbm < Formula
   desc "GNU database manager"
   homepage "https://www.gnu.org/software/gdbm/"
-  url "https://ftpmirror.gnu.org/gdbm/gdbm-1.12.tar.gz"
-  mirror "https://ftp.gnu.org/gnu/gdbm/gdbm-1.12.tar.gz"
-  sha256 "d97b2166ee867fd6ca5c022efee80702d6f30dd66af0e03ed092285c3af9bcea"
+  url "https://ftpmirror.gnu.org/gdbm/gdbm-1.13.tar.gz"
+  mirror "https://ftp.gnu.org/gnu/gdbm/gdbm-1.13.tar.gz"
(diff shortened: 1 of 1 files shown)


chdiza:
BTW, the new flag doesn't eliminate any functionality that *was* present in the previous version (1.12), as far as I know.

We could alternatively introduce an optional (or non-optional) `depends_on "readline"` here, but I'd rather wait to see what upstream says and add that later if need be.

Formula/gdbm.rb
```diff
@@ -19,6 +19,7 @@ def install
     args = %W[
       --disable-dependency-tracking
       --disable-silent-rules
+      --without-readline
```
fxcoudert: Please add a comment there + link to issue reported upstream

fxcoudert (changes requested):

Formula/gdbm.rb
chdiza: I'll add a comment, but there is no link for the issue because I reported it via email.

fxcoudert:
Thanks @chdiza for your contribution!

PR #10907, opened 2017-03-11
readline: drop universal

fxcoudert:
Former wine dependency.


MikeMcQuaid (approved):

tdsmith:
Please remember to pull and publish the new bottle when you bump the revision. (Should the revision have been bumped, here?)

ilovezfs:
>Should the revision have been bumped, here?

Yes, because it was ENV.universal_binary (i.e., always building universal), so the bottle contents are a tad different now :)

tdsmith:
Gotcha, thanks.

fxcoudert:
@tdsmith thanks for catching that… forgot this one was non optional universal

PR #10908, opened 2017-03-11
libffi 3.2.1

fxcoudert:
Removing universal option (wine does not need it anymore), updating to latest version.


ilovezfs:
I'll finish the original PR.

PR #10917, opened 2017-03-11
xz: drop universal

fxcoudert:


greensky01:
@fxcoudert any reason to drop universal support?

I need build rubygems  nokogiri on Mac, and it requires universal xz.

chdiza:
> I need build rubygems nokogiri on Mac, and it requires universal xz.

I don't think this is true.  I have built nokogiri repeatedly for over year and I don't have a universal `xz`.

MikeMcQuaid:
@greensky01 macOS has been 64-bit only since 10.6 so universal usage in Homebrew is deprecated.

JeanMertz:
FYI: [the docs of Nokogiri](http://www.nokogiri.org/tutorials/installing_nokogiri.html#mac_os_x) mention using `--universal`. I realise this is theirs to fix, but, I did have to do `brew unlink xz`, install Nokogiri, and link xz afterwards, for Nokogiri to work with the Homebrew installed xz.

This is 10.13 public beta, so that might be the reason, just thought I would share this tidbit of information here, for whomever this is useful.

PR #10927, opened 2017-03-11
y2j 1.1.1 (new formula)

trinitronx:
- Added new formula: `y2j`
- Uses python virtualenv to provide runtime dependency on PyYAML
- Uses optional dependency on docker for running on a system with only docker.

Closes wildducktheories/y2j#4

-----

### DIFF
--- a/Formula/y2j.rb
+++ b/Formula/y2j.rb
@@ -1,0 +1,60 @@
+class Y2j < Formula
+  include Language::Python::Virtualenv
+  desc "command-line tool for converting between YAML and JSON and vice versa."
+  homepage "https://github.com/wildducktheories/y2j"
+  url "https://github.com/wildducktheories/y2j/archive/v1.1.1.tar.gz"
+  sha256 "e11ac6886937b3c9784f61d3c77d00d9b4dbf3fa10ec6ddc6f847b68196d9f5d"
+  head "https://github.com/wildducktheories/y2j.git"
+
+  depends_on "jq" => :run
+  depends_on "base64" => :run
+  depends_on :python => :run
+  depends_on "docker" => :optional
+
+  resource "PyYAML" do
+    url "https://files.pythonhosted.org/packages/4a/85/db5a2df477072b2902b0eb892feb37d88ac635d36245a72a6a69b23b383a/PyYAML-3.12.tar.gz"
+    sha256 "592766c6303207a20efc445587778322d7f73b161bd994f227adaa341ba212ab"
+  end
+
+  def install
+    py_version = Language::Python.major_minor_version "python"
+    venv_site_packages_path = libexec/"lib/python#{py_version}/site-packages"
+    ENV["PYTHONPATH"] = venv_site_packages_path
+    venv = virtualenv_create(libexec, "python")
+
+    python_resources = resources.map(&:name).to_set
+    python_resources.each do |r|
+      venv.pip_install resource(r)
+    end
+
+    # Force our venv python for use with y2j
+    inreplace buildpath/"y2j.sh", /^#!.*bash/,
+                               "#!/usr/bin/env bash\nexport PYTHONPATH=#{venv_site_packages_path}:$PYTHONPATH"
+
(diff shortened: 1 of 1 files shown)


Formula/y2j.rb
```diff
@@ -0,0 +1,60 @@
...
+  def install
+    py_version = Language::Python.major_minor_version "python"
+    venv_site_packages_path = libexec/"lib/python#{py_version}/site-packages"
+    ENV["PYTHONPATH"] = venv_site_packages_path
+    venv = virtualenv_create(libexec, "python")
+
+    python_resources = resources.map(&:name).to_set
+    python_resources.each do |r|
+      venv.pip_install resource(r)
+    end
```
bfontaine: You can `venv.pip_install resources` instead.

Formula/y2j.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    # Force our venv python for use with y2j
+    inreplace buildpath/"y2j.sh", /^#!.*bash/,
+                               "#!/usr/bin/env bash\nexport PYTHONPATH=#{venv_site_packages_path}:$PYTHONPATH"
```
bfontaine: You can do so in the formula: `ENV["PYTHONPATH"] = …`.

Formula/y2j.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    install_script = `./y2j.sh installer #{prefix.to_s}`
+
+    File.open("install_y2j.sh", "w") do |f|
+      f.write(install_script)
+    end
```
bfontaine: Use `(buildpath/"install_y2j.sh").write install_script`.

Formula/y2j.rb
```diff
@@ -0,0 +1,60 @@
...
+    chmod 0755, "./install_y2j.sh"
+    system "./install_y2j.sh"
+
+    bin.install "y2j.sh", "y2j", "j2y", "yq"
```
bfontaine: Use `#{bin}` for the `install_script` so you can remove this line.

bfontaine (changes requested):

ahmetb:
@trinitronx thanks for contributing, please consider following up on the review comments.

fxcoudert:
Will be closing this one for now, as it seems inactive… please consider reopening once review comments are addressed.

Formula/y2j.rb
trinitronx: Not sure I understand what you mean on this one.  It looked like `#{bin}` just was an expansion of `#{prefix}/bin` which would be the formula's `bin` directory.

The problem here was that "`install_y2j.sh`" placed these as symlinks to the main `y2j.sh` script which lives in the upstream project's source repo as checked out by the `install_y2j.sh` script.  As such, the real bash script executable in this case is `y2j.sh` which is going to be located under something like: `/usr/local/Cellar/y2j/<version e.g.: 1.2.3>/y2j.sh`

It was my understanding that the usage of `bin.install` was meant for this use case.  That is: placing & setting the execute bit for executables such that they are [placed into the formula's `bin` dir][1] (e.g.: `/usr/local/Cellar/y2j/<version e.g.: 1.2.3>/bin/`).

[1]: https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#bininstall-foo

Formula/y2j.rb
trinitronx: wouldn't this only affect the runtime of the `brew install y2j` ruby script that is executed?  What do you mean by "in the formula"?  The need here was to ensure that the `y2j.sh` script environment uses the homebrew managed python virtualenv.  So my thought was:  just force a shebang line & `PYTHONPATH` variable set as a shim inside the `y2j.sh` script to set it for this tool.

trinitronx:
@fxcoudert: I can't seem to re-open this issue

trinitronx:
@bfontaine, @fxcoudert : I've made the changes requested and added this to a new Pull Request #12532 because I'm unable to re-open this.

PR #10931, opened 2017-03-12
remove unicorn glib dependency and add test

stephengroat:
unicorn no longer depends on glib in v1.0, should also use tests when needed

-----

### DIFF
--- a/Formula/unicorn.rb
+++ b/Formula/unicorn.rb
@@ -18,7 +18,7 @@ class Unicorn < Formula
   option "with-test", "Test build"
 
   depends_on "pkg-config" => :build
-  depends_on "cmocka" => "with-test"
+  depends_on "cmocka" => :build
 
   def install
     archs  = %w[x86 x86_64 arm mips]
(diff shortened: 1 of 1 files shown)


Formula/unicorn.rb
```diff
@@ -15,9 +15,10 @@ class Unicorn < Formula
   option "with-all", "Build with support for ARM64, Motorola 64k, PowerPC and "\
     "SPARC"
   option "with-debug", "Create a debug build"
+  option "with-test", "Test build"
 
-  depends_on "glib"
   depends_on "pkg-config" => :build
+  depends_on "cmocka" => :build
```
fxcoudert: Shouldn't it be `depends_on "cmocka" if build.with? "test"`?

fxcoudert (changes requested):

Formula/unicorn.rb
ilovezfs:
```
depends_on "cmocka" => :build if build.with? "test"
```

Issue #10920, opened 2017-03-11
option to skip fontconfig cache regeneration


vszakats:
First of all thanks for the new self-contained `wine` formula, I'm migrating from the wine-devel cask as we speak.

One concern with it, which I noticed on the first install: The infamous fontconfig cache regeneration is taking place as a post-install task and this may be a problem when installing the formula on CI machines. These are headless and certainly not GUI, so fonts should not matter. The regeneration process however takes several minutes (~2-5), which reduces the useful time to run the actual build and adds an uncomfortable wait to each CI session.

I had committed a patch to address this problem to the MSYS2 project here: Alexpux/MINGW-packages@fdea2f9
It solved the problem properly.

Do you think it would be feasible to implement a similar setting to the wine formula? I'm unsure about the best way, but maybe an --without-fc-cache or similar option would do. Or an envvar, like in MSYS2.

Any opinions?

vszakats:
Closing in favour of PR #10947.

vszakats:
Reopening this, as it is also an issue with `fontconfig` itself, even when it is being installed as a dependency.

The issue is that installing any of those packages that have `fontconfig` as a dependency, a several minutes long delay is required. This is fine on local machines, but quite a waste when using `homebrew` on a CI machine for automated builds (f.e. with Travis CI.)

The suggested `brew install --force-bottle --build-bottle wine` hack does work fine for `wine` but it's almost impossible to discover by a normal user and won't cover the cases when `fontconfig` is a normal dependency.

The previously proposed envvar-based solution is rather simple and painless.

vszakats:
Proposed patch: https://github.com/Homebrew/homebrew-core/compare/master...vszakats:fc

ilovezfs:
If it's a dependency, you can use the same arguments to install `fontconfig` before installing whatever is using it.

vszakats:
@ilovezfs That's true and thanks for the tip (which didn't occur me last night.)

I'd argue it's still not very ideal given that now it's required to know about the hack, to discover which packages may depend on `fontconfig` and manually install a dependency, if any of them do (in my case these are `cairo` and `gd`.)

I'm guessing there is quite a bit of CI cycles wasted on `fc-cache`.

mpsuzuki:
[...]

PR #10947, opened 2017-03-12
wine: add option to disable fc-cache run at post-install

vszakats:

-----

First of all thanks for the new self-contained `wine` formula, I'm migrating from the wine-devel cask as we speak.

One concern with it, which I noticed on the first install: The infamous fontconfig cache regeneration is taking place as a post-install task and this may be a problem when installing the formula on CI machines. These are headless and certainly not GUI, so fonts should not matter. The regeneration process however takes several minutes (~2-5), which reduces the useful time to run the actual build and adds an uncomfortable wait to each CI session.

I had committed a patch to address this problem to the MSYS2 project here: Alexpux/MINGW-packages@fdea2f9
It solved the problem properly.

Current patch is just an initial proposal and any advice for finding the best solution is much welcome.

Patch also contains an unrelated audit fix [UPDATE: moved to #10952), plus leaves this other audit error unfixed:
```
resource "openssl": The URL https://dl.bintray.com/homebrew/mirror/openssl-1.0.2k.tar.gz is not reachable (HTTP status code 404)
```

### DIFF
--- a/Formula/fontconfig.rb
+++ b/Formula/fontconfig.rb
@@ -62,8 +62,12 @@ class Fontconfig < Formula
   end
 
   def post_install
-    ohai "Regenerating font cache, this may take a while"
-    system "#{bin}/fc-cache", "-frv"
+    if ENV["HOMEBREW_FONTCONFIG_CACHE_SKIP"] = "yes"
+      ohai "HOMEBREW_FONTCONFIG_CACHE_SKIP set to 'yes', skipping font cache update"
+    else
+      ohai "Regenerating font cache, this may take a while"
+      system "#{bin}/fc-cache", "-frv"
+    end
   end
 
   test do
--- a/Formula/wine.rb
+++ b/Formula/wine.rb
@@ -491,8 +491,12 @@ class Wine < Formula
 
   def post_install
     # For fontconfig
-    ohai "Regenerating font cache, this may take a while"
-    system "#{libexec}/bin/fc-cache", "-frv"
+    if ENV["HOMEBREW_FONTCONFIG_CACHE_SKIP"] = "yes"
+      ohai "HOMEBREW_FONTCONFIG_CACHE_SKIP set to 'yes', skipping font cache update"
+    else
+      ohai "Regenerating font cache, this may take a while"
+      system "#{libexec}/bin/fc-cache", "-frv"
+    end
 
     # For net-snmp
     (var/"db/net-snmp_vendored_wine").mkpath


MikeMcQuaid:
The problem with making this an option is you won't use the binary package.

vszakats:
@MikeMcQuaid That's a stopper. What about using a properly named envvar for this purpose? (the way it's done in MSYS2)

MikeMcQuaid:
That will work for now but the longer-term plan is to filter out such things.

vszakats:
Updated the logic to use that till it is allowed. Naming/syntax/etc comments welcome.

fxcoudert:
This seems very ad hoc. Also, now that wine has much fewer dependencies (and even fewer uses), I don't see that it would be tested too often by the CI.

`fc-cache` takes 60 seconds on my laptop, which has 1360 fonts installed. On a CI system, I am sure there would be far fewer fonts.

vszakats:
Hard to tell how much time it takes exactly, but I meant using `brew install wine` as part of 3rd party build scripts running on CI systems (and not Homebrew Jenkins CI runs). F.e. I'm using `wine` in cross-builds, and it's being run on each commit to those particular projects. The total run time for `brew install` doesn't look [that bad](https://travis-ci.org/vszakats/harbour-core/jobs/210237972#L820), so maybe macOS systems are much less prone to this than Windows was. Anyhow, it's time wasted and it's easy to runs over CI limits, so I try to avoid any unnecessary waits.

fxcoudert:
Re-installing `wine` for every commit seems overkill, couldn't you find a build system where you don't do that?

If we were doing something like that, then I think it would be better to have a `brew install --skip-postinstall` option really, to make it generic at least.

vszakats:
Maybe I'm missing something fundamental, but can't see how installing a package could be avoided on each CI run, given that `wine` doesn't come installed with the provided OS image. If there is a way to do that I'm most definitely interested.

`brew install --skip-postinstall` sounds good to me.

ilovezfs:
Pretty sure `brew install --force-bottle --build-bottle wine` does exactly that.

vszakats:
I'm confused. How could this be used to avoid reinstalling `wine` let's say on each Travis CI run?

For context, here are the two lines that do the 'before install' tasks for a project cross-build (macOS to Windows): https://github.com/vszakats/harbour-core/blob/7174667dc85a83418f6fab4f210f8850d7a55b4e/.travis.yml#L69-L70

ilovezfs:
Because it's magic.
```
iMac-TMP:homebrew-core joe$ brew install --force-bottle --build-bottle wine
==> Downloading https://homebrew.bintray.com/bottles/wine-2.0.el_capitan.bottle.1.tar.gz
==> Pouring wine-2.0.el_capitan.bottle.1.tar.gz
==> Not running post_install as we're building a bottle
You can run it manually using `brew postinstall wine`
==> Caveats
You may want to get winetricks:
  brew install winetricks

By default Wine uses a native Mac driver. To switch to the X11 driver, use
regedit to set the "graphics" key under "HKCUSoftwareWineDrivers" to
"x11" (or use winetricks).

For best results with X11, install the latest version of XQuartz:
  https://xquartz.macosforge.org/
==> Summary
🍺  /usr/local/Cellar/wine/2.0: 8,227 files, 580.2MB
iMac-TMP:homebrew-core joe$
```

vszakats:
Cool, so it's a way to avoid running post-install I stand clarified. It's hackish as hell, but it will be super useful for a test run to see what is the fc-cache rebuild overhead.

ilovezfs:
>It's hackish as hell

Understatement.

vszakats:
[With fc-cache](https://travis-ci.org/vszakats/harbour-core/jobs/210262089#L824):
```
real	2m13.257s
user	1m17.520s
sys	0m22.499s
```
[Without fc-cache](https://travis-ci.org/vszakats/harbour-core/jobs/210263584#L823):
```
real	0m44.425s
user	0m18.297s
sys	0m11.554s
```

So it takes 1.5 minutes. It's the weekend, with lower than usual machine loads. Not terrible, but I'd be happy to shave it off as standard procedure.

MikeMcQuaid:
As a random anecdote: I also find this personally annoying on the `fontconfig` formula so I wonder if there's any way to speed up the cache building for both.

vszakats:
As a trivial extension of the proposed rough approach, the envvar renamed to `HOMEBREW_FC_CACHE_SKIP` (or `HOMEBREW_FONTCONFIG_CACHE_SKIP`) could control this in both formulae.

fxcoudert:
But in real use cases, we need the cache to be built…

vszakats:
I consider the CI use-case a real one.

ilovezfs:
semantics

vszakats:
Updated the patch to work on both `wine` and `fontconfig`.

vszakats:
Will use the hack @ilovezfs suggested. Closing.

vszakats:
Reopened the original Issue for re-evaluation: https://github.com/Homebrew/homebrew-core/issues/10920

Issue #10824, opened 2017-03-09
MacVim and Vim use different versions of Ruby


shadowhand:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is a problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

_You can erase any parts of this template not applicable to your Issue._

### Bug

MacVim is linking itself to system version of Ruby. Vim links itself to Homebrew Ruby. This causes any Vim plugin (such as [Command-T](https://github.com/wincent/Command-T)) to be usable in either MacVim *or* Vim, but not both.

MikeMcQuaid:
We don't consider this a bug, sorry.

shadowhand:
@MikeMcQuaid can you explain why?

MikeMcQuaid:
`vim` uses `depends_on :ruby` and MacVim does not. By default both will link against the system Ruby. `vim` provides an option to not do so or will not do so if building from source and MacVim does not. This is expected behaviour and I'd sooner remove the option from `vim` than add it to MacVim.

MikeMcQuaid:
Also on https://github.com/macvim-dev/macvim/issues/468 pretty sure it's not an upstream bug as if you use the `--enable-rubyinterp` I suspect it'll work. As you're building from source anyway you could maintain a custom formula fork with this change.

shadowhand:
@MikeMcQuaid 

> By default both will link against the system Ruby. 

I don't believe this is an accurate statement, as [vim formula](https://github.com/Homebrew/homebrew-core/blob/5d2d43ce5b534f4b209e1b35fa63b1db71326090/Formula/vim.rb#L26) clearly [states it will default to Homebrew Ruby](https://github.com/Homebrew/homebrew-core/blob/5d2d43ce5b534f4b209e1b35fa63b1db71326090/Formula/vim.rb#L98-L99).

Why have different defaults for two similar packages?

MikeMcQuaid:
> I don't believe this is an accurate statement, as vim formula clearly states it will default to Homebrew Ruby.

By "default" means "you're using /usr/local and the binary package".

> Why have different defaults for two similar packages?

Regardless of the past reasons: I've stated my reasons for why it's not considered a bug.

shadowhand:
@MikeMcQuaid 

[...]

PR #10951, opened 2017-03-12
MacVim: Link mvim with brewed ruby & perl

CarterLi:

-----

Like what [vim does](https://github.com/Homebrew/homebrew-core/blob/master/Formula/vim.rb#L39) for long time. This requires upstream change https://github.com/macvim-dev/macvim/commit/9d23d58d4c8800b395986fbcdeba4c80e7afe4d7
Tested with `--HEAD`

Also add `gettext` option.
Note: MacVim won't bundle `.po` files by default, you may copy them into `MacVim.app/Contents/Resources/vim/runtime/lang` manually

Also fix tests when built with python3 ;)

### DIFF
--- a/Formula/macvim.rb
+++ b/Formula/macvim.rb
@@ -14,6 +14,7 @@ class Macvim < Formula
   end
 
   option "with-override-system-vim", "Override system vim"
+  option "with-gettext", "Build MacVim with +gettext."
 
   deprecated_option "override-system-vim" => "with-override-system-vim"
 
@@ -21,13 +22,18 @@ class Macvim < Formula
   depends_on "cscope" => :recommended
   depends_on "lua" => :optional
   depends_on "luajit" => :optional
+  depends_on "gettext" => :optional
 
   if MacOS.version >= :mavericks
     option "with-custom-python", "Build with a custom Python 2 instead of the Homebrew version."
+    option "with-custom-ruby", "Build with a custom Ruby instead of the Homebrew version."
+    option "with-custom-perl", "Build with a custom Perl instead of the Homebrew version."
   end
 
   depends_on :python => :recommended
   depends_on :python3 => :optional
+  depends_on :ruby => "1.8" # Can be compiled against 1.8.x or >= 1.9.3-p385.
+  depends_on :perl => "5.3"
 
   def install
     # Avoid "fatal error: 'ruby/config.h' file not found"
@@ -87,6 +93,8 @@ class Macvim < Formula
       args << "--enable-pythoninterp"
     end
 
+    args << "--disable-nls" if build.without? "gettext"
+
     system "./configure", *args
     system "make"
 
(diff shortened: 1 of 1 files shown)


CarterLi:
Related issues: https://github.com/Homebrew/homebrew-core/issues/10824

ilovezfs:
I have déjà vu.

Formula/macvim.rb
```diff
@@ -14,20 +14,29 @@ class Macvim < Formula
   end
 
   option "with-override-system-vim", "Override system vim"
+  option "with-gettext", "Build MacVim with +gettext."
 
   deprecated_option "override-system-vim" => "with-override-system-vim"
 
+  conflicts_with "vim",
+    :because => "They both override system vim" if build.with? "override-system-vim"
```
ilovezfs: This won't work properly unless you also inspect the `tab` (INSTALL_RECEIPT) of vim to see whether it was built with `override-system-vi`.

CarterLi: Huh?
[override-system-vi](https://github.com/Homebrew/homebrew-core/blob/master/Formula/vim.rb#L115) only generates `symlink "vim" => "vi"`, system global symlink `vim` is always generated whatever override-system-vi or not.

ilovezfs: Yes you're right.

ilovezfs:
I suspect these changes are likely to be quite controversial since you're adding rather heavy new dependencies here.

CarterLi:
I can make brewed ruby & perl optional @ilovezfs

ilovezfs:
Let's see what other people think.

Formula/macvim.rb
```diff
@@ -14,20 +14,29 @@ class Macvim < Formula
   end
 
   option "with-override-system-vim", "Override system vim"
+  option "with-gettext", "Build MacVim with +gettext."
```
dunn: The `depends_on "gettext" => :optional` below will generate the option text, so we don't need this line.

Formula/macvim.rb
```diff
@@ -14,20 +14,29 @@ class Macvim < Formula
...
 
   deprecated_option "override-system-vim" => "with-override-system-vim"
 
+  conflicts_with "vim",
```
dunn: It's been a while, but [last I checked](https://github.com/Homebrew/legacy-homebrew/pull/42314) we couldn't do conditional formulae conflicts.  Has that changed?

ilovezfs: @dunn I don't know if anything has changed, but look at the current version `moreutils` in core. It's baaaack.

Formula/macvim.rb
CarterLi: Removed

Formula/macvim.rb
dunn: D:

chdiza:
> I suspect these changes are likely to be quite controversial since you're adding rather heavy new dependencies here.

This.  Don't add those, please.

CarterLi:
Upstream have released [a new version](https://github.com/macvim-dev/macvim/releases/tag/snapshot-128);this PR is ready for merging.
For new dependencies: [vim formula can](https://github.com/Homebrew/homebrew-core/blob/master/Formula/vim.rb#L39), why MacVim can't?

ilovezfs:
@DomT4 any thoughts here?

DomT4:
On the one hand I lean towards @chdiza's & your position pretty strongly, but on the other hand when this was done to `vim` I think we got <15 complaints out of the thousands of people who install `vim` regularly, so. I think _most people_ care less here than we expected, and I think you'd get similar results again.

The backticks in `test do` are nasty though 😝.

ilovezfs:
Fits of domitus. This is definitely encouraging.

chdiza:
> For new dependencies: vim can, why MacVim can't?

Vim should not have ever been made to depend on ruby (or any other scripting language, for that matter).

Also, MacVim is historically much more finicky than Vim about building against Ruby, both in and out of Homebrew.

CarterLi:
I've made dependencies `ruby` & `perl` optional.
Any chance to merge?

thefonso:
@CarterLi  this crazy sauce with vim and macvim has persisted for years. The macvim team seems to be stuck. If you cats on homebrew have a workable solution, I hope you can make this happen! ^_^

PR #10967, opened 2017-03-12
add test for wget

stephengroat:

-----

### DIFF
--- a/Formula/wget.rb
+++ b/Formula/wget.rb
@@ -25,6 +25,7 @@ class Wget < Formula
   deprecated_option "enable-debug" => "with-debug"
 
   option "with-debug", "Build with debug support"
+  option "with-test", "Build with test"
 
   depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


Formula/wget.rb
```diff
@@ -33,6 +34,8 @@ class Wget < Formula
   depends_on "libmetalink" => :optional
   depends_on "gpgme" => :optional
 
+  patch :p0, :DATA
```
fxcoudert: If the test fails, we need it reported upstream and use the upstream commit as a `patch`

stephengroat: will do, working out the patch

stephengroat: added

MikeMcQuaid:
@stephengroat I think we talked about this before, no? We don't want `make check` tests as they are inferior to `test do` ones and don't actually catch bugs. They are for upstream to be running as part of CI and the release process and not us. The tiny exceptions may be crypto software. That this requires a patch to work should indicate that this test is not useful.

PR #10968, opened 2017-03-12
scsh: migrate to boneyard

fxcoudert:
Created with `brew boneyard-formula-pr` because it is unmaintained.


fxcoudert:
`scsh` is unmaintained (last release on May 16, 2006). It only builds in 32-bit mode and has very few users according to analytics.

ilovezfs (approved):

PR #11023, opened 2017-03-13
recipes 0.22.0

tschoonj:


Formula/recipes.rb
```diff
@@ -23,11 +25,16 @@ def install
                           "--disable-debug",
                           "--prefix=#{prefix}",
                           "--disable-autoar",
-                          "--disable-gspell"
+                          "--disable-gspell",
+                          "--disable-schemas-compile"
```
fxcoudert: Why compile the schemas at post-install?

Formula/recipes.rb
```diff
@@ -1,17 +1,19 @@
 class Recipes < Formula
   desc "Formula for GNOME recipes"
   homepage "https://wiki.gnome.org/Apps/Recipes"
-  url "https://download.gnome.org/sources/recipes/0.4/recipes-0.4.2.tar.xz"
-  sha256 "9554d4f5d97eb9cd4032de0e4f9cc27a218c32d022dd1917a7e9efbd379c5bc1"
+  url "https://download.gnome.org/sources/gnome-recipes/0.22/gnome-recipes-0.22.0.tar.xz"
+  sha256 "ae4f669b1f1d20f2846ad1c9f6ba6580a15347288ef3d044972c2ba589d9c7b9"
```
fxcoudert: Should we rename the formula to `gnome-recipes`?

Formula/recipes.rb
tschoonj: Formulas that use glib schemas all end up being compiled into /usr/local/share/glib-2.0/schemas/gschemas.compiled, which is shared by all.

I we allow formulas to produce their own gschemas.compiled files you will get a `brew link` error, so it needs to happen during post-install.

Formula/recipes.rb
tschoonj: AFAIK we don't do actual formula renames. At least in the past this was handled with an alias.

fxcoudert: I think it has happened in the past: rename + alias for the old name
https://github.com/Homebrew/homebrew-core/commit/2c6fd9c6c0e97c68debe2b07593c6e61558f9713

tschoonj: This is new to me.

I certainly don't mind doing something similar for this Formula.

fxcoudert (approved):

fxcoudert:
Oops, I didn't realise there were other reviews requested. Not reverting right now cause I think it might be worse, sorry guys :(

MikeMcQuaid:
👍 to me.

PR #9814, opened 2017-02-09
elasticsearch 2.3

maxpsq:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #11048, opened 2017-03-13
elasticsearch 5.0.2

ygale:
This is an exact copy of the formula for elasticsearch 5.2.2, currently the latest on homebrew, except modified to install a recent previous version of Elasticsearch. This version of elasticsearch is the one currently supported by [bloodhound](https://hackage.haskell.org/package/bloodhound), an elasticsearch bindings library for Haskell, so it would be nice to have this version in homebrew.

Note that `brew test` requires elasticsearch to be running already on localhost. I assume that is the case for existing elasticsearch formulas as well.

Also note that elasticsearch will not run (and therefore `brew test` will fail) if there is existing data in `/usr/local/var/elasticsearch` created by a previous `brew install` of a different version of elasticsearch, since elasticsearch data files are version-specific. This probably ought to be managed automatically by `brew link`, but it is not. That also appears to be the case for the existing formulas.

-----

### DIFF
--- a/Formula/elasticsearch@5.0.2.rb
+++ b/Formula/elasticsearch@5.0.2.rb
@@ -1,0 +1,130 @@
+class ElasticsearchAT502 < Formula
+  desc "Distributed search & analytics engine"
+  homepage "https://www.elastic.co/products/elasticsearch"
+  url "https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.0.2.tar.gz"
+  sha256 "bbe761788570d344801cb91a8ba700465deb10601751007da791743e9308cb83"
+
+  head do
+    url "https://github.com/elasticsearch/elasticsearch.git"
+    depends_on "gradle" => :build
+  end
+
+  bottle :unneeded
+
+  depends_on :java => "1.8+"
+
+  def cluster_name
+    "elasticsearch_#{ENV["USER"]}"
+  end
+
+  def install
+    if build.head?
+      # Build the package from source
+      system "gradle", "clean", ":distribution:tar:assemble"
+      # Extract the package to the tar directory
+      mkdir "tar"
+      cd "tar"
+      system "tar", "--strip-components=1", "-xf", Dir["../distribution/tar/build/distributions/elasticsearch-*.tar.gz"].first
+    end
+
+    # Remove Windows files
+    rm_f Dir["bin/*.bat"]
+    rm_f Dir["bin/*.exe"]
+
+    # Install everything else into package directory
+    libexec.install "bin", "config", "lib", "modules"
+
+    # Set up Elasticsearch for local development:
+    inreplace "#{libexec}/config/elasticsearch.yml" do |s|
+      # 1. Give the cluster a unique name
+      s.gsub!(/#\s*cluster\.name\: .*/, "cluster.name: #{cluster_name}")
+
+      # 2. Configure paths
+      s.sub!(%r{#\s*path\.data: /path/to.+$}, "path.data: #{var}/elasticsearch/")
+      s.sub!(%r{#\s*path\.logs: /path/to.+$}, "path.logs: #{var}/log/elasticsearch/")
+    end
+
+    inreplace "#{libexec}/bin/elasticsearch.in.sh" do |s|
+      # Configure ES_HOME
+      s.sub!(%r{#\!/bin/bash\n}, "#!/bin/bash\n\nES_HOME=#{libexec}")
+    end
+
+    inreplace "#{libexec}/bin/elasticsearch-plugin" do |s|
+      # Add the proper ES_CLASSPATH configuration
+      s.sub!(/SCRIPT="\$0"/, %Q(SCRIPT="$0"\nES_CLASSPATH=#{libexec}/lib))
+      # Replace paths to use libexec instead of lib
+      s.gsub!(%r{\$ES_HOME/lib/}, "$ES_CLASSPATH/")
+    end
+
+    # Move config files into etc
+    (etc/"elasticsearch").install Dir[libexec/"config/*"]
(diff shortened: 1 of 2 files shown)


formula_renames.json
```diff
@@ -36,6 +36,7 @@
   "eigen32": "eigen@3.2",
   "elasticsearch17": "elasticsearch@1.7",
   "elasticsearch24": "elasticsearch@2.4",
+  "elasticsearch502": "elasticsearch@5.0.2",
```
neutric: This is not necessary as you are not renaming a formula previously called 'elasticsearch502'.

Formula/elasticsearch@5.0.2.rb
```diff
@@ -0,0 +1,130 @@
...
+  desc "Distributed search & analytics engine"
+  homepage "https://www.elastic.co/products/elasticsearch"
+  url "https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.0.2.tar.gz"
+  sha256 "bbe761788570d344801cb91a8ba700465deb10601751007da791743e9308cb83"
+
+  head do
+    url "https://github.com/elasticsearch/elasticsearch.git"
+    depends_on "gradle" => :build
+  end
+
```
neutric: The `head` block is not needed, one would use the main `elasticsearch` formula to build from head.

Formula/elasticsearch@5.0.2.rb
```diff
@@ -0,0 +1,130 @@
...
+  def install
+    if build.head?
+      # Build the package from source
+      system "gradle", "clean", ":distribution:tar:assemble"
+      # Extract the package to the tar directory
+      mkdir "tar"
+      cd "tar"
+      system "tar", "--strip-components=1", "-xf", Dir["../distribution/tar/build/distributions/elasticsearch-*.tar.gz"].first
+    end
+
```
neutric: See above comments on `head`.

neutric (commented):
@ygale: Thanks for your contribution to Homebrew! I commented in a few places, but it is probably wise to wait with changes until other maintainers chime in as I am not sure that this meets the criteria for [acceptable versioned formulae](https://github.com/Homebrew/brew/blob/master/docs/Versions.md#acceptable-versioned-formulae).

fxcoudert:
Why do we want to maintain a specific 5.0.2 versioned formula?

ilovezfs:
@fxcoudert https://github.com/Homebrew/brew/blob/master/docs/Versions.md

Specific patch versions are not acceptable, as specified in the docs.

Since the current version of elasticsearch is 5.2.2, that means this would need to be named `elasticsearch@5.0` (even though it would use the 5.0.2 tarball).

fxcoudert:
But the, is elastic search 5.0 _“used by a large number of people and still supported by their upstream projects”_? (which is the criterion for inclusion of versioned formulas)

ilovezfs:
@fxcoudert indeed. But: https://github.com/Homebrew/homebrew-core/pull/9814. So :man_shrugging:

neutric:
@ygale We'll wait and see what @MikeMcQuaid has to say. Should this get a thumbs down, consider hosting this formula in a [personal tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

Formula/elasticsearch@5.0.2.rb
```diff
@@ -0,0 +1,130 @@
+class ElasticsearchAT502 < Formula
```
MikeMcQuaid: `ElasticsearchAT50` and the file named `elasticsearch@5.0`

MikeMcQuaid (changes requested):

MikeMcQuaid:
> This version of elasticsearch is the one currently supported by bloodhound, an elasticsearch bindings library for Haskell, so it would be nice to have this version in homebrew.

When you say "currently supported" does Bloodhound fail to work with the latest formula version?

ilovezfs:
@jasontedor what is upstream's policy regarding which prior version(s) of the software are or are not supported?

jasontedor:
> what is upstream's policy regarding which prior version(s) of the software are or are not supported?

We support this, but we do not maintain it (except for critical security patches). Does that make sense? Effectively what this means is that we encourage users to be on the latest maintained release (currently 5.2.2). There are some bad bugs in 5.0.2 (I know of at least two), I'd prefer that we keep it simple and only provide latest.

ilovezfs:
@jasontedor thanks for that. Just to make sure you have all the facts, there are currently elasticsearch 1.7.x, elasticsearch 2.3.x, and elasticsearch 2.4.x formulae in homebrew.

jasontedor:
We can remove 2.3.x. I'll open a PR for this.

jasontedor:
And 1.7, that is also EOL.

To be clear, we still maintain and support 2.4.x, and we support 2.3.x but do not maintain it (except for critical security patches).

ilovezfs:
@jasontedor thanks. So if I'm following, there are

maintained, supported, current: 5.2.x
maintained, supported, not current: 2.4.x
unmaintained, supported, not current: 2.3.x, 5.0.x
unsupported, unmaintained, and EOL: 1.7

jasontedor:
Correct, with one small addition: there is also 5.1.x which is unmaintained, supported, and not correct. Perhaps:

5.2.x: maintained, supported, current (until 5.3.0 is released)
all prior 5.x are unmaintained, supported, current
2.4.x: maintained, supported. not current
all prior 2.x are unmaintained, supported, not current
all prior 1.x and 0.x are EOL

jasontedor:
I opened #11361.

fxcoudert:
Thanks @jasontedor for commenting, closing this PR accordingly.

PR #11054, opened 2017-03-13
Updated sshguard formula to use new config file, Sierra unified logging.

jinschoi:
SSHguard v2 requires a config file that does not get installed with make install. This commit creates that file and configures it for ipfw/pf and specifies the proper logfile, or a LOGREADER to use macOS unified logging for Sierra. Also removes some configure flags that are no longer valid.

### DIFF
--- a/Formula/sshguard.rb
+++ b/Formula/sshguard.rb
@@ -75,6 +75,6 @@ class Sshguard < Formula
   end
 
   test do
-    assert_match "SSHGuard #{version.to_s}", shell_output("#{sbin}/sshguard -v 2>&1")
+    assert_match "SSHGuard #{version}", shell_output("#{sbin}/sshguard -v 2>&1")
   end
 end
--- a/Formula/sshguard.rb
+++ b/Formula/sshguard.rb
@@ -23,7 +23,7 @@ class Sshguard < Formula
     inreplace "examples/sshguard.conf" do |s|
       s.gsub! /^#BACKEND=.*$/, "BACKEND=\"#{libexec}/sshg-fw-#{firewall}\""
       if MacOS.version >= :sierra
-        s.gsub! %r{^#LOGREADER="/usr\/bin\/log}, "LOGREADER=\"/usr/bin/log"
+        s.gsub! %r{^#LOGREADER="/usr/bin/log}, "LOGREADER=\"/usr/bin/log"
       else
         s.gsub! /^#FILES.*$/, "FILES=#{log_path}"
       end
@@ -75,6 +75,6 @@ class Sshguard < Formula
   end
 
   test do
-    assert_match "SSHGuard #{version.to_s}", shell_output("#{sbin}/sshguard -v 2>&1", 0)
+    assert_match "SSHGuard #{version.to_s}", shell_output("#{sbin}/sshguard -v 2>&1")
   end
 end
--- a/Formula/sshguard.rb
+++ b/Formula/sshguard.rb
@@ -12,18 +12,14 @@ class Sshguard < Formula
     sha256 "20918ad422229e3d3fec35af56916d396963014b95341206bd1e905a3f553384" => :yosemite
   end
 
-  depends_on "automake" => :build
-  depends_on "autoconf" => :build
-
   def install
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
                           "--prefix=#{prefix}",
                           "--sysconfdir=#{etc}"
     system "make", "install"
-    mkdir "etc"
-    cp "examples/sshguard.conf.sample", "etc/sshguard.conf"
-    inreplace "etc/sshguard.conf" do |s|
+    cp "examples/sshguard.conf.sample", "examples/sshguard.conf"
+    inreplace "examples/sshguard.conf" do |s|
       s.gsub! /^#BACKEND=.*$/, "BACKEND=\"#{libexec}/sshg-fw-#{firewall}\""
       if MacOS.version >= :sierra
         s.gsub! %r{^#LOGREADER="/usr\/bin\/log}, "LOGREADER=\"/usr/bin/log"
@@ -31,7 +27,7 @@ class Sshguard < Formula
         s.gsub! /^#FILES.*$/, "FILES=#{log_path}"
       end
     end
-    etc.install "etc/sshguard.conf"
+    etc.install "examples/sshguard.conf"
   end
 
   def firewall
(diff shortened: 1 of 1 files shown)


fxcoudert:
```
==> /usr/local/Cellar/sshguard/2.0.0/sbin/sshguard -v 2>&1
Error: sshguard: failed
<78> expected but was
<0>.
```

Could we also have a test that isn't just `sshguard -v`?

Formula/sshguard.rb
```diff
@@ -16,12 +16,22 @@ class Sshguard < Formula
   depends_on "autoconf" => :build
```
fxcoudert: Does it actually need `automake` and `autoconf`? It seems that the `configure` script is already present.

Formula/sshguard.rb
```diff
@@ -16,12 +16,22 @@ class Sshguard < Formula
...
-    system "./configure", "--disable-debug",
-                          "--disable-dependency-tracking",
+    system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
                           "--prefix=#{prefix}",
-                          "--with-firewall=#{firewall}"
+                          "--sysconfdir=#{etc}"
     system "make", "install"
+    mkdir "etc"
+    cp "examples/sshguard.conf.sample", "etc/sshguard.conf"
```
fxcoudert: `mkdir` and `cp` can be combined into: `etc.install "examples/sshguard.conf.sample" => "sshguard.conf"`

Formula/sshguard.rb
```diff
@@ -16,12 +16,22 @@ class Sshguard < Formula
...
+    inreplace "etc/sshguard.conf" do |s|
+      s.gsub! /^#BACKEND=.*$/, "BACKEND=\"#{libexec}/sshg-fw-#{firewall}\""
+      if MacOS.version >= :sierra
+        s.gsub! %r{^#LOGREADER="/usr\/bin\/log}, "LOGREADER=\"/usr/bin/log"
+      else
+        s.gsub! /^#FILES.*$/, "FILES=#{log_path}"
+      end
+    end
+    etc.install "etc/sshguard.conf"
```
fxcoudert: Remove the `etc.install` from here

Formula/sshguard.rb
```diff
@@ -16,12 +16,22 @@ class Sshguard < Formula
...
-                          "--disable-dependency-tracking",
+    system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
                           "--prefix=#{prefix}",
-                          "--with-firewall=#{firewall}"
+                          "--sysconfdir=#{etc}"
     system "make", "install"
+    mkdir "etc"
+    cp "examples/sshguard.conf.sample", "etc/sshguard.conf"
+    inreplace "etc/sshguard.conf" do |s|
```
fxcoudert: Then `inreplace etc/"sshguard.conf"`

fxcoudert (changes requested):

fxcoudert:
Also: please edit commit message to be `sshguard: new config file, logging scheme`.
And add a `revision 1` line.

Formula/sshguard.rb
jinschoi: Oh yes. I did this because by default it would look for the conf script in the Cellar directory. I changed the sysconfdir prefix to /usr/local, then forgot about this.

Formula/sshguard.rb
jinschoi: This fails if you already have an etc/sshguard.conf that has been rewritten. That etc/"path" is interesting, looks like it creates a new version named ".default" if there is something already there when used with install, but it doesn't change the reference on read for use with inreplace.

Formula/sshguard.rb
jinschoi: So unless you have a suggestion for how to deal with etc/"sshguard.conf" referring to the pre-existing version, I think making a copy in the build tree and etc.install'ing it is the way to go. I've changed it in my version to make the copy in the examples/ directory instead of creating an etc/ because there is no point to that now that I'm pointing the binary at the proper etc directory.

Formula/sshguard.rb
jinschoi: Probably not. I didn't submit this formula, I'm just modifying it to work. I'll remove those dependencies.

jinschoi:
I've updated the test to expect the proper exit code. There is really nothing more that can be done, because it needs to be run as root to do anything.

Formula/sshguard.rb
```diff
@@ -12,16 +13,22 @@ class Sshguard < Formula
...
                           "--disable-silent-rules",
                           "--prefix=#{prefix}",
-                          "--with-firewall=#{firewall}"
+                          "--sysconfdir=#{etc}"
     system "make", "install"
+    cp "examples/sshguard.conf.sample", "examples/sshguard.conf"
+    inreplace "examples/sshguard.conf" do |s|
+      s.gsub! /^#BACKEND=.*$/, "BACKEND=\"#{libexec}/sshg-fw-#{firewall}\""
+      if MacOS.version >= :sierra
+        s.gsub! %r{^#LOGREADER="/usr\/bin\/log}, "LOGREADER=\"/usr/bin/log"
```
DomT4: Shouldn't need to escape forward slashes inside `%r{}`?

Formula/sshguard.rb
```diff
@@ -70,6 +75,6 @@ def plist; <<-EOS.undent
   end
 
   test do
-    assert_match version.to_s, shell_output("#{sbin}/sshguard -v 2>&1", 78)
+    assert_match "SSHGuard #{version.to_s}", shell_output("#{sbin}/sshguard -v 2>&1", 0)
```
DomT4: `0` is assumed with `shell_output`. Only need to write it explicitly if it isn't `0`.

Formula/sshguard.rb
jinschoi: Yes, left over from when I converted that after a brew audit. Updated.

Formula/sshguard.rb
jinschoi: Fixed.

fxcoudert (approved):

Formula/sshguard.rb
```diff
@@ -70,6 +75,6 @@ def plist; <<-EOS.undent
   end
 
   test do
-    assert_match version.to_s, shell_output("#{sbin}/sshguard -v 2>&1", 78)
+    assert_match "SSHGuard #{version.to_s}", shell_output("#{sbin}/sshguard -v 2>&1")
```
fxcoudert: `W: 78: 38: Redundant use of Object#to_s in interpolation.`

Just use `"SSHGuard #{version}"` and I believe we are good to go!

fxcoudert:
Thanks @jinschoi for your contribution!

PR #11065, opened 2017-03-14
gtk+3 3.22.10

tschoonj:


fxcoudert:
Yosemite build failure reported upstream: https://bugzilla.gnome.org/show_bug.cgi?id=780019

tschoonj:
Thanks, I was going to do that myself.

Your explanation on bugzilla is incorrect though: the header is not the issue, but the use of NSWindowStyleMask is, as this was added only in macOS 10.12, so Xcode 8 is required.

tschoonj:
I will push a patch to fix this in a bit

Formula/gtk+3.rb
```diff
@@ -111,3 +113,27 @@ def post_install
...
+diff --git a/gdk/quartz/gdkwindow-quartz.c b/gdk/quartz/gdkwindow-quartz.c
+index f868f35..db2e58f 100644
+--- a/gdk/quartz/gdkwindow-quartz.c
++++ b/gdk/quartz/gdkwindow-quartz.c
+@@ -2478,7 +2478,7 @@ gdk_quartz_window_set_functions (GdkWindow    *window,
+   gboolean min, max, close;
+ 
+   g_return_if_fail (GDK_IS_WINDOW (window));
+-
++#if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_12
```
fxcoudert: Are you sure it's 10.12 minimum? It was working on El Capitan…

tschoonj: It will compile on El Capitan as long as Xcode 8 is installed. There is no guarantee that people who compile their packages from source on El Cap have actually done that.

Interestingly, Apple has removed the `setStyleMask` method from the 10.12 API, in favour of `@property styleMask`, which itself was changed in 10.12 from `NSUInteger` to `NSWindowStyleMask` (an NSUInteger enum).

tschoonj:
@fxcoudert I am using the patch from the bugzilla bugreport. Feel free to merge this one in if you agree that this is a good solution until they release a new version.

fxcoudert (approved):
Good to merge!

PR #11074, opened 2017-03-14
readline: remove no-op --enable-multibyte

fxcoudert:
`--enable-multibyte` is a no-op in readline's `configure` script (it is the default).

### DIFF
--- a/Formula/readline.rb
+++ b/Formula/readline.rb
@@ -33,7 +33,7 @@ class Readline < Formula
   EOS
 
   def install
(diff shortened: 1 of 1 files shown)


maxim-belkin:
what's up with this commit?

fxcoudert:
What's your question?

maxim-belkin:
I guess, your `remove no-op --enable-multibyte` is really `remove the default --enable-multibyte`. Got confused by `no-op` part.

fxcoudert:
Yes, the option is removed because it was unnecessary. This does not change the behaviour.

Issue #11080, opened 2017-03-14
gnupg@2.1 marked keg-only


FiloSottile:
Hi! I don't understand why gnupg@2.1 would be marked keg-only. It is a valid (actually, the latest) version of a command line tool, so it makes no sense unless linked.

This also happened to break systems on `brew update` but I think that's a `brew` bug, and is being tracked as Homebrew/brew#2340

FiloSottile:
In particular if Homebrew/brew#2340 is deemed "work as intended", then gnupg@2.1 should really not be keg-only.

ilovezfs:
By default all formulae imported from homebrew/versions will be or have been made `keg_only` unless they're known not to conflict with any of the non-versioned formulae in already in homebrew/core. In this case, gnupg@2.1 does indeed conflict with the `gnupg2` formula.

You're welcome to open a PR upgrading `gnupg2` (or `gnupg` for that matter!) to the 2.1 series.

```
==> Finishing up
/usr/local/Homebrew/Library/Homebrew/brew.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/gnupg2.rb
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/addgnupghome
Target /usr/local/bin/addgnupghome
is a symlink belonging to gnupg2. You can unlink it:
  brew unlink gnupg2

To force the link and overwrite all conflicting files:
  brew link --overwrite gnupg@2.1

To list all files that would be deleted:
  brew link --overwrite --dry-run gnupg@2.1

Possible conflicting files are:
/usr/local/bin/addgnupghome -> /usr/local/Cellar/gnupg2/2.0.30_3/bin/addgnupghome
/usr/local/bin/applygnupgdefaults -> /usr/local/Cellar/gnupg2/2.0.30_3/bin/applygnupgdefaults
/usr/local/bin/gpg-connect-agent -> /usr/local/Cellar/gnupg2/2.0.30_3/bin/gpg-connect-agent
/usr/local/bin/gpg2 -> /usr/local/Cellar/gnupg2/2.0.30_3/bin/gpg2
/usr/local/bin/gpgconf -> /usr/local/Cellar/gnupg2/2.0.30_3/bin/gpgconf
/usr/local/bin/gpgparsemail -> /usr/local/Cellar/gnupg2/2.0.30_3/bin/gpgparsemail
[...]

Issue #637, opened 2016-04-27
Gnupg 2.1


maletor:
https://github.com/Homebrew/legacy-homebrew/issues/36104 references updating to 2.1 but the issue is closed and locked claiming that updating breaks backwards compatibility. 

...I mean, ya, that's sometimes that happens when software is updated. We shouldn't be afraid to update a package manager's software versions because the newest one doesn't work with the version before it.

Homebrew should package Gnupg 2.1 and, if necessary, provide a caveat about how migrating back to 2.0 would be painful and backups should be made. There are a great many improvements in 2.1 that users will want to take advantage of.

DomT4:
It wasn't locked as a personal thing, one of the maintainers wrote a script using a GitHub API to lock a lot of old Issues because people would comment in them and we'd miss those notifications.

`brew install homebrew/versions/gnupg21` will get you what you want. It also comes shipped with a giant explanatory caveat. This has been [noted in the core formula](https://github.com/Homebrew/homebrew-core/blob/7e36b9d5885fd710eb3139de2acca506dea59301/Formula/gnupg2.rb#L1-L3). It'll likely take over the core formulae for `gnupg2` at some point but we are packaging `gnupg21` in the meantime.

PR #11083, opened 2017-03-14
gnupg 2.1.19, gnupg@2.0 (new formula), gnupg@1.4 (new formula)

ilovezfs:

-----

CC @DomT4 @JCount @MikeMcQuaid


chdiza:
Moving to 2.1 from 2.0 can potentially mess up a user's keychain (as the caveats in 2.1 say).  However this ends up working, I don't think it should auto-upgrade someone from 2.0 to 2.1.

JCount:
To provide some clarity on the keyring issue when upgrading, from https://www.gnupg.org/faq/whats-new-in-2.1.html#nosecring:

> Removal of the secret keyring
> gpg used to keep the public key pairs in two files: pubring.gpg and secring.gpg. The only difference is that secring stored in addition to the public part also the private part of the key pair. The secret keyring thus contained only the keys for which a private key is available, that is the user’s key. It required a lot of code to keep both versions of the key in sync and led to sometimes surprising inconsistencies.
> 
> The design of GnuPG-2 demands that only the gpg-agent has control over the private parts of the keys and the actual encryption engine (gpg or gpgsm) does not know about the private key but care only about session keys and keys for symmetric encryption. This has been implemented about 10 years ago for gpgsm (the S/MIME part of GnuPG). However, gpg (the OpenPGP part) used the gpg-agent only as passphrase entry and cache device but handles the private key itself.
> 
> With GnuPG 2.1 this changed and gpg now also delegates all private key operations to the gpg-agent. Thus there is no more code in the gpg binary for handling private keys. En passant this allows the long time requested “merging of secret keys” and several other advanced key management techniques.
> 
> To ease the migration to the no-secring method, gpg detects the presence of a secring.gpg and converts the keys on-the-fly to the the key store of gpg-agent (this is the private-keys-v1.d directory below the GnuPG home directory (~/.gnupg)). This is done only once and an existing secring.gpg is then not anymore touched by gpg. This allows co-existence of older GnuPG versions with GnuPG 2.1. However, any change to the private keys using the new gpg will not show up when using pre-2.1 versions of GnuPG and vice versa.
> 
> Note that the command --export-secret-keys still creates an OpenPGP compliant file with the secret keys. This is achieved by asking gpg-agent to convert a key and return it in the OpenPGP protected format. The export operation requires that the passphrase for the key is entered so that gpg-agent is able to change the protection from its internal format to the OpenPGP required format.

ilovezfs:
@chdiza as written, this will indeed upgrade `gnupg2` people to 2.1.19.

chdiza:
@Jcount I was thinking specifically about the new "keybox" format for the public keyring (described later on down in that webpage), not about the secret keyring.  The former could really mess people up, *especially* if they're not expecting their normal `brew` updating/upgrading to make non-backwards-compatible keyring format changes.

> this will indeed upgrade gnupg2 people to 2.1.19.

Well, that's bad.  If I recall correctly, this keybox thing is pretty much the whole reason why 2.1 was kept out in versions for so long, and then not made the default `gnupg`.  Nothing has changed, as far as I know.  I defer to @DomT4 on this.

ilovezfs:
>Nothing has changed, as far as I know.

2.0 will be EOL 31 Dec 2017.

DomT4:
Yes, @chdiza is correct. Whilst the move by default to `2.1.x` will not destroy _existing_ keys it will convert them to the new format and then keep them in the new format; IIRC any changes made within the new format cannot be accessed by `1.x` or `2.0.x` ever.

So for example if you have the more portable `gnupg1` installed alongside the newer `gnupg2.1.x` and you rely on the former being able to see what's going on with the latter things will break for you very suddenly.

The big sticking point for me on this was always that Homebrew does not intentionally mess with stuff in people's `$HOME` folders, and this _almost_ violates that principle because of both Homebrew's automatic update system & the `upgrade --all` functionality baked into a vanilla `brew upgrade`.

It's not impossible, to say the least, that someone runs `brew upgrade`, ignores the significant wad of output because most people don't update wildly often, and then runs some `gpg` command without thinking about it and the backwards-incompatible changes happen then. No warning at all if you miss the `upgrade` messages.

Longer-term this also makes it necessary to rethink Homebrew's _(incomplete because I'm a doofus)_ GPG verification, and Cask's _(incomplete because nobody finished it)_ GPG verification if the plan there is to allow using existing keychains so people can still `brew install` stuff that needs/wants GPG verification offline, in the same way Homebrew technically supports you `brew fetch`-ing things in advance now and then `brew install xyz` whilst you're sat on a bus in the middle of nowhere.

ilovezfs:
@DomT4 Mind reviewing the PR assuming the upgrade is a foregone conclusion ;)

DomT4:
I'll look it over. Just wanted to lay out that this wouldn't be the end of the story if the update was pushed through, assumptions made elsewhere would need to change as well.

FWIW my long-term plan here prior to departing was to get to the point where `gnupg2.1` didn't need patching or tests disabling to actually run a clean testsuite on macOS, then announce it somewhere a month or two in advance in both `caveat` form and a more explicit text document somewhere to reference, and then push it through after that notice period. That plan still has the issues of adjusting long-term assumptions, but my aim was less sudden change rather than less change 😄.

ilovezfs:
It would be nice if I could wave a magic wand and undo the years of anti-2.1 propaganda we've been putting out there.

DomT4:
Well, to be fair, there were some pretty gnarly issues towards the start of this branch, not least around the agent on upgrade. It was a legitimate decision to stop casual users going _"Ooh new shiny thing, let's try it out!"_ without understanding the changes around that.

JCount:
@chdiza I meant to post the part about the public keybox format but lost track of doing it. However, I think that its impacts on the end user are generally less in the event of an upgrade. This is because if there is an existing pubring.gpg file and no prior existing keybox, as would most likely be the case if you had been using the gnupg2 formula, as there is no automatic conversion of pubring.gpg. GnuPG would continue to use the the pubring.gpg file and remain backwards compatible. It would only be an issue if prior use of gpgsm had generated a keybox. In this case, if the user were unaware it could potentially be a serious issue.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
> However, I think that its impacts on the end user are generally less in the event of an upgrade. This is because if there is an existing pubring.gpg file and no prior existing keybox, as would most likely be the case if you had been using the gnupg2 formula, as there is no automatic conversion of pubring.gpg. GnuPG would continue to use the the pubring.gpg file and remain backwards compatible.

Aye, this is correct, with some caveats. The biggest caveat is that `gpg1` cannot automatically see new secret keys created by `gpg2`, so once you change your secret key _(which theoretically we should all be doing a lot more than anyone actually does)_ `gpg1` won't know about it unless you do the export/import dance.

This is only true for formats it actually understands though, if you use ECC _(which IMO is half the point of the new branch)_ and then try to use that with `gpg1` at all it does some weird weird things:

```
sec      0C/0xFFFFFFFFFFFFFFFF 2017-03-14
      Key fingerprint = 0000 0000 0000 0000 0000  0000 0000 0000 0000 0000
uid                            Shaun Sheep <shaunthesheep@gmail.com>
ssb      0e/0xFFFFFFFFFFFFFFFF 2017-03-14
```

So whilst it's true that the pain for existing users isn't immediately in your face, the longer you actually use the `2.1.x` branch the more it becomes _practically difficult_ to go back to `1.x` or `2.0.x`.

DomT4:
> @DomT4 PR refreshed.

You fixed some things I was about to point out. That's very efficient 😄.

Formula/gnupg.rb
```diff
@@ -13,70 +13,92 @@ class Gnupg < Formula
...
+    args << "--with-readline=#{Formula["readline"].opt_prefix}" if build.with? "readline"
 
     system "./configure", *args
-    system "make"
-    system "make", "check"
 
-    # we need to create these directories because the install target has the
-    # dependency order wrong
-    [bin, libexec/"gnupg"].each(&:mkpath)
     system "make", "install"
```
DomT4: Can you copy the logic from the existing `gnupg2` to install vanilla symlinks to `gpg` and `gpgv`.

Formula/gnupg.rb
```diff
@@ -13,70 +13,92 @@ class Gnupg < Formula
...
+    of your `~/.gnupg` directory.
 
-        PATH="#{opt_libexec}/gpgbin:$PATH"
+    For full details of the changes, please visit:
+      https://www.gnupg.org/faq/whats-new-in-2.1.html
 
-    Note that doing so may interfere with GPG-using formulae installed via
-    Homebrew.
+    If you are upgrading to gnupg 2.1.x from gnupg 2.0.x you should execute:
+      `killall gpg-agent && gpg-agent --daemon`
```
DomT4: Solid chance Mike makes a disapproving face at me here, but if you're going to mass-update people it might be polite to do this automatically in `postinstall` given how much we both know people bother to read caveats.

Formula/gnupg.rb
```diff
@@ -13,70 +13,92 @@ class Gnupg < Formula
...
-      Name-Real: Testing
-      Name-Email: testing@foo.bar
-      Expire-Date: 1d
-      %commit
-    EOS
-    system bin/"gpg1", "--batch", "--gen-key", "batchgpg"
-    (testpath/"test.txt").write "Hello World!"
-    system bin/"gpg1", "--armor", "--sign", "test.txt"
-    system bin/"gpg1", "--verify", "test.txt.asc"
+    system bin/"gpgconf"
```
DomT4: It should be safe to keep the stronger existing test from the original formula here given Homebrew captures `$HOME` for testing.

DomT4:
Filippo is correct on this point FWIW: https://github.com/Homebrew/homebrew-core/issues/11080#issuecomment-286496206. I don't have an especially elaborately-modified GPG configuration, but I still get yelled at every time I run a command under the new `gpg`.

Maybe that's not a bad thing, means people might actually read up and adjust their configurations, but it will require user intervention and user knowledge of what needs to change and why, which could be interesting.

ilovezfs:
It may require some hand holding but I'm sure upstream will help pick up some of the support burden.

JCount:
@DomT4 Thank you for your constructive explanation of how I was underestimating the potential for issues that the end user faces. I appreciate candor, especially when it comes to things I may have misinterpreted or misunderstood. 😄

maletor:
This looks right to me, in general.

zmwangx:
Since this doesn't seem to have been mentioned in the thread, I'll point it out: [GPG Suite](https://gpgtools.org/), a well established GnuPG distribution on macOS that provides Mail.app integration and so on, currently uses GPG 2.0.30. I'm afraid that upgrading to 2.1.x (which automatically upgrades the keyring format in place) would destroy interoperability with GPG Suite.

I think a caveat should be added to warn users to **uninstall brewed gnupg 2.1.x** if they use GPG Suite.

ilovezfs:
>I think a caveat should be added to warn users to uninstall brewed gnupg 2.1.x if they use GPG Suite.

That's not a particularly constructive suggestion.

zmwangx:
> That's not a particularly constructive suggestion.

What would you suggest? Data loss (okay, maybe not loss, but close) is a serious issue.

ilovezfs:
I would suggest you contact GPG upstream and recommend that they share their plan for 2.0 EOL with everyone.

zmwangx:
All I know about "2.0 EOL" is from https://www.gnupg.org/download/index.html:

> GnuPG modern (2.1) is the brand new version with enhanced features like support for Elliptic Curve Cryptography. It will eventually replace the current stable (2.0).

"Eventually" can mean anything, hence nothing. Have you heard of a more concrete plan? Or are you suggesting that I torture a plan out of the GPG team first?

Nevertheless, I'm not sure how that suggestion conflicts with my practical suggestion of not having both brewed gpg 2.1.x (assuming it will land soon) and GPG Suite installed at the same time. You simply can't use both without messing up, unless you have separate `GNUPGHOME`s, which is unlikely.

Also note that GPG Suite can be installed via `brew-cask` (`brew cask install gpgtools`) so you can't tie up your hands and say "none of our business".

ilovezfs:
https://www.gnupg.org/

"2.0.30 is the stable version from an often used branch. This branch will reach end-of-life on 2017-12-31."

zmwangx:
Okay, created support ticket https://gpgtools.tenderapp.com/discussions/problems/51244. (Code is at https://github.com/GPGTools/MacGPG2, but issue tracker isn't enabled.)

In the mean time, "my practical suggestion of not having both brewed gpg 2.1.x (assuming it will land soon) and GPG Suite installed at the same time" still stands.

ilovezfs:
>In the mean time, "my practical suggestion of not having both brewed gpg 2.1.x (assuming it will land soon) and GPG Suite installed at the same time" still stands.

How about after every installation of software that has a data store of some kind we print
```
Caveat: if you choose to use incompatible software that tries to work with the same database this formula uses you may want to rethink your choices.
```

zmwangx:
> How about after every installation of software that has a data store of some kind 

Care to give an example of a pair of notable, non-version packages in Homebrew/Caskroom's official repositories that require conflicting formats of the exact same data store?

DomT4:
>  Or are you suggesting that I torture a plan out of the GPG team first?

Citation on legality needed 😉.

On the `GPG Suite` issue and that alone, I designed the `gpg` requirement and such to work with it, but I don't think Homebrew has to be bound by that choice forever.

I'm more concerned about the factors Homebrew can and does control. It works both ways really; `GPG Suite` could update to `2.1.x` tomorrow and bork a bunch of formulae presuming they can access a set of current, up-to-date secret keys, bearing in mind you wouldn't get a Homebrew `gpg` shipped until you `brew reinstall` or `brew install`'d one of the `gpg` using formulae in that situation.

It's a valid concern but at some point this change will have to go through and some sacrifices will need to be made to get there. That's not an endorsement of this right now, and it's also not not an endorsement of this right now, but I think whichever way this goes there will be some pain and it's easier addressing pain that Homebrew can mitigate _to an extent_.

Maybe the new `gnupg` needs a conflict with the `GPG Suite` similar to `osxfuse` when there was a formula for that still around or something, that would at least prevent the problem, even if not in a particularly gentle way.

zmwangx:
> It's a valid concern but at some point this change will have to go through and some sacrifices will need to be made to get there.

I never tried to stop this. All I'm proposing is an additional sentence in the caveats (which is feature-length already).

ilovezfs:
>Or are you suggesting that I torture a plan out of the GPG team first?

That sounds like a reasonable action plan to me.

DomT4:
> All I'm proposing is an additional sentence in the caveats (which is feature-length already).

This is my concern, partly. All three of us here that are maintainers or used to be maintainers know how thoroughly caveats get read 😕

zmwangx:
By the way,

> GPG Suite could update to 2.1.x tomorrow and bork a bunch of formulae

Having used GPG Suite for a few years, I'd say (with >95% confidence) it won't be updated to 2.1.x until the very last minute. Not really blaming them, but they've always been playing catch-up, apparently due to lack of resources.

ilovezfs:
Surely someone has it working with 2.1.x already …

zmwangx:
> Surely someone has it working with 2.1.x already …

and compatible with all other components of the GPG Suite? Possible, but I won't be so sure.

ilovezfs:
In any case, we're not doing anyone any favors by making it easier for GPG Suite to delay the inevitable passed the EOL date.

ilovezfs:
@zmwangx I expect your ticket will be closed with a "we're stabilizing it on Sierra so we'll get to it when we get to it" message and will be linked to https://gpgtools.lighthouseapp.com/projects/66001/tickets/142-update-to-gnupg-21.

Formula/gnupg.rb
ilovezfs: done

ilovezfs:
@DomT4 PR refreshed.

ilovezfs:
@BrewTestBot test this please

teythoon:
Hello, GnuPG upstream developer here.  We feel your pain with the migration.  Note that major Linux distributions have made the switch, and it worked out allright for them.

The most intrusive change for users is that GnuPG 2.1 uses a different way to store secret keys.  When a user uses 2.1 for the first time, the old secring will be migrated.  From that point on, 1.4 or 2.0 will no longer be able to use the private keys, though operations requiring only the public keys will still work (e.g. signature verification).

[Having said that, I believe GnuPG 2.0 (maybe even 1.4) can use the agent from 2.1, and then possibly use the secret keys stored in the new store.  But that is certainly not something we recommend or test.]

I'm sorry, but I cannot give you a definitive release schedule, because we (the GnuPG project) could not agree on a concrete schedule yet.  I have tried to bring up the issue repeatedly, but no luck yet.  Please don't torture us ;)

What I can say is that we are going to release it soonish, definitively this year because our old stable version EOLs as you pointed out.  We are very close to 2.2 though, and are mostly stabilizing it, adding small features here and there, fixing regressions.  Input from a larger user set (Debian testing, currently in freeze) that switched to 2.1 was most insightful.

MikeMcQuaid:
> IIRC certain changes made within the new format cannot be accessed by 1.x or 2.0.x ever.

> Care to give an example of a pair of notable, non-version packages in Homebrew/Caskroom's official repositories that require conflicting formats of the exact same data store?

MySQL.

zmwangx:
> MySQL.

That's not a **pair** of notable, **non-version** packages.

MikeMcQuaid:
> That's not a pair of notable, non-version packages.

Yup, don't care. MySQL upgrades present (automatically, no less) exactly the same issues described here and we've not cared for years. Similarly, check out moving from `mysql` to `mysql@5.6` on a database with data on it.

Lots of software decides to break backwards compatibility. It's only our problem if we decide to make it so. GnuPG 2 has been out for a long time at this point and I agree with @ilovezfs that we need to be moving forward on this.

maletor:
Agree with @MikeMcQuaid.

In fact, [I've been advocating this](https://github.com/Homebrew/homebrew-core/issues/637#issue-151271596) since April of last year.

Formula/gpgme.rb
```diff
@@ -12,7 +13,7 @@ class Gpgme < Formula
     sha256 "4a4c3e90175b02ded974a8a385b6d9799bcc2873c5a1087c558f350eb103934d" => :yosemite
   end
 
-  depends_on "gnupg2"
+  depends_on "gnupg"
```
DomT4: This seems incorrect based on the `inreplace` below? One or the other needs to change.

ilovezfs: I imagine it's not even needed now?

DomT4: The `inreplace`? Unsure. I added it when we moved from `1.x` to `2.0.x`, partly to ensure it was actually using `2.0.x` and partly to ensure it was always linked to a Homebrew `gpg` rather than interacting with some other one in the environment.

But yes, if the latter isn't a concern and the new `gnupg` 2.1.x ships "vanilla" `gpg` and `gpgsm` executables it should work okay without the `inreplace`.

Formula/gnupg@2.0.rb
```diff
@@ -79,7 +71,17 @@ def post_install
   end
 
   test do
-    Gpg.create_test_key(testpath)
```
DomT4: Does this not work any more?

Formula/gpgme.rb
ilovezfs: Strangely it seems to puke
```
<"/usr/local/opt/gnupg/bin/gpg"> expected but was
<"@GPG@">.
```
why even run configure? lol

ilovezfs: Interesting
```
        --get-gpg)
            # Deprecated
            output="$output @GPG@"
            ;;
```

Formula/gnupg@2.0.rb
ilovezfs: It may work again once @JCount's PR goes through https://github.com/Homebrew/brew/pull/2343

Formula/gnupg.rb
```diff
@@ -13,53 +13,83 @@ class Gnupg < Formula
...
-    needing GnuPG 1.x we recommend:
-      brew install gnupg2
+  def post_install
+    (var/"run").mkpath
+    Utils.popen_read("killall", "gpg-agent")
+  end
 
-    If you really need to use these tools without the "1" suffix you can
-    add a "gpgbin" directory to your PATH from your #{shell_profile} like:
+  def caveats; <<-EOS.undent
```
DomT4: This needs rewriting a bit. It has for a while, but I forgot to do it, and then I ran away without mentioning it 😅.

Maybe something like:
```ruby
def caveats; <<-EOS.undent
  Once you run this version of gpg you may find it difficult to return to using
  a prior 1.4.x or 2.0.x. Most notably the prior versions will not automatically
  know about new secret keys created or imported by this version. We recommend
  creating a backup of your `~/.gnupg` prior to first use.

  For full details on each change & how it could impact you please see:
    https://www.gnupg.org/faq/whats-new-in-2.1.html
  EOS
end
```

Formula/gnupg.rb
```diff
@@ -13,53 +13,83 @@ class Gnupg < Formula
...
-  def caveats; <<-EOS.undent
-    This formula does not install either `gpg` or `gpgv` executables into
-    the PATH.
-
-    If you simply require `gpg` and `gpgv` executables without explicitly
-    needing GnuPG 1.x we recommend:
-      brew install gnupg2
+  def post_install
+    (var/"run").mkpath
+    Utils.popen_read("killall", "gpg-agent")
```
DomT4: Maybe `quiet_system` instead? If `gpg-agent` isn't running `killall` throws an annoying message.

ilovezfs: The issue is the exit code ;)

Formula/gnupg@2.0.rb
ilovezfs: Scratch that. It doesn't work when keg_only, which would require other changes.

Formula/gnupg.rb
DomT4: Welp. Forgot we can't manipulate `quiet_system` to accept a "bad" exit code. Maybe something like:
```ruby
quiet_system "killall gpg-agent; true"
```
Heh 😅. Nasty.

ilovezfs: Note that as written, this prints nothing. To get the ugly message I'd need to do
```
puts Utils.popen_read("killall gpg-agent 2>&1")
```

DomT4: Oh, am I forgetting that `killall` prints only to `stderr` or am I forgetting how `poepn_read` works? 😆

Formula/gnupg.rb
```diff
@@ -72,11 +102,16 @@ def caveats; <<-EOS.undent
...
       Expire-Date: 1d
+      %no-protection
       %commit
     EOS
-    system bin/"gpg1", "--batch", "--gen-key", "batchgpg"
-    (testpath/"test.txt").write "Hello World!"
-    system bin/"gpg1", "--armor", "--sign", "test.txt"
-    system bin/"gpg1", "--verify", "test.txt.asc"
+    begin
+      system bin/"gpg2", "--batch", "--gen-key", "batchgpg"
```
DomT4: Ideally this would use `gpg`, even though that's just a symlink. Trying to get people to expect that a non-keg-only `gpg` from Homebrew will always be `gpg2`.

Formula/gnupg@1.4.rb
```diff
@@ -0,0 +1,48 @@
...
+    args << "--with-libusb=no" if build.without? "libusb-compat"
+
+    system "./configure", *args
+    system "make"
+    system "make", "check"
+
+    # we need to create these directories because the install target has the
+    # dependency order wrong
+    [bin, libexec/"gnupg"].each(&:mkpath)
+    system "make", "install"
```
DomT4: Likewise I think it's worth keeping the `gpg1` symlinks available for this formula. That way you can have `gnupg` (2.1.x) in your `$PATH` and also have a `keg_only` `gnupg@1.4` in your `$PATH` and still be able to access this formula without a `gpg` execution always being intercepted by the 2.1.x version.

Formula/gnupg.rb
ilovezfs: It returns the string, which isn't going to make it to the shell unless we use `puts` or `ohai` or whatever. But `quiet_system` may work too. Not entirely sure why it seems to disregard the exit code.

Formula/gnupg@2.0.rb
DomT4: Yes, the existing system was designed around everything being publicly accessible, for better or worse.

Formula/gpgme.rb
DomT4: Colour me unsurprised 😕. Probably want to rewrite the `inreplace` and the `test`, I guess.

Formula/gnupg@1.4.rb
ilovezfs: I mean … at that point we might as well get rid of this keg_only thing, no?

DomT4: I'm not actually against that idea, where the system remains semi-similar to how it is now, except people are shunted across to `gnupg` (2.1.x) which becomes what `gnupg2` currently is in this system.

That would avoid significant rewrites of the `gpg` stuff in `brew`, but I'm mostly wary that 1.4.x remains quite popular because of how portable it is. On a modern system you require zero dependencies at all.

DomT4:
I think that's everything I can find to discuss, which you're probably thankful for 🙈. I think this is fairly close to working as well as it can.

Theoretically everything in core currently using `gpg` might need testing afresh given we're forcing a new major `gpg` version onto everyone (which I think is the "right thing to do" ™️ under this model), but I'll leave that call to you.

Formula/gnupg@1.4.rb
ilovezfs: OK let's try keeping the OG gnupg formula as it is, just renamed.

Formula/gpgme.rb
ilovezfs: may just punt?
```
assert_match version.to_s, shell_output("#{bin}/gpgme-tool --lib-version")
```

Formula/gnupg.rb
ilovezfs: Sure.

ilovezfs:
@DomT4 PR refreshed.

DomT4:
Will look this over again in an hour or so. My headache needs a break from Ruby 😄.

Formula/gnupg.rb
```diff
@@ -13,53 +13,85 @@ class Gnupg < Formula
...
+    bin.install_symlink "gpgv2" => "gpgv"
+    man1.install_symlink "gpg2.1" => "gpg.1"
+    man1.install_symlink "gpgv2.1" => "gpgv.1"
 
-    # Although gpg2 support should be pretty universal these days
-    # keep vanilla `gpg` executables available, at least for now.
-    %w[gpg-zip1 gpg1 gpgsplit1 gpgv1].each do |cmd|
-      (libexec/"gpgbin").install_symlink bin/cmd => cmd.to_s.sub(/1/, "")
-    end
+    bin.install "tools/gpgsplit" => "gpgsplit2" if build.with? "gpgsplit"
```
DomT4: What's the argument against installing this by default/keeping it unversioned/installing an unversioned symlink?

DomT4:
One last thing questioned, but otherwise I think this is 💯 in terms of the actual code changes. I'm happy at least. It became rather a long thread so some last self-quotes, and I'm aware self-quoting in the same thread makes me the worst human 😄.

> Theoretically everything in core currently using gpg might need testing afresh given we're forcing a new major gpg version onto everyone (which I think is the "right thing to do" ™️ under this model), but I'll leave that call to you.

> Ideally here once everything's ready, you'll want to merge https://github.com/Homebrew/brew/pull/2343, cut a new tag, and then merge Homebrew/homebrew-core#11083 immediately after. That should keep things fairly smooth for everyone.

> Maybe that's not a bad thing, means people might actually read up and adjust their configurations, but it will require user intervention and user knowledge of what needs to change and why, which could be interesting.

I think those are the big three issues left to work out what to do about/how to guide people if they ask questions/etc.

Formula/gnupg.rb
ilovezfs: Only that upstream doesn't include it by default and the man page was removed.

DomT4: Cool. Makes sense as-is then.

ilovezfs:
@DomT4 I think `gnupg` may also need
```
  link_overwrite "bin/dirmngr"
  link_overwrite "bin/dirmngr-client"
  link_overwrite "bin/gpg-agent"
  link_overwrite "share/man/man1/dirmngr-client.1"
```

ilovezfs:
hmm that didn't seem to work … :(

ilovezfs:
The problem is that `gnupg` 2.1.x doesn't depend on `gpg-agent`, so even though it's being made `keg_only` to avoid the conflict, the upgrade of `gnupg` goes first when `brew upgrade` runs, so `gnupg` cannot successfully link since at that point gpg-agent is not `keg_only` yet.

I could add an explicit `conflicts_with` but that is really annoying to have to force everyone to manually unlink `gpg-agent` first when we're making `gpg-agent` keg_only anyway.

ilovezfs:
I guess I could artificially depend on `gpg-agent`, which would get the order of the upgrades correct, although it's absurd.

DomT4:
> hmm that didn't seem to work … :(

Because `update` handles migration rather than `upgrade` I presume, so the migration always happens first? Or am I misunderstanding your point there?

Ideally `brew` should prioritise anything becoming `keg_only` above everything else on `upgrade`, rather than simply plowing through a list by dependency order and then alphabetical order.

ilovezfs:
@DomT4 https://github.com/Homebrew/brew/pull/2358

DomT4:
@ilovezfs Nice 👍. I forgot for a moment `gnupg` would be sorted alphabetically before `gpg-agent`; I'm not 100% well so bear with any glaring stupidity 😄.

ilovezfs:
@DomT4 I considered changing its name to `agent-du-gpg`

ilovezfs:
@DomT4 Also: https://github.com/Homebrew/brew/pull/2359.

DomT4:
I vaguely recall someone on the team _(can't remember who though so maybe not)_ had intentions to leave migrations permanently available via the old name, but I'll leave that down to y'all to resolve 🙈.

ilovezfs:
Technically that would still be the case here, since gnupg@1.4 and gnupg@2.0 are new formulae, not migrations from anywhere.

DomT4:
@ilovezfs Just a heads-up that I fly out Wednesday and won't be taking my laptop with me, so I'll likely be off-the-grid or confined to short emailed replies till early April. This may be a relief for you :trollface:.

Formula/dirmngr.rb
```diff
@@ -13,6 +13,8 @@ class Dirmngr < Formula
     sha256 "47fe29be8ca19eeb4d4a3e3434cd35ef7b13e1c1a9e8696f5ebd4434dc8cc062" => :mavericks
   end
 
+  keg_only "GPG2.1.x ships an internal dirmngr which it must use."
```
JCount: "GPG 2.1.x ships an internal dirmngr which it must use."

ilovezfs: Copy-pasted out of the old gnupg21 formula:
```
  conflicts_with "gnupg2",
        :because => "GPG2.1.x is incompatible with the 2.0.x branch."
  conflicts_with "gpg-agent",
        :because => "GPG2.1.x ships an internal gpg-agent which it must use."
  conflicts_with "dirmngr",
        :because => "GPG2.1.x ships an internal dirmngr which it it must use."
  conflicts_with "fwknop",
        :because => "fwknop expects to use a `gpgme` with Homebrew/Homebrew's gnupg2."
  conflicts_with "gpgme",
        :because => "gpgme currently requires 1.x.x or 2.0.x."
```
CC @DomT4 for orthographical wisdom.

DomT4:
JCount is probably right here in terms of most common expected usage.

ilovezfs:
The [new tag](https://github.com/Homebrew/brew/releases/tag/1.1.12) has shipped, and we're back to :green_apple: here.

MikeMcQuaid (approved):

ilovezfs:
Shipped. The die is cast!

Thanks to everyone who helped here, particularly @JCount, @DomT4, and @MikeMcQuaid.

PR #11086, opened 2017-03-14
Notmuch 0.24

benjaminfrank:

-----

### DIFF
--- a/Formula/notmuch.rb
+++ b/Formula/notmuch.rb
@@ -1,8 +1,8 @@
 class Notmuch < Formula
   desc "Thread-based email index, search, and tagging"
   homepage "https://notmuchmail.org"
-  url "https://notmuchmail.org/releases/notmuch-0.23.7.tar.gz"
-  sha256 "f11bb10d71945f6c3f16d23117afc70810aa485878e66bb4bf43cc3f08038913"
+  url "https://notmuchmail.org/releases/notmuch-0.24.tar.gz"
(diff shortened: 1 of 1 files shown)


benjaminfrank:
Sorry, didn't know that deleting a branch automatically closes an pull request

Formula/notmuch.rb
```diff
@@ -1,8 +1,8 @@
 class Notmuch < Formula
   desc "Thread-based email index, search, and tagging"
   homepage "https://notmuchmail.org"
-  url "https://notmuchmail.org/releases/notmuch-0.23.7.tar.gz"
-  sha256 "f11bb10d71945f6c3f16d23117afc70810aa485878e66bb4bf43cc3f08038913"
+  url "https://notmuchmail.org/releases/notmuch-0.24.tar.gz"
```
fxcoudert: Shouldn't the patch be removed?

dunn: Has it been merged? It hadn't the last time I checked.

fxcoudert: If upstream doesn't accept the patch, we shouldn't be carrying it anyway, as I understand our policy.

benjaminfrank:
The patch isn't merged yet.
I think they're awaiting a reply from the person who did the python bindings but either he went missing or forgot about it.

fxcoudert:
Thanks for the pull request. I've merged it, but in the long term we would like upstream to include the fix or we should drop the python bindings if they have stopped working and upstream doesn't care.

PR #11150, opened 2017-03-16
zbar: add HEAD option

mjbshaw:

-----

Adds the option to build from HEAD for zbar.

### DIFF
--- a/Formula/zbar.rb
+++ b/Formula/zbar.rb
@@ -1,15 +1,33 @@
 class Zbar < Formula
   desc "Suite of barcodes-reading tools"
   homepage "https://zbar.sourceforge.io"
-  url "https://downloads.sourceforge.net/project/zbar/zbar/0.10/zbar-0.10.tar.bz2"
-  sha256 "234efb39dbbe5cef4189cc76f37afbe3cfcfb45ae52493bfe8e191318bdbadc6"
   revision 4
+
+  stable do
+    url "https://downloads.sourceforge.net/project/zbar/zbar/0.10/zbar-0.10.tar.bz2"
+    sha256 "234efb39dbbe5cef4189cc76f37afbe3cfcfb45ae52493bfe8e191318bdbadc6"
+
+    # Fix JPEG handling using patch from
+    # https://sourceforge.net/p/zbar/discussion/664596/thread/58b8d79b#8f67
+    # already applied upstream but not present in the 0.10 release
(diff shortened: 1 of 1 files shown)


Formula/zbar.rb
```diff
@@ -21,12 +39,13 @@ class Zbar < Formula
...
   depends_on "libtool" => :run
 
-  # Fix JPEG handling using patch from
-  # https://sourceforge.net/p/zbar/discussion/664596/thread/58b8d79b#8f67
-  # already applied upstream but not present in the 0.10 release
-  patch :DATA
-
   def install
+    if build.head?
+      inreplace "configure.ac", "-Werror", ""
```
fxcoudert: Why do we need that. Has a bug be filed upstream?

Formula/zbar.rb
```diff
@@ -21,12 +39,13 @@ class Zbar < Formula
...
+      gettext = Formula["gettext"]
+      system "autoreconf", "-fvi", "-I", "#{gettext.opt_share}/aclocal"
```
fxcoudert: Why do we need gettext and why only at build-time?

mjbshaw: [gettext provides the `AM_ICONV` macro](https://www.gnu.org/software/gettext/manual/html_node/AM_005fICONV.html), which is used in the configure.ac.

Formula/zbar.rb
mjbshaw: The upstream project is rather defunct and hasn't been touched since 2012. [This issue has been brought up on the mailing list](https://sourceforge.net/p/zbar/discussion/664596/thread/f4b52988/) back in 2012, so they should be aware of it.

The reason we need it is because without it, `autoreconf` ends up failing because:
```
automake: warnings are treated as errors
gtk/Makefile.am.inc:13: warning: '%'-style pattern rules are a GNU make extension
Makefile.am:31:   'gtk/Makefile.am.inc' included from here
gtk/Makefile.am.inc:17: warning: '%'-style pattern rules are a GNU make extension
Makefile.am:31:   'gtk/Makefile.am.inc' included from here
pygtk/Makefile.am.inc:20: warning: '%'-style pattern rules are a GNU make extension
Makefile.am:34:   'pygtk/Makefile.am.inc' included from here
...etc...
```

fxcoudert:
Thanks @mjbshaw for your contribution!

PR #11152, opened 2017-03-16
Add zxing-cpp formula

mjbshaw:

-----

Adds the ZXing C++ port to Homebrew. The C++ port was deprecated and removed from the main ZXing project after v2.3.0, and its development has moved to a new repo ([see details here](https://github.com/zxing/zxing#zxing-based-third-party-open-source-projects)).


Formula/zxing-cpp.rb
```diff
@@ -0,0 +1,38 @@
...
+  def install
+    if build.stable?
+      cd "cpp"
+    end
+
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make"
+
+      if build.stable?
```
fxcoudert: Isn't there a `make install` for stable?

Formula/zxing-cpp.rb
```diff
@@ -0,0 +1,38 @@
...
+        lib.install "libzxing.a"
```
fxcoudert: Can we build a shared library instead of static?

Formula/zxing-cpp.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      if build.head?
+        system "make", "install"
+      end
+    end
+  end
+
+  test do
+    system bin/"zxing"
```
fxcoudert: Please include a test that actually does something. Consider installing a test image for that.

fxcoudert (changes requested):

mjbshaw (commented):
I added a proper test.

Formula/zxing-cpp.rb
mjbshaw: Done.

Formula/zxing-cpp.rb
kriansa: @mjbshaw I believe you made a typo on **install** (not _intall_)

mjbshaw: Oops, that command had a typo. But it really doesn't have `make install`:

```
$ make install
make: *** No rule to make target `install'.  Stop.
$ grep install Makefile
# Prepare targets for installation.
preinstall: all
	$(MAKE) -f CMakeFiles/Makefile2 preinstall
.PHONY : preinstall
# Prepare targets for installation.
preinstall/fast:
	$(MAKE) -f CMakeFiles/Makefile2 preinstall
.PHONY : preinstall/fast
```

dunn: Could you ask the developers if they'll add one?

mjbshaw: I don't think that would work, given that [the C++ port was deleted in 2013](https://github.com/zxing/zxing/commit/4856e86a46a5836d7ea793a78bea5637c85fe7c6#diff-ac731f2fedd74089d1149d83620eeb92) from the main repo. [zxing-cpp](https://github.com/glassechidna/zxing-cpp) forked that code, and is the unofficial-official active C++ zxing repo (but with no releases, currently).

I know it's a sorry state of affairs, but it's one that is common to barcode/QR decoding libraries, and unfortunately what C++ developers have to work with for the time being.

fxcoudert:
Thanks @mjbshaw for your contribution to Homebrew!

PR #11177, opened 2017-03-16
Install all Bash completion files for docker-machine-completion

JoeNyland:
Just as it's done in the main [docker-machine formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/docker-machine.rb#L25)

### DIFF
--- a/Formula/docker-machine-completion.rb
+++ b/Formula/docker-machine-completion.rb
@@ -3,6 +3,7 @@ class DockerMachineCompletion < Formula
   homepage "https://docs.docker.com/machine/completion/"
   url "https://github.com/docker/machine/archive/v0.10.0.tar.gz"
   sha256 "ed64698723436e68341a1ebd8486006293eac306bfa8628e924b38bfcb5b039d"
+  revision 1
   head "https://github.com/docker/machine.git"
 
   bottle :unneeded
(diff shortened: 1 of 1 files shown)


Formula/docker-machine-completion.rb
```diff
@@ -11,7 +11,7 @@ class DockerMachineCompletion < Formula
     :because => "docker-machine already includes completion scripts"
 
   def install
-    bash_completion.install "contrib/completion/bash/docker-machine.bash"
+    bash_completion.install Dir["contrib/completion/bash/*.bash"]
```
fxcoudert: Please add a `revision 1` line before `head`.

fxcoudert (changes requested):
Why do we have a specific formula for completion if the full formula exists and contains completion?

JoeNyland:
@fxcoudert Since homebrew/completions was deprecated and merged into homebrew/core [here][ref].

Having a separate formula for just the completion allows users to install just the completion without the rest of Docker Machine, for example you're using Docker Machine that's installed from [Docker for Mac][docker-for-mac].

[ref]: c474a0d86f290fd793cfe6eb359cf811eb7899b6
[docker-for-mac]: https://docs.docker.com/docker-for-mac/

fxcoudert (approved):

PR #11204, opened 2017-03-17
llvm@3.9 3.9.1 (new formula)

ilovezfs:

### DIFF
--- a/Formula/llvm@3.9.rb
+++ b/Formula/llvm@3.9.rb
@@ -1,0 +1,380 @@
+class CodesignRequirement < Requirement
+  include FileUtils
+  fatal true
+
+  satisfy(:build_env => false) do
+    mktemp do
+      cp "/usr/bin/false", "llvm_check"
+      quiet_system "/usr/bin/codesign", "-f", "-s", "lldb_codesign", "--dryrun", "llvm_check"
+    end
+  end
+
+  def message
+    <<-EOS.undent
+      lldb_codesign identity must be available to build with LLDB.
+      See: https://llvm.org/svn/llvm-project/lldb/trunk/docs/code-signing.txt
+    EOS
+  end
+end
+
+class LlvmAT39 < Formula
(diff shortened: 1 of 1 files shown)


ilovezfs:
@tmcdonell Let me know if you'd prefer I re-open https://github.com/Homebrew/homebrew-core/pull/9414.

ilovezfs:
CC @@SeanTAllen

ilovezfs:
CC @SeanTAllen

tmcdonell:
I think it's worthwhile to try the keg_only plus `llvm-config-3.9` solution, since you should be able to automate that (and llvm drops a release every 6 months or so, which is a lot of work for you guys). Might as well see how it goes, and reopen #9414 if it doesn't work for people.

ilovezfs:
I'm not sure if it's actually possible to install anything in /usr/local/bin when it's marked keg_only as I think that would be a sandbox violation, but most build systems should allow it to be specified via an environment variable (e.g. `LLVM_CONFIG` for `ponyc`) or a configure option (e.g. `--llvm-root` for `rust`).

ilovezfs:
Actually, it looks like we have HOMEBREW/"bin" white-listed in `postinstall` so I can do that.

ilovezfs:
It's also important to note, I think, that the main `llvm` formula is itself `keg_only :provided_by_osx` so having the versioned ones be `keg_only` too really shouldn't be that big of a deal.

ilovezfs:
In the interest of getting this shipped, I'm going to punt on linking llvm-config-3.9 since it technically violates the idea of keg_only-ness but I'm open to reconsidering in a future PR, especially if we do the same for the main llvm formula as well.

yiakwy:
This formula has problems to support libcxxabi, because the ruby formula does not contain the libcxxabi sources. Could you show me the minimum codes to support that. Here is a example from homebrew legacy but I don't know how to generate the hashcode. Look forward to your reply.

PR #11209, opened 2017-03-17
pyenv: fix missing caveats

JasonQSY:

-----

As shown in <https://github.com/pyenv/pyenv#homebrew-on-mac-os-x>, we need add `eval "$(pyenv init -)"` to our profile manually after the installation. The original formula does not provide such a hint, which makes me confusing when installing it. What I do is just adding caveats about it.

### DIFF
--- a/Formula/pyenv.rb
+++ b/Formula/pyenv.rb
@@ -21,6 +21,13 @@ class Pyenv < Formula
     end
   end
 
+  def caveats; <<-EOS.undent
+    After installation, you'll need to add
+      eval "$(pyenv init -)"
+    to your profile (like ~/.bashrc, ~/.zshrc) manually.
+    EOS
+  end
+
   test do
     shell_output("eval \"$(#{bin}/pyenv init -)\" && pyenv versions")
   end


DomT4:
This was intentionally removed in https://github.com/Homebrew/homebrew-core/commit/d82ef4e9a5105c9c5da086025c57f89929811763 FWIW.

ilovezfs:
lol

DomT4:
🤷‍♂️

JasonQSY:
Sorry about inconvience since I do not read the history carefully. But it seems adding such an 'export xxx' is general in homebrew and convenience for users to install formula without reading the manual. Why is it non-hb-specific? :confused:

MikeMcQuaid:
> Why is it non-hb-specific?

because everyone installing `pyenv` will have to do this, not just Homebrew users.

JasonQSY:
OK.

joshfriend:
> everyone installing pyenv will have to do this, not just Homebrew users.

That may be true, but how will people know that there are additional setup steps that must be taken? They might eventually figure it out by googling, but lots of people will just create duplicate issues on pyenv or similarly affected projects. The caveats section at least presents the additional info for a moment where the user might read it and save themselves (and package maintainers) some trouble. I guess I don't understand why the caveats section is now "considered harmful".

MikeMcQuaid:
Sorry, upstream documentation is a better place for this.

PR #11211, opened 2017-03-17
guile: Update to 2.2.0

ts826848:

-----

Devel is now older than stable, but doesn't seem a "real" devel is available, since 2.2.0 just came out?

### DIFF
--- a/Formula/autogen.rb
+++ b/Formula/autogen.rb
@@ -4,6 +4,7 @@ class Autogen < Formula
   url "https://ftpmirror.gnu.org/autogen/autogen-5.18.7.tar.xz"
   mirror "https://ftp.gnu.org/gnu/autogen/autogen-5.18.7.tar.xz"
   sha256 "a7a580a5e18931cb341b255cec2fee2dfd81bea5ddbf0d8ad722703e19aaa405"
+  revision 1
 
   bottle do
     sha256 "34f6959cbd858533dd3d86fa88663247f619fe2650df135616f983479da6cb68" => :sierra
--- a/Formula/autogen.rb
+++ b/Formula/autogen.rb
@@ -16,6 +16,12 @@ class Autogen < Formula
   depends_on "pkg-config" => :build
   depends_on "guile"
 
+  # Allow guile 2.2 to be used
+  patch do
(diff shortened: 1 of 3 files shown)


ilovezfs:
other patch is still needed?

ilovezfs:
Also the devel spec is older than the stable spec … is there a new devel version that actually is > 2.2.0?

ilovezfs:
@DomT4 I hate to even ask this question, but … do you think we're going to need a guile@2.0?

aw1621107:
Don't know about the patch. Trying right now...

Latest tag in the guile repo appears to be 2.2.0, so doesn't seem to be a devel spec at the moment. Should I just get rid of it?

ilovezfs:
>Should I just get rid of it?

Yup.
```
iMac-TMP:~ joe$ brew audit guile
guile:
  * devel version 2.1.8 is older than stable version 2.2.0
Error: 1 problem in 1 formula
```

aw1621107:
Looks like the other patch shouldn't be necessary. From libguile/filesys.c:1497:

```
     /* mkostemp(2) only defines O_APPEND, O_SYNC, and O_CLOEXEC to be
         useful, as O_RDWR|O_CREAT|O_EXCL are implicitly added.  It also
         notes that other flags may error on some systems, which turns
         out to be the case.  Of those flags, O_APPEND is the only one
         of interest anyway, so limit to that flag.  */
      open_flags &= O_APPEND;
```

ilovezfs:
@aw1621107 looks like we need revision bumps for mdk and autogen

ilovezfs:
and actually autogen doesn't build :(

aw1621107:
Should we hold off on this until autogen is fixed?

ilovezfs:
push the revision bumps and I'll add my patch :)

aw1621107:
Done!

ilovezfs:
patch added

DomT4:
> @DomT4 I hate to even ask this question, but … do you think we're going to need a guile@2.0?

Hopefully not 😖. Possibly for `gnutls`; the compatibility there hasn't always been great.

JCount:
@aw1621107 Thank you for another great contribution! 😸

PR #6480, opened 2016-11-01
terraform 0.7.8

cblecker:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

Bump version to v0.7.8
Fix dependency formatting, per @MikeMcQuaid 's comment [here](https://github.com/Homebrew/homebrew-core/pull/6149#discussion_r84594821).

PR #11253, opened 2017-03-18
vault: leave build tool fetching to Makefile

cblecker:

-----

### DIFF
--- a/Formula/vault.rb
+++ b/Formula/vault.rb
@@ -1,5 +1,3 @@
-require "language/go"
-
 # Please don't update this formula until the release is official via
 # mailing list or blog post. There's a history of GitHub tags moving around.
 # https://github.com/hashicorp/vault/issues/1051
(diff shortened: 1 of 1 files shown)


DomT4:
Calling this a "cleanup" isn't really an accurate explanation for the changes. The changes here seem to be simply removing the build tool fetching from the Homebrew formula and relying on the upstream `Makefile` to handle that.

DomT4:
I'm still not really sure of the merits here. Is there an issue with the current version that is resolved by using this method? I don't see an upside off the top of my head, but I see the downside that this breaks Homebrew's caching.

cblecker:
I was operating based on https://github.com/Homebrew/homebrew-core/pull/6480#issuecomment-257783714, which noted that the goal was to eliminate go_resource blocks where we can. I was able to work this out this for the terraform formula, and noticed it was also possible to do it for packer and vault.

DomT4:
The goal is to eliminate `go_resource`s with upstream support though, or at least it used to be.

The idea being that `go_resource`s would die in favour of upstream `vendor` support, rather than simply ditching them and going back to a system where things get to run `go get blah` themselves, which breaks caching and also is almost never pinned to a specific commit/tag.

CC @ilovezfs though; it's his decision to make here. Things may have changed in recent times.

ilovezfs:
what Dom said.

cblecker:
This all makes sense -- It looks like I just didn't understand clearly the feedback and how to go about it. Thank you both for explaining it!

DomT4:
It should become less ambiguous if upstream ever tag a new release for `gox` and that can be turned into a new formula and used as a `:build` dep, hopefully. That's one of the most persistent ones sticking around, along with its `iochan` dep.

ilovezfs:
Yeah, I'm tempted to just create a formula with an invented version since upstream seems to be ignoring all the issues on GitHub currently.

DomT4:
Aye, wouldn't say that was a bad idea at this point.

PR #11256, opened 2017-03-18
magic-wormhole 0.9.2 (new formula)

tdsmith:
CLI tool for securely sending data between computers. https://github.com/warner/magic-wormhole

### DIFF
--- a/Formula/magic-wormhole.rb
+++ b/Formula/magic-wormhole.rb
@@ -1,0 +1,153 @@
+class MagicWormhole < Formula
+  include Language::Python::Virtualenv
+
+  desc "Securely transfers data between computers"
+  homepage "https://github.com/warner/magic-wormhole"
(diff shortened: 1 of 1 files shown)


Formula/magic-wormhole.rb
```diff
@@ -0,0 +1,153 @@
...
+  end
+
+  def install
+    ENV["SODIUM_INSTALL"] = "system"
+    virtualenv_install_with_resources
+  end
+
+  test do
+    n = Time.now.to_i
+    send_thread = Thread.new do
```
ilovezfs: why not fork exec?

tdsmith: Oh, I see that pattern now; I forgot how we did this.

tdsmith: Well, the output is nicer with `system`. Unlike most of these, the `send` process isn't long-running, and terminates on its own once the `receive` process ends. Consistency's probably more important though.

PR #11260, opened 2017-03-19
Import of versioned formulas

fxcoudert:
Based on #10859 (because I couldn't push to @MikeMcQuaid's branch), adding the missing alias `kibana@5.2 -> kibana`.

The two other audit issues are:

```
kibana@4.1:
  * Use Language::Node for npm install args
kibana@4.4:
  * Use Language::Node for npm install args
```

which I would be inclined to ignore right now given that those are existing formulas.


MikeMcQuaid:
Nice work 👏

ilovezfs:
Why are we ignoring
```
kibana@4.1:
  * Use Language::Node for npm install args
kibana@4.4:
  * Use Language::Node for npm install args
```
?

fxcoudert:
@ilovezfs I am happy if people want to fix it, but given that those were working formulas I figured a rewrite would be considered too high a barrier for migration in the short term. Also: because I could not figure out how to fix it exactly :)

ilovezfs:
As far as I know, the main `kibana` formula isn't failing that audit.

fxcoudert:
Yes, and I could not figure out why, as the build process is pretty similar :(

ilovezfs:
It's possible it's a bug in the audit. There's something to be learned here.

ilovezfs:
@fxcoudert yes this is a bug in the audit: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L1179. It has a specific exemption for kibana and its versioned formulae, but the change to the `@` naming scheme hasn't been incorporated into that exemption yet.

PR #11282, opened 2017-03-19
Import homebrew/dupes into core.

MikeMcQuaid:
Part of the tap consolidation process. Related Homebrew/brew PR for checking that some of
these aren't depended on in https://github.com/Homebrew/brew/pull/2365.

### DIFF
--- a/Formula/angband.rb
+++ b/Formula/angband.rb
@@ -15,8 +15,6 @@ class Angband < Formula
 
   depends_on "autoconf" => :build
   depends_on "automake" => :build
-  depends_on :x11 => :optional
-  depends_on "homebrew/dupes/tcl-tk" => "with-x11" if build.with? :x11
   depends_on "sdl" => :optional
   if build.with? "sdl"
     depends_on "sdl_image"
@@ -36,7 +34,6 @@ class Angband < Formula
       --disable-sdltest
       --with-ncurses-prefix=#{MacOS.sdk_path}/usr
     ]
-    args << "--disable-x11" if build.without? :x11
     args << "--enable-sdl" if build.with? "sdl"
 
     system "./configure", *args
--- a/Formula/heimdal.rb
+++ b/Formula/heimdal.rb
@@ -1,0 +1,20 @@
+class Heimdal < Formula
+  desc "Free Kerberos 5 implementation"
+  homepage "https://www.h5l.org"
+  url "https://www.h5l.org/dist/src/heimdal-1.6rc2.tar.gz"
+  sha256 "8fe19e0f12ff0d0c2f20a6cf56727deaabd89db9c197288022fb489e103b730d"
+  revision 1
+
+  keg_only :provided_by_osx
+
+  def install
+    args = %W[
+      --disable-debug
+      --disable-dependency-tracking
+      --prefix=#{prefix}
+    ]
+
+    system "./configure", *args
+    system "make", "install"
+  end
+end
--- a/Formula/python.rb
+++ b/Formula/python.rb
@@ -34,7 +34,6 @@ class Python < Formula
   depends_on "openssl"
   depends_on "tcl-tk" => :optional
   depends_on "berkeley-db@4" => :optional
-  depends_on :x11 if build.with?("tcl-tk") && Tab.for_name("homebrew/dupes/tcl-tk").with?("x11")
 
   skip_clean "bin/pip", "bin/pip-2.7"
   skip_clean "bin/easy_install", "bin/easy_install-2.7"
@@ -159,7 +158,7 @@ class Python < Formula
     end
 
     if build.with? "tcl-tk"
-      tcl_tk = Formula["homebrew/dupes/tcl-tk"].opt_prefix
+      tcl_tk = Formula["tcl-tk"].opt_prefix
       cppflags << "-I#{tcl_tk}/include"
       ldflags  << "-L#{tcl_tk}/lib"
     end
@@ -270,8 +269,8 @@ class Python < Formula
     end
 
     if build.with? "tcl-tk"
-      include_dirs << Formula["homebrew/dupes/tcl-tk"].opt_include
-      library_dirs << Formula["homebrew/dupes/tcl-tk"].opt_lib
+      include_dirs << Formula["tcl-tk"].opt_include
+      library_dirs << Formula["tcl-tk"].opt_lib
     end
 
     cfg = lib_cellar/"distutils/distutils.cfg"
--- a/Formula/python.rb
+++ b/Formula/python.rb
@@ -32,7 +32,7 @@ class Python < Formula
   depends_on "sqlite" => :recommended
   depends_on "gdbm" => :recommended
   depends_on "openssl"
-  depends_on "homebrew/dupes/tcl-tk" => :optional
+  depends_on "tcl-tk" => :optional
   depends_on "berkeley-db@4" => :optional
   depends_on :x11 if build.with?("tcl-tk") && Tab.for_name("homebrew/dupes/tcl-tk").with?("x11")
 
--- a/Formula/tcl-tk.rb
+++ b/Formula/tcl-tk.rb
@@ -21,10 +21,6 @@ class TclTk < Formula
   option "with-threads", "Build with multithreading support"
   option "without-tcllib", "Don't build tcllib (utility modules)"
   option "without-tk", "Don't build the Tk (window toolkit)"
-  option "with-x11", "Build X11-based Tk instead of Aqua-based Tk"
-
-  depends_on :x11 => :optional
-  depends_on "pkg-config" => :build if build.with? "x11"
 
   resource "tk" do
     url "https://downloads.sourceforge.net/project/tcl/Tcl/8.6.6/tk8.6.6-src.tar.gz"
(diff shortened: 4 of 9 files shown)


MikeMcQuaid:
Let's hold off merging this until we have a new tag with https://github.com/Homebrew/brew/pull/1732.

tdsmith:
I merged an openssh upgrade into homebrew-dupes this morning & @JCount pointed out that you might not have been expecting that; sorry! https://github.com/Homebrew/homebrew-dupes/commit/628b3cd8833d74c28acf79a596ab971609dbf33a

MikeMcQuaid:
@tdsmith A delayed: no worries!

fxcoudert:
```
==> Verifying tcl-tk--tcllib-1.18.tar.gz checksum
Error: SHA256 mismatch
Expected: 6a87881f545afb69c1130f60984b5d35cc22f1593b0835b982871c188fde3de8
Actual: 9988b4385403c2aac78743fd3fce2d22e82686a56e6ca25942cb83c7d9e641db
```

What is even weirder is that the original formula in `dupes` had yet another shasum:

```
    url "https://downloads.sourceforge.net/project/tcllib/tcllib/1.18/tcllib-1.18.tar.gz"
    sha256 "72667ecbbd41af740157ee346db77734d1245b41dffc13ac80ca678dd3ccb515"
```

Formula/srm.rb
```diff
@@ -0,0 +1,28 @@
...
+  sha256 "7583c1120e911e292f22b4a1d949b32c23518038afd966d527dae87c61565283"
+
+  bottle do
+    cellar :any_skip_relocation
+    sha256 "3965dac338dab92d3c25306ec55dda0b1b75b78fb1f9060cd2e1c7b1e842f7a5" => :el_capitan
+    sha256 "bad23b6a0f2eebf37cc5b9f02f0001f5bf5af46fab55dd67aeb7b67796891b90" => :yosemite
+    sha256 "800ceccc9adecec1ceabfbc6c3d2aaa1da682db174ab02ae75cbcbba4cee1270" => :mavericks
+  end
+
+  keg_only :provided_by_osx
```
fxcoudert: `if MacOS.version < :sierra` since this is was removed in Sierra

Formula/heimdal.rb
```diff
@@ -0,0 +1,24 @@
+class Heimdal < Formula
+  desc "Free Kerberos 5 implementation"
+  homepage "https://www.h5l.org"
+  url "https://www.h5l.org/dist/src/heimdal-1.6rc2.tar.gz"
```
fxcoudert: There is now a 7.1.0 release

Formula/bc.rb
```diff
@@ -0,0 +1,32 @@
+class Bc < Formula
+  desc "arbitrary precision numeric processing language"
+  homepage "https://www.gnu.org/software/bc/"
+  url "https://ftpmirror.gnu.org/bc/bc-1.06.tar.gz"
```
fxcoudert: There is now a 1.07

chdiza:
I'm wondering whether `bzip` should be dropped.  It hasn't been updated since 2010 and I'm pretty sure that every Mac from Snow Leopard (or perhaps Lion at the outside) and up has the latest version of `bzip`.

Formula/srm.rb
ilovezfs: I think we should probably not import this at all since apple removed it to prevent people from killing SSDs.

chdiza: Yeah, but lots of people still have spinning platters, either in their machines or as external disks.

ilovezfs: say it isn't so

Formula/python.rb
```diff
@@ -32,7 +32,7 @@ class Python < Formula
   depends_on "sqlite" => :recommended
   depends_on "gdbm" => :recommended
   depends_on "openssl"
-  depends_on "homebrew/dupes/tcl-tk" => :optional
+  depends_on "tcl-tk" => :optional
   depends_on "berkeley-db@4" => :optional
   depends_on :x11 if build.with?("tcl-tk") && Tab.for_name("homebrew/dupes/tcl-tk").with?("x11")
```
ilovezfs: needs the homebrew/dupes reference removed here too

MikeMcQuaid:
> nano: build should be attempted without the ncurses dep

Fails, it wants unicode supporting ncurses (which seems a good enough reason).

MikeMcQuaid:
> tcpdump: build should be attempted without the libpcap dep

I think given this is optional it's OK?

MikeMcQuaid:
> awk: needs ENV.deparallelize

Test needs updated, no? No idea how to do that; doesn't work like my normal `awk` does...

ilovezfs:
analytics for last 30 days on tcpdump

* homebrew/dupes/tcpdump
329 installs
* homebrew/dupes/tcpdump --with-libpcap
4 installs
* homebrew/dupes/tcpdump --HEAD --with-libpcap
2 installs

So
>I think given this is optional it's OK?

yes. or drop the option.

ilovezfs:
>Test needs updated, no? No idea how to do that; doesn't work like my normal awk does...

Nope the test is fine. The build must be deparallelized or the result is a horribly broken end binary. (generated files + makefile that assumes -j1 = :boom:)

The entire build takes a matter of seconds so it's fine to deparallelize IMO.

fxcoudert:
Reproduced and reported the LAPACK issue upstream: https://github.com/Reference-LAPACK/lapack/issues/139

ilovezfs:
CC @dpo regarding LAPACK issue.

ilovezfs:
>Fails, it wants unicode supporting ncurses (which seems a good enough reason).

Yup.

MikeMcQuaid:
> I'm trying a clean build of lapack 3.7.0 with gfortran 6.3.0 and Xcode 8.3, with cmake .. -DLAPACKE:BOOL=ON. It fails with:

@fxcoudert @ilovezfs should we just not pass that? Is it still useful without?

fxcoudert:
It's the C interface, so it's typically *very* useful :)

MikeMcQuaid:
> It's the C interface, so it's typically very useful :)

I don't C where you're going with this 🥁 :trollface:

Formula/tcl-tk.rb
```diff
@@ -62,13 +58,7 @@ def install
         args = ["--prefix=#{prefix}", "--mandir=#{man}", "--with-tcl=#{lib}"]
         args << "--enable-threads" if build.with? "threads"
         args << "--enable-64bit" if MacOS.prefer_64_bit?
-
-        if build.with? "x11"
-          args << "--with-x"
-        else
-          args << "--enable-aqua=yes"
```
fxcoudert: Why did you remove `--enable-aqua=yes`?

MikeMcQuaid: I think that'll fix the audit failure, we'll see.

Formula/angband.rb
```diff
@@ -16,7 +16,7 @@ class Angband < Formula
   depends_on "autoconf" => :build
   depends_on "automake" => :build
   depends_on :x11 => :optional
-  depends_on "homebrew/dupes/tcl-tk" => "with-x11" if build.with? :x11
+  depends_on "tcl-tk" => "with-x11" if build.with? :x11
```
fxcoudert: We should remove that entirely, since we don't offer `tcl-tk` with X11 anymore…

MikeMcQuaid: Done, thanks.

Formula/heimdal.rb
MikeMcQuaid: Going to punt on it for now just to get this merged but: thanks.

Formula/tcl-tk.rb
MikeMcQuaid: It broke install instead 😁. Re-added.

MikeMcQuaid:
> It's the C interface, so it's typically very useful :)

Going to punt on `lapack` for now to get this merged before upstream figures something out. There's an argument that if it's been broken on Sierra for a while with no reports then that's a candidate for deletion.

fxcoudert:
What is weird is that it does have a Sierra bottle, so it worked at some point. I actually wonder if it's not a bug in `cmake 3.7.2`.

MikeMcQuaid:
> What is weird is that it does have a Sierra bottle, so it worked at some point. I actually wonder if it's not a bug in cmake 3.7.2.

Yeh, could be.

fxcoudert:
Lapack compiles (with same options) with cmake-3.6.3.

fxcoudert:
lapack is taken care of in #12033

MikeMcQuaid:
Pulled without `angband` changes and `heimdal` inclusion.

DomT4:
Probably a known thing but after something like this PR running `brew bundle cleanup` wants to uninstall all your dupe-installed formulae.

Not sure there's an easy fix given if you've installed something from a tap that duplicates a core formula you probably want to keep using the tap formula rather than matching the shorter core name & having that installed, but, raising it as an FYI.

MikeMcQuaid:
> Not sure there's an easy fix

It may be doable in Homebrew/bundle but it'd probably be low priority until someone gets confused enough by it. Eventually you'll want to rewrite your Brewfile anyway so chances are this will motivate you to do it.

PR #11283, opened 2017-03-19
Import from homebrew/fuse

MikeMcQuaid:
Import the contents of the homebrew/fuse tap.

Note that I will need to install `osxfuse` on our CI machines (without the kEXT) before this can be tested and merged.


MikeMcQuaid:
> Note that I will need to install osxfuse on our CI machines (without the kEXT) before this can be tested and merged.

This was done.

ilovezfs:
Non-test-block, non-notability audit issues:
- [ ] archivemount: build should be attempted without the libarchive dep
- [ ] ext2fuse: http://sourceforge.net/projects/ext2fuse could use HTTPS rather than HTTP

Missing test blocks:
- curlftpfs
- ext2fuse
- ext4fuse
- ifuse
- ntfs-3g
- rofs-filtered

MikeMcQuaid:
> archivemount: build should be attempted without the libarchive dep

Did that, failed.

ilovezfs (approved):

PR #11281, opened 2017-03-19
youtube-dl 2017.03.20

zmwangx:
Created with `brew bump-formula-pr`.

PR #11284, opened 2017-03-19
youtube-dl: skip make step for stable and stop bottling

zmwangx:

---

- Make artifacts are already included in release tarballs (at least for now);
- Stop bottling (#11281);
- Remove useless comments (all GitHub releases are published on PyPI).

### DIFF
--- a/Formula/youtube-dl.rb
+++ b/Formula/youtube-dl.rb
@@ -11,8 +11,6 @@ class YoutubeDl < Formula
     depends_on "pandoc" => :build
   end
 
-  depends_on "rtmpdump" => :optional
-
   def install
     system "make", "PREFIX=#{prefix}" if build.head?
     bin.install "youtube-dl"
@@ -20,10 +18,6 @@ class YoutubeDl < Formula
     bash_completion.install "youtube-dl.bash-completion"
     zsh_completion.install "youtube-dl.zsh" => "_youtube-dl"
     fish_completion.install "youtube-dl.fish"
-  end
-
-  def caveats
-    "To use post-processing options, `brew install ffmpeg` or `brew install libav`."
   end
 
   test do
--- a/Formula/youtube-dl.rb
+++ b/Formula/youtube-dl.rb
@@ -1,18 +1,10 @@
-# Please only update to versions that are published on PyPi as there are too
-# many releases for us to update to every single one:
-# https://pypi.python.org/pypi/youtube_dl
 class YoutubeDl < Formula
   desc "Download YouTube videos from the command-line"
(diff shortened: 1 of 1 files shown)


Formula/youtube-dl.rb
```diff
@@ -1,18 +1,10 @@
-# Please only update to versions that are published on PyPi as there are too
-# many releases for us to update to every single one:
-# https://pypi.python.org/pypi/youtube_dl
```
fxcoudert: Let's keep that comment

Formula/youtube-dl.rb
```diff
@@ -22,7 +14,7 @@ class YoutubeDl < Formula
   depends_on "rtmpdump" => :optional
```
fxcoudert: Can we remove `rtmpdump` here and treat it same as `ffmep` and `libav`, i.e. discuss it in `caveats`?

fxcoudert (changes requested):

Formula/youtube-dl.rb
zmwangx: Why? As I said, all GitHub releases are published to PyPI. https://pypi.org/project/youtube_dl/#history

Plus the fact that we're using GitHub release tarballs, not PyPI source tarballs.

Formula/youtube-dl.rb
zmwangx: Sure.

fxcoudert (approved):

ilovezfs:
Let's hold off on merge for a bit. @MikeMcQuaid any thoughts on the caveats?

Formula/youtube-dl.rb
```diff
@@ -1,37 +1,31 @@
...
     fish_completion.install "youtube-dl.fish"
   end
 
-  def caveats
-    "To use post-processing options, `brew install ffmpeg` or `brew install libav`."
+  def caveats; <<-EOS.undent
+    To use post-processing options, `brew install ffmpeg` or `brew install libav`.
+
+    Certain video formats may require rtmpdump, mplayer, or mpv to download.
+    All these packages can be installed with brew if necessary.
```
MikeMcQuaid: Don't think we need any more caveats here.

zmwangx: @MikeMcQuaid Just blow away the whole caveats section? I agree because readers are supposed to read https://github.com/rg3/youtube-dl#do-i-need-any-other-programs.

MikeMcQuaid: Yeh, go for it

ilovezfs:
🎆 🍡 🎊

PR #11285, opened 2017-03-19
avra: apply patch to avoid "abort trap 6" crash

adammck:

-----

Seems to be [an open bug](https://sourceforge.net/p/avra/bugs/75/) upstream. [The fix](https://sourceforge.net/p/avra/patches/16/) was submitted by one [olal](https://sourceforge.net/u/olalid/profile/) on SF.

### DIFF
--- a/Formula/avra.rb
+++ b/Formula/avra.rb
@@ -13,6 +13,13 @@ class Avra < Formula
     sha256 "2e208cec5f270c91c9afc0349236a9abb0622e1e8208c67d25c90f017fcecf65" => :mavericks
   end
 
+  # Crashes with 'abort trap 6' unless this fix is applied.
+  # See: https://sourceforge.net/p/avra/patches/16/
+  patch do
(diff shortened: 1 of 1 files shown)


ilovezfs:
This will need to be reported upstream as a patch link from 2015-01-19 is unlikely to get anyone's attention at this point without further action.

adammck:
The project hasn't been updated in a while. I'm trying to track down a maintainer. I'll leave this open for now, and replace it with a version update if I can get the patch merged and released.

fxcoudert:
Can you add a `revision 1` line after `sha256`? Then I think we ought to merge this: upstream isn't moving, so it's either that or removing the formula.

adammck:
Sure, done. Is that so we can force a rebuild if the patch changes?

fxcoudert:
The revision will mean that users running `upgrade` will get the new binaries after we merge this.

fxcoudert:
Thanks @adammck for your contribution!

PR #11291, opened 2017-03-20
ffmpeg: add alias for ffmpeg@3.2

msbit:
- ☑️  Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- ☑️  Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- ☑️  Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- ☑️  Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

### DIFF
--- a/Aliases/ffmpeg@3
+++ b/Aliases/ffmpeg@3
@@ -1,0 +1,1 @@
+../Formula/ffmpeg.rb


fxcoudert:
Why? As far as I know, we don't have any versioned `ffmpeg` formula, in which case an alias is not called for.

ilovezfs:
@fxcoudert 

```
$ ls `brew --repo homebrew/core`/Formula/ffmpeg*@*.rb
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/ffmpeg@2.8.rb
```

fxcoudert:
Shouldn't `ffmpeg@2.8` be renamed `ffmpeg@2` and the new alias be `ffmpeg@3`?

ilovezfs:
@fxcoudert you will need to consult the oracle. CC @MikeMcQuaid is your divining rod handy?

retokromer:
FYI: The currently maintained branches are: 2.8, 3.0, 3.1 and 3.2. The branch 2.9 is no longer maintained and 3.3 is imminent.

msbit:
@fxcoudert Seems the precedent is there for `ffmpeg@major.minor`, but I can see the value in avoiding a minor if that makes sense upstream; I guess not everyone does semver the same way.

Basically I was following the `brew audit` instructions:

```
$ brew audit --strict --online ffmpeg
ffmpeg:
      ln -s ../Formula/ffmpeg.rb ffmpeg@3.2
Error: 1 problem in 1 formula
```

In terms of the guidelines for versioning, is there a piece of official documentation regarding major/minor or is it on a case by case basis?

MikeMcQuaid:
@ilovezfs The Oracle Has Little Preference On This Matter

MikeMcQuaid:
> is it on a case by case basis?

This..

fxcoudert:
> The currently maintained branches are: 2.8, 3.0, 3.1 and 3.2.

So there will be no other maintained 2.x brand, and thus `ffmpeg@2` is unambiguous.

msbit:
@fxcoudert sounds sensible, though probably a topic for another day/PR. In this case, seems as though `@3.2` is  acceptable.

fxcoudert:
We should decide something, then stick to it. No point in changing that later in another PR.

msbit:
I meant that moving `ffmpeg@2.8` to `ffmpeg@2` is sensible, but outside the scope of this PR. Deciding to move `formula@major.minor` to `formula@major` when there is only one supported upstream version for `major` will only come into play with the `ffmpeg@3*` formulae/aliases when `ffmpeg` v4 becomes available and is added to `homebrew-core`. So, setting up this alias is in keeping with deciding on that course of action and sticking to it.

ilovezfs:
@msbit we prefer to keep the number of renames and alias changes to a minimum, so it's well within scope.

@zmwangx any opinions here?

msbit:
@ilovezfs @fxcoudert so, ideally, you'd have `ffmpeg@2` and `ffmpeg@3` (the latter an alias to `ffmpeg`), and you see the addition of an `ffmpeg@3.2` alias as an issue because it would require (at least) an additional rename to `ffmpeg@3` at a later date when it becomes the only supported 3.x version of `ffmpeg` in `homebrew-core` (in the same way that `ffmpeg@2.8` would now require a rename to `ffmpeg@2`).

The counter-argument I'd put forward is that the `ffmpeg` project supports multiple minor versions, and while there isn't a `@3.1` or `@3.0` in `homebrew-core` at the moment, someone may need to add one at a later date. Also, a future `@3.x` may be more breaking than expected, so there would be a need to differentiate between 3.x and 3.2, which would require renaming too.

Anyhow, that's my point of view as an infrequent contributor, I'll await the decision from the maintainer hive-mind 😄

zmwangx:
> The counter-argument I'd put forward is that the ffmpeg project supports multiple minor versions, and while there isn't a @3.1 or @3.0 in homebrew-core at the moment, someone may need to add one at a later date. Also, a future @3.x may be more breaking than expected, so there would be a need to differentiate between 3.x and 3.2, which would require renaming too.

For end users using the `ffmpeg` family of command line tools, there's little to no reason to use a legacy branch of FFmpeg. For developers using the `libav*` family of libraries, as far as I know it's been relatively painless to port code from 2 to 3 (in the grand scheme of things), let alone from 3.x to 3.x+1. Some projects also vendor `libav*`. IMO adding unnecessary complexity for the sake of one of the countless things that *may* happen in the future is a pit we all fall into way to often.

Disclosure: @ilovezfs might know that I've never been a big fan of versions in core, so I'm apparently biases towards fewer versions.

msbit:
Okay, I'll clean it up when I get back in front of my main computer, should be sometime this evening here (midday UTC).

msbit:
Updated.

fxcoudert:
Thanks @msbit for your contribution!

msbit:
👌🏼

PR #11296, opened 2017-03-20
add rxcpp formula

vigsterkr:


Formula/rxcpp.rb
```diff
@@ -0,0 +1,62 @@
...
+  desc "Reactive Extensions for C++"
+  homepage "https://rxcpp.codeplex.com/"
+  url "https://github.com/Reactive-Extensions/RxCpp.git",
+      :tag => "v3.0.0",
+      :revision => "8290f92f744f807e83b1bfe9e8c0ffd162140ec8"
+  head "https://github.com/Reactive-Extensions/RxCpp.git"
+
+  bottle :unneeded
+
+  option "with-test", "Build and run the test suite"
```
fxcoudert: How long do the tests take on a typical machine?

Formula/rxcpp.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  if build.with? "test"
+    depends_on "cmake" => :build
```
fxcoudert: `depends_on "cmake" => :build if build.with? "test"`

Formula/rxcpp.rb
```diff
@@ -0,0 +1,62 @@
...
+  def install
+    if build.with? "test"
+      ENV.cxx11
+      mkdir "build" do
+        system "cmake", "..", *std_cmake_args
+        system "make"
+        system "./build/test/rxcpp_test_all"
+      end
+    end
+    include.install "Rx/v2/src/rxcpp"
```
fxcoudert: Is this library header-only?
Can't we use `make install`?

Formula/rxcpp.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        int main(int argc, char** argv) {
+            auto sc = rxsc::make_test();
+            auto w = sc.create_worker();
+            const rxsc::test::messages<int> on;
+
+            auto res = w.start(
+                []() {
+                    return rx::observable<>::empty<int>()
+                        // forget type to workaround lambda deduction bug on msvc 2013
```
fxcoudert: We are not going to run with msvc 2013 :)

Formula/rxcpp.rb
vigsterkr: to compile it it's rather long... on my mbp pro 2015 took about 20 mins to get the test suit compiled

Formula/rxcpp.rb
vigsterkr: yeah we can... either way is fine.

Formula/rxcpp.rb
vigsterkr: obviously it's a copy-paste error from a test from them. i can remove if you think this'll actually make any difference :)

vigsterkr:
@fxcoudert thnx for the review. do you want `make install` then?

fxcoudert:
Yes, please use `make install`.

vigsterkr:
@fxcoudert ah yeah i totally forgot... if you want make install for that we one needs cmake for which you need to run `make` that will consequently run the compilation of all tests... it cannot be turned off as the cmake of the project is not so advanced... so either one creates a patch or just does simply `include.install "Rx/v2/src/rxcpp"` just as in case of `cereal`.

or we want to compile the tests? this is a header only library....

fxcoudert:
I'm actually starting to wonder: if it's a one-header only library, with installation process not clearly finalised, do we really want/need it in Homebrew? Isn't that the sort of library that people would simply include in other projects' source code?

Do we know of any other software that depends on it and does not include it directly?

vigsterkr:
@fxcoudert look i've just tried to be nice here with actually contributing this to brew. honestly since i can use this from my own tap i really... well neutral regarding whether this goes in or stays out... if you think that because of a simple `include.install "Rx/v2/src/rxcpp" just` vs `make install` - that is being used already in other formulas already in core - is a blocker here please feel free to close this PR.

PR #11297, opened 2017-03-20
v8 3.15 from homebrew/versions

msbit:

-----

### DIFF
--- a/Aliases/v8@5.1
+++ b/Aliases/v8@5.1
@@ -1,0 +1,1 @@
+../Formula/v8.rb
--- a/formula_renames.json
+++ b/formula_renames.json
@@ -143,5 +143,6 @@
   "stash-cli": "atlassian-cli",
   "tomcat6": "tomcat@6",
   "tomcat7": "tomcat@7",
+  "v8-315": "v8@3.15",
   "wires": "geckodriver"
 }
--- a/Formula/v8@3.15.rb
+++ b/Formula/v8@3.15.rb
@@ -1,0 +1,48 @@
+class V8AT315 < Formula
+  desc "Google's JavaScript engine"
+  homepage "https://github.com/v8/v8/wiki"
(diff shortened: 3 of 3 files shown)


Aliases/v8@5.1
```diff
@@ -0,0 +1 @@
+../Formula/v8.rb
```
fxcoudert: Why name the alias `v8@5.1` and not `v8@5` which seems more consistent with our policy?

formula_renames.json
```diff
@@ -143,5 +143,6 @@
   "stash-cli": "atlassian-cli",
   "tomcat6": "tomcat@6",
   "tomcat7": "tomcat@7",
+  "v8-315": "v8@3.15",
```
fxcoudert: Same here: shouldn't we rename it to `v8@3`?

msbit:
@fxcoudert hmm, I'm not sure about the major/minor for libv8. The project seems to run at a crazy pace, with 1751 tags under `v4`, and 3793 and counting under `v5`. I think the importance of that specific version (`v3.15.11.18`) was that it was the latest version that was compatible with a particular Ruby gem version (and perhaps a major Ruby version itself). I definitely remember going through the pain of getting that working in the past.

I know it's not entirely in keeping with trying to avoid minor version formulae/aliases, but if it were to go to `v8@3` from a glance I'd expect that to be `v3.32.7`, the latest 3.x release, so I _feel_ like it suits a `v8@3.15`. 

I can accept the status quo, and I'm not the person to make the bigger picture decision about `v8` in Homebrew in general, so perhaps it's best to drop this and it's sister PR (https://github.com/Homebrew/homebrew-versions/pull/1548).

msbit:
@fxcoudert closing this and related PR, as V8 support policy is for current and previous minor releases (5.7 and 5.8 at the moment) and a new minor is cut every 6 weeks. This makes this version unsupported so not really a good candidate for inclusion after all.

Issue #10839, opened 2017-03-10
audit --online failures


ilovezfs:
- [x] libtextcat: 
  * Stable: The URL http://software.wise-guys.nl/download/libtextcat-2.2.tar.gz is not reachable (HTTP status code 500)
- [x] man2html:
  * Stable: The URL http://dcssrv1.oit.uci.edu/indiv/ehood/tar/man2html3.0.1.tar.gz is not reachable
- [x] pce:
  * Devel: The URL http://www.hampa.ch/pub/pce/pre/pce-20160308-72f1e10.tar.gz is not reachable (HTTP status code 404)
- [x] pianod: (https://github.com/Homebrew/homebrew-core/pull/11634)
  * Devel: The URL http://deviousfish.com/Downloads/pianod2/Devel/pianod2-189.tar.gz is not reachable (HTTP status code 404)
- [x] polipo:
  * HEAD: The URL git://git.wifi.pps.jussieu.fr/polipo is not a valid git URL
- [ ] qt@5.5:
  * xcodebuild should be passed an explicit "SYMROOT"
- [x] riak:
  * Commented-out dep "openssl"
- [x] schroedinger:
  * Stable: The URL http://diracvideo.org/download/schroedinger/schroedinger-1.0.11.tar.gz is not reachable
- [x] spidermonkey:
  * HEAD: The URL https://hg.mozilla.org/tracemonkey/archive/tip.tar.gz is not reachable (HTTP status code 404)
- [x] webarchiver:
  * xcodebuild should be passed an explicit "SYMROOT"
- [ ] xctool:
  * xcodebuild should be passed an explicit "SYMROOT"
---
- [x] bigdata: (https://github.com/Homebrew/homebrew-core/pull/10929)
  * Stable: The URL https://downloads.sourceforge.net/project/bigdata/bigdata/1.5.3/bigdata-bundled.jar is not reachable (HTTP status code 404)
- [x] boom-completion: (https://github.com/Homebrew/homebrew-core/pull/10938)
  * The URL http://zachholman.com/boom could use HTTPS rather than HTTP
- [x] boost@1.59: (https://github.com/Homebrew/homebrew-core/pull/10939)
  * The URL http://www.boost.org could use HTTPS rather than HTTP
- [x] cabocha: (https://github.com/Homebrew/homebrew-core/pull/10942)
[...]

PR #11310, opened 2017-03-20
Fix some more `audit --online` issues

fxcoudert:
See https://github.com/Homebrew/homebrew-core/issues/10839

### DIFF
--- a/Formula/man2html.rb
+++ b/Formula/man2html.rb
@@ -1,7 +1,7 @@
 class Man2html < Formula
   desc "Convert nroff man pages to HTML"
   homepage "https://savannah.nongnu.org/projects/man2html/"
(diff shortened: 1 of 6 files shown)


fxcoudert:
Pushed all but `pianod` which will need to be adapted:

```
clang++ -DHAVE_CONFIG_H -I. -I../../..  -I../../../src/common -I../../../src/include -I../../../src/mediaunits   -g -O2 -c -o libaudiooutput_a-osxplayer.o `test -f 'osxplayer.mm' || echo './'`osxplayer.mm
In file included from osxplayer.mm:16:
../../../src/common/fundamentals.h:267:7: error: delegating constructors are permitted only in C++11
    : CommandError (r, message.c_str()) {};
      ^~~~~~~~~~~~
```

PR #11346, opened 2017-03-20
ecm: migrate to boneyard

fxcoudert:
Created with `brew boneyard-formula-pr` because it is unmaintained since 2012.


ilovezfs:
Is there a reason beyond that? That's not usually a sufficient reason to boneyard anything.

fxcoudert:
It doesn't have a homepage, the source code is not hosted anywhere and the project has not been continued or taken over by anyone else. Searching for the source code filename (`ecm100.zip`) in Google shows that FreeBSD and us are the top results. This doesn't exist as a software anymore.

We fall short of almost all of our own [criteria](http://docs.brew.sh/Acceptable-Formulae.html). This increase maintenance burden for almost no use.

ilovezfs:
122 installs in last 30 days.

PR #11350, opened 2017-03-20
xml2: migrate to boneyard

fxcoudert:
Created with `brew boneyard-formula-pr` because it is unmaintained since 2011.


zmwangx:
May I ask if there's more reason than "it is unmaintained since 2011"? It builds fine here, and Bintray statistics show quite a few downloads (330 over 30 days): https://bintray.com/homebrew/bottles/xml2#statistics.

To be clear, I don't use this formula myself, but it's worrisome to me that "X is unmaintained since 20XX" is used as the sole reason to trash a considerable number of formulae that might be working just fine (possibly indefinitely into the future). Many battle-hardened old Unix utilities can be called "unmaintained" — in many cases you can't find an upstream; you may find individual distros do a bit of downstream patching, and that's it. These utilities may be bug-free enough that they don't really need maintenance. A concrete example is Vixie Cron. It went unmaintained in the 90s (or 2004 if you count ISC Cron). Apple's port hasn't been touched since OS X 10.8, which you can confirm on opensource.apple.com. Yet to this day I use it literally every single minute.

In this specific case, the package is [carried by Debian](https://packages.debian.org/jessie/xml2), it still works (excuse me if I'm wrong), and people still use it, so I don't see why it should be trashed.

fxcoudert:
It doesn't have a homepage, the source code is not hosted anywhere, and the project has not been continued or taken over by anyone else.

We fall short of almost all of our own [criteria](http://docs.brew.sh/Acceptable-Formulae.html). This increase maintenance burden for little use, if we try to be a software conservation project.

ilovezfs:
97 installs in the last 30 days.

zmwangx:
> It doesn't have a homepage

Sure.

> the source code is not hosted anywhere

We have more than a few formulae grabbing source tarballs from the Debian repository. This could fall into the hosted by Debian category if you like.

> the project has not been continued or taken over by anyone else.

As I said, it doesn't have to be taken over by anyone. This is open source code on the open web, freely distributable by anyone.

> We fall short of almost all of our own criteria. 

Acceptable formulae criteria are mostly enforced at the time of *accepting* formulae. IMO retroactive enforcement *for the sake of enforcement* is misguided.

> This increase maintenance burden for little use, if we try to be a software conservation project.

The formula is not broken. If it's broken I'm all for throwing it out. But it's not.

MikeMcQuaid:
> We fall short of almost all of our own criteria.

The criteria for removal should be much higher than addition. I'm personally 👎 on boneyarding anything that still builds on Sierra with Clang and isn't getting somewhat regularly reported user upstream issues. Thoughts?

fxcoudert:
There is clear consensus…

PR #11359, opened 2017-03-21
fix ctop build

bcicen:
Update ctop to use makefile build, so version and appropriate build flags are included.


nijikon:
For future reference the preferred commit message format is `ctop 0.5.1`. Please complete the issue template in future PRs where this format is detailed in the linked Contributing guidelines.

PR #11361, opened 2017-03-21
elasticsearch: remove unmaintained versions

jasontedor:
This commit removes the formula for Elasticsearch 1.7 (EOL), and Elasticsearch 2.3 (unmaintained).


fxcoudert:
From our analytics: 12% of elasticsearch installs are for 2.4, 10% for 1.7, and 4% for 2.3.

JCount:
@jasontedor Would you mind explaining, if you can, what sort of a state Elasticsearch 2.3.5 is in now given the end of maintenance when 2.4.0 came out. Also, is the fact that its official EOL isn't until 2017-09-30 have any significance in this context?

Essentially, I suppose my question is whether or not it still has a valid use case, and if it would therefore be potentially worth keeping a bit longer for users who may be running outdated, legacy software. If you feel that we really should just stop distributing it, likely pushing people towards 2.4.4, please feel free to just say so.

(1.7 is definitely getting boneyarded, no question there.)

jasontedor:
> From our analytics: 12% of elasticsearch installs are for 2.4, 10% for 1.7, and 4% for 2.3.

Can you provide more detail? For example, over what time frame? What about the last two months (because 1.7 went EOL in January)?

jasontedor:
> Would you mind explaining, if you can, what sort of state Elasticsearch 2.3.5 is in now given the end of maintenance when 2.4.0 came out.

jasontedor:
> Would you mind explaining, if you can, what sort of state Elasticsearch 2.3.5 is in now given the end of maintenance when 2.4.0 came out. Also, does the fact that its official EOL is not until 2017-09-30 have any significance in this context?

It's supported, but unmaintained. This means that no bug fixes will be backported to 2.3.x, and only severe security fixes will be backported to 2.3.x.

> If you feel that we really should just stop distributing it, likely pushing people towards 2.4.4, please feel free to just say so.

Indeed, given that it is unmaintained, the best course of action for users would be to upgrade to 2.4.x. If users do need legacy versions, they are still accessible via the Elastic website. I would however prefer that only maintained versions be distributed here.

JCount:
@jasontedor Thanks for being clear. Not wanting us to be distributing unmaintained versions is perfectly reasonable, especially since any bugs or other issues would likely land in your lap. I will work on boneyarding both versions then.

jasontedor:
> Not wanting us to be distributing unmaintained versions is perfectly reasonable, especially since any bugs or other issues would likely land in your lap.

Precisely, thank you so much for understanding!

ilovezfs:
@jasontedor thanks for this PR!

PR #11379, opened 2017-03-21
Fix style for llvm@3.9 and ruby@2.3

maxim-belkin:
Before:
```sh
$ brew style *.rb
== llvm@3.9.rb ==
C: 82:  1: Extra blank line detected.
== ruby@2.3.rb ==
C: 25:  1: Extra blank line detected.

4091 files inspected, 2 offenses detected
```

After:
```sh
$ brew style llvm\@3.9.rb ruby\@2.3.rb 

2 files inspected, no offenses detected
```

### DIFF
--- a/Formula/llvm@3.9.rb
+++ b/Formula/llvm@3.9.rb
@@ -72,13 +72,13 @@ class LlvmAT39 < Formula
       sha256 "9ba5e61fc7bf8c7435f64e2629e0810c9b1d1b03aa5b5605b780d0e177b4cb46"
     end
   end
+
   bottle do
     cellar :any
     sha256 "367c8263a2ddb03b01026144cbacbb59a5c89678f0cd6b66a8ca7e54e3a6f09a" => :sierra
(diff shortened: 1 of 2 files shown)


fxcoudert:
Can you make those two commits, with commit message `llvm@3.9: fix style` (and same for ruby)

maxim-belkin:
sure...

ilovezfs:
It is odd this was no caught by `brew audit --new-formula` for llvm@3.9. I thought we ran `brew style` for that. :bug:?

ilovezfs:
Ah it's because it happened in the bottle block commit and `brew pull` doesn't re-check after applying that.

maxim-belkin:
but then it means that code that adds bottle block does it wrong

ilovezfs:
@maxim-belkin yup, there are definitely cases where it applies it incorrectly.

ilovezfs:
_ilovezfs will try to remember to be more careful next time._

fxcoudert:
Merged, thanks @maxim-belkin !

PR #11421, opened 2017-03-21
Compcert 3.0.1

beurdouche:

-----

### DIFF
--- a/Formula/compcert.rb
+++ b/Formula/compcert.rb
@@ -1,8 +1,8 @@
 class Compcert < Formula
   desc "Formally verified C compiler"
   homepage "http://compcert.inria.fr"
-  url "http://compcert.inria.fr/release/compcert-3.0.tgz"
-  sha256 "f5e42c528849fffa840a4c820ad4cc6257a7e518f3b3b0fe54c06fa60270dd0d"
+  url "http://compcert.inria.fr/release/compcert-3.0.1.tgz"
+  sha256 "09c7dc18c681231c6e83a963b283b66a9352a9611c9695f4b0c4b7df8c90f935"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:
The current version of `coq` in homebrew/core is 8.6, so the OPAM stuff can be removed in favor of `depends_on "coq" => :build`

beurdouche:
I was thinking that this would prevent problems in case of breaking changes from a Coq update as Compcert is really tied with a specific version of Coq. That said, I don't expect real breaking changes. @ilovezfs WDYT ?

ilovezfs:
Yup, good thought. Unfortunately, that will always be the case, so it's a dance we'll keep doing each time they're out of sync. But when the stars are aligned we might as well speed up the build for everyone by using the botted dependency.

beurdouche:
Ok, I'll let the build finish because I am curious and update the formula using brew dependencies soon ;)

ilovezfs:
Just shout if you need any help.

ilovezfs:
Failed because it needs a newer Menhir. So: https://github.com/Homebrew/homebrew-core/pull/11430.

ilovezfs:
https://github.com/Homebrew/homebrew-core/pull/11430 shipped.

ilovezfs:
@BrewTestBot test this please

fxcoudert:
Why do we not compile as 64-bit by default (when possible)?

ilovezfs:
@fxcoudert see https://github.com/Homebrew/homebrew-core/pull/9875

nijikon:

ilovezfs:
@beurdouche it looks like we got our wires crossed here. I'll handle changing it to `depends_on "coq" => :build`, so no worries.

PR #11442, opened 2017-03-21
lxc: Simplify formula

stgraber:
This does the following:
 - Use the official release tarballs from upstream
 - Use the embedded copies of all needed dependencies

That should greatly simplify the maintenance of this formula going
forward (only need to update the URL and sha256) and ensures that what's
shipped matches what upstream tested.

Upstream LXD CI does test the build on MacOS before generating the
release tarball, so using the dependencies we ship should be much more
reliable than pinning on random upstream hashes.

Signed-off-by: Stéphane Graber <stgraber@ubuntu.com>

-----

### DIFF
--- a/Formula/lxc.rb
+++ b/Formula/lxc.rb
@@ -1,10 +1,9 @@
-require "language/go"
-
 class Lxc < Formula
   desc "CLI client for interacting with LXD"
   homepage "https://linuxcontainers.org"
-  url "https://github.com/lxc/lxd/archive/lxd-2.12.tar.gz"
-  sha256 "8cc4143be017c1f6c70489c5745903fa8798acb7ed03f743b543689746eceb9f"
+  url "https://linuxcontainers.org/downloads/lxd/lxd-2.12.tar.gz"
+  sha256 "439bdadc7c63c523999a8d26b274a5c1c397d897197b43df8ce4b225a75585ff"
+  revision 1
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


ilovezfs:
Thanks @stgraber. Is this using a particular `go` vendor system or are you managing everything manually?

stgraber:
@ilovezfs so we don't try to pin any particular upstream commit of our dependencies. Instead we always test with whatever's latest and when we generate the release tarball, "make dist" includes the current version of all the dependencies. So that's not one of the usual go vendoring system which are typically used to pin your dependencies to particular commits.

ilovezfs:
Interesting approach. It lacks a bit of transparency in the git repository itself in terms of what revisions were used for a given tag but I'm not complaining.

The `prefix.install_metafiles` isn't needed since you're no longer moving things out of the buildpath. You get an automatic `buildpath.install_metafiles` when the install method returns.

stgraber:
@ilovezfs ah, thanks, pushed an updated version now.

As for what's actually in dist/, you're right. One way to improve that would be for "make dist" to record the HEAD hash for every dependency in a file in dist/src/.

fxcoudert:
How are the packages at the new URL different from the GitHub ones? Are there changes other than the included dependencies?

Formula/lxc.rb
```diff
@@ -1,10 +1,8 @@
-require "language/go"
-
 class Lxc < Formula
   desc "CLI client for interacting with LXD"
   homepage "https://linuxcontainers.org"
-  url "https://github.com/lxc/lxd/archive/lxd-2.12.tar.gz"
-  sha256 "8cc4143be017c1f6c70489c5745903fa8798acb7ed03f743b543689746eceb9f"
+  url "https://linuxcontainers.org/downloads/lxd/lxd-2.12.tar.gz"
+  sha256 "439bdadc7c63c523999a8d26b274a5c1c397d897197b43df8ce4b225a75585ff"
```
fxcoudert: Given the extension nature of the changes (even if it is the same version), we probably want a `revision 1` line after the `sha256` line.

stgraber:
@fxcoudert the tarballs at the new URL differ from the Github one in that those are the official, tested, signed tarballs, rather than an automatically generated tarball that Github generates on demand.

Github's tarballs are a simple git export as .tar.gz, whereas the ones generated by upstreams are done through the "make dist" target.

Formula/lxc.rb
stgraber: Updated

fxcoudert (approved):

ilovezfs (approved):

JCount:
Thanks a lot for this PR, @stgraber! I feel such a sense of joy when go_resource blocks disappear! 🎉

PR #11464, opened 2017-03-22
glib 2.52.0 and dependents

tschoonj:
Also: resurrection of latexila!
The glib formula fetches patches from my formula-patches fork. I will merge my branch into Homebrew/formula-patches and update the urls in the formula after this PR is merged in.

### DIFF
--- a/Formula/glib.rb
+++ b/Formula/glib.rb
@@ -1,8 +1,8 @@
 class Glib < Formula
   desc "Core application library for C"
   homepage "https://developer.gnome.org/glib/"
-  url "https://download.gnome.org/sources/glib/2.50/glib-2.50.3.tar.xz"
-  sha256 "82ee94bf4c01459b6b00cb9db0545c2237921e3060c0b74cff13fbc020cfd999"
+  url "https://download.gnome.org/sources/glib/2.52/glib-2.52.0.tar.xz"
+  sha256 "4578e3e077b1b978cafeec8d28b676c680aba0c0475923874c4c993403df311a"
 
   bottle do
     sha256 "53a03f3cae2e738580bf2d841325f8d8f8515f15c1dc8daff9e126f4b6cd2de0" => :sierra
@@ -15,6 +15,9 @@ class Glib < Formula
   deprecated_option "test" => "with-test"
 
   depends_on "pkg-config" => :build
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
   depends_on "gettext"
   depends_on "libffi"
   depends_on "pcre"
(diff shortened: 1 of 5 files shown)


Formula/glib.rb
```diff
@@ -15,6 +15,9 @@ class Glib < Formula
   deprecated_option "test" => "with-test"
 
   depends_on "pkg-config" => :build
+  depends_on "autoconf" => :build
```
fxcoudert: Can we add comments here to remember removing those after the patches go in?

Formula/glib.rb
```diff
@@ -76,6 +73,7 @@ def install
       --with-gio-module-dir=#{HOMEBREW_PREFIX}/lib/gio/modules
     ]
 
+    system "autoreconf", "-i", "-f"
```
fxcoudert: Same here, a comment would be helpful in the future.

Formula/glib.rb
```diff
@@ -33,35 +36,29 @@ class Glib < Formula
...
 
   def install
     inreplace %w[gio/gdbusprivate.c gio/xdgmime/xdgmime.c glib/gutils.c],
       "@@HOMEBREW_PREFIX@@", HOMEBREW_PREFIX
 
     # renaming is necessary for patches to work
-    mv "gio/gcocoanotificationbackend.c", "gio/gcocoanotificationbackend.m" unless MacOS.version < :mavericks
+    mv "gio/gcocoanotificationbackend.c", "gio/gcocoanotificationbackend.m"
     mv "gio/gnextstepsettingsbackend.c", "gio/gnextstepsettingsbackend.m"
+    # mv "gio/gosxappinfo.c", "gio/gosxappinfo.m"
```
fxcoudert: Why adding a commented-out line?

Formula/gobject-introspection.rb
```diff
@@ -14,10 +14,7 @@ class GobjectIntrospection < Formula
   depends_on "glib"
   depends_on "cairo"
   depends_on "libffi"
-  # System python in Mavericks or below has bug in distutils/sysconfig.py, which breaks the install.
-  #    Caught exception: <type 'exceptions.AttributeError'> AttributeError("'NoneType' object has no attribute 'get'",)
-  #    > /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/sysconfig.py(171)customize_compiler()
-  depends_on "python" if MacOS.version <= :mavericks
+  depends_on "python" # never switch back to system python!
```
fxcoudert: This seems like a strong new requirement. Is it upstream's policy that macOS system-provided python cannot work?

tschoonj: No, this is my personal experience speaking 😄 

A lot of gnome projects are migrating to meson as build system, and I noticed that when these projects need to generate gobject-introspection bindings, this fails because the system python adds flags that trigger a universal build of these bindings, which screws with some glib macros.

Formula/glib.rb
tschoonj: sure, I will push a commit to clean things up a bit

tschoonj:
Any objections to this one being merged in? 

The audit warning is bogus as `itstool` has indeed a dependency on `libxml2 --with-python`

fxcoudert:
The dependency on brewed python seems a bit odd to me: we should try to keep those as few as possible, I guess (currently 9 formulas, against 286 with `:python`). But you're probably a better judge than I.

tschoonj:
If Apple would make their python 64-bit only then there wouldn't be a need for this change... but until they do, `gobject-introspection` needs the brewed one.

MikeMcQuaid (approved):

Formula/gobject-introspection.rb
aparajita: You're forgetting that a lot of people use pyenv, which installs a non-universal python which shouldn't cause any problems, correct? If so, then there should at least be a build option to use the current python (or perhaps the path to a specific python binary).

brion:
This seems to have been undone in https://github.com/Homebrew/homebrew-core/pull/12226, causing builds using gobject-introspection to fail again.

PR #11503, opened 2017-03-23
glib: formula cleanup

tschoonj:


tschoonj:
Would anyone mind if I boneyard `gst-plugins-bad@0.10`? No other formula appears to be using it

MikeMcQuaid:
@tschoonj Check the usage numbers in analytics. Shout on Slack if you need help doing that.

tschoonj:
Feel free to merge this one in: the glib bottles don't need an update but the gst-plugins-bad@0.10 ones do.

Thanks

fxcoudert:
29 downloads in one month for `gst-plugins-bad@0.10`, vs 578 for `gst-plugins-bad`.
5% usage means I would support boneyarding.

PR #11509, opened 2017-03-23
pkg-config 0.29.2

tschoonj:


fxcoudert:
In local tests, the line `ENV.append "LDFLAGS", "-framework Foundation -framework Cocoa"` is not necessary anymore.

tschoonj:
Good catch!

PR #8019, opened 2016-12-19
Percona + Symbol not found: _clock_gettime

njt1982:
If Homebrew was updated on Aug 10-11th 2016 and `brew update` always says `Already up-to-date.` you need to run: `cd "$(brew --repo)" && git fetch && git reset --hard origin/master && brew update`.

# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

_You can erase any parts of this template not applicable to your Issue._

### Bug reports:

Getting the seemingly very common `Symbol not found: _clock_gettime` error. 

```
$ brew config
HOMEBREW_VERSION: 1.1.5-20-g0027ded5a
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 0027ded5aa6195dcc153ecf0d3d3b05e99f339af
Last commit: 13 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 7a7a1fe7ff55e3ef5d67c7ffbbc04d034bb35cc0
Core tap last commit: 3 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.11.0 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /Users/nthompson/.rbenv/shims/ruby => /usr/local/Cellar/ruby/2.3.3/bin/ruby
Java: 1.8.0_66
macOS: 10.11.6-x86_64
Xcode: 8.2
CLT: 8.2.0.0.1.1480973914
X11: 2.7.11 => /opt/X11
```

```
$ brew doctor
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
[...]

PR #11522, opened 2017-03-23
percona-server: fix compilation

fxcoudert:
Another solution to #8019, relying on `std_cmake_args`.


MikeMcQuaid (approved):
Don't reckon it needs a `revision` and also reckon we can just use `CMAKE_BUILD_TYPE=Release`

PR #11532, opened 2017-03-24
wireguard-tools 0.0.20170324

zx2c4:
Simple version bump.


neutric:
Thanks so much for your contribution, @zx2c4! 👍

PR #11533, opened 2017-03-24
fix sha256 in xplanetfx

blogabe:

-----

### DIFF
--- a/Formula/xplanetfx.rb
+++ b/Formula/xplanetfx.rb
@@ -3,7 +3,7 @@ class Xplanetfx < Formula
   homepage "http://mein-neues-blog.de/xplanetFX/"
   url "http://repository.mein-neues-blog.de:9000/archive/xplanetfx-2.6.12_all.tar.gz"
   version "2.6.12"
-  sha256 "bae39af674fc89b3fbe07ba6271bbf0a1c2ff64bbf63f4b04ff3d6f0bcc4380c"
+  sha256 "42b84623821032b7f31797d0187e9e0e298282831327fe3fc508dd1d85c493a7"
 
   bottle do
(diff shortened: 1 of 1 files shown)


neutric:
@blogabe Before we can proceed, we'll need to know why the file has changed. Is upstream aware of this issue? If not, would you be so kind and let them know about the checksum change and then report back to us? See [this](https://github.com/Homebrew/homebrew-core/pull/9778) for an example of what we are looking for.

blogabe:
I don't think it was working for some time.  Running `brew audit --strict` on the formula before fixing it didn't identify any errors.  It wasn't until I tried installing it that the alert came up.  My point is, I don't know if it's an upstream issue or if the formula had the wrong sha since the last update maybe 4 months ago.

neutric:
`sha256` was [changed](https://github.com/Homebrew/homebrew-core/commit/e4137bb511da38085b63b6af1ccd356c6c1bece8) during the version bump to 2.6.12 and bottles were built successfully, so what we have now was the correct checksum at some point. Upstream uses a comment form as "bug reporting facility". Would you mind asking real quick whether they are aware of the checksum change and why it changed?

blogabe:
From upstream: "I set up a new key for verification cause newer Ubuntu and derivates were complaining about the old one."

neutric:
[Link to upstream's response](http://mein-neues-blog.de/xplanetFX/#comment-1746) for good record keeping. 🗃 

I guess a version bump by upstream would have communicated this change better, as @ilovezfs [said recently](https://github.com/Homebrew/homebrew-core/pull/10842#issuecomment-285658549), but what are you gonna do. Thanks for getting clarification, @blogabe!

PR #11538, opened 2017-03-24
gobject-introspection: enforce use of brewed python

tschoonj:
This patch ensures that the full path to the brewed python executable is
used in the scripts shebang, instead of using whatever python is in the
PATH


dunn:
👍

ilovezfs:
I'd still like to see a more surgical solution for this.

ilovezfs:
Reverting this in https://github.com/Homebrew/homebrew-core/pull/12226.

PR #11577, opened 2017-03-25
passpie: migrate to boneyard

ilovezfs:
Created with `brew boneyard-formula-pr` because it doesn't work with gnupg 2.1.x. See https://github.com/marcwebbie/passpie/issues/112.


ilovezfs:
Upstream is working on a fix. Yay!

PR #11599, opened 2017-03-26
freetds: fix brew audit warning

razah-dev:

-----
``` brew audit freetds ``` gave the warning: 
``` 
      ln -s ../Formula/freetds.rb freetds@1.00
 ```

This PR resolves that warning.

### DIFF
--- a/Aliases/freetds@1.00
+++ b/Aliases/freetds@1.00
@@ -1,0 +1,1 @@
+../Formula/freetds.rb


fxcoudert:
Can we instead remove `freedts@0.91`? It's not the latest version from pre-1.0 branch, we do not use it as requirement in any other formula, and I do not see any indication on the website that it is still supported.

ilovezfs:
@fxcoudert not at this point. If you check Google Analytics, in the last 30 days the number of installs is

homebrew/versions/freetds091 307
freetds@0.91 235
freetds@0.91 --with-unixodbc 21
homebrew/versions/freetds091 --with-unixodbc 16

Since the current policy is not to remove the ones from versions until they reach zero, break, or have CVEs that aren't patched, it doesn't appear that freedts@0.91 is removable yet.

I suspect the reason for its usage is https://github.com/pymssql/pymssql/issues/432 and that people are finding http://stackoverflow.com/questions/37771434/mac-pip-install-pymssql-error

So maybe if someone can get pymssql working with the current version of freetds, it can be retired if that theory is correct.

ilovezfs (approved):

ilovezfs:
Thanks @raza15

PR #11601, opened 2017-03-26
vips 8.4.6 (new formula)

jcupitt:
vips.rb is currently in homebrew-science.

We've recently done a php binding, but homebrew-php are reluctant to
add a dependency on a non-core formula. They've suggested I should
ask for vips.rb to be moved into homebrew-core:

https://github.com/Homebrew/homebrew-php/pull/3757

What do you think? vips.rb used to be in core, it was moved out to
science a while ago, so this would reverse that.

vips.rb is becoming quite popular as an image processing library for the web,
with packages for Ruby, Python, PHP and node.js. The node one is probably
the highest-profile:

https://github.com/lovell/sharp

Now that vips has moved out of the scientific niche, perhaps core would
be a better home? What do you think?

-----


javian:
I was wondering about the naming of this. The code is on github https://github.com/jcupitt/libvips and it is called libvips there but the release package says vips. Is that intentional ?

bfontaine:
Ping re @javian’s comment above?

jcupitt:
Sorry, I was sure I'd responded on this, perhaps it was the same question coming up somewhere else.

My feeling is that the `vips.rb` name has been used for so long now it would probably cause more confusion to change it. Plus it is used elsewhere: it's `pecl install vips`, for example. On balance, I'd prefer to stick with plain `vips`.

jcupitt:
Oh, the formula installs some command line tools, like `vips`, `vipsthumbnail` and `vipsheader`, not just a library, so that's maybe another argument for plain `vips.rb`.

fxcoudert:
@BrewTestBot test this please

fxcoudert:
There is currently a failing test on all 3 macOS versions: `FAIL: test_python.sh`

jcupitt:
Hopefully it'll work now.

I made some other changes to the formula:

* vips is now using github for the site and hosting releases, so I've updated that
* removed the `revision` line
* restored the libmatio option, I think it must have been dropped accidentally

homebrew-science is on vips-8.5 now -- if this formula is merged, I'll make another PR to update to the latest version.

jcupitt:
Sorry, I'm not sure how to stop rubocop complaining about the duplicate name.

fxcoudert:
Great, thanks @jcupitt for your contribution! Because this is a migration, we need to:
 - have this PR be the same version as that currently in science (8.5)
 - have another PR in science, linked to this one, migrating the formula

jcupitt:
OK, it now installs the latest version, though that means my branch now has a misleading name. Perhaps that doesn't matter.

Unfortunately, due to some other changes, I had to do a new micro release of vips, so this one is actually ahead of science (8.5.5 here, vs. 8.5.4 in science). Does this matter? The changes are very minor (tweaks to the test suite, better handling of truncated png and jpg files).

I can put this on hold and get science/vips updated, or we can do the microbump as part of the migration. Which would be best?

jcupitt:
I've submitted a PR to update homebrew-science/vips to 8.5.5. When that's merged, I'll try a migration PR.

fxcoudert:
Great! I see the homebrew-science PR was merged, so I suggest closing this PR, opening a new one for the migration itself (the new formula), and a migration PR on homebrew-science.

ilovezfs:
@jcupitt happy to review a migration PR. Thanks!

PR #11606, opened 2017-03-26
alexjs 4.0.1 (new formula)

MikeMcQuaid:
This catches insensitive, inconsiderate writing. It’s available through
NPM but we package other things that are and this is sufficiently
useful to non-Javascript folks that it makes sense to add.

### DIFF
--- a/Formula/alexjs.rb
+++ b/Formula/alexjs.rb
@@ -1,0 +1,20 @@
+require "language/node"
+
+class Alexjs < Formula
+  desc "Catch insensitive, inconsiderate writing"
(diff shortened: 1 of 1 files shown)


ilovezfs:
The name is a bit generic. For instance: https://github.com/simonmar/alex

Maybe `alexjs`?

MikeMcQuaid:
@ilovezfs Good call. Renamed.

bfontaine (approved):

wooorm:
Awesome, thanks folks! 🎉

PR #11618, opened 2017-03-26
Remove all versioned unversioned formulae.

### LINKED ISSUE #11279: Some "@" formulas don't have un-@-ed names
Some formulas with "@" in their name don't have a corresponding formula with no "@".  This means that if you see e.g., `play@1.4` and wonder "huh, I wonder what `play` is" and do `brew desc play` or `brew edit play` you will get:
```
Error: No available formula with the name "play"
```
The same is true of two others:
```
Error: No available formula with the name "gst-ffmpeg"
Error: No available formula with the name "hdf"
```
I think this is subpar UX.  There shouldn't be an `x@n.m` unless there is a plain `x`, even if that plain `x` is an alias or a dummy formula or some such construction.

Also, for some reason `brew info play` will return some information, but `brew info [hdf | gst-ffmpeg]` just gives the "No available formula" error.

MikeMcQuaid:
Remove all the versioned formulae that lack a versioned one. These are not needed since the original formulae was deleted.

Fixes #11279.


ilovezfs:
hdf4 is being handled elsewhere:
https://github.com/Homebrew/homebrew-core/pull/11619
https://github.com/Homebrew/homebrew-science/pull/5335
https://github.com/Homebrew/homebrew-versions/pull/1551

MikeMcQuaid:
@ilovezfs This PR was opened before them, though?

ilovezfs:
Not really relevant since those PRs were correct and this one was not.

PR #11619, opened 2017-03-26
hdf@4: migrate to homebrew/science

JCount:
hdf5 and hdf5@1.8 already reside in science;
therefore rename to hdf4 as part of the transition


JCount:
CC @MikeMcQuaid @ilovezfs

ilovezfs (approved):

iltommi:
HDF4 is still alive and many programs still use it.

Would it be possible to resurrect the hdf4 formula? 

I updated the build to the latest version here:
https://github.com/iltommi/homebrew-brews/blob/master/hdf4.rb

If this is not the right place to complain and ask for a miraculous resurrection, could you please point me to the right one?

thanks

MikeMcQuaid:
> Would it be possible to resurrect the hdf4 formula?

@iltommi Not here but you could do so in your own tap.

PR #11637, opened 2017-03-27
gdk-pixbuf 2.36.6

tschoonj:


tschoonj:
Shall we just bite the bullet and revbump everything that depends on gdk-pixbuf?

This would affect packages in science and games too.

ilovezfs:
What's the alternative/counter-argument?

tschoonj:
Well we are missing out on a lot of bugfixes by not updating this formula. 

Basically the current release has the correct libtool version number and ours does not. The only alternative I can think of, which does not require a mass rebuild, is setting the libtool version numbers in 2.36.6 to those found in 2.36.2.

ilovezfs:
It sounds like if the ABI didn't actually change and it's only a matter of the number, we can use the "wrong" number again without a problem.

Formula/gdk-pixbuf.rb
```diff
@@ -38,6 +38,9 @@ def gdk_module_ver
   end
 
   def install
+    # fix libtool versions
+    # https://bugzilla.gnome.org/show_bug.cgi?id=776892
```
fxcoudert: Should we mention explicitly that this needs to be removed for 2.38.0 and later?

tschoonj:
No this one will need to stay in until we are willing to go through the pain of a mass recompilation of all packages that depend on gdk-pixbuf :smile:

fxcoudert:
When they release the 2.38 it will move the `LT_VERSION_INFO` to `3800` and we should be done, no? (since it is greater than `3602`)

tschoonj:
Hmmm, yeah that may be correct indeed. Let's see what happens when we get to that point.

ilovezfs (approved):

PR #11711, opened 2017-03-28
grc: Improved shell integration and caveats

KarlAmort:

-----

- `arc.zsh` was included in the current release and no longer needs the `if build.head?` qualification. The script for fish shell should also be copied to `#{etc}`.

- Expanded the caveats to address all three shells.

### DIFF
--- a/Formula/grc.rb
+++ b/Formula/grc.rb
@@ -21,7 +21,8 @@ class Grc < Formula
     man1.install %w[grc.1 grcat.1]
 
     etc.install "grc.bashrc"
-    etc.install "grc.zsh" if build.head?
+    etc.install "grc.zsh"
+    etc.install "grc.fish"
   end
 
   # Apply the upstream fix from garabik/grc@ddc789bf to preexisting config files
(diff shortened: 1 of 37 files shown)


Formula/grc.rb
```diff
@@ -34,8 +35,17 @@ def post_install
   end
 
   def caveats; <<-EOS.undent
-    New shell sessions will start using GRC after you add this to your profile:
-      . #{etc}/grc.bashrc
+    To use GRC automatically for new shell sessions:
+
+      Bash: add to ~/.profile:
+        . /usr/local/etc/grc.bashrc
```
bfontaine: Use `#{etc}` instead of `/usr/local/etc`. Not all users have Homebrew under `/usr/local`.

Formula/grc.rb
```diff
@@ -34,8 +35,17 @@ def post_install
...
+
+      ZSH: add to ~/.zshrc
+        [[ -s "/etc/grc.zsh" ]] && source /etc/grc.zsh
```
bfontaine: Use `#{etc}` here and below.

bfontaine (changes requested):

Formula/grc.rb
```diff
@@ -34,8 +35,17 @@ def post_install
   end
 
   def caveats; <<-EOS.undent
-    New shell sessions will start using GRC after you add this to your profile:
-      . #{etc}/grc.bashrc
+    To use GRC automatically for new shell sessions:
+
```
MikeMcQuaid: Remove this empty line.

Formula/grc.rb
```diff
@@ -34,8 +35,17 @@ def post_install
...
+      Bash: add to ~/.profile:
+        . /usr/local/etc/grc.bashrc
+
+      ZSH: add to ~/.zshrc
```
MikeMcQuaid: I'm not convinced we need to add these per-shell. You could instead say:
```
New shells will use GRC if you add the relevant file to your profile e.g.:
  . #{etc}/grc.bashrc
```

PR #11751, opened 2017-03-28
mlton: Add --HEAD build target

jez:
After the most recent changes to the mlton formula (see
Homebrew/legacy-homebrew#48694), mlton basically supports building from
HEAD for free.

The only tricky change here is that the patch block is only needed for
the stable (because the patch has been merged already on HEAD).

I don't think the idea behind this change is unwelcome, but if it is I'm open
to suggestions.

/cc @MatthewFluet and @apjanke, who were involved in the previous PR.

### DIFF
--- a/Formula/mlton.rb
+++ b/Formula/mlton.rb
@@ -1,9 +1,19 @@
 class Mlton < Formula
   desc "Whole-program, optimizing compiler for Standard ML"
   homepage "http://mlton.org"
-  url "https://downloads.sourceforge.net/project/mlton/mlton/20130715/mlton-20130715.src.tgz"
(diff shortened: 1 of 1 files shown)


fxcoudert (approved):

fxcoudert:
This looks good to me

jez:
> You'll want to move version into the stable block too.

Alright I was wondering about that. `brew audit --strict mlton` told me to move the `url` and the `sha256` there, but it didn't tell me to move the `version` so I didn't. Maybe you'd like to add this check.

ilovezfs:
Thanks @jez

PR #11809, opened 2017-03-29
delete android-sdk and -ndk dependencies due to migration to caskroom

yurikoles:

-----

>I had tried to put caskroom/android-sdk as dependency, but brew can't resolve it. Is there any official way for formulae to depend on cask?

@MikeMcQuaid maybe we can introduce special case like `:xquartz` os `:osxfuse`?


MikeMcQuaid:
This won't work. You need to just remove all these dependencies and, where not removable, formulae.

yurikoles:
@MikeMcQuaid done.

JCount (dismissed):
Both `crystal-ndk` and `fdroidserver` need revision bumps, since the removed dependencies were `:recommended`.

yurikoles:
@JCount done

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @yurikoles! I was unable to preserve commit attribution, unfortunately, because we need only one formula changed per commit.

PR #11810, opened 2017-03-29
multirust: removed deprecated formula

HarryLovesCode:
Removed multirust as it has been deprecated and is no longer supported.

[Multirust repository](https://github.com/brson/multirust)

- Removed formula
- Removed reference to formula as conflict

---

-----


fxcoudert:
How about providing `rustup`, and migrating `multirust` users to `rustup`? Would that make sense, from a user perspective? I.e. are the two tools sufficiently close that a user could transition without effort?

JCount:
There has been a previous discussion about adding a `rustup` formula to Homebrew, so if you would like to try and migrate `multirust` users the way @fxcoudert suggested, you should read https://github.com/Homebrew/homebrew-core/pull/9617

Even without the addition of a `rustup` formula, this should probably be merged, given `multirust`'s deprecated state.

alex:
Given that `multirust` was deprecated in favor of `rustup`, it seems a bit unfortunate to remove this before the replacement is available.

PR #11815, opened 2017-03-29
minio: Fix to use go1.7 instead of go1.8

harshavardhana:

-----

### DIFF
--- a/Formula/minio.rb
+++ b/Formula/minio.rb
@@ -13,7 +13,10 @@ class Minio < Formula
     sha256 "157af8ad8cde930743a2033c723ba1ecf05079d9eca39bddafdf735321aee6e5" => :yosemite
   end
 
-  depends_on "go" => :build
+  ## Minio depends on go1.7.x only because of the apparent incompatibilities with go1.8
+  ## for now, until the upstream https://github.com/minio/minio/tree/go1.8 is merged
+  ## and a new release is out we don't need to change the below line.
+  depends_on "go@1.7" => :build
 
   def install
     ENV["GOPATH"] = buildpath
@@ -83,6 +86,41 @@ class Minio < Formula
   end
 
   test do
-    system "#{bin}/minio", "version"
+    (testpath/"config.json").write <<-EOS.undent
+{
+        "version": "14",
+        "credential": {
+                "accessKey": "minio",
+                "secretKey": "minio123"
+        },
+        "region": "us-east-1",
+        "browser": "on",
+        "logger": {
+                "console": {
+                        "level": "error",
+                        "enable": true
+                },
+                "file": {
+                        "level": "error",
+                        "enable": false,
+                        "filename": ""
+                }
+        },
+        "notify": {
+                "redis": {
+                        "1": {
+                                "enable": true,
+                                "address": "127.0.0.1:6379",
+                                "password": "",
+                                "key": "minio_events"
+                        }
+                }
+        }
+}
+EOS
+    minio_io = IO.popen("#{bin}/minio --config-dir #{testpath} server #{testpath}/export", :err=>[:child, :out])
+    sleep 1
+    Process.kill("INT", minio_io.pid)
+    assert_match("getsockopt: connection refused", minio_io.read)
   end
 end
--- a/Formula/minio.rb
+++ b/Formula/minio.rb
@@ -5,6 +5,7 @@ class Minio < Formula
     :tag => "RELEASE.2017-03-16T21-50-32Z",
     :revision => "6e7d33df203d0c11d014b6625bfd18aada22328a"
   version "20170316215032"
+  revision 1
 
   bottle do
     cellar :any_skip_relocation


fxcoudert:
Is go 1.8 (or later versions) officially unsupported for minio? Are you targeting a release for latest go support? We would ideally need a comment above the changed line so we know, in the future, when we can move back to `go`.

MikeMcQuaid:
Can you explain what this "fixes"?  Thanks!

harshavardhana:
Go 1.8 has bugs apparently they broke url parsing rules etc. Which doesn't happen in go1.7.5 . Just the tip of the ice berg of issues with go1.8. Minio is not officially supporting go 1.8 at the moment. It was a mistake to update the release without depending on go1.7 package.

harshavardhana:
Yes we will move to go1.8 once we have fully migrated the code to the newer expectations in stdlib. Would it better if it a comment in the formula ?

fxcoudert:
A comment with a link to a PR/bug tracking the process of go 1.8 support would be perfect.

MikeMcQuaid:
> It was a mistake to update the release without depending on go1.7 package.

Releases often will be upgraded by a member of our team. Please add something into the `test do` block that reproduces the bug to stop this happening in future, thanks.

harshavardhana:
> Releases often will be upgraded by a member of our team. Please add something into the test do block that reproduces the bug to stop this happening in future, thanks.

Let me check.

harshavardhana:
> A comment with a link to a PR/bug tracking the process of go 1.8 support would be perfect.

Will add.

harshavardhana:
Added a simple test to look for the url parsing. Is that good @MikeMcQuaid ?

fxcoudert:
Test would require runtime go dependency, which we don't want…

harshavardhana:
> Test would require runtime go dependency, which we don't want…

Are the tests run when the `brew installs` ? i thought this was for only build. @fxcoudert

MikeMcQuaid:
Indeed, this test should be for a `minio` failure rather than a `go` one.

harshavardhana:
Let's see added a test case.. using minio @MikeMcQuaid @fxcoudert - Thanks for the feedback.

harshavardhana:
Looks like it did the trick - can you guys verify?. Thanks

MikeMcQuaid:
@harshavardhana Thanks! To confirm: that test fails with the current version of `go`? If so: 👍 from me.

harshavardhana:
> @harshavardhana Thanks! To confirm: that test fails with the current version of go? If so:  from me.

Let me confirm once again.

harshavardhana:
Confirmed build has been fixed and also added a case where go1.8 would fail @fxcoudert @MikeMcQuaid

ilovezfs (changes requested):
:-1: to depending on `go@1.7`

191 formulae are in core that do not need a specific go version, and then there would be this one. Please let's get this fixed upstream.

harshavardhana:
> 191 formulae are in core that do not need a specific go version, and then there would be this one. Please let's get this fixed upstream.

Completely agree upstream should be fixed. The reason to do this is simple go1.8 broke a bunch of functionality which were working fine before - while that is being fixed upstream?. 

Don't you have packages which rely on specific versions of certain libraries? classic example i see is Java libraries. I certainly do not understand what is the issue here. Also what does this has to do with other 191 packages - i don't see the relation here.

ilovezfs:
We have zero formulae in homebrew/core with `depends_on "go@*"`. I'd like to keep it that way.

harshavardhana:
> We have zero formulae in homebrew/core with depends_on "go@*". I'd like to keep it that way.

@ilovezfs - i am not sure i follow -  i don't see any technical argument or merit is doing so.  I don't see why there shouldn't be exceptions when it comes to not keeping broken bottles for a particular package, migration to go1.8 should also complete by the next release upstream so until then this is the best solution. 

On the same note i would like to point out - 61 different formulas sticking to a particular version, they are not `go@` but it is not uncommon. 

```
$ git grep depends_on | awk {'print $3'} | grep '@'  | wc -l
61
"openssl@1.1"
"boost@1.59"
"boost@1.59"
"openssl@1.1"
"openssl@1.1"
"openssl@1.1"
"gmp@4"
"isl@0.12"
"gmp@4"
"ppl@0.11"
"openssl@1.1"
"imagemagick@6"
"berkeley-db@4"
"openssl@1.1"
"gmp@4"
"libmpc@0.8"
"mpfr@2"
"ppl@0.11"
"cloog@0.15"
"gmp@4"
"libmpc@0.8"
"mpfr@2"
"ppl@0.11"
"cloog@0.15"
"gmp@4"
"libmpc@0.8"
"mpfr@2"
"isl@0.12"
"gmp@4"
"libmpc@0.8"
"mpfr@2"
"isl@0.12"
"isl@0.14"
"qt@5.7"
"qt@5.7"
"gst-plugins-base@0.10"
"gstreamer@0.10"
"gst-plugins-base@0.10"
"gst-plugins-base@0.10"
"gmp@4"
"gmp@4"
"openssl@1.1"
"gmp@4"
"mpfr@2"
"lua@5.1"
"lua@5.1"
"openssl@1.1"
"openssl@1.1"
"openssl@1.1"
"gmp@4"
"openssl@1.1"
"protobuf@3.1"
"qt@5.7"
"openssl@1.1"
"llvm@3.9"
"gmp@4"
"openssl@1.1"
"berkeley-db@4"
"lua@5.1"
"lua@5.1"
"lua@5.1"
```

mantoni:
This was opened because of a bug I reported in Minion. It broke on a new install without any change on their end, but due to braking changes introduced in Go. There is no other way to resolve this kind of problem other than specifying the Go dependency explicitly. It's a compile time dependency, not a runtime thing, and although the version number implies that it's a feature release, it changed some behaviors significantly.

From my (the users) perspective, I don't have Go installed, I just depend on a package that uses the Go compiler, I make a fresh install, things break. @ilovezfs please provide instructions on how you think this issues should be resolved. Fixing it upstream is not an option if breaking changes are introduced.

ilovezfs:
It should be fixed with a go 1.8 compatible release ASAP.

MikeMcQuaid:
> Fixing it upstream is not an option if breaking changes are introduced.

It is, actually, because an upstream patch can then be applied to the formula. Our introduction of versioned formulae into Homebrew/homebrew-core is a relatively new thing and we're still trying to avoid depending on them whenever necessary.

mantoni:
This means a breaking release is enforced on Minio due to changes in Go that they never intended to depend on. Interesting view. I will have to be a lot more careful with brew upgrades in the future then. Thanks for explaining.

ilovezfs:
@mantoni yes, this is why robust `test do` blocks are critical to brew's overall QA.

mantoni:
For my own understanding: What is going to happen if Go 2.0 comes out? As I understand the discussion here, the latest version is used and all depending formulaes might break. Is that correct? The specific issue on Go 1.8 aside, wouldn't it make sense to at least specify `go@1` as the dependency?

fxcoudert:
There is a difference between minor releases, which shouldn't break compatibility (_“Go programs that work today should continue to work even as future "point" releases of Go 1 arise”_), and new major version. When that issue arises, we will deal with it, by testing our Go software and using `go@1` as dependency when that becomes necessary.

ilovezfs:
For 2.0, we'll probably revision bump everything with either `depends_on "go" => :build` or `depends_on "go"` in CI, and see where we stand at that point.

mantoni:
Ok. Thanks for clarifying.

harshavardhana:
> It is, actually, because an upstream patch can then be applied to the formula. Our introduction of versioned formulae into Homebrew/homebrew-core is a relatively new thing and we're still trying to avoid depending on them whenever necessary.

This only works if the patch is simple enough or relatively has marginal changes, go1.8 has some wide changes ranging from rpc behavior, url parsing and other stdlib changes.  It shouldn't be expected for any project to move to something which was just released a month ago and clearly not from a packaging point of view. It is a thumb rule to wait for 1.8.x releases to iron out the issues of GA release. 

Now to the topic of versioning - I see even go1.4 is shipped in brew. If there is no need for an explicit dependency expected from packages why even version them or ship them. go should have been kept just 1.8 IMHO - that way this PR would have never emerged. From what i can see you do want packages to depend on specific versions.

Now regarding dependencies clearly you have allowed for other packages to depend on specific releases especially compiler ones such as `llvm`, `lua` , `clang` etc. so exceptions do exist. 

I presented my ideas here on why this is a bad idea.  In any-case we do have a work-around there is a binary release https://dl.minio.io/server/minio/release/darwin-amd64/

ilovezfs:
Most of the versioned dependencies in core are for _other_ versioned formulae recently imported from homebrew/versions, since we're consolidating all the taps into core. I admire your persistence here, but your efforts would be better directed at getting this working with 1.8 rather than arguing.

ilovezfs:
In particular, excluding old versioned formulae depending on other old versioned formulae, and excluding the long-term openssl 1.1 transition:  
```
$ (cd `brew --repo homebrew/core`;  grep -r depends_on .|grep '\@'|awk -F ':' '{print $1}'|sort|uniq|grep -v '@'|while read f; do brew cat $f|grep '@'|grep -q 'openssl@1.1' || echo $f; done|while read f; do echo "$(basename $f .rb):"; cat $f|grep '@'; done)
Error: No available formula with the name "./Aliases/letsencrypt" 
letsencrypt:
  depends_on "openssl@1.1"
cloog:
  depends_on "gmp@4"
  depends_on "isl@0.12"
      "--with-gmp-prefix=#{Formula["gmp@4"].opt_prefix}",
      "--with-isl-prefix=#{Formula["isl@0.12"].opt_prefix}",
emacs:
  deprecated_option "imagemagick" => "imagemagick@6"
  depends_on "imagemagick@6" => :optional
    if build.with? "imagemagick@6"
        exec #{prefix}/Emacs.app/Contents/MacOS/Emacs "$@"
exim:
  depends_on "berkeley-db@4"
    bdb4 = Formula["berkeley-db@4"]
gnuplot:
    url ":pserver:anonymous:@gnuplot.cvs.sourceforge.net:/cvsroot/gnuplot", :using => :cvs
  deprecated_option "qt" => "with-qt@5.7"
  deprecated_option "with-qt" => "with-qt@5.7"
  deprecated_option "with-qt5" => "with-qt@5.7"
  depends_on "qt@5.7" => :optional
  needs :cxx11 if build.with? "qt@5.7"
    ENV.cxx11 if build.with? "qt@5.7"
    if build.with? "qt@5.7"
gpsbabel:
  depends_on "qt@5.7"
lighttpd:
  option "with-lua@5.1", "Include Lua scripting support for mod_magnet"
  deprecated_option "with-lua51" => "with-lua@5.1"
  depends_on "lua@5.1" => :optional
    args << "--with-lua" if build.with? "lua@5.1"
luabind:
  depends_on "lua@5.1"
    ENV["LUA_PATH"] = Formula["lua@5.1"].opt_prefix
ola:
  depends_on "protobuf@3.1"
    protobuf_pth = Formula["protobuf@3.1"].opt_lib/"python2.7/site-packages/homebrew-protobuf.pth"
pc6001vx:
  depends_on "qt@5.7"
ponyc:
  depends_on "llvm@3.9"
    ENV["LLVM_CONFIG"] = "#{Formula["llvm@3.9"].opt_bin}/llvm-config"
python:
  depends_on "berkeley-db@4" => :optional
      s.gsub! "/usr/include/db4", Formula["berkeley-db@4"].opt_include
radare2:
  depends_on "lua@5.1" # It seems to latch onto Lua@5.1 rather than Lua. Enquire this upstream.
stone-soup:
  depends_on "lua@5.1"
uwsgi:
  deprecated_option "with-lua51" => "with-lua@5.1"
  depends_on "lua@5.1" => :optional
    plugins << "lua" if build.with? "lua@5.1"
```

harshavardhana:
> Most of the versioned dependencies in core are for other versioned formulae recently imported from homebrew/versions, since we're consolidating all the taps into core. I admire your persistence here, but your efforts would be better directed at getting this working with 1.8 rather than arguing.

Looks like we are too late for the party, there is a work-around by providing a link to the binary posted earlier - so we will live with it for now.  On to go1.8.

ilovezfs:
Not late to the party. Versioned formulae only exist for formulae with un-versioned counterparts.

harshavardhana:
Is it possible to remove and deprecate `minio` package present in `homrebrew-core`? I looked through documents couldn't get any detail.   

We build binaries anyway so it seemed like a better idea to https://github.com/minio/homebrew-stable keep our own brew tap.

ilovezfs:
With 583 installs in the last 30 days, that's unlikely something we'd remove.

PR #9928, opened 2017-02-13
sqlite 3.17.0

neutric:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #11818, opened 2017-03-29
pastebinit: use secure protocol in test

vszakats:

-----
also:
- actually verify the uploaded text
- override default username with something neutral

### DIFF
--- a/Formula/pastebinit.rb
+++ b/Formula/pastebinit.rb
@@ -30,7 +30,7 @@ class Pastebinit < Formula
 
   test do
     text = "Hello, world!"
-    # Returned URL is cleartext as of version 1.5
+    # Returned URL is cleartext as of 2017-03
     url = pipe_output("#{bin}/pastebinit -a test -b paste.ubuntu.com", text).chomp
     url.sub! "http:", "https:"
     assert_match text, shell_output("curl " + url)
--- a/Formula/pastebinit.rb
+++ b/Formula/pastebinit.rb
@@ -29,9 +29,10 @@ class Pastebinit < Formula
   end
 
   test do
-    url = pipe_output("#{bin}/pastebinit", "Hello, world!").chomp
-    assert_match "http://pastebin.com/", url
-    # We can't actually fetch the URL to check the paste's success because
-    # pastebin blocks our fetches with curl, probably based on the user agent.
+    text = "Hello, world!"
+    # Returned URL is cleartext as of version 1.5
+    url = pipe_output("#{bin}/pastebinit -a test -b paste.ubuntu.com", text).chomp
+    url.sub! "http:", "https:"
+    assert_match text, shell_output("curl " + url)
   end
 end


neutric (approved):

fxcoudert:
Why does `pastebinit` return a HTTP URL if pastebin has switched to HTTPS? Isn't that a bug in `pastebinit` itself?

vszakats:
It's a bug in the server service, not in `pastebinit`. The other HTTPS option is the debian one, but that one suffers from the same (plus the downloads fails to return the uploaded text.)

vszakats:
`pastebin.com` now does support HTTPS and even returns the correct secure URL, but to use it, `pastebinit` needs to be patched with this:
```diff
--- pastebin.d/pastebin.com.conf-ORI	2017-03-30 09:01:37.000000000 +0000
+++ pastebin.d/pastebin.com.conf	2017-03-30 09:00:06.000000000 +0000
@@ -1,6 +1,7 @@
 [pastebin]
 basename = pastebin.com
-regexp = http://((([a-zA-Z0-9\-_\.]*)(pastebin\.com)))
+regexp = https://((([a-zA-Z0-9\-_\.]*)(pastebin\.com)))
+https = True
 
 [format]
 content = api_paste_code
```

fxcoudert:
I guess this should be the default then, if `pastebin.com` has migrated. Can you file a PR with upstream? If so, I guess we let the test fail for now, given how often they release new versions.

vszakats:
Barrier of entry is too high for me to start dealing with upstream PR, so I'd pass this now.

But I think current PR is still useful and can be adapted for `pastebin.com` easily, once a patch similar to above is applied by upstream (and a new version eventually released).

vszakats:
Well, correction: The current comment says that `pastebin.com` doesn't support automated downloads, like `curl`, so the verification is not possible. I suggest accepting this PR as it solves this problem, too.

fxcoudert:
I'd rather keep a test that fails, to remind us that there is a problem, than "fix" the test to work around the software :)

vszakats:
The current failure is valid and the problem is _our test_. It relies on a value received from an external server and that external server changed its behaviour, hence the failure. This PR fixes this, too.

neutric:
@fxcoudert Why keep a test that fails if the reason for failure as indicated by @vszakats is the value that we hardcoded and not so much the tool under test?

neutric:
@fxcoudert poke :D

vszakats:
The updated comment should have probably read:
```
# URL returned by this service is a cleartext (HTTP) one as of 2017-03-30
```

fxcoudert:
I must be really dense… I still don't understand how a test resorting to `url.sub` is a good test.

If the service requires HTTPS, it should say so. Failure to do so is either a problem with `pastebinit` (which we should fix), or a problem with the online server (in which case we should switch to another one).

vszakats:
`pastebinit` just forwards the result coming back as the answer from the server request. It's not a problem in `pastebinit`. Speaking of `paste.debian.net` and `paste.ubuntu.com` it's _sloppiness_ on the server-side whereas it returns an HTTP URL while it supports HTTPS just fine, but it's not a bug per se, and it's fine to correct this locally using `gsub` till they don't update the server code. Once they update it, the `gsub` will not break anything, though it can then be removed. BTW the `gsub` is only necessary to ensure HTTPS protocol in the verification phase, the test should work fine without it, though data will leak to the internets.

vszakats:
@neutric already approved this change and by now I've deleted the branch. So in case this may be interesting, please shout and I'm going to create another PR (though maybe it can be merged via the web interface as-is?)

vszakats:
Or, another view of this is that `paste.debian.net`'s HTTPS support is accidental (the client cert request certainly isn't reassuring when using a browser), so here's an updated patch that uses `paste.ubuntu.com` which doesn't have this problem:
```diff
--- ../pastebinit.rb	2017-03-29 20:36:34.000000000 +0000
+++ pastebinit.rb	2017-03-31 09:20:41.000000000 +0000
@@ -29,9 +29,10 @@
   end
 
   test do
-    url = pipe_output("#{bin}/pastebinit", "Hello, world!").chomp
-    assert_match "http://pastebin.com/", url
-    # We can't actually fetch the URL to check the paste's success because
-    # pastebin blocks our fetches with curl, probably based on the user agent.
+    text = "Hello, world!"
+    # URL returned by this service is a cleartext (HTTP) one as of 2017-03-30
+    url = pipe_output("#{bin}/pastebinit -a test -b paste.ubuntu.com", text).chomp
+    url.sub! "http:", "https:"
+    assert_match text, shell_output("curl " + url)
   end
 end
```

Test:
```
$ brew test ./pastebinit.rb 
Testing pastebinit
==> Using the sandbox
==> /usr/local/Cellar/pastebinit/1.5/bin/pastebinit -a test -b paste.ubuntu.com
==> curl https://paste.ubuntu.com/24286931/
100  1066    0  1066    0     0   4261      0 --:--:-- --:--:-- --:--:--  4264
```

fxcoudert:
What happens if we don't `gsub`? We're going to download the paste from the HTTP address, and it should work, right?

vszakats:
Yes, it will work without `gsub`. But the point of this patch was to secure communications done in the tests, and this would remove it from the download pass. It's still an improvement, or the download can even be removed altogether since it's missing from current live formula anyways.)

MikeMcQuaid:
> If the service requires HTTPS, it should say so. Failure to do so is either a problem with pastebinit (which we should fix), or a problem with the online server (in which case we should switch to another one).

I agree with this.

vszakats:
Guys, it appears that I completely fail to get through the information that none of these is a problem with `pastebinit`. Anyways let's leave it at that.

MikeMcQuaid:
yeh, it's cool, just wanted to drop in my comments that I missed it too so @fxcoudert isn't stupid 😁

vszakats:
Here's the actual config file for `paste.ubuntu.com`, maybe this helps to understand:
```
[pastebin]
basename = paste.ubuntu.com
regexp = https://paste.ubuntu.com
https = True

[format]
user = poster
content = content
format = syntax

[defaults]
```

neutric:
@vszakats No, you didn't fail. I am with you. 

I approved your changes because I figured that using a non-default server and then validating that the returned url actually contained what we are looking for and is not just a random paste that someone else managed to get through where `pastebinit` might have failed is a good thing. I didn't care for doing the request as HTTPS but was okay with it.

To fix the problem at hand, can we agree on just changing `assert_match "http://pastebin.com/", url` to `assert_match "https://pastebin.com/", url`, which made the test pass for me.

vszakats:
@neutric Yes, that will sync the test with the `pastebin.com` server-side changes and thus fix the current failure.

The goal of this patch however was to secure communication in `test` and implement the download phase which was (according to the existing comment) not workable with the `pastebin.com` service.

fxcoudert:
👍 for actually downloading, but given it's a test I don't see that doing it over HTTPS is so important that it warrants some "patching" of the URL.

neutric:
On second thought: why would you want to download and test the content? If it returns a url, all should be good, no?

vszakats:
It's just restoring the protocol which is proved to be supported under this domain. I agree that it would be the server's job, so it'd be best to inform server-side about it. But because it's unlikely I can do this, a `gsub` is an obvious solution without any downsides (except for the maintenance burden of this single line.)

vszakats:
@neutric Just though it'd be fun to implement what the comment said doesn't work. It can be just dropped as-is.

vszakats:
Either case I think it's enough to verify if the server returned `://pastebin.com` so it will not assume any protocol whatsoever.

neutric:
Until they get paste.bin, that is ;)

neutric:
@vszakats Do you want to restore your branch and do the tiny change that makes all of us happy or the slightly larger one that leaves you a bit sad about doing the unsecured request?

vszakats:
@neutric That's one more in favour of selecting a service explicitly via `-b` option.

What about restoring one that secures the upload using `paste.ubuntu.com`, then does the return value check in a protocol-agnostic way and just omits the download phase?

vszakats:
That would be:
```diff
--- a/pastebinit.rb	2017-03-29 20:36:34.000000000 +0000
+++ b/pastebinit.rb	2017-03-31 10:09:42.000000000 +0000
@@ -29,9 +29,7 @@
   end
 
   test do
-    url = pipe_output("#{bin}/pastebinit", "Hello, world!").chomp
-    assert_match "http://pastebin.com/", url
-    # We can't actually fetch the URL to check the paste's success because
-    # pastebin blocks our fetches with curl, probably based on the user agent.
+    url = pipe_output("#{bin}/pastebinit -a test -b paste.ubuntu.com", "Hello, world!").chomp
+    assert_match "://paste.ubuntu.com/", url
   end
 end
```

neutric:
If minimizing unsecured connections in tests is more important than doing the content validation unsecured, then I guess this is it. It improves over the current situation, so you have my +1.

vszakats:
Thank you @neutric!

(I think it's safe to consider the upload successful if the upload tool returned a valid-looking URL.)

vszakats:
It looks I need to open another PR, as simply force pushing a branch with the same name won't restore the branch in this PR. (if there is a trick for this, pls let me know.)

Here's a new one #11894 anyway.

vszakats:
Thanks to you all for wading through this!

neutric:
Up there where it says
> vszakats deleted the vszakats:patch-1 branch a day ago
should be a `Restore branch` button.

vszakats:
Thanks, I knew about that one, but it was no longer there by now. (not sure what its exact expiration time is)

neutric:
You probably created another `patch-1` branch by now? I still have the button on e.g. #9928 from 6 weeks ago, but the branch name is more unique.

vszakats:
Hm, I'm sure I checked that one before pushing the new branch. But maybe I've made a mistake somewhere. Sadly, it's lost now :( Let's continue in #11894. It's a slightly different patch anyway.

(I can also see the restore button in other previous PRs.)

PR #11821, opened 2017-03-29
Make sourcery not depend on system Xcode

Pr0Ger:

-----

Right now this formula is building sourcery with Swift Package Manager. It will dynamically link binary with Swift runtime (Which is most likely located inside Xcode.app). Unfortunately right now Swift doesn't have ABI compatibility so `sourcery` will most likely fail. This is what happens for current bottled version and Xcode 8.3 (released yesterday) installed 

Also, users who have Xcode 8.3 installed can't build this formula from sources because of this issue https://github.com/jpsim/SourceKitten/issues/348

I changed build step to building with xcodebuild instead of SPM. It will build a dummy *.app with CLI app inside and Xcode will bundle all required frameworks inside app bundle and create a shell script as a shortcut for a binary inside this bundle.


Formula/sourcery.rb
```diff
@@ -13,11 +13,10 @@ class Sourcery < Formula
   depends_on :xcode => ["8.0", :build]
 
   def install
-    ENV.delete("CC")
-    ENV["SDKROOT"] = MacOS.sdk_path
-    system "swift", "build", "-c", "release"
-    bin.install ".build/release/sourcery"
-    lib.install Dir[".build/release/*.dylib"]
```
fxcoudert: This is changing the installed files: with your change, we will not be installing the dylibs anymore, but would install the GUI app. Why?

Pr0Ger: As I said, it's not an actually GUI app, it's a CLI app wrapped inside *.app for tricking Xcode to bundle all required libSwift*.dylib inside an app bundle

This app can be unwrapped from *.app bundle (Sourcery.app/Contents/MacOS/Sourcery => prefix/bin/sourcery; Sourcery.app/Contents/Frameworks/* => prefix/Frameworks/*) and this will work too, but brew will try to link Frameworks dir to /usr/local/Frameworks/ and this can break other formulas (for example, Sourcery is using sourcekitten as a dependency, so SourceKittenFramework.framework will be built by this formula, but installing `sourcekitten` formula separately will build this formula too)

I didn't found the way how to make brew link only binaries from formula (and keep frameworks inside cellar folder only) so I made it this way.

fxcoudert: Install all private files into `libexec` (which Homebrew will not try to put elsewhere) and then symlink what you need into `bin`, `lib`, etc.

MikeMcQuaid: > Install all private files into libexec (which Homebrew will not try to put elsewhere) and then symlink what you need into bin, lib, etc.

Agreed. Also, we'd rather not install a `.app` if at all possible.

Pr0Ger: I tried to make it work without installing *.app bundle, but it will not work in some cases.
Sourcery has support for three types of templates, and two of them require some files bundled inside *.app bundle (https://github.com/krzysztofzablocki/Sourcery/blob/master/Sourcery/Generating/Template/JavaScript/JavaScriptTemplate.swift#L38 https://github.com/krzysztofzablocki/Sourcery/blob/master/Sourcery/Generating/Template/Swift/SwiftTemplate.swift#L123). Using this kind of templates will lead to a crash with unbundled binary (current version built with "swift build -c release" will crash too).
I think having an `.app` is a lesser evil than doesn't work in 2 of 3 cases

MikeMcQuaid: Ok, seems fine then, just consider installing it into `libexec` too.

Pr0Ger: I moved  an `*.app` to `libexec`, do I need to do something else?

ilovezfs:
This has been fixed by https://github.com/Homebrew/homebrew-core/pull/12203 but thank you for the PR nonetheless @Pr0Ger

PR #11468, opened 2017-03-22
migrate android-ndk, -sdk, -platform-tools to caskroom

yurikoles:
According to discussion in #7860.
Please don't merge till https://github.com/caskroom/homebrew-cask/pull/31126.

PR #11840, opened 2017-03-30
fix android-platform-tools migration path

yurikoles:
-----
The corresponding cask has been added in https://github.com/caskroom/homebrew-cask/pull/31501

### DIFF
--- a/tap_migrations.json
+++ b/tap_migrations.json
@@ -1,7 +1,7 @@
 {
   "adobe-air-sdk": "caskroom/cask",
   "android-ndk": "caskroom/cask/android-sdk",
-  "android-platform-tools": "caskroom/cask/android-sdk",
+  "android-platform-tools": "caskroom/cask",
   "android-sdk": "caskroom/cask",
   "aeskeyfind": "homebrew/boneyard",
   "afuse": "homebrew/fuse",


MikeMcQuaid:
Thanks again @yurikoles!

MikeMcQuaid:
Will be great when we have this for `android-ndk` too.

yurikoles:
@MikeMcQuaid it's not approved yet, as well as `crystax-ndk`

gtklocker:
I noticed that `android-platform-tools` from Cask got automatically installed when I ran `brew update`, which is really unexpected behaviour and requires no user intervention. ([output](https://gist.github.com/gtklocker/7a789bdd3b94f7eeb5ec07fd1978e507))

However while I had [manually verified the URL and shasum of the old cask](https://github.com/Homebrew/homebrew-core/pull/11468/files#diff-a8cf25419c1c7d406bf8cdd5f16e4caaL8), the new one has a [different link and no checksum at all](https://github.com/caskroom/homebrew-cask/commit/00ad65b19e6ff5c3cdb01fb3b26833880cdda498#diff-76080a65c896f608f66d37222aeed0bbR3). Maybe [this](https://github.com/caskroom/homebrew-cask/pull/31501) shouldn't have been merged, since the option of versioning and also having a checksum as available?

I think at least asking the user before installing a different formula would make sense.

MikeMcQuaid:
> However while I had manually verified the URL and shasum of the old cask, the new one has a different link and no checksum at all. Maybe this shouldn't have been merged, since the option of versioning and also having a checksum as available?
> 
> I think at least asking the user before installing a different formula would make sense.

It's a migration so we're moving users from something we don't support to something we do. We don't prompt users for actions but take the sensible default for the majority.

PR #11844, opened 2017-03-30
libtecla: test added

lordqwerty:

-----

### DIFF
--- a/Formula/libtecla.rb
+++ b/Formula/libtecla.rb
@@ -18,4 +18,22 @@ class Libtecla < Formula
     system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
     system "make", "install"
   end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <locale.h>
+      #include <libtecla.h>
+
+      int main(int argc, char *argv[]) {
+        GetLine *gl;
+        setlocale(LC_CTYPE, "");
+        gl = new_GetLine(1024, 2048);
+        if (!gl) return 1;
+        return 0;
+      }
+    EOS
+
+    system ENV.cc, "test.c", "-L#{lib}", "-ltecla", "-o", "test"
+    system "./test"
+  end
 end


Formula/libtecla.rb
```diff
@@ -14,8 +14,36 @@ class Libtecla < Formula
   end
 
   def install
-    ENV.deparallelize
-    system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
-    system "make", "install"
+    ENV.deparallelize do
```
fxcoudert: `ENV.deparallelize` was already present, and making it into a block is a no-op.

Formula/libtecla.rb
```diff
@@ -14,8 +14,36 @@ class Libtecla < Formula
...
+      system "./configure", "prefix=#{prefix}", "mandir=#{man}"
```
fxcoudert: This is a traditional `configure`, the options need to be of the form `--prefix`

Formula/libtecla.rb
```diff
@@ -14,8 +14,36 @@ class Libtecla < Formula
...
+      system "make", "install"
+    end
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <stdio.h>
```
fxcoudert: `stdio.h` and `string.h` should not be necessary

fxcoudert (changes requested):
Thanks for your contribution! The test is a very good idea.

fxcoudert:
Thanks, can you squash the commits into one, with commit message `libtecla: add test`

Formula/libtecla.rb
```diff
@@ -18,4 +18,29 @@ def install
...
+        char *line;
+        GetLine *gl;
+
+        setlocale(LC_CTYPE, "");
+
+        gl = new_GetLine(1024, 2048);
+        if(!gl)
+          return 1;
+
+        gl = del_GetLine(gl);
```
maxim-belkin: what value is `gl` assigned to in case of a failure? can you add something similar to lines 35-36 above?

fxcoudert: It always returns NULL, says the doc.

lordqwerty: I don't think this matters. 35 - 36 cover failure enough for test. Line 38 deleting as not doing anything of significance.

maxim-belkin: then just don't assign this to `gl`

lordqwerty:
I've done something horribly wrong... *gulp*

fxcoudert:
You have picked up an unrelated commit.

Formula/libtecla.rb
fxcoudert: It's fine…

lordqwerty:
Sorry for the trouble.

maxim-belkin:
How about making this a tad more compact?
```ruby
  test do
    (testpath/"test.c").write <<-EOS.undent
      #include <locale.h>
      #include <libtecla.h>

      int main(int argc, char *argv[]) {
        GetLine *gl;
        setlocale(LC_CTYPE, "");
        gl = new_GetLine(1024, 2048);
        if (!gl) return 1;
        return 0;
      }
    EOS

    system ENV.cc, "test.c", "-L#{lib}", "-ltecla", "-o", "test"
    system "./test"
  end
```

lordqwerty:
Please let me know if you would like further changes. If not I believe ready to merge.

maxim-belkin:
I think it is worth combining lines 28 and 30

fxcoudert (approved):

fxcoudert:
Thanks @lordqwerty for your contribution!

PR #11855, opened 2017-03-30
thrift@0.9 0.9.3 (new formula)

fernandomora:
* Based on https://github.com/Homebrew/homebrew-core/blob/74063d85f4da579d02e34d8770756273707aec0b/Formula/thrift.rb

-----


fernandomora:
Automated build gives the following error:
```Error: 1 problem in 1 formula
thrift@0.9:
  * Dependency 'bison' may be unnecessary as it is provided by macOS; try to build this formula without it.```

But building the formula without `bison` dependency fails since build process fails with:
```checking for bison version >= 2.5... no
configure: error: Bison version 2.5 or higher must be installed on the system!```

How could tell brew bot to skip that check in the automated build?

**Comment by MikeMcQuaid** [comment]
> How could tell brew bot to skip that check in the automated build?

@fernandomora You can ignore it, we'll merge regardless. Thanks for checking ❤️

**Comment by dunn** [comment]
The test is failing, though:
```
==> /usr/local/Cellar/thrift@0.9/0.9.3/bin/thrift --version
Error: thrift@0.9: failed
<1> expected but was
<0>.
```
(A more thorough test would be nice, too!)

**Comment by fernandomora** [comment]
Sorry, I fixed the test, it should work now
```
$ brew test thrift@0.9
Testing thrift@0.9
==> Using the sandbox
==> /usr/local/Cellar/thrift@0.9/0.9.3/bin/thrift --version
==> /usr/local/Cellar/thrift@0.9/0.9.3/bin/thrift --version
```

**Comment by fernandomora** [comment]
@dunn Regarding the test, I'm following what the old thrift090 formula was doing and what the Thrift documentation suggests: http://thrift-tutorial.readthedocs.io/en/latest/installation.html#verify-installation
I think It should be enough, but I'm open to any suggestion to perform a simple but more thorough test.

File: Formula/thrift@0.9.rb [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+      gem install thrift
+
+    To install PHP extension for e.g. PHP 5.5:
+      brew install homebrew/php/php55-thrift
+  EOS
+  end
+
+  test do
+    assert_match /Thrift/, shell_output("#{bin}/thrift --version", 0)
+    assert_match /0\.9\.3/, shell_output("#{bin}/thrift --version", 0)
```
neutric: Why not combine this to `assert_match "Thrift version #{version}", shell_output("#{bin}/thrift --version", 0)`?

Formula/thrift@0.9.rb
```diff
@@ -0,0 +1,63 @@
...
+  depends_on "bison" => :build
+  depends_on "boost"
+  depends_on "openssl"
+  depends_on "libevent" => :optional
+  depends_on :python => :optional
+
+  def install
+    system "./bootstrap.sh" unless build.stable?
+
+    exclusions = ["--without-ruby", "--disable-tests", "--without-php_extension"]
```
neutric: The above two lines can go as stable is all there is now.

Formula/thrift@0.9.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  depends_on "bison" => :build
+  depends_on "boost"
+  depends_on "openssl"
+  depends_on "libevent" => :optional
+  depends_on :python => :optional
+
+  def install
+    system "./bootstrap.sh" unless build.stable?
+
```
neutric: The above two lines can go as stable is all there is now.

neutric: The above two lines can go as stable is all there is now.

Formula/thrift@0.9.rb
neutric: Why not combine this to `assert_match "Thrift version #{version}", shell_output("#{bin}/thrift --version", 0)`?

neutric (changes requested):

fxcoudert:
We do we need yet another versioned formula for this?

fxcoudert:
I don't think there is a real use case made for a versioned formula here, closing the PR.

PR #11861, opened 2017-03-31
ems-flasher: fix HEAD URL and installation

pulsemod:
The homepage doesn't indicate it, but ems-flasher has a github
repository where it has been receiving commits. I've switched the HEAD
URL to it, as well as written a different install process for it.
Hopefully I've done this in the preferred way.


dunn:
Thanks! Merged as d15b89e8e211ce7d6a6ee8c7cde67bd3c7cdaf7a.

PR #9918, opened 2017-02-13
grafana 4.1.2

bergquist:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #6936, opened 2016-11-15
Remove npm install dependency

torifat:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

POC for https://github.com/Homebrew/brew/pull/1408

PR #11881, opened 2017-03-31
yarn: fix to not use npm to install

SimenB:

-----

This doesn't pass it's test, and I'm unable to use it locally after install. There seems to be some issue  with the symlink. Instead of following it it's just prepended to the current path. So if I try to run yarn from a deep path:

```sh-session
$ ls -la /usr/local/Cellar/yarn/0.21.3/bin
total 16
drwxr-xr-x  4 simen  admin  136 Mar 31 08:21 .
drwxr-xr-x  8 simen  admin  272 Mar 31 08:21 ..
lrwxr-xr-x  1 simen  admin   19 Mar 31 08:21 yarn -> ../libexec/bin/yarn
lrwxr-xr-x  1 simen  admin   22 Mar 31 08:21 yarnpkg -> ../libexec/bin/yarnpkg
$ ls -la /usr/local/Cellar/yarn/0.21.3/libexec/bin
total 32
drwxr-xr-x   7 simen  admin   238 Feb 27 16:36 .
drwxr-xr-x  10 simen  admin   340 Mar 31 08:21 ..
drwxr-xr-x   4 simen  admin   136 Feb 27 16:36 node-gyp-bin
-rwxr-xr-x   1 simen  admin   906 Feb 27 16:29 yarn
-rw-r--r--   1 simen  admin    34 Feb 27 16:29 yarn.cmd
-rwxr-xr-x   1 simen  admin  1874 Feb 27 16:29 yarn.js
-rwxr-xr-x   1 simen  admin    42 Feb 27 16:29 yarnpkg
$ pwd
/Users/simen
$ /usr/local/Cellar/yarn/0.21.3/libexec/bin/yarn --version
0.21.3
$ /usr/local/Cellar/yarn/0.21.3/bin/yarn --version
module.js:472
    throw err;
    ^

Error: Cannot find module '/Users/libexec/bin/yarn.js'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:423:7)
    at startup (bootstrap_node.js:147:9)
    at bootstrap_node.js:538:3
```

Why does this happen? `brew test` fails because of this. Might be that the old npm solution put up some extra magic symlinks, but I can't find them, beyond the fact that old `libexec/bin` also was a symlink.

(this is using the old install)
```sh-session
$ ls -la /usr/local/Cellar/yarn/0.21.3/libexec/bin
total 16
drwxr-xr-x  4 simen  admin  136 Mar 31 08:33 .
drwxr-xr-x  5 simen  admin  170 Mar 31 08:33 ..
lrwxr-xr-x  1 simen  admin   36 Mar 31 08:33 yarn -> ../lib/node_modules/yarn/bin/yarn.js
lrwxr-xr-x  1 simen  admin   36 Mar 31 08:33 yarnpkg -> ../lib/node_modules/yarn/bin/yarn.js
```

/cc @Daniel15

### DIFF
--- a/Formula/yarn.rb
+++ b/Formula/yarn.rb
@@ -1,26 +1,20 @@
-require "language/node"
-
 class Yarn < Formula
   desc "JavaScript package manager"
   homepage "https://yarnpkg.com/"
   url "https://yarnpkg.com/downloads/0.21.3/yarn-v0.21.3.tar.gz"
   sha256 "0946a4d1abc106c131b700cc31e5c3aa5f2205eb3bb9d17411f8115354a97d5d"
+  revision 1
   head "https://github.com/yarnpkg/yarn.git"
 
-  bottle do
-    cellar :any_skip_relocation
-    rebuild 2
-    sha256 "9858ac5a34367b65a20ee828cba84ded38ac1bd9920f6c11c28e5d3a896755cf" => :sierra
-    sha256 "de303f2b4ab5fe273082118825b01c3ef64bb4e109fc62d36eeb37d23dfe7b81" => :el_capitan
-    sha256 "9b0b0e12931a58f5ef2a29372a353f46a65d81de492cf919cb9ece34019cd999" => :yosemite
-  end
+  bottle :unneeded
 
   depends_on "node"
 
   def install
-    system "npm", "install", *Language::Node.std_npm_install_args(libexec)
-    bin.install_symlink Dir["#{libexec}/bin/*"]
-    inreplace "#{libexec}/lib/node_modules/yarn/package.json", '"installationMethod": "tar"', '"installationMethod": "homebrew"'
+    libexec.install Dir["*"]
+    bin.install_symlink "#{libexec}/bin/yarn.js" => "yarn"
+    bin.install_symlink "#{libexec}/bin/yarn.js" => "yarnpkg"
+    inreplace "#{libexec}/package.json", '"installationMethod": "tar"', '"installationMethod": "homebrew"'
   end
 
   test do


SimenB:
Ah, figured it out! The clue was in the last one. The old links pointed directly to the `yarn.js` file, not to the incuded `yarn` and `yarnpkg`. It works now 😄 Sorry about the noise in the OP, but I'll keep it for reference.

SimenB:
This might be partly #6936 /cc @torifat

ilovezfs (changes requested):
This has 42,055 installs in the last 30 days. I'd prefer we do such "experiments" on something a bit more obscure than this.

SimenB:
Not sure how this is "an experiment"? It exchanges an unsupported way of installing yarn with a supported one. Aren't the tests to be trusted?

And the fact that npm is used to install broke the installation from brew for me yesterday since I got a new npm from a newer node. So even though the tarball itself was unchanged, I was unable to use a brew-installed yarn without resetting brew to a place in history with another node version

ilovezfs:
>It exchanges an unsupported way of installing yarn with a supported one.

That could describe a huge number of formulae, so not really convincing.

>And the fact that npm is used to install broke the installation from brew for me yesterday since I got a new npm from a newer node. So even though the tarball itself was unchanged, I was unable to use a brew-installed yarn without resetting brew to a place in history with another node version.

Not sure I understand what that means but it's not really relevant to this specific PR and sounds like something you should open an issue about.

Formula/yarn.rb
```diff
@@ -18,9 +16,10 @@ class Yarn < Formula
   depends_on "node"
 
   def install
-    system "npm", "install", *Language::Node.std_npm_install_args(libexec)
-    bin.install_symlink Dir["#{libexec}/bin/*"]
-    inreplace "#{libexec}/lib/node_modules/yarn/package.json", '"installationMethod": "tar"', '"installationMethod": "homebrew"'
+    libexec.install Dir["*"]
+    bin.install_symlink "#{libexec}/bin/yarn.js" => "yarn"
+    bin.install_symlink "#{libexec}/bin/yarn.js" => "yarnpkg"
```
MikeMcQuaid: Pretty much my only objection here is if they add additional commands that we now need to remember to update/manually add them. Thoughts?

SimenB: Those are extracted from the `bin` object in package.json. Is there some easy way to use something like `jq` to extract it?

https://github.com/yarnpkg/yarn/blob/76c9fc181e819d76d6f0833f57313c646b9b5406/package.json#L82-L85

MikeMcQuaid: You could use Ruby's JSON support to do so but it worries me a little if we get to the point that we're parsing these files rather than letting `npm` or similar do so...

ilovezfs: Note this is handled without any additional fuss by the current formula.

SimenB: That said, I doubt they'll add new ones. `yarn` is the name of the thing, but since some debian distros already have a global executable called `yarnpkg`, that's added as an alias. @Daniel15 can probably comment

MikeMcQuaid: In general @Daniel15 having very strong feelings on this PR either way would be valued input.

Daniel15: I doubt we'll add additional executables other than `yarn` and `yarnpkg` so I think it's fine to just hard-code those two. I don't know enough about Homebrew to have strong feelings either way. 

The executables are in a `bin` directory, could Homebrew just link all the extensionless files in the `bin` directory? https://github.com/yarnpkg/yarn/tree/master/bin

SimenB: @Daniel15 That's what I had at the start, but the `yarn` executable in there was unable to properly follow symlinks (or it followed symlinks it shouldn't, not sure?), so I ended in the situation I describe in the OP

SimenB:
```sh-session
$ readlink /usr/local/bin/yarn
../Cellar/yarn/0.21.3/bin/yarn
$ realpath /usr/local/bin/yarn
/usr/local/Cellar/yarn/0.21.3/libexec/bin/yarn.js
```

Can `realpath` be trusted to exist on the platforms that runs that shellscript?

Although `yarnpkg` just runs `yarn.js`, it doesn't do the stuff `yarn` does. Is that by design?

SimenB:
@MikeMcQuaid Changed the symlinking to read paths and binaries from package.json

ilovezfs:
That seems overly complex given what Daniel said above.

SimenB:
Like I replied there, it doesn't work. That was the whole trouble I had in the OP, since I just symlinked the whole bin folder. In addition to non-working scripts, I also got extra files linked. I can filter those out, or just use a whitelist, but then it isn't really any less complex, IMO 😄

Daniel15:
The code in this PR seems reasonable to me. It's not overly complex.

ilovezfs:
I would hard code the list not parse json.

Daniel15:
The code in this PR seems reasonable to me. It's up to the Homebrew maintainers though :)

Formula/yarn.rb
```diff
@@ -18,9 +16,13 @@ class Yarn < Formula
   depends_on "node"
 
   def install
-    system "npm", "install", *Language::Node.std_npm_install_args(libexec)
-    bin.install_symlink Dir["#{libexec}/bin/*"]
-    inreplace "#{libexec}/lib/node_modules/yarn/package.json", '"installationMethod": "tar"', '"installationMethod": "homebrew"'
+    libexec.install Dir["*"]
+    package_json = JSON.parse File.read("#{libexec}/package.json")
```
ilovezfs: If we're going down this road of parsing these json packaging files ourselves, which I don't think is advisable, it would belong in the brew backend with a DSL, not proliferated throughout formulae.

Daniel15:
Yeah hardcoding `yarn` and `yarnpkg` would be fine too. We do that in Yarn itself for things like the Debian package.

ilovezfs:
>Yeah hardcoding yarn and yarnpkg would be fine too. We do that in Yarn itself for things like the Debian package.

This.

Formula/yarn.rb
SimenB: That I can agree on, I was actually thinking the exact same thing 😄

SimenB:
> I would hard code the list not parse json.

That's what I had. And again, the extensionless files in `/bin` doesn't work, so that would require some work over in Yarn as well.
And the old solution symlinked the symlinks I'm now creating manually.

If it's ok to hardcode to `bin/yarn.js` I can do that? That is what I had before though, see https://github.com/Homebrew/homebrew-core/pull/11881#discussion_r109109418

ilovezfs:
@SimenB Yes, but now we have new information since Daniel said, "I doubt we'll add additional executables other than yarn and yarnpkg so I think it's fine to just hard-code those two."

Daniel15:
@SimenB Here's the Debian package script, for reference: https://github.com/yarnpkg/yarn/blob/fda53e97b07d32982d9ada611588d4baedc7ff51/scripts/build-deb.sh#L57-L59. In the Debian package, we symlink `/usr/bin/yarn` to `/usr/share/yarn/bin/yarn`, where `/usr/share/yarn/bin/yarn` is `bin/yarn` from the Yarn repo/tarball. Couldn't you do the same thing?

SimenB:
Cool, I'll revert the changes I did today 😄 

@Daniel15 No, `readlink` returns a relative path, and while yarnpkg/yarn#2441 fixes it for GNU `readlink`s, on Mac you get `illegal option -- f`. If `yarn` can use `realpath` instead of `readlink` that would fix it, but I'm not sure how viable that is.

Daniel15:
Oh yeah, I always forget that Mac uses BSD tools rather than GNU tools 😛 

If you like, you could modify Yarn's `bin/yarn` shell script to use the right command based on the environment? It already has some hacks for Cygwin/MinGW on Windows.

SimenB:
I can try sending a PR for it sure, but then we'd have to hold off on this PR until a new (stable) release of yarn. No biggie of course 😄

ilovezfs:
>If you like, you could modify Yarn's bin/yarn shell script to use the right command based on the environment? It already has some hacks for Cygwin/MinGW on Windows.

Yeah it would be great to have a BSD friendly shell script. Note that `realpath` itself isn't a command on macOS, but only available as
```
char * realpath(const char *restrict file_name, char *restrict resolved_name);
```
(unless you've installed `coreutils`, but that's a pretty heavy dependency to add, which wouldn't pass the smell test.)

ilovezfs:
>I can try sending a PR for it sure, but then we'd have to hold off on this PR until a new (stable) release of yarn.

Assuming upstream indicated an intent to accept that PR, we could always apply it as a patch here.

SimenB:
We can merge this linking yarn.js and if we can make the `bin/yarn` behave better update what we link?

ilovezfs:
>We can merge this linking yarn.js and if we can make the bin/yarn behave better update what we link?

Yes, whatever works properly without adding unnecessary dependencies! :)

SimenB:
Ok, removed the changes involving package.json

ilovezfs (approved):

ilovezfs:
This seems fine to me as long as it's well tested before it goes in.

SimenB:
I have tested it on my personal Mac and the one I use at work. Is there any more testing I can do?

ilovezfs:
@SimenB You could try it with https://github.com/Homebrew/homebrew-core/pull/11991 which wants to use `yarn` as a dependency, and with `lumo`, which already does https://github.com/Homebrew/homebrew-core/blob/master/Formula/lumo.rb#L17 but won't have been tested by CI since it's build-time only.

SimenB:
Lumo worked fine. Grafana had some issue. I don't think it's the installed yarn's fault, though, as I get the same error with the yarn installed from master.

```
$ brew install grafana
==> Downloading https://homebrew.bintray.com/bottles/grafana-4.2.0.sierra.bottle.tar.gz

curl: (22) The requested URL returned error: 401 Unauthorized
Error: Failed to download resource "grafana"
Download failed: https://homebrew.bintray.com/bottles/grafana-4.2.0.sierra.bottle.tar.gz
Warning: Bottle installation failed: building from source.
==> Using the sandbox
==> Downloading https://github.com/grafana/grafana/archive/v4.2.0.tar.gz
==> Downloading from https://codeload.github.com/grafana/grafana/tar.gz/v4.2.0
==> go run build.go build
==> yarn install
==> npm install grunt-cli --verbose
==> node_modules/grunt-cli/bin/grunt build
Last 15 lines from /Users/simen/Library/Logs/Homebrew/grafana/04.grunt:

05 04 2017 08:25:24.031:ERROR [launcher]: Cannot start PhantomJS

05 04 2017 08:25:24.034:INFO [launcher]: Trying to start PhantomJS again (2/2).
05 04 2017 08:25:24.950:ERROR [phantomjs.launcher]: 2017-04-05 08:25:24.950 phantomjs[88158:459644] -[_NSXPCDistantObjectWithError performScanWithChannels:ssidList:legacyScanSSID:includeHiddenNetworks:mergeScanResults:maxAge:maxMissCount:maxWakeCount:maxAutoJoinCount:interfaceName:waitForWiFi:waitForBluetooth:token:priority:reply:]: unrecognized selector sent to instance 0x7fe033808f40

05 04 2017 08:25:24.951:ERROR [phantomjs.launcher]: PhantomJS has crashed. Please read the bug reporting guide at
<http://phantomjs.org/bug-reporting.html> and file a bug report.

05 04 2017 08:25:24.953:ERROR [launcher]: Cannot start PhantomJS

05 04 2017 08:25:24.955:ERROR [launcher]: PhantomJS failed 2 times (cannot start). Giving up.
Warning: Task "karma:test" failed. Use --force to continue.

Aborted due to warnings.

READ THIS: http://docs.brew.sh/Troubleshooting.html

These open issues may also help:
Grafana v4.2.0 https://github.com/Homebrew/homebrew-core/pull/11991
```

ilovezfs:
@SimenB cool. Yeah grafana needs that `--force`. See https://github.com/Homebrew/homebrew-core/pull/9918/files#diff-7473b50efa5c9abb0b107de1cb79c8f7R37

ilovezfs:
@SimenB I think you can also make this ` bottle :unneeded` now.

SimenB:
What does `bottle :unneeded` do?

Building grafana with this diff from #11991 

```diff
diff --git i/Formula/grafana.rb w/Formula/grafana.rb
index 49b03993f..0e15a2d14 100644
--- i/Formula/grafana.rb
+++ w/Formula/grafana.rb
@@ -27,7 +27,10 @@ class Grafana < Formula
       system "go", "run", "build.go", "build"
       system "yarn", "install"
       system "npm", "install", "grunt-cli", *Language::Node.local_npm_install_args
-      system "node_modules/grunt-cli/bin/grunt", "build"
+      args = ["build"]
+      # Avoid PhantomJS error "unrecognized selector sent to instance"
+      args << "--force" unless build.bottle?
+      system "node_modules/grunt-cli/bin/grunt", *args
 
       bin.install "bin/grafana-cli"
       bin.install "bin/grafana-server"
```

ilovezfs:
>What does bottle :unneeded do?

It's a replacement for the bottle block, which you'll delete as well.

The users will `brew install yarn`, triggering a download of https://yarnpkg.com/downloads/0.21.3/yarn-v0.21.3.tar.gz instead of a Bintray bottle, and then it will execute your install method.

SimenB:
Grafana passed with `--force`. And added `bottle :unneeded`, thanks!

ilovezfs:
>Grafana passed with --force.

Great!

Do you mind opening an issue in the yarn issue tracker about the script issue with BSD userland?

Formula/yarn.rb
```diff
@@ -1,26 +1,19 @@
-require "language/node"
-
 class Yarn < Formula
   desc "JavaScript package manager"
   homepage "https://yarnpkg.com/"
   url "https://yarnpkg.com/downloads/0.21.3/yarn-v0.21.3.tar.gz"
```
ilovezfs: Since it's `bottle :unneeded` now, please also add
```
mirror "https://github.com/yarnpkg/yarn/releases/download/v0.21.3/yarn-v0.21.3.tar.gz"
```

Daniel15: What does `mirror` do? The Yarn download URL just redirects to the GitHub one (we use the yarnpkg.com URL so we're not tightly-coupled with Github's release system)

ilovezfs: Plot foiled! In that case it would only help if yarnpkg.com went down.

I guess we could use `http://distfiles.gentoo.org/distfiles/yarn-v0.21.3.tar.gz` or be mirror-less.

ilovezfs:
@Daniel15 any changes you'd like to see here before we :ship:?

Formula/yarn.rb
```diff
@@ -1,26 +1,19 @@
...
   sha256 "0946a4d1abc106c131b700cc31e5c3aa5f2205eb3bb9d17411f8115354a97d5d"
```
ilovezfs: @SimenB let's add
```
revision 1
```

Formula/yarn.rb
Daniel15: You could keep the mirror just in case `yarnpkg.com` goes down, but the Windows Chocolatey package also uses yarnpkg.com and I haven't seen any reports of being unable to hit yarnpkg.com.

Daniel15:
@ilovezfs It seems reasonable to me, but as I mentioned earlier I'm not very familiar with Homebrew (I used it back when I had a Mac, but I switched back to a PC around a year ago).

By the way, do you really love ZFS? 😄

ilovezfs:
>but I switched back to a PC around a year ago

Heresy!

>By the way, do you really love ZFS?

Yes: https://openzfsonosx.org/

MikeMcQuaid:
@Daniel15 @SimenB I'm not happy with the current confidence level of "I doubt we'll add additional executables other than yarn and yarnpkg so I think it's fine to just hard-code those two." personally. 

What will happen for updates of Yarn is that someone will submit a PR to bump to the next version whenever it comes out. If Yarn ever does include additional executables, manpages, etc. then someone will need to notice and update the formula and until then Yarn and Homebrew both end up with a substandard situation.

This is why for similar e.g. C/C++ projects we strongly recommend having and using a `make install` so that upstream is control of what files get installed.

Thoughts?

MikeMcQuaid:
> I would think that if other binaries are added, and the Debian script needs to be updated, someone™️ would remember to also make sure the linked binaries are correct.

I also just saw the Debian script. I think ideally there'd be a Unix install script (which the Debian one could call, even) which installed to a given `PREFIX`. If this logic lives in `yarn`'s repo: I have no issue with it 👍 

> I did have a solution that checked the bin entry in package.json, but we concluded that it was a bit too complex for little gain

Yeh, I appreciate you investigating that 🙇.

ilovezfs:
>I think ideally there'd be a Unix install script (which the Debian one could call, even) which installed to a given PREFIX.

@Daniel15 that sure would be nice! :innocent:

Formula/yarn.rb
ilovezfs: @SimenB let's add
```
revision 1
```

ilovezfs (changes requested):

Formula/yarn.rb
SimenB: Done. What about `mirror`?

ilovezfs (approved):

Formula/yarn.rb
ilovezfs: I think we can go without for now since it's unlikely yarnpkg.com won't resolve

ilovezfs: You'll need to move `revision` above `head` or it won't pass `brew audit --strict`

SimenB: I'm currently getting this:
```
$ brew audit --strict yarn
Error: Error running `rubocop --format json --force-exclusion --config /usr/local/Homebrew/Library/.rubocop.yml /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/yarn.rb`
```

What am I doping wrong?

ilovezfs: You can probably fix that by blowing away your `~/.gem` directory

MikeMcQuaid (approved):

ilovezfs:
Shipped! :rocket:

Thanks @SimenB for the PR and @Daniel15 for the help.

PR #11903, opened 2017-03-31
mpv: add options to re-enable optical media support

gtm87x:
Since mpv-player/mpv@77cbb35 optical media libs are disabled by default, so explicit flags needs to be passed to the configure script to build dvd/bd support.


dunn:
Good catch, thanks!

PR #11793, opened 2017-03-29
wine: include mpg123

fxcoudert:
Trying to fix #11592

PR #11911, opened 2017-03-31
wine 2.5 (devel)

uyjulian:


dunn:
cc @woodruffw 
```
wine:
  * Use ruby-macho instead of calling lipo
  * Use ruby-macho instead of calling lipo
  * Use ruby-macho instead of calling lipo

woodruffw:
#11793 is replacing `lipo` in Wine with the correct ruby-macho calls, so it's safe to ignore this audit warnings and merge anyways or wait until the aforementioned is merged and then rebase to silence them.

dunn:
Merged as 7e32b582cc158656a0df139686b05201715f8676. Thanks @uyjulian!

PR #11973, opened 2017-04-02
stdman 2017-02-14

stek29:
Also versioning was changed, because it makes more sense to sync versions with cpprefrence archive updates rather than with repo's tags because author of original repo doesn't seem to care much about updates.

And url:revision is now used instead of url:sha, because there are no releases, see first line.
Minor edit for description, because it now has pages for C++17 too (try `man std::variant`)

-----

-----

### DIFF
--- a/Formula/stdman.rb
+++ b/Formula/stdman.rb
@@ -1,8 +1,11 @@
 class Stdman < Formula
-  desc "Formatted C++11/14 stdlib man pages from cppreference.com"
+  desc "Formatted C++11/14/17 stdlib man pages from cppreference.com"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> Also versioning was changed, because it makes more sense to sync versions with cpprefrence archive updates rather than with repo's tags because author of original repo doesn't seem to care much about updates.

Sorry, we need the author to tag releases, we're not willing to just declare new releases based on dates. Please submit an issue upstream asking if they'll do that. Thanks!

stek29:
@MikeMcQuaid Sure.

PR #11979, opened 2017-04-03
libgosu 0.11.3.1

pmer:


ilovezfs (changes requested):
I noticed this tag was created, but it's not designated as the latest release. That is still 0.11.3. Upstream will need to designate the latest release as 0.11.3.1 before we merge this.

See https://github.com/gosu/gosu/releases/latest

ilovezfs:
I've asked upstream what's up here https://github.com/gosu/gosu/issues/396

jlnr:
0.11.3.1 is a Ruby-only hotfix release. On the C++ side, nothing has changed, and a formula update is not necessary.

I will update gosu's release list to clear that up too. Sorry for the confusion.

ilovezfs:
Thanks for the clarification @jlnr

ilovezfs:
@pmer thanks for this PR nonetheless.

PR #11999, opened 2017-04-03
Import homebrew/dupes into core, part 1

fxcoudert:
Splitting PR #11282 to get it through CI…


fxcoudert:
3 style audits that do not really apply, and `crosstool-ng` test failure is not related (because we only change an optional dependency). @MikeMcQuaid what say you?

MikeMcQuaid:
If the test is failing it's probably better to kill it rather than leave it broken. That can be done on pull, though. 👍 to 🚢

fxcoudert:
Do we wait until all parts are ready to ship, and merge the dupes export at the same time?

PR #12033, opened 2017-04-04
lapack: import from homebrew/dupes

fxcoudert:
- Vendoring cmake 3.6.3 because lapack won't build with 3.7 or later
- Removing `with-doxygen` option: never used, requires special `doxygen` llvm build

### DIFF
--- a/Formula/lapack.rb
+++ b/Formula/lapack.rb
@@ -1,0 +1,57 @@
+class Lapack < Formula
+  desc "Linear Algebra PACKage"
+  homepage "http://www.netlib.org/lapack/"
+  url "http://www.netlib.org/lapack/lapack-3.7.0.tgz"
+  sha256 "ed967e4307e986474ab02eb810eed1d1adc73f5e1e3bc78fb009f6fe766db3be"
+  head "https://github.com/Reference-LAPACK/lapack.git"
+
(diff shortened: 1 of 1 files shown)


fxcoudert:
Test links fine, but fails to run with `dyld: Library not loaded: @rpath/liblapacke.dylib`

MikeMcQuaid (approved):

ilovezfs:
I think we may want to modify that rpath to be the opt path since this is keg_only. Really learning to regret https://github.com/Homebrew/brew/pull/2036.

ilovezfs:
CC @woodruffw for thoughts.

woodruffw:
> I think we may want to modify that rpath to be the opt path since this is keg_only.

Agreed, and I think https://github.com/Homebrew/brew/pull/2036 is going to continue to give us pain. I can make a PR to undo it tonight.

MikeMcQuaid:
> Agreed, and I think Homebrew/brew#2036 is going to continue to give us pain. I can make a PR to undo it tonight.

Done in https://github.com/Homebrew/brew/pull/2450

MikeMcQuaid (approved):

ilovezfs (approved):

PR #12060, opened 2017-04-04
giflib 4->5 migration, part 1

fxcoudert:
Creating a new `giflib@4` formula for `swftools`, which depends on that version.

### DIFF
--- a/Formula/giflib@4.rb
+++ b/Formula/giflib@4.rb
@@ -1,0 +1,20 @@
+class GiflibAT4 < Formula
+  desc "GIF library using patented LZW algorithm"
+  homepage "https://giflib.sourceforge.io/"
+  url "https://downloads.sourceforge.net/project/giflib/giflib-4.x/giflib-4.2.3.tar.bz2"
+  sha256 "0ac8d56726f77c8bc9648c93bbb4d6185d32b15ba7bdb702415990f96f3cb766"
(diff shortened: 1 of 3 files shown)


fxcoudert:
@BrewTestBot test this please

fxcoudert:
Looking good to me.

ilovezfs:
We may be able to avoid adding `giflib@4` altogether. What're the analytics on those two? Have you tried to fix their issues with giflib 5? Were those issues with giflib 5 reported upstream? Has upstream responded? etc.

ilovezfs (changes requested):
Let's not add a new versioned formula we're not already providing and no one has asked for unless we have to and it meets the guidelines for versioned formulae.

fxcoudert:
We are already providing giflib 4 anyway (see #12061), the goal here is to change which is the default version and which is the versioned formula.

- `swftools` and `pdf2htmlex` both get decent numbers of installs, so we can't just throw them out.
- Both appear to have very slow development.
- Giflib 5 support was [added to `swftools` in 2013](http://lists.nongnu.org/archive/html/swftools-common/2013-06/msg00004.html) but there was no release since then.
- I have reported the issue to `pdf2htmlex`: https://github.com/coolwanglu/pdf2htmlEX/issues/713

PR #12061, opened 2017-04-04
giflib 4->5 migration, part 2

fxcoudert:
Goes after #12060


fxcoudert:
`pdf2htmlex` linkage and test failures will not be a problem after #12060 is committed.
`mapserver` is preexisting :(

Formula/giflib.rb
```diff
@@ -1,42 +1,22 @@
...
+  desc "Library and utilities for processing GIFs"
   homepage "https://giflib.sourceforge.io/"
-  url "https://downloads.sourceforge.net/project/giflib/giflib-4.x/giflib-4.2.3.tar.bz2"
-  sha256 "0ac8d56726f77c8bc9648c93bbb4d6185d32b15ba7bdb702415990f96f3cb766"
+  url "https://downloads.sourceforge.net/project/giflib/giflib-5.1.4.tar.bz2"
+  sha256 "df27ec3ff24671f80b29e6ab1c4971059c14ac3db95406884fc26574631ba8d5"
 
   bottle do
     cellar :any
-    rebuild 1
```
ilovezfs: You don't need to manually modify the bottle block.

Formula/xplanet.rb
```diff
@@ -44,6 +45,9 @@ class Xplanet < Formula
     sha256 "3f95ba8d5886703afffdd61ac2a0cd147f8d659650e291979f26130d81b18433"
   end
 
+  # Support for giflib 5 is incomplete
```
ilovezfs: Was this reported?

Also, a patch of that size needs to go in formula-patches.

formula_renames.json
```diff
@@ -48,6 +48,7 @@
   "gcc@6": "gcc",
   "geode": "apache-geode",
   "giflib5": "giflib@5",
```
ilovezfs: I don't think chaining works.

ilovezfs:
>mapserver is preexisting

It has a Sierra bottle, so that sounds fishy.

formula_renames.json
JCount: it does

fxcoudert:
`mapserver` has a Sierra bottle, but it fails to build locally for me.

PR #12062, opened 2017-04-04
micro: fixing formula to work correctly when reinstalling from HEAD

samdmarshall:

-----

when attempting to run `brew reinstall micro` to reinstall the current version of `HEAD`, the existing brew formula fails to build due to changes in dependencies and not correctly updating the runtime.go file to account for asset changes.

Original Install Log: https://gist.github.com/samdmarshall/853bc356e5b402345c7f441fb199be75
Fixed Install Log: https://gist.github.com/samdmarshall/b966b5944d90334b7e291b551b922b30

The core problem here is that the `quick-build` target in the makefile does not cause `go` to check the dependency graph, which was fixed by changing `quick-build` to `build`. This presented an additional problem of not having built the runtime.go file out correctly, so an additional target (`runtime`) as added prior to `build` to fix this issue.

### DIFF
--- a/Formula/micro.rb
+++ b/Formula/micro.rb
@@ -102,7 +102,7 @@ class Micro < Formula
     mkdir_p buildpath/"src/github.com/zyedidia"
     ln_s buildpath, buildpath/"src/github.com/zyedidia/micro"
     Language::Go.stage_deps resources, buildpath/"src"
-    system "make", "build-quick"
+    system "make", "runtime", "build"
     bin.install "micro"
   end
 


JCount:
@samdmarshall This will have no effect on building the stable version of the formula at different times? Essentially, this really only effects the output if the HEAD spec is being built? If this is not the case, we probably can, and should, see about integrating this change in a different way.

DomT4:
> when attempting to run brew reinstall micro to reinstall the current version of HEAD, the existing brew formula fails to build due to changes in dependencies

And this is why `go` projects should move completely over to the `vendor` system introduced way back in Go 1.5 IIRC. Homebrew cannot be expected to sensibly handle formulae that use `go`, have both `stable` and `head` specs and require `go_resources` IMO.

ilovezfs:
HEAD specs for go formulae without vendoring are fishy. I suppose we could just use `go get` for them though that points to the question whether they're worth having at all.

DomT4:
I used to recommend just killing `HEAD` if they don't vendor. Never found it worth the hassle, for anyone involved; it's a bad deal in terms of user experience, chewing up maintainer time to fix & causing unnecessary bug reports upstream.

samdmarshall:
@JCount Correct, however if the stable version is not installed from a bottle then it will break as it currently doesn't fetch the dependencies prior to building. This fixes that eventuality as well as it not being able to build from HEAD at all.

As it stands, this project is not using a dependency manager to handle the installation of dependencies, that is all entirely done within the Makefile.

ilovezfs:
@samdmarshall let's remove HEAD here. If upstream later switches to using one of the go vendor systems, then we can always consider re-adding HEAD.

samdmarshall:
@ilovezfs No, as previously mentioned this still breaks non-bottles so removing HEAD wouldn't do anything and that is actively used by users of this project.

JCount:
The current `stable` spec builds and reinstalls fine from source.

samdmarshall:
@JCount But it once after the next release happens because of the change in dependencies that i mentioned 🙃

ilovezfs:
@samdmarshall the go resource blocks get updated at each release. That is the reason that using one of the vendor systems is preferred, since it means we don't have to do that each time.

samdmarshall:
which currently requires someone to go and write new blocks for each release, whereas the fix i provide eliminates that additional maintenance. I don't see why this is such a contentious fix for a formula that wasn't written correctly to start with.

ilovezfs:
@samdmarshall it is specifically our practice not to use `go get`, which is what you're describing as the correct way. This formula is written as I requested it to be here https://github.com/Homebrew/homebrew-core/pull/9840#issuecomment-278995836

samdmarshall:
@ilovezfs surprisingly, that doesn't make the make-file targets used any less wrong regardless of who was the one that asked for them to be written that way, that isn't how the project is managed.

JCount:
Have you actually looked at the link?

samdmarshall:
@JCount yes, I have. As mentioned in my original post; there are two problems here:

1. Building from `HEAD` fails because of the way that the formula trying to manage the dependencies of the project by duplicating the work of the makefile. This works for stable builds so long as the formula is updated correctly, by hand. Using the project's preferred method of fetching dependencies it needs would resolve this.

2. Regardless of building from stable or from `HEAD`, the build from source can still break if the `runtime.go` file that is stored in the repo isn't correctly up to date; to protect against this the additional `runtime` target is run when the project is being built from source.

As I understand it, the request was made to change `build` to `build-quick` in the linked thread, as it did a tiny bit of extra work where `go get` would validate and then not do anything as all of the dependencies fetched by `go_resources` were already satisfied. I am saying that this was incorrect as it doesn't safely handle the project's dependencies when building from source and assumes that the formula is current and correct. This makes the proposed change a preventative fix to stop making this formula be as fragile for the sake of a few additional CPU cycles where `go get` sees we already have the on-disk dependencies we want to build with.

Now, it seems that the ideal solution is to have the project (`micro`) use a go dependency manager or somehow vendor the dependencies into the source tree to stop this being necessary. However the project hasn't decided on doing that yet, and it isn't planned for being solved in the short term as far as I am aware.

ilovezfs:
@samdmarshall the reason for not using `go get` in the stable specs of go formulae is to ensure they do not pull tip of tree revisions when users do from source builds, and instead have pinned the exact revisions that were current at the time of the tag.

The use of the make file targets is correct if the dependencies have already been fetched, which is what the go resource blocks do in combination with the `Language::Go.stage_deps resources, buildpath/"src"` in the install method.

If you'd like to change the install method to be

```
  def install
    ...
    if build.head?
      system "make", "runtime", "build"
    else
      Language::Go.stage_deps resources, buildpath/"src"
      system "make", "build-quick"
    end
    ...
```
I'm fine with that as an alternative to removing the HEAD spec.

If the `runtime.go` that is checked into source control at the time of a tag is incorrect, that would be an upstream issue that we'd address by creating an issue on GitHub.

If the go resource blocks aren't properly updated, or aren't updated at all, when we update the formula in the future for new releases, that would be a mistake on the part of one of the maintainers, and something we'd address by updating the relevant resource blocks, and bumping the formula `revision`.

samdmarshall:
@ilovezfs I think that would be an appropriate fix here, thank you. I guess I don't fully understand the homebrew codepaths in regards to the management of go dependencies when building from source; this PR was made in the spirit of trying to take burden off y'all from getting issues with a formula breaking in the (near) future.

ilovezfs:
Fixed in https://github.com/Homebrew/homebrew-core/pull/12368.

PR #12063, opened 2017-04-04
packer 1.0.0

ed-flanagan:

-----

### DIFF
--- a/Formula/packer.rb
+++ b/Formula/packer.rb
@@ -4,8 +4,8 @@ class Packer < Formula
   desc "Tool for creating identical machine images for multiple platforms"
   homepage "https://packer.io"
   url "https://github.com/mitchellh/packer.git",
-      :tag => "v0.12.3",
-      :revision => "c0f2b4768d7f9265adb1aa3c91a0206ca707d91e"
+      :tag => "v1.0.0",
+      :revision => "e3d65a193e96c7b228589fc9699e39998ec12e4e"
   head "https://github.com/mitchellh/packer.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


fxcoudert:
Shouldn't `golang.org/x/tools` be updated?

yurikoles:
Please update urls according to move of upstream to haschicorp: https://github.com/hashicorp/packer/commit/81522dced0b25084a824e79efda02483b12dc7cd

Formula/packer.rb
```diff
@@ -4,8 +4,8 @@ class Packer < Formula
   desc "Tool for creating identical machine images for multiple platforms"
   homepage "https://packer.io"
   url "https://github.com/mitchellh/packer.git",
```
ttrifonov: As pointed in the comments, this line should probably become `url "https://github.com/hashicorp/packer.git"`

Formula/packer.rb
```diff
@@ -4,8 +4,8 @@ class Packer < Formula
...
-      :tag => "v0.12.3",
-      :revision => "c0f2b4768d7f9265adb1aa3c91a0206ca707d91e"
+      :tag => "v1.0.0",
+      :revision => "e3d65a193e96c7b228589fc9699e39998ec12e4e"
   head "https://github.com/mitchellh/packer.git"
```
ttrifonov: Ditto.

ed-flanagan:
Sorry for the late response.
It looks like the v1.0.0 release was built before it moved to the Hashicorp organization.
So while the URLs to download work, build through the Makefile will fail.

There's a [commit](https://github.com/hashicorp/packer/commit/81522dced0b25084a824e79efda02483b12dc7cd) which does the renaming, however it does not patch cleanly (based off commits ahead of the release).

For now it might be best to just keep the old "mitchellh" URLs, or instead of building from source, just download the pre-built bin from https://www.packer.io/downloads.html

In the mean time, I can updated the `go_resource` revisions as @fxcoudert suggested

yurikoles:
>@ed-flanagan pushed 250 commits.

please rebase your branch, you seems to merge commits from master here.

cancan101:
would love to see this merged!

fxcoudert:
Thanks @ed-flanagan for the pull request!

PR #12069, opened 2017-04-05
flif 0.3 (new formula)

stek29:
It currently has two issues: 
- It has a patch
Upstream Makefile  uses `install -s`, which calls `strip`, which fails, while `strip -x` works fine.
I'm not sure what's the right way to solve this. I looked through install(1), but haven't found anything. If the way I did it is correct, I'd fill an upstream PR.

- Tests are missing
FLIF already has tests in Makefile. I'm not sure what's the proper way to "port" them, because they need some files in repo's `test` dir.
Should I add a resource for them (which would be the same as formula's resource), patch makefile so it doesn't rebuild everything before testing, and then call `make test`?

-----

-----

### DIFF
--- a/Formula/flif.rb
+++ b/Formula/flif.rb
@@ -1,0 +1,32 @@
+class Flif < Formula
+  desc "Free Loseless Image Format"
+  homepage "http://flif.info/"
+  # When updating, please check if FLIF switched to CMake yet
+  url "https://github.com/FLIF-hub/FLIF/archive/v0.3.tar.gz"
+  sha256 "aa02a62974d78f8109cff21ecb6d805f1d23b05b2db7189cfdf1f0d97ff89498"
+  head "https://github.com/FLIF-hub/FLIF.git"
+
+  depends_on "pkg-config" => :build
+  depends_on "libpng"
+  depends_on "sdl2"
+
+  resource "test_c" do
+    url "https://raw.githubusercontent.com/FLIF-hub/FLIF/dcc2011/tools/test.c"
+    sha256 "a20b625ba0efdb09ad21a8c1c9844f686f636656f0e9bd6c24ad441375223afe"
+  end
+
+  def install
+    system "make", "PREFIX=#{prefix}", "install", "install-dev"
+    doc.install "doc/flif.pdf"
+  end
+
+  test do
+    testpath.install resource("test_c")
+    system ENV.cc, "test.c", "-I#{include}", "-L#{lib}", "-lflif", "-o", "test"
+    system "./test", "dummy.flif"
+    system bin/"flif", "-i", "dummy.flif"
+    system bin/"flif", "-I", test_fixtures("test.png"), "test.flif"
+    system bin/"flif", "-d", "test.flif", "test.png"
+    assert_predicate testpath/"test.png", :exist?, "Failed to decode test.flif"
+  end
+end


stek29:
So, does anyone have any suggestions? Is `install -s` a known problem? Maybe in worked on older OS X versions, or something.

bfontaine:
> FLIF already has tests in Makefile. I'm not sure what's the proper way to "port" them, because they need some files in repo's `test` dir.
> Should I add a resource for them (which would be the same as formula's resource), patch makefile so it doesn't rebuild everything before testing, and then call `make test`?

Are those unit tests or integration tests? We only care about the latter.

Formula/flif.rb
```diff
@@ -0,0 +1,57 @@
+class Flif < Formula
+  desc "Formatted C++11/14/17 stdlib man pages from cppreference.com"
```
MikeMcQuaid: `"Free Lossless Image Format"`

Formula/flif.rb
```diff
@@ -0,0 +1,57 @@
...
+  depends_on "sdl2"
+
+  def install
+    make_args = %W[
+      PREFIX=#{prefix}
+    ]
+    cd "src" do
+      system "make", *make_args
+      system "make", "install", *make_args
+    end
```
bfontaine: You should be able to strip all of this down to:
```ruby
def install
  system "make", "-C", "src", "install", "PREFIX=#{prefix}"
end
```

bfontaine (changes requested):

MikeMcQuaid:
> FLIF already has tests in Makefile. I'm not sure what's the proper way to "port" them, because they need some files in repo's test dir.

Check out `jpeg` or `libpng` for similar libraries and emulate their test blocks.

Formula/flif.rb
bfontaine: Or even `system "make", "install", "PREFIX=#{prefix}"` since the top-level `Makefile` uses the one in `src` anyway: https://github.com/FLIF-hub/FLIF/blob/fe2c1558a7245d9efaafd082f6237e9e06b72e2d/Makefile#L7-L8

Formula/flif.rb
stek29: Oh lol.

stek29:
@MikeMcQuaid Not sure if I should add a test for `flif` executable too.

MikeMcQuaid:
@stek29 Could do that instead or as well, up to you.

stek29:
Done.

bfontaine (approved):

Formula/flif.rb
```diff
@@ -0,0 +1,75 @@
+class Flif < Formula
+  desc "Free Loseless Image Format"
+  homepage "http://flif.info/"
+  head "https://github.com/FLIF-hub/FLIF.git"
+
+  stable do
+    # patch 'install -s', see https://github.com/FLIF-hub/FLIF/issues/369
+    patch :DATA
```
MikeMcQuaid: Can you please submit this patch to the upstream developers of this project as a PR and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

MikeMcQuaid (dismissed):

Formula/flif.rb
stek29: @MikeMcQuaid I asked in my first comment about this patch -- is it common issue that `install -s` fails?
Anyways, I've did it: FLIF-hub/FLIF#382.

stek29:
Patch was merged into Upstream
Sorry for not noticing that trailing whitespace :\

Formula/flif.rb
MikeMcQuaid: @stek29 It's common that upstreams don't test on macOS/BSD versions of `install`.

ilovezfs (dismissed):
@stek29 please inquire upstream about whether they would be willing to create a new release so we can avoid merging a new formula with a patch.

stek29:
@ilovezfs FLIF-hub/FLIF#401

ilovezfs:
@stek29 thanks you!

stek29:
@ilovezfs @MikeMcQuaid The patch is no longer needed and was removed, since it was patched in upstream and a new release was made.

ilovezfs (approved):

ilovezfs:
@stek29 I've made some minor changes to the formula if you'd like to look it over.

stek29:
@ilovezfs Awesome, thanks :D

fxcoudert:
Thank you @stek29 for your contribution to Homebrew!

PR #12076, opened 2017-04-05
ntp 4.2.8p10: new formula

johnhawkinson:

The Network Time Protocol (NTP) Distribution.

-----

This pull request ported over from https://github.com/Homebrew/homebrew-dupes/pull/744 per @MikeMcQuaid 

> I appreciate this is a massive pain and I genuinely apologise for that but during this PR homebrew/dupes has been migrated to homebrew/core so after addressing the feedback could you recreate this there?

----

Addressing comments from the prior PR; I've made all these changes, although I'm not sure it is advisable to do so.

> Can remove all these comments

w/r/t https://github.com/Homebrew/homebrew-dupes/pull/744/files/ae3ce6715c032da50409c99f45f54df5ce01af5d#r109661078 I think it's necessary to keep some of the comments in the `head` section, otherwise it makes no sense that there's no URL there. The alternative, of course, is to remove the head section entirely. I've removed all the remaining such comments.

> Let's remove this; it's fine to use the system Perl here.

For the record, it's not fine on my system. As I've articulated previously in #10133, the result is I get commands that don't work:

```
pb3:~ jhawk$ ntptrace
Usage: DynaLoader::dl_find_symbol(libhandle, symbolname) at /Users/jhawk/perl5/lib/perl5/darwin-thread-multi-2level/XSLoader.pm line 89.
Compilation failed in require at /usr/local/Cellar/ntp/4.2.8p10/share/ntp/lib/NTP/Util.pm line 16.
BEGIN failed--compilation aborted at /usr/local/Cellar/ntp/4.2.8p10/share/ntp/lib/NTP/Util.pm line 24.
Compilation failed in require at /usr/local/sbin/ntptrace line 8.
BEGIN failed--compilation aborted at /usr/local/sbin/ntptrace line 8.
pb3:~ jhawk$ 
```

I guess the answer is that this something I have to resolve locally and Homebrew is not interested in fixing? I guess I need to maintain my own private fork of the formula.

> Turn this into an if/else to make it easier to follow.

```diff
 +    args << (build.with?("snmp") ? "-with-net-snmp-config" : "-with-net-snmp-config=no") 
```

Huh. Not only do I find the one-line syntax easier to follow, but there are only 3 other formulas that test for this condition (`lldpd`, `nut`, and `powerman`), and all of them use the one-line syntax.

### DIFF
--- a/Formula/ntp.rb
+++ b/Formula/ntp.rb
@@ -10,22 +10,6 @@ class Ntp < Formula
     sha256 "a07e73d7a3ff139bba33ee4b1110d5f3f4567465505d6317c9b50eefb9720c42"
   end
 
-  head do
-    ## The git repo is broken/ancient, per
-    ## https://github.com/ntp-project/ntp/issues/17 but the others
-    ## aren't much better.
-    ##
-    ## bk://bk.ntp.org/ntp-dev
-    ## (4.3.93) 2016/06/02 Released by Harlan Stenn <stenn@ntp.org>
-
-    # Unsupported syntax?: depends_on cask: "bitkeeper"
-
-    depends_on "autoconf" => :build
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
-    depends_on "lynx" => :build
-  end
-
   option "with-net-snmp", "Build net-snmp support"
 
   depends_on "openssl"
--- a/Formula/ntp.rb
+++ b/Formula/ntp.rb
@@ -43,7 +43,7 @@ class Ntp < Formula
       "--with-openssl-incdir=#{Formula["openssl"].include}",
     ]
     if build.with?("net-snmp")
-      args <<  "-with-net-snmp-config"
+      args << "-with-net-snmp-config"
     else
       args << "-with-net-snmp-config=no"
     end
--- a/Formula/ntp.rb
+++ b/Formula/ntp.rb
@@ -52,58 +52,6 @@ class Ntp < Formula
     system "make", "install"
   end
 
-  def caveats
-    s = <<-EOS.undent
-    By default, OS X ships a broken version of ntpd that by design
-    does not control the system clock, but maintains the network
-    protocol and is intended to write a drift file (which it fails to
-    do under OS X 10.10). The drift file is read by pacemaker(8),
-    which is supposed to call adjtime() in a battery-efficient fashion
-    for laptops.
-
-    Installing this formula implies disabling the system ntpd as well as pacemaker:
-
-      sudo launchctl disable system/com.apple.pacemaker
-      sudo launchctl disable system/org.ntp.ntpd
-    EOS
-    s
-  end
-
-  plist_options :startup => true
-
-  # OS X uses /usr/libexec/ntp-wrapper to wait for appropriate network conditions and then
-  # invoke ntpd with various args. It's not clear that we need to resort to that. It also
-  # runs sntp first, which is definitely unnecessary given ntpd -g.
-  def plist; <<-EOS.undent
-    <?xml version="1.0" encoding="UTF-8"?>
-    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-    <plist version="1.0">
-    <dict>
-      <key>KeepAlive</key>
-      <true/>
-      <key>Label</key>
-      <string>#{plist_name}</string>
-      <key>ProgramArguments</key>
-      <array>
-        <string>#{opt_sbin}/ntpd</string>
-        <string>-c</string>
-        <string>/private/etc/ntp-restrict.conf</string>
-        <string>-n</string>
-        <string>-g</string>
-        <string>-p</string>
-        <string>/var/run/ntpd.pid</string>
-        <string>-f</string>
-        <string>/var/db/ntp.drift</string>
-        <string>-l</string>
-        <string>/var/log/ntp.log</string>
-      </array>
-      <key>RunAtLoad</key>
-      <true/>
-    </dict>
-    </plist>
-    EOS
-  end
-
   test do
     assert_match "step time server ", shell_output("#{sbin}/ntpdate -bq pool.ntp.org")
   end
--- a/Formula/ntp.rb
+++ b/Formula/ntp.rb
@@ -59,7 +59,11 @@ class Ntp < Formula
       "--with-openssl-libdir=#{Formula["openssl"].lib}",
       "--with-openssl-incdir=#{Formula["openssl"].include}",
     ]
-    args << (build.with?("net-snmp") ? "-with-net-snmp-config" : "-with-net-snmp-config=no")
+    if build.with?("net-snmp")
+      args <<  "-with-net-snmp-config"
+    else
+      args << "-with-net-snmp-config=no"
+    end
 
     system "./configure", *args
 
--- a/Formula/ntp.rb
+++ b/Formula/ntp.rb
@@ -1,0 +1,117 @@
+class Ntp < Formula
+  desc "The Network Time Protocol (NTP) Distribution"
+  homepage "https://www.eecis.udel.edu/~mills/ntp/html/"
+  # Arguments could be made for:
+  # http://ntp.org/documentation.html (has more unofficial stuff)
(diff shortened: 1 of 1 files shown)


johnhawkinson:
So, Jenkins fails this with:

```
failed: brew audit ntp --online --new-formula
...
ntp:
  * Dependency 'net-snmp' may be unnecessary as it is provided by macOS; try to build this formula without it.
  * Dependency 'net-snmp' may be unnecessary as it is provided by macOS; try to build this formula without it.
```

I don't understand what this means; the dependency is there because otherwise there are problems linking against the system openssl's libcrypto which the system net-smmp depends on, but only in El Capitan and later. My previous version didn't have this problem before I marked `net-snmp` as optional per Mike (unless the `homebrew-core` j

Note that `nut` and `powerman` both fail in this way.

---

Also, I am confused. Why does http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html tell me to run `brew audit --strict` but not `--new-formula`? I guess this is just an oversight since the Cookbook indicates use of `--new-formula`? Somehow I overlooked that.

Also, why does Jenkins invoke this as `--online --new-formula` given that the usage says:

```
    If --new-formula is passed, various additional checks are run that check
    if a new formula is eligible for Homebrew. This should be used when creating
    new formulae and implies --strict and --online.
```
So specifying `--online` is superfluous?

Formula/ntp.rb
```diff
@@ -0,0 +1,58 @@
...
+  sha256 "ddd2366e64219b9efa0f7438e06800d0db394ac5c88e13c17b70d0dcdf99b99f"
+
+  devel do
+    url "https://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-dev/ntp-dev-4.3.93.tar.gz"
+    sha256 "a07e73d7a3ff139bba33ee4b1110d5f3f4567465505d6317c9b50eefb9720c42"
+  end
+
+  head do
+    ## The git repo is broken/ancient, per
+    ## https://github.com/ntp-project/ntp/issues/17 but the others
```
bfontaine: It doesn’t make sense to have a `head do` block without URL. Please remove the whole block.

Formula/ntp.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    # Unsupported syntax?: depends_on cask: "bitkeeper"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+    depends_on "lynx" => :build
+  end
+
+  option "with-net-snmp", "Build net-snmp support"
```
bfontaine: This options is automatically created by the `depends_on "net-snmp" => :optional` below.

bfontaine (changes requested):

Formula/ntp.rb
MikeMcQuaid: It just if you want to e.g. expand on the text to note what it does that you may want to keep this.

Formula/ntp.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    args = [
+      "--disable-debug",
+      "--disable-dependency-tracking",
+      "--disable-silent-rules",
+      "--prefix=#{prefix}",
+      "--with-openssl-libdir=#{Formula["openssl"].lib}",
+      "--with-openssl-incdir=#{Formula["openssl"].include}",
+    ]
+    if build.with?("net-snmp")
```
MikeMcQuaid: Could even do `args << if build.with?("net-snmp")` and then remove `args <<` in the lines below.

Formula/ntp.rb
```diff
@@ -0,0 +1,58 @@
...
+    else
+      args << "-with-net-snmp-config=no"
+    end
+
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  test do
+    assert_match "step time server ", shell_output("#{sbin}/ntpdate -bq pool.ntp.org")
```
MikeMcQuaid: Nice test!

MikeMcQuaid (changes requested):

MikeMcQuaid:
> I don't understand what this means; the dependency is there because otherwise there are problems linking against the system openssl's libcrypto which the system net-smmp depends on, but only in El Capitan and later.

It's fine to ignore in this case; this audit exists to just ensure the maintainers do a manual check.

> Also, I am confused. Why does http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html tell me to run brew audit --strict but not --new-formula? I guess this is just an oversight since the Cookbook indicates use of --new-formula?

It's mostly omitted just for brevity and because the e.g. notability checks aren't something users can do something about (not without messing with GitHub anyway which they will try on occasion.

ilovezfs:
Yes, it's superfluous. See https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L515-L517

Formula/ntp.rb
johnhawkinson: Removed.

Formula/ntp.rb
johnhawkinson: OK, let's expand it in a less generic way: 

```
==> Options
--with-net-snmp
	Build ntpsnmpd, the SNMP MIB agent for ntpd
```

Formula/ntp.rb
johnhawkinson: Done.

johnhawkinson (commented):
Addressed review comments.

bfontaine (approved):

johnhawkinson:
> It's mostly omitted just for brevity and because the e.g. notability checks aren't something users can do something about (not without messing with GitHub anyway which they will try on occasion.

I guess I was remarking more on the inconsistency between http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html and http://docs.brew.sh/Formula-Cookbook.html. If it's omitted for the above reason, then it should be the same in both documents.

After sleeping, another obvious reason occurs to me, which is that it's desired that new formulas receive more rigorous checks than modifications to existing formulas. If that's the reasoning, then it's fine as-is. But if it's really that an issue about users not having access to resolve the issue, then the two should probably be harmonized?

ilovezfs:
>After sleeping, another obvious reason occurs to me, which is that it's desired that new formulas receive more rigorous checks than modifications to existing formulas.

Right. There are checks `--new-formula` does that `--online --strict` does not do.

Formula/ntp.rb
```diff
@@ -0,0 +1,42 @@
...
+    args = [
+      "--disable-debug",
+      "--disable-dependency-tracking",
+      "--disable-silent-rules",
+      "--prefix=#{prefix}",
+      "--with-openssl-libdir=#{Formula["openssl"].lib}",
+      "--with-openssl-incdir=#{Formula["openssl"].include}",
+    ]
+    args << if build.with?("net-snmp")
+      "-with-net-snmp-config"
```
MikeMcQuaid: Just checking: this shouldn't be `--`?

MikeMcQuaid:
> I guess I was remarking more on the inconsistency between http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html and http://docs.brew.sh/Formula-Cookbook.html. If it's omitted for the above reason, then it should be the same in both documents.

One of them is more geared towards making changes to existing formulae and the other for creating new ones.

Formula/ntp.rb
johnhawkinson: It should be `--` yes. It appears that `configure` accepts it both ways and behaves appropriately (so I just learned by testing; I wondered how I could have not caught this in my testing of the formula if otherwise). I seem to have messed up my git workflow sufficiently enough that I can't pull your changes to my branch, so I'm just replying here for now instead of pushing another commit addressing this until I can figure that out.

w/r/t your indentation tweak, that essentially reverts 1865b56 and generates an auditing failure, does it not?

MikeMcQuaid: > w/r/t your indentation tweak, that essentially reverts 1865b56 and generates an auditing failure, does it not?

Yes. Stupid `rubocop`. I put it back to how it was before because the Rubocop indentation makes my eyes bleed. Sorry for the time wastage.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks for your contribution to Homebrew (and patience) @johnhawkinson! Without people like you submitting PRs we couldn't run this project. You rock!

Issue #5751, opened 2016-10-10
Installing uwsgi with python3 still uses python2


maneyko:
After doing a clean install

``` bash
brew uninstall --force uwsgi
brew install uwsgi --with-python --with-python3
```

and using a test flask app

``` python
# hello.py

from flask import Flask
app = Flask(__name__)

@app.route('/')
def index():
    return 'Hello world!'
```

uwsgi **still** uses python2 even when specifying python3 as a command line argument

``` bash
uwsgi --plugin python3 --http-socket :8080 --module hello --callable app
```

The following patch fixed the problem on my machine, it seems that running python2 to install the python3 plugin [doesn't do anything](https://stackoverflow.com/questions/16264213/uwsgi-python3-plugin-doesnt-work).

``` patch
diff --git a/Formula/uwsgi.rb b/Formula/uwsgi.rb
index 4e191d7..5632032 100644
--- a/Formula/uwsgi.rb
+++ b/Formula/uwsgi.rb
@@ -124,7 +124,7 @@ class Uwsgi < Formula
     python_versions = ["python", "python2"]
     python_versions << "python3" if build.with? "python3"
     python_versions.each do |v|
-      system "python", "uwsgiconfig.py", "--verbose", "--plugin", "plugins/python", "brew", v
+      system v, "uwsgiconfig.py", "--verbose", "--plugin", "plugins/python", "brew", v
     end

     bin.install "uwsgi"
```

dunn:
Would you like to try opening a PR? Take a look at https://github.com/Homebrew/brew/blob/master/docs/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md and let us know if you have questions!

ilovezfs:
Happy to accept a PR for this as @dunn said.

PR #12100, opened 2017-04-05
uwsgi fix --with-python3 to build plugin using python3

deinspanjer:

-----

Without this PR, if you `brew install uwsgi --with-python3`, the plugin is compiled using the `python` interpreter which is likely version 2.  This means that if you run `uwsgi --plugin python3`, it will not actually be running with the python3 interpreter.

Issue #5751 mentions this problem, but the fix suggested in that issue did not work for me because it threw an error when trying to build a variant of the plugin with `python2 uwsgiconfig.py --verbose --plugin plugins/python brew python2`.

I resolved that error by using a map for the python plugins building the bare `python` and the `python2` plugins with the `python` interpreter and `python3` plugin with the `python3` interpreter.

I'd like to add a test that examines the output of the exec call to verify that uwsgi is using the Python version 3 interpreter when run with the `--plugin python3` argument, but I wasn't sure how to capture the output from exec as opposed to IO.popen.


fxcoudert:
poke @tdsmith

tdsmith:
Thanks for the PR and sorry for the delay! If we tested that, it would only apply in a non-default configuration, so it wouldn't make this very much less fragile. (If we should start building Python 3 versions by default, otoh...)

PR #12107, opened 2017-04-05
giflib migration

fxcoudert:
- Upgrading the `giflib` formula from 4.2.3 to 5.1.4
- Migrating the previous `giflib@5` formula to `giflib`
- Bumping revision on formulas that depend on `giflib`
- Patching `pdf2htmlex` and `xplanet` because they don't compile with giflib ≥ 5
- Removing `--with-giflib` from `swftools` because analytics show it's unused


Formula/xplanet.rb
```diff
@@ -44,6 +45,13 @@ class Xplanet < Formula
     sha256 "3f95ba8d5886703afffdd61ac2a0cd147f8d659650e291979f26130d81b18433"
   end
 
+  # Fix compilation with giflib 5
+  # http://xplanet.sourceforge.net/FUDforum2/index.php?t=msg&th=592
```
vszakats: Can this be updated to?:
`https://xplanet.sourceforge.io/FUDforum2/index.php?t=msg&th=592`

fxcoudert: Do we really care whether comment URLs are HTTP or HTTPS?

vszakats: Sorry, it should be `https://xplanet.sourceforge.io/` (not `https://xplanet.sourceforge.net/`), former works, latter doesn't.

fxcoudert:
`mapserver` and `swftools` do not compile, even before the patch. What's the right thing to do then? Delay I can get them fixed (no idea when), or skip them and commit the rest?

ilovezfs:
`'Python.h' file not found` is probably caused by superenv filtering (https://github.com/Homebrew/brew/pull/1746) and having brew python opportunistically selected. A fix in `moose` was `"-DPYTHON_EXECUTABLE:FILEPATH=#{which("python")}"` (https://github.com/Homebrew/homebrew-science/blob/master/moose.rb#L50)

formula_renames.json
```diff
@@ -48,6 +48,7 @@
   "gcc@6": "gcc",
   "geode": "apache-geode",
   "giflib5": "giflib@5",
+  "giflib@5": "giflib",
```
MikeMcQuaid: This may be weird. CC @ilovezfs

MikeMcQuaid (approved):
Looks good here.

fxcoudert:
I am dropping the mapserver revision bump since it fails to build…

fxcoudert:
@ilovezfs thanks, that fixes `mapserver`

formula_renames.json
ilovezfs: I'd probably retire `"giflib5": "giflib@5",` to avoid the chaining but I believe @JCount said it sort of works.

ilovezfs:
`xplanetfx` probably needs a revision bump.

ilovezfs:
And `swftools` needs `depends_on "freetype"`

ilovezfs:
Going to pull this as-is and use it as an opportunity to test https://github.com/Homebrew/brew/pull/1975.

ilovezfs (approved):

ilovezfs:
Oh well. No warning at all from brew pull about swftools not getting new bottles. CC @MikeMcQuaid

ilovezfs:
terminal output from brew pull https://gist.github.com/ilovezfs/c8751efb03ab5bc7b9c9a4d9f4963663

ilovezfs:
Pulled without the swftools commit.

MikeMcQuaid:
> No warning at all from brew pull about swftools not getting new bottles. CC @MikeMcQuaid

```
==> Publishing on Bintray: swftools 0.9.2_1
{"files":0}
...
Verifying bottle: swftools-0.9.2_1.sierra.bottle.tar.gz
```

ilovezfs:
Yes I know. Bottle fails to build, no bottle commit json, no rebuild bump, and so grabs the old bottle without a peep.

MikeMcQuaid:
Yeh. I'm open to a PR for that but for me `{"files":0}` has always been a/the red flag.

ilovezfs:
in which case I will consider https://github.com/Homebrew/brew/pull/1975 to be for some other use case and will look for some other opportunity to test it since this clearly wasn't it

MikeMcQuaid:
@ilovezfs Ok, I see. Yeh, that's an additional edge case there where a previous bottle exists but we didn't update it. I guess we could check if `--bottle` is passed for the bottle `rebuild` before and after the pull and go :rage4: if it doesn't change?

ilovezfs:
>I guess we could check if --bottle is passed for the bottle rebuild before and after the pull and go :rage4: if it doesn't change?

That sounds right since `--bottle` implies new bottles for all formulae except those with an explicit `bottle :unneeded` or `bottle :disable`. Sometimes, that's a hassle because we actually don't necessarily want a new bottle for everything, so some unnecessary bottles get pulled, but in the absence of some sort --skip-bottles-for=foo,bar list, it seems like the absence of a new  bottle commit for something that has a bottle block "should" be a hard `brew pull` failure without `--warn-on-publish-failure ` and a warning with `--warn-on-publish-failure `.

ilovezfs:
I guess `--warn-on-publish-failure` is somewhat strangely named if it's going to cover that case too, though, in terms of converting error to warning.

PR #12129, opened 2017-04-06
freedink 108.4 / freedink-data 1.08.20170409

djpadz:


Formula/freedink-data.rb
```diff
@@ -0,0 +1,18 @@
...
+  desc "Data for FreeDink."
+  homepage "https://www.gnu.org/software/freedink/"
+  url "https://ftpmirror.gnu.org/freedink/freedink-data-1.08.20170401.tar.xz"
+  sha256 "e8d3f3ff55b1a86661949d08d901740d65a6092027886fd791f6c315b9c7a7ff"
+
+  def install
+    system "sed", "-i.bak", "-e", "s,^VERSION,#VERSION,", "-e", "s,^SOURCE_DATE_EPOCH,#SOURCE_DATE_EPOCH,", "-e", "s,xargs -0r,xargs -0,", "Makefile"
+    version=Dir.pwd.split("/")[-1].split("-")[-1]
+    source_date = Dir.pwd.split("/")[-1].split(".")[-1].match(/(\d{4})(\d{2})(\d{2})/).to_a[1..-1].join("-")
+    source_date_epoch = Time.parse("#{source_date} 0:00 UTC").to_i
```
bfontaine: Why do we need to do all this parsing?

Formula/freedink-data.rb
```diff
@@ -0,0 +1,18 @@
+class FreedinkData < Formula
+  desc "Data for FreeDink."
```
bfontaine: You can make this a resource of the `freedink` formula instead of a formula in itself.

Formula/freedink.rb
```diff
@@ -0,0 +1,32 @@
...
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+
+    system "make", "install"
+    ln_s "#{HOMEBREW_PREFIX}/freedink-data/share/dink", "#{share}/dink"
+  end
+
+  test do
+    FileTest.executable? "src/freedink"
```
bfontaine: Please write a test that executes something instead of just checking if a file exists.

Formula/freedink.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on "freedink-data"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+
+    system "make", "install"
+    ln_s "#{HOMEBREW_PREFIX}/freedink-data/share/dink", "#{share}/dink"
```
bfontaine: This should be done as a post-install step (in `def post_install`) but making `freeding-data` a resource of this formula is simpler.

bfontaine (changes requested):

Formula/freedink-data.rb
djpadz: Believe me, I wasn't happy about this, either.  The author puts the build date into the version, which would lead to different people getting different versions of the code, depending on when they installed it.  I decided to rewrite it, so that the version would reflect the author's release date, rather than the build date.

Formula/freedink.rb
djpadz: I'm not quite sure I understand the second part of that.  How would I do that?

Formula/freedink-data.rb
djpadz: I'm not sure how to do that.  Can you send me a link to some documentation?

Formula/freedink-data.rb
bfontaine: You should report this upstream.

Formula/freedink-data.rb
bfontaine: See e.g. [`cracklib.rb`](https://github.com/Homebrew/homebrew-core/blob/312a38050f7bf3fa4f644a9d6d23d2d9dabac506/Formula/cracklib.rb#L18). You have to declare the external resource in a `resource do` block:

```ruby
resource "freedink-data" do
  url "..."
  sha256 "..."
end
```

Then in your `def install` you can unpack it with `Resource#stage`:
```ruby
resource("freedink-data").stage do
  # execute some code here to install the additional resource
  # the freedink-data resource is unpacked in a temporary
  # directory and your code is run in that directory.
end
```

Formula/freedink.rb
bfontaine: See my other comment: https://github.com/Homebrew/homebrew-core/pull/12129#discussion_r114116644

Formula/freedink.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "make", "install"
+  end
+
+  def post_install
+    ln_s "#{HOMEBREW_PREFIX}/freedink-data/share/dink", "#{share}/dink"
+  end
+
+  test do
+    FileTest.executable? "src/freedink" &&
+    %x[src/freedink -vwis].split("\n").first == 'GNU FreeDink 108.4'
```
bfontaine: Please use `shell_output("…")` instead of `%x[...]`. Also, `freedink` should be installed in `bin/` if it’s executable.

djpadz:
Sorry I let this go so long...  Competing priorities...

ilovezfs:
@BrewTestBot test this please

djpadz:
@bfontaine I think I've addressed all of the issues.  The upstream issue isn't really one that's blocking the formula; more of a style decision on the part of the author, which I decided to work around.

bfontaine:
Please report the issue upstream anyway; it’d be better if they could provide a way to do so instead of us relying of this code.

(sorry for the late response, I was in holidays away from my computer)

djpadz:
I dropped the following email to `freedink-bugs@gnu.org`:

> Hi there-
> 
> I’m building a package for freedink and freedink-data for HomeBrew, and I received the following feedback from the maintainers, which I thought I’d pass along.
> 
> In the Makefile for freedink-data, the build date is included in the version string.  This is probably not the best idea, since it means that my version of freedink-data will appear to be newer than yours, simply by virtue of the fact that I happened to type make after you did.  It might be a good idea to set your versions to static strings.  If you want the build date in there somewhere, then put it elsewhere, but part of the version string probably isn’t the best place.
> 
> In any case, for HomeBrew, I’m munging the date portion of the source directory, and using that in lieu of a build date.
> 
> Best,
> 
> —Dj
> 

I've also updated the `freedink-data` resource to the latest version.

ilovezfs:
The current test block gets stuck on CI indefinitely until the build is aborted.

djpadz:
Found the cause of the test hang.  Should be okay now.

djpadz:
@bfontaine I got an email back from the developer, and, as it turns out, I didn't look at the `Makefile` hard enough.  The parsing stuff wasn't necessary, so I've removed that stuff.  All that's needed is the change from `xargs -0r` to `xargs -0`.

Formula/freedink.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+
+    system "make", "install"
+    resource("freedink-data").stage do
+      system "sed", "-i.bak", "-e", "s,xargs -0r,xargs -0,", "Makefile"
```
dunn: We have an `inreplace` method that can be used instead of `sed`.  But this should also be reported upstream.

djpadz: I mentioned the `sed` statement to the author. I haven't heard back, so I'd say we can file that under "lesser sins."  As for the `inreplace` method, I'll get a look at it, and push another commit later tonight.

bfontaine (approved):

Formula/freedink.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    system "make", "install"
+    resource("freedink-data").stage do
+      inreplace "Makefile", "xargs -0r", "xargs -0"
+      system "make", "install", "PREFIX=#{prefix}"
+    end
+  end
+
+  test do
+    FileTest.executable?("#{bin}/freedink") && shell_output("#{bin}/freedink -vwis").split("\n").first == "GNU FreeDink 108.4" && FileTest.exists?("#{share}/dink/dink/Dink.dat")
```
dunn: These can all be on separate lines; if any fail the test fails.

djpadz: I don't think that's the case.  In a ruby code block, the return value is the last value in the block.  For example, I changed the `test do` block to this:

```
  test do
    FileTest.exists?("/foo/bar/baz")
    FileTest.executable?("#{bin}/freedink")
    shell_output("#{bin}/freedink -vwis").split("\n").first == "GNU FreeDink 108.4"
    FileTest.exists?("#{share}/dink/dink/Dink.dat")
  end
```

This test passes (you can try it yourself), even though the first test fails, because the block returns `true`.

A viable alternative would be:

```
  test do
    return false unless FileTest.executable?("#{bin}/freedink")
    return false unless shell_output("#{bin}/freedink -vwis").split("\n").first == "GNU FreeDink 108.4"
    return false unless FileTest.exists?("#{share}/dink/dink/Dink.dat")
  end
```

djpadz:
_Moving my response outside of the obsolete block_

@dunn I don't think that's the case.  In a ruby code block, the return value is the last value in the block.  For example, I changed the `test do` block to this:

```
  test do
    FileTest.exists?("/foo/bar/baz")
    FileTest.executable?("#{bin}/freedink")
    shell_output("#{bin}/freedink -vwis").split("\n").first == "GNU FreeDink 108.4"
    FileTest.exists?("#{share}/dink/dink/Dink.dat")
  end
```

This test passes (you can try it yourself), even though the first test fails, because the block returns `true`.

A viable alternative would be:

```
  test do
    return false unless FileTest.executable?("#{bin}/freedink")
    return false unless shell_output("#{bin}/freedink -vwis").split("\n").first == "GNU FreeDink 108.4"
    return false unless FileTest.exists?("#{share}/dink/dink/Dink.dat")
  end
```

dunn:
My mistake, sorry.  But instead of returns, do `assert File.exist? file` etc.

djpadz:
That seems reasonable :-)

Formula/freedink.rb
```diff
@@ -0,0 +1,41 @@
...
+    system "make", "install"
+    resource("freedink-data").stage do
+      inreplace "Makefile", "xargs -0r", "xargs -0"
+      system "make", "install", "PREFIX=#{prefix}"
+    end
+  end
+
+  test do
+    assert FileTest.executable?("#{bin}/freedink")
+    assert shell_output("#{bin}/freedink -vwis").split("\n").first == "GNU FreeDink 108.4"
```
bfontaine: Use `assert_match "GNU FreeDink 108.4", shell_output("#{bin}/freedink -vwis")` and remove the `FileTest.executable?` above; `shell_output` will fail if it’s not executable anyway.

djpadz: Done.

Formula/freedink.rb
```diff
@@ -34,8 +34,7 @@ def install
   end
 
   test do
-    assert FileTest.executable?("#{bin}/freedink")
-    assert shell_output("#{bin}/freedink -vwis").split("\n").first == "GNU FreeDink 108.4"
+    assert_match "GNU FreeDink 108.4", shell_output("#{bin}/freedink -vwis").split("\n").first
```
bfontaine: You can remove `.split("\n").first`; `assert_match` already ensures `GNU FreeDink 108.4` is *somewhere* in that output. We have `assert_equal` for strict equality.

djpadz:
@bfontaine Done.

djpadz:
@bfontaine @dunn So, are you waiting on me for anything?  I believe we've cleared the upstream issue.

bfontaine:
I stepped away from maintainership since you opened the issue so you’ll have to wait for @dunn or others :wink:

dunn:
Merged as 89b2d82defa415f8106b25010a389119760d198b, thanks again!

djpadz:
Thanks, @dunn & @bfontaine !

PR #12144, opened 2017-04-06
logstash: fix lack of plist

robertjrodger:
Elasticsearch and Kibana both offer out-of-the-box run-as-a-service functionality, so I added it for Logstash.

-----

### DIFF
--- a/Aliases/logstash@5.3
+++ b/Aliases/logstash@5.3
@@ -1,0 +1,1 @@
+../Formula/logstash.rb
--- a/Formula/logstash.rb
+++ b/Formula/logstash.rb
@@ -33,6 +33,37 @@ class Logstash < Formula
     EOS
   end
 
+  plist_options :manual => "logstash"
+
+  def plist; <<-EOS.undent
+      <?xml version="1.0" encoding="UTF-8"?>
+      <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+      <plist version="1.0">
+        <dict>
+          <key>KeepAlive</key>
+          <false/>
+          <key>Label</key>
+          <string>#{plist_name}</string>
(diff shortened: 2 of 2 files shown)


fxcoudert:
@MikeMcQuaid shouldn't the versions be `logstash@2` (instead of `@2.4`) and `logstash@5` (instead of `@5.3`)? to keep at least some logic in the versioned formulas…

robertjrodger:
Where do you see `logstash@2`? I used `logstash@5.3` because that was the suggestion given by the `brew audit` tool results.

MikeMcQuaid:
> @MikeMcQuaid shouldn't the versions be logstash@2 (instead of @2.4) and logstash@5 (instead of @5.3)? to keep at least some logic in the versioned formulas…

@fxcoudert I think it's worth avoiding renaming logstash@2.4 again for now therefore having a major.minor alias is fine for me.

fxcoudert:
@MikeMcQuaid what happens when 2.5.0 or 5.4.0 is released? do we keep all minor versions? do we migrate for every new minor version? I'd like that we have a consistent policy for those versioned formulas (or maybe we have, but I totally missed it)

MikeMcQuaid:
> @MikeMcQuaid what happens when 2.5.0 or 5.4.0 is released? do we keep all minor versions? do we migrate for every new minor version? I'd like that we have a consistent policy for those versioned formulas (or maybe we have, but I totally missed it)

http://docs.brew.sh/Versions.html should provide guidance here.

MikeMcQuaid:
@robertjrodger It's been sorted so you can ignore it now, thanks.

robertjrodger:
@jasontedor Hi, I see that you're the one who's been keeping the Logstash formula up-to-date recently. What do you think of the addition of a plist?

MikeMcQuaid:

jasontedor:
> What do you think of the addition of a plist?

Sounds good to me, thanks for doing this.

PR #12039, opened 2017-04-04
gnupg 2.1.20

neutric:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [ ] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [ ] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

-----

PR #12145, opened 2017-04-06
gnupg 2.1.20: enable larger secure memory

adam-moss:

-----

GnuPG 2.1.x includes support for up-to 8k RSA keys via the
`--enable-large-rsa` option.  To make use of this function GnuPG must
be built with the `--enable-large-secmem` option.

### DIFF
--- a/Formula/gnupg.rb
+++ b/Formula/gnupg.rb
@@ -4,6 +4,7 @@ class Gnupg < Formula
   url "https://gnupg.org/ftp/gcrypt/gnupg/gnupg-2.1.20.tar.bz2"
   mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/gnupg/gnupg-2.1.20.tar.bz2"
   sha256 "24cf9a69369be64a9f6f8cc11a1be33ab7780ad77a6a1b93719438f49f69960d"
+  revision 1
 
   bottle do
     sha256 "da72dd30eff4131ff8177528b5bf03f6e0273e553bdf34e8daa490696e24a92c" => :sierra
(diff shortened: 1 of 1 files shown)


fxcoudert:
What's the downside? Or, in other terms, why do we want to enable it in Homebrew if upstream decided not to make it the default?

adam-moss:
Its a fair question.

So the downside is you have a hard limit of 4k RSA keys, and if you include the `--enable-large-rsa` option on the command line or your `gpg.conf` it will error out.

On the other hand if you don't do either of those things, you would never know 😃 

In terms of whether it is a *good idea* to have keys > 4k I personally have no strong feelings either way, there is https://wiki.gnupg.org/LargeKeys, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=739424#30, and pg52 of the manual (https://gnupg.org/documentation/manuals/gnupg/GPG-Configuration-Options.html#GPG-Configuration-Options) which expresses some opinions.

I guess its a bit like ECC being hidden by default currently.

fxcoudert:
Usually we prefer to ship packages as close as possible to upstream's choices…
paging @DomT4 @ilovezfs on that one

DomT4:
I don't think there's any practical application where 4096 bit RSA keys aren't secure enough. I think the default is reasonable and I worry there's an assumption with newer crypto users that bigger is always better, whereas in this case bigger makes no appreciable difference & is likely to actually slow down encryption/decryption operations.

I guess if there's a desire for it I'd probably say it should be an option. I'd be reluctant to turn it on by default.

adam-moss:
>I worry there's an assumption with newer crypto users that bigger is always better, whereas in this case bigger makes no appreciable difference & is likely to actually slow down encryption/decryption operations

I would certainly agree with that statement.  https://www.keylength.com/en/compare/ is good to point people at for that.

>I guess if there's a desire for it I'd probably say it should be an option. I'd be reluctant to turn it on by default.

I'm certainly happy to rework the PR to do it that way (I did actually ask that in https://github.com/Homebrew/homebrew-core/pull/12039#issuecomment-291637512 although I appreciate that was probably the wrong place).  I'm equally happy if its closed - its definitely an edge case.

PR #12153, opened 2017-04-06
mariadb 10.2.5 (devel)

makigumo:

There is already #12127. But this should fix the build error.

-----

### DIFF
--- a/Formula/mariadb.rb
+++ b/Formula/mariadb.rb
@@ -11,8 +11,8 @@ class Mariadb < Formula
   end
 
   devel do
-    url "https://ftp.osuosl.org/pub/mariadb/mariadb-10.2.4/source/mariadb-10.2.4.tar.gz"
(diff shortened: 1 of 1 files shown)


neutric:
@makigumo: It certainly does! @ilovezfs, you cool with this?

neutric:
@makigumo We can go ahead and merge this here. Since we try to keep formulae as close to the default build as possible and you are regularly updating this formula, would you mind reporting the failing build upstream? That'd be awesome! :1st_place_medal:

makigumo:
@neutric Looks like it was already reported and fixed upstream: https://jira.mariadb.org/browse/MDEV-12452

As 'rocksdb' is a newly added component and nothing old got broken, I would propose to wait for 10.2.6 to (re)enable it.

ilovezfs:
@makigumo Thank you for looking into this! Yes, as long as it's been reported we're good.

Issue #11898, opened 2017-03-31
Formulae without a `test do` test block


ilovezfs:
Please help add `test do` test blocks to the formulae missing them. If you see something on the list that now _does_ have a test block, just comment below, and one of the maintainers will check the box.

- [x] aces_container
- [x] ahcpd
- [x] aldo
- [x] alure
- [ ] anttweakbar
- [ ] aoeui
- [x] apachetop
- [x] apollo
- [ ] app-engine-python
- [x] arabica
- [x] argtable
- [x] argus-clients
- [x] argus
- [x] arm
- [x] arp-sk
- [x] arpoison
- [x] arss
- [ ] at-spi2-atk
- [x] at-spi2-core
- [x] atf
- [x] atomicparsley
- [x] autoenv
- [ ] autopsy
- [x] aview
- [x] avro-c
- [x] avro-cpp
- [x] avro-tools
- [x] backupninja
- [x] balance
- [ ] ballerburg
- [x] bash-git-prompt
- [ ] bastet
- [x] batik
- [x] bbe
- [x] bdw-gc
- [x] beanstalkd
- [x] blahtexml
- [x] blitz
- [ ] blitzwave
- [x] bogofilter
- [x] bokken
- [x] bonnie++
- [x] boost-build
- [ ] bpm-tools
- [x] brew-pip
- [x] briss
- [ ] bsdsfv
- [ ] bsponmpi
- [x] cadaver
- [ ] camellia
- [ ] cataclysm
- [x] cclive
- [ ] cctools-headers
- [x] cdargs
- [x] cdparanoia
- [x] cdpr
- [ ] cdrdao
- [x] cgdb
- [x] cherokee
- [x] chgems
- [ ] chmlib
- [x] chocolate-doom
- [ ] chromaprint
- [x] chruby
- [x] cityhash
- [ ] classads
- [ ] clens
- [ ] clucene
- [ ] cmigemo
- [ ] cmockery
- [ ] cmu-pocketsphinx
- [ ] cmu-sphinxbase
- [ ] cmuclmtk
- [ ] cntlm
- [x] coccinelle
- [ ] cocot
- [ ] collada-dom
- [x] contacts
- [ ] convertlit
- [x] corsixth
- [ ] cpansearch
- [ ] crash
- [ ] crm114
- [ ] cronolog
- [x] crunch
- [ ] cssembed
- [ ] ctemplate
- [ ] curlftpfs
- [ ] cvsps
- [ ] davmail
[...]

PR #12164, opened 2017-04-07
autoenv: add test

zmwangx:
#11898.

### DIFF
--- a/Formula/autoenv.rb
+++ b/Formula/autoenv.rb
@@ -16,4 +16,10 @@ class Autoenv < Formula
       source #{opt_prefix}/activate.sh
     EOS
   end
+
+  test do
+    (testpath/"test/.env").write "echo it works\n"
+    testcmd = "yes | bash -c '. #{prefix}/activate.sh; autoenv_cd test'"
+    assert_match "it works", shell_output(testcmd)
+  end
 end


Formula/autoenv.rb
```diff
@@ -16,4 +16,11 @@ def caveats; <<-EOS.undent
       source #{opt_prefix}/activate.sh
     EOS
   end
+
+  test do
+    (testpath/"test").mkdir
```
ilovezfs: mkpath

Formula/autoenv.rb
```diff
@@ -16,4 +16,11 @@ def caveats; <<-EOS.undent
...
+    (testpath/"test/.env").write "echo it works"
```
ilovezfs: Should this have \n so it's a proper text file?

Formula/autoenv.rb
```diff
@@ -16,4 +16,11 @@ def caveats; <<-EOS.undent
...
+    testcmd = "yes | bash -c '. #{opt_prefix}/activate.sh; autoenv_cd test'"
```
ilovezfs: Should we use pipe_output?

Formula/autoenv.rb
DomT4: Shouldn't be needed at all given the `.write` below.

Formula/autoenv.rb
zmwangx: I think `shell_output` makes much more sense here, because you should treat the entire thing as a test command, rather than, this part is the test command, that part is the input we feed into it (a y to answer a prompt...).

ilovezfs: Why doesn't that argument apply to all uses of pipe_output?

Formula/autoenv.rb
```diff
@@ -16,4 +16,11 @@ def caveats; <<-EOS.undent
       source #{opt_prefix}/activate.sh
     EOS
   end
+
+  test do
+    (testpath/"test").mkpath
+    (testpath/"test/.env").write "echo it works\n"
+    testcmd = "yes | bash -c '. #{opt_prefix}/activate.sh; autoenv_cd test'"
```
ilovezfs: if it's for the formula being tested, not some other formula, I prefer `prefix` over `opt_prefix`

Formula/autoenv.rb
zmwangx: Because a typical use of pipe_output concerns a test command that can handle any input, or at least a class of input. You specify one (usually simple) input and check the output. Here, there's only one input that makes any sense at all.

ilovezfs: OK.

Formula/autoenv.rb
```diff
@@ -16,4 +16,11 @@ def caveats; <<-EOS.undent
       source #{opt_prefix}/activate.sh
     EOS
   end
+
+  test do
+    (testpath/"test").mkpath
```
ilovezfs: per Dom's comment I guess this line isn't actually needed due to the `.write` which creates the directory for you

Formula/autoenv.rb
zmwangx: Oh, and `pipe_output("bash -c 'blah blah'", "y\n")` is much less understandable as far as I am concerned. I know what `yes` commonly does. I don't know what the heck `y\n` is.

Formula/autoenv.rb
zmwangx: Wooh TIL.

Formula/autoenv.rb
ilovezfs: I think generally our use of a mixture of pipe_output and shell_output depending on circumstances when `|` could be used instead is kind of weird but :man_shrugging:

ilovezfs:
:robot:

PR #12166, opened 2017-04-07
iftop: add test

zmwangx:
#11898.

### DIFF
--- a/Formula/iftop.rb
+++ b/Formula/iftop.rb
@@ -27,4 +27,8 @@ class Iftop < Formula
     You should be certain that you trust any software you grant root privileges.
     EOS
   end
+
+  test do
+    assert_match "interface:", pipe_output("#{sbin}/iftop -t -s 1 2>&1")
+  end
 end


Formula/iftop.rb
```diff
@@ -27,4 +27,8 @@ def caveats; <<-EOS.undent
     You should be certain that you trust any software you grant root privileges.
     EOS
   end
+
+  test do
+    assert_match "interface:", pipe_output("#{sbin}/iftop -t -s 1 2>&1")
```
ilovezfs: what's the exit code? we should be able to use shell_output and assert it

zmwangx: There are two possible exit codes. 1 if you use sandbox and it fails to read the device file. 0 if you --no-sandbox. Which is why I'm using pipe_output.

ilovezfs: OK

ilovezfs:
:tophat:

PR #12169, opened 2017-04-07
kite: add test

zmwangx:
#11898.

### DIFF
--- a/Formula/kite.rb
+++ b/Formula/kite.rb
@@ -21,6 +21,10 @@ class Kite < Formula
                           "--prefix=#{prefix}"
     system "make", "install"
   end
+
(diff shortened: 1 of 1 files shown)


Formula/kite.rb
```diff
@@ -21,6 +21,10 @@ def install
                           "--prefix=#{prefix}"
     system "make", "install"
   end
+
+  test do
+    assert_equal "hello, world\n", pipe_output(bin/"kite", "'hello, world'|print;", 0)
```
ilovezfs: same

ilovezfs:
Thanks for adding all of these tests @zmwangx :cookie:

PR #12170, opened 2017-04-07
libev: add test

zmwangx:
#11898.

### DIFF
--- a/Formula/libev.rb
+++ b/Formula/libev.rb
@@ -20,4 +20,37 @@ class Libev < Formula
     # Remove compatibility header to prevent conflict with libevent
     (include/"event.h").unlink
   end
+
+  test do
+    (testpath/"test.c").write <<-'EOS'.undent
+      /* Wait for stdin to become readable, then read and echo the first line. */
+
+      #include <stdio.h>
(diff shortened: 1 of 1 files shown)


Formula/libev.rb
```diff
@@ -20,4 +20,37 @@ def install
...
+
+      int main() {
+        ev_io_init(&stdin_watcher, stdin_cb, STDIN_FILENO, EV_READ);
+        ev_io_start(EV_DEFAULT, &stdin_watcher);
+        ev_run(EV_DEFAULT, 0);
+        return 0;
+      }
+    EOS
+    system ENV.cc, "-I#{include}", "-L#{lib}", "-lev", "-o", "test", "test.c"
+    input = "hello, world\n"
```
ilovezfs: same

zmwangx: It makes more sense to use a complete line as input.

ilovezfs: I'd probably do
```
output = pipe_output("./test", "hello world\n", 0)
assert_equal "hello world", output.chomp
```
but this is fine.

ilovezfs:
:evergreen_tree:

PR #12187, opened 2017-04-07
swftools: remove giflib option

ilovezfs:

-----

CC @fxcoudert

### DIFF
--- a/Formula/swftools.rb
+++ b/Formula/swftools.rb
@@ -18,7 +18,6 @@ class Swftools < Formula
   depends_on :x11 if build.with? "xpdf"
   depends_on "jpeg" => :optional
   depends_on "lame" => :optional
-  depends_on "giflib" => :optional
   depends_on "fftw" => :optional
 
   resource "xpdf" do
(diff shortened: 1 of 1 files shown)


fxcoudert:
You win by 6 minutes #12188
Although from the documentation, it seems that freetype is required for the same tool as jpeg is required, so if we're going to depend on that we should as well include jpeg (low cost) to make it useful.

ilovezfs:
Will wait for users to request that before adding dependencies.

fxcoudert:
But why was build failing before and not now?

ilovezfs:
Because freetype was installed due to the other things going on in the PR, leading to its opportunistic use, superenv filtering, and :boom:

ilovezfs:
@fxcoudert this is why the superenv filtering is there. We would have ended up with a bottle linked to freetype, but no `depends_on "freetype"` in the formula or any of its dependencies, leading to broken dylib errors for users but not CI.

ilovezfs:
Um ... seems squash and merge decided to remove your authorship somehow. wtf? Sorry

PR #12188, opened 2017-04-07
swftools: remove giflib, add freetype dependency

fxcoudert:
Following #12107
Freetype and jpeg are required dependencies. Fixes build.


ilovezfs:
duplicates https://github.com/Homebrew/homebrew-core/pull/12187

and actually freetype turns out to only be an issue if it's already installed.

PR #12189, opened 2017-04-07
Add minikube formula

bonifaido:

-----


Formula/minikube.rb
```diff
@@ -0,0 +1,16 @@
...
+  version "v0.17.1"
+  sha256 "b175c355d377a6ce2fefdd19201c865a7e628581261ac949fffb725af459c389"
+
+  def install
+    mv "minikube-darwin-amd64", "minikube"
+    bin.install "minikube"
+  end
+
+  test do
+    system "minikube"
```
bfontaine: Please use the full path instead: `bin/"minikube"`

Formula/minikube.rb
```diff
@@ -0,0 +1,16 @@
+class Minikube < Formula
+  desc "Run Kubernetes locally"
+  homepage "https://github.com/kubernetes/minikube"
+  url "https://github.com/kubernetes/minikube/releases/download/v0.17.1/minikube-darwin-amd64"
+  version "v0.17.1"
+  sha256 "b175c355d377a6ce2fefdd19201c865a7e628581261ac949fffb725af459c389"
+
+  def install
+    mv "minikube-darwin-amd64", "minikube"
+    bin.install "minikube"
```
bfontaine: Use `bin.install "foo" => "bar"` to do the equivalent of `mv "foo", "bar"; bin.install "bar"`.

Formula/minikube.rb
```diff
@@ -0,0 +1,16 @@
+class Minikube < Formula
+  desc "Run Kubernetes locally"
+  homepage "https://github.com/kubernetes/minikube"
+  url "https://github.com/kubernetes/minikube/releases/download/v0.17.1/minikube-darwin-amd64"
```
bfontaine: We need to build from source here.

bfontaine (changes requested):

eneveu:
Note that there is already a cask for Minikube:

```
 ~/code > brew cask info minikube
minikube: 0.18.0
https://github.com/kubernetes/minikube
Not installed
From: https://github.com/caskroom/homebrew-cask/blob/master/Casks/minikube.rb
==> Name
Minikube
==> Artifacts
minikube-darwin-amd64 (binary)
```

bfontaine:
Let’s close this PR, then.

commitay:
@chases Feel free to open a new PR that builds `minikube` from source.

chases:
I notice that the Nomad formula (for example) was migrated from cask to core, would that be an acceptable solution here?

Formula/minikube.rb
chases: Sorry for my ignorance, couldn't we do something like `OS.mac? ? MAC_URL : LINUX_URL`. Or is there a reason to build from source other than OS compatibility?

commitay: The formula must build from source.

PR #12199, opened 2017-04-07
lapack: depend on cmake instead of vendoring it

ilovezfs:

-----

vendoring CMake 3.6.3 is no longer necessary now that the CMake
-lto_library bug is fixed.

See
https://github.com/Homebrew/homebrew-core/pull/12198
https://github.com/Reference-LAPACK/lapack/issues/139

CC @fxcoudert

### DIFF
--- a/Formula/lapack.rb
+++ b/Formula/lapack.rb
@@ -3,6 +3,7 @@ class Lapack < Formula
   homepage "http://www.netlib.org/lapack/"
   url "http://www.netlib.org/lapack/lapack-3.7.0.tgz"
   sha256 "ed967e4307e986474ab02eb810eed1d1adc73f5e1e3bc78fb009f6fe766db3be"
+  revision 1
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):

tenomoto:
Fortran build error occurs when MACOSX_DEPLOYMENT_TARGET is set. https://trac.macports.org/ticket/53565

ilovezfs:
@tenomoto since there are bottles, and we set MACOSX_DEPLOYMENT_TARGET, I'm not sure I follow.

ilovezfs:
@tenomoto so the reason for the discrepancy is that in Homebrew we were setting both  MACOSX_DEPLOYMENT_TARGET and SDKROOT. The bug only occurs if MACOSX_DEPLOYMENT_TARGET is set and SDKROOT is not set.

PR #12204, opened 2017-04-07
brew-pip: add test

doniarobinson:
Add test to brew-pip formula

### DIFF
--- a/Formula/brew-pip.rb
+++ b/Formula/brew-pip.rb
@@ -9,4 +9,8 @@ class BrewPip < Formula
   def install
     bin.install "bin/brew-pip"
   end
+
+  test do
+    system "#{bin}/brew-pip", "help"
+  end
 end


Formula/brew-pip.rb
```diff
@@ -9,4 +9,8 @@ class BrewPip < Formula
   def install
     bin.install "bin/brew-pip"
   end
+
+  test do
+    system "#{bin}/brew-pip", "help"
```
bfontaine: Can’t we write a test that does something more substantial than e.g. `--help` or `--version`?

MikeMcQuaid: @bfontaine Given there's no test here and `brew-gem` uses a similar format I think this is fine for now 👍

neutric:
Thanks for your contribution to Homebrew, @doniarobinson 🏆

PR #6022, opened 2016-10-17
zyre 1.2.0 (new formula)

colstrom:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)?
- [x] Have you checked that there aren't other open [pull requests](https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change?
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting?
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)?

---

Adds formula for current stable release of zyre, a library for
reliable group messaging over local area networks.

PR #12207, opened 2017-04-08
new formula: Zyre (zeromq project)

wesyoung:

-----

adding a Formula for the ZeroMQ 'Zyre' (peer to peer stack) messaging. 

 is a slightly extended take of closed [inactivity] PR: #6022. i tried to address the original intent and the followup concerns. this is my first PR to the project, i think i covered your req's but let me know if i missed something.

incidentally; the `zyre_test` function is a self-test function that actually spins up and tests the nodes end to end. it abstracts a lot of the necessarily C code from test stuff like this.

the only thing i couldn't really figure out was being able to specify a minimal version for the "depends_on" function(?). tried hunting down doc, but wasn't able to specify that you really need `zeromq >= 4.2.2` and `czmq >= 4.0.2` for zyre2 to work correctly. if i'm missing something stupid there, let me know that too and ill fix it. it didn't seem like you could specify that for "non special" tags.


Formula/zyre.rb
```diff
@@ -0,0 +1,61 @@
+# from https://github.com/Homebrew/homebrew-core/pull/6022/files
```
bfontaine: Please remove this comment.

Formula/zyre.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  head do
+    url "https://github.com/zeromq/zyre.git"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+  end
+
+  option "with-drafts", "Build and install draft classes and methods"
```
bfontaine: Is there a reason not to use this option?

Formula/zyre.rb
```diff
@@ -0,0 +1,61 @@
...
+  homepage "https://github.com/zeromq/zyre"
+  url "https://github.com/zeromq/zyre/releases/download/v2.0.0/zyre-2.0.0.tar.gz"
+  sha256 "8735bdf11ad9bcdccd4c4fd05cebfbbaea8511e21376bc7ad22f3cbbc038e263"
+
+  head do
+    url "https://github.com/zeromq/zyre.git"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
```
bfontaine: You can remove these three lines since they’re already required below anyway. Replace the whole block with `head "https://github.com/zeromq/zyre.git"`.

bfontaine (changes requested):

Formula/zyre.rb
wesyoung: done.

Formula/zyre.rb
wesyoung: good catch. done.

Formula/zyre.rb
wesyoung: yes.

wesyoung:
updated; ty for the feedback. :)

Formula/zyre.rb
bfontaine: And what is it? :slightly_smiling_face:

wesyoung: within the ZeroMQ "zproject" framework (a framework that auto-generates C based APIs based on XML doc, which generates the APIs and code around CZMQ and Zyre) there's the concept of a "draft API" which enables developers to include "alpha/beta" APIs within the various [stable] ZeroMQ frameworks. these API's are considered unstable, susceptible to change/breakage and are usually turned off by default in stable / production releases. so while messing with development it's nice to enable these API's (esp when working between the different underlying libraries such as CZMQ), in production mode they're generally defaulted to off so production code doesn't accidentally start using an API that's being tested.

there's a lot of C based code generation in the various ZMQ frameworks, this was a way to help automate some of the heavy lifting when it came to testing new APIs fast(er).

there's a lot more detail around how we do this here:

https://github.com/zeromq/zproject#draft-api-support

does that sorta make sense?

bfontaine: Sorta, yes :smile: Thanks for the detailed explanation!

bfontaine (approved):

Formula/zyre.rb
wesyoung: np. was trying to not get too far into the weeds, the zmq community has a bit of a "merge first ask questions later" sort of culture to it. draft APIs was a way to enable "merge faster" while at the same time provide some stability while the APIs got tested, sorted, etc. good stuff floats to stable and bad stuff eventually gets ripped out.

just a different way to do design, very specific to that culture. a "production" release is triggered when enough people clamour that "This DRAFT is correct, we're using it in production and we need it to be moved to STABLE". if it never gets that much traction, it eventually just gets removed. but a way to thwart usage by those who may not quite understand the difference in their early adoption.

fxcoudert:
Thanks @wesyoung for the pull request!

PR #12210, opened 2017-04-08
Bump Phoronix-Test-Suite to 7.0.1

BenjaminHCCarr:
Big jump from version `5.2.1` ~2012 to -current which is `7.0.1`
Minor tweaks were required; but formula now provides a fully functioning PTS 7.0.1 system. As well as testing.
-----


Formula/phoronix-test-suite.rb
```diff
@@ -1,8 +1,9 @@
 class PhoronixTestSuite < Formula
   desc "Automated testing framework"
   homepage "https://www.phoronix-test-suite.com/"
-  url "https://www.phoronix-test-suite.com/download.php?file=phoronix-test-suite-5.2.1"
-  sha256 "1186f460691e2fe7a07df5edb8d8ed1ac0c65327512e646da2b2e3a60dda6cd9"
+  version "7.0.1"
+  url "http://www.phoronix-test-suite.com/releases/phoronix-test-suite-#{version}.tar.gz"
```
fxcoudert: Include the real URL, and let `version` be automatically determined.

Formula/phoronix-test-suite.rb
```diff
@@ -29,38 +30,38 @@ def install
  mkdir -p $DESTDIR$INSTALL_PREFIX/share/doc/phoronix-test-suite/
 -mkdir -p $DESTDIR$INSTALL_PREFIX/../etc/bash_completion.d/
 +mkdir -p $DESTDIR$INSTALL_PREFIX/etc/bash_completion.d/
-
```
fxcoudert: Why is the patch still necessary? Has it been reported upstream? Has it been accepted/merged?

fxcoudert (changes requested):

Formula/phoronix-test-suite.rb
neutric: And keep the HTTPS, please.

Formula/phoronix-test-suite.rb
BenjaminHCCarr: Yes, this whole block is still needed to use the install script.
The other option would be to remove the bash-completion line and use `bash_completion.install` to do the completion install, but that would change the syntax more than the current patch.

Formula/phoronix-test-suite.rb
BenjaminHCCarr: Changed to https for download script

BenjaminHCCarr:
Still working, I seem to have broken the patching again.

Formula/phoronix-test-suite.rb
dunn: > The other option would be to remove the bash-completion line and use bash_completion.install to do the completion install

That sounds preferable to me.

fxcoudert:
ping?

BenjaminHCCarr:
🏓 

Sorry, sick kid, and an deadlined project... it's coming.

Will try to push an update by 2359 EST. Please don't close, our current version is hopelessly out of date.

joanwa:
As suggested by the creator, the EULA check can also be omitted via a patch:

https://www.phoronix.com/forums/forum/phoronix/phoronix-test-suite/936918-phoronix-test-suite-7-0-released

--- a/pts-core/objects/client/pts_client.php	2017-05-02 18:30:16.000000000 +0200
+++ b/pts-core/objects/client/pts_client.php	2017-05-02 18:31:53.000000000 +0200
@@ -803,6 +803,8 @@
 	}
 	public static function user_agreement_check($command)
 	{
+		return true;
+
 		$pso = pts_storage_object::recover_from_file(PTS_CORE_STORAGE);

 		if($pso == false)

and I also don't know why the "echo -e" execute flag has been canned. IMO this has to stay as is...

Formula/phoronix-test-suite.rb
BenjaminHCCarr: switching to `bash_completion.install "./pts-core/static/bash_completion"`

BenjaminHCCarr:
Sorry. I am having issues with the sed block. For some reason this works fine on my system, works fine on my tap. But in homebrew the \1 is escaping to unicode rather than ASCII:
https://github.com/Homebrew/homebrew-core/pull/12210/commits/2c1e9346267949c4a345ea62e1473cb2ca9e71f0#L23

The correct is:
`assert_match "Ringsaker", "#{bin}/phoronix-test-suite version | grep -v ^$ | sed 's/.*(\(.*\)).*/\1/'"`
What is expanding in Travis to:
`"/usr/local/Cellar/phoronix-test-suite/7.0.1/bin/phoronix-test-suite version | grep -v ^$ | sed 's/.*((.*)).*/\u0001/'"`
for some reason.

The unicode expansion breaks the test.

Note, the 5 year old formula had no `test` at all, so this simple fix by a committer/ reviewer/ maintainer bumps us five years in PTS, dozens of versions, and gains us a test block that is reproducible by swapping only the version name in the future; we only have to do this sed unicode malarchy once.

BenjaminHCCarr:
so this is the CI failure:
```
Error Message

failed: brew test phoronix-test-suite --verbose
Stacktrace

        Testing phoronix-test-suite
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20170531-31836-gdumdq.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/phoronix-test-suite.rb --verbose
Error: phoronix-test-suite: failed
Expected /Ringsaker/ to match "/usr/local/Cellar/phoronix-test-suite/7.0.1/bin/phoronix-test-suite version | grep -v ^$".
```
the command when run is:
```
tyr:~ benc$ /usr/local/Cellar/phoronix-test-suite/7.0.1/bin/phoronix-test-suite version | grep -v ^$
Phoronix Test Suite v7.0.1 (Ringsaker)
```
which is why I was doing the sed; it appears to me that `assert_match` isn't working as intended.

ilovezfs:
>it appears to me that assert_match isn't working as intended.

Given it's used 1,113 times in homebrew/core, that's probably not the case.

```
 -    assert_match "Ringsaker", "#{bin}/phoronix-test-suite version | grep -v ^$ | sed 's/.*(\(.*\)).*/\1/'" 
```

That does not say to execute `"#{bin}/phoronix-test-suite version | grep -v ^$ | sed 's/.*(\(.*\)).*/\1/'" `. It says to check for the substring `"Ringsaker"` anywhere in the string  `"#{bin}/phoronix-test-suite version | grep -v ^$ | sed 's/.*(\(.*\)).*/\1/'"`.

BenjaminHCCarr:
@ilovezfs you can see I got it to work, I needed the wrap my test in `shell_output()` which passed it.

http://www.rubydoc.info/github/Homebrew/brew/master/Formula#test-class_method

doesn't list an `assert_match` option at all; you can see why I was mucking with sed for so long.

I've done a bunch of PR's over time, and while I love that we keep iterating, the formulas are a moving target to everyone but committers basically.

the `shell_output()` is understandable, but is different syntax from other blocks when you can call `#{[s]bin]` directly.

That said I'm not the only one confused:
```
tyr:/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula (master) benc$ grep assert_match *.rb | grep bin | grep -v shell | grep -v pipe
cf.rb:    assert_match /Jan 20 00:35:44/, `echo 1074558944 | #{bin}/cf -u`
gitup.rb:    assert_match %r{#{Dir.pwd}/third}, `#{bin}/gitup --list`.strip
masscan.rb:    assert_match(/adapter =/, `#{bin}/masscan --echo | head -n 6 | tail -n 1`)
pstree.rb:    assert_match "#{bin}/pstree", lines[1]
sendemail.rb:    assert_match /sendemail-#{Regexp.escape(version)}/, `#{bin}/sendemail`.strip
zmap.rb:    assert_match /redis-csv/, `#{sbin}/zmap --list-output-modules` if build.with? "hiredis"
zmap.rb:    assert_match /mongo/, `#{sbin}/zmap --list-output-modules` if build.with? "mongo-c-driver"
```
admittedly that this represents just a few confused test blocks.

The multiple locations for documentation (no ones favorite thing to work on I know) makes it hard to know why you are failing:
http://docs.brew.sh/Formula-Cookbook.html (links to):
http://www.rubydoc.info/github/Homebrew/brew/master/Formula#test-class_method
which only offers `assert_equal`

It would also be useful to allow `test do` to fail on localhost rather than wait for the CI to flag it.
`audit` `style` `test` and `reinstall` all passing makes one think that it should pass CI; maybe this is an issue to be raised, if testing is only happening on merge and not on build from source.

Anyway, this one is good to go just as 7.0.2 is about to come out, but that should be be a three line change!

BenjaminHCCarr:
@fxcoudert  @dunn @neutric cc: @ilovezfs

Can we merge this into -core, it's a huge step up, and a 5yr leap in package target.

dunn:
Sorry for the delay; merged as 9e7171d89170b7ee63de3e128d741eff17841c90!

Issue #11536, opened 2017-03-24
Forced Python Install with gobject-introspection


blogabe:
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [X] Confirmed this is a problem with specific formulae and not Homebrew/brew? If it's a general Homebrew/brew problem please file this issue at https://github.com/Homebrew/brew/issues/new

May I ask why the change in commit https://github.com/Homebrew/homebrew-core/commit/1ab01953810fa04d65d0f961dc34898077f0ec7f#diff-626b321c06d31d0bffed408d358c97cf to force homebrew's python over system python?

I'd prefer to use system python and am unsure why specifically v1.52 requires a more modern python.

neutric:
Does [this comment](https://github.com/Homebrew/homebrew-core/pull/11464#discussion_r107407934) by @tschoonj answer your question?

blogabe:
It does, thank you.  Went looking for the gobject-introspection pull request by name and couldn't find it.  Didn't realize it was part of a broader change related to glib.

I agree with the question, it does seem a bit heavy-handed to require, and I now understand the reason (although I don't personally like it).

blogabe:
Reason I created this issue in the first place is because homebrew can't link python upon installation:
```
>~$ brew upgrade
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (caskroom/cask).
No changes to formulae.

==> Upgrading 1 outdated package, with result:
gobject-introspection 1.52.0
Error: You must `brew link python` before gobject-introspection can be installed
>:~$ brew link python
Linking /usr/local/Cellar/python/2.7.13... Error: Permission denied - /usr/local/Frameworks
```

tschoonj:
I have just opened #11538, which hardcodes Homebrew's python full path to the executable in `gobject-introspection`'s executables' shebangs.

So even if you don't like the brewed python, you can still unlink it, and it will still get used by `gobject-introspection`.

Why the linking doesn't work on your system I cannot say though: looks like you cannot write to /usr/local...

ilovezfs:
@tschoonj I'm wondering if we can intervene with a shim to prevent the universal-whatever rather than resorting to full replacement. This doesn't seem like a particularly surgical approach.

fxcoudert:
There are other formulas that work around the universal python issue, IIRC

tschoonj:
[...]

PR #12226, opened 2017-04-08
gobject-introspection: do not require brew python

ilovezfs:

-----

Reverts #11538.

### DIFF
--- a/Formula/gobject-introspection.rb
+++ b/Formula/gobject-introspection.rb
@@ -15,9 +15,6 @@ class GobjectIntrospection < Formula
   depends_on "glib"
   depends_on "cairo"
   depends_on "libffi"
-  # never switch back to system python!
-  # https://github.com/Homebrew/homebrew-core/pull/11464#discussion_r107407934
-  depends_on "python"
 
   resource "tutorial" do
     url "https://gist.github.com/7a0023656ccfe309337a.git",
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @tschoonj @fxcoudert 

Closes https://github.com/Homebrew/homebrew-core/issues/11536.

JCount:
Note, this still requires brewed python.

JCount:
Note that this still depends upon brewed python.

ilovezfs:
fixed.

tschoonj:
Hmmm... is there already a fix to strip those universal arch build flags from the Python config output?

ilovezfs:
Nope, but since this passes CI, I'm still not sure I've understood the problem.

tdsmith:
AIUI the extra arch flags won't be a problem in Homebrew because superenv strips them but they can cause trouble if you're building things outside of Homebrew using Homebrew-installed stuff. Is that right?

ilovezfs:
If the only reason for the dependency is things outside of Homebrew, I think we should remove it.

tschoonj:
IIRC the problems that forced me to use the brewed python occurred when trying to port an existing to formula to the meson build system, not while building something outside of Homebrew.

ilovezfs:
@tschoonj if you can recall which one that is, I'll take a look.

tschoonj:
Try graphene

ilovezfs:
@tschoonj thanks

ilovezfs:
`graphene` builds fine with system python plus `meson` if we use `ENV.refurbish_args`. (The current `graphene` formula is already effectively doing that because it uses `make`, which causes brew to apply `ENV.refurbish_args` automatically, whereas using `meson` does not.)

tschoonj:
Is there a fix now ensuring that `meson` based formulas use `ENV.refurbish_args`?

ilovezfs:
Nope. We only do that for make. PRs welcome.

tschoonj:
Any pointers on where to find the relevant code? I am finding a reference in `formula.rb` but that only appears to kick in for python commands...

brion:
This causes g-ir-scanner to always fail when compiling anything using glib.h, as in https://gitlab.gnome.org/GNOME/librsvg/issues/212. I would tend to recommend restoring the use of a brew-installed python2.

PR #12236, opened 2017-04-08
readline: restore universal

georgebrock:

-----

This reverts commit ae391d59dc47007786c561fd04592fba40820d05.

Having a universal version of libreadline is useful when linking against macOS system files, many of which are still universal binaries. For example, building C extensions for the system versions of Ruby or Python fails with a non-universal readline build.

This PR was motivated by https://github.com/thoughtbot/gitsh/issues/287, which reported that the [gitsh formula][1] had suddenly stopped working. I discovered #10907 had changed the default build architecture of readline, which is what was causing the formula to fail.

As an alternative to merging this PR, I could include ruby as a dependency in the gitsh formula; the ruby formula is non-universal, so it has no problem with a non-universal readline library. I thought it was worthwhile trying this PR first, since gitsh is unlikely to be the only program affected by the readline formula's architecture change.

[1]: https://github.com/thoughtbot/homebrew-formulae/blob/master/Formula/gitsh.rb

### DIFF
--- a/Formula/readline.rb
+++ b/Formula/readline.rb
@@ -5,7 +5,7 @@ class Readline < Formula
   mirror "https://ftp.gnu.org/gnu/readline/readline-7.0.tar.gz"
   version "7.0.3"
   sha256 "750d437185286f40a369e1e4f4764eda932b9459b5ec9a731628393dd3d32334"
-  revision 1
+  revision 2
 
   bottle do
     cellar :any
@@ -33,6 +33,7 @@ class Readline < Formula
   EOS
 
   def install
+    ENV.universal_binary
     system "./configure", "--prefix=#{prefix}"
     system "make", "install"
   end


fxcoudert:
I do not understand this statement: _“Having a universal version of libreadline is useful when linking against macOS system files”_. We have many formulas that link against macOS system libraries, with no problem whatsoever: the fact that macOS libraries are universal _allows_ you to link universal binaries against them, but it never _forces_ you to only link universal binaries. That's how all of our current Homebrew software works.

If you want a build system with universal support, you can host a `readline-universal` formula in a tap. Or vendor the universal readline inside the formula (the way we do it for `wine`). Or include readline inside the gitsh build process.

georgebrock:
> he fact that macOS libraries are universal allows you to link universal binaries against them, but it never forces you to only link universal binaries

Sounds like I'm missing something on my side. I tried various ways of building my Ruby C extension with a universal Ruby and a x86_64-only readline, and so far everything failed. I'll go read the Ruby extension configuration code more closely, and see what I'm missing.

georgebrock:
Yep, my fault. I was missing the `--with-arch-flag` option for Ruby `mkmf` scripts, which triggers some specific behaviour beyond what you get when you pass `--with-cppflags="-arch x86_64" --with-ldflags="-arch x86_64"`.

Sorry for the spurious PR. I'll update my formula to handle a non-universal Readline correctly.

Thanks for pointing me in the right direction.

PR #12251, opened 2017-04-09
haproxy: add lua support

dmvk:


Formula/haproxy.rb
```diff
@@ -11,8 +11,11 @@ class Haproxy < Formula
     sha256 "07eb7eaf898e5855614d8da183a9fca3250cae0cbb3b7eeec957a6cac94aed60" => :yosemite
   end
 
+  option "with-lua@5.3", "Build with lua support"
+
   depends_on "openssl"
   depends_on "pcre"
+  depends_on "lua@5.3" if build.with? "lua@5.3"
```
DomT4: `depends_on "lua@5.3" => :optional` is what you're looking for, combined with dropping the option wording above. I think Homebrew is trying to avoid using `@` dependencies for now, but I'll let ILZ chime in on that.

Formula/haproxy.rb
```diff
@@ -24,6 +27,13 @@ def install
       USE_ZLIB=1
       ADDLIB=-lcrypto
     ]
+    if build.with? "lua@5.3"
+      args << "USE_LUA=1"
+      args << "LUA_LIB_NAME=lua.5.3"
+      args << "LUA_LIB=#{Formula["lua@5.3"].lib}"
```
DomT4: `opt_lib`, `opt_include`, and so on. Surprised these are necessary once you tell it what version of `lua` you want it to use though, most build scripts are pretty smart in terms of handling the potential for multiple `lua`s sitting around.

Formula/haproxy.rb
dmvk: 👍

Formula/haproxy.rb
dmvk: Unfortunately lua@5.3 is little bit non-standard (include path and lib version suffix), only property i was able to omit is `LUA_LIB`.

`LUA_LD_FLAGS` needs to be replaced because `--export-dynamic` flag does not seem to be supported on OS X.

https://github.com/haproxy/haproxy/blob/25f067ccec52f53b0248a05caceb7841a3cb99df/Makefile#L605

PR #12260, opened 2017-04-09
sc-im 0.5.0 (new formula)

jez:
- [x] Have you followed the [guidelines for contributing][guidelines]?
- [x] Have you checked that there aren't other open [pull requests][pulls] for the same formula update/change?

[guidelines]: https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md
[pulls]: https://github.com/Homebrew/homebrew-core/pulls

- - -

One thing to note is that `sc-im` doesn't pass this check:

```
❯ brew audit --new-formula sc-im
sc-im:
  * Dependency 'ncurses' may be unnecessary as it is provided by macOS; try to build this formula without it.
  * Dependency 'ncurses' may be unnecessary as it is provided by macOS; try to build this formula without it.
```

When I built it against the macOS default ncurses, there were some
visual glitches where the screen wouldn't always reflow correctly. I'm
not sure if I'm supposed to silence this error by modifying the formula
somehow, or if it's fine as it is.


Formula/sc-im.rb
```diff
@@ -0,0 +1,20 @@
...
+  sha256 "d7a31c5225e02239e925b50b414d7e69d12bc3554f218621823782872ccc5e4d"
+  head "https://github.com/andmarti1424/sc-im.git"
+
+  depends_on "ncurses"
+
+  def install
+    cd "src" do
+      system "make", "prefix=#{prefix}", "LDFLAGS=-L#{Formula["ncurses"].lib}"
+      system "make", "prefix=#{prefix}", "install"
+    end
```
bfontaine: You can use `system "make", "-C", "src", …` and avoid this `cd "src" do` block.

jez: I actually looked into what Homebrew's preference was regarding using a `cd` block vs a `-C` flag. Based on some rough grepping, it seemed like `cd` was used about twice as often as `-C` when calling `make` more than once in that directory.

I'm fine changing it, but I thought the way it's written currently was more idiomatic Homebrew and more readable.

Formula/sc-im.rb
```diff
@@ -0,0 +1,20 @@
+class ScIm < Formula
+  desc "Spreadsheet program for the terminal, using ncurses"
+  homepage "https://github.com/andmarti1424/sc-im"
+  url "https://github.com/andmarti1424/sc-im/archive/v0.5.0.tar.gz"
+  sha256 "d7a31c5225e02239e925b50b414d7e69d12bc3554f218621823782872ccc5e4d"
+  head "https://github.com/andmarti1424/sc-im.git"
```
neutric: This needs a `, :branch => "freeze"` appended.

neutric (changes requested):

Formula/sc-im.rb
jez: Good catch; should be fixed now.

Formula/sc-im.rb
bfontaine: cc @MikeMcQuaid for thought here.

jez:
Hmm, I'm having some issues with the tests. It's saying something like "Device not configured". `brew audit sc-im` works for me locally.

I have a hunch that it's related to the fact that `scim` is designed to only be run interactively. There's no `--version` or `--help` option that just runs and quits right after.

I have two ideas:

- Are there any other formulas that are interactive-only that I could use as a reference to design a test block?
- Would it be possible to omit the test, or check instead for the existence of the binary, instead of actually running it?

dunn:
Some formulae like [`ne.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ne.rb) use `script` to fake a TTY; for providing input other formulae use `Open3`: [`euler-py.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/euler-py.rb)

claui:
@jez The `scim` binary [accepts a script from standard input](https://github.com/jez/homebrew-core/pull/1/files#diff-866c4a23cfb68fb183dbcdb1d60b94d2R20).  
To inject it, we have `Homebrew::Assertions#pipe_output`.

I have updated the test accordingly, and [double-parked a small PR](https://github.com/jez/homebrew-core/pull/1) which can be merged into this one.

jez:
Oh cool; I'll take a look at it then.

Know that I'm quite busy for the next week and a half (final exams and
graduation!) but I'll try to sneak in some time for it.

Formula/sc-im.rb
MikeMcQuaid: I'm fine with either.

fxcoudert:
@BrewTestBot test this please

fxcoudert:
And 👍 to merging @claui 's PR for the test, it's exactly what we need

jez:
I'm no longer using this program. Someone who is still using sc-im might want to recreate this PR with passing tests, but for now I'm signing off.

PR #12264, opened 2017-04-10
sbt 0.13.15

eed3si9n:

-----

```
$ brew install --build-from-source sbt
==> Using the sandbox
==> Downloading https://dl.bintray.com/sbt/native-packages/sbt/0.13.15/sbt-0.13.15.tgz
Already downloaded: /Users/eugene/Library/Caches/Homebrew/sbt-0.13.15.tgz
==> Caveats
You can use $SBT_OPTS to pass additional JVM options to SBT:
   SBT_OPTS="-XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=256M"

This formula is now using the standard lightbend sbt launcher script.
Project specific options should be placed in .sbtopts in the root of your project.
Global settings should be placed in /usr/local/etc/sbtopts
==> Summary
🍺  /usr/local/Cellar/sbt/0.13.15: 377 files, 63.3MB, built in 1 second
$ brew audit --strict sbt
$
```

### DIFF
--- a/Formula/sbt.rb
+++ b/Formula/sbt.rb
@@ -21,6 +21,8 @@ class Sbt < Formula
     end
 
     inreplace "bin/sbt-launch-lib.bash" do |s|
+      # Upstream issue "Replace realpath with something Mac compatible"
+      # Reported 10 Apr 2017 https://github.com/sbt/sbt-launcher-package/issues/149
       s.gsub! "$(dirname \"$(realpath \"$0\")\")", "#{libexec}/bin"
       s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
 
--- a/Formula/sbt.rb
+++ b/Formula/sbt.rb
@@ -24,9 +24,9 @@ class Sbt < Formula
       s.gsub! "$(dirname \"$(realpath \"$0\")\")", "#{libexec}/bin"
       s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
 
-      ## See https://github.com/sbt/sbt-launcher-package/issues/150
-      ## This is a workaround for `brew test sbt` failing to detect java -version.
-      if s.include?("[[ \"$java_version\" > \"8\" ]]")
+      # Workaround for `brew test sbt` failing to detect java -version
+      # Reported 10 Apr 2017 https://github.com/sbt/sbt-launcher-package/issues/150
+      if build.stable?
         s.gsub! "[[ \"$java_version\" > \"8\" ]]", "[[ \"$java_version\" == \"9\" ]]"
       end
     end
--- a/Formula/sbt.rb
+++ b/Formula/sbt.rb
@@ -23,6 +23,12 @@ class Sbt < Formula
     inreplace "bin/sbt-launch-lib.bash" do |s|
       s.gsub! "$(dirname \"$(realpath \"$0\")\")", "#{libexec}/bin"
       s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
+
+      ## See https://github.com/sbt/sbt-launcher-package/issues/150
+      ## This is a workaround for `brew test sbt` failing to detect java -version.
+      if s.include?("[[ \"$java_version\" > \"8\" ]]")
+        s.gsub! "[[ \"$java_version\" > \"8\" ]]", "[[ \"$java_version\" == \"9\" ]]"
+      end
     end
 
     libexec.install "bin"
--- a/Formula/sbt.rb
+++ b/Formula/sbt.rb
@@ -26,8 +26,8 @@ class Sbt < Formula
     end
 
     libexec.install "bin"
-    etc.install "conf/sbtopts"
     libexec.install "lib" if build.stable? # remove `if` when devel > 1.0.0-M4
+    etc.install "conf/sbtopts"
 
     (bin/"sbt").write <<-EOS.undent
       #!/bin/sh
--- a/Formula/sbt.rb
+++ b/Formula/sbt.rb
@@ -31,10 +31,7 @@ class Sbt < Formula
 
     libexec.install "bin"
     etc.install "conf/sbtopts"
-    ## if should be removed when devel is > 1.0.0-M4
-    if build.stable?
-      libexec.install "lib"
-    end
+    libexec.install "lib" if build.stable? # remove `if` when devel > 1.0.0-M4
 
     (bin/"sbt").write <<-EOS.undent
       #!/bin/sh
--- a/Formula/sbt.rb
+++ b/Formula/sbt.rb
@@ -23,10 +23,6 @@ class Sbt < Formula
     inreplace "bin/sbt-launch-lib.bash" do |s|
       s.gsub! "$(dirname \"$(realpath \"$0\")\")", "#{libexec}/bin"
       s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
-      ## This is required to pass the test
-      if s.include?("[[ \"$java_version\" > \"8\" ]]")
-        s.gsub! "[[ \"$java_version\" > \"8\" ]]", "[[ \"$java_version\" == \"9\" ]]"
-      end
     end
 
     libexec.install "bin"
--- a/Formula/sbt.rb
+++ b/Formula/sbt.rb
@@ -21,15 +21,9 @@ class Sbt < Formula
     end
 
     inreplace "bin/sbt-launch-lib.bash" do |s|
+      s.gsub! "$(dirname \"$(realpath \"$0\")\")", libexec
+      s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
       s.gsub! "${sbt_home}/bin/sbt-launch.jar", "#{libexec}/sbt-launch.jar"
-      if s.include?("${sbt_bin_dir}/java9-rt-export.jar")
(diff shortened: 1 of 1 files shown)


ilovezfs (dismissed):
Please explain in prose why all of these inreplace changes are needed so that we can find a more sustainable way of handling this formula.

Formula/sbt.rb
```diff
@@ -20,11 +20,30 @@ def install
       s.gsub! "/etc/sbt/sbtopts", "#{etc}/sbtopts"
     end
 
-    inreplace "bin/sbt-launch-lib.bash", "${sbt_home}/bin/sbt-launch.jar", "#{libexec}/sbt-launch.jar"
+    inreplace "bin/sbt-launch-lib.bash" do |s|
+      s.gsub! "${sbt_home}/bin/sbt-launch.jar", "#{libexec}/sbt-launch.jar"
+      if s.include?("${sbt_bin_dir}/java9-rt-export.jar")
+        s.gsub! "${sbt_bin_dir}/java9-rt-export.jar", "#{libexec}/java9-rt-export.jar"
```
eed3si9n: I'm following the existing pattern of calling `inreplace` to change where `sbt_home` (or `sbt_bin_dir`) appear in the original bash script. These are calculated in the original as follows:

```
declare -r sbt_bin_dir="$(dirname "$(realpath "$0")")"
declare -r sbt_home="$(dirname "$sbt_bin_dir")"
```

I'm guessing these are needed because the bash files and jar files exist directly under `libexec` while normal `sbt` directory would have `bin`, `config`, and `lib` directory under it.

eed3si9n:
Hi, I'm one of the maintainers of sbt/sbt.

sbt 0.13.15 adds JDK9 support and also the tgz package now includes necessary jar files (instead of downloading from Maven Central on the first use). These two changes were implemented partially in the bash script, so that resulted to all the changes. The root cause I think is the layout difference between `bin`/`config`/`lib` and `libexec`/`libexec/lib`. In the future version of sbt we can add some variable in bash script to abstract these differences.

ilovezfs:
Note the `realpath` command isn't available on macOS  (unless you've run `brew install coreutils` but that would be a very heavy dependency to add).

eed3si9n:
I tried to simplify it a bit by substituting the RHS of `sbt_bin_dir` and `sbt_home` to be `"#{libexec}"`.

```
declare -r sbt_bin_dir="/usr/local/Cellar/sbt/0.13.15/libexec"
declare -r sbt_home="/usr/local/Cellar/sbt/0.13.15/libexec"
```

This is more or less what I was thinking.

ilovezfs:
Would simplify it further to allow some things to be in libexec/bin?

eed3si9n:
> Would it simplify it further to allow some things to be in libexec/bin?

That would probably make the file copying simpler for sure.

Formula/sbt.rb
```diff
@@ -20,11 +20,24 @@ def install
...
+        s.gsub! "[[ \"$java_version\" > \"8\" ]]", "[[ \"$java_version\" == \"9\" ]]"
+      end
+    end
 
-    libexec.install "bin/sbt", "bin/sbt-launch-lib.bash", "bin/sbt-launch.jar"
+    libexec.install "bin/sbt", "bin/sbt-launch-lib.bash"
+    libexec.install Dir["bin/*.jar"]
     etc.install "conf/sbtopts"
 
+    if File.directory?("lib")
```
ilovezfs: Why `if`?

Formula/sbt.rb
```diff
@@ -20,11 +20,24 @@ def install
       s.gsub! "/etc/sbt/sbtopts", "#{etc}/sbtopts"
     end
 
-    inreplace "bin/sbt-launch-lib.bash", "${sbt_home}/bin/sbt-launch.jar", "#{libexec}/sbt-launch.jar"
+    inreplace "bin/sbt-launch-lib.bash" do |s|
+      s.gsub! "$(dirname \"$(realpath \"$0\")\")", libexec
+      s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
+      s.gsub! "${sbt_home}/bin/sbt-launch.jar", "#{libexec}/sbt-launch.jar"
+      ## This is required to pass the test
```
ilovezfs: Which test?

Formula/sbt.rb
eed3si9n: devel version `sbt-1.0.0-M4` doesn't come with `lib` yet.

ilovezfs: So it specifically is for stable only? In that case you can make it `if build.stable?` and add a comment that the if should be removed when devel is > 1.0.0-M4.

Formula/sbt.rb
eed3si9n: `$ brew test sbt` that calls `("#{bin}/sbt sbt-version")` fails to detect the java version correctly. It prints out

```
Picked up _JAVA_OPTIONS: -Dsbt.log.noformat=true -Duser.home=/Users/eugene/Library/Caches/Homebrew/java_cache
```

Formula/sbt.rb
```diff
@@ -20,18 +20,29 @@ def install
...
+        s.gsub! "[[ \"$java_version\" > \"8\" ]]", "[[ \"$java_version\" == \"9\" ]]"
+      end
+    end
 
-    libexec.install "bin/sbt", "bin/sbt-launch-lib.bash", "bin/sbt-launch.jar"
+    libexec.install "bin"
     etc.install "conf/sbtopts"
+    ## if should be removed when devel is > 1.0.0-M4
+    if build.stable?
+      libexec.install "lib"
```
ilovezfs: you can do this on one line
```
     libexec.install "lib" if build.stable? # remove `if` when devel > 1.0.0-M4
```

Formula/sbt.rb
```diff
@@ -20,18 +20,26 @@ def install
       s.gsub! "/etc/sbt/sbtopts", "#{etc}/sbtopts"
     end
 
-    inreplace "bin/sbt-launch-lib.bash", "${sbt_home}/bin/sbt-launch.jar", "#{libexec}/sbt-launch.jar"
+    inreplace "bin/sbt-launch-lib.bash" do |s|
+      s.gsub! "$(dirname \"$(realpath \"$0\")\")", "#{libexec}/bin"
+      s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
+      ## This is required to pass the test
```
ilovezfs: The test appears to pass without this.

ilovezfs:
@eed3si9n CI passed without the replacement for the java version. Nonetheless, is that an upstream change that needs to be made for the next version, regardless? I see it does produce an undesirable error to be printed so I'm not necessarily against including a fix for it, but as you can see the reason for including it wouldn't be that it's needed for the test block to pass.

Formula/sbt.rb
```diff
@@ -20,9 +20,13 @@ def install
       s.gsub! "/etc/sbt/sbtopts", "#{etc}/sbtopts"
     end
 
-    inreplace "bin/sbt-launch-lib.bash", "${sbt_home}/bin/sbt-launch.jar", "#{libexec}/sbt-launch.jar"
+    inreplace "bin/sbt-launch-lib.bash" do |s|
+      s.gsub! "$(dirname \"$(realpath \"$0\")\")", "#{libexec}/bin"
```
ilovezfs: Is there an upstream issue tracking "abstracting" this? Note that `realpath` won't work on macOS regardless of Homebrew since it's not even a command in the BSD userland, only a C function

```
     #include <stdlib.h>

     char *
     realpath(const char *restrict file_name, char *restrict resolved_name);
```

eed3si9n:
>  CI passed without the replacement for the java version. Nonetheless, is that an upstream change that needs to be made for the next version, regardless?

Probably yes.

Formula/sbt.rb
eed3si9n: Now there is - https://github.com/sbt/sbt-launcher-package/issues/149

ilovezfs: Thanks.

ilovezfs:
>Probably yes.

Do you mind creating an issue for that as well, and re-adding your fix along with a ruby comment with a link to the issue? Thanks!

eed3si9n:
> Do you mind creating an issue for that as well, and re-adding your fix along with a ruby comment with a link to the issue? 

Done

ilovezfs:
Thanks.

Formula/sbt.rb
```diff
@@ -20,9 +20,19 @@ def install
...
+      s.gsub! "$(dirname \"$sbt_bin_dir\")", libexec
 
-    libexec.install "bin/sbt", "bin/sbt-launch-lib.bash", "bin/sbt-launch.jar"
+      ## See https://github.com/sbt/sbt-launcher-package/issues/150
+      ## This is a workaround for `brew test sbt` failing to detect java -version.
+      if s.include?("[[ \"$java_version\" > \"8\" ]]")
```
ilovezfs: Why the `if`?

eed3si9n: Could be `if build.stable?`

ilovezfs: Yes, that sounds better.

ilovezfs (approved):

ilovezfs:
@eed3si9n Shipped. Thanks for the pull request! :rocket:

eed3si9n:
Thanks for your help!

PR #12286, opened 2017-04-10
stone-soup: Added option to build without Lua

ronmrdechai:
Formula requires Lua@5.1 which is outdated. Added option to build
without Lua as Lua is only required for specific features.

-----

### DIFF
--- a/Formula/stone-soup.rb
+++ b/Formula/stone-soup.rb
@@ -13,7 +13,7 @@ class StoneSoup < Formula
   end
 
   option "with-tiles", "Enable graphic tiles and sound"
-  option "without-lua", "Disable Lua bindings for user scripts"
+  option "without-lua@5.1", "Disable Lua bindings for user scripts"
 
   depends_on "pkg-config" => :build
   depends_on "pcre"
@@ -27,7 +27,7 @@ class StoneSoup < Formula
   end
 
   if build.with? "lua"
-    depends_on "lua@5.1"
+    depends_on "lua@5.1" => :recommended
   end
 
   needs :cxx11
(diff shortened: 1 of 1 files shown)


Formula/stone-soup.rb
```diff
@@ -13,9 +13,9 @@ class StoneSoup < Formula
   end
 
   option "with-tiles", "Enable graphic tiles and sound"
+  option "without-lua", "Disable Lua bindings for user scripts"
```
DomT4:
```ruby
option "without-lua@5.1", "Disable Lua bindings for user scripts"
```
And then:
```ruby
depends_on "lua@5.1" => :optional
```

Formula/stone-soup.rb
```diff
@@ -26,6 +26,10 @@ class StoneSoup < Formula
     depends_on "freetype"
   end
 
+  if build.with? "lua"
```
DomT4: You can kill the `if build.with? "lua"` and `end`. Just the `depends_on "lua@5.1" => :recommended` needs to remain with the tweaked option text.

Formula/stone-soup.rb
```diff
@@ -53,6 +53,12 @@ def install
         args << "SOUND=y"
       end
 
+      if build.with? "lua"
```
DomT4: And `if build.with? "lua@5.1"` down here too. That should be the last change necessary in terms of correct style/usage. After that this should be fine for a maintainer to review for acceptability in general.

DomT4 (approved):
LGTM. Up to ILZ/a maintainer to make a decision about actual acceptability rather than code style now.

ilovezfs:
@ronmrdechai thanks for the PR. Shipped!

@DomT4 thanks for reviewing this.

DomT4:
@ilovezfs No worries at all ❤️

PR #12289, opened 2017-04-10
dnscrypt-proxy: enable plugins by default

DomT4:

-----

Since https://github.com/Homebrew/homebrew-core/commit/dde7062cbb5cdd2b073dd185406b062179fdd036 Homebrew has been using the default configuration file instead of passing arguments directly through the plist. This default configuration file assumes you're building `dnscrypt-proxy` with plugin support. This keeps causing confusion upstream with users, so let's just change it so everything works out of the box.

Ref: https://github.com/jedisct1/dnscrypt-proxy/issues/539#issuecomment-291905685

### DIFF
--- a/Formula/dnscrypt-proxy.rb
+++ b/Formula/dnscrypt-proxy.rb
@@ -3,7 +3,8 @@ class DnscryptProxy < Formula
   homepage "https://dnscrypt.org"
   url "https://github.com/jedisct1/dnscrypt-proxy/archive/1.9.4.tar.gz"
   sha256 "a79d5da0133344d38f8b3d3355c16269f11c15fbeedd0521e1a657b00ac503bb"
-  revision 1
+  revision 2
+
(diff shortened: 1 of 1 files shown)


DomT4:
@ilovezfs Can you take a look at this when you get time? ❤️. Think it got buried under the PR list.

ilovezfs:
Any reason to think anyone will actually pass the new option?

DomT4:
Doubtful. If you do you're basically promising you'll edit the default configuration, which well, a reasonably small amount of people bother to do I suspect.

ilovezfs (approved):
OK. We can always remove the option later if no one uses it.

ilovezfs:
:electric_plug:

DomT4:
> We can always remove the option later if no one uses it.

Aye, if you don't see many people using it by the next upstream release I'd probably kill it then, unless upstream cut a new release tomorrow or something.

PR #12290, opened 2017-04-10
influxdb: fix duplicate 'v' of version name

darklore:

-----
Fix head "vv" of version number in output of "influxd version" command.

```
$ influxd version
InfluxDB vv1.2.2 (git: master 1bcf3ae74c6b9c4897dab68d513d056277eb24f7)
```

And also fix a URL to influxdb documentation page embedded in influxd admin page (localhost:8083)  as follows.

```
https://docs.influxdata.com/influxdb/vv1.2/introduction/getting_started/
```
to
```
https://docs.influxdata.com/influxdb/v1.2/introduction/getting_started/
```


Formula/influxdb.rb
```diff
@@ -23,6 +23,11 @@ def install
     revision = `git rev-parse HEAD`.strip
     version = `git describe --tags`.strip
 
+    # "main.version" must have version number only.
+    if version[0] == "v"
+      version = version[1..-1]
+    end
```
bfontaine: You can fix this by using `version "1.2.2"` at the top of the formula.

ilovezfs: will that break HEAD builds?

bfontaine: HEAD builds shouldn’t be affected by this; their "version" looks like `HEAD-<commit>`.

ilovezfs: They would be affected though since currently `version` is set to
```
version = `git describe --tags`.strip
```
for stable and HEAD alike in this formula.

bfontaine: Right, I missed that line.

PR #12291, opened 2017-04-10
add maven@3.3 based on previous maven

freebeer42:
- [x] Have you followed the [guidelines for contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md)? Yes
- [x] Have you checked that there aren't other open [pull requests] (https://github.com/Homebrew/homebrew-core/pulls) for the same formula update/change? Yes
- [x] Have you built your formula locally with `brew install --build-from-source <formula>`, where `<formula>` is the name of the formula you're submitting? Yes
- [x] Does your build pass `brew audit --strict <formula>` (after doing `brew install <formula>`)? Yes

-----

The maven recently switched from 3.3.9 to 3.5.0. This formula allows users to install version 3.3.9 again. The contents are based on maven.rb before adding 3.5.0 alpha as a "devel", plus specifying 1.7+ as the Java dependency (which is true starting with maven 3.3). I also remove the conflicts_with as directed by the audit.

### DIFF
--- a/Formula/maven@3.3.rb
+++ b/Formula/maven@3.3.rb
@@ -1,0 +1,52 @@
+class MavenAT33 < Formula
+  desc "Java-based project management"
+  homepage "https://maven.apache.org/"
+  url "https://www.apache.org/dyn/closer.cgi?path=maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz"
+  mirror "https://archive.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
Does 3.5 actually not work for your project?

freebeer42:
I haven't checked. It's out of scope for me at this point for a few reasons, and I thought others might also want/need to stay on 3.3 until they are able to switch to the latest version. If not, I can live without this.

ilovezfs:
OK, I'm fine with shipping this regardless and we'll see whether it ends up getting any significant use.

ilovezfs:
@freebeer42 Shipped. Thanks for the PR!

PR #12306, opened 2017-04-11
ledger: add alias for ledger@3.1

geoffmacintosh:

-----
This fixes the `brew audit --strict ledger` error asking for an alias. I didn't build the formula from source to test the fix, but that shouldn't matter, right?

### DIFF
--- a/Aliases/ledger@3.1
+++ b/Aliases/ledger@3.1
@@ -1,0 +1,1 @@
+../Formula/ledger.rb


geoffmacintosh:
Sorry to add to your workload @neutric, but it says you modified the pull request, and I cannot figure out what was changed, especially since I deleted my branch with reckless abandon. I think there's a handful more of these that I can fix in the future, so if there's something I can do to make it less work to merge, that would be good to know. Thanks!

neutric:
@geoffmacintosh I modified the commit msg and PR title to "ledger: add alias for ledger@3.1".

For simple version bumps, the preferred commit message format is e.g. `cassandra 3.9` or e.g. `dub 1.3.0-rc.1 (devel)` if it concerns not the stable but the development version. All other things are `<formula name>: <short description of change>` when they concern just one formula or free form like `Import homebrew/dupes into core` if multiple formulae are changed in one PR. Special case: when the revision counter is increased because of a new version of a dependency, then we use e.g. `nvc: revision for llvm`. That should cover most cases. It doesn't hurt to consult the [commit log](https://github.com/Homebrew/homebrew-core/commits/master) if in doubt.

geoffmacintosh:
Thanks for the explanation! I'll try to use your guide as a reference for future commit messages.

PR #12336, opened 2017-04-12
swift 3.1

JCount:
This includes the addition of a number of somewhat
optional components in an effort to encompass more of the
feature set of the standard release toolchain.

-----

**NOTE:** 
This formula will fail `brew audit`, and this is expected. The missing dylibs mentioned as the reason for failure are a part of the toolchain for compiling for non-macOS targets, such as iOS. The two missing frameworks, UIKit and AssetsLibrary, are simply not components of macOS.

-----
cc @ilovezfs

### DIFF
--- a/Formula/swift.rb
+++ b/Formula/swift.rb
@@ -2,45 +2,47 @@ class Swift < Formula
   desc "High-performance system programming language"
   homepage "https://github.com/apple/swift"
 
+  # This formula is expected to have broken/missing linkage to
+  # both UIKit.framework and AssetsLibrary.framework. This is
+  # simply due to the nature of Swift's SDK Overlays.
   stable do
-    url "https://github.com/apple/swift/archive/swift-3.0.2-RELEASE.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs (approved):

ilovezfs:
@woodruffw any ideas on how to "fix" the audit here? The best I can come up with is white-listing, but ew.

woodruffw:
We could just remove those dylib load commands, although I'm a bit confused: are those dylibs not *available& on OS X, or are they available but part of the iOS toolchain?

JCount:
Yes

woodruffw:
> Yes

To which one? :wink:

JCount:
See edit. 😆

JCount:
To make certain this wasn't the result of my doing something incorrectly, I downloaded and installed the toolchain directly from swift.org, but it had exactly the same "broken" linkage.

JCount:
So, what is being identified as "broken" or "missing" linkage is actually correct and intentional, and is simply part of Swift's stdlib's SDK overlay.

PR #12351, opened 2017-04-12
httplab 0.2.1 (new formula)

m1ome:

-----


Formula/httplab.rb
```diff
@@ -0,0 +1,21 @@
...
+    ENV["GOPATH"] = buildpath
+    (buildpath/"src/github.com/gchaincl/httplab").install buildpath.children
+    cd "src/github.com/gchaincl/httplab" do
+      system "go", "build", "-o", bin/"httplab"
+      prefix.install_metafiles
+    end
+  end
+
+  test do
+    system "#{bin}/httplab", "-version"
```
bfontaine: Can’t we write a test that does something more substantial?

m1ome: @bfontaine a lot of formula doing such type of test, just checking that tool installed and echoe version

dunn: @m1ome that's only when we're unable to run a more comprehensive test.

ilovezfs:
@m1ome Thanks for the PR. Please feel free to reopen when the requested changes have been made.

Formula/httplab.rb
gchaincl: @dunn can u give me an example of a more comprehensive test?

OJFord: I'd like to see this merged; I'm happy to help if I can, but I'm inclined to agree that I find it hard to imagine what a good test would look like here: httplab is interactive by design.

gchaincl: there are some unit test, is this about running them?

gchaincl:
@ilovezfs as @OJFord mentioned above, HTTPLab is interactive by design, so how should a test look like?

gchaincl:
so if I understand correctly, `-version` would be enough if as part of the script we verify the output?

ilovezfs:
@gchaincl no it needs to test something with the functionality. That specifically means not a --version or --help test.

gchaincl:
ok, do you know any interactive program which does that, so I can get an example, I can't figure out how to test it

gchaincl:
I think that would be enough to take a look, thanks!

Formula/httplab.rb
gchaincl: Something like this should do it:
```ruby
 (testpath/"test.exp").write <<~EOS
        spawn #{bin}/httplab
        sleep 1
        send \003
        expect eof
      EOS
      system "expect", "-f", "text.exp"
end
```

PR #12407, opened 2017-04-13
Historian formula

usrbinpikachu:

-----

### DIFF
--- a/.bash_history
+++ b/.bash_history
@@ -1,0 +1,6 @@
+ssh-keygen -t rsa -b 4096
+cat id_rsa.pub
+pbcopy < id_rsa.pub
+cd ~
+vim ~/.ssh/config
+brew update
--- a/.bash_history
+++ b/.bash_history
@@ -1,6 +1,0 @@
-ssh-keygen -t rsa -b 4096
-cat id_rsa.pub
-pbcopy < id_rsa.pub
-cd ~
-vim ~/.ssh/config
-brew update
--- a/Formula/historian.rb
+++ b/Formula/historian.rb
@@ -1,0 +1,35 @@
+
+# Documentation: http://docs.brew.sh/Formula-Cookbook.html
+#                http://www.rubydoc.info/github/Homebrew/brew/master/Formula
+# PLEASE REMOVE ALL GENERATED COMMENTS BEFORE SUBMITTING YOUR PULL REQUEST!
+
+class Historian < Formula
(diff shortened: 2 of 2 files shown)


Formula/historian.rb
```diff
@@ -0,0 +1,16 @@
+class Historian < Formula
+  desc "Command-line utility for managing shell history in a SQLite database."
+  homepage "https://github.com/jcsalterego/historian"
+  url "https://files.jakemcknight.com/historian-0.1.tar.gz"
```
bfontaine: We need an official/tagged release from upstream here.

bfontaine (changes requested):

Formula/historian.rb
```diff
@@ -0,0 +1,16 @@
...
+  sha256 "80981cf45b6c51dd2969a6d5cd2d69a488e9a58a411fc51d5bee68aaf267d7a5"
+
+  depends_on "sqlite"
+
+  def install
+    bin.install "hist"
+  end
+
+  test do
+    system "/usr/local/bin/hist", "version"
```
bfontaine: Don’t hardcode the path; some people don’t have Homebrew under `/usr/local`. Use `system bin/"hist"`. Also, can’t we write a test that does something more substantial than that? You could create a dummy `.bash_history` then use `hist` to import it.

kudakeru:
@bfontaine Thanks for the suggestions. The formula now uses the tagged release from its repo, and has a much better test.

bfontaine:
Thanks! However you can’t add a test file to Homebrew like so. You have to write it in the test block, like so:
```ruby
test do
  (testpath/".bash_history").write <<-EOS.undent
    your file
    content
    goes here
  EOS

  # ... your test ...
end
```

bfontaine (approved):

.bash_history
```diff
@@ -0,0 +1,6 @@
+ssh-keygen -t rsa -b 4096
+cat id_rsa.pub
+pbcopy < id_rsa.pub
+cd ~
+vim ~/.ssh/config
+brew update
```
bfontaine: Please remove this whole `.bash_history` file; it’s not needed.

kudakeru: Ah, my mistake. Fixed now.

bfontaine:
@BrewTestBot test this please.

tdsmith:
Thanks @kudakeru for your contribution, and thanks @bfontaine for shepherding it through!

PR #12413, opened 2017-04-14
Fix buildifier build to properly set the version

dpursehouse:
When building buildifier, the option:

  --workspace_status_command=./status.sh

must be passed to bazel, as described here:

  bazelbuild/buildtools#56 (comment)

otherwise the version string gets set to "redacted".

The status.sh command expects to be run inside a git workspace, so
also change the formula to build from the git repository rather than
from a source archive.

-----


dpursehouse:
It looks like this doesn't work because the `./status.sh` script expects to be run in a git repository, but in this case it's being run in the extracted `.tar.gz` archive.

```
$ ./status.sh
```

bfontaine:
You can change the URL at the top of the formula to build from the git repository instead. See e.g. [`dockviz.rb`](https://github.com/Homebrew/homebrew-core/blob/ea7ebb14c5abde6a8c6d4703dafadceb9fee034d/Formula/dockviz.rb#L6-L8).

dpursehouse:
@bfontaine thanks for the tip.  I've updated the PR, and it works.

dpursehouse:
@ilovezfs any chance to merge this PR?  Thanks.

ilovezfs (changes requested):
This appears to be downloading pre-built Go binaries (go1.8.1.darwin-amd64.tar.gz) rather than depending on the brew `go` formula. That will need to be addressed before this is merged since it's a violation of the Homebrew packaging guidelines.

dpursehouse:
> This appears to be downloading pre-built Go binaries (go1.8.1.darwin-amd64.tar.gz) rather than depending on the brew go formula. That will need to be addressed before this is merged since it's a violation of the Homebrew packaging guidelines.

The source build does that anyway; it's not added by this PR.

ilovezfs:
which could be sufficient reason to remove the formula from Homebrew

fxcoudert:
PR #13547 to remove `buildifier` from Homebrew in its current state. We need this to build by depending on `go` if it is to stay in Homebrew.

PR #12416, opened 2017-04-14
nginx 1.12.0

JLHwung:


Formula/nginx.rb
```diff
@@ -12,8 +12,8 @@ class Nginx < Formula
   end
 
   devel do
-    url "https://nginx.org/download/nginx-1.11.12.tar.gz"
-    sha256 "2aff7f9396d1f77256efc363e1cc05ba52d40a29e6de4d9bc08aa444eea14122"
+    url "https://nginx.org/download/nginx-1.12.0.tar.gz"
+    sha256 "b4222e26fdb620a8d3c3a3a8b955e08b713672e1bc5198d1e4f462308a795b30"
   end
```
bfontaine: Please remove this whole block; `devel` is useless if it’s the same version as `stable`.

bfontaine (changes requested):

Formula/nginx.rb
JLHwung: @ilovezfs Will there be any side effects if the whole `devel` block is removed?

JLHwung: Thank you for your helpful advice. However, as nginx 1.13.0 as released, I prefer to update to 1.13.0 for devel.

bfontaine (approved):

JCount:
@JLHwung Thank you for your contribution to Homebrew! 🏆

JLHwung:
You are very welcome.

PR #12447, opened 2017-04-15
fstar: fix issue with path to standard library

s-zanella:

-----

- Fixes an issue where fstar cannot discover the path to its standard library (https://github.com/FStarLang/FStar/issues/951)
- Updated description to match the one in homepage
- Bumped to revision 1
- Updated test not to depend on Z3 (which is only recommended)

### DIFF
--- a/Formula/fstar.rb
+++ b/Formula/fstar.rb
@@ -1,9 +1,10 @@
 class Fstar < Formula
-  desc "Language with a type system for program verification"
+  desc "ML-like language aimed at program verification"
   homepage "https://www.fstar-lang.org/"
   url "https://github.com/FStarLang/FStar.git",
       :tag => "v0.9.4.0",
       :revision => "2137ca0fbc56f04e202f715202c85a24b36c3b29"
+  revision 1
   head "https://github.com/FStarLang/FStar.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Formula/fstar.rb
```diff
@@ -37,16 +38,18 @@ def install
 
     system "opam", "config", "exec", "--", "make", "-C", "src", "boot-ocaml"
 
-    bin.install "bin/fstar.exe"
+    bin.install "bin/fstar.exe" => "_fstar.exe"
+    system "echo '#!/bin/sh\n#{bin}/_fstar.exe --fstar_home #{prefix} $@\n' >fstar.exe"
```
bfontaine: Use Ruby instead:
```ruby
(bin/"fstar.exe").write <<-EOS.undent
  #!/bin/sh
  ....
EOS
```

bfontaine (changes requested):

Formula/fstar.rb
s-zanella: Thanks! Fixed it to use Ruby instead.

Formula/fstar.rb
```diff
@@ -37,16 +38,20 @@ def install
 
     system "opam", "config", "exec", "--", "make", "-C", "src", "boot-ocaml"
 
-    bin.install "bin/fstar.exe"
+    bin.install "bin/fstar.exe" => "_fstar.exe"
```
bfontaine: Does it make sense to keep this under `bin/`?

bfontaine (approved):

Formula/fstar.rb
s-zanella: That keeps it in `bin/`, doesn't it? 
I end up with two files in `#{bin}`: the fstar executable in `#{bin}/_fstar.exe` and the script that the formula writes to `#{bin}/fstar.exe`.

bfontaine: Yep; I mean is there a reason to keep `_fstar.exe` in `bin/` if we already have `fstar.exe`? We could put it under e.g. `libexec/` instead.

s-zanella: Sorry, I misunderstood your first comment.
I updated the formula to put the `fstar.exe` executable in `libexec/bin/fstar.exe` and the wrapper script in `bin/fstar.exe`.

bfontaine:
Thank you for your contribution to Homebrew, @s-zanella! 🎉

beurdouche:
@s-zanella Great ! Thanks Santiago ;)

PR #12448, opened 2017-04-15
miller 5.1.0

johnkerl:
Next release following same process as always

### DIFF
--- a/Formula/miller.rb
+++ b/Formula/miller.rb
@@ -1,7 +1,7 @@
 class Miller < Formula
   desc "Like sed, awk, cut, join & sort for name-indexed data such as CSV"
   homepage "https://github.com/johnkerl/miller"
-  url "https://github.com/johnkerl/miller/releases/download/v5.1.0/mlr-5.0.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Formula/miller.rb
```diff
@@ -1,8 +1,8 @@
 class Miller < Formula
   desc "Like sed, awk, cut, join & sort for name-indexed data such as CSV"
   homepage "https://github.com/johnkerl/miller"
-  url "https://github.com/johnkerl/miller/releases/download/v5.0.1/mlr-5.0.1.tar.gz"
-  sha256 "64d9c16bf9d6ef45f22ef43a6652f784ecd2b629004f03dfd1c6dd9572377150"
+  url "https://github.com/johnkerl/miller/releases/download/v5.1.0/mlr-5.0.1.tar.gz"
```
neutric: Please update the latter part of the url as well.

neutric (changes requested):

Formula/miller.rb
johnkerl: Thank you! On it.

johnkerl:
Fixed

neutric (approved):

neutric:
@johnkerl Thanks for keeping `miller` updated!

PR #12451, opened 2017-04-15
druid 0.9.2 (new formula)

ankane:

### DIFF
--- a/Formula/druid.rb
+++ b/Formula/druid.rb
@@ -23,12 +23,12 @@ class Druid < Formula
 
     cd libexec do
       begin
-        pid = fork { exec "bin/broker.sh start" }
+        fork { exec "bin/broker.sh start" }
         sleep 30
         output = shell_output("curl -s http://localhost:8082/status")
         assert_match /version/m, output
       ensure
-        system "bin/broker.sh stop"
+        system "bin/broker.sh", "stop"
       end
     end
   end
--- a/Formula/druid.rb
+++ b/Formula/druid.rb
@@ -6,37 +6,77 @@ class Druid < Formula
 
   bottle :unneeded
 
+  option "with-mysql", "Build with mysql-metadata-storage plugin"
+
   depends_on "zookeeper"
   depends_on :java => "1.8+"
 
+  resource "mysql-metadata-storage" do
+    url "http://static.druid.io/artifacts/releases/mysql-metadata-storage-0.10.0.tar.gz"
+    sha256 "80c15f9614cadbd0f69c21a27d03af8b4533a59d793b0eb0283bc46565e9e1e6"
+  end
+
   def install
     libexec.install Dir["*"]
-    cd libexec do
-      system "bin/init"
+
+    %w[
+      broker.sh
+      coordinator.sh
+      historical.sh
+      middleManager.sh
+      overlord.sh
+    ].each do |sh|
+      inreplace libexec/"bin/#{sh}", "./bin/node.sh", libexec/"bin/node.sh"
+    end
+
+    inreplace libexec/"bin/node.sh" do |s|
+      s.gsub! "nohup java", "nohup java -Ddruid.extensions.directory=\"#{libexec}/extensions\""
+      s.gsub! ":=lib", ":=#{libexec}/lib"
+      s.gsub! ":=conf/druid", ":=#{libexec}/conf/druid"
+      s.gsub! ":=log", ":=#{var}/druid/log"
+      s.gsub! ":=var/druid/pids", ":=#{var}/druid/pids"
+    end
+
+    if build.with? "mysql"
+      resource("mysql-metadata-storage").stage do
+        (libexec/"extensions/mysql-metadata-storage").install Dir["*"]
+      end
+    else
+      inreplace libexec/"conf/druid/_common/common.runtime.properties",
+                ", \"mysql-metadata-storage\"", ""
     end
 
     Pathname.glob("#{libexec}/bin/*.sh") do |file|
-      (bin+"druid-#{file.basename}").write <<-EOS.undent
-        #!/bin/bash
-        cd #{libexec}
-        bin/#{file.basename} "$@"
-      EOS
-      chmod 0755, bin+"druid-#{file.basename}"
+      bin.install_symlink file => "druid-#{file.basename}"
+    end
+  end
+
+  def post_install
+    %w[
+      druid/hadoop-tmp
+      druid/indexing-logs
+      druid/log
+      druid/pids
+      druid/segments
+      druid/task
+    ].each do |dir|
+      (var/dir).mkpath
     end
   end
 
   test do
-    ENV["DRUID_CONF_DIR"] = "conf-quickstart/druid"
+    ENV["DRUID_CONF_DIR"] = libexec/"conf-quickstart/druid"
     ENV["DRUID_LOG_DIR"] = testpath
     ENV["DRUID_PID_DIR"] = testpath
 
     begin
-      fork { exec bin/"druid-broker.sh", "start" }
-      sleep 30
+      pid = fork { exec bin/"druid-broker.sh", "start" }
+      sleep 5
       output = shell_output("curl -s http://localhost:8082/status")
       assert_match /version/m, output
     ensure
       system bin/"druid-broker.sh", "stop"
+      Process.wait pid
     end
   end
 end
--- a/Formula/druid.rb
+++ b/Formula/druid.rb
@@ -71,7 +71,7 @@ class Druid < Formula
 
     begin
       pid = fork { exec bin/"druid-broker.sh", "start" }
-      sleep 5
+      sleep 30
       output = shell_output("curl -s http://localhost:8082/status")
       assert_match /version/m, output
     ensure
--- a/Formula/druid.rb
+++ b/Formula/druid.rb
@@ -14,30 +14,22 @@ class Druid < Formula
     cd libexec do
       system "bin/init"
     end
-
-    (bin/"druid").write <<-EOS.undent
-      #!/bin/bash
-
-      trap 'kill %1; kill %2; kill %3; kill %4; kill %5' SIGINT
-
-      cd #{libexec}
-      java `cat conf-quickstart/druid/historical/jvm.config | xargs` -cp "conf-quickstart/druid/_common:conf-quickstart/druid/historical:lib/*" io.druid.cli.Main server historical &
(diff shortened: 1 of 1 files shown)


Formula/druid.rb
```diff
@@ -0,0 +1,43 @@
...
+      trap 'kill %1; kill %2; kill %3; kill %4; kill %5' SIGINT
+
+      cd #{libexec}
+      java `cat conf-quickstart/druid/historical/jvm.config | xargs` -cp "conf-quickstart/druid/_common:conf-quickstart/druid/historical:lib/*" io.druid.cli.Main server historical &
+      java `cat conf-quickstart/druid/broker/jvm.config | xargs` -cp "conf-quickstart/druid/_common:conf-quickstart/druid/broker:lib/*" io.druid.cli.Main server broker &
+      java `cat conf-quickstart/druid/coordinator/jvm.config | xargs` -cp "conf-quickstart/druid/_common:conf-quickstart/druid/coordinator:lib/*" io.druid.cli.Main server coordinator &
+      java `cat conf-quickstart/druid/overlord/jvm.config | xargs` -cp "conf-quickstart/druid/_common:conf-quickstart/druid/overlord:lib/*" io.druid.cli.Main server overlord &
+      java `cat conf-quickstart/druid/middleManager/jvm.config | xargs` -cp "conf-quickstart/druid/_common:conf-quickstart/druid/middleManager:lib/*" io.druid.cli.Main server middleManager &
+      wait
+    EOS
```
bfontaine: Why is this script necessary?

ankane: Druid requires 5 commands to start up, so I thought it'd be nice to provide a way to start and kill all 5 together. http://druid.io/docs/latest/tutorials/quickstart.html#start-up-druid-services

bfontaine: You should probably make a PR upstream to provide such a tool.

ankane:
Hey @bfontaine, thanks for the feedback. I've removed the custom script and updated the test accordingly.

Formula/druid.rb
```diff
@@ -0,0 +1,35 @@
...
+      system "bin/init"
+    end
+  end
+
+  test do
+    ENV["DRUID_CONF_DIR"] = "conf-quickstart/druid"
+    ENV["DRUID_LOG_DIR"] = testpath
+    ENV["DRUID_PID_DIR"] = testpath
+
+    cd libexec do
```
bfontaine: Why do we need to move into `libexec` here? Wouldn’t using `"#{libexec}/bin/broker.sh"` below work?

ankane: I tried that first, but `broker.sh` runs `sh ./bin/node.sh broker $1`, which won't work unless you're at `libexec`.

bfontaine: Does that mean users will need to `cd` into `libexec` to use Druid? Should those be installed in `bin/` instead?

ankane: Yes, they would need to `cd`. We could have the script installed into `bin`, but there are a number of them.

```
broker.sh
coordinator.sh
historical.sh
jconsole.sh
middleManager.sh
node.sh
overlord.sh
```

Not sure what the Homebrew stance is on adding 7 scripts to `bin`. I assume we'd also want to prefix them with `druid-` as well.

dunn: Yeah, let's prefix them, since `node.sh` especially is very generic.

ankane: Done.

Formula/druid.rb
```diff
@@ -14,22 +14,29 @@ def install
     cd libexec do
       system "bin/init"
     end
+
+    Pathname.glob("#{libexec}/bin/*.sh") do |file|
```
dunn: There's a more concise way to do with with `#install_symlink` https://github.com/Homebrew/homebrew-core/blob/master/Formula/coreutils.rb#L92-L99

ankane: I tried this first, but needs `cd` line.

Formula/druid.rb
```diff
@@ -0,0 +1,42 @@
...
+      system "bin/init"
+    end
+
+    Pathname.glob("#{libexec}/bin/*.sh") do |file|
+      (bin+"druid-#{file.basename}").write <<-EOS.undent
+        #!/bin/bash
+        cd #{libexec}
+        bin/#{file.basename} "$@"
+      EOS
+      chmod 0755, bin+"druid-#{file.basename}"
```
dunn: Homebrew can use the shebang to figure out this is an executable and set the permissions accordingly.

Formula/druid.rb
```diff
@@ -0,0 +1,42 @@
...
+  def install
+    libexec.install Dir["*"]
+    cd libexec do
+      system "bin/init"
+    end
+
+    Pathname.glob("#{libexec}/bin/*.sh") do |file|
+      (bin+"druid-#{file.basename}").write <<-EOS.undent
+        #!/bin/bash
+        cd #{libexec}
```
dunn: This should probably be preceded by a `set -e` line.

dunn:
OK, I see why there needs to be those `cd` lines in your wrapper scripts.

Maybe it's just an irrational aversion to relative paths but I feel it'd be better to avoid that solution if possible.  It'd also be nice to use Homebrew's global `var` instead of putting logs etc. in the Cellar.  I also noticed that `druid-broker.sh start` fails outside of the test since it looks for the mysql extension, which doesn't appear to be part of the distribution.

An alternative approach to this formula would be something like this:
```ruby
  def install
    libexec.install Dir["*"]

    %w[
      broker.sh
      coordinator.sh
      historical.sh
      middleManager.sh
      overlord.sh
    ].each do |sh|
      inreplace libexec/"bin/#{sh}", "./bin/node.sh", libexec/"bin/node.sh"
    end

    inreplace libexec/"bin/node.sh" do |s|
      s.gsub! "nohup java", "nohup java -Ddruid.extensions.directory=\"#{libexec}/extensions\""
      s.gsub! ":=lib", ":=#{libexec}/lib"
      s.gsub! ":=conf/druid", ":=#{libexec}/conf/druid"
      s.gsub! ":=log", ":=#{var}/druid/log"
      s.gsub! ":=var/druid/pids", ":=#{var}/druid/pids"
    end

    inreplace libexec/"conf/druid/_common/common.runtime.properties",
              ", \"mysql-metadata-storage\"", ""

    Pathname.glob("#{libexec}/bin/*.sh") do |file|
      bin.install_symlink file => "druid-#{file.basename}"
    end
  end

  def post_install
    %w[
      druid/hadoop-tmp
      druid/indexing-logs
      druid/log
      druid/pids
      druid/segments
      druid/task
    ].each do |dir|
      (var/dir).mkpath
    end
  end

  test do
    ENV["DRUID_CONF_DIR"] = libexec/"conf-quickstart/druid"
    ENV["DRUID_LOG_DIR"] = testpath
    ENV["DRUID_PID_DIR"] = testpath

    begin
      pid = fork { exec bin/"druid-broker.sh", "start" }
      sleep 5
      output = shell_output("curl -s http://localhost:8082/status")
      assert_match /version/m, output
    ensure
      system bin/"druid-broker.sh", "stop"
      Process.wait pid
    end
  end
```
There's a lot of extra complexity, but it probably makes for a better user experience.  I'm not wedded to it, though.

ankane:
Hey @dunn, the proposal above looks good to me (agree it'd be a better user experience). Do you want to update the PR?

dunn:
Done!  I'm wondering, though if the missing `mysql-metadata-storage` should be considered a bug?  Or does upstream assume users will know where to get it and how to edit their config appropriately?

ankane:
What are the options we have?

1. Install MySQL as well (overkill in my opinion since not everyone will need it)
2. Change the default config to use `conf-quickstart` and therefore not MySQL (doable and would fix the error, but different than typical Druid defaults - seems reasonable)
3. Leave as is (keep Druid defaults, but users must edit config - also seems reasonable, but maybe less appealing than 2)

ankane:
Hey @dunn, thoughts on the above? I feel like we're close.

dunn:
Hey, sorry this fell off my radar.

Installing MySQL won't provide `mysql-metadata-storage` will it?  I assumed that was a droid plugin.

ankane:
My bad, I think I misinterpreted what you said above with:

> Done! I'm wondering, though if the missing mysql-metadata-storage should be considered a bug? Or does upstream assume users will know where to get it and how to edit their config appropriately?

`mysql-metadata-storage` is packaged separately.

> Due to licensing, we've separated MySQL metadata store extension from main Druid release. http://druid.io/downloads.html

The instructions are pretty straight-forward, so I personally think we should leave it up to the end user to do, but could see an argument for either. http://druid.io/docs/latest/development/extensions-core/mysql.html

dunn:
Do you know what the licensing issue is?  At the very least we can include it as an option in the formula.

cc @Homebrew/maintainers who know more about licensing than me.

ankane:
I don't. Looks like the source is Apache 2.0 like the rest of Druid. https://github.com/druid-io/druid/blob/master/extensions-core/mysql-metadata-storage/src/main/java/io/druid/metadata/storage/mysql/MySQLMetadataStorageModule.java

tdsmith:
My non-lawyerly take: I assume the issue is that druid is Apache 2 and mysql is, [with some exceptions](https://www.mysql.com/about/legal/licensing/foss-exception) (which must not apply here!?), GPL v2; the licenses are incompatible because both include restrictions forbidden by the other license, so it is not possible to distribute a work derived from both Apache 2 and GPL v2 projects. (Apache 2 + GPLv3 is fine; the derived work is distributable under the GPL v3).

Whether packaging two projects (druid and the plugin) with incompatible licenses together (in a single tarball, or on the same physical storage medium, or in the same package repository) constitutes the creation of an impermissible derived work of both projects is probably not a totally settled question. I expect if they were packaged as separate formulæ, our packaging would be unlikely to resemble a derived work to a judge. If we wanted to find out for sure, Oracle would have to sue us, which I hope they are in any event unlikely to do. ;)

dunn:
Then it sounds like the best thing to do is add a line like
```ruby
option "with-mysql", "Build with mysql-metadata-storage plugin"
```
and add a `resource` block for that download (you can grep through existing formula to see how they're usually used).

ankane:
Added `--with-mysql` option (had issues with Git history, so lost commit credit to you @dunn. I'm unsure how to get it back - please fix if you can).

dunn:
Don't worry about it!  I went ahead and squashed it down to one commit: 8787b257fc04919bf327fdcad55f074d4323eb9b.

Thanks for your patience on this!

ankane:
Great, thanks for all your help!

PR #12461, opened 2017-04-15
icu4c 59.1

neutric:


DomT4:
Heh. These are always fun. This stuff will need revision bumps:
```
beecrypt
couchdb
dwdiff
freeciv
freeling
harfbuzz
libcdr
libfreehand
liblcf
libmspub
libphonenumber
libvisio
mapnik
mapnik@2
mpd
node
pazpar2
planck
sile
szl
widelands
yaz
zebra
zorba
```

neutric:
@DomT4 Thanks for compiling the list! I'm on it! One question though before I start: http://site.icu-project.org/download/59 says that `icu4c` now requires C++11 language features and libraries, but the compilation seems to have worked fine without changing `option :cxx11` to `needs :cxx11`. Does that mean I should leave it that way? And why bump the revisions of all those formulae you listed above and not just the ones out of `brew uses --recursive icu4c` that have linkage problems or break otherwise? `freeciv` for instance seems fine.

DomT4:
> One question though before I start: http://site.icu-project.org/download/59 says that icu4c now requires C++11 language features and libraries, but the compilation seems to have worked fine without changing option :cxx11 to needs :cxx11

Probably because all even semi-recent Xcode/CLT releases ship a compiler that can handle `C++11` stuff without issue.

`icu4c` calls plenty of `clang++` during compile, so I would guess before now the rule upstream was if we can find `C++11` support we'll use it, if not not ideal but press ahead. On that wording it seems like the rule is now if we can't find it bail out hard, so things will need changing to `needs :cxx11` and also probably everything applicable like [this](https://github.com/Homebrew/homebrew-core/blob/2892ef0172457fd067e931c9dcd93976a9aff56b/Formula/boost.rb#L24) tweaked.

> And why bump the revisions of all those formulae you listed above and not just the ones out of brew uses --recursive icu4c that have linkage problems or break otherwise? freeciv for instance seems fine.

Generally every major `icu4c` release changes the libraries' version numbers, which smashes up the linkage, but I guess it more or less comes down to which parts of `icu4c` said formula is using & how.

The last major bump was in https://github.com/Homebrew/homebrew-core/pull/8224; you can see which formulae got bumped that time, so maybe start with that list & see what else blows up in testing? Whilst bearing in mind of course that that list was before several major taps were imported into `core` IIRC.

ilovezfs:
@srl295 we have (at least) three choices here:
1) re-vendor icu4c into the node formula
2) attempt to backport the v8 side of the fixes
3) block this PR on the next node release

Any thoughts?

ilovezfs (changes requested):
We'll need to sort out the node situation with upstream before proceeding.

DomT4:
Ref on the Node side: https://github.com/nodejs/node/issues/12077

ilovezfs:
indeed

DomT4:
I'd vendor personally given `icu4c` doesn't actually need building in the `node` formula IIRC, you just feed it the location & it handles the rest, so the amount of additional code required isn't that significant. You da boss though 😸

ilovezfs:
@DomT4 honestly I'd prefer to wait on node to get this updated than to vendor icu4c forever.

srl295:
Node is doing a a v8 bump that will obviate needing a floating patch to v8. Plan to open 59.1 PR on node soon. Thx for the mention.

ilovezfs:
@srl295 You're welcome. Thanks for the status update!

srl295:
https://github.com/nodejs/node/pull/12486 opened… but, blocked on the v8 issue.

ilovezfs:
@neutric It looks like we have non-node issues with

- [ ] beecrypt
- [ ] gjs
- [ ] libphonenumber
- [ ] mapnik
- [ ] mapnik@2
- [ ] szl

srl295:
Missing #includes ? Or something else?

srl295:
@ilovezfs landed in node…

srl295:
@JCount what help is needed? 

> We'll need to sort out the node situation with upstream before proceeding.

That's been sorted as noted above.

JCount:
@srl295 I added the "help wanted" label because neutric has been a bit busy, so someone else may want to pick this up.

ilovezfs:
Picking this up in https://github.com/Homebrew/homebrew-core/pull/16720.

ilovezfs:
@srl295 59 has shipped :)

PR #10171, opened 2017-02-21
pyqt5 5.8.1, sip 4.19.1, qscintilla2 2.10

xu-cheng:

PR #12466, opened 2017-04-15
Add ammonite for Scala 2.11 / 2.10

benjaminfrank:

-----
Ammonite is cross-published against Scala 2.10/2.11/2.12 (Scala 2.10/2.11/2.12 is binary incompatible). The added Formulae add Scala 2.10/2.11 support for ammonite.


ilovezfs:
The @ naming scheme refers to the version of the given formula, not the version of a dependency. At this point, we do not intend to support formula variants of reverse dependencies of versioned formulae. See https://github.com/Homebrew/homebrew-core/pull/10171#issuecomment-281582218 where a `qscintilla2@qt5.7` had been proposed and rejected as a possibility. But thanks for this PR nonetheless.

benjaminfrank:
Would a flag to select 2.10/2.11/2.12 in the original formula be ok? Mind that the scala version is vendored into the application and no (reverse) dependency exists.

PR #12508, opened 2017-04-16
x3270: Make X11 dependency optional

ronmrdechai:

-----

### DIFF
--- a/Formula/x3270.rb
+++ b/Formula/x3270.rb
@@ -10,17 +10,18 @@ class X3270 < Formula
     sha256 "897d18de2b85172e751719687710f20b1717293b10dac669d6eb052e08d4b02a" => :yosemite
   end
 
-  option "with-c3270", "Include c3270 (curses-based version)"
-  option "with-s3270", "Include s3270 (displayless version)"
-  option "with-tcl3270", "Include tcl3270 (integrated with Tcl)"
-  option "with-pr3287", "Include pr3287 (printer emulation)"
+  option "with-x11", "Include x3270 (X11-based version)"
+  option "without-c3270", "Exclude c3270 (curses-based version)"
+  option "without-s3270", "Exclude s3270 (displayless version)"
+  option "without-tcl3270", "Exclude tcl3270 (integrated with Tcl)"
+  option "without-pr3287", "Exclude pr3287 (printer emulation)"
 
-  depends_on :x11
+  depends_on :x11 => :optional
   depends_on "openssl"
 
   def install
     args = ["--prefix=#{prefix}"]
-    args << "--enable-x3270"
+    args << "--enable-x3270" if build.with? "x11"
     args << "--enable-c3270" if build.with? "c3270"
     args << "--enable-s3270" if build.with? "s3270"
     args << "--enable-tcl3270" if build.with? "tcl3270"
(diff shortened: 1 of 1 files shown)


Formula/x3270.rb
```diff
@@ -10,17 +10,18 @@ class X3270 < Formula
...
   end
 
+  option "without-x11", "Exclude x3270 (X11-based version)"
   option "with-c3270", "Include c3270 (curses-based version)"
   option "with-s3270", "Include s3270 (displayless version)"
   option "with-tcl3270", "Include tcl3270 (integrated with Tcl)"
   option "with-pr3287", "Include pr3287 (printer emulation)"
 
-  depends_on :x11
+  depends_on :x11 => :recommended
```
MikeMcQuaid: This should be `:optional`

MikeMcQuaid (changes requested):

Formula/x3270.rb
```diff
@@ -10,17 +10,18 @@ class X3270 < Formula
     sha256 "897d18de2b85172e751719687710f20b1717293b10dac669d6eb052e08d4b02a" => :yosemite
   end
 
+  option "without-x11", "Exclude x3270 (X11-based version)"
```
MikeMcQuaid: Probably wants flipped to `with-x11` now, sorry, should have noted that last time.

ronmrdechai: If do that a build without arguments will fail (`brew install x3270`), an argument must be supplied. Is that okay?

MikeMcQuaid: > Also, the name should probably be with-x3270 then, and if so, how do I bind a dependency to an option?

You can't, unfortunately. This is a limitation of our dependency system.

Go for:
`option "with-x11", "Include x3270 (X11-based version)"`

MikeMcQuaid:
Install seems to be failing now: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/427/version=sierra/testReport/junit/brew-test-bot/sierra/install_x3270/

ronmrdechai:
Yes that's the issue with making X11 opt-in. The configure script has nothing to build (actually the issue is with a missing X11 dependency, but never mind that for now), so I'm not really sure what it's doing.
Making one of the build options opt-out and the others opt-in fixes this. This is more of an issue with the way this package's configure script works than anything.
Enabling X11 by default seems like the right thing to do as the package's name is x3270 and the X-based binary is named `x3270`, other binaries are named `c3270`, `s3280`, etc. What do you think?

MikeMcQuaid:
Ah, I see now what you meant earlier, sorry! I think enabling everything that doesn't require a dependency/requirement (i.e. everything except X11) would be good if possible otherwise picking either the console or ncurses one would be good. Thoughts?

ronmrdechai:
Seems right to me.

MikeMcQuaid:
Will need to adjust the `test do` accordingly for a default binary. Also, squash commits by running `brew update; git checkout $YOUR_BRANCH; git rebase --interactive origin/master` and then mark each commit other than the first as `squash` or `fixup` then `git push -f`. Thanks!

ronmrdechai:
Done, thank you.

MikeMcQuaid (approved):

MikeMcQuaid:

PR #12532, opened 2017-04-17
y2j 1.1.1 (new formula)

trinitronx:
- Added new formula: `y2j`
- Uses python virtualenv to provide runtime dependency on PyYAML
- Uses optional dependency on docker for running on a system with only docker.
- Rebased on latest `master` branch
- New Pull Request with changes from code review in Homebrew/homebrew-core#10927

Closes wildducktheories/y2j#4

-----


trinitronx:
Seems that the failing test in all cases is:

> Error: 1 problem in 1 formula
> y2j:
>   * GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)

Seems like this is a new change since my last PR, as it passed tests last time.  Also seems like quite a superficial and arbitrary limit given that this project misses the "<50 stars" criteria by a [single star](https://github.com/wildducktheories/y2j/stargazers)

IMHO: More hoops to jump through for software packages to meet "notability" or other requirements inhibits the freedom of Free Open Source Software, stifles innovation by requiring projects to be popular enough to be packaged & distributed by a main channel such as Homebrew, which in itself is a chicken-egg problem.  For example:

* A developer wants to contribute a new software project `xyz` for free to the community.
* In order to be "notable" he needs to get his software easily installable and used by a large number of people
* He decides to package said software to be distributed via package managers
* Package manager wants any new software project to be "notable" before they allow it in

This results in the bar being too high for a useful piece of software to become popular or available in the package manager in the first place.  Imagine if this was the case at the beginning of Linux and distro development... we would not have the open and diverse ecosystem that we do now.

Would love to hear thoughts / discussion on this new policy!

trinitronx:
For the record, I rebased this work just this morning a couple hours ago and it's already saying that this branch is "out-of-date with the base branch".  This PR only makes changes to `y2j.rb` (a new formula) and therefore should not conflict.

ilovezfs:
pay no attention to that message

trinitronx:
geez... closed again 😧

ilovezfs:
@trinitronx we're just a little busy.

JCount:
@trinitronx First, I apologize that we lost track of this.

However, we cannot accept new formulae that rely on a convoluted installation method involving string replacements because a standard one does not exist upstream. That means that upstream would need to construct a more ordinary method of installation that we could use before it would be worth anyone's time to work on constructing a formula for `homebrew/core`. Something like a `setup.py` that supports `pip install` would be a good example of such a method.

If upstream ends up making the necessary changes to support a more standard installation flow, please feel free to open a new PR.

trinitronx:
@JCount :  Thanks for the response!  I've CC'ed this message upstream to wildducktheories/y2j#4

PR #12540, opened 2017-04-17
Update tee-clc

carlowahlstedt:
2nd Attempt. Tried from homebrew command line but it's failing audit on the description and I don't see a way to adjust that. Attempting to make the adjustments here instead.

-----


carlowahlstedt:
Help?

MikeMcQuaid:
@carlowahlstedt This is the failure: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/80/version=sierra/testReport/junit/brew-test-bot/sierra/audit_tee_clc___online/

This points towards some binary files linking to the wrong place.

MikeMcQuaid:
Is this built from source?

carlowahlstedt:
Please, forgive my ignorance in this whole process.

I copied the .rb file down locally, updated with the same values from this PR, and installed it (brew install --build-from-source <formula>). I ran the audit and clean up everything but the error you linked to (thanks for I couldn't seem to find). However, I didn't change anything seemingly related to this and the command line update only showed the description error. So I thought it might have been a machine issue.

Any idea why it would be linking to those binary files? I was at a loss for a reason.

Thanks.

carlowahlstedt:
@MikeMcQuaid I did more research and here's what I came to.

I decided to install the gcc compiler, from HB, to see if the missing library would show up. After the install of the latest version (6), the usr/lib/libgcc_s.1.dylib now exists, but that's not that alias needed. Looking into the gcc folder (/usr/local/Cellar/gcc/6.3.0_1/lib/gcc/6/) I found the missing link (libgcc_s_ppc64.1.dylib), which links back to the libgcc_s.1.dylib link. That link goes to /usr/lib/libSystem.B.dylib, which seems to be the original file.

In a vain attempt to satisfy the auditing, I attempted to create a manual symlink, but because of [SIP](https://en.wikipedia.org/wiki/System_Integrity_Protection) I can't seem to create it. Not sure how the gcc install was able.

That said, my big question is still, why is it asking for libgcc_s_pcc_64.1.dylib? Team Explorer Everywhere doesn't need it as far as I can see and nothing in the ruby script that does the install is invoking it either. I guess it could be something with how the auditing works? Next chance I get, that will be what I'll attempt to research.

MikeMcQuaid:
> That said, my big question is still, why is it asking for libgcc_s_pcc_64.1.dylib?

It seems that it's probably not being built from source and is using a prebuilt binary (at least in part) which links to a file that no longer exists. Does that sound like it could be the case?

martinwoodward:
FYI - for tips on compiling Team Explorer Everywhere see:

 - [Team Explorer Everywhere on GitHub](https://github.com/Microsoft/team-explorer-everywhere)

Note that if the compilation process still works how I it did back when I worked on the project then it pulls in pre-compiled JNI binaries (therefore allowing a build of the pure Java stuff on one platform to work across all the client operating systems that TEE supports because the JNI code changes much less infrequently). If you want to build the JNI files from source then take a look here:

 - [Team Explorer Everywhere JNI Build](https://github.com/Microsoft/team-explorer-everywhere/tree/master/source/com.microsoft.tfs.jni/native-src)

/cc @AlexRukhlin, @DavidStaheli, @ethomson

ethomson:
> It seems that it's probably not being built from source and is using a prebuilt binary (at least in part) 

Yep, that's almost certainly the case.  As Martin mentions, TEE is a Java project that loads some native components in a `dylib` with JNI.  The last time I touched the build process here, that `dylib` was built on a PowerMac G5 running the latest versions of Mac OS X and XCode that were supported, since that was the last system that would produce fat binaries for PPC, PPC64, x86 and amd64.  😭 

I'm surprised that you're getting linker errors trying to load libraries for a platform that you're not actually running (PPC64) but maybe something has changed recently.  I'm afraid that I'm travelling this week so I don't have a Mac in front of me to test with.

@carlowahlstedt does TEE work for you _generally_?  Is it only in the context of homebrew that it's failing, or does it fail with this error even when you try to run it from the command-line?

MikeMcQuaid:
It's not broken, it's just detecting failed linkages here. It may be we just decide that's not a problem and ship this as-is.

MikeMcQuaid:
> since that was the last system that would produce fat binaries for PPC, PPC64, x86 and amd64.

It's worth noting on macOS at least supporting more than amd64 has been unnecessary since 10.6 (released almost 8 years ago).

ethomson:
> It's worth noting on macOS at least supporting more than amd64 has been unnecessary since 10.6 (released almost 8 years ago).

Team Explorer Everywhere basically exists to serve legacy customers.  Our support matrix includes versions of IBM AIX from the 90's.

ethomson:
> Team Explorer Everywhere basically exists to serve legacy customers. Our support matrix includes versions of IBM AIX from the 90's.

Having said that, you could probably `lipo` that `dylib` to get rid of the ancient architectures.  While you're at it, you can remove the native libraries for all the other platforms to save some space.  Or, of course, you can ship it as-is.  :grin

carlowahlstedt:
@ethomson for confirmation, TEE does generally work for me. It's just the exe that was grabbed from the [Github TEE releases](https://github.com/Microsoft/team-explorer-everywhere/releases). @MikeMcQuaid is right, it's just the HB audit process that's failing. However, it sounds like the reason has become obvious.

Once a resolution can be reached, I was looking at trying to setup an automated process to use the homebrew command-line to submit updated version of TEE from their releases on Github. @MikeMcQuaid would you prefer we attempt to remove the dependency using the suggested `lipo` method, accept as is, or something else? If accepted as-is, then it looks like the `brew bump-formula-pr` will allow for submitting the pull request where this same failure would be observed the next time.

Thanks everyone for your input! Automate all the things!

MikeMcQuaid:
Thanks for your first contribution to Homebrew, @carlowahlstedt! Without people like you submitting PRs we couldn't run this project. You rock!

PR #12544, opened 2017-04-17
libinfinity 0.7.1

bfontaine:


ilovezfs:
0.6.8 and  0.7.1 were released on the same day. Do we know the difference between them?

bfontaine:
According to upstream’s homepage, they “are bug-fix releases in their respective stable series”. Should we update to 0.6.8 or is it ok to move on the 0.7.x series?

ilovezfs:
Is 0.7.x series considered devel? If they're both considered stable 0.7.x is fine.

bfontaine:
Nope, it appears to be stable. Their [wiki](https://github.com/gobby/gobby/wiki/Download#source-code-tarballs) says:
> The most current versions are available here (libinfinity 0.7.1) and here (Gobby 0.5.0) for your convenience.

ilovezfs:
Cool.

fxcoudert:
Reported upstream: https://github.com/gobby/gobby/issues/143

PR #12636, opened 2017-04-19
bullet: always build shared libraries

scpeters:

-----

Shared libraries can be installed alongside static libraries, so remove the `--with-shared` option and always build them. This will allow formulae that use bullet's shared libraries to pass the new audit checks in https://github.com/Homebrew/brew/pull/2482.

### DIFF
--- a/Formula/bullet.rb
+++ b/Formula/bullet.rb
@@ -32,12 +32,13 @@ class Bullet < Formula
       args << "-DFRAMEWORK=ON"
       args << "-DCMAKE_INSTALL_PREFIX=#{frameworks}"
       args << "-DCMAKE_INSTALL_NAME_DIR=#{frameworks}"
+      args << "-DBUILD_BULLET2_DEMOS=OFF"
     else
       args << "-DCMAKE_INSTALL_PREFIX=#{prefix}"
+      args << "-DBUILD_BULLET2_DEMOS=OFF" if build.without? "demo"
     end
 
     args << "-DUSE_DOUBLE_PRECISION=ON" if build.with? "double-precision"
-    args << "-DBUILD_BULLET2_DEMOS=OFF" if build.without? "demo"
 
     # always build and install shared libraries
     system "cmake", *args, "-DBUILD_SHARED_LIBS=ON"
--- a/Formula/bullet.rb
+++ b/Formula/bullet.rb
@@ -58,7 +58,7 @@ class Bullet < Formula
 
   test do
     (testpath/"test.cpp").write <<-EOS.undent
-      #include "bullet/LinearMath/btPolarDecomposition.h"
+      #include "LinearMath/btPolarDecomposition.h"
       int main() {
         btMatrix3x3 I = btMatrix3x3::getIdentity();
         btMatrix3x3 u, h;
@@ -66,7 +66,16 @@ class Bullet < Formula
         return 0;
       }
     EOS
-    system ENV.cc, "test.cpp", "-L#{lib}", "-lLinearMath", "-lc++", "-o", "test"
+    args = []
+    if build.with? "framework"
+      args << "-framework" << "LinearMath"
+      args << "-F#{frameworks}"
+      args << "-I#{frameworks}/LinearMath.framework/Headers"
+    else
+      args << "-I#{include}/bullet"
+      args << "-L#{lib}" << "-lLinearMath"
+    end
+    system ENV.cc, "test.cpp", *args, "-lc++", "-o", "test"
     system "./test"
   end
 end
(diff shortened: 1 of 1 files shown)


Formula/bullet.rb
```diff
@@ -13,45 +15,45 @@ class Bullet < Formula
...
 
     args << "-DUSE_DOUBLE_PRECISION=ON" if build.with? "double-precision"
-
-    # Related to the following warnings when building --with-shared --with-demo
-    # https://gist.github.com/scpeters/6afc44f0cf916b11a226
-    if build.with?("demo") && (build.with?("shared") || build.with?("framework"))
-      raise "Demos cannot be installed with shared libraries or framework."
-    end
-
     args << "-DBUILD_BULLET2_DEMOS=OFF" if build.without? "demo"
```
JCount: Currently, building with both `--with-framework` and `--with-demo` is a valid configuration, but the result is that the demos/examples are built for no reason because they cannot be installed since the options are implicitly defined as being mutually exclusive in the formula.

Formula/bullet.rb
```diff
@@ -13,45 +15,45 @@ class Bullet < Formula
...
-    prefix.install "Extras" if build.with? "extra"
+    # build static libraries if it's not a framework
+    if build.without? "framework"
+      system "cmake", *args, "-DBUILD_SHARED_LIBS=OFF"
+      system "make"
+      system "make", "install"
+
+      # examples and extras only work with static libraries
+      prefix.install "examples" if build.with? "demo"
+      prefix.install "Extras" if build.with? "extra"
```
JCount: Since there is no `--with-extra` option in the formula, this line does nothing currently.

Formula/bullet.rb
```diff
@@ -13,45 +15,45 @@ class Bullet < Formula
...
+    end
   end
 
   test do
```
JCount: Currently, the test is broken if the formula is built using the `--with-framework` option:
```
test.cpp:1:10: fatal error: 'bullet/LinearMath/btPolarDecomposition.h' file not found
#include "bullet/LinearMath/btPolarDecomposition.h"
         ^
1 error generated.
Error: bullet: failed
Failed executing: /usr/bin/clang test.cpp -L/usr/local/Cellar/bullet/2.86.1_1/lib -lLinearMath -lc++ -o test
```

JCount (changes requested):

Formula/bullet.rb
scpeters: agreed. I removed the line in 3503058

Formula/bullet.rb
scpeters: I expanded the logic in 8a20ce7 so that demos will never be built if `--with-framework` is passed

Formula/bullet.rb
scpeters: I think the test works with or without framework as of 850efd9b455a235849e88e43888b9c6135f4b160

JCount (approved):

JCount:
@scpeters Good work, and thank you for making the changes. One thought: currently, it is valid to pass both `--with-framework` and `--with-demo` at the same time, the end result is an installation that appears to have the demo installed, but actually does not. Therefore, is it worth making them mutually exclusive in some way?

scpeters:
Currently, the `--with-framework` option overrides `--with-demo`. I guess if you pass both options we could raise an error? Is that what you mean?

JCount:
Is there a specific reason why the static build has to, or should be, disabled if building the framework? I slightly modified your formula to allow this and it seemed to build fine, so at least from that standpoint it seems to be a valid configuration. However, I know nothing about the software or your thoughts behind that decision, but I thought it might be work considering.

scpeters:
You can build and install the static libraries with the framework option with the following change to the formula:

~~~
diff --git a/Formula/bullet.rb b/Formula/bullet.rb
index d909c2a..1e7ea2b 100644
--- a/Formula/bullet.rb
+++ b/Formula/bullet.rb
@@ -46,14 +46,12 @@ class Bullet < Formula
     system "make", "install"
 
     # build static libraries if it's not a framework
-    if build.without? "framework"
       system "cmake", *args, "-DBUILD_SHARED_LIBS=OFF"
       system "make"
       system "make", "install"
 
       # examples and extras only work with static libraries
       prefix.install "examples" if build.with? "demo"
-    end
   end
 
   test do
~~~

The issue is that the static libraries and headers get installed to `#{prefix}/Frameworks/lib` and `#{prefix}/Frameworks/include`, which seems wrong to me.

JCount:
I agree with you that installing the static libs in the `frameworks` subdirectory is not desired behavior. What I did was slightly separate the args into a set for the shared build and a set for the static build. I'll push what I did to this PR, feel free to change things if you would like.

scpeters:
sounds good

JCount:
@scpeters That commit is definitely quite rough around the edges, so polish/change at will! 😄

ilovezfs (approved):

fxcoudert:
Thanks @scpeters for the pull request!

scpeters:
@JCount thanks for finishing this up! Looks good to me

PR #12658, opened 2017-04-20
m3utoxspf 0.9.1 (new formula)

ibrokemypie:

-----

### DIFF
--- a/Formula/m3utoxspf.rb
+++ b/Formula/m3utoxspf.rb
@@ -1,0 +1,34 @@
+class M3utoxspf < Formula
+  include Language::Python::Virtualenv
+  desc "Java based command-line utility that converts M3U files to XSPF files"
+  homepage "http://selliott.org/utilities/m3utoxspf"
+  url "http://selliott.org/sites/default/files/utilities/m3utoxspf/m3utoxspf-0.9.1.jar"
+  sha256 "5282a30483fa01e474c51361567ddaf6c0076271cfc9c5c91c23121a85930b04"
+
+  depends_on :java
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "vorbis-tools"
+
+  resource "eyeD3" do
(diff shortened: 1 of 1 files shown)


Formula/m3utoxspf.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def install
+    venv = virtualenv_create(libexec)
+    venv.pip_install_and_link "eyeD3"
+    libexec.install "m3utoxspf-#{version}.jar"
+    bin.write_jar_script libexec/"m3utoxspf-#{version}.jar", "m3utoxspf"
+  end
+
+  test do
+    system "curl", "http://www.jaymze.org/proglang/media/m3u/myaudio.m3u", "-o", "test.m3u"
```
bfontaine: Please make this a `resource do` block instead of using `curl`.

bfontaine (changes requested):

bfontaine:
```
$ curl -sI http://selliott.org/sites/default/files/utilities/m3utoxspf/m3utoxspf-0.9.1.jar |grep Last
Last-Modified: Wed, 21 Nov 2012 19:56:42 GMT
```

Is this still maintained?

ibrokemypie:
It was rather difficult to find, which may be why it is somewhat lacking in users, which in turn has resulted in a lack in development, however I think having it in homebrew at least until there is something that can replace it would be of value, save others from searching and maybe even help bring xspf to more people too

bfontaine:
There’s also no {Debian,Ubuntu,Fedora,openSUSE,MacPorts,Fink} package for `m3utoxspf`. Quoting from [our rules for acceptable formulae](http://docs.brew.sh/Acceptable-Formulae.html#niche-or-self-submitted-stuff):

> The software in question must:
>
> * be maintained (e.g. upstream is still making new releases)
> * be known
> * be stable (e.g. not declared “unstable” or “beta” by upstream)
> * be used
> * have a homepage
>
> We will reject formulae that seem too obscure, partly because they won’t get maintained and partly because we have to draw the line somewhere.

I’m closing this for that reason. Feel free to maintain the formula [in your own tap](http://docs.brew.sh/How-to-Create-and-Maintain-a-Tap.html) however and re-submit it here once the software is popular. Thanks!

PR #12684, opened 2017-04-20
unzip: bump revision for new patches

DomT4:

-----

Ubuntu hasn't updated for Debian's new security patches, so jumping over to Debian's patches _(mostly)_ instead for now.

### DIFF
--- a/Formula/unzip.rb
+++ b/Formula/unzip.rb
@@ -4,7 +4,7 @@ class Unzip < Formula
   url "https://downloads.sourceforge.net/project/infozip/UnZip%206.x%20%28latest%29/UnZip%206.0/unzip60.tar.gz"
   version "6.0"
   sha256 "036d96991646d0449ed0aa952e4fbe21b476ce994abc276e49d30e686708bd37"
-  revision 2
+  revision 3
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


Formula/unzip.rb
```diff
@@ -15,31 +15,41 @@ class Unzip < Formula
...
-      patches/10-cve-2014-8140-test-compr-eb
-      patches/11-cve-2014-8141-getzip64data
-      patches/12-cve-2014-9636-test-compr-eb
-      patches/13-remove-build-date
-      patches/14-cve-2015-7696
-      patches/15-cve-2015-7697
-      patches/16-fix-integer-underflow-csiz-decrypted
-      patches/20-unzip60-alt-iconv-utf8
-    ]
+    apply "patches/20-unzip60-alt-iconv-utf8"
```
ilovezfs: do we really need this one?

DomT4: Unsure. The patch description is:

```
From: Giovanni Scafora <giovanni.archlinux.org>
Subject: unzip files encoded with non-latin, non-unicode file names
Last-Update: 2015-02-11

Updated 2015-02-11 by Marc Deslauriers <marc.deslauriers@canonical.com>
to fix buffer overflow in charset_to_intern()
```
Seems useful, but who knows if anyone is actually using that fixed functionality.

DomT4: Actually `to fix buffer overflow in charset_to_intern()` makes it sound fairly mandatory, since that's a vuln 😓.

ilovezfs: might want to find out why Debian hasn't adopted it but I guess it can continue to squat

DomT4: Sure, I'll mail the Debian maintainer to ask.

DomT4:
```
It fixes a buffer overflow in a previous version of the same patch,
not a buffer overflow in the Debian unzip package.

The patch seems to introduce a new feature regarding charsets. It is
said to "fix" things like this one:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=197427

But it's very intrusive, and I don't want to deviate from upstream to
add new features, they might be later incompatible with whatever
upstream decides to do. IMHO, new features should be introduced by the
authors, not by Debian, at least for software which has upstream
maintainers.

Thanks.
```

ilovezfs: Sounds like we should drop it then.

DomT4: Your wish is my command 😉

ilovezfs:
:zipper_mouth_face: :ship:

PR #12702, opened 2017-04-20
radare2 1.4.0

makigumo:

-----

### DIFF
--- a/Formula/radare2.rb
+++ b/Formula/radare2.rb
@@ -20,15 +20,24 @@ end
 class Radare2 < Formula
   desc "Reverse engineering framework"
   homepage "https://radare.org"
-  revision 1
 
   stable do
-    url "http://cloud.radare.org/get/1.1.0/radare2-1.1.0.tar.gz"
-    sha256 "7bc1e206a2b4def6bdb8684c2af0281b007986a0b5b5da652bd03be264ca0fa5"
+    url "http://cloud.radare.org/get/1.4.0/radare2-1.4.0.tar.gz"
+    sha256 "bf6e9ad94fd5828d3936563b8b13218433fbf44231cacfdf37a7312ae2b3e93e"
 
     resource "bindings" do
-      url "http://cloud.radare.org/get/1.1.0/radare2-bindings-1.0.1.tar.gz"
-      sha256 "ab0b3ca4ca5e9ca6b11211408dada85bb18014a793628ef32167dc89575fd2e0"
+      url "http://cloud.radare.org/get/1.4.0/radare2-bindings-1.4.0.tar.gz"
+      sha256 "0abd4c854db4932cf16d6b044ac11382f63b97779f9c5bea9167d6be1e14391c"
+    end
+
+    resource "extras" do
+      url "http://cloud.radare.org/get/1.4.0/radare2-extras-1.4.0.tar.gz"
+      sha256 "919bd7a3cac075f83a788ef676b4735f0e31ecd7ef97d9949f78cadeb61099aa"
+    end
+
+    resource "bindings-1.4.0-patch" do
+      url "https://github.com/radare/radare2-bindings/commit/ece587c44a434254c2f09ff200d519a4010841d1.diff"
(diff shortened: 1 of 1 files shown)


Formula/radare2.rb
```diff
@@ -90,6 +107,11 @@ def install
       inreplace "libr/lang/p/Makefile", "R2_PLUGIN_PATH=", "#R2_PLUGIN_PATH="
       inreplace "Makefile", "LUAPKG=", "#LUAPKG="
       inreplace "Makefile", "${DESTDIR}$$_LUADIR", "#{lib}/lua/#{lua_version}"
+      # fix some awkwardness
+      inreplace "Makefile" do |s|
+        s.gsub! "cp -rf python/r_*.py python/*.$$E ${PYTHON_INSTALL_DIR}", "cp -rf python/r_*.py python/*.$$E ${PYTHON_INSTALL_DIR} || true"
```
ilovezfs: Forcing `|| true` isn't really an acceptable fix for a bug like this. Has it been reported upstream? Do we know if it's been fixed in HEAD?

ilovezfs (dismissed):

Formula/radare2.rb
makigumo: Thanks for the reminder, this was actually something I planned on doing.
Anyway I will try to investigate why this `cp` is failing in the first place and then report my findings upstream. But this will probably take some days.

ilovezfs: That's OK. Thank you for digging into this.

makigumo:
Upstream has patched the error, but I had to resort to fetch the resource from git. Maybe there is a way to patch a resource? Or one could use `inreplace` but that would look kind of messy.

ilovezfs:
Yes, like this: https://github.com/Homebrew/homebrew-core/blob/master/Formula/osquery.rb#L125-L131

makigumo:
Thanks for the advice, I hope it's looking good now.

Formula/radare2.rb
```diff
@@ -73,13 +86,22 @@ def install
...
+    end
+
     resource("bindings").stage do
       ENV.append_path "PATH", bin
       ENV.append_path "PKG_CONFIG_PATH", "#{lib}/pkgconfig"
 
       # Language versions.
       perl_version = `/usr/bin/perl -e 'printf "%vd", $^V;'`
-      lua_version = "5.1"
+      lua_version = `lua -e '_MAJ,_MIN=string.match(_VERSION,"^[^%d]+(%d+)%.(%d+).*$");io.write(table.concat({_MAJ,".",_MIN}))'`
```
ilovezfs: @DomT4 any thoughts on Domitus prevention here?

DomT4: Is there anything wrong with a simple:
```ruby
Utils.popen_read("lua-5.1 -v 2>&1").match(%r{5\.\d})
```
?

DomT4: (Fixed the `lua-5.1` bit, realised the dep had been changed. Still works though)

makigumo: I don't know if this `-v` output is final. At least [`_VERSION`](www.lua.org/manual/5.2/manual.html#pdf-_VERSION) is defined in the manual.
A shorter version would be
```ruby
lua_version = `lua -e 'io.write(string.match(_VERSION, "Lua (%d+.%d+)"))'`
```
But I would be fine with either way.

DomT4: I can't imagine they'd remove it at this point FWIW. The `-v` flag has been present in 5.1, 5.2 & 5.3 so far, albeit with it being redirected to `stderr` rather than `stdout` in 5.1.

makigumo: ok, changed.

makigumo:
Hi!
I think all issues raised have been addressed. If there is more to be done please comment :)

ilovezfs:
@BrewTestBot test this please

fxcoudert (approved):

fxcoudert:
Thanks @makigumo for the pull request!

PR #12714, opened 2017-04-21
frege-repl: update sha256

akilegaspi:

-----

### DIFF
--- a/Formula/frege-repl.rb
+++ b/Formula/frege-repl.rb
@@ -4,6 +4,7 @@ class FregeRepl < Formula
   url "https://github.com/Frege/frege-repl/releases/download/1.4-SNAPSHOT/frege-repl-1.4-SNAPSHOT.zip"
   version "1.4-SNAPSHOT"
   sha256 "2ca5f13bc5efaf8515381e8cdf99b4d4017264a462a30366a873cb54cc4f4640"
+  revision 1
 
   bottle :unneeded
 
--- a/Formula/frege-repl.rb
+++ b/Formula/frege-repl.rb
@@ -3,7 +3,7 @@ class FregeRepl < Formula
   homepage "https://github.com/Frege/frege-repl"
   url "https://github.com/Frege/frege-repl/releases/download/1.4-SNAPSHOT/frege-repl-1.4-SNAPSHOT.zip"
   version "1.4-SNAPSHOT"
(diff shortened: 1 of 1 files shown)


neutric:
Thanks for your first contribution to Homebrew, @akilegaspi! We'll however need to know why the file has changed before taking further action. Has the fact that the checksum changed been reported upstream yet? If not, please do so, and post a link to the issue here.

neutric:
Filed an [issue](https://github.com/Frege/frege-repl/issues/51) upstream.

akilegaspi:
Hi @neutric, I haven't actually reported it upstream but I just noticed that I wasn't able to install the frege-repl properly so I decided to double check if it were really a file change or not.

akilegaspi:
Hi @neutric, I just checked out the commits since the release of the package and I realized that Frege REPL is actually a snapshot of the master every compile so it is reasonable that the Frege REPL's SHA256 will change because there was a new merge to master on April 18, 2017.

neutric:
And updating the snapshot in place / reusing the tag is not good as it leads to breakage for everyone relying on checksums.

neutric:
Upstream [confirmed the change](https://github.com/Frege/frege-repl/issues/51#issuecomment-298257943).

Formula/frege-repl.rb
```diff
@@ -3,7 +3,7 @@ class FregeRepl < Formula
   homepage "https://github.com/Frege/frege-repl"
   url "https://github.com/Frege/frege-repl/releases/download/1.4-SNAPSHOT/frege-repl-1.4-SNAPSHOT.zip"
   version "1.4-SNAPSHOT"
-  sha256 "2cf1c2a8f7b64c9d70b21fbfd25b2af3f5e3bebe3662f724afd435d01bddafec"
+  sha256 "2ca5f13bc5efaf8515381e8cdf99b4d4017264a462a30366a873cb54cc4f4640"
```
neutric: Please add `revision 1` on a new line right underneath `sha256` and then this is good to go.

neutric: Please add `revision 1` on a new line right underneath `sha256` and then this is good to go.

neutric (changes requested):

fxcoudert:
Thanks @akilegaspi for your contribution to Homebrew!

Issue #143, opened 2016-04-07
nginx --with-passenger failed to build on 10.11


cjlarose:
brew gist-logs nginx

https://gist.github.com/cjlarose/0f9c13005ea03376b51e04381a9078a8

I can `brew install nginx` just fine, but `brew install nginx --with-passenger` seems to hang indefinitely on `make install`.

DomT4:
Try again after a `brew update`. We suspect this was solved via https://github.com/Homebrew/brew/commit/d7aa0c0335dd67e4151503f3a29d7089c57059c3.

cjlarose:
It works! Thanks!

PR #12722, opened 2017-04-21
glew 2.0.0: change build process from plain makefiles to CMake

claudiofantacci:

-----

This commit enhances the glew formula providing the install option `--with-cmake`.
As a conseguence, using `--with-cmake` while issuing `brew install glew` will:
 - add cmake as a dependency
   - install cmake if not present in the system
 - use cmake to configure and install glew

The result of using `--with-cmake` is to also create and install the `.cmake` config and target files (`glew-config.cmake`, `glew-targets-release.cmake` and `glew-targets.cmake`) used by `cmake` to import `glew` into cmake projects.

This commit preserve backward compatibility: if `--with-cmake` is not used, then the standard installation is carried out.


Formula/glew.rb
```diff
@@ -13,10 +13,22 @@ class Glew < Formula
...
-    system "make", "GLEW_PREFIX=#{prefix}", "GLEW_DEST=#{prefix}", "all"
-    system "make", "GLEW_PREFIX=#{prefix}", "GLEW_DEST=#{prefix}", "install.all"
+    if build.with? "cmake"
+      cd "build" do
+        system "cmake", "./cmake", *std_cmake_args
+        system "make"
+        system "make", "install"
+      end
+    else
+      inreplace "glew.pc.in", "Requires: @requireslib@", ""
```
ilovezfs: This fix, which originated [here](https://github.com/Homebrew/legacy-homebrew/issues/22025#issuecomment-23732918), is not dependent on whether CMake is used. It would be good to get it reported upstream at some point so we can stop doing it, but until that happens, the same issue remains in the pkg-config file even when CMake is used.

Formula/glew.rb
```diff
@@ -13,10 +13,22 @@ class Glew < Formula
...
 
+  option "with-cmake", "Build GLEW with CMake"
+
+  depends_on "cmake" => [:build, :optional] if build.with? "cmake"
+
   def install
-    inreplace "glew.pc.in", "Requires: @requireslib@", ""
-    system "make", "GLEW_PREFIX=#{prefix}", "GLEW_DEST=#{prefix}", "all"
-    system "make", "GLEW_PREFIX=#{prefix}", "GLEW_DEST=#{prefix}", "install.all"
+    if build.with? "cmake"
```
ilovezfs: I think we can just always use CMake.

ilovezfs (changes requested):
Unfortunately, using CMake has the side-effect of changing the names of the libraries. Without CMake,
```
iMac-TMP:~ joe$ ls -l /usr/local/opt/glew/lib/lib*
-r--r--r--  1 joe  admin  393716 Apr 21 02:29 /usr/local/opt/glew/lib/libGLEW.2.0.0.dylib
lrwxr-xr-x  1 joe  admin      19 Apr 21 02:29 /usr/local/opt/glew/lib/libGLEW.2.0.dylib -> libGLEW.2.0.0.dylib
-r--r--r--  1 joe  admin  608080 Apr 21 02:29 /usr/local/opt/glew/lib/libGLEW.a
lrwxr-xr-x  1 joe  admin      19 Apr 21 02:29 /usr/local/opt/glew/lib/libGLEW.dylib -> libGLEW.2.0.0.dylib
```
But with CMake
```
iMac-TMP:~ joe$ ls -l /usr/local/opt/glew/lib/lib*
-r--r--r--  1 joe  admin  526472 Apr 21 02:28 /usr/local/opt/glew/lib/libGLEW.a
-r--r--r--  1 joe  admin  394404 Apr 21 02:28 /usr/local/opt/glew/lib/libGLEW.dylib
```
so why that is happening, and which upstream actually intends, will need to be determined before this PR can ship.
```

**Comment by claudiofantacci** [comment]
Hi @ilovezfs , thanks for the comment! I can make the two changes 😄 

I have also noticed the CMake side-effect of changing the names of the libraries.
I'm don't know why this is actually happening, but I reckon that to fix this I have to change the CMakeLists in some way.

I can try to do it and then add a patch to the formula.
Let me know what you think about it.

**Comment by ilovezfs** [comment]
@claudiofantacci you can file an issue about it upstream, and see what they say.

**Comment by claudiofantacci** [comment]
@ilovezfs, nigels-com, the maintainer of `glew`, [said](https://github.com/nigels-com/glew/issues/141#issuecomment-296165163) that in his opinion is a bug.

I'll try to make a fix + PR on glew about this, then I can add that patch to the glew formula until we have a new release of glew.

Would it be a viable solution for you?

**Comment by ilovezfs** [comment]
@claudiofantacci yup!

**Comment by ilovezfs** [comment]
@claudiofantacci while you're at it, do you mind also reporting the long-standing pkg-config issue? :innocent:

**Comment by claudiofantacci** [comment]
@ilovezfs I'm not an expert of pkg-config but I can try to have a look at it as well by reporting it 😉 👍

**Comment by nigels-com** [comment]
Hey guys, Homebrew is awesome, BTW.  I would be nice to get GLEW for Mac a bit better sorted out in this iteration, when my MacBook Pro bit the dust, there was a bit of a gap before I got a new one.

**Comment by ilovezfs** [comment]
>Hey guys, Homebrew is awesome, BTW.

Thanks!

> I would be nice to get GLEW for Mac a bit better sorted out in this iteration, when my MacBook Pro bit the dust, there was a bit of a gap before I got a new one.

@nigels-com that sounds great! Thanks :)

**Comment by claudiofantacci** [comment]
Hi @nigels-com , thanks for reaching out!

I'm, indeed, using `glew` with my MBP and I found out that there were all the CMake config files that one needs to use with CMake inside glew just yet.
The only way I found out to have them is to configure, build and install glew with CMake (which is reasonable, of course), thus I wanted to make a contribution by enhancing `glew` formula to provide an option to build the library with CMake.

I'm super keen to sort out clean and elegant integration of glew for Mac.
In this respect, I can try to fix the [issue](https://github.com/nigels-com/glew/issues/141) of the library names as soon as possible.

Meanwhile, as pointed out by @ilovezfs , we could think how we can sort out the [ong-standing pkg-config issue](https://github.com/Homebrew/legacy-homebrew/issues/22025#issuecomment-23732918) I just told you on the other issue.
Are there any suggestion on this side?

**Comment by nigels-com** [comment]
Concerning the pkg-config issue.  Linux seems to have a `glu.pc`, so why not Homebrew?
Because GLU is Apple-provided?

```
$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/glu.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: glu
Description: Mesa OpenGL Utility library
Requires: gl
Version: 9.0.0
Libs: -L${libdir} -lGLU
Libs.private: -lm
```

**Comment by nigels-com** [comment]
And to clarify with `cmake` - I'm glad to have it harmonised with `make` for homebrew goodness, but overall I do sometimes regret all the (needless?) maintenance that has come along with the cmake support.

**Comment by claudiofantacci** [comment]
`glu.pc` is shipped with X11 which is [optional on macOS](https://support.apple.com/en-us/HT201341).
```
Claudio@MBP-di-Claudio:~ $ locate glu.pc
/opt/X11/lib/pkgconfig/glu.pc


Claudio@MBP-di-Claudio:~ $ cat /opt/X11/lib/pkgconfig/glu.pc
prefix=/opt/X11
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: glu
Description: Mesa OpenGL Utility library
Requires: gl
Version: 9.0.0
Libs: -L${libdir} -lGLU
Libs.private: -lm
```

I'm trying to think of a possible solution.

**Comment by claudiofantacci** [comment]
>And to clarify with cmake - I'm glad to have it harmonised with make for homebrew goodness, but overall I do sometimes regret all the (needless?) maintenance that has come along with the cmake support.

I do understand this. Thanks for the effort spent so far! Overall, I think that supporting CMake will be of benefit for a lot of people/developers.

**Comment by ilovezfs** [comment]
Such a file could potentially be added [here](https://github.com/Homebrew/brew/tree/master/Library/Homebrew/os/mac/pkgconfig) but it would be unusual unless it were needed by at least three formulae. Also, it seems such a file is not actually needed on macOS given that we have 14 formulae in core using glew with that pkg-config requirement having been excised.

**Comment by nigels-com** [comment]
Including `glu.h` always works on Mac because it's always there.  Therefore no need for glew.pc on Mac.

However, `glu.h` is optional on Linux, so pkg-config needs to mention the requirement.

So, we need one .pc for Linux, a different one for Mac?

-- Nigel

**Comment by ilovezfs** [comment]
@nigels-com yeah, we're doctoring the existing one with `inreplace "glew.pc.in", "Requires: @requireslib@", ""` but a separate upstream .pc for macOS would be preferable to our downstream patching.

**Comment by claudiofantacci** [comment]
> So, we need one .pc for Linux, a different one for Mac?

I think this would be the most clean way to tackle this issue.
The `Requires: glu` should not be present on a macOS machine.

**Comment by claudiofantacci** [comment]
In the `makefile`, if `SYSTEM` is `darwin` we can cut out `Requires: glu`.

Not an expert of makefiles here though 😄

**Comment by nigels-com** [comment]
Well, we can omit `Requires: glu` unless we're in GLX/X11 mode?

**Comment by claudiofantacci** [comment]
Yes, we can check whether X11 is present in the system or not and then omit or include `Requires: glu`.

**Comment by ilovezfs** [comment]
Our CI has X11 installed so that will definitely confuse it, I think.

**Comment by claudiofantacci** [comment]
Confuse in the sense that only the X11 mode will be tested?

**Comment by ilovezfs** [comment]
Confuse in the sense that we will almost always want a build not to use X11 even though X11 happens to be installed on the CI boxes. The only cases where the X11 dependency would potentially be OK would be when a formula has explicitly declared `depends_on :x11`, which we almost never allow these days.

In other words, think of our CI as itself having X11 but nearly always cross compiling for non-X11-bearing systems.

**Comment by nigels-com** [comment]
It's not a problem from a GLEW Makefile build perspective.  X11/GLX mode is opt-in.  cmake currently doesn't support it.  So I don't think it will be a problem for Homebrew for GLEW Makefile to be precisely correct about it.

**Comment by ilovezfs** [comment]
:+1:

**Comment by claudiofantacci** [comment]
@ilovezfs ok, got it, thanks for the explanation.

@nigels-com seems to be perfect then! 👍 I will look into the CMakeFiles (hopefully) during the next week. I will make a PR on `glew` and modify this one accordingly. As far as the `makefiles` are concerned, I fear I'm not of great help 😟

**Comment by claudiofantacci** [comment]
@ilovezfs @nigels-com , I take the chance to having you both here to have your opinion about the change I made in the CMakeLists of glew.

By compiling with CMake the output of the compilation is:
``` bash
Claudio $ ls lib/
total 2784
-rwxr-xr-x  1 Claudio  staff   585K 22 Apr 19:52 libGLEW.2.0.0.dylib*
lrwxr-xr-x  1 Claudio  staff    19B 22 Apr 19:52 libGLEW.2.0.dylib@ -> libGLEW.2.0.0.dylib
-rw-r--r--  1 Claudio  staff   796K 22 Apr 19:52 libGLEW.a
lrwxr-xr-x  1 Claudio  staff    17B 22 Apr 19:52 libGLEW.dylib@ -> libGLEW.2.0.dylib
```
which still differs from the output of using `makefile` in that `libGLEW.dylib@ -> libGLEW.2.0.dylib` instead of symlinking to `libGLEW.2.0.0.dylib`, but by the user point of view should be the same.

I set `VERSION` (2.0.0) and `SOVERSION` (2.0) in `set_target_properties()` and cmake does the rest. Thus the result is how symlinks are concieved to be by cmake.

Let me know what you think about it 😄 !

nigels-com:
@claudiofantacci That sounds fine.  I'll follow up for Linux etc to see if the makefile build ought to match what cmake is doing.

claudiofantacci:
@nigels-com Ok! I'll push the CMake changes within today!

claudiofantacci:
Created [glew PR #143](https://github.com/nigels-com/glew/pull/143).

claudiofantacci:
@ilovezfs, [glew PR #143](https://github.com/nigels-com/glew/pull/143) was merged, I pushed new commits for applying a temporary patch (until new version of glew is released) and for taking into account your reviews.

ilovezfs:
@claudiofantacci you should be able to use a `patch do` block
```
  patch do
    url "https://github.com/nigels-com/glew/pull/143.patch"
    sha256 "7761252b504e869e66edfe1615b8eef40311b9eebd6268ae164d4c83af1b31e0"
  end
```

claudiofantacci:
@ilovezfs I have integrated the two CMake upstream patches to deal with the `glew.pc` `glu` requirement  and the wrong symlinks compiled library.

Everything should be fine now 😄 
Let me know if you require other changes 👍

fxcoudert:
Many thanks @claudiofantacci for your contribution to Homebrew! We look forward to other pull requests from you in the future :)

I'm sorry, I had trouble merging this one in due to changes to a spurious `.DS_Store` file… and so the final commit ended up loosing your authorship (although I mentioned it in the commit comment). I apologise for that.

claudiofantacci:
Hi @fxcoudert, it is unfortunate that the PR could not be merged smoothly and that the authorship was lost in the process 😭 . Anyway, I'm super happy that I could contribute to the `homebrew` project and I'm looking forward to do more contribution 👍.
Thanks for pointing out my name in your commit 😄, I appreciated that 😉!

PR #12736, opened 2017-04-21
emacs 25.2

dunn:

-----

### DIFF
--- a/Formula/emacs.rb
+++ b/Formula/emacs.rb
@@ -1,20 +1,15 @@
 class Emacs < Formula
   desc "GNU Emacs text editor"
   homepage "https://www.gnu.org/software/emacs/"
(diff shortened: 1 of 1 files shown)


dunn:
That's weird.  One of the mirrors didn't have the new version when I built it locally, but I'm seeing a 404 from all of them.

dunn:
Oh dear, looks like I just got lucky.  I think the only updated mirror so far is http://gnu.askapache.com/emacs/emacs-25.2.tar.xz

ilovezfs:
This happens all the time. @MikeMcQuaid since we use the GNU mirrors as the primary `url` for all GNU formulae, the `brew audit --online` test for mirrors, which is only supposed to run when `--strict` is passed, is "leaking" into the non-strict audits, and causing almost every GNU PR to fail CI when we open the PR before the mirrors have all synced, which is most of the time.

PR #12755, opened 2017-04-22
moodbar 0.1.2 (new formula)

ibrokemypie:

-----

### DIFF
--- a/Formula/moodbar.rb
+++ b/Formula/moodbar.rb
@@ -9,10 +9,10 @@ class Moodbar < Formula
   depends_on "pkg-config" => :build
   depends_on "automake" => :build
   depends_on "libtool" => :build
-  depends_on "gst-plugins-base@0.10" => [:recommended, "with-libogg", "with-libvorbis", "with-theora"]
-  depends_on "gst-plugins-good@0.10" => [:recommended, "with-flac"]
+  depends_on "gst-plugins-base@0.10" => :recommended
+  depends_on "gst-plugins-good@0.10" => :recommended
   depends_on "gst-plugins-bad@0.10" => :recommended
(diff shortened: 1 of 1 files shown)


formula/moodbar.rb
```diff
@@ -0,0 +1,38 @@
...
+                          "--disable-silent-rules",
+                          "--with-plugindir=#{lib}/gstreamer-0.10",
+                          "--with-gstreamer=#{prefix}",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    testpath.install resource("testfile")
+    system bin/"moodbar", "-o", "test.mood", "sample.wav"
```
bfontaine: You can use `test_fixtures("test.wav")` instead of `"sample.wav"` for a sample WAV file. It’ll return a path to this file: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/support/fixtures/test.wav

fxcoudert (approved):

fxcoudert:
```
moodbar:
  * Dependency gst-plugins-base@0.10 should not use option with-theora
  * Dependency gst-plugins-good@0.10 should not use option with-flac
  * Dependency gst-plugins-ugly@0.10 should not use option with-mad
```

Could we avoid these non-default options in plugins dependencies?

ibrokemypie:
How do you mean? Without those options we lose support for those filetypes

bfontaine:
The issue is since these options are non-default they force each of these dependencies to be built from source instead of poured from a bottle.

ibrokemypie:
What would be the best solution?

fxcoudert:
It may loose some file formats, but we need to build with default options of the dependencies.

fxcoudert:
Thanks @ibrokemypie for the pull request!

PR #12768, opened 2017-04-22
mpd: add option --with-upnp

alite:
Add option to compile mpd with support for upnp database plugin.
Add conditional requirements of 'expat' and 'libupnp'

-----

### DIFF
--- a/Formula/mpd.rb
+++ b/Formula/mpd.rb
@@ -26,6 +26,7 @@ class Mpd < Formula
   option "with-opus", "Build with opus support (for Opus encoding and decoding)"
   option "with-libmodplug", "Build with modplug support (for decoding modules supported by MODPlug)"
   option "with-pulseaudio", "Build with PulseAudio support (for sending audio output to a PulseAudio sound server)"
+  option "with-upnp", "Build with upnp database plugin support"
 
   deprecated_option "with-vorbis" => "with-libvorbis"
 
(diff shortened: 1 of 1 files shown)


Formula/mpd.rb
```diff
@@ -92,6 +97,11 @@ def install
     args << "--enable-nfs" if build.with? "libnfs"
     args << "--enable-modplug" if build.with? "libmodplug"
     args << "--enable-pulse" if build.with? "pulseaudio"
+    if build.with? "upnp"
+      args << "--enable-upnp"
+      args << "--enable-expat"
+      args << "--with-expat=/usr/local/opt/expat"
```
fxcoudert: Given that there is a `depends_on "expat"` above, the configure script should find it without help. Please try if it works that way.

If it does not, we'll keep the `--with-expat` but please use `--with-expat=#{Formula["openssl"].opt_prefix}` rather than hardcoding the path.

fxcoudert (changes requested):

Formula/mpd.rb
AliTe: Indeed, '--with-expat=' option is not neccessary, fixed it.
Thanks for noticing…

fxcoudert (approved):

fxcoudert:
Thanks @AliTe for your contribution to Homebrew!

PR #12771, opened 2017-04-22
audacious 3.8.2 (new formula)

ibrokemypie:

-----

### DIFF
--- a/Formula/audacious.rb
+++ b/Formula/audacious.rb
@@ -1,47 +1,57 @@
 class Audacious < Formula
   desc "Free and advanced audio player based on GTK+"
   homepage "http://audacious-media-player.org"
-  url "http://distfiles.audacious-media-player.org/audacious-3.8.2.tar.bz2"
-  sha256 "bdf1471cce9becf9599c742c03bdf67a2b26d9101f7d865f900a74d57addbe93"
-  head "https://github.com/audacious-media-player/audacious.git"
+  
+  stable do
+    url "http://distfiles.audacious-media-player.org/audacious-3.8.2.tar.bz2"
+    sha256 "bdf1471cce9becf9599c742c03bdf67a2b26d9101f7d865f900a74d57addbe93"
 
-  depends_on "make" => :build
-  depends_on "automake" => :build
-  depends_on "autoconf" => :build
+    resource "plugins" do
+      url "http://distfiles.audacious-media-player.org/audacious-plugins-3.8.2.tar.bz2"
+      sha256 "d7cefca7a0e32bf4e58bb6e84df157268b5e9a6771a0e8c2da98b03f92a5fdd4"
+    end
+  end
+  
+  head do
+    url "https://github.com/audacious-media-player/audacious.git"
+
+    resource "plugins" do
+      url "https://github.com/audacious-media-player/audacious-plugins.git"
+    end
+
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+    depends_on "libtool" => :build
+  end
+
   depends_on "gettext" => :build
-  depends_on "libtool" => :build
+  depends_on "make" => :build
   depends_on "pkg-config" => :build
+
   depends_on "neon"
-  depends_on "sdl2" => :recommended
-  depends_on "wavpack" => :recommended
   depends_on "glib"
-  depends_on "gtk+" => :optional
-  depends_on "ffmpeg" => :recommended
-  depends_on "libnotify" => :recommended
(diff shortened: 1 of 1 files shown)


Formula/audacious.rb
```diff
@@ -0,0 +1,56 @@
...
+    resource("audacious-plugins").stage do
+      ENV.prepend "PKG_CONFIG_PATH", "#{lib}/pkgconfig:"
+      system "./configure", "--prefix=#{prefix}"
+      system "make"
+      system "make", "install"
+    end
+  end
+
+  test do
+    system "false"
```
bfontaine: Please write a test.

Formula/audacious.rb
```diff
@@ -0,0 +1,56 @@
...
+    sha256 "d7cefca7a0e32bf4e58bb6e84df157268b5e9a6771a0e8c2da98b03f92a5fdd4"
+  end
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+
+    resource("audacious-plugins").stage do
+      ENV.prepend "PKG_CONFIG_PATH", "#{lib}/pkgconfig:"
```
bfontaine: Use `ENV.prepend_path "PKG_CONFIG_PATH", "#{lib}/pkgconfig"` instead.

bfontaine (changes requested):

ibrokemypie:
any ideas why head wont build? seems like the same environment variables arent being set. I opened a thread with more detail over here https://discourse.brew.sh/t/formula-fails-with-head-but-succeeds-with-the-same-content-without-head/718/4

Formula/audacious.rb
```diff
@@ -0,0 +1,96 @@
...
+  def caveats; <<-EOS.undent
+    audtool does not work due to a broken dbus implementation on macOS, so is not built
+    coreaudio output has been disabled as it does not work (Fails to set audio unit input property.)
+    GTK+ gui is not built by default as the QT gui has better integration with macOS, and when built, the gtk gui takes precedence
+    EOS
+  end
+
+  test do
+    system bin/"audacious", "-H", "-V", "-q", "-E", test_fixtures("test.wav"), "-p"
+    system "true"
```
bfontaine: You can remove this line now.

Formula/audacious.rb
```diff
@@ -0,0 +1,96 @@
+class Audacious < Formula
+  desc "Free and advanced audio player based on GTK+"
+  homepage "http://audacious-media-player.org"
+  url "http://distfiles.audacious-media-player.org/audacious-3.8.2.tar.bz2"
+  sha256 "bdf1471cce9becf9599c742c03bdf67a2b26d9101f7d865f900a74d57addbe93"
+  head "https://github.com/audacious-media-player/audacious.git"
+
+  option "with-app", "Build .app"
```
bfontaine: We don’t build `.app` files in Homebrew. Use Homebrew Cask for that.

bfontaine (changes requested):

Formula/audacious.rb
ibrokemypie: cask does not support building from source, which is the only options as there are no .apps provided from upstream

bfontaine: True. Still, please remove this option. Quoting from our ["Acceptable Formulae" guidelines](http://docs.brew.sh/Acceptable-Formulae.html):

> Don’t make your formula build an .app (native macOS Application); we don’t want those things in Homebrew (but Homebrew Cask does). Make it build a command-line tool or a library.

fxcoudert:
Description says *“based on GTK+”* but it's actually built with Qt?

ibrokemypie:
By default and originally it is a GTK app, however the QT build works better on macOS, which is why it builds with QT in this formula

bfontaine (approved):

fxcoudert:
Thanks @ibrokemypie for the pull request, this has now been committed!

ilovezfs:
The test doesn't work, and hangs until the test block itself times out after 5 minutes, and even then a process is left running permanently. (I just manually killed more than 12 of them on the CI nodes.)

It seems the test never worked including in the original CI (https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/1478/version=sierra/testReport/junit/brew-test-bot/sierra/test_audacious/).

As a stop-gap I'm at least temporarily switching this over to a simple test in https://github.com/Homebrew/homebrew-core/pull/13852 rather than removing the formula, but it would be good to get a non-trivial test working at some point.

PR #12810, opened 2017-04-24
gmime 3.0.0

tschoonj:


tschoonj:
@ilovezfs any thoughts on what's the best approach here to deal with `notmuch` and `mu` not being compatible with gmime 3.0.0? Shall I make `gmime@2.6` and link these two formula against while we wait for new releases?

ilovezfs:
@tschoonj I'd either wait for the new versions, or vendor gmime 2.6.23 into those two formulae, since a `gmime@2.6` is unlikely to be needed by users and the install method is simple.

fxcoudert:
Don't `notmuch` and `mu` need a revision bump?

tschoonj:
Sorry, I have been a bit sloppy lately... Should be fixed now

fxcoudert:
@BrewTestBot test this please

PR #12835, opened 2017-04-24
openshift-cli 1.5.0

gilesdring:

-----

### DIFF
--- a/Formula/openshift-cli.rb
+++ b/Formula/openshift-cli.rb
@@ -39,6 +39,9 @@ class OpenshiftCli < Formula
 
   test do
     assert_match /^oc v#{version}/, shell_output("#{bin}/oc version")
-    assert_match /^oadm v#{version}/, shell_output("#{bin}/oadm version")
-  end
+    if version >= '3.6.0-alpha.0'
+      assert_match /^oc v#{version}/, shell_output("#{bin}/oadm version")
+    else
+      assert_match /^oadm v#{version}/, shell_output("#{bin}/oadm version")
+    end
 end
--- a/Formula/openshift-cli.rb
+++ b/Formula/openshift-cli.rb
@@ -44,4 +44,5 @@ class OpenshiftCli < Formula
     else
       assert_match /^oadm v#{version}/, shell_output("#{bin}/oadm version")
     end
+  end
 end
--- a/Formula/openshift-cli.rb
+++ b/Formula/openshift-cli.rb
@@ -39,7 +39,7 @@ class OpenshiftCli < Formula
 
   test do
     assert_match /^oc v#{version}/, shell_output("#{bin}/oc version")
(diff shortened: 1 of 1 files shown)


gilesdring:
I noticed the tests failed on the devel version of the code. It looks like openshift client has changed version reporting for the `oadm` command - it now reports as `oc`, which is causing the failure. This may be an alpha issue. Any way of defining a different test for devel versions? I could weaken the regex, but don't want do do this without discussion.

JCount:
@gilesdring You can use `if build.devel?` and/or `if build.stable?` in the test block to branch the code depending upon the current spec, which should allow you to do what you want.

For a basic example, see https://github.com/Homebrew/homebrew-core/blob/master/Formula/libvirt.rb#L74-L82

elyscape:
@gilesdring As @JCount says, you could change the test to something like this:
```ruby
test do
  assert_match /^oc v#{version}/, shell_output("#{bin}/oc version")
  if build.devel?
    assert_match /^oc v#{version}/, shell_output("#{bin}/oadm version")
  else
    assert_match /^oadm v#{version}/, shell_output("#{bin}/oadm version")
  end
end
```
Or you could filter on version like this:
```ruby
test do
  assert_match /^oc v#{version}/, shell_output("#{bin}/oc version")
  if version >= '3.6.0-alpha.0'
    assert_match /^oc v#{version}/, shell_output("#{bin}/oadm version")
  else
    assert_match /^oadm v#{version}/, shell_output("#{bin}/oadm version")
  end
end
```
That option has the benefit of being more obvious that it should be removed when 3.6.0 hits release.

gilesdring:
Thanks - have updated. Let's see if that works. Do you want me to squash and resubmit the PR?

JCount:
@gilesdring Don't bother, I don't mind squashing the commits for you when I merge this. The goal general goal is to have 1 commit if possible, but that is not always the case. Also, in this instance, it is probably better to preserve the state of the PR.

JCount:
@gilesdring Don't bother, I don't mind squashing them when I merge this. The general aim is to have one commit per PR if possible, but that is not always feasible.

JCount:
@gilesdring Does the version based differentiation in the test still work if the HEAD spec is built? Or does it execute the test in the 'else' branch and invariably fail?

elyscape:
@JCount [As per `version.rb`][version], `HEAD` is assumed to be newer than everything that isn't `HEAD`.

[version]: https://github.com/Homebrew/brew/blob/1.1.13/Library/Homebrew/version.rb#L225-L227

JCount:
Right, haha. So many things to remember. 😆

JCount:
@gilesdring  Thank you for your for your contribution to Homebrew! 😄

gilesdring:
No problem - sorry about the snafus along the way! ;-)

PR #12844, opened 2017-04-24
add instructions for auto-activating pyenv in shell

tejasmanohar:
Without this, the user has to `eval "$(pyenv init -)"` before
every pyenv command, which is rarely desirable. The pyenv official
instructions recommend adding this to your shell's profile.

### DIFF
--- a/Formula/pyenv.rb
+++ b/Formula/pyenv.rb
@@ -21,6 +21,12 @@ class Pyenv < Formula
     end
   end
 
+  def caveats; <<-EOS.undent
+    To enable auto-activation, add the following to your profile:
+      eval "$(pyenv init -)"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Our caveats are only for things that are Homebrew-specific so passing on this, sorry.

tejasmanohar:
@MikeMcQuaid I see. I created this due to
https://github.com/segmentio/homebrew-core/blob/02f61b9a9273ce4b8152da4e8a62da207398f28c/Formula/pyenv-virtualenv.rb#L20

MikeMcQuaid:
@tejasmanohar Yeh, it's still present in some formulae and should be removed from them all.

PR #12847, opened 2017-04-25
proj: fix allow cmake helpers install

leolchat:
Not fully there is a warning from before, about the description having the name of the formula in it.
-----

When using the configure, make route, proj4 doesn't install the cmake helper scripts, while using cmake to build it does.

I chose to set it as an option since it adds a dependency on cmake, but it might be that most people are willing to have the cmake helpers installed (a quick google for FindProj shows that many are reimplementing it because it is not installed by default).

### DIFF
--- a/Formula/proj.rb
+++ b/Formula/proj.rb
@@ -67,6 +67,12 @@ class Proj < Formula
 
   skip_clean :la
 
+  option "with-cmake", "Install the CMake library helpers to allow find_package(PROJ4)"
+
+  if build.with? "cmake"
+    depends_on "cmake" => :build
+  end
+
   def install
     resources.each do |r|
       if r.name == "datumgrid"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I reckon because CMake is a build-time-only dependency that many other things depend on it's worth making this on unconditionally.

leolchat:
You mean no option just switch the build to using cmake ? I can certainly do that.

ilovezfs (changes requested):
Switching to cmake seems to change the library number, which is a bug that needs to be reported upstream and fixed before we can merge this, unless we revision bump all the reverse dependencies to accommodate the different library number.

fxcoudert:
Upstream's bug report is there: https://github.com/OSGeo/proj.4/issues/378
Nothing moving so far, so let's close this PR for now and wait until upstream gets it fixed!

PR #12850, opened 2017-04-25
gmime - make gobject-introspection dependency optional; add test

dcreemer:
An earlier commit: https://github.com/Homebrew/homebrew-core/commit/5fb70c657bd39a007529510efaec7714508040aa introduced a hard dependency to the gobject-introspection lib. That lib in turn pulls in cairo, python, and all their dependencies -- which seems unwise. Other tools, such as mu, require gmime, but do not need all of the gobject-introspection capability.

This PR makes the dependency merely recommended, rather than required. This should have no other impact, but allows for people like me to build gmime without the many dependencies.
In addition, the PR adds a test for a working gmime library.

There is an existing PR #12810 which updates gmime to version 3.0. I think it's probably best to merge this first, then upgrade (that way you get the test), but either order would probably work.

### DIFF
--- a/Formula/gmime.rb
+++ b/Formula/gmime.rb
@@ -12,17 +12,44 @@ class Gmime < Formula
 
   depends_on "pkg-config" => :build
   depends_on "libgpg-error" => :build
+  depends_on "gobject-introspection" => :recommended
   depends_on "glib"
-  depends_on "gobject-introspection"
 
   def install
-    system "./configure", "--disable-dependency-tracking",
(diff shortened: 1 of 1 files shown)


tschoonj:
I do not object to this PR being merged in before my #12810. I have only one comment: the commit title should start with `gmime` to clarify which formula it deals with.

dcreemer:
thanks @tschoonj . Change commit title

fxcoudert:
Thanks @dcreemer for the pull request!

PR #12861, opened 2017-04-25
profanity: add option --with-python-plugin-support

alite:
Add option to compile profanity with support for python plugins.
Set dependency to 'python' to always use homebrew version of python.

-----

### DIFF
--- a/Formula/profanity.rb
+++ b/Formula/profanity.rb
@@ -19,6 +19,8 @@ class Profanity < Formula
     depends_on "libtool" => :build
   end
 
+  option "with-python-plugin-support", "Build with support for python plugins"
+
   depends_on "pkg-config" => :build
   depends_on "ossp-uuid"
   depends_on "libstrophe"
@@ -29,13 +31,19 @@ class Profanity < Formula
   depends_on "libotr" => :recommended
   depends_on "gpgme" => :recommended
   depends_on "terminal-notifier" => :optional
+  # this always uses homebrewed python version
+  depends_on "python" if build.with? "python-plugin-support"
 
   def install
+    args = %W[
+      --disable-dependency-tracking
+      --disable-silent-rules
+      --prefix=#{prefix}
+    ]
+    args << "--disable-python-plugins" if build.without? "python-plugin-support"
+
     system "./bootstrap.sh" if build.head?
-    system "./configure", "--disable-dependency-tracking",
-                          "--disable-silent-rules",
-                          "--disable-python-plugins",
-                          "--prefix=#{prefix}"
+    system "./configure", *args
     system "make", "install"
   end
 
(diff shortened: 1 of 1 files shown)


Formula/profanity.rb
```diff
@@ -29,13 +31,19 @@ class Profanity < Formula
   depends_on "libotr" => :recommended
   depends_on "gpgme" => :recommended
   depends_on "terminal-notifier" => :optional
+  # this always uses homebrewed python version
```
DomT4: Why?

AliTe: OK, I just learned that I can use `depends_on :python ...` which automatically selects the python version.
That solves my problem, I think.

Formula/profanity.rb
```diff
@@ -29,13 +31,18 @@ class Profanity < Formula
   depends_on "libotr" => :recommended
   depends_on "gpgme" => :recommended
   depends_on "terminal-notifier" => :optional
+  depends_on :python if build.with? "python-plugin-support"
```
ilovezfs: Are the plugins unable to build against system Python?

AliTe: It works both, against system and brewed python.
This way it will compile against brewed version if installed, against system version otherwise.
I'm not shure what should be the recommended behaviour …

Formula/profanity.rb
```diff
@@ -19,6 +19,8 @@ class Profanity < Formula
     depends_on "libtool" => :build
   end
 
+  option "with-python-plugin-support", "Build with support for python plugins"
```
dunn: How about just `depends_on :python => :optional`, which will create option text automatically and is less buggy than `depends_on ... if`.

AliTe: Sounds good, but `:python` is only required, if profanity is build with option `--with-pyton-plugin-support` - there's no need to choose python otherwise. Couldn't that be confusing.

But if you think the 'optional' solution is best, I will patch it this way.

Formula/profanity.rb
dunn: If it works with system Python we shouldn't even need this dependency; `:python` is used to require 2.7 or higher.  Right @ilovezfs ?

ilovezfs: Right.
```
```

Formula/profanity.rb
ilovezfs: Why would someone want a build without the python plugin support?

Formula/profanity.rb
AliTe: Sorry, I'm getting a little confused, so let me summarize …

I found the (original) homebrew formula of profanity comes with `--disable-python-plugins` configure option by default and so comes the bottled version.

Compiling profanity against python is needed if - and only if - python plugin support is desired.

Using `depends_on :python if build.with? "python-plugin-support` (i.e. the symbol `:python`) homebrew will automatically select either the system python (if no homebrew python is installed) or the homebrewed python if it is installed - what seems coherent to me (since I didn't want to change the default behaviour of the original formula).

It will compile against both (system and homebrew python) - at least on macOS sierra (couldn't test it on older systems …)

Formula/profanity.rb
AliTe: This was the default of the original formula.

